PR #35, opened 2015-06-09
add a bridge-router-bridge test with router implemented as a namespace

yonghong-song:
This is to implement brb with bcc.

Brenden, ethernet->dst (probably other mac address extraction as well) does not work.
Please investigate. At this point, I use load_half/load_word as the workaround.

### DIFF
--- a/tests/cc/test_brb.py
+++ b/tests/cc/test_brb.py
@@ -1,0 +1,281 @@
+#!/usr/bin/env python
+# Copyright (c) PLUMgrid, Inc.
+# Licensed under the Apache License, Version 2.0 (the "License")
+
+# This program implements a topology likes below:
+#   pem: physical endpoint manager, implemented as a bpf program
+#
+#     vm1 <--------+  +----> bridge1 <----+
+#                  V  V                   V
+#                  pem                  router
+#                  ^  ^                   ^
+#     vm2 <--------+  +----> bridge2 <----+
+#
+# The vm1, vm2 and router are implemented as namespaces.
+# The bridge is implemented with limited functionality in bpf program.
+# 
+# vm1 and vm2 are in different subnet. For vm1 to communicate to vm2,
+# the packet will have to travel from vm1 to pem, bridge1, router, bridge2, pem, and
+# then come to vm2.
+#
+# When this test is run with verbose mode (ctest -R <test_name> -V),
+# the following printout is observed on my local box:
+#
+# ......
+# 8: ARPING 100.1.1.254 from 100.1.1.1 eth0
+# 8: Unicast reply from 100.1.1.254 [E6:5F:05:95:4B:41]  0.532ms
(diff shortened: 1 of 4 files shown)


tests/cc/test_brb.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def setup_router_ns(self, ns, veth1_in, veth1_out, veth2_in, veth2_out):
+        subprocess.call(["ip", "netns", "add", ns])
+        subprocess.call(["ip", "link", "add", veth1_in, "type", "veth", "peer", "name", veth1_out])
+        subprocess.call(["ip", "link", "set", veth1_in, "netns", ns])
+        subprocess.call(["ip", "netns", "exec", ns, "ip", "link", "set", veth1_in, "name", "eth0"])
+        subprocess.call(["ip", "link", "add", veth2_in, "type", "veth", "peer", "name", veth2_out])
+        subprocess.call(["ip", "link", "set", veth2_in, "netns", ns])
+        subprocess.call(["ip", "netns", "exec", ns, "ip", "link", "set", veth2_in, "name", "eth1"])
+        subprocess.call(["ifconfig", veth1_out, "up"])
```
4ast: pls use 'ip link set dev up' instead of ifconfig

tests/cc/test_brb.py
```diff
@@ -0,0 +1,232 @@
...
+
+        # get vm mac address
+        self.vm1_mac = commands.getoutput('ip netns exec ' + self.ns1 + ' cat /sys/class/net/eth0/address')
+        self.vm2_mac = commands.getoutput('ip netns exec ' + self.ns2 + ' cat /sys/class/net/eth0/address')
+
+        # load the program and configure maps
+        self.config_maps()
+
+    def test_brb(self):
+        # our bridge is not smart enough, so send arping for router learning to prevent router
```
4ast: would be good to do an example with real linux bridge

tests/cc/test_brb.py
```diff
@@ -0,0 +1,232 @@
...
+        self.config_router_ns(self.ns_router, self.vm1_rtr_ip, self.vm2_rtr_ip)
+
+        # get vm mac address
+        self.vm1_mac = commands.getoutput('ip netns exec ' + self.ns1 + ' cat /sys/class/net/eth0/address')
+        self.vm2_mac = commands.getoutput('ip netns exec ' + self.ns2 + ' cat /sys/class/net/eth0/address')
+
+        # load the program and configure maps
+        self.config_maps()
+
+    def test_brb(self):
```
4ast: may be include example 'make test' output in commit log and/or this .py file, so users know what to expect?

tests/cc/test_brb.py
yonghong-song: Yes, this will be my next step.

PR #44, opened 2015-06-10
use ethernet->src/dst to get mac address instead of primitive load_* …

yonghong-song:
…functions

### DIFF
--- a/tests/cc/test_brb.c
+++ b/tests/cc/test_brb.c
@@ -2,9 +2,7 @@
 // Licensed under the Apache License, Version 2.0 (the "License")
 #include <bcc/proto.h>
 
-/* compiler workaround */
-#define _htonl __builtin_bswap32
-#define _htons __builtin_bswap16
+#define _memcpy __builtin_memcpy
 
 // meta data passed between bpf programs
 typedef struct bpf_metadata {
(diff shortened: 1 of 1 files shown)


tests/cc/test_brb.c
```diff
@@ -129,9 +128,8 @@ static int br_common(struct __sk_buff *skb, int which_br) {
 
     BEGIN(ethernet);
     PROTO(ethernet) {
-        // ethernet->dst seems not working, so tentatively use the primitive C API.
-        *(__u32 *)&dmac.addr[0] = _htonl(load_word(skb, 0));
-        *(__u16 *)&dmac.addr[4] = _htons(load_half(skb, 4));
+        mac_addr = bpf_htonll(ethernet->dst);
```
4ast: why swap is needed?

drzaeus77:
Please remove uses of htonll, they are not required. The endian conversion should (by design) be taken care of by the library/language without impacting the user. For instance, if you assign a number in host format (as from a table lookup result) to the ethernet->dst/src field, the bpf_dins_pkt helper function and some bits of the BFrontendAction class in the compiler will convert this for you.

drzaeus77:
I was too slow! Anyway, please followup the changes with what was mentioned.

PR #46, opened 2015-06-10
add another bridge-router-bridge test (test_brb2)

yonghong-song:
o router is implemented as a namespace, similar to test_brb.
  o using linux bridge to provide bridge functionality.
    In test_brb, a simple bpf program implemented limited
    bridge functionality.
  o on my local box, test_brb performance is roughly 12% better
    than test_brb2.

### DIFF
--- a/tests/cc/test_brb2.c
+++ b/tests/cc/test_brb2.c
@@ -1,0 +1,29 @@
+// Copyright (c) PLUMgrid, Inc.
+// Licensed under the Apache License, Version 2.0 (the "License")
+#include <bcc/proto.h>
+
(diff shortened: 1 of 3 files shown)


tests/cc/test_brb2.c
```diff
@@ -0,0 +1,29 @@
...
+        /* accumulate stats */
+        u32 index = 0;
+        u32 *value = pem_stats.lookup(&index);
+        if (value)
+            lock_xadd(value, 1);
+#endif
+        bpf_clone_redirect(skb, *ifindex_p, 0);
+    }
+
+    return 0;
```
4ast: it passes packets to the stack. more accurate would be to drop them.

PR #47, opened 2015-06-11
Add a Learning VLAN program example

drzaeus77:
* Create a simulation of a physical network with multiple vlans, which
  can accept and distribute connections from multiple clients in
  overlapping subnets. The program will assign to a worker based on mac
  address.
* Remove a hardcoded SO_ATTACH_BPF literal, and fix one spacing

### DIFF
--- a/examples/vlan_learning.c
+++ b/examples/vlan_learning.c
@@ -1,0 +1,54 @@
+// Copyright (c) PLUMgrid, Inc.
+// Licensed under the Apache License, Version 2.0 (the "License")
+
+#include <bcc/proto.h>
+
+struct ifindex_leaf_t {
+  int out_ifindex;
+  u64 tx_pkts;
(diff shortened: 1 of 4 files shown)


examples/vlan_learning.c
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) PLUMgrid, Inc.
+// Licensed under the Apache License, Version 2.0 (the "License")
+
+#include <bcc/proto.h>
+
+struct ifindex_leaf_t {
+  int out_ifindex;
```
4ast: there is a 4 byte gap in the struct. I guess llvm decided that it's cheaper to initialize the whole thing including gaps. Better to avoid.

drzaeus77: Not sure I understand. Are you suggesting to pack the structure, or put some pad fields in there, or something else?

PR #48, opened 2015-06-11
avoid include file linux/bpf.h in helpers.h

yonghong-song:
o use uapi/linux/bpf.h instead
   o to be consistent with kernel/samples/bpf/

### DIFF
--- a/src/cc/export/helpers.h
+++ b/src/cc/export/helpers.h
@@ -16,8 +16,7 @@
 #ifndef __BPF_HELPERS_H
 #define __BPF_HELPERS_H
 
-#include <linux/bpf.h>
-#include <linux/filter.h>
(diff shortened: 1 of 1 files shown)


drzaeus77:
Nice, this speeds up compilation considerably! Later, I will see about enabling PCH in clang to speed it up even further.

src/cc/export/helpers.h
```diff
@@ -16,8 +16,7 @@
 #ifndef __BPF_HELPERS_H
 #define __BPF_HELPERS_H
 
-#include <linux/bpf.h>
-#include <linux/filter.h>
+#include <uapi/linux/bpf.h>
 #include <linux/if_packet.h>
```
4ast: this should also say 'uapi', since that's where the header is coming from.

PR #62, opened 2015-06-16
Compatibility fixes and pyroute2 examples

drzaeus77:


examples/simple_ipdb.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python
```
4ast: only one example is needed, right? two that do the same stuff is confusing, since not clear which one is recommended to use. Could you delete one that not recommended?

drzaeus77: Sure, both actually work fine, just the IPDB one has more features, which are shown in the vlan_learning example. For the simple case, using ipr is preferred and less likely to lock up.

drzaeus77:
Redundant example removed, please review.

Issue #59, opened 2015-06-13
Create a TC example that shows how to do complex classifications more simply


drzaeus77:
For example, try implementing Figure 12 from http://people.netfilter.org/pablo/netdev0.1/papers/Linux-Traffic-Control-Classifier-Action-Subsystem-Architecture.pdf

PR #63, opened 2015-06-16
Add example for BPF+tc rate limiting

drzaeus77:
* Adding an example to address #59, simulation of bandwidth sharing

### DIFF
--- a/examples/tc_neighbor_sharing.c
+++ b/examples/tc_neighbor_sharing.c
@@ -1,0 +1,61 @@
+// Copyright (c) PLUMgrid, Inc.
+// Licensed under the Apache License, Version 2.0 (the "License")
+
+#include <bcc/proto.h>
+
+struct ipkey {
+  u32 client_ip;
+};
+
+BPF_TABLE("hash", struct ipkey, int, learned_ips, 1024);
+
+// trivial action
+int pass(struct __sk_buff *skb) {
+  return 1;
+}
+
+// Process each wan packet, and determine if the packet is in the IP
+// table or not. Learned IPs are rate-limited and unclassified are not.
+// returns: > 0 when an IP is known
+//          = 0 when an IP is not known, or non-IP traffic
+int classify_wan(struct __sk_buff *skb) {
+  BEGIN(ethernet);
+  PROTO(ethernet) {
+    switch (ethernet->type) {
+      case 0x0800: goto ip;
+    }
+    goto EOP;
+  }
+  PROTO(ip) {
+    u32 dip = ip->dst;
+    struct ipkey key = {.client_ip=dip};
+    int *val = learned_ips.lookup(&key);
+    if (val)
+      return *val;
+    goto EOP;
+  }
+EOP:
+  return 0;
+}
+
+// Process each neighbor packet, and store the source IP in the learned table.
+// Mark the inserted entry with a non-zero value to be used by the classify_wan
+// lookup.
+int classify_neighbor(struct __sk_buff *skb) {
+  BEGIN(ethernet);
+  PROTO(ethernet) {
+    switch (ethernet->type) {
+      case 0x0800: goto ip;
+    }
+    goto EOP;
+  }
+  PROTO(ip) {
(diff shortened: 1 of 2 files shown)


examples/tc_neighbor_sharing.c
```diff
@@ -0,0 +1,61 @@
...
+    switch (ethernet->type) {
+      case 0x0800: goto ip;
+    }
+    goto EOP;
+  }
+  PROTO(ip) {
+    u32 sip = ip->src;
+    struct ipkey key = {.client_ip=sip};
+    int val = 1;
+    learned_ips.lookup_or_init(&key, &val);
```
4ast: update(,NOEXIST) would be cleaner?

examples/tc_neighbor_sharing.py
```diff
@@ -0,0 +1,122 @@
...
+               prio=2, name=pass_fn.name, parent="ffff:", action="drop",
+               classid=2, rate="1024kbit", burst=1024 * 32, mtu=16 * 1024)
+        self.wan_if = wan_if
+
+    # helper function to create a namespace and a veth connecting it
+    def _create_ns(self, name, ipaddr, fn):
+        ns_ipdb = IPDB(nl=NetNS(name))
+        ipdb.create(ifname="%sa" % name, kind="veth", peer="%sb" % name).commit()
+        with ipdb.interfaces["%sb" % name] as v:
+            v.net_ns_fd = ns_ipdb.nl.netns
```
4ast: this assignment is moving interface into netns? would be good to add a comment

examples/tc_neighbor_sharing.py
```diff
@@ -0,0 +1,122 @@
...
+        ipr.tc("add-filter", "bpf", wan_if["index"], ":1", fd=wan_fn.fd,
+               prio=1, name=wan_fn.name, parent="ffff:", action="drop",
+               classid=1, rate="128kbit", burst=1024 * 32, mtu=16 * 1024)
+        ipr.tc("add-filter", "bpf", wan_if["index"], ":2", fd=pass_fn.fd,
+               prio=2, name=pass_fn.name, parent="ffff:", action="drop",
+               classid=2, rate="1024kbit", burst=1024 * 32, mtu=16 * 1024)
+        self.wan_if = wan_if
+
+    # helper function to create a namespace and a veth connecting it
+    def _create_ns(self, name, ipaddr, fn):
```
4ast: probably can be shared with vlan example?

examples/tc_neighbor_sharing.py
```diff
@@ -0,0 +1,122 @@
...
+try:
+    sim = SharedNetSimulation()
+    sim.start()
+    print("Network ready. Create a shell in the wan0 namespace and test with iperf")
+    print(" e.g.: ip netns exec wan0 iperf -t 2 -c 172.16.1.100")
+    input("Press enter when finished: ")
+finally:
+    if "sim" in locals(): sim.release()
+    if "br100" in ipdb.interfaces: ipdb.interfaces.br100.remove().commit()
+    sleep(2)
```
4ast: what sleep is for?

svinota: @drzaeus77 — in the case when IPDB doesn't work well w/o `sleep()` syncs, pls report the bug, since the goal of `commit()` calls is to be reliably synchronous and to guarantee the system state when the `commit()` call is finished.

drzaeus77: Will do! Let me see if there is a less complicated test script that captures the scenario that I see.

On another note, I am also working on a related set of changes to pyroute2 which this test depends on. It will add policer support for the bpf filter types.

examples/tc_neighbor_sharing.py
drzaeus77: Not that easily, unfortunately, the vlan example is doing some things with mac addresses that makes it less portable.

drzaeus77:
Addressed some comments. Will attempt to address the sleep(2) after some investigation in a follow-up change.

PR #65, opened 2015-06-17
add get_current_pid/tgid/uid/gid/comm helpers

4ast:
Signed-off-by: Alexei Starovoitov <ast@plumgrid.com>


drzaeus77:
Would be good to add a test later.

PR #72, opened 2015-06-23
use "raw_input" instead "input" to avoid python 2.7 exception

yonghong-song:


drzaeus77:
Please use the approach in vlan_learning.py instead.

`from builtins import input`

This is a backwards/forwards compatible approach to avoid the same issue.

yonghong-song:
Just updated the patch with the suggestion.

drzaeus77:
Thanks!

PR #78, opened 2015-07-02
Drop PROTO macro syntax and introduce alternative

drzaeus77:
* Feedback given by some external users was that the PROTO syntax was
  opaque and unintuitive. As such, drop that macro and introduce an
  alternative syntax.
* Convert all to use 'cursor_advance' macro API
* Add kernel in_ether.h header to proto.h, and convert many literals to use the macro equivalent.
  Considered adding TCP/UDP definitions as well, but the header including that has more dependencies
  that I would rather not force on the user.

### DIFF
--- a/examples/tunnel_monitor/monitor.c
+++ b/examples/tunnel_monitor/monitor.c
@@ -57,9 +57,9 @@ int handle_egress(struct __sk_buff *skb) {
 
 // parse the outer vxlan frame
 int handle_outer(struct __sk_buff *skb) {
-  u8 *skb_cursor = 0;
+  u8 *cursor = 0;
 
-  struct ethernet_t *ethernet = bpf_consume_skb(skb_cursor, sizeof(*ethernet));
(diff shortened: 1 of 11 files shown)


examples/tunnel_monitor/monitor.c
```diff
@@ -57,37 +57,43 @@ int handle_egress(struct __sk_buff *skb) {
...
-  PROTO(ip) {
-    skb->cb[CB_SIP] = ip->src;
-    skb->cb[CB_DIP] = ip->dst;
-    switch (ip->nextp) {
-      case 17: goto udp;
-    }
-    goto EOP;
+  u8 *skb_cursor = 0;
+
+  struct ethernet_t *ethernet = bpf_consume_skb(skb_cursor, sizeof(*ethernet));
```
4ast: left over of previous experiments?

drzaeus77: Crap, didn't test after manual search/replace

drzaeus77:
Please check updated version.

PR #79, opened 2015-07-06
fix a test error for brb/brb2

yonghong-song:


drzaeus77:
As per another conversation, please try using pyroute2 as well as some of the automatic cleanup functions in simulation.py.

PR #80, opened 2015-07-07
Add multiple build support styles

drzaeus77:
* Add RPM and DEB packaging targets (using CPack from CMake) to build
  binary packages for Fedora and Ubuntu targets.
* Add Docker build scripts for each of the above that run the build in
  the right environment (assuming docker is available).
 - In Ubuntu, build against the LLVM 3.7 nightly snapshots
 - In Fedora, build against LLVM 3.7 from git (takes longer)
* Depending on packages installed on the build machine, it may be
  possible to cross-package for other targets without invoking Docker.
* Re-introduce src/cc/compat directory to keep the build stable for the
  time being. Similarly, it was necessary to #define some ugly constants
  that should eventually show up in libc.
* Add a few simple version checks to allow a partially working (really
  tracing only) libbcc in 4.1 kernels.

TODO (followup commit): Re-work the READMEs


Dockerfile.fedora
```diff
@@ -0,0 +1,21 @@
...
+MAINTAINER Brenden Blanco <bblanco@plumgrid.com>
+
+RUN dnf -y install @rpm-development-tools @c-development @development-tools cmake libstdc++-static
+
+WORKDIR /root
+RUN git clone https://github.com/llvm-mirror/llvm.git
+RUN git clone https://github.com/llvm-mirror/clang.git llvm/tools/clang
+RUN mkdir -p /root/llvm/build
+WORKDIR /root/llvm/build
+RUN cmake .. -DBUILD_SHARED_LIBS=OFF -DCMAKE_BUILD_TYPE=Release -DLLVM_TARGETS_TO_BUILD="ARM;CppBackend;X86;BPF"
```
4ast: do we need to build arm,cpp,x86 backends?

drzaeus77: It's been a while since I tried optimizing this. I will give it a shot to see.

drzaeus77: With no X86, I see the following:

Native target X86 is not selected; lli will not JIT code

BPF is not shown as enabled in this case.

PR #85, opened 2015-07-14
Change to simulator arguments for readability

drzaeus77:
* Let one argument be optional
* Disable ipv6 for whole namespace once
* Python 2/3 compatibility fix

### DIFF
--- a/examples/simulation.py
+++ b/examples/simulation.py
@@ -26,7 +26,8 @@ class Simulation(object):
         else:
             ns_ipdb = IPDB(nl=NetNS(name))
             if disable_ipv6:
-                cmd = ["sysctl", "-q", "-w", "net.ipv6.conf.default.disable_ipv6=1"]
+                cmd = ["sysctl", "-q", "-w",
+                       "net.ipv6.conf.default.disable_ipv6=1"]
                 nsp = NSPopen(ns_ipdb.nl.netns, cmd)
                 nsp.wait(); nsp.release()
             ns_ipdb.interfaces.lo.up().commit()
@@ -48,6 +49,10 @@ class Simulation(object):
             if ipaddr: v.add_ip("%s" % ipaddr)
             if macaddr: v.address = macaddr
             v.up()
+        if disable_ipv6:
+            cmd = ["sysctl", "-q", "-w",
+                   "net.ipv6.conf.%s.disable_ipv6=1" % out_ifc.ifname]
+            subprocess.call(cmd)
         if fn and out_ifc:
             self.ipdb.nl.tc("add", "ingress", out_ifc["index"], "ffff:")
             self.ipdb.nl.tc("add-filter", "bpf", out_ifc["index"], ":1",
--- a/examples/simulation.py
+++ b/examples/simulation.py
@@ -18,13 +18,18 @@ class Simulation(object):
     # if called directly outside Simulation class, "ifc_base_name" should be
     # different from "name", the "ifc_base_name" and "name" are the same for
     # the first ifc created by namespace
-    def _ns_add_ifc(self, name, ns_ifc, ifc_base_name, in_ifc=None, out_ifc=None,
-                    ipaddr=None, macaddr=None, fn=None, cmd=None, action="ok",
-                    disable_ipv6=False):
+    def _ns_add_ifc(self, name, ns_ifc, ifc_base_name=None, in_ifc=None,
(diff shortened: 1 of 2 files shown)


examples/simulation.py
```diff
@@ -18,13 +18,18 @@ def __init__(self, ipdb):
...
+                    action="ok", disable_ipv6=False):
         if name in self.ipdbs:
             ns_ipdb = self.ipdbs[name]
         else:
             ns_ipdb = IPDB(nl=NetNS(name))
+            if disable_ipv6:
+                cmd = ["sysctl", "-q", "-w", "net.ipv6.conf.default.disable_ipv6=1"]
+                nsp = NSPopen(ns_ipdb.nl.netns, cmd)
+                nsp.wait(); nsp.release()
+            ns_ipdb.interfaces.lo.up().commit()
```
yonghong-song: To disable ipv6, do we need to disable both veth end ifc, or just the ifc inside netns?

drzaeus77: Don't all of the ifcs end up inside a netns?

tests/cc/test_brb.py
```diff
@@ -109,7 +109,7 @@ def connect_ports(self, prog_id_pem, prog_id_br, curr_pem_pid, curr_br_pid,
         br_dest_map[c_uint(curr_br_pid)] = br_dest_map.Leaf(prog_id_pem, curr_pem_pid)
         self.pem_port[c_uint(curr_pem_pid)] = c_uint(ifindex)
         self.pem_ifindex[c_uint(ifindex)] = c_uint(curr_pem_pid)
-        mac_addr = br_mac_map.Key(int(EUI(vm_mac.decode())))
+        mac_addr = br_mac_map.Key(int(EUI(vm_mac)))
```
yonghong-song: int(EUI(vm_mac)) won't work on python 2.7.3.
8:   File "/home/plumgrid/iovisor/bcc/tests/cc/test_brb.py", line 181, in test_brb
8:     self.config_maps()
8:   File "/home/plumgrid/iovisor/bcc/tests/cc/test_brb.py", line 136, in config_maps
8:     self.ns1_eth_out.index, self.vm1_mac, self.vm1_ip)
8:   File "/home/plumgrid/iovisor/bcc/tests/cc/test_brb.py", line 112, in connect_ports
8:     mac_addr = br_mac_map.Key(int(EUI(vm_mac())))
8: TypeError: 'str' object is not callable

This is yet another python 2/3 compatibility issue.

examples/simulation.py
yonghong-song: out_ifc is not in the namespace.

tests/cc/test_brb.py
drzaeus77: No, it looks like you have a typo. Did you type or apply the full patch? It should not be "vm_mac()" with the parentheses.

yonghong-song: You are correct. It is my bad. Just tried and works perfectly fine.

examples/simulation.py
yonghong-song: I just tried the patch (git checkout bblanco_dev) and tried on my local box:

9: F
9: ======================================================================
9: FAIL: test_brb2 (__main__.TestBPFSocket)
9: ----------------------------------------------------------------------
9: Traceback (most recent call last):
9:   File "/home/plumgrid/iovisor/bcc/tests/cc/test_brb2.py", line 165, in
test_brb2
9:     self.assertEqual(self.pem_stats[c_uint(0)].value, 12)
9: AssertionError: 23L != 12

So it does seem we also need to disable ipv6 for outer ifc as well.

drzaeus77: I see. You are using a bridge to link the two veth pairs together. It could be done by inserting one end of a veth pair into the other namespace directly, such that neither of them show up in the abstract namespace. In fact, the support for this is exactly what this function is supposed to achieve, but in a different set of code example (in a private codebase, so you don't see it here). I will amend the patch to use that approach, and eliminate the extra bridges.

yonghong-song: This will be fine too. Please do add some comments in simulation.py though
to warn that if users do have one end of veth not in the namespace, they
will be responsible to disable ipv6 if it is desirable to them.

drzaeus77: I went back to the original, since I realized that you actually wanted the outer bridges to exist as part of the test scenario.

PR #86, opened 2015-07-15
Map functions are now pass-by-value

drzaeus77:
As a convenience to avoid extra variables, trying out if passing by
value and using the rewriter to construct the pointer is cleaner. It
seems that the code has fewer lines, though in some cases the lines
are longer.

The best example of a cleaner source file is in tc_neighbor_sharing.c.

### DIFF
--- a/examples/tc_neighbor_sharing.c
+++ b/examples/tc_neighbor_sharing.c
@@ -29,9 +29,7 @@ int classify_wan(struct __sk_buff *skb) {
   }
   ip: {
     struct ip_t *ip = cursor_advance(cursor, sizeof(*ip));
-    u32 dip = ip->dst;
-    struct ipkey key = {.client_ip=dip};
-    int *val = learned_ips.lookup(&key);
+    int *val = learned_ips.lookup((struct ipkey){ip->dst});
     if (val)
       return *val;
     goto EOP;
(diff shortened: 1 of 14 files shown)


drzaeus77:
Please consider this a RFC.

examples/tc_neighbor_sharing.c
```diff
@@ -29,9 +29,7 @@ int classify_wan(struct __sk_buff *skb) {
   }
   ip: {
     struct ip_t *ip = cursor_advance(cursor, sizeof(*ip));
-    u32 dip = ip->dst;
-    struct ipkey key = {.client_ip=dip};
-    int *val = learned_ips.lookup(&key);
+    int *val = learned_ips.lookup((struct ipkey){ip->dst});
```
yonghong-song: Maybe, you can use lookup(&(struct ipkey){ip->dst}) to still maintain pointers?

examples/tc_neighbor_sharing.c
```diff
@@ -54,10 +52,7 @@ int classify_neighbor(struct __sk_buff *skb) {
   }
   ip: {
     struct ip_t *ip = cursor_advance(cursor, sizeof(*ip));
-    u32 sip = ip->src;
-    struct ipkey key = {.client_ip=sip};
-    int val = 1;
-    learned_ips.update(&key, &val);
+    learned_ips.update((struct ipkey){ip->src}, 1);
```
yonghong-song: This one definitely favors pass-by-value. "int val = 1; ... &val ..."  is no match to just "... 1 ...".

PR #88, opened 2015-07-20
Fix cmake problem if using pre-packaged libllvm3.7/llvm-3.7-dev etc. …

yonghong-song:
…from ubuntu

  o The ubuntu package build cmake uses -DBUILD_SHARED_LIBS=ON flag which generates
    different LLVMExports.cmake than with -DBUILD_SHARED_LIBS=OFF.
  o This patch uses a version of LLVMExports.cmake generated with -DBUILD_SHARED_LIBS=OFF
    to keep track of dependencies of different LLVM libraries so our makefile does not
    need to keep track of all needed LLVM libraries,


drzaeus77:
The inclusion of pre-baked cmake files worries me a little bit. Before merging, I will try my hand at a different solution.

drzaeus77:
This is a bug in the debian packaging from LLVM. The bug is already open: https://llvm.org/bugs/show_bug.cgi?id=24154. I'm not going to merge this workaround, in the hopes that LLVM is fixed in the next month before it reaches 3.7 final.

PR #90, opened 2015-07-22
simply vlan_learning example with newer vlan_push/pop helpers

yonghong-song:

### DIFF
--- a/examples/vlan_learning.c
+++ b/examples/vlan_learning.c
@@ -5,6 +5,8 @@
 
 struct ifindex_leaf_t {
   int out_ifindex;
+  int vlan_tci; // populated by phys2virt and used by virt2phys
+  int vlan_proto; // populated by phys2virt and used by virt2phys
   u64 tx_pkts;
   u64 tx_bytes;
 };
@@ -16,6 +18,9 @@ BPF_TABLE("hash", int, struct ifindex_leaf_t, egress, 4096);
 BPF_TABLE("hash", u64, struct ifindex_leaf_t, ingress, 4096);
 
 int handle_phys2virt(struct __sk_buff *skb) {
+  // only handle vlan packets
+  if (!skb->vlan_present)
+    return 1;
   u8 *cursor = 0;
   ethernet: {
     struct ethernet_t *ethernet = cursor_advance(cursor, sizeof(*ethernet));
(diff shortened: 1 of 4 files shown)


examples/vlan_learning.c
```diff
@@ -31,6 +36,13 @@ int handle_phys2virt(struct __sk_buff *skb) {
       // relearn when mac moves ifindex
       if (out_leaf->out_ifindex != skb->ifindex)
         out_leaf->out_ifindex = skb->ifindex;
+      // relearn when vlan_tci/vlan_proto changes
+      if (out_leaf->vlan_tci != skb->vlan_tci)
+        out_leaf->vlan_tci = skb->vlan_tci;
+      if (out_leaf->vlan_proto != skb->vlan_proto)
+        out_leaf->vlan_proto = skb->vlan_proto;
+      // pop the vlan header and send to the destination
+      bpf_skb_vlan_pop(skb);
```
4ast: looks great. so what was the problem with rst+ack ?

examples/vlan_learning.c
```diff
@@ -31,6 +36,13 @@ int handle_phys2virt(struct __sk_buff *skb) {
       // relearn when mac moves ifindex
       if (out_leaf->out_ifindex != skb->ifindex)
         out_leaf->out_ifindex = skb->ifindex;
+      // relearn when vlan_tci/vlan_proto changes
+      if (out_leaf->vlan_tci != skb->vlan_tci)
```
drzaeus77: Can be assignment without conditional, no? Same for above line. Fewer branches.

examples/vlan_learning.c
yonghong-song: In my old python changes, I disabled ipv6 in various places so that tcpdump
packets will be clean.
However, looks like pyroute2 does not really work well with "sysctl ..."
(This happens before with confusing messages), and
the worker http server (in worker namespace: python -m SimpleHTTPServer 80)
does not come up, thus tcp connection cannot be
established.

Without disabling ipv6, everything works fine.

I tried to use "ip" command to create veth, create namespace and disable
ipv6 inside namespace and launch SimpleHTTPServer
inside the namespace, there is no issue.

examples/vlan_learning.c
yonghong-song: Yes, we can do. I thought this before. But finally I just want to conform
to the existing style.
Let me change that.

PR #92, opened 2015-07-28
Add index-based function references to c api

drzaeus77:
* Add the index based API so that an external api (bcc-fuse) can iterate
  through the maps and functions that were compiled in this program.
* Add a more unique section prefix to disambiguate exportable functions.

### DIFF
--- a/src/cc/b_frontend_action.cc
+++ b/src/cc/b_frontend_action.cc
@@ -25,9 +25,7 @@
 
 #include "b_frontend_action.h"
 
-extern "C"
-int bpf_create_map(enum bpf_map_type map_type, int key_size, int value_size,
-                   int max_entries);
+#include "libbpf.h"
 
 namespace ebpf {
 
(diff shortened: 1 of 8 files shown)


src/cc/bpf_module.cc
```diff
@@ -76,6 +76,8 @@ using std::unique_ptr;
 using std::vector;
 using namespace llvm;
 
+const string BPFModule::FN_PREFIX = ".bpf.fn.";
```
4ast: can the same constant be used in all three places? via ifdef or something?

drzaeus77:
Updated.

src/cc/b_frontend_action.cc
```diff
@@ -90,7 +90,7 @@ bool BTypeVisitor::VisitFunctionDecl(FunctionDecl *D) {
   // put each non-static non-inline function decl in its own section, to be
   // extracted by the MemoryManager
   if (D->isExternallyVisible() && D->hasBody()) {
-    string attr = string("__attribute__((section(\".") + D->getName().str() + "\")))\n";
+    string attr = string("__attribute__((section(\".bpf.fn.") + D->getName().str() + "\")))\n";
```
4ast: and here too?

PR #108, opened 2015-08-04
"Ubuntu - From source" section no longer works.

weichunc:
Comment it out until LLVM is fixed.

Signed-off-by: Wei-Chun Chao <weichunc@plumgrid.com>


drzaeus77:
Hmm, this approach doesn't seem too helpful. Maybe we can add the manual build steps instead of the strike-through? We can always get the previous instructions from git history later.

weichunc:
ok, let me try out the manual steps and update.

PR #110, opened 2015-08-04
better clean up of created namespace/interfaces if ctrl-c is pressed

yonghong-song:
o There are still some cases when ctrl-c is pressed, the program does not exit
    gracefully, with a however lower probability. In my random manual experiments,
    one in ten cases it still hangs and needs further ctrl-z.


drzaeus77:
Nice! The vlan and tc_neighbor_sharing examples work well for me with this. However, distributed_bridge may still need some fine tuning.

PR #118, opened 2015-08-05
Add cmake checks for required test programs

drzaeus77:
Set as warning, since the build will still work fine and be able to
create a binary package.

### DIFF
--- a/tests/cc/CMakeLists.txt
+++ b/tests/cc/CMakeLists.txt
@@ -1,5 +1,19 @@
 # Copyright (c) PLUMgrid, Inc.
 # Licensed under the Apache License, Version 2.0 (the "License")
+
+find_program(ARPING arping)
+if(ARPING STREQUAL "ARPING-NOTFOUND")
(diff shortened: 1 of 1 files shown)


tests/cc/CMakeLists.txt
```diff
@@ -1,5 +1,19 @@
 # Copyright (c) PLUMgrid, Inc.
 # Licensed under the Apache License, Version 2.0 (the "License")
+
+find_program(ARPING arping)
+if(ARPING STREQUAL "ARPING-NOTFOUND")
+  message(WARNING "Required test program 'arping' not found")
```
4ast: if it's warning then it should say 'recommended' instead of 'required'?

drzaeus77: I guess...it's "required" for test but not for build?

yonghong-song: Yes. arping/iperf/netperf are used in python codes in several tests. They
will be needed in test running time.

PR #119, opened 2015-08-09
Patch series: reorganize bpf_module and add sscanf feature for fuse

drzaeus77:


src/cc/bpf_module.cc
```diff
@@ -95,6 +98,8 @@ class MyMemoryManager : public SectionMemoryManager {
 
 BPFModule::BPFModule(unsigned flags)
     : flags_(flags), ctx_(new LLVMContext) {
+  InitializeNativeTarget();
```
4ast: I was hoping we can eventually get rid of x86 backend dependency, but this forces it in. Why do we need it?

drzaeus77: The idea is to run a native sscanf function that is tailored to to the key/leaf type. The problem with building a custom parser is that the alignment of the struct is convoluted, but the Type that is kept in the llvm::Module is accurate according to how the frontend (clang or lex, doesn't matter) compiled the struct, taking alignments, padding, and so on into account. This implementation leverages that knowledge, but as a downside uses native JITed code. We can discuss in person the overall approach if this doesn't satisfy. I acknowledge that the dependency is an ugly one.

4ast: I see. I think
b.update_table("stats", "2", "{ 2 3 0x1000000004 { 5 6 }}")
is getting too fancy. Using native sscanf for the string to get it into binary blob to send via bpf syscall... I don't see much advantage of the string vs creating the blob manually. I thought this is to help reading? then it would make a lot more sense.

drzaeus77: The use case is not in python at all. The use case is for fuse, where the daemon process has no facility to understand the types inside of the maps. Suppose I write to a file, in order to create an entry (key=1, value=2, types are ints):
echo "1 2" > new_entry
My problem statement is how can I have a string-based, simple input/ouput format that works with file-system syntax. The example above is easy enough to implement by hand, but breaks down with structs/unions/nested structs etc. I would like to support those. The implementation seen in this patch is not the best or most portable one, but it lets us prototype such an API on top. Whichever implementation can achieve that I would be happy with. JIT is a fancy trick but just a means to an end. Complex formatted string input is the goal.

Maybe it would help to see a more complete use case?

4ast: ahh. ok. I hope both key and value will be possible to represent this way eventually, like
echo "{1 2}  {3 4}" > new_entry
where key and value are structs of two u32s.

drzaeus77: Yup! Exactly.

PR #150, opened 2015-08-18
some bcc examples and tools

brendangregg:


examples/bitehist.py
```diff
@@ -0,0 +1,109 @@
...
+			raise
+		if len(argv) > 2:
+			count = int(argv[2])
+	except:	# also catches -h, --help
+		usage()
+
+# load BPF program
+b = BPF(src_file = "bitehist.c")
+BPF.attach_kprobe(b.load_func("do_request", BPF.KPROBE), "blk_start_request")
+dist = b.get_table("dist", c_int, c_ulonglong)
```
drzaeus77: Are the explicit arguments (c_int, c_ulonglong) to get_table intentional? We do support automatic key/leaf detection, so this is a minor breakage of DRY

brendangregg: No, explicit was not intentional. I'll drop these and update the PR - thanks!

tools/syncsnoop
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/python
```
drzaeus77: Can you convert these to /usr/bin/env python?

brendangregg: I can, although it wastes some syscalls....

drzaeus77:
Unfortunately, this code is not python 3 compatible. I have been trying to keep BPF working for both versions. I will test a few more files, but so far I have found the following diff necessary:

```diff
diff --git a/tools/syncsnoop b/tools/syncsnoop
index 25f1e8c..78806de 100755
--- a/tools/syncsnoop
+++ b/tools/syncsnoop
@@ -24,13 +24,14 @@ int do_sync(void *ctx) {
 BPF.attach_kprobe(b.load_func("do_sync", BPF.KPROBE), "sys_sync")
 
 # header
-print "%-18s %s" % ("TIME(s)", "CALL")
+print("%-18s %s" % ("TIME(s)", "CALL"))
 
 # open trace pipe
 try:
        trace = open("/sys/kernel/debug/tracing/trace_pipe", "r")
 except:
-       print >> sys.stderr, "ERROR: opening trace_pipe"
+       sys.stdout = sys.stderr
+       print("ERROR: opening trace_pipe")
        exit(1)
 
 # format output
@@ -43,4 +44,4 @@ while 1:
        prolog, time_s, colon, word = line.rsplit(" ", 3)
        time_s = time_s[:-1]    # strip trailing ":"
 
-       print "%-18s %s" % (time_s, word)
+       print("%-18s %s" % (time_s, word))
```

tools/syncsnoop
drzaeus77: I spoke too soon! I should have looked at some actual examples, since I found that most (all?) of my installed machine uses /usr/bin/python explicitly. Sorry for flip-flopping, but since I already have the other comment pending, can you change this back? Shame on me :)

tools/vfscount
```diff
@@ -0,0 +1,73 @@
...
+			continue
+		addr = int(cols[0], 16)
+		ksym_addrs.append(addr)
+		ksym_names.append(name)
+	syms.close()
+def ksym(addr):
+	start = -1
+	end = len(ksym_addrs)
+	while end != start + 1:
+		mid = (start + end) / 2
```
drzaeus77: I get a traceback when mid is used as an index in the line below, please change to:
mid = int((start + end) / 2)

brendangregg:
Oh, right. (Our systems default to 2, but I can test on both). I'll fix.

I think the stderr is better as:

```Python
print("ERROR: opening trace_pipe", file=sys.stderr)
```

tools/vfsstat
```diff
@@ -0,0 +1,87 @@
...
+# Written as a basic example of counting multiple events as a stat tool.
+#
+# USAGE: vfsstat [interval [count]]
+#
+# Copyright (c) 2015 Brendan Gregg.
+# Licensed under the Apache License, Version 2.0 (the "License")
+#
+# 14-Aug-2015	Brendan Gregg	Created this.
+
+from __future__ import print_function
```
drzaeus77: As mentioned in another comment, the un-parenthesized print breaks in python 3. Importing print_function is a valid solution, as well as the more hackish sys.stdout = sys.stderr that I mentioned. Whichever compatibility approach you choose, please choose consistently :) Probably the import is the right one.

drzaeus77:
Yes, please go with file=sys.stderr. Thanks!

brendangregg:
Yes, I think the future import (when needed) is easier to delete in the future when cleaning this up.

tools/vfscount
brendangregg: got it, thanks

tools/syncsnoop
brendangregg: All fixed!

Issue #149, opened 2015-08-18
Add a read_trace_fields() Python function


brendangregg:
As a follow on from #136, a function could be added that returned the fields printed by bpf_trace_printk(). Eg, so one can use:

```Python
b = BPF(...)
[...]
task, pid, cpu, flags, timestamp, function, arguments = b.read_trace_fields()
```

This is slightly harder than it sounds: I believe the task name can contain spaces.

drzaeus77:
I added the suggested change, as well as an option to trace_print:
example simplistic json printer:
```python
b.trace_print(fmt='{{"pid":{1}, "msg":"{5}"}}')
```

```bash
sudo ./hello_world.py 
{"pid":1118, "msg":"Hello, World!"}
{"pid":1118, "msg":"Hello, World!"}
{"pid":8696, "msg":"Hello, World!"}
```

brendangregg:
Thanks, trace_readline_fields() works for me.

I'm not sure about the fmt= option's stability. Eg, the number of fields have changed in the past, which if it happened again would break indexes. But that's probably very unlikely...

Here's an example of using the new function (disksnoop.py):

Before:

```Python
while 1:
        try:
                line = trace.readline().rstrip()
        except KeyboardInterrupt:
                pass; exit()
        prolog, time_s, colon, bytes_s, flags_s, us_s = \
                line.rsplit(" ", 5)

        time_s = time_s[:-1]    # strip trailing ":"
```

After:

```Python
while 1:
        (task, pid, cpu, flags, ts, msg) = b.trace_readline_fields()
        (bytes_s, bflags_s, us_s) = msg.split()
```

So that's getting much better. And this should be commonly used.

Maybe as a follow on issue, we could have trace_readline_fields() split the msg on whitespace, so the above could be one line. I also think this may be better called just trace_fields().

drzaeus77:
Shortening to trace_fields is fine to me.

Splitting msg doesn't seem right to me, since the formatting of msg is completely up to the user...trace_fields() can't know to split or not to split.

PR #156, opened 2015-08-21
Add BPF python helpers for reading trace_pipe

drzaeus77:

### DIFF
--- a/examples/hello_world.py
+++ b/examples/hello_world.py
@@ -16,7 +16,4 @@ int hello(void *ctx) {
 """
 b = BPF(text=prog)
 b.attach_kprobe(event="sys_clone", fn_name="hello")
-try:
-    call(["cat", "/sys/kernel/debug/tracing/trace_pipe"])
-except KeyboardInterrupt:
-    pass
+b.trace_print()
--- a/src/python/bpf/__init__.py
+++ b/src/python/bpf/__init__.py
@@ -15,6 +15,7 @@
 import atexit
 from collections import MutableMapping
 import ctypes as ct
+import fcntl
 import json
 import os
 import sys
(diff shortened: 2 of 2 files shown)


brendangregg:
Sorry to overload, but here's 3 small suggestions:

1) I'd default trace_readline() to blocking, given the nature of these tools. I forgot to set it to false (like it is in trace_print()) and got this:

```Shell
# python3 syncsnoop
TIME(s)            CALL
  File "syncsnoop", line 33, in <module>
    line = b.trace_readline()
  File "/usr/lib/python3/dist-packages/bpf/__init__.py", line 422, in trace_readline
    trace = BPF.trace_open(nonblocking)
  File "/usr/lib/python3/dist-packages/bpf/__init__.py", line 409, in trace_open
    fd = trace.fileno()
NameError: name 'trace' is not defined
```

2) I'd also seriously consider making this exception handling inbuilt to trace_readline():

```Python
        try:
                line = b.trace_readline()
        except KeyboardInterrupt:
                pass; exit()
```

So we can just call b.trace_readline().

This means a Ctrl-C will always exit() the program if you are using trace_readline(). I think this is desirable, since the signal means a human has hit Ctrl-C, and the common expected behavior would be to exit. (This exception handling should be the same with #149 as well). It'll save some boilerplate in many common scripts.

3) Is there a way to define BlockingIOError for python 2.6?:

```Shell
# ../examples/hello_world.py
            bash-14044 [001] d... 29450763.501397: : Hello, World!
^CTraceback (most recent call last):
  File "../examples/hello_world.py", line 19, in <module>
    b.trace_print()
  File "/usr/lib/python2.7/dist-packages/bpf/__init__.py", line 435, in trace_print
    line = BPF.trace_readline(nonblocking=False)
  File "/usr/lib/python2.7/dist-packages/bpf/__init__.py", line 427, in trace_readline
    except BlockingIOError:
NameError: global name 'BlockingIOError' is not defined
```

PR #161, opened 2015-08-24
examples: use IPDB `wait_ip()` feature

svinota:
The feature allows to have a bit cleaner code.

This particular code piece is just a proposal, it wasn't tested, though the feature itself is working and covered by the the CI. Actually, it was made after this sample script, but may be useful in other cases as well.


drzaeus77:
Thanks, much cleaner!

PR #169, opened 2015-08-27
Add example of live kernel assembly debugging

drzaeus77:

### DIFF
--- a/examples/reg_debugger.py
+++ b/examples/reg_debugger.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python
+# Copyright (c) PLUMgrid, Inc.
+# Licensed under the Apache License, Version 2.0 (the "License")
+#
+# This example shows how a BPF program can single step through the calculations
+# in a single function. The use case in this scenario was to see the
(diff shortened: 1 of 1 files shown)


examples/reg_debugger.py
```diff
@@ -0,0 +1,46 @@
...
+regs = ["di", "si", "dx", "cx", "ax", "bx", "bp"]
+regs += ["r%d" % i for i in range(8, 15)]
+
+
+prog = """
+#include <linux/ptrace.h>
+"""
+for r in regs:
+    prog += reg_printer("print_%s" % r, r)
+    prog += reg_printer("print_%s_net" % r, r, "bpf_ntohll")
```
4ast: this line is unused.
In general, the whole .py doesn't look like an example. Most likely it works with only one specific kernel. No one can even run it to get meaningful results. If it's here to demonstrate sed of C program via python then some other example is needed.

drzaeus77: I'm ok leaving this code out of the branch. I just wanted to get it out of my private workspace and somewhere visible. We can close the pull request and refer to it as needed or resurrect when a valid use case is found.

PR #172, opened 2015-08-29
initial ksym() and ksymaddr()

### LINKED ISSUE #140: Kernel symbol translation
There should be a Python function for translating kernel addresses to symbol names: eg, BPF.ksym(addr). As an example, here's some code I was using (feel free to use/improve):

```Python
# kernel symbol translation
ksym_addrs = []			# addresses for binary search
ksym_names = []			# same index as ksym_addrs
def load_kallsyms():
	symfile = "/proc/kallsyms"
	try:
		syms = open(symfile, "r")
	except:
		print >> stderr, "ERROR: reading " + symfile
		exit()
	line = syms.readline()
	for line in iter(syms):
		cols = line.split()
		name = cols[2]
		addr = int(cols[0], 16)
		ksym_addrs.append(addr)
		ksym_names.append(name)
	syms.close()
def ksym(addr):
	start = -1
	end = len(ksym_addrs)
	while end != start + 1:
		mid = (start + end) / 2
		if addr < ksym_addrs[mid]:
			end = mid
		else:
			start = mid
	if start == -1:
		return "[unknown]"
	return ksym_names[start]
load_kallsyms()
```

The python program could then call ksym() on an address, to get the symbol name from /proc/kallsyms.

In hindsight I'd adjust this so that ksym() checked if symbols were loaded, and if not, called load_kallsyms() itself. That would avoid needing to call load_kallsyms() explicitly. The end-user would simply use ksym() if they needed it.

It may also be useful to provide a ksymaddr() function, which returned the symbol name followed by the instruction offset. Eg, "vfs_read+0x18" instead of just "vfs_read".

brendangregg:
Fixes #140

### DIFF
--- a/tools/vfscount
+++ b/tools/vfscount
@@ -16,43 +16,6 @@ from ctypes import c_ushort, c_int, c_ulonglong
 from time import sleep, strftime
 from sys import stderr
 
-# kernel symbol translation
-ksym_addrs = []			# addresses for binary search
-ksym_names = []			# same index as ksym_addrs
-def load_kallsyms():
-	symfile = "/proc/kallsyms"
-	try:
-		syms = open(symfile, "r")
-	except:
-		print("ERROR: reading " + symfile, file=sys.stderr)
-		exit()
-	line = syms.readline()
-	for line in iter(syms):
-		cols = line.split()
-		name = cols[2]
-		if name[:4] != "vfs_":	# perf optimization
-			continue
-		addr = int(cols[0], 16)
-		ksym_addrs.append(addr)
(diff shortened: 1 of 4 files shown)


tools/vfscount
```diff
@@ -16,47 +16,9 @@ from ctypes import c_ushort, c_int, c_ulonglong
 from time import sleep, strftime
 from sys import stderr
```
drzaeus77: With this cleanup, ctypes, sys, and strftime can be removed from the import list.

drzaeus77:
Awesome, thanks!

PR #173, opened 2015-09-01
Add RPM packaging for el6,el7,fedora22

drzaeus77:


4ast:
looks good, but non-empty commit log would be even better. My understanding that neither deb nor rpm packages are run by buildbot yet, right?

yonghong-song:
deb packages are generated from source and run by buildbot.

Issue #143, opened 2015-08-18
Print log2 histograms from Python


brendangregg:
There should be a print_log2_hist() function. Example tool output:

```
# ./bitehist.py
Tracing... Hit Ctrl-C to end.

     kbytes          : count     distribution
       0 -> 1        : 0        |                                      |
       2 -> 3        : 0        |                                      |
       4 -> 7        : 26       |*************                         |
       8 -> 15       : 3        |*                                     |
      16 -> 31       : 5        |**                                    |
      32 -> 63       : 6        |***                                   |
      64 -> 127      : 7        |***                                   |
     128 -> 255      : 75       |**************************************|
```

The Python code used (feel free to take/improve):

```Python
last = {}
for i in range(1, dist_max + 1):
	last[i] = 0

# functions
stars_max = 38
def stars(val, val_max, width):
	i = 0
	text = ""
	while (1):
		if (i > (width * val / val_max) - 1) or (i > width - 1):
			break
		text += "*"
		i += 1
	if val > val_max:
		text = text[:-1] + "+"
	return text

def print_log2_hist(d, val_type):
	idx_max = -1
	val_max = 0
	for i in range(1, dist_max + 1):
		try:
			val = dist[c_int(i)].value - last[i]
			if (val > 0):
				idx_max = i
			if (val > val_max):
				val_max = val
		except:
			break
	if idx_max > 0:
		print("     %-15s : count     distribution" % val_type);
	for i in range(1, idx_max + 1):
		low = (1 << i) >> 1
		high = (1 << i) - 1
		if (low == high):
			low -= 1
		try:
			val = dist[c_int(i)].value - last[i]
			print("%8d -> %-8d : %-8d |%-*s|" % (low, high, val,
			    stars_max, stars(val, val_max, stars_max)))
			last[i] = dist[c_int(i)].value
		except:
			break

[...]
	print_log2_hist(dist, "kbytes")
```

The only hitch here is that I'm printing a per-interval histogram, and needed to subtract the previous map values, hence the use of the "last" dictionary.

If there was a way to clear a map (or pretend to; this is issue #142), then the "last" logic can be dropped, simplifying this a bit.

brendangregg:
Something else to think about: support of histograms for each key. Imagine the bitehist histogram was saved per-PID, per-comm, per-device-name, or some combination?

[...]

PR #185, opened 2015-09-07
print_log2_hist() for #143

brendangregg:
I made it a Table function, I think this makes sense. The new bitehist example is much nicer.

### DIFF
--- a/examples/bitehist.py
+++ b/examples/bitehist.py
@@ -5,8 +5,6 @@
 #
 # Written as a basic example of using a histogram to show a distribution.
 #
-# USAGE: bitehist.py [interval [count]]
-#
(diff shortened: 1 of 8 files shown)


brendangregg:
not sure why the checks failed -- maybe because I pushed an update while the build was in progress?

drzaeus77:
Yes, that is exactly why.
```
From https://github.com/iovisor/bcc
 * branch            refs/pull/185/head -> FETCH_HEAD
git reset --hard 08af9e47963e3ee23756e11776c24ffa24476e85 --
fatal: Could not parse object '08af9e47963e3ee23756e11776c24ffa24476e85'.
```

Looks like it sorted itself out.

PR #194, opened 2015-09-09
Support automatic kprobe event detection in common case

drzaeus77:
* In the simple case, a user only creates 1 C function to be used with
  kprobes. Detect this common case and don't require the user to repeat
  themselves by passing the fn_name to attach_kprobe().
  e.g.: BPF(text='int sys_clone(void *ctx) {/*do stuff*/}').trace_print()

### DIFF
--- a/examples/hello_world.py
+++ b/examples/hello_world.py
@@ -8,6 +8,4 @@
 
 from bcc import BPF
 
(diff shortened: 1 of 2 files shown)


brendangregg:
That's great, thanks, but as soon as people start doing this with 1 function, they'll likely expect 2 and more to work as well, if the C function name matches something from /sys/kernel/debug/tracing/available_filter_functions.

PR #195, opened 2015-09-09
Change auto-loading behavior of trace_print

drzaeus77:
* As @brendangregg pointed out, users will probably assume that
  handily-named C functions that can be auto-loaded will be all the
  time, rather than just in the singleton case. This is pretty easy to
  implement, so changing the behavior.

Fixes: #191 
Signed-off-by: Brenden Blanco <bblanco@plumgrid.com>

PR #196, opened 2015-09-09
Add debug and fix the inline replace of kprobe args

drzaeus77:


drzaeus77:
This change depends on #195, I will re-trigger tests after that merges.

src/cc/bpf_module.cc
```diff
@@ -295,7 +295,7 @@ unique_ptr<ExecutionEngine> BPFModule::finalize_rw(unique_ptr<Module> m) {
 
 // load an entire c file as a module
 int BPFModule::load_cfile(const string &file, bool in_memory) {
-  clang_loader_ = make_unique<ClangLoader>(&*ctx_);
+  clang_loader_ = make_unique<ClangLoader>(&*ctx_, flags_);
```
4ast: flags are propagated, but I don't see who sets them or how they used?

drzaeus77: flags is the debug=?? field of the BPF __init__ constructor. Currently there are users of 0x1, 0x2, and 0x4 bits, but not necessarily in all classes that flags_ is passed is it used.

Issue #191, opened 2015-09-08
auto attach BPF functions to kprobes


brendangregg:
I think this is worth suggesting... I keep declaring C function names that only exist for an attach_kprobe() mapping, and wonder if this could be simplified.

Eg, here's the existing hello_world:

```Python
b = BPF(text='void hello(void *ctx) { bpf_trace_printk("Hello, World!\\n"); }')
b.attach_kprobe(event="sys_clone", fn_name="hello")
b.trace_print()
```

imagine:

```Python
b = BPF(text='kprobe:sys_clone(void *ctx) { bpf_trace_printk("Hello, World!\\n"); }')
b.trace_print()
```

I think this will work well with tracepoints, when their support is added; eg, imagine tracing the "sched:sched_process_fork" tracepoint using:

```Python
b = BPF(text='sched:sched_process_fork(void *ctx) { bpf_trace_printk("Hello, World!\\n"); }')
b.trace_print()
```

brendangregg:
I should note that attach_kprobe() could stay around. This would be an optional shortcut.

drzaeus77:
I am hesitant to instrument the C syntax in the way mentioned (`sched:sched_process_fork`). Though we mangle a little bit the programs that are created, they are still all valid C syntax. If/when we go down the route of a custom tracing language, then this can be introduced.

That said, there is an optimization in the api to be done without changing the supported C syntax. What do you think about this compromise:

```python
BPF(text='int sys_clone(void *ctx) { bpf_trace_printk("Hello, World!\\n"); }').attach_kprobe("sys_clone").trace_print()
```

The change to support a single function BPF module is a pretty small diff:

```patch
diff --git a/src/python/bcc/__init__.py b/src/python/bcc/__init__.py
index 36ef17e..013da1a 100644
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -468,6 +468,13 @@ class BPF(object):
                         cpu=cpu, group_fd=group_fd)
             return
 
+        # If fn_name is omitted, auto-detect the function in singular case
+        if not fn_name:
+            if len(self.funcs) == 1:
+                fn_name = self.funcs.items()[0][0]
+            else:
+                raise Exception("Missing parameter fn_name not provided")
+
         fn = self.load_func(fn_name, BPF.KPROBE)
         ev_name = "p_" + event.replace("+", "_")
         desc = "p:kprobes/%s %s" % (ev_name, event)
```

[...]

PR #198, opened 2015-09-09
Autoload kprobes for all types of trace_* functions

drzaeus77:
The previous patch #195 for autoloading of kprobes only did it for
trace_print. Turn this feature on for all trace_* functions. This
requires that these functions are also no longer staticmethods.

Enable the feature in examples/disksnoop.py

Fixes: #191 


drzaeus77:
@brendangregg last change for this feature, I hope!

PR #201, opened 2015-09-10
Always autoload k[ret]probe__ prefixed functions

drzaeus77:
This will shorten some examples, no longer requiring them to call
attach_kprobe.

Fixes: #191 


4ast:
interface is shaping up nicely.

PR #210, opened 2015-09-13
add ebpf method to retrieve bytecode

rlane:
This is useful if you want to use bcc as a compiler without running the
program.

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -392,7 +392,7 @@ class BPF(object):
 
         return fn
 
-    def ebpf(self, func_name):
+    def dump_func(self, func_name):
         """
         Return the eBPF bytecodes for the specified function as a string
         """
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -392,6 +392,17 @@ class BPF(object):
 
         return fn
 
+    def ebpf(self, func_name):
+        """
+        Return the eBPF bytecodes for the specified function as a string
+        """
+        if lib.bpf_function_start(self.module, func_name.encode("ascii")) == None:
+            raise Exception("Unknown program %s" % func_name)
+
+        start, = lib.bpf_function_start(self.module, func_name.encode("ascii")),
+        size, = lib.bpf_function_size(self.module, func_name.encode("ascii")),
+        return ct.string_at(start, size)
+
     str2ctype = {
         u"_Bool": ct.c_bool,
         u"char": ct.c_char,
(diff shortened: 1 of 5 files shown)


drzaeus77:
I have a question..how will the program dump work with maps? There is a quirk with how eBPF programs deal with maps, in that they are referred to in the program by a process-specific fd. If a program that references a BPF_TABLE is dumped as a binary string and passed to another process, the other process won't be able to run that program since the maps haven't been opened by that process, and/or the fds may have a different number.

Maybe if you can explain the use case a little bit, we can discuss if the approach that you are taking here is good enough, or if there is some alternative that can be worked towards.

rlane:
Thanks for taking a look. I didn't realize these fds were hard coded in the
bytecode, though that makes sense. It looks like the fd is inserted in
b_frontend_action.cc.

My goal is to compile B source into eBPF bytecode and run it on my
userspace JIT (http:// <http://github.com/rlane/ubpf>github.com
<http://github.com/rlane/ubpf>/rlane/ <http://github.com/rlane/ubpf>ubpf
<http://github.com/rlane/ubpf>). It doesn't support maps yet.

Ignoring alternate implementations, is there interest in being able to
compile B to eBPF and ship the bytecode to other machines? Are there other
roadblocks besides map fds? I'm assuming that eBPF bytecode programs are
portable between kernel versions and architectures, though that may be a
bad assumption.

It may be too late to add this to the kernel api, but I'd suggest an
indirection between table ids and map fds, so that the userspace program
loading eBPF would bind an fd to each table id.

drzaeus77:
I'm ok taking the patch more or less as is, but would suggest a different name, something like 'dump_func' as a corollary to load_func.

Also, if you want this API to remain functional, I would highly suggest adding a test case into the repo as well.

Regarding shipping code across machines, BCC is intended to be usable as a library on whichever target machine. The libbcc.so library only has a runtime dependency on libc, and I intend to keep it that way. The dependency on clang/llvm for instance is build-only, and those libraries have a wide array of supported cross-compile architectures. Shipping bytecode is probably still in the realm of possibility, but that would definitely need to be the design goal of some other tool besides BCC.

Pertaining to the kernel API, there's no way that a new mechanism besides FD-based will make it in. The instruction set of eBPF is considered part of the kernel ABI, meaning it is set in stone. In addition, adding extensions such as ID->FD mapping wouldn't fly, since that would effectively introduce two ways to do the same thing, which is also against kernel philosophy.

I would suggest that the userspace side of things hack around FDs to give them meaning.

rlane:
I renamed the method to 'dump_func' and added a test for it. Unfortunately the test is failing on the Ubuntu buildbot. Could you change the buildbot to pass the '-V' flag to ctest so that I can see the reason for the failure?

drzaeus77:
I've turned on -V in the buildbot for both flavors, and also fixed the underlying issue (directory permissions). You can see the failure [here](https://52.8.191.206/buildbot/builders/runtests-ubuntu/builds/72/steps/shell_7/logs/stdio).

Unfortunately, I cannot merge this change since your fork is out of date with master. Can you rebase it?

rlane:
Ok, rebased.

PR #211, opened 2015-09-13
updated  mainline version to support bridge

affansyed:
This version of the mainline kernel supports the bridge and vlan learning examples (i.e. the additional APIs). Will allow new users to run all examples provided.


drzaeus77:
First change looks good, but please fetch again from master and recreate the branch to avoid the extra delta.

Issue #188, opened 2015-09-07
Error digging out d_name from struct file


brendangregg:
Fetching a file pathname from a struct file should be a common routine, but I can't get it to work.

vfs_read.py:
```Python
#!/usr/bin/python

from bcc import BPF

# load BPF program
b = BPF(src_file="vfs_read.c")
b.attach_kprobe(event="vfs_read", fn_name="run_read")

# format output
while 1:
	(task, pid, cpu, flags, ts, msg) = b.trace_fields()
	print("got: %s" % msg)
```

vfs_read.c:
```C
# cat vfs_read.c
#include <linux/fs.h>

void run_read(struct pt_regs *ctx, struct file *file)
{
	void *name = 0;

	if (file != NULL && file->f_path.dentry != NULL) {
		//bpf_trace_printk("read %s\n", file->f_path.dentry->d_iname);
		bpf_trace_printk("read %s\n", file->f_path.dentry->d_name.name);
	} else {
		bpf_trace_printk("was null");
	}
}
```

The line commented out works, but when I dig further to the d_name it does this:

```Shell
# ./vfs_read.py 
bpf: Permission denied
0: (bf) r6 = r1
1: (79) r3 = *(u64 *)(r6 +112)
2: (15) if r3 == 0x0 goto pc+29
 R1=ctx R3=inv R6=ctx R10=fp
3: (b7) r7 = 0
4: (7b) *(u64 *)(r10 -8) = r7
5: (7b) *(u64 *)(r10 -16) = r7
6: (07) r3 += 16
7: (bf) r1 = r10
8: (07) r1 += -16
9: (b7) r2 = 16
10: (85) call 4
11: (79) r1 = *(u64 *)(r10 -8)
12: (15) if r1 == 0x0 goto pc+19
 R0=inv R1=inv R6=ctx R7=imm0 R10=fp
13: (18) r1 = 0x64616572
15: (7b) *(u64 *)(r10 -32) = r1
16: (73) *(u8 *)(r10 -24) = r7
17: (7b) *(u64 *)(r10 -40) = r7
18: (7b) *(u64 *)(r10 -48) = r7
19: (79) r3 = *(u64 *)(r6 +112)
20: (07) r3 += 16
21: (bf) r1 = r10
22: (07) r1 += -48
23: (b7) r2 = 16
24: (85) call 4
25: (79) r1 = *(u64 *)(r10 -40)
26: (79) r3 = *(u64 *)(r1 +40)
[...]

PR #218, opened 2015-09-15
Translate multiple pointer dereference into bpr_probe_read

drzaeus77:
This commit adds support for multiple consecutive and nested pointer
dereference of function arguments that should be converted to
bpf_probe_read. The logic works by marking variables as needing a
probe_read if they come from the register argument, and then applying
this property transitively.

Supported syntax:
```
int trace_entry(struct pt_regs *ctx, struct file *file) {
    struct vfsmount *mnt = file->f_path.mnt;
    struct super_block *k = mnt->mnt_sb;
    const char *name = file->f_path.dentry->d_name.name;
```

Not supported: probe reads from map leaves, probe reads after explicit casts.

Fixes: #188


4ast:
looks great!

Issue #219, opened 2015-09-16
Need a bpf_get_current_comm example


brendangregg:
I don't know if I'm using bpf_get_current_comm incorrectly, or if it doesn't work. My code:

```C
BPF_HASH(commbyreq, struct request *, char *);

int trace_start(struct pt_regs *ctx, struct request *req) {
        char comm[TASK_COMM_LEN];
        if (bpf_get_current_comm(&comm, sizeof(comm))) {
                commbyreq.update(&req, (char **)&comm);
        }
}
```

```Shell
invalid indirect read from stack off -48+0 size 16
```

It's complaining about the bpf_get_current_comm() line.

brendangregg:
Oh, I need:

```C
char comm[TASK_COMM_LEN] = {};
```

brendangregg:
I can bpf_trace_printk() comm, but can I store it as a map value then later fetch it?

brendangregg:
```C
BPF_HASH(commbyreq, struct request *, char[TASK_COMM_LEN]);
```

```Shell
# ./biosnoop_ext4
error: brackets are not allowed here; to declare an array, place the brackets after the identifier
error: brackets are not allowed here; to declare an array, place the brackets after the identifier
<bcc-memory-buffer>:11:1: error: expected ')'
BPF_HASH(commbyreq, struct request *, char[TASK_COMM_LEN]);
^
/usr/share/bcc/include/bcc/helpers.h:56:26: note: expanded from macro 'BPF_HASH'
  BPF_HASHX(__VA_ARGS__, BPF_HASH3, BPF_HASH2, BPF_HASH1)(__VA_ARGS__)
                         ^
<bcc-memory-buffer>:11:1: note: to match this '('
/usr/share/bcc/include/bcc/helpers.h:56:26: note: expanded from macro 'BPF_HASH'
  BPF_HASHX(__VA_ARGS__, BPF_HASH3, BPF_HASH2, BPF_HASH1)(__VA_ARGS__)
[...]
```

drzaeus77:
Can you put char[] inside a struct?

brendangregg:
I got far along...

```C
struct val_t {
        char name[TASK_COMM_LEN];
};

BPF_HASH(commbyreq, struct request *, struct val_t);

int kprobe__blk_account_io_start(struct pt_regs *ctx, struct request *req)
{
        struct val_t val = {};

        if (bpf_get_current_comm(&val.name, sizeof(val.name)) == 0) {
[...]

PR #221, opened 2015-09-16
Fix probe reads on char[] types

drzaeus77:
It is easy enough to wrap the type in a typeof(), otherwise the rewriter
would need to do a deeper parsing of the type information to place it
properly next to the variable name.

Fixes: #219


4ast:
I've learned a new C syntax today :)

Issue #224, opened 2015-09-17
struct assignment multiple errors


brendangregg:
For this work in progress, I'm stashing syscall entry arguments for later reference on return. Hit errors.

```Python
from __future__ import print_function
from bcc import BPF

# load BPF program
b = BPF(text = """
#include <uapi/linux/ptrace.h>

struct args_t {
	const char *filename;
	int flags;
	int mode;
};

int kprobe__sys_open(struct pt_regs *ctx, const char __user *filename,
    int flags, int mode) {
	struct args_t args = {};
	args.filename = filename;
	args.flags = flags;
	args.mode = mode;
	// XXX hash args by pid
	bpf_trace_printk("%s\\n", filename);
	return 0;
};
""")

# header
print("%-18s %s" % ("TIME(s)", "CALL"))

# format output
while 1:
	(task, pid, cpu, flags, ts, msg) = b.trace_fields()
	print("%-18.9f %s" % (ts, msg))
```

```Shell
# ./opensnoop 
<bcc-memory-buffer>:14:105: error: operand of type 'struct args_t' where arithmetic or pointer type is required
        ({ typeof(const char *) _val; memset(&_val, 0, sizeof(_val)); bpf_probe_read(&_val, sizeof(_val), (u64)args + offsetof(NULL TYPE, filename)); _val; }) = ...
                                                                                                               ^~~~
<bcc-memory-buffer>:14:121: error: expected a type
        ({ typeof(const char *) _val; memset(&_val, 0, sizeof(_val)); bpf_probe_read(&_val, sizeof(_val), (u64)args + offsetof(NULL TYPE, filename)); _val; }) = ...
                                                                                                                               ^
include/linux/stddef.h:7:14: note: expanded from macro 'NULL'
#define NULL ((void *)0)
             ^
<bcc-memory-buffer>:14:121: error: expected ')'
include/linux/stddef.h:7:23: note: expanded from macro 'NULL'
#define NULL ((void *)0)
                      ^
<bcc-memory-buffer>:14:121: note: to match this '('
include/linux/stddef.h:7:14: note: expanded from macro 'NULL'
#define NULL ((void *)0)
             ^
<bcc-memory-buffer>:14:153: error: expression is not assignable
        ({ typeof(const char *) _val; memset(&_val, 0, sizeof(_val)); bpf_probe_read(&_val, sizeof(_val), (u64)args + offsetof(NULL TYPE, filename)); _val; }) = ...
[...]

Issue #217, opened 2015-09-15
kprobe function wrapping error


brendangregg:
Having issues wrapping a static function in a kprobe'd function. test.py:

```Python
#!/usr/bin/python

from bcc import BPF

b = BPF(text = '''
#include <uapi/linux/ptrace.h>
#include <linux/blkdev.h>
#include <linux/mount.h>

BPF_HASH(start, u32);
BPF_HASH(filebypid, u32, struct dentry *);

static int trace_process(struct pt_regs *ctx, struct file *file)
{
    u64 ts;
    u32 pid;

    pid = bpf_get_current_pid_tgid();

    // map pid to dentry pointer, for later filename lookups
    if (file != NULL && file->f_path.dentry != NULL){
        struct dentry *dent = file->f_path.dentry;
        filebypid.update(&pid, &dent);
    } else {
        // cleanup now (see return)
        filebypid.delete(&pid);
    }

    // save start timestamp
    ts = bpf_ktime_get_ns();
    start.update(&pid, &ts);
    return 0;
}

int trace_readpages_entry(struct pt_regs *ctx, struct file *file)
{
    return trace_process(ctx, file);
}
''')
b.attach_kprobe(event="ext4_readpages", fn_name="trace_readpages_entry")
```

output:

```Shell
bpf: Permission denied
0: (79) r6 = *(u64 *)(r1 +112)
1: (85) call 14
2: (63) *(u32 *)(r10 -12) = r0
3: (15) if r6 == 0x0 goto pc+12
 R0=inv R6=inv R10=fp
4: (79) r1 = *(u64 *)(r6 +24)
R6 invalid mem access 'inv'

Traceback (most recent call last):
  File "./ext4slower", line 193, in <module>
    b.attach_kprobe(event="ext4_readpages", fn_name="trace_readpages_entry")
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 498, in attach_kprobe
    fn = self.load_func(fn_name, BPF.KPROBE)
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 388, in load_func
    raise Exception("Failed to load BPF program %s" % func_name)
Exception: Failed to load BPF program trace_readpages_entry
```

Note the exception is for trace_readpages_entry(), which is pretty simple.

brendangregg:
Oh, if I remove static from trace_process() it works...

drzaeus77:
Hmmm, that's definitely weird. How about static inline?

[...]

PR #227, opened 2015-09-17
Add support for static helper functions

drzaeus77:
This adds support for static helper functions that can be reused. It is
not necessary to include pt_regs in the helper functions, even though
external pointers may be dereferenced. Arguments in the helpers can also
be reordered.

Fixes: #224 #217

### DIFF
--- a/tests/cc/test_clang.py
+++ b/tests/cc/test_clang.py
@@ -172,5 +172,40 @@ int kprobe__blk_update_request(struct pt_regs *ctx, struct request *req) {
     return 0;
 }""")
 
+    def test_probe_read_helper(self):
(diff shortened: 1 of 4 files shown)


tests/cc/test_clang.py
```diff
@@ -172,5 +172,40 @@ def test_char_array_probe(self):
     return 0;
 }""")
 
+    def test_probe_read_helper(self):
+        b = BPF(text="""
+#include <linux/fs.h>
+static void print_file_name(struct file *file) {
```
4ast: would  be nice to add few more tests for this to cover the rest of possibilities that rewriter already supports.

Issue #193, opened 2015-09-09
helper for per-task variables


brendangregg:
It will likely be common to store data with the current on-CPU task (ie, PID/thread), for example, timestamps, arguments, etc., that are then later retrieved by another event for the same task.

Eg, the vfsreadlat.c example does this to store a timestamp from function entry to return:

```C
struct key_t {
	u32 pid;
};

BPF_HASH(start, struct key_t);

int do_entry(struct pt_regs *ctx)
{
        struct key_t key = {};
        u64 ts, *val;

        key.pid = bpf_get_current_pid_tgid();
        ts = bpf_ktime_get_ns();
        start.update(&key, &ts);
        return 0;
}

int do_return(struct pt_regs *ctx)
{
        struct key_t key = {};
        u64 *tsp, delta;

        key.pid = bpf_get_current_pid_tgid();
        tsp = start.lookup(&key);
[...]
```

Imagine something like:

```C
BPF_TASK_VAR(start, u64);

void do_entry(struct pt_regs *ctx)
{
        start.set(bpf_get_current_pid_tgid());
}

void do_return(struct pt_regs *ctx)
{
        u64 *tsp, delta;

        tsp = start.get();
[...]
```

Or better ("start = bpf_get_current_pid_tgid()", etc).

PR #229, opened 2015-09-18
opensnoop

brendangregg:


brendangregg:
Oh, bpf_get_current_pid_tgid() isn't returning the task ID (per-thread) version of the PID? If not, we need a bpf_get_current_tid(). This will get changed when(/if) #193 is done, as this will be a common operation (saving something by-task) and there could be a simpler way.

brendangregg:
Looks like bpf_get_current_pid_tgid() is returning the thread-ID as the PID, so opensnoop should already be working fine.

Simple multi-threaded test program, with 9 threads:

```Shell
# ps -Lp 15550 
  PID   LWP TTY          TIME CMD
15550 15550 pts/2    00:00:00 threads
15550 15551 pts/2    00:00:00 threads
15550 15552 pts/2    00:00:00 threads
15550 15553 pts/2    00:00:00 threads
15550 15554 pts/2    00:00:00 threads
15550 15555 pts/2    00:00:00 threads
15550 15556 pts/2    00:00:00 threads
15550 15557 pts/2    00:00:00 threads
15550 15558 pts/2    00:00:00 threads
```

So the PID is 15550, and the LWPs (which is really the thread-IDs) are from 15550 to 15558.

8 of these threads open a file called "thread" with its own thread ID appended, in a loop. These files don't exist. opensnoop:

```Shell
# ./opensnoop -x
15553  threads            -1   2 thread2
15558  threads            -1   2 thread7
15554  threads            -1   2 thread3
15552  threads            -1   2 thread1
15551  threads            -1   2 thread0
15556  threads            -1   2 thread5
15557  threads            -1   2 thread6
15555  threads            -1   2 thread4
```

It shows these as coming from the LWPs (thread-IDs), and not the single PID.

brendangregg:
... I should note that I did modify opensnoop to print the pid from bpf_get_current_pid_tgid(), and not just use the pid from BPF.trace_fields().

drzaeus77:
Perfect, thanks for confirming.

4ast:
On 9/18/2015 1:28 AM, Brendan Gregg wrote:
> Looks like bpf_get_current_pid_tgid() is returning the thread-ID as the
> PID, so opensnoop should already be working fine.

pid and tgid are kernel names.

https://www.kernel.org/doc/Documentation/accounting/taskstats.txt

"pid", "tid" and "task" are used interchangeably and refer to the standard
Linux task defined by struct task_struct.  per-pid stats are the same as
per-task stats.

"tgid", "process" and "thread group" are used interchangeably and refer 
to the
tasks that share an mm_struct i.e. the traditional Unix process. Despite the
use of tgid, there is no special treatment for the task that is thread group
leader - a process is deemed alive as long as it has any task belonging 
to it.

brendangregg:
Thanks for the link. That clears the confusion.

Now to dig up some more confusion and roll around in it... So ps(1) on Linux is all lies:

```Shell
# ps -Lp 3263
  PID   LWP TTY          TIME CMD
 3263  3263 pts/4    00:00:00 threads
 3263  3264 pts/4    00:00:00 threads
 3263  3265 pts/4    00:00:00 threads
 3263  3266 pts/4    00:00:00 threads
 3263  3267 pts/4    00:00:00 threads
 3263  3268 pts/4    00:00:00 threads
 3263  3269 pts/4    00:00:00 threads
 3263  3270 pts/4    00:00:00 threads
 3263  3271 pts/4    00:00:00 threads
```

The behavior here is what people probably expect. Except that we know that the "-p PID" is really matching a tgid, and the "PID" column should say "TGID", and the "LWP" column should say "PID". :)

I probably should update my scripts that include a "-p PID" filter to actually match on the tgid, so it follows ps(1) behavior. Or provide some other way to match on the tgid ("-g TGID").

4ast:
user space developers always think of 'pid' as process, but kernel developers think of 'pid' as thread. performance tools probably should follow user space naming, but bpf programs are following the kernel. Inevitable confusion?

Issue #144, opened 2015-08-18
Easily save log2 histograms


brendangregg:
Power-of-2 histograms should be a common use case for analyzing distributions. Here is how they can be populated in C (code largely from https://github.com/torvalds/linux/blob/master/samples/bpf/tracex2_kern.c):

```C
BPF_TABLE("array", int, u64, dist, 64);

static unsigned int log2(unsigned int v)
{
	unsigned int r;
	unsigned int shift;

	r = (v > 0xFFFF) << 4; v >>= r;
	shift = (v > 0xFF) << 3; v >>= shift; r |= shift;
	shift = (v > 0xF) << 2; v >>= shift; r |= shift;
	shift = (v > 0x3) << 1; v >>= shift; r |= shift;
	r |= (v >> 1);
	return r;
}

static unsigned int log2l(unsigned long v)
{
	unsigned int hi = v >> 32;
	if (hi)
		return log2(hi) + 32 + 1;
	else
		return log2(v) + 1;
}

[...]
	int index = log2l(req->__data_len / 1024);
	u64 *leaf = dist.lookup(&index);
	if (leaf) (*leaf)++;
```

Adding a BPF_LOG2_IDX macro for the above two C functions would be a big improvement, eg:

```C
BPF_TABLE("array", int, u64, dist, 64);
[...]
	int index = BPF_LOG2_IDX(req->__data_len / 1024);
	u64 *leaf = dist.lookup(&index);
	if (leaf) (*leaf)++;
```

Even better would be a BPF_HIST macro, and a BPF helper:

```C
BPF_HIST(dist, log2);
[...]
	dist.increment(req->__data_len / 1024);
```

brendangregg:
Something else to think about: support of histograms for each key. Imagine we want to save this log2 histogram per-PID, per-comm, per-device-name, or some combination?

brendangregg:
Could something like this be added in the meantime? ... I assume a final implementation will need bpf_log() functions anyway...

```Shell
# git diff src/cc/export/helpers.h
diff --git a/src/cc/export/helpers.h b/src/cc/export/helpers.h
index 4571166..f0aaf9f 100644
--- a/src/cc/export/helpers.h
+++ b/src/cc/export/helpers.h
@@ -185,6 +185,28 @@ static inline void bpf_store_dword(void *skb, u64 off, u64 val) {
 #define MASK(_n) ((_n) < 64 ? (1ull << (_n)) - 1 : ((u64)-1LL))
[...]

PR #240, opened 2015-09-21
add bpf_log2 functions

brendangregg:
This code is from Alexei's tracex2. :)


brendangregg:
thanks; added to aid #144

PR #241, opened 2015-09-21
biolatency, funclatency, and bpf_log2l usage

brendangregg:


brendangregg:
this is my last PR today. :)

Issue #243, opened 2015-09-23
/usr/lib64 seems to be hardcoded.


Earnestly:
Looking at the `CMakeList.txt` I am having a hard time trying to figure out how to prevent it from using the `lib64/` library install paths by default.

Would you perhaps consider switching to [GNUInstallDirs](http://cmake.org/cmake/help/v3.3/module/GNUInstallDirs.html) instead?

With that change I could simply use `-DCMAKE_INSTALL_LIBDIR{,_FULL}` to define it as `/usr/lib`.

drzaeus77:
Please test and close if successful. Thanks!

Earnestly:
Yup, this solves the issue nicely.  Thank you for your efforts.

PR #244, opened 2015-09-23
Use GNUInstallDirs for LIBDIR variable

drzaeus77:
This should be supported in older versions of cmake, but haven't
explicitly tested those.

Fixes: #243


4ast:
looks like ubuntu buildbot doesn't like it

drzaeus77:
Yes definitely don't merge this. Investigating.

PR #248, opened 2015-09-25
Helper to get open k[ret]probes. Fixes #236

### LINKED ISSUE #236: attach_kprobe/kretprobe return number of attached probes
For information, it could be useful for these functions to return the number of probes actually attached. When using unfamiliar event_re's, I'm not sure if it has matched 0 probes, 1 probe, or how many. Tools where I'm using attach_k[ret]probe, I'd likely print "attached to %d functions..." on initialization.

tuxology:


drzaeus77:
This change needs a rebase before I will be able to merge it, in addition to the inline comment.

PR #255, opened 2015-09-28
Add ability to consume perf events in python

drzaeus77:
This adds the ability to consume perf events in libbpf/python using the
ring buffer. For now, this is the only way to get access to the function
call graph. Only kernel functions are supported.

It does this by introducing a new set of libbpf helper functions that
can open the perf fd, mmap it, and poll over the events as they are
submitted by the kernel. This allow for faster event processing than
trace_printks, but has not been tested.

The functionality is disabled by default, the user can enable it by
passing a non-empty cb parameter into the BPF constructor. That cb
function will be invoked for each event that is read from the buffer.
Buffers are per-fd, so each event is distinct as well as separated from
other processes that may be running simultaneously.

The initial test case uses this functionality to build a histogram of
events keyed by the callchain.

### DIFF
--- a/src/cc/perf_reader.c
+++ b/src/cc/perf_reader.c
@@ -54,7 +54,8 @@ void perf_reader_free(void *ptr) {
   if (ptr) {
     struct perf_reader *reader = ptr;
     munmap(reader->base, reader->page_size * (reader->page_cnt + 1));
-    close(reader->fd);
+    if (reader->fd >= 0)
+      close(reader->fd);
     free(reader->buf);
     free(ptr);
   }
--- a/src/cc/perf_reader.c
+++ b/src/cc/perf_reader.c
@@ -1,0 +1,220 @@
+/*
+ * Copyright (c) 2015 PLUMgrid, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <linux/perf_event.h>
+#include <poll.h>
+#include <stdio.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <unistd.h>
+
+#include "libbpf.h"
+#include "perf_reader.h"
+
+struct perf_reader {
+  perf_reader_cb cb;
+  void *cb_cookie; // to be returned in the cb
+  void *buf; // for keeping segmented data
+  size_t buf_size;
+  void *base;
+  int page_size;
+  int page_cnt;
+  int fd;
+  uint64_t sample_type;
+};
+
+struct perf_reader * perf_reader_new(int fd, int page_cnt, perf_reader_cb cb, void *cb_cookie) {
+  struct perf_reader *reader = calloc(1, sizeof(struct perf_reader));
+  if (!reader)
+    return NULL;
+  reader->cb = cb;
+  reader->cb_cookie = cb_cookie;
+  reader->fd = fd;
+  reader->page_size = getpagesize();
+  reader->page_cnt = page_cnt;
+  return reader;
+}
+
+void perf_reader_free(void *ptr) {
+  if (ptr) {
+    struct perf_reader *reader = ptr;
+    munmap(reader->base, reader->page_size * (reader->page_cnt + 1));
+    close(reader->fd);
+    free(reader->buf);
+    free(ptr);
+  }
+}
+
+int perf_reader_mmap(struct perf_reader *reader, int fd, uint64_t sample_type) {
+  int mmap_size = reader->page_size * (reader->page_cnt + 1);
+
+  if (!reader->cb)
+    return 0;
+
+  reader->base = mmap(NULL, mmap_size, PROT_READ | PROT_WRITE , MAP_SHARED, fd, 0);
+  if (reader->base == MAP_FAILED) {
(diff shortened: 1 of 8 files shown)


src/cc/perf_reader.c
```diff
@@ -0,0 +1,220 @@
...
+
+  if (!reader->cb)
+    return 0;
+
+  reader->base = mmap(NULL, mmap_size, PROT_READ | PROT_WRITE , MAP_SHARED, fd, 0);
+  if (reader->base == MAP_FAILED) {
+    perror("mmap");
+    return -1;
+  }
+  reader->fd = fd;
```
4ast: wasn't it inited as in _new ? line 47

drzaeus77: Initial value is -1, here it will be valid. But, that reminds me that there should be a check in _free.

src/cc/perf_reader.c
```diff
@@ -0,0 +1,220 @@
...
+    ptr += sizeof(cc->nr) + sizeof(*cc->ips) * cc->nr;
+    // size sanity check
+    if (ptr > (uint8_t *)data + size) {
+      fprintf(stderr, "%s: corrupt callchain sample\n", __FUNCTION__);
+      return;
+    }
+    int i;
+    // don't include magic numbers in the call chain
+    for (i = 0; i < cc->nr; ++i) {
+      if (cc->ips[i] == PERF_CONTEXT_USER)
```
4ast: hmm, it's in the trace instead of PERF_CONTEXT_KERNEL ?

src/cc/perf_reader.c
```diff
@@ -0,0 +1,220 @@
...
+
+  // sanity check
+  if (ptr != (uint8_t *)data + size) {
+    fprintf(stderr, "%s: extra data at end of sample\n", __FUNCTION__);
+    return;
+  }
+
+  // call out to the user with the parsed data
+  if (reader->cb)
+    reader->cb(reader->cb_cookie, tk ? tk->common.pid : -1, num_callchain, callchain);
```
4ast: hmm, I thought  perf_sample_trace_common header is always there.

src/cc/perf_reader.c
drzaeus77: The callchain typically looks like:

```
PERF_CONTEXT_KERNEL
 0xffffffff8102b320
  0xffffffff81030780
   ...
    PERF_CONTEXT_USER
     ...
```

I pass along the bits between those two markers, since those are the only bits that will resolve correctly in ksym() function.

src/python/bcc/__init__.py
```diff
@@ -360,6 +367,8 @@ def __init__(self, src_file="", hdr_file="", text=None, debug=0):
                 0x2: print BPF bytecode to stderr
         """
 
+        self._reader_cb_impl = _CB_TYPE(BPF._reader_cb)
```
4ast: and that is enough to call python from c? wow

tests/cc/test_callchain.py
```diff
@@ -0,0 +1,32 @@
...
+    return 1;
+}
+""", cb=cb)
+        start = time.time()
+        while time.time() < start + 1:
+            b.kprobe_poll()
+
+        for k, v in hist.items():
+            syms = [b.ksym(addr) for addr in k]
+            print("%-08d:" % v, syms)
```
4ast: nice. example output would be great adding to commit log too.

src/cc/perf_reader.c
drzaeus77: It has always been present in my experience, I just included the check for safety in case somehow the sample_types is mistakenly disabled.

src/cc/perf_reader.c
```diff
@@ -0,0 +1,221 @@
...
+      fprintf(stderr, "%s: corrupt callchain sample\n", __FUNCTION__);
+      return;
+    }
+    int i;
+    // don't include magic numbers in the call chain
+    for (i = 0; i < cc->nr; ++i) {
+      if (cc->ips[i] == PERF_CONTEXT_USER)
+        break;
+      if (cc->ips[i] >= PERF_CONTEXT_MAX)
+        continue;
```
yonghong-song: What does PERF_CONTEXT_MAX represent here?
How do we deal with module symbols which /proc/kallsyms does not have mapping? Do we still include them in the output call chain?

drzaeus77: It just means the address is a special value for the kernel to interpret, I will have to dig more to understand better than I do now. So far _USER and _KERNEL fall into this category.

If such values are given to ksym(), it will print a wrong symbol name, since it approximates symbols by doing a search in a sorted array. Breaking out of the loop as I do here should give valid kernel symbols for all of the cases I have tried so far.

yonghong-song: I checked the kernel code, PERF_CONTEXT_MAX is not set inside the kernel. So cc->ips[i] >= PERF_CONTEXT_MAX seems common practice to preclude special kernel markings. So the current code is okay.

For the code below,

    // don't include magic numbers in the call chain
    for (i = 0; i < cc->nr; ++i) {
      if (cc->ips[i] == PERF_CONTEXT_USER)
        break;
      if (cc->ips[i] >= PERF_CONTEXT_MAX)
        continue;
      if (!callchain)
        callchain = &cc->ips[i];
      ++num_callchain;
    }

It looks like address from kernel modules will be included in callchain if they are included in kernel return callstack.
Maybe I missed some places to filter this out?

drzaeus77: For me, kernel modules fall in the range 0xffffffffa0000000-0xffffffffa0900000, which will be < PERF_CONTEXT_MAX, therefore it will be reported properly in the call stack.

yonghong-song: That is right. Thanks for explanation. Your kernel module range is probably wrong. It is too small. 
From arch/x86/include/asm/pgtable_64_types.h,

#define MODULES_VADDR    (__START_KERNEL_map + KERNEL_IMAGE_SIZE)
#define MODULES_END      _AC(0xffffffffff000000, UL)
#define MODULES_LEN   (MODULES_END - MODULES_VADDR)

#define __START_KERNEL_map      _AC(0xffffffff80000000, UL)
The range for kernel module + kernel image size is 2GB-16M.

PR #272, opened 2015-10-14
perf_reader : add timeout argument to perf_reader_poll

SunKyu:
add timeout argument to exit from poll after given time.


drzaeus77:
Thanks!

PR #286, opened 2015-11-10
Fixes #285 - build process assumes git use

### LINKED ISSUE #285: Build process assumes git checkout for version (REVISION)
I'm creating a numbered release package for Arch; the cmake/version.cmake file assumes that you're running out of git, and tries to perform git operations that fail. The issue here is that it affects the soname in the results like so (notice how it chops off the first letter too due to the substring extraction):

````
cmake .. -DREVISION=0.1.8 -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_LIBDIR=/usr/lib
...
build/src/cc/cmake_install.cmake:

  foreach(file
      "$ENV{DESTDIR}/usr/lib/libbcc.so.EAD-HASH-NOTFOUND"
      "$ENV{DESTDIR}/usr/lib/libbcc.so.0"
      "$ENV{DESTDIR}/usr/lib/libbcc.so"
      )
````
A minor amount of rearrangement fixes up the problem, pull request to follow.

troyengel:
Packagers will supply the revision on the commandline with the -DREVISION=x.y.z flag, this rearranges the logic to only run the git commands if the REVISION is missing.


drzaeus77:
Looks good, thanks!

PR #288, opened 2015-11-11
Edit README, link path of some example files

drzix:
As some example files were reorganized into subdirecotries, some
pathes in README bring connection on 404. So it have to be new pathes.


drzaeus77:
Wow, huge brain fart on my side. In my head I was about to do this, and wasn't going to allow myself to merge the offending code without doing this update as well. Thanks for fixing it!

PR #295, opened 2015-11-25
Some minor fixes for packaging and testing

lcp:


lcp:
I did it for a few python 3 modules such as builtins and http.server. Besides, the function "input()" is the python 3 input().

drzaeus77:
Can you split the two patches into separate pull requests so the first one (which looks fine) can be merged while we iterate on the 2nd?

I'm still not convinced that the python code has a compatibility problem. The use of http.server should be handled, and the use of "input()" is also handled by the use of "from builtins import input", which I see present in multiple places. For reference, this is my favorite reference for finding ways to keep the code compatible: http://python-future.org/compatible_idioms.html.

What would be most helpful to converge on this would be to share the traceback that you see in these files when the code is run unmodified.

lcp:
OK, "pip2 install future" works for me, and it seems openSUSE didn't package python-future. I'll send another pull request for the first patch and fix the python dependency in openSUSE.

Issue #281, opened 2015-11-06
could not open bpf map: Operation not permitted


asifhasnain:
I am using Ubuntu 14.04 with latest kernel 4.3.0-999-generic (Build Date 4/Nov/2015) as a VM on the virtual box.

Linux pgUbuntu14 4.3.0-999-generic #201511042100 SMP Thu Nov 5 02:02:35 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux

When I try to execute the distributed-bridge code using the root account I receive following error.

plumgrid@pgUbuntu14:/usr/share/bcc/examples/distributed_bridge$ sudo python main.py 
multicast 1 dhcp 0 gretap 0
Launching host 1 of 3
Launching host 2 of 3
/usr/share/bcc/examples/distributed_bridge/tunnel.c:18:1: error: could not open bpf map: Operation not permitted
BPF_TABLE("hash", struct vni_key, struct host, mac2host, 10240);
^
/usr/share/bcc/include/bcc/helpers.h:41:4: note: expanded from macro 'BPF_TABLE'
}; \
   ^
/usr/share/bcc/examples/distributed_bridge/tunnel.c:38:29: error: expected initialized handle for bpf_table
    struct host *src_host = mac2host.lookup_or_init(&vk,
                            ^
/usr/share/bcc/examples/distributed_bridge/tunnel.c:60:27: error: expected initialized handle for bpf_table
  struct host *dst_host = mac2host.lookup(&vk);
                          ^
3 errors generated.
Traceback (most recent call last):
  File "tunnel.py", line 19, in <module>
    b = BPF(src_file="tunnel.c")
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 392, in __init__
    raise Exception("Failed to compile BPF module %s" % src_file)
Exception: Failed to compile BPF module tunnel.c
Launching host 3 of 3
/usr/share/bcc/examples/distributed_bridge/tunnel.c:18:1: error: could not open bpf map: Operation not permitted
BPF_TABLE("hash", struct vni_key, struct host, mac2host, 10240);
^
/usr/share/bcc/include/bcc/helpers.h:41:4: note: expanded from macro 'BPF_TABLE'
}; \
   ^
/usr/share/bcc/examples/distributed_bridge/tunnel.c:38:29: error: expected initialized handle for bpf_table
    struct host *src_host = mac2host.lookup_or_init(&vk,
                            ^
[...]

PR #298, opened 2015-11-30
Automatically bump memlock ulimit

drzaeus77:
Instead of requiring the user to bump the ulimit in their shell before
starting a bcc script, try to setrlimit automatically when a failure
occurs. Since there is no getrusage for memlock limit, unfortunately we
have to brute force setting the limit. For now, just try bpf() once and
then try to set unlimited ulimit, then try bpf() again.

Fixes: #281


4ast:
looks good to me.

PR #301, opened 2015-12-03
Fix breakage from LLVM 3.8 API change

drzaeus77:
There is some difference in how to take a pointer from an iterator. This
code should work for both 3.7 and 3.8 APIs.


4ast:
I did the same diff. Worked for me. Nice that it's transparent on 3.7

PR #306, opened 2016-01-01
add new Arch packages, new TOC and spiffy up the formatting

troyengel:
The official Arch kernel has been released with CONFIG_BPF_SYSCALL=y, I was able to complete my AUR packages and test functionality. hello_world.py works perfectly. :)

```
$ sudo python2 /usr/share/bcc/examples/hello_world.py 
         gkrellm-1801  [002] d..1 43877.766400: : Hello, World!
  NetworkManager-655   [004] d..1 43880.638182: : Hello, World!

$ sudo python3 /usr/share/bcc/examples/hello_world.py 
         gkrellm-1801  [000] d..1 43892.821996: : Hello, World!
         gkrellm-1801  [001] d..1 43897.810282: : Hello, World!
```

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,6 +1,16 @@
-# Kernel requirements
+# Installing BCC
 
-## Requirements
+* [Kernel Configuration](#kernel-configuration)
+* [Packages](#packages)
+  - [Ubuntu](#ubuntu---binary)
+  - [Fedora](#fedora---binary)
+  - [Arch](#arch---aur)
+* [Source](#source)
+  - [Ubuntu](#ubuntu---source)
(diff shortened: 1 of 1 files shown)


drzaeus77:
Awesome work @troyengel, just tried it myself as well.

One note: looking at the PKGBUILD file, it uses its own logic for setting -jN, whereas I think the convention is to set MAKEFLAGS="-jN" in /etc/makepkg.conf.

troyengel:
Ooh, thanks Brenden - you can guess which config I've never touched. :) I just uploaded a new 0.1.7-3 PKGBUILD removing the -jN flags from make. Thanks!

PR #307, opened 2016-01-04
Remove runtime dependency on gcc stdarg headers and make

drzaeus77:
Remove the runtime dependency on gcc's stdarg.h. Do this by packaging
the (non-GPL) version shipped with clang. This allows the bulk of
kbuild_helper to be removed, at the expense of hardcoding the kernel
include paths. If in the future the kernel make system changes again to
require different -I paths, we'll have to think this through again.


4ast:
awesome

PR #308, opened 2016-01-06
Add ability to set custom cflags when loading programs

drzaeus77:
Example:
b = bcc.BPF("myprog.c", cflags=["-DMYCUSTOMFLAG1", ...])

### DIFF
--- a/src/cc/bpf_common.cc
+++ b/src/cc/bpf_common.cc
@@ -26,18 +26,18 @@ void * bpf_module_create_b(const char *filename, const char *proto_filename, uns
   return mod;
 }
 
-void * bpf_module_create_c(const char *filename, unsigned flags) {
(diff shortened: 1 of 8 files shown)


4ast:
I guess that's need for arm which needs -D__ASM_SYSREG_H ?
but why ask user to do it manually?

drzaeus77:
My use case is to define flags at runtime, not necessarily depending on architecture but depending on feature. For instance, when I load a program through the iomodule server, the server may define a constant which is my module ID. I can refer to that id in my program via MODULE_ID or some such, which the server defines when it is loading the program to bcc on my behalf.

Another use case may be a #define which holds the number of CPUs in the system. I can use that to define my own table rather than write architecture specific code that for instance does s/NUMCPU/.../.

4ast:
fair enough, though some of the tools/biolatency is using text.replace() as more generic
way of achieving the same and golang could have done it too.

PR #313, opened 2016-01-16
stack walking scripts

brendangregg:


tools/offcputime
```diff
@@ -0,0 +1,201 @@
...
+    start.delete(&pid);
+    delta = delta / 1000;
+    if (delta < MINBLOCK_US)
+        return 0;
+
+    bpf_get_current_comm(&key.name, sizeof(key.name));
+    bp = ctx->bp;
+
+    // unrolled loop (MAXDEPTH):
+    if (!(key.ret[depth++] = get_frame(&bp))) goto out;
```
4ast: wow. that's a cool hack :)
proper stack support is #1 on my todo list.

brendangregg: thanks! yes, once proper support exists (bpf_ routines) these scripts can be edited.

4ast:
I was afraid that we'd have to wait for kernel stack trace support to get this type of tools working, but you've managed to get it working with unrolled probe_read. Awesome. Starting to play with it...

tools/stackcount
```diff
@@ -0,0 +1,179 @@
...
+static u64 get_frame(u64 *bp) {
+    if (*bp) {
+        // The following stack walker is x86_64 specific
+        u64 ret = 0;
+        if (bpf_probe_read(&ret, sizeof(ret), (void *)(*bp+8)))
+            return 0;
+        if (!ret || ret < __START_KERNEL_map)
+            return 0;
+        if (bpf_probe_read(bp, sizeof(*bp), (void *)*bp))
+            bp = 0;
```
drzaeus77: Just noticed (belatedly) while reading through the blog entry, shouldn't this be `*bp = 0;`?

brendangregg:
Good catch, thanks. Plus I can simplify the test on line 89. Pull request forthcoming...

PR #316, opened 2016-01-20
offcputime improvements

brendangregg:


tools/offcputime
```diff
@@ -82,32 +88,32 @@ static u64 get_frame(u64 *bp) {
...
-int oncpu(struct pt_regs *ctx) {
-    u32 pid = bpf_get_current_pid_tgid();
-    FILTER
-    u64 ts = bpf_ktime_get_ns();
-    struct key_t key = {};
-    u64 zero = 0, *val, bp = 0, *tsp, delta;
-    int depth = 0;
+    // record previous thread sleep time
+    if (FILTER) {
+        pid = prev->pid;
```
4ast: nice. lower overhead.
do you think it makes sense to add key.pid,
since key.comm muxes all threads of given process into one
and for multi threaded apps the final numbers are hard to interpret.
Also '-p pid' is also useless for such apps, since it filters by kernel pid,
but user spaces guys think of tgid when they say pid

brendangregg: It's useful both ways. I'll add key.pid as a command line option. Usually I want just comm first, and then pid(tgid) if need be. One example where I really want comm is profiling software builds, which churn short-lived processes.

Issue #317, opened 2016-01-24
CMake tools - use FILE + GLOB?


troyengel:
I'm working on the bcc-git packaging and noticed that certain parts of the tools/ tree were not propagating in the build; for example the new 'offtimecpu' utilities are missing from the DESTDIR install target. I'm reading commit c175cf8096523949b3e41400180f04313d147308 here:
````
set(FILES biolatency.8 biosnoop.8 funccount.8 funclatency.8 hardirqs.8
  killsnoop.8 opensnoop.8 pidpersec.8 softirqs.8 syncsnoop.8 tcpaccept.8
  tcpconnect.8 vfscount.8 vfsstat.8)
````
(and the other ones just like it for the other parts), and wondered -- is the better solution for us to use the file(GLOB ...) instead as outlined here? (disclaimer: I'm a cmake newbie)

* https://cmake.org/cmake/help/v3.0/command/file.html
````
file(GLOB FILES *.8)
file(GLOB PROGRAMS *.c *.py)
file(GLOB EXAMPLE_FILES *.txt)
list(REMOVE_ITEM EXAMPLE_FILES CMakeLists.txt)
````
...or something along those lines? The PROGRAMS is a bit difficult to glob out without regex since there's a mix of files with no extensions and .c in there. The bigger point being to try and make the CMake logic generic so that every change/add/delete won't require updating it later. Thoughts?

drzaeus77:
Definitely worth looking into. I've been delinquent and working on some of the stuff in iovisor/iomodules, so there's a bit of a backlog in bcc.

The points you raise about PROGRAMS vs FILES is indeed why I kept the explicit list in this directory. There is probably a way to solve it, cmake provides a mechanism to parse lists, maybe I could glob the files with no extensions in some way. Let me look into it.

Besides that, it is past due time to tag 0.1.8, so my plan is to fix this bug one way or another and then tag.

4ast:
Should we rename the tools/xx to tools/xx.py ? then GLOB pattern would be much easier to apply and during install into /usr we can drop .py suffix?
as for 0.1.8 tag... could you keep it 0.1.7 for now. Let's reserve version bump for things like changes in API and probably the next one should be 0.2

troyengel:
Awesome, thanks - I'll just go with the correct logic then like so for the bcc-tools-git split child:
````
  cd "${srcdir}/${pkgbase}/build/tools"
  make DESTDIR="${pkgdir}/" install

[...]

PR #319, opened 2016-01-26
Updates to use cmake GLOB and libbcc.so.0 in python init

drzaeus77:
In order not to miss some files in the tools and examples source
directories, use cmake file(GLOB) to collect relevant files. To ease the
implementation, move all tools to be .py suffixed in the source, but
sans-suffix in the installation (same as before)

In addition, to prevent future API breakage confusion (though of course
that may still happen), use CDLL("libbcc.so.0") in the bcc __init__.py.

Fixes: #317


4ast:
Thanks! LGTM
not sure what's going on with conflicts. rebase needed?

PR #321, opened 2016-01-28
wakeuptime

brendangregg:
follow-on from offcputime. This shows the waker stack, and waker & target process names.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,23 +65,24 @@ Examples:
 
 Tools:
 
-- tools/[biolatency](tools/biolatency): Summarize block device I/O latency as a histogram. [Examples](tools/biolatency_example.txt).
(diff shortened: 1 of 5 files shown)


4ast:
looks great. The wake stack by itself is interesting, but makes the most sense only when combined with off-cpu stack, right? That would be the next step?

brendangregg:
Yes, and I've already prototyped that tool, called offwaketime. I'll push it after some more testing.

Having wakertime separate may be useful for adhoc investigations, where you know you just want waker stacks. It may also be useful for generating a series of flame graphs: imagine having a script that generated a HTML page with a CPU flame graph, followed by an off-CPU flame graph, and then the waker flame graph. They won't be integrated, but it might be a useful option for presenting all this new information.

PR #322, opened 2016-01-28
Uprobe support

drzaeus77:
This adds uprobes support in two commits.

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -765,7 +765,7 @@ class BPF(object):
         if not addr:
             raise Exception("could not determine address of symbol %s" % sym)
 
-        return (path, load_addr+addr)
+        return (path, addr-load_addr)
 
     def attach_uprobe(self, name="", sym="", addr=None,
             fn_name="", pid=-1, cpu=0, group_fd=-1):
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -20,7 +20,9 @@ import fcntl
 import json
 import multiprocessing
 import os
+import re
 from subprocess import Popen, PIPE
+import struct
 import sys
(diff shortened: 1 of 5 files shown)


drzaeus77:
Please leave comments on the attach_uprobe calling convention. Docstring will be my first change.

drzaeus77:
Looks like the buildbots don't support uprobes. Might need to upgrade the kernel.

tests/cc/test_uprobes.py
```diff
@@ -0,0 +1,87 @@
...
+
+import bcc
+import ctypes
+import os
+import re
+import struct
+import time
+import unittest
+
+## code from ctypes impl
```
4ast: these lines are left over?

drzaeus77: Thanks, removed.

drzaeus77:
@4ast this code should be ready now

src/python/bcc/__init__.py
```diff
@@ -670,6 +692,141 @@ def detach_kretprobe(event):
...
+        cls._libsearch_cache[name] = path
+        return path
+
+    @classmethod
+    def find_load_address(cls, path):
+        if path in cls._lib_load_address_cache:
+            return cls._lib_load_address_cache[path]
+
+        # "LOAD off    0x0000000000000000 vaddr 0x0000000000400000 paddr 0x..."
+        with os.popen("""/usr/bin/objdump -x %s | awk '$1 == "LOAD" && $3 ~ /^[0x]*$/ { print $5 }'""" % path) as f:
```
4ast: either full path to objump in both places or both without?

drzaeus77: Fixed in next patch.

4ast:
Nice!

PR #325, opened 2016-01-29
bashreadline

brendangregg:
Snoop bash commands system wide!


4ast:
that's ultimate snoop tool :)

PR #331, opened 2016-01-30
offwaketime

brendangregg:


4ast:
This is awesome!
Is there a way to make the time easier to interpret?
Currently it's "The number beneath the stacks is the total time spent from the blocking event to the wakeup event. This is summed for all occurances with the same stack pairs." but it's hard to translate that to wall-clock nsec.

brendangregg:
The time is the total time. Conceptually, maybe we can explain it as an off-CPU analogue to CPU sampling: imagine we're sampling when threads are off-CPU, and showing the totals in different paths.

Time gets more complex with more waker stacks (chain graphs), and I'm putting more thought into it...

PR #338, opened 2016-02-05
contributing scripts guide

brendangregg:


brendangregg:
We should refer people to it, and if they have reasonable exceptions then that's ok, and can be even better: guidelines + exceptions > guidelines alone.

A different type of exception would be if an engineer who wrote a body of code gives us some bcc tooling of their own code, then I'd be inclined to write documentation on their behalf, if they didn't have the time. Eg, an ext4 engineer gives us some ext4 tools, or a Cassandra engineer gives us some Cassandra tools, etc. Again, guidelines + exceptions > guidelines alone.

PR #340, opened 2016-02-05
Add bpf_module.h to CMakeLists.txt

linuxfood:
The C++ API should also be exported, in addition to the C API.
Because type safety is nice.


4ast:
Thanks! Could you kill an extra 'merge' commit and rebase.

PR #355, opened 2016-02-08
New tool: memleak

goldshtn:
Trace user or kernel allocations and their call stacks to display outstanding allocations that haven't been freed. Should be useful for tracing memory leaks in both user-space programs and kernel modules.


brendangregg:
LGTM

drzaeus77:
Nicely done! I think as a future work, I (or someone) can split out the symbol parsing code into a helper inside the bcc module, with some tests. Before that, however, I would like to refactor BPF a bit into a hierarchy, so stay tuned.

PR #356, opened 2016-02-08
ping reply in ebpf

muhammadzaheer:
ping_reply.py demonstrates checksum calculation and packet header manipulation in an ebpf program using bcc.

This example:
- Hooks at Traffic Control (ingress queue) in Kernel using eBPF
- If packet is an ICMP echo (ping)
    * Intercepts the packet at traffic control ingress queue
    * Modifies ICMP packet type to ICMP reply
    * Swaps src/dst ip and mac and updates checksum
    * Redirects packet to the same interface as egress


muhammadzaheer:
I've included the headers now. Thanks for the feedback.

examples/networking/ping_reply.py
```diff
@@ -0,0 +1,137 @@
...
+        // Packet is available on the 0th location of the memory.
+        u8 *cursor = 0;
+
+        struct ethernet_t * ethernet = cursor_advance(cursor,
+                sizeof(*ethernet));
+        // If packet type is not IP, return
+        if (ethernet->type != ETH_P_IP)
+                return 0;
+
+        struct ip_t * ip = cursor_advance(cursor, sizeof(*ip));
```
drzaeus77: You might need to further advance by '(ip->hlen << 2) - sizeof(*ip)', in the case of optional headers.

examples/networking/ping_reply.py
```diff
@@ -0,0 +1,137 @@
...
+        // If ICMP packet is not echo, return
+        if (icmp->type != ICMP_ECHO)
+                return 0;
+
+        /*
+        Converting ICMP echo into ICMP reply by changing the type to 0
+        Since we're changing packet contents, we need to update the checksum
+        */
+        unsigned short type = ICMP_ECHOREPLY;
+        bpf_l4_csum_replace(skb,36,icmp->type, type,sizeof(type));
```
drzaeus77: It would be better to use incr_cksum_l4 helper function from helpers.h, which invokes some logic inside clang to understands bitfields and network byte order better.

examples/networking/ping_reply.py
```diff
@@ -0,0 +1,137 @@
...
+        bpf_skb_store_bytes(skb, 34, &type, sizeof(type),0);
+
+        /*
+        Swapping Source and Destination in IP header
+        We don't need to update checksum since we're just swapping.
+        However to demonstrate the use of bpf_l3_csum_replace, the checksum
+        is recomputed after each change
+        */
+        unsigned int old_src = bpf_ntohl(ip->src);
```
drzaeus77: If you eliminate bpf_ntohl() and also incorporate incr_cksum_l3/4, the code should shrink overall. Please see 'tests/cc/test_xlate1.c' for uses of the mentioned functions.

examples/networking/ping_reply.py
```diff
@@ -0,0 +1,137 @@
...
+
+        bpf_l3_csum_replace(skb, IP_CKSUM_OFF, old_src, old_dst,
+                sizeof(old_dst));
+
+        /*
+        Demonstrating the use of bpf_skb_store_bytes(...)
+        There are other easier ways to swap ip->src and ip->dst, one of which
+        follows while swapping mac addresses
+        */
+        bpf_skb_store_bytes(skb, IP_SRC_OFF, &old_dst,
```
drzaeus77: It is possible to direct assign to ip->src/ip->dst, example also in tests/cc/test_xlate1.c.

examples/networking/ping_reply.py
```diff
@@ -0,0 +1,137 @@
...
+
+        ethernet->src = old_dst_mac;
+        ethernet->dst = old_src_mac;
+
+        u64 ret = bpf_clone_redirect(skb, skb->ifindex,0 /*For Egress */);
+        /*
+        This output to the kernel trace_pipe which can also be read by:
+        cat /sys/kernel/debug/tracing/trace_pipe
+        */
+        bpf_trace_printk("Replying to ICMP echo: Result=%u\\n", ret);
```
drzaeus77: This printk seems like overkill, a stat seems more appropriate.

examples/networking/ping_reply.py
```diff
@@ -0,0 +1,137 @@
...
+
+        return 1;
+}
+"""
+ipr = IPRoute()
+ipdb = IPDB(nl=ipr)
+ifc = ipdb.interfaces.eth0
+
+b = BPF(text=prog)
+pbr = b.load_func("ping_reply", BPF.SCHED_CLS)
```
drzaeus77: The convention we are going with is to use SCHED_ACT, and from inside the program you should call bpf_redirect and then return TC_ACT_REDIRECT instead of TC_ACT_DROP. This is more efficient in that it completely avoids cloning the packet.

examples/networking/ping_reply.py
```diff
@@ -0,0 +1,137 @@
...
+from pyroute2 import IPRoute, IPDB
+import sys
+
+prog = """
+  #include <bcc/proto.h>
+  #include <uapi/linux/if_ether.h>
+  #include <uapi/linux/in.h>
+  #include <uapi/linux/icmp.h>
+ 
+  #define IP_SRC_OFF 26
```
drzaeus77: The use of static offsets and u8 *cursor at the same time is mixing two different flavors. I would try to use one approach or the other. If you go with cursor (preferred), then you can do something like 'incr_cksum_l3(&ip->hchecksum, orig_icmp_type, new_icmp_type);', which would support dynamic parsing of the ip header in case of different ethertypes.

examples/networking/ping_reply.py
```diff
@@ -0,0 +1,137 @@
...
+        if (icmp->type != ICMP_ECHO)
+                return 0;
+
+        /*
+        Converting ICMP echo into ICMP reply by changing the type to 0
+        Since we're changing packet contents, we need to update the checksum
+        */
+        unsigned short type = ICMP_ECHOREPLY;
+        bpf_l4_csum_replace(skb,36,icmp->type, type,sizeof(type));
+        bpf_skb_store_bytes(skb, 34, &type, sizeof(type),0);
```
drzaeus77: Another case for `icmp->type = type`.

muhammadzaheer:
@drzaeus77 Thanks for the detailed feedback. I've fixed parts of code accordingly. I still can't get incr_cksum_l4 work for icmp header. If I replace bpf_l4_csum replace with incr_cksum_l4, the packet just gets dropped.

drzaeus77:
@muhammadzaheer I'm going to close this PR, since some parts of the review comments have not been addressed and it has been sitting here for some time. If you would like to work on this again, please ask further questions on how to make the corrections and open a new PR.

PR #360, opened 2016-02-10
3 tools: oomkill, dcstat, dcsnoop

brendangregg:


tools/dcsnoop.py
```diff
@@ -0,0 +1,117 @@
...
+    return 1;
+}
+
+int kprobe__d_lookup(struct pt_regs *ctx, const struct dentry *parent,
+    const struct qstr *name)
+{
+    u32 pid = bpf_get_current_pid_tgid();
+    struct entry_t entry = {};
+    if (name->name) {
+        bpf_probe_read(&entry.name, sizeof(entry.name), (void *)name->name);
```
4ast: clang::rewriter currently cannot be optimized, so this name->name and the one above are two different probe_reads that compiler cannot combine. You can try 'char * tmp = name->name; if (tmp)...' to save one call.

tools/oomkill.py
```diff
@@ -0,0 +1,42 @@
...
+loadavg = "/proc/loadavg"
+
+# initialize BPF
+b = BPF(text="""
+#include <uapi/linux/ptrace.h>
+#include <linux/oom.h>
+void kprobe__oom_kill_process(struct pt_regs *ctx, struct oom_control *oc,
+    struct task_struct *p, unsigned int points, unsigned long totalpages)
+{
+    bpf_trace_printk("OOM kill of PID %d (\\"%s\\"), %d pages\\n", p->pid,
```
4ast: trace_printk() might fail in oom... better to use bpf_perf_event_output() which won't be allocating memory.
And in general better to use bpf_perf_event_output() whenever possible to avoid use trace_pipe, since it can be used by some other daemon or another bcc/tools/* script. See examples/tracing/trace_perf_output.py It's also way faster than trace_printk.

man/man8/oomkill.8
```diff
@@ -0,0 +1,55 @@
+.TH oomkill 8  "2016-02-09" "USER COMMANDS"
+.SH NAME
+oomkill \- Trace oom_kill_process(). Uses Linux eBPF/bcc.
+.SH SYNOPSIS
+.B bashreadline
```
goldshtn: I think this should be `oomkill`.

tools/oomkill.py
brendangregg: Ok, what kernel version did that start working in?

```Shell
# ./trace_perf_output.py 
PERF_COUNT_SW_BPF_OUTPUT feature unsupported
```

drzaeus77: 4.4

brendangregg: ok, thanks, I'm on the wrong 4.4. :) I'm going to try Alexei's stackid branch...

brendangregg: Does bpf_perf_event_output() provide any freebies, like how pid and task come with bpf_trace_printk(), or, does everything need to be put in the data struct?

Also, if I'm printing strings, this means the data struct will need char[MAX_LEN] storage for strings. Would the entire char[] array be copied to user-level, or does it only go up to the NULL termination for efficiency? (comparing this to bpf_trace_printk()). It's a minor detail but something I know I'll be wondering about.

drzaeus77: No freebies :(

If you know the length, you could pass a different value other than `events.perf_submit(ctx, &data, sizeof(data))`. But for strings, since you can't write a loop to do strlen for you, I imagine the cases in which you know the proper length are limited?

brendangregg: Thanks, yes, limited. Although there is __builtin_strlen(), although I haven't seen it work yet.

In case you know offhand:

```
struct data_t {
    u64 fpid;
    u64 tpid;
    u64 pages;
    char fcomm[TASK_COMM_LEN];
    char tcomm[TASK_COMM_LEN];
};
[...]
class Data(ct.Structure):
    _fields_ = [
        ("fpid", ct.c_ulonglong),
        ("tpid", ct.c_ulonglong),
        ("pages", ct.c_ulonglong),
        ("fcomm", ct.c_char),
        ("tcomm", ct.c_char)
    ]
```

Should I be passing these in as ct.c_char or ct.c_char_p or something else?

drzaeus77: ct.c_char * 16

brendangregg: thanks, works!

man/man8/oomkill.8
brendangregg: thanks!

tools/dcsnoop.py
brendangregg: Makes sense (I've done a lot of similar caching in DTrace).

PR #362, opened 2016-02-10
Enhancements to memleak.py

goldshtn:
Added a bunch of new switches: -z and -Z for filtering captures by size, -d for controlling stack depth, and -T for printing only the top *N* stacks by size. Also added timestamp output for each message printed, and changed the command-line arguments to accept the interval and count as positional arguments, per the convention of other *stat tools.


tools/memleak.c
```diff
@@ -1,6 +1,11 @@
-#include <uapi/linux/ptrace.h>
+/*
+ * memleak.c    Trace and display outstanding allocations to detect 
+ *              memory leaks in user-mode processes and the kernel.
+ *
+ * Copyright (C) 2016 Sasha Goldshtein.
```
drzaeus77: With which license are you releasing this?

brendangregg:
LGTM!

FWIW: with so many new options, one thing I'm keeping an eye on is if they are are consistent with the tool's primary purpose (which they are). As soon as you add something that feels different, a bit out of place, like another tool, then I'd suggest splitting it out to another tool.

tools/memleak.py
```diff
@@ -206,6 +236,19 @@ def run_command_get_pid(command):
 bpf_source = open("memleak.c").read()
```
drzaeus77: I didn't catch this the first time around, but this will only work if the tool is run from the tools source directory. See the _find_files() function of BPF for a possible helper you could use. We need a utils submodule it seems...

brendangregg: ... or consider inlining. There's a few .c files in /tools from the earlier days, but I'll do an update and merge them into the .py.

tools/memleak.c
goldshtn: Apache. Will fix.

tools/memleak.py
goldshtn: Will inline.

4ast: On Wed, Feb 10, 2016 at 10:26 PM, Brendan Gregg <notifications@github.com>
wrote:

... or consider inlining. There's a few .c files in /tools from the earlier
> days, but I'll do an update and merge them into the .py.
>
> +1
when .c is embedded in python it makes distribution of the tools easier.

PR #369, opened 2016-02-11
argdist.py: Trace a function and its parameters into a histogram or frequency count

goldshtn:
gentrace.py attaches to functions and traces their parameter values into a histogram or a raw data collection. A simple command-line syntax specifies the probe point, the parameter values to capture, and whether a histogram or raw data is desired.

From the examples -- print a histogram of buffer sizes passed to `write()` across all processes, where the file descriptor was 1 (STDOUT):

```
# ./gentrace.py "hist:c:write(int fd, const void *buf, size_t count):size_t:count:fd==1"
[11:25:28]
hist:c:write(int fd, const void *buf, size_t count):size_t:count:fd==1
     count               : count     distribution
         8 -> 15         : 1        |****************************************|
        64 -> 127        : 1        |****************************************|
[11:25:29]
hist:c:write(int fd, const void *buf, size_t count):size_t:count:fd==1
     count               : count     distribution
         8 -> 15         : 2        |********                                |
        32 -> 63         : 1        |****                                    |
        64 -> 127        : 9        |****************************************|
[11:25:30]
hist:c:write(int fd, const void *buf, size_t count):size_t:count:fd==1
     count               : count     distribution
         8 -> 15         : 3        |*******                                 |
        32 -> 63         : 2        |****                                    |
        64 -> 127        : 17       |****************************************|
```


brendangregg:
This is pretty awesome; I'm still reading, but a couple of quick comments:

- this really feels like 2 tools in one: raw and hist. If I were writing it, I'd explore having this as two tools: arghist and argtrace. 
- The "-ret" option. Well, lets say this becomes two tools, I'd do this using "p" for entry and "r" for return, because we already have that convention from kprobes (https://github.com/torvalds/linux/blob/master/Documentation/trace/kprobetrace.txt). I can't say I love "p" and "r", but it's there.
- I don't like tools that write tools (I did this with my errinfo script years ago, which was a pain to maintain). I understand this would be an exception, because of what it does, but I'd put comments in the code to say something like "this is a tool that generates code then executes it; in general, you should try to avoid doing this unless you have a really good reason. Explicit code is easier to debug."
- it should have a debug mode to print the generated code. Could even be an option ("-v" for verbose).

goldshtn:
Thanks for the quick feedback! Totally agree about the last two points. Will add. 

Re two tools: I've given this some thought initially. There are two reasons why I wanted it in the same tool:

1. With the current infra we have for building scripts, there would be no way to share the 80%+ of common code between the tools. Unless we are ok with adding Python modules to the tools/ directory. 

2. It is actually useful to combine a probe that collects a histogram with a probe that collects raw data. E.g., off the top of my head -- `malloc` retprobe that captures raw addresses and `malloc` entry probe that captures a histogram of allocated sizes... Or a probe that collects `write()` buflen histogram and a retprobe that collects `write()` return codes to detect errors. This could be accomplished by using two tools in two separate windows, but it just feels clumsy to me. 

What do you think?

brendangregg:
- Why wouldn't -P work for kernel probes? I know a lot of stuff is async, but a lot of kernel stuff is in PID context (VFS, sockets, ...), so I'd use it there too.
- See how the -i interval stuff starts to feel weird with a tracer? Imagine these synopses:

argtrace [-p PID] [-z STRING_SIZE] {p,r}:[library]:function(signature)[:type:expr[:filter]]
arghist [-p PID] [-z STRING_SIZE] {p,r}:[library]:function(signature)[:type:expr[:filter]] [interval [count]]

Then this:

```
 ./gentrace.py -i 3 -c 2 "hist::__kmalloc(size_t size):size_t:size"
```

Becomes:

```
./arghist "p::__kmalloc(size_t size):size_t:size" 3 2
```

It has crossed my mind to spit these further into separate user & kernel versions. This may be something we can do later on if it makes sense. I'm thinking of what happens when we have tracepoint support, and USDT (user tracepoint) support, and how we'd add those in. Or, maybe the static tracepoint versions should be the ones split out, and leave arghist/argtrace as-is, for dynamic tracing. (In which case, maybe they should be called darghist/dargtrace? names are hard).

goldshtn:
Makes sense. I will add support for -p in kernel probes. 

Re -i and -c, it was a deliberate decision because I want to support multiple probes in one command line invocation. I don't see how it would work with interval and count being positional arguments. I mean, it could work if I circumvent argparser, but I rather wouldn't ...

brendangregg:
Regarding 1. I wouldn't have a separate module, I'd have just duplicated the code. It's a 200 line program; maintaining both in lockstep shouldn't be hard, and if it's a better end-user experience it's worth the small inconvenience by us. Although in the longer term, it may be a better experience for us: the individual programs should get a bit shorter, and easier to modify and test (smaller range of tests).

I haven't tried, but imagine the tool is arghist, and argtrace is a symlink. The tool itself checks how it was invoked, and acts accordingly. Maybe that's a compromise.

2. Hmm. Certainly possible, I'd have to see good use-cases. Usually on an ad hoc investigation, I'm doing summaries and summaries (hist), and then when narrowed it down, I switch to tracing.

brendangregg:
... ah, yes, I'm not sure how to get argparser to be sane, so can be -i for now. We can always fix later on if it's too annoying.

brendangregg:
BTW, if it stays as gentrace, something that may help address breaking the "do one thing and do it well" convention would be to pitch it as something that clearly does a lot more, like supertrace, or arganalyze. The other tools we have are nice and short; we haven't used the word "analyze" yet, so maybe that would help distinguish it and mentally prepare the user for something different.

brendangregg:
arganalyzer

goldshtn:
Well then, maybe there are really three tools :-P

arghist for histograms, arggroup for simple by-key grouping (value 14 seen 173 times etc.), and argtrace for direct tracing a la perf (except it doesn't go through a file). The third one I haven't implemented yet but it's the easiest one. 

Are you convinced it won't be useful to combine multiple probe types in the same invocation? My intuition says it would be, but you have more experience to back it up.

brendangregg:
Ah, sorry, I'd misread the examples! (Although if I can, so can other people ;). I thought you had a per-event dump in there, and "raw" seemed to corroborate my assumption (meaning raw, un-summarized).

So what it really does right now is an arghist and argcount. I generally use "count" as shorthand for frequency counting. Then there could be argtrace, for raw per-event output.

I'm less concerned about mixing arghist and argcount, since they are both doing summaries. You could merge them and call it something ... not arganalyzer (that would imply everything, incl. tracing)... trying to think of something shorter than argsummary.

goldshtn:
Cool. I'll try to clarify the examples.

As for the trace tool, I'm not sure how I would dump the output. I'd need to call some form of printk or perf_event_output and pass the user-defined expression to it. The type of that expression can be anything. Should I try to figure out the right format specifier based on the type? Feels very ugly to me.

brendangregg:
If arghist & argcount stay as one tool, usage could be:

argsum [-H] [-p PID] [-z STRING_SIZE] {p,r}:[library]:function(signature)[:type:expr[:filter]]

-H for histograms, otherwise frequency count by default (I'd stop calling it raw data collection - that's what perf does with perf.data :)

p & r for probe-entry and probe-return

goldshtn:
OK - so you don't see a scenario where you would want more than one probe on the same command line? Or to mix one hist probe and one count probe?

brendangregg:
up to you, but you can not worry about tracing for now. It's really a different use case than summaries, and this summary tool is awesome, so we should get it in.

If you do start thinking about tracing, then here's a starting point:

https://github.com/brendangregg/perf-tools/blob/master/examples/kprobe_example.txt
https://github.com/brendangregg/perf-tools/blob/master/examples/uprobe_example.txt

Versions of these with better argument support would be good. I haven't tried much bcc on a system without kernel debuginfo -- and those existing kprobe and uprobe tools work without debuginfo, registers only -- so it would be good to keep that working as a fallback.

brendangregg:
more than one summary, and mixing summaries, sure. All part of a high level investigation. Will lead to some pretty powerful one-liners. Again, sorry for the earlier confusion, but I thought you wanted to mix per-event trace dumps with summaries. :)

goldshtn:
But if we are doing mixed hist and count, we can't have a single -H switch and just p/r specifiers. We would still need something like {hist,count}:{p,r}...

goldshtn:
Or maybe:

-H "p:whatever..." -C "r:whoever..."

Instead of just positional arguments.

brendangregg:
hrm... good point...

I think frequency count will be the common case, and we try to not have a switch for the common case (eg, not doing "cat -i file").

Here's two acutal one-liners I think we'd use:

```
# reads and writes by file descriptor:
argsum 'p:c:read(int fd):int:fd' 'p:c:write(int fd):int:fd'
# read and write returned size histogram:
argsum -H 'r:c:read():int:@retval' -H 'r:c:write():int:$retval'
```

- PS the kprobe/uprobe convention is '$retval', not '@retval'

goldshtn:
Makes perfect sense. Re @retval, I know it's the convention, but it has the weakness of being expanded in weak quotes; if a user ends up doing "...:$retval" it will not have the desired effect. If you don't think it's a big deal, I'll go back to $retval.

brendangregg:
I'd go with $retval, and " ' "s in all examples. It'll be like using awk and the command line.

Maybe it should be "argdist" (not argsum). Showing frequency and histogram distributions? I've used "dist" for frequency count tools before (fddist).

goldshtn:
What do you think about argstat? I also like argdist. Argsum, not so much.

brendangregg:
No, I've been using *stat for per-second column summaries. Convention from vmstat, iostat, mpstat, etc. argdist is good. Short for argument distributions. Which is short for function argument distributions, supporting frequency and histogram distributions.

brendangregg:
PS, if you don't like -H, I just remembered that I did something similar with softirq/hardirq, where normally it prints summary totals but a -d will split these into distributions. I don't actually mind either way, H for histograms or d for distributions.

goldshtn:
Argh. It's actually a problem with argparse to do multiple arguments with -H and multiple arguments without. For example:

```
argdist.py -H 'p:c:malloc()' 'p:c:free()'
```

Might seem like it's doing a histogram of malloc and a count of free, but it's in fact doing a histogram of both. OTOH:

```
argdist.py 'p:c:free()' -H 'p:c:malloc()'
```

This grabs a frequency count of free and a histogram of malloc.

I think this is rather confusing, so I'm going to go with -C (count) and -H (histogram) for now. Maybe in the future we'll find a more flexible argparser...

brendangregg:
ok, fair enough, we can deal with it later if it's a pain. Could also use -c and -d for now, if not -C and -H.

goldshtn:
I think I'm done with all the work we talked about earlier. Check out the updated examples and man page. I want to add a couple of additional features later, but I think this PR is ready for review.

/cc @brendangregg

drzaeus77:
LGTM.

brendangregg:
Thanks Sasha, LGTM!

brendangregg:
... I think you have some blog posts to write now. ;) I copy and paste from my _example files to save time.

goldshtn:
Absolutely! It's a challenge though, bcc is such fun I just want to keep going and going :-)

brendangregg:
Understandable; maybe we should coordinate a little; we're usually on irc.freenode.net #iovisor (although can take some hours to respond).

brendangregg:
sorry that's #iovisor at irc.oftc.net

PR #373, opened 2016-02-13
Enhancements to argdist.py

goldshtn:
Added -v switch to print the resulting eBPF program, -T switch to print only top N values from a frequency count collection, and added `$entry(paramname)` and `$latency` keywords for tracing in return probes.

Print histogram of nanoseconds per byte allocated:
```
argdist.py -H 'r::__kmalloc(size_t size):u64:$latency/$entry(size);ns per byte'
```

Print histogram of vfs read sizes that were slower than 0.1ms (inspired by vfsslower.py):
```
argdist.py -H 'r::__vfs_read(void *file, void *buf, size_t count):size_t:$entry(count):$latency > 1000000'
```

P.S. I think at this point there are many one-liners from SystemTap (or DTrace) that can be converted to argdist, although obviously this probe language is not as expressive. From my perspective, the next step -- and I'm not going to start working on it without discussing it with you -- is to think about further ways of making eBPF programs easier to write for users whose primary skill isn't development. I particularly don't like the SystemTap syntax, I don't particularly hate it either, but it's much easier than the current combination of C and Python that we have.

### DIFF
--- a/tools/argdist.py
+++ b/tools/argdist.py
@@ -74,22 +74,30 @@ int PROBENAME(struct pt_regs *ctx SIGNATURE)
                 text = text.replace("PID_FILTER", pid_filter)
                 collect = ""
                 for pname in self.args_to_probe:
-                        collect += "%s.update(&pid, &%s);\n" % \
-                                   (self.hashname_prefix + pname, pname)
+                        param_hash = self.hashname_prefix + pname
+                        if pname == "__latency":
+                                collect += """
+u64 __time = bpf_ktime_get_ns();
+%s.update(&pid, &__time);
+"""                             % param_hash
+                        else:
+                                collect += "%s.update(&pid, &%s);\n" % \
+                                           (param_hash, pname)
                 text = text.replace("COLLECT", collect)
                 return text
 
         def _generate_entry_probe(self):
-                # TODO $latency as a special keyword that should be traced
                 # Any $entry(name) expressions result in saving that argument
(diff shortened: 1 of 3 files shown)


brendangregg:
Useful!

Yes, please shoot me an email (bgregg@netflix.com), and I'll also get a conversation started on github about a higher level interface.

tools/argdist.py
```diff
@@ -200,28 +372,42 @@ def display(self):
...
 
-argdist.py -C 'r:c:gets():char*:$retval;snooped strings'
+argdist.py -C 'r:c:gets():char*:$retval#snooped strings'
         Snoop on all strings returned by gets()
 
-argdist.py -p 1005 -C 'p:c:write(int fd):int:fd'
+argdist.py -H 'r::__kmalloc(size_t size):u64:$latency/$entry(size)#ns per byte'
+        Print a histogram of nanoseconds per byte from kmalloc allocations
+
+argdist.py -I 'linux/slab.h' \\
```
4ast: This swiss army knife looks great, but -I flag is too hard for the average user who won't know which header defines gfp_t. We can either include a lot of headers by default or parse the command line for known symbols and add .h automatically. I guess we can keep -I for obscure cases of driver debugging, but things mentioned in the example file (like gfp_t and timespec) should be automatic.

goldshtn:
Closing this one because I've added a couple of more enhancements, including tuple support. Opening a new PR for this.

PR #377, opened 2016-02-14
Enhancements to argdist

goldshtn:
Added -v switch to print the resulting eBPF program, -T switch to print only top N values from a frequency count collection, and added `$entry(paramname)` and `$latency` keywords for tracing in return probes. Also added tuples support, so multiple parameters can be traced. (This replaces the closed PR #373.)

Print histogram of nanoseconds per byte allocated:
```
argdist.py -H 'r::__kmalloc(size_t size):u64:$latency/$entry(size);ns per byte'
```

Print histogram of vfs read sizes that were slower than 0.1ms (inspired by fileslower.py):
```
argdist.py -H 'r::__vfs_read(void *file, void *buf, size_t count):size_t:$entry(count):$latency > 1000000'
```

Print reads from files in __vfs_read grouped by file name and number of bytes:
```
./argdist.py -I 'linux/fs.h' -C 'p::__vfs_read(struct file *file, void *buf, size_t count):char*,size_t:file->f_path.dentry->d_iname,count:file->f_path.dentry->d_iname[0]!=0'
```

### DIFF
--- a/man/man8/argdist.8
+++ b/man/man8/argdist.8
@@ -49,7 +49,7 @@ include in the BPF program, e.g. 'linux/blkdev.h' or 'linux/time.h'.
 .SH SPECIFIER SYNTAX
 The general specifier syntax is as follows:
 
-.B {p,r}:[library]:function(signature)[:type:expr[:filter]][#label]
+.B {p,r}:[library]:function(signature)[:type[,type...]:expr[,expr...][:filter]][#label]
 .TP
 .B {p,r}
 Probe type \- "p" for function entry, "r" for function return;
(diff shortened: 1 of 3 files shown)


goldshtn:
@4ast 
Regarding your comment on -I, I agree that it should be as automatic as possible. I'm not sure how easy it would be to parse the header files and figure out which one is required for each type declaration. In fact, I'm pretty sure it would be rather hard. I'm willing to try, but it probably requires a separate PR, so I'd still appreciate if you could merge this first.

4ast:
@goldshtn I didn't meant that we need to parse .h that would add too much startup cost. We can grep through command line of the tool. If *alloc* is found, add slab.h, if *time* is found, add time.h,
if *fs* or *file* is found, add fs.h, ...

goldshtn:
But that would make like a million little rules, don't you think? Unless you think it's 20-30 of those, in which case I suppose it's reasonable. Do you have a specific list in mind except what you listed above?

goldshtn:
In the last commit, I've added automatic includes for a handful of keywords.

4ast:
LGTM

PR #383, opened 2016-02-15
3 tools: tcpretrans, zfsslower, zfsdist

brendangregg:


brendangregg:
I don't know why that fedora build failed...

4ast:
zfs on linux? some out of tree kernel module? and bcc works fine with it. progress :)

brendangregg:
Yes, bcc worked fine with zfs on linux, although zfs didn't work so fine with linux 4.5 (I had to hack some things to make it compile). I should try btrfs next.

Issue #333, opened 2016-02-04
avoid dependency on share/bcc/include/


4ast:
Installing helpers.h, proto.h and stdarg.h into default location
makes distribution and versioning of libbcc very difficult.
libbcc should embed them as text string as part of the library
and only depend on the presence of kernel headers in /lib/modules/
high-pri since it's a major road block for production deployment.

PR #396, opened 2016-02-18
Embed runtime header files in libbcc.so

drzaeus77:
To avoid installing header files needed by clang to disk
(/usr/share/bcc), embed the files as strings inside the library and feed
them to clang as memory buffers. The mechanism that we use to do this
retains all of the existing features, as in one can still do `#include
<bcc/helpers.h>`, even though it is redundant, and clang will pick up
the embedded file.

Fixes: #333

### DIFF
--- a/src/cc/frontends/clang/loader.cc
+++ b/src/cc/frontends/clang/loader.cc
@@ -47,6 +47,7 @@
 
 #include "common.h"
 #include "exception.h"
+#include "exported_files.h"
 #include "kbuild_helper.h"
 #include "b_frontend_action.h"
 #include "loader.h"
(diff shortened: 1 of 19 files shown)


src/cc/frontends/clang/loader.cc
```diff
@@ -156,11 +169,17 @@ int ClangLoader::parse(unique_ptr<llvm::Module> *mod, unique_ptr<vector<TableDes
                                           const_cast<const char **>(ccargs.data()) + ccargs.size(), diags))
     return -1;
 
+  // This option instructs clang whether or not to free the file buffers that we
+  // give to it. Since the embedded header files should be copied fewer times
+  // and reused if possible, set this flag to true.
+  invocation1->getPreprocessorOpts().RetainRemappedFileBuffers = true;
+  for (const auto &f : remapped_files_)
+    invocation1->getPreprocessorOpts().addRemappedFile(f.first, &*f.second);
```
4ast: awesome!

PR #402, opened 2016-02-19
Migrated filelife to bpf_perf_event

mcaleavya:
Looks ok with testing. Not sure how to test the -p option at the moment but looks a straight forward update.


brendangregg:
Thanks

PR #403, opened 2016-02-19
syncsnoop migrated to use bpf_perf_events

mcaleavya:
I have added in a strcpy to copy in the Sync() text, I am thinking it might be best to just print this from python. Thoughts?

Al


brendangregg:
Yeah, I'd just print that from Python. Missed a " " too. ;)

mcaleavya:
I forgot to remove the string.h include, doing so now and will push.

brendangregg:
Sorry, I'd also combine a couple of lines:

```
data.ts = bpf_ktime_get_ns() / 1000;
````

Rest is good.

mcaleavya:
done , going through checks now

brendangregg:
thanks

PR #404, opened 2016-02-19
Migrated killsnoop to bpf_perf_event

mcaleavya:


brendangregg:
It's good, but adding infotmp should really also include tpid and sig now, and then delete args_pid and args_sig -- we won't need them, as it'll all be in infotmp.

mcaleavya:
Done

brendangregg:
thanks

PR #405, opened 2016-02-20
tcpconnlat

brendangregg:


mcaleavya:
Typo : We can trace this for handshack completion. (handshake)

drzaeus77:
I prefer handshack. Can you change it back?

PR #407, opened 2016-02-21
Documentation and tcp4connect enhancements

yadutaf:
I started to play around with kprobes this week-end. Here are some enhancements from my local notes. Hope it can help other people getting started!

- on Ubuntu Wily: LLVM dependencies are in the default repositories, no need for extra ones
- ignore ``/build`` build directory. It is manually created but it is also the recommended setup, so makes sense to ignore it
- add a reference to the FAQ.txt which looks like a good resource for early troubleshoot :)

Regarding ``tcpv4connect.py``, it may not be the best thing to do, I'm just starting to experiment with this API. I'm using Ubuntu Wily stock kernel (``Linux jt-laptop 4.2.0-22-generic #27-Ubuntu SMP Thu Dec 17 22:57:08 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux``). Under this kernel I observed that all probes seems to share a single output pipe so that competing probe may cause other to fail parsing their output.

A possible workaround is to tag messages and filter based on the tag. I wonder if there is a more efficient way to do it?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
+# Editor's files
 *.swp
 *.swo
 *.pyc
+
+# Build artefacts
+/build/
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -112,12 +112,15 @@ To build the toolchain from source, one needs:
 
 ### Install build dependencies
 ```
+# Trusty and older
 VER=trusty
 echo "deb http://llvm.org/apt/$VER/ llvm-toolchain-$VER-3.7 main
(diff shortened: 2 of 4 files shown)


drzaeus77:
Looks good @yadutaf ! The deficiency you noticed is known, which is why you'll see things like #404 which are migrating away from trace_pipe. However, this functionality is only available in kernel 4.5 and above, so we expect to have some transition period. Which means, the solution you present in this patch is unfortunately still needed. Thanks.

PR #415, opened 2016-02-23
funccount: use array map instead of hash

4ast:
hash map is not safe when used in recursive context.
If kprobes are placed in the middle of memory allocator functions,
the recursive call of bpf_prog->map_update->kmalloc may crash.
Example: funccount '*spin*' may dead lock the cpu.
Instead use array map as a hash table.
After the fix it works fine:
$ sudo ./funccount.py '*spin*'
Tracing 33 functions for "*spin*"... Hit Ctrl-C to end.
^C
ADDR             FUNC                          COUNT
ffffffff818dc121 _raw_spin_unlock_bh             779
ffffffff818dc461 _raw_spin_lock_bh              1359
ffffffff810990a1 mutex_spin_on_owner           11510
ffffffff818dc3f1 _raw_spin_trylock             14687
ffffffff810990f1 mutex_optimistic_spin         28618
ffffffff818dc2a1 _raw_spin_lock_irq            57933
ffffffff818dc211 _raw_spin_lock_irqsave       337914
ffffffff818dc0a1 __lock_text_start            360507
ffffffff818dc431 _raw_spin_lock               729764
Detaching...

### DIFF
--- a/tools/funccount.py
+++ b/tools/funccount.py
@@ -57,18 +57,25 @@ def signal_ignore(signal, frame):
 bpf_text = """
 #include <uapi/linux/ptrace.h>
 
-struct key_t {
+#define MAP_SIZE 10000
+
+struct val_t {
     u64 ip;
+    u64 value;
 };
-BPF_HASH(counts, struct key_t);
+BPF_TABLE("array", int, struct val_t, counts, MAP_SIZE);
 
 int trace_count(struct pt_regs *ctx) {
     FILTER
-    struct key_t key = {};
-    u64 zero = 0, *val;
-    key.ip = ctx->ip;
-    val = counts.lookup_or_init(&key, &zero);
-    (*val)++;
+    struct val_t *val;
+    u64 zero = 0;
+    u32 key = ctx->ip % MAP_SIZE;
+    val = counts.lookup(&key);
+    if (!val || (val->ip != 0 && val->ip != ctx->ip))
+        return 0;
+    if (val->ip == 0)
+            val->ip = ctx->ip;
+    val->value++;
     return 0;
 }
 """
@@ -108,7 +115,8 @@ while (1):
     print("%-16s %-26s %8s" % ("ADDR", "FUNC", "COUNT"))
     counts = b.get_table("counts")
     for k, v in sorted(counts.items(), key=lambda counts: counts[1].value):
-        print("%-16x %-26s %8d" % (k.ip, b.ksym(k.ip), v.value))
+        if v.ip:
+            print("%-16x %-26s %8d" % (v.ip, b.ksym(v.ip), v.value))
     counts.clear()
 
     if exiting:


goldshtn:
AFAICT this may be an issue with other tools that use a hash, such as argdist, memleak, etc. 

Am I right or is there something specific about this tool in particular? If not, perhaps it's a good idea to do a code review and open issues where necessary so we can fix them...

drzaeus77:
I imagine this also deserves some protection on the kernel side. Is that in the works?

4ast:
memleak is fine as-is. argdist needs a workaround.
The fix on the kernel side is going to be intrusive and won't be backported.
If anyone has an idea how to fix it cheaply, I'm all ears.

goldshtn:
I don't know enough to say it with confidence, but isn't it just a reentrancy check before entering the probe? Unless there are issues that don't involve reentrancy, or you want to preserve the option for reentrancy. 

Also I'm a bit worried about how to fix this class of issues. Treating arbitrary values as hash table keys is very convenient. Converting to arrays requires coming up with something better than % SIZE, because the input isn't necessarily uniformly distributed...

4ast:
there is already reentrancy checks at kprobe, ftrace and bpf_prog levels.
that's why memleak works fine even though it adds kprobe to kfree and call kfree from the program.

goldshtn:
Then what exactly is the issue?

4ast:
What's happening:
some bit calls kfree() which goes into slow path of slub and grabs kmem_cache_node->lock, does some work and calls _raw_spin_unlock_irqrestore() which trigger kprobe and bpf program which
is doing bpf_map_update_elem()->kmalloc() and unlucky hitting of slow path again which deadlocks on spin_lock for the same kmem_cache_node->lock.

brendangregg:
LGTM.

I might adjust it later and add an overflow counter, so if you are unlucky to trace something that hashes to the same key, at least it gives you a count of what it missed, along with a suggestion "trace fewer events".

I was also already thinking we should limit matched probes, because startup/teardown times are too long. eg, funccount says "too many matches" if you try to match > 10k. Can do this as well.

I don't know of a cheap kernel fix offhand, other than writing your own version of kfree... Is it possible to to mark as deathrow instead of kfree, then have another thread do the kfree's (a thread you control, rather than letting this get triggered in probe context).

tools/funccount.py
```diff
@@ -57,18 +57,25 @@ def signal_ignore(signal, frame):
...
 int trace_count(struct pt_regs *ctx) {
     FILTER
-    struct key_t key = {};
-    u64 zero = 0, *val;
-    key.ip = ctx->ip;
-    val = counts.lookup_or_init(&key, &zero);
-    (*val)++;
+    struct val_t *val;
+    u64 zero = 0;
+    u32 key = ctx->ip % MAP_SIZE;
```
goldshtn: I don't know if modulus map size is the best "hash" value to use. I don't think function start addresses are uniformly distributed. Aren't they even aligned to some power of two?

4ast:
yep, 10k kprobes don't work either. we gotta limit regex. attach_kprobe() should never try to do more than 1000 or so. @drzaeus77 

was thinking about kmalloc/kfree in a separate thread as well.
The possible fixes so far:
1. pre-allocate all hash map elements, but that is the waste for majority of use cases
where we can kmalloc/kfree from the program as usual
2. kmem_cache_create for every map would add too much noise to /proc/slabinfo
3. add recursion check to slow-path of slub and return NULL from kmalloc
4. remember irqs_disabled_flags() before calling into bpf prog and do preventive actions
  in bpf_map_update_elem() like using reserved memory or simply failing to update -
  sort of ok, but there is no cross-arch irqs_disabled_flags() that operates on pt_regs,
  since pt_regs is the only thing kprobe is seeing. The simple in_irq() or irqs_disabled()
  won't work, since they're always return true (kprobes run with irq off)
5. same but using preempt_count() as indicator with extra hacks to undo pc increments
  done by kprobe logic to compute original preempt count - too hacky and too fragile
  irq off and/or non-atomic is not an indicator that kmalloc cannot be called.
6. separate thread for alloc/free

Implemented 3,4,5 already and discarded, since they have other corner cases.
1 is not applicable in all cases.
6 is hard as well, since bpf_map_update() cannot sleep, so not clear how to
request an allocation from other thread.

tools/funccount.py
4ast: probably not, but we're not trying to optimize hash distribution here. It's a workaround for crash.
Doing extra math here will only make it slower.

drzaeus77: Does the % divisor need to be power of 2 (I know the array size does)? Maybe choose a prime number, or does bpf inst

brendangregg: @goldshtn right, so this update drops events if there was a collision:

```
    if (!val || (val->ip != 0 && val->ip != ctx->ip))
        return 0;
```

If, instead, it:

```
    if (!val || (val->ip != 0 && val->ip != ctx->ip))
        if (val->ip != ctx->ip)
            collisions++;
        return 0;
```
And then told the end-user the collisions counter if it was non-zero, then this solution is workable. If you see collisions, then match fewer functions. If there were no collisions, then no problem. This will probably have to do until the kernel supports recursive context hashs.

brendangregg:
@4ast ok, need something like a match_kprobe() that tells us the matched count, before the actual attach_kprobe(), so we can exit if it's over the 1k threshold.

goldshtn:
The other thread could try to keep a cache of ready-made chunks for the probe thread to grab with an atomic op. If it's exhausted the probe thread could spin a bit and then fail the allocation? But as long as the other thread keeps up with a reasonable rate it should be replenished soon enough?

tools/funccount.py
goldshtn: @brendangregg As long as your key is the instruction pointer and nothing else, you could simply precompute the array indices -- even from the Python program. You basically don't need the hashing at all.

4ast: @drzaeus77 the % can be any number. Internal hash implementation is using power of 2 because integer divide is too slow. So %10000 here has better distribution than hash with default=10240 which gets rounded to power of 2.

4ast: @brendangregg to do 'collisions++' we'd need another array of one element to keep track of this.
I'm working on more native support of globals. Underneath they will be using arrays, but interface in C will be nicer.

drzaeus77: Ok, anyway I like @goldshtn suggestion for pre-indexing the functions. That can be done either via a read-only map or with python str.replace(), but doesn't really help the generic case.

4ast:
@goldshtn how another thread will know that main thread exhausted the pool? we can do queue_work(), but quick map inserts will anyway exhaust the pool, so inserts will be failing for some time and then suddenly start to succeed ? that's not really usable. Note wait is not possible, since we may be in interrupt.

tools/funccount.py
4ast: what do you guys mean by pre-indexing? have different program with hard coded index for every counted function?

drzaeus77: Or have a read-only (from kernel point of view) map that gives pc to idx
translation.

goldshtn: Yes, or a series of hardcoded if statements comparing the ip to predefined values and generating the index. For a small number of elements it would probably be as fast as a modulus op... 
Or we could do a read only array of ip to index.

4ast: imo a set of 'if' statements or extra ip->index hash map or different program for every function is complete overkill.

goldshtn: But isn't allocating a 10K element array when you might only have two functions you're tracing also a bit overkill? :-)

4ast: yep, it is, but it's a difference of typing 5 characters vs whole new set of things.
the default of BPF_HASH is 10240, suboptimal, but ok.

drzaeus77: Overkill is subjective. Is funccount.py meant to be fast, accurate, or safe? Seems that current implementation can't have all 3, so we need to choose. What is the purpose for funccount.py, and which of those 3 is the most appropriate tradeoff?

4ast: safe. everything else is secondary.

goldshtn: What I'm worried about is that in case of collision, the user really doesn't have any choice but to trace fewer and fewer functions. And if two functions out of 50 end up colliding, it's not going to be very pleasant figuring out which.

goldshtn:
Suppose a few inserts fail and then start working again... I assume for some programs that would be acceptable actually. So maybe you can do two modes of hash allocation - preallocated when you need assured success, and on-demand allocated when you can cope with occasional failures?

brendangregg:
For funccount, if it ends up limited for a while, we can refer people to the ftrace version https://github.com/brendangregg/perf-tools/blob/master/kernel/funccount

```
# /apps/perf-tools/bin/funccount '*spin*'
Tracing "*spin*"... Ctrl-C to end.
^C
FUNC                              COUNT
mutex_optimistic_spin                 3
mutex_spin_on_owner.isra.3            3
rwsem_spin_on_owner                  27
__pv_queued_spin_lock_slowpath      122
_raw_spin_unlock_bh                 255
_raw_spin_lock_bh                   303
_raw_spin_trylock                  1201
_raw_spin_lock_irq                 3680
_raw_spin_lock_irqsave            14457
_raw_spin_unlock_irqrestore       15869
_raw_spin_lock                    87916

Ending tracing...
#
```

Although, perhaps a different way to fix funccount now is to do something like this (similar to @goldshtn 's idea), pseudocode:

```
for func in matches:
    i++;
    idx2name[i] = func;
    bpf_text .= "count_$i { bpfmap[$i]++ }";
    b.attach_kprobe(func, count_$i);
```

ie, adding a separate function per probe. Would that break any BPF limits? We'd be declaring up to 1000 functions. ... Then on print, lookup index in idx2name.

brendangregg:
More testing: it's just the counts.clear() that triggers the issue. It's not that funccount is exiting. This also hits it, after printing the first interval:

```
# ./funccount.py -i 1 '*spin*'
```

brendangregg:
This also seems to fix the issue (thanks Brenden):

```diff
diff --git a/tools/funccount.py b/tools/funccount.py
index f417087..749d4b7 100755
--- a/tools/funccount.py
+++ b/tools/funccount.py
@@ -108,9 +108,13 @@ while (1):
     print("%-16s %-26s %8s" % ("ADDR", "FUNC", "COUNT"))
     counts = b.get_table("counts")
     for k, v in sorted(counts.items(), key=lambda counts: counts[1].value):
+        if v.value == 0:
+            continue
         print("%-16x %-26s %8d" % (k.ip, b.ksym(k.ip), v.value))
-    counts.clear()
 
     if exiting:
         print("Detaching...")
         exit()
+    else:
+        # zero hash
+        for k in counts.keys(): counts[k] = counts.Leaf()
```

drzaeus77:
Ok, not sure which approach will win out, @4ast please tell me what you think between the two. I was going for accuracy of the counts (well, still need to support per-cpu hash) without adding any extra tables or 'overkill'. :)

4ast:
superseded by https://github.com/iovisor/bcc/pull/417

PR #416, opened 2016-02-23
Enforce limit of 1000 open [uk]probes

drzaeus77:
The number of open probes has bitten users already, so lets enforce a
limit with a meaningful(ish) message.


brendangregg:
This is a good fallback.

I'd also want tools (like funccount) to test the matched count and produce a better custom message (and not a Python traceback). Could we also get a BPF function to test the match count? (doesn't have to be this PR).

PR #417, opened 2016-02-23
Remove use of update() in funccount

drzaeus77:
This is an alternative fix for #415, funccount being called in a recursive context.

### DIFF
--- a/src/python/bcc/table.py
+++ b/src/python/bcc/table.py
@@ -186,6 +186,10 @@ class TableBase(MutableMapping):
         for k in self.keys():
             self.__delitem__(k)
 
+    def zero(self):
+        for k in self.keys():
+            self[k] = self.Leaf()
+
     def __iter__(self):
         return TableBase.Iter(self, self.Key)
 
--- a/tools/funccount.py
+++ b/tools/funccount.py
@@ -65,9 +65,13 @@ BPF_HASH(counts, struct key_t);
 int trace_count(struct pt_regs *ctx) {
     FILTER
     struct key_t key = {};
-    u64 zero = 0, *val;
-    key.ip = ctx->ip;
-    val = counts.lookup_or_init(&key, &zero);
+    u64 *val;
+    // the kprobe pc is slightly after the function starting address, align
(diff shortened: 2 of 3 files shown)


src/python/bcc/table.py
```diff
@@ -186,6 +186,10 @@ def clear(self):
         for k in self.keys():
             self.__delitem__(k)
 
+    def zero(self):
+        for k in self.keys():
+            self[k] = self.Leaf()
```
4ast: I don't think this part actually clear it in a hash map.
Isn't zero-ing it in python memory only?

drzaeus77: No, everything implemented in this file is write-through.

drzaeus77: To be specific, this will invoke __setitem__(key, value)

4ast: yeah found it. so it will call update_elem() which will do kmalloc in user context which is ok.

4ast: hmm then delete should be ok as well.

tools/funccount.py
```diff
@@ -81,6 +85,16 @@ def signal_ignore(signal, frame):
...
 b = BPF(text=bpf_text)
+counts = b.get_table("counts")
+
+# pre-insert the function addresses into the counts table
+fns = b._get_kprobe_functions(pattern)
+for fn in fns:
+    addr = b.ksymname(fn)
+    if addr == -1:
+        raise Exception("Unknown symbol name %s" % fn)
+    counts[counts.Key(addr)] = counts.Leaf()
```
4ast: without checking for collisions this approach is no better than converting array.
can you add collision check here?

drzaeus77: Collision at which level? Two ksyms can have the same address (SyS_write and sys_write), but the address itself will have a unique entry in the bpf hlist underlying the map, correct? Are you suggesting to use BPF_NOEXIST? I don't think that would matter here.

4ast: got it. yeah should work.

goldshtn:
How would you propose to deal with this issue in argdist? Unlike here, the keys are actually arbitrary structures and aren't known in advance...

PR #418, opened 2016-02-24
tools map

brendangregg:
Add a tools map, and also make "Tools" a small heading, so that we have the hyperlink https://github.com/iovisor/bcc#tools


4ast:
awesome

PR #426, opened 2016-03-03
Move cc/*py tests to python subdir and introduce c standalone test binary

drzaeus77:
Moves the python tests to the python subdir. Adds a libbcc.a target for those statically inclined.


src/cc/CMakeLists.txt
```diff
@@ -32,8 +32,13 @@ if (CMAKE_COMPILER_IS_GNUCC)
   endif()
 endif()
 
-add_library(bcc SHARED bpf_common.cc bpf_module.cc libbpf.c perf_reader.c shared_table.cc exported_files.cc)
-set_target_properties(bcc PROPERTIES VERSION ${REVISION_LAST} SOVERSION 0)
+add_library(bcc-shared SHARED bpf_common.cc bpf_module.cc libbpf.c perf_reader.c shared_table.cc exported_files.cc)
+set_target_properties(bcc-shared PROPERTIES VERSION ${REVISION_LAST} SOVERSION 0)
+set_target_properties(bcc-shared PROPERTIES OUTPUT_NAME bcc)
+
+add_library(bcc-loader-static libbpf.c perf_reader.c)
```
4ast: why split it into another lib?

tests/cc/CMakeLists.txt
```diff
@@ -1,50 +1,9 @@
...
-add_test(NAME py_test_histogram WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-  COMMAND ${TEST_WRAPPER} py_histogram sudo ${CMAKE_CURRENT_SOURCE_DIR}/test_histogram.py)
-add_test(NAME py_test_callchain WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-  COMMAND ${TEST_WRAPPER} py_callchain sudo ${CMAKE_CURRENT_SOURCE_DIR}/test_callchain.py)
-add_test(NAME py_array WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-  COMMAND ${TEST_WRAPPER} py_array sudo ${CMAKE_CURRENT_SOURCE_DIR}/test_array.py)
-add_test(NAME py_uprobes WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-  COMMAND ${TEST_WRAPPER} py_uprobes sudo ${CMAKE_CURRENT_SOURCE_DIR}/test_uprobes.py)
+add_executable(test_static test_static.c)
+target_link_libraries(test_static bcc-static)
```
4ast: that only works for .c that are part of the bcc build unfortunately.

src/cc/CMakeLists.txt
drzaeus77: There's a dependency (bpf_map_create) required by both bcc and clang_frontend. This gets resolved at final link-time.

tests/cc/CMakeLists.txt
drzaeus77: Of course, the next step will be to include libbcc.a in the list of installed targets. Some packaging and further testing required there. I would plan to do that in a followup independent change.

drzaeus77:
Buildbots failed, that's fixed now.

PR #427, opened 2016-03-04
Fix stacksnoop PID filter

markdrayton:
The function FILTER is injected into (trace_stack()) returns void, not int.


brendangregg:
thanks

brendangregg:
wait, this function is a void, and has many other "return void"s.

drzaeus77:
@brendangregg his change makes sense to me, what's your confusion?

brendangregg:
oh, sorry, read the diff the wrong way around. Yes, LGTM.

PR #428, opened 2016-03-04
[RFC] Basic implementation of stacktrace table

drzaeus77:
This is the initial implementation of bpf_get_stackid with python
helpers. This doesn't expose the different methods of calling
(kernel/user, fast, or reuse).

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -352,16 +352,20 @@ bool BTypeVisitor::VisitCallExpr(CallExpr *Call) {
                                                                      Call->getArg(2)->getLocEnd()));
           txt = "bpf_perf_event_output(" + arg0 + ", bpf_pseudo_fd(1, " + fd + ")";
           txt += ", bpf_get_smp_processor_id(), " + args_other + ")";
+        } else if (memb_name == "get_stackid") {
+            if (table_it->type == BPF_MAP_TYPE_STACK_TRACE) {
+              txt = "bpf_get_stackid(";
+              txt += "bpf_pseudo_fd(1, " + fd + "), " + args + ")";
+            } else {
+              unsigned diag_id = C.getDiagnostics().getCustomDiagID(DiagnosticsEngine::Error,
+                                                                    "get_stackid only available on stacktrace maps");
+              C.getDiagnostics().Report(Call->getLocStart(), diag_id);
+              return false;
+            }
         } else {
           if (memb_name == "lookup") {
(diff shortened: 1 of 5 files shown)


src/cc/frontends/clang/b_frontend_action.cc
```diff
@@ -354,8 +354,14 @@ bool BTypeVisitor::VisitCallExpr(CallExpr *Call) {
           txt += ", bpf_get_smp_processor_id(), " + args_other + ")";
         } else {
           if (memb_name == "lookup") {
```
4ast: why override .lookup ?
programs can do lookup on stackmap and it will return a set of ip addresses.
I was planning to use 'bounded loop' to do a search in there.
@brendangregg mentioned that it would be nice to filter based on ip address or range of addresses.
so that's what stackmap.lookup() from bpf program is for.

drzaeus77: Ok, I was recalling the discussion about what operations on the table would be allowed, and I must have remembered wrong. I will add a new op (name suggestions?). Any thoughts on how to expose the flags?

4ast: smap.get_stackid(ctx, flags) ?

drzaeus77:
API updated, buildbots will fail since they're running 4.5 kernel.

tests/python/test_stackid.py
```diff
@@ -0,0 +1,39 @@
...
+class TestStackid(unittest.TestCase):
+    def test_simple(self):
+        b = bcc.BPF(text="""
+#include <uapi/linux/ptrace.h>
+#include <linux/bpf.h>
+BPF_STACK_TRACE(stack_traces, 10240);
+BPF_HASH(stack_entries, int, int);
+BPF_HASH(stub);
+int kprobe__htab_map_delete_elem(struct pt_regs *ctx, struct bpf_map *map, u64 *k) {
+    int id = stack_traces.get_stackid(ctx, (BPF_F_REUSE_STACKID));
```
4ast: extra () were not really needed, right?

4ast:
so you're going to update the build bots, so it can merged or ...?

tests/python/test_stackid.py
drzaeus77: It was for now, am fighting the clang rewriter...since BPF_F_REUSE_STACKID is a macro it doesn't like to replace it at that particular boundary.

drzaeus77:
I don't want to build custom kernels for the buildbots, I'd rather mark it an optional test and update them later. I'll send a v3 with some cmake flags.

drzaeus77:
Ok, done.

Issue #380, opened 2016-02-15
add inet_ntoa functions for IP addresses to strings


brendangregg:
inet_ntoa should be in bcc, so we can call it via b.inet_ntoa(). Code:

```Python
def inet_ntoa(addr):
    dq = ''
    for i in range(0, 4):
        dq = dq + str(addr & 0xff)
        if (i != 3):
            dq = dq + '.'
        addr = addr >> 8
    return dq
```

We need another function for IPv6, inet_ntoa6(), following the standard for IPv6 address representation as strings https://tools.ietf.org/html/rfc5952 .

This will shorten the tcp* tools.

brendangregg:
Closing; we'll use the socket library instead. See #582 and:

```Python
from socket import inet_ntop, AF_INET, AF_INET6 
[...]
inet_ntop(AF_INET, pack("I", event.saddr)
[...]
inet_ntop(AF_INET6, event.saddr)
```

PR #429, opened 2016-03-04
add tcpv4listen.py tracing example, with comments

yadutaf:
This is yet-another-example(tm). It was originally written as a toy to experiment with bcc. The goal was to see how easily I could move from a per-namespace poller to a kernel event streaming for new TCP listener, in any namespace / container.

It includes a cleaned version of the notes I took while testing. I believe this could be valuable to new comers (like me) and avoid some pitfalls. Hence the example section.

Real world use case includes
- dynamic load balancer re-configuration when the application is *ready*.
- monitor all listens, even short lived sockets.

UPDATE:

As discussed below, the PR is being rewritten as a tool. Pending tasks:

- [X] Migrate to perf events
- [X] TCP v4
- [X] TCP v6
- [X] TCP backlog
- [X] netns
- [x] Documentation
- [x] Command line filters (protocol / ip version / pid / container)
- [x] Move to /tools and rename to "solisten"

Deferred:
- [ ] UDP v4
- [ ] UDP v6

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,7 @@ Examples:
 - tools/[pidpersec](tools/pidpersec.py): Count new processes (via fork). [Examples](tools/pidpersec_example.txt).
 - tools/[runqlat](tools/runqlat.py): Run queue (scheduler) latency as a histogram. [Examples](tools/runqlat_example.txt).
 - tools/[softirqs](tools/softirqs.py):  Measure soft IRQ (soft interrupt) event time. [Examples](tools/softirqs_example.txt).
+- tools/[solisten](tools/solisten.py): Trace TCP socket listen. [Examples](tools/solisten_example.txt).
 - tools/[stackcount](tools/stackcount.py): Count kernel function calls and their stack traces. [Examples](tools/stackcount_example.txt).
 - tools/[stacksnoop](tools/stacksnoop.py): Trace a kernel function and print all kernel stack traces. [Examples](tools/stacksnoop_example.txt).
 - tools/[statsnoop](tools/statsnoop.py): Trace stat() syscalls. [Examples](tools/statsnoop_example.txt).
--- a/tools/solisten.py
+++ b/tools/solisten.py
@@ -1,0 +1,217 @@
+#!/usr/bin/env python
+#
+# solisten	Trace TCP listen events
+#		For Linux, uses BCC, eBPF. Embedded C.
+#
+# USAGE: solisten.py [-h] [-p PID] [--show-netns]
+#
+# This is provided as a basic example of TCP connection & socket tracing.
+# It could be usefull in scenarios where load balancers needs to be updated
+# dynamically as application is fully initialized.
+#
+# All IPv4 listen attempts are traced, even if they ultimately fail or the
+# the listening program is not willing to accept().
+#
+# Copyright (c) 2016 Jean-Tiare Le Bigot.
+# Licensed under the Apache License, Version 2.0 (the "License")
+#
+# 04-Mar-2016	Jean-Tiare Le Bigot	Created this.
+
+import os
+import socket
+import netaddr
+import argparse
+from bcc import BPF
+import ctypes as ct
+
+# Arguments
+examples = """Examples:
+    ./solisten.py              # Stream socket listen
+    ./solisten.py -p 1234      # Stream socket listen for specified PID only
+    ./solisten.py --netns 4242 # Stream socket listen for specified network namespace ID only
+    ./solisten.py --show-netns # Show network namespace ID. Probably usefull if you run containers
+"""
+
+parser = argparse.ArgumentParser(
+    description="Stream sockets listen",
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    epilog=examples)
+parser.add_argument("--show-netns", action="store_true",
+    help="show network namespace")
+parser.add_argument("-p", "--pid", default=0, type=int,
+    help="trace this PID only")
+parser.add_argument("-n", "--netns", default=0, type=int,
+    help="trace this Network Namespace only")
+
+
+# BPF Program
+bpf_text = """ 
+#include <net/sock.h>
+#include <net/inet_sock.h>
+#include <net/net_namespace.h>
+#include <bcc/proto.h>
+
+// Endian conversion. We can't use kernel version here as it uses inline
+// assembly, neither libc version as we can't import it here. Adapted from both.
+#if defined(__LITTLE_ENDIAN)
+#define bcc_be32_to_cpu(x) ((u32)(__builtin_bswap32)((x)))
+#define bcc_be64_to_cpu(x) ((u64)(__builtin_bswap64)((x)))
+#elif defined(__BIG_ENDIAN)
+#define bcc_be32_to_cpu(x) (x)
+#define bcc_be64_to_cpu(x) (x)
+#else
+#error Host endianness not defined
+#endif
+
+// Common structure for UDP/TCP IPv4/IPv6
+struct listen_evt_t {
+    u64 ts_us;
+    u64 pid_tgid;
+    u64 backlog;
+    u64 netns;
+    u64 proto;    // familiy << 16 | type
+    u64 lport;    // use only 16 bits
+    u64 laddr[2]; // IPv4: store in laddr[0]
+    char task[TASK_COMM_LEN];
+};
+BPF_PERF_OUTPUT(listen_evt);
+
+// Send an event for each IPv4 listen with PID, bound address and port
+int kprobe__inet_listen(struct pt_regs *ctx, struct socket *sock, int backlog)
+{
+        // cast types. Intermediate cast not needed, kept for readability
+        struct sock *sk = sock->sk;
+        struct inet_sock *inet = inet_sk(sk);
+
+        // Built event for userland
+        struct listen_evt_t evt = {
+            .ts_us = bpf_ktime_get_ns() / 1000,
+            .backlog = backlog,
+        };
+
+        // Get process comm. Needs LLVM >= 3.7.1 see https://github.com/iovisor/bcc/issues/393
+        bpf_get_current_comm(evt.task, TASK_COMM_LEN);
+
+        // Get socket IP family
+        u16 family = sk->__sk_common.skc_family;
+        evt.proto = family << 16 | SOCK_STREAM;
+
+        // Get PID
+        evt.pid_tgid = bpf_get_current_pid_tgid();
+
+        ##FILTER_PID##
+
+        // Get port
+        bpf_probe_read(&evt.lport, sizeof(u16), &(inet->inet_sport));
+        evt.lport = ntohs(evt.lport);
+
+        // Get network namespace id, if kernel supports it
+#ifdef CONFIG_NET_NS
+        evt.netns = sk->__sk_common.skc_net.net->ns.inum;
+#else
+        evt.netns = 0;
+#endif
+
+        ##FILTER_NETNS##
+
+        // Get IP
+        if (family == AF_INET) {
+            bpf_probe_read(evt.laddr, sizeof(u32), &(inet->inet_rcv_saddr));
+            evt.laddr[0] = bcc_be32_to_cpu(evt.laddr[0]);
+        } else if (family == AF_INET6) {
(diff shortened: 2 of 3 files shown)


brendangregg:
I like most of this. I don't like pulling event data from /proc/<pid>/ns/net. I did that in my perf-tools only because of limitations with ftrace; in bcc we can walk data structures and fetch things directly. A big problem with my perf-tools tcpretrans that reads from /proc/net/tcp is that it can lag behind the kernel event. I've had many emails about tcpretrans printing zero values, because the /proc/net/tcp session information was missing by the time it was read. Which doesn't make a lot of sense -- the session has gone missing in-between the retransmit being sent (when these sessions are usually in ESTABLISHED) and then user-space trying to read /proc/net/tcp. You'd think I'd get a TIME_WAIT grace, but no...

So I'm not sure we should have another /proc reading hack as an /example for others to follow. There may be a time where we use the /proc stuff in a /tool because there's no other sensible way to do it.

Incidentally, the namespace might be available from socket->sk->sk_socket->file->f_cred->user_ns (which may be a pain to fully fetch).

We also really should do #380 so that we can more easily make these into tools.

brendangregg:
BTW, this could be turned into an solisten tool (short for socket listen; along with a man page, and more columns from the listen call, including backlog). I was going to say tcplisten, but can inet_listen handle other stuff other than TCP?

yadutaf:
I love the perspective of turning toy notes into a real tool. I'll also investigate adding "net_cls" and "net_prio" cgroup informations. It would make it very easy to derive any potential container.

I've tried to access "current" task from eBPF with no success. Is it possible? Would probably make it easier to get the ``current``  task ? ``current->nsproxy->net_ns->ns->inum``

AFAIK, ``inet_listen`` can handle more than TCP, although I don't know these protocols. I'd rather try an equivalent for UDP and hide it behind a switch to protect perf as I expect it to be for each packet.

This said, I'm doing this on my free time, so may take a while ;)

4ast:
accessing 'current' is not possible yet. It's on todo list.

yadutaf:
Turns out reading the namespace name from the kernel was much easier than expected. That's the good news.

The not-so-good-news is ``bpf_trace_printk`` accepts at most 3 arguments. It does not accept the "%pI6" formater which works by reference and has (unsurprisingly) no 128 bits hex formatter so I can't print the full IPv6 address.

Spliting between 2 calls to ``bpf_trace_printk`` would work but it confuses the reader (2 subsequent reads) and is racy, so that it would need some sort of synchronisation or counter that would be atomically incremented on each run (!)

Running out of ideas, I'll sleep over this one.

4ast:
bpf_trace_printk() should be used for debugging only. bpf_perf_event_output() is a better interface and it can send any raw data. Then user space can pretty print ipv6 addresses.

yadutaf:
OK. I may come up with a hack for older versions afterward as we're not going to see 4.5 kernel in LTS distributions anytime soon...

drzaeus77:
perf_event_output is in 4.4, so LTS's should see it.

yadutaf:
I've updated the PR to use ``perf_event_output`` and support IPv6. Can you review the event part ?

examples/tracing/tcpv4listen_example.txt
```diff
@@ -0,0 +1,25 @@
+Demonstrations of tcpv4listen.py, the Linux eBPF/bcc version.
+
+
+This example traces the kernel function called when a program wants to listen
+for TCP IPv4 connections. It will not see UDP neither IPv6 but may server as
```
4ast: obsolete comment? It seems ipv6 should work.

examples/tracing/tcpv4listen.py
```diff
@@ -0,0 +1,174 @@
...
+def inet6_ntoa(addr):
+    s = "%x%x" % (addr[1], addr[0])
+    d = ['']*8
+    for i in range(8):
+        b = s[30-8*i-2 : 30-8*i] + s[30-8*i-4 : 30-8*i-2]
+        d[i] = b
+
+    return ":".join(d)
+
+import socket
```
drzaeus77: This should go at the top of the file.

examples/tracing/tcpv4listen.py
```diff
@@ -0,0 +1,174 @@
...
+        ("proto", ct.c_ulonglong),
+        ("lport", ct.c_ulonglong),
+        ("laddr", ct.c_ulonglong * 2),
+        ("task", ct.c_char * TASK_COMM_LEN)
+    ]
+
+    # TODO: properties to unpack protocol / ip / pid / tgid ...
+
+# Convert big endian IPv4 address into a.b.c.d
+def inet_ntoa(addr):
```
drzaeus77: Have you tried the netaddr package?

yadutaf: That will work. If I find a way to deal with the endian. Either in C either in python. The tunnel examples uses bpf_skb_get_tunnel_key which does the conversion under the hood using be32_to_cpu. BUT when I want to use it with something like:

```C
u32 laddr = 0;
bpf_probe_read(&laddr, sizeof(laddr), &(inet->inet_rcv_saddr));
evt.laddr[0] = be32_to_cpu(laddr);
``` 

I get:

```
error: couldn't allocate output register for constraint 'r' at line 667935
``` 

With netaddr, the best I could find so far:

```python
'.'.join(map(str, reversed(netaddr.IPAddress('0x0100007f').words)))
```

Or I could use "struct" to deal with the endian instead of ctypes.

This said, the IPv6 version is completely broken. I'll fix it.

4ast:
instead of be32_to_cpu use __builtin_bswap32
be32_to_cpu is using x86 inline asm, which isn't going to work.

examples/tracing/tcpv4listen.py
drzaeus77: Tell me what you think of this:

```patch
diff --git a/examples/tracing/tcpv4listen.py b/examples/tracing/tcpv4listen.py
index 8233b6f..19a004b 100644
--- a/examples/tracing/tcpv4listen.py
+++ b/examples/tracing/tcpv4listen.py
@@ -20,6 +20,9 @@
 import os
 from bcc import BPF
 import ctypes as ct
+import netaddr
+import socket
+from struct import unpack
 
 # Status:
 # - [TODO] UDP v4
@@ -46,7 +49,7 @@ struct listen_evt_t {
     u64 netns;
     u64 proto;    // familiy << 16 | type
     u64 lport;    // use only 16 bits
-    u64 laddr[2]; // IPv4: store in laddr[0]
+    u8 rawaddr[16]; // IPv4 store in rawaddr[:4]
     char task[TASK_COMM_LEN];
 };
 BPF_PERF_OUTPUT(listen_evt);
@@ -87,9 +90,9 @@ int kprobe__inet_listen(struct pt_regs *ctx, struct socket *sock, int backlog)
 
         // Get IP
         if (family == AF_INET) {
-            bpf_probe_read(evt.laddr, sizeof(u32), &(inet->inet_rcv_saddr));
+            bpf_probe_read(evt.rawaddr, sizeof(u32), &(inet->inet_rcv_saddr));
         } else if (family == AF_INET6) {
-            bpf_probe_read(evt.laddr, sizeof(evt.laddr), sk->__sk_common.skc_v6_rcv_saddr.in6_u.u6_addr32);
+            bpf_probe_read(evt.rawaddr, sizeof(evt.rawaddr), sk->__sk_common.skc_v6_rcv_saddr.in6_u.u6_addr32);
         }
 
         // Send event to userland
@@ -109,34 +112,12 @@ class ListenEvt(ct.Structure):
         ("netns", ct.c_ulonglong),
         ("proto", ct.c_ulonglong),
         ("lport", ct.c_ulonglong),
-        ("laddr", ct.c_ulonglong * 2),
+        ("rawaddr", ct.c_ubyte * 16),
         ("task", ct.c_char * TASK_COMM_LEN)
     ]
 
     # TODO: properties to unpack protocol / ip / pid / tgid ...
 
-# Convert big endian IPv4 address into a.b.c.d
-def inet_ntoa(addr):
-        dq = ''
-        for i in range(0, 4):
-                dq = dq + str(addr & 0xff)
-                if (i != 3):
-                        dq = dq + '.'
-                addr = addr >> 8
-        return dq
-
-# Convert big endian IPv6 address into string
-def inet6_ntoa(addr):
-    s = "%x%x" % (addr[1], addr[0])
-    d = ['']*8
-    for i in range(8):
-        b = s[30-8*i-2 : 30-8*i] + s[30-8*i-4 : 30-8*i-2]
-        d[i] = b
-
-    return ":".join(d)
-
-import socket
-
 # Format output
 def print_event(cpu, data, size):
     # Decode event
@@ -156,10 +137,11 @@ def print_event(cpu, data, size):
     address = ""
     if proto_type == socket.AF_INET:
         protocol += "v4"
-        address = inet_ntoa(event.laddr[0])
+        rawaddr = unpack(">I", bytes(event.rawaddr[:4]))[0]
     elif proto_type == socket.AF_INET6:
-        address = inet6_ntoa(event.laddr)
+        rawaddr = ":".join(["%x"%x for x in unpack(">HHHHHHHH", event.rawaddr)])
         protocol += "v6"
+    address = netaddr.IPAddress(rawaddr).format()
 
     # Display
     print("%-6d %-12.12s %-12s %-6s %-8s %-39s %-4s" % (
```

drzaeus77:
If this supports v6 now too, perhaps the filename should drop the "v4" part?

examples/tracing/tcpv4listen.py
```diff
@@ -0,0 +1,178 @@
+#!/usr/bin/env python
```
drzaeus77: I think these should be /usr/bin/python instead of /usr/bin/env, it will reduce the PATH searching and speed up startup time. Or at least that is what @brendangregg argued when he critiqued my use of env in an old patch :)

yadutaf: ``/usr/bin/env`` makes the script virtualenv proof. I'm not sure the extra PATH lookup is much of an issue given Python and Clang and ... overhead anyway ;)

yadutaf:
@drzaeus77 Sure. I added a pending tasks list in the initial comment. Thanks for the ``struct`` example.
@4ast Thanks! I didn't know this one.

brendangregg:
Still think this should be called solisten, as it's tracing at the socket layer, not TCP.

I'm still not sure if this is an example or a tool. If there isn't a strong use case for the namespace column, but we wanted to show how it was done as a neat example, then it's an /example. If it's for /tools, then I'd explain why/when NETNS gets used in the docs, and I'd also add a column for backlog (since it's another trivial argument to listen). If it's tricky to explain the NETNS column but you want to keep it in the tool, then drop it from the default output and only show it with a "-v" mode. The default output should be "ah-hah, I get it", not "what the heck is that?", whenever possible (not always possible).

Did inet6_ntoa() get added to bcc and I missed it? Needs an example in the example file for this tool.

yadutaf:
@brendangregg aggree on the rename. I'll add use case for the netns. Basically, this is interesting when you want to update a load balancer in a container based environment as '0.0.0.0' does not mean the same depending on the container. It would be more intuitive for a user if we could catch the cgroup name. Unfortunately, it does not seem reachable without access to ``current``. But I may miss something. Ok to hide it behind a '--netns' or '--verbose'.

inet6_ntoa was a hack based on string manipulation to hide the endian conversion, I dropped it in favor of eBPF side endian conversion + netaddr formating with the help of @4ast. netaddr is already used in other tools so seems not an issue.

Time to have some sleep. French time here.

yadutaf:
Moved to tools, updated documentation, added arguments. Remaining: add more protocols.

4ast:
lgtm

brendangregg:
LGTM.

In the future, we might want to change --show-netns to be just -v or -x, following the convention in other Unix tools (where -v or -x gives you extended mode; even though this currently only adds one column). Also, in the future, I suspect we'll want to drop the netaddr dependency and use bcc built ins, when they are available.

drzaeus77:
Do you think you can squash and rebase this change to clean up the final commit log that goes into master? Then we will merge it as-is.

yadutaf:
Sure. Do you prefer to merge as is with TCP only or to wait for other protocols ?

drzaeus77:
Just TCP for now.

yadutaf:
Done. Not sure why the fedora build failed.

Issue #431, opened 2016-03-08
biosnoop.py binary error


brendangregg:
From comments on my post http://www.brendangregg.com/blog/2016-03-05/linux-bpf-superpowers.html :

"
Hi, using biosnoop.py actually returns a bunch of TypeError exceptions for me.
Linux kernel version: 4.4.3-1
Python version: 3.5.1

sudo ./biosnoop.py
...
Traceback (most recent call last):
File "_ctypes/callbacks.c", line 234, in 'calling callback function'
File "/usr/lib/python3.5/site-packages/bcc/table.py", line 376, in <lambda>
fn = _RAW_CB_TYPE(lambda _, data, size: callback(cpu, data, size))
File "./biosnoop.py", line 157, in print_event
if not re.match('\?', event.name):
File "/usr/lib/python3.5/re.py", line 163, in match
return _compile(pattern, flags).match(string)

Problem seems to be that event.name contains binary chars, while the regexp at line 157 of biosnoop.py expects printable characters.
Rewriting line 157 as follows:
if not re.match(b'\?', event.name):
solves the issue on my machine.
Disclaimer: I'm not a Python expert.
"

drzaeus77:
That change is fine. Want me to commit on their behalf?

PR #432, opened 2016-03-08
Enable python3 compatible strings in biosnoop

drzaeus77:
Fixes: #431 

### DIFF
--- a/tools/biosnoop.py
+++ b/tools/biosnoop.py
@@ -154,7 +154,7 @@ def print_event(cpu, data, size):
     if event.rwflag == 0:
         rwflg = "R"
 
-    if not re.match('\?', event.name):
+    if not re.match(b'\?', event.name):
         val = event.sector
 
     if start_ts == 0:


4ast:
lgtm

brendangregg:
The change is ok, but I don't understand the logic. If the task name does not contain '?', then trust the sector value? What case was this for? //cc @mcaleavya

mcaleavya:
By looking at the original I took this to mean we didn't find a request lookup for the task name, we print ?

    if (valp == 0) {

        bpf_trace_printk("0 0 ? %d\\n", req->__data_len);

I added a strcpy to populate the ? as I was using a string for the task name in my struct. It may be best to remove the strcpy, but at the moment I am not sure how many times now we hit the missing of the task name now it's no longer using the printk.

brendangregg:
oh, right, thanks, I was using '?' to mean "start not seen". I was thrown by the re.match() thinking it was looking for ? elsewhere in the string, and I was wondering what kernel daemon / process that could be.

Later on we might want to make that a direct comparison (event.name != "?"), as well as renaming "val = event.sector" to be "nr_sector = ...". But we can do that later.

Issue #419, opened 2016-02-24
Kernel tracepoint support (ugly)


goldshtn:
I understand that adding kernel tracepoint support in BPF is going to take a while. In the meantime, would it be possible to put together a tool that relies on kprobes? What I had in mind was to enable the tracepoint using ftrace and then place a kprobe in the `trace_event_raw_event_*` function or in `trace_event_buffer_commit`, or some similar location. The kprobe would have access to the entry structure defined by `TRACE_EVENT`. 

Is that a direction worth pursuing? Would it be possible (for efficiency) to instruct the ftrace infrastructure to discard the event and not even put it in the ftrace buffer?

4ast:
:) I was hoping it won't get to it, but that's a reasonable 'workaround'.
Probably enabling tracepoints via perf and adding kprobe to perf_trace_*() will be a tiny bit better, since the code that enables tracepoint via perf_event_open will be reused when tracepoint+bpf is ready.
but going via ftrace is also ok-ish. Both are ugly hacks, of course, and will be slower than normal kprobe at the beginning of the function, since trace_event_raw* and perf_trace* are 'notrace', so kprobe will do full 'int 3'.
Also note that they will _not_ have access to struct of trace_event, but rather 'proto' part of it only.

goldshtn:
OK, suppose I go through perf - would it be possible to have perf discard the events and not write them out to perf.data? I'm looking to eliminate the inefficiency in doing analysis and filtering in user space, otherwise it isn't really worth pursuing...

4ast:
if perf_event_open is done for the task, then perf_trace* will be called, but it will not produce any output, since pid != pid_of_our_python_process. Another approach is to add a filter to it that always returns false. It's a bit slower, since alloc + copy of arguments and registers will be happening before discarding event. The main overhead is int3 anyway.

goldshtn:
OK, so just to make sure I understand the recommended approach. Turn on the tracepoints using `perf record` with a filter that always returns false so that no data actually goes into a file. Place a kprobe on one of the methods that process the perf event, such as `perf_tp_event` and try to capture data out of the provided record. Makes sense?

Also adding @brendangregg if you have any thoughts.

4ast:
[...]

PR #439, opened 2016-03-20
tracepoint support for argdist and trace, and new tplist tool

goldshtn:
Following up on the discussion in #419, this is an implementation of tracepoint support for argdist and trace. Additionally, the tplist tool is a simple wrapper on top of debugfs that prints the available tracepoints and the format variables that are available to argdist and trace as the `tp` structure.

A couple of examples:

```
# ./tplist.py -v block:block_rq_complete
block:block_rq_complete
    dev_t dev;
    sector_t sector;
    unsigned int nr_sector;
    int errors;
    char rwbs[8];

# ./argdist.py -H 't:block:block_rq_complete():int:tp.nr_sector'
[07:11:36]
     tp.nr_sector        : count     distribution
         0 -> 1          : 8        |****************************************|
        16 -> 31         : 1        |*****                                   |
^C

# ./trace.py 't:irq:irq_handler_entry (tp.irq != 27) "irq = %d", tp.irq'
07:13:25 0      swapper/0    irq_handler_entry irq = 24
07:13:25 0      swapper/0    irq_handler_entry irq = 24
^C
```

The code is currently duplicated and not shared in the BPF module, because adding the `BPF.attach_tracepoint` method would require the BPF module to add its own probe, which is something the module currently doesn't do -- and would seem like something that would be removed anyway when proper tracepoint support is introduced in the kernel.

### DIFF
--- a/tools/trace.py
+++ b/tools/trace.py
@@ -5,6 +5,7 @@
 #
 # USAGE: trace [-h] [-p PID] [-v] [-Z STRING_SIZE] [-S] [-M MAX_EVENTS] [-o]
 #              probe [probe ...]
+#
 # Licensed under the Apache License, Version 2.0 (the "License")
 # Copyright (C) 2016 Sasha Goldshtein.
 
@@ -13,6 +14,7 @@ from time import sleep, strftime
 import argparse
 import re
 import ctypes as ct
+import multiprocessing
 import os
 import traceback
 import sys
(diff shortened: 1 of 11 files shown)


tools/trace.py
```diff
@@ -42,6 +44,122 @@ def monotonic_time():
...
+        PERF_FLAG_FD_CLOEXEC = 8
+        PERF_EVENT_IOC_SET_FILTER = 1074275334
+        PERF_EVENT_IOC_ENABLE = 9216
+
+        libc = ct.CDLL('libc.so.6', use_errno=True)
+        syscall = libc.syscall          # not declaring vararg types
+        ioctl = libc.ioctl              # not declaring vararg types
+
+        @staticmethod
+        def _open_for_cpu(cpu, attr):
```
4ast: way too much copy paste. Since two tools need the same functions. Move them into some common part.

tools/trace.py
```diff
@@ -42,6 +44,122 @@ def monotonic_time():
...
+                attr = Perf.perf_event_attr()
+                attr.config = tpoint_id
+                attr.type = Perf.PERF_TYPE_TRACEPOINT
+                attr.sample_type = Perf.PERF_SAMPLE_RAW
+                attr.sample_period = 1
+                attr.wakeup_events = 1
+                for cpu in range(0, multiprocessing.cpu_count()):
+                        Perf._open_for_cpu(cpu, attr)
+
+class Tracepoint(object):
```
4ast: same for this whole class.

tools/trace.py
```diff
@@ -308,6 +486,31 @@ def generate_program(self, pid, include_self):
         def _time_off_str(cls, timestamp_ns):
                 return "%.6f" % (1e-9 * (timestamp_ns - cls.first_ts))
 
+        auto_includes = {
+                "linux/time.h"      : ["time"],
+                "linux/fs.h"        : ["fs", "file"],
+                "linux/blkdev.h"    : ["bio", "request"],
+                "linux/slab.h"      : ["alloc"],
+                "linux/netdevice.h" : ["sk_buff"]
```
4ast: copy paste mistake. Please move this to common as well

4ast:
looks good, but common code shouldn't be duplicated.

goldshtn:
It has been discussed before, and @brendangregg 's position has been that it's okay to duplicate because we get tools that are more stand-alone that way. I would gladly move common code to a module, but where would it be? In the tools folder? An additional part of the BPF module?

4ast:
Brendan's point was about pieces that going to evolve differently in different tools, whereas here this part of code should be common across tools and any bug fix must be applied to all, so can only be in common. imo src/python/bcc/tracepoint.py is the best place.

goldshtn:
Sure. While I'm at it - can I also move the user symbol resolution code from memleak to somewhere in the BPF module?

4ast:
yes please. usym resolution definitely common. There are pieces in bcc/__init__.py already.

brendangregg:
Yes, I think this stuff (tracepoints and usym) should be in the BPF module.

As an example of when I think duplication is ok: ext4dist, btrfsdist, xfsdist, and zfsdist, all pretty much do the same thing. So I could write a generic library file (fsdist) then simplify each tool. But I don't think that's worth it: it has value to have stand-alone tools, and, updating 4 in parallel is not as time consuming as it might sound. If I a -z option to one, then had to add it to 3 others, well, if it took 2 minutes to add it to each of the remaining 3, then it only costs 6 minutes.

But if it's something that's going to be used in dozens of tools, then it should probably go into BPF module.

brendangregg:
BTW, tplist looks good. Can't wait to use tracepoints!

goldshtn:
With all due respect, I disagree. I was very surprised to see that you added four tools that should have 99% shared code. But maybe for write-once tools that are pretty much done on v1 the bar should be different that for more reusable tools.

4ast:
I agree with Sasha. *fsslower could have been one tool to cover 4 file systems. There is really not much difference between them.

brendangregg:
It's a separate thread, but the advantages of separate tools include:

- easier adoption: if you're debugging a btrfs problem, then it's easy to find the btrfs tools.
- easier specific tool growth and maintenance: lets say the btrfs folk want to add a btrfs-specific option to the btrfs tools. No problem. But if they are modifying a unified *fsslower/*fsdist, then they need to test how this option behaves with other file system types, including zfs, which they may have no interest in doing.

I've written lots of multi-tools and single-purpose tools. Most of the execsnoop/opensnoop/etc could just be dtruss oneliners. But people prefered the execsnoop/opensnoop/etc ones. Personally, I like having both types.

goldshtn:
With the two latest commits, I moved the tracepoint-related stuff to tracepoint.py under the bcc module, and the auto-includes stuff to __init__.py. Should be ready to merge unless you have any other comments. The user symbol resolution will be in a separate PR.

4ast:
thanks!

PR #440, opened 2016-03-20
Fix compilation with llvm 3.8.0

ismail:
Add the needed LLVM libraries to the link line. Otherwise, we are getting
a lot of undefined symbols.

Tested on openSUSE Tumbleweed with llvm 3.8.0 (runtime tested with kernel 4.5.0).

### DIFF
--- a/src/cc/CMakeLists.txt
+++ b/src/cc/CMakeLists.txt
@@ -42,8 +42,8 @@ set_target_properties(bcc-static PROPERTIES OUTPUT_NAME bcc)
 
 # BPF is still experimental otherwise it should be available
 #llvm_map_components_to_libnames(llvm_libs bpf mcjit irreader passes)
-llvm_map_components_to_libnames(llvm_libs mcjit irreader passes linker
-  instrumentation objcarcopts bitwriter option x86codegen)
+llvm_map_components_to_libnames(llvm_libs bpfinfo bpfdesc core executionengine mcjit ipo irreader passes linker
+  instrumentation objcarcopts runtimedyld support bitwriter option x86codegen x86desc x86info)
 # order is important
 set(clang_libs ${libclangFrontend} ${libclangSerialization} ${libclangDriver} ${libclangParse}
   ${libclangSema} ${libclangCodeGen} ${libclangAnalysis} ${libclangRewrite} ${libclangEdit}


4ast:
I don't think that patch is needed. 3.8.0 and 3.9svn work fine without this.
Likely you have incompatibility in cmake versions building bcc and that one used to build suse llvm packages.

ismail:
Actually this is nothing about cmake (I use the same cmake as the llvm build version because I am the llvm/cmake maintainer). And it might be work for you because your ld doesn't do --as-needed (https://wiki.gentoo.org/wiki/Project:Quality_Assurance/As-needed) by default. I am attaching the build log so you can see whats going on.

[build.log.txt](https://github.com/iovisor/bcc/files/181438/build.log.txt)

4ast:
we've seen these errors in the past. They're due to llvm_map_components_to_libnames() failing to extract the dependencies. why cmake failing to do that is not clear.

ismail:
If you check my PR you'll see that I am adding libraries to llvm_map_components_to_libnames so its not a dependency extraction problem, its a whole new dependency its missing. Do me a favor and do a 

```
seferogullari ~/src/bcc/build > ldd src/cc/libbcc.so | grep BPF
        libLLVMBPFInfo.so.3.8 => /usr/lib64/libLLVMBPFInfo.so.3.8 (0x00007fe31e923000)
        libLLVMBPFDesc.so.3.8 => /usr/lib64/libLLVMBPFDesc.so.3.8 (0x00007fe31e719000)
        libLLVMBPFCodeGen.so.3.8 => /usr/lib64/libLLVMBPFCodeGen.so.3.8 (0x00007fe31bba8000)
        libLLVMBPFAsmPrinter.so.3.8 => /usr/lib64/../lib64/libLLVMBPFAsmPrinter.so.3.8 (0x00007fe319356000)
```

You see it needs to link to bpfinfo, bpfdesc, bpfcodegen. Codegen part is done in

```
src/cc/CMakeLists.txt
53:target_link_libraries(bcc-shared b_frontend clang_frontend ${clang_libs} ${llvm_libs} LLVMBPFCodeGen)
```

but bpfinfo, bpfdesc is missing. My patch adds these missing deps (among others).

drzaeus77:
We should not link against shared llvm libs. A static llvm install does not
have this problem. I would like to keep bcc having only a build time
dependency on llvm.

4ast:
$ ldd src/cc/libbcc.so |grep -i bpf|wc -l
0
it is llvm/cmake issue. The proper fix would be to call llvm_expand_dependencies() on the libs
instead of adding them manually.
I have a patch that does that but it's not clean enough to send pr.

ismail:
Well if you are linking to static llvm, of course you won't have this problem. That's a whole different story. I don't understand why you won't merge this patch because obviously it's right for shared llvm but, for now, I'll just maintain this as a local patch.

4ast:
> because obviously it's right for shared llvm 
it's not. we cannot chase llvm library dependencies manually. There is llvm_expand_dependencies() to address that in a generic way.

ismail:
llvm_expand_dependencies() seems to be only used in LLVM-Config.cmake and nowhere else (checked all the sub-project clang/lldb/lld etc.). Do you have any sample usage somewhere?

4ast:
the following works for me:
llvm_map_components_to_libnames(llvm_libs mcjit irreader passes linker
  instrumentation objcarcopts bitwriter option x86codegen bpfcodegen)
llvm_expand_dependencies(expanded_libs ${llvm_libs})
target_link_libraries(bcc b_frontend clang_frontend ${clang_libs} ${expanded_libs})

Probably initial set of llvm_libs can also be reduced and llvm_expand_dependencies()
will do all the heavy work of figuring out dependencies.
clang_libs is still ugly. hopefully there is an equivalent to llvm_expand_dependencies()
for clang.

Note that llvm_expand_dependencies() is implicitly called today already when cmake is not broken.
What I'm proposing it to call it explicitly to avoid all these linker issues.

ismail:
Ok, your proposed version doesn't work for me but I understand the general solution. For now, I'll stick to my version until a cleaner approach can be done. Thanks.

4ast:
why it didn't work?

ismail:
I am getting the same undefined reference errors, its not expanding to right set of libraries.

4ast:
hmm. let's debug it together? could you print expanded_libs ?

ismail:
Actually it was a copy/paste error, tried again and it works fine :) This looks pretty clean to me, would be nice to commit it to repo.

ismail:
And even better, you can trim llvm libs to:

```
llvm_map_components_to_libnames(llvm_libs mcjit passes x86codegen bpfcodegen)
```

And it works.

4ast:
awesome. could you respin your PR like this then?

PR #442, opened 2016-03-21
Moved user symbol decoding from memleak into bcc module

goldshtn:
User symbol decoding is a bit more complicated than with kernel symbols, because it depends on the pid. I added an inner class `BPF.ProcessSymbols` that represents the symbol info for a particular process. Users of the module can work with that directly (which is what memleak currently does), or they can use a simpler static method `BPF.usymaddr(pid, addr)` which uses a cache of `BPF.ProcessSymbols` instances per pid.

There's also the matter of refreshing the symbols. If the user process loads a new module, or unloads an existing module, the symbolic information we cached becomes stale. What memleak does is refresh the symbol cache on every reporting interval; so I added a method `ProcessSymbols.refresh_code_ranges()` which callers (currently: memleak) can use.

### DIFF
--- a/src/python/bcc/usyms.py
+++ b/src/python/bcc/usyms.py
@@ -1,0 +1,129 @@
+# Copyright 2016 Sasha Goldshtein
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from subprocess import Popen, PIPE, STDOUT
+
+class ProcessSymbols(object):
+    def __init__(self, pid):
+        """
+        Initializes the process symbols store for the specified pid.
+        Call refresh_code_ranges() periodically if you anticipate changes
+        in the set of loaded libraries or their addresses.
+        """
+        self.pid = pid
+        self.refresh_code_ranges()
+
+    def refresh_code_ranges(self):
+        self.code_ranges = self._get_code_ranges()
+        self.ranges_cache = {}
+        self.exe = self._get_exe()
+        self.start_time = self._get_start_time()
+
+    def _get_exe(self):
+        return ProcessSymbols._run_command_get_output(
+                "readlink -f /proc/%d/exe" % self.pid)
+
+    def _get_start_time(self):
+        return ProcessSymbols._run_command_get_output(
+                "cut -d' ' -f 22 /proc/%d/stat" % self.pid)
+
+    @staticmethod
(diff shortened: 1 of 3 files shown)


drzaeus77:
Code looks nice! Two questions:
1. Would you mind moving the new class to a new file? The original bcc lacked organization, but I'd like to structure it a bit more.
2. What happens on pid wrap?

goldshtn:
Sure, I'll extract it to a separate file. Not sure what to do about pid wrap though. I suppose it's possible to detect that the original pid exited and clear it from the cache, but I don't know if that's a responsibility that the bcc module wants to assume. The user of this class can call `ProcessSymbols.refresh_code_ranges()`, and I can have it clear the whole cache. Do you have some other suggestion?

drzaeus77:
Can you just keep the pid's comm in the hash, and flush the cache or warn if the commands don't match?

goldshtn:
Sure, although it's still possible for the comm to repeat when the pid wraps. Maybe it would be even better to store the full command line (/proc/PID/cmdline) and compare that. What do you think?

drzaeus77:
How about `readlink -f /proc/$PID/exe` + `awk '{print $22}' /proc/$PID/stat`? The binary name + starttime.

goldshtn:
Yeah, that would do. BTW, I could read procfs, but there's also [psutil](https://github.com/giampaolo/psutil), which provides a nice abstraction. What's your position on using an external module, that's not bundled with a typical Python distribution? I suspect you'd rather avoid it :-)

drzaeus77:
Yup, would rather avoid it in that case, but it's not a hard and fast rule. Bang for buck, etc. /proc is a stable api.

4ast:
no psutil please. it's a mix of c+python we cannot really use it without breaking it apart.

goldshtn:
All right, added pid wrap detection and moved the `ProcessSymbols` class to usyms.py.

src/python/bcc/usyms.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def refresh_code_ranges(self):
+        self.code_ranges = self._get_code_ranges()
+        self.ranges_cache = {}
+        self.exe = self._get_exe()
+        self.start_time = self._get_start_time()
+
+    def _get_exe(self):
+        return ProcessSymbols._run_command_get_output(
+                "readlink -f /proc/%d/exe" % self.pid)
```
4ast: /proc/self/exe is cleaner

src/python/bcc/usyms.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def _get_start_time(self):
+        return ProcessSymbols._run_command_get_output(
+                "cut -d' ' -f 22 /proc/%d/stat" % self.pid)
```
4ast: same here

src/python/bcc/usyms.py
goldshtn: But it's not just self, it's any pid you want to monitor. So it could be an arbitrary pid.

4ast: yeah, misread :)

4ast:
lgtm

goldshtn:
Now fixed to make it auto-mergable. Hope I didn't screw anything up in the process :-$

PR #448, opened 2016-03-26
Better stack walking APIs

vmg:
Hi folks!

Here's a hot take on replacing @brendangregg's lovingly crafted eBPF stack walker code (teehee) with the brand new StackTrace table APIs.

I totally get the kernel is not quite there yet, but the APIs in Torvald's mainline seem to be stabilized now and running a nightly kernel makes this work like a charm. I'll just twiddle my fingers until 4.6 comes out, but I wanted to open this PR early to discuss it. 👌

Looking at the (minimal) changes, the design for the new table seems just right, and the resulting Python APIs make all the stack-related code significantly simpler. Details on the changes are described in each individual commit message.

There's a couple points I'm not sure about -- I'm leaving inline comments with questions.

Thanks for the amazing work with `bcc`. This is by far the coolest thing going on right now re. perf work in Linux.

### DIFF
--- a/src/python/bcc/table.py
+++ b/src/python/bcc/table.py
@@ -412,8 +412,36 @@ class PerCpuArray(ArrayBase):
         raise Exception("Unsupported")
 
 class StackTrace(TableBase):
+    MAX_DEPTH = 127
+
     def __init__(self, *args, **kwargs):
         super(StackTrace, self).__init__(*args, **kwargs)
+
+    class StackWalker(object):
+        def __init__(self, stack, resolve=None):
+            self.stack = stack
+            self.n = -1
+            self.resolve = resolve
+
+        def __iter__(self):
+            return self
+
+        def __next__(self):
+            return self.next()
+
+        def next(self):
+            self.n += 1
+            if self.n == StackTrace.MAX_DEPTH:
+                raise StopIteration()
+
+            addr = self.stack.ip[self.n]
+            if addr == 0 :
+                raise StopIteration()
+
+            return self.resolve(addr) if self.resolve else addr
+
+    def walk(self, stack_id, resolve=None):
+        return StackTrace.StackWalker(self[self.Key(stack_id)], resolve)
 
     def __len__(self):
         i = 0
--- a/examples/tracing/mallocstacks.py
+++ b/examples/tracing/mallocstacks.py
@@ -1,0 +1,60 @@
+#!/usr/bin/python
+#
+# mallocstacks  Trace malloc() calls in a process and print the full
+#               stack trace for all callsites.
+#               For Linux, uses BCC, eBPF. Embedded C.
+#
+# This script is a basic example of the new Linux 4.6+ BPF_STACK_TRACE
+# table API.
+#
+# Copyright 2016 GitHub, Inc.
+# Licensed under the Apache License, Version 2.0 (the "License")
+
+from __future__ import print_function
+from bcc import BPF, ProcessSymbols
+from time import sleep
+import sys
+
+if len(sys.argv) < 2:
+    print("USAGE: mallocstacks PID")
+    exit()
+pid = int(sys.argv[1])
+
+# load BPF program
+b = BPF(text="""
+#include <uapi/linux/ptrace.h>
+
+BPF_HASH(calls, int);
+BPF_STACK_TRACE(stack_traces, 1024)
+
+int alloc_enter(struct pt_regs *ctx, size_t size) {
+    int key = stack_traces.get_stackid(ctx,
+        BPF_F_USER_STACK|BPF_F_REUSE_STACKID);
+    if (key < 0)
+        return 0;
+
+    u64 zero = 0, *val;
+    val = calls.lookup_or_init(&key, &zero);
+    (*val) += size;
+    return 0;
+};
+""")
+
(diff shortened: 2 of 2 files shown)


src/python/bcc/table.py
```diff
@@ -412,9 +412,37 @@ def __init__(self, *args, **kwargs):
         raise Exception("Unsupported")
 
 class StackTrace(TableBase):
+    MAX_DEPTH = 127
```
vmg: This value is hardcoded in the kernel. I am not sure whether it'd be worth it to dig into the FFI ctypes to figure it out dynamically. Thoughts?

tools/memleak.py
```diff
@@ -223,12 +203,12 @@ def run_command_get_pid(command):
         sizes.delete(&pid);
 
         info.timestamp_ns = bpf_ktime_get_ns();
-        info.num_frames = grab_stack(ctx, &info) - 2;
```
vmg: We were dropping two frames here, and I'm not exactly sure why. At least for userland stacks, these frames seem perfectly fine.

tools/offcputime.py
```diff
@@ -183,24 +140,17 @@ def signal_ignore(signal, frame):
...
     for k, v in sorted(counts.items(), key=lambda counts: counts[1].value):
         if folded:
             # print folded stack output
-            line = k.name.decode() + ";"
-            for i in reversed(range(0, maxdepth)):
-                if k.ret[i] == 0:
-                    continue
-                line = line + b.ksym(k.ret[i])
-                if i != 0:
-                    line = line + ";"
```
vmg: Same here -- we were dropping the topmost frame (which ought to always be `finish_task_switch`). Shall I drop it too?

tools/memleak.py
goldshtn: Yep, that was a workaround. I don't remember the exact details anymore :-) ...but I think memleak would out irrelevant values in the last two entries. If we can grab the full stack without any of these hacks, I'm all for it.

goldshtn:
Looks awesome!

By the way, when this gets merged, the man and example files also need to be updated to point out that the stack support is not experimental anymore, and to remove the stack depth command line arguments.

src/python/bcc/table.py
drzaeus77: @4ast it makes sense to me to leave this unless the kernel value is ever likely to change. How likely is that?

The problem with getting this dynamically will be that the kernel headers may update but the libc-headers, where the userspace value of this will be most readily available, is very slow to update. I tend to come across (and am one myself) lagging userspace header files, which is why bcc includes a bunch of compat header files and hardcoded things.

drzaeus77:
It might make sense to split out the updates to table.py into a separate PR, since that can merge right away, while the others we might want to discuss about backwards compat.

tools/offcputime.py
brendangregg: This was dropped specifically for offcputime, since it decorated the top edge of the final flame graph with the finish_task_switch function -- which was redundant and increased the SVG size without value (I'm worried about SVG size since they can get so large they get sluggish in browsers).

src/python/bcc/table.py
4ast: PERF_MAX_STACK_DEPTH is highly unlikely to change. All sorts of abi concerns will pop up, including increase in size of perf.data files, since 127 is also default for 'perf record -g'

brendangregg:
Thanks! I was looking for this specific support yesterday and saw it hadn't been done yet.

Might want a hello world simple example in /examples, or, (either now or later) we can update stackcount/stacksnoop in /tools which can serve as the same thing.

Also, as tools get rewritten to use the new stack support, I'd put old copies of those tools (which used unrolling hacks) in /tools/old -- particularly for Ubuntu Xenial users.

FWIW, the stackdepth limit of 127 will be suitable for many cases, but we're already exceeding it for Java/Groovy stacks (which can get as deep as >512).

goldshtn:
While we are at it, would it make sense to extend the user symbol resolution support with data from perf-map-agent? That would give us call stacks for Java and Node, right? @brendangregg

brendangregg:
@goldshtn right, it does make sense, now or later. User symbol resolution should check for /tmp/perf-PID.map files. Format is tools/perf/Documentation/jit-interface.txt .

tools/offcputime.py
vmg: 👌 Will fix.

src/python/bcc/table.py
vmg: Good! Let's leave it hardcoded then.

vmg:
> It might make sense to split out the updates to table.py into a separate PR, since that can merge right away, while the others we might want to discuss about backwards compat.

Done! This PR now only contains the changes to `table.py` and a Hello World example (`examples/tracing/mallocstacks.py`) as suggested by @brendangregg.

I've opened https://github.com/iovisor/bcc/pull/450 where I'm working on converting the actual tools to use 4.6.

> By the way, when this gets merged, the man and example files also need to be updated to point out that the stack support is not experimental anymore, and to remove the stack depth command line arguments.

This is done in #450.

> Also, as tools get rewritten to use the new stack support, I'd put old copies of those tools (which used unrolling hacks) in /tools/old -- particularly for Ubuntu Xenial users.

Also done in #450 👌

PR #449, opened 2016-03-27
Fixed bug around local assignment from char array

goldshtn:
Missed that test case when creating automatic local variables from the tracepoint struct. For example, `trace 't:block:block_rq_complete'` wouldn't work at all because it tries to do `char rwbs[8] = tp.rwbs[8];` which makes no sense. Fixed in this PR.

### DIFF
--- a/src/python/bcc/tracepoint.py
+++ b/src/python/bcc/tracepoint.py
@@ -147,6 +147,11 @@ struct %s {
         def _generate_struct_locals(self):
                 text = ""
                 for field_type, field_name in self.struct_fields:
+                        if field_type == "char" and field_name.endswith(']'):
+                                # Special case for 'char whatever[N]', should
+                                # be assigned to a 'char *'
+                                field_type = "char *"
+                                field_name = re.sub(r'\[\d+\]$', '', field_name)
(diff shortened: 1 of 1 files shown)


src/python/bcc/tracepoint.py
```diff
@@ -147,6 +147,11 @@ def generate_struct(self):
         def _generate_struct_locals(self):
                 text = ""
                 for field_type, field_name in self.struct_fields:
+                        if field_type == "char" and field_name.endswith(']'):
+                                # Special case for 'char whatever[N]', should
+                                # be assigned to a 'char *'
+                                field_type = "char *"
+                                field_name = field_name[:-3]
```
drzaeus77: What is -3? Is it for `[N]`? In that case, what if N is 2 digits or larger?

brendangregg:
LGTM

src/python/bcc/tracepoint.py
goldshtn: Yeah, I wasn't thinking straight. Fixed in the latest push.

goldshtn:
@drzaeus77: Just wanted to make sure you haven't missed this, I fixed the issue you pointed out and updated the PR.

PR #450, opened 2016-03-27
Stack Traces: TNG

vmg:
This PR is a followup on https://github.com/iovisor/bcc/pull/448 and updates several tools to use the new stack trace APIs in 4.6+.

As always, details on the changes are on each individual commit message. The man pages have been updated to reflect the changes, and the old versions using the manual stack walker have been saved to `tools/old`.

### TODO
- [x] `memleak.py`
- [x] `offcputime.py`
- [x] `stackcount.py`
- [x] `stacksnoop.py`

### DIFF
--- a/man/man8/memleak.8
+++ b/man/man8/memleak.8
@@ -3,7 +3,7 @@
 memleak \- Print a summary of outstanding allocations and their call stacks to detect memory leaks. Uses Linux eBPF/bcc.
 .SH SYNOPSIS
 .B memleak [-h] [-p PID] [-t] [-a] [-o OLDER] [-c COMMAND] [-s SAMPLE_RATE]
-[-d STACK_DEPTH] [-T TOP] [-z MIN_SIZE] [-Z MAX_SIZE] [INTERVAL] [COUNT]
+[-T TOP] [-z MIN_SIZE] [-Z MAX_SIZE] [INTERVAL] [COUNT]
 .SH DESCRIPTION
 memleak traces and matches memory allocation and deallocation requests, and
 collects call stacks for each allocation. memleak can then print a summary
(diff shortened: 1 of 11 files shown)


4ast:
lgtm. should the man page say that kversion<=4.5 can be found in tools/old ? probably not. I guess that's implicit assumption for all tools in old and we don't need to partition it further per kernel version.

vmg:
@4ast: Thank you! I've updated `stacksnoop` and `stackcount` too, which afaict are the two remaining tools that still had a manual stack walker.

@brendangregg: For `stacksnoop`, I've significantly simplified the output of the tool, given that we no longer log a line for each frame in the trace. The examples and manpage have been updated accordingly. Please let me know if you think the new output is acceptable.

Also, 4.6-rc1 is out! Perfect timing to start trying this stuff out. http://lkml.iu.edu/hypermail/linux/kernel/1603.3/01187.html 🎉

brendangregg:
@vmg great, thanks! New output of stacksnoop looks good. (Old output was partially for debug anyway.)

@4ast I'm thinking we need a /docs/Ubuntu-Xenial.md for specific instructions, which points people to tools/old. And /docs can be for more misc stuff that isn't high level (like /INSTALL.md).

brendangregg:
... wow, so BPF_MAP_TYPE_STACK_TRACE is in 4.6-rc1. I must have missed the final integration email.

evverx:
Oh, I can simply run `tools/old/offcputime.py` instead of
```
$ git checkout e82fb1baa26b0~1
$ cmake .. -DCMAKE_INSTALL_PREFIX=/usr
```
:)

So, yeah,
> @4ast I'm thinking we need a /docs/Ubuntu-Xenial.md for specific instructions, which points people to tools/old

makes sense.

PR #451, opened 2016-03-28
argdist, trace, and tplist support for USDT probes

goldshtn:
These tools now support USDT probes with the 'u:provider:probe' syntax.
Probes in a library or process can be listed with 'tplist -l LIB' or 'tplist -p PID'.
Probe arguments are also parsed and available in both argdist and trace as arg1,
arg2, etc., regardless of the probe attach location.

The same USDT probe can be used at multiple locations, which means the attach infra-
structure must probe all these locations. argdist and trace register thunk probes
at each location, which call a central probe function (which is static inline) with
the location id (__loc_id). The central probe function checks the location id to
determine how the arguments should be retrieved -- this is location-dependent.

Finally, some USDT probes must be enabled first by writing a value to a memory
location (this is called a "semaphore"). This value is per-process, so we require a
process id for this kind of probes. WARNING: This whole business of writing a value
to another process' memory is inherently unsafe, and needs more thorough testing,
especially when the semaphore is in a shared library whose load address can depend
on the pid.

Along with trace and argdist tool support, this commit also introduces new classes
in the bcc module: ProcStat handles pid-wrap detection, whereas USDTReader,
USDTProbe, USDTProbeLocation, and USDTArgument are the shared USDT-related
infrastructure that enables enumeration, attachment, and argument retrieval for
USDT probes.

### DIFF
--- a/src/python/bcc/usdt.py
+++ b/src/python/bcc/usdt.py
@@ -1,0 +1,433 @@
+# Copyright 2016 Sasha Goldshtein
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import struct
+import re
+
+from . import BPF
+from . import ProcStat
+
+class USDTArgument(object):
+        def __init__(self, size, is_signed, register=None, constant=None,
+                     deref_offset=None, deref_name=None):
+                self.size = size
+                self.is_signed = is_signed
+                self.register = register
+                self.constant = constant
+                self.deref_offset = deref_offset
+                self.deref_name = deref_name
+
+        def _normalize_register(self):
+                normalized = self.register
+                if normalized is None:
+                        return None
+                if normalized.startswith('%'):
+                        normalized = normalized[1:]
+                if normalized in USDTArgument.translations:
+                        normalized = USDTArgument.translations[normalized]
+                return normalized
+
+        translations = {
+                "rax": "ax", "rbx": "bx", "rcx": "cx", "rdx": "dx",
+                "rdi": "di", "rsi": "si", "rbp": "bp", "rsp": "sp",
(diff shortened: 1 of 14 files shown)


goldshtn:
Updated this PR with somewhat-tested support for semaphores in shared objects. I used /proc/PID/maps to find the .so's load address, and then added the semaphore address (from the stapsdt note) to the load address. It seems to be working :clap:

```
# ./trace.py 'u:/home/vagrant/libusdt_shared.so:loop_iter "i=%d msg=%s", arg1, arg2' -p $(pidof usdt_main)
06:30:24 26059  usdt_main    loop_iter        i=42 msg=All right
06:30:24 26059  usdt_main    loop_iter        i=0 msg=Hello!
06:30:24 26059  usdt_main    loop_iter        i=1 msg=Hello!
06:30:24 26059  usdt_main    loop_iter        i=2 msg=Hello!
06:30:24 26059  usdt_main    loop_iter        i=3 msg=Hello!
06:30:24 26059  usdt_main    loop_iter        i=4 msg=Hello!
06:30:24 26059  usdt_main    loop_iter        i=5 msg=Hello!
06:30:24 26059  usdt_main    loop_iter        i=6 msg=Hello!
06:30:24 26059  usdt_main    loop_iter        i=7 msg=Hello!
06:30:24 26059  usdt_main    loop_iter        i=8 msg=Hello!
06:30:24 26059  usdt_main    loop_iter        i=9 msg=Hello!
^C
```

src/python/bcc/usdt.py
```diff
@@ -0,0 +1,433 @@
...
+                        text += (" "*16) + \
+                                arg.generate_assign_to_local(
+                                                "%s%d" % (prefix, i)) + "\n"
+                return text
+
+        def _parse_args(self):
+                for arg in self.raw_args.split():
+                        self._parse_arg(arg.strip())
+
+        def _parse_arg(self, arg):
```
4ast: where do you get these text args to feed into this func? are they embedded as strings in notes? 'readelf -n' | re for stap, right? notes section is preserved in binaries? Would be good to describe how the whole thing works.

goldshtn: Yes, it's coming from readelf (see below in USDTReader). By "describe" do you mean in code comments or a separate documentation file?

4ast: Something like 'usdt prerequisites' or user focused doc is needed. Or faq that describes what should be in the binary for this mechanism to work. May be including 'readelf -n|grep magic', so that users can do quick sanity test that stap style usdts are there.

goldshtn:
I agree. I'll work on docs and submit a separate PR.

src/python/bcc/usdt.py
goldshtn: I agree. I'll work on docs and submit a separate PR.

PR #453, opened 2016-03-28
expose buitin byte swap methods

yadutaf:
This PR makes it possible to directly use byte swap macro from the kernel headers by advertising the compiler's builtin byte swap method. In this commit, this is simply a forced ``CFLAGS`` in the python lib.

Not sure this is the best way to do. It should probably go in the clang frontend. Especially if we want to use it from non python programs.


4ast:
I think that's a good thing to do regardless of other discussion around ntoa. The macro is kernel specific and clang won't be advertising then. @drzaeus77 could you re-trigger bots?

PR #454, opened 2016-03-28
IPv6 strings with an inet6_ntoa

brendangregg:
This has a simple implementation of an inet6_ntoa() in tcpconnect. If this proves suitable, we can move it to bcc __init.__.py.

Known shortcomings: requires "re" package (on my systems that's usually there, whereas netaddr usually isn't). Also doesn't shorten maximum run, so is RFC4291 but not RFC5952 (missing 4.2.1), however, most of the time it is producing RFC5952 addresses. I think it works fine for now, but someone later might drop in a better Python implementation.

### DIFF
--- a/tools/tcpconnect.py
+++ b/tools/tcpconnect.py
@@ -24,6 +24,7 @@
 from __future__ import print_function
 from bcc import BPF
 import argparse
+import re
 import ctypes as ct
 
 # arguments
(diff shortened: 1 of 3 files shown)


4ast:
lgtm

tools/tcpconnect.py
```diff
@@ -228,6 +234,35 @@ def inet_ntoa(addr):
         addr = addr >> 8
     return dq
 
+def inet6_ntoa(addr):
```
drzaeus77: I think this is overly complicated. An approach with struct.unpack and an array of bytes along with your compacting regex would be a lot less code and still eliminate netaddr dependency. The struct module is also a python stdlib. Something like:

```python
s = ":".join(["%x"%x for x in unpack(">HHHHHHHH", rawaddr)])
s = re.sub(r'(^|:)0:(0:)+', r'::', s, 1)
```

brendangregg:
ok; I haven't gotten unpack to work yet but I'll keep trying.

brendangregg:
Forgot to note it earlier: this PR will be to fix #380 .

drzaeus77:
That looks better. If addr were a byte[] then the pack("QQ" could be avoided, but this version works.

PR #456, opened 2016-03-29
Minor endian and debug enhancement

yadutaf:
- add DEBUG_* constants in place of the hardcoded values. Should help making self-documenting code
- move endian fixes to clang, previous PR was merged... fast :)

I was trying to automatically insert ``bpf_probe_read`` in situations like https://github.com/iovisor/bcc/blob/master/tools/solisten.py#L105 but did not find the right way. I tried tweaking https://github.com/iovisor/bcc/blob/master/src/cc/frontends/clang/b_frontend_action.cc with no success. Do you have a guide on how to run it into a debugger ?

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -688,7 +688,7 @@ BFrontendAction::BFrontendAction(llvm::raw_ostream &os, unsigned flags)
 }
 
 void BFrontendAction::EndSourceFileAction() {
-  if (flags_ & 0x4)
+  if (flags_ & DEBUG_PREPROCESSOR)
     rewriter_->getEditBuffer(rewriter_->getSourceMgr().getMainFileID()).write(llvm::errs());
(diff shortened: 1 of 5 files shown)


4ast:
lgtm

drzaeus77:
I generally used `llvm::errs()` to debug, since there is a lot of tree-walking going on such that following in gdb can get confusing. It was easier for me to get a dump of the control flow and stare at it after the fact. That said, the easiest way to get gdb into b_frontend_action is probably to instrument or attach to tests/cc/test_static, since that will be straightforward to test and run.

The probe stuff happens mostly in the ProbeVisitor class (which you may have already come across). Try for instance adding `E->dump()` wherever `memb_visited_.insert(E);` occurs. In your case, what I suspect happens is that the call to `inet_sk(sk)` hides the `ProbeChecker::needs_probe_` semantic from the analysis. Otherwise doing `evt.lport = inet->inet_sport` might have worked automatically.

PR #458, opened 2016-03-29
Percpu support

zaafar:
This patch enables **percpu** tables (hash/array) support in bcc. Following are the notes related to this patch.

1. It provides two additional keywords **pc_hash** and **pc_array** so that users can create percpu hash/array table.

2. These tables use an array of Leaf rather than single Leaf object to set/get the data from the kernel, so pc_hash/pc_array **Leaf** data type is an array of Leaf with array size equal to total number of cpu. Individual/original Leaf data type can be used via **sLeaf ** type. This change will not effect other tables type.

3. As only 8-bytes aligned data can be passed to kernel for getting/setting the key/value item, so only 8-bytes aligned data types (simple or structure) are allowed with exception to u32/int data types. In case of wrong/invalid leaf data types, this patch will throw an error. More data types can be added upon feedback.

4. pc_hash/pc_array tables setter (__setitem__) requires the user to create/initialize/pass the whole array (type Leaf) rather than just one instance (sLeaf).

5. When getting a key/value pair of pc_hash/array table user have the option to either get the whole array or apply some reducer function like **sum**, **average**, **max** and get the individual object (sLeaf data type). Default reducer functions are provided for simple data types, for complex data types (structs) user can define their own reducer function (in form of lambda function) and pass them to the pc_hash/array class.

6. This patch also includes a very simple test case, testing all of these features on u32, u64 and aligned structure data type.

### DIFF
--- a/src/python/bcc/table.py
+++ b/src/python/bcc/table.py
@@ -74,8 +74,8 @@ def _print_log2_hist(vals, val_type):
                       _stars(val, val_max, stars)))
 
 
-def Table(bpf, map_id, map_fd, keytype, leaftype, func_reducer):
-    """Table(bpf, map_id, map_fd, keytype, leaftype, func_reducer)
+def Table(bpf, map_id, map_fd, keytype, leaftype, **kwargs):
+    """Table(bpf, map_id, map_fd, keytype, leaftype, **kwargs)
 
     Create a python object out of a reference to a bpf table handle"""
 
@@ -90,11 +90,9 @@ def Table(bpf, map_id, map_fd, keytype, leaftype, func_reducer):
     elif ttype == BPF_MAP_TYPE_PERF_EVENT_ARRAY:
         t = PerfEventArray(bpf, map_id, map_fd, keytype, leaftype)
     elif ttype == BPF_MAP_TYPE_PERCPU_HASH:
-        t = PerCpuHash(bpf, map_id, map_fd, keytype,
-                            leaftype, func=func_reducer)
+        t = PerCpuHash(bpf, map_id, map_fd, keytype, leaftype, **kwargs)
     elif ttype == BPF_MAP_TYPE_PERCPU_ARRAY:
-        t = PerCpuArray(bpf, map_id, map_fd, keytype,
-                            leaftype, func=func_reducer)
+        t = PerCpuArray(bpf, map_id, map_fd, keytype, leaftype, **kwargs)
     elif ttype == BPF_MAP_TYPE_STACK_TRACE:
         t = StackTrace(bpf, map_id, map_fd, keytype, leaftype)
     if t == None:
@@ -149,9 +147,9 @@ class TableBase(MutableMapping):
             raise Exception("Could not scanf leaf")
         return leaf
 
-    def __getitem__(self, key, leaf=None):
+    def __getitem__(self, key):
         key_p = ct.pointer(key)
-        if not leaf: leaf = self.Leaf()
+        leaf = self.Leaf()
         leaf_p = ct.pointer(leaf)
         res = lib.bpf_lookup_elem(self.map_fd,
                 ct.cast(key_p, ct.c_void_p),
@@ -312,9 +310,9 @@ class ArrayBase(TableBase):
     def __len__(self):
         return self.max_entries
 
-    def __getitem__(self, key, leaf=None):
+    def __getitem__(self, key):
         key = self._normalize_key(key)
-        return super(ArrayBase, self).__getitem__(key, leaf)
+        return super(ArrayBase, self).__getitem__(key)
 
     def __setitem__(self, key, leaf):
         key = self._normalize_key(key)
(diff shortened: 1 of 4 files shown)


tests/python/test_percpu.py
```diff
@@ -0,0 +1,105 @@
...
+
+from bcc import BPF
+from time import sleep
+import unittest as ut
+
+class TestPercpu(ut.TestCase):
+
+    def test_u64(self):
+        test_prog1 = """
+        BPF_TABLE("pc_hash", u32, u64, stats, 1);
```
4ast: I don't like the names. Seeing 'pc_hash' the last thing comes to mind that they are 'percpu_hash'. please pick something else.

drzaeus77: how about `BPF_TABLE_PERCPU("hash", ...)`

4ast: yep. that's better. changing macro we probably won't need to touch rewriter?

drzaeus77: Probably still would, the macro would probably just `##` append a different name. The rewriter needs a hint to be able to pick the map_type variable.

4ast: then why new macro? just use "percpu_hash" as a string...

drzaeus77: That's fine too.

src/python/bcc/table.py
```diff
@@ -403,13 +404,123 @@ def close_perf_buffer(self, key):
             del(self.bpf.open_kprobes()[(id(self), key)])
         del self._cbs[key]
 
-class PerCpuHashTable(TableBase):
+class PerCpuHash(HashTable):
     def __init__(self, *args, **kwargs):
-        raise Exception("Unsupported")
+        self.reducer = kwargs["reducer"]
```
drzaeus77: You should probably use `kwargs.pop("reducer", None)` to avoid KeyError as well as combine this and the line below into a single statement.

src/python/bcc/table.py
```diff
@@ -403,13 +404,123 @@ def close_perf_buffer(self, key):
...
+            return ret
+
+    def __setitem__(self, key, leaf):
+        super(PerCpuHash, self).__setitem__(key, leaf)
+
+    def sum(self, key):
+        if isinstance(self.Leaf(), ct.Structure):
+            raise IndexError("Leaf must be an integer type for default sum functions")
+        temp = self.reducer
+        self.reducer = None
```
drzaeus77: This isn't very nice. For instance, consider what happens when an exception is raised...self.reducer is lost. Other issues such as parallelism bugs would occur as well.

src/python/bcc/table.py
```diff
@@ -403,13 +404,123 @@ def close_perf_buffer(self, key):
...
+
+    def __getitem__(self, key):
+        result = super(PerCpuHash, self).__getitem__(key)
+        if self.alignment is 0:
+            ret = result
+        else:
+            ret = (self.sLeaf * self.total_cpu)()
+            for i in range(0, self.total_cpu):
+                ret[i] = result[i]
+        if (self.reducer):
```
drzaeus77: parentheses are unnecessary

tests/python/test_percpu.py
```diff
@@ -0,0 +1,105 @@
...
+        }
+        """
+        self.addCleanup(self.cleanup)
+        bpf_code = BPF(text=test_prog1)
+        stats_map = bpf_code.get_table("stats")
+        bpf_code.attach_kprobe(event="sys_clone", fn_name="hello_world")
+        sleep(1)
+        self.assertEqual(len(stats_map),1)
+        for x in range(0, 10):
+            ini = stats_map.Leaf(0,0,0,0,0,0,0,0)
```
drzaeus77: This test case should be num_processors aware.

tests/python/test_percpu.py
```diff
@@ -0,0 +1,105 @@
...
+                reducer=lambda x,y: stats_map.sLeaf(x.c1+y.c1))
+        bpf_code.attach_kprobe(event="sys_clone", fn_name="hello_world")
+        sleep(1)
+        self.assertEqual(len(stats_map),1)
+        for x in range(0, 10):
+            ini = stats_map.Leaf()
+            for i in ini:
+                i = stats_map.sLeaf(0,0)
+            stats_map[ stats_map.Key(0) ] = ini
+            sleep(1)
```
drzaeus77: Can you come up with a test case with no sleeps, and especially no sleeps in the middle of a for loop? I would like to keep every test case within the repo under 1 second, if possible.

tests/python/test_percpu.py
```diff
@@ -0,0 +1,105 @@
...
+            k = stats_map[ stats_map.Key(0) ]
+            self.assertGreater(k.c1, 1L)
+
+    def cleanup(self):
+        BPF.detach_kprobe("sys_clone")
```
drzaeus77: This shouldn't be necessary.

tests/python/test_percpu.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+if __name__ == "__main__":
+    ut.main()
```
drzaeus77: Please add a newline.

tests/python/test_percpu.py
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env python
+# Copyright (c) PLUMgrid, Inc.
+# Licensed under the Apache License, Version 2.0 (the "License")
+
+from bcc import BPF
+from time import sleep
+import unittest as ut
```
drzaeus77: No need to abbreviate unittest, renames like this should really be used sparingly. 'ut' only shows up twice in the file and is unlikely to grow.

tests/python/test_percpu.py
zaafar: Without this you get resource busy error.
Either I do this or execute detach_kprobe instruction at the end of each function.

tests/python/test_percpu.py
zaafar: done.

src/python/bcc/table.py
zaafar: done.

src/python/bcc/table.py
zaafar: done.

tests/python/test_percpu.py
zaafar: updated, now time is under 1 second.

4ast:
lgtm

PR #467, opened 2016-03-30
Support for global variable arguments in USDT probes

goldshtn:
I previously thought it wouldn't be possible to probe arbitrary user-mode addresses,
but it turns out that it is. This was required to support global variable arguments
in USDT probes. For example, there are a few probes in libpthread whose arguments
refer to the `mp_` global variable. These are now supported.

Additionally, refactored process-related utils into a `ProcUtils` class in **procstat.py**,
and implemented a version of `which` (`ProcUtils.which`) that is as portable as
possible -- it only relies on traversing the PATH environment variable.

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -26,7 +26,7 @@ import sys
 basestring = (unicode if sys.version_info[0] < 3 else str)
 
 from .libbcc import lib, _CB_TYPE
-from .procstat import ProcStat
+from .procstat import ProcStat, ProcUtils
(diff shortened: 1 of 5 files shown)


brendangregg:
Thanks! BTW, tools/trace.py still has the /usr/bin/which --skip-alias.

goldshtn:
@brendangregg Thanks, I fixed it in argdist but forgot to fix in trace, which is the one I was supposed to fix in the first place. Completed in #468. :tired_face:

PR #468, opened 2016-03-30
Replaced `which --skip-alias` with `ProcUtils.which`

goldshtn:
(Fixed in `argdist` but forgot to fix in `trace`. This completes #467.)


brendangregg:
Works, thanks!

PR #470, opened 2016-03-31
Standalone Lua tracing tool

vmg:
Here's something that @4ast requested. We can package all the lua tooling into its own executable with no dependencies, that can be used in production systems without having to install anything.

Some notes:

1. I've simplified the modules in the Lua wrapper to make it easier to package
2. The packaging happens in 3 steps (they can be followed in the `CMakeLists.txt` file for the lua folder)
    1. We use the `squish` tool to pre-compile and minimize all the Lua code into a single file
    2. We run `luajit -b` on the generated lua bundle -- `luajit` is smart enough to write an `.o` object with the compiled bytecode. This object is ready to link into our tool. How cool is that?
    3. We package a simple wrapper for the C API (src/lua/src/main.c`) and our Lua bytecode object into the standalone tool

Observant minds have probably noticed that we're not statically linking `libbcc.a` into the tool. This is because the LuaJIT FFI kinda expects to be able to `dlopen` the `so`. Since the `so` we're building has no dependencies already, it should be safe to ship with our tool. The tool is capable of finding this shared object if it's in the same folder as the tool itself, or through the `LIBBCC_SO_PATH` environment var, if it's set.

I'm 90% happy with this solution. It will work reliably in production by simply shipping the tool and the `so` to any machine. I believe a fully standalone solution may be feasible if I could manage to export all the `libbcc` symbols when linking `libbcc.a` into the tool -- then I could trick LuaJIT into `dlopen`-ing itself and it should be able to access the library from Lua. I'm not sure at all of how to accomplish this with CMake, however. I can successfully link against `libbcc.a`, but none of the symbols are exported on the resulting binary.

Feedback welcome!


src/lua/src/main.c
```diff
@@ -0,0 +1,186 @@
+#include <signal.h>
```
4ast: please add license banner to the top and format the first part of the file to match the 2nd part. Like spaces after funcname and before (. Just to be consistent.

src/lua/src/main.c
```diff
@@ -0,0 +1,186 @@
...
+
+		strcpy(dirname + 1, "libbcc.so");
+		libbcc = buffer;
+	}
+
+	if (access(libbcc, F_OK|R_OK|X_OK) != 0)
+		return -1;
+
+	lua_pushstring(L, libbcc);
+	lua_setglobal(L, "LIBBCC_SO_PATH");
```
4ast: the approach is good for now. The concern with lua-bcc+bcc.so is that we cannot easily have two versions of lua-bcc on the box. Say I would like to install lua-bcc-1.0 everywhere in /usr/local/bin/, then lua-bcc-2.0 cannot go in there, since it would need to look for libbcc-2.0.so. In this case we don't want to change .so versions for every git commit. Anyway, we need to have libbcc.a to include all clang/llvm .a first. It's been on my todo list and I have a hacky patch to do that that merges all static libs into single libbcc.a

tests/lua/test_uprobes.lua
```diff
@@ -68,4 +67,4 @@ function TestUprobes:teardown()
   BPF.cleanup_probes()
 end
 
-os.exit(LuaUnit.run())
+suite("TestUprobes", TestUprobes)
```
4ast: looks like the shiny new bcc-lua will be built, but not a single test for it?

4ast:
squish part is impressive!

src/lua/src/main.c
drzaeus77: What is missing from libbcc.a?

```bash
ldd tests/cc/test_static
        linux-vdso.so.1 (0x00007ffcb3760000)
        librt.so.1 => /usr/lib/librt.so.1 (0x00007fc74b882000)
        libdl.so.2 => /usr/lib/libdl.so.2 (0x00007fc74b67e000)
        libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007fc74b461000)
        libz.so.1 => /usr/lib/libz.so.1 (0x00007fc74b24b000)
        libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x00007fc74aec9000)
        libm.so.6 => /usr/lib/libm.so.6 (0x00007fc74abc4000)
        libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00007fc74a9ae000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007fc74a60d000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fc74ba8a000)
```

4ast: yep, my merge static libs stuff is not necessary for static bcc-lua.
libbcc.a has only bcc bits and none of clang/llvm.
so for building tests/cc/test_static or static bcc-lua it is fine,
but any external repo that wants to use bcc c++/c api it needs to be a part of bcc cmake system, otherwise clang/llvm dependency is lost.

vmg: Yeah, as far as I was able to tell, the `libbcc.a` we're building gets LLVM static right already. It's only depending on system libs, and runs standalone...

4ast: yes. libstdc++.so is another concern, but we can build libbcc.so to staticly link it already. Ideally static bcc-lua would embed it as well.

src/lua/src/main.c
vmg: Done!

vmg:
Tests added and formatting fixed. 👌 

You will need to install `libluajit-dev` on the buildbot, otherwise the tests are skipped, since `bcc-lua` is not built.

drzaeus77:
I did install luajit-dev on the buildbots, but looks like they're still not being picked up. Ubuntu gets:
```
-- Found LuaJIT: /usr/lib/x86_64-linux-gnu/libluajit-5.1.a;/usr/lib/x86_64-linux-gnu/libdl.so;/usr/lib/x86_64-linux-gnu/libm.so  
```
but later:
```
23: + echo 'bcc-lua not built --- skipping'
23: bcc-lua not built --- skipping
```

And Fedora gets stuck earlier at:
```
-- Could NOT find LuaJIT (missing:  LUAJIT_LIBRARIES) 
```


For reference:
```
$ dpkg -l | grep luajit
ii  libluajit-5.1-2:amd64              2.0.2+dfsg-1                     amd64        Just in time compiler for Lua - library version
ii  libluajit-5.1-common               2.0.2+dfsg-1                     all          Just in time compiler for Lua - common files
ii  libluajit-5.1-dev:amd64            2.0.2+dfsg-1                     amd64        Just in time compiler for Lua - development files
ii  luajit                             2.0.2+dfsg-1                     amd64        Just in time compiler for Lua programming language version 5.1
```

```
$ rpm -qa | grep luajit
luajit-devel-2.0.3-4.fc22.x86_64
luajit-2.0.3-4.fc22.x86_64
```

tests/lua/test_standalone.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+# Copyright (c) GitHub, Inc.
+# Licensed under the Apache License, Version 2.0 (the "License")
+
+set -xe
+cd "src/lua"
```
4ast: are you building in-src ? may be that's why buildbots don't pick it up?

vmg:
@drzaeus77: I'm afraid Fedora does not ship a static `libluajit.a`, so it's not possible to build or test the standalone tool there. :sweat:

Any ideas on how to work around this? Or shall we stick to testing the tool in Ubuntu?

vmg:
@drzaeus77: Also, the buildbot cannot test the standalone tool in ubuntu because you're running a sub-makefile:

```
argv: ['make', '-j2', '-C', 'tests/cc']
```

If you run `make -j2` instead, the standalone tool will be built too (not only the C tests).

drzaeus77:
> If you run make -j2 instead, the standalone tool will be built too (not only the C tests).

Well, this is actually intentional. If the tool were part of the packaging, it should be available to the test scripts. That way we actually test what goes into the end-user system. If we forgot some bits, we will catch it here.

If it's premature to add your binary to the deb/rpm, a workaround would be to build the static target in tests/lua, and I can update the buildbot to do `make -C tests`.

I'm fine either way (make -C tests or packaging lua binary tool in the release).

drzaeus77:
> Any ideas on how to work around this? Or shall we stick to testing the tool in Ubuntu?

I could build from source on the buildbot.

vmg:
> Well, this is actually intentional. If the tool were part of the packaging, it should be available to the test scripts. That way we actually test what goes into the end-user system. If we forgot some bits, we will catch it here.

Oh, interesting. Thanks for pointing that out. I think the standalone binary should be safe to ship in the list. Could you give me a brief overview of what changes to I need to make that happen?

> I could build from source on the buildbot.

That sounds great.

drzaeus77:
> Could you give me a brief overview of what changes to I need to make that happen?

* add an `install()` line to your src/lua/CMakeLists.txt
* add a new package in `SPECS/bcc*.spec`
* add a new .install file in debian/

vmg:
@drzaeus77: Done! I think `bcc-lua` should now be part of a normal installation, and a `bcc-lua` package has been added to the specs. How is this looking?

debian/bcc-lua.install
```diff
@@ -0,0 +1 @@
+usr/bin/bcc-lua
```
drzaeus77: Also needed is an entry in debian/control

tests/lua/test_standalone.sh
drzaeus77: With the latest addition of packaging, bcc-lua can come either from a package or in-tree. Both should continue to work. Probably what you want is like we do with the wrapper.sh, which is to update the PATH accordingly, and here just check for `which bcc-lua`.

debian/bcc-lua.install
vmg: Done!

drzaeus77:
This version looked great! There are still some nice-to-haves (libbcc.a), but that is completely fine as a follow-on if someone gets around to it.

PR #474, opened 2016-04-02
migrated gethostlatency to use bpf_perf_event

mcaleavya:


brendangregg:
haven't run it yet, but the code LGTM.

PR #477, opened 2016-04-05
Add support for nixos

ragnard:
Two changes that enables bcc to build and work nicely on nixos:

- add `-I/build/include/generated` when calculating cflags. This is where `autoconf.h` with `#define CONFIG_BPF_SYSCALL 1` lives so it is critical. This appears to be true not only for nixos, but I checked Ubuntu 14.04.3 LTS too and the location was the same.
- bcc quite reasonably assumes that the kernel modules dir is `/lib/modules/`, encoded using `#define KERNEL_MODULES_DIR "/lib/modules"`. This is not true for nixos however. The proposed patch:
  - introduces a `BCC_KERNEL_MODULES_DIR` cmake flag
  - defaults that flag to `/lib/modules`, so no breaking change
  - sets `KERNEL_MODULES_DIR` based on that flag.

  This allows us to override `KERNEL_MODULES_DIR` when building bcc for nixos.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,12 @@ FOREACH(DIR ${LLVM_INCLUDE_DIRS})
   include_directories("${DIR}/../tools/clang/include")
 ENDFOREACH()
 
+# Set to a string path if system places kernel lib directory in
+# non-default location.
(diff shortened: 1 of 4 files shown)


4ast:
lgtm

drzaeus77:
Thanks!

Issue #478, opened 2016-04-06
attach_uprobe fails to resolve symbols in some cases


LucaCanali:
attach_uprobe fails to resolve symbols in some cases, in particular when trying to resolve functions placed in subsections text.hot and text.unlikely. This is an example of this issue in a test I did with tracing function kskthewt in the oracle binaries:
```
Traceback (most recent call last):
  File "./my_uprobeora.py", line 43, in <module>
    b.attach_uprobe(name="/u01/app/oracle/product/11.2.0.4/rdbms/bin/oracle", sym="kskthewt", fn_name="printval")
  File "/usr/lib/python2.7/site-packages/bcc/__init__.py", line 512, in attach_uprobe
    (path, addr) = BPF._check_path_symbol(name, sym, addr)
  File "/usr/lib/python2.7/site-packages/bcc/__init__.py", line 490, in _check_path_symbol
    raise Exception("could not determine address of symbol %s" % sym)
Exception: could not determine address of symbol kskthewt
```

The issue can be tracked down to "find_symbol method" defined in `src/python/bcc/__init__.py` and seems easy to fix. One possible fix that I tested is to change (in line 467) 
`awk -v sym=%s '$NF == sym && $4 == ".text"  \` with
 `awk -v sym=%s '$NF == sym && ($4 == ".text" || $4 == "text.hot" || $4 == "text.unlikely")  \`

Another possible option is to switch to using nm -D instead of objdump -tT, with something like:
```
        with os.popen("""/usr/bin/nm -D %s | \
                awk -v sym=%s '$NF == sym && $2 == "T"  \
                { print $1; exit }'""" % (path, sym)) as f:
```

drzaeus77:
The first option looks good to me. Would you like to send a PR?

LucaCanali:
Thanks for the quick reaction on this.
Luca

PR #479, opened 2016-04-06
Fix for find_symbol method, issue #478

LucaCanali:
As discussed in issue #478 this is a proposed fix for "find_symbol method" to solve the issue that attach_uprobe fails to resolve symbols in some cases, in particular when trying to resolve functions placed in subsections text.hot and text.unlikely.

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -464,7 +464,8 @@ class BPF(object):
             return symbols[sym]
 
         with os.popen("""/usr/bin/objdump -tT %s | \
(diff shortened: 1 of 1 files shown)


src/python/bcc/__init__.py
```diff
@@ -464,7 +464,7 @@ def find_symbol(cls, path, sym):
             return symbols[sym]
 
         with os.popen("""/usr/bin/objdump -tT %s | \
-                awk -v sym=%s '$NF == sym && $4 == ".text"  \
+                awk -v sym=%s '$NF == sym && ($4 == ".text" || $4 == "text.hot" || $4 == "text.unlikely") \
```
drzaeus77: Can you line wrap here to keep under 80 columns?

drzaeus77:
Thanks!

PR #481, opened 2016-04-09
bcc-lua: switch to quiet by default

4ast:
use --verbose to turn verbose output


vmg:
Yes, I agree. The noisy output was useful while developing but I also find myself always using `-q` when testing probes in production.

4ast:
Thanks. Pushing then.

PR #482, opened 2016-04-10
bcc-lua: build bcc.lua and bcc.o in the build tree

markdrayton:
The squishing/luajiting process during the build of bcc-lua writes a couple of files (`bcc.lua` and `bcc.o`) into the source tree. This change moves these files into the build tree so the source tree is untouched, consistent with the rest of the BCC build.

I'm not sure if this is the right way to do things (I've never touched cmake) as it looks like the rest of the build artifacts are written into `src/lua/CMakeFiles/bcc-lua.dir/src`:

    mark@ubuntu:~/bcc/build$ find src/lua/ \( -name bcc.lua -o -name '*.o' \)
    src/lua/CMakeFiles/bcc-lua.dir/src/main.c.o
    src/lua/bcc.lua
    src/lua/bcc.o

It's not immediately obvious how (or if I should) write these files into `src/lua/CMakeFiles/bcc-lua.dir`. Suggestions welcome if this isn't right (cc @vmg).

### DIFF
--- a/src/lua/CMakeLists.txt
+++ b/src/lua/CMakeLists.txt
@@ -5,21 +5,19 @@ if (LUAJIT_LIBRARIES)
 	FILE(GLOB_RECURSE SRC_LUA ${CMAKE_CURRENT_SOURCE_DIR}/bcc/*/*.lua)
 
 	ADD_CUSTOM_COMMAND(
-		OUTPUT ${CMAKE_CURRENT_SOURCE_DIR}/src/bcc.lua
-		COMMAND ${LUAJIT} src/squish.lua
+		OUTPUT bcc.lua
+		COMMAND ${LUAJIT} ${CMAKE_CURRENT_SOURCE_DIR}/src/squish.lua ${CMAKE_CURRENT_SOURCE_DIR}
 		DEPENDS ${SRC_LUA} ${CMAKE_CURRENT_SOURCE_DIR}/squishy
-		WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
 	)
 
 	ADD_CUSTOM_COMMAND(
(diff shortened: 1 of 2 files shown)


drzaeus77:
The approach looks ok to me.

vmg:
This looks like a nice change, but I'm afraid it breaks the build dependencies. Editing any of the lua files now doesn't cause the `bcc-lua` target to rebuild!

Let me see if I can figure out what exactly went wrong. :)

markdrayton:
Hrm, did my change break this or was it already broken? I checked out the commit before my change and it seems the problem of changed dependencies not rebuilding `bcc-lua` existed then too:

    $ cd ~/bcc
    $ git checkout c98ffd634ffa5c795e617e7fc0b2297709bed3b0
    $ mkdir build; cd build
    $ cmake .. -DCMAKE_INSTALL_PREFIX=/usr
    $ make
    $ md5sum src/lua/bcc-lua 
    24418d605f0d6038fab1a77b49b7c2c1  src/lua/bcc-lua
    $ vim ../src/lua/bcc/table.lua
    $ git diff
    diff --git a/src/lua/bcc/table.lua b/src/lua/bcc/table.lua
    index 094a783..fcefd58 100644
    --- a/src/lua/bcc/table.lua
    +++ b/src/lua/bcc/table.lua
    @@ -377,7 +377,7 @@ local function NewTable(bpf, name, key_type, leaf_type)
         leaf_type = _decode_table_type(desc)
       end
     
    -  log.info("key = %s value = %s", key_type, leaf_type)
    +  log.info("key = %s value = %s, id = %s", key_type, leaf_type, id)
       return table:new(bpf, id, fd, key_type, leaf_type)
     end
     
    $ make
    $ md5sum src/lua/bcc-lua                                                                                                                                      
    24418d605f0d6038fab1a77b49b7c2c1  src/lua/bcc-lua

PR #487, opened 2016-04-14
Install all required packages on ubuntu

bobrik:
* `binutils` is needed for `objdump`
* `bcc` and `bcc-tools` are installed on other distros


4ast:
lgtm @drzaeus77

Issue #488, opened 2016-04-15
bcc-lua: 64 bit addresses and Lua numbers


markdrayton:
I ported `stacksnoop` to Lua. It was easy and works but its stacks are misleading:

    146355.560800000   submit_bio
            ffffffff813bd800 blk_init_queue
            ffffffff812b1800 ext4_writepages
            ffffffff811a3800 do_writepages
            ffffffff8124e000 wbc_detach_inode
            ffffffff8124e800 writeback_inodes_wb.constprop.59
            ffffffff8124e800 writeback_inodes_wb.constprop.59
            ffffffff8124f000 wb_start_writeback
            ffffffff8124f800 dirtytime_interval_handler
            ffffffff8109c800 create_worker
            ffffffff8109d000 rescuer_thread
            ffffffff810a3000 flush_kthread_work
            ffffffff8183a000 _raw_read_lock

Note the addresses all end `00` (and the symbols are probably wrong). This is because Lua's [number type is double](http://www.lua.org/pil/2.3.html) and can't represent all > 52 bit integers:

    $ luajit -e 'print(string.format("%x", tonumber("20000000000000", 16)))'
    20000000000000
    $ luajit -e 'print(string.format("%x", tonumber("20000000000001", 16)))'
    20000000000000
    $ luajit -e 'print(string.format("%x", tonumber("20000000000002", 16)))'
    20000000000002

@vmg, do you have any handy workarounds planned for this? It looks like [Lua 5.3 has a 64 bit integer type](http://www.lua.org/manual/5.3/readme.html#changes) but as far as I can see that hasn't made it to LuaJIT.

Other than this the Lua integration is amazing :)

vmg:
Well, that throws quite a wrench in my plans. Thanks for bringing this to my attention... Let me research and figure out how can we work around this. I can't believe I didn't spot the issue during production testing. :sweat:

> Other than this the Lua integration is amazing :)

😍

vmg:
OK, I have an attack plan. I'll work on this during the weekend. :)

vmg:
This was less painful than I thought! Check out https://github.com/iovisor/bcc/issues/490

PR #490, opened 2016-04-17
lua: Properly support high-range 64 addresses

vmg:
Lua's native Number type is a 64-bit double with 52-bit precision, which
was causing rounding errors when storing and working with high-range
memory addresses (namely, addresses from the kernel, which are all in
the `0xffffffff........` range).

To work around this, we've made sure to never call `tonumber` on any
variables that represent memory addresses, and instead continue
operating on them with their native types: LuaJIT can work with the
underlying `uint64_t` type for these values and transparently perform
all kinds of numeric operations.

The only limitation of working with native 64-bit types in LuaJIT is
that they cannot be printed with the language's default `string.format`
API. To give better UX to probe writers, these APIs have been
monkeypatched so the `%p` format specifier will now properly handle
64-bit addresses and print them in an appropriate format.

Fixes https://github.com/iovisor/bcc/issues/488

### DIFF
--- a/src/lua/bcc/vendor/helpers.lua
+++ b/src/lua/bcc/vendor/helpers.lua
@@ -1,3 +1,29 @@
+do
+  local ffi = require("ffi")
+  local ptrtype = ffi.typeof("uint64_t")
+  local strformat = string.format
+  function string.format(format, ...)
+    local args = {...}
+    local match_no = 1
+    local newfmt, count = string.gsub(format, "()%%(.-)(%a)",
+      function(_, mods, t)
+        local n = match_no
+        match_no = match_no + 1
+        if t == 'p' and ffi.istype(ptrtype, args[n]) then
+          local lo = tonumber(args[n] % 4294967296ULL)
+          local hi = tonumber(args[n] / 4294967296ULL)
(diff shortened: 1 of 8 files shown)


src/lua/bcc/vendor/helpers.lua
```diff
@@ -1,3 +1,29 @@
+do
+  local ffi = require("ffi")
+  local ptrtype = ffi.typeof("uint64_t")
+  local strformat = string.format
+  function string.format(format, ...)
```
vmg: This is a bit of Rocky's Horror Show, but IMO it's worth it to let probe authors use `%p` like they'd expect to.

4ast: fancy magic.

src/lua/bcc/vendor/posix.lua
```diff
@@ -62,9 +64,15 @@ local function cpu_count()
   return tonumber(ffi.C.get_nprocs())
 end
 
+local function tonumber64(n, base)
+  assert(type(n) == "string")
+  return ffi.C.strtoull(n, nil, base or 10)
```
drzaeus77: Would you want `base or 0` to invoke strtoull's automatic base logic? Although so far everybody is calling it with base=16.

PR #495, opened 2016-04-19
[RFC] Remove all dependencies on external binaries

vmg:
In its current implementation, BCC depends on a large set of external binaries for normal operation. Namely:

- `objdump` for symbol lookup
- `objdump` for load address resolution
- `readelf` for USDT notes parsing
- `ldconfig` for system library resolution
- `file` and `magic` to separate ELF binaries and SOs
- `readlink` to follow symlinks

This approach has IMO several shortcomings, amongst them:

- They add a large layer of runtime dependencies for the library and bindings which are not exactly specified anywhere (and may or may not be assumed to exist on production systems).
- The textual interface to every one of these tools is ill-defined. This makes parsing their output require regexps, which need to be continuously tweaked and which may break between versions of the tools.
- The boilerplate to interact with these tools is cumbersome, and needs to be re-implemented by every author of bindings to `libbcc`.
- They are *slow* to work with, since they are really generic utilities that process much more information than we require (particularly noticeable for `readelf` and `objdump`).

This work-in-progress RFC/Pull Request proposes an alternative take on these dependencies: they have all been replaced with native C/C++ code, inside of `libbcc` itself, which makes the APIs readily available for all existing and future binding implementations without having to re-implement any of the parsing boilerplate.

The new programmatic APIs have (presumingly) the same behavior as their binary counterparts, with none of the cruft. A new external dependency, `libelf`, has been added to libbcc, but this is a development dependency that can be linked statically and should be readily available in all major distributions.

Work so far has been focused on removing the ProcessSymbols and KernelSymbols caches in both the Python and Lua wrappers, and unifying them with a native SymbolCache class, accessed through C APIs. The `check_path_symbol` API, which resolves libraries and hook addresses for the probes has also been re-implemented in C, again unifying the code between the Python and Lua wrappers.

A very advanced and performant API for USDT probe resolution based on libelf has also been written (removing the dependency on `readelf` and the janky parsing), although right now it has not been wired up to either wrapper. Likewise for a `file`/`magic` replacement for shared-object detection.

There's still a bit of work to be done (particularly when testing this new implementation), but I'd like to open this PR early to gather feedback and see whether the maintainers deem this approach interesting. I'm hoping that the chance to unify large swaths of code for all `libbcc` wrappers will be as appealing to you as it is to me!

As always, all the work here has been broken up in small meaningful comments, so please check the commit messages for specifics. Thanks for your consideration!


src/cc/bcc_elf.c
```diff
@@ -0,0 +1,269 @@
+#include <sys/types.h>
```
4ast: I like the direction a lot. Please add apache2 banner to all new .h and .c
Also we need to agree on formatting of .c files. Take a look at existing style.
I don't think it's pretty (I prefer kernel style with tabs), but that's what bcc is using already. So better be consistent. @drzaeus77 we need pre-commit lint checker.

src/cc/bcc_helpers.h
```diff
@@ -0,0 +1,53 @@
+#ifndef LIBBCC_ELF_H
+#define LIBBCC_ELF_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <stdint.h>
```
4ast: unnecessary?

src/cc/bcc_syms.cc
```diff
@@ -0,0 +1,258 @@
...
+	struct Symbol {
+		Symbol(const char *name, uint64_t addr) :
+			name(name), addr(addr) {}
+		std::string name;
+		uint64_t addr;
+
+		bool operator<(const Symbol& rhs) const { return addr < rhs.addr; }
+	};
+
+	std::vector<Symbol> _syms;
```
4ast: what's the meaning of _ ? @drzaeus77 we need a coding style doc.

vmg:
@drzaeus77: Could you install `libelf-dev` on the buildbots? I believe that should make the build green and displaying the new features. :)

drzaeus77:
@vmg Can you first add a rpm/deb builddep on libelf-dev? Ideally it would have been caught before cmake got to run.

src/cc/bcc_syms.cc
vmg: Member variables are prefixed with `_` in some coding standards (particularly the one we use internally). Other standards are `m_` or suffixed with `_`. The point is to differentiate them from locally-declared variables.

...This is obviously arbitrary. I think a clang-formatter file for the repository would be neat, with whichever standard you deem more appropriate.

vmg:
@drzaeus77: I've added `libelf-dev` to `bcc.spec` -- is that what you had in mind?

src/cc/bcc_syms.cc
drzaeus77: I'm more used to the google C++ style guide, which is a bit old but easy to verify (single python script). I am considering adding their cpplint.py to our tree and wrapping it in a ctest.

vmg: Sadly Google's Python script does not work well with plain C. Is there a way we could specify Google's C++ rules through `clang-format`? That would also work well for C.

src/cc/bcc_helpers.h
```diff
@@ -0,0 +1,53 @@
+#ifndef LIBBCC_ELF_H
```
4ast: ifdef doesn't match file name. Should it be bcc_elf.h ?
bcc_helpers.h sounds too abstract and eventually other non-elf things
will start piling up in here.
Is it possible to add some simple c+c++ demo?
Everything so far is c+py or c+lua with focus of 'extern C' api
while c++ api is little bit of afterthought. 
With this new addition of elf api, something like offwaketime in c+c++ is now possible, would be really useful and it will help to stabilize c/c++ API.

drzaeus77:
@vmg debian/control update is also needed.

src/cc/bcc_syms.cc
vmg: > When using clang-format command line utility or clang::format::reformat(...) functions from code, one can either use one of the predefined styles (LLVM, **Google**, Chromium, Mozilla, WebKit) or create a custom style by configuring specific style options.

It seems like the ideal choice here. Particularly since this whole project is already based on LLVM.

vmg:
> dpkg-buildpackage: host architecture amd64
> dpkg-checkbuilddeps: Unmet build dependencies: libelf-dev
> dpkg-buildpackage: warning: build dependencies/conflicts unsatisfied; aborting
> dpkg-buildpackage: warning: (Use -d flag to override.)

Done. Looks like it picked this up alright! Now we gotta actually install the package. :)

SPECS/bcc+clang.spec
```diff
@@ -14,7 +14,7 @@ Source1:        http://llvm.org/releases/%{llvmver}/llvm-%{llvmver}.src.tar.xz
 Source2:        http://llvm.org/releases/%{llvmver}/cfe-%{llvmver}.src.tar.xz
 
 BuildArch:      x86_64
-BuildRequires:  bison, cmake >= 2.8.7, flex, gcc, gcc-c++, libxml2-devel, python2-devel
+BuildRequires:  bison, cmake >= 2.8.7, flex, gcc, gcc-c++, libxml2-devel, python2-devel, libelf-dev
```
drzaeus77: Looks like the name should be elfutils-libelf-devel

vmg: That is not an Ubuntu package. I'm afraid I'm a bit lost here -- where are you getting these packages from? On Debian and Ubuntu, we'd require `libelf-dev`.

drzaeus77: .spec is for rpm based systems, the debian/control file is perfectly fine

drzaeus77: Also, looking at the details of elfutils-libelf-devel, this is going to be an extra runtime dependency as well. Might need to switch it to elfutils-libelf-devel-static.

vmg: Got it. I think the RPM and Deb requirements should be right now. Sorry for the fuss!

vmg:
It's green! I'm going to work on fixing the C/C++ style tomorrow. Thanks for your patience. :)

vmg:
I've fixed the formatting, and added a `.clang-format` file (which basically says "use Google's style"). I do agree with @4ast that this 2-space style is awful looking and the Kernel style beats it in every way but heeeeey I'm not here for the Holy War.

I've also added license headers and split up/reorganized the header files to make them clearer and more obvious.

> Is it possible to add some simple c+c++ demo?

I don't think so... There's still a _very_ significant amount of boilerplate in Lua and Python that we'd have to port over. Would that be worth it? I'm not going to write probes in C++ -- the Lua wrapper is essentially just as fast as a native C++ probe anyway, which is one of the main reasons I wrote it. I could write a "hello world" in C++ if that really interests you, but it's most likely not going to cover all the C APIs we're exporting, so it won't do much for API stabilization.

I think a better use of our time would be porting the USDT code to C++, so it can be used for Python and Lua at the same time. I've already started working on that (in fact, this PR is just a yak shave to allow me to port all the USDT parsing code into C++ with pretty APIs -- I want to use USDTs from Lua probes).

Assuming you're happy with the quality of this implementation, I can wrap up this PR by adding some more tests for the C++ code (particularly for Symbol resolution, I want to make sure we're getting this right).

src/cc/bcc_elf.c
drzaeus77: I don't mind the kernel style for .c files either, but I would really rather use it only when in a pure C project. Mixing C and C++ files with different formats is more aggravating that suffering with a lesser format, IMO.

4ast:
@vmg agree that usdt to c++ is higher priority.
but not sure about 'significant amount of boilerplate' to make c++ usable. Once there is ksym, usdt, elf apis, the only thing left will be key/value api, but it cannot be auto-generated dynamically for c++ anyway, so c++ would have to copy-paste structs between c and c++. Not a concern to me.
What we get out of proper c++ support is that large services that already written in c++ can start using bcc apis. Few folks actually tried to use existing c++ api, but it was too hard for newcomers and lack of demo programs didn't help.

As far as is PR, I think it's ready, but it can silently break existing tools, so the tests are strongly encouraged :)

vmg:
Yoohoo! So I've added quite a few tests for this new functionality, and already caught a (serious) bug in that was present in the old Python implementation: Address->Symbol resolution would fail for a lot of loaded modules because we were detecting many modules as executables instead of shared objects (hence we weren't adjusting the symbol address based on the load address for the SO). The issue is that we were checking if a module ended in `.so`, when the module could very well end in `.so.X.X` for versioning.

Anyway, tests are extensive and green. I think this is good to merge. :)

tests/cc/sput.h
```diff
@@ -0,0 +1,272 @@
+/*
+ *  sput - Simple, Portable Unit Testing Framework for C/C++ v1.3.1
```
4ast: interesting. I thought we're already using gtest, but since not, I think this one is preferred, since it's so tiny.

tests/cc/test_c_api.c
```diff
@@ -0,0 +1,116 @@
...
+
+  sput_fail_unless(
+      bcc_symcache_resolve(resolver, (uint64_t)libc_fptr, &sym) == 0,
+      "resolve a function in libc in our current process");
+
+  sput_fail_unless(
+      sym.module && sym.module[0] == '/' && strstr(sym.module, "libc"),
+      "resolve a function to linked libc module");
+
+  sput_fail_unless(strcmp("strtok", sym.name) == 0,
```
4ast: nice tests. thanks

PR #496, opened 2016-04-19
Add infra and ctest for code style formatting

drzaeus77:
This change adds a test for ensuring that changed files conform to some form of style guideline. Currently, that style is LLVM's.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,4 @@
+---
+BasedOnStyle: Google
+AllowShortIfStatementsOnASingleLine: false
+IndentCaseLabels: false
--- a/scripts/git-clang-format
+++ b/scripts/git-clang-format
@@ -1,0 +1,485 @@
+#!/usr/bin/env python
+#
+#===- git-clang-format - ClangFormat Git Integration ---------*- python -*--===#
+#
+#                     The LLVM Compiler Infrastructure
+#
+# This file is distributed under the University of Illinois Open Source
+# License. See LICENSE.TXT for details.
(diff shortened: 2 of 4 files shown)


4ast:
I'm missing something. Where is actual format spec? Once this pr lands than all non-llvm formated bits will be failing?

drzaeus77:
The format spec is LLVM (not Google) defaults. I already started working on a few tweaks, and will take Vicent's .clang-format as well. The test should run only against .c, .h, and .cc files. Others will give incorrect results, especially python files.

And no, only if you touch a file will clang-format check it. That's why it's passing now, because nothing is in the changeset. Actually, it would be a good test, if you make a trivial change to some c/cc file and send a PR for it. It should flag it and give some helpful output.

4ast:
Then Vincent's clang-format should be part of the diff as well, otherwise interim other contributors will be confused?

drzaeus77:
Yes yes, I'm working on it. Don't review this one any further until the next update.

drzaeus77:
@4ast updated

.clang-format
```diff
@@ -0,0 +1,4 @@
+---
+BasedOnStyle: LLVM
```
4ast: should it be google?

drzaeus77: You're right, thinking about it a second time I hate their capitalization rules. Switching back to google.

4ast:
hope it's a step in the right direction. If it causes too much pain, we can reconsider.

PR #498, opened 2016-04-21
[WIP] Native implementation for USDT probes

vmg:
Opening this PR early, as I start porting the USDT probe implementation (`src/python/bcc/usdt.py`) to libbcc itself.

As @4ast pointed out on the previous PR, we don't use GTest in the library -- in fact, we didn't have a test framework for C/C++ because we didn't have any C/C++ tests (do note that the behavior of the library is quite thoroughly tested through the Python and Lua tests, however). When porting over the Elf and Process APIs into C, I added `sput` as a small test framework for some basic tests.

Now, whilst writing more native code, it's becoming obvious that `sput` doesn't quite cut it (it's a very basic test framework), so I've decided to go with Catch (https://github.com/philsquared/Catch), which is the header-only framework I usually use for heavy C++ testing because:

1. It's lighter than google-test
2. The assert/expect interface is fucking gorgeous (it decomposes the statements to show you the actual test failure).
3. It's a single-header so it can be drag and dropped into projects

Consequently, I've ported the existing tests for the C API to Catch, and added a new test suite for my ongoing USDT work.

So far, I've only implemented parsing the probe arguments (more work on the way). I decided to hand-write a small parser because the original regexps used in the Python implementation were missing a lot of corner cases. I think this new implementation is more accurate and not overly complex -- I'm of course open to feedback.

I'll keep updating the PR as I implement more USDT functionality natively.


tests/cc/test_c_api.cc
```diff
@@ -0,0 +1,95 @@
+#include <stdint.h>
```
4ast: sorry to be picky, but please add the license banner.
it matters for big companies adopting the software.

Disappointing that sput didn't last a week.
What are the chances that 'catch' will last longer?
I think it makes sense to move to gtest and stick with it.

4ast:
hand coded parser looks good, but gtest vs catch arguments don't resonate. Don't see a concern with gtest as a dependency whereas 10k lines copy-paste of catch is not pretty.

vmg:
> Don't see a concern with gtest as a dependency whereas 10k lines copy-paste of catch is not pretty.

Are you sure about this? GTest is a very heavy dependency and Google strongly advises against using distribution packages (see https://github.com/google/googletest/blob/master/googletest/docs/FAQ.md) for application tests. Ideally we'd have to submodule it, or paste it like we did with Catch.

I don't feel too strongly about this, but Catch is well tested and used in production by a lot of people (including us). It works very well and the assertion API is miles ahead of GTest's IMO. It is designed to be vendored on projects as a single header, so I particularly find it to be much prettier and usable than GTest. GTest also does some funky stuff with object isolation (to prevent crashes from affecting other tests). I'm not sure at all of how is this going to affect our local symbol resolution tests.

I know I made a mistake bringing `sput` so early in the development, when it clearly didn't have enough features to test C++ code. But I think Catch fills all our needs for for foreseeable future and saves us from the dependency hell of GTest. Is this something you're willing to reconsider?

Thanks! :)

drzaeus77:
I think Catch is fine. I've had my time with gtest and am ready for something different.

vmg:
Regarding gtest: I've been evaluating how to vendor the library. It's not pretty. For starters the Debian/Ubuntu packages for GTest don't ship built libraries (following Google's recommendation). See: http://askubuntu.com/questions/145887/why-no-library-files-installed-for-google-test

This means we cannot depend on them to build `bcc`. The alternatives are vendoring all of GTest (a few megabytes of data), or the GTest amalgamation (which is about 5x larger than Catch). I think this whole situation makes Catch significantly more appealing.

Regarding the Fedora test suite: why is it crashing? Simple (well, not really simple to figure out, but let me explain). It's crashing because it runs with two versions of the C++ STD.

This is easy to visualize if we run the test suite through Valgrind on Ubuntu (where it does finish without crashing).

```
==17771== Mismatched free() / delete / delete []
==17771==    at 0x4C2F24B: operator delete(void*) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==17771==    by 0x400C13: USDT::Argument::~Argument() (usdt.h:23)
==17771==    by 0x400AB3: main (simpletest.cc:14)
==17771==  Address 0x8b7f5b0 is 0 bytes inside a block of size 31 alloc'd
==17771==    at 0x4C2DB8F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==17771==    by 0x69AC707: operator new(unsigned long) (in /home/vmg/src/bcc/build/src/cc/libbcc.so.0.1.8)
==17771==    by 0x6A1EDC8: std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_mutate(unsigned long, unsigned long, char const*, unsigned long) (in /home/vmg/src/bcc/build/src/cc/libbcc.so.0.1.8)
==17771==    by 0x6A1F93A: std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_replace(unsigned long, unsigned long, char const*, unsigned long) (in /home/vmg/src/bcc/build/src/cc/libbcc.so.0.1.8)
==17771==    by 0x51558C7: USDT::ArgumentParser::parse_identifier(long, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) (usdt_args.cc:33)
==17771==    by 0x5155AE7: USDT::ArgumentParser::parse_expr(long, USDT::Argument&) (usdt_args.cc:64)
==17771==    by 0x5155C0E: USDT::ArgumentParser::parse_1(long, USDT::Argument&) (usdt_args.cc:83)
==17771==    by 0x5155D1E: USDT::ArgumentParser::parse(USDT::Argument&) (usdt_args.cc:100)
==17771==    by 0x400A8D: main (simpletest.cc:13)
```

You can see how we're working with `std::string`s inside of `libbcc`... But look at the actual `operator new` that the STD uses to allocate internal storage for the strings. It's mapped in `libbcc.so`, not in `libstdc++`! So then, when we free the allocated string *from our actual program that links against `libbcc.so`*, it's using the `operator delete` from `libstdc++`.

The underlying case is obvious in retrospect: we're building `libcc.so` with a statically linked `libstdc++`, which means that any applications that use any of the C++ APIs exported by the library will crash! Our tests are obviously one of those applications.

We have several possible ways to fix this:

1. Stop linking `libstdc++` statically by default (this "feature" can be hidden behind a flag for people who want to redistribute truly standalone `libbcc`).
2. Change our test suite so it doesn't *link* against `libbc.so` but instead builds the relevant `.cc` files from scratch for the test binary. That will fix the underlying issue. Not very elegant though.
3. Write clean C wrappers for these C++ APIs. If the test suite only interacts with `libbcc.so` through C APIs, the two implementations (static vs system's) of `libstdc++` will not collide. This is kinda lame because these C++ APIs are meant to be internal -- we'd have to build C wrappers just for testing, and these wrappers wouldn't be used by the bindings, since the bindings require higher level APIs.
4. Some `g++` flag that I'm missing so we can compile our test suite reusing the `libstdc++` that has already been linked statically in `libbc.so`. This would be neat -- I'm not sure it is possible.

I'm leaning towards option 1 because it's the simplest and safest. Thoughts?

4ast:
I wasn't aware of this gtest recommendation. Ok. let's try catch.

4ast:
missing something... why on fedora /lib/stdc++.so used by test program is different from the one statically linked in libbcc.so ?

vmg:
@4ast: It happens in all OSes. It's because one is statically linked, but the other is not. Same code, different allocation paths.

4ast:
yes, it can happen everywhere. The question why it happened on our fedora buildbot. The usdt_test should have used the same compiler with the same flags as libbcc... so even if std:string of libstdc++ is using global objects (which I think it's not), it should still work even though 'one-def rule' is potentially violated.

vmg:
I think this goes beyond the "one definition rule". It's using different allocation pools between two implementations. Even if it is "the same code", the system's `libstdc++` cannot `delete` an object that was allocated by the statically linked `libstdc++` in `libbcc`. I don't know the specifics of what tracking does `libstdc++` do on top of `malloc` for the C++ allocators, but it's really clear if you try to hook `malloc` (i.e. with valgrind) that this code can and will crash in all OSes.

4ast:
I don't think so. operator new is a simple wrapper on top of malloc with extra _cxa_throw. There is no special pool. The code doesn't crash on ubuntu and crashes on fedora, right? So there must be something else.

4ast:
it could be that Catch is doing something special, but then it's a Catch problem.

drzaeus77:
I may have a solution for this, but currently am away from my code. The
change would be to move the con/destructor for the Argument class into the
cc file, as well as to prevent the client library from allocating strings
outside. The strings would be private members and only expose themselves
through const &. Besides, I don't like the use of non const reference
passing that is going on here.

drzaeus77:
It's not catch, its normal code, though it didn't crash on my machine,
valgrind did complain.

drzaeus77:
Please take a look at https://github.com/iovisor/bcc/commit/cf2ed1be388a300dfc643ed91b39615a200e2627, which fixes the valgrind complaint for me.

4ast:
it's good to fix valgrind, but it looks like valgrind is trying to ld_preload 'operator delete' and complaining.  I think it's a false positive, since 'operator delete' of libbcc is statically linked, and the cause of segfault is something else. Can we try asan the whole thing?

vmg:
@drzaeus77: Thanks for that patch. I've applied it on top of this branch and the Fedora build is now green. Valgrand and ASAN don't complain either.

@4ast: I am not sure at this point, but I think Valgrind was onto something. Looking at the code for GNU's `libstdc++`, it makes no sense to me either (because `new` and `delete` are indeed thin wrappers around `malloc` and `free`). But the fact is that hiding the `std::string` allocation from the header fixes all of Valgrind's complaints and the Fedora crash, *and/or* removing the static link for `libstdc++` also fixes Valgrind and the crash.

So this has something to do with statically linking `libstdc++`. I have a feeling there's something wrong with the static `libstdc++` that ships with Fedora... But since @drzaeus77's patch fixes the issue and is quite reasonable, I'm just going to proceed with this. :)

tests/cc/test_c_api.cc
```diff
@@ -0,0 +1,95 @@
+#include <dlfcn.h>
```
4ast: license banner please. other than that it looks good.

vmg:
Yey! Green! So the C++ implementation is "finished" (modulo your kind review feedback --thanks in advance). And it's quite thoroughly tested on the new test suite. I'm gonna start wrapping it up (Python first, I guess) and see how's this looking. And then I'll start with @brendangregg's suggestions.

@drzaeus77: Some notes for the buildbot: as you can see the build is green, but the most important USDT tests are being skipped right now. The dependencies are:

- If the test suite finds `ruby` on the path, it will run a quite extensive coverage test for Ruby's USDTs. I picked Ruby as an example of "production software that is widely available and comes with USDT probes". So installing a new version of Ruby (2.3) on the buildbot will enable a bunch of new USDT tests. I know Node.js could have been another candidate, but Ruby felt lighter.

- If the test suite finds the `systemtap-sdt-dev` Ubuntu package (or Fedora equivalent) when compiling, it will insert USDT probes on the test suite itself, and "self-test". It's pretty cool. Works well in practice.

So, yeah. Feedback welcome. Thanks!

tests/cc/test_usdt_probes.cc
```diff
@@ -0,0 +1,219 @@
...
+    char *const argv[2] = {_ruby, NULL};
+
+    ChildProcess ruby(argv[0], argv);
+    if (!ruby.spawned())
+      return;
+
+    USDT::Context ctx(ruby.pid());
+    REQUIRE(ctx.num_probes() >= mri_probe_count);
+
+    SECTION("get probe in running process") {
```
4ast: nice tests

src/cc/usdt.cc
```diff
@@ -0,0 +1,233 @@
...
+  ::close(memfd);
+  return true;
+}
+
+bool Probe::enable(int pid) {
+  if (!add_to_semaphore(pid, +1))
+    return false;
+
+  // TODO: what happens if we enable this twice?
+  enabled_semaphores_.emplace(pid, std::move(ProcStat(pid)));
```
aaronpuchert: Clang is warning here: `moving a temporary object prevents copy elision [-Wpessimizing-move]`.

Calling `std::move` here doesn't make sense, since the object created by `ProcStat(pid)` is already an R-Value.

vmg: Thank you, you are right. Fixed.

Issue #504, opened 2016-04-25
new dependency on libelf is not captured in package


drzaeus77:
We added the dependency on libelf-dev into the debian/control file, but the downstream clients are now getting a linker error if libelf is not installed. Obviously this needs to be fixed.

The easy fix is to add a dependency in the deb on libelf1, but I would much rather have statically linked libelf.a. However, that package doesn't compile with -fPIC.

The same question probably goes for rpm, but I haven't tested that yet.

@vmg can you look into this, and see if somehow we can do it statically?

vmg:
Hey! Sorry I didn't get to this issue early.

As @drzaeus77 accurately points out, the `libelf.a` that ships with Debian and Ubuntu is not compiled with `-fPIC`, so it's not possible to link it statically into a library. Boo! :/

I think the only possible workaround is vendoring libelf as part of `bcc`. It's not a big dependency, and IMO it's better than having to install `libelf.so` on production systems.

Let me try vendor the library and see how this turns out.

vmg:
Hmpf. So the `elfutils` package in Debian does build a PIC version of `libelf.a` (aptly named `libelf_pic.a`), it just chooses not to package it. Disappointing.

@drzaeus77: What would be your preferred course of action? Vendor the libelf sources into `vendor/libelf`? Or add the `elfutils` git repository as a submodule?

PR #505, opened 2016-04-26
Remove external dependency on libelf

vmg:
This pull request uses the CMake "external project" tooling to fetch `libelf` dynamically at runtime, allowing us to link it statically against our main project and removing the runtime dependency on `libelf.so`.

CMake makes this whole process rather straightforward and painless, so I'm quite happy with the solution. `libelf-dev` (or the Fedora equivalent) is no longer required for the build process.

Fixes https://github.com/iovisor/bcc/issues/504

cc @drzaeus77


drzaeus77:
Hmm, interesting approach, but now we're statically vendoring a GPLv3 library. @4ast with your stricter packaging requirements, which is easier for you?

vmg:
@drzaeus77: LGPLv3, actually. And we're statically linking it. AFAICT this is completely abiding by the license's terms.

vmg:
Also note that we're directly fetching the Ubuntu source package. This is as close as "depending on a deb" as we can get, considering that we need to tune the build options for that deb.

...On a related topic, it seems like the source package for `libelf` fails to configure on Fedora: it complains about not finding a GNU C99 compliant compiler.

It's kinda weird, because the check on the `configure` script actually tries to build a simple `.c` file with some C99 features. And yet it seems to be failing in Fedora. Any version of GCC greater than 4.x should succeed!

drzaeus77:
Thanks, just wanted to make sure the question was asked.

I'll look into the buildbot in a bit.

cmake/BuildLibElf.cmake
```diff
@@ -0,0 +1,15 @@
+include(ExternalProject)
+
+ExternalProject_Add(libelf
+  PREFIX "vendor/elfutils"
+  URL http://archive.ubuntu.com/ubuntu/pool/main/e/elfutils/elfutils_0.163.orig.tar.bz2
```
4ast: that won't work for us, since this type of 'during the make' fetching is impossible in dc environment. All source has to be present during the build. We cannot do uploads during the build. Sorry. I guess we have to live with libelf.so or ask people build one independently as libelf_pic.a with -fpic

vmg:
@drzaeus77: OK, I know why is this failing. You're forcing `CC` to `clang` in the RPM-building step of the buildbot.

From Clang's C features page:

> clang does not support nested functions; this is a complex feature which is infrequently used, so it is unlikely to be implemented anytime soon. In C++11 it can be emulated by assigning lambda functions to local variables, e.g:

This is one of the "features" that the `libelf` configure file is testing for... Even though I'm 99% sure that it's _not_ needed to build `libelf` itself.

I'm going to verify whether this is the case, and fix the `configure` script accordingly.

cmake/BuildLibElf.cmake
vmg: Ouch! Well, that settles it I guess.

@drzaeus77: What would you prefer? Runtime dependency on `libelf.so`?

vmg: ...Alternatively, I could check in `elfutils_0.163.orig.tar.bz2` on the repository. It's a tad unorthodox, but I know of a lot of people who do just that for upstream dependencies.

4ast: imo that's ugly too. imo libelf.so is not a big deal, since it's only a dependency in libbcc.so.
The fully static bcc-lua should be statically linking both libbcc.a and libelf.a
and in the future that will be the main tracing tool that won't have any deps.

drzaeus77: This split makes sense. Let's keep the .so dependency, and let static be static.

vmg: Settled. Will open followup PR. 👌

PR #507, opened 2016-04-26
Explicitly list the libelf runtime dependency

vmg:
The shared `libbcc.so` depends on `libelf.so`. So list this package dependency explicitly for Ubuntu and Fedora.

cc @4ast @drzaeus77


4ast:
lgtm

PR #513, opened 2016-04-29
Powerpc support

rnav:
Greetings--
This patchset enables basic powerpc support for bcc. The core library now builds and seems to work fine. I have also fixed a few tests, but there are a few more tests and tools that are failing as they seem to have a hard dependency on x86.

Kindly take a look and consider merging this.

### DIFF
--- a/src/cc/CMakeLists.txt
+++ b/src/cc/CMakeLists.txt
@@ -44,7 +44,7 @@ set_target_properties(bcc-static PROPERTIES OUTPUT_NAME bcc)
 # BPF is still experimental otherwise it should be available
 #llvm_map_components_to_libnames(llvm_libs bpf mcjit irreader passes)
 llvm_map_components_to_libnames(llvm_libs bitwriter bpfcodegen irreader linker
-  mcjit objcarcopts option passes x86codegen)
+  mcjit objcarcopts option passes x86codegen powerpccodegen)
 llvm_expand_dependencies(expanded_libs ${llvm_libs})
 
 # order is important
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -35,8 +35,14 @@ namespace ebpf {
 const char *calling_conv_regs_x86[] = {
   "di", "si", "dx", "cx", "r8", "r9"
 };
+const char *calling_conv_regs_ppc[] = {"gpr[3]", "gpr[4]", "gpr[5]",
+                                       "gpr[6]", "gpr[7]", "gpr[8]"};
 // todo: support more archs
+#if defined(__powerpc__)
+const char **calling_conv_regs = calling_conv_regs_ppc;
+#else
 const char **calling_conv_regs = calling_conv_regs_x86;
+#endif
 
 using std::map;
 using std::set;
--- a/src/cc/frontends/clang/loader.cc
+++ b/src/cc/frontends/clang/loader.cc
@@ -128,7 +128,11 @@ int ClangLoader::parse(unique_ptr<llvm::Module> *mod, unique_ptr<vector<TableDes
   DiagnosticsEngine diags(DiagID, &*diag_opts, diag_client);
 
   // set up the command line argument wrapper
+#if defined(__powerpc64__)
+  driver::Driver drv("", "ppc64le-unknown-linux-gnu", diags);
+#else
   driver::Driver drv("", "x86_64-unknown-linux-gnu", diags);
+#endif
   drv.setTitle("bcc-clang-driver");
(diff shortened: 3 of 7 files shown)


src/cc/CMakeLists.txt
```diff
@@ -44,7 +44,7 @@ set_target_properties(bcc-static PROPERTIES OUTPUT_NAME bcc)
 # BPF is still experimental otherwise it should be available
 #llvm_map_components_to_libnames(llvm_libs bpf mcjit irreader passes)
 llvm_map_components_to_libnames(llvm_libs bitwriter bpfcodegen irreader linker
-  mcjit objcarcopts option passes x86codegen)
+  mcjit objcarcopts option passes x86codegen powerpccodegen)
```
4ast: what's the difference in .so size after inclusion of ppc backend?
is it really needed?
can we select x86codegen or powerpccodegen at build time?
I think we should try to remove x86. We can probably just stub LLVMInitializeX86Target*

drzaeus77: x86 is used in the implementation of scanf/sprintf helpers.

4ast: ahh yeah, it was for fuse only, right?

drzaeus77: No I use it in iomodules as well, since right now this is the only way to get keys in/out of the maps without having the struct layout known in the particular binding. I don't like it, this has been itching me for a while, and as soon as I find a way to do this more cleanly I will remove the old way.

4ast: agree.
I think if the cost of extra ppc backend is in Kbytes then it's fine to link the whole thing, otherwise we need to pick native backend for fuse/iomodule usage.

rnav: There is no change in the size of libbcc - the powerpc bits only get included when built on powerpc.

rnav: Oh and thanks for the review!

4ast: hmm. not following. including powerpccodegen into llvm_libs makes cmake include them on x86 too. It's probably better to do something like 'nativecodegen'.

rnav: Yes, it is included, but not linked in. BPFModule() uses InitializeNativeTarget() which translates to LLVMInitializeX86Target on x86 and LLVMInitializePowerPCTarget on powerpc. So, while linking, only the platform-specific library is picked up.

```
[root@x86node lib64]# readelf -s libbcc.so.0.1.8 | grep LLVMInitialize
  2708: 00000000011ae7b0     3 FUNC    LOCAL  DEFAULT   11 _ZZ27LLVMInitializeBPFTar
 20644: 0000000001031080    41 FUNC    LOCAL  DEFAULT   11 LLVMInitializeBPFTarget
 25265: 0000000000df3f70    30 FUNC    LOCAL  DEFAULT   11 LLVMInitializeX86Target
 27952: 00000000011adb10   192 FUNC    LOCAL  DEFAULT   11 LLVMInitializeBPFTargetMC
 28406: 0000000000f04490   185 FUNC    LOCAL  DEFAULT   11 LLVMInitializeX86TargetMC
<snip>
```

```
[root@rhel71le build]# readelf -s /usr/lib64/libbcc.so.0.1.8 | grep LLVMInitia
    29: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND LLVMInitializeBPFTarget
   125: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND LLVMInitializeBPFTargetMC
   142: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND LLVMInitializeBPFAsmPrint
   229: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND LLVMInitializeBPFTargetIn
   287: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND LLVMInitializePowerPCAsmP
   316: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND LLVMInitializePowerPCTarg
<snip>
```

I could be wrong, but that's my understanding.

4ast: that's correct for .so, but .a will include it all. Eventually fully static bcc-lua should stay the same though. Right.
The other concern is that some distros may not ship powerpc backend by default and the build will break on x86 though ppc backend is unused.
Can we do an 'if' inside cmake to include x86codegen or ppccodegen?

rnav: Sure. I just tested with 'nativecodegen' and it actually works. It would also be appropriate for other architectures as well. I'll send a new pull request with this change. Thanks.

rnav:
Raised a new pull request #520

Issue #512, opened 2016-04-28
lua install steps


brendangregg:
Either in INSTALL.md, or in docs/lua.md and linked from INSTALL.md. ... Might want a docs/python.md as well at some point.

I haven't figured it out yet. Here's what I get when I run an example:

```
/mnt/src/bcc/examples/lua# lua bashreadline.lua 
lua: bashreadline.lua:16: module 'ffi' not found:
	no field package.preload['ffi']
	no file './ffi.lua'
	no file '/usr/local/share/lua/5.1/ffi.lua'
	no file '/usr/local/share/lua/5.1/ffi/init.lua'
	no file '/usr/local/lib/lua/5.1/ffi.lua'
	no file '/usr/local/lib/lua/5.1/ffi/init.lua'
	no file '/usr/share/lua/5.1/ffi.lua'
	no file '/usr/share/lua/5.1/ffi/init.lua'
	no file './ffi.so'
	no file '/usr/local/lib/lua/5.1/ffi.so'
	no file '/usr/lib/x86_64-linux-gnu/lua/5.1/ffi.so'
	no file '/usr/lib/lua/5.1/ffi.so'
	no file '/usr/local/lib/lua/5.1/loadall.so'
stack traceback:
	[C]: in function 'require'
	bashreadline.lua:16: in main chunk
	[C]: ?
```

Ok, am I supposed to be running it from lua or luajit?:

```
/mnt/src/bcc/examples/lua# /mnt/LuaJIT-2.0.4/src/luajit bashreadline.lua 
/mnt/src/bcc/examples/lua#
```

I'm probably just doing it wrong. //CC @vmg

vmg:
Hi Brendan! Sorry about the lack of documentation!

The recommended way to run the lua probes is the `bcc-lua` tool. If CMake detects the `luajit` and `luajit-5.1-dev` packages (these are the Ubuntu names), it will create a standalone `bcc-lua` tool with no dependencies that run any of the Lua probes and has an user friendly interface.

You'll find it at `build/src/lua/bcc-lua` (again, if the Lua packages are available). This tool has 0 dependencies (so it can be deployed to servers) and will load `libbcc.so` dynamically at runtime (to let you switch between libbcc versions). Setting the `LIBBCC_SO_PATH` environment variable will let you point the tool at the version of the library you want to be using.

From there it's smooth sailing. Try `./bcc-lua --help` to see the available options.

4ast:
[...]

PR #515, opened 2016-04-30
Lua tooling improvements

vmg:
Some misc fixes which were bothering me:

- `bcc-lua` now links statically against `libbcc.a`. The trick is passing `--whole-archive` to `ld`, so all the symbols get linked in and can be exported. When running `bcc-lua`, the `--so-path` CLI option is disabled, and we no longer try to find the SO anywhere in the system.

- The example `lua` probes are now `+x`, and will try to run themselves through `bcc-lua` if it's in the path. Thanks to @brendangregg for the neat idea.

- `bcc-lua` now rebuilds itself when you modify the Lua sources! Finally! Turns out we were not globbing for the Lua sources properly.

- (unrelated) there was a bug in `table.py` where you couldn't manually close open perf buffers. It was calling the parent class' `__init__` instead of `__delitem__`. :)

- And last, but not least, fix the shutdown noise that @brendangregg was seeing in in #512. We were trying to "detach" the perf buffers -- perf buffers are listed as kprobes for polling, but they don't need "detaching".

### DIFF
--- a/src/lua/CMakeLists.txt
+++ b/src/lua/CMakeLists.txt
@@ -19,6 +19,7 @@ if (LUAJIT_LIBRARIES AND LUAJIT)
 	include_directories(${LUAJIT_INCLUDE_DIR})
 	add_executable(bcc-lua src/main.c bcc.o)
 	target_link_libraries(bcc-lua ${LUAJIT_LIBRARIES})
+	target_link_libraries(bcc-lua -Wl,--whole-archive bcc-static -Wl,--no-whole-archive)
 
 	install(TARGETS bcc-lua RUNTIME DESTINATION bin)
 endif()
--- a/src/lua/CMakeLists.txt
+++ b/src/lua/CMakeLists.txt
@@ -2,7 +2,9 @@ find_package(LuaJIT)
 find_program(LUAJIT luajit)
 
 if (LUAJIT_LIBRARIES AND LUAJIT)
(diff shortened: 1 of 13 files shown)


src/python/bcc/table.py
```diff
@@ -372,7 +372,7 @@ def __init__(self, *args, **kwargs):
         super(PerfEventArray, self).__init__(*args, **kwargs)
 
     def __delitem__(self, key):
-        super(PerfEventArray, self).__init__(key)
+        super(PerfEventArray, self).__delitem__(key)
```
vmg: Copypaste issue :)

src/lua/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ if (LUAJIT_LIBRARIES AND LUAJIT)
 
 	include_directories(${LUAJIT_INCLUDE_DIR})
 	add_executable(bcc-lua src/main.c bcc.o)
+	set_target_properties(bcc-lua PROPERTIES LINKER_LANGUAGE C)
```
4ast: so it was using g++ to link and failing? is this an optimization or fix?

vmg: It's a WIP-fix, actually. Sorry I didn't make it obvious. It was picking `g++` to compile the source program, which caused `bcc-lua` to depend on `libstdc++` (when ideally it wouldn't, because we're already linking is statically).

Changing it to compile using `gcc` now parses `main.c` as C code (which is a good thing), but for some reason CMake still insists on linking `libstdc++`... So I need to figure out Step 2 to remove that dependency.

Issue #327, opened 2016-01-29
USDT probes


brendangregg:
User-level statically defined tracing probes have been placed in various applications and runtimes, including Java, Node.js, MySQL, and PostgreSQL. These allow API-stable scripts to be written, that do not depend on tracing raw user-level functions (uprobes).

As an example of hacking in USDT tracing using ftrace, see: http://www.brendangregg.com/blog/2015-07-03/hacking-linux-usdt-ftrace.html . The unpublished script I referred to is: https://gist.github.com/brendangregg/f1b3d09c14088522065b

For a simple example to trace:

##### 1. Create tick-dtrace.d:
```
provider tick {
        probe loop(int);
}

#pragma D attributes Evolving/Evolving/ISA provider node provider
#pragma D attributes Private/Private/Unknown provider node module
#pragma D attributes Private/Private/Unknown provider node function
#pragma D attributes Private/Private/ISA provider node name
#pragma D attributes Evolving/Evolving/ISA provider node args
```

##### 2. Then create an object file:

```
# apt-get install -y systemtap-sdt-dev        # adds "dtrace"
# dtrace -G -s tick-dtrace.d -o tick-dtrace.o
```

##### 3. Create the target program, tick-main.c:

```
#include <stdio.h>
#include <unistd.h>
/* from systemtap-sdt */
#include <sys/sdt.h>

int
main(int argc, char *argv[])
{
        int i;

        for (i = 0; i < 5; i++) {
                DTRACE_PROBE1(tick, loop, i);
                printf("hi: %d\n", i);
                sleep(1);
        }

        return (0);
}
```

##### 4. Compile tick-main:

```
gcc -c tick-main.c
gcc -o tick tick-main.o tick-dtrace.o
```

##### 5. Check it has USDT probes:

```
readelf -n tick

Notes at offset 0x0000021c with length 0x00000020:
  Owner                 Data size       Description
  GNU                  0x00000010       NT_GNU_ABI_TAG (ABI version tag)
    OS: Linux, ABI: 2.6.35

Notes at offset 0x0000023c with length 0x00000024:
  Owner                 Data size       Description
  GNU                  0x00000014       NT_GNU_BUILD_ID (unique build ID bitstring)
    Build ID: 2b94c0e7c684a001a34d685c862b33aa51ff7672

Notes at offset 0x00000a08 with length 0x00000044:
  Owner                 Data size       Description
  stapsdt              0x0000002e       NT_STAPSDT (SystemTap probe descriptors)
[...]

PR #518, opened 2016-05-01
[RFC] New USDT API

vmg:
Hey everyone,

Here's a proposal on what would a potential new USDT API look like. This is based on top off the C++ USDT code which I've previously written. I've implemented this API in Lua instead of Python because I find it easier to iterate the design like that, but if people think it's acceptable, I will port it to Python too.

### The existing API

The current interface for writing, attaching and working with USDT probes is a bit complex (*understatement.jpg*).

In https://github.com/iovisor/bcc/issues/327#issuecomment-215540322, @brendangregg writes a "hello world" USDT probe which highlights many shortcomings on the API. Namely:

1. You need to create a USDT reader to list/load the probes

2. You need to randomly insert two sections of "boilerplate" into your C text. The USDT reader generates this boilerplate in two phases: a set of wrapper functions for your actual C probe, which are used to notify your probe of the USDT calling location, and a "case" statement that assigns the arguments of the USDT probe to local variables named `argX`.

    This is IMO the biggest design flaw in the existing interface. Your C probe needs to be acutely aware of this implementation detail; you need to adjust your probe to take a magical `__loc_id` argument (which needs to be named exactly that), and you need to pick a place in the body of your probe to insert the argument cases.

3. Once you've generated your C probe with the boilerplate, there is no easy way to attach the corresponding UProbes. Again, the user needs to know the names of the magical wrappers generated by the USDT Reader, and attach those wrappers as UProbes. The user also needs to know that a probe can appear in different locations, and understand that we need to attach to specific addresses (i.e. the probe locations) instead to function names.

4. Last, but not least, the user must set up a signal handler to ensure that the probes that have been activated on a given PID are disabled once tracing is done.

### The proposal

As an alternative, I propose the following API, exemplified in a commented "hello world" example:

```lua
local program = [[
// C code for the probe; note that there's no reference to the probe location
// also note that there's no text replacement performed
int trace_method(struct pt_regs *ctx) {
    char fn_name[128] = {};
    // read the 2nd argument to the `method__entry` USDT probe
    // the `_p` variant of this helper dereferences the argument as a pointer
    // and reads its contents
    bpf_usdt_readarg_p(method__entry_2, ctx, &fn_name, sizeof(fn_name));

    // you can also call the following API (which is typesafe on its return value)
    int32_t line_number = bpf_usdt_readarg(method__entry_4, ctx);

    // print to the trace log as usual
    bpf_trace_printk("%s(...)\n", &fn_name);
    return 0;
};
]]

return function(BPF, util)
  -- create a USDT context for the given PID
  -- (you can also target a binary instead of a PID)
  local u = util.USDT:new{pid=tonumber(arg[1])}

  -- enable the `method__entry` USDT probe, calling the `trace_method` of our C text
  -- this will automagically do probe resolution and enabling of semaphores
  u:enable_probe{probe="method__entry", fn_name="trace_method"}

  -- now we create our BPF context, and pass our C text *and* the USDT context
  local b = BPF:new{text=program, usdt=u}

  -- we're ready. the BPF code has automatically attached to all the USDT probes
  -- which we enabled in our USDT context, and we can start tracing
  local pipe = b:pipe()
  while true do
    print(pipe:trace_fields())
  end

  -- the USDT semaphores will be automatically disabled when quitting
end
```

### The new codegen phase

Inquiring minds will probably notice that I've gotten rid of all text substitution for the C text, and instead any BPF program that has been attached to an USDT probe has access to generic, typesafe APIs to query the arguments for the probe, regardless of source location. How is this possible?

Well, let's take a look at the generated BPF program as it goes into the C compiler to understand this better:

```c
static inline uint64_t _bpf_readarg_method__entry_1(struct pt_regs *ctx) {
  uint64_t result = 0x0;
  switch(ctx->ip) {
  case 0x55c79eeed834ULL: result = (uint64_t)ctx->bx; break;
  case 0x55c79eef68f4ULL: result = (uint64_t)ctx->r13; break;
  }
  return result;
}
static inline uint64_t _bpf_readarg_method__entry_2(struct pt_regs *ctx) {
  uint64_t result = 0x0;
  switch(ctx->ip) {
  case 0x55c79eeed834ULL: result = (uint64_t)ctx->bp; break;
  case 0x55c79eef68f4ULL: { u64 __addr = ctx->sp + (8); uint64_t __res = 0x0; bpf_probe_read(&__res, sizeof(__res), (void *)__addr); result = __res; } break;
  }
  return result;
}
static inline uint64_t _bpf_readarg_method__entry_3(struct pt_regs *ctx) {
  uint64_t result = 0x0;
  switch(ctx->ip) {
  case 0x55c79eeed834ULL: result = (uint64_t)ctx->r12; break;
  case 0x55c79eef68f4ULL: result = (uint64_t)ctx->dx; break;
  }
  return result;
}
static inline int32_t _bpf_readarg_method__entry_4(struct pt_regs *ctx) {
  int32_t result = 0x0;
  switch(ctx->ip) {
  case 0x55c79eeed834ULL: result = (int32_t)ctx->ax; break;
  case 0x55c79eef68f4ULL: result = (int32_t)ctx->ax; break;
  }
  return result;
}
int trace_method(struct pt_regs *ctx) {
    char fn_name[128] = {};
    bpf_usdt_readarg_p(method__entry_2, ctx, &fn_name, sizeof(fn_name));
    bpf_trace_printk("%s(...)\n", &fn_name);
    return 0;
};
```

Oooh. That makes sense! We don't need to insert the argument parsing code throughout our C text; we can simply prepend a set of helpers to read each argument, and the helpers will be called magically by the `bpf_usdt_readarg` macro. This process is transparent to the user.

```c
#define bpf_usdt_readarg(probearg, ctx) _bpf_readarg_##probearg(ctx)
#define bpf_usdt_readarg_p(probearg, ctx, buf, len) {\
  u64 __addr = bpf_usdt_readarg(probearg, ctx); \
  bpf_probe_read(buf, len, (void *)__addr); }
```

Also note that there's no longer any magical wrappers for "probe location" or `__loc_id`. Instead, we can tell what's the location of an USDT probe because... well... we know where the probe is being called from. That's why `%rip` is there. :)

It is also worth noting that for the common case of probes with a single source location, the `%rip` check is skipped. Overall, very little code that generates typesafe and efficient APIs that IMO are a pleasure to use.

### Conclusion

That's what I'm proposing here today. I'm quite happy with this design. It can be integrated into a language wrapper with very few changes, and the surface of the C API is minimal, so I believe the abstraction level is quite adequate.

I'd love to hear your thoughts on this. The existing `examples/lua/ruby_usdt.lua` would be a good starting point for local testing.

cc @drzaeus77 @4ast @brendangregg

### DIFF
--- a/src/cc/usdt.cc
+++ b/src/cc/usdt.cc
@@ -38,11 +38,12 @@ Probe::Location::Location(uint64_t addr, const char *arg_fmt) : address_(addr) {
 }
 
 Probe::Probe(const char *bin_path, const char *provider, const char *name,
-             uint64_t semaphore)
+             uint64_t semaphore, const optional<int> &pid)
     : bin_path_(bin_path),
       provider_(provider),
       name_(name),
-      semaphore_(semaphore) {}
+      semaphore_(semaphore),
+      pid_(pid) {}
 
 bool Probe::in_shared_object() {
   if (!in_shared_object_)
@@ -50,45 +51,36 @@ bool Probe::in_shared_object() {
   return in_shared_object_.value();
 }
 
-bool Probe::resolve_global_address(uint64_t *global, const uint64_t addr,
-                                   optional<int> pid) {
+bool Probe::resolve_global_address(uint64_t *global, const uint64_t addr) {
   if (in_shared_object()) {
-    return (pid &&
-            bcc_resolve_global_addr(*pid, bin_path_.c_str(), addr, global) ==
-                0);
+    return (pid_ && !bcc_resolve_global_addr(
+          *pid_, bin_path_.c_str(), addr, global));
   }
 
   *global = addr;
   return true;
 }
 
-bool Probe::lookup_semaphore_addr(uint64_t *address, int pid) {
-  auto it = semaphores_.find(pid);
-  if (it != semaphores_.end()) {
-    *address = it->second;
-    return true;
+bool Probe::add_to_semaphore(int16_t val) {
+  assert(pid_ && attached_semaphore_);
+
+  if (!attached_semaphore_) {
+    uint64_t addr;
+    if (!resolve_global_address(&addr, semaphore_))
+      return false;
+    attached_semaphore_ = addr;
   }
 
-  if (!resolve_global_address(address, semaphore_, pid))
-    return false;
+  off_t address = static_cast<off_t>(attached_semaphore_.value());
 
-  semaphores_[pid] = *address;
-  return true;
-}
-
-bool Probe::add_to_semaphore(int pid, int16_t val) {
-  uint64_t address;
-  if (!lookup_semaphore_addr(&address, pid))
-    return false;
-
-  std::string procmem = tfm::format("/proc/%d/mem", pid);
+  std::string procmem = tfm::format("/proc/%d/mem", pid_.value());
   int memfd = ::open(procmem.c_str(), O_RDWR);
   if (memfd < 0)
     return false;
 
-  int16_t original;  // TODO: should this be unsigned?
+  int16_t original;
 
-  if (::lseek(memfd, static_cast<off_t>(address), SEEK_SET) < 0 ||
+  if (::lseek(memfd, address, SEEK_SET) < 0 ||
       ::read(memfd, &original, 2) != 2) {
     ::close(memfd);
     return false;
@@ -96,7 +88,7 @@ bool Probe::add_to_semaphore(int pid, int16_t val) {
 
   original = original + val;
 
-  if (::lseek(memfd, static_cast<off_t>(address), SEEK_SET) < 0 ||
+  if (::lseek(memfd, address, SEEK_SET) < 0 ||
       ::write(memfd, &original, 2) != 2) {
     ::close(memfd);
     return false;
@@ -106,28 +98,33 @@ bool Probe::add_to_semaphore(int pid, int16_t val) {
   return true;
 }
 
-bool Probe::enable(int pid) {
-  if (enabled_semaphores_.find(pid) != enabled_semaphores_.end())
-    return true;
-
-  if (!add_to_semaphore(pid, +1))
+bool Probe::enable(const std::string &fn_name) {
+  if (attached_to_)
     return false;
 
-  enabled_semaphores_.emplace(pid, std::move(ProcStat(pid)));
+  if (need_enable()) {
+    if (!pid_)
+      return false;
+
+    if (!add_to_semaphore(+1))
+      return false;
+  }
+
+  attached_to_ = fn_name;
   return true;
 }
 
-bool Probe::disable(int pid) {
-  auto it = enabled_semaphores_.find(pid);
-  if (it == enabled_semaphores_.end())
+bool Probe::disable() {
+  if (!attached_to_)
     return false;
 
-  bool result = true;
-  if (!it->second.is_stale())
-    result = add_to_semaphore(pid, -1);
+  attached_to_ = nullopt;
 
-  enabled_semaphores_.erase(it);
-  return result;
+  if (need_enable()) {
+    assert(pid_);
+    return add_to_semaphore(-1);
+  }
+  return true;
 }
 
 std::string Probe::largest_arg_type(size_t arg_n) {
@@ -143,9 +140,11 @@ std::string Probe::largest_arg_type(size_t arg_n) {
   return largest->ctype();
 }
 
-bool Probe::usdt_getarg(std::ostream &stream,
-    const std::string &fn_name, const optional<int> &pid) {
+bool Probe::usdt_getarg(std::ostream &stream) {
   const size_t arg_count = locations_[0].arguments_.size();
+
+  if (!attached_to_)
+    return false;
 
   if (arg_count == 0)
     return true;
@@ -158,13 +157,13 @@ bool Probe::usdt_getarg(std::ostream &stream,
         "static inline int _bpf_readarg_%s_%d("
         "struct pt_regs *ctx, void *dest, size_t len) {\n"
         "  if (len != sizeof(%s)) return -1;\n",
-        fn_name, arg_n + 1, ctype);
+        attached_to_.value(), arg_n + 1, ctype);
 
     if (locations_.size() == 1) {
       Location &location = locations_.front();
       stream << "  ";
       if (!location.arguments_[arg_n].assign_to_local(stream, cptr,
-                                                      bin_path_, pid))
+                                                      bin_path_, pid_))
         return false;
       stream << "\n  return 0;\n}\n";
     } else {
@@ -172,12 +171,12 @@ bool Probe::usdt_getarg(std::ostream &stream,
       for (Location &location : locations_) {
         uint64_t global_address;
 
-        if (!resolve_global_address(&global_address, location.address_, pid))
+        if (!resolve_global_address(&global_address, location.address_))
           return false;
 
         tfm::format(stream, "  case 0x%xULL: ", global_address);
         if (!location.arguments_[arg_n].assign_to_local(stream, cptr,
-                                                        bin_path_, pid))
+                                                        bin_path_, pid_))
           return false;
 
         stream << " return 0;\n";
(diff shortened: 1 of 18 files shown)


goldshtn:
I like the new design. Getting rid of code is usually a good thing, especially when it's my code ;-)

vmg:
Haha, technically we're not getting rid of code, we're just porting it to C++. ;)

drzaeus77:
The implementation of bpf_usdt_readarg_p belongs in b_frontend_action, otherwise future uses of that helper will conflict with other code that gets clang-rewritten. Generally speaking, macros don't mix well with the clang rewriter, since it makes the text read-only.

src/cc/usdt.cc
```diff
@@ -137,32 +146,79 @@ bool Probe::usdt_cases(std::ostream &stream, const optional<int> &pid) {
...
+
+    if (locations_.size() == 1) {
+      Location &location = locations_.front();
+      stream << "  ";
+      if (!location.arguments_[arg_n].assign_to_local(stream, "result",
+                                                      bin_path_, pid))
+        return false;
+      stream << "\n";
+    } else {
+      stream << "  switch(ctx->ip) {\n";
```
4ast: looks like the loop below will handle different registers at different locations properly. neat approach. I guess the switch(ip) has to stay dynamic unless we can duplicate the program for different ips. The cost of switch is small comparing to the rest. How common is 'locations==1' case?

goldshtn: From my experiments with Node, libc, JVM - most probes have only one location.

brendangregg:
Looks pretty good!

```
int trace_method(struct pt_regs *ctx) {
    char fn_name[128] = {};
    bpf_usdt_readarg_p(method__entry_2, ctx, &fn_name, sizeof(fn_name));
    int32_t line_number = bpf_usdt_readarg(method__entry_4, ctx);
[...]
  u:enable_probe{probe="method__entry", fn_name="trace_method"}
```

The "method__entry"s in bpf_usdt_readarg seem redundant, although I know why they are there.

Maybe a future way to address it is adding, in the same vein as the kprobe__ and kretprobe__, optional shortcuts (eg, see opensnoop.py):

```
int usdtprobe__method__entry(struct pt_regs *ctx) {
    char fn_name[128] = {};
    bpf_usdt_readarg_p(2, ctx, &fn_name, sizeof(fn_name));
    int32_t line_number = bpf_usdt_readarg(4, ctx);
[...]
```

Again, similar to why we added kprobe__ -- for cases where you just had one probe and one C function. Future addition.

vmg:
> The implementation of bpf_usdt_readarg_p belongs in b_frontend_action, otherwise future uses of that helper will conflict with other code that gets clang-rewritten. Generally speaking, macros don't mix well with the clang rewriter, since it makes the text read-only.

Hm. I see. I'm afraid the code in `b_frontend_action` is particularly arcane to me. I'd love some pointers on how to implement the argument helper as a rewrite rule instead of a macro... By doing so, I think we could also handle @brendangregg's complaint that having to specify the name of the USDT probe is redundant, since we're only attaching one uprobe for each USDT address.

drzaeus77:
Note to other reviewers: I'd like to let this RFC sit here for a few days. I'd like to spend some time looking at how to integrate this a little better with the rewriter.

vmg:
@drzaeus77: OK, I've given another stab at implementing the arguments API as a frontend action.

Here's what the probe looks like. The call now looks like `bpf_usdt_readarg(2, ctx, &addr);`. As @brendangregg pointed out, you no longer need to pass the name of the USDT probe, because the frontend knows what function is attached to each probe.

```c
int trace_method(struct pt_regs *ctx) {
  uint64_t addr;
  bpf_usdt_readarg(2, ctx, &addr);

  char fn_name[128] = {};
  bpf_probe_read(&fn_name, sizeof(fn_name), (void *)addr);

  bpf_trace_printk("%s(...)\n", fn_name);
  return 0;
};
```

And this is the generated code:

```c
__attribute__((always_inline))
static inline int _bpf_readarg_trace_method_1(struct pt_regs *ctx, void *dest, size_t len) {
  if (len != sizeof(uint64_t)) return -1;
  switch(ctx->ip) {
  case 0x555dfcc63834ULL: *((uint64_t *)dest) = (uint64_t)ctx->bx; return 0;
  case 0x555dfcc6c8f4ULL: *((uint64_t *)dest) = (uint64_t)ctx->r13; return 0;
  }
  return -1;
}
__attribute__((always_inline))
static inline int _bpf_readarg_trace_method_2(struct pt_regs *ctx, void *dest, size_t len) {
  if (len != sizeof(uint64_t)) return -1;
  switch(ctx->ip) {
  case 0x555dfcc63834ULL: *((uint64_t *)dest) = (uint64_t)ctx->bp; return 0;
  case 0x555dfcc6c8f4ULL: { u64 __addr = ctx->sp + (8); uint64_t __res = 0x0; bpf_probe_read(&__res, sizeof(__res), (void *)__addr); *((uint64_t *)dest) = __res; } return 0;
  }
  return -1;
}
__attribute__((always_inline))
static inline int _bpf_readarg_trace_method_3(struct pt_regs *ctx, void *dest, size_t len) {
  if (len != sizeof(uint64_t)) return -1;
  switch(ctx->ip) {
  case 0x555dfcc63834ULL: *((uint64_t *)dest) = (uint64_t)ctx->r12; return 0;
  case 0x555dfcc6c8f4ULL: *((uint64_t *)dest) = (uint64_t)ctx->dx; return 0;
  }
  return -1;
}
__attribute__((always_inline))
static inline int _bpf_readarg_trace_method_4(struct pt_regs *ctx, void *dest, size_t len) {
  if (len != sizeof(int32_t)) return -1;
  switch(ctx->ip) {
  case 0x555dfcc63834ULL: *((int32_t *)dest) = (int32_t)ctx->ax; return 0;
  case 0x555dfcc6c8f4ULL: *((int32_t *)dest) = (int32_t)ctx->ax; return 0;
  }
  return -1;
}
__attribute__((section(".bpf.fn.trace_method")))
int trace_method(struct pt_regs *ctx) {
  uint64_t addr;
  _bpf_readarg_trace_method_2(ctx, &addr, sizeof(*(&addr)));

  char fn_name[128] = {};
  bpf_probe_read(&fn_name, sizeof(fn_name), (void *)addr);

  ({ char _fmt[] = "%s(...)\n"; bpf_trace_printk_(_fmt, sizeof(_fmt), fn_name); });
  return 0;
};
```

I think this is a strict improvement. And it should not collide with other rewrite phases. Thoughts @drzaeus77 @brendangregg?

src/cc/usdt.h
```diff
@@ -159,6 +164,8 @@ class Probe {
 
 class Context {
   std::vector<Probe *> probes_;
+  std::vector<std::pair<Probe *, std::string>> uprobes_;
```
goldshtn: Can the `probes_` and `uprobes_` vectors contain `std::unique_ptr<Probe>` instead of raw pointers?

vmg: Done. I've revamped a chunk of of the USDT code. I think it makes more sense now, and there are no more `new` or `delete`s in sight.

drzaeus77:
@vmg Thanks for converting, sorry I left you hanging without helping to explain how it should be done. You seem to have figured it out.

Unfortunately on my system I wasn't able to get a version of ruby with STAP symbols, even though ./configure of ruby from source seemed to claim it was enabling dtrace support. Lies! I wonder if other folks had better experience with nodejs (I did), and if that should be our default example for USDT, to avoid others pestering us with FAQs.

vmg:
@drzaeus77: Totally. I'll add a Node.js example then.

Looking forward to your feedback about the rewrite phase. :)

src/cc/usdt.cc
```diff
@@ -210,14 +229,44 @@ std::string Context::resolve_bin_path(const std::string &bin_path) {
   return result;
 }
 
-Probe *Context::find_probe(const std::string &probe_name) {
-  for (Probe *p : probes_) {
+std::shared_ptr<Probe> Context::get(const std::string &probe_name) {
```
drzaeus77: As a personal choice I tend to avoid shared_ptr, partly due to bad history and partly due to design philosophy. I haven't read through this code in enough depth to say whether it can be done differently, so I'll just ask: can it be done differently?

drzaeus77:
The rewrite implementation looks fine.

src/cc/usdt.cc
vmg: It can. The only reason I'm using `shared_ptr` is to be able to return Probes to use on the test suite. Would you prefer if I used `unique_ptr` and returned bare pointers instead?

drzaeus77: Yes, if you would please.

vmg:
@drzaeus77: Replaced `shared_ptr` with `unique_ptr`. 👌

vmg:
Nice. Now to port the Python code. :D

PR #521, opened 2016-05-03
Adjust pid filtering/display in runqlat

drzaeus77:
The filtering and display of pids in the runqlat tool was not correct.
Internally, the kernel keeps pid and tgid, which correspond to thread-id
and user process-id, respectively. The runqlat tool was filtering and
displaying pid instead of tgid.

Change -P and -p options to filter by tgid, and add a new option to give
a breakdown by pid (thread-id).

Update the docs with the -L option.

### DIFF
--- a/tools/runqlat.py
+++ b/tools/runqlat.py
@@ -4,7 +4,7 @@
 # runqlat   Run queue (scheduler) latency as a histogram.
 #           For Linux, uses BCC, eBPF.
 #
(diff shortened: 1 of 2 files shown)


4ast:
lgtm

brendangregg:
thanks!

brendangregg:
... will remember to add -L to the man page later on...

Issue #336, opened 2016-02-05
kprobe/uprobe aliases for function arguments


brendangregg:
Aliases for accessing arguments (eg, arg0, arg1, ...) instead of doing so via registers, so that scripts aren't tied to a particular processor architecture. This is complementary to #225.

References: https://en.wikipedia.org/wiki/X86_calling_conventions#System_V_AMD64_ABI and http://x86-64.org/documentation/abi.pdf

So for x86_64 this may just be: RDI, RSI, RDX, RCX, R8, R9, XMM0–7, unless the compiler messes with them (is that possible?).

brendangregg:
This looks fixed by #529, as we can now use PT_REGS_PARM1..N (from src/cc/export/helpers.h).

4ast:
PT_REGS_PARM1 actually didn't work when I tried it few days ago. Didn't have time to debug

Issue #225, opened 2015-09-17
kretprobe alias for retval


brendangregg:
On kprobe, I can declare and use the function arguments. On kretprobe, is there an alias for the return value? I'm currently using ```ctx->ax```, which is platform dependent.

drzaeus77:
That's it for now. It's a fair ask.

brendangregg:
x86_64 ABI references (these just confirm that %ax is correct to use): https://en.wikipedia.org/wiki/X86_calling_conventions#System_V_AMD64_ABI and http://x86-64.org/documentation/abi.pdf

brendangregg:
This should also work for uretprobe

brendangregg:
Extending on #529, perhaps we just need to add a PT_REGS_RVAL alias?

brendangregg:
We now have PT_REGS_RC(ctx). Closing ticket.

PR #522, opened 2016-05-03
[RFC] Introduce helpers to access pt_regs in an arch-independent manner

rnav:
This should help address #225 and #336.
Tested lightly.

### DIFF
--- a/src/cc/export/helpers.h
+++ b/src/cc/export/helpers.h
@@ -410,5 +410,27 @@ int bpf_num_cpus() asm("llvm.bpf.extra");
 
 #define lock_xadd(ptr, val) ((void)__sync_fetch_and_add(ptr, val))
 
+#ifdef __powerpc__
+#define PT_REGS_PARM1	gpr[3]
+#define PT_REGS_PARM2	gpr[4]
+#define PT_REGS_PARM3	gpr[5]
+#define PT_REGS_PARM4	gpr[6]
+#define PT_REGS_PARM5	gpr[7]
+#define PT_REGS_PARM6	gpr[8]
+#define PT_REGS_RC	gpr[3]
+#define PT_REGS_IP	nip
+#define PT_REGS_SP	sp
+#else
+#define PT_REGS_PARM1	di
+#define PT_REGS_PARM2	si
+#define PT_REGS_PARM3	dx
+#define PT_REGS_PARM4	cx
+#define PT_REGS_PARM5	r8
+#define PT_REGS_PARM6	r9
+#define PT_REGS_RC	ax
+#define PT_REGS_IP	ip
+#define PT_REGS_SP	sp
+#endif
+
 #endif
 )********"
--- a/tests/python/test_trace2.c
+++ b/tests/python/test_trace2.c
@@ -6,7 +6,7 @@ struct Counters { u64 stat1; };
 BPF_TABLE("hash", struct Ptr, struct Counters, stats, 1024);
 
 int count_sched(struct pt_regs *ctx) {
-  struct Ptr key = {.ptr=ctx->bx};
+  struct Ptr key = {.ptr = ctx->PT_REGS_PARM1};
   struct Counters zleaf = {0};
   stats.lookup_or_init(&key, &zleaf)->stat1++;
   return 0;
--- a/examples/lua/bashreadline.c
+++ b/examples/lua/bashreadline.c
@@ -11,11 +11,11 @@ int printret(struct pt_regs *ctx)
 {
 	struct str_t data  = {};
 	u32 pid;
-	if (!ctx->ax)
+	if (!ctx->PT_REGS_RC)
 		return 0;
 	pid = bpf_get_current_pid_tgid();
 	data.pid = pid;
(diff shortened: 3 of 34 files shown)


tests/python/test_trace2.c
```diff
@@ -6,7 +6,7 @@ struct Counters { u64 stat1; };
 BPF_TABLE("hash", struct Ptr, struct Counters, stats, 1024);
 
 int count_sched(struct pt_regs *ctx) {
-  struct Ptr key = {.ptr=ctx->bx};
+  struct Ptr key = {.ptr = ctx->PT_REGS_PARM1};
```
rnav: @drzaeus77: Need feedback on whether this is the right change. I am not sure how bx register is useful here -- I am assuming this was meant to use di (the first and the only argument to finish_task_switch()), but since bx is a non-volatile register, I may be wrong.

goldshtn:
Can we handle platforms where parameters are passed on the stack?

src/cc/export/helpers.h
```diff
@@ -410,5 +410,27 @@ int bpf_num_cpus() asm("llvm.bpf.extra");
...
+#define PT_REGS_PARM1	gpr[3]
+#define PT_REGS_PARM2	gpr[4]
+#define PT_REGS_PARM3	gpr[5]
+#define PT_REGS_PARM4	gpr[6]
+#define PT_REGS_PARM5	gpr[7]
+#define PT_REGS_PARM6	gpr[8]
+#define PT_REGS_RC	gpr[3]
+#define PT_REGS_IP	nip
+#define PT_REGS_SP	sp
+#else
```
goldshtn: Shouldn't this be an `#elif` and the `#else` case would error out?

rnav:
The buildbot failures are to do with clang-format checks which seem to insist on changing the leading tabs on lines to spaces in examples/lua/bashreadline.c. Is that the recommended coding style or is clang-format throwing wrong errors?

rnav:
@goldshtn Good question - that will need more involved macros working off the pt_regs structure. In that case, we may be better off having the macros take pt_regs as an argument and returning parameters or return values.

src/cc/export/helpers.h
rnav: Sure, that's an option, but the current approach errors out too since the x86 specific register names won't be present in other architecture's pt_regs.

tests/python/test_trace2.c
drzaeus77: I can't honestly recall why `ctx->bx` got put in there. This is one of the older test cases. Best guess is that I needed a non-zero value to serve as a key, but the specifics here aren't really important. Probably rather than using the arch-independent macro, the function should be changed to use `u64 prev` in the count_sched prototype, rather than directly pulling from ctx-> directly.

Something like:
```
int count_sched(struct pt_regs *ctx, u64 prev) {
  struct Ptr Key = {.ptr = prev};
```

src/cc/export/helpers.h
drzaeus77: Wouldn't it be cleaner to have an else at the end so that any unsupported architectures are caught with a saner error message?

In either case, I will see about adding a top-level supported architecture check to prevent the compiler trying too hard before getting this far.

aaronpuchert:
Sorry for weighing in here, but I'm not sure if using macros for the field names is a good idea. Partly because macros always create problems, and as @goldshtn pointed out, because some architectures pass parameters via the stack.

An obvious improvement would be to pass the context as a parameter:
```cpp
#define PT_REGS_SP(ctx)	(ctx)->sp
```
to allow accessing parameters on the stack and to avoid notations like `ctx->SOME_MACRO`. (Which looks weird.)

But this can also be written as a bunch of functions, which is (I think) generally preferable:
```cpp
u64 pt_sp(struct pt_regs* ctx) {return ctx->sp;}
```

aaronpuchert:
While we're here: this would be very nice:
```cpp
#ifdef __powerpc__
// I'm indexing the parameters beginning with 0 ...
u64 pt_param(struct pt_regs* ctx, unsigned index)
{
    if (index < 6)
        return ctx->gpr[3+index];
    else {
        // get other parameters from the stack
    }
}
#endif
```

tests/python/test_trace2.c
rnav: Nice. However, it looks like we convert those parameters to point to the proper register in pt_regs. If so, this won't help with architectures that pass parameters on the stack like @goldshtn pointed out?

src/cc/export/helpers.h
rnav: Ok, I will add an #else clause to explicitly throw an error.

rnav:
@aaronpuchert Thanks for weighing in. I agree with your suggestion and I did allude to using a PT_REGS_XYZ(ctx) form for enabling architectures needing to do some work to return the proper parameter value in response to @goldshtn.
I still prefer macros to functions though -- most of the architectures should just need a pt_regs reference for the parameters, so macros seem to be simpler. Can we start with macros now and consider moving to functions as and when the need arises?

rnav:
Opened new pull request #529 with some of the review comments addressed.

PR #524, opened 2016-05-04
Force lua main.c to include luajit headers

drzaeus77:
On my system, when both lua and luajit are installed, multiple headers
exist in /usr/include. However, my lua installation is 5.3, which has
some api-breaking changes, which means that the standard lua.h must not
be included.

Simply force lua.h to come from luajit-2.0 directory.

Cc: @vmg

### DIFF
--- a/src/lua/src/main.c
+++ b/src/lua/src/main.c
@@ -29,9 +29,9 @@
 #include <string.h>
 #include <unistd.h>
 
-#include "lauxlib.h"
-#include "lua.h"
-#include "lualib.h"
+#include <luajit-2.0/lauxlib.h>
+#include <luajit-2.0/lua.h>
+#include <luajit-2.0/lualib.h>
 
 static lua_State *globalL = NULL;
 static const char *progname = NULL;


vmg:
Works on my Ubuntu. I'm not sure if this is gonna break Fedora but...

4ast:
I'm not sure why this is needed. The original code works for me.

drzaeus77:
It likely works for you because your non-jit lua is 5.1. lua 5.2+ is not compatible with luajit 2.0.

http://stackoverflow.com/questions/10087226/lua-5-2-lua-globalsindex-alternative

4ast:
we have many installed. I just do export LUAJIT_DIR= prior to cmake and it seems to pick the right one. Or am I just lucky?

drzaeus77:
Why should such a step be necessary? It makes building bcc a pain.

4ast:
I have a feeling that you'll still need to use luajit_dir, but fine for now.

markdrayton:
I'm not sure this was the correct fix. In our environment we find the LuaJIT headers and libraries by setting `LUAJIT_DIR` to something like `/some/other/prefix/LuaJIT/`, where that directory contains `include/luajit-2.0` and `lib` and so on. `FindLuaJIT.cmake` sets `LUAJIT_INCLUDE_DIR` to `/some/other/prefix/LuaJIT/include/luajit-2.0` and passes that as `-I` to gcc. Compilation fails because `main.c` tells the compiler to look in a `luajit-2.0` subdirectory that can't be found relative to the include directory:

    src/main.c:32:32: fatal error: luajit-2.0/lauxlib.h: No such file or directory
    #include <luajit-2.0/lauxlib.h>
                                   ^

In other words, we have an extra `luajit-2.0/` is in there.

However, as other commenters noticed that this diff worked on their systems. Why? I think it's because LuaJIT's headers are usually found in somewhere like `/usr/include/luajit-2.0` (as on my vanilla Ubuntu host), and `/usr/include` is in the compiler's default include path. In other words, `lua.h` and friends are found through `/usr/include/` (compiler default) + `luajit-2.0/lua.h`, not through `${LUAJIT_INCLUDE_DIR}/luajit-2.0/lua.h`. I just tried to compile `main.c` on my vanilla Ubuntu host and it works even with a totally bogus include path:

    $ rm CMakeFiles/bcc-lua.dir/src/main.c.o
    $ /usr/bin/cc -Wall -O3 -DNDEBUG -I/usr/lib/llvm-3.7/include/../tools/clang/include  -I/home/mark/bcc/src  -I/home/mark/bcc/build/src  -I/blah/blah/include/luajit-2.0  -o CMakeFiles/bcc-lua.dir/src/main.c.o  -c /home/mark/bcc/src/lua/src/main.c
    $ echo $?
    0

I think the correct fix is to adjust `FindLuaJIT.cmake` to pick up the right files. I'm guessing that before this change in @drzaeus77's case it found a non-LuaJIT-2.0 `lua.h` in a path further up the PATHS list here:

https://github.com/iovisor/bcc/blob/master/cmake/FindLuaJIT.cmake#L31

Presumably then the compiler preferred this non-LuaJIT-2.0 path over the system default, used the wrong code and failed to build. After the change I suspect that `LUAJIT_INCLUDE_DIR` is still set to the non-LuaJIT-2.0 path but the compiler doesn't find `luajit-2.0/lua.h` there (note the prefix added by this change) and falls back to trying to get it from the default include path, which happily picks up something like `/usr/include/` + `luajit-2.0/lua.h`.

@drzaeus77, can you a) try diddling `src/lua/CMakeLists.txt` to print `LUAJIT_INCLUDE_DIR` when cmake runs, b) show where all the `lua.h` files are on your system?

drzaeus77:
That analysis sounds correct to me, and sorry for breaking the build on your system. Indeed even before playing with the .cmake file I can tell you that /usr/include/lua.h exists, and contains 5.3 implementation.

drzaeus77:
As expected:

```
$ cmake .. -DCMAKE_INSTALL_PREFIX=/usr | grep LUAJIT_INCLUDE_DIR
-- LUAJIT_INCLUDE_DIR=/usr/include
```

versus

```
$ LUAJIT_DIR=/usr/include/luajit-2.0 cmake .. -DCMAKE_INSTALL_PREFIX=/usr | grep LUAJIT_INCLUDE_DIR
-- LUAJIT_INCLUDE_DIR=/usr/include/luajit-2.0
```

markdrayton:
> sorry for breaking the build on your system

No problem! It was kind of fun to track down and I learned a few things on the way.

Thanks for reverting, and sorry that you now have a bit more hassle while building :-/

Issue #446, opened 2016-03-24
LLVM ERROR with reading names of a dns packet


valkum:
Part of my bachelor thesis is to obtain the names of a dns question. As i got some errors when trying to print them with bpf_trace_print I thought it would be a better way to write them into a map and print them on the python side.

But with my code i get somehow the following LLVM ERROR. Im not sure if this is a bug in bcc or a bug in my code. I read something about #188, but another issues, says this has been fixed in December.

```
LLVM ERROR: Cannot select: 0x1ea6d90: ch,glue = BPFISD::CALL 0x1ea6c68, 0x1ea68f0, 0x1ed2ed0, 0x1dcd680, 0x1ec7398, 0x1ea6c68:1 [ORD=7] [ID=25]
  0x1ea68f0: i64 = TargetExternalSymbol'memset' [ID=10]
  0x1ed2ed0: i64 = Register %R1 [ID=7]
  0x1dcd680: i64 = Register %R2 [ID=8]
  0x1ec7398: i64 = Register %R3 [ID=9]
  0x1ea6c68: ch,glue = CopyToReg 0x1ea60d8, 0x1ec7398, 0x168ca38, 0x1ea60d8:1 [ORD=7] [ID=24]
    0x1ec7398: i64 = Register %R3 [ID=9]
    0x168ca38: i64 = add 0x1ec7cd8, 0x1ed3498 [ORD=6] [ID=21]
      0x1ec7cd8: i64 = and 0x1ed3120, 0x1ed3938 [ORD=5] [ID=19]
        0x1ed3120: i64 = sub 0x1d78c10, 0x1d78d38 [ORD=4] [ID=17]
          0x1d78c10: i64 = Constant<28> [ID=12]
          0x1d78d38: i64 = AssertSext 0x1ea6450, 0x1dccf90 [ORD=2] [ID=16]
            0x1ea6450: i64,ch = CopyFromReg 0x168a420, 0x1ec7a88 [ORD=2] [ID=13]
              0x1ec7a88: i64 = Register %vreg6 [ID=1]
        0x1ed3938: i64 = Constant<65535> [ID=11]
      0x1ed3498: i64 = Constant<1> [ID=4]
    0x1ea60d8: ch,glue = CopyToReg 0x1d78f88, 0x1dcd680, 0x168cc88, 0x1d78f88:1 [ORD=7] [ID=23]
      0x1dcd680: i64 = Register %R2 [ID=8]
[...]

PR #536, opened 2016-05-11
Add new example with test case for for unroll, as requested.

valkum:
Add new example with test case for for unroll, as requested at iovisor-dev@lists.iovisor.org.

### DIFF
--- a/examples/networking/dns_matching/dns_matching.c
+++ b/examples/networking/dns_matching/dns_matching.c
@@ -1,0 +1,105 @@
+/*
+ * dns_matching.c  Drop DNS packets requesting DNS name contained in hash map
+ *    For Linux, uses BCC, eBPF. See .py file.
+ *
+ * Copyright (c) 2016 Rudi Floren.
+ * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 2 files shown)


examples/networking/dns_matching/dns_matching.c
```diff
@@ -0,0 +1,105 @@
...
+        // Do nothing if packet is not a request.
+        if((dns_hdr->flags >>15) != 0) {
+          // Exit if this packet is not a request.
+          return -1;
+        }
+
+        u16 i = 0;
+        struct dns_char_t *c;
+        // This unroll worked not in latest BCC version.
+        for(u8 j = 0; i<255;i++){
```
4ast: I think different llvm versions might do different things here.
The only sure way is to add '#pragma unroll' here.

valkum: I had an issue open (#446) about this and the #pragma unroll seems not to work properly on this.

4ast: got it. thanks for the test.

PR #539, opened 2016-05-11
Lua ports of biosnoop and stacksnoop

markdrayton:
I ported a couple of tools while testing out the Lua stuff.

What should we do about tools now that we have Python and Lua wrappers? For me Lua is easier to manage in production as bcc-lua is completely static but it might get confusing if we have two potentially unsynchronized implementations of each tool.

### DIFF
--- a/tools/biosnoop.lua
+++ b/tools/biosnoop.lua
@@ -1,0 +1,183 @@
+#!/usr/bin/env bcc-lua
+--[[
+Copyright 2016 GitHub, Inc
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+--]]
+
+local program = [[
+#include <uapi/linux/ptrace.h>
+#include <linux/blkdev.h>
+
+struct val_t {
+    u32 pid;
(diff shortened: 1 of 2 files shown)


4ast:
looks great to me.
I think in this case biosnoop.lua and stacksnoop.lua are fully equivalent to their python versions, so probably should live in tools/ directory. Then hopefully whoever modifies foo.py will not forget to do the same in foo.lua and may be eventually we'll just get rid of .py versions of biosnoop and stacksnoop if bcc-lua stands up to its potential.

mcaleavya:
Quick? should the original copyrights / updates not be included in the lua version as the c code is replicated? And looks like I need to start learning lua now if that's the way forward for the tools.

examples/lua/biosnoop.lua
```diff
@@ -0,0 +1,182 @@
...
+    "DISK", "T", "SECTOR", "BYTES", "LAT(ms)"})
+
+  local rwflg = ""
+  local start_ts = 0
+  local prev_ts = 0
+  local delta = 0
+
+  local function print_event(cpu, event)
+    local val = -1
+    local event_pid = tonumber(event.pid)
```
vmg: You don't really need to call `tonumber` on any of these variables. The JIT will automatically convert them to native Lua numbers. `tonumber` is only useful for `uint64_t` data because it cannot be trivially converted to a Lua integer, so calling it will force the loss of precision.

vmg:
Looks like a beautiful port. Thank you!

markdrayton:
I moved the tools to tools/, fixed the superfluous `tonumber()`, and wrapped the lines at 80 cols.

What's the etiquette for updating a PR, BTW? In this case it doesn't seem helpful to have two commits so I just squashed locally and force pushed the branch again.

markdrayton:
Removed the unneeded `ffi` import from `stacksnoop.lua`. This is all good to go now.

4ast:
squash and force update is the right call. Thanks

PR #546, opened 2016-05-20
Fix a few fd leaks

iamkafai:
This diff fixes a few fd leaks after reading files under /sys/kernel/debug/tracing.


drzaeus77:
Seems obvious in retrospect. Thanks.

4ast:
yeah goto fail; goto fail :)

Issue #547, opened 2016-05-21
PIDs and TIDs in filters and output


markdrayton:
Lots of the scripts in `tools/` take a PID argument to apply filtering to a particular process. The filtering is usually on `task->pid` (i.e. the lower 32 bits of `bpf_get_current_pid_tgid()`) and consequently only matches a single thread within a process. IME most userspace apps treat PID as TGID, rather than a specific thread within the process (e.g. think `pgrep`, `ps aux`, etc).

Sometime it's handy to filter events by a single task but very often I want to do something for a whole process without filtering by threads within it. E.g., if I wanted to see which stacks in my webserver are doing block IO I'd run `stacksnoop -p $(pgrep webserver) submit_bio` and expect it to match all threads by default. If I was doing this in `perf` its `-p` option would match the process, not a particular thread (there is `-t` to limit to certain threads).

Any objections to changing tools to:

a) make `-p PID` filter on TGID
b) adding a `-t TID` option to limit to a single task (i.e. what `-p` does now)
c) print both TID and PID in output

I realise this is changing existing behavior but I'd argue it'd be a net improvement overall.

4ast:
makes sense to me. I think we discussed pid/tgid controversy few times.
Though it may be a bit confusing to script writers that kernel pid is compared with command line tid
and kernel tgid is compared with command line pid.

brendangregg:
Makes sense to me. I think -p should match whatever pidstat says is the PID, and -t should match what "pidstat -t" says is the TID.

I likely got this wrong many, many times (copy-n-pasted) in /tools, sorry. Would be happy to see it fixed!

As for printing both PID and TID, definitely print TID if -t is used (as the user has signaled that it is important). As for other tools and default output, I guess it would depend on the tool. I try to keep the default output to a minimum. Some tools probably have -v for verbose. Plus I also try to keep the default output to < 80 chars width.

Eg, execsnoop default output probably shouldn't have TID. Nor should killsnoop. ... Actually, I probably wouldn't add TID to most of the scripts' default output. It could show up with -t, or -v.

This will require updates to the /examples files and man pages...

Issue #540, opened 2016-05-11
Replace use of trace pipe with bpf_perf_event


markdrayton:
BCC tools that read data from the trace pipe are often confused when they come across unexpected output from other tracepoints. Where possible, we should replace use of the trace pipe with perf_event_event. It looks like only 5 scripts still use it (via the `trace_fields` Python function):

    $ grep trace_fields tools/*.py                                                                                          
    tools/dcsnoop.py:    (task, pid, cpu, flags, ts, msg) = b.trace_fields()
    tools/execsnoop.py:    (task, pid, cpu, flags, ts, msg) = b.trace_fields()
    tools/fileslower.py:    (task, pid, cpu, flags, ts, msg) = b.trace_fields()
    tools/memleak.py:                print(bpf_program.trace_fields())
    tools/stacksnoop.py:    (task, pid, cpu, flags, ts, msg) = b.trace_fields()

brendangregg:
Now down to 2:

```
$ grep trace_fields tools/*.py    
tools/dcsnoop.py:    (task, pid, cpu, flags, ts, msg) = b.trace_fields()
tools/memleak.py:                print(bpf_program.trace_fields())
```

markdrayton:
I'm not sure it makes sense to convert memleak.py as it only uses the trace pipe to print information. Nothing's parsing the output and converting the script to use the perf system will add a good chunk more code.

goldshtn:
Yep, and besides the printing isn't actually useful for anything but debugging the script itself, because of the immense overhead of these printouts.

markdrayton:
Perfect. Let's call this one done then.

PR #550, opened 2016-05-24
Use BPF_PERF_OUTPUT for fileslower and stacksnoop

markdrayton:
This PR updates `fileslower` and  `stacksnoop` (both Python and Lua) to use BPF_PERF_OUTPUT instead of the trace pipe. The only noticeable differences in output are:

a) `COMM` is more reliable. More often than not when reading the trace pipe this would be set to `<...>`. `bpf_get_current_comm()` works all the time, as far as I can tell.

b) The `TIME(s)` column is relative to the program's start, rather than the timestamp embedded in the lines read from the trace pipe.

c) Renamed the `SYSCALL` column to `FUNCTION` since we may not be tracing a syscall.

This goes towards #540. I didn't fix the PID/TID issue I raised in #547 — will do that in another PR.

### DIFF
--- a/tools/stacksnoop.lua
+++ b/tools/stacksnoop.lua
@@ -17,16 +17,29 @@ limitations under the License.
 
 local program = [[
 #include <uapi/linux/ptrace.h>
+#include <linux/sched.h>
+
+struct data_t {
+    u64 stack_id;
+    u32 pid;
+    char comm[TASK_COMM_LEN];
+};
 
 BPF_STACK_TRACE(stack_traces, 128)
+BPF_PERF_OUTPUT(events);
 
 void trace_stack(struct pt_regs *ctx) {
+    u32 pid = bpf_get_current_pid_tgid();
     FILTER
-    int stack_id = stack_traces.get_stackid(ctx, BPF_F_REUSE_STACKID);
-    if (stack_id >= 0)
-        bpf_trace_printk("stack_id=%d\n", stack_id);
+    struct data_t data = {};
+    data.stack_id = stack_traces.get_stackid(ctx, BPF_F_REUSE_STACKID),
+    data.pid = pid;
+    bpf_get_current_comm(&data.comm, sizeof(data.comm));
+    events.perf_submit(ctx, &data, sizeof(data));
 }
 ]]
+
+local ffi = require("ffi")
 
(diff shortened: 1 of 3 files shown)


vmg:
Love how you've cleaned up the Lua code. 👌 😄

drzaeus77:
LGTM

markdrayton:
I didn't copy `stacksnoop.py` to the old dir as we already have an old one (pre-`get_stackid()`) there. I'm happy to copy one or both of the old versions of there tools to `old/` if we want that, though.

drzaeus77:
The change in this PR requires 4.3 features only, so if anything this update could be added into the version in old/, but I don't think having a 3rd version is necessary. I'd say let the change stand.

tools/stacksnoop.lua
```diff
@@ -41,52 +54,58 @@ return function(BPF, utils)
...
+
+  local TASK_COMM_LEN = 16 -- linux/sched.h
+
+  bpf:get_table("events"):open_perf_buffer(print_event, [[
+    struct {
+      uint64_t stack_id;
+      uint32_t pid;
+      char comm[%d];
+    }
+  ]] % {TASK_COMM_LEN})
```
4ast: imo the c->lua binding are less verbose and cleaner than c->py. Just curious luajit dislikes 'char comm[TASK_COMM_LEN]' because the macro is undefined? any other way to pass it down to lua other than this '%' approach?
overall LGTM. thank you for fixing it all.

markdrayton: Yeah, it blows up thusly without it:

    [ERROR] bcc.lua:2737: unexpected symbol near 'TASK_COMM_LEN' at line 4
    stack traceback:
            [C]: in function 'typeof'
            bcc.lua:2737: in function 'open_perf_buffer'
            tools/stacksnoop.lua:103: in function <tools/stacksnoop.lua:44>
            [C]: in function 'xpcall'
            bcc.lua:3014: in function <bcc.lua:2963>
            [C]: at 0x0074e4f0

@vmg any suggestions to make this cleaner?

vmg: https://github.com/iovisor/bcc/pull/551 is the best I can do. Explanation included. :)

PR #552, opened 2016-05-26
Install tools/old

evverx:
* Install `tools/old` too
* Point people to `tools/old`

This is a followup for #450 

@brendangregg , @vmg, please have a look


4ast:
makes sense

evverx:
I created a wrapper:
```diff
diff --git a/tools/offcputime.py b/tools/offcputime.py
index 751349f..7b75dcb 100755
--- a/tools/offcputime.py
+++ b/tools/offcputime.py
@@ -16,6 +16,19 @@ from sys import stderr
 from time import sleep, strftime
 import argparse
 import signal
+import sys
+import os
+
+uname = os.uname()
+try:
+    release = uname.release()
+except AttributeError:
+    release = uname[2]
+
+major, minor = map(int, release.split('.')[:2])
+if major < 4 or minor < 6:
+    old_tool = os.path.join(os.path.dirname(__file__), 'old', os.path.basename(__file__))
+    os.execl(old_tool, old_tool, *sys.argv[1:])

 # arg validation
 def positive_int(val):
```
Maybe, this is a better option.

brendangregg:
It's nice and short, but I'm a bit worried about going too far in that direction, and having nasty scripts like you see sometimes in SystemTap, which tries to work across multiple kernel versions.

Could this lump of code be in __init__.py, and then called with a single BPF.some_unique_name()? Then it's not too messy, and in the distant future we can search and delete them all when they are no longer needed.

4ast:
version based checks don't work. Not too long ago we deleted all of them. Kernel version says nothing about actual feature availability. Backports are common.

evverx:
Agree.
So,
> This tool only works on Linux 4.6+. Stack traces are obtained using the new BPF_STACK_TRACE` APIs. For kernels older than 4.6, see the version under tools/old.

is a bit misleading. But it doesn't really matter, I think.

4ast:
doc saying 4.6+ is a reasonable suggestion. For most 4.6+ kernels bpf_stack_trace will be there,
but it doesn't mean that the newer tool will never work for older kernels. imo it's useful to give such guidance in the doc, but it's not ok to do rely on version in the code.

PR #554, opened 2016-05-27
Fix tools[/old]/offcputime -p PID

evverx:
We should extract pid before THREAD_FILTER

### DIFF
--- a/tools/offcputime.py
+++ b/tools/offcputime.py
@@ -96,12 +96,11 @@ BPF_HASH(start, u32);
 BPF_STACK_TRACE(stack_traces, STACK_STORAGE_SIZE)
 
 int oncpu(struct pt_regs *ctx, struct task_struct *prev) {
-    u32 pid;
+    u32 pid = prev->pid;
     u64 ts, *tsp;
 
     // record previous thread sleep time
     if (THREAD_FILTER) {
-        pid = prev->pid;
         ts = bpf_ktime_get_ns();
         start.update(&pid, &ts);
     }
--- a/tools/old/offcputime.py
+++ b/tools/old/offcputime.py
@@ -89,12 +89,11 @@ static u64 get_frame(u64 *bp) {
 }
 
 int oncpu(struct pt_regs *ctx, struct task_struct *prev) {
-    u32 pid;
+    u32 pid = prev->pid;
     u64 ts, *tsp;
 
     // record previous thread sleep time
     if (FILTER) {
-        pid = prev->pid;
         ts = bpf_ktime_get_ns();
         start.update(&pid, &ts);
     }


evverx:
`-p` was broken by d364d044f912eab
See https://github.com/iovisor/bcc/commit/d364d044f912eab552399bea00d9713da06a3568#diff-3cb26705b3e0cad7984a1d66f90950fbL93

@brendangregg , please have a look

evverx:
Rebased.
(actually, this is the same as https://github.com/iovisor/bcc/pull/438)

tools/offcputime.py
```diff
@@ -96,12 +96,11 @@ def signal_ignore(signal, frame):
 BPF_STACK_TRACE(stack_traces, STACK_STORAGE_SIZE)
 
 int oncpu(struct pt_regs *ctx, struct task_struct *prev) {
-    u32 pid;
+    u32 pid = prev->pid;
```
4ast: shouldn't it be 'pid = bpf_get_current_pid_tgid();' instead?
moving line 109 into here...

evverx: At this point `prev->pid` is "offcpu", `bpf_get_current_pid_tgid` is "oncpu". We update "offcpu"-start-time table:
```c
    if (THREAD_FILTER) {
        ts = bpf_ktime_get_ns();
        start.update(&pid, &ts);
    }
```
so, we really need `prev->pid` here.

4ast: I see. Will we be wasting time doing lookup for current_pid down in lines 109-111 that is not going to succeed? I guess that can be an optimization for the future?
Anyway I'm fine with it as-is.
@brendangregg

brendangregg:
will hopefully check this out soon; at a conference and don't have access to my bpf testing server. Setting up another one...

evverx:
@brendangregg , ok

Rebased.

My tests:

* before
```sh
$ stty -tostop

$ sudo /usr/share/bcc/tools/old/offcputime -f -p $$ 10 &
[1] 17752

$ sleep 3
$
[1]+  Done                    sudo /usr/share/bcc/tools/old/offcputime -f -p $$ 10
```

* after:
```sh
$ stty -tostop

$ sudo /usr/share/bcc/tools/old/offcputime -f -p $$ 10 &
[1] 18566

$ sleep 3
$
bash;int_ret_from_sys_call;syscall_return_slowpath;exit_to_usermode_loop;schedule 3219
bash;__key.26431;entry_SYSCALL_64_fastpath;SyS_read;vfs_read;__vfs_read;tty_read;n_tty_read;wait_woken;schedule_timeout;schedule 1619339
bash;entry_SYSCALL_64_fastpath;sys_wait4;do_wait;schedule 3003124

[1]+  Done                    sudo /usr/share/bcc/tools/old/offcputime -f -p $$ 10
```

brendangregg:
Tested, works, thanks! Except the user/kernel stack order I think is backwards -- easy fix, I'll file a separate ticket :)

PR #555, opened 2016-05-27
execsnoop: use BPF_PERF_OUTPUT, add PPID

markdrayton:
This adjusts execsnoop to use BPF_PERF_OUTPUT rather than `bpf_trace_printf()`. It's WIP because I can't get it to run with all the calls to `submit_arg` loop uncommented. If I uncomment line 99 I get:

    $ sudo tools/execsnoop.py 
    bpf: Permission denied
    0: (bf) r6 = r1
    1: (79) r1 = *(u64 *)(r6 +112)
    2: (7b) *(u64 *)(r10 -528) = r1
    invalid stack off=-528 size=8

I assume this means we're trampling on the stack somewhere. Any suggestions on how to work around this?

(this goes towards #540)

### DIFF
--- a/tools/execsnoop.py
+++ b/tools/execsnoop.py
@@ -145,7 +145,7 @@ b = BPF(text=bpf_text)
 # header
 if args.timestamp:
     print("%-8s" % ("TIME(s)"), end="")
-print("%-16s %-6s %3s %s" % ("PCOMM", "PID", "RET", "ARGS"))
+print("%-16s %-6s %-6s %3s %s" % ("PCOMM", "PID", "PPID", "RET", "ARGS"))
 
 TASK_COMM_LEN = 16      # linux/sched.h
 ARGSIZE = 128           # should match #define in C above
@@ -166,6 +166,19 @@ class EventType(object):
 start_ts = time.time()
 argv = defaultdict(list)
 
+# TODO: This is best-effort PPID matching. Short-lived processes may exit
+# before we get a chance to read the PPID. This should be replaced with fetching
+# PPID via C when available (#364).
+def get_ppid(pid):
+    try:
+        with open("/proc/%d/status" % pid) as status:
+            for line in status:
+                if line.startswith("PPid:"):
+                    return int(line.split()[1])
+    except IOError:
+        pass
+    return 0
+
 # process event
 def print_event(cpu, data, size):
     event = ct.cast(data, ct.POINTER(Data)).contents
@@ -183,7 +196,9 @@ def print_event(cpu, data, size):
         if not skip:
             if args.timestamp:
                 print("%-8.3f" % (time.time() - start_ts), end="")
-            print("%-16s %-6s %3s %s" % (event.comm, event.pid, event.retval,
+            ppid = get_ppid(event.pid)
+            print("%-16s %-6s %-6s %3s %s" % (event.comm, event.pid,
+                    ppid if ppid > 0 else "?", event.retval,
                     ' '.join(argv[event.pid])))
 
         del(argv[event.pid])
--- a/tools/execsnoop.py
+++ b/tools/execsnoop.py
@@ -19,7 +19,10 @@
 from __future__ import print_function
 from bcc import BPF
 import argparse
+import ctypes as ct
 import re
+import time
+from collections import defaultdict
 
 # arguments
 examples = """examples:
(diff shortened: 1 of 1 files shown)


markdrayton:
@4ast suggested that making one `struct data_t` and passing it to `submit_arg()` via a pointer might be more successful. It is! We can can now print all 20 arguments. To celebrate, I bumped the ARGSIZE limit to 128 to increase the odds of printing full arguments (I saw some truncated args when using the current execsnoop in production). This works fine:

    # in one terminal
    $ ls $(python -c "import string; print ' '.join(l * 256 for l in string.ascii_lowercase)")

    # in another
    $ sudo tools/execsnoop.py                                                                         
    python           3560   1714     0 /usr/bin/python -c import string; print ' '.join(l * 256 for l in string.ascii_lowercase)
    ls               3561   1714     0 /bin/ls --color=auto aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa <snip> rrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrr ...

I also added rudimentary support for getting the PPID of a process that execve()d, albeit in a slightly hacky `/proc`-scraping way. Personally I find this pretty useful but if it's deemed too hacky I can remove it.

I'm done with tinkering here so feel free to merge or return for changes when ready.

tools/execsnoop.py
```diff
@@ -70,42 +83,54 @@
...
+    if (submit_arg(ctx, (void *)&__argv[i], &data) == 0) goto out; i++;
+    if (submit_arg(ctx, (void *)&__argv[i], &data) == 0) goto out; i++;
+    if (submit_arg(ctx, (void *)&__argv[i], &data) == 0) goto out; i++;
+    if (submit_arg(ctx, (void *)&__argv[i], &data) == 0) goto out; i++;
+    if (submit_arg(ctx, (void *)&__argv[i], &data) == 0) goto out; i++; // XX
+
+    // handle truncated argument list
+    char ellipsis[] = "...";
+    bpf_probe_read(data.argv, sizeof(data.argv), (void*)&ellipsis);
+    events.perf_submit(ctx, &data, sizeof(data));
```
markdrayton: I also tried:

    char ellipsis[] = "...";
    submit_arg(ctx, (void *)&ellipsis, &data);

but for whatever reason `bpf_probe_read()` seems to read an empty string into `argp`.

tools/execsnoop.py
```diff
@@ -70,42 +83,54 @@
...
-    bpf_probe_read(&fname, sizeof(fname), (void *)(filename));
-    bpf_trace_printk("ARG %s\\n", fname);
+    // create data here and pass to submit_arg to save space on the stack (#555)
+    struct data_t data = {};
+    data.pid = bpf_get_current_pid_tgid() >> 32;
+    bpf_get_current_comm(&data.comm, sizeof(data.comm));
+    data.type = EVENT_ARG;
 
-    int i = 1;  // skip first arg, as we printed fname
+    submit_arg(ctx, (void *)&filename, &data);
```
4ast: can we split submit_arg() into two functions, so we don't have to do '&filename' here?
since extra probe_read is not needed in this case and makes the code confusing to read.
Like
int submit_arg(ctx, ptr, data) {
 const char *argp;
bpf_probe_read(&argp, sizeof(argp), ptr);
if (argp)
 return __submit_arg(ctx, argp, data);
}
and call __ version here ?

tools/execsnoop.py
4ast: since ellipsis is an array, &ellipsis doesn't take another ref, so first call to probe_read in
submit_arg() will read 8 bytes of the string instead of the pointer.
If '&' is removed and __submit_arg() called then it should work.

markdrayton:
Updated with @4ast's suggestions (thanks!)

4ast:
looks great. thanks!

Issue #559, opened 2016-06-06
offcputime user/kernel stack order


brendangregg:
```
# ./offcputime.py -p 21077
Tracing off-CPU time (us) of PID 21077 by user + kernel stack... Hit Ctrl-C to end.
^C
    00007f4ff2a4f810 __read
    000000000040866f read_token
    000000000040816d yylex
    000000000040580e yyparse
    000000000040546b parse_command
    000000000040553c read_command
    0000000000405270 reader_loop
    00000000004038e7 main
    00007f4ff2985ec5 __libc_start_main
    ffffffff810a16d1 finish_task_switch
    ffffffff817b3255 schedule
    ffffffff817b5d1b schedule_timeout
    ffffffff810bc0f3 wait_woken
    ffffffff814b6308 n_tty_read
    ffffffff814b0df2 tty_read
    ffffffff81205168 __vfs_read
    ffffffff81205fcf vfs_read
    ffffffff812073c6 sys_read
    ffffffff817b6d76 entry_SYSCALL_64_fastpath
    -                bash (21077)
        124267
```

Note that the output begins with user, then shows kernel. I think it should be the other way around, kernel first, then user, to preserve the logical code flow from the bottom of the stack to the top. You always begin with "main" at the bottom, then work up, and sometimes enter the kernel.

Same with folded:

```
# ./offcputime.py -fp 21077
^Cbash;entry_SYSCALL_64_fastpath;sys_read;vfs_read;__vfs_read;tty_read;n_tty_read;wait_woken;schedule_timeout;schedule;__libc_start_main;main;reader_loop;read_command;parse_command;yyparse;yylex;read_token 239532
```

Oops, another bug (maybe mine): note that the folded output is missing "__read" and "finish_task_switch" -- must be an off-by-one bug in the loops.

I'd also add an option to insert a delimiter frame, "-", which is supported by flamegraph.pl (it colors it grey). Eg, as -d:

```
# ./offcputime.py -d -p 21077
Tracing off-CPU time (us) of PID 21077 by user + kernel stack... Hit Ctrl-C to end.
^C
    ffffffff810a16d1 finish_task_switch
    ffffffff817b3255 schedule
    ffffffff817b5d1b schedule_timeout
    ffffffff810bc0f3 wait_woken
    ffffffff814b6308 n_tty_read
    ffffffff814b0df2 tty_read
    ffffffff81205168 __vfs_read
    ffffffff81205fcf vfs_read
    ffffffff812073c6 sys_read
[...]

PR #561, opened 2016-06-07
offcputime -f: don't miss backtrace bounds

evverx:
See #559

### DIFF
--- a/tools/offcputime.py
+++ b/tools/offcputime.py
@@ -225,8 +225,8 @@ for k, v in sorted(counts.items(), key=lambda counts: counts[1].value):
 
     if folded:
         # print folded stack output
-        user_stack = list(user_stack)[1:]
-        kernel_stack = list(kernel_stack)[1:]
+        user_stack = list(user_stack)
+        kernel_stack = list(kernel_stack)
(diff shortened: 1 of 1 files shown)


tools/offcputime.py
```diff
@@ -225,8 +225,8 @@ def signal_ignore(signal, frame):
 
     if folded:
         # print folded stack output
-        user_stack = list(user_stack)[1:]
-        kernel_stack = list(kernel_stack)[1:]
+        user_stack = list(user_stack)
+        kernel_stack = list(kernel_stack)
```
evverx: hm, I should remove these lines totally. Will fix

evverx: no, I was wrong:)
```
argument to reversed must be a sequence
```

brendangregg:
thanks; I think I remembered why it was done originally -- to elide the redundant "finish_task_switch" frame from the top of all stacks in off-CPU flame graphs. However, -f means "folded" and not "folded and filtered", so I think the former behavior was confusing (it confused me!). As for filtering finish_task_switch, that can be done in another separate command in the pipeline (sed/awk) after "offcputime -f"...

PR #563, opened 2016-06-08
libbpf.c: Prepend BPF syscall error to log_bug when debug is enabled.

qmonnet:
Commit 759029fea8066b41b54be5447137db95cb1313c4 provided an option to store the output from BPF syscall into a buffer (and not to print it systematically to standard output) on program load in libbpf.c.

But doing so, it only stores the content of attr.log_buf, while the error string resulting from a failed BPF syscall is no more displayed when the DEBUG_BPF flag is used in the Python script responsible to
convert and to inject the code.

The commit in this PR proposes a fix to this bug by prepending the error string to the log_buf buffer when the latter exists, so that the error can also be displayed when DEBUG_BPF flag has been set to true.
When no buffer is provided to bpf_prog_load(), the message error is prepended during the second (recursive) call and then displayed to stderr once this second call has returned.

----

To reproduce the bug, one can use the following snippet to call a BPF filter:

    #!/usr/bin/env python
    # Run with e.g. `./my_script.py my_bpf_prog.c`

    from bcc import BPF, DEBUG_BPF
    import sys

    b = BPF(src_file=sys.argv[1], debug=DEBUG_BPF)
    fn = b.load_func("ebpf_filter", BPF.SCHED_CLS)

If one tries to run this code without root permission,
- *Without* flag DEBUG_BPF: `bpf: Operation not permitted` is displayed.
- *With* flag DEBUG_BPF: nothing (only the Python backtrace). This is
  because a specific log_buf buffer is passed to the C function
  bpf_prog_load(), where the error string is displayed only if no buffer
  has been provided.

----

Note: It might be simpler / cleaner to append the error string to the buffer instead of prepending it. I chose to display the information (error string and then buffer filled in the kernel) in the same order as
it was before, but maybe we can rearrange it.

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -211,7 +211,7 @@ class BPF(object):
             print(log_buf.value.decode(), file=sys.stderr)
 
         if fd < 0:
-            if self.debug & DEBUG_BPF and not log_buf.value:
+            if self.debug & DEBUG_BPF:
                 errstr = os.strerror(ct.get_errno())
                 raise Exception("Failed to load BPF program %s: %s" %
                                 (func_name, errstr))
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -207,11 +207,16 @@ class BPF(object):
                 lib.bpf_module_kern_version(self.module),
                 log_buf, ct.sizeof(log_buf) if log_buf else 0)
 
-        if self.debug & DEBUG_BPF:
+        if self.debug & DEBUG_BPF and log_buf.value:
             print(log_buf.value.decode(), file=sys.stderr)
 
         if fd < 0:
-            raise Exception("Failed to load BPF program %s" % func_name)
+            if self.debug & DEBUG_BPF and not log_buf.value:
+                errstr = os.strerror(ct.get_errno())
+                raise Exception("Failed to load BPF program %s: %s" %
+                                (func_name, errstr))
+            else:
+                raise Exception("Failed to load BPF program %s" % func_name)
 
         fn = BPF.Function(self, func_name, fd)
         self.funcs[func_name] = fn


src/cc/libbpf.c
```diff
@@ -173,12 +173,22 @@ int bpf_prog_load(enum bpf_prog_type prog_type,
...
   }
 
-  if (ret < 0 && !log_buf) {
-    // caller did not specify log_buf but failure should be printed,
-    // so call recursively and print the result to stderr
-    bpf_prog_load(prog_type, insns, prog_len, license, kern_version,
-        bpf_log_buf, LOG_BUF_SIZE);
-    fprintf(stderr, "bpf: %s\n%s\n", strerror(errno), bpf_log_buf);
+  if (ret < 0) {
+    const char *str_error = strerror(errno);
```
4ast: strerror is not thread safe. pls use strerror_r.
Also I don't think the malloc/free pain is worth it.
When there is output in log_buf, it's self explanatory.
Additional errno string will only be confusing,
so I think it makes sense to fill it into the buffer only when log_buf stayed empty after syscall.

drzaeus77: How about saving `errno` in a local variable after the first failure, and then changing the caller (py/lua) to display strerror in their own manner. For instance, `attach_raw_socket()` in `__init__.py` already does this.

4ast: how about changing all syscall wrappers to directly return errno ?
the -1 return code isn't useful anyway.

drzaeus77: Sure, but that would be outside scope of this bugfix. For now I would adopt the existing convention of `attach_raw_socket`.

qmonnet:
Hi @4ast, @drzaeus77, and thank you for your feedback!

Last commit is a new version of the fix, in which the `errstr` is printed by the Python caller, based on what `attach_raw_socket()` already does—as suggested by @drzaeus77.

Following @4ast's first remark, the error string is displayed only if `log_buf` is empty. But I tend to believe that since it is now printed by the Python script, the distinction from the rest of the `log_buf` is clearer than before (if nothing else, there is the Python backtrace between the logs and the error string, as we raise a Python exception), and it might be cleaner in the Python code to print it even if `log_buf` contains some data?

4ast:
@qmonnet you mean to drop 'and not log_buf.value' from current patch?
I think it's fine both ways.

qmonnet:
@4ast yes, this is exactly what I meant. It's in the commit I have just pushed, if you want to add it to the fix.

(Note: the commit log assumes it gets squashed with the previous commits of the PR before merging).

drzaeus77:
> (Note: the commit log assumes it gets squashed with the previous commits of the PR before merging).
You'll want to do this from your side, by `git push -f`to this same branch. Github will automatically update the PR, while still preserving the review comment chain.

Otherwise the PR LGTM.

4ast:
Out of 4 commits the last two I think is ok to keep separate.
but libbpf.c change plus revert change shouldn't be committed.
So pls squash either all of them or remove the first two.

qmonnet:
@drzaeus77 Indeed I feared the forced update could suppress the comments from the PR. Thanks for the precision!

So I rebased my branch with the last two commits on top of `master` branch, and updated commit logs. Now it should be ready to be merged.

Issue #548, opened 2016-05-22
Support resolving addresses via /tmp/perf-pid.map


markdrayton:
It's common for JITs (Java, HHVM, etc) to write a `/tmp/perf-<pid>.map` to map addresses to JITed code. BCC should support symbolizing using this data.

brendangregg:
Also consider container environments, where we are running bcc/BPF from the host, and the container /tmp/perf*.map files are in other directories for each container, and the PID may not match (what the container guest thinks is the PID, is not the same as what the host thinks is the PID).

markdrayton:
> Also consider container environments, where we are running bcc/BPF from the host, and the container /tmp/perf*.map files are in other directories for each container ..

Yes, I've been encountering this complication a lot recently :-/ I have a working diff that reads symbols from `/tmp/perf-pid.map` but I need to put some more effort into making it work with containers/different namespaces. I'll try to find time to wrap it up.

abirchall:
>  I have a working diff that reads symbols from /tmp/perf-pid.map but I need to put some more effort into making it work with containers/different namespaces. I'll try to find time to wrap it up.

Is it worth getting some basic PID support out (for non-container environments) first and iterating?
I ask this question partly as a learning opportunity because I'm not sure whether iteration like this is encouraged / discouraged in the open source community.

I'm also curious what we'd like to support and at what level. LXC? Chroot? Namespace? Cgroup?

markdrayton:
> I'm also curious what we'd like to support and at what level. LXC? Chroot? Namespace? cgroup?

I think we just need to support finding the `/tmp/perf-pid.map` when the target process isn't in the same PID namepace or filesystem root. We can find the PID inside the target namespace by reading `NSpid` in `/proc/pid/status` on the outside, and we get get the path to the process's root dir from `readlink /proc/pid/root`. I'm not aware of any special handling needed for LXC or cgroups (but please shout if there is something I missed!)

brendangregg:
Should we close this issue now?

markdrayton:
Whoops, yes.

PR #573, opened 2016-06-15
Add support for reading symbols from /tmp/perf-pid.map

markdrayton:
For #548, this adds basic support for reading symbols from `/tmp/perf-pid.map` (e.g., from a JIT or other code relocation). To cope with processes in containers, it supports:

* mapping from BCC's PID namespace to the target process's PID namespace
  using `/proc/pid/status`
* resolving a target process's root filesystem using `/proc/pid/root`

Note that mapping to a target's PID in a separate namespace requires a `NSpid:` line in `/proc/pid/status`. IIRC this was added sometime in 4.0 but I can't remember exactly when.

### DIFF
--- a/src/cc/bcc_perf_map.c
+++ b/src/cc/bcc_perf_map.c
@@ -1,0 +1,102 @@
+/*
+ * Copyright (c) 2016 Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include <ctype.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "bcc_perf_map.h"
+
+int bcc_perf_map_nspid(int pid) {
+  char status_path[64];
+  FILE *status;
+
+  snprintf(status_path, sizeof(status_path), "/proc/%d/status", pid);
+  status = fopen(status_path, "r");
+
+  if (!status)
+    return -1;
+
+  // return the original PID if the NSpid line is missing
+  int nspid = pid;
+
+  size_t size;
+  char *line = NULL;
+  while (getline(&line, &size, status) != -1)
+    if (strstr(line, "NSpid:") != NULL)
+      // PID namespaces can be nested -- last number is innermost PID
+      nspid = (int)strtol(strrchr(line, '\t'), NULL, 10);
+  free(line);
+
+  return nspid;
+}
+
+bool bcc_perf_map_path(char *map_path, size_t map_len, int pid) {
+  char source[64];
+  snprintf(source, sizeof(source), "/proc/%d/root", pid);
+
+  char target[4096];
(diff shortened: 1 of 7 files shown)


src/cc/bcc_perf_map.h
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright (c) 2016 GitHub, Inc.
```
4ast: copy paste? ;)

markdrayton:
Argh! Yes, copy/paste fail :-/ Fixed now.

src/cc/bcc_perf_map.h
markdrayton: Whoops, yes. Fixed now.

4ast:
LGTM
@vmg pls take a look

src/cc/bcc_perf_map.c
```diff
@@ -0,0 +1,102 @@
...
+
+  if (!status)
+    return -1;
+
+  // return the original PID if the NSpid line is missing
+  int nspid = pid;
+
+  size_t size;
+  char *line = NULL;
+  while (getline(&line, &size, status) != -1)
```
vmg: Not quite right here. According to POSIX.1-200, `getline` will only allocate a new buffer to store the line when `line == NULL` _and_ `size == 0`. Since `size` is uninitialized here, this can break in really unexpected ways.

On top of that, `getline` allocates a buffer for each call (even in failure cases), and you're calling `free` only once outside the loop, hence leaking significant amounts of memory.

Please fix the initialization and deallocation, and add some curly braces to prevent further bugs like this in the future. Thanks! :)

src/cc/bcc_perf_map.c
```diff
@@ -0,0 +1,102 @@
...
+    while (*cursor && isspace(*cursor)) cursor++;
+
+    newline = strchr(cursor, '\n');
+    if (newline)
+        newline[0] = '\0';
+
+    callback(cursor, begin, len - 1, 0, payload);
+  }
+
+  free(line);
```
vmg: Again, leaking memory here. `getline` always allocates a buffer.

vmg:
Sorry I couldn't get to review this earlier. @markdrayton: I quite dig the tests, but you're not using `getline` properly. See the attached comments.

src/cc/bcc_perf_map.c
```diff
@@ -0,0 +1,102 @@
...
+}
+
+int bcc_perf_map_foreach_sym(const char *path, bcc_perf_map_symcb callback,
+                             void* payload) {
+  FILE* file = fopen(path, "r");
+  if (!file)
+    return -1;
+
+  char *line = NULL;
+  size_t size;
```
vmg: Same here. This needs zero initialization.

vmg:
Everything else looks good. Thanks for adding this feature!

src/cc/bcc_perf_map.c
markdrayton: Thanks for the feedback!

Fixing `size` is fine. My `getline` man page actually says it's ignored if `*lineptr == NULL`, but http://man7.org/linux/man-pages/man3/getline.3.html says it must be zero. Regardless, there's clearly there's no downside to intializing to zero so I'll do that.

Re: `getline` allocating on every call: are you sure about that? http://man7.org/linux/man-pages/man3/getline.3.html says `getline` will use the buffer passed and `realloc` it if necessary. It doesn't say it'll allocate every time around and the example code in the man page just has a single `free` outside the loop. Additionally, my own simple test with valgrind appears to confirm this is safe:

https://gist.github.com/markdrayton/1ef00109ff2092898ab3b908d0598c41

vmg: Ah, you're totally right, I misread the manpage. It will indeed realloc if the original pointer is not `NULL`... That is a much more friendly API to use! So yes, the only frisky thing is initializing the value of `size`. Sorry for wasting your time!

PR #576, opened 2016-06-21
Fix nits for /tmp/perf-pid.map support

markdrayton:
Fixes a couple of nits that @vmg pointed out in #573 (see also #540).


vmg:
Looks good to me. Thank you!

Issue #503, opened 2016-04-23
move inet_ntoa and inet6_ntoa to bcc library


brendangregg:
Code is in tcpconnect.py.

brendangregg:
Closing; we'll use the socket library instead. See #582 and:

```Python
from socket import inet_ntop, AF_INET, AF_INET6 
[...]
inet_ntop(AF_INET, pack("I", event.saddr)
[...]
inet_ntop(AF_INET6, event.saddr)
```

PR #579, opened 2016-06-23
Do not throw exception on ipv6 packets in tcpretrans

bobrik:
Really not sure how come I get ipv6 retransmissions when ipv6 is disabled, but I don't like the exception either way 😕

### DIFF
--- a/tools/tcpretrans.py
+++ b/tools/tcpretrans.py
@@ -198,7 +198,7 @@ def print_ipv4_event(cpu, data, size):
         tcpstate[event.state]))
 def print_ipv6_event(cpu, data, size):
     event = ct.cast(data, ct.POINTER(Data_ipv6)).contents
-    print("%%-8s -6d %-2d %-20s %1s> %-20s %s" % (
+    print("%-8s %-6d %-2d %-20s %1s> %-20s %s" % (
         strftime("%H:%M:%S"), event.pid, event.ip,
         "...%x:%d" % (event.saddr, event.lport),
         type[event.type],


bobrik:
With python 3 or netaddr we can print ipv6 addresses nicely.

markdrayton:
Hah. I just fixed this too :) I also fixed up the v6 handling — at the moment we're only sending the last 4 bytes of the address from the kernel to userspace. I'll put my PR up in a few hours but if this one is committed before then I can just rebase on top.

brendangregg:
yes, see #503. This code should be moved into bcc, and tools updated to use it:

```Python
def inet_ntoa(addr):
    # u32 to dotted quad string
    dq = ''
    for i in range(0, 4):
        dq = dq + str(addr & 0xff)
        if (i != 3):
            dq = dq + '.'
        addr = addr >> 8
    return dq

def inet6_ntoa(addr):
    # see RFC4291 summary in RFC5952 section 2
    s = ":".join(["%x" % x for x in unpack_from(">HHHHHHHH",
        pack("QQ", addr & 0xffffffff, addr >> 64))])

    # compress left-most zero run only (change to most for RFC5952):
    s = re.sub(r'(^|:)0:(0:)+', r'::', s, 1)
    return s
```

If someone wants to do that (and/or improve the code!), please go ahead!

brendangregg:
I'll accept this PR for now and look out for the next one. Really looking forward to having these inet things in __init__.py. Can't remember why I didn't do it -- maybe I wanted to write a test suite first?

brendangregg:
Correction, I won't accept it as github is not letting me (branch is out of date)!...

bobrik:
@brendangregg I just rebased.

PR #582, opened 2016-06-26
IPv6 support for tcp* tools

markdrayton:
Currently these tools just grab the last 2 or 4 bytes of IPv6 addresses. This diff changes them to grab the whole address, as well as fixing a few other things here and there (timestamps were off by 10 and not floats, --lossprobe wasn't used).

I had to move things about in the v6 struct to make `bpf_probe_read(&data.6.daddr, sizeof(data6.daddr), &sk->...)` work correctly. Without this many of the fields were garbage so I assume this is an alignment issue (@4ast can this be caught?)

@brendangregg, in #503 you mentioned moving `inet_ntoa()` into the library. Was that because you couldn't get the versions in `socket` working, or was it so we could reuse the same code outside of Python? The `socket` versions here seem to work so shout out if I missed something.

Lastly, the output with IPv6 addresses is kinda grim as they overflow the v4 columns (e.g. see tools/tcpconnlat_example.txt). I'm not sure what to do about this. Making all the columns wide enough for the widest possible v6 addresses is going to make the output much wider than 80 chars. I considered changing the format based on whether the host even had a v6 address but that's quite a lot more complexity and IPv6 is rapidly becoming the norm.

### DIFF
--- a/tools/tcpaccept.py
+++ b/tools/tcpaccept.py
@@ -9,10 +9,6 @@
 # This uses dynamic tracing of the kernel inet_csk_accept() socket function
 # (from tcp_prot.accept), and will need to be modified to match kernel changes.
 #
-# IPv4 addresses are printed as dotted quads. For IPv6 addresses, the last four
-# bytes are printed after "..."; check for future versions with better IPv6
-# support.
-#
 # Copyright (c) 2015 Brendan Gregg.
 # Licensed under the Apache License, Version 2.0 (the "License")
 #
@@ -21,6 +17,8 @@
 
 from __future__ import print_function
 from bcc import BPF
+from socket import inet_ntop, AF_INET, AF_INET6
+from struct import pack
 import argparse
 import ctypes as ct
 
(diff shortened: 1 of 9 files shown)


tools/tcpconnlat.py
```diff
@@ -44,6 +42,7 @@
 bpf_text = """
 #include <uapi/linux/ptrace.h>
 #include <net/sock.h>
+#include <net/tcp_states.h>
```
markdrayton: Without this `tcpconnlat` wouldn't work on our pre-4.6 kernels. I double-checked it still worked under 4.6 with this.

brendangregg:
Yes, why not just use socket...

From memory, I'd searched for inet_ntoa() etc, found them in a python library, but then found that was not installed on our systems, so wrote our own implementations. But I just checked socket, and that seems to be everywhere. So I don't know what it was I'd found earlier!

Yes, I'd run into alignment issues earlier too. Thanks for the other fixes. The use of __int128 is much better .

LGTM!

tools/tcpaccept.py
```diff
@@ -52,21 +50,20 @@
...
     u64 lport;
     char task[TASK_COMM_LEN];
 };
 BPF_PERF_OUTPUT(ipv4_events);
 
 struct ipv6_data_t {
-    // XXX: update to transfer full ipv6 addrs
     u64 ts_us;
     u64 pid;
+    unsigned __int128 saddr;
```
4ast: nice. surprised this actually works. I assume the align of the whole struct is still 8. We probably need to add llvm test to make sure it doesn't regress.

PR #583, opened 2016-06-27
ext4slower: fix getting kallsyms address

ygrek:
It didn't work on `Linux 4.6.0-1-amd64 #1 SMP Debian 4.6.1-1 (2016-06-06) x86_64 GNU/Linux` out of the box, because of fourth field : 
```
$ grep ext4_file_operations /proc/kallsyms
ffffffffc0331340 r ext4_file_operations	[ext4]
```


brendangregg:
thanks; I'm guessing we'll need this for ext4dist.py, btrfsdist.py, and btrfsslower.py as well.

Issue #500, opened 2016-04-22
tool request: offcpudist


brendangregg:
The counterpart to cpudist ( #499 ), this tool would print histograms of off-CPU (blocked) time by process. 

Again, the existing code for runqlat should be a good starting point.

mcaleavya:
would you see this as using the add_wait_queue and remove_wait_queue functions?

brendangregg:
If I were using kprobes, I'd be doing it from the scheduler switch routine. Something like finish_task_switch. The add/remove_wait_queue functions would be interesting to trace, but for other reasons -- digging into queueing latency, length, etc. Here we just want total time off-CPU.

But now we have tracepoints, it would be great to use them. Eg, maybe sched:sched_switch.

Issue #567, opened 2016-06-13
tracepoint example


brendangregg:
I was looking for a simple example of instrumenting a tracepoint, and didn't find one, so wrote the following. Is this ok to submit for examples/tracing, or am I using the wrong interface? (I'm wondering if I've used the kprobe hack that predates tracepoint support. Is there a much better way nowadays?)

```Python
#!/usr/bin/python
#
# tracepoint  Example of instrumenting a kernel tracepoint.
#             For Linux, uses BCC, BPF. Embedded C.
#
# test by running this, then in another shell:
#   dd if=/dev/urandom of=/dev/null bs=1k count=10
#
# Copyright 2016 Netflix, Inc.
# Licensed under the Apache License, Version 2.0 (the "License")

from __future__ import print_function
from bcc import BPF, Tracepoint
import argparse

# arguments
parser = argparse.ArgumentParser(
    description="Tracepoint example: traces random:urandom_read")
parser.add_argument("-v", "--verbose", action="store_true",
    help="print BPF program")
args = parser.parse_args()

# 1. tracepoint argument stashing
tp = Tracepoint.enable_tracepoint("random", "urandom_read")
bpf_text = '#include <uapi/linux/ptrace.h>\n'
bpf_text += tp.generate_decl()
bpf_text += tp.generate_entry_probe()
bpf_text += tp.generate_struct()

# 2. BPF program
bpf_text += """
int printarg(struct pt_regs *ctx) {
    // Fetch struct of tracepoint args. The following references a map and
    // a struct defined by (1). Run with -v to see the full BPF program.
    u64 tid = bpf_get_current_pid_tgid();
    u64 *di = __trace_di.lookup(&tid);
    if (di == 0) { return 0; }
    struct urandom_read_trace_entry args = {};
    bpf_probe_read(&args, sizeof(args), (void *)*di);

    bpf_trace_printk("%d\\n", args.got_bits);
    return 0;
};
"""
if args.verbose:
    print(bpf_text)

# 3. load BPF program
b = BPF(text=bpf_text)
Tracepoint.attach(b)
b.attach_kprobe(event="perf_trace_urandom_read", fn_name="printarg")

# header
print("%-18s %-16s %-6s %s" % ("TIME(s)", "COMM", "PID", "GOTBITS"))

# format output
while 1:
    try:
        (task, pid, cpu, flags, ts, msg) = b.trace_fields()
    except ValueError:
        continue
[...]

PR #585, opened 2016-06-28
cpudist: Summarize task on-CPU time as histograms

### LINKED ISSUE #499: tool request: cpudist
Similar to cpudists from the DTraceToolkit, this should show histograms of on-CPU time by process, by tracing the scheduler from when a thread begins running to when it is descheduled.

Should be possible to use the runqlat code as a starting point.

Maybe this should be called "rundist" instead, and leave "cpudist" for a tool that does per-CPU histograms. I'm not sure. The downside there is the next ticket I'll file, which will be for offcpudist, for blocking time. So a pair of cpudist and offcpudist sounds consistent. Or maybe they should be rundist and blockdist?. Hm. cpudist and offcpudist sounds better I think.

goldshtn:
This summarizes task on-CPU time as histograms and displays it with a variety of switches, modeled on `runqlat`. Basic example:

```
# ./cpudist.py
^C
         2 -> 3          : 1        |                                        |
         4 -> 7          : 1        |                                        |
         8 -> 15         : 13       |**                                      |
        16 -> 31         : 187      |****************************************|
        32 -> 63         : 89       |*******************                     |
        64 -> 127        : 26       |*****                                   |
       256 -> 511        : 1        |                                        |
```

With this, we can close #499.

I don't know about what the relationship between #500 and #499 should be. @brendangregg I think you wanted these as separate tools for discoverability -- but the implementation can be 99% shared. Basically we'd have to reverse the histogram update location: update the histogram when the task gets *on* the CPU, instead of when it's getting *off*. Do you think it's worth the effort of duplicating the tool itself, its man page, and its examples file? Or can we maybe turn it into a switch, e.g. `-O`, which would reverse the meaning to off-CPU time instead of on-?

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ Examples:
 - tools/[btrfsdist](tools/btrfsdist.py): Summarize btrfs operation latency distribution as a histogram. [Examples](tools/btrfsdist_example.txt).
 - tools/[btrfsslower](tools/btrfsslower.py): Trace slow btrfs operations. [Examples](tools/btrfsslower_example.txt).
 - tools/[cachestat](tools/cachestat.py): Trace page cache hit/miss ratio. [Examples](tools/cachestat_example.txt).
+- tools/[cpudist](tools/cpudist.py): Summarize on-CPU time per task as a histogram. [Examples](tools/cpudist_example.txt)
 - tools/[dcsnoop](tools/dcsnoop.py): Trace directory entry cache (dcache) lookups. [Examples](tools/dcsnoop_example.txt).
 - tools/[dcstat](tools/dcstat.py): Directory entry cache (dcache) stats. [Examples](tools/dcstat_example.txt).
 - tools/[execsnoop](tools/execsnoop.py): Trace new processes via exec() syscalls. [Examples](tools/execsnoop_example.txt).
--- a/man/man8/cpudist.8
+++ b/man/man8/cpudist.8
@@ -1,0 +1,100 @@
+.TH cpudist 8  "2016-06-28" "USER COMMANDS"
+.SH NAME
+cpudist \- On-CPU task time as a histogram.
+.SH SYNOPSIS
+.B cpudist [\-h] [\-T] [\-m] [\-P] [\-L] [\-p PID] [interval] [count]
+.SH DESCRIPTION
+This measures the time a task spends on the CPU before being descheduled, and
+shows the times as a histogram. Tasks that spend a very short time on the CPU
+can be indicative of excessive context-switches and poor workload distribution,
+and possibly point to a shared source of contention that keeps tasks switching
+in and out as it becomes available (such as a mutex).
+
+This tool uses in-kernel eBPF maps for storing timestamps and the histogram,
+for efficiency. Despite this, the overhead of this tool may become significant
+for some workloads: see the OVERHEAD section.
+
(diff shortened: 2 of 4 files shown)


goldshtn:
Oh, I forgot to add that this is the first tool (I think) that was implemented using a kernel tracepoint instead of a kprobe. I think it works just fine. And of course, when we get proper tracepoint support in BCC, we could update this tool to require Linux 4.7 and use BPF tracepoint support instead of the hack we have right now.

4ast:
lgtm, though would be great if you can add proper tracepoint support sooner than later :)

goldshtn:
@4ast I'd be delighted to, but I will need a walkthrough ... As far as I can tell, there is a new BPF program type for tracepoints now. What would be required on the bcc side to use that?

4ast:
on bcc side just needs to support new prog_type. fancy clang bits can stay as-is for now, though all auto probe_read insertion will be failing, since there is no 'struct pt_regs *ctx' anymore. that can be done later.

goldshtn:
Sure, but where do I get the tracepoint struct from? The user will provide a probe function that takes void* and cast it? Should the clang side generate the struct?

4ast:
eventually yes, clang rewriter can and probably should (to avoid mistakes) generate appropriate input struct, but for now manual struct is fine. Just as samples/bpf/offwaketime_kern.c does:
```
struct sched_switch_args {
        unsigned long long pad;
        char prev_comm[16];
        int prev_pid;
        int prev_prio;
        long long prev_state;
        char next_comm[16];
        int next_pid;
        int next_prio;
};
SEC("tracepoint/sched/sched_switch")
int oncpu(struct sched_switch_args *ctx)
{
        /* record previous thread sleep time */
        u32 pid = ctx->prev_pid;
```
no type casting necessary.

brendangregg:
Thanks Sasha, looks great! Good usage: I like the choice and use of -p, -P, -L, etc.

Yes, it's pretty close to offcpudist #500, so maybe that can be added as an option (maybe "-o"; I'd usually reserve that for output files, but these tools generally just do STDOUT). For discoverability, we could later on add a offcpudist symlink.

I tried coming up with a tracepoint example (#567), but it was hard going to process arguments without the new 4.7 stuff. This tool, cpudist, looks much better because you aren't reading args. So I think I'd include this tool as is, and move it to old & updated to 4.7 later on. But I'll still resist adding many more other tracepoint scripts until the 4.7 stuff.

LGTM.

goldshtn:
@brendangregg The usage is all yours -- it's pretty much a verbatim copy from `runqlat` :-)

Do you prefer -o or -O? I thought the capital O would be more appropriate, no mixup with -o for output, and capitalized because it's a significant behavior change. But it's very subjective obviously. 

I'll try to add bcc tracepoint support in the next few days. Need to set up latest 4.7, had some problems building bcc on Fedora with 4.7rc5 lately.

brendangregg:
We can go with -O. I guess at some point we might have a tool that uses -o. (eg, imagine a tool that traced network packets and wrote a tcpdump file!).

I think -O, and proper tracepoints, can be added later. This works and is useful on older kernels.

Another later addition could be to add execnames to the -P output, which could be read from /proc during output (man page can say these are best effort, and may be inaccurate).

brendangregg:
I think this is ok to merge, as there'll be follow-ons. // @4ast

PR #586, opened 2016-06-29
cpudist: Support off-cpu time reports

### LINKED ISSUE #500: tool request: offcpudist

goldshtn:
Add -O switch, which directs cpudist to collect off-CPU time
statistics. Also restructure the code slightly and added examples
as appropriate.

Can now close #500. I don't know if we should add a symlink for
`offcpudist` or a simple script that forwards arguments to `cpudist`
and adds `-O`. // @brendangregg

### DIFF
--- a/tools/cpudist.py
+++ b/tools/cpudist.py
@@ -69,22 +69,16 @@ BPF_HASH(start, u32, u64);
 BPF_HASH(tgid_for_pid, u32, u32);
 STORAGE
 
-#define INVALID_TGID 0xffffffff
-
-static inline u32 get_tgid_if_missing(u32 tgid, u32 pid)
+static inline u32 get_tgid(u32 pid)
 {
-    if (tgid == INVALID_TGID) {
-        u32 *stored_tgid = tgid_for_pid.lookup(&pid);
-        if (stored_tgid != 0)
-            return *stored_tgid;
-    }
-    return tgid;
+    u32 *stored_tgid = tgid_for_pid.lookup(&pid);
+    if (stored_tgid != 0)
+        return *stored_tgid;
+    return 0xffffffff;
 }
 
 static inline void store_start(u32 tgid, u32 pid, u64 ts)
 {
-    tgid = get_tgid_if_missing(tgid, pid);
-
     if (FILTER)
         return;
 
@@ -93,8 +87,6 @@ static inline void store_start(u32 tgid, u32 pid, u64 ts)
 
 static inline void update_hist(u32 tgid, u32 pid, u64 ts)
 {
-    tgid = get_tgid_if_missing(tgid, pid);
-
     if (FILTER)
         return;
 
@@ -123,17 +115,24 @@ int sched_switch(struct pt_regs *ctx)
 
     struct sched_switch_trace_entry args = {};
     bpf_probe_read(&args, sizeof(args), (void *)*di);
-    // TODO: Store the comm as well
 
-    if (args.prev_state == TASK_RUNNING) {
-        u32 prev_pid = args.prev_pid;
 #ifdef ONCPU
-        update_hist(INVALID_TGID, prev_pid, ts);
+    if (args.prev_state == TASK_RUNNING) {
 #else
-        store_start(INVALID_TGID, prev_pid, ts);
+    if (1) {
+#endif
+        u32 prev_pid = args.prev_pid;
+        u32 prev_tgid = get_tgid(prev_pid);
+        if (prev_tgid == 0xffffffff)
+            goto BAIL;
+#ifdef ONCPU
+        update_hist(prev_tgid, prev_pid, ts);
+#else
+        store_start(prev_tgid, prev_pid, ts);
 #endif
     }
 
+BAIL:
 #ifdef ONCPU
     store_start(tgid, pid, ts);
 #else
--- a/tools/cpudist.py
+++ b/tools/cpudist.py
@@ -48,12 +48,9 @@ args = parser.parse_args()
 countdown = int(args.count)
 debug = 0
 
-tp = Tracepoint.enable_tracepoint("sched", "sched_switch")
-bpf_text = "#include <uapi/linux/ptrace.h>\n"
-bpf_text += "#include <linux/sched.h>\n"
-bpf_text += tp.generate_decl()
-bpf_text += tp.generate_entry_probe()
-bpf_text += tp.generate_struct()
+bpf_text = """#include <uapi/linux/ptrace.h>
+#include <linux/sched.h>
+"""
 
 if not args.offcpu:
     bpf_text += "#define ONCPU\n"
@@ -66,16 +63,7 @@ typedef struct pid_key {
 
 
 BPF_HASH(start, u32, u64);
-BPF_HASH(tgid_for_pid, u32, u32);
 STORAGE
-
-static inline u32 get_tgid(u32 pid)
-{
-    u32 *stored_tgid = tgid_for_pid.lookup(&pid);
-    if (stored_tgid != 0)
-        return *stored_tgid;
-    return 0xffffffff;
-}
 
 static inline void store_start(u32 tgid, u32 pid, u64 ts)
 {
@@ -99,32 +87,19 @@ static inline void update_hist(u32 tgid, u32 pid, u64 ts)
     STORE
 }
 
-int sched_switch(struct pt_regs *ctx)
+int sched_switch(struct pt_regs *ctx, struct task_struct *prev)
 {
     u64 ts = bpf_ktime_get_ns();
     u64 pid_tgid = bpf_get_current_pid_tgid();
     u32 tgid = pid_tgid >> 32, pid = pid_tgid;
-    // Keep a mapping of tgid for pid because when sched_switch hits,
-    // we only have the tgid information for the *current* pid, but not
-    // for the previous one.
-    tgid_for_pid.update(&pid, &tgid);
-
-    u64 *di = __trace_di.lookup(&pid_tgid);
-    if (di == 0)
-        return 0;
-
-    struct sched_switch_trace_entry args = {};
-    bpf_probe_read(&args, sizeof(args), (void *)*di);
 
 #ifdef ONCPU
-    if (args.prev_state == TASK_RUNNING) {
+    if (prev->state == TASK_RUNNING) {
 #else
     if (1) {
 #endif
-        u32 prev_pid = args.prev_pid;
-        u32 prev_tgid = get_tgid(prev_pid);
-        if (prev_tgid == 0xffffffff)
-            goto BAIL;
+        u32 prev_pid = prev->pid;
+        u32 prev_tgid = prev->tgid;
 #ifdef ONCPU
         update_hist(prev_tgid, prev_pid, ts);
 #else
@@ -173,8 +148,7 @@ if debug:
     print(bpf_text)
 
 b = BPF(text=bpf_text)
-Tracepoint.attach(b)
-b.attach_kprobe(event="perf_trace_sched_switch", fn_name="sched_switch")
+b.attach_kprobe(event="finish_task_switch", fn_name="sched_switch")
 
 print("Tracing %s-CPU time... Hit Ctrl-C to end." %
       ("off" if args.offcpu else "on"))
--- a/tools/cpudist.py
+++ b/tools/cpudist.py
@@ -82,6 +82,11 @@ static inline void update_hist(u32 tgid, u32 pid, u64 ts)
     if (tsp == 0)
         return;
 
+    if (ts < *tsp) {
+        // Probably a clock issue where the recorded on-CPU event had a
+        // timestamp later than the recorded off-CPU event, or vice versa.
+        return;
+    }
     u64 delta = ts - *tsp;
     FACTOR
     STORE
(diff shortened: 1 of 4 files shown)


goldshtn:
@4ast @drzaeus77 I think the Ubuntu build bot doesn't like me this morning. Can you re-trigger the build please?

brendangregg:
LGTM. The comment about storing the execname: when someone adds this feature later, I'd probably do that post-process from /proc rather than in BPF. Sched events can be very frequent (>1M/sec), so I wouldn't make the BPF program more expensive than really needed. That's just my intuition: I haven't perf tested it.

brendangregg:
I tested it (Linux 4.7-rc2), and it's not catching all events. Eg, use -p on the PID of another bash shell, and type in a few keystrokes. It should be measuring inter-keystroke latency as bash blocks on reads, but it isn't showing up.

```
# ./cpudist.py -O -p 12769
Tracing off-CPU time... Hit Ctrl-C to end.
^C
# /apps/perf-tools/bin/tpoint -p 12769 sched:sched_switch
Tracing sched:sched_switch. Ctrl-C to end.
           <...>-12769 [003] d... 22409373.184270: sched_switch: prev_comm=bash prev_pid=12769 prev_prio=120 prev_state=S ==> next_comm=swapper/3 next_pid=0 next_prio=120
           <...>-12769 [003] d... 22409373.391972: sched_switch: prev_comm=bash prev_pid=12769 prev_prio=120 prev_state=S ==> next_comm=swapper/3 next_pid=0 next_prio=120
           <...>-12769 [003] d... 22409373.624062: sched_switch: prev_comm=bash prev_pid=12769 prev_prio=120 prev_state=S ==> next_comm=swapper/3 next_pid=0 next_prio=120
           <...>-12769 [003] d... 22409373.920014: sched_switch: prev_comm=bash prev_pid=12769 prev_prio=120 prev_state=S ==> next_comm=swapper/3 next_pid=0 next_prio=120
           <...>-12769 [003] d... 22409374.071857: sched_switch: prev_comm=bash prev_pid=12769 prev_prio=120 prev_state=S ==> next_comm=swapper/3 next_pid=0 next_prio=120
^C
Ending tracing...
```

It's also not showing up with -P:

```
# ./cpudist.py -O -P
Tracing off-CPU time... Hit Ctrl-C to end.
^C

pid = 0
         0 -> 1          : 64       |****************************************|
         2 -> 3          : 10       |******                                  |
         4 -> 7          : 27       |****************                        |
         8 -> 15         : 12       |*******                                 |
        16 -> 31         : 3        |*                                       |
        32 -> 63         : 6        |***                                     |
        64 -> 127        : 10       |******                                  |
       128 -> 255        : 7        |****                                    |
       256 -> 511        : 4        |**                                      |
       512 -> 1023       : 4        |**                                      |
      1024 -> 2047       : 1        |                                        |
      2048 -> 4095       : 29       |******************                      |
      4096 -> 8191       : 3        |*                                       |
      8192 -> 16383      : 1        |                                        |
     16384 -> 32767      : 2        |*                                       |
     32768 -> 65535      : 1        |                                        |
     65536 -> 131071     : 13       |********                                |
    131072 -> 262143     : 11       |******                                  |
    262144 -> 524287     : 3        |*                                       |
```

Looking at the BPF logic, maybe it's filtering more than it should. The C program (debug=1) when doing -O -p PID is:

```C
#include <linux/sched.h>

BPF_HASH(__trace_di, u64, u64);

int __trace_entry_update(struct pt_regs *ctx)
{
        u64 tid = bpf_get_current_pid_tgid();
        u64 val = PT_REGS_PARM1(ctx);
        __trace_di.update(&tid, &val);
        return 0;
}

struct sched_switch_trace_entry {
        u64 __do_not_use__;
        char prev_comm[16];
        pid_t prev_pid;
        int prev_prio;
        long prev_state;
        char next_comm[16];
        pid_t next_pid;
        int next_prio;

};
                
typedef struct pid_key {
    u64 id;
    u64 slot;
} pid_key_t;


BPF_HASH(start, u32, u64);
BPF_HASH(tgid_for_pid, u32, u32);
BPF_HISTOGRAM(dist);

#define INVALID_TGID 0xffffffff

static inline u32 get_tgid_if_missing(u32 tgid, u32 pid)
{
    if (tgid == INVALID_TGID) {
        u32 *stored_tgid = tgid_for_pid.lookup(&pid);
        if (stored_tgid != 0)
            return *stored_tgid;
    }
    return tgid;
}

static inline void store_start(u32 tgid, u32 pid, u64 ts)
{
    tgid = get_tgid_if_missing(tgid, pid);

    if (tgid != 12769)
        return;

    start.update(&pid, &ts);
}

static inline void update_hist(u32 tgid, u32 pid, u64 ts)
{
    tgid = get_tgid_if_missing(tgid, pid);

    if (tgid != 12769)
        return;

    u64 *tsp = start.lookup(&pid);
    if (tsp == 0)
        return;

    u64 delta = ts - *tsp;
    delta /= 1000;
    dist.increment(bpf_log2l(delta));
}

int sched_switch(struct pt_regs *ctx)
{
    u64 ts = bpf_ktime_get_ns();
    u64 pid_tgid = bpf_get_current_pid_tgid();
    u32 tgid = pid_tgid >> 32, pid = pid_tgid;
    // Keep a mapping of tgid for pid because when sched_switch hits,
    // we only have the tgid information for the *current* pid, but not
    // for the previous one.
    tgid_for_pid.update(&pid, &tgid);

    u64 *di = __trace_di.lookup(&pid_tgid);
    if (di == 0)
        return 0;

    struct sched_switch_trace_entry args = {};
    bpf_probe_read(&args, sizeof(args), (void *)*di);
    // TODO: Store the comm as well

    if (args.prev_state == TASK_RUNNING) {
        u32 prev_pid = args.prev_pid;
#ifdef ONCPU
        update_hist(INVALID_TGID, prev_pid, ts);
#else
        store_start(INVALID_TGID, prev_pid, ts);
#endif
    }

#ifdef ONCPU
    store_start(tgid, pid, ts);
#else
    update_hist(tgid, pid, ts);
#endif

    return 0;
}
```

PS. tpoint is my ftrace tool; I went to use trace.py but hit:

```
# ./trace.py -p 12769 t:sched:sched_switch
global name 'pid' is not defined
```

goldshtn:
Thanks, that's a bug. The test for TASK_RUNNING makes sense for on-CPU (we want to trace when a running task gets descheduled) but not for off-CPU (where the task could wake up even if the previous task wasn't running, which is typical on an idle system). I'll fix it later today. 

Re trace, I'll take a look -- must have broken the -p flag at some point. This is a good time to think about writing at least some basic smoke tests for the tools, not just the bcc core...

goldshtn:
@brendangregg TASK_RUNNING test fixed in this PR, the trace issue was fixed in #589. Thanks again for testing this.

tools/cpudist.py
```diff
@@ -52,60 +55,89 @@
...
-        if (stored_tgid == 0)
-            goto BAIL;
-        tgid = *stored_tgid;
-
-        if (FILTER)
+#else
+    if (1) {
+#endif
+        u32 prev_pid = args.prev_pid;
+        u32 prev_tgid = get_tgid(prev_pid);
```
4ast: extra update+lookup for task switch... that is quite heavy.
I was assuming that trace_di.lookup will go away with proper tracepoint support, but these two will stay... hmm
I think it's probably better to use kprobe on sched_switch instead then we can fetch pid and tgid from task_struct pointer via probe_read. That will be much lower overhead.

goldshtn: There's the tradeoff between using a tracepoint, which is less likely to break, and the perf hit in this case because the tracepoint args don't have the tgid. I guess the overhead consideration is more important here. So should I rewrite this with a kprobe on sched_switch?

4ast: kprobe on finish_task_switch() can get pid/tgid from prev pointer and current_pid_tgid via helper. If that's the only thing we need then the speed wins... but accessing prev_state accurately is problematic, since 'preempt' flag of schedule() is lost.

goldshtn: The reason I'm using prev_state here is only because I want to track the current pid going off-CPU only if it was previously running, and not if the system was idle. 

So the question really is, what do we know about situations when prev_state is not running? What's the meaning of prev_pid in that case? Is it just the pid that happened to be there before we went idle?

4ast: actually comparing to TASK_RUNNING is probably doing not what you expected. Take a look at __trace_sched_switch_state(). It will set tracepoint's pref_state to TASK_RUNNING | TASK_STATE_MAX and our 'if' will miss it. Looking at prev->state is cleaner.

goldshtn: Yep, I see now that it's not always TASK_RUNNING. Still, not sure what you mean by "looking at prev->state". Do you mean placing a kprobe on `finish_task_switch` and looking at `prev->state` there?

4ast: > Do you mean placing a kprobe on finish_task_switch and looking at prev->state there?

yes

goldshtn: Done in 06d90d3.

brendangregg:
Thanks, almost perfect!

```
# ./cpudist.py.1 -P
^C

pid = 0
               usecs                         : count     distribution
                   0 -> 1                    : 18       |******              |
                   2 -> 3                    : 24       |********            |
                   4 -> 7                    : 44       |****************    |
                   8 -> 15                   : 37       |*************       |
                  16 -> 31                   : 5        |*                   |
                  32 -> 63                   : 17       |******              |
                  64 -> 127                  : 20       |*******             |
                 128 -> 255                  : 10       |***                 |
                 256 -> 511                  : 9        |***                 |
                 512 -> 1023                 : 10       |***                 |
                1024 -> 2047                 : 2        |                    |
                2048 -> 4095                 : 50       |******************  |
                4096 -> 8191                 : 5        |*                   |
                8192 -> 16383                : 13       |****                |
               16384 -> 32767                : 22       |********            |
               32768 -> 65535                : 55       |********************|
               65536 -> 131071               : 0        |                    |
              131072 -> 262143               : 0        |                    |
              262144 -> 524287               : 0        |                    |
              524288 -> 1048575              : 0        |                    |
             1048576 -> 2097151              : 0        |                    |
             2097152 -> 4194303              : 0        |                    |
             4194304 -> 8388607              : 0        |                    |
             8388608 -> 16777215             : 0        |                    |
            16777216 -> 33554431             : 0        |                    |
            33554432 -> 67108863             : 0        |                    |
            67108864 -> 134217727            : 0        |                    |
           134217728 -> 268435455            : 0        |                    |
           268435456 -> 536870911            : 0        |                    |
           536870912 -> 1073741823           : 0        |                    |
          1073741824 -> 2147483647           : 0        |                    |
          2147483648 -> 4294967295           : 0        |                    |
          4294967296 -> 8589934591           : 0        |                    |
          8589934592 -> 17179869183          : 0        |                    |
         17179869184 -> 34359738367          : 0        |                    |
         34359738368 -> 68719476735          : 0        |                    |
         68719476736 -> 137438953471         : 0        |                    |
        137438953472 -> 274877906943         : 0        |                    |
        274877906944 -> 549755813887         : 0        |                    |
        549755813888 -> 1099511627775        : 0        |                    |
       1099511627776 -> 2199023255551        : 0        |                    |
       2199023255552 -> 4398046511103        : 0        |                    |
       4398046511104 -> 8796093022207        : 0        |                    |
       8796093022208 -> 17592186044415       : 0        |                    |
      17592186044416 -> 35184372088831       : 0        |                    |
      35184372088832 -> 70368744177663       : 0        |                    |
      70368744177664 -> 140737488355327      : 0        |                    |
     140737488355328 -> 281474976710655      : 0        |                    |
     281474976710656 -> 562949953421311      : 0        |                    |
     562949953421312 -> 1125899906842623     : 0        |                    |
    1125899906842624 -> 2251799813685247     : 0        |                    |
    2251799813685248 -> 4503599627370495     : 0        |                    |
    4503599627370496 -> 9007199254740991     : 0        |                    |
    9007199254740992 -> 18014398509481983    : 0        |                    |
   18014398509481984 -> 36028797018963967    : 1        |                    |

pid = 20737

pid = 20738

pid = 20739

pid = 20740
        64 -> 127        : 1        |****************************************|
[...]
```

I'm not sure how we got the "18014398509481984 -> 36028797018963967" by looking at the BPF program. Maybe a map read is returning -1 in Python, and Python isn't handling it? Just a guess.

goldshtn:
@brendangregg Not sure how this can happen, but I managed to reproduce it too after a few stress runs. It is a result of the current timestamp being SMALLER than the previous timestamp. One theory I have is that it could be the event firing on two different CPUs -- on one CPU, you'd get the thread switching out, and on another CPU the thread is switching in. If the clock isn't guaranteed monotonic between the two CPUs and the time interval is very small, we could observe a negative delta.

In any case, I added an explicit check that discards events in that case (see bee8d36).

brendangregg:
Ah, right, I've set clocksource to tsc on this Xen system for performance, but I don't think tsc can be trusted in this context.

Ok, tested again, and now works. Thanks!

brendangregg:
LGTM. I'd merge it but the "out-of-date with base branch" thing is stopping me. :)

PR #590, opened 2016-06-30
bcc: Tracepoint support in libbpf and BPF

### LINKED ISSUE #587: Proper tracepoint support in BCC
(Discussion from #585 about tracepoint support should move here if necessary.)

Kernel 4.7 will have native BPF support for tracepoints, using a new BPF program type. BCC should support this as well. Here are the pieces I think are necessary -- @brendangregg @4ast @drzaeus77 please let me know if I'm missing anything.

- [x] Add `attach_tracepoint` to the BPF module, which would take the tracepoint name (probably as two components, e.g. "sched" and "sched_switch") and a function name from the BPF program
- [x] Add `attach_tracepoint` to libbbc, which is very similar to the other attach functions, but it uses the proper BPF program type and the proper debugfs configuration to enable the probe
- [x] Add special treatment to the rewriter so that when a tracepoint is attached, the tracepoint arguments structure is automatically generated with some agreed-upon name that the BPF program can reference (see below)

This is the ideal API, C side:

```c
int probe_sched_switch(struct TP__sched__sched_switch__args *args) {
  bpf_trace_printk("switching from %u to %u\n", args->prev_pid, args->next_pid);
  return 0;
}
```

Python side:

```python
b = BPF(...)
b.attach_tracepoint(category="sched", event="sched_switch", fn_name="probe_sched_switch")
```

goldshtn:
Introduce tracepoint support in libbpf via new `bpf_attach_tracepoint`
API, which takes the tracepoint category and name (e.g. "sched",
"sched_switch"). Attach the tracing program to the tracepoint's id
and proceed as usual.

Add `attach_tracepoint` API to Python BPF module, which takes the
tracepoint description as a single string (e.g. "sched:sched_switch").
Load the BPF program with bpf_prog_type set to TRACEPOINT and then
call `bpf_attach_tracepoint` to attach it.

Simple test:

```python
#!/usr/bin/env python

from bcc import BPF
from time import sleep

bpf = BPF(text="""
#include <linux/ptrace.h>

struct tp_args {
    unsigned long long __do_not_use__;
    int irq;
};

BPF_HASH(count, u32, u32);

int probe(struct tp_args *args) {
    u32 key = args->irq;
    u32 val = 0;
    u32 *existing_val = count.lookup_or_init(&key, &val);
    (*existing_val)++;
    return 0;
}
""")

bpf.attach_tracepoint("irq:irq_handler_entry", "probe")

print("Counting interrupts, Ctrl-C to quit...")
try:
    sleep(999999)
except KeyboardInterrupt:
    pass

print("")
counts = bpf["count"]
print("%-8s %-8s" % ("IRQ", "COUNT"))
for k, v in counts.items():
    print("%-8d %-8d" % (k.value, v.value))
```

Test output:

```
# ./test_bcc_tp.py
Counting interrupts, Ctrl-C to quit...
^C
IRQ      COUNT
24       2
27       4
```

Partially closes #587 -- no rewriter support for generating the tracepoint
structure yet, it needs to be provided manually.

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -342,6 +342,34 @@ int bpf_detach_uprobe(const char *event_desc) {
   return bpf_detach_probe(event_desc, "uprobe");
 }
 
+void * bpf_attach_tracepoint(int progfd, const char *tp_category,
+                             const char *tp_name, int pid, int cpu,
+                             int group_fd, perf_reader_cb cb, void *cb_cookie) {
+  char buf[256];
+  struct perf_reader *reader = NULL;
+
+  reader = perf_reader_new(cb, NULL, cb_cookie);
+  if (!reader)
+    goto error;
+
+  snprintf(buf, sizeof(buf), "/sys/kernel/debug/tracing/events/%s/%s",
+           tp_category, tp_name);
+  if (bpf_attach_tracing_event(progfd, buf, reader, pid, cpu, group_fd) < 0)
+    goto error;
+
+  return reader;
+
+error:
+  perf_reader_free(reader);
+  return NULL;
+}
+
+int bpf_detach_tracepoint(const char *tp_category, const char *tp_name) {
+  // Right now, there is nothing to do, but it's a good idea to encourage
+  // callers to detach anything they attach.
+  return 0;
+}
+
 void * bpf_open_perf_buffer(perf_reader_raw_cb raw_cb, void *cb_cookie, int pid, int cpu) {
   int pfd;
   struct perf_event_attr attr = {};
--- a/src/libbpf.h
+++ b/src/libbpf.h
@@ -54,6 +54,11 @@ void * bpf_attach_uprobe(int progfd, const char *event, const char *event_desc,
                          void *cb_cookie);
 int bpf_detach_uprobe(const char *event_desc);
 
+void * bpf_attach_tracepoint(int progfd, const char *tp_category,
+                             const char *tp_name, int pid, int cpu,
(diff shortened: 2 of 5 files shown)


4ast:
nice! lgtm
@drzaeus77 pls take a look

brendangregg:
I like it. I don't mind adding manual tracepoint structs for now.

I assume USDT tracepoints are a different beast?

drzaeus77:
LGTM...except please add a test case for the awesome new feature.

goldshtn:
@brendangregg: There's no BPF support for USDT tracepoints, so I don't have anything new to add there. But what we have right now, with uprobes, is actually quite decent.

@drzaeus77: Thanks! Added a test that exercises `attach_tracepoint`.

goldshtn:
Ugh, I picked a bad tracepoint. Don't know why I assumed kmem:kmalloc would necessarily be hit. Will use something more reliable...

goldshtn:
Hmpf. @drzaeus77 The test is failing with `bpf: Invalid argument`. Do the build bots have 4.7 already?

drzaeus77:
Nope, mostly 4.4. There should be a way to mask the error if the feature is unsupported, just in the test itself. As in, in the contents of class MyTest, when the particular error that happens when the feature doesn't exist, raise unittest.SkipTest instead, otherwise let the other types of errors propagate normally.

goldshtn:
I'm followed the example of test_stackid.py and asserted kernel version >= 4.7. It's a bit difficult to determine at runtime whether the test failed because the feature is unavailable -- all I'd get is an error from bpf_load_program.

PR #594, opened 2016-07-01
ProcSyms: fix off-by-ones, use binary search to resolve addresses

markdrayton:
This PR:

a) fixes a couple of off-by-one bugs in matching modules and symbols. Specifically, when given two adjacent symbols resolving the start address of the second symbol would incorrectly return the first symbol. This diff fixes the ProcSyms::Module code, removes the workaround in bcc_perf_map_foreach_sym(), and adds a test.

b) switches to a binary search in ProcSyms::Module::find_addr(). The current linear search isn't very good on large mappings (some of our JIT maps are hundreds of thousands of entries).

### DIFF
--- a/src/cc/bcc_perf_map.c
+++ b/src/cc/bcc_perf_map.c
@@ -93,7 +93,7 @@ int bcc_perf_map_foreach_sym(const char *path, bcc_perf_map_symcb callback,
     if (newline)
         newline[0] = '\0';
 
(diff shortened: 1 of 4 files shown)


vmg:
Ah yes, `std::upper_bound` seems like a good fit here. The original Python version used the naive loop, but this is most definitely a worthy optimization. 👌

PR #598, opened 2016-07-05
ProcSyms: deduplicate symbol names

markdrayton:
`/tmp/perf-pid.map` files can contain the same function name on many lines. Often this is because the map is written by a JIT which compiles parts of a function as it needs them, writing each to a different memory location. Currently `ProcSyms` just reads them all in and creates a `Symbol` for every line, regardless of whether the name is duplicate. This can easily add up to a lot of RAM. I saw cases in which a large, heavily duplicated map and this simple example would use nearly 400MB RAM:

    struct bcc_symbol sym
    void *resolver = bcc_symcache_new(pid);
    bcc_symcache_resolve(resolver, 0xf00, &sym);

This is enough to make tools which themselves have a negative impact on the system they're observing.

This PR deduplicates the symbol names by putting names an `unordered_set` and storing pointers to the values instead of string copies. This is safe — references aren't invalidated during insertion (http://en.cppreference.com/w/cpp/container/unordered_set/emplace) — but doesn't feel very clean. Suggestions for improvement are welcome.

The same test program is now 130MB RSS. I'll try to work out if there is more to save.

### DIFF
--- a/src/cc/bcc_syms.cc
+++ b/src/cc/bcc_syms.cc
@@ -132,7 +132,8 @@ bool ProcSyms::resolve_name(const char *module, const char *name,
 int ProcSyms::Module::_add_symbol(const char *symname, uint64_t start,
                                   uint64_t end, int flags, void *p) {
   Module *m = static_cast<Module *>(p);
-  m->syms_.emplace_back(symname, start, end, flags);
+  auto res = m->symnames_.emplace(symname);
+  m->syms_.emplace_back(&*(res.first), start, end, flags);
   return 0;
 }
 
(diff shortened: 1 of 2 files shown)


markdrayton:
cc @vmg (forgot this earlier)

4ast:
a bit hack-ish, but I couldn't think of anything better.
Was a bit concerned with how refresh()/delete logic will deal with these cross pointers, but it looks fine.
lgtm.

4ast:
pending for too long without any other feedback. merging.

markdrayton:
Thanks.

> a bit hack-ish, but I couldn't think of anything better.

Yeah, I agree. Over time I hope we can come up with something nicer.

PR #603, opened 2016-07-08
Fix test failure in test_libbcc

drzaeus77:
On some systems, was seeing a failure at tests/cc/test_c_api.cc:172 due
to failure to open the /tmp/perf-pid.map file. Looking through the code,
narrowed it down to an invalid use of c_str() on a temporary
std::string. Fix it by storing the string in a variable.


4ast:
good catch. Thanks!
@markdrayton

Issue #605, opened 2016-07-09
Python module unregisters probes only at-exit, not when the BPF object is destroyed


goldshtn:
There's a somewhat un-intuitive behavior in the Python bcc module. Namely, probes are added to a `BPF` object instance -- either automatically through `__init__` or manually via methods such as `attach_kprobe` etc. However, probes are automatically deleted only at program exit -- in the `cleanup_probes` global function, which is registered with `@atexit.register` to run on shutdown.

This makes for brittle code, and in fact we have 3 tests in test_histogram.py, which rely on this behavior unintentionally. The first test attaches a probe, doesn't detach it, and the subsequent tests work because the probe is still attached -- even though the wrong probe function is being invoked. (Discovered as I broke the build in #602.)

I propose that in addition to the `@atexit.register` cleanup function, we also add an explicit `close` method to the `BPF` object that will unregister probes that were added through that `BPF` object. We should encourage probe cleanup at the `BPF` object level.

/cc @drzaeus77

4ast:
makes sense to me

markdrayton:
> I propose that in addition to the @atexit.register cleanup function, we also add an explicit close method to the BPF object that will unregister probes that were added through that BPF object.

Do we need anything fancier than calling `cleanup_kprobes` from the `BPF.close` method of a `BPF` object here? As far as I can see this'll only unregister probes that were registered by that object.

goldshtn:
I might be missing something trivial, but why would it only unregister probes that were registered by that object?

goldshtn:
Oh and if we could call it "cleanup" or something while we're at it. It's not cleaning only kprobes for a while now.

markdrayton:
Oh, doh, you're right — I forgot that `open_kprobes` and friends are global to the module, not a `BPF` instance.

> Oh and if we could call it "cleanup" or something while we're at it. It's not cleaning only kprobes for a while now.

Yeah, I spotted that one too.

markdrayton:
I think my earlier comments were confusing. By

> As far as I can see this'll only unregister probes that were registered by that object.

I meant that calling the global `cleanup_kprobes` from a `cleanup` method on `BPF` would only remove probes that were added by the instantiated `BPF` object. In other words, no probes/tracepoints that existed before the `BPF` module was instantiated would be touched. It'd just do what the current cleanup does but on demand.

[...]

PR #614, opened 2016-07-13
Fix probe detaching and auto-kprobes

markdrayton:
619fc14 (part of #602) added support for auto-tracepoint probes along the lines
of auto-kprobes. Unfortunately, this seems to have broken automatic kprobes
except when the automatic probe is first to be added (i.e., when `open_kprobes`
is empty). This diff fixes that and adds a test, though the test currently
breaks other tests because of the probe unregistering issues described in #605
(see https://gist.github.com/markdrayton/5be240a9043982a3d10f2a59d8ed7275).

A real-life test is `killsnoop` and `execsnoop`: both add more than one
auto-kprobe and stopped working after #602. They work now.

It's quite possible that I don't understand the intent of the change that broke this — @goldshtn, feel free to jump in and correct me if this isn't the right approach.

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -30,43 +30,19 @@ from .table import Table
 from .tracepoint import Perf, Tracepoint
 from .usyms import ProcessSymbols
 
-open_kprobes = {}
-open_uprobes = {}
-open_tracepoints = {}
-tracefile = None
-TRACEFS = "/sys/kernel/debug/tracing"
 _kprobe_limit = 1000
+_num_open_probes = 0
+
+# for tests
+def _get_num_open_probes():
+    global _num_open_probes
+    return _num_open_probes
+
+TRACEFS = "/sys/kernel/debug/tracing"
 
 DEBUG_LLVM_IR = 0x1
 DEBUG_BPF = 0x2
 DEBUG_PREPROCESSOR = 0x4
-
-@atexit.register
-def cleanup_kprobes():
-    for k, v in open_kprobes.items():
-        lib.perf_reader_free(v)
-        # non-string keys here include the perf_events reader
-        if isinstance(k, str):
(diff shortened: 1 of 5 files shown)


goldshtn:
Good catch. So, I think the whole check for no open kprobes is unnecessary, but I didn't want to break the histogram tests just yet. In the process, I botched the check -- instead of checking once that there are no kprobes outside of the loop, I put the test inside the loop, which means it only works for the first kprobe. 

I guess if you want to fix this for good, it would be best to address the probe unreg issues and fix the histogram tests. Or, for now, you could just move the check for no open kprobes out of the loop...

Sorry about the mess.

markdrayton:
> Sorry about the mess.

No problem :)

I'll have a go at fixing the detaching and tests.

markdrayton:
Okay, these two commits should clean up all the probe detaching. The USDT tests don't work locally for me (need to find the right Ruby packages) so I'll have to leave those for the buildbots. All other tests passed and I ran the bulk of the BCC tools to make sure they still work.

goldshtn:
Looks good. However, the probe quota logic is not global anymore, and I'm not sure that's desirable. Maybe keep the quota global?

markdrayton:
> However, the probe quota logic is not global anymore, and I'm not sure that's desirable. Maybe keep the quota global?

Yes, true. The latest patches address this by adding a module-level counter and helpers to do the work of keeping the global counter and object-local dicts up to date.

goldshtn:
@drzaeus77 @4ast I think this looks good. And it also seems that the build bots need to be retriggered?

4ast:
thanks for the extra test! LGTM

drzaeus77:
Will re-run them, one sec...

markdrayton:
Tests look good. @4ast, @drzaeus77, can this be merged?

4ast:
LGTM. thanks!

PR #618, opened 2016-07-18
Use errno symbols in offcputime.py

palmtenor:
The script uses hard-coded values to handle cases like stack trace size limit exceeded or requesting user stack on kernel thread.
Change it to use the errno symbols instead.

Tested by running: http://pastebin.com/44axT43A


4ast:
lgtm

brendangregg:
thanks!

PR #625, opened 2016-07-23
Fix tools still referencing ProcUtils

palmtenor:
Recent [change](https://github.com/iovisor/bcc/commit/4f88a9401357d7b75e917abd994aa6ea97dda4d3) removed `procstat.py`, but the `tools/argdist.py` and `tools/trace.py` are still referencing it. This Diff moves the only method (`ProcUtils.which()`) that is still being used to those tools.

Also, make `BPF.find_library()` not to call `decode()` on `None`.

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -116,6 +116,35 @@ class BPF(object):
                     raise Exception("Could not find file %s" % filename)
         return filename
 
+    @staticmethod
+    def _find_exe(cls, bin_path):
+        """
+        _find_exe(bin_path)
+
+        Traverses the PATH environment variable, looking for the first
+        directory that contains an executable file named bin_path, and
+        returns the full path to that file, or None if no such file
+        can be found. This is meant to replace invocations of the
+        "which" shell utility, which doesn't have portable semantics
+        for skipping aliases.
+        """
+        # Source: http://stackoverflow.com/a/377028
+        def is_exe(fpath):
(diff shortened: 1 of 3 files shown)


brendangregg:
thanks @palmtenor, that was quick! @goldshtn what do you think? ... I thought I'd fixed argdist, but I see it works for some arguments but not others.

```
# ./argdist.py -C 't:irq:irq_handler_entry():int:tp.irq'
global name 'Perf' is not defined
```

goldshtn:
@brendangregg: That's odd. You moved the `Perf` class to the bcc module, so that should work. Did you update the bcc module before running `argdist`? 

@palmtenor: This duplication of `find_exe` feels a bit hacky. Maybe put it in libbcc?

palmtenor:
Moved the helper function to a common place according to @goldshtn's suggestion:)

goldshtn:
Thanks, looks good (although I would like it even better if it was on the C++ side; but this can be done later). 

Really tiny nitpick -- could you rebase and squash the merge commits please? :-)

markdrayton:
Another nit: can you indent the code in __init__.py with 4 spaces so it matches the rest of the file? Thanks!

palmtenor:
Fixed indentation and commits:)

palmtenor:
Merge commits into one

4ast:
agree with @goldshtn that long term such things we need to have on c++ side, but it can be done later. Thanks!

PR #629, opened 2016-07-26
[tools] code refactoring between cachestat and cachetop

chantra:
Also move cachetop.py sleep after the displaying block so UI shows
something from the beginning and not afer the first `interval` seconds.

### DIFF
--- a/tools/cachestat.py
+++ b/tools/cachestat.py
@@ -17,38 +17,12 @@
 # 13-Jan-2016   Allan McAleavy  run pep8 against program
 
 from __future__ import print_function
-from bcc import BPF
 from time import sleep, strftime
-import signal
-import re
+from libs import cache, utils
 from sys import argv
 
-# signal handler
-def signal_ignore(signal, frame):
-    print()
-
-# Function to gather data from /proc/meminfo
-# return dictionary for quicker lookup of both values
-def get_meminfo():
-    result = dict()
-
-    for line in open('/proc/meminfo'):
-        k = line.split(':', 3)
-        v = k[1].split()
-        result[k[0]] = int(v[0])
-    return result
 
 # set global variables
-rtaccess = 0
-wtaccess = 0
-mpa = 0
-mbd = 0
-apcl = 0
-apd = 0
-access = 0
-misses = 0
-rhits = 0
-whits = 0
 debug = 0
 
 # args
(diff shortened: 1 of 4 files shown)


brendangregg:
I'm not sure if putting this in a library is a good or a bad thing. This is instrumenting 4 kprobes that can easily break (and I've seen reports that they don't work on some kernels, but that may be 3.x series). It's one thing to say a tool is broken, but another to say a bcc library is broken. Tests? Also, are there implications in having a tool refer to extra libraries for packaging and distribution under /usr/share/bcc/tools?

It also feels a bit like SystemTap's tapsets, which became massively complex as it tried to satisfy different kernel versions. We _could_ do such library support for tracepoints, which shouldn't change, but kprobes will.

Having extra functions in utils.py, again, I'm not sure if it's good or bad. IMHO, SystemTap became very difficult to use as the API grew, and grew, and grew, and trying to make sense of any example often meant learning yet-another way to do kind-of the same thing. I guess those two functions are ok, but we'd really want to resist adding too much.

chantra:
Well, yes I hear your concern and I guess it goes both ways. When something is getting fixed, it needs to be fixed in multiple places, while having common components in one place require 1 file to be fixed.

I don't have strong opinions and will happily through this branch away though.

> Also, are there implications in having a tool refer to extra libraries for packaging and distribution under /usr/share/bcc/tools?

I believe setup.py will need to be updated, that is about it. I did not test that bit though, just tested when ran locally.

brendangregg:
There's also the overhead factor (just to put these reasons in one place for future visitors). And this is a particularly bad example; from the man page:

```
.SH OVERHEAD
This traces various kernel page cache functions and maintains in-kernel counts, which
are asynchronously copied to user-space. While the rate of operations can
be very high (>1G/sec) we can have up to 34% overhead, this is still a relatively efficient way to trace 
these events, and so the overhead is expected to be small for normal workloads.
```

So it would appear to be a neat library function that returned cache statistics, but in reality would be a brittle (kprobe) high overhead function (up to 34%). Keeping the kprobes (or tracepoints) explicitly in the tool source makes it easier for new users to evaluate overhead, rather than implicitly in library functions. You can scan each tool to see what it instruments.

Plus I'm still worried that I got those kprobes wrong. Original post: http://www.brendangregg.com/blog/2014-12-31/linux-page-cache-hit-ratio.html

4ast:
agree with @brendangregg ... moving compute_cache_stats() into lib makes c/py split hard to follow with hidden dependency on layout of k and v in 'counts' map.

chantra:
ok, let me drop that then.

Issue #623, opened 2016-07-22
Fix string type handling in attach_kprobe()


markdrayton:
As discovered in #615, `attach_kprobe()` with `from __future__ import unicode_literals` makes an assertion failure:

```
# ./cachetop.py
Traceback (most recent call last):
  File "./cachetop.py", line 255, in <module>
    curses.wrapper(handle_loop, args)
  File "/usr/lib/python2.7/curses/wrapper.py", line 43, in wrapper
    return func(stdscr, *args, **kwds)
  File "./cachetop.py", line 177, in handle_loop
    b.attach_kprobe(event="add_to_page_cache_lru", fn_name="do_count")
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 362, in attach_kprobe
    assert isinstance(event, str), "event must be a string"
AssertionError: event must be a string
```

The assert exists because we later use the key type to determine how to unregister a kprobe:

https://github.com/iovisor/bcc/blob/f4bf27517d1fe59ed915d7f8cb3610d4a58583ae/src/python/bcc/__init__.py#L770

We can probably make this more robust by splitting the probe types or filtering in a way that matches all string types.

markdrayton:
closed via #632.

PR #632, opened 2016-07-27
Probe registration fixes

markdrayton:
The first diff fixes the py3 incompatibilities @drzaeus77 pointed out in https://github.com/iovisor/bcc/commit/8fa8c0e8a0daeb4ab9eb2cc5fb308f996351965a. In testing (which I did using Ubuntu's [alternatives](https://linuxconfig.org/how-to-change-from-default-to-alternative-python-version-on-debian-linux) support) I found a couple of other things that needed fixing too.

The second removes the asserts that force `str` names for kprobes (#623). I'm on the fence on this one. On one hand I think it's better to make sure all the probes have string keys, particularly as we use the type to decide which probes to remove, but on the other hand it makes the API unforgiving if you use py2's unicode_literals, or you try to register a probe with a unicode string you already had to hand. @drzaeus77 any opinion here?

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -334,11 +334,12 @@ class BPF(object):
         fn.sock = sock
 
     def _get_kprobe_functions(self, event_re):
-        blacklist = set([line.rstrip().split()[1] for line in
-                open("%s/../kprobes/blacklist" % TRACEFS)])
+        with open("%s/../kprobes/blacklist" % TRACEFS) as blacklist_file:
+            blacklist = set([line.rstrip().split()[1] for line in
+                    blacklist_file])
         fns = []
-        with open("%s/available_filter_functions" % TRACEFS) as f:
-            for line in f:
+        with open("%s/available_filter_functions" % TRACEFS) as avail_file:
+            for line in avail_file:
                 fn = line.rstrip().split()[0]
                 if re.match(event_re, fn) and fn not in blacklist:
                     fns.append(fn)
@@ -604,7 +605,7 @@ class BPF(object):
 
     def _trace_autoload(self):
         for i in range(0, lib.bpf_num_functions(self.module)):
-            func_name = lib.bpf_function_name(self.module, i)
+            func_name = str(lib.bpf_function_name(self.module, i).decode())
             if func_name.startswith("kprobe__"):
                 fn = self.load_func(func_name, BPF.KPROBE)
                 self.attach_kprobe(event=fn.name[8:], fn_name=fn.name)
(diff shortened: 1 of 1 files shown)


chantra:
>  but on the other hand it makes the API unforgiving if you use py2's unicode_literals, or you try to register a probe with a unicode string you already had to hand.

Eventually all string will be unicode, why not rather enforcing unicode. If you want the api to be forgiving, decorate the api to properly convert from str to unicode if needs be.

I seems safe to assume that the unicode strings will only be made of ascii char and will safely convert to the c binding variable using `desc.encode("ascii")`

src/python/bcc/__init__.py
```diff
@@ -768,15 +766,15 @@ def kprobe_poll(self, timeout = -1):
...
+        for k in list(self.open_kprobes):
+            lib.perf_reader_free(self.open_kprobes[k])
+            # perf buffers have tuple keys and don't need detaching
+            if not isinstance(k, tuple):
                 desc = "-:kprobes/%s" % k
                 lib.bpf_detach_kprobe(desc.encode("ascii"))
             self._del_kprobe(k)
-        for k, v in self.open_uprobes.items():
-            lib.perf_reader_free(v)
+        for k in list(self.open_uprobes):
```
drzaeus77: Ah, good, I had noticed this one as well, and my fix would have been the same.

drzaeus77: Actually, you could also have done:
```
for k, v in list(self.open_uprobes.items()):
```
Which would have avoided the `[k]` below.

markdrayton: Ah. I didn't realise that was possible. I can make that change.

src/python/bcc/__init__.py
```diff
@@ -388,7 +388,6 @@ def attach_kprobe(self, event="", fn_name="", event_re="",
         return self
 
     def detach_kprobe(self, event):
-        assert isinstance(event, str), "event must be a string"
```
drzaeus77: Was this being hit in practice? If str() is being wrapped around all internal decode() uses, wouldn't str be the default option?

markdrayton: Not from anything internal, as far as I could tell. It was primarily there to make sure callers passed in strings.

drzaeus77: If it's not being hit for some existing tests or obvious use cases, I'd say leave the check then. Or is there some other concern?

markdrayton: I *think* the only case in which it's awkward is when using py2 with unicode_literals. In both py2 and 3 strings (e.g. "__kprobe_foo") are type `str`. Only when importing unicode_literals are they something else (a `unicode`). As long as we're fine with asserting in this case they can stay.

Alternatively, we can cast event to `str` in `attach_*`.

PR #634, opened 2016-07-27
Added XDP support to BCC

Eichhoernchen:
This PR adds XDP support as currently supported in net-next.

Concretely, it adds two functions to bcc, namely: 
`attach_xdp` and `remove_xdp`
which allows to attach an XDP program to a device (given via its name, e.g., en0) (in the future this might change to a specific queue on a device once the kernel offers this interface)
and `remove_xdp` removes a XDP program from a device. Please note that there can currently be only one program attached to the device and attaching another program replaces the previous.

One example is available to test XDP, in networking/xdp which drops all packets an counts for which protocol a packet was dropped (this is taken from the kernel xdp1 example). Please note that you cannot use the network headers defined in `<bcc/proto.h>` as they cause llvm/clang to generate instructions not available on XDP layer. On XDP layer you do not have an skb yet, so you are operating on the bare packet data.

XDP support is currently limited to only some network adapters, there is the `mlx4` and there is also a patch available for the `e1000` driver.

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -408,3 +408,104 @@ error:
 
   return NULL;
 }
+
+
+
+int bpf_attach_xdp(const char *dev_name, int progfd) {
+    struct sockaddr_nl sa;
+    int sock, seq = 0, len, ret = -1;
+    char buf[4096];
+    struct nlattr *nla, *nla_xdp;
+    struct {
+        struct nlmsghdr  nh;
+        struct ifinfomsg ifinfo;
+        char             attrbuf[64];
+    } req;
+    struct nlmsghdr *nh;
+    struct nlmsgerr *err;
+
+    memset(&sa, 0, sizeof(sa));
+    sa.nl_family = AF_NETLINK;
+
+    sock = socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE);
+    if (sock < 0) {
+        fprintf(stderr, "bpf: opening a netlink socket: %s\n", strerror(errno));
+        return -1;
+    }
+
+    if (bind(sock, (struct sockaddr *)&sa, sizeof(sa)) < 0) {
+        fprintf(stderr, "bpf: bind to netlink: %s\n", strerror(errno));
+        goto cleanup;
+    }
+
+    memset(&req, 0, sizeof(req));
+    req.nh.nlmsg_len = NLMSG_LENGTH(sizeof(struct ifinfomsg));
+    req.nh.nlmsg_flags = NLM_F_REQUEST | NLM_F_ACK;
+    req.nh.nlmsg_type = RTM_SETLINK;
+    req.nh.nlmsg_pid = 0;
+    req.nh.nlmsg_seq = ++seq;
+    req.ifinfo.ifi_family = AF_UNSPEC;
+    req.ifinfo.ifi_index = if_nametoindex(dev_name);
+    if (req.ifinfo.ifi_index == 0) {
+        fprintf(stderr, "bpf: Resolving device name to index: %s\n", strerror(errno));
+        goto cleanup;
+    }
+
+    nla = (struct nlattr *)(((char *)&req)
+                            + NLMSG_ALIGN(req.nh.nlmsg_len));
+    nla->nla_type = NLA_F_NESTED | 43/*IFLA_XDP*/;
+
+    nla_xdp = (struct nlattr *)((char *)nla + NLA_HDRLEN);
+
+    // we specify the FD passed over by the user
+    nla_xdp->nla_type = 1/*IFLA_XDP_FD*/;
+    nla_xdp->nla_len = NLA_HDRLEN + sizeof(int);
+    memcpy((char *)nla_xdp + NLA_HDRLEN, &progfd, sizeof(progfd));
+    nla->nla_len = NLA_HDRLEN + nla_xdp->nla_len;
+
+    req.nh.nlmsg_len += NLA_ALIGN(nla->nla_len);
+
+    if (send(sock, &req, req.nh.nlmsg_len, 0) < 0) {
+        fprintf(stderr, "bpf: send to netlink: %s\n", strerror(errno));
+        goto cleanup;
+    }
+
+    len = recv(sock, buf, sizeof(buf), 0);
+    if (len < 0) {
+        fprintf(stderr, "bpf: recv from netlink: %s\n", strerror(errno));
+        goto cleanup;
+    }
+
+    for (nh = (struct nlmsghdr *)buf; NLMSG_OK(nh, len);
+         nh = NLMSG_NEXT(nh, len)) {
+        if (nh->nlmsg_pid != getpid()) {
+            fprintf(stderr, "bpf: Wrong pid %d, expected %d\n",
+                   nh->nlmsg_pid, getpid());
+            errno = EBADMSG;
+            goto cleanup;
+        }
+        if (nh->nlmsg_seq != seq) {
+            fprintf(stderr, "bpf: Wrong seq %d, expected %d\n",
+                   nh->nlmsg_seq, seq);
+            errno = EBADMSG;
+            goto cleanup;
+        }
+        switch (nh->nlmsg_type) {
+            case NLMSG_ERROR:
+                err = (struct nlmsgerr *)NLMSG_DATA(nh);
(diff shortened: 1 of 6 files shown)


4ast:
awesome. Thanks! lgtm @drzaeus77

examples/networking/xdp/xdp_drop_count.py
```diff
@@ -0,0 +1,123 @@
+from bcc import BPF
```
drzaeus77: I think it would help to chmod +x the flie that is checked in, as well as add `#/usr/bin/env python` as the first line.

examples/networking/xdp/xdp_drop_count.py
```diff
@@ -0,0 +1,123 @@
...
+import time
+import sys
+
+
```
drzaeus77: Can you please remove the whitespace?

Also, it would be good to add your own copyright notice to the top of the file.

examples/networking/xdp/xdp_drop_count.py
```diff
@@ -0,0 +1,123 @@
...
+    print("Usage: {0} <ifdev>\n\ne.g.: {0} eth0".format(sys.argv[0]))
+    exit(1)
+
+
+device = sys.argv[1]
+
+
+# load BPF program
+b = BPF(text = """
+    #include <uapi/linux/bpf.h>
```
drzaeus77: We've been taking the convention of aligning the embedded strings starting from column 0. That is, treat the string as if it were a C file starting from scratch, with indentation rules for C in effect after the opening """.

examples/networking/xdp/xdp_drop_count.py
```diff
@@ -0,0 +1,123 @@
...
+    #include <linux/in.h>
```
drzaeus77: When I ran this, I had to add:
```
#define KBUILD_MODNAME "foo"
```
In order to silence some warnings.

examples/networking/xdp/xdp_drop_count.py
```diff
@@ -0,0 +1,123 @@
...
+        struct ipv6hdr *ip6h = data + nh_off;
+
+        if ((void*)&ip6h[1] > data_end)
+            return 0;
+        return ip6h->nexthdr;
+    }
+
+
+
+    int xdp_prog1(struct xdp_md *ctx) {
```
drzaeus77: Opening braces here are inconsistent. In bcc, we're mostly using same-line opening braces, meaning this line is correct and the `parse_ipv[46]` functions should be compacted by 1 line.

src/cc/libbpf.c
```diff
@@ -408,3 +408,104 @@ void * bpf_open_perf_buffer(perf_reader_raw_cb raw_cb, void *cb_cookie, int pid,
...
+	    errno = ESPIPE;
+            goto cleanup;
+        }
+        switch (nh->nlmsg_type) {
+            case NLMSG_ERROR:
+                err = (struct nlmsgerr *)NLMSG_DATA(nh);
+                if (!err->error)
+                    continue;
+                fprintf(stderr, "bpf: nlmsg error %s\n", strerror(-err->error));
+		errno = -err->error;
```
drzaeus77: Some hard tabs got mixed in here, please remove them.

src/cc/libbpf.c
```diff
@@ -408,3 +408,104 @@ void * bpf_open_perf_buffer(perf_reader_raw_cb raw_cb, void *cb_cookie, int pid,
...
+        if (nh->nlmsg_pid != getpid()) {
+            fprintf(stderr, "bpf: Wrong pid %d, expected %d\n",
+                   nh->nlmsg_pid, getpid());
+	    errno = ESRCH;
+            goto cleanup;
+        }
+        if (nh->nlmsg_seq != seq) {
+            fprintf(stderr, "bpf: Wrong seq %d, expected %d\n",
+                   nh->nlmsg_seq, seq);
+	    errno = ESPIPE;
```
drzaeus77: These seem like odd choices to me, and besides there is already a printf with a better error message. Might make more sense just to leave the printf and remove the os.strerror from the python side, and then leave errno alone.

examples/networking/xdp/xdp_drop_count.py
Eichhoernchen: Agree will fix!

examples/networking/xdp/xdp_drop_count.py
Eichhoernchen: Okay, can do

src/cc/libbpf.c
Eichhoernchen: Hm, I disagree. The c-printf only appears to the user when debug=4, otherwise noting meaningful would be printed. I chose these errno codes as they they came close to what happens there.

src/cc/libbpf.c
Eichhoernchen: Will do!

examples/networking/xdp/xdp_drop_count.py
Eichhoernchen: Will do!

examples/networking/xdp/xdp_drop_count.py
Eichhoernchen: I also get a warning and I could not figure out the reason for it, can you give some insights why this silences the warning?

examples/networking/xdp/xdp_drop_count.py
Eichhoernchen: Okay, I will fix this

Eichhoernchen:
I will incorporate Brendens comments and update the PR tomorrow!

Eichhoernchen:
Hi,
I incorporated @drzaeus77 remarks and I also changed the error behavior.
Now errno is set to EBADMSG and the python part checks that message and recommends increasing the debug level

examples/networking/xdp/xdp_drop_count.py
```diff
@@ -0,0 +1,135 @@
+#/usr/bin/env python
+#
+# xdp_drop_count.py Drop incomming packets on XDP layer and count for which
+#                   protocol type
+#
+# C code adapted from Brenden Blanco as found in kernel:samples/bpf/xdp1_kern.c
+#
+# Copyright (c) 2016 Jan Rüth
+# Licensed under the Apache License, Version 2.0 (the "License")
```
drzaeus77: Hmm, I don't think you are allowed to do this (change license). Please drop this file and I will resubmit with the changed license.

drzaeus77:
Also, if you would, please squash the commits. You can do it locally and just force push to your same branch, and github will pick up the new git log automatically, while still preserving these comments.

examples/networking/xdp/xdp_drop_count.py
Eichhoernchen: Hm, yes I was unsure about it, as I only adapted parts in the C file, I think I have to adapt the GPL license, however the C file is actually not really part of the python file. I thought it would be ok to include the GPL note in the C part and just use the apache license (as it is used also in the other files) for the python part. would it be better to have them as separate files? Would that solve things? Do you mean drop the whole python file?
It would also be fine with me to just have the GPL license in the file, I don't know if the project requires anything.

drzaeus77: 1. I would prefer to have this file checked in as ASL2
2. The original file is GPL (C) PLUMgrid, which you can include as-is but not re-license
3. I can re-license the original file as ASL2, on behalf of PLUMgrid (my employer)
4. Therefore, to achieve 1, I would prefer to be the author of this file, and will include a more appropriate license for the bcc project.
5. You can then later adapt this file, and add your (C), since it will be ASL2 already.

drzaeus77: In other words, yes please drop the whole python file, the end result will be cleaner from a licensing point of view.

Eichhoernchen:
Okay, I removed the example and squashed the commits together.
Thanks for your patience and for making me aware how to use git properly.
Also thanks for your code review and your comments. This is my first push request, I hope I won't make the same mistakes the next time.

If there is still anything you like changed let me know

drzaeus77:
Nicely done!

examples/networking/xdp/xdp_drop_count.py
Eichhoernchen: ok great, thank you

PR #615, opened 2016-07-14
[cachetop] top-like cachestat

chantra:
Alike cachestat.py but providing cache stats at the process level.

PR #635, opened 2016-07-28
[cachetop] fix stats computation per processes.

chantra:
The current logic was only initializing page accesses, mark dirty.. at
the beginning of the method, preventing counters to be ever reset for
each PIDs.

Piggyback https://github.com/iovisor/bcc/pull/615#discussion_r71056842

Tested by running both tools manually.

### DIFF
--- a/tools/cachestat.py
+++ b/tools/cachestat.py
@@ -139,24 +139,16 @@ while 1:
     for k, v in sorted(counts.items(), key=lambda counts: counts[1].value):
 
         if re.match('mark_page_accessed', b.ksym(k.ip)) is not None:
-            mpa = v.value
-            if mpa < 0:
-                mpa = 0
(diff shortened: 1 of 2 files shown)


tools/cachestat.py
```diff
@@ -139,24 +139,16 @@ def usage():
     for k, v in sorted(counts.items(), key=lambda counts: counts[1].value):
 
         if re.match('mark_page_accessed', b.ksym(k.ip)) is not None:
-            mpa = v.value
-            if mpa < 0:
-                mpa = 0
+            mpa = max(0, mpa)
```
4ast: should it be max(0, v.value) ?

chantra: /facepalm. totally.... I only have 1 line right which was enough to visually look sane :s

PR #638, opened 2016-07-29
Try to demangle C++ symbols

palmtenor:
For C++ programs, this would make outputted stack traces look nicer.
Example: http://pastebin.com/LqT0nP67

### DIFF
--- a/src/cc/bcc_syms.cc
+++ b/src/cc/bcc_syms.cc
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+#include <cxxabi.h>
 #include <string.h>
 #include <sys/stat.h>
 #include <sys/types.h>
@@ -52,6 +53,7 @@ bool KSyms::resolve_addr(uint64_t addr, struct bcc_symbol *sym) {
 
   if (syms_.empty()) {
     sym->name = nullptr;
+    sym->demangle_name = nullptr;
     sym->module = nullptr;
     sym->offset = 0x0;
     return false;
@@ -59,6 +61,7 @@ bool KSyms::resolve_addr(uint64_t addr, struct bcc_symbol *sym) {
 
   auto it = std::upper_bound(syms_.begin(), syms_.end(), Symbol("", addr)) - 1;
   sym->name = (*it).name.c_str();
+  sym->demangle_name = sym->name;
   sym->module = "[kernel]";
   sym->offset = addr - (*it).addr;
   return true;
@@ -108,11 +111,19 @@ bool ProcSyms::resolve_addr(uint64_t addr, struct bcc_symbol *sym) {
 
   sym->module = nullptr;
   sym->name = nullptr;
+  sym->demangle_name = nullptr;
   sym->offset = 0x0;
 
   for (Module &mod : modules_) {
-    if (addr >= mod.start_ && addr < mod.end_)
-      return mod.find_addr(addr, sym);
+    if (addr >= mod.start_ && addr < mod.end_) {
+      bool res = mod.find_addr(addr, sym);
+      if (sym->name) {
+        sym->demangle_name = abi::__cxa_demangle(sym->name, nullptr, nullptr, nullptr);
+        if (!sym->demangle_name)
+          sym->demangle_name = sym->name;
+      }
+      return res;
+    }
   }
   return false;
 }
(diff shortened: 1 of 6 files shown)


src/python/bcc/__init__.py
```diff
@@ -746,6 +748,19 @@ def _sym_cache(pid):
         return BPF._sym_caches[pid]
 
     @staticmethod
+    def _demangle(name):
+        if name not in BPF._demangle_caches:
+            try:
+                pipe = subprocess.Popen(
```
4ast: nice and useful, but can we do it on c++ side just calling a function instead of a process from python?

palmtenor:
Do the demangling in C++ side instead.
I have little knowledge about Lua so I will leave that part untouched for now...
@4ast: I don't know what would be the effect if I change the `bcc_symbol->name` in-place, so I added a new field.

src/cc/bcc_syms.cc
```diff
@@ -108,11 +111,22 @@ bool ProcSyms::resolve_addr(uint64_t addr, struct bcc_symbol *sym) {
...
   sym->offset = 0x0;
 
   for (Module &mod : modules_) {
-    if (addr >= mod.start_ && addr < mod.end_)
-      return mod.find_addr(addr, sym);
+    if (addr >= mod.start_ && addr < mod.end_) {
+      bool res = mod.find_addr(addr, sym);
+      if (sym->name) {
+        sym->demangle_name = abi::__cxa_demangle(
+          sym->name, nullptr, nullptr, nullptr
```
4ast: the coding style looks a bit odd... why not in the same line? other than that it looks great!

src/cc/bcc_syms.cc
```diff
@@ -52,13 +53,15 @@ bool KSyms::resolve_addr(uint64_t addr, struct bcc_symbol *sym) {
...
     sym->name = nullptr;
+    sym->demangle_name = nullptr;
     sym->module = nullptr;
     sym->offset = 0x0;
     return false;
   }
 
   auto it = std::upper_bound(syms_.begin(), syms_.end(), Symbol("", addr)) - 1;
   sym->name = (*it).name.c_str();
+  sym->demangle_name = sym->name;
```
palmtenor: I assume no demangling are needed for kernel stack, right?

src/cc/bcc_syms.cc
```diff
@@ -108,11 +111,22 @@ bool ProcSyms::resolve_addr(uint64_t addr, struct bcc_symbol *sym) {
...
-      return mod.find_addr(addr, sym);
+    if (addr >= mod.start_ && addr < mod.end_) {
+      bool res = mod.find_addr(addr, sym);
+      if (sym->name) {
+        sym->demangle_name = abi::__cxa_demangle(
+          sym->name, nullptr, nullptr, nullptr
+        );
+        if (!sym->demangle_name) {
+          sym->demangle_name = sym->name;
+        }
```
4ast: extra { } are also unnecessary

src/cc/bcc_syms.cc
4ast: yeah. no demangling for kernel :)

src/cc/bcc_syms.cc
palmtenor: I have hard 80-char limit marker on my editor :P

4ast:
LGTM

src/lua/bcc/libbcc.lua
```diff
@@ -101,6 +101,7 @@ void perf_reader_set_fd(struct perf_reader *reader, int fd);
 ffi.cdef[[
 struct bcc_symbol {
 	const char *name;
+    const char *demangle_name;
```
markdrayton: indentation :)

markdrayton:
As far as I can tell you also need to change https://github.com/iovisor/bcc/blob/d516406d1be4355f009d2bd6ca79c08bf015666d/src/lua/bcc/sym.lua#L28 to make the Lua side use the demangled name.

markdrayton:
Also thanks for doings this :)

palmtenor:
Thanks @markdrayton! Changed according to comment. Didn't locally verify Lua builds though, hoping the tests would cover:)

4ast:
can you please squash all of the commits into one?
github will preserve the discussion/comments and git history will be less noisy.
Thanks!

palmtenor:
Done:)

4ast:
Thanks!

PR #639, opened 2016-07-31
tcpretrans: ntohs dport

markdrayton:
`dport` is network order in the kernel and needs `ntohs` to be unravelled for human consumption. `sport` is host order and doesn't need changing (see http://lists.openwall.net/netdev/2013/10/10/44).


4ast:
good catch!

PR #642, opened 2016-08-01
MySQL USDT tool and example

brendangregg:


4ast:
lgtm.
I wonder how we can do uprobes (and usdt) tracing without full context switch into kernel.

PR #643, opened 2016-08-01
add open_perf_event api for reading perf counters

drzaeus77:
Though the rewriter table method existed, we were not yet opening up the
perf counter in the proper way for it to be read.

Introduce a c function to enable attaching perf counters to cpu-indexed
table slots. The python side opens, assigns, then immediately closes the
fd. Only the kernel keeps a reference, so closing the table or deleting
the entry will be sufficient to free up the fd when finished.


4ast:
LGTM

PR #645, opened 2016-08-03
funccount: filter bcc function count

oujunli:


drzaeus77:
ok to test

brendangregg:
style issues (these tools are included in other environments that will run pep8):

```
# pep8 --show-source --ignore=E123,E125,E126,E127,E128,E302 funccount.py 
funccount.py:9:80: E501 line too long (80 > 79 characters)
# The pattern is a string with optional '*' wildcards, similar to file globbing.
                                                                               ^
funccount.py:81:80: E501 line too long (103 > 79 characters)
filter_text = 'u32 pid; pid = bpf_get_current_pid_tgid(); if (pid != %s) { return 0; }' % (os.getpid())
                                                                               ^
funccount.py:83:51: E225 missing whitespace around operator
    filter_text += ' if (pid != %s){ return 0;}' %(args.pid)
                                                  ^
funccount.py:124:24: E701 multiple statements on one line (colon)
        if v.value == 0: continue
                       ^
```

Also, when using "-p PID", the logic has redundancies; eg, from the compiled BPF script:

```
    u32 pid; pid = bpf_get_current_pid_tgid(); if (pid != 19045) { return 0; } if (pid != 11){ return 0;}
```

I passed it "-p 11".

oujunli:
I fix the style issues and logical error, generate commit.
`u32 pid; pid = bpf_get_current_pid_tgid(); if (pid == 10075 || pid != 2716) { return 0; }`
run python tools/funccount.py -p 2716 "vfs_*",  It filter the current bcc process 10075 or pid not equals 2716.
Sorry about that. Thanks.

brendangregg:
I'd fix this further: these are early examples, and people are likely to copy-n-paste what we create to make other tools, so it's worth getting as nice as possible to start with.

That filter should be:

```C
 if (pid != 2716) { return 0; }
```

If you've specified -p, I don't see a reason to also check that it's not its own PID. ie, the code could be:

```Python
if args.pid:
    filter_text += "if (pid != %s) { return 0; }" % args.pid
else:
    filter_text += "if (pid == %s) { return 0; }" % os.getpid()
```

unless I've missed something? Thanks.

oujunli:
Yes, you are right. when pid special, no reason to check it own PID again,  fix it:

`u32 pid; pid = bpf_get_current_pid_tgid(); if (pid == 15357) { return 0; }`

run `python funccount.py  "vfs_*"`, and 

`u32 pid; pid = bpf_get_current_pid_tgid(); if (pid != 2716) { return 0; }`

run `python funccount.py  -p 2716 "vfs_*"`.

Thanks.

brendangregg:
ok to test

brendangregg:
ok, LGTM.

Issue #633, opened 2016-07-27
python 3.4, llvm 3.8.1 -> segfault


Serafean:
backtrace (likely incomplete, if needed I'll recompile with debug symbols) :
(gdb) bt
#0  0x00007f3c4e7eafa7 in raise () from /lib64/libc.so.6
#1  0x00007f3c4e7ec30a in abort () from /lib64/libc.so.6
#2  0x00007f3c4e829394 in ?? () from /lib64/libc.so.6
#3  0x00007f3c4e82ec06 in ?? () from /lib64/libc.so.6
#4  0x00007f3c4e82f963 in ?? () from /lib64/libc.so.6
#5  0x00007f3c4c8f81c2 in ?? () from /usr/lib64/libclangFrontend.so.3.8
#6  0x00007f3c4c8ff266 in clang::CompilerInvocation::CreateFromArgs(clang::CompilerInvocation&, char const* const*, char const* const*, clang::DiagnosticsEngine&) () from /usr/lib64/libclangFrontend.so.3.8
#7  0x00007f3c4cc0eef4 in ebpf::ClangLoader::parse(std::unique_ptr<llvm::Module, std::default_delete<llvm::Module> >*, std::unique_ptr<std::vector<ebpf::TableDesc, std::allocator<ebpf::TableDesc> >, std::default_delete<std::vector<ebpf::TableDesc, std::allocator<ebpf::TableDesc> > > >*, std::string const&, bool, char const**, int) () from /usr/lib64/libbcc.so.0
#8  0x00007f3c4cbbf831 in ebpf::BPFModule::load_cfile(std::string const&, bool, char const**, int) () from /usr/lib64/libbcc.so.0
#9  0x00007f3c4cbc5722 in ebpf::BPFModule::load_c(std::string const&, char const**, int) () from /usr/lib64/libbcc.so.0
#10 0x00007f3c4cbbe9f4 in bpf_module_create_c () from /usr/lib64/libbcc.so.0
#11 0x00007f3c4da99f8e in ffi_call_unix64 () from /usr/lib64/libffi.so.6
#12 0x00007f3c4da999f8 in ffi_call () from /usr/lib64/libffi.so.6
#13 0x00007f3c4dd051a6 in _ctypes_callproc () from /usr/lib64/python3.4/lib-dynload/_ctypes.cpython-34m.so
#14 0x00007f3c4dcff68f in ?? () from /usr/lib64/python3.4/lib-dynload/_ctypes.cpython-34m.so
[...]

PR #647, opened 2016-08-03
Disable static-libstdc++ when clang is linked dynamically

drzaeus77:
Based on the bug report in
https://bugs.gentoo.org/show_bug.cgi?id=582770, mixing static+non-static
libstdc++ can lead to crashes. Disable such combinations. Choosing to
leave out the llvm check, since in practice clang is less likely to be
provided statically, so the check should cover both cases.

Fixes: #633


drzaeus77:
@Serafean @mlen would this approximate a fix? Based on the bug report, mixing static/non-static libstdc++ is the root cause.

4ast:
so it crashes because libstdc++ is static, but clang libs are using libstdc++.so ?

drzaeus77:
That was how I read it, yes. And I think we have heard that before, when diagnosing some crashes when ProcUtils was introduced, and strings were being passed by pointer from one type of allocation to another.

4ast:
the fix makes sense to me then

PR #651, opened 2016-08-08
Remove unnecessary delete() in sync_timing.py

pchaigno:
This pull request removes an unnecessary `.delete(&key)` in `sync_timing.py` and updates the tutorial accordingly.


drzaeus77:
ok to test

4ast:
the delete() is actually necessary at the moment due to the kernel bug in the update() that was introduced in May and bugfix wasn't merged yet.
Otherwise update() will eventually fill the whole map and new updates will stop working.

drzaeus77:
Link to bugfix?

Also, in that case, @pchaigno could you update this PR to be a comment mentioning the reason for the extra .delete().

pchaigno:
Thanks for the explanation @4ast! 

@drzaeus77 I'm on it!

4ast:
> link to bugfix

https://git.kernel.org/cgit/linux/kernel/git/davem/net.git/commit/?id=a6ed3ea65d9868fdf9eff84e6fe4f666b8d14b02
it's in net tree, but not in linus tree yet.

pchaigno:
Pull request updated!

PR #654, opened 2016-08-11
Add some explanation for functions in BPF program in Tutorial

palmtenor:
I was confused by this when I first started. The compiler will complain about the helper functions having wrong arguments, which is not very obvious what went wrong. Talked with @4ast offline, he suggested adding this to the Tutorial.


brendangregg:
ok to test

brendangregg:
Thanks! Just keep being brief. If you add a lot of text in the future, try putting in headings to avoid a wall of text.

Issue #653, opened 2016-08-10
error: too many arguments to function call


markdrayton:
I'm trying to update dcsnoop.py to use PERF_EVENT_OUTPUT (#540). Since there are two places that submit data back to userspace I made a simple wrapper function to deduplicate this code:

``` c
enum lookup_type {
    LOOKUP_MISS,
    LOOKUP_REFERENCE,
};

struct data_t {
    u32 pid;  // PID as in the userspace term (i.e. task->tgid in kernel)
    enum lookup_type type;
    char comm[TASK_COMM_LEN];
    char name[MAX_FILE_LEN];
};

BPF_PERF_OUTPUT(events);

void submit_event(struct pt_regs *ctx, void *name, int type, u32 pid)
{
    struct data_t data = {
        .pid = pid,
        .type = type,
    };
    bpf_get_current_comm(&data.comm, sizeof(data.comm));
    bpf_probe_read(&data.name, sizeof(data.name), name);
    events.perf_submit(ctx, &data, sizeof(data));
}

// later

submit_event(ctx, (void *)nd->last.name, LOOKUP_REFERENCE, pid);
```

(full code at https://gist.github.com/markdrayton/9c7751b1bc64184af90e83ee52140c0c)

When running I get:

```
/virtual/main.c:51:23: error: too many arguments to function call, expected single argument 'ctx', have 4 arguments
    submit_event(ctx, (void *)({ typeof(const unsigned char *) _val; memset(&_val, 0, sizeof(_val)); bpf_probe_read(&_val, sizeof(_val), (u64)nd + offsetof(struct nameidata, last.name)); _val; }), LOOKUP_REFERENCE, pid);
    ~~~~~~~~~~~~      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/virtual/main.c:35:1: note: 'submit_event' declared here
__attribute__((section(".bpf.fn.submit_event")))
^
/virtual/main.c:77:23: error: too many arguments to function call, expected single argument 'ctx', have 4 arguments
[...]

PR #656, opened 2016-08-11
Disable non-static function calls

drzaeus77:
Compiled BPF programs must consist of a single contiguous code block,
meaning trying to call other function entry points (besides the
kernel-defined helpers) is not possible. The bcc frontend didn't
explicitly prohibit this, even though the program would fail to
compile/load. Add an explicit check and error message.

Fixes: #653
Signed-off-by: Brenden Blanco bblanco@plumgrid.com

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -462,6 +462,15 @@ bool BTypeVisitor::VisitCallExpr(CallExpr *Call) {
           rewriter_.ReplaceText(expansionRange(Call->getSourceRange()), text);
         }
       }
+    } else if (FunctionDecl *F = dyn_cast<FunctionDecl>(Decl)) {
+      if (F->isExternallyVisible() && !F->getBuiltinID()) {
+        auto start_loc = rewriter_.getSourceMgr().getFileLoc(Decl->getLocStart());
+        if (rewriter_.getSourceMgr().getFileID(start_loc)
+            == rewriter_.getSourceMgr().getMainFileID()) {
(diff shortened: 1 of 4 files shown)


drzaeus77:
retest this please

4ast:
which tools/* did you check that are still working fine with this extra check ?
I don't think there should be any breakage, but...

drzaeus77:
None, found a couple that will need fixing, please hold this PR.

goldshtn:
@drzaeus77 How come we never got this merged? It looks good.

goldshtn:
@drzaeus77 Ping?

drzaeus77:
Rebased, updated, and went through tools/* testing.

4ast:
i guess that's fine for now.
To future proof things can we force inline them even when they're non-static ?
we'll try to do this in llvm backend. Can bcc pass do that as well?

drzaeus77:
@4ast 
See BPFModule::annotate()

```
  for (auto fn = mod_->getFunctionList().begin(); fn != mod_->getFunctionList().end(); ++fn)
    if (!fn->hasFnAttribute(Attribute::NoInline))
      fn->addFnAttr(Attribute::AlwaysInline);
```

4ast:
yes. I remember that bit and there is LLVMAddAlwaysInlinerPass(), but it still doesn't quite work.
The most recent llvm change switched default for -O0 to disable inline, so even functions marked with inline become non-inlinable and not eligible for alwaysinline somehow.
May be 'if (!fn->hasFnAttribute(Attribute::NoInline))' is causing the issues with new llvm ?
@iamkafai @palmtenor @shodoco

PR #506, opened 2016-04-26
update debian packaging

lnussbaum:
Hi,

The changes attached are currently required to build the Debian package properly.

If you are ever interested in getting this into Debian, don't hesite to ask!

Lucas

Issue #397, opened 2016-02-18
Location of the kernel headers


ismail:
I am trying to run bcc on openSUSE Linux and there is a problem with the kernel header path it seems:

```
> ./hello_world.py
In file included from <built-in>:316:
<command line>:2:10: fatal error: './include/linux/kconfig.h' file not found
#include "./include/linux/kconfig.h"
         ^
1 error generated.
Traceback (most recent call last):
  File "./hello_world.py", line 11, in <module>
    BPF(text='void kprobe__sys_clone(void *ctx) { bpf_trace_printk("Hello, World!\\n"); }').trace_print()
  File "/usr/local/lib/python2.7/site-packages/bcc/__init__.py", line 126, in __init__
    raise Exception("Failed to compile BPF module %s" % src_file)
Exception: Failed to compile BPF module
```

The file exists at /usr/src/linux/include/linux/kconfig.h , I am wondering if there is a special setup I should be doing before using bcc.

drzaeus77:
On most systems, I find the kernel headers in `/lib/modules/$(uname -r)/build`, and therefore the bcc sources (specifically `kbuild_helper.{cc,h}`) are implemented to search at that location. Usually the `build` file is a symlink to the architecture specific location. In SUSE, is there a kernel-headers package that can be installed in addition to the kernel image?

ismail:
There is kernel-default-devel which provides /lib/modules/$(uname -r)/build but the header structure is very different than what I expect:

```
> ls -al /lib/modules/$(uname -r)/build/include/
total 52
drwxr-xr-x   4 root root  4096 Feb  5 19:03 ./
drwxr-xr-x   5 root root  4096 Feb  8 10:19 ../
drwxr-xr-x 743 root root 36864 Feb  5 19:03 config/
drwxr-xr-x   3 root root  4096 Feb  5 19:03 generated/
```

ismail:
This is obviously a SUSE related problem, I am closing the bug. Will try to ask our kernel devs.

drzaeus77:
:) I imagine there is a reasonable answer. My hope is that you aren't required to install the full kernel sources. Let us know when you find out.

lcp:
I have a quick fix.
https://build.opensuse.org/package/view_file/home:gary_lin:iovisor/bcc/bcc-suse-kernel-headers.patch?expand=1

ismail:
@lcp thats awesome! Maybe it can be upstreamed :-)

lcp:
[...]

PR #657, opened 2016-08-11
update debian packaging

drzaeus77:
Updated version of PR #506 with included clang-format-3.7 fix.


drzaeus77:
retest this please

drzaeus77:
@lnussbaum @taem The last commit in this series breaks Ubuntu packaging (which doesn't have `/lib/modules/$(uname -r)/source`, so this would need to be made into a toggle for debian vs ubuntu. Do you have a preferred mechanism (env variables, multiple files, etc.) to toggle this in the relevant build system?

taem:
Hi @drzaeus77,

You can use dpkg-vendor(1) to change the build behavior as per distribution.

lnussbaum:
@taem, yes that would be the way to go.

However, I'm surprised that this breaks on Ubuntu. Maybe there's another way to achieve the same thing that works on both distributions.

drzaeus77:
In our previous experiments, most of the distros have unique preferences w.r.t. kernel header directories. See, for example, #397. In other words, I'm not at all surprised that it broke :)

lnussbaum:
Yeah, but I suspected that there's a way to make this work automatically on both Debian and Ubuntu. Given Ubuntu tries to pull as many packages as possible from Debian without modification, I would be surprised if they introduced a difference that would cause some stuff to break.

PR #661, opened 2016-08-13
killsnoop: use current time replace timestamp and default output

oujunli:
the option -t give the delta from the first trace kill sig, current time get  more info when solve problem.


brendangregg:
Timestamps from python will be slightly inaccurate, but in this case -- since we're printing it to whole seconds -- it shouldn't matter, and the reduction in code is good. I also don't like increasing the default output unless the fields are really necessary, but in this case it makes sense.

brendangregg:
ok to test

brendangregg:
thanks!

PR #663, opened 2016-08-16
Tool to sniff data contents before encrypted with OpenSSL

adrianlzt:
Add tool as talked in iovisor-dev 'BCC: bpf_probe_read read function arguments'

### DIFF
--- a/man/man8/sslsniff.8
+++ b/man/man8/sslsniff.8
@@ -1,0 +1,51 @@
+.TH sslsniff 8  "2016-08-16" "USER COMMANDS"
+.SH NAME
+sslsniff \- Print data passed to OpenSSL. Uses Linux eBPF/bcc.
+.SH SYNOPSIS
+.B sslsniff
+.SH DESCRIPTION
+sslsniff prints data sent to SSL_write and SSL_read OpenSSL functions, allowing
+us to read plain text content before encryption (when writing) and after
+decryption (when reading).
+
+This works reading the second parameter of both functions (*buf).
+
+Since this uses BPF, only the root user can use this tool.
+.SH REQUIREMENTS
+CONFIG_BPF and bcc.
+.SH EXAMPLES
+.TP
+Print all calls to SSL_write and SSL_read system-wide:
+#
+.B sslsniff
+.SH FIELDS
+.TP
+FUNC
+Which function is being called (SSL_write or SSL_read)
+.TP
+TIME
+Time of the command, in seconds.
+.TP
+COMM
+Entered command.
+.TP
+PID
+Process ID calling OpenSSL.
+.TP
+LEN
+Bytes written or read by OpenSSL functions.
+.SH SOURCE
+This is from bcc.
+.IP
+https://github.com/iovisor/bcc
+.PP
+Also look in the bcc distribution for a companion _examples.txt file containing
+example usage, output, and commentary for this tool.
+.SH OS
+Linux
+.SH STABILITY
+Unstable - in development.
+.SH AUTHORS
+Adrian Lopez and Mark Drayton
+.SH SEE ALSO
+trace(8)
--- a/tools/sslsniff.py
+++ b/tools/sslsniff.py
@@ -1,16 +1,46 @@
 #!/usr/bin/python
 #
-# sslsniff  Captures data on SSL_READ or SSL_WRITE functions of OpenSSL
+# sslsniff  Captures data on read/recv or write/send functions of OpenSSL and
+#           GnuTLS
 #           For Linux, uses BCC, eBPF.
+#
+# USAGE: sslsniff.py [-h] [-p PID] [-c COMM] [-o] [-g] [-d]
 #
 # Licensed under the Apache License, Version 2.0 (the "License")
 #
 # 12-Aug-2016    Adrian Lopez   Created this.
 # 13-Aug-2016    Mark Drayton   Fix SSL_Read
+# 17-Aug-2016    Adrian Lopez   Capture GnuTLS and add options
+#
 
 from __future__ import print_function
 import ctypes as ct
 from bcc import BPF
+import argparse
+
+# arguments
+examples = """examples:
+    ./sslsniff              # sniff OpenSSL and GnuTLS functions
+    ./sslsniff -p 181       # sniff PID 181 only
+    ./sslsniff -c curl      # sniff curl command only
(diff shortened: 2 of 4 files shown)


adrianlzt:
@markdrayton I take your version and clean it a little

markdrayton:
Excellent, thanks! I have a few review remarks — will add them inline :)

Also, can you make a man page? It should be easy to copy and modify an example from `man/man8`.

tools/sslsniff.py
```diff
@@ -0,0 +1,136 @@
...
+        perf_SSL_read.perf_submit(ctx, &__data, sizeof(__data));
+        return 0;
+}
+"""
+
+b = BPF(text=prog)
+libpath = "/usr/lib/libssl.so.1.0.0"
+
+# Join to ssl_write
+b.attach_uprobe(name=libpath, sym="SSL_write", fn_name="probe_SSL_write")
```
markdrayton: You can just use `name="ssl"` here. It'll be resolved into the correct path to libssl (which may differ from yours) automagically.

tools/sslsniff.py
```diff
@@ -0,0 +1,136 @@
...
+{
+        u64 timestamp_ns;
+        u32 pid;
+        char comm[TASK_COMM_LEN];
+        char v0[476];
+};
+
+BPF_PERF_OUTPUT(perf_SSL_write);
+
+int probe_SSL_write(struct pt_regs *ctx)
```
markdrayton: In my quick test it appears that listing args on the function definition works, e.g.:

    int probe_SSL_write(struct pt_regs *ctx, void *ssl, void *buf, int num) {

(same for the other probes)

This is (IMO) clearer than using the PT_REGS_* macros to access the function args.

tools/sslsniff.py
```diff
@@ -0,0 +1,136 @@
...
+        __data.timestamp_ns = bpf_ktime_get_ns();
+        __data.pid = bpf_get_current_pid_tgid();
+
+        bpf_get_current_comm(&__data.comm, sizeof(__data.comm));
+
+        int ret;
+
+        if ( (void *) PT_REGS_PARM2(ctx) != 0) {
+                bpf_probe_read(&__data.v0, sizeof(__data.v0), (void *) ctx->si);
+                ret = bpf_probe_read(&__data.v0, sizeof(__data.v0), (unsigned char*) PT_REGS_PARM2(ctx));
```
markdrayton: Do we need both `bpf_probe_reads`? Also, `ret` doesn't seem to be used.

tools/sslsniff.py
```diff
@@ -0,0 +1,136 @@
...
+        struct probe_SSL_data_t __data = {0};
+        __data.timestamp_ns = bpf_ktime_get_ns();
+        __data.pid = pid;
+
+        bpf_get_current_comm(&__data.comm, sizeof(__data.comm));
+
+        int ret;
+        if (bufp != 0) {
+                ret = bpf_probe_read(&__data.v0, sizeof(__data.v0), (char *)*bufp);
+        }
```
markdrayton: same for `ret`.

markdrayton:
It'd also be nice if there was an indicator when data is truncated, just so people aren't surprised that only the first N bytes are captured.

Also, @4ast, are there any clever ways to work around the stack size limitation when sending data back through perf?

markdrayton:
One last request: can you PEP8 format the code and add `from __future__ import print_function` for consistency with the other tools?

tools/sslsniff.py
```diff
@@ -0,0 +1,150 @@
...
+        u32 pid = bpf_get_current_pid_tgid();
+        u64 *bufp = bufs.lookup(&pid);
+        if (bufp == 0) {
+                return 0;
+        }
+
+        struct probe_SSL_data_t __data = {0};
+        __data.timestamp_ns = bpf_ktime_get_ns();
+        __data.pid = pid;
+        __data.len = num;
```
markdrayton: `num` is the number of bytes the user asked for, which isn't always the number read. The return value of `SSL_read` (e.g. `PT_REGS_RC(ctx)`) will give you the number read.

tools/sslsniff.py
```diff
@@ -0,0 +1,150 @@
...
+        start = event.timestamp_ns
+    time_s = (float(event.timestamp_ns - start)) / 1000000000
+
+    s_mark = "-" * 5 + " DATA " + "-" * 5
+
+    e_mark = "-" * 5 + " END DATA " + "-" * 5
+
+    # If data length is exactly MAX_BUF_SIZE, it will be reported
+    # as truncated when it is not
+    if len(event.v0) == MAX_BUF_SIZE:
```
markdrayton: Use `if event.len > MAX_BUF_SIZE` (after changing `len` as above) to accurately tell if we truncated.

tools/sslsniff.py
```diff
@@ -0,0 +1,150 @@
...
+
+int probe_SSL_write(struct pt_regs *ctx, void *ssl, void *buf, int num) {
+        struct probe_SSL_data_t __data = {0};
+        __data.timestamp_ns = bpf_ktime_get_ns();
+        __data.pid = bpf_get_current_pid_tgid();
+        __data.len = num;
+
+        bpf_get_current_comm(&__data.comm, sizeof(__data.comm));
+
+        if ( (void *) buf != 0) {
```
markdrayton: I think you can drop the `void *` casts here and below.

man/man8/sslsniff.8
```diff
@@ -0,0 +1,48 @@
...
+Print all calls to SSL_write and SSL_read system-wide:
+#
+.B sslsniff
+.SH FIELDS
+.TP
+FUNC
+Which function is being called (SSL_write or SSL_read)
+.TP
+TIME
+Time of the command (HH:MM:SS).
```
markdrayton: nit: the timestamp is in seconds, not HH:MM:SS (one day we should make timestamps in all the tools consistent!)

markdrayton:
Looking good! Once the other things are finished can you squash this down to one commit and force push the branch?

brendangregg:
Nice! Please add a link in the tools list in README.md so people can quickly find the tool and example.

adrianlzt:
Thanks for the help improving this!
Will be great some _trick_ to be able to capture more data.

brendangregg:
ok to test

brendangregg:
LGTM, merging, thanks!

Issue #664, opened 2016-08-16
Segfault in RopePieceBTreeNode::erase


markdrayton:
This code segfaults in libbcc reproducibly:

``` python
from bcc import BPF

prog = """
#include <linux/memcontrol.h>
#include <linux/mm_types.h>

void trace(struct pt_regs *ctx, struct page *page) {
    bpf_trace_printk("%lu\\n", (unsigned long)page->mapping);
}
"""

b = BPF(text=prog)
b.attach_kprobe(event="mem_cgroup_commit_charge", fn_name="trace")
```

This is on a Ubuntu 14.04 host, BCC trunk, kernel 4.6.0, clang/LLVM 3.7 (as described in https://github.com/iovisor/bcc/blob/master/INSTALL.md).

Backtrace from GDB below. I tried `-DCMAKE_C_FLAGS=-g` but haven't been able to get the symbols working, unfortunately.

```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff4ce1564 in (anonymous namespace)::RopePieceBTreeNode::erase(unsigned int, unsigned int) () from /usr/lib/x86_64-linux-gnu/libbcc.so.0
(gdb) bt
#0  0x00007ffff4ce1564 in (anonymous namespace)::RopePieceBTreeNode::erase(unsigned int, unsigned int) () from /usr/lib/x86_64-linux-gnu/libbcc.so.0
#1  0x00007ffff4ce20be in clang::RewriteBuffer::ReplaceText(unsigned int, unsigned int, llvm::StringRef) () from /usr/lib/x86_64-linux-gnu/libbcc.so.0
#2  0x00007ffff4ce3d56 in clang::Rewriter::ReplaceText(clang::SourceLocation, unsigned int, llvm::StringRef) () from /usr/lib/x86_64-linux-gnu/libbcc.so.0
#3  0x00007ffff44221d0 in ebpf::ProbeVisitor::VisitMemberExpr(clang::MemberExpr*) () from /usr/lib/x86_64-linux-gnu/libbcc.so.0
#4  0x00007ffff44688d7 in clang::RecursiveASTVisitor<ebpf::ProbeVisitor>::TraverseMemberExpr(clang::MemberExpr*) () from /usr/lib/x86_64-linux-gnu/libbcc.so.0
[...]

PR #667, opened 2016-08-18
frontends/clang: Safety check for invalid opLoc in ProbeVisitor

drzaeus77:
As reported in #664:
In cases where the AST contains a MemberExpr with an invalid opLoc,
rewrite will fail but prior to this change would segfault. Fail a little
more gracefully, even though the program will still be rejected.


4ast:
LGTM

PR #672, opened 2016-08-22
Update references_guide.md kprobe section

juancb:
Updated the kprobe section to include the maximum number of args that can be passed is 4 including ctx.

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -99,7 +99,7 @@ This instruments the tcp_v4_connect() kernel function using a kprobe, with the f
 - ```struct pt_regs *ctx```: Registers and BPF context.
 - ```struct sock *sk```: First argument to tcp_v4_connect().
 
-The first argument is always ```struct pt_regs *```, the remainder are the arguments to the function (they don't need to be specified, if you don't intend to use them).
+The first argument is always ```struct pt_regs *```, the remainder are the arguments to the function (they don't need to be specified, if you don't intend to use them). Limitations: maximum of 4 args including ctx.
 
 Examples in situ:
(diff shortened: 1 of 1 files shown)


docs/reference_guide.md
```diff
@@ -99,7 +99,7 @@ This instruments the tcp_v4_connect() kernel function using a kprobe, with the f
 - ```struct pt_regs *ctx```: Registers and BPF context.
 - ```struct sock *sk```: First argument to tcp_v4_connect().
 
-The first argument is always ```struct pt_regs *```, the remainder are the arguments to the function (they don't need to be specified, if you don't intend to use them).
+The first argument is always ```struct pt_regs *```, the remainder are the arguments to the function (they don't need to be specified, if you don't intend to use them). Limitations: maximum of 4 args including ctx.
```
4ast: @drzaeus77 can we fix this limitation? It's only on rewriter side, right? since backend sees only 'ctx'

drzaeus77: It should be higher, 6 + ctx is what the rewriter supports. The test for overflow is in https://github.com/iovisor/bcc/blob/master/tests/python/test_clang.py#L345

juancb: Sorry, I ran into what I thought was a limitation on the number of args
being passed to BPF when in fact it was a bpf_trace_printk throwing the
error. Is there any reason we shouldn't check for the number of args passed
and throw a more helpful error? I wouldn't mind trying to implement that if
no one else opposed. Sorry if that last question is naive but I have never
worked with clang/llvm before, but it looks like it should be possible to
perform a check like in BTypeVisitor::VisitFunctionDecl()


Cryptic error thrown when passing more than 3 args to bpf_trace_printk()

error: <unknown>:0:0: in function kprobe__tcp_rcv_established i32
(%struct.pt_regs*): too many args to 0x7fa3ae20de90: i64 = Constant<6>

drzaeus77: Yes, the printk limitation is indeed as you say (3 parameters). A better error message would be most welcome, the check should be pretty simple. Try putting it into `BTypeVisitor::VisitCallExpr` in the `if (Decl->getName() == "bpf_trace_printk")` section. A test case to go with it in the same style as `tests/python/test_clang.py:test_too_many_args()` would be ideal.

Issue #665, opened 2016-08-17
funccount "SyS_.*" stuck on-CPU


oujunli:
When run python funccount.py "SyS_.*",  at the beginning, log:
```
write(/sys/kernel/debug/tracing/kprobe_events, "p:kprobes/p_SyS_sigsuspend SyS_sigsuspend") failed: Device or resource busy
```
when Ctrl-C, the tracing not end:
```
Tracing 337 functions for "SyS_.*"... Hit Ctrl-C to end.
^C
ADDR             FUNC                          COUNT
ffffffff81097230 sys_getrusage                     1
ffffffff81253d10 SyS_fstatfs                       1
ffffffff81092cc0 sys_kill                          1
ffffffff81095170 sys_getpriority                   1
ffffffff8121d030 sys_creat                         1
ffffffff8108bd30 sys_capset                        1
ffffffff81238c00 sys_getcwd                        1
ffffffff8170d4c0 sys_getpeername                   1
ffffffff81267d20 sys_epoll_create                  1
ffffffff81228220 SyS_pipe                          2
ffffffff810af5e0 sys_sched_getaffinity             2
ffffffff8170d6f0 SyS_shutdown                      2
ffffffff8121c2f0 sys_chdir                         2
ffffffff8113f250 sys_seccomp                       2
ffffffff8122fc60 sys_rmdir                         2
ffffffff8121d010 sys_openat                        2
ffffffff81223db0 SyS_readlink                      2
ffffffff810808f0 sys_set_tid_address               3
ffffffff8122fb60 sys_mkdir                         3
ffffffff81097140 sys_setrlimit                     3
ffffffff81228170 SyS_pipe2                         3
ffffffff812269e0 SyS_execve                        3
ffffffff8170da10 sys_sendmmsg                      3
ffffffff8170d470 sys_accept4                       3
ffffffff81097280 sys_umask                         3
ffffffff8123d620 sys_dup2                          4
ffffffff81085f50 sys_exit_group                    5
ffffffff81234d60 sys_ppoll                         6
ffffffff8170d480 sys_accept                        6
ffffffff8126bd10 sys_eventfd2                      7
ffffffff81086120 sys_wait4                         9
ffffffff8170d3b0 sys_bind                          9
ffffffff8122fcb0 sys_unlink                       10
ffffffff81096590 sys_newuname                     12
ffffffff81230d30 sys_rename                       12
ffffffff8170d4b0 sys_getsockname                  13
[...]

PR #676, opened 2016-08-25
Fix function keys enter an infinite loop (#665)

oujunli:


4ast:
that is still a race. how is better?

oujunli:
I log out the table.py zero routime:
```Python
    def zero(self):
        count = 0
        for k in self.keys():
            self[k] = self.Leaf()
            count += 1
            print("count: %d"%(count))
```
run python tools/funccount.py  "SyS_*":
```
Tracing 335 functions for "SyS_*"... Hit Ctrl-C to end.
^C
ADDR             FUNC                          COUNT
ffffffff811fc1d0 sys_execve                        1
ffffffff811f3f60 sys_lseek                         1
ffffffff8107cf30 sys_wait4                         1
...
ffffffff810f4690 SyS_futex                      7450
count: 1
count: 2
count: 3
...
count: 333308
count: 333309
count: 333310
...
````
Tracing 335 functions, but the zero for larger the 335 and enter an infinite loop. I set the map to zero when funccount  loops use counts.items(), avoid this problem and less call counts.Leaf(). If I missing something, please checkout. 

Thank you.

4ast:
then we need to fix zero() function instead of trying to workaround it in the script

drzaeus77:
The code should already be working around this. The way keys() works is relies on Iter(), which itself relies on the fact that bpf_get_next_key returns the least upper bound for the last position in the iteration, so it only loops through the list once, even if items are being added back into the map.

oujunli:
should we change the `zero` function loops through the list once as the `items` and `values` function?  
because funccount is used `items`, so depend on it's `item` loop solve this problem and no need to loop the map again to set value zero. can make another commit let `zero` function  loops through the list once.  
thanks.

oujunli:
I try to fix this issue by through the keys list once as the items and values function do, not solve by funccount script.  update the commit.  
thanks.

drzaeus77:
ok to test
[message to buildbot]

4ast:
@drzaeus77 pls review

drzaeus77:
FYI, I'm currently investigating the kernel behavior w.r.t. this patch. It may possibly be a kernel-side issue, and we'll decide best how/if to workaround this in userspace after getting a better understanding.

oujunli:
OK. If want more information and test, I am very happy to provide.  

thanks.

oujunli:
a better way in [780](https://github.com/iovisor/bcc/pull/780#issuecomment-255966580), close.

PR #680, opened 2016-09-02
 BPF buffer log

davidefdl:
This fix is addressing the buffer log  returned by bpf_prog_load api. The hard-coded value of 65k might not be large enough to contain all bpf program instruction. The new length is based on number of instruction times 128 bytes per instruction. 
In case the bpf program is bigger than 4096 instruction, kernel returns an error without filling up the the log buffer. A meaningful message is now returned, explaining the problem.

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -25,6 +25,7 @@
 #include <linux/rtnetlink.h>
 #include <linux/unistd.h>
 #include <linux/version.h>
+#include <linux/bpf_common.h>
 #include <net/ethernet.h>
 #include <net/if.h>
 #include <stdio.h>
@@ -33,6 +34,7 @@
 #include <sys/ioctl.h>
 #include <sys/resource.h>
 #include <unistd.h>
+#include <stdbool.h>
 
 #include "libbpf.h"
 #include "perf_reader.h"
@@ -134,14 +136,16 @@ int bpf_get_next_key(int fd, void *key, void *next_key)
 
 #define ROUND_UP(x, n) (((x) + (n) - 1u) & ~((n) - 1u))
 
-char bpf_log_buf[LOG_BUF_SIZE];
-
 int bpf_prog_load(enum bpf_prog_type prog_type,
                   const struct bpf_insn *insns, int prog_len,
                   const char *license, unsigned kern_version,
                   char *log_buf, unsigned log_buf_size)
 {
   union bpf_attr attr;
+  char *bpf_log_buffer = NULL;
+  unsigned buffer_size = 0;
+  int ret = 0;
+
   memset(&attr, 0, sizeof(attr));
   attr.prog_type = prog_type;
   attr.insns = ptr_to_u64((void *) insns);
@@ -155,7 +159,17 @@ int bpf_prog_load(enum bpf_prog_type prog_type,
   if (log_buf)
     log_buf[0] = 0;
 
-  int ret = syscall(__NR_bpf, BPF_PROG_LOAD, &attr, sizeof(attr));
+  if (attr.insn_cnt > BPF_MAXINSNS) {
+    ret = -1;
+    errno = EINVAL;
+    fprintf(stderr,
+            "bpf: %s. Program too large (%d insns), at most %d insns\n\n",
+            strerror(errno), attr.insn_cnt, BPF_MAXINSNS);
+    return ret;
+  }
+
+  ret = syscall(__NR_bpf, BPF_PROG_LOAD, &attr, sizeof(attr));
+   
   if (ret < 0 && errno == EPERM) {
     // When EPERM is returned, two reasons are possible:
     //  1. user has no permissions for bpf()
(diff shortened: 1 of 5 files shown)


drzaeus77:
add to whitelist
[to buildbot]

drzaeus77:
retest this please
[to buildbot]

src/cc/libbpf.c
```diff
@@ -177,9 +187,14 @@ int bpf_prog_load(enum bpf_prog_type prog_type,
   if (ret < 0 && !log_buf) {
     // caller did not specify log_buf but failure should be printed,
     // so call recursively and print the result to stderr
+    buffer_size = attr.insn_cnt * 128;
```
4ast: I'd rather not have such formula propagated everywhere and copy pasted by other projects. The size of verifier log_buf has no correlation to program size.

src/python/bcc/__init__.py
```diff
@@ -218,12 +218,12 @@ def load_funcs(self, prog_type=KPROBE):
...
             return self.funcs[func_name]
-
         if not lib.bpf_function_start(self.module, func_name.encode("ascii")):
             raise Exception("Unknown program %s" % func_name)
 
-        log_buf = ct.create_string_buffer(65536) if self.debug else None
-
+        prog_len = lib.bpf_function_size(self.module, func_name.encode("ascii"))
+        instruc_c = prog_len / 8 # sizeof(struct bpf_insn)
+        log_buf = ct.create_string_buffer(instruc_c * 128) if self.debug else None
```
4ast: same here. if 65k was too small, make a logic to increase try load few times until kernel stops returning enospc

src/cc/libbpf.c
davidefdl: I understand your point. If we have to try few times do you have a suggestion regarding the memory limit? Basically when shall we give up ?
Also if we need to implement this for a c api, who owns the memory in this case?

drzaeus77: > Also if we need to implement this for a c api, who owns the memory in this case?

I think it's a good time to fix up this API a little bit. Right now either the callee or bpf_prog_load itself can allocate memory, which is why the `insn_cnt * 128` got duplicated in multiple places. If instead someone needs to allocate an increasing buffer, the exponential increase logic would get duplicated as well. I would like to avoid that.

I see a couple options, would appreciate folks thoughts:

1. Always allocate memory inside bpf_prog_load, and if the caller passed in a pointer, assign the buffer to that pointer and let the caller print+free the memory, otherwise print+free locally. This has the downside that the memory allocation escapes the function, making different language bindings slightly more complex.

2. Remove the log_buf parameter, and just always print to stderr. This breaks python's ability to print a customized string, and could break other callers as well.

3. Pass in a file handle to print to instead of a buffer. This might also not work well for all language bindings.

drzaeus77: @4ast any thoughts?

4ast: I think duplicating exponential increase logic is a lesser evil than passing a pointer to allocated memory from c side into python.

davidefdl:
retest this please
[to buildbot]

src/cc/libbpf.c
```diff
@@ -175,11 +186,36 @@ int bpf_prog_load(enum bpf_prog_type prog_type,
...
+
+    buffer_size = LOG_BUF_SIZE;
     // caller did not specify log_buf but failure should be printed,
-    // so call recursively and print the result to stderr
-    bpf_prog_load(prog_type, insns, prog_len, license, kern_version,
-        bpf_log_buf, LOG_BUF_SIZE);
-    fprintf(stderr, "bpf: %s\n%s\n", strerror(errno), bpf_log_buf);
+    // so repeat the syscall and print the result to stderr
+    for (;; repeat++) {
+         buffer_size = buffer_size * repeat;
```
drzaeus77: Would do `buffer_size <<= 1` here instead.

src/cc/libbpf.c
```diff
@@ -156,6 +160,13 @@ int bpf_prog_load(enum bpf_prog_type prog_type,
     log_buf[0] = 0;
 
   int ret = syscall(__NR_bpf, BPF_PROG_LOAD, &attr, sizeof(attr));
+  if (attr.insn_cnt > BPF_MAXINSNS) {
```
drzaeus77: I like the fact that this error case is checked, but it is problematic when done after the syscall. Can you move the check to be above the syscall?

src/cc/libbpf.c
```diff
@@ -175,11 +186,36 @@ int bpf_prog_load(enum bpf_prog_type prog_type,
...
+    buffer_size = LOG_BUF_SIZE;
     // caller did not specify log_buf but failure should be printed,
-    // so call recursively and print the result to stderr
-    bpf_prog_load(prog_type, insns, prog_len, license, kern_version,
-        bpf_log_buf, LOG_BUF_SIZE);
-    fprintf(stderr, "bpf: %s\n%s\n", strerror(errno), bpf_log_buf);
+    // so repeat the syscall and print the result to stderr
+    for (;; repeat++) {
+         buffer_size = buffer_size * repeat;
+         bpf_log_buffer = (char *)malloc(buffer_size * sizeof(char));
```
4ast: please drop '* sizeof(char)' and unnecessary '(char *)' cast

src/python/bcc/__init__.py
```diff
@@ -218,29 +219,32 @@ def load_funcs(self, prog_type=KPROBE):
...
+        while True:
+            log_buf = ct.create_string_buffer(buffer_len) if self.debug else None
+            fd = lib.bpf_prog_load(prog_type,
+                    lib.bpf_function_start(self.module, func_name.encode("ascii")),
+                    lib.bpf_function_size(self.module, func_name.encode("ascii")),
+                    lib.bpf_module_license(self.module),
+                    lib.bpf_module_kern_version(self.module),
+                    log_buf, ct.sizeof(log_buf) if log_buf else 0)
+            if fd < 0 and ct.get_errno() == errno.ENOSPC and self.debug:
+                buffer_len = buffer_len * 2
```
drzaeus77: Same power of two logic here as well.

drzaeus77:
LGTM, @4ast ?

4ast:
does lua need an update or not?

drzaeus77:
Not needed, it passes nil so printing is done on C side.

PR #682, opened 2016-09-07
add CONFIG_FUNCTION_PROFILER for cachestat

jhgorse:
cachestat seems to need the CONFIG_FUNCTION_PROFILER=y kernel option.


4ast:
I don't think that's needed. CONFIG_FUNCTION_PROFILE is ftrace functionality and pretty much orthogonal to kprobe+bpf that cachestat is using.

brendangregg:
I agree with @4ast ; maybe you're thinking of the perf-tools version of cachestat? https://github.com/brendangregg/perf-tools

jhgorse:
@brendangregg, that was it. Closing. Cheers!

PR #685, opened 2016-09-08
examples: fix indentation in tracing/tcpv4connect

iaguis:
Python doesn't like mixing spaces and tabs.


brendangregg:
ok to test

brendangregg:
thanks

PR #688, opened 2016-09-10
minor cleanup + process partial name matching

KarimAllah:


brendangregg:
ok to test

brendangregg:
Thanks, looked like some of that stuff (delta) was copied from another tool, and was redundant.

The -n option needs to be added to the man page, and the USAGE example at the end of opensnoop_example.txt updated.

Does it need to do regular expressions? Can -n just do a partial string match?

Here's why I ask: -n will probably make sense for other tools. Some will have a higher frequency rate than open()s, and so, we'll want to do the string match in kernel context (I'd accept the user space match in opensnoop, only because open()s are low frequency). Doing a process name string match in other tools, in kernel context, might require some BPF enhancements first, but when they arrive, I doubt we'll have regex's in-kernel. We'll probably have strstr() and friends. So, I'm not sure I want to set a precedent of -n means regex name search. If -n means just string match, then we'll be ok when we add it to other tools that implement it in kernel.

KarimAllah:
>> Does it need to do regular expressions? Can -n just do a partial string match?

I only used regex because the only other tool that have "-n" on the tools directory is "execsnoop.py" and it was using regex already so I was trying to be coherent.

If you prefer a partial string match, I can do that instead and also switch the one in "execsnoop.py" to that.

>> Doing a process name string match in other tools, in kernel context, might require some BPF enhancements first ..

I'm not very familiar with BPF internals, but which enhancements are you talking about here ? is it something already in lkml ?

brendangregg:
Ok, right, yes, it might be safer to switch execsnoop to a simple partial string match, although you don't need to do that right now. I wonder if anyone is depending on it being REs. Its predecessor in perf-tools was REs.

So, some tools use __builtin_memcpy() (from the compiler), and I think some builtin str functions might work as well. But I think the issue is that these can become too many instructions, and can hit the 4k BPF limit. There's been the idea of adding some BPF string functions, but we don't have a patchset on lkml yet. When we do, I'm guessing we won't have regex's in the kernel. :)

KarimAllah:
>> Ok, right, yes, it might be safer to switch execsnoop to a simple partial string match, although you don't need to do that right now. I wonder if anyone is depending on it being REs. Its predecessor in perf-tools was REs.

Sure, I updated 3/3 to only do partial matches instead of regex (+the man page and example). I also updated the commit subject in 1/3 (s/diff/delta)

brendangregg:
tested, works. Just some style nits:

```
# pep8 --show-source --ignore=E123,E125,E126,E127,E128,E302 opensnoop.py 
opensnoop.py:161:22: E713 test for membership should be 'not in'
    if args.name and not args.name in event.comm:
                     ^
opensnoop.py:162:1: E101 indentation contains mixed spaces and tabs
	return
^
opensnoop.py:162:1: W191 indentation contains tabs
	return
^
opensnoop.py:164:1: E101 indentation contains mixed spaces and tabs
    if args.timestamp:
^
```

KarimAllah:
>> tested, works. Just some style nits:

Fixed

brendangregg:
thanks!

PR #689, opened 2016-09-12
[tcpconnect] filter traced connection based on destination ports

### LINKED ISSUE #681: tcpconnect: filter based on destination port
Currently one can specify a process id to monitor for tcpconnects. Would there be any value doing so for a specific port. The specific case this is trying to solve is "tell me what is initiating connection to port X".

Currently, this can be achieve with a bit of `grep`ping, but it may be convenient to not track every connections within `trace_connect_entry` if we don't care about them in the first place.

I would assume we can access the destination port through `struct sock *sk` in  `trace_connect_entry` and return early if this is not matching the port of interest.

chantra:
Test:
While running:
while [ 1 ]; do nc -w 1 100.127.0.1 80; nc -w 1 100.127.0.1 81; done

root@vagrant:/mnt/bcc# ./tools/tcpconnect.py
PID    COMM         IP SADDR            DADDR            DPORT
19978  nc           4  10.0.2.15        100.127.0.1      80
19979  nc           4  10.0.2.15        100.127.0.1      81
19980  nc           4  10.0.2.15        100.127.0.1      80
19981  nc           4  10.0.2.15        100.127.0.1      81
root@vagrant:/mnt/bcc# ./tools/tcpconnect.py  -P 80
PID    COMM         IP SADDR            DADDR            DPORT
19987  nc           4  10.0.2.15        100.127.0.1      80
19989  nc           4  10.0.2.15        100.127.0.1      80
19991  nc           4  10.0.2.15        100.127.0.1      80
19993  nc           4  10.0.2.15        100.127.0.1      80
19995  nc           4  10.0.2.15        100.127.0.1      80
root@vagrant:/mnt/bcc# ./tools/tcpconnect.py  -P 80 81
PID    COMM         IP SADDR            DADDR            DPORT
8725   nc           4  10.0.2.15        100.127.0.1      80
8726   nc           4  10.0.2.15        100.127.0.1      81
8727   nc           4  10.0.2.15        100.127.0.1      80
8728   nc           4  10.0.2.15        100.127.0.1      81
8729   nc           4  10.0.2.15        100.127.0.1      80

Fixes #681

### DIFF
--- a/tools/tcpconnect.py
+++ b/tools/tcpconnect.py
@@ -4,7 +4,7 @@
 # tcpconnect    Trace TCP connect()s.
 #               For Linux, uses BCC, eBPF. Embedded C.
 #
-# USAGE: tcpconnect [-h] [-t] [-p PID]
+# USAGE: tcpconnect [-h] [-t] [-p PID] [-P PORT [PORT ...]]
 #
 # All connection attempts are traced, even if they ultimately fail.
 #
@@ -20,7 +20,7 @@
 from __future__ import print_function
 from bcc import BPF
 import argparse
-from socket import inet_ntop, AF_INET, AF_INET6
+from socket import inet_ntop, ntohs, AF_INET, AF_INET6
 from struct import pack
 import ctypes as ct
 
@@ -29,6 +29,8 @@ examples = """examples:
     ./tcpconnect           # trace all TCP connect()s
     ./tcpconnect -t        # include timestamps
     ./tcpconnect -p 181    # only trace PID 181
+    ./tcpconnect -P 80     # only trace port 80
+    ./tcpconnect -P 80,81  # only trace port 80 and 81
 """
 parser = argparse.ArgumentParser(
     description="Trace TCP connects",
@@ -38,6 +40,8 @@ parser.add_argument("-t", "--timestamp", action="store_true",
     help="include timestamp on output")
 parser.add_argument("-p", "--pid",
     help="trace this PID only")
+parser.add_argument("-P", "--port",
+    help="comma-separated list of destination ports to trace.")
 args = parser.parse_args()
 debug = 0
 
@@ -76,7 +80,7 @@ BPF_PERF_OUTPUT(ipv6_events);
 int trace_connect_entry(struct pt_regs *ctx, struct sock *sk)
 {
     u32 pid = bpf_get_current_pid_tgid();
-    FILTER
+    FILTER_PID
 
     // stash the sock ptr for lookup on return
     currsock.update(&pid, &sk);
@@ -106,6 +110,8 @@ static int trace_connect_return(struct pt_regs *ctx, short ipver)
     struct sock *skp = *skpp;
     u16 dport = 0;
     bpf_probe_read(&dport, sizeof(dport), &skp->__sk_common.skc_dport);
+
+    FILTER_PORT
 
     if (ipver == 4) {
         struct ipv4_data_t data4 = {.pid = pid, .ip = ipver};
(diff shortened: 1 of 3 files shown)


brendangregg:
Thanks. Need to update the man page, and (at least) the USAGE message at the end of the _example.txt file.

chantra:
@brendangregg doc updated. the build failure seems unrelated.

tools/tcpconnect.py
```diff
@@ -107,6 +111,8 @@
     u16 dport = 0;
     bpf_probe_read(&dport, sizeof(dport), &skp->__sk_common.skc_dport);
 
+    FILTER_PORT
```
chantra: @brendangregg I have tried to put this in `trace_connect_entry` but did not manage do it there. I can try to dig deeper if needed.

brendangregg:
retest this please

chantra:
> retest this please

What do you mean? I did re-run the tests.

drzaeus77:
This is a message to the buildbot, we use
https://wiki.jenkins-ci.org/display/JENKINS/GitHub+pull+request+builder+plugin
in the backend. No cause for concern to you :)

brendangregg:
@drzaeus77 is there a way we can prefix the buildbot messages with "bot:" or something, to not confuse people? :)

drzaeus77:
@brendangregg Yes, I have updated the buildbot regex to the following:
```
.*\[buildbot,\W+add\W+to\W+whitelist\].*
.*\[buildbot,\W+ok\W+to\W+test\].*
.*\[buildbot,\W+test\W+this\W+please\].*
```

As an example, since this commit has run up against the kernel bug that requires iovisor-buildbot-slave05 to be rebooted (which I have done), I will ask [buildbot, test this please].

Patch has merged into stable 4.4 queue as of this morning, so hopefully soon we can update slave05 to pick up the change.

http://www.gossamer-threads.com/lists/linux/kernel/2523907

tools/tcpconnect.py
brendangregg: If the bpf_probe_read() really didn't pull a valid port on the entry, then maybe struct sock isn't fully initialized yet. But I think it's ok as it is for now (if it were a little higher frequency, I'd be more interested in making the earlier filter work).

tools/tcpconnect_example.txt
```diff
@@ -41,16 +41,20 @@ process to various other addresses. A few connections occur every minute.
...
   -p PID, --pid PID  trace this PID only
+  -P PORT [PORT ...], --port PORT [PORT ...]
+                     trace these ports only
 
 examples:
     ./tcpconnect           # trace all TCP connect()s
     ./tcpconnect -t        # include timestamps
     ./tcpconnect -p 181    # only trace PID 181
+    ./tcpconnect -P 80     # only trace port 80
+    ./tcpconnect -P 80 81  # only trace port 80 and 81
```
brendangregg: what other tool does this? has an unquoted list as an optional argument. Is it hard to make this either:

- -P "80 81"
- -P 80,81

which I think better matches expectations? (1st is from ps(1)'s -p behavior). Plus will work when copying to other tools, where we have positional parameters.

tools/tcpconnect.py
chantra: yeah, struct sock is not fully initialized on entry.

tools/tcpconnect_example.txt
chantra: This is the default behaviour of argparse with `nargs='+'` from line 43 and comes free with it: https://docs.python.org/3/library/argparse.html#nargs 
I can make -P an optional string parameter and handle it by splitting around space of parameter. Whichever is preferred.

brendangregg:
I'd go with -P "80 81" style. "-P 80 81" would be the first and only tool I know with that behavior, plus I think it breaks nargs="?". I'll have to keep an eye out for it if argparse encourages it!

I was fooling with it and added interval and count as nargs="?":

```
usage: tcpconnect.py [-h] [-t] [-p PID] [-P PORT [PORT ...]]
                     [interval] [count]
```

But it pulls them in as ports.

interval & count doesn't make sense for tcpconnect.py, but may do for other tools that borrow the -P code.

chantra:
> interval & count doesn't make sense for tcpconnect.py, but may do for other tools that borrow the -P code.

Actually, I wanted to add a `-c` for count in a separate diff :) The reasoning is that I have been trying to find which shortlive process was connecting to port X. I could:
```
cat /proc/$(tcpconnect -P 1234 -c 1 | sed 1d | awk '{print $1}')/cmdline
```

With a continous trace, this is harder to get done.

chantra:
I went with `-P 80,81` and realize that you wanted `-P "80 81"`. Do you have strong opinion about comma-separated vs space-separated?

brendangregg:
no strong opinion

brendangregg:
buildbot, ok to test

brendangregg:
tested, works, thanks!

Issue #588, opened 2016-06-29
tool feature request: offcputime min and max thresholds


brendangregg:
offcputime is a great tool for analyzing all blocking paths, but the output can be polluted with less interesting short and long blocking events. Eg, for short events: minor lock contention. Eg, for long events: idle threads waiting for work.

Min and max thresholds could be added to allow the user to dial up the latency range they are interested in. Eg,

```
./offcputime -m 4 -M 400
```

To only include events between 4 and 400 milliseconds.

The non-folded output should begin with an informational line about the filters:

```
Only showing off-CPU times between 4 and 400 milliseconds.
```

You could also imagine using this when a specific latency problem has been noticed in the application, eg, 90ms, and then dialing up a tight range to see if it's a simple single blocking call.

PR #696, opened 2016-09-20
[offcputime] allow for the speficication of a threshold for collection

glommer:
There are situations, specially when using non-folded mode, where we are
interested only in very high latencies that happen due to blocking.

While we can certainly filter out the very small ones out of the output, it is a
lot more convenient to do this from the tool itself, as it would be difficult
from an external filter to do this in one pass.

But if we are to discard unused measurements, we can do even better: we can
change the bpf code itself not to grab those traces, and gain a bit of
efficiency in scenarios in which we are only concerned about peak latencies.

### DIFF
--- a/tools/offcputime.py
+++ b/tools/offcputime.py
@@ -78,6 +78,9 @@ parser.add_argument("--stack-storage-size", default=1024,
 parser.add_argument("duration", nargs="?", default=99999999,
     type=positive_nonzero_int,
     help="duration of trace, in seconds")
+parser.add_argument("--min-block-time", default=1,
+    type=positive_nonzero_int,
+    help="the amount of time in microseconds over which we store traces (default 1)")
 args = parser.parse_args()
 if args.pid and args.tgid:
     parser.error("specify only one of -p and -t")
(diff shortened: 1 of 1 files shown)


brendangregg:
yes, did you see #588 ?

glommer:
No, sorry @brendangregg !

Had I seen this I would have easily implemented a max boundary as well. Since all I needed was the min threshold, that's what I've done. Would you like me to handle max boundaries as well and respin?

brendangregg:
If you don't mind, please do. Will be very useful to have these options.

glommer:
Sorry @brendangregg , I suck at githubbing.

Should I close this pull request and send a new one for a new branch, or is there a way to  update this?

brendangregg:
I'm probably not much better.

One way (which removes history btw) is to do your changes, then git add, git commit, and then "git rebase -i HEAD~2", and edit the mesage to "pick" the earlier commit message, and "squash" the newer ones. Save, and then a "git push --force" should refresh this PR.

glommer:
uh oh, git push --force is contrarian to thousands of years of evolutionary instinct. I have pushed my updated code to a new branch. Maybe I can just create a new PR and reference this one from there?

I wish we could just have an easy way to get something similar to "PATCH v2" here.

PR #697, opened 2016-09-20
[offcputime] allow for the speficication of a threshold for collection

glommer:
There are situations, specially when using non-folded mode, where we are
interested only in very high latencies that happen due to blocking.

While we can certainly filter out the very small ones out of the output, it is a
lot more convenient to do this from the tool itself, as it would be difficult
from an external filter to do this in one pass.

But if we are to discard unused measurements, we can do even better: we can
change the bpf code itself not to grab those traces, and gain a bit of
efficiency in scenarios in which we are only concerned about peak latencies.

This scheme can be easily extended to also allow a maximum cap in the latencies
we are interested in.

After this patch is applied, the options -m and -M can be used to set those
limits respectively.

Fixes: #588

### DIFF
--- a/tools/offcputime.py
+++ b/tools/offcputime.py
@@ -40,6 +40,8 @@ examples = """examples:
     ./offcputime             # trace off-CPU stack time until Ctrl-C
     ./offcputime 5           # trace for 5 seconds only
     ./offcputime -f 5        # 5 seconds, and output in folded format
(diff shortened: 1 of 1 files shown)


glommer:
Previous discussion for v1 at #696

brendangregg:
buildbot, ok to test

brendangregg:
Tested, works.

I hit #704 , but I think that's a separate bug.

PR #698, opened 2016-09-22
Fix argdist, trace, tplist to use the libbcc USDT support

goldshtn:
UPDATED: The task list below is pretty much done, and I think that's the pieces we need to update the three tools to the new USDT support. Any comments and feedback will be appreciated. From my side, it's ready to merge.

So, here's the task list:

- [x] Fix argdist so that it fully supports USDT probes, including when no pid is specified
- [x] Add API to libbcc that returns a list of the USDT probes in a process or binary
- [x] Fix tplist to print USDT probe information on processes or binaries
- [x] Fix trace so that it doesn't bail for kprobes, uprobes, and tracepoints (not including USDT)
- [x] Migrate trace to the new libbcc USDT support
- [x] Modify `BPF` Python class to allow multiple USDT contexts
- [x] Fix trace to support multiple USDT probes
- [x] Fix argdist to support multiple USDT probes

### DIFF
--- a/src/python/bcc/usdt.py
+++ b/src/python/bcc/usdt.py
@@ -48,7 +48,9 @@ class USDT(object):
 
     def enable_probe(self, probe, fn_name):
         if lib.bcc_usdt_enable_probe(self.context, probe, fn_name) != 0:
-            raise Exception("failed to enable probe '%s'" % probe)
+            raise Exception(("failed to enable probe '%s'; a possible cause " +
+                            "can be that the probe requires a pid to enable") %
+                            probe)
 
     def get_text(self):
         return lib.bcc_usdt_genargs(self.context)
--- a/src/python/bcc/usdt.py
+++ b/src/python/bcc/usdt.py
@@ -12,7 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from .libbcc import lib, _USDT_CB
+from .libbcc import lib, _USDT_CB, _USDT_PROBE_CB
+
+class USDTProbe(object):
+    def __init__(self, usdt):
+        self.provider = usdt.provider
+        self.name = usdt.name
+        self.bin_path = usdt.bin_path
+        self.semaphore = usdt.semaphore
+        self.num_locations = usdt.num_locations
+        self.num_arguments = usdt.num_arguments
+
+    def __str__(self):
+        return "%s %s:%s [sema 0x%x]\n  %d location(s)\n  %d argument(s)" % \
+               (self.bin_path, self.provider, self.name, self.semaphore,
+                self.num_locations, self.num_arguments)
 
 class USDT(object):
     def __init__(self, pid=None, path=None):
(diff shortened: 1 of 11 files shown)


4ast:
lgtm.
btw the latest python 3.6 beta includes usdt probes:
https://docs.python.org/3.6/whatsnew/3.6.html
could you please check that bcc usdt parser understand them as well?
Thanks!
cc @ambv @brendangregg

goldshtn:
@4ast Just for reference, [this](https://docs.python.org/3.6/whatsnew/3.6.html#dtrace-and-systemtap-probing-support) looks like the specific changelog for the `--with-dtrace` compilation switch (same as Node), and then [these](https://docs.python.org/3.6/howto/instrumentation.html#instrumentation) are the instructions for building and then using these probes. I'm looking into it right now.

goldshtn:
@4ast Yep, it works with Python-3.6.0b1!

First shell:

```
$ ./python
Python 3.6.0b1 (default, Sep 22 2016, 15:40:09)
[GCC 5.3.1 20160406 (Red Hat 5.3.1-6)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> def factorial(n):
...     if n == 1: return 1
...     return factorial(n-1) * n
...
```

Second shell:

```
# ./trace.py 'u:/home/vagrant/Python-3.6.0b1/python:function__entry "%s", arg2' -p $(pidof python)
```

Back in first shell:

```
>>> factorial(5)
120
```

Back in second shell:

```
15:45:34 29354  python       function__entry  factorial
15:45:34 29354  python       function__entry  factorial
15:45:34 29354  python       function__entry  factorial
15:45:34 29354  python       function__entry  factorial
15:45:34 29354  python       function__entry  factorial
```

4ast:
@goldshtn nice nice nice!
cc @ambv
please indicate when this PR can be merged, so we can give a try as well
fedora error "perf_event_open(/sys/kernel/debug/tracing/events/uprobes/p__lib64_libc_so_6_0x86b40/id): Cannot allocate memory"
is unrelated. It's known kernel bug that was fixed in the later versions.
Workaround is to reboot the box, but I don't have access to the buildbot while @drzaeus77 is on pto.
@dark do you have the access by any chance?

goldshtn:
@4ast it's good to go from my end.

tools/tplist_example.txt
```diff
@@ -17,25 +17,18 @@ $ tplist -l basic_usdt
...
 
 $ tplist '*loop_iter' -l basic_usdt -v
 /home/vagrant/basic_usdt basic_usdt:loop_iter [sema 0x601036]
-  location 0x400550 raw args: -4@$42 8@%rax
-    4   signed bytes @ constant 42
-    8 unsigned bytes @ register %rax
-  location 0x40056f raw args: 8@-8(%rbp) 8@%rax
-    8 unsigned bytes @ -8(%rbp)
-    8 unsigned bytes @ register %rax
+  2 location(s)
```
4ast: btw, we're rewriting sdt.h and getting rid of all the craft in there. Part of it is removing '-' form size, since it's of dubious value. The feature request for the future patch is to add 'very verbose' output that can print raw args for debugging.

src/python/bcc/usdt.py
```diff
@@ -12,34 +12,67 @@
...
-                raise Exception("USDT failed to instrument PID %d" % pid) 
+                raise Exception("USDT failed to instrument PID %d" % pid)
         elif path:
             self.path = path
             self.context = lib.bcc_usdt_new_frompath(path)
             if self.context == None:
-                raise Exception("USDT failed to instrument path %s" % path) 
+                raise Exception("USDT failed to instrument path %s" % path)
+        else:
+            raise Exception("either a pid or a binary path must be specified")
```
4ast: nice hints!

brendangregg:
I have access to the build bot server & slaves. I'm not sure how to fix it yet though... Checking some logs.

PR #699, opened 2016-09-23
build: use old libstdc++ ABI for GCC 5.1 and later

htbegin:
According to https://gcc.gnu.org/gcc-5/changes.html#libstdcxx,
GCC 5.1 and later introduce a dual ABI for C++ runtime library,
and the new ABI is enabled by default.

On Fedora 23 the version of gcc is 5.3.1, and the build of libbcc.so
and tests would fail with the following errors:

......
../../src/cc/libbcc.a(bpf_module.cc.o): In function `ebpf::BPFModule::dump_ir(llvm::Module&)':
bpf_module.cc:(.text+0x44c0): undefined reference to `llvm::createPrintModulePass(llvm::raw_ostream&,
	std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool)'
../../src/cc/libbcc.a(bpf_module.cc.o): In function `llvm::RTDyldMemoryManager::findSymbol(
	std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
bpf_module.cc:(.text._ZN4llvm19RTDyldMemoryManager10findSymbolERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE[
	_ZN4llvm19RTDyldMemoryManager10findSymbolERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE]+0x18):
undefined reference to `llvm::RTDyldMemoryManager::getSymbolAddressInProcess(
	std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'
......

Fix it by adding -D_GLIBCXX_USE_CXX11_ABI=0 to CMAKE_CXX_FLAGS to
restore to the old ABI.


4ast:
Interesting. Thanks for the fix.
@markdrayton I thought we build it with gcc 5 as well and don't see this issue ?

markdrayton:
Yes, we build with GCC 5.3 but I don't recall seeing these errors.

dark:
@htbegin Thanks for the patch.
That error suggests that the version of libLLVM might have been compiled with an older version of GCC.
Could you please detail how was LLVM installed? From a package, or did you compile it yourself?
You can also confirm the hypotesis by running (pls fix the path if cmake outputs a different one while configuring):
`readelf -Ws /usr/lib/llvm-3.7/lib/libLLVM-3.7.1.so.1  | grep createPrintModulePass`
In my case (Ubuntu 16.04.1 LTS, LLVM installed from upstream packages), I see a symbol name that uses the new __cxx11 ABI, so it confirms that LLVM was built with a new gcc:
>   2085: 000000000063a600   100 FUNC    GLOBAL DEFAULT   12 _ZN4llvm21createPrintModulePassERNS_11raw_ostreamERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEb

htbegin:
Hi, @dark. You are right.

The LLVM libraries I used were build with the old ABI and I downloaded them from LLVM download page.
The version of LLVM libraries is 3.7.1, and the following is the output of readelf cmd:
# readelf -Ws *.a | grep "FUNC.*createPrintModulePass" | c++filt  
......llvm::createPrintModulePass(llvm::raw_ostream&, std::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool)

I tried both LLVM 3.9.0 from LLVM download page and the llvm-static package from Fedora 23.
New ABI is enabled for both of them. The build succeded for LLVM 3.9.0, but the build
failed when using llvm-static.

I will close the pull request. Thanks.

PR #677, opened 2016-08-27
fileslower: use de->d_name.name, add filtering

markdrayton:
I think this PR goes some way towards solving #609 by reading `de->d_name.name` instead of `de->d_iname`.

In my experience the current fileslower.py prints a lot of lines for the file `TCPv6`. I dug into this and many of these reads are actually for regular disk files, not TCP sockets. I'm not quite sure what's going on but my guess is that TCP dentries get reused for regular file dentries and the short name (`d_iname`) isn't destroyed, leading to a misleading attribution in fileslower.py. This PR switches to reading `de->d_name.name`, which AFAICT is the source of truth (it either points to external storage for long file names or `d_iname` for short names; http://lxr.free-electrons.com/source/fs/dcache.c#L1638).

Other fixes:

* by default, only show regular files. Add a `-a` option to show all files (sockets, FIFOs, etc)
* make `-p` match process IDs instead of thread IDs
* remove unused signal handling code
* indicate when a name was longer than the 32 byte buffer by replacing the last 3 chars with `...` (fits within 80 chars assuming `comm` and `bytes` don't overflow)

E.g.

```
$ ./tools/fileslower.py 
Tracing sync read/writes slower than 1 ms
TIME(s)  COMM           TID    D BYTES   LAT(ms) FILENAME
1.790    dd             3843   W 8192       1.33 paddingpaddingpaddingpadding...
```

Issue #609, opened 2016-07-12
fileslower and filetop report TCP instead of file name


bobrik:
```
root@myhost:/# /usr/share/bcc/tools/fileslower 1
Tracing sync read/writes slower than 1 ms
TIME(s)  COMM           PID    D BYTES   LAT(ms) FILENAME
0.000    foobarba       25738  R 5         90.12 TCP
0.010    foobarba       25018  R 5         97.56 TCP
0.020    foobarba       24538  R 5         92.43 TCP
```

I'm running it on 4.4.10.

brendangregg:
This is because it's tracing at the VFS layer, which catches file read/writes and other read/write types.

This could either be filtered, or, documented. Perhaps the most intuitive option would be to filter by default, unless a verbose option is passed at the command line.

markdrayton:
I think this could be a consequence of using `d_iname` for the name lookup in `fileslower` rather than/as well as catching non file IO.

A few days ago I was investigating a writeback bug that slowed down `write` syscalls. `fileslower` showed slow writes for lots of files but it **didn't** show slow writes to the application log files. This seemed really weird as off-CPU profiling showed the app was clearly blocking on writing its log. Today I dug into what was going on.

I wrote a basic script to trace `write`s to a given FD for a given process. For these writes I printed `de->d_iname` as well as `de->d_name.name` ([full code](https://gist.github.com/markdrayton/f57f7124e36ff8e2966235bef63a31be)).

Here's the workload generator (write.py):

```python
import tempfile
import time

with tempfile.NamedTemporaryFile(prefix='/tmp/paddingpaddingpaddingpadding') as f:
    while True:
        f.write('test\n')
        f.flush()
        time.sleep(1)
```

```
# python write.py &
[1] 1122305

# python print.py 1122305 3
fd: 3, count: 5, iname: TCPv6, name: paddingpaddingpaddingpaddingrGCX
fd: 3, count: 5, iname: TCPv6, name: paddingpaddingpaddingpaddingrGCX
fd: 3, count: 5, iname: TCPv6, name: paddingpaddingpaddingpaddingrGCX
```

The string in `d_iname` varies. On our production hosts it's usually TCPv6; on my little-used Ubuntu VM it's the empty string. I'm not really sure why it ends up as this other value — perhaps dentry swapping/copying (see `fs/dcache.c`)? It seems the right thing to do is use `de->d_name.name`, since as far as I could tell this is the authoritative source.

[...]

PR #700, opened 2016-09-23
filelife: use d_name.name and d_name.len instead of d_iname

dark:
This is similar in spirit to what was done in PR #677 to fix the problem reported in #609.

filelife.py is now converted to use the right struct field.

### DIFF
--- a/tools/filelife.py
+++ b/tools/filelife.py
@@ -82,13 +82,14 @@ int trace_unlink(struct pt_regs *ctx, struct inode *dir, struct dentry *dentry)
     delta = (bpf_ktime_get_ns() - *tsp) / 1000000;
     birth.delete(&dentry);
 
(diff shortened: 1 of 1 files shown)


brendangregg:
```
# pep8 --show-source --ignore=E123,E125,E126,E127,E128,E302 ~/filelife.py 
/root/filelife.py:91:80: E501 line too long (85 > 79 characters)
        bpf_probe_read(&data.fname, sizeof(data.fname), (void*) dentry->d_name.name);
```

plus "void*" should have a space in there.

I think it otherwise works, although vfs_create() isn't being called on my 4.8-rc4 kernel. I'll create a separate ticket.

dark:
Thanks, will fix.

brendangregg:
Thanks, LGTM.

PR #707, opened 2016-09-27
revert "Flag ${LLVM_INCLUDE_DIRS} as a system include directory"

4ast:
it breaks the build as:
"
c++: fatal error: no input files
compilation terminated.
/bin/sh: /w/llvm/bld/include: Is a directory
"

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,7 +66,7 @@ if(NOT DEFINED BCC_KERNEL_MODULES_SUFFIX)
 endif()
 
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -isystem ${LLVM_INCLUDE_DIRS}")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall")
 endif()
 
 add_subdirectory(examples)


4ast:
cc @dark

dark:
Ouch.

@4ast, do you have more info about the failure? Distro/packages versions/etc?
Or is it one of the bots?

4ast:
that was my normal build on vanilla centos7 with gcc 6.1
I'm guessing it has a different way of handling -isystem

dark:
I will spin a VM and try and reproduce the issue.

dark:
@4ast 
I have tried to reproduce the issue, but unfortunately I couldn't.
I have spawned a base centos7 VM, and installed gcc `6.1.1-2.fc24` from `http://mirrors.fedoraproject.org/mirrorlist?repo=fedora-24&arch=$basearch` and LLVM from `http://llvm.org/releases/3.7.1/clang+llvm-3.7.1-x86_64-fedora22.tar.xz`. I had to hack something in CMake because I could not upgrade to `2.8.12`, and I was stuck with `2.8.11`. The whole `src` directory built fine, though.

Looking again at the error message, it looks like a parameter might be missing from the commandline (maybe it's empty?), rather than isystem behaving differently. Have you tried deleting the whole `build` directory and starting from a clean one?

Since this is not a critical commit, if we cannot figure out what's going on a revert is likely the fastest solution.

4ast:
The problem is that LLVM_INCLUDE_DIRS is not a single dir.
On my machine it's /w/llvm/include;/w/llvm/bld/include
If I do:
-isystem '${LLVM_INCLUDE_DIRS}'
it compiles, but I don't think gcc is doing the right thing.
Thoughts?

dark:
Aaah, I suspect CMake does some list-expansion when it encounters semicolons in a string. Let me do some experiments.

dark:
I was wrong. The semicolon goes all the way to the shell, and it's interpreted as a command separator. Hence, we have:

> c++ .... -isystem /w/llvm/include;/w/llvm/bld/include ... [the rest of the command]
 
So we need to parse LLVM_INCLUDE_DIRS and expand it into:

> -isystem /w/llvm/include -isystem /w/llvm/bld/include  

I will try and write a patch.

dark:
Please see #709 for a fix.

PR #709, opened 2016-09-28
cmake: fix build when LLVM_INCLUDE_DIRS includes multiple directories

dark:
If LLVM_INCLUDE_DIRS includes multiple directories, separated by
semicolon, the string would be incorrectly propagated all the way down
to the shell, that would interpret such semicolon as a command
separator. E.g. we would have:

 c++ ... -isystem /w/llvm/include;/w/llvm/bld/include ...

Instead, we need to parse the string as a CMake list (that are defined
as strings composed by semicolon-separated tokens) and build a string
in the form:

 c++ ... -isystem /w/llvm/include -isystem /w/llvm/bld/include ...

This bug was introduced in d19e0cb.
This commit fixes #707.

Signed-off-by: Marco Leogrande <marcol@plumgrid.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,7 +66,12 @@ if(NOT DEFINED BCC_KERNEL_MODULES_SUFFIX)
 endif()
 
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -isystem ${LLVM_INCLUDE_DIRS}")
+# iterate over all available directories in LLVM_INCLUDE_DIRS to
+# generate a correctly tokenized list of parameters
+foreach(ONE_LLVM_INCLUDE_DIR ${LLVM_INCLUDE_DIRS})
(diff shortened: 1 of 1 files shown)


4ast:
builds, though I still see a ton of warnings:
/w/llvm/include/../tools/clang/include/clang/AST/ASTContext.h:529:13: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]

dark:
@4ast 
That directory structure looks a bit odd, maybe you installed manually?

`/w/llvm/include/../tools` is neither under `/w/llvm/include`, nor under `/w/llvm/bld/include`, so `-isystem` will not catch it.

For me, `ASTContext.h` is at `/usr/lib/llvm-3.7/include/clang/AST/ASTContext.h` (and LLVM_INCLUDE_DIRS is `/usr/lib/llvm-3.7/include`).

4ast:
yes. it's not installed. I'm pointing LLVM_DIR to llvm build directory to avoid copying Gbytes of libraries and binaries.
I think it should be good for everyone, but llvm developers. Thanks!

PR #710, opened 2016-09-28
Fix trace.py USDT argument filtering

palmtenor:
@goldshtn fixed general USDT support in `trace.py` in #698 . However, filtering is still not able to function since we are not reading the arguments correctly.

This Diffs replaces `argX` variables in filtering conditions to `argX_filter`, and use the `bpf_readarg_N` macros to read argument values into `argX_filter` variables. 

In order to define the variables correctly, this Diff also exposes the ability to get USDT argument type (i.e., size) in Python `USDT` class. I think this would be useful elsewhere as well.

Tested by defining an USDT probe with two parameters in my test binary and run
```
trace.py 'u:pathToBinary:theProbeName (arg1!=0) "Arg1 %lld Arg2 %s", arg1, arg2'
```
Generated BPF program looks like:
```
int probe_contextSwitch_1(struct pt_regs *ctx)
{

        u32 __pid = bpf_get_current_pid_tgid();
        if (__pid == 3787616) { return 0; }

        uint64_t arg1_filter;
        bpf_usdt_readarg(1, ctx, &arg1_filter);

        if (!((arg1_filter!=0))) return 0;

        struct probe_contextSwitch_1_data_t __data = {0};
        __data.timestamp_ns = bpf_ktime_get_ns();
        __data.pid = bpf_get_current_pid_tgid();
        bpf_get_current_comm(&__data.comm, sizeof(__data.comm));
        u64 arg1;
        bpf_usdt_readarg(1, ctx, &arg1);
        __data.v0 = (long long)arg1;
        u64 arg2;
        bpf_usdt_readarg(2, ctx, &arg2);

        if (arg2 != 0) {
                bpf_probe_read(&__data.v1, sizeof(__data.v1), (void *)arg2);
        }

        probe_contextSwitch_1_events.perf_submit(ctx, &__data, sizeof(__data));
        return 0;
}
```

### DIFF
--- a/tools/trace.py
+++ b/tools/trace.py
@@ -299,6 +299,26 @@ BPF_PERF_OUTPUT(%s);
                                         (idx, Probe.c_type[field_type], expr)
                 self._bail("unrecognized field type %s" % field_type)
 
+        def _generate_usdt_filter_read(self):
+            text = ""
+            if self.probe_type == "u":
+                    for arg, _ in Probe.aliases.items():
+                        if not (arg.startswith("arg") and (arg in self.filter)):
+                                continue
+                        arg_index = int(arg.replace("arg", ""))
+                        arg_ctype = self.usdt.get_probe_arg_ctype(
+                                self.usdt_name, arg_index)
+                        if not arg_ctype:
+                                self._bail("Unable to determine type of {} "
+                                           "in the filter".format(arg))
+                        text += """
+        {} {}_filter;
+        bpf_usdt_readarg({}, ctx, &{}_filter);
+                        """.format(arg_ctype, arg, arg_index, arg)
+                        self.filter = self.filter.replace(
+                                arg, "{}_filter".format(arg))
+            return text
+
         def generate_program(self, include_self):
                 data_decl = self._generate_data_decl()
                 # kprobes don't have built-in pid filters, so we have to add
@@ -331,6 +351,7 @@ int %s(%s)
 {
         %s
         %s
+        %s
         if (!(%s)) return 0;
 
         struct %s __data = {0};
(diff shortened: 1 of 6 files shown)


4ast:
Teng, Thanks for the fix!
@goldshtn please review.

tools/trace.py
```diff
@@ -299,6 +299,26 @@ def _generate_field_assign(self, idx):
                                         (idx, Probe.c_type[field_type], expr)
                 self._bail("unrecognized field type %s" % field_type)
 
+        def _generate_usdt_filter_read(self):
+            text = ""
+            if self.probe_type == "u":
+                    for arg, _ in Probe.aliases.items():
```
goldshtn: Why not just check `arg[0:3] == "arg"` as we do elsewhere? It's just that depending on `Probe.aliases` is a bit fragile, because the arg aliases are not really used for USDT probes and might change or move at some point.

tools/trace.py
```diff
@@ -299,6 +299,26 @@ def _generate_field_assign(self, idx):
...
+                        if not (arg.startswith("arg") and (arg in self.filter)):
+                                continue
+                        arg_index = int(arg.replace("arg", ""))
+                        arg_ctype = self.usdt.get_probe_arg_ctype(
+                                self.usdt_name, arg_index)
+                        if not arg_ctype:
+                                self._bail("Unable to determine type {} "
```
goldshtn: Maybe `"Unable to determine type of {} "` to be a bit clearer. (Add "of".)

tools/trace.py
palmtenor: Ideally we should do this. The issue here is for filters, unlike the return (printf) values, we don't actually parse them. We just take what's in between brackets in the CLI argument and put it in the `if` statement directly, with some direct find-and-replace on variable names (as you can see here). We would have to parse the filter expression in order to get type and expr as we do for printf values:(

tools/trace.py
palmtenor: Done:)

tools/trace.py
goldshtn: Right, forgot about that. The filter is indeed quite unstructured. Good to go then.

Issue #591, opened 2016-06-30
Make trace and argdist use the new tracepoint support


goldshtn:
Currently, there's the ugly `perf_trace_nnn` hack used by these tools, and there's a whole `Tracepoint` class in bcc that generates a bunch of boilerplate. Replace that with the new tracepoint support, and move the old versions into **old/**.

goldshtn:
Good to close when #724 is merged.

goldshtn:
Closing -- completed in #724.

PR #724, opened 2016-10-04
argdist, trace: Native tracepoint support

goldshtn:
This PR modifies argdist and trace to use the `TRACEPOINT_PROBE` infrastructure rather than the hackish kprobe-based perf functions for enabling tracepoints.


goldshtn:
Rebased so it can be merged automatically. This also resolves #591.

brendangregg:
Tested both argdist and trace, (and with -v to confirm TRACEPOINT_PROBE()), and they both work. Thanks!

brendangregg:
I think I broke the land speed record for time from PR to test. :)

4ast:
should we remove old 'tracepoint' hacks? or keep them around for old kernels?
I don't think anyone really used them (or planning to use), so I'd remove.

goldshtn:
@brendangregg Yes you did! :)

@4ast Do you mean removing the tracepoint.py and perf.py files that aren't used anymore? I think there are no references to them in any of the tools at this point.

4ast:
yes. so they don't accumulate dust.

brendangregg:
can the old versions be put under tools/old?

goldshtn:
@brendangregg I was thinking of putting them in tools/old, but that won't work if we get rid of the tracepoint.py and perf.py files. Also, it's unlikely that the old tools will actually work as the bcc module diverges going forward... So maybe it's best to just preserve them in git history.

goldshtn:
OK, so the `Perf` class is actually used elsewhere. The `Tracepoint` class is now completely unused, so I'm removing it.

brendangregg:
@goldshtn ok, if the tools are unlikely to work, then people's best bet is to grab an older bcc release, instead of putting them in tools/old

PR #733, opened 2016-10-06
funclatency: user functions support

goldshtn:
This introduces user functions support in `funclatency`. It's quite useful as a mini-profiler or just another way to quickly probe the latency of a particular function in user space.

Usage examples:

```
funclatency pthread:*mutex* -p 6449 -F
funclatency c:read
funclatency dd:* -p $(pidof dd) -uF
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ Examples:
 - tools/[fileslower](tools/fileslower.py): Trace slow synchronous file reads and writes. [Examples](tools/fileslower_example.txt).
 - tools/[filetop](tools/filetop.py): File reads and writes by filename and process. Top for files. [Examples](tools/filetop_example.txt).
 - tools/[funccount](tools/funccount.py): Count kernel function calls. [Examples](tools/funccount_example.txt).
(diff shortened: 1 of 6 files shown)


4ast:
LGTM cc @brendangregg

brendangregg:
LGTM, and tested, works, and is pretty amazing. Thanks @goldshtn !

@4ast please merge (I don't have that button)

@goldshtn funccount next? :) I actually use funccount more than funclatency -- funccount is good for checking the rate of function calls before doing anything more expensive. Plus, my old perf-tools funccount is kernel only (ftrace); this will be the first funccount that can do both.

brendangregg:
... we'll eventually need to update the reference-guide for the other bits that have been added...

goldshtn:
@brendangregg Thanks :) Yes, `funccount` is next. Planning to add uprobes, tracepoints, and USDT to that one. Probably going to copy-pasta most of `stackcount` and then remove pieces. It's actually kind of annoying that there's so much duplication in all the general-purpose tools, but I think most of the reusable pieces are already in libbcc.

PR #734, opened 2016-10-06
tplist: Print USDT locations and arguments

### LINKED ISSUE #711: tplist: Print USDT raw arguments for debugging in super-verbose mode
This requires adding a function in libbcc to expose the raw argument data. @4ast Did you want this to be printed per location (and indeed the argument sizes and values can be different in each location), or just the largest argument size as we report in #710?

goldshtn:
This adds support in libbcc (on the C++ and Python side) and in tplist for printing USDT probe locations and arguments in super-verbose (`-vv`) mode, as requested by @4ast.

Example:

```
# ./tplist.py -vv -l c *mmap*
/lib64/libc.so.6 libc:memory_mallopt_mmap_max [sema 0x0]
  location #0 0x85448
    argument #0 4 signed   bytes @ bp
    argument #1 4 signed   bytes @ *(&mp_ + 44)
    argument #2 4 signed   bytes @ *(&mp_ + 52)
/lib64/libc.so.6 libc:memory_mallopt_mmap_threshold [sema 0x0]
  location #0 0x8546e
    argument #0 4 signed   bytes @ bp
    argument #1 8 unsigned bytes @ *(&mp_ + 16)
    argument #2 4 signed   bytes @ *(&mp_ + 52)
```

Resolves #711.

### DIFF
--- a/man/man8/tplist.8
+++ b/man/man8/tplist.8
@@ -22,7 +22,8 @@ Display the USDT probes from the specified library or executable. If the librar
 or executable can be found in the standard paths, a full path is not required.
 .TP
 \-v
-Display the variables associated with the tracepoint or USDT probe.
+Increase the verbosity level. Can be used to display the variables, locations,
+and arguments of tracepoints and USDT probes.
 .TP
 [filter]
 A wildcard expression that specifies which tracepoints or probes to print.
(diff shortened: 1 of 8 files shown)


4ast:
awesome. Thanks a lot!

PR #739, opened 2016-10-07
Update opensnoop to filter by PID and TID

dinazil:
I updated opensnoop in to support #547: 
* Now supports passing TID and PID as filters where PID is the actual process ID and TID is the thread ID (note that this is a change on the behavior which currently used -p to filter by thread ID!)
* Output displays PID or TID according to the input parameters
* Added some missing documentation about the -n parameter

```
$ sudo ./opensnoop.py -Tt 2359
TIME(s)       TID    COMM               FD ERR PATH
0.000000000   2359   opens               3   0 /tmp/myfile.txt
```

If this looks good, I can update other tools similarly. @brendangregg

### DIFF
--- a/man/man8/opensnoop.8
+++ b/man/man8/opensnoop.8
@@ -2,7 +2,7 @@
 .SH NAME
 opensnoop \- Trace open() syscalls. Uses Linux eBPF/bcc.
 .SH SYNOPSIS
-.B opensnoop [\-h] [\-t] [\-x] [\-p PID] [\-n name]
+.B opensnoop [\-h] [\-T] [\-x] [\-p PID] [\-t TID] [\-n name]
 .SH DESCRIPTION
 opensnoop traces the open() syscall, showing which processes are attempting
 to open which files. This can be useful for determining the location of config
@@ -24,7 +24,7 @@ CONFIG_BPF and bcc.
 \-h
 Print usage message.
 .TP
-\-t
+\-T
 Include a timestamp column.
 .TP
 \-x
@@ -32,6 +32,9 @@ Only print failed opens.
 .TP
 \-p PID
 Trace this process ID only (filtered in-kernel).
+.TP
+\-t TID
+Trace this thread ID only (filtered in-kernel).
 .TP
 \-n name
 Only print processes where its name partially matches 'name'
@@ -43,7 +46,7 @@ Trace all open() syscalls:
 .TP
 Trace all open() syscalls, and include timestamps:
 #
-.B opensnoop \-t
+.B opensnoop \-T
 .TP
 Trace only open() syscalls that failed:
 #
@@ -63,6 +66,9 @@ Time of the call, in seconds.
 .TP
 PID
 Process ID
+.TP
+TID
+Thread ID
 .TP
 COMM
 Process name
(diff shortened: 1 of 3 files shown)


brendangregg:
Thanks, tested, works! Some minor style things:

```
# pep8 --show-source --ignore=E123,E125,E126,E127,E128,E302 opensnoop.py 
opensnoop.py:114:13: E261 at least two spaces before inline comment
if args.tid: # TID trumps PID
            ^
opensnoop.py:115:42: W291 trailing whitespace
    bpf_text = bpf_text.replace('FILTER',
                                         ^
opensnoop.py:147:80: E501 line too long (92 > 79 characters)
print("%-6s %-16s %4s %3s %s" % ("TID" if args.tid else "PID", "COMM", "FD", "ERR", "PATH"))
                                                                               ^
opensnoop.py:175:36: W291 trailing whitespace
    print("%-6d %-16s %4d %3d %s" %
                                   ^
opensnoop.py:176:66: W291 trailing whitespace
          (event.id & 0xffffffff if args.tid else event.id >> 32,
                                                                 ^
```

... I went to test it but couldn't find a nice threaded program, so went to write one that would, say, spawn a few threads, then have each thread do "open()" in a loop to a pathname containing a thread ID. Then I found I'd written exactly this over a year ago:

```C
#include <stdio.h>
#include <unistd.h>
#include <pthread.h>
#include <sys/types.h>
#include <sys/stat.h>
#include <fcntl.h>

#define	MAX_THREADS	8

void *thread_function(void *arg);

int
main(int argc, char *argv[])
{
	int i;
	pthread_t threads[MAX_THREADS];

	/* create threads */
	for (i = 0; i < MAX_THREADS; i++) {
		if (pthread_create(&threads[i], NULL, thread_function,
		    (void *)i) != 0)
			perror("thread create failed");
	}
	
	/* wait for threads to complete */
	for (i = 0; i < MAX_THREADS; i++)
		pthread_join(threads[i], NULL);

	return (0);
}

void *
thread_function(void *arg)
{
	int id, fd;
	char file[128], *filep;
	filep = &file;

	id = (int)arg;
	printf("thread started: %d\n", id);
	sprintf(filep, "thread%d", id);
	for (;;) {
		fd = open(filep, O_RDONLY);
		sleep(1);
	}

	(void) sleep(id);

	printf("thread stopped: %d\n", id);

	return (arg);
}
```

Testing -p:

```
# ./opensnoop.py -p 824
824    threads            -1   2 thread5
824    threads            -1   2 thread4
824    threads            -1   2 thread3
824    threads            -1   2 thread6
824    threads            -1   2 thread0
824    threads            -1   2 thread2
824    threads            -1   2 thread1
824    threads            -1   2 thread7
824    threads            -1   2 thread1
824    threads            -1   2 thread5
824    threads            -1   2 thread3
824    threads            -1   2 thread6
824    threads            -1   2 thread0
824    threads            -1   2 thread7
[...]
```

Testing -t:

```
# ./opensnoop.py -t 830
TID    COMM               FD ERR PATH
830    threads            -1   2 thread4
830    threads            -1   2 thread4
830    threads            -1   2 thread4
830    threads            -1   2 thread4
[...]
```

all good.

Yes, it'd be good to have accurate and consistent -p and -t across all tools that have them, and you're welcome to do that.

dinazil:
@brendangregg 
pep8 --show-source --ignore=E123,E125,E126,E127,E128,E302 opensnoop.py 
on my machine now seems satisfied.

brendangregg:
Thanks @dinazil , all LGTM.

PR #742, opened 2016-10-09
trace: Add %K and %U format specifiers

goldshtn:
The %K and %U format specifiers can be used in a trace
format string to resolve kernel and user symbols,
respectively. For example, the pthread_create USDT probe
has an argument pointing to the new thread's function.
To trace pthread_create and print the symbolic name of
the new thread's function, use:

```
trace 'u:pthread:pthread_create "%U", arg3'
```

The %U specifier resolves addresses in the event's process,
while the %K specifier resolves kernel addresses.

### DIFF
--- a/man/man8/trace.8
+++ b/man/man8/trace.8
@@ -92,6 +92,12 @@ number of arguments as there are placeholders in the format string. The
 format specifier replacements may be any C expressions, and may refer to the
 same special keywords as in the predicate (arg1, arg2, etc.).
 
+In addition to the above format specifiers, you can also use %K and %U when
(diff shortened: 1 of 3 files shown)


brendangregg:
Tested, works, thanks!

... at one point you might want to "pep8 --show-source --ignore=E123,E125,E126,E127,E128,E302 trace.py", although this commit didn't introduce the warnings...

goldshtn:
Yeah, it's not just trace though. I'll create an issue to clean up linter issues in the tools. Can we merge this in the meantime?

Issue #748, opened 2016-10-13
Can't use syscall tracepoints


brendangregg:
```
 # ./tools/trace.py t:syscalls:sys_enter_newfstat
Ioctl(PERF_EVENT_IOC_SET_BPF): Invalid argument
Failed to attach BPF to tracepoint
```

Here's newfstat.py:

``` Python
#!/usr/bin/python

from __future__ import print_function
from bcc import BPF

# load BPF program
b = BPF(text="""
TRACEPOINT_PROBE(syscalls, sys_enter_newfstat) {
    bpf_trace_printk("%d\\n", args->fd);
    return 0;
};
""", debug=0)

# header
print("%-18s %-16s %-6s %s" % ("TIME(s)", "COMM", "PID", "FD"))

# format output
while 1:
    try:
        (task, pid, cpu, flags, ts, msg) = b.trace_fields()
    except ValueError:
        continue
    print("%-18.9f %-16s %-6d %s" % (ts, task, pid, msg))
```

output:

```
# ./newfstat.py 
ioctl(PERF_EVENT_IOC_SET_BPF): Invalid argument
Traceback (most recent call last):
  File "./newfstat.py", line 12, in <module>
    """, debug=0)
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 212, in __init__
    self._trace_autoload()
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 767, in _trace_autoload
    self.attach_tracepoint(tp=tp, fn_name=fn.name)
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 585, in attach_tracepoint
    raise Exception("Failed to attach BPF to tracepoint")
Exception: Failed to attach BPF to tracepoint
```

Testing on 4.8-rc4. Same problem with other syscall tracepoints.

I can use kprobes as a workaround in the meantime, but this should be a nice example of using tracepoints...

4ast:
hmm. looks like syscalls:sys_(enter|exit)_* are not considered proper 'tracepoint' from kernel point of view, though perf thinks they are.

goldshtn:
@4ast: Yes, and what's interesting is that perf works just fine with these tracepoints, e.g. `perf record -e syscalls:*` works (albeit a bit slowly) and records all syscalls. Is there something we need to do differently to record these?

goldshtn:
[...]

Issue #569, opened 2016-06-14
man page notes about bpf_perf_event_output(): 4.5 or 4.4?


brendangregg:
Many man pages have:

> This makes use of a Linux 4.5 feature (bpf_perf_event_output());
> for kernels older than 4.5, see the version under tools/old,
> which uses an older mechanism.

Which I think I wrote. But I just checked on 4.4 (Ubuntu Xenial), and these tools are working fine. I think bpf_perf_event_output() did make it into 4.4 ( http://lxr.free-electrons.com/source/kernel/trace/bpf_trace.c?v=4.4#L223 ), which would mean these man pages need a s/4.5/4.4/g.

4ast:
looks like it got into 4.4 somehow :)

matthew-olson-intel:
It looks like this is fixed now. Time to close it?

```
$ grep -R "4\.5"
biolatpcts.8: read     95us 175us 305us 515us 895us 985us 995us 1.5ms 2.5ms 3.5ms 4.5ms  10ms
biolatpcts.8: flush     5us   5us   5us   5us   5us   5us   5us   5us   5us 1.5ms 4.5ms 5.5ms
```

brendangregg:
Thanks for checking @matthew-olson-intel, closing.

PR #750, opened 2016-10-14
Add new mountsnoop tool

osandov:
I've needed global mount namespace tracing many times, so here's a BPF tool to
do just that.

This tool actually points out a couple of gaps in BPF. Firstly, I had to
duplicate the definition of `struct mnt_namespace` in the BPF program since
it's not defined in any public headers. Secondly, emitting strings is a pain
without proper string support. The comments in the script have more details.

There's some stuff in there which might be useful in general for writing BCC
tools (decoding flags/errnos, formatting bytestrings in C syntax), but I kept
it private to the script for now.

Please pull. Thanks!

### DIFF
--- a/man/man8/mountsnoop.8
+++ b/man/man8/mountsnoop.8
@@ -1,0 +1,55 @@
+.TH mountsnoop 8  "2016-10-14" "USER COMMANDS"
+.SH NAME
+mountsnoop \- Trace mount() and umount() syscalls. Uses Linux eBPF/bcc.
+.SH SYNOPSIS
+.B mountsnoop
+.SH DESCRIPTION
+mountsnoop traces the mount() and umount() syscalls, showing which processes
+are mounting and unmounting filesystems in what mount namespaces. This can be
+useful for troubleshooting system and container setup.
+
+This works by tracing the kernel sys_mount() and sys_umount() functions using
+dynamic tracing, and will need updating to match any changes to this function.
+
+This makes use of a Linux 4.4 feature (bpf_perf_event_output()).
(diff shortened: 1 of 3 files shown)


brendangregg:
Looks good.

Style warnings, at some point these need to be cleaned up: pep8 --show-source --ignore=E123,E125,E126,E127,E128,E302 mountsnoop.py

In the future, when #748 is fixed, it might make sense to switch this to the syscall tracepoints instead of kprobes.

osandov:
Fixed the style warnings. Not sure why pep8 cares about how I formatted inside of a string literal.

man/man8/mountsnoop.8
```diff
@@ -0,0 +1,55 @@
...
+.B mountsnoop
+.SH DESCRIPTION
+mountsnoop traces the mount() and umount() syscalls, showing which processes
+are mounting and unmounting filesystems in what mount namespaces. This can be
+useful for troubleshooting system and container setup.
+
+This works by tracing the kernel sys_mount() and sys_umount() functions using
+dynamic tracing, and will need updating to match any changes to this function.
+
+This makes use of a Linux 4.5 feature (bpf_perf_event_output()).
```
goldshtn: I think this is actually in 4.4.

osandov: Huh, I lifted that from opensnoop.8, but you're right. Fixed and force pushed.

brendangregg: Yes, sorry, still haven't fixed all the man pages. #569

PR #753, opened 2016-10-15
cc: Support glob+offset format in USDT arguments

goldshtn:
Modern versions of USDT probes (such as what's found in
PostgreSQL when compiled with `--enable-dtrace`) may have
the offset listed after the global symbol for USDT
arguments of the format `4@symbol+8(%rip)`. This commit
extends the argument parser to support these cases, adds
tests for these cases, and makes sure that in case of a
parse error, the parser always moves forward and consumes
at least one character. Presently, the parser would get
stuck on the problematic position and enter an infinite
loop.


4ast:
[buildbot, add to whitelist]

PR #756, opened 2016-10-16
Support filtering by process ID in the filesystem slower tools

dinazil:
Changed the -p parameter to filter by real process ID rather than just first thread in the following tools:
* ext4slower.py
* btrfsslower.py
* xfsslower.py
* zfsslower.py

@brendangregg


brendangregg:
Thanks, LGTM!

Issue #728, opened 2016-10-04
bpf_ktime_get_ns() odd returns in 4.8-rc4


brendangregg:
Something is very amiss -- has this been found and fixed already post 4.8-rc4?

I was fixing filelife.py, and got the following:

```
# ./filelife.py
TIME     PID    COMM             AGE(s)  FILE
23:25:22 20376  rm               0.99    foo
23:25:24 20380  rm               1.00    foo
23:25:26 20384  rm               18446744073.70 foo
23:25:29 20388  rm               18446744073.70 foo
23:25:32 20392  rm               1.00    foo
23:25:34 20396  rm               1.00    foo
```

Test workload:

```
# time (touch foo; sleep 0.5; rm foo)

real	0m0.508s
user	0m0.000s
sys	0m0.000s
```

So none of filelife's output is correct. I added debug statements: https://gist.github.com/brendangregg/0e125937a48e9f27979e8a6942ced635

which print:

```
(root) /mnt/src/bcc/tools # cat /sys/kernel/debug/tracing/trace_pipe 
           touch-20374 [007] d... 6771453.064892: : at 2938902477515565, create of 8b929240
              rm-20376 [007] d... 6771453.569953: : unlink of 8b929240
              rm-20376 [007] d... 6771453.569959: : at 2938903469266385, delta 991
           touch-20378 [006] d... 6771454.872515: : at 2938904476374014, create of 8b929240
              rm-20380 [006] d... 6771455.377425: : unlink of 8b929240
              rm-20380 [006] d... 6771455.377429: : at 2938905478901276, delta 1002
           touch-20382 [006] d... 6771457.617427: : at 2938907478801260, create of 8b929240
              rm-20384 [006] d... 6771458.122352: : unlink of 8b929240
              rm-20384 [006] d... 6771458.122357: : at 2938907470212787, delta -140462620
           touch-20386 [006] d... 6771460.553749: : at 2938910479575068, create of 8b929240
              rm-20388 [006] d... 6771461.058745: : unlink of 8b929240
              rm-20388 [006] d... 6771461.058750: : at 2938910471184565, delta -140462619
           touch-20390 [006] d... 6771462.930027: : at 2938912469068110, create of 8b929240
              rm-20392 [006] d... 6771463.434816: : unlink of 8b929240
              rm-20392 [006] d... 6771463.434820: : at 2938913471254426, delta 1002
           touch-20394 [006] d... 6771465.162422: : at 2938914470088433, create of 8b929240
[...]

PR #757, opened 2016-10-16
Fix negative delta time

Nicop06:
While investigating slow Btrfs operations, I noticed a very high latency range that didn't seem right:

```
    2251799813685248 -> 4503599627370495     : 0        |                    |
    4503599627370496 -> 9007199254740991     : 0        |                    |
    9007199254740992 -> 18014398509481983    : 0        |                    |
   18014398509481984 -> 36028797018963967    : 1206     |*                   |
```

It seems that the start time is after the end time, leading to negative elapse time. I didn't have time to investigate more and I only have Btrfs on this laptop so I only submitted a quick fix for this filesystem, but it may impacts others.

Did it happen to anyone else? Do you have any idea to investigate further? I was thinking about tracing all events with their occurrence time to see if I can find something.

Also, I am using a 4.8 kernel with custom config (Gentoo), so maybe I forgot some debugging options. Do you have any idea which one I can check?

### DIFF
--- a/tools/btrfsdist.py
+++ b/tools/btrfsdist.py
@@ -119,15 +119,17 @@ int trace_open_entry(struct pt_regs *ctx, struct inode *inode,
 
 static int trace_return(struct pt_regs *ctx, const char *op)
 {
-    u64 *tsp;
+    u64 *tsp, ts;
     u32 pid = bpf_get_current_pid_tgid();
 
     // fetch timestamp and calculate delta
     tsp = start.lookup(&pid);
-    if (tsp == 0) {
+    ts = bpf_ktime_get_ns();
+
+    if (tsp == 0 || ts < *tsp) {
         return 0;   // missed start or filtered
     }
-    u64 delta = (bpf_ktime_get_ns() - *tsp) / FACTOR;
+    u64 delta = (ts - *tsp) / FACTOR;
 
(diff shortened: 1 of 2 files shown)


brendangregg:
Thanks, but I think this may be #728 -- the kernel time source BPF uses broke in 4.8. It's since been patched.

goldshtn:
I saw this multiple times on 4.8, quite consistently even. It seems fixed on 4.9 (since upgrading a couple of days ago I never saw this again).

brendangregg:
It's been fixed in 4.9-rc1. https://github.com/torvalds/linux/commit/58bfea9532552d422bde7afa207e1a0f08dffa7d#diff-03ab89da17a203808142f6a1f0272153

@Nicop06 are you able to check 4.9-rc1 or with that patch?

Nicop06:
Yes, this is exactly that. Also, this is fixed 4.8.2 according to the [changelog](https://www.kernel.org/pub/linux/kernel/v4.x/ChangeLog-4.8.2). I just upgraded today as it was released yesterday and so far it seems to work. I'd rather use a kernel supported by my distribution.

brendangregg:
I'll close this PR as it looks like the kernel bug was responsible instead. If you determine otherwise, please reopen. Thanks!

PR #759, opened 2016-10-17
add slabratetop

brendangregg:
Pity the cache name is not in the kmem:kmem_cache_alloc tracepoint.

In the example file I pulled out stacks for the top cache, kmalloc-4096, using my old ftrace tool:

```
# ./perf-tools/bin/kprobe -s 'p:kmem_cache_alloc name=+0(+96(%di)):string' 'name == "kmalloc-4096' | head -100
Tracing kprobe kmem_cache_alloc. Ctrl-C to end.
          kprobe-3892  [002] d... 7888274.478331: kmem_cache_alloc: (kmem_cache_alloc+0x0/0x1b0) name="kmalloc-4096"
          kprobe-3892  [002] d... 7888274.478333: <stack trace>
 => kmem_cache_alloc
 => user_path_at_empty
 => vfs_fstatat
 => SYSC_newstat
 => SyS_newstat
 => entry_SYSCALL_64_fastpath
```

Imagine we enhance trace.py to allow argdist-style digging, we could get as far as the cache name, but couldn't do the strcmp() comparison. Easily anyway. :)

A future addition to kmemcachetop might be to dig out the previous frame and print that as a column, so you can see the calling function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,7 @@ Examples:
 - tools/[gethostlatency](tools/gethostlatency.py): Show latency for getaddrinfo/gethostbyname[2] calls. [Examples](tools/gethostlatency_example.txt).
 - tools/[hardirqs](tools/hardirqs.py):  Measure hard IRQ (hard interrupt) event time. [Examples](tools/hardirqs_example.txt).
 - tools/[killsnoop](tools/killsnoop.py): Trace signals issued by the kill() syscall. [Examples](tools/killsnoop_example.txt).
+- tools/[slabratetop](tools/slabratetop.py): Kernel SLAB/SLUB memory cache allocation rate top. [Examples](tools/slabratetop_example.txt).
 - tools/[mdflush](tools/mdflush.py): Trace md flush events. [Examples](tools/mdflush_example.txt).
 - tools/[mysqld_qslower](tools/mysqld_qslower.py): Trace MySQL server queries slower than a threshold. [Examples](tools/mysqld_qslower_example.txt).
 - tools/[memleak](tools/memleak.py): Display outstanding memory allocations to find memory leaks. [Examples](tools/memleak_example.txt).
--- a/man/man8/slabratetop.8
+++ b/man/man8/slabratetop.8
@@ -1,0 +1,76 @@
+.TH slabratetop 8  "2016-10-17" "USER COMMANDS"
+.SH NAME
+slabratetop \- Kernel SLAB/SLUB memory cache allocation rate top.
+Uses Linux BPF/bcc.
+.SH SYNOPSIS
+.B slabratetop [\-h] [\-C] [\-r MAXROWS] [interval] [count]
+.SH DESCRIPTION
+This is top for the the rate of kernel SLAB/SLUB memory allocations.
+It works by tracing kmem_cache_alloc() calls, a commonly used interface for
+kernel memory allocation (SLAB or SLUB). It summarizes the rate and total bytes
+allocated of these calls per interval: the activity. Compare this to
+slabtop(1), which shows the current static volume of the caches.
+
+This tool uses kernel dynamic tracing of the kmem_cache_alloc() function.
+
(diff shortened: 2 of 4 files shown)


goldshtn:
Looks cool!

BTW, trace can do the same argument digging as argdist. For example you could do `trace 'p::kmem_cache_alloc "%x", arg1'` and get the `kmem_cache` pointer. The only problem is getting the name out because the `kmem_cache` struct is declared in a header that's not getting auto-included. You could modify the `_auto_includes` map in the BPF Python module so that the right header is included automatically. It might also be worth adding a `--include` option to trace, similar to argdist, so you could provide additional headers from the command line.

Of course, this doesn't help with the string comparison. I was thinking of experimenting with a helper function that would do an unrolled comparison against a constant string. This might provide a temporary solution until we have something better from the BPF side.

goldshtn:
So, I ran a little test. It looks like we can definitely use loops that implement `strcmp` manually if the string is known at compile-time. This opens up a bunch of use cases for scripts where we don't really need to compare two run-time strings, but rather compare one run-time string to one compile-time string (like your example above).

Here's a function that compares a run-time string to "true". It can be used from a bcc program.

```c
static inline bool equal_to_true(char *str) {
  char comparand[4];
  bpf_probe_read(&comparand, sizeof(comparand), str);
  char compare[] = "true";
  for (int i = 0; i < 4; ++i)
    if (compare[i] != comparand[i])
      return false;
  return true;
}
```

This can be generalized into a macro:

```c
#include <linux/ptrace.h>                                                       
                                                                                 
 #define DECLARE_EQUAL_TO(s) static inline bool equal_to_##s(char *str) { \      
     char comparand[sizeof(#s)]; \                                               
     bpf_probe_read(&comparand, sizeof(comparand), str); \                       
     char compare[] = #s; \                                                      
     for (int i = 0; i < sizeof(comparand); ++i) \                               
         if (compare[i] != comparand[i]) \                                       
             return false; \                                                     
     return true; \                                                              
 } \                                                                             
                                                                                 
 #define IS_EQUAL_TO(str, s) equal_to_##s(str)                                   
                                                                                 
 DECLARE_EQUAL_TO(true)                                                          
                                                                                 
 int test_strings(struct pt_regs *ctx, char *str) {                              
     bpf_trace_printk("String printed: %s, equal to 'true': %d\\n",              
         str, IS_EQUAL_TO(str, true));                                           
     return 0;                                                                   
 }                               
```

It's not elegant, but seems to be working. I think it can be incorporated into `trace` and `argdist` so we can compare strings. Thoughts? :)

4ast:
based on the name I assumed it's a faster version of slabtop, but it's a very different one. Since it tracks the rate of allocation. I guess the assumption that the rate of free is roughly similar (otherwise memory leaks eventually for top active users of slab).
Should it be kmemalloctop? or kmemallocratetop? or some way to make it less ambiguous that it's about rate of access and not current volume of allocated memory.

brendangregg:
Right; slabtop shows the static volume, this shows the rate of usage. I've debugged slab issues with the rate before (back on Solaris, when storage config was switching to the overflow slab instead of the 128k slab, and tracing alloc rate let you associate this with workloads and tests).

kmemcachealloctop is more descriptive, but pretty long. How about kcatop?

brendangregg:
@goldshtn Looks useful. Maybe one of us should put that macro into a simple /examples/tracing example along with a block comment to say "This is a workaround that works. It might get better in the future with BPF string support, it might not. At least there is a way today that works."

4ast:
> kmemcachealloctop is more descriptive, but pretty long. How about kcatop?

may be slabratetop ?

brendangregg:
slaberatetop... I think that might work...

Oh, I just added doc update and it put it on this PR, sorry. Should have used a branch.

brendangregg:
I'll fix this PR -- too many commits.

brendangregg:
PR fixed!

PR #762, opened 2016-10-18
tools: add tcptracer

iaguis:
This allows tracking TCP connections by tracking TCP connects, closes
and accepts.

This is different from existing tools like tcpconnect or tcpaccept in
that:

* It includes more information like network namespace or source ports
for tcpconnects or remote ports for tcpaccepts
* It traces tcp_close allowing to see when a connection ends
* It only shows information about established connections

### DIFF
--- a/tools/tcptracer_example.txt
+++ b/tools/tcptracer_example.txt
@@ -1,0 +1,26 @@
+Demonstrations of tcptracer, the Linux eBPF/bcc version.
+
+
+This tool traces the kernel function performing TCP connections (eg, via a
+connect() or accept() syscalls) and closing them (explicitly or if the process
+dies). Some example output (IP addresses are fake):
+
+```
+# ./tcptracer
+Tracing TCP established connections. Ctrl-C to end.
+T  PID    COMM             IP SADDR            DADDR            SPORT  DPORT
+C  28943  telnet           4  192.168.1.2      192.168.1.1      59306  23
+C  28818  curl             6  [::1]            [::1]            55758  80
+X  28943  telnet           4  192.168.1.2      192.168.1.1      59306  23
+A  28817  nc               6  [::1]            [::1]            80     55758
+X  28818  curl             6  [::1]            [::1]            55758  80
+X  28817  nc               6  [::1]            [::1]            80     55758
+A  28978  nc               4  10.202.210.1     10.202.109.12    8080   59160
+X  28978  nc               4  10.202.210.1     10.202.109.12    8080   59160
+```
+
+This output shows three conections, one outgoing from a "telnet" process, one
+outgoing from "curl" to a local netcat, and one incoming received by the "nc"
+process. The output details show the kind of event (C for connection, X for
+close and A for accept), PID, IP version, source address, destination address,
+source port and destination port.
--- a/README.md
+++ b/README.md
@@ -133,6 +133,7 @@ pair of .c and .py files, and some are directories of files.
 - tools/[tcplife](tools/tcplife.py): Trace TCP sessions and summarize lifespan. [Examples](tools/tcplife_example.txt).
 - tools/[tcpretrans](tools/tcpretrans.py): Trace TCP retransmits and TLPs. [Examples](tools/tcpretrans_example.txt).
 - tools/[tcptop](tools/tcptop.py): Summarize TCP send/recv throughput by host. Top for TCP. [Examples](tools/tcptop_example.txt).
+- tools/[tcptracer](tools/tcptracer.py): Trace TCP established connections (connect(), accept(), close()). [Examples](tools/tcptracer_example.txt).
 - tools/[tplist](tools/tplist.py): Display kernel tracepoints or USDT probes and their formats. [Examples](tools/tplist_example.txt).
 - tools/[trace](tools/trace.py): Trace arbitrary functions, with filters. [Examples](tools/trace_example.txt).
 - tools/[ttysnoop](tools/ttysnoop.py): Watch live output from a tty or pts device. [Examples](tools/ttysnoop_example.txt).
--- a/man/man8/tcptracer.8
+++ b/man/man8/tcptracer.8
@@ -1,0 +1,98 @@
+.TH tcptracer 8  "2017-03-27" "USER COMMANDS"
+.SH NAME
+tcptracer \- Trace TCP established connections. Uses Linux eBPF/bcc.
+.SH SYNOPSIS
+.B tcptracer [\-h] [\-v] [\-p PID] [\-N NETNS]
+.SH DESCRIPTION
+This tool traces established TCP connections that open and close while tracing,
+and prints a line of output per connect, accept and close events. This includes
+the type of event, PID, IP addresses and ports.
+
+This tool works by using kernel dynamic tracing, and will need to be updated if
+the kernel implementation changes. Only established TCP connections are traced,
+so it is expected that the overhead of this tool is rather low.
+
+Since this uses BPF, only the root user can use this tool.
+.SH REQUIREMENTS
+CONFIG_BPF and bcc.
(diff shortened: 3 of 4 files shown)


brendangregg:
Thanks; some quick comments:

- This isn't an /example, it's a /tool. Please see CONTRIBUTING_SCRIPTS.md.
- It uses 5 kernel dynamic tracepoints just for IPv4. This makes it especially brittle. tcpconnect uses 2, and tcpaccept uses 1, and they both cover IPv4 & IPv6.

This is the kind of tool I'd rather wait until we have TCP tracepoints to attempt -- speaking from experience. My tcpsnoop script from Solaris 10 used several dynamic kernel tracepoints, and it broke at least 7 times, and I ended up with a lot of upset end users. Once we had static tracepoints, I was able to write such tools once and they never broke again. We need them on Linux. And, should they arrive, supporting IPv6 should be much easier -- this would become a "tcptracer" tool.

Although, Solaris is obviously not Linux. The real question here is how likely these are to change:

- int kprobe__tcp_v4_connect(struct pt_regs *ctx, struct sock *sk)
- int kprobe__tcp_close(struct pt_regs *ctx, struct sock *sk)
- int kretprobe__tcp_v4_connect(struct pt_regs *ctx)
- int kretprobe__tcp_close(struct pt_regs *ctx)
- int kretprobe__inet_csk_accept(struct pt_regs *ctx)

We can add details like ports and namespaces to the existing tcpconnect/tcpaccept. And there's another ticket to implement a tcplife tool, for tcp_close() tracing.

brendangregg:
The output is also over 80 chars, by _default_, when it doesn't need to be. That breaks a lot of things, and should only be done if it's unavoidable.

If I were writing it, I'd have made the TYPE column a single character "T" with "C" == connect, etc. NETNS only printed with a -n or -v option. Truncated COMM a little (12 chars) if need be. Etc.

4ast:
> Although, Solaris is obviously not Linux. The real question here is how likely these are to change:

    int kprobe__tcp_v4_connect(struct pt_regs *ctx, struct sock *sk)
    int kprobe__tcp_close(struct pt_regs *ctx, struct sock *sk)
    int kretprobe__tcp_v4_connect(struct pt_regs *ctx)
    int kretprobe__tcp_close(struct pt_regs *ctx)
    int kretprobe__inet_csk_accept(struct pt_regs *ctx)

These funcs and their args are highly unlikely to change any time soon.
I think it's ok to go with kprobe approach for now, since there is no one actively proposing tracepoints for tcp stack and it's not clear what kind of push back there will be.

brendangregg:
Thanks @4ast. Yes, I really want TCP tracepoints, and maybe I'll just have to do propose them, but that's all in the future at the moment.

@iaguis So I'd check CONTRIBUTING_SCRIPTS.md, which will mean pep8 styling, an examples file, and a man page. Try to adjust the output width to keep it <80, which probably means putting the NETNS column in a verbose mode (-v). After all, if you run it with -n and have selected a NETNS, then you don't need to print it out by default.

I'd also consider changing -n to -N. -n is often used to match a process name, and that's a reasonable addition to this tool at some point ("tcpv4tracer java").

brendangregg:
Also, why do you bother with the kretprobe of tcp_close()? Is something not valid on the kprobe?

iaguis:
Thanks everybody for the quick responses!

> Also, why do you bother with the kretprobe of tcp_close()? Is something not valid on the kprobe?

You're right, the kprobe should be enough. Thanks!

I'm sorry I haven't read CONTRIBUTING_SCRIPTS.md before submitting this, I'll rework this PR taking that and https://github.com/iovisor/bcc/pull/762#issuecomment-254598664 into account.

goldshtn:
@iaguis This PR is still open -- is it ready for another review?

iaguis:
> @iaguis This PR is still open -- is it ready for another review?

I'll try to rebase it and update by the end of the week.

4ast:
@iaguis waiting for rebase and addressing @brendangregg comments re: contributing_scripts.md

goldshtn:
@iaguis Ping? :)

iaguis:
Rebased, addressed comments (I think I didn't forget any).

goldshtn:
[buildbot, ok to test]

goldshtn:
[buildbot, test this please]

goldshtn:
Copying @brendangregg @4ast to take a look.

brendangregg:
It's pretty long, at 649 lines, which would make it the 3rd longest tool in bcc (coming after the multi-tools trace and argdist). My longest tool is tcplife, at 360 lines, and my average is 194. Can the header include something like?:

```
# You should generally try to avoid writing long scripts that measure multiple functions and
# walk multiple kernel structures, as they will be a burden to maintain as the kernel changes.
# The following code should be replaced, and simplified, when static TCP probes exist.
```

My hope is to discourage others copying this and making it longer/worse. We need static TCP tracepoints, which should cut down a lot of these lines, and make the script stable.

BTW:

```
# ./tcptracer.py -h
usage: tcptracer.py [-h] [-p PID] [-N NETNS] [-v]

Trace TCP connections

optional arguments:
  -p PID, --pid PID     trace this PID only
  -N NETNS, --netns NETNS
                        trace this Network Namespace only
  -v, --verbose         include Network Namespace in the output
```

It doesn't do that much right now, either. When people start adding options, this tool will quickly become the biggest in bcc.

Other comments...

-v has a \n in the header:

```
# ./tcptracer.py -v
Tracing TCP established connections. Ctrl-C to end.
TYPE         PID    COMM             IP SADDR            DADDR            SPORT  DPORT 
NETNS   
```

And the T column:

```
./tcptracer.py 
Tracing TCP established connections. Ctrl-C to end.
T  PID    COMM             IP SADDR            DADDR            SPORT  DPORT 
AC 1171   sshd             4  100.66.11.247    100.72.22.55     22     27692 
CL 19801  sshd             4  100.66.11.247    100.72.22.55     22     27692 
```

When I see two chars, I think flags (like the RWBS field from the block tracepoints). I'd consider changing it to a single character (which matches the header, "T"):

- A: accept
- C: connect
- X: close

or, to be TCP-terminology pedantic:

- A: active open (what at the socket layer would be a connect())
- P: passive open (what at the socket layer would be an accept())
- C: close

Maybe the A/C/X is better, as those terms are more commonly used.

iaguis:
Thanks for the comments. Updated.

goldshtn:
[buildbot, ok to test]

tools/tcptracer_example.txt
```diff
@@ -0,0 +1,25 @@
...
+# ./tcptracer
+T  PID    COMM             IP SADDR            DADDR            SPORT  DPORT
+C 2389    telnet           4  192.168.1.2      192.168.1.1      53404  23
+C 23323   curl             6  [::1]            [::1]            36226  80
+X 2389    telnet           4  192.168.1.2      192.168.1.1      53404  23
+A 23278   nc               6  [::1]            [::1]            80     36226
+X 23323   curl             6  [::1]            [::1]            36226  80
+X 23278   nc               6  [::1]            [::1]            80     36226
+A 15195   nc               4  10.202.210.1     10.202.109.12    8080   43904
+X 15195   nc               4  10.202.210.1     10.202.109.12    8080   43904
```
goldshtn: Is the PID column header really misaligned now with the first digit of the pid?

iaguis: Ooops. No, that was me being off-by-one when editing the file. Let me re-run the test scenario.

goldshtn:
[buildbot, ok to test]

drzaeus77:
@goldshtn 
I updated one more setting on the buildbot, your [] should work next time.

brendangregg:
LGTM

brendangregg:
[buildbot, ok to test]

PR #763, opened 2016-10-18
funccount: Generalized for uprobes, tracepoints, and USDT

goldshtn:
This PR introduces the same set of changes done in `stackcount` and `funclatency` -- support user functions, kernel tracepoints, and USDT probes in addition to kernel functions as targets for `funccount`.

Examples (see _example.txt file for example output):

```
funccount t:block:*
funccount u:node:gc*
funccount -r 'c:(read|write)$'
funccount -p 142 u:ruby:object__create
```

Resolves #731.

PR #766, opened 2016-10-18
add reset-trace

brendangregg:
this came up in #763

### DIFF
--- a/tools/reset-trace.sh
+++ b/tools/reset-trace.sh
@@ -1,0 +1,135 @@
+#!/bin/bash
+#
+# reset-trace - reset state of tracing, disabling all tracing.
+#               Written for Linux.
+#
+# If a bcc tool crashed and you suspect tracing is partially enabled, you
+# can use this tool to reset the state of tracing, disabling anything still
+# enabled. Only use this tool in the case of error, and, consider filing a
+# bcc ticket so we can fix the error.
+#
+# bcc-used tracing facilities are reset. Other tracing facilities (ftrace) are
+# checked, and if not in an expected state, a note is printed. All tracing
+# files can be reset with -F for force, but this will interfere with any other
+# running tracing sessions (eg, ftrace).
+#
+# USAGE: ./reset-trace [-Fhqv]
+#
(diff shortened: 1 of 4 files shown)


palmtenor:
Thanks! @drarmstr and I encountered a bunch of such issues when developing our tools. We wrote something our own but it would be good to have something available to public.

tools/reset-trace.sh
```diff
@@ -0,0 +1,91 @@
...
+done
+shift $(( $OPTIND - 1 ))
+
+### reset tracing state
+vecho "Reseting tracing state..."
+vecho
+cd $tracing || die "ERROR: accessing tracing. Root user? /sys/kernel/debug?"
+writefile kprobe_events
+writefile uprobe_events
+writefile trace         # clears trace_pipe
```
4ast: what about adding something like:
echo nop > current_tracer
echo nobin > trace_options
echo context-info > trace_options

brendangregg:
@4ast Yes, I had them in my original ftrace one, but stripped them out, on the notion that if we wern't setting them, then why reset them (in case someone is using them). This is what I also had:

```
writefile current_tracer nop
writefile set_ftrace_filter
writefile set_graph_function
writefile set_ftrace_pid
writefile events/enable 0
writefile tracing_thresh 0
writefile tracing_on 1
```

Where there is no argument, it's a "> file".

4ast:
makes sense.
can we warn that the options are not in the state that bcc expects?
All bpf_trace_printk examples will dump junk when trace_pipe is in binary mode...

brendangregg:
ok, so you mean not adding them back makes sense? :)

I'll add warnings if it finds it needs to reset something, and change cat to "cat -v" to handle binary.

4ast:
yes. I mean it makes sense not to 'reset' the global state that bcc scripts didn't touch :)
may be separate script for expected bcc global state is needed... or a flag to this reset-trace... though warn about unexpected global configuration is probably sufficient.

brendangregg:
Updated. It'll now warn on detecting unrelated (ftrace) files are in use, and can reset them with -F for force.

4ast:
nice!

Issue #767, opened 2016-10-18
Support BPF_PROG_TYPE_PERF_EVENT in 4.9 for profiling


brendangregg:
This is the new profiling/PMCs support that's in 4.9-rc1. I'm guessing it'll require several changes, including:

src/python/bcc/\_\_init\_\_.py:
- \_trace\_autoload() to match on "perf_event", and call a new attach_perfevent() function.
- maybe passing a perf FD to the BPF creator. I'm imagining something like:

```Python
bpf_text = """
perf_event(struct pt_regs *ctx) {
[...]
}
"""

pfd = Perf.perf_event_open(0, pid=-1, ptype=Perf.PERF_TYPE_SOFTWARE,
                freq=args.frequency)
b = BPF(text=bpf_text, perf=pfd)
```

So updating perf_event_open() to return the fd, and BPF's init to parse a "perf" option and ioctl PERF_EVENT_IOC_SET_BPF on the passed FD.

And a tests/python/test_perfevent.py.

brendangregg:
this just merged in #770, closing ticket.

PR #768, opened 2016-10-18
fix profile.py page_offset_base breakage

### LINKED ISSUE #760: profile error: page_offset_base could not be resolved
```
# ./profile
LLVM ERROR: Program used external function 'page_offset_base' which could not be resolved!
```

This is due to this kernel change: https://github.com/torvalds/linux/commit/021182e52fe01c1f7b126f97fd6ba048dc4234fd#diff-2c06e1fa0d80d36f7530356f4e2a1865

profile needs to be updated anyway for the 4.9 perf_event_open() support.

brendangregg:
fixes #760. Will revisit this tool later after #767 (and this "fixed" version will be moved to tools/old, for kernels prior to 4.9).

### DIFF
--- a/man/man8/profile.8
+++ b/man/man8/profile.8
@@ -18,20 +18,16 @@ kernel context, rather than passing each stack to user space for frequency
 counting there. Only the unique stacks and counts are passed to user space
 at the end of the profile, greatly reducing the kernel<->user transfer.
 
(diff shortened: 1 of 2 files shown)


brendangregg:
I'll fix this PR. seems to have picked up my other PR as well.

brendangregg:
Fixed.

PR #770, opened 2016-10-19
Add basic support for BPF perf event 

palmtenor:
- Enum changes for perf event program type.
- Add support of attaching to PMU counter overflow events in C `libbpf`.
- Add support of attaching to PMU counter overflow events in Python.
- Add an example to demonstrate the usage.

Tested using the new `llcstat.py` tool


examples/tracing/cache_miss.py
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/python
+#
+# cache_miss.py Example of using BPF perf event.
+#               For Linux, uses BCC, eBPF. Embedded C.
+#
+# REQUIRES: Linux 4.7+ (BPF_PROG_TYPE_PERF_EVENT support).
```
palmtenor: Is this correct? What is the exact Kernel version that his is supported?

src/python/bcc/__init__.py
```diff
@@ -62,13 +62,47 @@ def resolve_name(self, name):
...
+    CPU_MIGRATIONS = 4
+    PAGE_FAULTS_MIN = 5
+    PAGE_FAULTS_MAJ = 6
+    ALIGNMENT_FAULTS = 7
+    EMULATION_FAULTS = 8
+    DUMMY = 9
+    BPF_OUTPUT = 10
+
 class BPF(object):
+    # From bpf_prog_type in uapi/linux/bpf.h
```
palmtenor: I want to move these to a separated class, or may all these enums to a separated `enums.py` or `constants.py`. But that involves changing a lot of existing tools. Should I do that?

examples/tracing/cache_miss.py
brendangregg: This should be Linux 4.9+.

brendangregg:
Wow, neat, thanks. Is it coincidence that I filed #767 earlier? :)

I see this approach does:

```
 b.attach_perf_event(
     ev_type=BPFPerfType.HARDWARE, ev_config=BPFPerfHWConfig.CACHE_MISSES,
     fn_name="on_cache_miss", sample_period = opts.sample_period,
     sample_freq = opts.sample_freq)
```

Which follows attach_kprobe() style. Whereas I was doing something that felt more like the BPF(..., usdt_context=) style. I'm fine with this attach_perf_event() instead.

This works for sampling too, right? ie, this:

```
	struct perf_event_attr pe_sample_attr = {
		.type = PERF_TYPE_SOFTWARE,
		.freq = 1,
		.sample_period = freq,
		.config = PERF_COUNT_SW_CPU_CLOCK,
		.inherit = 1,
	};
```

more comments in the review...

examples/tracing/cache_miss.py
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/python
+#
+# cache_miss.py Example of using BPF perf event.
```
brendangregg: This is a good example, but it should be a good tool as well, so this might belong under /tools, with a man page and examples file.

examples/tracing/cache_miss.py
```diff
@@ -0,0 +1,123 @@
...
+import argparse
+from bcc import BPF, BPFPerfType, BPFPerfHWConfig
+import signal
+from time import sleep
+
+parser = argparse.ArgumentParser(
+    description="Summarize cache misses by PID or stack trace",
+    formatter_class=argparse.RawDescriptionHelpFormatter)
+parser.add_argument(
+    "--duration", default=10, type=int, help="Time to run, in seconds")
```
brendangregg: I think duration would be the most commonly used option, in which case, I like the style of making the most commonly used option an optional positional parameter instead of an flag option. Like I did in profile.py. So "cachemiss" uses a default, and "cachemiss 5" runs for 5 seconds.

examples/tracing/cache_miss.py
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/python
```
brendangregg: I think it'd be ok to name this cachemiss.py. If you look at /tools, and also Linux /usr/bin, underscores are rare.

examples/tracing/cache_miss.py
```diff
@@ -0,0 +1,123 @@
...
+import signal
+from time import sleep
+
+parser = argparse.ArgumentParser(
+    description="Summarize cache misses by PID or stack trace",
+    formatter_class=argparse.RawDescriptionHelpFormatter)
+parser.add_argument(
+    "--duration", default=10, type=int, help="Time to run, in seconds")
+parser.add_argument(
+    "--sample_period", default=0, type=int,
```
brendangregg: I think this is better called "--sample_count", with a shortened version as well, "-c". This also makes it similar to "perf record".

examples/tracing/cache_miss.py
```diff
@@ -0,0 +1,123 @@
...
+    help="Sample one in this many events")
+parser.add_argument(
+    "--sample_freq", default=0, type=int,
```
brendangregg: add "-F", copying "perf record".

examples/tracing/cache_miss.py
```diff
@@ -0,0 +1,123 @@
...
+    help="Try to sample this many events per second")
+parser.add_argument(
+    "--stack", default=False, action="store_true",
```
brendangregg: I'd probably add "-s", or maybe "-g" after perf record (although -g is harder to remember. Call graph?).

examples/tracing/cache_miss.py
```diff
@@ -0,0 +1,123 @@
...
+    sleep(opts.duration)
+except KeyboardInterrupt:
+    signal.signal(signal.SIGINT, lambda signal, frame: print())
+
+count = b.get_table('count')
+if opts.stack:
+    stack_traces = b.get_table("stack_traces")
+
+for k, v in sorted(count.items(), key=lambda count: count[1].value):
+    print('PID {}-{} on CPU {}: {} cache misses'.format(
```
brendangregg: FWIW, I don't like how this feels like two tools in one: a statistical tool (perf stat), and a stack reporting tool (perf record). If I were to make it, I might have made a cachemissstat, and then a cachemissstacks, especially if the division made the code much cleaner for each. In this case, it doesn't look like the code is too bad...

And a cachemissstat could instrument both CACHE_MISS and CACHE_REFERENCES, and become just cachestat with a column for each, and another column for the ratio as a percent.

pmcprofile is split into a pmcstat and pmcstacks. I'll add another comment to the PR directly and explain this more.

examples/tracing/cache_miss.py
```diff
@@ -0,0 +1,123 @@
...
+    return 0;
+}
+"""
+text = text.replace("DATA_REPLACE", DATA)
+text = text.replace("HASH_REPLACE", HASH)
+text = text.replace("GET_REPLACE", GET)
+b = BPF(text=text)
+
+b.attach_perf_event(
+    ev_type=BPFPerfType.HARDWARE, ev_config=BPFPerfHWConfig.CACHE_MISSES,
```
brendangregg: Huh. So you could make CACHE_MISS an argument (-e), and turn this tool into a multi-tool:

pmcprofile -e cache_miss
pmcprofile -e instructions -s 5

etc. More on line 102...

examples/tracing/cache_miss.py
```diff
@@ -0,0 +1,123 @@
...
+count = b.get_table('count')
+if opts.stack:
+    stack_traces = b.get_table("stack_traces")
+
+for k, v in sorted(count.items(), key=lambda count: count[1].value):
+    print('PID {}-{} on CPU {}: {} cache misses'.format(
+        k.pid, k.name, k.cpu, v.value))
+    if not opts.stack:
+        continue
+    print("  Kernel Stack:")
```
brendangregg: This appears in another tool (trace.py) and I feel like it could be more concise. In profile.py I don't decorate each stack at all, and only put a delimiter with a -d option.

examples/tracing/cache_miss.py
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/python
+#
+# cache_miss.py Example of using BPF perf event.
+#               For Linux, uses BCC, eBPF. Embedded C.
+#
```
brendangregg: Need a comment to say something like: this instruments CACHE_MISS, which comes from uapi/linux/perf_event.h, and is really defined for each specific architecture. For example, on x86_64 it means LLC miss (LONGEST_LAT_CACHE.MISS).

examples/tracing/cache_miss.py
brendangregg: (oh, "cachestat" is taken).

brendangregg:
Ok, so I think there's an awesome opportunity here. I'll just jump to the future and say what I think this could become:

- **/examples/tracing/cachemiss.py**: remove the stack stuff, and just do the PID summary. A shorter, neat, example. (I'm right that the current "perf stat" can't do breakdowns by PID, right?)
- **/tools/cpucachestat.py**: remove the stack stuff, add CACHE_REFERENCES, then have a column for REFS, MISSES, and RATIO. cpucachestat is a mouthful. llcstat.py may be a better name (it's measuring the LLC on x86_64, at least).
- **/tools/pmcstat.py**: remove the stack stuff, and make CACHE_MISSES a "-e" "--event" option. Powerful multi-tool. Positional parameters can be [interval [count]]. Maybe support other breakdowns: by-PID, or by-CPU_ID.
- **/tools/pmcstacks.py**: remove the PID stuff, and make CACHE_MISSES a "-e" "--event" option. Powerful multi-tool. Positional parameters can be [duration].

I think you've already done most of the work, these are minor customizations. And having cachemiss, pmcstat, pmcstacks, and llcstat would be great additions: an example, two powerful multitools, and one single purpose tool.

src/python/bcc/__init__.py
goldshtn: What's the motivation for moving them?

examples/tracing/cache_miss.py
goldshtn: Absolutely! This can be a great tool. I'm actually thinking this can be a generalized tool for multiple kinds of events?

goldshtn:
I also think it could be nice to have this tool report at the IP level, not the full stack. Just to figure out which functions (and potentially lines) are causing cache misses, page faults, and so on. Can be cheaper to collect than the full stack.

brendangregg:
@goldshtn right, for pmcstacks. It could have either:
- **"-l" "--stack-depth"**: where you can truncate to 5 levels deep (that's where the "-l" is from), or 1 for just the IP and not the stack trace that follows.
- **"-i"**: to switch from stacks to just the IP.

I think I'd prefer the "-l" approach.

src/python/bcc/__init__.py
```diff
@@ -62,13 +62,47 @@ def resolve_name(self, name):
             return -1
         return addr.value
 
+class BPFPerfType:
```
goldshtn: I think this should probably be just `PerfType`, which aligns better with other classes not having the `BPF` prefix.

src/python/bcc/__init__.py
```diff
@@ -62,13 +62,47 @@ def resolve_name(self, name):
...
+    # From perf_type_id in uapi/linux/perf_event.h
+    HARDWARE = 0
+    SOFTWARE = 1
```
goldshtn: Are `PERF_TYPE_HW_CACHE` and `PERF_TYPE_RAW` not supported by the underlying implementation? It could be nice to have these as well.

brendangregg:
Oh, and profile.py already has some useful functionality that a pmcstack could borrow from.

```
optional arguments:
  -p PID, --pid PID     profile this PID only
  -U, --user-stacks-only
                        show stacks from user space only (no kernel space
                        stacks)
  -K, --kernel-stacks-only
                        show stacks from kernel space only (no user space
                        stacks)
  -F FREQUENCY, --frequency FREQUENCY
                        sample frequency, Hertz (default 49)
  -d, --delimited       insert delimiter between kernel/user stacks
  -a, --annotations     add _[k] annotations to kernel frames
  -f, --folded          output folded format, one line per stack (for flame
                        graphs)
  --stack-storage-size STACK_STORAGE_SIZE
                        the number of unique stack traces that can be stored
                        and displayed (default 2048)
  -S KERNEL_SKIP, --kernel-skip KERNEL_SKIP
                        skip this many kernel frames (default 3)
```

Things like -af or -df are good for making flamegraphs (depending on the style of highlighting you want: either breaking on the user/kernel boundary, or just annotating kernel frames; flamegraphs.pl already parses either of these: -d delimiters become grey frames; -a annotations change the hue).

palmtenor:
@brendangregg @goldshtn Thanks a lot for the review! The example is definitely far from good, I wrote it solely for the purpose of testing. I think what I'm gonna do now is to just move it to `tools/`, remove the stack stuffs and add cache reference count so that it would become a simple cache hit ration stat tool:)

brendangregg:
@palmtenor thanks, that's a good start (I'd go with llcstat), and you can either follow on and do pmcstat/pmcstack or someone else can.

Just remember that we don't need system-wide LLC ratio tool -- "perf stat -a" already does that. The value here is you can easily breakdown by PID using a BPF map.

palmtenor:
Updated the example to be `tools/llcstat.py`, as well as the test plan in the summary.
I have no idea why some times we get more cache miss events than reference events. I will try to figure it out......

palmtenor:
Add (some sort of) detaching for completeness

src/python/bcc/__init__.py
```diff
@@ -62,13 +62,47 @@ def resolve_name(self, name):
             return -1
         return addr.value
 
+class PerfType:
+    # From perf_type_id in uapi/linux/perf_event.h
+    HARDWARE = 0
+    SOFTWARE = 1
```
goldshtn: @palmtenor Should this have cache and raw types as well? I asked the same thing on the previous diff, but the force push destroyed it 💣

src/cc/libbpf.h
```diff
@@ -64,6 +64,13 @@ void * bpf_open_perf_buffer(perf_reader_raw_cb raw_cb, void *cb_cookie, int pid,
 /* attached a prog expressed by progfd to the device specified in dev_name */
 int bpf_attach_xdp(const char *dev_name, int progfd);
 
+// attach a prog expressed by progfd to run on a specific perf event, with
+// certain sample period or sample frequency
+int bpf_attach_perf_event(int progfd, uint32_t ev_type, uint32_t ev_config,
+                          uint64_t sample_period, uint64_t sample_freq,
+                          pid_t pid, int cpu, int group_fd);
+int bpf_deatch_perf_event(uint32_t ev_type, uint32_t ev_config);
```
goldshtn: @palmtenor This looks like a typo, should be `detach`.

src/python/bcc/__init__.py
palmtenor: According to https://github.com/torvalds/linux/commit/0515e5999a466dfe6e1924f460da599bb6821487 I think so. @4ast can confirm

src/cc/libbpf.h
palmtenor: Thanks!

palmtenor:
Why the build checks won't run for this PR?

goldshtn:
@palmtenor I think one of the repo owners needs to add you to the whitelist, or instruct the buildbots to test on a case-by-case basis...

brendangregg:
Tested, works. Some minor suggestions:

- The -F/--freq should be -c/--count, to match "perf record". (I'd use -F/--freq for profiling tools)
- Instead of "RATIO" I'd call the column "MISS%" (better still, change it to be a HIT%, which is what people might expect).
- "{:>8d} {:>16s} {:>4d} {:>12d} {:>12d} {:>8.2f}%" -> "{:<8d} {:<16s} {:<4d} {:>12d} {:>12d} {:>8.2f}%" -- maybe I should put this in CONTRIBUTING_SCRIPTS.md, but I love using justification to separate static workload attributes from the measurements. PID & COMM get left justified, and measurements like MISS and HIT% would be right justified.

It still needs a man page and examples file, but I'm so interested in this merged ASAP I'll offer to write them and add them as a follow up PR.

src/python/bcc/__init__.py
4ast: right now it's hw and sw only. Adding support for hw_cache and raw would be nice. If somebody can craft a kernel patch and test case, I can help upstreaming it.

brendangregg:
buildbot, retest this please

brendangregg:
@palmtenor also, do you have your original cache_miss.py handy somewhere? I just wanted to check how stacks were handled.

brendangregg:
... I figured out the stack stuff. I updated profile.py to use this new support, and it works great. I'll add it once this PR lands.

palmtenor:
Change `-F` to `-c`, add `llcstat_example.txt`

palmtenor:
Add `llcstat_example.txt` and `man` file

brendangregg:
Thanks, LGTM. Does @drzaeus77 or @4ast want to merge?

drzaeus77:
[buildbot, test this please]

drzaeus77:
[buildbot, add to whitelist]

@brendangregg fyi, the "[]" are currently necessary to satisfy the pull request plugin regex, I notice you've been omitting them.

brendangregg:
@drzaeus77 ok thanks

PR #771, opened 2016-10-19
funccount: Verify probe max limit

goldshtn:
Because `funccount` doesn't use the direct regex attach infrastructure
in the BPF module, it needs its own checking for a maximum probe
limit that would make sense. We use 1000 because that's what the
BPF module uses as well. When trying to attach to more than 1000
probes, we bail out early. (This is better than the behavior in the last
merged PR because it doesn't attach probes one-by-one and then fail,
it bails out immediately.)

@brendangregg You were so quick to merge that I didn't manage to get
this in the last PR 😄

### DIFF
--- a/tools/funccount.py
+++ b/tools/funccount.py
@@ -27,6 +27,12 @@ import traceback
 
 debug = False
 
+def verify_limit(num):
+    probe_limit = 1000
+    if num > probe_limit:
+        raise Exception("maximum of %d probes allowed, attempted %d" %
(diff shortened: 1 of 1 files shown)


brendangregg:
LGTM

brendangregg:
Tested. Good error message.

```
# ./funccount.py.1 'bash:*'
maximum of 1000 probes allowed, attempted 1470
```

Issue #691, opened 2016-09-15
string functions


brendangregg:
This is on BPF tasks #574 as string factory (stringmap). Can we propose a minimal set of functions we'd like, to begin with? Here's my list:

- strstr() or strncmp() (or both)
- strncat() and strlen(), or a basic snprintf()

Examples of current workarounds: strcmp():

- https://github.com/iovisor/bcc/blob/master/examples/networking/http_filter/http-parse-simple.c#L71 
- https://github.com/weaveworks/scope/blob/master/examples/plugins/http-requests/http-requests.c#L95

strncat():

- https://github.com/iovisor/bcc/blob/master/tools/execsnoop.py#L73 

This emits arguments one by one. This is tricky, since we need to walk argv\[] (currently using an unrolled loop). I was thinking strncat() would let us build a concatenated args string, but it probably needs to use strlen() as well to keep checking the destination string size. A basic snprintf() sounds much better, but .. gah this is awful.

How about a helper just for the argv[] case? :) Ie, take an argv\[] argument and return a truncated string of the arguments. bpf_strarray_to_str().

Then, in lieu of another use case, my list is just:

- strstr() or strncmp() (or both)

@4ast may have another thread of discussion about this. I'm posting this here as it was suggested in IRC.

alban:
Thanks for filing this issue. /cc @alban @asymmetric @alepuccetti @iaguis @2opremio

goldshtn:
I'd really like `strcpy`, which is very needed in argdist (see #607) so we don't do aggregations based on extraneous characters (which almost always leads to wrong results when aggregating on strings without specifying the exact string size).

brendangregg:
from @goldshtn on #759:

> So, I ran a little test. It looks like we can definitely use loops that implement strcmp manually if the string is known at compile-time. This opens up a bunch of use cases for scripts where we don't really need to compare two run-time strings, but rather compare one run-time string to one compile-time string (like your example above).
> 
> Here's a function that compares a run-time string to "true". It can be used from a bcc program.

```
static inline bool equal_to_true(char *str) {
  char comparand[4];
  bpf_probe_read(&comparand, sizeof(comparand), str);
  char compare[] = "true";
  for (int i = 0; i < 4; ++i)
    if (compare[i] != comparand[i])
      return false;
  return true;
}
```

> This can be generalized into a macro:

```
 #include <linux/ptrace.h>                                                       

[...]

PR #772, opened 2016-10-19
trace, argdist: STRCMP helper function

goldshtn:
Hacky solution, but works: STRCMP is replaced by a static inline function call that compares a runtime string to a compile-time literal constant. This is good enough to unroll so we don't need a backward jump, which is forbidden by BPF. Quite useful examples can arise, such as tracing opens to a specific file, aggregating operations on a specific device, and so on.

/cc @brendangregg


brendangregg:
Neat workaround. I'd be inclined to put a comment above it to say "# this is a workaround implementation of strcmp(). In the future this might get replaced. See #691."

We can see how painful there workarounds are in practice. They make be good enough...

brendangregg:
I'm happy to merge; I don't know if @4ast / @drzaeus77 wanted to comment, or at least know we're trying out this workaround. :)

4ast:
lgtm

PR #775, opened 2016-10-20
fix junk filenames

### LINKED ISSUE #747: Fix *slower tools to use d_name.name instead of dentry->d_iname
I don't think d_iname is reliable: it sometimes contains junk. See #237.

brendangregg:
fixes #747 -- the use of d_iname instead of d_name.name in ext4slower, etc.


4ast:
thanks for taking care of it!

PR #781, opened 2016-10-24
C++ helper class for BCC

palmtenor:
We have pretty good helper class available in Python and Lua. However, for C++ developers, they would need to go into the details of `libbpf` and `bcc_module` to figure out what needs to be done to attach their probes. This Diff adds some basics of a C++ helper class, that includes:
- Basic attaching and detaching of kprobes, uprobes and tracepoints.
- Reading of hash table and stack trace table.
- A few examples to demonstrate how to use the APIs.

```
$ sudo ./HelloWorld
The BPF Program is:

int probe_sys_clone(void *ctx) {
  bpf_trace_printk("Hello, World!");
  return 0;
}
           <...>-1105667 [010] d... 555825.171913: : Hello, World!           <...>-1105668 [009] d... 555825.173268: : Hello, World!           <...>-1105668 [014] d... 555825.174373: : Hello, World!           <...>-1105670 [022] d... 555825.174582: : Hello, World!           <...>-1105670 [022] d... 555825.174698: : Hello, World!           <...>-1105668 [000] d... 555825.176054: : Hello, World!           <...>-1105673 [028] d... 555825.176271: : Hello, World!           <...>-1105673 [028] d... 555825.176392: : Hello, World!           <...>-1105668 [014] d... 555825.177785: : Hello, World!           <...>-1105676 [022] d... 555825.177996: : Hello, World!           <...>-1105676 [022] d... 555825.178099: : Hello, World!           <...>-1105668 [000] d... 555825.179444: : Hello, World!           <...>-1105668 [013] d... 555825.180261: : Hello, World!           <...>-1105668 [013] d... 555825.180339: : Hello, World!           <...>-1105681 [007] d... 555825.187444: : Hello, World!           <...>-1105681 [007] d.
```

```
$ sudo ./CPUDistribution
Probing for 10 seconds
CPU  0 worked for 465.048 ms.
CPU  1 worked for 2083.19 ms.
CPU  2 worked for 494.222 ms.
CPU  3 worked for 438.929 ms.
CPU  4 worked for 520.142 ms.
CPU  5 worked for 484.19 ms.
CPU  6 worked for 321.532 ms.
CPU  7 worked for 418.621 ms.
CPU  8 worked for 919.519 ms.
CPU  9 worked for 785.386 ms.
CPU 10 worked for 753.927 ms.
CPU 11 worked for 1942.52 ms.
CPU 12 worked for 716.44 ms.
CPU 13 worked for 817.163 ms.
CPU 14 worked for 915.345 ms.
CPU 15 worked for 1357.1 ms.
CPU 16 worked for 16.329 ms.
CPU 17 worked for 9.93461 ms.
CPU 18 worked for 8.63813 ms.
CPU 19 worked for 12.4068 ms.
CPU 20 worked for 41.3855 ms.
CPU 21 worked for 14.8725 ms.
CPU 22 worked for 5.20463 ms.
CPU 23 worked for 17.3686 ms.
CPU 24 worked for 15.5483 ms.
CPU 25 worked for 59.5226 ms.
CPU 26 worked for 114.46 ms.
CPU 27 worked for 285.454 ms.
CPU 28 worked for 22.2104 ms.
CPU 29 worked for 20.6931 ms.
CPU 30 worked for 11.7063 ms.
CPU 31 worked for 20.1535 ms.
```

```
$ sudo ./TCPSendStack 1
Probing for 1 seconds
[...]
PID: 5357 (IOThreadPool0) made 6 TCP sends on following stack:
  Kernel Stack:
    tcp_sendmsg
    do_sock_sendmsg
    ___sys_sendmsg
    __sys_sendmsg
    sys_sendmsg
    system_call_fastpath
  User Stack:
    [UNKNOWN]
    folly::AsyncSocket::performWrite(iovec const*, unsigned int, folly::WriteFlags, unsigned int*, unsigned int*)
    folly::AsyncSSLSocket::performWrite(iovec const*, unsigned int, folly::WriteFlags, unsigned int*, unsigned int*)
    folly::AsyncSocket::writeImpl(folly::AsyncWriter::WriteCallback*, iovec const*, unsigned long, std::unique_ptr<folly::IOBuf, std::default_delete<folly::IOBuf> >&&, folly::WriteFlags)
    folly::AsyncSocket::writeChain(folly::AsyncWriter::WriteCallback*, std::unique_ptr<folly::IOBuf, std::default_delete<folly::IOBuf> >&&, folly::WriteFlags)
    apache::thrift::TAsyncTransportHandler::write(wangle::HandlerContext<folly::IOBufQueue&, std::unique_ptr<folly::IOBuf, std::default_delete<folly::IOBuf> > >*, std::unique_ptr<folly::IOBuf, std::default_delete<folly::IOBuf> >)
    wangle::ContextImpl<apache::thrift::TAsyncTransportHandler>::write(std::unique_ptr<folly::IOBuf, std::default_delete<folly::IOBuf> >)
    wangle::OutboundContextImpl<wangle::OutputBufferingHandler>::fireWrite(std::unique_ptr<folly::IOBuf, std::default_delete<folly::IOBuf> >)
    wangle::OutputBufferingHandler::runLoopCallback()
    folly::EventBase::runLoopCallbacks()
    folly::EventBase::loopBody(int)
    folly::EventBase::loopForever()
    wangle::IOThreadPoolExecutor::threadRun(std::shared_ptr<wangle::ThreadPoolExecutor::Thread>)
    void folly::detail::function::FunctionTraits<void ()>::callBig<std::_Bind<std::_Mem_fn<void (wangle::ThreadPoolExecutor::*)(std::shared_ptr<wangle::ThreadPoolExecutor::Thread>)> (wangle::ThreadPoolExecutor*, std::shared_ptr<wangle::ThreadPoolExecutor::Thread>)> >(folly::detail::function::Data&)
    execute_native_thread_routine
    start_thread
    __clone
```

```
$ sudo ./RecordMySQLQuery [...]/mysqld 25
Using mysqld path: [...]/mysqld
Probing for 25 seconds
9 queries recorded:
Time: 556598409159660 PID: 723991 Query: SET SQL_SAFE_UPDATES=1,SQL_SELECT_LIMIT=1000,MAX_JOIN_SIZE=1000000
Time: 556598425292767 PID: 723991 Query: select USER()
Time: 556602820110928 PID: 723991 Query: show processlist
Time: 556605435971397 PID: 723991 Query: show databases
Time: 556607386278462 PID: 539142 Query: /* /usr/local/bin/dbatool:base.pyc */ SHOW GLOBAL VARIABLES
Time: 556607410458669 PID: 539264 Query: /* /usr/local/bin/dbatool:base.pyc */ SHOW GLOBAL VARIABLES
Time: 556609363724478 PID: 723991 Query: SELECT DATABASE()
Time: 556611171660239 PID: 723991 Query: show tables
Time: 556614523422863 PID: 723991 Query: describe INNODB_FT_CONFIG
```

```
$ sudo ./RandomRead
Started tracing, hit Ctrl-C to terminate.
PID: 5830 (SREventBase3) Read 128 bits
PID: 5831 (SREventBase4) Read 128 bits
PID: 5828 (SREventBase1) Read 96 bits
PID: 5829 (SREventBase2) Read 128 bits
PID: 5830 (SREventBase3) Read 128 bits
PID: 5831 (SREventBase4) Read 96 bits
PID: 6079 (SREventBase3) Read 96 bits
PID: 6132 (sasl-thread-32) Read 32 bits
PID: 6132 (sasl-thread-32) Read 256 bits
PID: 6132 (sasl-thread-32) Read 128 bits
PID: 6079 (SREventBase3) Read 96 bits
PID: 955806 (hhvm.node.0) Read 24 bits
PID: 955806 (hhvm.node.0) Read 16 bits
PID: 955806 (hhvm.node.0) Read 24 bits
PID: 955806 (hhvm.node.0) Read 16 bits
PID: 949264 (SREventBase2) Read 96 bits
PID: 5970 (sasl-thread-21) Read 32 bits
PID: 5970 (sasl-thread-21) Read 256 bits
PID: 5970 (sasl-thread-21) Read 128 bits
PID: 5979 (sasl-thread-30) Read 32 bits
```

### DIFF
--- a/examples/cpp/CPUDistribution.cc
+++ b/examples/cpp/CPUDistribution.cc
@@ -1,0 +1,97 @@
+/*
+ * CPUDistribution Show load distribution across CPU cores during a period of
+ *                 time. For Linux, uses BCC, eBPF. Embedded C.
+ *
+ * Basic example of BCC and kprobes.
+ *
+ * USAGE: CPUDistribution [duration]
+ *
+ * Copyright (c) Facebook, Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ */
+
+#include <unistd.h>
+#include <cstdlib>
+#include <iomanip>
+#include <iostream>
+#include <string>
+
+#include "BPF.h"
+
+const std::string BPF_PROGRAM = R"(
+#include <linux/sched.h>
+#include <uapi/linux/ptrace.h>
+
+BPF_HASH(pid_to_cpu, pid_t, int);
+BPF_HASH(pid_to_ts, pid_t, uint64_t);
+BPF_HASH(cpu_time, int, uint64_t);
+
+int task_switch_event(struct pt_regs *ctx, struct task_struct *prev) {
+  pid_t prev_pid = prev->pid;
+  int* prev_cpu = pid_to_cpu.lookup(&prev_pid);
+  uint64_t* prev_ts = pid_to_ts.lookup(&prev_pid);
+
+  pid_t cur_pid = bpf_get_current_pid_tgid();
+  int cur_cpu = bpf_get_smp_processor_id();
+  uint64_t cur_ts = bpf_ktime_get_ns();
+
+  uint64_t this_cpu_time = 0;
+  if (prev_ts) {
+    pid_to_ts.delete(&prev_pid);
+    this_cpu_time = (cur_ts - *prev_ts);
+  }
+  if (prev_cpu) {
+    pid_to_cpu.delete(&prev_pid);
+    if (this_cpu_time > 0) {
+      int cpu_key = *prev_cpu;
+      uint64_t* history_time = cpu_time.lookup(&cpu_key);
+      if (history_time)
+        this_cpu_time += *history_time;
+      cpu_time.update(&cpu_key, &this_cpu_time);
+    }
+  }
+
+  pid_to_cpu.update(&cur_pid, &cur_cpu);
+  pid_to_ts.update(&cur_pid, &cur_ts);
+
+  return 0;
+}
+)";
+
+int main(int argc, char** argv) {
+  ebpf::BPF bpf;
+  auto init_res = bpf.init(BPF_PROGRAM);
+  if (std::get<0>(init_res) != 0) {
+    std::cerr << std::get<1>(init_res) << std::endl;
+    return 1;
+  }
+
+  auto attach_res =
+      bpf.attach_kprobe("finish_task_switch", "task_switch_event");
+  if (std::get<0>(attach_res) != 0) {
+    std::cerr << std::get<1>(attach_res) << std::endl;
+    return 1;
+  }
+
+  int probe_time = 10;
+  if (argc == 2) {
+    probe_time = atoi(argv[1]);
+  }
+  std::cout << "Probing for " << probe_time << " seconds" << std::endl;
+  sleep(probe_time);
+
+  auto table = bpf.get_hash_table<int, uint64_t>("cpu_time");
+  auto num_cores = sysconf(_SC_NPROCESSORS_ONLN);
+  for (int i = 0; i < num_cores; i++) {
+    std::cout << "CPU " << std::setw(2) << i << " worked for ";
+    std::cout << (table[i] / 1000000.0) << " ms." << std::endl;
+  }
+
+  auto detach_res = bpf.detach_kprobe("finish_task_switch");
+  if (std::get<0>(detach_res) != 0) {
+    std::cerr << std::get<1>(detach_res) << std::endl;
+    return 1;
+  }
+
+  return 0;
+}
--- a/examples/cpp/HelloWorld.cc
+++ b/examples/cpp/HelloWorld.cc
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) Facebook, Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ */
+
+#include <unistd.h>
+#include <fstream>
+#include <iostream>
+#include <string>
+
+#include "BPF.h"
+
+const std::string BPF_PROGRAM = R"(
+int on_sys_clone(void *ctx) {
+  bpf_trace_printk("Hello, World! Here I did a sys_clone call!\n");
+  return 0;
+}
+)";
+
+int main() {
+  ebpf::BPF bpf;
+  auto init_res = bpf.init(BPF_PROGRAM);
+  if (std::get<0>(init_res) != 0) {
+    std::cerr << std::get<1>(init_res) << std::endl;
+    return 1;
+  }
+
+  std::ifstream pipe("/sys/kernel/debug/tracing/trace_pipe");
+  std::string line;
+
+  auto attach_res = bpf.attach_kprobe("sys_clone", "on_sys_clone");
+  if (std::get<0>(attach_res) != 0) {
+    std::cerr << std::get<1>(attach_res) << std::endl;
+    return 1;
+  }
+
+  while (true) {
+    if (std::getline(pipe, line)) {
+      std::cout << line << std::endl;
+      // Detach the probe if we got at least one line.
+      auto detach_res = bpf.detach_kprobe("sys_clone");
+      if (std::get<0>(detach_res) != 0) {
+        std::cerr << std::get<1>(detach_res) << std::endl;
+        return 1;
+      }
+      break;
+    } else {
+      std::cout << "Waiting for a sys_clone event" << std::endl;
+      sleep(1);
+    }
+  }
+
+  return 0;
+}
--- a/src/cc/BPF.cc
+++ b/src/cc/BPF.cc
@@ -1,0 +1,397 @@
+/*
+ * Copyright (c) 2016 Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <linux/bpf.h>
+#include <unistd.h>
+#include <cstdio>
+#include <exception>
+#include <iostream>
+#include <memory>
+#include <sstream>
+#include <vector>
+
+#include "bcc_syms.h"
+#include "bpf_module.h"
+#include "common.h"
+#include "exception.h"
+#include "libbpf.h"
+#include "perf_reader.h"
+
+#include "BPF.h"
+
+namespace ebpf {
+
+std::string uint_to_hex(uint64_t value) {
+  std::stringstream ss;
+  ss << std::hex << value;
+  return ss.str();
+}
+
+std::string sanitize_str(std::string str, bool (*validator)(char),
+                         char replacement = '_') {
+  for (size_t i = 0; i < str.length(); i++)
+    if (!validator(str[i]))
+      str[i] = replacement;
+  return str;
+}
+
+StatusTuple BPF::init(const std::string& bpf_program,
+                      std::vector<std::string> cflags) {
+  auto flags_len = cflags.size();
+  const char* flags[flags_len];
+  for (size_t i = 0; i < flags_len; i++)
+    flags[i] = cflags[i].c_str();
+  if (bpf_module_->load_string(bpf_program, flags, flags_len) != 0)
+    return mkstatus(-1, "Unable to initialize BPF program");
+  return mkstatus(0);
+};
+
+BPF::~BPF() {
+  auto res = detach_all();
+  if (std::get<0>(res) != 0)
+    std::cerr << "Failed to detach all probes on destruction: " << std::endl
+              << std::get<1>(res) << std::endl;
+}
+
+StatusTuple BPF::detach_all() {
+  bool has_error = false;
+  std::string error_msg;
+
+  for (auto it : kprobes_) {
+    auto res = detach_kprobe_event(it.first, it.second);
+    if (std::get<0>(res) != 0) {
+      error_msg += "Failed to detach kprobe event " + it.first + ": ";
+      error_msg += std::get<1>(res) + "\n";
+      has_error = true;
+    }
+  }
+
+  for (auto it : uprobes_) {
+    auto res = detach_uprobe_event(it.first, it.second);
+    if (std::get<0>(res) != 0) {
+      error_msg += "Failed to detach uprobe event " + it.first + ": ";
+      error_msg += std::get<1>(res) + "\n";
+      has_error = true;
+    }
+  }
+
+  for (auto it : tracepoints_) {
+    auto res = detach_tracepoint_event(it.first, it.second);
+    if (std::get<0>(res) != 0) {
+      error_msg += "Failed to detach Tracepoint " + it.first + ": ";
+      error_msg += std::get<1>(res) + "\n";
+      has_error = true;
+    }
+  }
+
+  for (auto it : perf_buffers_) {
+    auto res = it.second->close();
+    if (std::get<0>(res) != 0) {
+      error_msg += "Failed to close perf buffer " + it.first + ": ";
+      error_msg += std::get<1>(res) + "\n";
+      has_error = true;
+    }
+    delete it.second;
+  }
+
+  if (has_error)
+    return mkstatus(-1, error_msg);
+  else
+    return mkstatus(0);
+}
+
+StatusTuple BPF::attach_kprobe(const std::string& kernel_func,
+                               const std::string& probe_func,
+                               bpf_attach_type attach_type,
+                               pid_t pid, int cpu, int group_fd,
+                               perf_reader_cb cb, void* cb_cookie) {
+  int probe_fd;
+  TRY2(load_func(probe_func, BPF_PROG_TYPE_KPROBE, probe_fd));
+
+  std::string probe_event = get_kprobe_event(kernel_func, attach_type);
+  if (kprobes_.find(probe_event) != kprobes_.end()) {
+    TRY2(unload_func(probe_func));
+    return mkstatus(-1, "kprobe %s already attached", probe_event.c_str());
+  }
+
+  std::string probe_event_desc = attach_type_prefix(attach_type);
+  probe_event_desc += ":kprobes/" + probe_event + " " + kernel_func;
+
+  void* res =
+      bpf_attach_kprobe(probe_fd, probe_event.c_str(), probe_event_desc.c_str(),
+                        pid, cpu, group_fd, cb, cb_cookie);
+
+  if (!res) {
+    TRY2(unload_func(probe_func));
+    return mkstatus(-1, "Unable to attach %skprobe for %s using %s",
+                    attach_type_debug(attach_type).c_str(), kernel_func.c_str(),
+                    probe_func.c_str());
+  }
+
+  open_probe_t p = {};
+  p.reader_ptr = res;
+  p.func = probe_func;
+  kprobes_[probe_event] = std::move(p);
+  return mkstatus(0);
+}
+
+StatusTuple BPF::attach_uprobe(const std::string& binary_path,
+                               const std::string& symbol,
+                               const std::string& probe_func,
+                               uint64_t symbol_addr,
+                               bpf_attach_type attach_type,
+                               pid_t pid, int cpu, int group_fd,
+                               perf_reader_cb cb, void* cb_cookie) {
+  int probe_fd;
+  TRY2(load_func(probe_func, BPF_PROG_TYPE_KPROBE, probe_fd));
+
+  bcc_symbol sym = bcc_symbol();
+  TRY2(check_binary_symbol(binary_path, symbol, symbol_addr, &sym));
+
+  std::string probe_event =
+      get_uprobe_event(sym.module, sym.offset, attach_type);
+  if (uprobes_.find(probe_event) != uprobes_.end()) {
+    TRY2(unload_func(probe_func));
+    return mkstatus(-1, "uprobe %s already attached", probe_event.c_str());
+  }
+
+  std::string probe_event_desc = attach_type_prefix(attach_type);
+  probe_event_desc += ":uprobes/" + probe_event + " ";
+  probe_event_desc += binary_path + ":0x" + uint_to_hex(sym.offset);
+
+  void* res =
+      bpf_attach_uprobe(probe_fd, probe_event.c_str(), probe_event_desc.c_str(),
+                        pid, cpu, group_fd, cb, cb_cookie);
+
+  if (!res) {
+    TRY2(unload_func(probe_func));
+    return mkstatus(
+        -1,
+        "Unable to attach %suprobe for binary %s symbol %s addr %lx using %s\n",
+        attach_type_debug(attach_type).c_str(), binary_path.c_str(),
+        symbol.c_str(), symbol_addr, probe_func.c_str());
+  }
+
+  open_probe_t p = {};
+  p.reader_ptr = res;
+  p.func = probe_func;
+  uprobes_[probe_event] = std::move(p);
+  return mkstatus(0);
+}
+
+StatusTuple BPF::attach_tracepoint(const std::string& tracepoint,
+                                   const std::string& probe_func,
+                                   pid_t pid, int cpu, int group_fd,
+                                   perf_reader_cb cb, void* cb_cookie) {
+  int probe_fd;
+  TRY2(load_func(probe_func, BPF_PROG_TYPE_TRACEPOINT, probe_fd));
+
+  if (tracepoints_.find(tracepoint) != tracepoints_.end())
+    return mkstatus(-1, "Tracepoint %s already attached", tracepoint.c_str());
+
+  auto pos = tracepoint.find(":");
+  if ((pos == std::string::npos) || (pos != tracepoint.rfind(":")))
+    return mkstatus(-1, "Unable to parse Tracepoint %s", tracepoint.c_str());
+  std::string tp_category = tracepoint.substr(0, pos);
+  std::string tp_name = tracepoint.substr(pos + 1);
+
+  void* res =
+      bpf_attach_tracepoint(probe_fd, tp_category.c_str(), tp_name.c_str(), pid,
+                            cpu, group_fd, cb, cb_cookie);
+
+  if (!res) {
+    TRY2(unload_func(probe_func));
+    return mkstatus(-1, "Unable to attach Tracepoint %s using %s",
+                    tracepoint.c_str(), probe_func.c_str());
+  }
+
+  open_probe_t p = {};
+  p.reader_ptr = res;
+  p.func = probe_func;
+  tracepoints_[tracepoint] = std::move(p);
+  return mkstatus(0);
+}
+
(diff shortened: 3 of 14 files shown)


examples/cpp/HelloWorld.cc
```diff
@@ -0,0 +1,38 @@
...
+int main() {
+  std::cout << "The BPF Program is: " << std::endl << BPF_PROGRAM;
+  ebpf::BPF bpf(BPF_PROGRAM);
+
+  if (!bpf.attach_kprobe("sys_clone", "probe_sys_clone")) {
+    std::cout << "Error" << std::endl;
+    return 1;
+  }
+
+  auto f = fopen("/sys/kernel/debug/tracing/trace_pipe", "r");
```
goldshtn: I thought it might actually be nice to abstract this away as well, the way the Python module does. Not necessarily parse the fields, but at least just hide the details with opening the file and reading from it.

examples/cpp/CPUDistribution.cc
```diff
@@ -0,0 +1,85 @@
...
+  sleep(probe_time);
+
+  auto table = bpf.get_hash_table<int, uint64_t>("cpu_time");
+  auto num_cores = sysconf(_SC_NPROCESSORS_ONLN);
+  for (int i = 0; i < num_cores; i++) {
+    std::cout << "CPU " << std::setw(2) << i << " worked for ";
+    std::cout << (table[i] / 1000000.0) << " ms." << std::endl;
+  }
+
+  bpf.detach_kprobe("finish_task_switch");
```
goldshtn: I was about to ask if this can this be done in the destructor automatically, but then I noticed that the destructor does it already. So why is this `detach_kprobe` call required here?

src/cc/BPF.cc
```diff
@@ -0,0 +1,377 @@
...
+
+namespace ebpf {
+
+  const static std::regex KPROBE_EV_NAME_REPLACE("[+.]");
+  const static std::regex UPROBE_EV_PATH_REPLACE("[^a-zA-Z0-9_]");
+
+  BPF::~BPF() {
+    for (auto it : kprobes_)
+      if (!detach_kprobe_event(it.first, it.second))
+        fprintf(
```
goldshtn: `std::cerr`?

src/cc/BPF.cc
```diff
@@ -0,0 +1,377 @@
...
+          stderr, "Failed to detach uprobe event %s in BPF destructor\n",
+          it.first.c_str()
+        );
+    for (auto it : tracepoints_)
+      if (!detach_tracepoint_event(it.first, it.second))
+        fprintf(
+          stderr, "Failed to detach Tracepoint %s in BPF destructor\n",
+          it.first.c_str()
+        );
+    delete bpf_module_;
```
goldshtn: Should probably be a smart pointer, and then you won't need this `delete` call here.

src/cc/BPF.cc
```diff
@@ -0,0 +1,377 @@
...
+  }
+
+  bool BPF::attach_kprobe(
```
goldshtn: I don't know if I like this error management strategy. Printing errors to `stderr` and returning `false` isn't very composable in a library. I think I'd actually prefer exceptions.

src/cc/BPF.cc
```diff
@@ -0,0 +1,377 @@
...
+        kernel_func.c_str(), probe_func.c_str()
+      );
+      unload_func(probe_func);
+      return false;
+    }
+
+    open_probe_t p = {};
+    p.reader_ptr = res;
+    p.func = probe_func;
+    kprobes_[probe_event] = p;
```
goldshtn: `std::move(p)`

src/cc/BPF.h
```diff
@@ -0,0 +1,164 @@
...
+#include <string>
+
+#include "bcc_syms.h"
+#include "bpf_module.h"
+#include "BPFTable.h"
+#include "libbpf.h"
+
+namespace ebpf {
+
+enum bpf_attach_type {
```
goldshtn: `enum class`?

src/cc/BPFTable.cc
```diff
@@ -0,0 +1,84 @@
...
+  std::vector<std::string> res;
+  res.reserve(addresses.size());
+
+  if (pid < 0)
+    pid = -1;
+  if (pid_sym_.find(pid) == pid_sym_.end())
+    pid_sym_[pid] = bcc_symcache_new(pid);
+  void* cache = pid_sym_[pid];
+
+  bcc_symbol* symbol = new bcc_symbol();
```
goldshtn: Why not a stack object?

src/cc/BPFTable.h
```diff
@@ -0,0 +1,110 @@
...
+    return get_value(key);
+  }
+
+  std::vector<std::pair<KeyType, ValueType>> get_table_offline() {
+    std::vector<std::pair<KeyType, ValueType>> res;
+    res.reserve(max_size());
+
+    KeyType* key = new KeyType();
+    KeyType* next_key = new KeyType();
+    ValueType* value = new ValueType();
```
goldshtn: Why not stack objects?

examples/cpp/HelloWorld.cc
palmtenor: Makes sense. This was my first `HelloWorld` Demo to make sure things work, so it is pretty primitive. Will try to improve it in the future.

examples/cpp/CPUDistribution.cc
palmtenor: You are right, the destructor already does it. Just want to demonstrate what should be called:)

src/cc/BPF.cc
palmtenor: 👍

src/cc/BPF.cc
palmtenor: Talked about this with @4ast and we want to avoid using `shared_ptr` for build size. It feels a bit weird if I use `unique_ptr` here but then pass the underlying raw pointer around (for tables etc.)

src/cc/BPFTable.cc
palmtenor: What do you mean by a stack object?

palmtenor:
@goldshtn Thanks a lot for review. I talked with @4ast about error handling in the beginning. Since the underlying `libbcc` is compiled without RTTI and have documented behavior of not throwing exceptions, he thought it would be good to keep it the same way for C++ interface (I'm cleaning up the remaining two throwing call sites). We can discuss this here, I'm OK either way:)

examples/cpp/CPUDistribution.cc
goldshtn: Sure, but if this is the example -- people will assume they have to call it. That's the way of examples 😄

src/cc/BPF.cc
goldshtn: I mean, smart pointers are for expressing ownership. If you don't have or share ownership over an object, raw pointers are fine. In this case, I think the only object with ownership of the `bpf_module_` is the `BPF` object.

src/cc/BPFTable.cc
goldshtn: I mean, why allocate a `bcc_symbol` with `new` and not just pass a pointer to a local variable allocated on the stack?

goldshtn:
OK, I'm perfectly fine with the decision to not throw exceptions, but then we need to figure out a way to return errors to the client without printing them to `stderr`.

palmtenor:
For the build failure... I guess no GCC4.9 for ubuntu1404?

src/cc/BPFTable.cc
palmtenor: Ah, makes sense:)

4ast:
re: exceptions...
I think API shouldn't be using exceptions as a mechanism to pass errors. I would say that even non-api class shouldn't be using them either. I like golang approach in that regard. Something like parser with hairy stack unwind that will be ugly without exceptions is a valid case for exceptions. Error return is not.
For bcc c++ api we can use golang style of errors where the error code and the error string are passed to the caller. Cleaner than printing to stderr.
To avoid exceptions in constructor, we can have private constructor and make_bpf or create_bpf static method that returns (err_code, err_string) pair.
cc @goldshtn @drzaeus77 @palmtenor

drzaeus77:
@palmtenor gcc 4.8.4 :(

Also agree on the no-exceptions bit. In the codegen for the "b" language, we use a few helper functions to return `std::tuple<int, std::string>`, which I found to be quite easy to use. Maybe take a look at how that is used, and tell us what you think about the resulting API?

See src/cc/exception.h, and src/cc/common.h for `typedef std::tuple<int, std::string> StatusTuple`. and src/cc/frontends/b/codegen_llvm.cc for the heavy user.

palmtenor:
Some style upgrades according to suggestions.

Thanks a lot for the input on exception handling. Will work on revise them soon. The `StatusTuple` thing does look reasonable.

palmtenor:
ubuntu1404 doesn't even have `linux/bpf.h`? :( :( :(

src/cc/BPF.cc
```diff
@@ -0,0 +1,398 @@
...
+  }
+
+  BPF::BPF(
+    const std::string& bpf_program,
+    unsigned int debug_flag, const char* cflags[], int ncflags
+  ) :
+    bpf_module_(new BPFModule(debug_flag))
+  {
+    if (bpf_module_->load_string(bpf_program, cflags, ncflags) != 0)
+      throw std::runtime_error("Unable to initialize BPF program");
```
jsedgwick: would be nice for there to be a map of error numbers to messages somewhere and to throw to msg in with the exception

bonus points if there's a BPFException class

src/cc/BPF.cc
```diff
@@ -0,0 +1,398 @@
...
+  };
+
+  BPF::~BPF() {
+    for (auto it : kprobes_)
```
jsedgwick: i'll leave general style guidelines to repo owners but as you know the FB way is to always enclose these scopes in braces, for IMO good reasons (see 'goto fail')

src/cc/BPF.h
```diff
@@ -0,0 +1,162 @@
...
+  probe_entry,
+  probe_return
+};
+
+struct open_probe_t {
+  void* reader_ptr;
+  std::string func;
+};
+
+class BPF {
```
jsedgwick: I'd recommend running clang-format on everything in this commit for stylistic consistency

src/cc/BPFTable.h
```diff
@@ -0,0 +1,105 @@
...
+ public:
+  size_t max_size() {
+    return max_size_;
+  }
+
+ protected:
+  BPFTableBase(BPFModule* bpf_module, const std::string& name);
+
+  bool lookup(void* key, void* value);
+  bool next(void* key, void* next_key);
```
jsedgwick: include an iterator class based on this?

src/cc/BPFTable.h
```diff
@@ -0,0 +1,105 @@
...
+#include <utility>
+#include <vector>
+
+#include "bpf_module.h"
+
+namespace ebpf {
+
+class BPFTableBase {
+ public:
+  size_t max_size() {
```
jsedgwick: capacity()?

src/cc/BPFTable.h
```diff
@@ -0,0 +1,105 @@
...
+    return res;
+  }
+
+  ValueType operator[](const KeyType& key) {
+    return get_value(key);
+  }
+
+  std::vector<std::pair<KeyType, ValueType>> get_table_offline() {
+    std::vector<std::pair<KeyType, ValueType>> res;
+    res.reserve(max_size());
```
jsedgwick: i would remove this and just let auto-resize do the right thing

src/cc/BPFTable.h
```diff
@@ -0,0 +1,105 @@
...
+
+    KeyType cur, nxt;
+    ValueType value;
+
+    while (true) {
```
jsedgwick: this is just an internal example of where an iterator class would come in handy

src/cc/BPFTable.h
```diff
@@ -0,0 +1,105 @@
...
+class BPFStackTable : protected BPFTableBase {
+ public:
+  BPFStackTable(BPFModule* bpf_module, const std::string& name)
+    : BPFTableBase(bpf_module, name) {}
+  ~BPFStackTable();
+
+  std::vector<uint64_t> get_stack_addr(int stack_id);
+  std::vector<std::string> get_stack_symbol(int stack_id, int pid);
+ private:
+  // From src/cc/export/helpers.h
```
jsedgwick: can you just include and use that instead then?

src/cc/BPFTable.h
```diff
@@ -0,0 +1,105 @@
...
+    ValueType value;
+
+    while (true) {
+      if (!next(static_cast<void*>(&cur), static_cast<void*>(&nxt)))
+        break;
+      if (!lookup(static_cast<void*>(&nxt), static_cast<void*>(&value))) {
+        fprintf(stderr, "Failed reading table %d\n", fd_);
+        break;
+      }
+      res.push_back(std::make_pair(nxt, value));
```
jsedgwick: res.emplace_back(std::move(nxt), std::move(value))

src/cc/BPFTable.cc
```diff
@@ -0,0 +1,83 @@
...
+  stacktrace_t stack;
+  if (!lookup(static_cast<void*>(&stack_id), static_cast<void*>(&stack)))
+    return res;
+  int i = 0;
+  while (true) {
+    if ((i == BPF_MAX_STACK_DEPTH) || (stack.ip[i] == 0))
+      break;
+    res.push_back(stack.ip[i]);
+    i++;
+  }
```
jsedgwick: refactor this as a for loop

src/cc/BPFTable.cc
```diff
@@ -0,0 +1,83 @@
...
+  int stack_id, int pid
+) {
+  auto addresses = get_stack_addr(stack_id);
+  std::vector<std::string> res;
+  res.reserve(addresses.size());
+
+  if (pid < 0)
+    pid = -1;
+  if (pid_sym_.find(pid) == pid_sym_.end())
+    pid_sym_[pid] = bcc_symcache_new(pid);
```
jsedgwick: beware pid recycling

src/cc/BPF.cc
```diff
@@ -0,0 +1,398 @@
...
+  bool BPF::attach_kprobe(
+    const std::string& kernel_func,
+    const std::string& probe_func,
+    bpf_attach_type attach_type,
+    pid_t pid, int cpu, int group_fd,
+    perf_reader_cb cb, void* cb_cookie
+  ) {
+    int probe_fd = load_func(probe_func, BPF_PROG_TYPE_KPROBE);
+    if (probe_fd == -1) {
+      fprintf(stderr, "Unable to load probe function %s\n", probe_func.c_str());
```
jsedgwick: you have a strange mix of stderr prints and exceptions... IMO exceptions are the way to go although i see i might be alone in thinking that

src/cc/BPF.cc
```diff
@@ -0,0 +1,398 @@
...
+        kernel_func.c_str(), probe_func.c_str()
+      );
+      unload_func(probe_func);
+      return false;
+    }
+
+    open_probe_t p = {};
+    p.reader_ptr = res;
+    p.func = probe_func;
+    kprobes_[probe_event] = std::move(p);
```
jsedgwick: what if kprobes_[probe_event] already exists?

src/cc/BPF.cc
```diff
@@ -0,0 +1,398 @@
...
+  bool BPF::detach_uprobe_event(const std::string& event, open_probe_t& attr) {
+    if (attr.reader_ptr) {
+      perf_reader_free(attr.reader_ptr);
+      attr.reader_ptr = nullptr;
+    }
+    if (!unload_func(attr.func))
+      return false;
+    std::string detach_event = "-:uprobes/" + event;
+    if (bpf_detach_uprobe(detach_event.c_str()) < 0) {
+      fprintf(stderr, "Unable to detach kprobe %s\n", event.c_str());
```
jsedgwick: s/kprobe/uprobe

palmtenor:
Thanks @jsedgwick. Re iterator: I wanted to do this as well. But there are limited thing we can do with the BPF table, and I can't find a good way of implementing an iterator class that satisfies standard constructible / assignable / swappable constraints:(

src/cc/BPF.cc
palmtenor: 👍

src/cc/BPFTable.h
palmtenor: This is an installed header. If I reference it directly I would have to install that header as well, which would be confusing since that header is to be used **inside** BPF program code...

src/cc/BPFTable.cc
palmtenor: I think `SymbolCache` does handle refresh internally. I would have to check to confirm.

palmtenor:
First pass of refactor exception handling. Working on the rest.
Also address some of the comments above.

palmtenor:
- Fixed Tracepoint

palmtenor:
OK, I feel that I'm putting to much stuff into this one single PR. Let me stop here and ask for another round of review @4ast @goldshtn @brendangregg @drzaeus77 @jsedgwick 

I still have some error handling left to do, constructors, but also functions that suppose to return value (like `get_table`, `get_value`, `[]` opeartor etc.).

I'm also planning to add an iterator class for tables. But maybe I will separate it out to another chance after this lands.

src/cc/BPF.h
```diff
@@ -0,0 +1,171 @@
...
+
+  explicit BPF(
+    const std::string& bpf_program,
+    unsigned int debug_flag = 0,
+    std::vector<std::string> cflags = {}
+  );
+
+  ~BPF();
+
+  StatusTuple attach_kprobe(
```
drarmstr: The use of StatusTuple as the return value for everything leads to awkward C++ syntax for accessing members with hard-coded index numbers.  Can you use a class/struct instead?  

But really, if this is supposed to be a C++ interface, it would be much more appropriate to use exceptions for reporting errors.

src/cc/BPF.cc
```diff
@@ -0,0 +1,418 @@
...
+  const std::string& bpf_program,
+  unsigned int debug_flag,
+  std::vector<std::string> cflags
+) :
+  bpf_module_(new BPFModule(debug_flag))
+{
+  const char* flags[cflags.size()];
+  for (size_t i = 0; i < cflags.size(); i++)
+    flags[i] = cflags[i].c_str();
+  if (bpf_module_->load_string(bpf_program, flags, cflags.size()) != 0)
```
drarmstr: What about just:

    if (bpf_module_->load_string(bpf_program, cflags.c_str(), cflags.size()) != 0)

src/cc/BPF.cc
```diff
@@ -0,0 +1,418 @@
...
+
+  open_probe_t p = {};
+  p.reader_ptr = res;
+  p.func = probe_func;
+  kprobes_[probe_event] = std::move(p);
+  return mkstatus(0);
+}
+
+StatusTuple BPF::attach_uprobe(
+  const std::string& binary_path, const std::string& symbol,
```
4ast: please reformat the whole thing similar to bpf_module.*
bcc doesn't have 80 char limit. Use the following style:
```
int foo(int arg1, int arg2,
        int arg3) {
}
```
which is pretty much google coding style.

src/cc/BPF.cc
```diff
@@ -0,0 +1,418 @@
...
+  if (uprobes_.find(probe_event) != uprobes_.end()) {
+    TRY2(unload_func(probe_func));
+    return mkstatus(-1, "uprobe %s already attached", probe_event.c_str());
+  }
+
+  std::string probe_event_desc = attach_type_prefix(attach_type);
+  probe_event_desc += ":uprobes/" + probe_event + " ";
+  probe_event_desc += binary_path + ":0x" + uint_to_hex(sym.offset);
+
+  void* res = bpf_attach_uprobe(
```
4ast: same here. Please reformat.

src/cc/BPF.cc
```diff
@@ -0,0 +1,418 @@
...
+    flags[i] = cflags[i].c_str();
+  if (bpf_module_->load_string(bpf_program, flags, cflags.size()) != 0)
+    throw std::runtime_error("Unable to initialize BPF program");
+};
+
+BPF::~BPF() {
+  for (auto it : kprobes_) {
+    auto res = detach_kprobe_event(it.first, it.second);
+    if (std::get<0>(res) != 0)
+      std::cerr << "Failed to detach kprobe event " << it.first
```
palmtenor: The idea here is that if we are failing in destructor, then there would be nothing else we could do, so I'd rather inform the user to let them know. But I see your point, will move this out so people could do clean up whenever they want.

palmtenor:
- Construction and destruction changes according to @4ast 
- `clang-format` on all newly added code. Manually fixed some style discrepancy. I think it would be good if we have a `.clang-format` file in the repository so that contributors could just run a script before committing:)

drzaeus77:
We do have a .clang-format file.
```
$ git log --oneline .clang-format
0865f94 usdt: Implement probe resolution [WIP]
d3d2a57 Add clang-format file, tweak style-check.sh
```

Also, when adding new code, you can run ctest -I 1,1 -V to see what clang-format thinks about the changes between your HEAD and master.
```
$ ctest -I 1,1 -V
test 1
    Start 1: style-check
...
1: old tree: 95ca49877d0525992e486b4955499fd88de50ed1
1: new tree: 86793b72e873d60e1a2fd8769d15c84776daca76
1/1 Test #1: style-check ......................   Passed    0.10 sec
```

Initially this test was gating, but people complained and the test was ignored. So, the code quality has generally deteriorated since.

palmtenor:
@drzaeus77 Thanks! I think my checkout was a bit broken previously so I didn't noticed that. Looks like a lot of files doesn't pass `clang-format` now:(

src/cc/BPF.cc
```diff
@@ -0,0 +1,403 @@
...
+  return str;
+}
+
+BPF::BPF(const std::string& bpf_program, unsigned int debug_flag,
+         std::vector<std::string> cflags)
+    : bpf_module_(new BPFModule(debug_flag)),
+      bpf_program_(bpf_program),
+      cflags_(cflags) {}
+
+StatusTuple BPF::init() {
```
4ast: could you move 'bpf_program' string into init() method instead of constructor ?
The constructor can act on flags that can be applied immediately,
whereas storing 'bpf_program' string just to be used in init() method is suspicious.
The user of the api will be left to wonder 'what's going on with that string during constructor'...

src/cc/BPF.h
```diff
@@ -0,0 +1,153 @@
...
+  template <class KeyType, class ValueType>
+  BPFHashTable<KeyType, ValueType> get_hash_table(const std::string& name) {
+    return BPFHashTable<KeyType, ValueType>(bpf_module_.get(), name);
+  }
+
+  BPFStackTable get_stack_table(const std::string& name) {
+    return BPFStackTable(bpf_module_.get(), name);
+  }
+
+  StatusTuple open_perf_buffer(const std::string& name, perf_reader_raw_cb cb,
```
4ast: I'm not a fan of the inband signaling with cpu = -1. Especially when further down close(-1) is used.
What is the use case of specifying a cpu here? Can open_on_cpu() be private and we always create ring buffer on all cpus by default? If later some use case pops up that needs exact cpu number, we can add another explicit method for that.

palmtenor:
- Move `bpf_program` parameter to `init()`
- Perf buffer `open` and `close` do not take `cpu` as argument anymore, until we have use case for that in the future.

4ast:

4ast:
@palmtenor 
sudo ./HelloWorld
bpf: Invalid argument
H�9"�
Failed to load on_sys_clone: -1

sudo ./hello_world.py
bpf: Invalid argument
H�md

not your fault, of course.
Could you please investigate the source of that junk ?

palmtenor:
@4ast I noticed similar failure message when not running under root. Will try to find out where it comes from. What caused your failure?

PR #794, opened 2016-10-30
Fix argument type for increment() in documentation

pchaigno:
`map.increment()` actually takes the key itself and not a reference to it.

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -509,7 +509,7 @@ Examples in situ:
 
 ### 10. map.increment()
 
-Syntax: ```map.increment(&key)```
+Syntax: ```map.increment(key)```
 
 Increments the key's value by one. Used for histograms.
 


docs/reference_guide.md
```diff
@@ -509,9 +509,9 @@ Examples in situ:
 
 ### 10. map.increment()
 
-Syntax: ```map.increment(&key)```
+Syntax: ```map.increment(key)```
 
-Increments the key's value by one. Used for histograms.
+Increments the key's value by one. Used for histograms. This is actually translated to `lookup()` and `update()` calls.
```
4ast: if we want the doc to clarify such details then it probably needs to be more accurate and say that it's update(&key,&zero,noexist); value=lookup(&key); value++. Otherwise "lookup and update" is confusing.

pchaigno: Hm, you're right. It's confusing and not coherent with the rest of the doc, which doesn't give such details. I'll remove.

drzaeus77:
[buildbot, add to whitelist]

PR #798, opened 2016-11-02
trace: add pid/tid filtering, fix symbolizing, misc nits

markdrayton:
`trace` is an excellent exploratory tool but it has a few shortcomings when dealing with multithreaded applications. This PR fixes up the issues I run into most often.

* support filtering by process ID (-p) or thread ID (-t); previously -p
  actually filtered on thread ID (aka "pid" in kernel-speak)
* include process and thread ID in output
* flip order of user and kernel stacks to flow more naturally
* resolve symbols using process ID instead of thread ID so only one symbol
  cache is instantiated per process
* misc aesthetic fixes here and there

Output before:

```
# trace -UK -p 28047 osq_lock
TIME     PID    COMM         FUNC
10:02:36 28047  load-monitor osq_lock         
    User Stack Trace:
        00000000024e966b get_cpu
        [..snipped..]
        00007fa2739677f1 start_thread
        00007fa272f7846d __clone
    Kernel Stack Trace:
        ffffffff810ac881 osq_lock
        ffffffff8177667b __mutex_lock_slowpath
        ffffffff8177676b mutex_lock
        ffffffff810b34ec kstat_irqs_usr
        ffffffff8122743a show_stat
        ffffffff811db22a seq_read
        ffffffff8121e0fd proc_reg_read
        ffffffff811b8a10 __vfs_read
        ffffffff811b8ad6 vfs_read
        ffffffff811b8bd6 sys_read
        ffffffff81778232 system_call_fastpath
```

After:

```
# trace -UK -t 28047 osq_lock
TIME     PID    TID    COMM         FUNC
10:03:26 28041  28047  load-monitor osq_lock         
        ffffffff810ac881 osq_lock
        ffffffff8177667b __mutex_lock_slowpath
        ffffffff8177676b mutex_lock
        ffffffff810b34ec kstat_irqs_usr
        ffffffff8122743a show_stat
        ffffffff811db22a seq_read
        ffffffff8121e0fd proc_reg_read
        ffffffff811b8a10 __vfs_read
        ffffffff811b8ad6 vfs_read
        ffffffff811b8bd6 sys_read
        ffffffff81778232 system_call_fastpath
        00000000024e966b get_cpu
        [..snipped..]
        00007fa2739677f1 start_thread
        00007fa272f7846d __clone
```

### DIFF
--- a/tools/trace.py
+++ b/tools/trace.py
@@ -3,8 +3,9 @@
 # trace         Trace a function and print a trace message based on its
 #               parameters, with an optional filter.
 #
-# USAGE: trace [-h] [-p PID] [-v] [-Z STRING_SIZE] [-S] [-M MAX_EVENTS] [-o]
-#              [-K] [-U] [-I header] probe [probe ...]
+# usage: trace [-h] [-p PID] [-t TID] [-v] [-Z STRING_SIZE] [-S]
+#              [-M MAX_EVENTS] [-o] [-K] [-U] [-I header]
+#              probe [probe ...]
 #
 # Licensed under the Apache License, Version 2.0 (the "License")
 # Copyright (C) 2016 Sasha Goldshtein.
@@ -51,6 +52,7 @@ class Probe(object):
         event_count = 0
         first_ts = 0
         use_localtime = True
+        tgid = -1
         pid = -1
 
         @classmethod
@@ -58,6 +60,7 @@ class Probe(object):
                 cls.max_events = args.max_events
                 cls.use_localtime = not args.offset
                 cls.first_ts = Time.monotonic_time()
+                cls.tgid = args.tgid or -1
                 cls.pid = args.pid or -1
 
         def __init__(self, probe, string_size, kernel_stack, user_stack):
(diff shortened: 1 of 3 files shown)


tools/trace.py
```diff
@@ -410,14 +418,17 @@ def generate_program(self, include_self):
 
                 text = heading + """
 {
+        u32 __tgid = bpf_get_current_pid_tgid() >> 32;
+        u32 __pid = bpf_get_current_pid_tgid();
```
goldshtn: I don't know if the optimizer can take care of this or not, but I'd probably read into a u64 local and then grab the bottom and top halves from that local.

markdrayton: Ah yes, can do. I need to duck out but will update when I get back.

markdrayton:
> One thing I'm not sure about is if we want to print the tid unconditionally, or only if the user asks.

I commonly find I am trying to narrow something down from systemwide -> process -> thread and having both numbers shown at once makes that easier. We're comfortably within 80 chars unless the function names are huge so my preference is to keep both as always shown.

markdrayton:
Okay, this is ready for review. Thanks!

goldshtn:
LGTM.

PR #804, opened 2016-11-08
lookup perf symbol table generated by perf-map-agent

base698:
Noticed there were lots of unknowns using *offcputime.py* Added a symbol lookup for the process in /tmp/perf-pid.map.

### DIFF
--- a/tools/offcputime.py
+++ b/tools/offcputime.py
@@ -17,6 +17,19 @@ from time import sleep, strftime
 import argparse
 import errno
 import signal
+import os
+
+def get_symbol_hash(perf_map_file):
+    symbol_hash = {}
+    if os.path.isfile(perf_map_file):
+      with open(perf_map_file, 'r') as f:
+        lines = f.readlines()
+        for line in lines: 
+          line = line.split(' ')
+          k = line[0]
+          name = line[2]
+          symbol_hash[k] = name
+    return symbol_hash
 
 # arg validation
 def positive_int(val):
@@ -97,7 +110,6 @@ duration = int(args.duration)
 # signal handler
 def signal_ignore(signal, frame):
     print()
-
 # define BPF program
 bpf_text = """
 #include <uapi/linux/ptrace.h>
@@ -269,13 +281,22 @@ for k, v in sorted(counts.items(), key=lambda counts: counts[1].value):
             [b.ksym(addr) for addr in reversed(kernel_stack)]
         print("%s %d" % (";".join(line), v.value))
     else:
+        symbol_hash = {}
+        # build filename with k.pid
+        # build symbol hash
+        if(args.pid):
+          perf_map_file = "/tmp/perf-%d.map" % args.pid
+          symbol_hash = get_symbol_hash(perf_map_file)
+
         # print default multi-line stack output
         for addr in kernel_stack:
             print("    %016x %s" % (addr, b.ksym(addr)))
         if need_delimiter:
             print("    --")
         for addr in user_stack:
-            print("    %016x %s" % (addr, b.sym(addr, k.tgid)))
+            addr_key = "%016x" % addr
+            name = symbol_hash[addr_key] if symbol_hash.has_key(addr_key) else b.sym(addr, k.tgid)
+            print("    %016x %s" % (addr, name))
         print("    %-16s %s (%d)" % ("-", k.name, k.pid))
         print("        %d\n" % v.value)
 


markdrayton:
Hrm, this should already be working through `BPF.sym()`. I added support for `/tmp/perf-pid.map` to the libbcc code a few months ago (#573). Do you have a simple repro for the issue?

4ast:
cc @markdrayton

markdrayton:
There is one issue with resolving symbols from maps when calling `BPF.sym(addr, tid)` (#805) but it doesn't look like that is the case here — offcputime explicitly calls `sym()` with `k.tgid`, which should be the process ID.

base698:
I think the reason it isn't resolving is this code
https://github.com/iovisor/bcc/blob/master/tools/offcputime.py#L279

The k.pid there is the thread process id.  Maybe if the symbol table build
with sym isn't taking that into account?

Was specifically doing this with Java code using PreserveFramePointer.

markdrayton:
Right, but the symbol lookup happens on the previous line, where `k.tgid` is used:

https://github.com/iovisor/bcc/blob/master/tools/offcputime.py#L278

Can you try running the script with `strace -eopen offcputime ..` and see which `/tmp/perf-*` files are opened, then compare that to the actual name of your map? It's not impossible that there's something wrong with the script or this is related to #805 but on the surface it seems like the whole thing should be working correctly. #805 shouldn't apply here as we use the process ID already.

base698:
I see nothing with tmp or .map as an extension

Just a bunch of files like this:

> open("include/linux/task_io_accounting.h", O_RDONLY) = 7
> open("include/linux/latencytop.h", O_RDONLY) = 7
> open("include/linux/cred.h", O_RDONLY)  = 7
> open("include/linux/key.h", O_RDONLY)   = 7
> open("include/linux/assoc_array.h", O_RDONLY) = 7
> open("include/linux/selinux.h", O_RDONLY) = 7
> open("include/linux/gfp.h", O_RDONLY)   = 7
> open("include/linux/magic.h", O_RDONLY) = -1 ENOENT (No such file or directory)
> open("./include/uapi/linux/magic.h", O_RDONLY) = 7
> open("include/linux/cgroup-defs.h", O_RDONLY) = 7
> open("include/linux/limits.h", O_RDONLY) = -1 ENOENT (No such file or directory)
> open("./include/uapi/linux/limits.h", O_RDONLY) = 7
> open("include/linux/idr.h", O_RDONLY)   = 7
> open("include/linux/percpu-refcount.h", O_RDONLY) = 7
> open("include/linux/percpu-rwsem.h", O_RDONLY) = 7
> open("include/linux/rcu_sync.h", O_RDONLY) = 7
> open("include/linux/cgroup_subsys.h", O_RDONLY) = 7
> open("/sys/kernel/debug/tracing/kprobe_events", O_WRONLY|O_APPEND) = 8
> open("/sys/kernel/debug/tracing/events/kprobes/p_finish_task_switch/id", O_RDONLY) = 8
>

ceeaspb:
Hi, couple more checks:

1)
seeing the last line of strace being 

open("/sys/kernel/debug/tracing/events/kprobes/p_finish_task_switch/id", O_RDONLY) = 8

-> I found that offcputime without a seconds duration parameter meant you control-c'd (did you control-c ?) the execution, which truncated the strace output to not show anything about the perf-map post processing.

could you run the following?:

root@ubuntu-yakkety:~# echo "java is "$(pgrep java); strace offcputime 1 2>&1 |grep perf-

java process is 2736
open("/tmp/perf-2795.map", O_RDONLY)    = -1 ENOENT (No such file or directory)
open("/tmp/perf-2795.map", O_RDONLY)    = -1 ENOENT (No such file or directory)
open("/tmp/perf-2736.map", O_RDONLY)    = 11
open("/tmp/perf-2795.map", O_RDONLY)    = -1 ENOENT (No such file or directory)
open("/tmp/perf-2795.map", O_RDONLY)    = -1 ENOENT (No such file or directory)


2)
"Was specifically doing this with Java code using PreserveFramePointer."

Java symbol resolution needs perf-map-agent.

Can you confirm you also ran the perf-map-agent after starting java and before running offcputime?
https://github.com/jrudolph/perf-map-agent

markdrayton:
Can you also include the `offcputime` output (you can make a gist, https://gist.github.com/) and `ls /tmp/perf-*`?

Does the java process run in a separate PID namespace?

base698:
I ran perf-map-agent, and the paste I made with the strace output was an example.  Will make a gist, hold please ♫ ♫ ♫

base698:
Using:

```
$ strace ./offcputime.py 1 2>&1 | grep perf-
```

I see:
```
open("/tmp/perf-20080.map", O_RDONLY)   = 43
```

I also ran it with -p 20080 and saw it as well.

However: The stacks all have unknown.  

https://gist.github.com/base698/b397ee332a1dc79fb011c281999ade6b

markdrayton:
Hrm. That all looks correct. The map is named after the PID and is opened successfully. Is `00007f3612abfd0b` (unknown) actually in the perf map? Does this all work with your PR?

base698:
A few addresses resolve from the perf map.  This one:  _Lsun/net/spi/DefaultProxySelector$3;::run_ is one example.

I ran it with my pull request and updated the gist (https://gist.github.com/base698/b397ee332a1dc79fb011c281999ade6b) with the stacks from this pull request.

markdrayton:
It's not clear from the two gists which symbols were resolved correctly with your PR — both have some unknowns and some that seem to have resolved. With your PR `Lsun/net/spi/DefaultProxySelector$3;::run` was resolved, but we don't see the corresponding address (`00007f36191085bc`) in the normal output so can't tell if would have worked or not. The workload doesn't seem to be the same across both runs.

Can you come up with an easily reproducible test that proves what does and doesn't work? If not, maybe you can wait for a PR for #805 and see if that helps.

base698:
I ran the same offcpu commands, but reran perf-map-agent in between.

Updated the gist with the Sample Java code I'm trying to profile.  It's
just connecting to a local server that waits a couple of seconds before
sending a response.  Can add that too if need be.

drzaeus77:
[buildbot, ok to test]

markdrayton:
The fix for #805 has been merged so feel free to see if that helps with this issue.

base698:
Didn't seem to fix it.

markdrayton:
Okay, then it seems highly likely the issue is in your environment. Can you check that your perf map actually contains symbols for the unresolved addresses? E.g., in the `my.pull.request.out.http.stacks` part of the gist I see:

```
00007f36191085bc Lsun/net/spi/DefaultProxySelector$3;::run
..
00007f360befc8c0 [unknown]
```

Is there a corresponding symbol in the perf map for `0x00007f360befc8c0`? Is there one for `0x00007f36191085bc` (I assume so, since we resolved that address)? If there's not one for the former then that is your issue.

4ast:
what's the status here? This pr is still necessary or not?

base698:
It doesn't work for me as of Nov 10.  I can try again soon.

4ast:
@markdrayton @brendangregg what's your take?

markdrayton:
As far as I can tell the code for reading /tmp/perf-pid.map should work as intended. If there are problems with it a concrete repro from @base698 would really help to diagnose the issue.

brendangregg:
The existing code works for me, without this PR. Eg:

```
# ./profile.py -p 12049
Sampling at 49 Hertz of PID 12049 by user + kernel stack... Hit Ctrl-C to end.
^C
[...]
    00007fb1591129f4 LFunc_loop;::func_b
    00007fb1591127b4 LFunc_loop;::func_a
    00007fb159110818 LFunc_loop;::main
    00007fb1590004e7 call_stub
    00007fb16dca3c66 JavaCalls::call_helper(JavaValue*, methodHandle*, JavaCallArguments*, Thread*)
    00007fb16dce5072 jni_invoke_static(JNIEnv_*, JavaValue*, _jobject*, JNICallType, _jmethodID*, JNI_ArgumentPusher*, Thread*)
    00007fb16dd018da jni_CallStaticVoidMethod
    00007fb16ebd9bdf JavaMain
    00007fb16edef70a start_thread
    -                java (12049)
        13
```

Ran my program with:

```
java -XX:-Inline -XX:+PreserveFramePointer Func_loop &
```

And then got the perf-map-agent dump using jmaps:  https://raw.githubusercontent.com/brendangregg/Misc/master/java/jmaps

Although I noticed I think the PID is wrong in profile.py: I had to fix it first so that "-p" matched the right thing. I'll do a PR.

goldshtn:
This has gone a bit stale; @base698 is there something we can do to advance this? If not I think this PR can be closed.

4ast:
closing. please reopen if the issue can be reliably reproduced.

PR #808, opened 2016-11-09
cc, python: Clean up BPF module and tables

### LINKED ISSUE #806: bpf-map fd leak
The following program leaks a bpf-map file descriptor every second (verify with `lsof | grep bpf-map`). I've added `None` assignments and even a `gc.collect` call to rule out any possibility of some object not getting reclaimed because of the Python runtime. It looks like we genuinely have a leak somewhere whereby the map fd is not getting closed. 

```
#!/usr/bin/env python

from bcc import BPF
from time import sleep
import gc

prog = """
BPF_HASH(map);

int probe(struct pt_regs *ctx) {
    return 0;
}
"""

while True:
    bpf = BPF(text=prog)
    bpf.attach_kprobe(event="__kmalloc", fn_name="probe")
    m = bpf["map"]
    m.items()
    bpf.cleanup()
    m = None
    bpf = None
    gc.collect()
    sleep(1)
```

/cc @4ast @drzaeus77

goldshtn:
When the `BPFModule` object is destroyed, shared tables
were closed, but non-shared tables were not. Add `close`
call to close non-shared tables in `~BPFModule`.

Make sure the `BPF.cleanup` function in the Python module
calls `bpf_module_destroy` to clean up the module. Otherwise,
we have hanging table fds that aren't destroyed.

Fixes #806. This would have been an issue for any tool
that repeatedly runs a BPF program with new maps. It's just
that most tools we have haven't exhibited this behavior.

### DIFF
--- a/src/cc/bpf_module.cc
+++ b/src/cc/bpf_module.cc
@@ -120,6 +120,8 @@ BPFModule::~BPFModule() {
     for (auto table : *tables_) {
       if (table.is_shared)
         SharedTables::instance()->remove_fd(table.name);
+      else
+        close(table.fd);
     }
   }
 }
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -1049,6 +1049,10 @@ class BPF(object):
             self.detach_perf_event(ev_type, ev_config)
         if self.tracefile:
             self.tracefile.close()
+            self.tracefile = None
+        if self.module:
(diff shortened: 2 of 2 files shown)


dark:
Thanks for the quick turnaround, this indeed fixes the problem I saw while trying out #774

palmtenor:
@goldshtn Do we also need to close the program FDs?

PR #815, opened 2016-11-21
Eliminate rpmlint errors and allow bcc srpm to build on Fedora koji b…

wcohen:
…uild system

The nightly bcc srpm does not build as-is on Fedora 24. To get the
build to work the following changes were made in the bcc.spec file:

-Specify where to find luajit headers and libraries
-Correctly limit the build to the x86_64 architecture
-Add needed build requires so the srpm can build on Fedora koji
-Run ldconfig so shared library information is updated
-Follow spec conventions by putting %changelog at the end of the spec file

### DIFF
--- a/SPECS/bcc.spec
+++ b/SPECS/bcc.spec
@@ -1,3 +1,4 @@
+%bcond_with local_clang_static
 %define debug_package %{nil}
 
 Name:           bcc
@@ -11,12 +12,14 @@ URL:            https://github.com/iovisor/bcc
 Source0:        bcc.tar.gz
 
 ExclusiveArch: x86_64
-BuildRequires: bison cmake >= 2.8.7 flex
+BuildRequires: bison cmake >= 2.8.7 flex make
 BuildRequires: gcc gcc-c++ python2-devel elfutils-libelf-devel-static
 BuildRequires: luajit luajit-devel
+%if %{without local_clang_static}
 BuildRequires: llvm-devel llvm-static
 BuildRequires: clang-devel
-BuildRequires: pkgconfig
+%endif
+BuildRequires: pkgconfig ncurses-devel
 
 %description
 Python bindings for BPF Compiler Collection (BCC). Control a BPF program from
@@ -47,33 +50,40 @@ Requires: elfutils-libelf
 %description -n libbcc
 Shared Library for BPF Compiler Collection (BCC)
 
-%package -n libbcc-examples
-Summary: Examples for BPF Compiler Collection (BCC)
-%description -n libbcc-examples
-Examples for BPF Compiler Collection (BCC)
-
 %package -n python-bcc
 Summary: Python bindings for BPF Compiler Collection (BCC)
+Requires: libbcc = %{version}-%{release}
 %description -n python-bcc
 Python bindings for BPF Compiler Collection (BCC)
 
-%package -n bcc-tools
-Summary: Command line tools for BPF Compiler Collection (BCC)
-Requires: python-bcc
-%description -n bcc-tools
-Command line tools for BPF Compiler Collection (BCC)
-
 %package -n bcc-lua
 Summary: Standalone tool to run BCC tracers written in Lua
+Requires: libbcc = %{version}-%{release}
 %description -n bcc-lua
 Standalone tool to run BCC tracers written in Lua
 
-%files -n python-bcc
-%{python_sitelib}/bcc*
+%package -n libbcc-examples
+Summary: Examples for BPF Compiler Collection (BCC)
+Requires: python-bcc = %{version}-%{release}
+Requires: bcc-lua = %{version}-%{release}
+%description -n libbcc-examples
+Examples for BPF Compiler Collection (BCC)
+
+%package -n bcc-tools
+Summary: Command line tools for BPF Compiler Collection (BCC)
+Requires: python-bcc = %{version}-%{release}
+%description -n bcc-tools
+Command line tools for BPF Compiler Collection (BCC)
 
 %files -n libbcc
 /usr/lib64/*
 /usr/include/bcc/*
+
+%files -n python-bcc
+%{python_sitelib}/bcc*
+
+%files -n bcc-lua
+/usr/bin/bcc-lua
 
 %files -n libbcc-examples
 /usr/share/bcc/examples/*
@@ -87,9 +97,6 @@ Standalone tool to run BCC tracers written in Lua
 %files -n bcc-tools
 /usr/share/bcc/tools/*
 /usr/share/bcc/man/*
-
-%files -n bcc-lua
-/usr/bin/bcc-lua
 
 %post -n libbcc -p /sbin/ldconfig
 
--- a/SPECS/bcc.spec
+++ b/SPECS/bcc.spec
@@ -10,8 +10,13 @@ License:        ASL 2.0
 URL:            https://github.com/iovisor/bcc
 Source0:        bcc.tar.gz
 
-BuildArch:      x86_64
-BuildRequires:  bison, cmake >= 2.8.7, flex, gcc, gcc-c++, python2-devel, elfutils-libelf-devel-static
+ExclusiveArch: x86_64
+BuildRequires: bison cmake >= 2.8.7 flex
+BuildRequires: gcc gcc-c++ python2-devel elfutils-libelf-devel-static
+BuildRequires: luajit luajit-devel
+BuildRequires: llvm-devel llvm-static
+BuildRequires: clang-devel
+BuildRequires: pkgconfig
 
 %description
 Python bindings for BPF Compiler Collection (BCC). Control a BPF program from
@@ -19,32 +24,22 @@ userspace.
 
 
 %prep
-%setup -n bcc
+%setup -q -n bcc
 
 %build
 
 mkdir build
 pushd build
-cmake .. -DREVISION_LAST=%{version} -DREVISION=%{version} -DCMAKE_INSTALL_PREFIX=/usr
+cmake .. -DREVISION_LAST=%{version} -DREVISION=%{version} \
+      -DCMAKE_INSTALL_PREFIX=/usr \
+      -DLUAJIT_INCLUDE_DIR=`pkg-config --variable=includedir luajit` \
+      -DLUAJIT_LIBRARIES=`pkg-config --variable=libdir luajit`/lib`pkg-config --variable=libname luajit`.so
 make %{?_smp_mflags}
 popd
 
 %install
 pushd build
 make install/strip DESTDIR=%{buildroot}
-
-%changelog
-* Mon Apr 04 2016 Vicent Marti <vicent@github.com> - 0.1.4-1
-- Add bcc-lua package
-
-* Sun Nov 29 2015 Brenden Blanco <bblanco@plumgrid.com> - 0.1.3-1
-- Add bcc-tools package
-
-* Mon Oct 12 2015 Brenden Blanco <bblanco@plumgrid.com> - 0.1.2-1
-- Add better version numbering into libbcc.so
-
-* Fri Jul 03 2015 Brenden Blanco <bblanco@plumgrid.com> - 0.1.1-2
-- Initial RPM Release
 
 %package -n libbcc
 Summary: Shared Library for BPF Compiler Collection (BCC)
@@ -54,13 +49,11 @@ Shared Library for BPF Compiler Collection (BCC)
 
 %package -n libbcc-examples
 Summary: Examples for BPF Compiler Collection (BCC)
-Requires: libbcc
 %description -n libbcc-examples
 Examples for BPF Compiler Collection (BCC)
 
 %package -n python-bcc
 Summary: Python bindings for BPF Compiler Collection (BCC)
-Requires: libbcc
 %description -n python-bcc
 Python bindings for BPF Compiler Collection (BCC)
 
@@ -72,7 +65,6 @@ Command line tools for BPF Compiler Collection (BCC)
 
 %package -n bcc-lua
 Summary: Standalone tool to run BCC tracers written in Lua
-Requires: libbcc
 %description -n bcc-lua
 Standalone tool to run BCC tracers written in Lua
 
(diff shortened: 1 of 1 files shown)


drzaeus77:
[buildbot, ok to test]

drzaeus77:
With this change, we now build against the Fedora-supplied llvm/clang libs, which do not provide a static implementation of clang. On the Fedora buildbots, we have built clang/llvm from source with clang-static enabled to workaround this. The .spec file now no longer supports that; can we get a knob to support both? From what I understand, the RedHat style builds shun linking statically, but I consider this a useful feature to have in the specific case of this type of tool (intended to have fewer runtime dependencies, standalone).

wcohen:
The Fedora guidelines discourage static linking:  https://fedoraproject.org/wiki/Packaging:Guidelines#Statically_Linking_Executables

The proposed spec changes were to designed to work with the default clang that is packaged with the stock Fedora. It is possible to have some switches in the spec file to change features in the build for example disable certain features that need rpms or features that are not available in earlier versions of fedora or for particular architectures.  However, I am not sure how that is going to work between the standard fedora build environment and the custom fedora environment with the clang static libraries used for the buildbot.  

Building with the dynamic clang-lib add that dependency and pulls in 29MB.  However, it does make the bcc rpms a bit smaller.  The statically linked nightly ones:

$ rpm -qp ./*bcc* --queryformat '%{SIZE} %{NAME} \n'
35763024 bcc-lua 
1127170 bcc-tools 
35559570 libbcc 
436502 libbcc-examples 
170363 python-bcc 

The dynamically linked build:
$ rpm -qp ./*bcc* --queryformat '%{SIZE} %{NAME} \n'
21337200 bcc-lua 
1127170 bcc-tools 
20580554 libbcc 
436502 libbcc-examples 
170363 python-bcc

About 14.4MB smaller bcc-lua and 15.0MB smaller libbcc rpms for the shared library versions.  Looks kind of a wash in terms of space unless other shared libraries are getting pulled in as a result of the static linking.

wcohen:
It is possible to control the rpmbuild from the command line.  http://people.redhat.com/wcohen/bcc/bcc-bcond.diff puts a %bcond_with local_clang_static to control the rpm depenencies.  To disable rpm dependencies in jenkins build would run with:

rpmbuild -ba bcc.spec --with local_clang_static

Is changing the command that jenkins uses to start the rpm build acceptable?  Or does that need to remain the same?

SPECS/bcc.spec
```diff
@@ -54,13 +49,11 @@ Shared Library for BPF Compiler Collection (BCC)
 
 %package -n libbcc-examples
 Summary: Examples for BPF Compiler Collection (BCC)
-Requires: libbcc
 %description -n libbcc-examples
 Examples for BPF Compiler Collection (BCC)
 
 %package -n python-bcc
 Summary: Python bindings for BPF Compiler Collection (BCC)
-Requires: libbcc
```
drzaeus77: Is this dependency handled magically some other way?

SPECS/bcc.spec
```diff
@@ -10,42 +10,37 @@ License:        ASL 2.0
...
 Source0:        bcc.tar.gz
 
-BuildArch:      x86_64
-BuildRequires:  bison, cmake >= 2.8.7, flex, gcc, gcc-c++, python2-devel, elfutils-libelf-devel-static
+ExclusiveArch: x86_64
+BuildRequires: bison cmake >= 2.8.7 flex
+BuildRequires: gcc gcc-c++ python2-devel elfutils-libelf-devel-static
+BuildRequires: luajit luajit-devel
+BuildRequires: llvm-devel llvm-static
+BuildRequires: clang-devel
```
drzaeus77: Once doing this, don't you also need a `Requires: clang-libs` under libbcc to make up for the new .so requirement?

drzaeus77:

drzaeus77:
[buildbot, test this please]

SPECS/bcc.spec
wcohen: As part of the rpmbuild process the executables are scanned for shared libraries requirements. This is can be seen toward the end of https://kojipkgs.fedoraproject.org//work/tasks/4695/16554695/build.log  .  rpmlint will complain if there are explicit dependencies on the rpms that contain those libraries.

drzaeus77:
[buildbot, test this please]

drzaeus77:
[buildbot, test this please]

drzaeus77:
So, after configuring jenkins to ignore the static clang installation, and also after installing the dependency on ncurses-devel (link-time dependency of libclang I believe), the error at https://buildbot.iovisor.org/jenkins/job/bcc-fc24-pr/180/consoleFull is observed.

Though the buildbot is named fc24, this is actually a rawhide VM, with the following library+gcc versions:
```
$ rpm -q llvm-devel clang-devel gcc
llvm-devel-3.9.0-6.fc26.x86_64
clang-devel-3.9.0-3.fc26.x86_64
gcc-6.2.1-2.fc26.x86_64
```

wcohen:
Is there additional details on how jenkins/fc24 is configured, so one can recreate the environment that the build is failing in?  In particular the rpms installed and other software installed.  How does this setup differ from the successful jenkins/fc23?

drzaeus77:
> Comparing the console logs the rawhide version using diff. It looks like the rawhide version is not setting the path to locally installed llvm code.

This was me testing. I wanted to see the result of the .spec file when not using the local llvm. The fc23 buildbot still has the local llvm pointed to, which is the reason why it still works.

I did try manually building fc23 without /opt/local/llvm, but that failed with some new linktime errors (seems like it tried to use /usr/lib instead of /usr/lib64 for libLLVMLTO.a). I don't intend to pursue that error further, and will keep the fc23 buildbot using /opt/local/llvm.

> The rpm builds fine on a local rawhide image, but missing ncurses-devel caused problem with the koji rawhide build:
> http://koji.fedoraproject.org/koji/taskinfo?taskID=16564325
> https://kojipkgs.fedoraproject.org//work/tasks/4325/16564325/build.log
>
> However, adding the ncurses-devel build requires resulted in successful build on koiji rawhide:
>
> http://koji.fedoraproject.org/koji/taskinfo?taskID=16564459
>
> Are there any differences between the koji root.log and the jenkin environment that look like they would cause the link failure on jenkins?
> https://kojipkgs.fedoraproject.org//work/tasks/4459/16564459/root.log

Possibly the use of c++ static:
```
-- Found LibElf: /usr/lib64/libelf.so  
-- Performing Test ELF_GETSHDRSTRNDX
-- Performing Test ELF_GETSHDRSTRNDX - Success
Building for a split build/source kernel headers package
-- Using static-libstdc++
-- Found LuaJIT: /usr/lib64/libluajit-5.1.so  
```

I'm currently bringing up a proper fc24 as opposed to rawhide. I'll add that to the buildbot pool and see if that has any different errors.

drzaeus77:
I just did a fresh rpmbuild (no linker errors) on fc24, and have a few comments:

1. Add `make` requires
2. The removal of libbcc from the requires of the sub packages breaks them. After building, I am able to install python-bcc without installing libbcc, and then we get obvious errors like:
```
$ sudo python
Python 2.7.12 (default, Sep 29 2016, 13:30:34)
[GCC 6.2.1 20160916 (Red Hat 6.2.1-2)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import bcc
...
OSError: libbcc.so.0: cannot open shared object file: No such file or directory
```

Whatever magic was supposed to happen, in fact, did not. The requires of python-bcc rpm is clearly too short:
```
$ rpm -qp ./python-bcc-0.2.0-107.git.a633fe0.x86_64.rpm --requires
python(abi) = 2.7
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PartialHardlinkSets) <= 4.0.4-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsXz) <= 5.2-1
```

3. We probably want a dependency on the kernel-devel package, otherwise we'll get:
```
$ sudo /usr/share/bcc/tools/opensnoop
chdir(/lib/modules/4.5.5-300.fc24.x86_64/build): No such file or directory
  File "/usr/share/bcc/tools/opensnoop", line 126, in <module>
    b = BPF(text=bpf_text)
  File "/usr/lib/python2.7/site-packages/bcc/__init__.py", line 240, in __init__
    raise Exception("Failed to compile BPF module %s" % src_file)
Exception: Failed to compile BPF module
```

drzaeus77:
> It is possible to control the rpmbuild from the command line. http://people.redhat.com/wcohen/bcc/bcc-bcond.diff puts a %bcond_with local_clang_static to control the rpm depenencies. To disable rpm dependencies in jenkins build would run with:
> 
> rpmbuild -ba bcc.spec --with local_clang_static
> 
> Is changing the command that jenkins uses to start the rpm build acceptable? Or does that need to remain the same?

This is fine, just include the change to scripts/build-rpm.sh in your diff and the buildbot will pick it up.

drzaeus77:
> Building with the dynamic clang-lib add that dependency and pulls in 29MB. However, it does make the bcc rpms a bit smaller.

The combined clang-libs+llvm-libs adds 65M on disk, so depending on how libbcc was built (llvm static/dynamic) that is all overhead or partially overhead. When llvm is linked statically into libbcc, it is all overhead, since llvm-static (~17M) is included already in libbcc and clang-libs has its own dependency on llvm-libs, so it gets double counted. Seems a bit messy to me.

Time will tell whether it bothers people or not to have this extra baggage.

wcohen:
Will split the build-rpm.sh change as a separate patch.

I have a test bcc.spec file with subpackage requires added back for python-bcc and bcc-lua. As mentioned earlier for executables that use shared libraries rpm should find things.  However, it doesn't catch the python (or lua) imports.

Where should the "make" requires go?  It seems pretty difficult to setup up a system that doesn't have make installed on fedora.  Is there a particular case where this missing dependency cause a problem?  Is that a build requires or python-bcc/lua-bcc requires?

People could have a custom kernel installed and not require a kernel-devel installed or people might have the wrong kernel-devel installed.  It would be better to have a runtime check in the bcc code the describes the problem rather than trying to satistify this as a rpm requires.

It looks like the examples are a mixture of lua and python while the bcc-tools are python only.  Trying to make the dependencies as accurate as possible so currently have.  Do the following look correct:


%package -n libbcc
Summary: Shared Library for BPF Compiler Collection (BCC)
Requires: elfutils-libelf
%description -n libbcc
Shared Library for BPF Compiler Collection (BCC)

%package -n python-bcc
Summary: Python bindings for BPF Compiler Collection (BCC)
Requires: libbcc = %{version}-%{release}
%description -n python-bcc
Python bindings for BPF Compiler Collection (BCC)

%package -n bcc-lua
Summary: Standalone tool to run BCC tracers written in Lua
Requires: libbcc = %{version}-%{release}
%description -n bcc-lua
Standalone tool to run BCC tracers written in Lua

%package -n libbcc-examples
Summary: Examples for BPF Compiler Collection (BCC)
Requires: python-bcc = %{version}-%{release}
Requires: bcc-lua = %{version}-%{release}
%description -n libbcc-examples
Examples for BPF Compiler Collection (BCC)

%package -n bcc-tools
Summary: Command line tools for BPF Compiler Collection (BCC)
Requires: python-bcc = %{version}-%{release}
%description -n bcc-tools
Command line tools for BPF Compiler Collection (BCC)

drzaeus77:
> Will split the build-rpm.sh change as a separate patch.
> 
> I have a test bcc.spec file with subpackage requires added back for python-bcc and bcc-lua. As mentioned earlier for executables that use shared libraries rpm should find things. However, it doesn't catch the python (or lua) imports.

Sounds good to me. And I agree on the automatic nature of the library inclusion, it was working fine for libbcc->clang-libs dependency generation.
> 
> Where should the "make" requires go? It seems pretty difficult to setup up a system that doesn't have make installed on fedora. Is there a particular case where this missing dependency cause a problem? Is that a build requires or python-bcc/lua-bcc requires?

It's a build-requires. I simply took a fresh fc24 EC2 AMI and came across this missing dependency when trying to run build-rpm.sh. I had installed all of the things mentioned in the updated spec file. Agreed that the lack of make is unlikely, but still it's an easy fix.
> 
> People could have a custom kernel installed and not require a kernel-devel installed or people might have the wrong kernel-devel installed. It would be better to have a runtime check in the bcc code the describes the problem rather than trying to satistify this as a rpm requires.

It's not a specific kernel-devel that we'd depend on, just the meta kernel-devel. Anyway, I'm not religious about this one, if depending on kernel-devel is nonstandard, we can keep it out.
> 
> It looks like the examples are a mixture of lua and python while the bcc-tools are python only. Trying to make the dependencies as accurate as possible so currently have. Do the following look correct:
> 
> %package -n libbcc
> Summary: Shared Library for BPF Compiler Collection (BCC)
> Requires: elfutils-libelf
> %description -n libbcc
> Shared Library for BPF Compiler Collection (BCC)
> 
> %package -n python-bcc
> Summary: Python bindings for BPF Compiler Collection (BCC)
> Requires: libbcc = %{version}-%{release}
> %description -n python-bcc
> Python bindings for BPF Compiler Collection (BCC)
> 
> %package -n bcc-lua
> Summary: Standalone tool to run BCC tracers written in Lua
> Requires: libbcc = %{version}-%{release}
> %description -n bcc-lua
> Standalone tool to run BCC tracers written in Lua
> 
> %package -n libbcc-examples
> Summary: Examples for BPF Compiler Collection (BCC)
> Requires: python-bcc = %{version}-%{release}
> Requires: bcc-lua = %{version}-%{release}
> %description -n libbcc-examples
> Examples for BPF Compiler Collection (BCC)
> 
> %package -n bcc-tools
> Summary: Command line tools for BPF Compiler Collection (BCC)
> Requires: python-bcc = %{version}-%{release}
> %description -n bcc-tools
> Command line tools for BPF Compiler Collection (BCC)

Looks fine.

wcohen:
I am still learning the github work flow for bcc and don't know how to replace the old patch on wcohen/spec with the current patch, so the updated patch has been put on the wcohen/specv4 branch (https://github.com/wcohen/bcc-1/tree/wcohen/specv4).

drzaeus77:
> I am still learning the github work flow for bcc and don't know how to replace the old patch on wcohen/spec with the current patch, so the updated patch has been put on the wcohen/specv4 branch (https://github.com/wcohen/bcc-1/tree/wcohen/specv4).

Github should automatically update the PR if you push your changes to the same branch as the PR was created with. Don't delete the branch, but force pushing over it is ok. Adding a new commit on top is ok too, but a clean history is preferred once we merge the PR.

drzaeus77:
FYI, I'm currently working on fixing the buildbots. I ran some updates and rebooted into newer kernels, and subsequently some tests have started failing due to some issues that I will send a PR for shortly.

PR #817, opened 2016-11-24
Fixup test errors in clang, c api

drzaeus77:
Some errors seem to have cropped up due to updating kernels and clang
library versions.
1. test_c_api fails on kernels (seen on fedora 4.9) where startup_64
doesn't start at a particular address. Simplify the test to hopefully
work in more systems.
2. The lua and python test_clang did a dereference on a particular bio
field, which has been removed. Pick a different field that is more
uniquitous.
3. test_xlate1 was generating code where the verified pointer spilled.
Rewrite the example to store the result on the stack.


drzaeus77:
[buildbot, test this please]

PR #823, opened 2016-11-29
Determine kernel dirs at runtime (fix #743)

### LINKED ISSUE #743: Run-time detection of kernel headers location for containers (follow up of #397)
I am trying to run bcc from within a Docker container. The Dockerfile has the following:

```
FROM zlim/bcc
RUN apt-get update -y && apt-get install -y bash runit conntrack iproute2 \
    util-linux curl python bcc-tools python-bcc libbcc
```

And then, I run the Docker container with options like:

```
docker run --rm --privileged -v /var/run:/var/run \
    -v /lib/modules:/lib/modules --net=host --pid=host ...
```

It does not always find the file `./include/linux/kconfig.h`, depending on the host distributions. According to https://github.com/iovisor/bcc/issues/397#issuecomment-186033921, the headers can be in the `sources` or in the `build` directory, depending on distributions. And the compilation option `-DBCC_KERNEL_HAS_SOURCE_DIR=1` allows to distinguish that.

But at build-time of the Docker image, I don't know yet whether the host distribution will split the `source` and `build` directories.

Could bcc do the check of `source` or `build` at run-time instead of build-time? In my example, the bcc package in "zlim/bcc" image, based on Ubuntu 16.04, might not use the correct compilation option, depending on where the Docker image is executed. And this fails on CoreOS, which has `/lib/modules/4.6.3-coreos/{build,source}`.

/cc @iaguis @asymmetric

schu:
As some module build directories split headers between `source/` and
`build/` and this varies between distributions, determine the location
at runtime.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,24 +49,6 @@ endif()
 
 find_package(LibElf REQUIRED)
 
-# Set to non-zero if system installs kernel headers with split source and build
-# directories in /lib/modules/`uname -r`/. This is the case for debian and
-# suse, to the best of my knowledge.
-
-if(EXISTS "${ROOT}/lib/modules/${CMAKE_SYSTEM_VERSION}/build" AND EXISTS "${ROOT}/lib/modules/${CMAKE_SYSTEM_VERSION}/source")
-  message("Building for a split build/source kernel headers package")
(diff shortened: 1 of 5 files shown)


drzaeus77:
[buildbot, ok to test]

PR #824, opened 2016-11-29
Clean up bcc_exception.h

palmtenor:
- Remove unused headers and Macros.
- Implement a `StatusTuple` class that has `code()` and `msg()` method, instead of using `std::get` and `std::tuple`, as suggested by @4ast and @drarmstr in #781.
- `clang-format -i` on the file.

### DIFF
--- a/src/cc/bcc_exception.h
+++ b/src/cc/bcc_exception.h
@@ -37,6 +37,10 @@ public:
     msg_ = std::string(buf);
   }
 
+  void append_msg(const std::string& msg) {
+    msg_ += msg;
+  }
+
   int code() { return ret_; }
 
   std::string msg() { return msg_; }
--- a/src/cc/bcc_exception.h
+++ b/src/cc/bcc_exception.h
@@ -16,12 +16,9 @@
 
 #pragma once
 
-#include <exception>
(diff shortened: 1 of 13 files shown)


palmtenor:
Updated according to comment

src/cc/bcc_exception.h
```diff
@@ -39,30 +36,28 @@ static inline StatusTuple mkstatus(int ret, const char *msg) {
...
 }
 
-#define TRY(CMD) \
-  do { \
-    int __status = (CMD); \
-    if (__status != 0) { \
-      return __status; \
-    } \
-  } while (0)
+static int status_code(const StatusTuple &status_tuple) {
```
4ast: since it's in the header file, it probably needs 'static inline' ?

palmtenor:
Should I rename `mkstatus_` to be `mkstatus`? I guess originally the underscore is to avoid name conflict with the `mkstatus` functions from `exception.h`. But we don't have those anymore...

4ast:
you could have kept mkstatus() as inline func that returns StatusTuple, but since you've changed all of the callsites anyway, I think it's fine.

PR #830, opened 2016-11-30
hello_perf_output: match the data type of pid

lcp:


brendangregg:
makes sense. I imagine it worked by accident before (auto-padding by llvm?).

PS, this looked like a great talk! http://www.slideshare.net/suselab/ebpf-trace-from-kernel-to-userspace

PR #837, opened 2016-12-01
add Quick Start Guide for bcc docker

zlim:


zlim:
As requested by @4ast on iovisor-dev.

4ast:
Looks awesome!
Does it mean that you're signing up to maintain and update zlim/bcc docker image ? :)
How often do you update it?
cc @brendangregg @drzaeus77

brendangregg:
Maybe just me, but my test system ran the docker image, but not the tools:

```
# ./opensnoop 
chdir(/lib/modules/4.9.0-rc5-virtual/build): No such file or directory
  File "./opensnoop", line 126, in <module>
    b = BPF(text=bpf_text)
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 205, in __init__
    raise Exception("Failed to compile BPF module %s" % src_file)
Exception: Failed to compile BPF module 
```

So that directory doesn't exist in this docker image. Probably because I replaced the kernel to something odd, right?

brendangregg:
I tried this on Ubuntu Xenial, and it just worked.

Is there a way, for such systems with different kernels, to either:
- fix it,
- throw an error explaining why it doesn't work, or,
- at least document it in the README

4ast:
@zlim does your docker image include the latest patch that detects source/build directories at runtime instead of build time?

zlim:
@4ast, @brendangregg: Thanks for looking at this. bcc-docker contains the "Xenial nightly build". The intent is to keep it in sync with the .deb package. bcc-docker is already an automated build on docker hub. Need to figure out hook it up such that any .deb rebuild also triggers how to trigger a do

zlim:
@4ast, @brendangregg: Thanks for looking at this. bcc-docker contains the "Xenial nightly build". The intent is to keep it in sync with the .deb package. bcc-docker is already an automated build on docker hub. Need to figure out hook it up such that any .deb rebuild also triggers a docker rebuild.

What do you think? Is it better to keep it at bleeding edge? Or should we keep it at a "stable" version?

4ast:
@zlim I think the 'bleeding edge' is what folks would like to try if they're following these directions. Otherwise they will be ok with default distro packages and docker image won't help much to scale complexity wall ?

zlim:
@4ast Agreed. It's already using Xenial nightly builds, so I'll let that stand.

Right now:
1. bcc-docker rebuild automatically whenever the base image (ubuntu) is updated.
2. scheduled daily rebuilds to pick up the nightly .deb pkgs.

Is there value in exposing an endpoint (to trigger bcc-docker update) that the nightly buildbot can call?

4ast:
@zlim ok. makes sense. does it mean that this current PR is good as-is ?
@drzaeus77 is there a way to trigger bcc-docker from buildbot?

drzaeus77:
@4ast Setting up a REST hook from buildbot to docker is possible, but probably much easier (no need to set up a passphrase or certificate or w/e) to just coordinate the times. If there is a jenkins plugin that knows how to talk to the docker hub service, that might help, but I haven't looked into it yet.

zlim:
@4ast: Yes, you may pull this PR as is. FYI, I've also updated the description at [Docker Hub](https://hub.docker.com/r/zlim/bcc/).

@4ast, @drzaeus77: Based on observation that the nightly build completes ~02:30UTC, a job is setup to rebuild bcc-docker at ~03:30UTC. I also have a endpoint setup on AWS to manually trigger it. Happy to share API key with you privately.

PR #840, opened 2016-12-06
Add support for aarch64

zhiyisun:
ABI for aarch64: register x0-x7 are used for parameter and result. In bcc, there are 6 parameter registers are defined. So use x0-x5 as parameter. frame pointer, link register, stack pointer and pc are added in PT_REGS_xx according to arm64 architecture.

syscall number of bpf for aarch64 are defined in Kernel header uapi/asm-generic/unistd.h.

Signed-off-by: Zhiyi Sun <zhiyisun@gmail.com>


4ast:
LGTM
[buildbot, ok to test]

PR #845, opened 2016-12-09
Fix for issue 829 I reported 

affansyed:
Need the correct initialization without which the verifier would reject the code.


affansyed:
@4ast @brendangregg I will now update the tutorial that Brenden had setup so that everything is in sync as well

PR #847, opened 2016-12-09
fixing tutorial

affansyed:
based on commit 1b10c06


docs/tutorial_bcc_python_developer.md
```diff
@@ -603,7 +603,7 @@ TIME(s)            COMM             PID    ARGS
 24653340.510164998 node             24728  path:/images/favicon.png
 ```
 
-Relevant code from [examples/tracing/nodejs_http_server.py](../examples/tracing/nodejs_http_server.py):
+Relevant code from [examples/tracing/nodejs_http_server.py](../examples/tracing/nodes_http_server.py):
```
4ast: introducing a typo?

affansyed: dont know how that crept in. I though I just change the one line below. Lemme fix this.

affansyed:
hopefully acceptable now @4ast

docs/tutorial_bcc_python_developer.md
```diff
@@ -616,7 +616,7 @@ bpf_text = """
 #include <uapi/linux/ptrace.h>
 int do_trace(struct pt_regs *ctx) {
     uint64_t addr;
-    char path[128];
+    char path[128]={0};
```
4ast: docs should follow the coding convention that tools/* do.
Please use 'char path[128] = {};'... note spaces...
Also this init is not really needed. Several releases ago bpf_probe_read() started to accept uninitialized memory, so it's best to avoid zero init and use it only on old kernels.

goldshtn:
@affansyed Perhaps it would be better to add a note saying that the initialization might be needed on older kernels?

Issue #850, opened 2016-12-12
check effective userid on run


rickysarraf:
Hi,

Thank you for the BPF tools. They really are a great set, for a myriad of use cases. I've taken up the task of packaging it for Debian. It currently is under the NEW queue, and should hopefully clear in time to become part of the Debian Stretch release.

Meanwhile, here's an exception  that keeps throwing up, when run unprivileged.

```
rrs@learner:~/rrs-home/Community/Packaging/bpfcc (master)$ execsnoop 
In file included from /virtual/main.c:4:
include/linux/fs.h:2680:9: warning: comparison of unsigned enum expression < 0 is always false
      [-Wtautological-compare]
        if (id < 0 || id >= READING_MAX_ID)
            ~~ ^ ~
1 warning generated.
bpf: Operation not permitted

Traceback (most recent call last):
  File "/usr/sbin/execsnoop", line 148, in <module>
    b = BPF(text=bpf_text)
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 203, in __init__
    self._trace_autoload()
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 679, in _trace_autoload
    fn = self.load_func(func_name, BPF.KPROBE)
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 243, in load_func
    raise Exception("Failed to load BPF program %s" % func_name)
Exception: Failed to load BPF program kprobe__sys_execve
2016-12-12 / 14:26:44 ♒♒♒  ☹  => 1  
```

Can we make it cleaner? Something like the following. But it'd have to be added to all the command files.

```
if os.geteuid() != 0:
    sys.stderr.write("Need superuser privileges for %s\n" % (sys.argv[0]))
    sys.exit(1)
```

I thought of adding this check into the bcc module. But then, it'd have to be added to multiple functions. So instead I thought it'd make more sense to just update the command files, for once.

goldshtn:
I think it belongs in the BPF class. Various functions like `attach_kprobe` could just call this code as a helper method.

rickysarraf:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA512

On Mon, 2016-12-12 at 11:19 -0800, Sasha Goldshtein wrote:
> I think it belongs in the BPF class. Various functions like attach_kprobe
> could just call this code as a helper method.

I just added it to the initialization section. This helps cover almost all the
different callers.


[...]

PR #856, opened 2016-12-13
check effective userid on run #850

brendangregg:


4ast:
yeah. the python traces were not really helpful. thanks

PR #859, opened 2016-12-14
add runqlen tool

brendangregg:
also adds a linear histogram

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,7 @@ Examples:
 - tools/[profile](tools/profile.py): Profile CPU usage by sampling stack traces at a timed interval. [Examples](tools/profile_example.txt).
 - tools/[reset-trace](tools/reset-trace.sh): Reset the state of tracing. Maintenance tool only. [Examples](tools/reset-trace_example.txt).
 - tools/[runqlat](tools/runqlat.py): Run queue (scheduler) latency as a histogram. [Examples](tools/runqlat_example.txt).
+- tools/[runqlen](tools/runqlen.py): Run queue length as a histogram. [Examples](tools/runqlen_example.txt).
 - tools/[softirqs](tools/softirqs.py):  Measure soft IRQ (soft interrupt) event time. [Examples](tools/softirqs_example.txt).
 - tools/[solisten](tools/solisten.py): Trace TCP socket listen. [Examples](tools/solisten_example.txt).
 - tools/[sslsniff](tools/sslsniff.py): Sniff OpenSSL written and readed data. [Examples](tools/sslsniff_example.txt).
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -62,6 +62,7 @@ This guide is incomplete. If something feels missing, check the bcc and kernel s
         - [4. values()](#4-values)
         - [5. clear()](#5-clear)
         - [6. print_log2_hist()](#6-print_log2_hist)
+        - [7. print_linear_hist()](#6-print_linear_hist)
     - [Helpers](#helpers)
         - [1. ksym()](#1-ksym)
         - [2. ksymaddr()](#2-ksymaddr)
@@ -1009,6 +1010,65 @@ This is an efficient way to summarize data, as the summarization is performed in
 Examples in situ:
 [search /examples](https://github.com/iovisor/bcc/search?q=print_log2_hist+path%3Aexamples+language%3Apython&type=Code),
 [search /tools](https://github.com/iovisor/bcc/search?q=print_log2_hist+path%3Atools+language%3Apython&type=Code)
+
+### 6. print_linear_hist()
+
+Syntax: ```table.print_linear_hist(val_type="value", section_header="Bucket ptr", section_print_fn=None)```
+
+Prints a table as a linear histogram in ASCII. This is intended to visualize small integer ranges, eg, 0 to 100.
+
+Arguments:
+
+- val_type: optional, column header.
+- section_header: if the histogram has a secondary key, multiple tables will print and section_header can be used as a header description for each.
+- section_print_fn: if section_print_fn is not None, it will be passed the bucket value.
+
+Example:
+
+```Python
+b = BPF(text="""
+BPF_HISTOGRAM(dist);
+
+int kprobe__blk_account_io_completion(struct pt_regs *ctx, struct request *req)
+{
+	dist.increment(req->__data_len / 1024);
+	return 0;
+}
+""")
+[...]
+
+b["dist"].print_linear_hist("kbytes")
+```
+
+Output:
+
+```
+     kbytes        : count     distribution
+        0          : 3        |******                                  |
+        1          : 0        |                                        |
+        2          : 0        |                                        |
+        3          : 0        |                                        |
+        4          : 19       |****************************************|
+        5          : 0        |                                        |
+        6          : 0        |                                        |
+        7          : 0        |                                        |
+        8          : 4        |********                                |
+        9          : 0        |                                        |
+        10         : 0        |                                        |
+        11         : 0        |                                        |
+        12         : 0        |                                        |
+        13         : 0        |                                        |
+        14         : 0        |                                        |
+        15         : 0        |                                        |
+        16         : 2        |****                                    |
+[...]
+```
+
+This is an efficient way to summarize data, as the summarization is performed in-kernel, and only the values in the count column are passed to user space.
+
+Examples in situ:
+[search /examples](https://github.com/iovisor/bcc/search?q=print_linear_hist+path%3Aexamples+language%3Apython&type=Code),
+[search /tools](https://github.com/iovisor/bcc/search?q=print_linear_hist+path%3Atools+language%3Apython&type=Code)
 
 ## Helpers
 
--- a/man/man8/runqlat.8
+++ b/man/man8/runqlat.8
@@ -102,4 +102,4 @@ Unstable - in development.
 .SH AUTHOR
 Brendan Gregg
 .SH SEE ALSO
-pidstat(1)
+runqlen(8), pidstat(1)
--- a/man/man8/runqlen.8
+++ b/man/man8/runqlen.8
@@ -1,0 +1,86 @@
+.TH runqlen 8  "2016-12-12" "USER COMMANDS"
+.SH NAME
+runqlen \- Scheduler run queue length as a histogram.
+.SH SYNOPSIS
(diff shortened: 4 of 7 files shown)


brendangregg:
@4ast this is the first example of a bpf_get_current_task() from a sampling perf event. I ran into a couple of minor things that can be future feature requests:

I'm fetching the run queue length from task->se.cfs_rq->nr_running, which I don't think is ideal (cfs_rq always populated?). I'd rather something like a this_rq() or task_rq(p) BPF call: bpf_get_current_rq().

I'm also fetching the CPU from task->wake_cpu, which I'm assuming is right (seems so). Would also be nice to have a bpf_get_current_cpu(). We can get the CPU from perf_submit() for outputting records, but I was after one in BPF context to use in maps.

4ast:
useful tool !
as far as feature request... bpf_get_current_rq() probably won't fly, since 'struct rq' is the kernel internal data structure and scheduler folks are the biggest opponent of tracepoints. Exposing them to user space is pretty much unthinkable from their point of view, since it limits how scheduler can be changed in the future.
For 2nd part.. i'm missing how that's different from bpf_get_smp_processor_id() ?

brendangregg:
Ah, damn, I'd missed bpf_get_smp_processor_id(), I was looking for bpf.*cpu. I'll update the tool to use it (testing first).

brendangregg:
@4ast there's a few more scheduler tools I'd like to do. I really want a tool that shows the percent of time there were idle CPUs at the same time as tasks were queued waiting. Eg:

```
# ./runqaffinity.py -OT 1 
Sampling run queues... Hit Ctrl-C to end.

19:54:53
queued when idle: 0.00%

19:54:54
queued when idle: 0.05%

19:54:55
queued when idle: 0.01%
[...]
```

A little bit of this is ok (CPU affinity). A lot is not (Eg, an app trying to be clever and doing CPU binding, but then overloading the bound CPUs). Or just scheduler bugs.

But the BPF problem is knowing whether there's queued tasks and idle CPUs at the same time. I could setup perf sampling to only fire on one CPU, and then do unrolled loops.. but bleh.. I'm actually thinking of dumping timestamped samples to user-space and post-processing there...

brendangregg:
Oh, I think I've figured out at least one way. Double-sampling and alternating between recording and reporting. Will cause some extra interrupts (198 Hertz instead of 99), but should still be cheap. Some pseudocode.

```
// Timer fires at 2x desired frequency. CPU 0 toggles a global variable,
// "recording", between 0 and 1. On 1 cycles, all CPUs record. On 0 cycles,
// CPU 0 post processes the previous cycle's data.

// if not recording, and not CPU 0, return
if (cpu != 0 && recording == 0)
        return;

// do record
if (recording)
        idle++ if idle
        queued++ if queued

// recorders are now done
if (cpu != 0)
        return

// end recording
if (recording)
        recording = 0
        return

// this sample was processing
if (idle && queued)
        queued_when_idle++
recording = 1
```

4ast:
i'm not quite following why double sampling is needed.
can do the same with normal frequency and record into alternating variables?
two global idle and two global queued variables?
they're going to be incremented with xadd, right?
could you also add a flag to the current tool to use nmi hw cycles instead of sw timer?
i think it will be more accurate. I understand you cannot really test it due to hypervisor,
but datacenter folks would prefer hw cycles due to accuracy ;)
another question: idlea means nr_running==0 ?
do you think adding direct helper like idle_cpu() will be meaningfully better?
probably not.
I was also thinking that 'this_rq' feature request can be solved by adding generic helpers like
bpf_get_per_cpu_var(var, cpu_id)
bpf_get_this_cpu_var(var)
most of the global variables like runqueue and current_task are accessible this way.
Are there any other vars that would be interesting to fetch?
Not sure how to do it safely though...

brendangregg:
I'm not sure alternating variables works: at the same time, we'll have all CPUs updating counters, and one CPU trying to process them. If it flipped between variable sets, it could do so at slightly different times on each wakeup, and each variable set may or may not have been populated by the other CPUs yet.

With double sampling, there's a full sample cycle for the CPUs to quiesce. (I've now written the tool and it does work; man page next...)

Good point about HW cycles, thanks, I should make these tools default to HW if it's available.

> another question: idle means nr_running==0 ?

Yes. I just noticed I have a wrong code comment about that; I'll fix it in a following PR. runqlen is measuring the length beyond nr_running==1 (which is one non-idle running thread), hence a -1 in there, and other comments saying it is excluding the running thread.

There's also this in include/linux/sched.h:

```C
/**
 * is_idle_task - is the specified task an idle task?
 * @p: the task in question.
 *
 * Return: 1 if @p is an idle task. 0 otherwise.
 */
static inline bool is_idle_task(const struct task_struct *p)
{
        return p->pid == 0;
}
```

If I'm writing a tool that just needs to identify idle, I'd use that instead, since it avoids the task_struct->se->cfs_rq walk. I'm not here, since these tools have to walk that anyway to get to nr_running.

> I was also thinking that 'this_rq' feature request can be solved by adding generic helpers like
> bpf_get_per_cpu_var(var, cpu_id)
> bpf_get_this_cpu_var(var)
> most of the global variables like runqueue and current_task are accessible this way.

Accessing global variables is likely going to be a big help sooner or later, but I'll need to think of examples.

Thanks!

brendangregg:
... RE variable set switching: I should add, a constraint here is that we need all CPU's run queue information at the same point in time, so we can reason about queues on some CPUs at the same time as idle on others. I know variable set switching works for things like counts, where the timing isn't important.

4ast:
having been thinking that doing queued_when_idle this way is not quite correct, since sampling is not quite applicable here. There are many scheduling events in the system and 99 hz or any other frequency != kernel HZ. Also there are no_hz setups and no_hz when idle, etc. Would it make sense to do such accounting during scheduling events instead?

brendangregg:
It's a sampling (rough) approach, and avoids the overhead of scheduler event tracing. As with any sampler, there'll be give and take when hitting or missing short-lived events. If we can reliably sample at 99 Hz it should work.

But I think I did just hit an issue where SW samples are not firing during idle states, like you describe.

idle:

```
# ./cpuunclaimed.py 1
Sampling run queues... Hit Ctrl-C to end.

%CPU 0.00%, unclaimed idle 0.00%

%CPU 0.00%, unclaimed idle 0.00%

%CPU 0.00%, unclaimed idle 0.00%
[...]
```

Looks good.

All CPUs busy (8 hot threads, 8 CPUs):

```
# ./cpuunclaimed.py 1
Sampling run queues... Hit Ctrl-C to end.

%CPU 100.00%, unclaimed idle 0.00%

%CPU 100.00%, unclaimed idle 0.00%

%CPU 100.00%, unclaimed idle 0.00%

%CPU 99.87%, unclaimed idle 0.13%

%CPU 100.00%, unclaimed idle 0.00%
[...]
```

Looks good.

Inbalance: 8 hot threads bound to 1 CPU (via taskset), 8 CPUs, remaining 7 CPUs idle:

```
# ./cpuunclaimed.py 1
Sampling run queues... Hit Ctrl-C to end.

%CPU 12.50%, unclaimed idle 87.37%

%CPU 12.50%, unclaimed idle 87.37%

%CPU 16.42%, unclaimed idle 83.58%

%CPU 16.67%, unclaimed idle 82.83%

%CPU 19.57%, unclaimed idle 80.43%

%CPU 17.24%, unclaimed idle 82.76%

%CPU 16.84%, unclaimed idle 83.16%

%CPU 16.67%, unclaimed idle 83.33%
```

The first two lines are perfect, then it drifts. I added more columns, and I can see the number of samples decreases. Hmm...

brendangregg:
Yes, those events aren't firing. Easy to see with "perf record -F99 -a", "perf script". Some CPUs seem to go in a power down state (C state?).

brendangregg:
So I can just infer idle_cpu_count as cpu_count - running_cpu_count:

```
# ./cpuunclaimed.py 1
Sampling run queues... Hit Ctrl-C to end.

%CPU 12.50%, unclaimed idle 86.49%

%CPU 12.50%, unclaimed idle 87.50%

%CPU 12.50%, unclaimed idle 87.50%

%CPU 12.50%, unclaimed idle 87.50%

%CPU 12.50%, unclaimed idle 87.50%

%CPU 12.63%, unclaimed idle 87.37%

%CPU 12.50%, unclaimed idle 87.50%
[...]
```

Just need a nice way to know the CPU count (ncpu).

brendangregg:
Do I need to call xadd explicitly from bcc/BPF? Like:

```
            if (val) xadd(val, 1);                                              
[...]
error: couldn't allocate output register for constraint 'r' at line 2150480047
```

I was looking at the output of:

```
if (val) (*val)++;
[...]
57: (79) r1 = *(u64 *)(r0 +0)
58: (07) r1 += 1
```

which doesn't look atomic.

4ast:
sorry. it's lock_xadd() or use __sync_fetch_and_add() directly.
see src/cc/export/helpers.h

brendangregg:
thanks! I should put an example of this in bcc somewhere. Turns out that wasn't my bug -- I still have a race condition.

brendangregg:
... is it expected I can't use _sync_val_compare_and_swap()?

```
LLVM ERROR: Cannot select: 0x3c50070: i64,ch = AtomicCmpSwap 0x2a62a10, 0x2785d60, 0x2b3c5f0, 0x3c4fe20<Volatile LDST8[%18]> [ORD=3] [ID=11]
  0x2785d60: i64,ch = CopyFromReg 0x2a62a10, 0x3c25198 [ORD=2] [ID=8]
    0x3c25198: i64 = Register %vreg2 [ID=1]
  0x2b3c5f0: i64,ch = CopyFromReg 0x2a62a10, 0x2b3c150 [ORD=3] [ID=9]
    0x2b3c150: i64 = Register %vreg3 [ID=2]
  0x3c4fe20: i64,ch = CopyFromReg 0x2a62a10, 0x3c259b0 [ORD=3] [ID=10]
    0x3c259b0: i64 = Register %vreg1 [ID=3]
```

Trying to implement a tie-breaker when two samples fire at exactly the same time, and I want only one to then execute some code. Also tried __atomic_compare_exchange()...

4ast:
correct. there is only __sync_fetch_and_add.

brendangregg:
My work in progress so far... https://gist.github.com/brendangregg/731cf2ce54bf1f9a19d4ccd397625ad9

It works, but isn't perfect (without a way to do thread synchronization like a CAS). I might have to give up and dump samples to user space for perfection, although I don't know how much that matters, and that will cost more.

PR #862, opened 2016-12-16
Add USDT support to C++ API

palmtenor:
- Uses existing code to parse ELF and get addresses.
- Handles each probe location just as an `uprobe` with a particular address offset instead of symbol name.

Added an example that traces any binary contains the `request_context_switch_before` Tracepoint from [folly](https://github.com/facebook/folly/blob/master/folly/io/async/Request.cpp#L98)

```
$ sudo ./FollyRequestContextSwitch /home/qinteng/tester_server
Started tracing, hit Ctrl-C to terminate.
PID 2333652 (IOThreadPool23) folly::RequestContext switch from 0 to 139706199539968
PID 2333652 (IOThreadPool23) folly::RequestContext switch from 139706199539968 to 0
PID 2333652 (IOThreadPool23) folly::RequestContext switch from 0 to 139706199539968
PID 2333652 (IOThreadPool23) folly::RequestContext switch from 139706199539968 to 0
PID 2333652 (IOThreadPool23) folly::RequestContext switch from 0 to 139706199539968
PID 2333626 (ThreadViz Teste) folly::RequestContext switch from 0 to 139706199539968
PID 2333626 (ThreadViz Teste) folly::RequestContext switch from 139706199539968 to 0
PID 2333652 (IOThreadPool23) folly::RequestContext switch from 139706199539968 to 0
^CTerminating...
```

### DIFF
--- a/examples/cpp/CMakeLists.txt
+++ b/examples/cpp/CMakeLists.txt
@@ -26,3 +26,7 @@ install (TARGETS RandomRead DESTINATION share/bcc/examples/cpp)
 add_executable(LLCStat LLCStat.cc)
 target_link_libraries(LLCStat bcc-static)
 install (TARGETS LLCStat DESTINATION share/bcc/examples/cpp)
+
+add_executable(FollyRequestContextSwitch FollyRequestContextSwitch.cc)
+target_link_libraries(FollyRequestContextSwitch bcc-static)
+install (TARGETS FollyRequestContextSwitch DESTINATION share/bcc/examples/cpp)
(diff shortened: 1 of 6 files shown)


4ast:
no test cases?
it was clean when bpf.init() was for program only
i'm missing why usdt needs to add to program src ?
Only #include is added or more?

palmtenor:
Just been working on a test tool.

palmtenor:
Just been working on a test example.

The reason we need to add to the program source code is for the `_bpf_readarg_X_n` functions (rewriter changes `bpf_usdt_readarg` into them). They need to know the address of each tracepoint location to read arguments correctly, and name of the function to probe the Tracepoint so that we could name these helper functions accordingly. Therefore they have to be created by something knows about information of the binary containing the Tracepoints, and added to the program before compiling:(

PR #863, opened 2016-12-18
support macro in call arguments

shodoco:
Currently, table function calls don't accept macro arguments. E.g., for the added test case `test_call_macro_arg`, it fails with:
```
/virtual/main.c:13:49: error: expected expression
    bpf_tail_call_((void *)bpf_pseudo_fd(1, 3), );
                                                ^
1 error generated.
E
```
The clang rewritter fails to get rewritten text due to macro argument source location being not re-writtable ([clang doc](http://clang.llvm.org/doxygen/classclang_1_1Rewriter.html#a78971fe62c8cc759758158ca06586363)).

This pull request creates a lambda `getRewrittenText`, which applies expansion underneath before getting rewritten text.

The added test cases pass with fix.


4ast:
cc @drzaeus77

drzaeus77:
Thanks for this. Sorry for being MIA for a bit, currently undergoing $DAYJOB transition.

PR #865, opened 2016-12-19
trace: -L, -t, -T switches

### LINKED ISSUE #854: trace.py to have timestamp options
Something like:

```
  -T, --time            include time column on output (HH:MM:SS)
  -t, --timestamp       include timestamp on output (seconds.fraction)
```

Which might mean dropping the current timestamp column and only including it with -T.

Here's why I was prompted:

```
# ./trace.py 'p::__dev_get_by_name "lookup: %s", arg2' 'r::__dev_get_by_name "returned: 0x%x", retval'
TIME     PID    TID    COMM         FUNC             -
00:46:46 1703   1703   snmpd        __dev_get_by_name returned: 0xac89f000
00:46:46 1703   1703   snmpd        __dev_get_by_name returned: 0xa4bc2000
00:46:46 1703   1703   snmpd        __dev_get_by_name returned: 0xa4bc2000
00:46:46 1703   1703   snmpd        __dev_get_by_name lookup: lo
00:46:46 1703   1703   snmpd        __dev_get_by_name lookup: eth0
00:46:46 1703   1703   snmpd        __dev_get_by_name lookup: eth0
00:46:49 1703   1703   snmpd        __dev_get_by_name returned: 0xac89f000
00:46:49 1703   1703   snmpd        __dev_get_by_name returned: 0xa4bc2000
00:46:49 1703   1703   snmpd        __dev_get_by_name lookup: lo
00:46:49 1703   1703   snmpd        __dev_get_by_name lookup: eth0
00:46:49 1703   1703   snmpd        __dev_get_by_name lookup: eth0
00:46:49 1703   1703   snmpd        __dev_get_by_name returned: 0xa4bc2000
```

It's pretty awesome that I can trace the entry and return value so easily (BTW, you can reproduce this just by running "ifconfig lo0" in another window). But with output like the above, it's mixed up (I think), and I can't easily associate the entry to the return. With a -t option, and sufficient resolution (microseconds) I could sort by time and see what happened on the same PID/TIDs.

goldshtn:
The `-t` switch for specifying a thread id has been renamed
to `-L`. The `-T` switch specifies that a time column should
be provided, whereas the `-t` switch requests a relative
timestamp (from the beginning of the trace). This is more
consistent with other tools, and lets the user choose the
desired format for the time column, if any.

Fixes #854. /cc @brendangregg

### DIFF
--- a/man/man8/trace.8
+++ b/man/man8/trace.8
@@ -2,8 +2,8 @@
 .SH NAME
 trace \- Trace a function and print its arguments or return value, optionally evaluating a filter. Uses Linux eBPF/bcc.
 .SH SYNOPSIS
-.B trace [-h] [-p PID] [-t TID] [-v] [-Z STRING_SIZE] [-S]
-         [-M MAX_EVENTS] [-o] [-K] [-U] [-I header]
+.B trace [-h] [-p PID] [-L TID] [-v] [-Z STRING_SIZE] [-S]
+         [-M MAX_EVENTS] [-t] [-T] [-K] [-U] [-I header]
          probe [probe ...]
 .SH DESCRIPTION
 trace probes functions you specify and displays trace messages if a particular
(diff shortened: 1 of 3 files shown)


brendangregg:
thanks, tested, LGTM.

PR #867, opened 2016-12-21
Don't close extern table when a module destructs

shodoco:
Currently when a `BPFModule` destructs, it closes the extern table fd. This is inconsistent with the semantic of an extern table and renders the table invalid even in the module that actually owns the table (which has `maps/export`).

E.g. the new test case, `TestSharedTable.test_close_extern` fails with `KeyError` w/o the fix in `src/cc/bpf_module.cc`.
```
ERROR: test_close_extern (__main__.TestSharedTable)
  File "/data/users/hzhou/bcc/tests/python/test_shared_table.py", line 19, in test_close_extern
    self.assertEqual(t1[ct.c_int(1)].value, 10)
  File "/data/users/hzhou/toolchains/python-2.7.11/lib/python2.7/bcc/table.py", line 406, in __getitem__
    return super(ArrayBase, self).__getitem__(key)
  File "/data/users/hzhou/toolchains/python-2.7.11/lib/python2.7/bcc/table.py", line 191, in __getitem__
    raise KeyError
KeyError
```
With the fix, the added test case passes successfully.

This pull request also adds table type into `SharedTables` registry so we can retrieve it later in the extern table. Glue code is also added to python `BPF` class so we can use `with` statement.

### DIFF
--- a/tests/python/test_shared_table.py
+++ b/tests/python/test_shared_table.py
@@ -1,0 +1,23 @@
+#!/usr/bin/env python
+# Copyright (c) PLUMgrid, Inc.
+# Licensed under the Apache License, Version 2.0 (the "License")
+
+import ctypes as ct
+import unittest
+from bcc import BPF
+
+class TestSharedTable(unittest.TestCase):
+    def test_close_extern(self):
+        b1 = BPF(text="""BPF_TABLE_PUBLIC("array", int, int, table1, 10);""")
+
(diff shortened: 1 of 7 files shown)


tests/python/test_shared_table.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python
+# Copyright (c) PLUMgrid, Inc.
```
4ast: please use your own copyright or the company your work for

PR #871, opened 2016-12-22
Add optional kernel flags for running networking examples in INSTALL.md

zhiyisun:
All of these flags are used in networking examples. Like tc policing,
action, dummy interface, vxlan support These flags have
already been enabled in some Linux distributions, like Ubuntu.
But they are disabled by default on vanilla kernels. It would be nice to
mention them in INSTALL.md to make bcc easiser to use.

Signed-off-by: Zhiyi Sun <zhiyisun@gmail.com>


goldshtn:
@drzaeus77 @4ast This looks good, assuming that these are indeed all the flags required. Can this be merged?

PR #874, opened 2016-12-26
Fix variable names in exceptions

pchaigno:
Trying to detach an unattached uprobe results in the following incorrect error message:
```Python traceback
  File "tools/gethostlatency.py", line 88, in <module>
    b.detach_uprobe(name="c", sym="gethostbyname")
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 783, in detach_uprobe
    raise Exception("Uprobe %s is not attached" % event)
NameError: global name 'event' is not defined
```
This pull request fixes it.


goldshtn:
LGTM.

Issue #853, opened 2016-12-12
gethostlatency no output


oity:
gethostlatency on ubuntu Ubuntu 16.04.1 LTS (4.4.0-53-generic) after printinig colum headers produce no noticeable output

:/usr/share/bcc/tools# ./gethostlatency
TIME      PID    COMM                  LATms HOST
^C

while in other terminal a test program has been ran

ltrace ./a.out www.google.it
__libc_start_main(0x4006d6, 2, 0x7ffdd3ef79a8, 0x400790 <unfinished ...>
gethostbyname("www.google.it")                                                                                                = 0x7fe3808581c0
printf("Hostname: %s\n", "www.google.it"Hostname: www.google.it
)                                                                                     = 24
inet_ntoa({ 0x23c63ad8 })                                                                                                     = "216.58.198.35"
printf("IP Address: %s\n", "216.58.198.35"IP Address: 216.58.198.35
)                                                                                   = 26
+++ exited (status 0) +++

pchaigno:
I was able to reproduce the error (same OS and distrib).

The issue seem to be that bcc doesn't attach to the "proper" C library. If I hardcode the path to my C library (the one used by curl, `/lib/x86_64-linux-gnu/libc.so.6` for me), it works as expected:
```diff
diff --git a/tools/gethostlatency.py b/tools/gethostlatency.py
index 73e4ad2..eadd818 100755
--- a/tools/gethostlatency.py
+++ b/tools/gethostlatency.py
@@ -84,12 +84,13 @@ int do_return(struct pt_regs *ctx) {
 }
 """
 b = BPF(text=bpf_text)
-b.attach_uprobe(name="c", sym="getaddrinfo", fn_name="do_entry")
-b.attach_uprobe(name="c", sym="gethostbyname", fn_name="do_entry")
-b.attach_uprobe(name="c", sym="gethostbyname2", fn_name="do_entry")
-b.attach_uretprobe(name="c", sym="getaddrinfo", fn_name="do_return")
-b.attach_uretprobe(name="c", sym="gethostbyname", fn_name="do_return")
-b.attach_uretprobe(name="c", sym="gethostbyname2", fn_name="do_return")
+libc = "/lib/x86_64-linux-gnu/libc.so.6"
+b.attach_uprobe(name=libc, sym="getaddrinfo", fn_name="do_entry")
+b.attach_uprobe(name=libc, sym="gethostbyname", fn_name="do_entry")
[...]

PR #875, opened 2016-12-26
Use targeted process to resolve library names

pchaigno:
This pull request implements the first strategy discussed at #853 to improve the selection of libraries for `attach_uprobe`.

To resolve library names, `bcc_procutils_which_so` now leverages mapped libraries of the targeted process, if a PID is given. It uses the kernel's `/proc/$pid/maps` to extract mapped libraries.

Any other tests I should add?

/cc @vmg @goldshtn

### DIFF
--- a/src/cc/BPF.cc
+++ b/src/cc/BPF.cc
@@ -462,7 +462,7 @@ StatusTuple BPF::check_binary_symbol(const std::string& binary_path,
                                      const std::string& symbol,
                                      uint64_t symbol_addr, bcc_symbol* output) {
   int res = bcc_resolve_symname(binary_path.c_str(), symbol.c_str(),
-                                symbol_addr, output);
+                                symbol_addr, 0, output);
   if (res < 0)
     return StatusTuple(
         -1, "Unable to find offset for binary %s symbol %s address %lx",
--- a/src/cc/bcc_proc.c
+++ b/src/cc/bcc_proc.c
@@ -197,8 +197,6 @@ static struct ld_lib {
   int flags;
 } * lib_cache;
 
-static char libpath[4096];
-
 static int read_cache1(const char *ld_map) {
   struct ld_cache1 *ldcache = (struct ld_cache1 *)ld_map;
   const char *ldstrings =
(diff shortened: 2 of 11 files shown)


goldshtn:
Thanks, this looks good (haven't tested yet). Perhaps it would be good to add a test that actually verifies we're picking the right library if there are multiple options?

pchaigno:
> Perhaps it would be good to add a test that actually verifies we're picking the right library if there are multiple options?

I'd like that too. Unless there's already a library (`libc`?) with multiple versions on all test environments, we'd need to change the environments' setup before.

pchaigno:
I just updated the pull request to fix an error.

`load_mappings` loads the mapped libraries of a given process in memory. When `bcc_procutils_which_so` is called a second time, it reads the mapped libraries directly from memory. I hadn't considered that the given PID can change, in which case we should reload mapped libraries in memory.

goldshtn:
Not only can the given pid change, but even the same pid can be recycled for another process. I don't know if we want to cache the results at all, and if we do it would have to be careful caching that takes at least the process cmdline into account to try and avoid most recycling cases. 

What do others think? @4ast @drzaeus77 @brendangregg @vmg

4ast:
My understanding is bcc_procutils_which_so() will be called roughly once per script, so caching /proc/pid/maps doesn't help much and makes it error prone for cases when pid is recycled and libbcc is used in long running services. So I like the idea of reading /proc/pid/maps to discover specific .so of given pid, but caching is unnecessary.

vmg:
I agree. We already have some quirky workarounds in the library to spot and prevent bugs related PID reuse (which is a real issue in busy production systems), so naively caching the output of `maps` based on PID sounds like a recipe for disaster. This is not performance critical code, so it should be perfectly fine to read `maps` on each call and skip the caching altogether.

pchaigno:
I updated the branch to remove the cached libraries :innocent:

4ast:
lgtm
cc @goldshtn @vmg

Aszezo:
.

PR #882, opened 2017-01-02
add missing inclusion guards in bcc/proto.h

mvbpolito:
The lack of the these guards causes different issues.  For example, if a program loaded through hover includes bcc/proto.h, it fails to compile because proto.h is included twice.


drzaeus77:
[buildbot, ok to test]

PR #890, opened 2017-01-09
Add missing helper prototypes

mvbpolito:
bpf_skb_under_cgroup and bpf_current_task_under_cgroup are still missing as they receive a map as argument and I don't know how to deal with it exactly.


4ast:
[buildbot, ok to test]

PR #906, opened 2017-01-14
argdist, trace: Support non-C identifier names

### LINKED ISSUE #903: trace: more invalid function characters from golang
I was previously testing gccgo, now using just go:

```Shell
# ./funccount.py '/root/echotls:crypto.tls*'
Tracing 197 functions for "/root/echotls:crypto.tls*"... Hit Ctrl-C to end.
^C
crypto/tls.ecdheRSAKA                       1
crypto/tls.(*Config).ticketKeys             1
crypto/tls.(*Config).cipherSuites           1
crypto/tls.(*Config).getCertificate         1
crypto/tls.defaultCipherSuites              1
crypto/tls.(*halfConn).error                1
crypto/tls.(*Conn).sendAlertLocked          1
crypto/tls.(*Conn).sendAlert                1
crypto/tls.(*Conn).Write                    1
crypto/tls.(*Conn).Close                    1
crypto/tls.(*clientHelloMsg).marshal        1
crypto/tls.(*clientHelloMsg).unmarshal        1
crypto/tls.(*clientKeyExchangeMsg).unmarshal        1
crypto/tls.(*finishedMsg).unmarshal         1
[...]
```

Yes, @goldshtn was right, there's more. So that has "/", "(", "*" and ")". Maybe trace.py should have a list of these and encode them as ASCII hex, eg, "/" becomes "\_2f\_".

goldshtn:
When argdist or trace face a function that has characters
in its name that are not valid in C identifiers, they now
replace these characters with an underscore (`_`) when
generating function names and structure names to include
in the BPF program. As a result, it is now possible to
trace functions that have these identifiers in their names,
such as Golang functions like `fmt.Println`.

Resolves #903 reported by @brendangregg.

### DIFF
--- a/tools/argdist.py
+++ b/tools/argdist.py
@@ -159,7 +159,7 @@ u64 __time = bpf_ktime_get_ns();
                 if parts[0] not in ["r", "p", "t", "u"]:
                         self._bail("probe type must be 'p', 'r', 't', or 'u'" +
                                    " but got '%s'" % parts[0])
-                if re.match(r"\w+\(.*\)", parts[2]) is None:
+                if re.match(r"\S+\(.*\)", parts[2]) is None:
                         self._bail(("function signature '%s' has an invalid " +
                                     "format") % parts[2])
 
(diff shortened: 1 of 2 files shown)


4ast:
makes sense to me.

brendangregg:
thanks!

PR #908, opened 2017-01-16
fix iteration over CPUs

### LINKED ISSUE #893: iterations over CPU: problem with offline CPUs
Since kernel version 4.9.0 BPF stopped working in a KVM guest. The problem are calls to `perf_event_open` with CPU identifiers which do not exist (`ENODEV`). The root cause for this is that the current code assumes ascending numbered CPUs.

However, on my KVM guest I have these online CPUs:
```
$ cat /sys/devices/system/cpu/online
0-11,16-27
```

There are other reasons for non-ascending CPU numbers aswell, e.g. CPU hotplugging.

I already started working on a patch (see https://github.com/Catalysts/bcc/commit/00e3b797f9994d7fc207dd44a5a81939850c2391), but unfortunately I can't get the `BPF_PERF_OUTPUT` table working. All unit tests pass except the one I added (`test_perf_buffer_for_each_cpu`).

I suspect the error happens due to the indexing of the `PerfEventArray`; the array contains `num_cpus` entries, therefore I had to map CPU identifiers to an array index: https://github.com/Catalysts/bcc/commit/00e3b797f9994d7fc207dd44a5a81939850c2391#diff-8f7d7667d81c5d2a4dbabef190179d8fL521
Probably I have to map the incoming data from each CPU ring buffer to the correct index too, but I couldn't find the place in the code (seems that everything is handled directly by callbacks).
Any tips?

Probably the `PerCpu*` Python data structures need to be updated as well.

andihit:
Since kernel version 4.9.0 BPF stopped working in a KVM guest.
The problem are calls to perf_event_open with CPU identifiers which do
not exist (ENODEV). The root cause for this is that the current code
assumes ascending numbered CPUs. However, this is not always the case
(e.g. CPU hotplugging).

This patch introduces the get_online_cpus() and get_possible_cpus()
helper functions and uses the appropriate function for iterations over
CPUs. The BPF_PERF_OUTPUT table contains now an entry for each possible
CPU instead of for each online CPU.

Fixes #893

### DIFF
--- a/src/cc/BPF.cc
+++ b/src/cc/BPF.cc
@@ -30,6 +30,7 @@
 #include "bpf_module.h"
 #include "libbpf.h"
 #include "perf_reader.h"
+#include "common.h"
 #include "usdt.h"
 
 #include "BPF.h"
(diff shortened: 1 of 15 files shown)


4ast:
[buildbot, ok to test]
LGTM

PR #913, opened 2017-01-18
Fix python2/3 incompatible percpu helpers

drzaeus77:
The python3 version of the percpu helpers (average, sum, etc.) were
using a python2 function that has since moved to functools (reduce).

Worse, the test case for percpu functionality was not enabled in the
cmake file. Better turn that on and make it work.

### DIFF
--- a/src/python/bcc/table.py
+++ b/src/python/bcc/table.py
@@ -14,6 +14,7 @@
 
 from collections import MutableMapping
 import ctypes as ct
+from functools import reduce
 import multiprocessing
 import os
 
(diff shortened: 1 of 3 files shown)


drzaeus77:
@4ast does the num possible cpu logic (and other changes of course) look ok?

src/python/bcc/table.py
```diff
@@ -617,7 +618,8 @@ def __init__(self, *args, **kwargs):
         self.reducer = kwargs.pop("reducer", None)
         super(PerCpuArray, self).__init__(*args, **kwargs)
         self.sLeaf = self.Leaf
-        self.total_cpu = multiprocessing.cpu_count()
+        with open("/sys/devices/system/cpu/possible") as f:
```
4ast: pls use get_possible_cpus() helper that was recently added

drzaeus77: Added to where?

4ast: src/python/bcc/utils.py

drzaeus77:
Should be good to go now.

PR #914, opened 2017-01-19
[tools][memleak.py] add parameter for specifying object to load malloc/free from

mkacik:
This commit adds a way to specify object to load malloc/free from, leaving previous value as a default. In cases when binary statically links different implementation of malloc, or preloads libc from different location, memleak in current form won't capture any samples. For statically linked malloc implementation use full path to binary as OBJ.

### DIFF
--- a/tools/memleak.py
+++ b/tools/memleak.py
@@ -135,6 +135,8 @@ parser.add_argument("-z", "--min-size", type=int,
         help="capture only allocations larger than this size")
 parser.add_argument("-Z", "--max-size", type=int,
         help="capture only allocations smaller than this size")
+parser.add_argument("-O", "--obj", type=str, default="c",
+        help="attach to malloc & free in the specified object")
 
 args = parser.parse_args()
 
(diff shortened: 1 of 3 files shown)


tools/memleak_example.txt
```diff
@@ -203,3 +205,4 @@ EXAMPLES:
         allocations that are at least one minute (60 seconds) old
 ./memleak -s 5
         Trace roughly every 5th allocation, to reduce overhead
+
```
goldshtn: Empty line?

tools/memleak.py
```diff
@@ -135,6 +135,8 @@ def run_command_get_pid(command):
         help="capture only allocations larger than this size")
 parser.add_argument("-Z", "--max-size", type=int,
         help="capture only allocations smaller than this size")
+parser.add_argument("-O", "--obj", type=str, default="c",
+        help="attach to malloc & free in specified object instead of default")
```
goldshtn: In *the* specified object instead of *the* default?

mkacik: Sure, went for shorter help to not exceed 80 chars. Now looking at it "instead of default" does not seem to be necessary there, so maybe just "attach to malloc & free in the specified object"?

tools/memleak_example.txt
mkacik: Artifact of copy-pasting the whole help. Let me update.

4ast:
[buildbot, ok to test]

Issue #872, opened 2016-12-22
multiple concurrent probe users, fixing "Device or resource busy"


brendangregg:
bcc currently has a limitation of one program per probe. Eg, you can run execsnoop & opensnoop at the same time, but not two opensnoops, as the second gets:

```
# ./opensnoop.py 
write(/sys/kernel/debug/tracing/kprobe_events, "p:kprobes/p_sys_open sys_open") failed: Device or resource busy
Traceback (most recent call last):
  File "./opensnoop.py", line 127, in <module>
    b.attach_kprobe(event="sys_open", fn_name="trace_entry")
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 467, in attach_kprobe
    raise Exception("Failed to attach BPF to kprobe")
Exception: Failed to attach BPF to kprobe
```

As bcc/BPF gets adopted, I'm seeing the risk of clashes between 24x7 bcc/BPF monitoring and ad hoc analysis, when the latter tries to instrument the same probes.

I suspect it's fixable in just bcc, without kernel changes. For ftrace, anyway, we can create two probes for the same kprobe location, provided they have different aliases, and trace_pipe will see both events:

```
# cd /sys/kernel/debug/tracing
# echo 'p:myopen1 do_sys_open' >> kprobe_events 
# echo 'p:myopen2 do_sys_open' >> kprobe_events 
# echo 1 > events/kprobes/myopen1/enable 
# echo 1 > events/kprobes/myopen2/enable 
# cat trace_pipe
    redis-server-1651  [001] d... 4455463.122021: myopen1: (do_sys_open+0x0/0x210)
    redis-server-1968  [000] d... 4455463.222274: myopen2: (do_sys_open+0x0/0x210)
    redis-server-1651  [001] d... 4455463.222274: myopen2: (do_sys_open+0x0/0x210)
    redis-server-1651  [001] d... 4455463.222276: myopen1: (do_sys_open+0x0/0x210)
    redis-server-1968  [000] d... 4455463.222278: myopen1: (do_sys_open+0x0/0x210)
       snmp-pass-2218  [002] d... 4455463.243604: myopen2: (do_sys_open+0x0/0x210)
       snmp-pass-2218  [002] d... 4455463.243609: myopen1: (do_sys_open+0x0/0x210)
       snmp-pass-2218  [002] d... 4455463.243842: myopen2: (do_sys_open+0x0/0x210)
[...]

PR #918, opened 2017-01-23
Handling multiple concurrent probe users.

derek0883:
Handle multiple concurrent probe users, fixed issue #872.
Event naming pattern changed to $(eventname)_bcc_$(pid)
Detect /sys/kernel/debug/tracing/instances in bpf_attach_probe,
if it exist, then will create per-instance event, if failed then make global
event instance as same as before.

### DIFF
--- a/src/cc/BPF.h
+++ b/src/cc/BPF.h
@@ -51,23 +51,23 @@ public:
 
   StatusTuple attach_kprobe(
       const std::string& kernel_func, const std::string& probe_func,
-      bpf_probe_attach_type = probe_entry,
+      bpf_probe_attach_type = BPF_PROBE_ENTRY,
       pid_t pid = -1, int cpu = 0, int group_fd = -1,
       perf_reader_cb cb = nullptr, void* cb_cookie = nullptr);
   StatusTuple detach_kprobe(
       const std::string& kernel_func,
-      bpf_probe_attach_type attach_type = probe_entry);
+      bpf_probe_attach_type attach_type = BPF_PROBE_ENTRY);
 
   StatusTuple attach_uprobe(
       const std::string& binary_path, const std::string& symbol,
       const std::string& probe_func, uint64_t symbol_addr = 0,
-      bpf_probe_attach_type attach_type = probe_entry,
+      bpf_probe_attach_type attach_type = BPF_PROBE_ENTRY,
       pid_t pid = -1, int cpu = 0, int group_fd = -1,
       perf_reader_cb cb = nullptr, void* cb_cookie = nullptr);
   StatusTuple detach_uprobe(
       const std::string& binary_path, const std::string& symbol,
       uint64_t symbol_addr = 0,
-      bpf_probe_attach_type attach_type = probe_entry);
+      bpf_probe_attach_type attach_type = BPF_PROBE_ENTRY);
   StatusTuple attach_usdt(const USDT& usdt, pid_t pid = -1, int cpu = 0,
                           int group_fd = -1);
   StatusTuple detach_usdt(const USDT& usdt);
@@ -118,9 +118,9 @@ private:
 
   std::string attach_type_debug(bpf_probe_attach_type type) {
     switch (type) {
-    case probe_entry:
+    case BPF_PROBE_ENTRY:
       return "";
-    case probe_return:
+    case BPF_PROBE_RETURN:
       return "return ";
     }
     return "ERROR";
@@ -128,9 +128,9 @@ private:
 
   std::string attach_type_prefix(bpf_probe_attach_type type) {
     switch (type) {
-    case probe_entry:
+    case BPF_PROBE_ENTRY:
       return "p";
-    case probe_return:
+    case BPF_PROBE_RETURN:
       return "r";
     }
     return "ERROR";
--- a/src/cc/BPF.h
+++ b/src/cc/BPF.h
@@ -29,11 +29,6 @@
 
 namespace ebpf {
 
-enum class bpf_attach_type {
-  probe_entry,
-  probe_return
-};
-
 struct open_probe_t {
   void* reader_ptr;
   std::string func;
@@ -56,23 +51,23 @@ public:
 
   StatusTuple attach_kprobe(
       const std::string& kernel_func, const std::string& probe_func,
-      bpf_attach_type attach_type = bpf_attach_type::probe_entry,
+      int attach_type = BPF_PROBE_ENTRY,
       pid_t pid = -1, int cpu = 0, int group_fd = -1,
       perf_reader_cb cb = nullptr, void* cb_cookie = nullptr);
   StatusTuple detach_kprobe(
       const std::string& kernel_func,
-      bpf_attach_type attach_type = bpf_attach_type::probe_entry);
+      int attach_type = BPF_PROBE_ENTRY);
 
   StatusTuple attach_uprobe(
       const std::string& binary_path, const std::string& symbol,
       const std::string& probe_func, uint64_t symbol_addr = 0,
-      bpf_attach_type attach_type = bpf_attach_type::probe_entry,
+      int attach_type = BPF_PROBE_ENTRY,
       pid_t pid = -1, int cpu = 0, int group_fd = -1,
       perf_reader_cb cb = nullptr, void* cb_cookie = nullptr);
   StatusTuple detach_uprobe(
       const std::string& binary_path, const std::string& symbol,
       uint64_t symbol_addr = 0,
-      bpf_attach_type attach_type = bpf_attach_type::probe_entry);
+      int attach_type = BPF_PROBE_ENTRY);
   StatusTuple attach_usdt(const USDT& usdt, pid_t pid = -1, int cpu = 0,
                           int group_fd = -1);
   StatusTuple detach_usdt(const USDT& usdt);
@@ -111,9 +106,9 @@ private:
   StatusTuple unload_func(const std::string& func_name);
 
   std::string get_kprobe_event(const std::string& kernel_func,
-                               bpf_attach_type type);
+                               int attach_type);
   std::string get_uprobe_event(const std::string& binary_path, uint64_t offset,
-                               bpf_attach_type type);
+                               int attach_type);
 
   StatusTuple detach_kprobe_event(const std::string& event, open_probe_t& attr);
   StatusTuple detach_uprobe_event(const std::string& event, open_probe_t& attr);
@@ -121,21 +116,21 @@ private:
                                       open_probe_t& attr);
   StatusTuple detach_perf_event_all_cpu(open_probe_t& attr);
 
-  std::string attach_type_debug(bpf_attach_type type) {
-    switch (type) {
-    case bpf_attach_type::probe_entry:
+  std::string attach_type_debug(int attach_type) {
+    switch (attach_type) {
+    case BPF_PROBE_ENTRY:
       return "";
-    case bpf_attach_type::probe_return:
+    case BPF_PROBE_RETURN:
       return "return ";
     }
     return "ERROR";
   }
 
-  std::string attach_type_prefix(bpf_attach_type type) {
-    switch (type) {
-    case bpf_attach_type::probe_entry:
+  std::string attach_type_prefix(int attach_type) {
+    switch (attach_type) {
+    case BPF_PROBE_ENTRY:
       return "p";
-    case bpf_attach_type::probe_return:
+    case BPF_PROBE_RETURN:
       return "r";
     }
     return "ERROR";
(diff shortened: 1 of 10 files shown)


4ast:
lgtm
[buildbot, ok to test]
@drzaeus77 @palmtenor @markdrayton @goldshtn @brendangregg 
please review.
I think it's a good change and I don't expect a fallout from it,
but I may not have thought of all cases.

derek0883:
@4ast @brendangregg 
I have fixed the auto testing failure. :)

goldshtn:
I'm just wondering: why can't the logic for creating the "_bcc_$PID" pattern be on the C side, so that we don't have to change the external API at all? Right now it looks like we're going to have to add this for C++, Python, and Lua (BTW I don't see Lua on this PR).

derek0883:
@goldshtn 
That because when C code bpf_attach_kprobe get called, the argument is:
event_desc = p:kprobes/p_sys_open sys_open
event = sys_open
Front-end library will generate event_desc based on event name.
technically I can do that in C, But If I do that in C code,  I need replace
```
p:kprobes/p_sys_open           sys_open  
p:kprobes/p_sys_open_bcc_$PID  sys_open
```
I feel that This change will make that piece of code a little weird. 
I also thought that I can ignore event_desc, just do something like this:
snprintf(new_desc, "p:kprobe/p_%s_bcc_%d %s", event, getpid(), event);
Then I thought ignore a argument without front-end notice is not good neither. 
And we add _bcc_$PID in front-end library, not end-user script writer, this is 1 time change. 
Do you think this is acceptable?
Yes, I forgot to change LUA script, I will do that later tonight. thanks.

goldshtn:
Well, if we're making a change to the BCC library API anyway, maybe we can get rid of the `event_desc` and have the C side generate that? @4ast @drzaeus77

drzaeus77:
> Well, if we're making a change to the BCC library API anyway, maybe we can get rid of the event_desc and have the C side generate that? @4ast @drzaeus77

That sounds reasonable to me. Projects that link against bcc (e.g. gobpf) will have to adjust but I think that's ok.

derek0883:
> Well, if we're making a change to the BCC library API anyway, maybe we can get rid of the event_desc and have the C side generate that? 

So in bpf_attach_probe, remove event_desc argument, right?  This is easier for kprobe.
I just tried to make this change, I almost finished bpf_attach_kprobe, 
while I worked on bpf_attach_uprobe, I realized that bpf_attach_uprobe might want keep event_desc. 
e.g. in src/cc/BPF.cc attach_uprobe.  

```c++
  std::string probe_event_desc = attach_type_prefix(attach_type);
  probe_event_desc += ":uprobes/" + probe_event + " ";
  probe_event_desc += binary_path + ":0x" + uint_to_hex(sym.offset);
  void* res =
      bpf_attach_uprobe(probe_fd, probe_event.c_str(), probe_event_desc.c_str(),
                        pid, cpu, group_fd, cb, cb_cookie);
```
without probe_event_desc,  We need pass binary_path and sym.offset to bpf_attach_uprobe in libbpf.c

currently in src/cc/libbpf.c, kprobe and uprobe interface are same.  If we make this change, 
```c
void * bpf_attach_kprobe(int progfd, const char *event,
                        const char *event_desc,
                        pid_t pid, int cpu, int group_fd,
                        perf_reader_cb cb, void *cb_cookie) {
  return bpf_attach_probe(progfd, event, event_desc, "kprobe", pid, cpu, group_fd, cb, cb_cookie);
}

void * bpf_attach_uprobe(int progfd, const char *event,
                        const char *event_desc, 
                        pid_t pid, int cpu, int group_fd,
                        perf_reader_cb cb, void *cb_cookie) {
  return bpf_attach_probe(progfd, event, event_desc, "uprobe", pid, cpu, group_fd, cb, cb_cookie);
}   

1. either remove event_desc for kprobe, keep event_desc for uprobe
2. or remove event_desc for both, and pass binary_path and sym.offset for uprobe only.

this will make the interface different.  Maybe keep both interface same is better for front-end user ?
Another reason to keep in front-end is that, maybe kernel will add some more syntax for it,something like dynamic_debug the kernel current have. e.g.

```
# echo 'file svcsock.c line 1603 +p'  >  <debugfs>/dynamic_debug/control
```
line could be a line number or rang, e.g. 10-100. 
keep it in front-end, so user can change python script debug it without rebuild C lib?  :)

**Comment by derek0883** [comment]
@vmg 
Hi Vincent 
Would you please help me take a look how to make this changes for LUA.
I think the discussing and decision about  event_desc takes time. I'd like to make lua part for current changes. the change is simple. 
1. Add _bcc_$PID for ev_name.
2. for detach kprobe, currently pass event_desc, we need pass two argument,  event_desc, and ev_name.
I tried the changes by myself. I only added 1 line to  src/lua/bcc/bpf.lua
```
local posix = require("posix")
``` 
After rebuild 
```bash
build$ sudo ./src/lua/bcc-lua ../tools/biosnoop.lua
./src/lua/bcc-lua: attempt to write to undeclared variable posix
stack traceback:
	[C]: in function 'error'
	bcc.lua:1838: in function <bcc.lua:1837>
	[C]: at 0x7fd16c351ad0
	[C]: in function 'require'
	/usr/share/lua/5.1/posix.lua:2: in main chunk
	[C]: in function 'require'
	bcc.lua:1968: in function <bcc.lua:1946>
	[C]: in function 'require'
	bcc.lua:5776: in function <bcc.lua:5736>
	[C]: at 0x00793580
```
Would you please help me make the changes. :) Thanks!

4ast:
> need pass binary_path and sym.offset to bpf_attach_uprobe in libbpf.c?

yes. i think that's a good plan. Things like ":0x" + uint_to_hex(sym.offset) should
ideally be hidden within libbpf.c

derek0883:
@4ast 
Thanks for confirming that!  Then I'm going to remove event_desc for uprobe too, generate it inside libbpf.c. :)

brendangregg:
I'm back online. FWIW, I tested the current code and it works. :)

4ast:
@derek0883 you're still working on moving suffix gen into libbpf.c, I assume, right?

derek0883:
@brendangregg Thanks for confirm that.
@4ast  Yes, I'm going to send another PR for review this weekend. 
This version made more changes than previous one. :)

derek0883:
@4ast I submitted new fix for review.  
Moved event desc to libbpf.c, So front-end don't need to handle that.  But changed C API.  :)

src/cc/BPF.h
```diff
@@ -29,11 +29,6 @@
 
 namespace ebpf {
 
-enum class bpf_attach_type {
```
4ast: could you keep it as 'enum bpf_attach_type' inside libbpf.h instead of introducing #define -s ?
compiler unfortunately consider int == enum for argument checking purpose, but it's still cleaner api.

src/cc/BPF.h
```diff
@@ -111,31 +106,31 @@ class BPF {
...
   StatusTuple detach_uprobe_event(const std::string& event, open_probe_t& attr);
   StatusTuple detach_tracepoint_event(const std::string& tracepoint,
                                       open_probe_t& attr);
   StatusTuple detach_perf_event_all_cpu(open_probe_t& attr);
 
-  std::string attach_type_debug(bpf_attach_type type) {
-    switch (type) {
-    case bpf_attach_type::probe_entry:
+  std::string attach_type_debug(int attach_type) {
+    switch (attach_type) {
```
4ast: for example, by keeping enum compiler will be able to warn here that some enum values are not handled in the switch...

src/cc/libbpf.c
```diff
@@ -336,14 +338,18 @@ static int bpf_attach_tracing_event(int progfd, const char *event_path,
...
-                               pid_t pid, int cpu, int group_fd,
-                               perf_reader_cb cb, void *cb_cookie) {
+void * bpf_attach_kprobe(int progfd, int attach_type, const char *ev_name,
+                        const char *fn_name,
+                        pid_t pid, int cpu, int group_fd,
+                        perf_reader_cb cb, void *cb_cookie) 
+{
   int kfd;
-  char buf[256];
+  char buf[strlen(ev_name)+256];
```
4ast: i'm not sure what this 'strlen(ev_name)+256' math suppose to do.
if i understand it correctly, the size of this array is still a rough guess of how much buffer space is needed.
if so, just statically set it to 256. all ev_names suppose to be short?

src/cc/libbpf.c
```diff
@@ -355,52 +361,98 @@ static void * bpf_attach_probe(int progfd, const char *event,
...
     goto error;
   }
   close(kfd);
 
-  snprintf(buf, sizeof(buf), "/sys/kernel/debug/tracing/events/%ss/%s", event_type, event);
+  if (access("/sys/kernel/debug/tracing/instances", F_OK) != -1) {
+    snprintf(buf, sizeof(buf), "/sys/kernel/debug/tracing/instances/%s", new_name);
+    if (mkdir(buf, 0755) == -1) 
+      goto retry;
+    snprintf(buf, sizeof(buf), "/sys/kernel/debug/tracing/instances/%s/events/%ss/%s",
```
4ast: this one is probably the largest printed string, so makes to check the value of snprintf for overflow.

src/cc/libbpf.c
```diff
@@ -355,52 +361,98 @@ static void * bpf_attach_probe(int progfd, const char *event,
...
-                        const char *event_desc,
+void * bpf_attach_uprobe(int progfd, int attach_type, const char *ev_name,
+                        const char *binary_path, uint64_t offset,
                         pid_t pid, int cpu, int group_fd,
-                        perf_reader_cb cb, void *cb_cookie) {
-  return bpf_attach_probe(progfd, event, event_desc, "uprobe", pid, cpu, group_fd, cb, cb_cookie);
+                        perf_reader_cb cb, void *cb_cookie) 
+{
+  int kfd;
+  char buf[strlen(binary_path)+strlen(ev_name)+256];
```
4ast: same here. imo 256 + snprintf overflow check would be cleaner. no need to check all snprintfs only the largest.

vmg:
Hey @derek0883! Sorry I missed your ping. I don't see the `local posix = require("posix")` line in the diff. Could you please push the broken code so I can take a look at it? Thank you!

derek0883:
@4ast I have submitted new changes for review. 
1.  keep enum bpf_attach_type in libbpf.h, but I renamed it to bpf_probe_attach_type, as we have another "enum bpf_attach_type" in  src/cc/compat/linux/bpf.h. this will cause compiling error for b_frontend_action.cc.
2. using static buf size plus check for snprintf. In bpf_attach_uprobe, as binary_path is not fixed, 256 might not enough, especially for non-english system. So I set the size to linux pre-defined macro PATH_MAX.
Feel free to let me know if you have other suggestion :).  Thanks.

@vmg 
Hi vincent, Thanks for your response. I have moved _bcc_pid stuff to libbpf.c, So front-end side such as python or LUA don't need to handle that. So I don't have that require("posix") question any more. :)

vmg:
:+1: Fantastic.

src/cc/BPF.h
```diff
@@ -56,23 +51,23 @@ class BPF {
 
   StatusTuple attach_kprobe(
       const std::string& kernel_func, const std::string& probe_func,
-      bpf_attach_type attach_type = bpf_attach_type::probe_entry,
+      bpf_probe_attach_type = probe_entry,
```
4ast: how does this compile? or this is a github diff view bug?

src/cc/libbpf.h
```diff
@@ -24,6 +24,11 @@
 extern "C" {
 #endif
 
+enum bpf_probe_attach_type {
+	probe_entry,
```
4ast: please make it upper case. Like: BPF_PROBE_ENTRY
to reduce the chance of conflicts.

tools/argdist.py
```diff
@@ -20,7 +20,7 @@
 class Probe(object):
         next_probe_index = 0
         streq_index = 0
-        aliases = {"$PID": "bpf_get_current_pid_tgid()"}
+        aliases = {"$PID": "(bpf_get_current_pid_tgid() >> 32)"}
```
4ast: hmm. a merge conflict? this patch already applied.

derek0883:
@4ast 
1. Are you refer to :: in the bpf_attach_type::probe_entry?  As I moved the enum to libbpf.h, It protected by extern "C" {,  I removed the class key word. So we don't need :: to access the member. 
2. Changed it to BPF_PROBE_ENTRY and BPF_PROBE_RETURN.
3. That is a conflict, I I saw the page shows "This branch is out-of-date with the base branch". So I did  a "git rebase".  turns out not necessary.  Now no conflict.  
Feel free to let me know if you have other suggestion :). Thanks.

brendangregg:
many thanks!

PR #936, opened 2017-02-03
tools: add tool to detect potential deadlocks in running programs

kennyyu:
`deadlock_detector` is a new tool to detect potential deadlocks (lock order
inversions) in a running process. The program attaches uprobes on
`pthread_mutex_lock` and `pthread_mutex_unlock` to build a mutex wait directed
graph, and then looks for a cycle in this graph. This graph has the following
properties:

- Nodes in the graph represent mutexes.
- Edge (A, B) exists if there exists some thread T where lock(A) was called
  and lock(B) was called before unlock(A) was called.

If there is a cycle in this graph, this indicates that there is a lock order
inversion (potential deadlock). If the program finds a lock order inversion, the
program will dump the cycle of mutexes, dump the stack traces where each mutex
was acquired, and then exit.

The format of the output uses a similar output as ThreadSanitizer (See example:
https://github.com/google/sanitizers/wiki/ThreadSanitizerDeadlockDetector)

This program can only find potential deadlocks that occur while the program is
tracing the process. It cannot find deadlocks that may have occurred before the
program was attached to the process.

If the traced process has many mutexes and threads, this program will add a
very large overhead because every mutex lock/unlock and clone call will be
traced. This tool is meant for debugging only, and you should run this tool
only on programs where the slowdown is acceptable.

Note: This tool does not work for shared mutexes or recursive mutexes.

For shared (read-write) mutexes, a deadlock requires a cycle in the wait
graph where at least one of the mutexes in the cycle is acquiring exclusive
(write) ownership.

For recursive mutexes, lock() is called multiple times on the same mutex.
However, there is no way to determine if a mutex is a recursive mutex
after the mutex has been created. As a result, this tool will not find
potential deadlocks that involve only one mutex.

### DIFF
--- a/tools/deadlock_detector.py
+++ b/tools/deadlock_detector.py
@@ -387,11 +387,36 @@ def strlist(s):
 
 
 def main():
+    examples = '''Examples:
+    deadlock_detector 181        # Analyze PID 181
+
+    deadlock_detector 181 --binary /lib/x86_64-linux-gnu/libpthread.so.0
+                                 # Analyze PID 181 and locks from this binary.
+                                 # If tracing a process that is running from
+                                 # a dynamically-linked binary, this argument
+                                 # is required and should be the path to the
+                                 # pthread library.
+
+    deadlock_detector 181 --verbose
+                                 # Analyze PID 181 and print statistics about
+                                 # the mutex wait graph.
+
+    deadlock_detector 181 --lock-symbols my_mutex_lock1,my_mutex_lock2 \\
+        --unlock-symbols my_mutex_unlock1,my_mutex_unlock2
+                                 # Analyze PID 181 and trace custom mutex
+                                 # symbols instead of pthread mutexes.
+
+    deadlock_detector 181 --dump-graph graph.json
+                                 # Analyze PID 181 and dump the mutex wait
+                                 # graph to graph.json.
+    '''
     parser = argparse.ArgumentParser(
         description=(
             'Detect potential deadlocks (lock inversions) in a running binary.'
-            ' Must be run as root.'
-        )
+            '\nMust be run as root.'
+        ),
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        epilog=examples,
     )
     parser.add_argument('pid', type=int, help='Pid to trace')
     # Binaries with `:` in the path will fail to attach uprobes on kernels
@@ -401,7 +426,11 @@ def main():
         '--binary',
         type=str,
         default='',
-        help='If set, use this as the path to the binary for the process.',
+        help='If set, trace the mutexes from the binary at this path. '
+        'For statically-linked binaries, this argument is not required. '
+        'For dynamically-linked binaries, this argument is required and '
+        'should be the path of the pthread library the binary is using. '
+        'Example: /lib/x86_64-linux-gnu/libpthread.so.0',
     )
     parser.add_argument(
         '--dump-graph',
@@ -419,14 +448,14 @@ def main():
         type=strlist,
         default=['pthread_mutex_lock'],
         help='Comma-separated list of lock symbols to trace. Default is '
-        'pthread_mutex_lock',
+        'pthread_mutex_lock. These symbols cannot be inlined in the binary.',
     )
     parser.add_argument(
         '--unlock-symbols',
         type=strlist,
         default=['pthread_mutex_unlock'],
         help='Comma-separated list of unlock symbols to trace. Default is '
-        'pthread_mutex_unlock',
+        'pthread_mutex_unlock. These symbols cannot be inlined in the binary.',
     )
     args = parser.parse_args()
     if not args.binary:
--- a/tools/deadlock_detector.py
+++ b/tools/deadlock_detector.py
@@ -367,7 +367,7 @@ def print_cycle(binary, graph, edges, thread_info, print_stack_trace_fn):
 
 def symbolize_with_objdump(binary, addr):
     '''
-    Searches the biniary for the address using objdump. Returns the symbol if
+    Searches the binary for the address using objdump. Returns the symbol if
     it is found, otherwise returns empty string.
     '''
     try:
--- a/tools/deadlock_detector.py
+++ b/tools/deadlock_detector.py
@@ -1,0 +1,282 @@
+#!/usr/bin/env python
+#
+# deadlock_detector  Detects potential deadlocks (lock order inversions)
+#                    on a running process. For Linux, uses BCC, eBPF.
+#
+# USAGE: deadlock_detector.py [-h] [--dump-graph DUMP_GRAPH]
+#                             [--lock-symbols LOCK_SYMBOLS]
+#                             [--unlock-symbols UNLOCK_SYMBOLS]
+#                             binary pid
+#
+# This traces pthread mutex lock and unlock calls to build a directed graph
+# representing the mutex wait graph:
+#
+# - Nodes in the graph represent mutexes.
+# - Edge (A, B) exists if there exists some thread T where lock(A) was called
+#   and lock(B) was called before unlock(A) was called.
+#
+# If the program finds a potential lock order inversion, the program will dump
+# the cycle of mutexes and the stack traces where each mutex was acquired, and
+# then exit.
+#
+# Note: This tool does not work for shared mutexes or recursive mutexes.
+#
+# For shared (read-write) mutexes, a deadlock requires a cycle in the wait
+# graph where at least one of the mutexes in the cycle is acquiring exclusive
+# (write) ownership.
+#
+# For recursive mutexes, lock() is called multiple times on the same mutex.
+# However, there is no way to determine if a mutex is a recursive mutex
+# after the mutex has been created. As a result, this tool will not find
+# potential deadlocks that involve only one mutex.
+#
+# Copyright 2017 Facebook, Inc.
+# Licensed under the Apache License, Version 2.0 (the "License")
+#
+# 01-Feb-2017   Kenny Yu   Created this.
+
+from __future__ import (
+    absolute_import, division, unicode_literals, print_function
+)
+from bcc import BPF
+import argparse
+import json
+import networkx
+from networkx.readwrite import json_graph
+import sys
+import time
+
+
+def find_cycle(graph):
+    '''
+    Looks for a cycle in the graph. If found, returns the first cycle.
+    If nodes a1, a2, ..., an are in a cycle, then this returns:
+
+    [(a1,a2), (a2,a3), ... (an-1,an), (an, a1)]
+
+    Otherwise returns an empty list.
+    '''
+    cycles = list(networkx.simple_cycles(graph))
+    if cycles:
+        nodes = cycles[0]
+        nodes.append(nodes[0])
+        edges = []
+        prev = nodes[0]
+        for node in nodes[1:]:
+            edges.append((prev, node))
+            prev = node
+        return edges
+    else:
+        return []
+
+
+def find_cycle_and_print(graph, thread_to_parent, print_stack_trace_fn):
+    '''
+    Detects if there is a cycle in the mutex graph. If there is, dump
+    the following information and return True. Otherwise returns False.
+
+    Potential Deadlock Detected!
+
+    Cycle in lock order graph: M0 => M1 => M2 => M0
+
+    for (m, n) in cycle:
+        Mutex n acquired here while holding Mutex m in thread T:
+            [ stack trace ]
+
+        Mutex m previously acquired by thread T here:
+            [ stack trace ]
+
+    for T in all threads:
+        Thread T was created here:
+            [ stack trace ]
+    '''
+    edges = find_cycle(graph)
+    if not edges:
+        return False
+
+    # List of mutexes in the cycle, first and last repeated
+    nodes_in_order = []
+    # Map mutex address -> readable alias
+    node_addr_to_name = {}
+    for counter, (m, n) in enumerate(edges):
+        nodes_in_order.append(m)
+        node_addr_to_name[m] = 'Mutex M%d (0x%016x)' % (counter, m)
+    nodes_in_order.append(nodes_in_order[0])
+
+    print('----------------\nPotential Deadlock Detected!\n')
+    print(
+        'Cycle in lock order graph: %s\n' %
+        (' => '.join([node_addr_to_name[n] for n in nodes_in_order]))
+    )
+
+    # Set of threads involved in the lock inversion
+    thread_pids = set()
+
+    # For each edge in the cycle, print where the two mutexes were held
+    for (m, n) in edges:
+        thread_pid = graph[m][n]['thread_pid']
+        thread_comm = graph[m][n]['thread_comm']
+        first_mutex_stack_id = graph[m][n]['first_mutex_stack_id']
+        second_mutex_stack_id = graph[m][n]['second_mutex_stack_id']
+        thread_pids.add(thread_pid)
+        print(
+            '%s acquired here while holding %s in Thread %d (%s):' % (
+                node_addr_to_name[n], node_addr_to_name[m], thread_pid,
+                thread_comm
+            )
+        )
+        print_stack_trace_fn(second_mutex_stack_id)
+        print('')
+        print(
+            '%s previously acquired by the same Thread %d (%s) here:' %
+            (node_addr_to_name[m], thread_pid, thread_comm)
+        )
+        print_stack_trace_fn(first_mutex_stack_id)
+        print('')
+
+    # Print where the threads were created, if available
+    for thread_pid in thread_pids:
+        parent_pid, stack_id, parent_comm = thread_to_parent.get(
+            thread_pid, (None, None, None)
+        )
+        if parent_pid:
+            print(
+                'Thread %d created by Thread %d (%s) here: ' %
+                (thread_pid, parent_pid, parent_comm)
+            )
+            print_stack_trace_fn(stack_id)
+        else:
+            print(
+                'Could not find stack trace where Thread %d was created' %
+                thread_pid
+            )
+        print('')
+    return True
+
+
+def strlist(s):
+    '''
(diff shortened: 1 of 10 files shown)


kennyyu:
cc @4ast

4ast:
[buildbot, ok to test]

4ast:
@drzaeus77 suggestions on how to fix dependencies on builtbots?
'dpkg-checkbuilddeps: error: Unmet build dependencies: python-networkx'

tools/deadlock_detector_example.txt
```diff
@@ -0,0 +1,315 @@
...
+
+
+# ./deadlock_detector.py /path/to/program $(pidof program) --lock-symbols custom_mutex1_lock,custom_mutex2_lock --unlock_symbols custom_mutex1_unlock,custom_mutex2_unlock
+
+Tracing... Hit Ctrl-C to end.
+Nodes: 0, Edges: 0, Looking for cycle took 0.000062 seconds
+Nodes: 0, Edges: 0, Looking for cycle took 0.000066 seconds
+Nodes: 0, Edges: 0, Looking for cycle took 0.000065 seconds
+Nodes: 0, Edges: 0, Looking for cycle took 0.000053 seconds
+Nodes: 102, Edges: 4936, Looking for cycle took 0.155751 seconds
```
4ast: overall looks awesome! Few small nits:
what is the reason to measure and print time that cycle finder takes? Most of the time is spent by networkx library, right? and in some cases it can be slow? May be print it only when verbose mode on?
or may be make it adjustable interval instead of looking for cycles every second?
'nodes' and 'edges' in the output are not very human friendly.
May be print the number of mutexes traced during the interval?

goldshtn:
Looks very interesting. Is there a way to get rid of the external dependency? We really try to minimize any use of Python modules that aren't in-box for easier distribution and OOTB use ...

tools/deadlock_detector.py
```diff
@@ -0,0 +1,282 @@
...
+
+def main():
+    parser = argparse.ArgumentParser(
+        description=(
+            'Detect potential deadlocks (lock inversions) in a running binary.'
+            ' Must be run as root.'
+        )
+    )
+    parser.add_argument('binary', type=str, help='Absolute path to binary')
+    parser.add_argument('pid', type=int, help='Pid to trace')
```
goldshtn: Why do we need both the binary and the pid? Looks like one can be obtained from the other.

tools/deadlock_detector.py
```diff
@@ -0,0 +1,282 @@
...
+            }
+
+            start = time.time()
+            has_cycle = find_cycle_and_print(
+                graph, thread_to_parent_dict, print_stack_trace
+            )
+            print(
+                'Nodes: %d, Edges: %d, Looking for cycle took %f seconds' %
+                (len(graph.nodes()), len(graph.edges()), (time.time() - start))
+            )
```
goldshtn: This should probably be in verbose output only.

tools/deadlock_detector.py
```diff
@@ -0,0 +1,282 @@
...
+    if not edges:
+        return False
+
+    # List of mutexes in the cycle, first and last repeated
+    nodes_in_order = []
+    # Map mutex address -> readable alias
+    node_addr_to_name = {}
+    for counter, (m, n) in enumerate(edges):
+        nodes_in_order.append(m)
+        node_addr_to_name[m] = 'Mutex M%d (0x%016x)' % (counter, m)
```
goldshtn: Would be nice to try resolving the mutex address to a symbol, this will work nicely with global or static locks.

tools/deadlock_detector.py
kennyyu: We can resolve the pid to a binary by looking at /proc/pid/exe, but we had issues with this because:

- Some of our binaries have `:` in the path name (e.g. IP addresses)
- On kernels without this recent kernel patch (https://lkml.org/lkml/2017/1/13/585, submitted 2 weeks ago), we can't attach a uprobe on that binary because of the `:` in the path name
- However, if we have a symlink to the binary, it should work.

tools/deadlock_detector.py
kennyyu: Yes! agreed.

tools/deadlock_detector_example.txt
kennyyu: - I'll get rid of the time, it doesn't really help much. 
- I think it will be better to print the final graph size (number of mutexes and edges) only if the user presses Ctrl-C, or the program finds a deadlock.

tools/deadlock_detector.py
kennyyu: @goldshtn Do you know of an easy way of doing this? I tried `bpf.sym(address, pid)` (https://github.com/iovisor/bcc/blob/master/src/python/bcc/__init__.py#L947) on these mutex addresses, and they always returned `[unknown]`, even for global and static locks.

kennyyu:
@goldshtn @4ast Thanks for the feedback!

@goldshtn As for the networkx dependency: the only thing I really need from networkx is the cycle detection functionality: https://github.com/networkx/networkx/blob/b6dcdf979e8bcf3c0c0cd5d60bb70e5380e3b8b4/networkx/algorithms/cycles.py#L105

It might be feasible to copy and paste only the code that I need from networkx instead of pulling in the entire library as a dependency.

tools/deadlock_detector.py
goldshtn: Did you compile your test program with `-g`? Do you see these addresses in `objdump -tT yourbinary`?

tools/deadlock_detector.py
goldshtn: Hmm. Well it should work for most binaries I guess :) Maybe make the binary argument optional? Or even something like `-p PID [-x BIN]`?

tools/deadlock_detector.py
kennyyu: @goldshtn That works perfectly! I can get the symbol names for global and static mutexes, and I can also the `pthread_mutex_lock` symbol when previously, bpf would only return `[unknown]`. 

Example program:

```c++
#include <chrono>
#include <iostream>
#include <mutex>
#include <thread>

std::mutex global_mutex1;
std::mutex global_mutex2;

int main(void) {
  static std::mutex static_mutex3;
  std::mutex local_mutex4;

  std::cout << "sleeping for a bit to allow trace to attach..." << std::endl;
  std::this_thread::sleep_for(std::chrono::seconds(10));
  std::cout << "starting program..." << std::endl;

  auto t1 = std::thread([] {
    std::lock_guard<std::mutex> g1(global_mutex1);
    std::lock_guard<std::mutex> g2(global_mutex2);
  });
  t1.join();

  auto t2 = std::thread([] {
    std::lock_guard<std::mutex> g2(global_mutex2);
    std::lock_guard<std::mutex> g3(static_mutex3);
  });
  t2.join();

  auto t3 = std::thread([&local_mutex4] {
    std::lock_guard<std::mutex> g3(static_mutex3);
    std::lock_guard<std::mutex> g4(local_mutex4);
  });
  t3.join();

  auto t4 = std::thread([&local_mutex4] {
    std::lock_guard<std::mutex> g4(local_mutex4);
    std::lock_guard<std::mutex> g1(global_mutex2);
  });
  t4.join();

  std::cout << "sleeping to allow trace to collect data..." << std::endl;
  std::this_thread::sleep_for(std::chrono::seconds(5));
  std::cout << "done!" << std::endl;
}
```

Here's the output for that program:

```
Potential Deadlock Detected!

Cycle in lock order graph: Mutex M0 (global_mutex2 0x0000000000473c20) => Mutex M1 (main::static_mutex3 0x0000000000473c60) => Mutex M2 (0x00007ffca57f73d0) => Mutex M0 (global_mutex2 0x0000000000473c20)

Mutex M1 (main::static_mutex3 0x0000000000473c60) acquired here while holding Mutex M0 (global_mutex2 0x0000000000473c20) in Thread 3198691 (lockinversion):
@ 00000000004024d0 pthread_mutex_lock
@ 0000000000406dd0 std::mutex::lock()
@ 00000000004070d2 std::lock_guard<std::mutex>::lock_guard(std::mutex&)
@ 0000000000402dc9 main::{lambda()#2}::operator()() const
@ 0000000000406c0a void std::_Bind_simple<main::{lambda()#2} ()>::_M_invoke<>(std::_Index_tuple<>)
@ 0000000000406975 std::_Bind_simple<main::{lambda()#2} ()>::operator()()
@ 0000000000406758 std::thread::_Impl<std::_Bind_simple<main::{lambda()#2} ()> >::_M_run()
@ 00007f374b5df4e1 execute_native_thread_routine
@ 00007f374bd5e7f1 start_thread
@ 00007f374b02046d __clone

Mutex M0 (global_mutex2 0x0000000000473c20) previously acquired by the same Thread 3198691 (lockinversion) here:
@ 00000000004024d0 pthread_mutex_lock
@ 0000000000406dd0 std::mutex::lock()
@ 00000000004070d2 std::lock_guard<std::mutex>::lock_guard(std::mutex&)
@ 0000000000402db8 main::{lambda()#2}::operator()() const
@ 0000000000406c0a void std::_Bind_simple<main::{lambda()#2} ()>::_M_invoke<>(std::_Index_tuple<>)
@ 0000000000406975 std::_Bind_simple<main::{lambda()#2} ()>::operator()()
@ 0000000000406758 std::thread::_Impl<std::_Bind_simple<main::{lambda()#2} ()> >::_M_run()
@ 00007f374b5df4e1 execute_native_thread_routine
@ 00007f374bd5e7f1 start_thread
@ 00007f374b02046d __clone

Mutex M2 (0x00007ffca57f73d0) acquired here while holding Mutex M1 (main::static_mutex3 0x0000000000473c60) in Thread 3198692 (lockinversion):
@ 00000000004024d0 pthread_mutex_lock
@ 0000000000406dd0 std::mutex::lock()
@ 00000000004070d2 std::lock_guard<std::mutex>::lock_guard(std::mutex&)
@ 0000000000402e38 main::{lambda()#3}::operator()() const
@ 0000000000406ba8 void std::_Bind_simple<main::{lambda()#3} ()>::_M_invoke<>(std::_Index_tuple<>)
@ 0000000000406951 std::_Bind_simple<main::{lambda()#3} ()>::operator()()
@ 000000000040673a std::thread::_Impl<std::_Bind_simple<main::{lambda()#3} ()> >::_M_run()
@ 00007f374b5df4e1 execute_native_thread_routine
@ 00007f374bd5e7f1 start_thread
@ 00007f374b02046d __clone

Mutex M1 (main::static_mutex3 0x0000000000473c60) previously acquired by the same Thread 3198692 (lockinversion) here:
@ 00000000004024d0 pthread_mutex_lock
@ 0000000000406dd0 std::mutex::lock()
@ 00000000004070d2 std::lock_guard<std::mutex>::lock_guard(std::mutex&)
@ 0000000000402e22 main::{lambda()#3}::operator()() const
@ 0000000000406ba8 void std::_Bind_simple<main::{lambda()#3} ()>::_M_invoke<>(std::_Index_tuple<>)
@ 0000000000406951 std::_Bind_simple<main::{lambda()#3} ()>::operator()()
@ 000000000040673a std::thread::_Impl<std::_Bind_simple<main::{lambda()#3} ()> >::_M_run()
@ 00007f374b5df4e1 execute_native_thread_routine
@ 00007f374bd5e7f1 start_thread
@ 00007f374b02046d __clone

Mutex M0 (global_mutex2 0x0000000000473c20) acquired here while holding Mutex M2 (0x00007ffca57f73d0) in Thread 3198693 (lockinversion):
@ 00000000004024d0 pthread_mutex_lock
@ 0000000000406dd0 std::mutex::lock()
@ 00000000004070d2 std::lock_guard<std::mutex>::lock_guard(std::mutex&)
@ 0000000000402ea8 main::{lambda()#4}::operator()() const
@ 0000000000406b46 void std::_Bind_simple<main::{lambda()#4} ()>::_M_invoke<>(std::_Index_tuple<>)
@ 000000000040692d std::_Bind_simple<main::{lambda()#4} ()>::operator()()
@ 000000000040671c std::thread::_Impl<std::_Bind_simple<main::{lambda()#4} ()> >::_M_run()
@ 00007f374b5df4e1 execute_native_thread_routine
@ 00007f374bd5e7f1 start_thread
@ 00007f374b02046d __clone

Mutex M2 (0x00007ffca57f73d0) previously acquired by the same Thread 3198693 (lockinversion) here:
@ 00000000004024d0 pthread_mutex_lock
@ 0000000000406dd0 std::mutex::lock()
@ 00000000004070d2 std::lock_guard<std::mutex>::lock_guard(std::mutex&)
@ 0000000000402e97 main::{lambda()#4}::operator()() const
@ 0000000000406b46 void std::_Bind_simple<main::{lambda()#4} ()>::_M_invoke<>(std::_Index_tuple<>)
@ 000000000040692d std::_Bind_simple<main::{lambda()#4} ()>::operator()()
@ 000000000040671c std::thread::_Impl<std::_Bind_simple<main::{lambda()#4} ()> >::_M_run()
@ 00007f374b5df4e1 execute_native_thread_routine
@ 00007f374bd5e7f1 start_thread
@ 00007f374b02046d __clone

Thread 3198691 created by Thread 3198375 (lockinversion) here:
@ 00007f374b020431 __clone
@ 00007f374bd5eef5 pthread_create
@ 00007f374b5e1440 std::thread::_M_start_thread(std::shared_ptr<std::thread::_Impl_base>)
@ 00000000004033ac std::thread::thread<main::{lambda()#2}>(main::{lambda()#2}&&)
@ 000000000040308f main
@ 00007f374af330f6 __libc_start_main
@ 0000000000402ad8 [unknown]

Thread 3198692 created by Thread 3198375 (lockinversion) here:
@ 00007f374b020431 __clone
@ 00007f374bd5eef5 pthread_create
@ 00007f374b5e1440 std::thread::_M_start_thread(std::shared_ptr<std::thread::_Impl_base>)
@ 00000000004034b2 std::thread::thread<main::{lambda()#3}>(main::{lambda()#3}&&)
@ 00000000004030b9 main
@ 00007f374af330f6 __libc_start_main
@ 0000000000402ad8 [unknown]

Thread 3198693 created by Thread 3198375 (lockinversion) here:
@ 00007f374b020431 __clone
@ 00007f374bd5eef5 pthread_create
@ 00007f374b5e1440 std::thread::_M_start_thread(std::shared_ptr<std::thread::_Impl_base>)
@ 00000000004035b8 std::thread::thread<main::{lambda()#4}>(main::{lambda()#4}&&)
@ 00000000004030e6 main
@ 00007f374af330f6 __libc_start_main
@ 0000000000402ad8 [unknown]
```

In order to do that, I need to subshell out to objdump. This isn't very efficient, but this should only be run at the end of the program if a deadlock is found, so it's not so bad. Here's how I did it:

```python
def symbolize_with_objdump(binary, addr):
    '''
    Searches the symbol for the address using objdump. Returns the symbol if
    it is found, otherwise returns empty string.
    '''
    try:
        command = (
            'objdump -tT %s | grep %x | awk {\'print $NF\'} | c++filt' %
            (binary, addr)
        )
        output = subprocess.check_output(command, shell=True)
        return output.decode('utf-8').strip()
    except CalledProcessError:
        return ''
```

tools/deadlock_detector.py
kennyyu: I'll make the binary argument optional as a workaround for now!

kennyyu:
@4ast @goldshtn 

Addressed comments. Summary:

- Remove dependency on networkx. I did this by copying only the parts I needed
  from networkx, and adapting it to only use what I needed. These include:
  `DiGraph`, `strongly_connected_components`, `simple_cyles`

- Symbolize global and static mutexes. In order to do this, I subshell out to
  `subshell`. This isn't very efficient, but this only happens at the end of
  the program if a deadlock is found, so it's not too bad.

- `--verbose` mode to print graph statistics

- Make `--binary` flag optional. Not needed by default, However, this is needed
  on kernels without this recent kernel patch
  (https://lkml.org/lkml/2017/1/13/585, submitted 2 weeks ago): we can't attach
  a uprobe on a binary that has `:` in the path name. Instead, we can create a
  symlink without `:` in the path and pass that to the `--binary` argument
  instead.

brendangregg:
Very interesting! Some comments.

I'd like deadlock_detector.py to be chmod 755.

I can't get it to work with a PID alone.

```
# ./deadlock_detector.py `pgrep parprimes`
could not determine address of symbol pthread_mutex_unlock. Failed to attach to symbol: pthread_mutex_unlock
```

but I can do this:

```
# ./deadlock_detector.py --binary /lib/x86_64-linux-gnu/libpthread.so.0 `pgrep parprimes`
Tracing... Hit Ctrl-C to end.
```

Hm. Is it supposed to be like that?

I'd like -h to print these as examples, eg:

```
# deadlock_detector -h
[...]
examples:
    deadlock_detector 181         # analyze PID 181
    deadlock_detector /lib/x86_64-linux-gnu/libpthread.so.0 181
                                  # analyze PID 181 and locks from this binary

```

These examples should also be in the man page, which currently has:

```
              # deadlock_detector $(pidof binary) --binary /path/to/program
```

/path/to/program is how I'd write a synopsis. I think it'd be really helpful for people to see something like "/lib/x86_64-linux-gnu/libpthread.so.0" in the example.

kennyyu:
@brendangregg Thanks for the feedback!

I realize I've been testing mostly on statically-linked binaries, which do not require the `--binary` argument to be passed. This is because for statically-linked binaries, we're able to find the `pthread` symbols within it.

For dynamically-linked binaries, I think the `--binary` argument may be required in order to find the `pthread` symbols. I'll make this more explicit in the `-h` output and in the man page!

kennyyu:
@brendangregg Address round 2 of comments in #936

- Specify when `--binary` is needed (statically vs dynamically-linked
  binaries).

- Make `-h`, `_examples.txt`, and man page have concrete examples and
  be more user-friendly.

kennyyu:
@brendangregg @4ast Is there any more feedback before this can be merged? Thanks for taking a look!

4ast:
I think it's a solid start. All followups will be easier to review/fix as incremental patches on top. Hence merged. Thanks!

PR #937, opened 2017-02-03
docs: keep track of when prealloc of map elements were introduced

netoptimizer:
Kernel v4.6-rc1~91^2~108^2~6
 commit 6c9059817432 ("bpf: pre-allocate hash map elements")

Introduced default preallocation of mem elements to solve a deadlock
(when kprobe'ing the memory allocator itself).

This change is also a performance enhancement.

The commit also introduced a map_flags on BPF_MAP_CREATE, which can disable
this preallocation again BPF_F_NO_PREALLOC.

Signed-off-by: Jesper Dangaard Brouer <brouer@redhat.com>


brendangregg:
[buildbot, ok to test]

PR #941, opened 2017-02-03
remove symbol name storage optimization

jsedgwick:
Summary: not only is this implementation buggy (modifying the map invalidates pointers into it) but the cost of maintaining the map outstrips any performance benefits, according to perf. remove symnames_ and simplify implementation.

Test Plan: unit tests, used in my project


brendangregg:
[buildbot, ok to test]

markdrayton:
This optimisation is admittedly unpleasant but it saves a lot of RAM when using symbol maps that have many duplicate strings. There may be a small CPU cost but (IMO) saving up to hundreds of megabytes of RAM (see comments on #598) is a good tradeoff.

Also, as discussed offline with @jsedgwick, the code is okay -- emplace() can invalidate iterators but not references.

jsedgwick:
abandoning per offline conversation with @markdrayton, code is safe and my benchmarks didn't account for the common case of forking

PR #942, opened 2017-02-06
filetop: Fix TypeError by not mixing bytes and str.

r4f4:
When executing the filetop command tool, the following message was
generated:

  File "/usr/share/bcc/tools/filetop", line 190, in <module>
    name = name[:-3] + "..."
TypeError: can't concat bytes to str

Also, by decoding the bytes we print the strings without a leading "b'"
making the output more readable.

### DIFF
--- a/tools/filetop.py
+++ b/tools/filetop.py
@@ -185,14 +185,14 @@ while 1:
     line = 0
     for k, v in reversed(sorted(counts.items(),
                                 key=lambda counts: counts[1].rbytes)):
-        name = k.name
+        name = k.name.decode()
         if k.name_len > DNAME_INLINE_LEN:
             name = name[:-3] + "..."
 
         # print line
-        print("%-6d %-16s %-6d %-6d %-7d %-7d %1s %s" % (k.pid, k.comm,
-            v.reads, v.writes, v.rbytes / 1024, v.wbytes / 1024, k.type,
-            name))
+        print("%-6d %-16s %-6d %-6d %-7d %-7d %1s %s" % (k.pid,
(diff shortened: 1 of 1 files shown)


4ast:
cc @brendangregg

drzaeus77:
[buildbot, ok to test]

drzaeus77:
[buildbot, test this please]

(Please ignore failures, I am doing some test runs of a new set of buildbots)

brendangregg:
What's with the decode("utf-8")s? Other language character sets? Is there an easy example we can create to test?

If we need to do this on every process name and filename, that's going to be a big change, and litter the code with decode("utf-8")s. I'd hope there would be a way we could make it automatic for STDOUT, so we didn't need the explicit calls.

drzaeus77:
I haven't found a way to elide the `decode()` for python2/3 compatibility, but at least `decode()` should take no arguments (see #886 for example).

r4f4:
Yeah, if you just str() the bytes, they are printed as "b'<string>'". One option to get rid of some of those decodes is to save the values in the BPF class as strings instead of bytes. That way the decode has to be done only once for each attribute instead of every time you want to print them.

drzaeus77:
@r4f4 Did you decide a way to go on this? Either encode before filling the table or remove the (redundant) "utf-8" parameter to shorten the code.

r4f4:
@drzaeus77 Just update the PR. I went with the .decode with no params to avoid breaking the BPF class internals.

drzaeus77:
[buildbot, test this please]

Issue #940, opened 2017-02-03
With latest commits it's imposible to trace more than 200 syscalls.


vitalyvch:
The root of issue still is in an investigation, but if you have some ideas I would like to know them.

error from libbcc:

```
mmap: cannot allocate memory
```

@marcinslusarz, @GBuella, @ldorau, @plebioda, @sarahjelinek

brendangregg:
If you need to trace all the syscalls, then have you looked at the raw_syscalls:sys_enter, raw_syscalls:sys_exit tracepoints? 2 tracepoints to cover them all. I normally avoid them because I want to trace individual syscalls.

vitalyvch:
Yes, I tried them. Kernels, which are actual for us, reject attaching of eBPF to that tracepoints for this reason:

goldshtn:
I bisected and it looks like this was broken by b77915df95b971557dcb42606f493ffb5175a8ec -- this is the PR by @derek0883 that introduced support for concurrent probe users. Can you please take a look?

/cc @4ast @brendangregg

goldshtn:
To reproduce: `funccount 'SyS_*'` -- works before this commit, dies on this commit (probably due to out of memory but not sure).

vitalyvch:
From my point of view, the most suspicious changes were introduced into src/cc/libbpf.c, function "bpf_attach_kprobe()", lines 375-386.

derek0883:
Thank you guys. 
Just did a quick testing.
My commit [b77915d](https://github.com/iovisor/bcc/commit/b77915df95b971557dcb42606f493ffb5175a8ec)
will mkdir under /sys/kernel/debug/instances directory for each attach function. 
funccount 'SyS_*' will try to attach about 200 function, in current implementation, this will call instance_mkdir in kernel/trace/trace.c, 
```c
#define TRACE_BUF_SIZE_DEFAULT  1441792UL /* 16384 * 88 (sizeof(entry)) */
static unsigned long        trace_buf_size = TRACE_BUF_SIZE_DEFAULT;

static int instance_mkdir(const char *name)
{
if (allocate_trace_buffers(tr, trace_buf_size) < 0)
}
```
This will trigger out of memory, then will be killed.  you can get following error from dmesg or syslog.
```
Feb  5 13:54:13 ThinkPad-T430 kernel: [667147.087380] funccount.py invoked oom-killer: gfp_mask=0x240c0c0, order=2, oom_score_adj=0
[...]

PR #944, opened 2017-02-07
switch bcc to use single instance per bcc process, fixed issue #940

derek0883:
switch bcc to use single instance per bcc process, fixed issue #940
instance dir is bcc_pid, fixed issue #940


4ast:
[buildbot, ok to test]
[buildbot, add to whitelist]

brendangregg:
thanks! works!

PR #946, opened 2017-02-07
Update installation instructions for FC24 and FC25

goldshtn:
It turns out that for FC24 and FC25, there is a sufficiently recent
version of Clang in the official package sources, so we don't need
to fetch it from llvm.org using wget. Tested manually on a pair of
fresh FC24 and FC25 VMs, and confirmed that BCC builds and runs OK.


vitalyvch:
@ldorau

PR #949, opened 2017-02-08
tools/trace accepts abspath arguments to -I (and some doc tweaks)

ShelbyFrances:


4ast:
cc @goldshtn

drzaeus77:
[buildbot, ok to test]

tools/trace_example.txt
```diff
@@ -135,6 +135,16 @@ TIME     PID    COMM         FUNC             -
...
 for the Ruby method being invoked.
 
+You can also trace exported functions from shared libraries, or an imported
+function on the actual executable:
+
+# sudo ./trace.py 'r:/usr/lib64/libtinfo.so:curses_version "Version=%s", retval'
+# tput -V
+
+PID    TID    COMM         FUNC             -
+21720  21720  tput         curses_version   ret=ncurses 6.0.20160709
```
goldshtn: The trace format string says "Version=%s" but the output says "res=...", probably not intentional?

tools/trace_example.txt
```diff
@@ -164,6 +174,12 @@ PID    TID    COMM         FUNC             -
 777    785    automount    SyS_nanosleep    sleep for 500000000 ns
 ^C
 
+The tool automatically adds an `#include` for a few common types, but
+generally you should use the -I argument to have the requisite headers
+files included by the code generator, for compilation to succeed. Use
+either the filename relative to /usr/include, or an absolute path to
+the header file.
```
goldshtn: I don't know if this belongs in the examples file. I think it's better suited for the man page.

ShelbyFrances: I disagree. The one example here showing struct support "coincidentally" chose one of only 4 or so types that bcc has special support for. This kind of (partial) magic makes exposition easy, then creates frustration when the tool violates the user's expectations. 
So, I strongly feel some explanation like this belongs here, but I'll pare it down to just "Use -I to include the header file" as its shorter. Better to provide the general case that always works, than the convenient case that usually won't.

tools/trace_example.txt
ShelbyFrances: Fixed.

ShelbyFrances:
> could you also do argdist for uniformity?
Sure.

ShelbyFrances:
Pushed.

ShelbyFrances:
ping @goldshtn

tools/trace_example.txt
```diff
@@ -164,8 +174,12 @@ PID    TID    COMM         FUNC             -
 777    785    automount    SyS_nanosleep    sleep for 500000000 ns
 ^C
 
+Remember to use the -I argument include the appropriate header file. We didn't
```
goldshtn: "to include"?

goldshtn:
LGTM, just this tiny fix in the wording. Thanks!

ShelbyFrances:
Good catch, Fixed and pushed. Thank you!

@goldshtn

PR #951, opened 2017-02-08
trace: fix for python3

r4f4:
xrange does not exist in py3 and was replaced by range.

### DIFF
--- a/tools/trace.py
+++ b/tools/trace.py
@@ -483,10 +483,10 @@ BPF_PERF_OUTPUT(%s);
 
         def _format_message(self, bpf, tgid, values):
                 # Replace each %K with kernel sym and %U with user sym in tgid
-                kernel_placeholders = [i for i in xrange(0, len(self.types))
-                                       if self.types[i] == 'K']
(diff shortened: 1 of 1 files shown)


drzaeus77:
[buildbot, ok to test]

tools/trace.py
```diff
@@ -20,6 +20,9 @@
 import traceback
 import sys
 
+if sys.version_info >= (3, 0):
+    xrange=range
```
drzaeus77: Why not just change xrange to range in the body of the file? Or, looking at those uses, they would be good candidates for enumerate().

r4f4: What about 
                kernel_placeholders = filter(lambda t: t == 'K', self.types)
                user_placeholders = filter(lambda t: t == 'U', self.types)

drzaeus77: Sure, something along those lines would be good, although the intended list is meant to hold indices, so it might need some massaging.

r4f4: Oh, of course, my bad. I'll update the PR.

PR #955, opened 2017-02-09
tplist: don't fail if no verbosity is supplied

r4f4:
Initialise the verbosity argument to 0, so that when no verbosity is
provided we don't fail with

$ tplist
kvmmmu:kvm_mmu_pagetable_walk
unorderable types: NoneType() > int()


4ast:
cc @goldshtn

PR #956, opened 2017-02-09
Fix ucalls

r4f4:
Fix some compatibility issues with python3.

### DIFF
--- a/tools/ucalls.py
+++ b/tools/ucalls.py
@@ -236,12 +236,12 @@ if args.syscalls:
 def get_data():
     # Will be empty when no language was specified for tracing
     if args.latency:
-        data = map(lambda (k, v): (k.clazz + "." + k.method,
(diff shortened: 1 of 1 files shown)


4ast:
cc @goldshtn

drzaeus77:
[buildbot, ok to test]
[buildbot, add to whitelist]

goldshtn:
Hold on - did you test on Python 2.x after making these changes? Or does it break Python 2.x? (which is not acceptable)

r4f4:
It seems this [1] is the way to make it work on both py2 and py3. I'll update the patch.

[1] https://www.python.org/dev/peps/pep-3113/

goldshtn:
There's also the already merged patch that needs this fix

r4f4:
Just sent a separate PR addressing them.

PR #959, opened 2017-02-10
Fix uobjnew and ustat on python3

r4f4:


tools/ustat.py
```diff
@@ -239,10 +239,10 @@ def _loop_iter(self):
             counts.update(probe.get_counts(self.bpf))
             targets.update(probe.targets)
         if self.args.sort:
-            counts = sorted(counts.items(), key=lambda (_, v):
+            counts = sorted(counts.items(), key=lambda _, v:
```
goldshtn: This is now broken on Python 2.x

tools/uobjnew.py
```diff
@@ -147,13 +147,13 @@
...
-        data = sorted(data.items(), key=lambda (k, v): v.num_allocs)
+        data = sorted(data.items(), key=lambda k, v: v.num_allocs)
         data = data[-args.top_count:]
     elif args.top_size:
-        data = sorted(data.items(), key=lambda (k, v): v.total_size)
+        data = sorted(data.items(), key=lambda k, v: v.total_size)
         data = data[-args.top_size:]
     else:
-        data = sorted(data.items(), key=lambda (k, v): v.total_size)
+        data = sorted(data.items(), key=lambda k, v: v.total_size)
```
goldshtn: This is now broken on Python 2.x

PR #960, opened 2017-02-11
Add MapTypesVisitor for arbitrary inspection of types used as keys/va…

jsedgwick:
…lues in BPF tables

I have a C++ project that among other things attempts to read BPF tables without having the type information for table keys and values at compile time.
More specifically, arbitrary BPF programs can be passed to my program and loaded, and my program needs to be able to read data that the BPF programs put in their tables.
In order to have an idea of how to convert the void pointer returned by bpf_lookup_elem() into usable data, my program needs access to the layout of the types constituting the tables' keys/values.
My approach is to have a visitor interface that can be called with the key/value AST types for each table during one of the compiler passes over the BPF program.
As an aside, the visitor I use traverses the AST and creates a lambda of the type folly::dynamic(const void*) that is callable with the pointer returned by bpf_lookup_elem() on the table. See https://github.com/facebook/folly/blob/master/folly/docs/Dynamic.md - effectively, I'm creating converters to JSON-codable dynamic objects similar to python dicts etc.
I don't particularly like the intrusiveness of threading the visitor through to the appropriate AST pass - as you can see, it's quite a few levels of indirection down from the API
However, I found this preferable to the two other approaches I could think of -
  Adding folly::dynamic as a dependency to libbcc and building in the converters directly
    This seems too heavyweight for a feature that is unlikely to have additional customers
  Instead of injecting the visitor, exposing necessary AST data out through the API for external processing
    This has the same issues as my ultimate approach but worse, as multiple clang components (ASTContext, QualTypes) would have to be saved and exposed at an API level

If anyone has a cleaner approach, I'm all ears

Also exposes table_desc.h in include/ (for MapTypesVisitor interface), BPFModule from BPF, and the vector of TableDescs from the BPFModule

I tested this with my project and by compiling/running bcc tests.

### DIFF
--- a/src/cc/BPF.h
+++ b/src/cc/BPF.h
@@ -100,6 +100,10 @@ public:
   StatusTuple close_perf_buffer(const std::string& name);
   void poll_perf_buffer(const std::string& name, int timeout = -1);
 
+  BPFModule* get_module() const {
+    return bpf_module_.get();
+  }
+
 private:
   StatusTuple load_func(const std::string& func_name, enum bpf_prog_type type,
                         int& fd);
--- a/src/cc/CMakeLists.txt
+++ b/src/cc/CMakeLists.txt
@@ -67,7 +67,8 @@ target_link_libraries(bcc-static b_frontend clang_frontend bcc-loader-static ${c
 
 install(TARGETS bcc-shared LIBRARY COMPONENT libbcc
   DESTINATION ${CMAKE_INSTALL_LIBDIR})
-install(FILES bpf_common.h bpf_module.h bcc_syms.h bcc_exception.h libbpf.h perf_reader.h BPF.h BPFTable.h shared_table.h COMPONENT libbcc
+install(FILES bpf_common.h bpf_module.h bcc_syms.h bcc_exception.h libbpf.h
+perf_reader.h BPF.h BPFTable.h shared_table.h table_desc.h COMPONENT libbcc
   DESTINATION include/bcc)
 install(DIRECTORY compat/linux/ COMPONENT libbcc
   DESTINATION include/bcc/compat/linux
--- a/src/cc/bpf_module.cc
+++ b/src/cc/bpf_module.cc
@@ -326,6 +326,7 @@ unique_ptr<ExecutionEngine> BPFModule::finalize_rw(unique_ptr<Module> m) {
 // load an entire c file as a module
 int BPFModule::load_cfile(const string &file, bool in_memory, const char *cflags[], int ncflags) {
   clang_loader_ = make_unique<ClangLoader>(&*ctx_, flags_);
+  clang_loader_->set_map_types_visitor(map_types_visitor_);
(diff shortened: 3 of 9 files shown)


4ast:
@drzaeus77 could you please take a look.

drzaeus77:
Am looking, pondering better ways.

How is this different from the BMapDeclVisitor class which already constructs a json representation of the key/value type? Granted, that class isn't my favorite implementation and has been the source of many bugs, but the goal is the same, no?

jsedgwick:
pIt's more robust to allow full access to the AST when creating a byte-for-byte converter. The json encoding is somewhat lossy. I'd have to check with legal here to see if I could share my AST visitor class to demonstrate. For instance, I handle more builtin types (all int fact) correctly, bitfields more robustly, and have a specialization for char*. Plus I don't want to be dependent on the JSON format, which is not documented and could change under me (you have to read the code very carefully or run through all kinds of examples to see exactly what happens. Full AST access is safer.

drzaeus77:
I agree that the json approach is not ideal, but I guess my question is more about the placement of the API. Does the semantics of the json encoding make sense? As in, if you have a handle to a table, wouldn't you want a way to export the type of the key/leaf of that table without having to be an integral part of the compilation step. To me the biggest value add would be having a complete and accurate definition of the key/leaf types, which is independent from the clang step. I would be fine removing the json encoding in favor of something more robust, but I'm not aware of a well supported alternative that does:
 * Handling of all C types
 * Importable from Python or other bindings
 * Lightweight
 * Optionally something that can be stored with the map itself
Something like dwarf or CTF (C Type Format, used in dtrace) have been mentioned before, but no progress has been made.

jsedgwick:
I don't know what would reliably provide ALL info besides the clang QualType instances, but those require an ASTContext to be useful, which can't be extracted from a CompilerInstance. In other words, you'd have to keep the CompilerInstance around. You can share it, but that might still be heavyweight.

I'm just not sure that it's be possible to standardize a format that is separate from the AST - that's why I made an accessor for building my own.

jsedgwick:
Can anyone take another look at this. I'd like to get it (or some version of it) merged.

4ast:
ping @drzaeus77

drzaeus77:
I haven't forgotten about this. I'll send an update or an RFC-PR today with some framework for what I'd like to change.

drzaeus77:
The code is still a WIP, but I've been wanting to refactor how the shared table works, and this includes how the key/leaf types are visited. I have a WIP branch for those changes, which I've started to port over this change to as well. The next step for instance would be to re-implement BMapDeclVisitor as a visitor class of this type. See https://github.com/iovisor/bcc/tree/table-storage-wip for details.

drzaeus77:
What should we do with this PR? Is it possible to re-implement this easily against the updated TableStorage::AddMapTypesVisitor?

4ast:

PR #964, opened 2017-02-13
u* tools: Gracefully handle missing probes

### LINKED ISSUE #866: u* tools: gracefully handle missing USDT probes
It is more than likely that some people will try to use the u* tools on processes that don't have the required USDT probes (e.g. Oracle JDK, Node.js built without `--enable-dtrace` and so on). Make sure this condition is appropriately reported, in a way that isn't misleading. 

_Reported by @ceeaspb_

goldshtn:
When the target process is missing the required USDT probes, it can
be a simple mistake (e.g. attaching a script as Java to a Python
process), or a runtime that is not instrumented with the required
probes. Attempt to gracefully handle the error and print a helpful
message instructing the user why the error might have occurred.

Resolves #866.

```
$ uthreads -l java $(pidof python)
Error attaching USDT probes: the specified pid might not contain
the given language's runtime, or the runtime was not built with the
required USDT probes. Look for a configure flag similar to
--with-dtrace or --enable-dtrace. To check which probes are present
in the process, use the tplist tool.
```

### DIFF
--- a/tools/ucalls.py
+++ b/tools/ucalls.py
@@ -219,9 +219,9 @@ int syscall_return(struct pt_regs *ctx) {
 
 if args.language:
     usdt = USDT(pid=args.pid)
-    usdt.enable_probe(entry_probe, "trace_entry")
+    usdt.enable_probe_or_bail(entry_probe, "trace_entry")
     if args.latency:
-        usdt.enable_probe(return_probe, "trace_return")
+        usdt.enable_probe_or_bail(return_probe, "trace_return")
 else:
     usdt = None
 
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -1065,4 +1065,4 @@ class BPF(object):
         self.cleanup()
 
 
-from .usdt import USDT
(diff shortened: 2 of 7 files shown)


tools/ucalls.py
```diff
@@ -14,8 +14,20 @@
...
 import argparse
-from bcc import BPF, USDT
+from bcc import BPF, USDT, USDTException
 from time import sleep
+import sys
+
+def enable_probe_or_bail(usdt, probe, function):
+    try:
+        usdt.enable_probe(probe, function)
+    except USDTException:
```
4ast: should this be in common?

goldshtn: It probably should, but I wasn't sure if polluting the module with this check was worth it. Because it's essentially USDT::enable_probe with a custom behavior on failure. What do you think?

4ast: i think the copy paste avoidance trumps modularization concerns. i think of bcc/usdt.py as bcc specific usdt module.

goldshtn:
Fixed, rebased, and force-pushed.

PR #965, opened 2017-02-13
fixes #962 - fix lambda on both py2 and py3

### LINKED ISSUE #962: Some tools now broken on Python 2.x
After recent changes by @r4f4 to fix Python 3 compatibility issues, some tools are broken on Python 2.x, which is not good. For example, `ustat` now complains that the lambda on line 250 was given one argument when two were expected, after this change was applied:

```
- counts = sorted(counts.items(), key=lambda (k, _): k)
+ counts = sorted(counts.items(), key=lambda k, _: k)
```

@4ast @r4f4 @drzaeus77

r4f4:
See: https://www.python.org/dev/peps/pep-3113/

### DIFF
--- a/tools/ustat.py
+++ b/tools/ustat.py
@@ -239,10 +239,10 @@ class Tool(object):
             counts.update(probe.get_counts(self.bpf))
             targets.update(probe.targets)
         if self.args.sort:
-            counts = sorted(counts.items(), key=lambda _, v:
-                            -v.get(self.args.sort.upper(), 0))
+            counts = sorted(counts.items(), key=lambda kv:
+                            -kv[1].get(self.args.sort.upper(), 0))
         else:
-            counts = sorted(counts.items(), key=lambda k, _: k)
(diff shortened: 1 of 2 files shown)


tools/ustat.py
```diff
@@ -239,10 +240,10 @@ def _loop_iter(self):
             counts.update(probe.get_counts(self.bpf))
             targets.update(probe.targets)
         if self.args.sort:
-            counts = sorted(counts.items(), key=lambda _, v:
-                            -v.get(self.args.sort.upper(), 0))
+            counts = sorted(counts.items(), key=lambda kv:
+                            -kv[1].get(self.args.sort.upper(), 0))
         else:
-            counts = sorted(counts.items(), key=lambda k, _: k)
+            counts = sorted(counts.items(), key=itemgetter(1))
```
goldshtn: This is cute but maybe for simplicity stick to kv[1]? People reading and maintaining these tools aren't likely to be native pythonistas (myself included)...

PR #967, opened 2017-02-14
External debuginfo support and general symbols overhaul

### LINKED ISSUE #930: Symbol resolution code should support external debuginfo
A lot of binaries ship OOTB without debuginfo embedded, and require a debuginfo package (e.g. on FC, `dnf debuginfo-install pkg-name` will usually install the debuginfo into /usr/lib/debug). The current ELF scanning code that produces symbol names doesn't take debuginfo files into account. As a result, it's impossible to resolve symbols in any of our stack-tracing tools for any binary/SO that doesn't have debuginfo embedded, even if external debuginfo is installed.

It seems like FC and Ubuntu put debuginfo in /usr/lib/debug, and it seems like gdb on all platforms automatically knows to look there. So I'd say it would make sense for our symbol resolver to automatically try and find symbols there before giving up. Thoughts?

/cc @vmg @markdrayton @brendangregg

goldshtn:
This PR introduces support for external debuginfo (typically placed in /usr/lib/debug),
which is obtained by using either the `.note.gnu.build-id` or `.gnu_debuglink`
sections in the ELF. If debuginfo files are identified, they are loaded and parsed for
symbols, similarly to standard binaries. This resolves #930.

Additionally, the Python symbol API is simplified to just two main methods:
`BPF.sym` and `BPF.ksym` (the latter invokes the former). The caller can ask for the
module name to be included in the output, and/or for the instruction offset. Existing
tools are refactored to use this API. Any current callers of `BPF.sym` are not broken
by this change.

As part of looking at symbol resolution in some of the current tools, I removed
some overly verbose output from stacksnoop, offcputime, etc. --  specifically
I didn't feel that frame addresses are necessary most of the time, and they
contribute a lot of clutter.

The redundant and unnecessary usyms.py file with the `ProcessSymbols` class,
which was only used by memleak and one of the tracing examples, is also now removed,
and replaced with use of `SymbolCache`, offering almost identical functionality.

Lastly, a bug related to symbol resolution with PIE files is now fixed. (On my FC25, 
most system binaries are built with PIE, so this unbreaks a lot of stacks for me.)

Copying some persons of interest: @4ast @brendangregg @markdrayton


goldshtn:
Rebased, resolved conflicts, and force-pushed so this can be merged. Can someone take a look? @brendangregg @markdrayton @4ast @drzaeus77

markdrayton:
I'll review now. Sorry I didn't get to this sooner.

tools/old/memleak.py
```diff
@@ -11,36 +11,21 @@
 # Licensed under the Apache License, Version 2.0 (the "License")
 # Copyright (C) 2016 Sasha Goldshtein.
 
-from bcc import BPF, ProcessSymbols
+from bcc import BPF, SymbolCache
```
markdrayton: is `SymbolCache` used? I didn't see it in the added code.

tools/old/stackcount.py
```diff
@@ -145,10 +145,7 @@ def print_frame(addr):
     print("  ", end="")
     if verbose:
         print("%-16x " % addr, end="")
-    if offset:
-        print("%s" % b.ksymaddr(addr))
-    else:
-        print("%s" % b.ksym(addr))
+    print(b.ksym(addr, show_address=offset))
```
markdrayton: `=True`. It doesn't look like `offset` does anything special so seeing this instead of `True` may confuse.

tools/old/stacksnoop.py
```diff
@@ -119,10 +119,7 @@
     (task, pid, cpu, flags, ts, msg) = b.trace_fields()
     if msg != "":
         (reg, addr) = msg.split(" ")
-        if offset:
-            ip = b.ksymaddr(int(addr, 16))
-        else:
-            ip = b.ksym(int(addr, 16))
+        ip = b.ksym(int(addr, 16), show_address=offset)
```
markdrayton: `True` again

tools/stacksnoop.py
```diff
@@ -120,8 +120,8 @@ def print_event(cpu, data, size):
         print("%-18.9f %s" % (ts, function))
 
     for addr in stack_traces.walk(event.stack_id):
-        sym = b.ksymaddr(addr) if offset else b.ksym(addr)
-        print("\t%016x %s" % (addr, sym))
+        sym = b.ksym(addr, show_address=offset)
```
markdrayton: `True` again

src/cc/bcc_elf.c
```diff
@@ -196,20 +201,254 @@ static int listsymbols(Elf *e, bcc_elf_symcb callback, void *payload) {
...
+}
+
+static int find_debug_buildid(const char *buildid, char **debug_file ) {
+	char fullpath[PATH_MAX];
+
+	// Search for the file in the global debug directory with a sub-path:
+	// 	mm/nnnnnn...nnnn.debug
+	// Where mm are the first two characters of the buildid, and nnnn are the
+	// rest of the build id, followed by .debug.	
+	sprintf(fullpath, "/usr/lib/debug/.build-id/%c%c/%s.debug",
```
markdrayton: Is this path Linux distro independent?

tools/old/stacksnoop.py
markdrayton: Whoops, disregard this and the following ones. I misinterpreted `offset` as a string value.

tools/old/stackcount.py
markdrayton: Whoops, disregard this and the following similar comments. I misinterpreted `offset` as a string value.

docs/reference_guide.md
```diff
@@ -1122,11 +1105,11 @@ Examples in situ:
 [search /examples](https://github.com/iovisor/bcc/search?q=ksymname+path%3Aexamples+language%3Apython&type=Code),
 [search /tools](https://github.com/iovisor/bcc/search?q=ksymname+path%3Atools+language%3Apython&type=Code)
 
-### 4. sym()
+### 3. sym()
 
-Syntax: ```BPF.sym(addr, pid)```
+Syntax: ```BPF.sym(addr, pid, show_module=False, show_address=True)```
```
goldshtn: Thanks, fixed.

src/cc/bcc_elf.c
```diff
@@ -196,20 +201,254 @@ static int listsymbols(Elf *e, bcc_elf_symcb callback, void *payload) {
   return 0;
 }
 
-int bcc_elf_foreach_sym(const char *path, bcc_elf_symcb callback,
-                        void *payload) {
+static Elf_Data * get_section_elf_data(Elf *e, const char *section_name) {
+	Elf_Scn *section = NULL;
+	GElf_Shdr header;
+	char *name;
```
goldshtn: Absolutely. That's my punishment for making these edits on a box without my usual .vimrc :)

src/cc/bcc_elf.c
```diff
@@ -196,20 +201,254 @@ static int listsymbols(Elf *e, bcc_elf_symcb callback, void *payload) {
...
+/*
+  @name					the name of the debug information file
+  @binpath			the path to the original binary file for which we're resolving
+	@crc				  the expected checksum of the debuginfo file
+	@debug_file		receives the path to the file, or NULL if not found
+	
+	@returns			1 on success, 0 if not found
+*/
+static int find_debug_file(const char *name, const char *binpath,
+													 unsigned int crc, char **debug_file) {
```
goldshtn: Thanks, fixed.

src/cc/bcc_elf.c
goldshtn: You mean the .build-id part? It seems to be, according to the doc I quoted.

tools/old/memleak.py
goldshtn: Thanks, good catch.

goldshtn:
Thanks very much for the review!

Now for the general remarks:

> tests for the debug file handling would be great, though possibly tricky to arrange

I agree, but indeed tricky. We'd need debuginfo to be installed on the build bots. Some are Fedora, some are Ubuntu. I just don't see this happening easily. @drzaeus77 what do you think?

> would be nice to keep the formatting consistent in the C/C++ code

Absolutely, fixed.

> should show_address be called show_offset? AFAICT it controls whether or not the offset is shown, not the address

You're right, renamed to `show_offset`.

> did you see any slowdown from CRCing big debug files? I feel like in the past I've seen this take a long time with GDB/addr2line/etc but I can't easily repro the slowness now. It may well not be an issue here (especially as we only do it once) but it's something to be keep in the back of our minds.

I agree it can be a problem, but it's important to do -- this is why the checksum is there after all. We really don't want to be using the wrong debuginfo. And as you wrote, it only happens once when the module is loaded. I don't see a way to avoid it. Maybe parallelize the checksumming, but otherwise... not sure what we can do at all.

goldshtn:
Fixed comments from review, rebased, force-pushed. Thanks!

markdrayton:
> We'd need debuginfo to be installed on the build bots. Some are Fedora, some are Ubuntu. I just don't see this happening easily.

I think we could cover most of the new code by building a toy program that is compiled/rigged to have the right debug details.

E.g., we can set the GNU debug link with:

```
$ gcc -o foo foo.c
$ objcopy --only-keep-debug foo foo.debug
$ objcopy --add-gnu-debuglink=foo.debug foo
$ strip foo
```

We should be able to test the `find_debuglink`/`find_debug_file` path this way (we can also do `objcopy --add-gnu-debuglink=foo.debug foo.debug` to test the loop avoidance).

For the build ID-based path we can set a build ID using:

```
$ gcc -o foo -Xlinker --build-id=0xf00ff00f foo.c                                                                                                                                                   $ objdump --full-contents --section=.note.gnu.build-id foo
..
Contents of section .note.gnu.build-id:
 40023c 04000000 04000000 03000000 474e5500  ............GNU.
 40024c f00ff00f                             ....            
```

This would let us test `find_buildid`. The only bit that isn't covered then is `find_debug_buildid`, which wants to look in `/usr/lib`.

Obviously these would unit test the parts that look up the right debug build but wouldn't cover actually resolving an address from the debug file. That'd be doable by forking the toy program and using `BPF.sym(addr, pid)`, though. (We could objcopy/strip the test binary itself but that might make debugging or other things more annoying.)

Thanks for cleaning all this stuff up! It looks awesome.

drzaeus77:
@goldshtn 

> > tests for the debug file handling would be great, though possibly tricky to arrange


> I agree, but indeed tricky. We'd need debuginfo to be installed on the build bots. Some are Fedora, some are Ubuntu. I just don't see this happening easily. @drzaeus77 what do you think?

If you give me a list of packages for fc24, fc25, and ubuntu1604, I can add them to the metadata startup script to be installed the next time a build runs.

goldshtn:
All right, so I added a dummy.c and a test that builds it with debuglink and with build-id, and makes sure the symbol resolution can locate it. Also rebased and force-pushed. If this builds successfully on the server, I think it's ready for merging. @markdrayton

goldshtn:
It would be really great to get a final round of reviews on this and get this merged - it's a pretty big patchset and rebasing will not get any easier :) @markdrayton @drzaeus77 @4ast @brendangregg

src/cc/bcc_elf.c
```diff
@@ -196,20 +201,254 @@ static int listsymbols(Elf *e, bcc_elf_symcb callback, void *payload) {
...
 
   if (openelf(path, &e, &fd) < 0)
     return -1;
 
+  // If there is a separate debuginfo file, try to locate and read it, first
+  // using the build-id section, then using the debuglink section. These are
+  // also the rules that GDB folows.
+  // See: https://sourceware.org/gdb/onlinedocs/gdb/Separate-Debug-Files.html
+  if (!is_debug_file && (
+      (
```
4ast: the formatting style looks odd. please fix it in follow up

tests/python/test_debuginfo.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python
+# Copyright (c) PLUMgrid, Inc.
```
4ast: there is no such company anymore. Should be your copyright instead?

src/cc/bcc_elf.c
```diff
@@ -196,20 +201,254 @@ static int listsymbols(Elf *e, bcc_elf_symcb callback, void *payload) {
...
+       find_buildid(e, buildid) &&
+       find_debug_buildid(buildid, &debug_file))
```
4ast: i'm not sure why logic is split this way.
can find_debug_buildid() call find_buildid() and keep the knowledge of buildid private?
why expose it to this level?

src/cc/bcc_elf.c
```diff
@@ -196,20 +201,254 @@ static int listsymbols(Elf *e, bcc_elf_symcb callback, void *payload) {
...
+      ||
+      (
+       find_debuglink(e, &debug_name, &crc) &&
+       find_debug_file(debug_name, path, crc, &debug_file))
```
4ast: would it be cleaner to do:
debug_file = find_debug_via_buildid(e);
if (!debug_file)
  debug_file = find_debug_file(e);

markdrayton:
LGTM. Thanks for the tests and the rest of the work here! It'll be very helpful.

goldshtn:
@4ast Thanks for the comments, refactored, rebased, and force-pushed again. I think this is ready! 👍

PR #972, opened 2017-02-15
llcstat: fix TypeError on python3

r4f4:
The bytes object has no __format__ method of its own, inheriting it from
object, so an exception is thrown in python3 when it's passed to a
formatted string since formatting instructions are type specific.

$: ./llcstat
Running for 10 seconds or hit Ctrl-C to end.
PID      NAME             CPU     REFERENCE         MISS    HIT%
  File "./llcstat", line 108, in <module>
    (float(hit) / float(v.value)) * 100.0))
TypeError: non-empty format string passed to object.__format__


goldshtn:
[buildbot, test this please]

PR #978, opened 2017-02-15
Add --env, --status, and --ppid-tree to execsnoop

BMDan:
These flags add a bit of extra context for each line printed.

`--env`: Crucial for understanding highly environment-dependent execs (like `ssh`)
`--status`: All sorts of useful information, like the uid/gid, masks, etc.
`--ppid-tree`: Without this, trying to track down an ephemeral exec launched within another ephemeral exec (e.g. a program launched via `sudo`) is almost impossible.  This goes beyond the built-in PPID column by showing a full tree up to `init` or equivalent.

### DIFF
--- a/tools/execsnoop.py
+++ b/tools/execsnoop.py
@@ -236,7 +236,7 @@ def print_event(cpu, data, size):
     event = ct.cast(data, ct.POINTER(Data)).contents
 
     skip = False
-    
+
     if event.type == EventType.EVENT_ARG:
         argv[event.pid].append(event.argv)
     elif event.type == EventType.EVENT_RET:
--- a/tools/execsnoop.py
+++ b/tools/execsnoop.py
@@ -41,6 +41,12 @@ parser.add_argument("-x", "--fails", action="store_true",
     help="include failed exec()s")
 parser.add_argument("-n", "--name",
     help="only print commands matching this name (regex), any arg")
+parser.add_argument('--env', action='store_true',
+    help='Include the environment in the output')
+parser.add_argument('--status', action='store_true',
+    help='Include the contents of /proc/<pid>/status in the output (spammy!)')
+parser.add_argument('--ppid-tree', action='store_true',
+    help='Print every PPID up to 1 (or a loop)')
 args = parser.parse_args()
 
 # define BPF program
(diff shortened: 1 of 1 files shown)


brendangregg:
I think these capabilities would be a good addition, however, there may be a much better way to do this. The new code reads things out of /proc, like environ. But the tool already had this:

```
int kprobe__sys_execve(struct pt_regs *ctx, struct filename *filename,
    const char __user *const __user *__argv,
    const char __user *const __user *__envp)
{
```

So we already have *__envp. If you want the environment, have the code then (and only then) include a struct for outputting it (it's going to need a large buffer, so I wouldn't include it by default).

The problem this addresses is that reading /proc from bcc is asynchronous to kernel execution. I did this a lot in my ftrace perf-tools collection, because I didn't have a choice, and it caused problems: I've had lots of complaints about missing details because the process was too short lived, and vanished before the details could be read from /proc.

It could get worse in the future too:

```
while 1:
    b.kprobe_poll()
```

I've been thinking recently about changing all of these to introduce a deliberate small delay, so that output gets buffered. It was important for cpuunclaimed.py, but is probably good practice for all tools, otherwise we wakeup as fast as there is events (which may or may not buffer, depending on the event rate and burstyness, so it's hard to test; you might think reading /proc works ok now, but it doesn't work for someone's workload that's bursty and introduces a bigger delay).

So I'd investigate getting all these details from the kernel directly, and only going to /proc as a last resort.

brendangregg:
Also, doesn't work well for me with the latest bcc (ignore first 6 lines of warnings):

```
# ./execsnoop.py 
In file included from /virtual/main.c:4:
/lib/modules/4.9.0-rc5-virtual/build/include/linux/fs.h:2693:9: warning: comparison of unsigned enum expression < 0 is always false [-Wtautological-compare]
        if (id < 0 || id >= READING_MAX_ID)
            ~~ ^ ~
1 warning generated.
man              19355  12848    0 /usr/bin/man ls
man              19362  19355   -2 /usr/local/sbin/preconv -e UTF-8
man              19362  19355   -2 /usr/local/bin/preconv -e UTF-8
man              19362  19355   -2 /usr/sbin/preconv -e UTF-8
man              19364  19355   -2 /usr/local/sbin/tbl
man              19364  19355   -2 /usr/local/bin/tbl
  File "/usr/lib/python2.7/dist-packages/bcc/table.py", line 522, in <lambda>
    fn = _RAW_CB_TYPE(lambda _, data, size: callback(cpu, data, size))
  File "./execsnoop.py", line 251, in print_event
    ppid = get_ppid(event.pid)
  File "./execsnoop.py", line 183, in get_ppid
    for line in status:
TypeError: 'NoneType' object is not iterable
man              19365  19355   -2 /usr/local/sbin/nroff -mandoc -rLL=163n -rLT=163n -Tutf8
man              19365  19355   -2 /usr/local/bin/nroff -mandoc -rLL=163n -rLT=163n -Tutf8
man              19365  19355   -2 /usr/sbin/nroff -mandoc -rLL=163n -rLT=163n -Tutf8
man              19367  19355   -2 /usr/local/sbin/pager
man              19367  19355   -2 /usr/local/bin/pager
man              19367  19355   -2 /usr/sbin/pager
pager            19367  19355    0 /usr/bin/pager
troff            19371  19370    0 
  File "/usr/lib/python2.7/dist-packages/bcc/table.py", line 522, in <lambda>
    fn = _RAW_CB_TYPE(lambda _, data, size: callback(cpu, data, size))
  File "./execsnoop.py", line 251, in print_event
    ppid = get_ppid(event.pid)
  File "./execsnoop.py", line 183, in get_ppid
    for line in status:
TypeError: 'NoneType' object is not iterable
  File "/usr/lib/python2.7/dist-packages/bcc/table.py", line 522, in <lambda>
    fn = _RAW_CB_TYPE(lambda _, data, size: callback(cpu, data, size))
  File "./execsnoop.py", line 251, in print_event
    ppid = get_ppid(event.pid)
  File "./execsnoop.py", line 183, in get_ppid
    for line in status:
```

brendangregg:
When I run this:

```
for i in `seq 1 100`; do date > /dev/null; done
```

With execsnoop --env, it sometimes prints out different environments (fully of Netflix-y internal things, so I can't paste an example here). About 80% show one environment, and 20% show another. I don't know how that could happen from that shell one-liner -- so I assume it's a race when reading /proc.

Oh, and I just caught this:

```
debian-sa1       19824  19823    0 /usr/lib/sysstat/debian-sa1 1 1
  Env: {}
sa1              19824  19823    0 /usr/lib/sysstat/sa1 1 1
  Env: {}
sadc             19824  19823    0 /usr/lib/sysstat/sadc -F -L -S DISK 1 1 /var/log/sysstat
  Env: {}
```

Chances those environments really are null?

brendangregg:
Oh, --ppid-tree is nice. I imagine one might argue that it's less susceptible to a race with short-lived processes, since the parent should be longer lived. That's provided the event ppid -- the most fleeting -- is passed out on data_t, which it currently isn't, but should be.

It's also crooked:

```
date             19968  12848    0 /bin/date
PID: 19968
      \- 12848
       \- 12785
         \- 12769
           \- 1387
             \- 1
date             19969  12848    0 /bin/date
```

(see whitespace)

brendangregg:
--env should also have -e as an alternate (convention from ps(1)). --status could be -s, and --ppid-tree could be -P (-p is conventionally a PID filter).

```
  --status              Include the contents of /proc/<pid>/status in the
                        output (spammy!)
```

->

```
  --status              Include /proc/<pid>/status in output (verbose).
```

In summary, major things from a quick look:

- I think --env is doable in-kernel.
- The --ppid-tree should at least try to pass the current PPID out from the kernel directly. It could pass out a fixed number of parents too, using an unrolled loop (eg, up to 8 parents). If that is somehow impossible (I don't think it is), and instead uses some /proc, the man page should clearly state ppid-tree is best-effort and may be wrong.
- Needs updates to the man page and examples file.
- --status looks nice, but I don't have a great idea about how to implement that reliably. A man page "best effort" warning? People will complain if it gets it wrong. I do wonder why I was getting two different environments earlier...

goldshtn:
@BMDan Would you like to address the comments above? Or should we close the PR for now and create an issue to pick up in the future?

goldshtn:
@BMDan Are you still interested in looking into this?

goldshtn:
I'm closing for now, and creating an issue to track this in the future.

PR #983, opened 2017-02-16
Add xrange Python 2 -> 3 compatibility

r4f4:
In case returning a list in Py2 is acceptable, I can update the patch to replace xrange by range instead.


goldshtn:
I think in both cases the expected argument to xrange was quite small, so using range directly is perfectly acceptable rather than doing the test.

PR #984, opened 2017-02-16
usdt: fix argument passing on python3

r4f4:
This fixes the following error:

$: ./tplist -v -v -l /usr/lib64/dri/i965_dri.so
argument 1: <class 'TypeError'>: wrong type

### DIFF
--- a/src/python/bcc/usdt.py
+++ b/src/python/bcc/usdt.py
@@ -120,7 +120,7 @@ class USDT(object):
                 raise USDTException("USDT failed to instrument PID %d" % pid)
         elif path:
             self.path = path
-            self.context = lib.bcc_usdt_new_frompath(path)
+            self.context = lib.bcc_usdt_new_frompath(path.encode('ascii'))
             if self.context == None:
                 raise USDTException("USDT failed to instrument path %s" % path)
         else:


src/python/bcc/usdt.py
```diff
@@ -120,7 +120,7 @@ def __init__(self, pid=None, path=None):
                 raise USDTException("USDT failed to instrument PID %d" % pid)
         elif path:
             self.path = path
-            self.context = lib.bcc_usdt_new_frompath(path)
+            self.context = lib.bcc_usdt_new_frompath(path.encode('ascii'))
```
goldshtn: Does this still work on Python 2 with this change?

r4f4: It does in my tests.

$ python2 tplist -v -v -l /usr/lib64/dri/i965_dri.so
/usr/lib64/dri/i965_dri.so libstdcxx:throw [sema 0x0]
  location #1 0x58c1bd
    argument #1 8 unsigned bytes @ di
    argument #2 8 unsigned bytes @ si
/usr/lib64/dri/i965_dri.so libstdcxx:rethrow [sema 0x0]
  location #1 0x58c278
    argument #1 8 unsigned bytes @ ax
    argument #2 8 unsigned bytes @ *(di + 0)
/usr/lib64/dri/i965_dri.so libstdcxx:catch [sema 0x0]
  location #1 0x58d0ff
    argument #1 8 unsigned bytes @ ax
    argument #2 8 unsigned bytes @ *(bx - 80)

goldshtn:
LGTM then.

PR #985, opened 2017-02-17
support bcc on arm64(aarch64) platform

sjtuhjh:
As for aarch64 platform, the corresponding linux kernel header files might contain assemble code and register names directly. However clang compiler does not recognize them and will fail to compile them. Therefore, in order to make clang compiler happy, some header files will be revised in order to avoid including them. But these header file changes will be rollbacked automatically when bcc command has finished.

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -22,7 +22,6 @@ import re
 import struct
 import errno
 import sys
-import shutil
 
 basestring = (unicode if sys.version_info[0] < 3 else str)
 
@@ -121,8 +120,6 @@ class BPF(object):
         "linux/netdevice.h": ["sk_buff", "net_device"]
     }
     
-    _revised_header_files = []
-    
     # BPF timestamps come from the monotonic clock. To be able to filter
     # and compare them from Python, we need to invoke clock_gettime.
     # Adapted from http://stackoverflow.com/a/1205762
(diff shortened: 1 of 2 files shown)


4ast:
Thanks for testing on arm! Have you tried passing -D__ASM_SYSREG_H in preprocessor? That should remove asm.

drzaeus77:
Thanks for the effort on this, and glad to know that bcc can work on arm64 arch, but this approach just seems like a bad idea. We must not edit files on disk.

sjtuhjh:
@4ast  Just adding __ASM_SYSREG_H is not enough because some macros in sysreg.h are still required.  Therefore, the next solution is to read both sysreg.h and virt.h into load buffer and add some codes to filter assemble and register codes. 
However the latest solution wouldn't change the linux header source codes.

sjtuhjh:
@drzaeus77  Yes, you are right. We should try to avoid editing files on disk. I just added one new commit which wouldn't change linux header source files again.

src/python/bcc/__init__.py
```diff
@@ -155,6 +160,63 @@ def generate_auto_includes(cls, program_words):
...
+            kernel_filelist=[ "/usr/src/kernels/"+kernel_version+"/arch/arm64/include/asm/sysreg.h",
+                              "/usr/src/kernels/"+kernel_version+"/arch/arm64/include/asm/virt.h" ]
+            for filename in kernel_filelist:
+                if os.path.isfile(filename):
+                    filehandle = open(filename)
+       
+                    disable_assemble_flag = True
+                    cur_endif_index = 0
+                    for line in filehandle :
+                        if re.match("#if.*__ASSEMBLY__", line):
```
4ast: i don't understand why this is needed. This macro is active only when kernel headers are used inside asm files. Why do we need to replace them with if 0?

sjtuhjh: Let's use the sysreg.h(https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/arch/arm64/include/asm/sysreg.h?id=refs/tags/v4.10) for an example.  Usually there are assembly codes beween "#if.* __ASSEMBLY__" and "#endif" (such as line 57~61 in sysreg.h). These assembly codes couldn't be compiled by clang compiler. That is the reason why it added "#if 0" to disable these assembly code.

sjtuhjh: Let's use the sysreg.h(https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/arch/arm64/include/asm/sysreg.h?id=refs/tags/v4.10) for an example.  Usually there are assembly codes beween "#if.* __ASSEMBLY__" and "#endif" (such as line 57~61 in sysreg.h). These assembly codes couldn't be compiled by clang compiler. That is the reason why it added "#if 0" to disable these assembly code.

4ast: In this example the line 58 is not seen by clang, it's the line 60 that clang complains about,
but I don't understand how this patch removes it. It seems to me that it replaces
#ifdef __ASSEMBLY_
with #if 0,
so the line 60 will still be seen by clang.
What am i missing?
Is the goal to remove all of lines 57-61 ? and not just asm part before #else ?
so for sysreg.h file it will remove the lines 248-327 too ?

sjtuhjh: It doesnot remove the "#ifdef __ASSEMBLY". On the contrary, it addes "#if 0" before the "#ifdef __ASSEMBLY". In addition, it also addes the "#endif" after the "#endif" which matches the "#ifdef __ASSEMBLY".

sjtuhjh: Yes, it will also disable codes lines 248~327.

zhiyisun:
I met this issue when I ran bcc on ARM64 platform. My understanding is raw asm code between line 267 - 280 caused this problem. 

This patch modifies the kernel code "temporarily" to make bcc compilation successfully. But even it rolls back the code after that, we still can't guarantee there is no other process using these header files, especially for server environment. This patch supposes that the kernel source code is located in directory /usr/src/kernels/. That probably can't fit all scenarios.

src/python/bcc/__init__.py
4ast: got it. thanks. the /usr/src hard coding needs to improved, but good enough for now.
so LGTM. @drzaeus77 thoughts?

sjtuhjh: I just create one new pull request 998 which is based on latest bcc code, and resolve the /usr/src/ hard coding. In other words, it will use linux kernel source directory which is indicated by /lib/modules/${kernel_version}/source by default. If this directory doesn't exist, it will try to use /usr/src/${kernel_version} directory.

sjtuhjh:
@zhiyisun  The new pull request 998 will use linux kernel source directory which is indicated by /lib/modules/${kernel_version}/source by default. If this directory doesn't exist, it will try to use /usr/src/${kernel_version} directory. How do you think so?

sjtuhjh:
This pull request will be close, because the new pull request 998 is used to track latest code changes.

zhiyisun:
@sjtuhjh Thanks. That makes it more compatible on various system. But for the idea of this patch, I still think it would be better if we can find a way to let bcc toolchain to ignore raw asm code. Modify kernel source code make us hard to maintain between different Kernel versions. We don't know whether there will be any new raw asm code involved in other header files of aarch64 in the future.

4ast:
clang::rewriter can remove all asm from given .c on the fly, but i'm not sure
it can do it for .h files. cc @drzaeus77

zhiyisun:
@4ast FYI, for below code, if I compile it for bpf by using "clang -O2 -target bpf -c test.c -o test.o", it will return "Inline asm not supported". If I compile it for default target with "clang -O2 -c test.c -o test.o", it will not return any error. It is true for both x86 and arm64.

```code
/*    test.c */
asm ("nop");

int main(void) {
        int i = 1;
        return
        i;
}
```

PR #986, opened 2017-02-17
Fix bytes<->str mixing in python3.

r4f4:
I tried to fix all the tools I could execute and check the output.

These are the tools missing from this patch: mysql_qslower, btr*, solisten, oomkill, ucalls, uflow, ugc, uthreads, xfs*, zfs*.

I'll open an issue for ucalls, uflow, ugc and uthreads soon, since my fix attempt breaks the internal library finding procedure and I can't figure out why.

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -759,11 +759,12 @@ class BPF(object):
     def get_user_functions_and_addresses(name, sym_re):
         addresses = []
         def sym_cb(sym_name, addr):
-            if re.match(sym_re, sym_name):
-                addresses.append((sym_name, addr))
+            dname = sym_name.decode()
+            if re.match(sym_re, dname):
+                addresses.append((dname, addr))
             return 0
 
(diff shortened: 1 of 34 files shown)


drzaeus77:
[buildbot, test this please]

goldshtn:
@r4f4 Can you please rebase and force-push so we can get this merged?

goldshtn:
@r4f4 Can you please take a look at the build failure and also rebase this so we can get this merged?

r4f4:
Sure, I'm looking into it.

goldshtn:
Thanks very for much for this, I've taken the liberty of fixing up the commit message slightly. If you could follow up on the tools you said were showing some problems, it would be great.

r4f4:
@goldshtn It seems the internal library finding is working again, so I'm sending a patch for the remaining tools.

r4f4:
Well, never mind. The whole thing was reverted.

goldshtn:
@r4f4 We haven't reverted this, actually. And looks like we're not going to. So -- could you look into patching the remaining tools, like you wrote above?

PR #988, opened 2017-02-19
Support base + index * scale addressing for USDT arguments

goldshtn:
It turns out that some software will have USDT probe arguments
referencing memory using the full `nnn@(%basereg + %idxreg * scale)`
syntax. This is represented as `nnn@(%basereg,%idxreg,scale)` in
the `NT_STAPSDT` note, encountered in building a recent version of
PostgreSQL on FC25.

This format is now recognized by the USDT parser, and the correct
BPF code is emitted to retrieve arguments that reference memory
using this full addressing syntax.

### DIFF
--- a/src/cc/bcc_usdt.h
+++ b/src/cc/bcc_usdt.h
@@ -39,11 +39,13 @@ struct bcc_usdt_location {
     uint64_t address;
 };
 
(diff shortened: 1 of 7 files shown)


goldshtn:
@4ast If you have a moment, I'd love to get this merged too -- it broke my workflow a bit with Postgres. Thanks :)

PR #990, opened 2017-02-20
offwaketime: make sure to catch Ctrl-C

r4f4:
Otherwise the tool never stops running until you SIGKILL it.

### DIFF
--- a/tools/offwaketime.py
+++ b/tools/offwaketime.py
@@ -254,11 +254,13 @@ if not folded:
     else:
         print("... Hit Ctrl-C to end.")
 
-# as cleanup can take many seconds, trap Ctrl-C:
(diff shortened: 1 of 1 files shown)


brendangregg:
I'm fine with the change, but I haven't had that problem where Ctrl-C didn't work.

Just because I'm looking at it.. I wonder if my comment is wrong, and that it's printing the extra newline for non-folded output. And offwaketime.py should be chmod 755.

vitalyvch:
Hmmm, maybe we should do it in signal handler?

goldshtn:
@r4f4 @brendangregg What do we need to move forward on this?

goldshtn:
I'm closing for now, please feel free to reopen if there's any progress to be made.

PR #993, opened 2017-02-20
tools/opensnoop.py: Add options to filter for file names or patterns

MegaMaddin:
This commit adds two new options to the opensnoop.py tool - one for filtering
explicitely for files and the other one for file patterns.

### DIFF
--- a/tools/opensnoop.py
+++ b/tools/opensnoop.py
@@ -4,7 +4,7 @@
 # opensnoop Trace open() syscalls.
 #           For Linux, uses BCC, eBPF. Embedded C.
 #
-# USAGE: opensnoop [-h] [-T] [-x] [-p PID] [-t TID] [-n NAME]
+# USAGE: opensnoop [-h] [-T] [-x] [-p PID] [-t TID] [-n NAME] [-f FILE] [-r REGEX]
 #
 # Copyright (c) 2015 Brendan Gregg.
 # Licensed under the Apache License, Version 2.0 (the "License")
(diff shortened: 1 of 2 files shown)


brendangregg:
Thanks, but my first feelings on this is no.

I know a lot of people who want this feature, because they want to run it 24x7 to monitor for certain file accesses. This implementation passes all open events to user-level where it is filtered. How does this compare to doing the filtering in kernel level? Note that we don't have a regex engine in-kernel to use.

If I were to add this feature implemented in this way, I'd have made the implementation clear in the man page, something like:

> OVERHEAD: This passes all open() events to user-level where they are then filtered, costing overhead. There is likely a better way to code this in BPF, doing in-kernel filtering, but the author of this feature did not do it.

That not only sets expectations, but also doesn't mislead readers about BPF's overheads.

MegaMaddin:
You're absolutely right. Implementing this in Kernel, is probably the more convenient solution, at least for filtering for explicit file names. Feel free to decline this, I'll open another PR later, which implements the file parsing in-kernel.

MegaMaddin:
Will come up with another solution for this.

brendangregg:
Thanks, although you'll probably run into challenges with BPF's string support. (This sounds like it should be an easy place to get started with BPF/bcc work, but it's one of the harder areas to start with. I say this in case this apparently easy things seems disheartening!)

brendangregg:
I think the last time I saw some kernel string comparisons was trace.py:

```
# trace 'p:c:open (STRCMP("test.txt", arg1)) "opening %s", arg1' -T
01:43:15 10938  cat          open             opening test.txt
01:43:20 10939  cat          open             opening test.txt
^C
```

MegaMaddin:
Hmm, I see, they are explicitly re-implementing strcmp.

MegaMaddin:
Ok, it seem that I hit a bug regarding the builtin loop_detection of BPF. However I try to pass the data to BPF, if the file path gets longer than 23 chars, the loop detection kicks in. I can also reproduce this with the trace.py utility:

`# ./trace.py 'p:c:open (STRCMP("/var/lib/NetworkManager", arg1)) "opening %s", arg1'`
`bpf: Invalid argument`
`back-edge from insn 22 to 23`

It seems I need to open a ticket.

PR #998, opened 2017-02-22
support bcc on ARM64 platform

sjtuhjh:
This new pull request is based on previous pull request 985 (https://github.com/iovisor/bcc/pull/985) review comments.  In addition, the 985 pull request will be closed.


4ast:
lgtm
cc @drzaeus77 can clang::rewriter do this instead?
I mean removing 'asm' automatically ?

sjtuhjh:
@4ast  could we merge this pull request or wait for clang enhancement?

drzaeus77:
Yes, the clang rewriter should be able to handle this. Alternatively, you could read the files needing rewriting into c++ strings and modify them in-memory similar to how the exported_files.cc buffers are loaded. Basically, if you add a remapped file through invocation.getPreprocessorOpts().addRemappedFile() it should override the built-in version, without having to change anything on disk.

4ast:
there is another alternative:
https://www.spinics.net/lists/arm-kernel/msg567602.html

goldshtn:
@sjtuhjh Would you like to move this forward, or should I close for now while we examine the alternatives?

4ast:

vielmetti:
I'm still interested in this issue. Has there been any progress since May?

yonghong-song:
arm64 is okay with bcc now. This pull request is not needed, I think. Could you double check?

PR #999, opened 2017-02-22
Improve matching of file-backed memory mappings

markdrayton:
Use the same rules as perf to determine if a mapping in /proc/pid/maps is
file-backed. This allows mappings in anonymous huge pages and so on to fall
back to resolving from /tmp/perf-pid.map, if appropriate.

ref: http://lxr.free-electrons.com/source/tools/perf/util/map.c#L28

### DIFF
--- a/src/cc/bcc_proc.c
+++ b/src/cc/bcc_proc.c
@@ -70,6 +70,16 @@ char *bcc_procutils_which(const char *binpath) {
   return 0;
 }
 
+int bcc_mapping_is_file_backed(const char *mapname) {
+  return mapname[0] &&
+    strncmp(mapname, "//anon", sizeof("//anon") - 1) &&
+    strncmp(mapname, "/dev/zero", sizeof("/dev/zero") - 1) &&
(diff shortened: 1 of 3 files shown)


markdrayton:
Updated commit message to make things clearer. This is good to go now.

4ast:
nice fix!

4ast:
[buildbot, ok to test]
[buildbot, add to whitelist]

markdrayton:
Just force-pushed one more time to make the code formatting consistent.

markdrayton:
> Do we need this exported or can this function simply be static?

It doesn't need to be exported for normal use but it does if we want to test it. Of course, the test is pretty basic, but at least it'd catch someone inadvertently deleting or otherwise breaking the checks I added. I'm okay with making it static and dropping the test if you prefer.

4ast:
extra tests are always good. even for simple things like this.

palmtenor:
Last 3-digits PR!

PR #1002, opened 2017-02-23
cc: Handle nested functions correctly when resolving symbols

goldshtn:
`ProcSyms::Module::find_addr` incorrectly resolves symbols when
functions are nested in each other. Specifically, this was discovered
with libpthread, where there are multiple symbols for `write`,
and `write_nocancel` is strictly nested inside `write`. Fix by
explicitly going backward until we reach a matching symbol -- see
details in `ProcSyms::Module::find_addr` comments.

/cc @markdrayton

### DIFF
--- a/src/cc/bcc_syms.cc
+++ b/src/cc/bcc_syms.cc
@@ -190,16 +190,31 @@ bool ProcSyms::Module::find_addr(uint64_t addr, struct bcc_symbol *sym) {
   sym->offset = offset;
 
   auto it = std::upper_bound(syms_.begin(), syms_.end(), Symbol(nullptr, offset, 0));
-  if (it != syms_.begin())
-    --it;
-  else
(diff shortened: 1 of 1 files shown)


goldshtn:
@4ast @markdrayton Would be nice to get this merged, it's unbreaking a lot of stacks for me.

markdrayton:
Ugh sorry I missed this. Looks like a nice fix! I look forward to having more sensible stacks :-)

palmtenor:
@goldshtn Sasha, we are working on something related to this. Do you have an example of when such nesting would happen? It would be very helpful for our debugging. Thank you very much in ahead!

goldshtn:
I don’t remember all the details now (traveling), but the PR description says I saw it happen with write and write_nocancel in libpthread.

PR #1005, opened 2017-02-23
Restrict rewrite of unary operators to dereference operator

### LINKED ISSUE #344: possible rewriter issue with f_op
I have a possible rewriter issue. Full script:

``` Python
#!/usr/bin/python
# @lint-avoid-python-3-compatibility-imports
#
# fsslower  Trace slow file system sync reads and writes.
#           For Linux, uses BCC, eBPF.
#
# This script uses kernel dynamic tracing of two common file system functions:
# do_sync_read() and do_sync_write(). This provides a view of just two file
# system request types. There are typically many others: asynchronous I/O,
# directory operations, file handle operations, etc, that this script does
# not instrument.
#
# By default, a minimum millisecond threshold of 10 is used.
#
# Copyright 2016 Netflix, Inc.
#
#06-Feb-2016   Brendan Gregg   Created this.

from bcc import BPF
import argparse
import signal

# arguments
examples = """examples:
    ./fsslower             # trace sync I/O slower than 10 ms (default)
    ./fsslower 50          # trace sync I/O slower than 50 ms
    ./fsslower -p 185      # trace PID 185 only
"""
    description="Trace slow file system sync reads and writes",
    epilog=examples)
parser.add_argument("-p", "--pid",
    help="trace this PID only")
parser.add_argument("min_ms", nargs="?", default='10',
    help="minimum I/O duration to trace, in ms (default 10)")
args = parser.parse_args()
min_ms = int(args.min_ms)
pid = args.pid
debug = 1

# signal handler
def signal_ignore(signal, frame):
    print()

# define BPF program
bpf_text="""
#include <linux/fs.h>

#define TRACE_READ  0
#define TRACE_WRITE 1

struct val_t {
    u32 sz;
    u64 ts;
};

BPF_HASH(entryinfo, pid_t, struct val_t);

// store timestamp and size on entry
static int trace_rw_entry(struct pt_regs *ctx, struct file *file,
    char __user *buf, size_t count)
{
    u32 pid;
    struct val_t val = {};

    pid = bpf_get_current_pid_tgid();
    if (FILTER)
        return 0;

    val.sz = count;
    val.ts = bpf_ktime_get_ns();
    entryinfo.update(&pid, &val);

    return 0;
}

int trace_read_entry(struct pt_regs *ctx, struct file *file,
    char __user *buf, size_t count)
{
    if (!file->f_op->read_iter)
        return 0;     // not sync; see kernel code
    return trace_rw_entry(ctx, file, buf, count);
}

int trace_write_entry(struct pt_regs *ctx, struct file *file,
    char __user *buf, size_t count)
{
    if (!file->f_op->write_iter)
        return 0;     // not sync; see kernel code
    return trace_rw_entry(ctx, file, buf, count);
}

// output
static int trace_rw_return(struct pt_regs *ctx, int type)
{
    struct val_t *valp;

    valp = entryinfo.lookup(&pid);
    if (valp == 0) {
        // missed tracing issue or filtered
        return 0;
    }
    u64 delta_us = (bpf_ktime_get_ns() - valp->ts) / 1000;
    entryinfo.delete(&pid);
    if (delta_us < MIN_US)
        return 0;

    bpf_trace_printk("R %d %d\\n", valp->sz, delta_us);

    return 0;
}

int trace_read_return(struct pt_regs *ctx)
{
    return trace_rw_return(ctx, TRACE_READ);
}

int trace_write_return(struct pt_regs *ctx)
{
    return trace_rw_return(ctx, TRACE_WRITE);
}

"""
bpf_text = bpf_text.replace('MIN_US', str(min_ms * 1000))
if args.pid:
    bpf_text = bpf_text.replace('FILTER', 'pid != %s' % pid)
else:
    bpf_text = bpf_text.replace('FILTER', '0')
if debug:
    print(bpf_text)

# initialize BPF
b = BPF(text=bpf_text)

# I'd rather trace these via new_sync_read/new_sync_write (which used to be
# do_sync_read/do_sync_write), but those became static. So trace these from
# the parent functions, at the cost of more overhead, instead.
b.attach_kprobe(event="__vfs_read", fn_name="trace_read_entry")
b.attach_kprobe(event="__vfs_write", fn_name="trace_write_entry")
b.attach_kretprobe(event="__vfs_read", fn_name="trace_read_return")
b.attach_kretprobe(event="__vfs_write", fn_name="trace_write_return")

# header
print("Tracing sync read/writes slower than %d ms" % min_ms)
print("%-8s %-14s %-6s %-7s %-6s %7s" % ("TIME(s)", "COMM", "PID", "TYPE",
    "BYTES", "LAT(ms)"))

start_ts = 0

# format output
while 1:
    (type_s, sz, delta_us_s) = msg.split(" ")
    if start_ts == 0:
        start_ts = ts

    ms = float(int(delta_us_s, 10)) / 1000

    print("%-8.3f %-14.14s %-6s %-7s %-6s %7.2f" % (
        ts - start_ts, task, pid, type_s, sz, ms))
```

This bit:

``` C
    if (!file->f_op->read_iter)
        return 0;     // not sync; see kernel code
```

Errors with:

``` Shell
<bcc-memory-buffer>:38:384: error: no member named 'iter' in 'file'
  ...typeof(const struct file_operations *) _val; memset(&_val, 0, sizeof(_val)); bpf_probe_read(&_val, sizeof(_val), (u64)>rea + offsetof(struct file, iter)); _val; }) + ...
```

Which is odd since it I think it should be looking up read_iter in f_op. I'm happy to do this in steps, although, if I try:

``` C
    const struct file_operations *fop = file->f_op;
    if (!fop->read_iter)
        return 0;
```

I get:

``` Shell
<bcc-memory-buffer>:39:115: error: expected expression
    if (({ typeof(int) _val; memset(&_val, 0, sizeof(_val)); bpf_probe_read(&_val, sizeof(_val), (u64)fop->read_i({ typeof(ssize_t (*const)(struct kiocb *, struct iov...
                                                                                                                  ^
<bcc-memory-buffer>:39:108: error: no member named 'read_i' in 'struct file_operations'; did you mean 'read'?
    if (({ typeof(int) _val; memset(&_val, 0, sizeof(_val)); bpf_probe_read(&_val, sizeof(_val), (u64)fop->read_i({ typeof(ssize_t (*const)(struct kiocb *, struct iov...
                                                                                                           ^~~~~~
                                                                                                           read
include/linux/fs.h:1603:12: note: 'read' declared here
        ssize_t (*read) (struct file *, char __user *, size_t, loff_t *);
```

pchaigno:
This pull request limits the rewriting of unary operators to the
dereference operator and fixes #344.

Since the whole expression, unary operator included, is replaced by a
call to bpf_probe_read, the dereference operator is currently the
only unary operator properly rewritten anyway. When rewriting an
increment expression (++val) for instance, the increment operator is
lost in translation:
```c
// struct file *val = ++file;
struct file *val = ({ typeof(struct file *) _val;
                      memset(&_val, 0, sizeof(_val));
                      bpf_probe_read(&_val, sizeof(_val), (u64)file);
                      _val; });
```

Restricting the rewriting to dereference operators fixes #344 because
bcc won't try anymore to translate the logical negation in:
`!file->f_op->read_iter`.

/cc @drzaeus77 @brendangregg

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -200,7 +200,7 @@ bool ProbeVisitor::VisitBinaryOperator(BinaryOperator *E) {
   return true;
 }
 bool ProbeVisitor::VisitUnaryOperator(UnaryOperator *E) {
-  if (E->getOpcode() == UO_AddrOf)
+  if (E->getOpcode() != UO_Deref)
     return true;
   if (memb_visited_.find(E) != memb_visited_.end())
     return true;
--- a/tests/python/test_clang.py
+++ b/tests/python/test_clang.py
@@ -408,6 +408,17 @@ int dns_test(struct __sk_buff *skb) {
         """
         b = BPF(text=text)
 
+    def test_unary_operator(self):
+        text = """
+#include <linux/fs.h>
+#include <uapi/linux/ptrace.h>
+int trace_read_entry(struct pt_regs *ctx, struct file *file) {
+    return !file->f_op->read_iter;
+}
+        """
+        b = BPF(text=text)
+        b.attach_kprobe(event="__vfs_read", fn_name="trace_read_entry")
+
 if __name__ == "__main__":
     main()
 


pchaigno:
I've added a test case in `test_clang.py` based on @brendangregg's code in #344.

4ast:
cc @drzaeus77

drzaeus77:
I know I played with it once, but can you refresh my memory on whether clang treats `(*p).` and `p->` the same, and whether that would be a problem here?

Also, please rebase since there is now a conflict in test_clang.py.

pchaigno:
Clang sees `p->` as a `MemberExpr` (treated by bcc rewriter in [`b_frontend_action.cc#L202-L218`](https://github.com/iovisor/bcc/blob/d05e66de32beb49edeef69e811a1156237b9b6d7/src/cc/frontends/clang/b_frontend_action.cc#L202-L218)) whereas it sees `*p` as a `UnaryOperator` of opcode `UO_Deref` (treated in [`b_frontend_action.cc#L219-L256`](https://github.com/iovisor/bcc/blob/d05e66de32beb49edeef69e811a1156237b9b6d7/src/cc/frontends/clang/b_frontend_action.cc#L219-L256)).

So the rewrite isn't equivalent between `(*p).` and `p->`:
```c
return !(*file).f_op;
return !file->f_op;
```
is rewritten into:
```c
return !(({ typeof(struct file) _val;
            memset(&_val, 0, sizeof(_val));
            bpf_probe_read(&_val, sizeof(_val), (u64)file);
            _val; })).f_op;
return !({ typeof(const struct file_operations *) _val;
            memset(&_val, 0, sizeof(_val));
            bpf_probe_read(&_val, sizeof(_val), (u64)file + offsetof(struct file, f_op));
            _val; });
```

The new test case (`!file->f_op->read_iter`) serves as a check on the `!` more than on the `->`s. With this pull request, bcc doesn't try to replace the negation anymore (while the `->`s were already correctly replaced).

drzaeus77:
Ok, please proceed with the rebase.

drzaeus77:
[buildbot, test this please]

PR #1006, opened 2017-02-23
Fix bpf_dins_pkt rewrite in BinaryOperator

### LINKED ISSUE #537: Missing ')'
Hello again, this time with a real bug, I think.

```
/virtual/main.c:66:83: error: expected ')'
        bpf_dins_pkt(skb, (u64)ip+12, 0, 32, ibpf_dext_pkt(skb, (u64)ip+16, 0, 32);
                                                                                  ^
/virtual/main.c:66:21: note: to match this '('
        bpf_dins_pkt(skb, (u64)ip+12, 0, 32, ibpf_dext_pkt(skb, (u64)ip+16, 0, 32);
```

Here is a sample program that is causing the error.

```
#include <uapi/linux/bpf.h>
#include <uapi/linux/if_ether.h>
#include <uapi/linux/if_packet.h>
#include <uapi/linux/ip.h>
#include <uapi/linux/in.h>
#include <uapi/linux/udp.h>
#include <bcc/proto.h>

#define ETH_LEN 14

struct dns_hdr_t
{
    uint16_t id;
    uint16_t flags;
    uint16_t qdcount;
    uint16_t ancount;
    uint16_t nscount;
    uint16_t arcount;
} BPF_PACKET_HEADER;


struct dns_query_flags_t
{
  uint16_t qtype;
  uint16_t qclass;
} BPF_PACKET_HEADER;

struct dns_char_t
{
    char c;
} BPF_PACKET_HEADER;

struct Key {
  unsigned char p[32];
};

struct Leaf {
  // Not really needed in this example
  unsigned char p[4];
};

BPF_TABLE("hash", struct Key, struct Leaf, cache, 128);

int dns_test(struct __sk_buff *skb)
{
  u8 *cursor = 0;
  struct Key key = {};
  struct ethernet_t *ethernet = cursor_advance(cursor, sizeof(*ethernet));
  if(ethernet->type == ETH_P_IP) {

    u16 hlen_bytes = ip->hlen << 2;
    if(ip->nextp == IPPROTO_UDP) {

      struct udp_t *udp = cursor_advance(cursor, sizeof(*udp));
      if(udp->dport == 53){
        u8 *sentinel = cursor + udp->length - sizeof(*udp) - 4;

        struct dns_hdr_t *dns_hdr = cursor_advance(cursor, sizeof(*dns_hdr));
        if((dns_hdr->flags >>15) != 0) {
          return -1;
        }
         // Error is happening here.
          ip->src = ip->dst;
          return 0;
      }
    }
  }

  return -1;
}
```

pchaigno:
This pull request fixes #537 by changing the way `BinaryOperator`
expressions are rewritten.

Binary operator expressions where the left hand-side expression is a
reference to the packet are replaced by a call to the `bpf_dins_pkt`
helper. When replacing text, the Clang Rewriter tries to maintain a
list of offsets between the original and the new position of tokens.

However, replacing the whole binary operator expression with the call
to `bpf_dins_pkt` confuses the Rewriter and it is unable to track the
new position of the right hand-side expression. Rewriting the binary
operator expression in two times without rewriting the right
hand-side expression itself solves the issue.

I believe the second issue reported in #537 by @mvbpolito is a
different bug. It's specific to the use of a constant in a rewritten
expression. The constant seems "invisible" to the Clang rewriter.
~~I haven't found a proper way to fix it yet.~~ See #1008.

/cc @drzaeus77 @valkum

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -521,7 +521,6 @@ bool BTypeVisitor::VisitBinaryOperator(BinaryOperator *E) {
   if (!E->isAssignmentOp())
     return true;
   Expr *LHS = E->getLHS()->IgnoreImplicit();
-  Expr *RHS = E->getRHS()->IgnoreImplicit();
   if (MemberExpr *Memb = dyn_cast<MemberExpr>(LHS)) {
     if (DeclRefExpr *Base = dyn_cast<DeclRefExpr>(Memb->getBase()->IgnoreImplicit())) {
       if (DeprecatedAttr *A = Base->getDecl()->getAttr<DeprecatedAttr>()) {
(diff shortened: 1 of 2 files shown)


pchaigno:
I've added a test case in `test_clang.py` based on @valkum's code in #537.

drzaeus77:
Thanks!

PR #1008, opened 2017-02-26
Support for macros in rewriter

pchaigno:
This pull requests adds support for macros in the rewriter, thus fixing the
bug reported in #537 by @mvbpolito. 

It extends the scope of #863's fix to support macros anywhere in the rewriter.
All SourceRange objects are replaced to use macro expanded locations. [This
Stack Overflow response](http://stackoverflow.com/questions/24062989/clang-fails-replacing-a-statement-if-it-contains-a-macro#answer-24223347) explains the issue with macros in more details.

/cc @shodoco @mvbpolito


drzaeus77:
[buildbot, test this please]

PR #1011, opened 2017-02-28
cc: Use debuginfo symbols in preference to executable symbols

antonblanchard:
We currently look for some forms of debuginfo file before falling
back to the executable. Rearrange our search to look for all debuginfo
files before finally trying the executable.

This fixes an issue found on ppc64le Ubuntu where glibc is stripped
and I could only create uprobes for dynamic linker symbols.

Signed-off-by: Anton Blanchard <anton@samba.org>

### DIFF
--- a/src/cc/bcc_elf.c
+++ b/src/cc/bcc_elf.c
@@ -360,9 +360,9 @@ static char *find_debug_via_debuglink(Elf *e, const char *binpath) {
   bindir = strdup(binpath);
   bindir = dirname(bindir);
 
-  // Search for the file in 'binpath'
-  sprintf(fullpath, "%s/%s", bindir, name);
-  if (access(fullpath, F_OK) != -1) {
+  // Search for the file in 'binpath', but ignore it if it matches the
+  // binary itself.
+  if (strcmp(fullpath, binpath) && access(fullpath, F_OK) != -1) {
     res = strdup(fullpath);
     goto DONE;
   }


goldshtn:
As far as I can tell, you have just reordered the locations where bcc would look for the debuginfo file. All of this happens before trying the executable anyway. What am I missing?

Also, I would prefer to not do this reordering, because what we have right now is also the order that gdb is supposedly using: https://sourceware.org/gdb/onlinedocs/gdb/Separate-Debug-Files.html

> So, for example, suppose you ask gdb to debug /usr/bin/ls, which has a debug link that specifies the file ls.debug, and a build ID whose value in hex is abcdef1234. If the list of the global debug directories includes /usr/lib/debug, then gdb will look for the following debug information files, in the indicated order:
>
> /usr/lib/debug/.build-id/ab/cdef1234.debug
> /usr/bin/ls.debug
> /usr/bin/.debug/ls.debug
> /usr/lib/debug/usr/bin/ls.debug

antonblanchard:
I thought the first lookup in that function was for the executable itself (that was what ends up happening on my Ubuntu box). Perhaps the issue is:

```
# objdump -s -j .gnu_debuglink /lib/powerpc64le-linux-gnu/libc-2.24.so

/lib/powerpc64le-linux-gnu/libc-2.24.so:     file format elf64-powerpcle

Contents of section .gnu_debuglink:
 0000 6c696263 2d322e32 342e736f 00000000  libc-2.24.so....
 0010 bcd73013                             ..0.            
```

Do other distros add the trailing ".debug" in the debuglink string?

goldshtn:
On FC, pretty much all the binaries I have in /lib64 have the .debug at the end. On Ubuntu, I see some that do and some that don't. But the question is, what is the actual name of that debuginfo file? Does it include .debug?

antonblanchard:
```
/lib/powerpc64le-linux-gnu/libc-2.24.so
/usr/lib/debug/lib/powerpc64le-linux-gnu/libc-2.24.so
```

and the first lookup in find_debug_via_debuglink() hits the executable. Adding a print:

```
  // Search for the file in 'binpath'
  sprintf(fullpath, "%s/%s", bindir, name);
  printf("%s\n", fullpath);
```

outputs:

```
/lib/powerpc64le-linux-gnu/libc-2.24.so
```

goldshtn:
All right, so I suggest the fix should be checking if the lookup path is identical to the binary path, and skipping that lookup path if it is (but still doing the lookups in the same order as in the current code). We will end up checking the executable anyway if the normal debuginfo lookup doesn't get us anywhere. How's that sound?

rnav:
@goldshtn, that should work.

However, `perf` (*) prefers the distro debuginfo locations as well -- looking up /usr/lib/debug/ path with (fedora) or without (ubuntu) the .debug suffix for the filename. So, the current patch should generally work fine across architectures and may be simpler/faster.

(*) https://git.kernel.org/cgit/linux/kernel/git/powerpc/linux.git/commit/?id=a15ad2f5360c8

goldshtn:
Yeah, but OTOH why would we try to guess where to look and what the file should be named, if we have the debuglink and build-id sections in the binary? I feel that we should stick to gdb's search order. @4ast @markdrayton @brendangregg What do you think?

markdrayton:
@rnav thanks for raising this!

I'd go for @goldshtn's proposed fix — if it works, then great, and if not we can work out what we need to do to fix it. I'd definitely prefer to try to make things work the way they're supposed to before adding code to try to catch all possible eventualities.

antonblanchard:
@goldshtn Is this what you were thinking of?

goldshtn:
@antonblanchard Yes, that looks good. Although there might be symlinks involved, so I wonder if it would be even better to get the absolute path and then compare these?

goldshtn:
[buildbot, ok to test]

goldshtn:
[buildbot, test this please]

goldshtn:
[buildbot, test this please]

drzaeus77:
[buildbot, ok to test]

goldshtn:
@antonblanchard could you take a look at the test failures please?

goldshtn:
@antonblanchard Ping.

goldshtn:
I am closing this PR for now, because it seems to have been abandoned and there is a bug in the proposed logic. I will open my own PR with the same fix shortly.

PR #1012, opened 2017-02-28
cc: Fix SEGV when there is no build-id section

goldshtn:


4ast:
[buildbot, test this please]

PR #1015, opened 2017-03-01
cc: Retry symbol resolution using perfmap

goldshtn:
When a symbol lies within a module, and that module doesn't have
debuginfo (or doesn't even have an ELF header), the symbol will
always be resolved as [unknown]. However, the /tmp/perf-PID.map
(perf map) for that process might actually have an entry for that
symbol, if it was dynamically generated by some external tool.
This commit changes the resolution process such that if the desired
address lies in a module but that module doesn't have debuginfo,
we keep trying to resolve it in subsequent modules (including the
perf map). If we resolve it successfully using the perf map, the
reported symbol information will have the original module's name,
so we don't lose fidelity.

The motivation for this change is the way symbols work with .NET
Core on Linux. The runtime binaries are compiled ahead-of-time to
native code, but do not have debuginfo. There is an external tool,
which generates a file similar to a perf map (albeit with relative
addresses) for these binaries. This file can then be merged into
the main perf map for the process and used for symbol resolution,
but only if we keep trying to use the perf map when the symbol is
in a previously-seen module.

### DIFF
--- a/src/cc/bcc_syms.cc
+++ b/src/cc/bcc_syms.cc
@@ -114,6 +114,7 @@ bool ProcSyms::resolve_addr(uint64_t addr, struct bcc_symbol *sym) {
   sym->demangle_name = nullptr;
   sym->offset = 0x0;
 
+  const char *original_module = nullptr;
   for (Module &mod : modules_) {
     if (addr >= mod.start_ && addr < mod.end_) {
       bool res = mod.find_addr(addr, sym);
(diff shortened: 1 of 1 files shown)


markdrayton:
LGTM. I was initially concerned this would not handle anonymous mappings correctly, potentially replacing the perf map module name with the source mapping's path (e.g. the empty string or `//anon_hugepage`, etc). This isn't the case though — only file-backed mappings are added as modules and only they can replace the module name for code found in the perf map.

PR #1020, opened 2017-03-03
cc: Don't parse the same module multiple times for USDT probes

goldshtn:
If a module has more than one executable region, it is reported
multiple times by `bcc_procutils_each_module`. This is fine for
symbol resolution, but we don't need the duplicates for parsing
the ELF header looking for USDT probes: the first appearance of
that module is enough. This also prevents issues with the same
probe appearing multiple times with the same location, which
results in an invalid program when reading USDT arguments.

Fix by storing each visited module in the USDT::Context class,
and ignoring modules that were already visited.

### DIFF
--- a/src/cc/usdt.cc
+++ b/src/cc/usdt.cc
@@ -200,7 +200,13 @@ void Context::_each_probe(const char *binpath, const struct bcc_elf_usdt *probe,
 }
 
 int Context::_each_module(const char *modpath, uint64_t, uint64_t, void *p) {
-  bcc_elf_foreach_usdt(modpath, _each_probe, p);
+  Context *ctx = static_cast<Context *>(p);
+  // Modules may be reported multiple times if they contain more than one
+  // executable region. We are going to parse the ELF on disk anyway, so we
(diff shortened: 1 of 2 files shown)


goldshtn:
@drzaeus77 The fc25 build bot needs a nudge, it doesn't have 'git' again...

drzaeus77:
[buildbot, test this please]

goldshtn:
@drzaeus77 Same thing with the fc24 bot now. 😢

drzaeus77:
sigh...

sorry for the headache, this jenkins+gce plugin seems to be pretty buggy, it seems to be an intermittent race condition in the startup script for the fedora vms. Somehow they are not pulling the metadata from gce properly.

drzaeus77:
[buildbot, test this please]
And don't mess it up.

drzaeus77:
Of course, now that I'm watching it, the VMs initialized properly.

goldshtn:
Maybe you can add a magic command to restart the bots. :)

PR #1023, opened 2017-03-04
filetop: support specifying sort column via cmdline argument

dneiter:
I've found useful to be able to sort filetop results by write throughput instead of read throughput.

Let's add command line argument for this.

Testing:

I've tested the following filetop commands and they work as expected.
```
tools/filetop.py –help
tools/filetop.py
tools/filetop.py -s wbytes
```

### DIFF
--- a/man/man8/filetop.8
+++ b/man/man8/filetop.8
@@ -2,14 +2,15 @@
 .SH NAME
 filetop \- File reads and writes by filename and process. Top for files.
 .SH SYNOPSIS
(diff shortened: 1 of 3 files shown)


goldshtn:
[buildbot, test this please]

goldshtn:
LGTM and seems useful. @brendangregg what do you think?

brendangregg:
LGTM, tested, works.

[buildbot, test this please]

PR #1024, opened 2017-03-05
cmake: Explicitly mark static libraries as such

r4f4:
Some distros (e.g Fedora) override the default behaviour of building
static libraries to building dynamic ones instead. By explicitly
building the correct libraries as static, we make sure BCC properly
compiles everywhere.


4ast:
good catch. lgtm

Issue #1025, opened 2017-03-06
Build tests fail when run in parallel


fajs:
I followed the instructions from INSTALL.md for building and installing bcc from source on a Debian. (It's a Debian stretch, with a custom-compiled 4.10.1 kernel). Building the current master (40b881c) fails for me during the test phase. It's most often py_test_brb and py_test_brb2 failing, sometimes others, too. This changes from run to run. Example:

```
$ debuild -b -uc -us
[...]
he following tests FAILED:
         15 - py_test_brb (Failed)
         16 - py_test_brb2 (Failed)
Errors while running CTest
Makefile:86: recipe for target 'test' failed
make[1]: *** [test] Error 8
make[1]: Leaving directory '/tmp/bcc/obj-x86_64-linux-gnu'
dh_auto_test: make -j4 test ARGS+=-j4 returned exit code 2
debian/rules:12: recipe for target 'build' failed
make: *** [build] Error 2
dpkg-buildpackage: error: debian/rules build gave error exit status 2
````

or, on another run:
```
$ debuild -b -uc -s
[...]
The following tests FAILED:
         12 - py_test_trace4 (Failed)
         15 - py_test_brb (Failed)
         16 - py_test_brb2 (Failed)
Errors while running CTest
Makefile:86: recipe for target 'test' failed
make[1]: *** [test] Error 8
make[1]: Leaving directory '/tmp/bcc/obj-x86_64-linux-gnu'
dh_auto_test: make -j4 test ARGS+=-j4 returned exit code 2
debian/rules:12: recipe for target 'build' failed
make: *** [build] Error 2
dpkg-buildpackage: error: debian/rules build gave error exit status 2
```

It seems the test part can't be run in parallel properly. The problems disappear if I run ```debuild -b -uc -us -j1```.

Attached: full output of ```debuild -b -uc -us``` for [run1](https://github.com/iovisor/bcc/files/821206/run1.txt) and [run2](https://github.com/iovisor/bcc/files/821207/run2.txt)

drzaeus77:
This restriction seems okay to me for the time being.

fajs:
The above pull request gives a workaround for that problem. It only disables parallelization for the test step, so that building can still be done in parallel.

fajs:
Thanks for the quick merge. As far as I see it, this issue can be closed now.

PR #1027, opened 2017-03-06
debuild: Do not parallelize tests

fajs:
The tests in the test suits are not parallelizable and will fail if run
in parallel. Make the test step non-parallel to fix this issue.

This resolves issue #1025.


drzaeus77:
[buildbot, ok to test]

PR #1030, opened 2017-03-07
Added s390x support. Needs 4.10 Kernel

zvonkok:
Adding support for the s390x architecture (z Systems). Needs 4.10 Kernel for clean build, otherwise
VLAIS and out-of-bounds errors have to be eliminated. Verified most of the tools that are applicable.


4ast:
lgtm
[buildbot, ok to test]

PR #1040, opened 2017-03-10
RFC: rework of shared table into multiple classes

drzaeus77:
This patch series attempts to make somewhat more modular the way that tables are stored in memory, consumed by the clang rewriter, and exposed to c++ libraries that wish to control maps directly. Currently it comes in two parts, but I would like to break the diff into a few more parts if people find this hard to review.

At this point, I'm mostly looking for feedback on whether the approach seems viable. Some things I'd like to accomplish with this patch set:

1. Expose an interface to consume map types without intruding too much on the rewriter.
2. Create a a mechanism to support multiple table storage backends
 a. shared - a simple process-only mode that is currently the default and provides backwards compatibility
 b. bpffs - a persistent mode that allows multiple invocations of libbcc to share maps using bpf pin/get
 c. standalone - ?? - future work, a mode that doesn't involve using clang at all, so that tools can access maps without parsing the c files
3. Introduce a concept of paths/namespaces...this is something I have in mind that will enable multiple bpf programs to work in concert or to be isolated as needed. Maps will be stored with a local name and a global name (path). This would make a lot more sense with bpffs, but the API is exposed to the parser and the module class as well.

### DIFF
--- a/src/cc/BPF.h
+++ b/src/cc/BPF.h
@@ -26,6 +26,7 @@
 #include "bpf_module.h"
 #include "compat/linux/bpf.h"
 #include "libbpf.h"
+#include "table_storage.h"
 
 static const int DEFAULT_PERF_BUFFER_PAGE_CNT = 8;
 
@@ -38,12 +39,14 @@ struct open_probe_t {
 };
 
 class USDT;
+class TableStorage;
 
 class BPF {
 public:
   static const int BPF_MAX_STACK_DEPTH = 127;
 
-  explicit BPF(unsigned int flag = 0) : bpf_module_(new BPFModule(flag)) {}
+  explicit BPF(unsigned int flag = 0, TableStorage* ts = nullptr)
+      : bpf_module_(new BPFModule(flag, ts)), ts_(ts) {}
   StatusTuple init(const std::string& bpf_program,
                    std::vector<std::string> cflags = {},
                    std::vector<USDT> usdt = {});
@@ -90,6 +93,11 @@ public:
 
   template <class KeyType, class ValueType>
   BPFHashTable<KeyType, ValueType> get_hash_table(const std::string& name) {
+    if (ts_) {
+      TableStorage::iterator it;
+      if (ts_->Find(Path({name}), it))
+        return BPFHashTable<KeyType, ValueType>(it->second);
+    }
     return BPFHashTable<KeyType, ValueType>(bpf_module_.get(), name);
   }
 
@@ -152,6 +160,7 @@ private:
                                   uint64_t symbol_addr, bcc_symbol* output);
 
   std::unique_ptr<BPFModule> bpf_module_;
+  TableStorage* ts_;
 
   std::map<std::string, int> funcs_;
 
--- a/src/cc/BPF.h
+++ b/src/cc/BPF.h
@@ -46,7 +46,7 @@ public:
   static const int BPF_MAX_STACK_DEPTH = 127;
 
   explicit BPF(unsigned int flag = 0, TableStorage* ts = nullptr)
-      : bpf_module_(new BPFModule(flag, ts)), ts_(ts) {}
+      : bpf_module_(new BPFModule(flag, ts)) {}
   StatusTuple init(const std::string& bpf_program,
                    std::vector<std::string> cflags = {},
                    std::vector<USDT> usdt = {});
@@ -93,16 +93,17 @@ public:
 
   template <class KeyType, class ValueType>
   BPFHashTable<KeyType, ValueType> get_hash_table(const std::string& name) {
-    if (ts_) {
-      TableStorage::iterator it;
-      if (ts_->Find(Path({name}), it))
-        return BPFHashTable<KeyType, ValueType>(it->second);
-    }
-    return BPFHashTable<KeyType, ValueType>(bpf_module_.get(), name);
+    TableStorage::iterator it;
+    if (bpf_module_->table_storage().Find(Path({bpf_module_->id(), name}), it))
+      return BPFHashTable<KeyType, ValueType>(it->second);
+    return BPFHashTable<KeyType, ValueType>({});
   }
 
   BPFStackTable get_stack_table(const std::string& name) {
-    return BPFStackTable(bpf_module_.get(), name);
+    TableStorage::iterator it;
+    if (bpf_module_->table_storage().Find(Path({bpf_module_->id(), name}), it))
+      return BPFStackTable(it->second);
+    return BPFStackTable({});
   }
 
   StatusTuple open_perf_buffer(const std::string& name, perf_reader_raw_cb cb,
@@ -160,7 +161,6 @@ private:
                                   uint64_t symbol_addr, bcc_symbol* output);
 
   std::unique_ptr<BPFModule> bpf_module_;
-  TableStorage* ts_;
 
   std::map<std::string, int> funcs_;
 
--- a/src/cc/BPF.h
+++ b/src/cc/BPF.h
@@ -39,7 +39,6 @@ struct open_probe_t {
 };
 
 class USDT;
-class TableStorage;
 
 class BPF {
 public:
--- a/src/cc/BPFTable.h
+++ b/src/cc/BPFTable.h
@@ -28,6 +28,7 @@
 #include "bpf_module.h"
 #include "libbpf.h"
 #include "perf_reader.h"
+#include "table_desc.h"
 
 namespace ebpf {
 
@@ -44,6 +45,10 @@ protected:
     fd_ = bpf_module->table_fd(id_);
     capacity_ = bpf_module->table_max_entries(id_);
   };
+  explicit BPFTableBase(const TableDesc& desc) {
+    fd_ = desc.fd;
+    capacity_ = desc.max_entries;
+  }
 
   bool lookup(KeyType* key, ValueType* value) {
     return bpf_lookup_elem(fd_, static_cast<void*>(key),
@@ -70,7 +75,8 @@ protected:
 
 template <class KeyType, class ValueType>
 class BPFHashTable : protected BPFTableBase<KeyType, ValueType> {
-public:
+ public:
+  explicit BPFHashTable(const TableDesc& desc) : BPFTableBase<KeyType, ValueType>(desc) {}
   BPFHashTable(BPFModule* bpf_module, const std::string& name)
       : BPFTableBase<KeyType, ValueType>(bpf_module, name) {}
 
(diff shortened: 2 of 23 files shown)


CMakeLists.txt
```diff
@@ -63,7 +63,7 @@ if (GCC_VERSION VERSION_LESS 6.0)
   endforeach()
 endif()
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall ${CXX_ISYSTEM_DIRS}")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14 -Wall ${CXX_ISYSTEM_DIRS}")
```
4ast: which minimum version of gcc and clang is required ?

examples/cpp/CPUDistribution.cc
```diff
@@ -14,17 +14,27 @@
...
 #include <iostream>
+#include <memory>
 #include <string>
 
 #include "BPF.h"
+#include "shared_table.h"
+#include "table_desc.h"
+#include "common.h"
+
+const std::string BPF_TABLES = R"(
```
4ast: i don't understand the purpose of this split. are you suggesting that all .py scripts would need to split table def out of c code?

src/cc/shared_table.cc
```diff
@@ -14,52 +14,97 @@
...
+  return make_unique<iterator>(tables_.erase(i));
+}
+
+// All maps for this process are kept in global static storage.
+std::map<string, TableDesc> SharedTableStorage::tables_;
+
+unique_ptr<TableStorage> createSharedTableStorage() {
+  auto t = make_unique<TableStorage>();
+  t->Init(make_unique<SharedTableStorage>());
+  t->AddMapTypesVisitor(createJsonMapTypesVisitor());
```
4ast: this approach indeed looks cleaner than what we have today.

src/cc/table_desc.h
```diff
@@ -17,18 +17,75 @@
...
+  TableDesc(TableDesc&& that) = default;
+  TableDesc & operator=(TableDesc&& that) = default;
+  TableDesc dup() const {
+    return TableDesc(*this);
+  }
+
   std::string name;
-  int fd;
-  int type;
+  FileDesc fd;
```
4ast: is this really all that was needed to start using this new class?
many more places to go?

CMakeLists.txt
drzaeus77: I think 4.9, maybe 5.0. It has been 3 years. I would split this change out into a different PR, so we can ignore it for now.

examples/cpp/CPUDistribution.cc
drzaeus77: Umm, that was me playing around making sure this worked with the C++ API. I intend to remove it.

src/cc/table_desc.h
drzaeus77: Nope, it works just like this. The FileDesc class can probably be used in lots of places, and probably should go in a separate standalone header.

src/cc/BPF.h
```diff
@@ -38,12 +39,15 @@ struct open_probe_t {
...
 class USDT;
+class TableStorage;
 
 class BPF {
 public:
   static const int BPF_MAX_STACK_DEPTH = 127;
 
-  explicit BPF(unsigned int flag = 0) : bpf_module_(new BPFModule(flag)) {}
+  explicit BPF(unsigned int flag = 0, TableStorage *ts = nullptr)
+      : bpf_module_(new BPFModule(flag, ts)), ts_(ts)
```
palmtenor: Why do we even need to keep `ts_` in the C++ class? Can we always go to something like `bpf_module->get_table_storage()` if we want to access the table storage? Especially given the `BPFModule` holds `unique_ptr` of the table storage class......

src/cc/BPF.h
```diff
@@ -90,6 +94,11 @@ class BPF {
 
   template <class KeyType, class ValueType>
   BPFHashTable<KeyType, ValueType> get_hash_table(const std::string& name) {
+    if (ts_) {
+      TableStorage::iterator it;
+      if (ts_->Find(Path({name}), it))
+        return BPFHashTable<KeyType, ValueType>(it->second);
+    }
     return BPFHashTable<KeyType, ValueType>(bpf_module_.get(), name);
```
palmtenor: Then we can always refer to the table storage class pointer from `BPFModule` and don't need this fallback logic.

src/cc/BPFTable.h
```diff
@@ -44,6 +45,10 @@ class BPFTableBase {
     fd_ = bpf_module->table_fd(id_);
     capacity_ = bpf_module->table_max_entries(id_);
   };
+  explicit BPFTableBase(const TableDesc& desc) {
```
palmtenor: And hence we can keep only this constructor for `BPFTableBase`

src/cc/bpf_module.cc
```diff
@@ -349,10 +348,13 @@ int BPFModule::annotate() {
...
 
   // separate module to hold the reader functions
-  auto m = make_unique<Module>("sscanf", *ctx_);
+  auto m = std::make_unique<Module>("sscanf", *ctx_);
 
   size_t id = 0;
-  for (auto &table : *tables_) {
+  Path path({std::to_string((uintptr_t)this)});
+  for (auto it = ts_->lower_bound(path), up = ts_->upper_bound(path); it != up; ++it) {
+    TableDesc &table = it->second;
```
palmtenor: Is it possible for us not to keep a copy of `TableDesc` pointers in `BPFModule` class? Users need this table related methods (like get name, ID, size etc.) can get a reference to the table storage class from `BPFModule` (or from their own reference if they passed it to `BPFModule` like the BPFFS case), and then call relevant method from the source-of-truth directly.

We can even keep the methods in `BPFModule` for a while for backward-compatible, but just proxy through `ts_->corresponding_method()`?

src/cc/common.h
```diff
@@ -23,11 +23,7 @@
 
 namespace ebpf {
 
-template <class T, class... Args>
-typename std::enable_if<!std::is_array<T>::value, std::unique_ptr<T>>::type
-make_unique(Args &&... args) {
-  return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
-}
+using std::make_unique;
```
palmtenor: We are on C++14 already? ;))

src/cc/BPF.h
drzaeus77: Will do.

src/cc/bpf_module.cc
drzaeus77: For now we need it to provide a stable ordering for the users of the index-based API. If we phase that out, then sure.

My motivation for this changeset is in fact to allow the TableStorage class to be used externally to the JIT-time objects (BPFModule). Having a long-lived BPFFS that gets passed in and owned separately is one of the next steps.

src/cc/BPF.h
drzaeus77: Well, neither the new or existing logics handle cases where `name` is an invalid key, so this function needs some further changes.

drzaeus77:
Updated this to stick with C++11 features.

/cc @palmtenor @4ast

src/cc/BPF.h
drzaeus77: I've collapsed it to one constructor, but the error case is still unhandled.

src/cc/BPF.h
```diff
@@ -38,12 +39,14 @@ struct open_probe_t {
 };
 
 class USDT;
+class TableStorage;
```
palmtenor: We already include the header, no need for forward declare?

src/cc/CMakeLists.txt
```diff
@@ -67,7 +67,7 @@ target_link_libraries(bcc-static b_frontend clang_frontend bcc-loader-static ${c
 
 install(TARGETS bcc-shared LIBRARY COMPONENT libbcc
   DESTINATION ${CMAKE_INSTALL_LIBDIR})
-install(FILES bpf_common.h bpf_module.h bcc_syms.h bcc_exception.h libbpf.h perf_reader.h BPF.h BPFTable.h shared_table.h COMPONENT libbcc
+install(FILES bpf_common.h bpf_module.h bcc_syms.h bcc_exception.h libbpf.h perf_reader.h BPF.h BPFTable.h shared_table.h table_desc.h COMPONENT libbcc
```
palmtenor: Not installing `table_storage.h` given it's included in `BPF.h`?

src/cc/table_desc.h
```diff
@@ -16,18 +16,101 @@
...
+/// owned by this table are not implicitly copyable. One should call the dup()
+/// method if an explicit new handle is required. We define the move operators
+/// so that objects of this class can reside in stl containers.
+class TableDesc {
+ private:
+  TableDesc(const TableDesc &) = default;
+  TableDesc &operator=(const TableDesc &) = default;
+
+ public:
+  TableDesc()
```
palmtenor: `fd` is not initialized for the default empty constructor. What's the behavior if it's actually being used? Since we have fall-back like `BPFHashTable<>({})` which uses this......

src/cc/bpf_module.cc
```diff
@@ -101,30 +101,29 @@ class MyMemoryManager : public SectionMemoryManager {
...
   InitializeNativeTarget();
   InitializeNativeTargetAsmPrinter();
   LLVMInitializeBPFTarget();
   LLVMInitializeBPFTargetMC();
   LLVMInitializeBPFTargetInfo();
   LLVMInitializeBPFAsmPrinter();
   LLVMLinkInMCJIT(); /* call empty function to force linking of MCJIT */
+  if (!ts_) {
+    local_ts_ = createSharedTableStorage();
+    ts_ = &*local_ts_;
```
palmtenor: Would `local_ts_.get()` just work?

src/cc/table_desc.h
drzaeus77: It will be initialized with `-1`, by virtue of the FileDesc default constructor. In the current code, all the places where `ts.Insert` occur should check `fd >= 0` before inserting anyway, so for now we have the status quo (fails during compilation).

src/cc/bpf_module.cc
drzaeus77: It would, but this is my style preference (I've read too much LLVM code).

src/cc/CMakeLists.txt
drzaeus77: Good catch. In another project that depends on this, I only use bpf_module.h so didn't hit this error. Will update.

PR #1043, opened 2017-03-10
bpflist: Display processes with running BPF programs and maps

### LINKED ISSUE #1036: new tool proposal: bpflist
I'm proposing a new tool that would list enabled BPF programs.

This tool could get a lot better with kernel support. But we'd still have the problem of wanting this for older kernels.

Here's what I'm thinking of hacking up quickly for now. A tool based on:

```
# ls -l /proc/*/fd/* | grep bpf
ls: cannot access '/proc/10702/fd/255': No such file or directory
ls: cannot access '/proc/10702/fd/3': No such file or directory
ls: cannot access '/proc/self/fd/255': No such file or directory
ls: cannot access '/proc/self/fd/3': No such file or directory
ls: cannot access '/proc/thread-self/fd/255': No such file or directory
ls: cannot access '/proc/thread-self/fd/3': No such file or directory
lrwx------ 1 root     root     64 Mar  9 00:58 /proc/10700/fd/3 -> anon_inode:bpf-map
lrwx------ 1 root     root     64 Mar  9 00:58 /proc/10700/fd/4 -> anon_inode:bpf-prog
lrwx------ 1 root     root     64 Mar  9 00:58 /proc/10700/fd/6 -> anon_inode:bpf-prog
# ls -l /sys/fs/bpf/*/*
ls: cannot access '/sys/fs/bpf/*/*': No such file or directory
```

So FD symlinks with "bpf" reveal process IDs doing BPF, and I can then look up their /proc/PID/comm. Also, take a look in /sys/fs/bpf/\*/\* for any pinned BPF programs.

I (or whoever writes this tool) can add a verbose mode to print /sys/kernel/debug/tracing/enabled_functions, and/or kprobe_events/uprobe_events.

I haven't seen a way to dump enabled tracepoints yet. Are we missing a /sys/kernel/debug/tracing/tracepoint_events?

goldshtn:
This tool displays processes with running BPF programs and maps,
and also optionally kprobes and uprobes. This is a poor-man's version
that snoops BPF file descriptors, as proposed by @brendangregg.

Example:

```
PID    COMM             TYPE     COUNT
4058   fileslower       prog     4
4058   fileslower       map      2
4106   bashreadline     map      1
4106   bashreadline     prog     1
```

Resolves #1036.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@ pair of .c and .py files, and some are directories of files.
 - tools/[biotop](tools/biotop.py): Top for disks: Summarize block device I/O by process. [Examples](tools/biotop_example.txt).
 - tools/[biosnoop](tools/biosnoop.py): Trace block device I/O with PID and latency. [Examples](tools/biosnoop_example.txt).
 - tools/[bitesize](tools/bitesize.py): Show per process I/O size histogram. [Examples](tools/bitesize_example.txt).
+- tools/[bpflist](tools/bpflist.py): Display processes with active BPF programs and maps. [Examples](tools/bpflist_example.txt).
 - tools/[btrfsdist](tools/btrfsdist.py): Summarize btrfs operation latency distribution as a histogram. [Examples](tools/btrfsdist_example.txt).
 - tools/[btrfsslower](tools/btrfsslower.py): Trace slow btrfs operations. [Examples](tools/btrfsslower_example.txt).
 - tools/[capable](tools/capable.py): Trace security capability checks. [Examples](tools/capable_example.txt).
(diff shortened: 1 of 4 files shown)


goldshtn:
BTW I haven't added the code for listing pinned BPF maps because I don't have an easy sample that uses pinned BPF maps. Do you have something @brendangregg?

brendangregg:
Thanks! Looks like a good start.

In the future we might want to do a /proc/*/fd/* walker directly in python, since the "ls -l" may hit a "arglist too long" shell limit.

Can also add a -V at some point for really verbose, where it finishes by printing /sys/kernel/debug/tracing/kprobe_events and uprobe_events. It would be nice to associate that output with the PIDs, but that may need to wait for kernel support.

goldshtn:
@brendangregg Try -vv :-)

brendangregg:
Oh, great, thanks, it's already there!

PR #1044, opened 2017-03-10
syscount: Use ausyscalls if available to get syscall list

### LINKED ISSUE #1001: Use "ausyscall --dump" to generate dynamic map for syscount
syscount uses a static list ot map between syscall name and number.  The static list is tied to x86 and may miss new syscalls added to more recent kernels.  The "ausyscall --dump" can generate a list of event numbers and names.  Using ausyscall might provide a more portable way of getting that mapping information that avoids having to perform manual updates on a static list.

goldshtn:
If ausyscall is installed, it can provide a clean, up-to-date list of
syscall numbers for the current architecture. This is much more useful
than the default hardcoded list for x86-64, which is currently used by
syscount.

Try to run `ausyscall --dump` and parse the output before resorting to
the static list. Tested on FC/Linux 4.9 and produces 327 syscalls.

Resolves #1001.

### DIFF
--- a/tools/syscount.py
+++ b/tools/syscount.py
@@ -13,6 +13,7 @@ from bcc import BPF
 from itertools import izip_longest
 from time import sleep, strftime
 import argparse
+import subprocess
 import sys
 
 #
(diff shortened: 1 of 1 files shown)


4ast:
hmm. how is it better than hardcoded?
syscall numbers will never change.

goldshtn:
Yes, but:
1) if a syscall is added we don't have to update bcc
2) this works on more than just x86-64

PR #1051, opened 2017-03-17
bcc container improvements

kmjohansen:
The following commits add functionality to make bcc tracing of containers work a bit better.

- Resolve symbols in a container where the target process resides in a different mount namespace than the process running bcc.
- Resolve perf-map files either in the host container using the host's view of the target pid, or in the target container using its pid namespace's identifier
- Enable uprobes and USDT probes where the target process' files reside in a different mount namespace from the host.

The uprobe and USDT probe functionality only currently knows how to determine the target's namespcae if a pid is given to bpf_attach_uprobe.  A  future improvement would be to enhance this code path to let the tracing process specify a fd or pid from which namespace information should be gathered in situations where we want to trace all processes triggering a uprobe in a different namespace.  If the maintainers have some suggestions about the best way to accomplish this, I'm happy to take a stab at that too.

### DIFF
--- a/src/cc/bcc_syms.cc
+++ b/src/cc/bcc_syms.cc
@@ -100,12 +100,38 @@ void ProcSyms::refresh() {
 
 int ProcSyms::_add_module(const char *modname, uint64_t start, uint64_t end,
                           void *payload) {
+  struct ns_cookie nsc = {-1, -1};
+  bool ns_switch = false;
+  int arc;
   ProcSyms *ps = static_cast<ProcSyms *>(payload);
   auto it = std::find_if(ps->modules_.begin(), ps->modules_.end(),
                [=](const ProcSyms::Module &m) { return m.name_ == modname; });
-  if (it == ps->modules_.end())
-    it = ps->modules_.insert(ps->modules_.end(), Module(modname, ps->pid_));
+  if (it == ps->modules_.end()) {
+    // If modname references a perf-map, determine if we need to enter a mount
+    // namespace in order to read symbols from it later.
+    if (strstr(modname, ".map") != nullptr) {
+      ns_switch = bcc_procutils_enter_mountns(ps->pid_, &nsc);
+      if (ns_switch) {
+        char new_modname[4096];
+        arc = access(modname, R_OK);
+        bcc_procutils_exit_mountns(&nsc);
+
+        if (arc != 0) {
+          snprintf(new_modname, sizeof (new_modname), "/tmp/perf-%d.map",
+              ps->pid_);
+
+          it = ps->modules_.insert(ps->modules_.end(), Module(new_modname,
+              ps->pid_, false));
+          it->ranges_.push_back(ProcSyms::Module::Range(start, end));
+	  return 0;
+        }
+      }
+    }
+    it = ps->modules_.insert(ps->modules_.end(), Module(modname, ps->pid_,
+        ns_switch));
+  }
   it->ranges_.push_back(ProcSyms::Module::Range(start, end));
+
   return 0;
 }
 
@@ -159,9 +185,9 @@ bool ProcSyms::resolve_name(const char *module, const char *name,
   return false;
 }
 
-ProcSyms::Module::Module(const char *name, int pid)
-  : name_(name), pid_(pid) {
-  struct ns_cookie nsc = {-1, -1};
+ProcSyms::Module::Module(const char *name, int pid, bool in_ns)
+  : name_(name), pid_(pid), in_ns_(in_ns) {
+  struct ns_cookie nsc;
 
   bcc_procutils_enter_mountns(pid_, &nsc);
   is_so_ = bcc_elf_is_shared_obj(name) == 1;
@@ -183,15 +209,17 @@ bool ProcSyms::Module::is_perf_map() const {
 void ProcSyms::Module::load_sym_table() {
   struct ns_cookie nsc = {-1, -1};
 
-   if (syms_.size())
+  if (syms_.size())
     return;
 
-  bcc_procutils_enter_mountns(pid_, &nsc);
-
-  if (is_perf_map())
+  if (is_perf_map()) {
+    if (in_ns_)
+      bcc_procutils_enter_mountns(pid_, &nsc);
     bcc_perf_map_foreach_sym(name_.c_str(), _add_symbol, this);
-  else
+  } else {
+    bcc_procutils_enter_mountns(pid_, &nsc);
     bcc_elf_foreach_sym(name_.c_str(), _add_symbol, this);
+  }
 
   bcc_procutils_exit_mountns(&nsc);
 
--- a/src/cc/bcc_syms.cc
+++ b/src/cc/bcc_syms.cc
@@ -104,7 +104,7 @@ int ProcSyms::_add_module(const char *modname, uint64_t start, uint64_t end,
   auto it = std::find_if(ps->modules_.begin(), ps->modules_.end(),
                [=](const ProcSyms::Module &m) { return m.name_ == modname; });
   if (it == ps->modules_.end())
-    it = ps->modules_.insert(ps->modules_.end(), modname);
+    it = ps->modules_.insert(ps->modules_.end(), Module(modname, ps->pid_));
   it->ranges_.push_back(ProcSyms::Module::Range(start, end));
   return 0;
 }
@@ -159,9 +159,13 @@ bool ProcSyms::resolve_name(const char *module, const char *name,
   return false;
 }
 
-ProcSyms::Module::Module(const char *name)
-  : name_(name) {
+ProcSyms::Module::Module(const char *name, int pid)
+  : name_(name), pid_(pid) {
+  struct ns_cookie nsc = {-1, -1};
+
+  bcc_procutils_enter_mountns(pid_, &nsc);
   is_so_ = bcc_elf_is_shared_obj(name) == 1;
+  bcc_procutils_exit_mountns(&nsc);
 }
 
 int ProcSyms::Module::_add_symbol(const char *symname, uint64_t start,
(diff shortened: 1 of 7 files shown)


4ast:
[buildbot, ok to test]
@markdrayton @palmtenor pls take a lookt

markdrayton:
Thanks for the PR! Making BCC work in container environments is great.

I'm a little confused about the second commit:

> Allow perf-map files to exist either in the container under the pid
that's specific to the container's pid namespace, or in the host
container using the pid that's specific to the initial pid namespace.

I don't quite understand this — how do these changes allow us to find the map under either PID? It's very possible I'm just missing something about the mount namespace handling here.

Does (or can) any of this supersede the existing code to find maps inside different PID namespaces/chroots? (e.g. `bcc_perf_map_nstgid` and `bcc_perf_map_path`)

More generally the code looks fine bar a few nits but it would really benefit from some unit tests. I patched this locally and did some very basic testing but would feel more comfortable with unit tests.

kmjohansen:
Thanks for taking a look @markdrayton .  I appreciate the feedback.

Regarding the question about the 2nd commit, I may have done a poor job of explaining what's going on there.  That commit attempts to address the case where you have a process running in a different pid and mount namespace from the process where you're running the trace.  In that situation, a container may have /tmp mounted on a tmpfs that's separate from the /tmp where you're running bcc.  Without that 2nd commit, the code will always look in the target's /tmp using the target's pid.  So if the target is pid 2 in its namespace, but 4037 in bcc's namespace, it will enter the target's mount namespace and look for /tmp/perf-2.map.  If you managed to put the perf map in the /tmp that's accessible to bcc without changing mount namespaces, it won't look there by default.  The 2nd patch checks /tmp/perf-2.map in the target's mountns, and then also checks for /tmp/perf-4037.map in the mountns that bcc is using.  Does that explanation make more sense?

To the question about `bcc_perf_map_nstgid` and` bcc_perf_map_path`: I think these functions are complimentary, but it's also possible that I don't completely grasp their intent.  The code I added still relies upon `bcc_perf_map_nstgid` in order to determine the pid of the perf map in the target's pid namespace.  I'm less clear about `bcc_perf_map_path`.  I don't think it's harmful, since it would address the case where somebody put a process in a container and then chroot'd it, which might be overkill but it still permitted.

Finally, about unit tests:  I'll take a stab at putting something together and will update the pull request.  Most namespace stuff requires at least `CAP_SYS_ADMIN` in order to make modifications.  Do the tests expect to be running with elevated privileges, or is there something additional I need to do to make that happen?

goldshtn:
The tests should already run with elevated privileges, most of them anyway.

markdrayton:
Thanks for the explanation. It makes more sense now. I need to review the code a bit more but I'm pushed for time at the moment — will try to do so after the tests are ready.

Also, I just realised I forgot to send the inline comments I made. I didn't realise they are submitted independently of regular comments. I'll hit go on that after this.

kmjohansen:
Sorry this took longer to get to than I had hoped.  I've updated the original three patches with the cleanups that @markdrayton requested.  I tried to tidy things up a bit more while I was there too.  This latest update also includes a patch with unit tests.  I'm not enamored with them, but they work and they fail without the leading three patches present too.

markdrayton:
Thanks @kmjohansen! I should be able to look at this tomorrow during a flight.

goldshtn:
@kmjohansen There are two test failures on the Ubuntu buildbot. The first one is actually quite interesting because it looks like a USDT parsing error in libz, I'll try to take a look. The other one seems to be related to your new tests. Can you please review?

goldshtn:
Sorry, scratch that -- it actually looks like /lib64/libz.so.1 doesn't exist.

kmjohansen:
Thanks @markdrayton and @goldshtn for the feedback thus far.

I've uploaded a new version of the unit tests that should do a better job of attempting to locate libz.so.1.  It asks ld.so to do the dirty work instead of trying to guess about the location.

To the question about non perf-map modules in a namespace: yes.  One of the primary motivations was to be able to get stack traces of a nginx process that was running in a container.  This change should work for both java and standard ELF binaries that have symbols (or debug information) available in a container.

As far as consolidating the code that looks for perf maps in different places, I'm happy to make further changes to tidy this up. I agree that it's not as sleek as perhaps it could be, but I was at a bit of a loss as to the best way to improve it.  I'll respond to the specific outstanding code review comments above.

src/cc/bcc_syms.cc
```diff
@@ -100,12 +100,38 @@ void ProcSyms::refresh() {
...
+  int arc;
   ProcSyms *ps = static_cast<ProcSyms *>(payload);
   auto it = std::find_if(ps->modules_.begin(), ps->modules_.end(),
                [=](const ProcSyms::Module &m) { return m.name_ == modname; });
-  if (it == ps->modules_.end())
-    it = ps->modules_.insert(ps->modules_.end(), modname);
+  if (it == ps->modules_.end()) {
+    // If modname references a perf-map, determine if we need to enter a mount
+    // namespace in order to read symbols from it later.
+    if (strstr(modname, ".map") != nullptr) {
```
kmjohansen: I would have, except that is_perf_map() is specific to the Module class.  Since this is constructing the module, it's not yet available.  I can add a separate helper function, if you think that would be useful.

src/cc/bcc_syms.cc
```diff
@@ -177,13 +207,21 @@ bool ProcSyms::Module::is_perf_map() const {
...
+    if (in_ns_)
+      bcc_procutils_enter_mountns(pid_, &nsc);
     bcc_perf_map_foreach_sym(name_.c_str(), _add_symbol, this);
-  else
+  } else {
+    bcc_procutils_enter_mountns(pid_, &nsc);
     bcc_elf_foreach_sym(name_.c_str(), _add_symbol, this);
+  }
+
+  bcc_procutils_exit_mountns(&nsc);
```
kmjohansen: Ah, sorry.  Two different things going on here.  In the case of a perf-map, we want to try to look in the mount namespace if we've already determined that a perf.map is there.  If not, we'll double-check on the host container.  For standard ELF symbols, this code always attempts to look in the target namespace.

src/cc/bcc_syms.cc
```diff
@@ -100,12 +100,38 @@ void ProcSyms::refresh() {
...
+        bcc_procutils_exit_mountns(&nsc);
+
+        if (arc != 0) {
+          snprintf(new_modname, sizeof (new_modname), "/tmp/perf-%d.map",
+             ps->pid_);
+
+          it = ps->modules_.insert(ps->modules_.end(), Module(new_modname,
+              ps->pid_, false));
+          it->ranges_.push_back(ProcSyms::Module::Range(start, end));
+	  return 0;
```
kmjohansen: Sorry, struggling to see this one.  Which line is problematic? When I pull this up in my editor, it looks aligned to me.

goldshtn:
@markdrayton One more look before we merge this?

markdrayton:
Yep, was just looking at this now :-) I think it's more or less good to go at this point though — like @kmjohansen says we can probably consolidate things a bit more but I don't have specific recommendations now. We have good test coverage so refactoring down the road should be straightforward. I'm going to give it a last test in our environment just to make sure the perf map stuff we use works and, assuming that's fine, merging sounds good. Will report back soon!

markdrayton:
Yep, seems to work fine. I tested on processes in the same PID/mount namespace, different namespaces, and those in chroots. Thanks for the work @kmjohansen!

kmjohansen:
Totally welcome.  Thanks @markdrayton and @goldshtn for the review and for being willing to pull the change.  I'm a long time DTrace user and I'm really glad to have this project available for Linux.  Thanks for all the work you're doing to move this forward.

src/cc/bcc_syms.cc
```diff
@@ -100,12 +100,38 @@ void ProcSyms::refresh() {
...
+    // namespace in order to read symbols from it later.
+    if (strstr(modname, ".map") != nullptr) {
+      ns_switch = bcc_procutils_enter_mountns(ps->pid_, &nsc);
+      if (ns_switch) {
+        char new_modname[4096];
+        arc = access(modname, R_OK);
+        bcc_procutils_exit_mountns(&nsc);
+
+        if (arc != 0) {
+          snprintf(new_modname, sizeof (new_modname), "/tmp/perf-%d.map",
```
palmtenor: @kmjohansen: Quick question here. Is it possible that even in the new mount namespace, we should also use [NS-PID](https://github.com/iovisor/bcc/blob/master/src/cc/bcc_perf_map.c#L23) instead of just `ps->pid_`?

Also, I don't get when would this part of logic actually be hit. i.e., access to `modname` fails while access `/tmp/perf-PID.map` works. In fact, to my understanding `chroot` can also happen in combine with mount namespace, so `modname` should actually be better since it is the absolute path taking the Process's root into consideration.

kmjohansen: Sorry for the slow reply.  Not 100% sure I understand the question.   I believe this code was attempting to catch the case where the perf map was in the host system instead of a container, but the process that it referred to was inside a different PID namespace.  In that case, we were attempting to resolve the perf map relative to the primordial mount namespace.

To make things more confusing, there's no canonical definition of what has to be in a linux container, so it's possible to get a process that's in a different mount namespace but shares a pid namespace with the host.  Similarly, it's also possible that a process could be in a different pid namespace but share a mount namespace.  Or they both could be disjoint.  You're correct that chroot isn't out of scope after entering a mountns, but it is awfully redundant for a process to create a full chroot inside a container, which if properly scoped should be its own chroot on steroids.  If you really wanted to be thorough, you could attempt to look up the file relative to the mount namespace, and then resolved against any chroot applied by the process after it entered the mount namespace.  But again, I'm not quite sure I understand what's being asked.

The code above is just trying to locate a perf-map using a PID that would resolve on the host system.

PR #1052, opened 2017-03-18
tcplife for Lua.

monadbobo:
tcplife for Lua.


4ast:
[buildbot, ok to test]
@brendangregg @vmg

vmg:
This looks good to me! I assume the actual BPF C code was ported from the Python implementation, so I haven't reviewed it in depth.

PR #1053, opened 2017-03-18
Use epoll in BPFPerfBuffer

palmtenor:
`BPFPerfBuffer` class already hold all `perf_reader` pointers and FDs, but still now in every loop of `perf_reader_poll` we need to re-construct an FD array. Using `epoll` from the class directly should avoid this.
This would also give the class more control over polling, if we want to add them in the future.

Tested using the `RandomRead` example.

### DIFF
--- a/src/cc/BPFTable.cc
+++ b/src/cc/BPFTable.cc
@@ -14,18 +14,20 @@
  * limitations under the License.
  */
 
-#include <errno.h>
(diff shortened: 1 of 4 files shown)


palmtenor:
The build test seems failed on getting source code?

drzaeus77:
[buildbot, test this please]

PR #1058, opened 2017-03-22
snapcraft: add in some new tools missing from the snapcraft apps list

ColinIanKing:
Add in bpflist and mysqld_qslower to apps list and re-order
dbstat in the list.

Signed-off-by: Colin Ian King <colin.king@canonical.com>


goldshtn:
Should we make a note in the contrib script to update the snap file when adding or removing tools?

ColinIanKing:
I suspect that adding a note to the contrib script would be a useful way to do this.  Note that snapcraft bcc tool names cannot have a lower case _ in them, we need to also ensure they are replaced with -

PR #1065, opened 2017-03-25
cc: define load_func and unload_func public

mvbpolito:
With the current API it is not possible to load any kind of ebpf
program, for example BPF_PROG_TYPE_SCHED_CLS, this commit solves that by
defining load_func and unload_func as part of the public API.

Signed-off-by: Mauricio Vasquez B <mauricio.vasquez@polito.it>

### DIFF
--- a/src/cc/BPF.h
+++ b/src/cc/BPF.h
@@ -103,11 +103,11 @@ public:
   StatusTuple close_perf_buffer(const std::string& name);
   void poll_perf_buffer(const std::string& name, int timeout = -1);
 
-private:
   StatusTuple load_func(const std::string& func_name, enum bpf_prog_type type,
                         int& fd);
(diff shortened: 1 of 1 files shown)


goldshtn:
[buildbot, ok to test]

goldshtn:
@drzaeus77 Shouldn't the command above have triggered the build?

drzaeus77:
[buildbot, ok to test]

@goldshtn I've added you to the buildbot whitelist for [] commands

goldshtn:
@palmtenor Could you please take a look to make sure this makes sense for the C++ API?

palmtenor:
This makes sense. However users load a function and failed to do other things with it, they would need to call unload to clean-up. But we do full clean-up on class destruction anyways, so I wouldn't worry about this.

PR #1066, opened 2017-03-25
solisten: replace netaddr with socket library

### LINKED ISSUE #1035: solisten: No module named netaddr
```
# ./solisten.py 
  File "./solisten.py", line 22, in <module>
    import netaddr
ImportError: No module named netaddr
```

It's using netaddr for IP address to string. This could be switched to use the socket library like tcpconnect.py does, on the assumption it's more readily available than netaddr.

pchaigno:
Replaces the `netaddr` library with `socket`. `netaddr` isn't installed by
default on several distributions.

Fixes #1035.

/cc @brendangregg @goldshtn

### DIFF
--- a/tools/solisten.py
+++ b/tools/solisten.py
@@ -18,8 +18,8 @@
 # 04-Mar-2016	Jean-Tiare Le Bigot	Created this.
 
 import os
-import socket
-import netaddr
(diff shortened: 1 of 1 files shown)


goldshtn:
The fc25 bot needs a refresh... @drzaeus77
(I feel like this should be a [...] command :-))

PR #1067, opened 2017-03-26
u* tools: automatically detect the language

### LINKED ISSUE #864: u* tools: detect language/runtime automatically
By default, the u* tools (uobjnew, uflow, etc.) can detect the language/runtime present in the target process by inspecting its command line or its loaded libraries (/proc/PID/maps). The `-l` switch can then be used for overriding the default.

It might also make sense to make all the tools use `-l` for consistency, instead of having the language as a positional parameter.

_Suggested by @brendangregg_

pchaigno:
Looks into `/proc/$pid/cmdline`, `/proc/$pid/exe`, and
`/proc/$pid/maps` to determine the language. `-l` switch can override
the detected language. In `uthreads` and `ucalls`, the language can
be overwritten to 'none' to trace pthreads and syscalls respectively.

All tools now use the -l switch to set the language.

Should I put the `detect_language` function elsewhere? It's the same
in all tools except for `uobjnew.py` (to handle the C case).

Fixes #864.

/cc @goldshtn @brendangregg

### DIFF
--- a/src/cc/bcc_proc.c
+++ b/src/cc/bcc_proc.c
@@ -485,3 +485,59 @@ bool bcc_procutils_exit_mountns(struct ns_cookie *nc) {
 
   return rc;
 }
+
+/* Detects the following languages + C. */
+const char *languages[] = {"java", "python", "ruby", "php", "node"};
+const char *language_c = "c";
+const int nb_languages = 5;
+
+const char *bcc_procutils_language(int pid) {
+  char procfilename[22], line[4096], pathname[32], *str;
+  FILE *procfile;
+  int i, ret;
+
+  /* Look for clues in the absolute path to the executable. */
+  sprintf(procfilename, "/proc/%ld/exe", (long)pid);
+  if (realpath(procfilename, line)) {
+    for (i = 0; i < nb_languages; i++)
+      if (strstr(line, languages[i]))
+        return languages[i];
+  }
+
+
+  sprintf(procfilename, "/proc/%ld/maps", (long)pid);
+  procfile = fopen(procfilename, "r");
+  if (!procfile)
+    return NULL;
+
+  /* Look for clues in memory mappings. */
+  bool libc = false;
+  do {
+    char perm[8], dev[8];
+    long long begin, end, size, inode;
+    ret = fscanf(procfile, "%llx-%llx %s %llx %s %lld", &begin, &end, perm,
+                 &size, dev, &inode);
+    if (!fgets(line, sizeof(line), procfile))
+      break;
+    if (ret == 6) {
+      char *mapname = line;
+      char *newline = strchr(line, '\n');
+      if (newline)
+        newline[0] = '\0';
+      while (isspace(mapname[0])) mapname++;
+      for (i = 0; i < nb_languages; i++) {
+        sprintf(pathname, "/lib%s", languages[i]);
+        if (strstr(mapname, pathname))
+          return languages[i];
+        if ((str = strstr(mapname, "libc")) &&
+            (str[4] == '-' || str[4] == '.'))
+          libc = true;
+      }
+    }
+  } while (ret && ret != EOF);
+
+  fclose(procfile);
+
+  /* Return C as the language if libc was found and nothing else. */
+  return libc ? language_c : NULL;
+}
--- a/src/cc/bcc_proc.h
+++ b/src/cc/bcc_proc.h
@@ -41,6 +41,7 @@ int bcc_procutils_each_ksym(bcc_procutils_ksymcb callback, void *payload);
 void bcc_procutils_free(const char *ptr);
 bool bcc_procutils_enter_mountns(int pid, struct ns_cookie *nc);
 bool bcc_procutils_exit_mountns(struct ns_cookie *nc);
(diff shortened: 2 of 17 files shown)


goldshtn:
Oh this is amazing! I think the `detect_language` function should be shared, yes. This might actually be a good candidate for inclusion into bcc as a helper -- so it would go inside `bcc_proc.{h,c}`, get exported from libbcc, and then wrapped as a helper on the `BPF` class. I'm thinking there might be additional scenarios where we need this service, and also there's no reason why tools written in other languages (C++, Lua) shouldn't be able to benefit from this.

Does that make sense? Thanks so much for doing this!

goldshtn:
@pchaigno Just wanted to make sure you haven't missed my comment above -- are you OK with these changes?

goldshtn:
[buildbot, test this please]

pchaigno:
I just push an update with the changes :-)
I'll comment on the few pieces of code I'm not sure are needed.

src/cc/bcc_proc.c
```diff
@@ -392,3 +392,74 @@ char *bcc_procutils_which_so(const char *libname, int pid) {
...
+            (str[4] == '-' || str[4] == '.'))
+          libc = true;
+      }
+    }
+  } while (ret && ret != EOF);
+
+  fclose(procfile);
+
+  /* Return C as the language if libc was found and nothing else. */
+  return libc? strdup("c") : NULL;
```
goldshtn: Missing space here before the ?.

src/cc/bcc_proc.h
```diff
@@ -34,6 +34,7 @@ int bcc_procutils_each_module(int pid, bcc_procutils_modulecb callback,
                               void *payload);
 int bcc_procutils_each_ksym(bcc_procutils_ksymcb callback, void *payload);
 void bcc_procutils_free(const char *ptr);
+char *bcc_procutils_language(int pid);
```
goldshtn: We are returning a strdup'd string, so someone has to clean it up, no? Why not just return a const pointer to one of the static strings you have in the `languages` array?

tests/cc/test_c_api.cc
```diff
@@ -30,6 +30,13 @@
 
 using namespace std;
 
+TEST_CASE("language detection", "[c_api]") {
```
goldshtn: Would be nice to also have a Python-side test that makes sure we get "python".

src/cc/bcc_proc.c
```diff
@@ -392,3 +392,74 @@ char *bcc_procutils_which_so(const char *libname, int pid) {
...
+/* Detects the following languages + C. */
+const char *languages[] = {"java", "python", "ruby", "php", "node"};
+const int nb_languages = 5;
+
+char *bcc_procutils_language(int pid) {
+  char procfilename[22], line[4096], pathname[32], *str;
+  FILE *procfile;
+  int i, ret;
+
+  /* Look for clues in the process' command line. */
```
pchaigno: Is the following needed since I'm already looking into `/proc/$pid/exe`? I can't think of a case where the language would be part of the command line but not the `exe` absolute path.

src/cc/bcc_proc.c
```diff
@@ -392,3 +392,74 @@ char *bcc_procutils_which_so(const char *libname, int pid) {
...
+    if (ret == 6) {
+      char *mapname = line;
+      char *newline = strchr(line, '\n');
+      if (newline)
+        newline[0] = '\0';
+      while (isspace(mapname[0])) mapname++;
+      for (i = 0; i < nb_languages; i++) {
+        sprintf(pathname, "/bin/%s", languages[i]);
+        if (strstr(mapname, pathname))
+          return strdup(languages[i]);
```
pchaigno: Is this (looking for `/bin/$language`) the same as checking the absolute path of `/proc/$pid/exe`?

src/cc/bcc_proc.c
goldshtn: Probably not with the languages we currently recognize. BTW, the logic should be a bit more flexible. For example, `irb` (the Ruby REPL) will not be recognized as Ruby by this logic, right?

src/cc/bcc_proc.c
goldshtn: Probably yes, for the languages we currently recognize...

src/cc/bcc_proc.h
pchaigno: Hm. I don't know what I was thinking :/
Will fix.

tests/cc/test_c_api.cc
pchaigno: Done.

src/cc/bcc_proc.c
pchaigno: Not through the command line check but both `/proc/$pid/exe` and `/proc/$pid/maps` will point to `ruby` when using `irb`. `rake` has the same behavior.

When using `jruby`, however, `detect_language` will fail to return `Java` (only because of the `/proc/$pid/cmdline` check). I will remove the `/proc/$pid/cmdline` and `/bin/$language` checks for now; we can always add them back latter if some language requires it.

drzaeus77:
[buildbot, test this please]

PR #1071, opened 2017-03-27
Extend cpp table API

mvbpolito:
The table API used with the BPF class was incomplete, it only allowed to get the value of an element of a table. 

This PR extends the API for the maps of type hash, array and prog (LPM is still missing).

hash:
- provide methods to update, delete and get an element.
- the [] operator allows to update and get the value of an element using the "table[key] = value" and "value = table[key]" syntax. 

array:
- provide methods to update and get an element
- operator [] offers the same functionality as in the hash table

prog:
- provide a method to update an element
- operator [] allows to update an element using the "table[index] = fd" syntax.

### DIFF
--- a/src/cc/BPFTable.h
+++ b/src/cc/BPFTable.h
@@ -16,6 +16,7 @@
 
 #pragma once
 
+#include <cstring>
 #include <sys/epoll.h>
 #include <exception>
 #include <map>
@@ -70,16 +71,37 @@ template <class KeyType, class ValueType>
 class BPFHashTable : protected BPFTableBase<KeyType, ValueType> {
  public:
   explicit BPFHashTable(const TableDesc& desc)
-      : BPFTableBase<KeyType, ValueType>(desc) {}
-
-  ValueType get_value(const KeyType& key) {
-    ValueType res;
-    if (!this->lookup(const_cast<KeyType*>(&key), &res))
-      throw std::invalid_argument("Key does not exist in the table");
-    return res;
+      : BPFTableBase<KeyType, ValueType>(desc) {
+    if (desc.type != BPF_MAP_TYPE_HASH &&
+        desc.type != BPF_MAP_TYPE_PERCPU_HASH &&
+        desc.type != BPF_MAP_TYPE_LRU_HASH &&
+        desc.type != BPF_MAP_TYPE_LRU_PERCPU_HASH)
+      throw std::invalid_argument("Table '" + desc.name + "' is not a hash table");
   }
 
-  ValueType operator[](const KeyType& key) { return get_value(key); }
+  StatusTuple get_value(const KeyType& key, ValueType& value) {
+    if (!this->lookup(const_cast<KeyType*>(&key), &value))
+      return StatusTuple(-1, "Error getting value: %s", std::strerror(errno));
+    return StatusTuple(0);
+  }
+
+  StatusTuple update_value(const KeyType& key, const ValueType& value) {
+    if (!this->update(const_cast<KeyType*>(&key), const_cast<ValueType*>(&value)))
+      return StatusTuple(-1, "Error updating value: %s", std::strerror(errno));
+    return StatusTuple(0);
+  }
+
+  StatusTuple remove_value(const KeyType& key) {
+    if (!this->remove(const_cast<KeyType*>(&key)))
+      return StatusTuple(-1, "Error removing value: %s", std::strerror(errno));
+    return StatusTuple(0);
+  }
+
+  ValueType operator[](const KeyType& key) {
+    ValueType value;
+    get_value(key, value);
+    return value;
+  }
 
   std::vector<std::pair<KeyType, ValueType>> get_table_offline() {
     std::vector<std::pair<KeyType, ValueType>> res;
--- a/src/cc/BPFTable.h
+++ b/src/cc/BPFTable.h
@@ -67,6 +67,45 @@ class BPFTableBase {
   size_t capacity_;
 };
 
+template <class ValueType>
+class BPFArrayTable : protected BPFTableBase<int, ValueType> {
+public:
+  BPFArrayTable(const TableDesc& desc)
+      : BPFTableBase<int, ValueType>(desc) {
+    if (desc.type != BPF_MAP_TYPE_ARRAY &&
+        desc.type != BPF_MAP_TYPE_PERCPU_ARRAY)
+      throw std::invalid_argument("Table '" + desc.name + "' is not an array table");
+  }
+
+  StatusTuple get_value(const int& index, ValueType& value) {
+    if (!this->lookup(const_cast<int*>(&index), &value))
+      return StatusTuple(-1, "Error getting value: %s", std::strerror(errno));
+    return StatusTuple(0);
+  }
+
+  StatusTuple update_value(const int& index, const ValueType& value) {
+    if (!this->update(const_cast<int*>(&index), const_cast<ValueType*>(&value)))
+      return StatusTuple(-1, "Error updating value: %s", std::strerror(errno));
+    return StatusTuple(0);
+  }
+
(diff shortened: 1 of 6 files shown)


goldshtn:
[buildbot, ok to test]

/cc @palmtenor

src/cc/BPFTable.h
```diff
@@ -68,11 +68,104 @@ class BPFTableBase {
...
+    if (!this->update(const_cast<int*>(&index), const_cast<ValueType*>(&value)))
+      throw std::invalid_argument("Error updating element");
+  }
+
+  ValueType operator[](const int& index) const { return get_value(index); }
+
+  BPFArrayTableWriter operator[](const int& index) {
+    return BPFArrayTableWriter(*this, index);
+  }
+  std::vector<std::pair<int, ValueType>> get_table_offline() {
```
palmtenor: Since it is an array, how about we just return a vector of `ValueType`s?

src/cc/BPFTable.h
```diff
@@ -68,11 +68,104 @@ class BPFTableBase {
...
+    std::vector<std::pair<int, ValueType>> res;
+
+    int cur, nxt;
+    ValueType value;
+
+    while (true) {
+      if (!this->next(&cur, &nxt))
```
palmtenor: From what I read in the kernel-side code, it seems all arrays slots are pre-allocated with zero value. We should be able to just query every index directly instead of trying the `next()` calls. @4ast can you confirm?

src/cc/BPFTable.h
```diff
@@ -143,4 +250,41 @@ class BPFPerfBuffer : protected BPFTableBase<int, int> {
...
+      : BPFTableBase<int, int>(bpf_module, name) {
+    int type = bpf_module->table_type(name);
+    if (type != BPF_MAP_TYPE_PROG_ARRAY)
+      throw std::invalid_argument("Table '" + name + "' is not a prog table");
+  }
+
+  // updates an element
+  void update_value(const int& index, const int& value) {
+    if (!this->update(const_cast<int*>(&index), const_cast<int*>(&value)))
+      throw std::invalid_argument("Error updating element");
```
palmtenor: Same here for exception.

src/cc/BPFTable.h
```diff
@@ -68,11 +68,104 @@ class BPFTableBase {
...
+    ValueType res;
+    if (!this->lookup(const_cast<int*>(&index), &res))
+      throw std::invalid_argument("index out of range");
+    return res;
+  }
+
+  // updates an element
+  void update_value(const int& index, const ValueType& value) {
+    if (!this->update(const_cast<int*>(&index), const_cast<ValueType*>(&value)))
+      throw std::invalid_argument("Error updating element");
```
palmtenor: I think we try not to throw as much as possible but return StatusTuple. See some discussion in the original PR (#781)

src/cc/BPFTable.h
```diff
@@ -143,4 +250,41 @@ class BPFPerfBuffer : protected BPFTableBase<int, int> {
   std::unique_ptr<epoll_event[]> ep_events_;
 };
 
+class BPFProgTable : protected BPFTableBase<int, int> {
+private:
+  // this class is used to allow updating a table element using the
+  // table[index] = value syntax
+  class BPFProgTableWriter {
```
palmtenor: The prog array class is essentially the same as normal array. Do you think it would be a good idea to inherit and re-use code?

src/cc/BPFTable.h
```diff
@@ -143,4 +250,41 @@ class BPFPerfBuffer : protected BPFTableBase<int, int> {
...
+    public:
+      BPFProgTableWriter(BPFProgTable& t, const int& i):
```
palmtenor: Moreover, the update wrapper class are doing exact same thing across table types. Can we make it a templated class and reuse everywhere?

src/cc/BPFTable.h
```diff
@@ -68,11 +68,104 @@ class BPFTableBase {
...
+template <class ValueType>
+class BPFArrayTable : protected BPFTableBase<int, ValueType> {
+private:
+  // this class is used to allow updating a table element using the
+  // table[index] = value syntax
+  class BPFArrayTableWriter {
+    public:
+      BPFArrayTableWriter(BPFArrayTable<ValueType>& t, const int& i):
+      table(t), index(i) {}
+      BPFArrayTableWriter& operator=(ValueType const& rhs) {
```
palmtenor: Also, can the update wrapper just return `ValueType` directly instead of overloading the cast?

src/cc/BPFTable.h
mvbpolito: I based my code on the existing one, IMHO opinion we should continue to use exceptions, for example these are some issues if we only use return codes:
- In case of constructors we should implement a static method that instantiates objects as suggested by @4ast 
- It is impossible to use return codes with the [] operator
- It is impossible to return multiple values in a c++ function, then the return value should be passed as a parameter of the function. 

Of course these can be solved, but maybe the pain of doing that is more that remaining with the exception-based approach.

Said that, is it worthy to avoid using exceptions? Am I missing something about using exceptions?

src/cc/BPFTable.h
4ast: yes. it's an array, so just for(i) is good

src/cc/BPFTable.h
4ast: why is it impossible to return multiple values?
I'd really like to see existing throw in get_value() go away as well.
Dropping operator [] is also good thing, since it's semantics are ambiguous.
I especially don't like BPFHashTableWriter operator[].
I see no need to make bpf maps looks like std::maps. They are different and that should be obvious from api.

mvbpolito: Well, it is not that straightforward as in go, however it is also true that with C++11 and C++17 it could be easily done. 

Could you elaborate a little bit more about the why the [] operator's semantics are ambiguous?

According to my understanding what you would like to see as API is very similar to BPFTableBase, not exceptions, not [] operator, just a tiny wrapper over libbpf calls. Isn't it?

4ast: what map[key].field = 0 will do?

mvbpolito:
@palmtenor @4ast I updated the PR handling some of your comments, could you take a look?

@goldshtn it appears that the build was never done, could you ask the bot to redo it?

goldshtn:
[buildbot, test this please]

src/cc/BPFTable.h
```diff
@@ -68,20 +69,82 @@ class BPFTableBase {
...
+  }
+
+  StatusTuple update_value(const int& index, const ValueType& value) {
+    if (!this->update(const_cast<int*>(&index), const_cast<ValueType*>(&value)))
+      return StatusTuple(-1, "Error updating value: %s", std::strerror(errno));
+    return StatusTuple(0);
+  }
+
+  ValueType operator[](const int& key) {
+    int value;
```
palmtenor: Should be `ValueType` here?

mvbpolito: You are right, just waiting for any other comments before updating it.

palmtenor:
Looks good to me. Once someone kick the testbot I think we could land this:)

goldshtn:
[buildbot, test this please]

palmtenor:
@mvbpolito Can you rebase to current trunk so that we can get this merged? I think there are some conflicts after #1040. Thanks a lot:))

mvbpolito:
@palmtenor done!

Could somebody kick the testbot?

goldshtn:
[buildbot, test this please]

mvbpolito:
Any idea of why the build bot is trying to pick up a wrong commit id?

goldshtn:
Let's try again :-)
[buildbot, test this please]

drzaeus77:
[buildbot, test this please]

PR #1074, opened 2017-03-28
Update to -std=c++14, remove custom make_unique

drzaeus77:
This updates our compiler requirements to approximately gcc 4.9/5.0 and
clang 3.4.

### DIFF
--- a/src/cc/common.h
+++ b/src/cc/common.h
@@ -23,11 +23,7 @@
 
 namespace ebpf {
 
-template <class T, class... Args>
-typename std::enable_if<!std::is_array<T>::value, std::unique_ptr<T>>::type
-make_unique(Args &&... args) {
-  return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
-}
+using std::make_unique;
 
 std::vector<int> get_online_cpus();
 
--- a/src/cc/common.h
+++ b/src/cc/common.h
@@ -23,8 +23,6 @@
 
 namespace ebpf {
 
-using std::make_unique;
-
 std::vector<int> get_online_cpus();
 
 std::vector<int> get_possible_cpus();
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,7 +63,7 @@ if (GCC_VERSION VERSION_LESS 6.0)
   endforeach()
 endif()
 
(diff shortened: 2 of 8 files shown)


src/cc/common.h
```diff
@@ -23,11 +23,7 @@
 
 namespace ebpf {
 
-template <class T, class... Args>
-typename std::enable_if<!std::is_array<T>::value, std::unique_ptr<T>>::type
-make_unique(Args &&... args) {
-  return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
-}
+using std::make_unique;
```
goldshtn: Do we really need this in the header? Why?

drzaeus77: Less code churn. I don't mind tracking down the few uses and fixing those up.

4ast:
could you define precisely what is the minimum version of clang and gcc that would be required?
i'm worried about breaking distro builds. I think some are still running 4.8.
What is the advantage? other than it looks cleaner.

drzaeus77:
For clang it is 3.4 minimum required. For gcc it is 4.9.

C++14 brings mostly ergonomic improvements. For instance, see the automatic default member initializer section in http://en.cppreference.com/w/cpp/language/data_members#Member_initialization, which I make use of in the new definition of the TableDesc and FileDesc classes in https://github.com/iovisor/bcc/pull/1040/files#diff-518cfb5455fc552447ecf859f684527dR92.

drzaeus77:
> i'm worried about breaking distro builds. 

Is centos the weak link here?

4ast:
centos7 is still on gcc4.8 and I believe more or less recent fedora as well.
can we ask distro folks ?

drzaeus77:
Fedora 21 was the first to support gcc 4.9...not that recent. I'm pretty sure if centos7 hasn't updated by now then they never will. Ubuntu 14.04 also is stuck on 4.8. But we're not building packages for either of those, so I don't have any visibility into what would break, and who is tracking this. I guess based on the other issue that was getting comments today, debian folks are interested as well.

4ast:
imo it feels to early to dump gcc4.8
@brendangregg @goldshtn thoughts?

goldshtn:
I feel that C++ 14 is not worth breaking older distros. The cleanup introduced here is nice, but can easily be worked around with C++ 11 features.

PR #1078, opened 2017-03-30
funcslower: Trace slow kernel or user function calls

### LINKED ISSUE #915: new tool idea: funcslower
This ticket is to request a new tool, funcslower, that acts like the other *slower tools, but in this case for function time. A little like a per-event counterpart of funclatency. Mock-up output:

```
# funcslower vfs_write
Tracing the vfs_write() function slower than 1 ms. Ctrl-C to end.
COMM           PID     LAT(ms)    RVAL FUNC
dd             25570      3.92     640 vfs_write
dd             25570      1.41      -1 vfs_write
dd             25686     88.28     128 vfs_write
dd             25686    119.66   65536 vfs_write
```

Why RVAL (return value) as well? It's really useful (low-latency can mean error path), and will be there anyway on our function return when we're calculating the time delta, so easy to add.

I'd place the FUNC column last, as the function name can be quite long (so putting it last avoids an unnecessary wide column of padding), and, we can have an optional argument -a to show entry arguments (as hex), just tacked onto the end, eg:

```
# funcslower -a4 vfs_write
Tracing the vfs_write() function slower than 1 ms. Ctrl-C to end.
COMM           PID     LAT(ms)    RVAL FUNC ARGS
dd             25570      3.92     640 vfs_write 0xfffff816290 0xffff771810 0x280 0xffff67491
dd             25570      1.41      -1 vfs_write 0xfffff816290 0xffff771810 0x280 0xffff67491
dd             25686     88.28     128 vfs_write 0xfffff816290 0xffff771810 0x80 0xffff67491
dd             25686    119.66   65536 vfs_write 0xfffff816290 0xffff771810 0x10000 0xffff67491
```

There I said -a4 for 4 entry arguments. -a could default to 3 if no number is given. They can be presented in hex; anything beyond that can be left to trace.py. At some point, trace could get an option for printing function latency as well, although trace is complicated enough for what it does, and much more complicated should be left to custom programs.

Other options to funcslower should be similar to other tools. Eg:

```
-T	human readable timestamp column on left, HH:MM:SS
-t	timestamp column on left in seconds, to 6 or 9 decimal places (us or ns resolution)
-p PID	PID filter
```

The tool has an optional positional parameter for the minimum latency filter, which could default to 1 ms. Some example usage:

```
funcslower vfs_write           # only show kernel vfs_write() calls slower than, say, 1 ms
funcslower vfs_write 10        # ... slower than 10 ms
funcslower -u vfs_write 10     # ... slower than 10 us
funcslower vfs_write 0         # dump all calls
funcslower c:open              # trace libc open()
funcslower bash:readline       # trace bash's readline()
funcslower vfs_read vfs_write  # trace both these
```

I would not like to see something that allowed you to do this:

```
funcslower c:*         # time all functions in libc
funcslower vfs*        # time all vfs functions
```

... as the risk of shooting yourself in the foot by tracing too much is high. Even if the tool worked (and bcc didn't reject it for tracing too many probes), I'd stop trusting the latency measurements it reported.

The man page and examples file should stress that you should use funclatency and argdist first, as they summarize in-kernel and have lower overhead.

FWIW, there's an example in the argdist man page for its usage:

```
./argdist -C 'r::__vfs_read():u32:$PID:$latency > 100000'  
```

The man page should also make it clear that the overhead can be significant if you trace functiowns with a high frequency. No pulling punches. If you traced, say, libc malloc() latency, I'd imagine the target application (with the current uprobes implementation) could run 2x slower or more.

Difficulties: recursive or nested functions. funclatency doesn't handle them currently, and says so in the man page. It should be fixed one day to handle these, provided the additional overhead is acceptable (and if not, it'll remain something these tools have chosen not to do, and people can write custom tools if they must).

goldshtn:
This tool attaches to entry and return points of specified
functions and traces function calls slower than a given threshold.

Example:

```
# ./funcslower __kmalloc -a 2 -u 1
Tracing function calls slower than 1 us... Ctrl+C to quit.
COMM           PID    LAT(us)             RVAL FUNC ARGS
kworker/0:2    27077     7.46 ffff90054f9f8e40 __kmalloc 0x98 0x1400000
kworker/0:2    27077     6.84 ffff90054f9f8e40 __kmalloc 0x98 0x1400000
bash           11527     6.87 ffff90054f9f8e40 __kmalloc 0x90 0x1408240
bash           11527     1.15 ffff90054f9f8e40 __kmalloc 0x90 0x1408240
bash           11527     1.15 ffff90055a1b8c00 __kmalloc 0x2c 0x1400240
bash           11527     1.18 ffff90054b87d240 __kmalloc 0x1c 0x1400040
bash           11527    10.59 ffff900546d60000 __kmalloc 0x10000 0x14082c0
bash           11527     1.49 ffff90054fbd4c00 __kmalloc 0x280 0x15080c0
bash           11527     1.00 ffff90054789b000 __kmalloc 0x800 0x15012c0
bash           27128     3.47 ffff90057ca1a200 __kmalloc 0x150 0x1400240
bash           27128     1.82 ffff90054fbd4c00 __kmalloc 0x230 0x14000c0
bash           27128     1.17 ffff90054b87d5a0 __kmalloc 0x1c 0x14000c0
perf           27128     4.81 ffff90054f9f8e40 __kmalloc 0x90 0x1408240
perf           27128    24.71 ffff900566990000 __kmalloc 0x10000 0x14082c0
^C
```

Resolves #915.

/cc @brendangregg

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ pair of .c and .py files, and some are directories of files.
 - tools/[filetop](tools/filetop.py): File reads and writes by filename and process. Top for files. [Examples](tools/filetop_example.txt).
 - tools/[funccount](tools/funccount.py): Count kernel function calls. [Examples](tools/funccount_example.txt).
 - tools/[funclatency](tools/funclatency.py): Time functions and show their latency distribution. [Examples](tools/funclatency_example.txt).
+- tools/[funcslower](tools/funcslower.py): Trace slow kernel or user function calls. [Examples](tools/funcslower_example.txt).
 - tools/[gethostlatency](tools/gethostlatency.py): Show latency for getaddrinfo/gethostbyname[2] calls. [Examples](tools/gethostlatency_example.txt).
 - tools/[hardirqs](tools/hardirqs.py):  Measure hard IRQ (hard interrupt) event time. [Examples](tools/hardirqs_example.txt).
 - tools/[killsnoop](tools/killsnoop.py): Trace signals issued by the kill() syscall. [Examples](tools/killsnoop_example.txt).
--- a/man/man8/funcslower.8
+++ b/man/man8/funcslower.8
@@ -1,0 +1,112 @@
+.TH funcslower 8  "2017-03-30" "USER COMMANDS"
+.SH NAME
+fileslower \- Trace slow kernel or user function calls.
+.SH SYNOPSIS
+.B fileslower [\-h] [\-p PID] [-m MIN_MS] [-u MIN_US] [-a ARGUMENTS] [-T] [-t] [-v] function [function ...]
+.SH DESCRIPTION
+This script traces a kernel or user function's entry and return points, and
+prints a message when the function's latency exceeded the specified threshold.
+Multiple functions are supported, and you can mix kernel functions with user
+functions in different libraries.
+
+WARNING: See the OVERHEAD section.
+
+By default, a minimum millisecond threshold of 1 is used. Recursive functions
+are not supported: only the inner-most recursive invocation will be traced.
+
+Since this uses BPF, only the root user can use this tool.
+.SH REQUIREMENTS
+CONFIG_BPF and bcc.
+.SH OPTIONS
+\-p PID
+Trace this PID only.
+.TP
+\-m MIN_NS
+Minimum duration to trace, in milliseconds. Default is 1 ms.
+.TP
+\-u MIN_US
+Minimum duration to trace, in microseconds.
+.TP
+\-a ARGUMENTS
+Print the function's arguments, up to 6.
+.TP
+\-T
+Print a HH:MM:SS timestamp with each entry.
+.TP
+\-t
+Print a seconds timestamp with each entry, at microsecond resolution.
+.TP
+\-v
+Print the resulting BPF program, for debugging purposes.
+.TP
+function
+The function to trace -- multiple functions are supported. If a plain function
+name is provided, the function is assumed to be a kernel function. For user
+functions, provide the library name and the function name, e.g. bash:readline
+or c:malloc.
+.SH EXAMPLES
+.TP
+Trace vfs_write calls slower than 1ms:
+#
+.B funcslower vfs_write
+.TP
+Trace open() calls in libc slower than 10us:
+#
+.B funcslower \-u 10 c:open
+.TP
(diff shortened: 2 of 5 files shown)


brendangregg:
I like it, but I don't like needing a -u or -m for the default usage. Unix has a way of not needing options for default usage, eg:

```
ls *.zip
mv *.zip /tmp
chmod 644 /tmp/*.zip
```

So could funcslower do:

```
funcslower __kmalloc 1           # 1+ ms
funcslower __kmalloc 0.1         # 100+ us
funcslower __kmalloc -u 100      # 100 + us
```

ie, default to ```funcslower function milliseconds```. Optionally use -u to switch to microseconds.

goldshtn:
You wanted the function option to allow multiple functions, and that makes the syntax ambiguous. Unless I can tell ArgParser to somehow consider numbers as times and alphanumeric strings as function names. Consider:

funcslower __kmalloc 100

Could be a function called "100" or a time threshold. 

What do you think?

goldshtn:
[buildbot, test this please]

markdrayton:
Can it be:

```
funcslower func [func, ..]         # default (say 1ms)
funcslower -m 10 func [func, ..]   # 10ms
funcslower -u 100 func [func, ..]  # 100us
```

I'm not a fan of deducing arguments from types and prefer the positional args to be last (e.g. like `grep`). IMO predictability/simplicity trump flexibility here.

goldshtn:
That's what it is right now :-)

brendangregg:
chown deduces UID/username based on numbers...

Anyway, I can just alias funccount="funccount -m" in my .bashrc and then use it with the latency first and then functions.

Can it support fractions?

```
# ./funcslower.py -m 0.1 SyS_nanosleep
usage: funcslower.py [-h] [-p PID] [-m MIN_MS] [-u MIN_US] [-a ARGUMENTS] [-T]
                     [-t] [-v]
                     function [function ...]
funcslower.py: error: argument -m/--min-ms: invalid int value: '0.1'
```

goldshtn:
Done, the latest patchset allows fractional intervals, so your example with `-m 0.1` now works. If the tests are OK, this is ready for merging.

goldshtn:
@brendangregg OK to merge?

brendangregg:
LGTM. Good WARNING about overhead, did mention recursive problems... merging...

PR #1079, opened 2017-03-30
stacksnoop: Retire and move to examples/tracing

### LINKED ISSUE #737: retire stacksnoop
Now that we have trace.py -K and -U for kernel and user stacks, we might not need tools/stacksnoop.

I wanted to bring this up before @goldshtn fixes it :)

I think it might make sense to move stacksnoop to examples/tracing, since it's a short example of stack walking, and then delete stacksnoop's _example.txt file and man page (since it's no longer a tool). I can do a PR, just wanted to raise it here first.

goldshtn:
Move stacksnoop to examples/tracing, and leave a script in tools/
that redirects to trace.py with the appropriate arguments. The old
-s and -v switches are left for compatibility, but are ignored.

Resolves #737.


brendangregg:
Looks good, although I'd consider now or later taking the opportunity to add user stacks. I think it could work like profile.py (which means changing it's args a little):

- default: both user & kernel stacks
- -U: user only
- -K: kernel only

goldshtn:
I can do that of course, but the point sort of was to maintain compatibility and familiarity - the old stacksnoop didn't support user functions at all. I'll add and update the PR.

brendangregg:
right, thanks, I just think it's still early stages for tool adoption, so changing things isn't too painful yet. Plus I'm not sure stacksnoop had much usage yet anyway.

goldshtn:
Then maybe we could get rid of the tool entirely and leave just the example in examples/tracing? :-)

goldshtn:
[buildbot, test this please]

brendangregg:
Yes, we can get rid of tools/stacksnoop entirely and just have it as an example.

I don't want to lose your wrapper code if you update the PR, though, since we will probably want it again at some point. So I'll paste it here:

```Python
#!/usr/bin/python
#
# stacksnoop    Trace a kernel function and print all kernel stack traces.
#               For Linux, uses BCC, eBPF, and currently x86_64 only.
#
# USAGE: stacksnoop [-h] [-p PID] [-s] [-v] function
#
# Copyright 2016 Netflix, Inc.
#
# 12-Jan-2016   Brendan Gregg     Created this.
# 30-Mar-2017   Sasha Goldshtein  Migrated to a wrapper on top of `trace`.

import argparse
import os
import subprocess

examples = """examples:
    ./stacksnoop ext4_sync_fs    # print kernel stack traces for ext4_sync_fs
    ./stacksnoop -s ext4_sync_fs    # ... also show symbol offsets
    ./stacksnoop -v ext4_sync_fs    # ... show extra columns
    ./stacksnoop -p 185 ext4_sync_fs    # ... only when PID 185 is on-CPU
"""
    description="Trace and print kernel stack traces for a kernel function",
    epilog=examples)
parser.add_argument("-p", "--pid",
    help="trace this PID only")
parser.add_argument("-s", "--offset", action="store_true",
    help="show address offsets; ignored")
parser.add_argument("-v", "--verbose", action="store_true",
    help="print more fields; ignored")
parser.add_argument("function",
    help="kernel function name")
args = parser.parse_args()

if os.path.basename(__file__) == "stacksnoop.py":
    trace = os.path.join(os.path.dirname(__file__), "trace.py")
else:
    trace = os.path.join(os.path.dirname(__file__), "trace")

try:
    subprocess.call(trace + " -Kt " + args.function, shell=True)
except KeyboardInterrupt:
pass
```

goldshtn:
Done, got rid of `stacksnoop` in the tools directory. If the tests are OK this is ready to merge.

PR #1082, opened 2017-03-31
KVM hypercall analysis example

tuxology:
Small example to show how we can add associated trace events based on conditions being met for other trace events. Here we trace hypercalls and the associated `kvm_exit` and `kvm_extry` functions for a given exit reason. May be further useful for hypercall latency analysis.


goldshtn:
[buildbot, ok to test]

goldshtn:
[buildbot, ok to test]

tuxology:
@goldshtn Yes, indeed. I just added it. There was a dead-link which I removed as well

goldshtn:
Thanks for this!

PR #1084, opened 2017-03-31
Add an option to resolve address without demangling

palmtenor:
In some continuous profiling use cases, we would to minimize the overhead as much as possible. While we often need to resolve symbols right away, we can deferred the demangling work later or elsewhere.

On the other hand, on some large binary with complicated mangled names, the demangling could actually take non-negligible resource.

This PR adds a demangling option to `KSyms` and `ProcSyms`'s `resolve_addr` method, with a default `true` value to keep current behavior unchanged. When the option is `false`, the `demangle_name` pointer in the returned `bcc_symbol` structure will be left empty.

Because methods in `bcc_syms.h` are C `extern` function, we had to add a new `bcc_symcache_resolve_no_demangle` method for the new behavior.


goldshtn:
Looks good, but should this be exposed by the Python wrapper? There's precedent with show_address and show_offset.

palmtenor:
Makes sense, added:)
Tested by hard-coding the argument in `trace.py`

goldshtn:
Is there a way to test this automatically?

palmtenor:
There could be... But I looked at current Python symboling tests, they are all on a C-compiled binary. If we want to test demangling quite some things need to be changed:(

goldshtn:
Really? Isn't it just complying the dummy.c as dummy.cc? There aren't many tests either. I could take a look at this later...

palmtenor:
So... the test VM worked on the first two commit pushes, and gave up on the third. Slacker VM.

palmtenor:
@goldshtn I think you are right, moving the dummy file to C++ is quite easy. I added the `nm` step because theoritically mangling can differ across compilers and platform, so I don't want to hard-code.

goldshtn:
Thanks, this looks good! Merging.

PR #1085, opened 2017-04-01
map.insert bcc helper to expose the BPF_NOEXIST flag

pchaigno:
This pull request adds a new bcc helper, `map.insert()` to
expose the `BPF_NOEXIST` flag of the eBPF
`BPF_MAP_UPDATE_ELEM` helper.

It inserts element in map only if it does not already exist. The bcc
rewriter will throw a warning during if `.insert()` is used on a BPF
array (is that the right way to handle it?).

I found a single bcc script (`examples/networking/neighbor_sharing`)
where it makes sense to use `.insert()`.

I've been needing this to avoid allocating a new htab element when
updating a hash table whose values never change.

### DIFF
--- a/tests/python/test_clang.py
+++ b/tests/python/test_clang.py
@@ -3,7 +3,7 @@
 # Licensed under the Apache License, Version 2.0 (the "License")
 
 from bcc import BPF
-import ctypes
+import ctypes as ct
 from unittest import main, TestCase
 import os
 import sys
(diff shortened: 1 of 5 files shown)


goldshtn:
LGTM, but would it be possible to include a test that explicitly verifies the new behavior? I.e., that `map.insert` actually doesn't insert if the element already exists? I think you could pre-populate the map from Python with some value, make sure the BPF probe is run at least once, and then just verify that the original value was not overwritten. Or is there something that makes this particularly tricky?

4ast:
lgtm

tests/python/test_clang.py
```diff
@@ -485,5 +484,25 @@ def test_printk_2s(self):
...
+    dummy.update(&key, &val);
+}
+"""
+        b = BPF(text=text)
+        c_val = ct.c_ulong(1)
+        b["dummy"][ct.c_ulong(0)] = c_val
+        b["dummy"][ct.c_ulong(1)] = c_val
+        b.attach_kprobe(event="sys_sync", fn_name="do_trace")
+        libc = ct.CDLL("libc.so.6")
+        libc.sync()
```
pchaigno: Is there something "lighter" I could trace?

pchaigno:
> would it be possible to include a test that explicitly verifies the new behavior?

Done in `test_map_insert` by tracing a `sync` syscall.

tests/python/test_clang.py
goldshtn: I guess you could trace some cheaper syscall, like getuid or getpid, but I don't think it matters very much.

PR #1086, opened 2017-04-01
u* tools: per-language wrappers

### LINKED ISSUE #980: {java,ruby,node,php,python}{calls,flow,gc,objnew,stat,threads} wrappers for u* tools
Per @brendangregg 's [suggestion](https://github.com/iovisor/bcc/pull/970#issuecomment-279958323), move the u* tools to tools/lib and create wrapper scripts per language, making them more discoverable. E.g. javagc:

```bash
#!/bin/bash
lib=$(dirname)/lib
$lib/ugc -l java "$@"
```

The man pages and examples files can be symlinked and shared across the tools:

```
tools/javagc_example.txt -> tools/lib/ugc_example.txt
tools/nodegc_example.txt -> tools/lib/ugc_example.txt
...
man/man8/javagc.8 -> man/man8/ugc.8
man/man8/rubygc.8 -> man/man8/ugc.8
...
```

See also #864 for auto-detecting runtime and making all tools take `-l` switch for consistency.

pchaigno:
Fixes #980.

/cc @brendangregg @goldshtn


goldshtn:
Thanks, this is great! The tests are now failing because the u* tools have been moved into tools/lib, so this needs to be fixed. Also, did you make sure that the installer correctly copies the files into /usr/share/bcc/tools? I just ask because there's the whole .py stripping step and I want to make sure it works even if the scripts don't have the .py extension.

pchaigno:
Quite a few things were missing in the first version :s

It should be better now. `u*` tools will go in `/usr/share/bcc/tools/lib/`, `{python,java,node,php,ruby,c}*` tools in `/usr/share/bcc/tools/` (same for example files but in `/usr/share/bcc/tools/doc/lib/` and `/usr/share/bcc/tools/doc/`). All man pages go directly in `/usr/share/bcc/man/man8/`. Bash wrappers for `u*` tools will have their `.sh` file extension stripped during installation. How does that sound?

I've also made changes to the Snapcraft config, but I'm not sure they're correct/sufficient. @ColinIanKing Could you have a look?

goldshtn:
No, sorry, this still isn't quite right, because the symlinks are broken after installing (because the .py extension is removed):

```
# /usr/share/bcc/tools/cobjnew
/usr/share/bcc/tools/cobjnew: line 3: /usr/share/bcc/tools/lib/uobjnew.py: No such file or directory
```

pchaigno:
Okay, this should be fixed. It wasn't a problem with symlinks but with Bash wrapper scripts. I had to pull some of CMake's magic to rewrite them during the installation. It's kinda ugly and it means `.sh` files will be copied at each installation even if they were not modified. 

@drzaeus77 Looks like fc25 needs a reboot :-/

drzaeus77:
[buildbot, test this please]

goldshtn:
Let's hope this is transient. [buildbot, test this please]

goldshtn:
All right, thanks very much for doing this!

pchaigno:
Thanks for the careful review :p

brendangregg:
Thanks!

I'd actually add all these tools to the README.md. So you can see each of the the "java", "node", etc tools as separate lines.

PR #1091, opened 2017-04-03
softirqs: Migrate to kernel tracepoints instead of kprobes

### LINKED ISSUE #1031: softirqs is broken on newer kernels

goldshtn:
This commit migrates softirqs to use kernel tracepoints instead of
kprobes. Because tracepoints only provide the vector number and not
the function name, we use a conversion table, which is borrowed from
kernel/softirq.c, to translate the vector number to a display name.
This table is expected to be fairly stable. Notably, new names have
not been added since approximately 2009, and the last rename (without
adding or removing a name) was in 2014.

Resolves #1031.

### DIFF
--- a/man/man8/softirqs.8
+++ b/man/man8/softirqs.8
@@ -9,10 +9,10 @@ show this time as either totals or histogram distributions. A system-wide
 summary of this time is shown by the %soft column of mpstat(1), and soft IRQ
 event counts (but not times) are available in /proc/softirqs.
 
-WARNING: This currently uses dynamic tracing of various soft interrupt
-functions, and can easily not work with different kernel versions. Check and
-adjust the code as necessary. Also try in a test environment and ensure this
-tool is safe before use. Future versions should switch to tracepoints.
+This tool uses the irq:softirq_enter and irq:softirq_exit kernel tracepoints,
+which is a stable tracing mechanism. BPF programs can attach to tracepoints
+from Linux 4.7 only. An older version of this tool is available in tools/old,
+and uses kprobes instead of tracepoints.
 
 Since this uses BPF, only the root user can use this tool.
 .SH REQUIREMENTS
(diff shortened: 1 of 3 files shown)


goldshtn:
@brendangregg @4ast What do you think about this approach of converting softirq vec to name? I don't see anything more stable than the in-kernel array, which I can't read from the BPF program.

goldshtn:
@brendangregg @4ast Any objections to merging this?

brendangregg:
Thanks for taking this on. A side-by-side comparison:

New:

```
# ./softirqs.py 
^C
SOFTIRQ          TOTAL_usecs
net_tx                   175
net_rx                  1954
rcu                     4675
timer                   5044
sched                   5128
```

old:

```
# ./softirqs.py 
^C
SOFTIRQ                    TOTAL_usecs
net_tx_action                       56
net_rx_action                     1476
rcu_process_callbacks             1636
run_timer_softirq                 1974
run_rebalance_domains             3646
```

So I'm just wondering about run_rebalance_domains becomming sched, which seems less informative... thinking...

It's minor, but I'd have this comment:

```
    # from kernel/softirq.c
```

be something like:

```
    # copied from softirq_to_name() in kernel/softirq.c, may need updates:
```

vitalyvch:
Good work !-)

4ast:
I think the new naming is more concise and easier to understand. So lgtm

goldshtn:
Rebased and rephrased the comment according to @brendangregg 's suggestion.

brendangregg:
Yes, but "run_rebalance_domains" has become "sched". We can merge it, I'll just come up with another way (option or another tool) to split up sched if that becomes necessary.

PR #1092, opened 2017-04-04
Add callback in perf_reader to handle lost sample events

palmtenor:
Currently for `PERF_RECORD_LOST` events, we just print a log to `stderr`. That is not very helpful in production systems.
This PR adds a callback to handle such event, so that user could choose to monitor / export / log in different ways for sample lost events.


drzaeus77:
[buildbot, test this please]

palmtenor:
Fix Lua, update documentation.

PR #1102, opened 2017-04-07
Fix compiler warning on redefined Macro

palmtenor:
The `_GNU_SOURCE` flag is being passed by command-line argument most of the time, causing a warning:
```
[ 52%] Building C object src/cc/CMakeFiles/bcc-shared.dir/bcc_proc.c.o
/tmp/rpmbuild.hW1e0H/BUILD/bcc/src/cc/bcc_proc.c:16:0: warning: "_GNU_SOURCE" redefined
 #define _GNU_SOURCE
```

### DIFF
--- a/src/cc/BPFTable.cc
+++ b/src/cc/BPFTable.cc
@@ -136,12 +136,14 @@ StatusTuple BPFPerfBuffer::close_all_cpu() {
   std::string errors;
   bool has_error = false;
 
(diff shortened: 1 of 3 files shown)


palmtenor:
Piggy back a fix for `BPFPerfBuffer` destruction. Some times user explicitly close the buffer but we would call `close_all_cpu` again in destructor.

PR #1104, opened 2017-04-11
[RFC]: Extend table API with string support 

mauriciovasquezbernal:
Currently it is only possible to access the key/leaf from/to string functions using the bpf_module class.
This PR extends the high level c++ api to also support it.

First commit exposes the conversion functions. The second one exposes a new kind of table that can be accessed using only strings without knowing structure of the data saved on the map.

Please note that this feature is useful when implementing remote map access, as hover for example.

### DIFF
--- a/src/cc/BPFTable.cc
+++ b/src/cc/BPFTable.cc
@@ -31,6 +31,61 @@
 
 namespace ebpf {
 
+BPFTable::BPFTable(const TableDesc& desc) : BPFTableBase<void, void>(desc) {}
+
+StatusTuple BPFTable::get_value(const std::string& key_str,
+                                std::string& value_str) {
+  char key[desc.key_size];
+  char value[desc.leaf_size];
+
+  StatusTuple r(0);
+
+  r = string_to_key(key_str, key);
+  if (r.code() != 0)
+    return r;
+
+  if (!lookup(key, value))
+    return StatusTuple(-1, "error getting value");
+
+  return leaf_to_string(value, value_str);
+}
+
+StatusTuple BPFTable::update_value(const std::string& key_str,
+                                   const std::string& value_str) {
+  char key[desc.key_size];
+  char value[desc.leaf_size];
+
+  StatusTuple r(0);
+
+  r = string_to_key(key_str, key);
+  if (r.code() != 0)
+    return r;
+
+  r = string_to_leaf(value_str, value);
+  if (r.code() != 0)
+    return r;
+
+  if (!update(key, value))
+    return StatusTuple(-1, "error updating element");
+
+  return StatusTuple(0);
+}
+
+StatusTuple BPFTable::remove_value(const std::string& key_str) {
+  char key[desc.key_size];
+
+  StatusTuple r(0);
+
+  r = string_to_key(key_str, key);
+  if (r.code() != 0)
+    return r;
+
+  if (!remove(key))
+    return StatusTuple(-1, "error removing element");
+
+  return StatusTuple(0);
+}
+
 BPFStackTable::~BPFStackTable() {
   for (auto it : pid_sym_)
     bcc_free_symcache(it.second, it.first);
--- a/src/cc/BPFTable.h
+++ b/src/cc/BPFTable.h
@@ -36,39 +36,83 @@ namespace ebpf {
 template <class KeyType, class ValueType>
 class BPFTableBase {
  public:
-  size_t capacity() { return capacity_; }
+  size_t capacity() { return desc.max_entries; }
 
- protected:
-  explicit BPFTableBase(const TableDesc& desc) {
-    fd_ = desc.fd;
-    capacity_ = desc.max_entries;
+  StatusTuple string_to_key(const std::string& key_str, KeyType* key) {
+    if (!desc.key_sscanf)
+      return StatusTuple(-1, "Key sscanf not available");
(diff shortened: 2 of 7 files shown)


goldshtn:
Thanks! Would you mind including some tests that would also help demonstrate this feature? I assume updating the docs (reference guide) might also be a good idea.

goldshtn:
[buildbot, test this please]

src/cc/BPFTable.h
```diff
@@ -36,42 +36,83 @@ namespace ebpf {
...
 
   bool remove(KeyType* key) {
-    return bpf_delete_elem(fd_, static_cast<void*>(key)) >= 0;
+    return bpf_delete_elem(desc.fd, static_cast<void*>(key)) >= 0;
   }
 
-  int fd_;
-  size_t capacity_;
+  const TableDesc &desc;
+  BPFModule *bpf_module;
```
palmtenor: Change these to `desc_` and `bpf_module_` to go along with naming convention of other classes?
Also, what's the reason choosing pointer for `BPFModule` but reference for `TableDesc`? Did I miss something?

src/cc/BPFTable.h
```diff
@@ -36,42 +36,83 @@ namespace ebpf {
...
+  size_t id; // table id inside bpf_module
```
palmtenor: I think in the long run we want to eliminate the Table ID thing (which is essentially just index of the table in one particular `BPFModule`, and doesn't make so much sense globally), see discussions in #1040. 

Can you move the `scanf` / `printf` logic into `TableDesc` and just call that here? `BPFModule` can probably also just proxy the call for backward compatibility.

@drzaeus77 what do you think?

mvbpolito:
@goldshtn Sure, I'll add tests and documentation as soon as I have some feedback on the PR. Thanks.

src/cc/BPFTable.h
mvbpolito: I'll use {...}_ and use a reference instead of a pointer.

src/cc/BPFTable.h
mvbpolito: That was actually my first idea, use only the  `TableDesc`, but I found that functions saved in that class are `llvm::Function` that cannot be directly called, instead they have to be converted to functions pointers using the `getPointerToFunction()` method of the rewriter.
 
The idea I have is to generate those pointers to function and save them in the `TableDesc`, this within the `annotate` function, then we could call those directly without needing a reference to the `bpf_module`

My questions now is, is it possible from a technical point of view?, will those pointers remain valid for the `TableDesc` lifetime?

palmtenor: I think you are right. @drzaeus77 should know much better on how to work with the rewriter:)

mvbpolito:
I tried to implement it saving the function pointers directly in the TableDesc structure [1], it appears to work good, however I'd like to get somebody's confirmation about this approach.

[1] https://github.com/mvbpolito/bcc/commit/9ab244c3bea7fb10da75ecb95c0921011caafd30

drzaeus77:
The direct function pointer save approach looks good. The primary concerns are:
1. You remove some null checks, is it guaranteed that the pointer is initialized (I think so)?
2. The function JITed memory is bound to the `llvm::Module *` it was created in, so it is important that the scope of that pointer is sufficient. In this code change, the pointer is kept in the BPFTableBase via the `TableDesc &desc`, but the `BPFModule *` is in `BPF`, so they could be destroyed separately which would result in SIGSEGVs. An alternative could be to keep the `BPFModule *` created in the `annotate()` phase along with the TableDesc. Otherwise, one has to ensure that BPFTable is only used when the BPF object is also still in scope, since that is where the `BPFModule *` is kept.

mvbpolito:
I think I finally arrived to a solution. Pointers to functions are still saved in the `TableDesc` class, in this case the destructor of the `BPFModule` class sets to null those pointers, then they can be used as long as the `BPFModule` exists, otherwise they are null.

I added a test file that shows how it behaves when the `BPFModule` is destroyed and those functions are invoked.

mauriciovasquezbernal:
@palmtenor @drzaeus77  ping?

src/cc/BPFTable.cc
```diff
@@ -31,6 +31,60 @@
...
+BPFTable::BPFTable(const TableDesc& desc)
+  : BPFTableBase<void, void>(desc) {}
+
+StatusTuple BPFTable::get_value(const std::string &key_str, std::string &value_str) {
+  char key[desc.key_size];
+  char value[desc.leaf_size];
+
+  StatusTuple r(0);
+
+  r = this->string_to_key(key_str, key);
```
drzaeus77: using `this->` isn't a typical C++ idiom, can you remove it?

src/cc/BPFTable.cc
```diff
@@ -31,6 +31,60 @@
...
+  if (r.code() != 0)
+    return r;
+
+  if(!this->lookup(key, value))
```
drzaeus77: missing space between if and (...can you try running scripts/style-check.sh to see what other style issues may be introduced here?

drzaeus77:
[buildbot, ok to test]

mauriciovasquezbernal:
@drzaeus77 I removed "this->" and addressed style-issues using the script. (I didn't know it).

PR #1107, opened 2017-04-12
Fixed SEGFAULT due to perf buffer overflow if lost_cb unset

Eichhoernchen:
This fixes a SEGFAULT in the underlying libbcc when no lost_cb is set.
This occurs when the perf buffer is full and the lost_cb is called.
libbcc expects a NULL pointer as lost_cb to trigger its internal lost message.
This fixes the bug by providing an appropriate NULL pointer when lost_cb is unset.


4ast:
ouch. thanks for the fix
[buildbot, ok to test]

Eichhoernchen:
I noticed another mistake in the code, I added a fix to the PR.

palmtenor:
Oops >_< Thanks a lot for the fix!

PR #1111, opened 2017-04-13
cc: Prefer external debuginfo files to the binary itself

goldshtn:
On some distributions, the debuglink section in the binary will
not have a .debug file extension. As a result, we will try to look
for the debuginfo file in the binary itself, immediately find it,
and abort looking for any other alternatives. This is not good,
because the binary might contain stripped or partial symbols,
which precludes certain tools from realizing their full potential.

Fix by checking that the debuginfo file we're trying to use is
not the same as the binary file. In any case, if external debuginfo
can't be found, we will fall back to the symbols in the original
binary file, if present, so this should not regress any existing
scenario.

This was originally reported by @antonblanchard in #1011.

### DIFF
--- a/src/cc/bcc_elf.c
+++ b/src/cc/bcc_elf.c
@@ -360,9 +360,12 @@ static char *find_debug_via_debuglink(Elf *e, const char *binpath) {
   bindir = strdup(binpath);
   bindir = dirname(bindir);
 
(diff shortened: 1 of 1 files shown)


markdrayton:
Looks good to me.

PR #1113, opened 2017-04-14
Fix delitem operation for PROG_ARRAYs

pchaigno:
In bcc, `ProgArray.__delitem__` defaults to `ArrayBase.__delitem__`
which uses the `bpf_update_elem` helper to clear the item (override
with a null value). However, eBPF doesn't offer a `bpf_update_elem`
helper for prog arrays.

This pull request overrides `__delitem__` in `ProgArray` to use the
`bpf_delete_item` helper.


goldshtn:
Thanks!

PR #1115, opened 2017-04-16
Execsnoop cli args matching

tehnerd:
this diff adds option to match on a line in commands arguments list. could be usefull if you want to know which command is operate on your file/package (e.g. if you want to know why who starts some service/install specific package etc).

chnages:
1) added -l option and corresponding logic
2) added documentation and example for -l option
3) changes in documentation for -n option (as it's missleading, -n checks only if command's name matches specified value, but not command's line (as someone could rely on this behavior i have decided to implement new options instead of fixing/changing -n; could rework -n if it's acceptable)
4) adding try/catch over  del(argv[event.pid]) : as on startup we sometimes could have a situation where EventType.EVENT_RET has been received w/o corresponding EventType.EVENT_ARG. and because of this "argv" dict doesnt have an entry for specified pid (as we create it only on EVENT_ARG event). example of the error:

```
  File "_ctypes/callbacks.c", line 314, in 'calling callback function'
  File "bcc/table.py", line 527, in <lambda>
  File "/tmp/execsnoop.py", line 208, in print_event
    del(argv[event.pid])
KeyError: 3365839L
```


tools/execsnoop.py
```diff
@@ -192,16 +195,21 @@ def print_event(cpu, data, size):
...
+            skip = True
 
         if not skip:
             if args.timestamp:
                 print("%-8.3f" % (time.time() - start_ts), end="")
             ppid = get_ppid(event.pid)
             print("%-16s %-6s %-6s %3s %s" % (event.comm.decode(), event.pid,
                     ppid if ppid > 0 else "?", event.retval,
-                    b' '.join(argv[event.pid]).decode()))
+                    b' '.join(argv[event.pid])))
```
tehnerd: forget decode while were merging. will fix it

4ast:
[buildbot, ok to test]
cc @brendangregg

brendangregg:
All looks good, except this option needs to be added to the man page as well (man/man8/execsnoop.8).

I usually use "-l" for long format, but I can't think of a better letter here.

goldshtn:
@tehnerd Can you please fix the man page?

tehnerd:
sure, will fix man today

brendangregg:
it's execsnoop, not opensnoop! sorry, looks like that previous line in the man page is a copy-n-paste error.

man/man8/execsnoop.8
```diff
@@ -46,9 +49,13 @@ Include failed exec()s:
 #
 .B execsnoop \-x
 .TP
-Only trace exec()s where the filename or arguments contain "mount":
+Only trace exec()s where the filename contains "mount":
 #
 .B opensnoop \-n mount
```
goldshtn: execsnoop?

man/man8/execsnoop.8
```diff
@@ -46,9 +49,13 @@ Include failed exec()s:
...
+.TP
+Only trace exec()s where argument's line contains "testpkg":
+#
+.B opensnoop \-l testpkg
```
goldshtn: execsnoop?

tehnerd: ops. will fix other lines as well

goldshtn:
Thanks! Merging.

PR #1116, opened 2017-04-16
add the directive to check the compiler whether supporting c++11

monadbobo:
use the CheckCXXCompilerFlag module of Cmake to check the compiler whether supporting c++11.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,7 @@ enable_testing()
 include(cmake/GetGitRevisionDescription.cmake)
 include(cmake/version.cmake)
 include(GNUInstallDirs)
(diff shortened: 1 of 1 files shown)


goldshtn:
[buildbot, ok to test]

goldshtn:
Question: does this improve the situation considerably, compared to just getting an error when passing the std=c++11 switch to the compiler?

monadbobo:
@goldshtn yes, beacause some compilers do not fail with a bad flag . you could read the source of [CheckCXXCompilerFlag](https://github.com/Kitware/CMake/blob/c1f97af640231ff0cb90e4a1d9a567a2ee85daa5/Modules/CheckCXXCompilerFlag.cmake#L40).

PR #1117, opened 2017-04-18
runqlat add --pidnss option

brendangregg:
For container analysis.


goldshtn:
Cool. Is this something we'd want for all the tools that take `-p`? I'm just thinking this is a good time to make a decision and implement this in a way that would make it easy for other tools, not just this one, to work well with containers.

brendangregg:
I filed a ticket to discuss, #1119. It's just not clear yet.

This tool can certainly have a --pidns=PIDNS (not done yet) and --pidnss (done here), and, other tools already have a --netns=NETNS.

PR #1121, opened 2017-04-19
Use normal inlining in clang loader

shodoco:
This is a follow up to https://github.com/iovisor/bcc/pull/1114.

In clang -O0 the default inlining mode is `OnlyAlwaysInlining`. With a recent version of clang/llvm, this might break `static inline` functions from kernel headers, e.g. `ip_is_fragment` from `net/ip.h`.
```
error: <unknown>:0:0: in function process i32 (%struct.xdp_md*): A call to global function 'ip_is_fragment' is not supported. Please use __attribute__((always_inline) to make sure this function is inlined.
```

This pull request set the inlining mode to `NormalInlining` in second pass of clang loader, so that the hinted inline functions could be properly inlined.


4ast:
lgtm. Great debugging!

PR #1130, opened 2017-04-23
softirqs: focus CPU as disector

cherusk:
With having in mind that softirq processing is happeing in
ksoftirqd/<cpu> context, which is associated with a specific cpu over
the whole dynamic life time of a system, focussing on CPus as the
disector appears more sensical.

Quite helpful is this alternative angle of view on the softirqs
processing especially for surveilling the effectiveness of net stack
tunings, as this is highly dynamic depending on the actual net stack
configuration, e.g. scaling or steering.

Signed-off-by: Matthias Tafelmeier <matthias.tafelmeier@gmx.net>

### DIFF
--- a/tools/old/softirqs.py
+++ b/tools/old/softirqs.py
@@ -33,6 +33,8 @@ parser.add_argument("-N", "--nanoseconds", action="store_true",
     help="output in nanoseconds")
 parser.add_argument("-d", "--dist", action="store_true",
     help="show distributions as histograms")
+parser.add_argument("-C", "--bycpu", action="store_true",
+    help="break down softirqs to individual cpus")
 parser.add_argument("interval", nargs="?", default=99999999,
     help="output interval, in seconds")
 parser.add_argument("count", nargs="?", default=99999999,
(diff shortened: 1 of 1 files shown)


brendangregg:
Do you think it makes sense to update tools/softirq.py to key on the CPU id as well?

Note for others: to test tools/old/softirq.py on newer kernels, I had to delete the blk_iopoll_softirq kprobe.

So the output is now:

```
# ./softirqs.py.1
warning: unknown warning option '-Wno-pragma-once-outside-header'; did you mean '-Wno-private-header'? [-Wunknown-warning-option]
1 warning generated.
^C
SOFTIRQ                            CPU TOTAL_usecs
net_tx_action                        7          69
run_rebalance_domains                3         235
run_rebalance_domains                2         248
run_rebalance_domains                4         257
run_rebalance_domains                0         271
run_rebalance_domains                6         274
run_rebalance_domains                1         342
run_rebalance_domains                7         355
run_rebalance_domains                5         366
rcu_process_callbacks                5         804
rcu_process_callbacks                3         820
rcu_process_callbacks                2         822
rcu_process_callbacks                0         828
rcu_process_callbacks                1         912
rcu_process_callbacks                6         930
rcu_process_callbacks                4        1319
rcu_process_callbacks                7        1404
net_rx_action                        7        1451
run_timer_softirq                    6        2536
run_timer_softirq                    2        2554
run_timer_softirq                    3        2599
run_timer_softirq                    0        2642
run_timer_softirq                    1        2893
run_timer_softirq                    4        2901
run_timer_softirq                    7        2951
run_timer_softirq                    5        3269
```

I'd make this per-CPU breakdown an option, like -C. I do want the per-softirq summaries by default, especially as my larger systems have 64 CPUs and the output is going to get long.

An example of an optional breakdown is biolatency's -D.

I haven't encountered the word "dissector" for this feature, we usually use "breakdown", but I can see that dissector is more specific, so I think I like it. :)

cherusk:
> Do you think it makes sense to update tools/softirq.py to key on the
> CPU id as well?
>
Theopractically, it'd. Was a little selfish, since I was running on a
kernel < 4.7. Wanted to sound opinions first.

> ||
>
> I'd make this per-CPU breakdown an option, like -C.
>
Right, no objections.

> I do want the per-softirq summaries by default, especially as my
> larger systems have 64 CPUs and the output is going to get long.
>
Am handling equally equipped systems. On top, well, it's rather meant as
a "grepable" backend mechanism. Thought of NUMA aware agglomeration for
manual wielding. However, we can leave the status quo in place.
>
> [...] "breakdown", but I can see that dissector is more specific, so I
> think I like it. :)
>
No preferences, merely was the first word appearing to when commencing
on it.

-- 
Besten Gruß

Matthias Tafelmeier

cherusk:
Right, seems legit now ...

tools/old/softirqs.py
```diff
@@ -33,6 +33,8 @@
     help="output in nanoseconds")
 parser.add_argument("-d", "--dist", action="store_true",
     help="show distributions as histograms")
+parser.add_argument("-C", "--CPUidx", action="store_true",
```
goldshtn: I suggest a somewhat clearer long option name, perhaps `--by-cpu`?

tools/old/softirqs.py
```diff
@@ -48,58 +50,107 @@
...
-    u64 *tsp, delta, ip, *ipp;
-    u32 pid = bpf_get_current_pid_tgid();
-
-    // fetch timestamp and calculate delta
-    tsp = start.lookup(&pid);
-    ipp = iptr.lookup(&pid);
-    if (tsp == 0 || ipp == 0) {
-        return 0;   // missed start
+bpf_text = ""
+if args.CPUidx:
```
goldshtn: There is a lot of shared code across the two alternatives, which makes me worried about duplication and potential maintenance problems if we add more "dissectors". What we typically do in other tools is one of two options:

1) Embed string markers (like `REPLACEME`) in the C code string and then conditionally replace them with the appropriate code, kind of like you did with `COMMON` here but in the other direction
1) Use conditional compilation (`#ifdef ...`) in the C code and conditionally prepend the relevant macro definition

tools/old/softirqs.py
```diff
@@ -115,11 +166,18 @@
         "rcu_process_callbacks", "run_rebalance_domains", "tasklet_action",
         "tasklet_hi_action", "run_timer_softirq", "net_tx_action",
         "net_rx_action"):
-    b.attach_kprobe(event=softirqfunc, fn_name="trace_start")
-    b.attach_kretprobe(event=softirqfunc, fn_name="trace_completion")
+    if args.CPUidx:
```
goldshtn: Again, there's no need for this duplication if we give the C functions the same names in both cases. I don't know why not.

goldshtn:
[buildbot, add to whitelist]

cherusk:
> I have a couple of comments below, and also a bigger question: why
> does this PR only update tools/old/softirqs.py, and not the new tool
> that uses tracepoints? That's the one most people will be using by
> default (on 4.7+ anyway).
>
Answered that further up.

-- 
Besten Gruß

Matthias Tafelmeier

tools/old/softirqs.py
cherusk: No preferences, can do so ...

tools/old/softirqs.py
cherusk: Well, technically, I agree, though, I was refraining from making the replacement section a mess. So I had a tradeoff between readability and code duplication in mind.

brendangregg:
```
# ./softirqs.py 
  File "./softirqs.py", line 54, in <module>
    if args.by-cpu:
AttributeError: 'Namespace' object has no attribute 'by'
```

Seems that my versions of argparse doesn't like that. I found s/by-cpu/bycpu/ fixed it. (How come our test system didn't pick it up?).

-C also doesn't work with -d:

```
# softirqs.py -C -d
warning: unknown warning option '-Wno-pragma-once-outside-header'; did you mean '-Wno-private-header'? [-Wunknown-warning-option]
1 warning generated.
^C
  File "./softirqs.py", line 197, in <module>
    dist.print_log2_hist(label, "softirq", section_print_fn=distr_sec_fn)
  File "/usr/lib/python2.7/dist-packages/bcc/table.py", line 311, in print_log2_hist
    vals[slot] = v.value
IndexError: cannot fit 'long' into an index-sized integer
```

I filed #1146 for updating tools/softirqs.

goldshtn:
@brendangregg Our smoke tests only run the tools in tools/, not in tools/old/. 

I generally think of the tools in tools/old/ as unmaintained.

cherusk:
Amended things, also the general Softirq per CPU breakdown is working now, though, I don't deem this as very helpful. It might need a nested breakdown per CPU-softirq on top - feel free to add that later. Cheers!

4ast:

cherusk:
It is ... pls, reopen in order not to splinter the context. I've not procrastinated it.

cherusk:
I don't think so., seemingly, it hasn't been touched on master ever since.

yonghong-song:
@cherusk This still only changes tools/old/softirqs.py. Did any situation change in your side so you could make a change in tools/softirqs.py instead? As mentioned in the above, tools/old/... are considered unmaintained and hence will not be tested in bcc testing framework and will not be used by most people.

cherusk:
@yonghong-song I am prepared to align the tools/softirqs.py to the 'old' symmetrically. Though, landscapes on older kernels (enough out there for stability) will use this one, so please merge it in.

yonghong-song:
@4ast could you merge since you requested the change? Note that the code is using kprobe and it will not work on 4.14 kernel as some functions are already gone.

PR #1136, opened 2017-04-27
Remove unused symnames_ from ProcSyms

palmtenor:
It doesn't seem to be used anywhere, just taking memory space and makes things more complicated...


palmtenor:
Why didn't I look at blame history:(
So it kind of makes sense. But let me do some more testing

markdrayton:
Haha, I think this is the second time someone proposed removing this (the other was https://github.com/iovisor/bcc/pull/941). We probably need some better comments and/or a benchmark.

PR #1137, opened 2017-04-27
Only load non-zero and function symbols in ProcSyms

palmtenor:
`ProcSyms` is currently loading everything, which is slow and inefficient.
Tested on some common libraries.

Originally:
```
(trunk)                          Loaded <PATH>/libpthread-2.23.so   Entries: 2163    Unique names: 1066
(trunk)                          Loaded <PATH>/libstdc++.so.6.0.21  Entries: 18394   Unique names: 6553
(trunk)                          Loaded <PATH>/libc-2.23.so         Entries: 15980   Unique names: 7431
(trunk)                          Loaded <PATH>/hhvm                 Entries: 394265  Unique names: 188241
```
Not loading zero symbols:
```
(trunk)(no_zero_addr)            Loaded <PATH>/libpthread-2.23.so   Entries: 1676    Unique names: 709
(trunk)(no_zero_addr)            Loaded <PATH>/libstdc++.so.6.0.21  Entries: 17724   Unique names: 6113
(trunk)(no_zero_addr)            Loaded <PATH>/libc-2.23.so         Entries: 14495   Unique names: 6119
(trunk)(no_zero_addr)            Loaded <PATH>/hhvm                 Entries: 383567  Unique names: 181845
```
Further only load functions:
```
(trunk)(no_zero_addr)(only_func) Loaded <PATH>/libpthread-2.23.so   Entries: 1310    Unique names: 555
(trunk)(no_zero_addr)(only_func) Loaded <PATH>/libstdc++.so.6.0.21  Entries: 12682   Unique names: 4362
(trunk)(no_zero_addr)(only_func) Loaded <PATH>/libc-2.23.so         Entries: 11652   Unique names: 4985
(trunk)(no_zero_addr)(only_func) Loaded <PATH>/hhvm                 Entries: 260369  Unique names: 129575
```

### DIFF
--- a/src/cc/bcc_perf_map.c
+++ b/src/cc/bcc_perf_map.c
@@ -98,7 +98,7 @@ int bcc_perf_map_foreach_sym(const char *path, bcc_perf_map_symcb callback,
     if (newline)
         newline[0] = '\0';
 
-    callback(cursor, begin, len, len > 0 ? STT_FUNC : STT_OBJECT, payload);
+    callback(cursor, begin, len, STT_FUNC, payload);
   }
 
   free(line);
--- a/src/cc/bcc_perf_map.c
+++ b/src/cc/bcc_perf_map.c
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 #include <ctype.h>
+#include <linux/elf.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -97,7 +98,7 @@ int bcc_perf_map_foreach_sym(const char *path, bcc_perf_map_symcb callback,
     if (newline)
         newline[0] = '\0';
 
-    callback(cursor, begin, len, 0, payload);
+    callback(cursor, begin, len, len > 0 ? STT_FUNC : STT_OBJECT, payload);
   }
 
   free(line);
--- a/src/cc/bcc_syms.cc
+++ b/src/cc/bcc_syms.cc
@@ -15,6 +15,7 @@
  */
 
 #include <cxxabi.h>
+#include <linux/elf.h>
 #include <string.h>
 #include <sys/stat.h>
 #include <sys/types.h>
@@ -27,6 +28,8 @@
 
 #include "syms.h"
 #include "vendor/tinyformat.hpp"
+
+#define ELF_TYPE_IS_FUNCTION(flags) (((flags)&0xf) == STT_FUNC)
 
 ino_t ProcStat::getinode_() {
   struct stat s;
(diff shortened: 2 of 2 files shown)


palmtenor:
Test need to be fixed...

goldshtn:
Just to make sure, we don't need symbols for global variables for the USDT resolver? I think you can have USDT arguments addressed from globals.

palmtenor:
@goldshtn  Aren't USDT addresses come from note section, i.e. `do_note_segment` and `listprobes` in `bcc_elf.c`? Plus, this only changes the `ProcSyms::Module::_add_symbol` callback.

goldshtn:
No, see here: https://github.com/iovisor/bcc/blob/master/src/cc/usdt_args.cc

We use bcc_find_global_addr.

goldshtn:
Symbol addr. Argh. On my phone right now :)

goldshtn:
... and that seems to ignore the ProcSymbols and just iterate over the ELF again. Which is also a waste of time.

palmtenor:
@goldshtn You are right. I just have never seen a Tracepoint format like `-4@CheckpointStats+40(%rip)` so totally unaware of such. I think I can fix that part once I have more understanding what is it doing exactly. I think I can also make it break once it found the symbols, unlike in `ProcSyms` it always loads everything.

palmtenor:
@goldshtn Note that the USDT argument thing is probably broken for a long time. See [here](https://github.com/iovisor/bcc/blob/master/src/cc/usdt_args.cc#L43): we (I) added new field to `bcc_symbol` in #638

src/cc/bcc_perf_map.c
```diff
@@ -97,7 +98,7 @@ int bcc_perf_map_foreach_sym(const char *path, bcc_perf_map_symcb callback,
     if (newline)
         newline[0] = '\0';
 
-    callback(cursor, begin, len, 0, payload);
+    callback(cursor, begin, len, len > 0 ? STT_FUNC : STT_OBJECT, payload);
```
palmtenor: @markdrayton Any idea what's the correct logic here?

markdrayton: From a quick glance I'm not sure. I'll take a better look tomorrow morning.

palmtenor:
Talked with @markdrayton offline, all `perf-PID.map` entries should be viewed as possible functions.

goldshtn:
So, where are we going from here? @palmtenor

palmtenor:
@goldshtn I am planning to have another PR for USDT side because I also want to make some other small changes, then we can come back and land this

PR #1139, opened 2017-04-27
python helpers for common string treatment

drzaeus77:
This is the minimal code worked on so far to address usage of unicode strings: namely, don't call decode() on un-sanitized data. The first commit adds helpers for encoding/decoding command line arguments, implying we should always be dealing with bytes() objects. The second commit removes decode() of data that is received from the perf buffer, favoring instead an encode (into bytes) of the original command line arguments.

### DIFF
--- a/tools/execsnoop.py
+++ b/tools/execsnoop.py
@@ -18,6 +18,8 @@
 
 from __future__ import print_function
 from bcc import BPF
+from bcc.utils import ArgString, printb
+import bcc.utils as utils
 import argparse
 import ctypes as ct
 import re
@@ -41,8 +43,10 @@ parser.add_argument("-t", "--timestamp", action="store_true",
 parser.add_argument("-x", "--fails", action="store_true",
     help="include failed exec()s")
 parser.add_argument("-n", "--name",
+    type=ArgString,
     help="only print commands matching this name (regex), any arg")
 parser.add_argument("-l", "--line",
+    type=ArgString,
     help="only print commands where arg contains this line (regex)")
 args = parser.parse_args()
 
@@ -193,19 +197,19 @@ def print_event(cpu, data, size):
     elif event.type == EventType.EVENT_RET:
         if args.fails and event.retval == 0:
             skip = True
-        if args.name and not re.search(args.name, event.comm):
+        if args.name and not re.search(bytes(args.name), event.comm):
             skip = True
-        if args.line and not re.search(args.line,
-                                       b' '.join(argv[event.pid]).decode()):
+        if args.line and not re.search(bytes(args.line),
+                                       b' '.join(argv[event.pid])):
             skip = True
 
         if not skip:
             if args.timestamp:
                 print("%-8.3f" % (time.time() - start_ts), end="")
             ppid = get_ppid(event.pid)
-            print("%-16s %-6s %-6s %3s %s" % (event.comm.decode(), event.pid,
-                    ppid if ppid > 0 else "?", event.retval,
-                    b' '.join(argv[event.pid]).decode()))
+            ppid = b"%d" % ppid if ppid > 0 else b"?"
+            printb(b"%-16s %-6d %-6s %3d %s" % (event.comm, event.pid,
+                   ppid, event.retval, b' '.join(argv[event.pid])))
         try:
             del(argv[event.pid])
         except Exception:
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -27,7 +27,7 @@ basestring = (unicode if sys.version_info[0] < 3 else str)
 from .libbcc import lib, _CB_TYPE, bcc_symbol, bcc_symbol_option, _SYM_CB_TYPE
 from .table import Table
 from .perf import Perf
-from .utils import get_online_cpus
+from .utils import get_online_cpus, printb, _assert_is_bytes, ArgString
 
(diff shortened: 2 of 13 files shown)


4ast:
@brendangregg pls review

tools/execsnoop.py
```diff
@@ -193,19 +197,19 @@ def print_event(cpu, data, size):
...
         if not skip:
             if args.timestamp:
                 print("%-8.3f" % (time.time() - start_ts), end="")
             ppid = get_ppid(event.pid)
-            print("%-16s %-6s %-6s %3s %s" % (event.comm.decode(), event.pid,
-                    ppid if ppid > 0 else "?", event.retval,
-                    b' '.join(argv[event.pid]).decode()))
+            ppid = b"%d" % ppid if ppid > 0 else b"?"
+            printb(b"%-16s %-6d %-6s %3d %s" % (event.comm, event.pid,
+                   ppid, event.retval, b' '.join(argv[event.pid])))
```
brendangregg: Is there a way to do this without changing print() to printb()? I can get my head around kernel strings needing b"", but this is affecting non-kernel strings too (like printing an int, or b"?"). Something like this would seem easier:

```Python
            print("%-16s %-6d %-6s %3d %s" % (b(event.comm), event.pid,
                   ppid, event.retval, b' '.join(argv[event.pid])))
```

We only special case the kernel strings.

Plus, since these are pulled in via:

```Python
 class Data(ct.Structure):
     _fields_ = [
         ("pid", ct.c_uint),
         ("comm", ct.c_char * TASK_COMM_LEN),
         ("type", ct.c_int),
         ("argv", ct.c_char * ARGSIZE),
         ("retval", ct.c_int),
]
```

It makes me wonder if we can deal with it there, and leave the entire print() call unchanged. eg, something like:

```Python
 class Data(ct.Structure):
     _fields_ = [
         ("pid", ct.c_uint),
         ("comm", b.k_char * TASK_COMM_LEN),
         ("type", ct.c_int),
         ("argv", b.k_char * ARGSIZE),
         ("retval", ct.c_int),
]
```

b.k_char == kernel character.

drzaeus77: The simplest is just to leave it as `print("%s", event.comm)` which will give you `b'/usr/bin/bash'` or similar output (if run from python3).

For the other suggestions, the issue remains that in order to convert the byte string to a unicode string, some encoding must be chosen. Simply wrapping these in utf-8 can still fail (e.g. try `b"\xff".decode("utf8")`), so I don't see that path working out.

drzaeus77:
I've gone through `__init__.py`to remove all of the encode/decode usages in a way that we can retain support for python3 (I hope). The idea is hopefully self evident enough in the code, but the plan would be to:

1. Call out all arguments in the api that should take a bytes(), and enforce those checks with _assert_is_bytes(). For now, coerce but don't warn/error.
   1a. fix breakages
2. Turn on warnings in that assertion helper, and go through and clean up all the tools/tests.
3. Turn on assertions

So far, this code is at step 1, 1a is not done yet (and may need others to help test on their respective environments). I'd like feedback despite this incompleteness. We will also need to go and find in the tools any place where decode is being used improperly, or where python2 is being assumed. The answer for such cases should be to use b"" syntax.

/cc @torgil

4ast:

torgil:
Do we have a plan here?  I'll have to address this again sooner or later. For me python might be a dead end due to numerous issues, also it's not very multithreading- and embedded-friendly. I've started looking into the C++ classes and it strikes me that we have a Python and a C++ implementation of roughly the same thing so the obvious solution is to move also the higher level code to a C API (implementation can still be in C++) that could be used from either C / Python / C++ or any language with an C/API bridge.

This way we get a clean flexible solution and the decode/encode madness is removed as the code is moved down to C level. Performance will improve as well. Any thoughts?

yonghong-song:
We do have core C library libbpf.so to have core functions in libbpf.c and perf_reader.c. Its interface is not the same as C++, but close and have more. So instead of designing a new C API, people uses C can use this libbpf.so and the interface exposed by it.

yonghong-song:
cc @drzaeus77 @brendangregg for common string issue.

torgil:
So you suggesting moving current Python interface down to c?  I'm well aware of the libbpf.so (last time I pulled a few month ago this was also used from Python through a wrapper). This way there would also be less pain for people (like me) wanting to migrate from python to a more efficient platform.

I was also playing around refactoring c++ interface to get more features in. C++ as an API is a bit of pain as the ABI is varying compiler to compiler and there is no common way to wrap these in a higher level language, this might also have changed lately though as there have been years since last time I made an effort looking t this.

yonghong-song:
> So you suggesting moving current Python interface down to c? 
No, I am not suggesting this. I means we have different APIs for C compared to C++/Python and it should work in most cases. C++ have nice properties and may make code easy to maintain, and LLVM headers are all C++, so I guess for many codes, we will have to use C++.

torgil:
That's correct. My point is to reach the following scenario
* One lean high-level code-base (today is a two separate implementations of the same in C++ and Python where Python has more features but is inefficient and encode/decode-broken. Features added benefits many languages.
* Easy "cdecl"-entry to high-level code from other languages than Python (instead of writing yet another implementation). May just be a thin wrapper-layer to C++ - classes. May be auto-generated etc.

I can reach the second without the first but then I may need to track features going into Python side and re-implement.

PR #1140, opened 2017-04-27
frontend/b: fixes for LLVM 4.0 API change

drzaeus77:
Upstream LLVM added a new parameter to the AllocaInst constructor, which
breaks compilation. Fix it by using the IRBuilder method.


drzaeus77:
[buildbot, test this please]

PR #1141, opened 2017-04-28
RFC: Improve mount namespace handling

palmtenor:
This diff:
- Adds `ProcMountNS` and `ProcMountNSGuard`, that automatically opens and closes FDs, and enters and exists mount namespace on construction and destruction. So we won't make the mistake for forget to exit / clean-up.
- Move mount namespace checking to only done once in `ProcSyms` construction (by constructing a `ProcMountNS` instance), instead of doing it for every `Module` during the iteration.

This gives us a nicer and cleaner code, also much easier to re-use such as for USDT symbol lookup.


src/cc/CMakeLists.txt
```diff
@@ -39,8 +39,8 @@ add_library(bcc-shared SHARED bpf_common.cc bpf_module.cc libbpf.c perf_reader.c
 set_target_properties(bcc-shared PROPERTIES VERSION ${REVISION_LAST} SOVERSION 0)
 set_target_properties(bcc-shared PROPERTIES OUTPUT_NAME bcc)
 
-add_library(bcc-loader-static STATIC libbpf.c perf_reader.c bcc_elf.c bcc_perf_map.c bcc_proc.c)
-add_library(bcc-static STATIC bpf_common.cc bpf_module.cc shared_table.cc bpffs_table.cc json_map_decl_visitor.cc table_storage.cc exported_files.cc bcc_syms.cc usdt_args.cc usdt.cc common.cc BPF.cc BPFTable.cc)
+add_library(bcc-loader-static STATIC libbpf.c perf_reader.c bcc_elf.c bcc_perf_map.c bcc_proc.c bcc_syms.cc)
```
palmtenor: @drzaeus77: I looked back to #426 and saw that you said we have to have this separated target because `bpf_map_create` is shared. Can we just make the `clang_frontend` target not `STATIC` hence it could be linked later (like in line 66)? I'm quite a n00b on this so I'm not entirely sure if other bad thing would happen if I do that......

src/cc/bcc_syms.cc
```diff
@@ -13,8 +13,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#ifndef _GNU_SOURCE
```
palmtenor: @4ast : Do we need this? Which part is GNU source (I moved this from `bcc_proc.c`)

4ast: don't remember. some glibc function is only available if this is defined.
we can probably drop it here.

markdrayton:
I will look at this this morning.

markdrayton:
Looks good to me. A minor nit: most of the code puts `*` next to the type, not the variable name.

palmtenor:
That's actually `clang-format`'s decision. I can either change the `.clang-format` file, for format the entire file to make it consistent.

palmtenor:
Fixed formatting. I was hoping `clang-format` would do that for me...

markdrayton:
Doh. We should fix the clang-format config to do this automatically.

src/cc/bcc_syms.cc
```diff
@@ -92,51 +95,112 @@ bool KSyms::resolve_name(const char *_unused, const char *name,
   return true;
 }
 
-ProcSyms::ProcSyms(int pid) : pid_(pid), procstat_(pid) { load_modules(); }
+ProcMountNS::ProcMountNS(int pid) : self_fd_(-1), target_fd_(-1) {
+  if (pid < 0)
+    return;
+
+  int self_fd = -1;
```
drzaeus77: Can I suggest using the FileDesc class from src/cc/table_desc.h? You would have a local `FileDesc self_fd`, that at the end of the function gets `self_fd_ = std::move(self_fd)`.

src/cc/bpf_module.cc
```diff
@@ -329,7 +329,7 @@ unique_ptr<ExecutionEngine> BPFModule::finalize_rw(unique_ptr<Module> m) {
 
 // load an entire c file as a module
 int BPFModule::load_cfile(const string &file, bool in_memory, const char *cflags[], int ncflags) {
-  clang_loader_ = make_unique<ClangLoader>(&*ctx_, flags_);
+  clang_loader_ = ebpf::make_unique<ClangLoader>(&*ctx_, flags_);
```
palmtenor: This is because in (new) LLVM, there is also a `make_unique` and we `using namespace llvm` in this file:((

src/cc/common.cc
```diff
@@ -47,5 +49,43 @@ std::vector<int> get_possible_cpus() {
...
+
+FileDesc &FileDesc::operator=(int fd) {
+  if (fd_ >= 0)
+    ::close(fd_);
+  fd_ = fd;
+  return *this;
+}
+
+FileDesc &FileDesc::operator=(FileDesc &&that) {
+  if (fd_ >= 0)
```
palmtenor: Added cleanup in move constructor and `int` assign operator since this class should be more generic now.

src/cc/table_desc.h
```diff
@@ -93,10 +81,10 @@ class TableDesc {
         leaf_snprintf(nullptr),
         is_shared(false),
         is_extern(false) {}
-  TableDesc(const std::string &name, FileDesc &&fd, int type, size_t key_size,
+  TableDesc(const std::string &name, int fd, int type, size_t key_size,
```
drzaeus77: This looks wrong to me, and was explicitly done this way initially. In particular, I don't want any possibility of invoking `FileDesc::operator int()` implicitly anywhere that a FileDesc must instead be move()'ed into the TableDesc. Such possibility would lead to double-close() of the underlying fd.

I don't have a problem with moving the class to common.h, but the `friend` and original implementation should remain.

src/cc/common.cc
drzaeus77: Yes, good catch.

src/cc/bpf_module.cc
drzaeus77: Why are you addressing this now, compile error or generic cleanup? I also have 4.0+ llvm and either make_unique implementation should be fine.

palmtenor: Because the usage of `common.h` (which also has our version of `make_unique`) in more files lead to compiler complain about duplicated overload etc.

drzaeus77: Fair enough.

src/cc/table_desc.h
palmtenor: I didn't want to add a `TableDesc` as `friend class` to this generic common helper. Actually, the only [call site](https://github.com/iovisor/bcc/blob/master/src/cc/frontends/b/codegen_llvm.cc#L1245) of the constructor I can found passes `int` value on this parameter.

But you are right on potential double-closing. How about we make this still take a `FileDesc &&`, but explicitly construct a `FileDesc` instance to pass to constructor in the call site?

drzaeus77: That call site implicitly invoked the FileDesc(int) constructor originally. Explicitly adding FileDesc(fd) won't help the compiler after you restore the `FileDesc &&` version.

drzaeus77:
```
9: Traceback (most recent call last):
9:   File "/home/fedora/jenkins/workspace/bcc-pr/label/fc24/tests/python/test_trace1.py", line 30, in setUp
9:     b.attach_kprobe(event="sys_write", fn_name="sys_wr", pid=0, cpu=-1)
9:   File "/home/fedora/jenkins/workspace/bcc-pr/label/fc24/src/python/bcc/__init__.py", line 513, in attach_kprobe
9:     fn = self.load_func(fn_name, BPF.KPROBE)
9:   File "/home/fedora/jenkins/workspace/bcc-pr/label/fc24/src/python/bcc/__init__.py", line 344, in load_func
9:     (func_name, errstr))
9: Exception: Failed to load BPF program sys_wr: Bad file descriptor
9: 
9: ----------------------------------------------------------------------
9: Ran 1 test in 0.088s
9: 
9: FAILED (errors=1)
9: Failed
 9/32 Test  #9: py_test_trace1 ...................***Failed    0.16 sec
```

I've already debugged all of these errors. That is how I arrived at the original implementation.

src/cc/table_desc.h
palmtenor: @drzaeus77 Sorry I'm a bit confused now. My main goal is to mark the int constructor as `explict` to avoid this kind of implicit conversions. It doesn't seem that we actually want to duplicate the fd here, so what have I missed? I construct the temporary object in the call site with FD, and move it to the `TableDesc` local member. That seems to be exactly the behavior as your current implementation.

I also can't reproduce the test failure locally. It seems to be only happening in fedora hosts? Is that compiler related?

Sorry for all the trouble and thanks a lot for your patience!

palmtenor:
@drzaeus77 Sorry I'm a bit confused now. My main goal is to mark the `int` constructor as `explict` to avoid this kind of implicit conversions. I manually implemented the `TableDesc`'s copy constructor so that I can mark the `FileDesc`'s copy construct / assign and `delete` in addition to just private.

It doesn't seem that we actually want to duplicate the fd for this call site, so what have I missed? The temporary object constructed in the call site with FD, and moved to `TableDesc` local member. That seems to be exactly the behavior of current implementation?

I also can't reproduce the test failure locally. It seems to be only happening in fedora hosts? Is that compiler related?

Sorry for all the trouble and thanks a lot for your patience!

src/cc/common.cc
palmtenor: The issue causing the test failure is actually this added cleanup. We call this Move Assign Operator in the Move Constructor, where the `fd_` value may not be initialized and we could incorrectly close a random fd here. Correctly initialize the pre-move`fd_` to -1 in the Move Constructor fixes the issue.

4ast: nice debugging!

palmtenor:
Friendly ping:)

kmjohansen:
Sorry for being late to the review, but this looks good.  Thanks for cleaning entry/exit handling up to make it more automatic and easier to understand.

PR #1143, opened 2017-04-28
bfp, xdp: make type explicit

baloo:
explicit the type of the nlattr


drzaeus77:
[buildbot, ok to test]

PR #1145, opened 2017-04-30
Install sudo inside the debian build container

skottler:
The jessie image used for the build container does not include sudo, but `scripts/build-deb.sh` [requires it](https://github.com/iovisor/bcc/blob/master/scripts/build-deb.sh#L32:L33). This installs the sudo package so that containerized package builds work as expected on jessie.


drzaeus77:
[buildbot, ok to test]

PR #1150, opened 2017-05-03
Avoid walking all symbols for unknown address

palmtenor:
#1002 is an awesome fix for nested functions. However it causes that for unknown symbols, we will bascially go through the entire list of symbols, which could be bad for large binaries / perf maps.
This change fixes the issue.

cc @markdrayton @goldshtn


goldshtn:
[buildbot, add to whitelist]

goldshtn:
LGTM; @markdrayton what do you think?

markdrayton:
Nice catch. LGTM!

danobi:
Perhaps a bit late to the party, but will this work if a function has multiple nested functions inside? `limit > it->start + it->size` doesn't seem to work if we look backwards and see another nested function that ends before `limit`.

Issue #1133, opened 2017-04-26
Verifier failure on reading ctx


palmtenor:
One of my `trace.py` call failed with an USDT tracepoint because of verifier issues.
And the verifier failed on:
```
bpf: Permission denied
0: (bf) r6 = r1
1: (85) call bpf_get_current_pid_tgid#14
2: (bf) r7 = r0
3: (bf) r8 = r7
4: (77) r8 >>= 32
5: (15) if r8 == 0xdb365 goto pc+46
 R0=inv R6=ctx R7=inv R8=inv32 R10=fp
6: (b7) r1 = 0
7: (7b) *(u64 *)(r10 -8) = r1
8: (7b) *(u64 *)(r10 -16) = r1
9: (7b) *(u64 *)(r10 -24) = r1
10: (7b) *(u64 *)(r10 -32) = r1
11: (7b) *(u64 *)(r10 -40) = r1
12: (7b) *(u64 *)(r10 -48) = r1
13: (7b) *(u64 *)(r10 -56) = r1
14: (7b) *(u64 *)(r10 -64) = r1
15: (7b) *(u64 *)(r10 -72) = r1
16: (7b) *(u64 *)(r10 -80) = r1
17: (7b) *(u64 *)(r10 -88) = r1
18: (7b) *(u64 *)(r10 -96) = r1
19: (85) call bpf_ktime_get_ns#5
20: (7b) *(u64 *)(r10 -112) = r0
21: (63) *(u32 *)(r10 -104) = r8
22: (63) *(u32 *)(r10 -100) = r7
23: (bf) r1 = r10
24: (07) r1 += -96
25: (b7) r2 = 16
26: (85) call bpf_get_current_comm#16
27: (79) r1 = *(u64 *)(r6 +128)
28: (15) if r1 == 0x111be2b goto pc+5
 R0=inv R1=inv R6=ctx R7=inv R8=inv32 R10=fp
29: (15) if r1 == 0x111bd1a goto pc+1
 R0=inv R1=inv R6=ctx R7=inv R8=inv32 R10=fp
30: (55) if r1 != 0x111bbfd goto pc+11
 R0=inv R1=inv,min_value=17939453,max_value=17939453 R6=ctx R7=inv R8=inv32 R10=fp
31: (bf) r1 = r6
32: (07) r1 += 96
33: (05) goto pc+2
36: (79) r3 = *(u64 *)(r1 +0)
R1 invalid mem access 'inv'
```
[...]

Issue #829, opened 2016-11-30
nodejs_http_server.py fails to load do_trace


affansyed:
@brendangregg your example for USDT tracing of a node application is failing to load the ebpf program. 

It seems to be related to bug here (https://github.com/iovisor/bcc/issues/751), but really cant figure out how to resolve. 

Using the nightly build for bcc from  Nov 28th, 2016. 

commenting out this line allows the ebpf code to be hooked. 

//bpf_probe_read(&path, sizeof(path), (void *)addr);

Affan
`$ sudo ./nodejs_http_server.py 6615
bpf: Permission denied
0: (79) r3 = *(u64 *)(r1 +32)
1: (b7) r7 = 0
2: (7b) *(u64 *)(r10 -8) = r7
3: (07) r3 += -4312
4: (bf) r1 = r10
5: (07) r1 += -8
6: (b7) r2 = 8
7: (85) call 4
8: (79) r3 = *(u64 *)(r10 -8)
9: (bf) r6 = r10
10: (07) r6 += -144
11: (bf) r1 = r6
12: (b7) r2 = 128
13: (85) call 4
invalid indirect read from stack off -144+0 size 128

Traceback (most recent call last):
  File "./nodejs_http_server.py", line 44, in <module>
    b = BPF(text=bpf_text, usdt_contexts=[u])
  File "/usr/local/lib/python2.7/dist-packages/bcc/__init__.py", line 243, in __init__
    usdt_context.attach_uprobes(self)
  File "/usr/local/lib/python2.7/dist-packages/bcc/usdt.py", line 156, in attach_uprobes
    addr=addr, pid=pid)
  File "/usr/local/lib/python2.7/dist-packages/bcc/__init__.py", line 758, in attach_uprobe
    fn = self.load_func(fn_name, BPF.KPROBE)
  File "/usr/local/lib/python2.7/dist-packages/bcc/__init__.py", line 290, in load_func
    raise Exception("Failed to load BPF program %s: %s" % (func_name, errstr))
Exception: Failed to load BPF program do_trace: Permission denied`

goldshtn:
I'm assuming this is fixed in #1151.

PR #1151, opened 2017-05-03
cc: Work around verifier error when reading USDT probe arguments

### LINKED ISSUE #751: BPF program load error
I think I found a bug. Can't load a very simple BPF program, getting a verifier error.

Repro:

``` python
# This is error.py
from bcc import BPF                                                             

 prog = """                                                                      
 #include <linux/ptrace.h>                                                       

 static inline int dummy(struct pt_regs *ctx, void *dest, size_t len) {          
   switch(ctx->ip) {                                                             
   case 0x55e29a4ce2a4ULL: *((uint64_t *)dest) = (uint64_t)ctx->si; return 0;    
   case 0x55e29a4ce7abULL: *((uint64_t *)dest) = (uint64_t)ctx->bp; return 0;    
   }                                                                             
   return -1;                                                                    
 }                                                                               

 int do_start(struct pt_regs *ctx)                                               
 {                                                                               
         u64 arg1 = 0;                                                           
         dummy(ctx, &arg1, sizeof(arg1));                                        
         if (arg1 != 0) {                                                        
             bpf_trace_printk("Hi\\n");                                          
         }                                                                       

         return 0;                                                               
 }                                                                                                                                                         
 """                                                                             

 BPF(text=prog).load_func("do_start", BPF.KPROBE)                                
```

Results:

```
# uname -r
4.8.0-1.vanilla.knurd.1.fc23.x86_64

# python error.py
bpf: Permission denied
0: (79) r2 = *(u64 *)(r1 +128)
1: (18) r3 = 0x9a4ce7ab
3: (1d) if r2 == r3 goto pc+5
 R1=ctx R2=inv R3=inv R10=fp
4: (18) r3 = 0x9a4ce2a4
6: (5d) if r2 != r3 goto pc+11
 R1=ctx R2=inv R3=inv R10=fp
7: (07) r1 += 104
8: (05) goto pc+1
10: (79) r1 = *(u64 *)(r1 +0)
R1 invalid mem access 'inv'

  File "/home/vagrant/error.py", line 27, in <module>
    BPF(text=prog).load_func("do_start", BPF.KPROBE)
  File "/usr/lib/python2.7/site-packages/bcc/__init__.py", line 255, in load_func
    raise Exception("Failed to load BPF program %s: %s" % (func_name, errstr))
Exception: Failed to load BPF program do_start: Permission denied
```

Adding a bit more code so that `arg1` isn't the first local being used makes the error go away. Removing one of the cases makes the error go away. 

/cc @4ast @drzaeus77 @brendangregg

goldshtn:
The code generated by the USDT helpers for reading probe arguments
occasionally caused a verifier error due to a construct similar to
the following:

```c
switch (ctx->ip) {
case 0xaaaa: *dest = ctx->cx; break;
case 0xbbbb: *dest = ctx->dx; break;
}
```

This would generate an instruction sequence that attempts to access
ctx through an offset that is not statically known, which confuses
the verifier. This was reported in #751, #829, and #1133, and likely
seen by others as well.

The workaround, suggested by @yonghong-song, is to force memory
writes by using the volatile modifier, which precludes this specific
optimization.

Tested by running`trace -p 1234 'u::mutex__acquired (arg1!=0) "%llx", arg1'`
with and without the patch. Without the patch, on my system, the probe
has four locations and arg1 is in bx, r8, bx, r8 respectively, so the generated
code hits the bug. With the patch, everything works OK. Testing this
automatically would require us to have control not only over exactly which
probes are present in the target binary, but also which registers are used
by the compiler for the arguments.

Resolves #751, #829, #1133.

### DIFF
--- a/src/cc/usdt.cc
+++ b/src/cc/usdt.cc
@@ -152,10 +152,10 @@ bool Probe::usdt_getarg(std::ostream &stream) {
 
   for (size_t arg_n = 0; arg_n < arg_count; ++arg_n) {
     std::string ctype = largest_arg_type(arg_n);
-    std::string cptr = tfm::format("*((%s *)dest)", ctype);
+    std::string cptr = tfm::format("*((volatile %s *)dest)", ctype);
 
     tfm::format(stream,
-                "static inline int _bpf_readarg_%s_%d("
+                "static __always_inline int _bpf_readarg_%s_%d("
                 "struct pt_regs *ctx, void *dest, size_t len) {\n"
                 "  if (len != sizeof(%s)) return -1;\n",
                 attached_to_.value(), arg_n + 1, ctype);


yonghong-song:
lgtm.

Yonghong

PR #1152, opened 2017-05-03
Add documentation to guide future scripts to be Python3-ready

palmtenor:
I assume this is what we agreed per today's discussion, that we are going to move to Python3 eventually?
Having new scripts Python3-ready reduces our future work to migrate.


4ast:
+1 :)

PR #1153, opened 2017-05-03
Update the changes of flag HAVE_BPF_JIT in kernel 4.7 to the INSTALL

secmobi:
Since Linux kernel 4.7, the flag HAVE_BPF_JIT has been replaced by HAVE_CBPF_JIT and HAVE_EBPF_JIT:

https://github.com/torvalds/linux/commit/6077776b5908e0493a3946f7d3bc63871b201e87

I've updated the INSTALL.md to follow the changes.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -31,6 +31,8 @@ CONFIG_HAVE_BPF_JIT=y
 CONFIG_BPF_EVENTS=y
 ```
 
+For Linux kernel version 4.7 or newer, the flag `CONFIG_HAVE_BPF_JIT` has been replaced by `CONFIG_HAVE_EBPF_JIT` (for x86, arm64 and s390) or by `CONFIG_HAVE_CBPF_JIT` (for arm, mips, powerpc and sparc.)
+
 There are a few optional kernel flags needed for running bcc networking examples on vanilla kernel:
 
 ```


INSTALL.md
```diff
@@ -31,6 +31,8 @@ CONFIG_HAVE_BPF_JIT=y
 CONFIG_BPF_EVENTS=y
 ```
 
+For Linux kernel version 4.7 or newer, the flag `CONFIG_HAVE_BPF_JIT` has been replaced by `CONFIG_HAVE_EBPF_JIT` (for x86, arm64 and s390) or by `CONFIG_HAVE_CBPF_JIT` (for arm, mips, powerpc and sparc.)
```
4ast: why is it relevant to users who install bcc?

secmobi: In two cases the information is useful:

1) when users are trying bcc on any Linux distribution, they may check the distribution's pre-installed kernel's compile config to ensure BPF is supported by the kernel. Especially, they could know which kind of BPF (cBPF or eBPF) is supported, which may determine whether some new BPF features are available or not. 

2) when users want to build their own Linux kernel with fully BPF support, these are essential flags that they have to enable in the .config file. 

The second case is exactly the situation I met: 

I was compiling the newest kernel for some bcc experiments. In the kernel's default config for some hardware, BPF isn't fully enabled (e.g., `BPF_SYSCALL` is not set). Hence I checked all flags listed in this `INSTALL.md` document and enabled them one by one. However, I couldn't find the `CONFIG_HAVE_BPF_JIT` option in newest kernel source and was pretty confused. Finally realize the flag was deprecated.

4ast: 'misinformation' is my problem with this change.
re: 1) cbpf is always supported. ebpf is supported when CONFIG_BPF_SYSCALL is on
(and that bit is already covered by this INSTALL.md)
re: 2) CONFIG_HAVE_EBPF_JIT and CONFIG_HAVE_CBPF_JIT has nothing to do with kernel providing either cbpf or ebpf api to user space. These config flags are hidden and not selectable by users. CONFIG_BPF_JIT is a different flag that is selectable and it's on by default in all distros. Depending on (c|e)bpf jit availability in the architecture it will
improve performance of one or both types of programs, but doesn't affect functionality
and availability of JITs is different with kernel releases.
So I don't think there is a need to mention these info in this install doc.

secmobi: If one checkout kernel source code and build it by default configuration for either x86_64 or for arm, the `CONFIG_BPF_JIT` is NOT on by default. I double checked this by generating the `.config`  with `make ARCH=x86_64 defconfig` and `make ARCH=arm versatile_defconfig`. 

You're right that `CONFIG_HAVE_(E|C)BPF_JIT` is enabled by default. However, the existing `INSTALL.md` misleads by guiding people to check config flag of `CONFIG_HAVE_BPF_JIT` cause the flag wasn't exist anymore in newest kernel:

https://github.com/iovisor/bcc/blame/master/INSTALL.md#L29

Another way to avoid the misleading is, I think, to delete `CONFIG_HAVE_BPF_JIT` from INSTALL.md since the flag is not selectable by users. Just don't think it's a good idea to provide an outdated information here.

secmobi: Anyway, if you think user don't need to consider about status of the `CONFIG_HAVE_` flags, that's fine. Please just close the pull request. 
And I do hope you could delete the `CONFIG_HAVE_BPF_JIT` in existing document due to the same reason. 
Thanks!

4ast: ahh. didn't realize we already mention CONFIG_HAVE_BPF_JIT in there. Yeah. Let's delete it.

divergentdave: FYI, `INSTALL.md` still says to look for CONFIG_HAVE_BPF_JIT. My kernel's configuration doesn't have that, but it does have `CONFIG_HAVE_EBPF_JIT`.

yonghong-song: Since you are looking at this, could you help submit a pull request to replace `CONFIG_HAVE_BPF_JIT` with `CONFIG_HAVE_EBPF_JIT`? Thanks!

PR #1155, opened 2017-05-04
Improve perf map parsing logic

palmtenor:
- If the function start address is 0, skip regardless of error or not.
- Check range overflows.

@markdrayton

### DIFF
--- a/src/cc/bcc_perf_map.c
+++ b/src/cc/bcc_perf_map.c
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 #include <ctype.h>
+#include <errno.h>
+#include <limits.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


src/cc/bcc_perf_map.c
```diff
@@ -82,13 +84,17 @@ int bcc_perf_map_foreach_sym(const char *path, bcc_perf_map_symcb callback,
...
-    if (*sep != ' ' || (sep == cursor && begin == 0))
+    if (begin == 0)
+      continue;
+    if (*sep != ' ' || (begin == ULLONG_MAX && errno == ERANGE))
       continue;
     cursor = sep;
     while (*cursor && isspace(*cursor)) cursor++;
 
     len = strtoull(cursor, &sep, 16);
-    if (*sep != ' ' || (sep == cursor && begin == 0))
```
palmtenor: Also fix a typo here `begin` -> `len`

markdrayton:
lgtm!

PR #1157, opened 2017-05-04
Fedora compile

drepper:
I need the two changes on this branch to compile bcc on a Fedora system.  The first is needed by all Fedora/RHEL users.  The second is needed in my case but most likely is only by chance not needed by everyone else.

### DIFF
--- a/cmake/FindLuaJIT.cmake
+++ b/cmake/FindLuaJIT.cmake
@@ -44,7 +44,7 @@ FIND_PATH(LUAJIT_INCLUDE_DIR lua.h
 )
 
 FIND_LIBRARY(LUAJIT_LIBRARY
-  NAMES libluajit-51.a libluajit-5.1.a libluajit.a
+  NAMES libluajit-51.a libluajit-5.1.a libluajit.a libluajit-5.1.so
   HINTS
   $ENV{LUAJIT_DIR}
   PATH_SUFFIXES lib64 lib
--- a/src/cc/bcc_proc.c
+++ b/src/cc/bcc_proc.c
@@ -320,7 +320,7 @@ static bool match_so_flags(int flags) {
     return (sizeof(void *) == 8);
   }
 
-  return true;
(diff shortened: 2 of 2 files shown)


drzaeus77:
[buildbot, ok to test]

drzaeus77:
Can you please give more details on how Fedora is breaking? I haven't seen particular issues with the FC24/25 builds I've tried. It will also help others that have problems to know that they need to pick up this change or not.

drepper:
On Thu, May 4, 2017 at 10:58 PM, Brenden Blanco <notifications@github.com>
wrote:

> Can you please give more details on how Fedora is breaking? I haven't seen
> particular issues with the FC24/25 builds I've tried. It will also help
> others that have problems to know that they need to pick up this change or
> not.
>

 There are three different breakages:

- there is no libluajit.a, with or without version number added.  So luajit
support is not added.  The build works but is incomplete

- the ld.so cache issue manifests itself at runtime.  When I use some of
the example scripts (like strlen_count.py) on a x86-64 machine the routines
in libbcc identify /libx32/libc.so.6 as the library to instrument.  That's
certainly special to my machine (not many people have a x32 environment)
but the same problem would be encountered if the ld.so cache would list
/lib/libc.so.6 before /lib64/libc.so.6.  There is no guarantee of any
ordering.  The patch I provided at least makes sure the pointer size
matches.  It's still not correct but better.

- the problem with finding lua.h is that the incorrect subdir names will
all cause the tests to fail.  In the end the cmake system will try
/usr/include/lua.h which does exist on my system but this is not the file
from luajit but instead the normal lua implementation.  /usr/include/lua.h
does not provide, for instance, LUA_GLOBALSINDEX.  This is where the build
fails for me.  I was only able to debug this issue by running cmake under
strace and examine which files it accesses but then it became clear.

drzaeus77:
@drepper TYVM!

Issue #1154, opened 2017-05-03
long compilation time if table key or value structure size is big


yonghong-song:
The following test case can demonstrate the problem:

```
[yhs@localhost test]$ cat test.py
#!/usr/bin/env python

from bcc import BPF

prog = """
#define STACK_MAX_LEN 64
#define STR_MAX_LEN 128

typedef struct {
  char name[STR_MAX_LEN];
  char file[STR_MAX_LEN];
} Symbol;

typedef struct {
  uint32_t pid;
  uint32_t tid;
  int tls_key;
  uint32_t stack_len;
  int32_t kernel_stack_id;
  int32_t user_stack_id;
  uint64_t pthread_id;
  uint64_t pthread_created;
  bool thread_state_different;
  bool stack_complete;
  Symbol stack[STACK_MAX_LEN];
} Event;

BPF_TABLE("percpu_array", int, Event, eventTable, 1);
__attribute__((section(".bpf.fn.log_py_stack")))
int log_py_stack(struct pt_regs* ctx) {
  return 0;
}
;
"""

# load BPF program
b = BPF(text=prog, debug=1)
[yhs@localhost test]$ 
```

Here the value type `Event` is pretty big. Internally, for each key and value type of each structure, BCC generates `readerX` and `writerX` functions, which utilizes `sscanf` and `snprintf`. For data structure with large size (e.g., array or array of array). The number of parameters to `sscanf` and `snprintf` could be pretty big, and compiler seems having a hard time to digest, thus resulting in pretty big compilation time.

@drzaeus77 what is your opinion on this?

yonghong-song:
CC the group: @iovisor/iovisor

drzaeus77:
Have been staring at this for a while today, and so far my best proposal is to use "%s" instead of "%i %i %i %i %i ..." for input of arrays of type u8. This reduces the running time on my machine from 50s to <1s. However, I haven't quite figured out how to convince sscanf to handle empty strings in a clean way.

Besides this approach, the other would be to conditionally compile these helper functions on an as-needed basis.

yonghong-song:
On Thu, May 4, 2017 at 5:23 PM, Brenden Blanco <notifications@github.com>
wrote:

> Have been staring at this for a while today, and so far my best proposal
> is to use "%s" instead of "%i %i %i %i %i ..." for input of arrays of type
> u8. This reduces the running time on my machine from 50s to <1s. However, I
[...]

PR #1160, opened 2017-05-07
Compile internal "sscanf" module only if needed (#1154)

yonghong-song:


src/cc/bpf_module.cc
```diff
@@ -396,16 +389,27 @@ int BPFModule::annotate() {
...
     }
   }
 
   rw_engine_ = finalize_rw(move(m));
+
+  return 0;
+}
+
+int BPFModule::finalize_annotate() {
+  fn_ptrs_lock_.lock();
```
mauriciovasquezbernal: What about using a lock guard here to avoid any future problem?

src/cc/bpf_module.cc
```diff
@@ -396,16 +389,27 @@ int BPFModule::annotate() {
...
+  if (has_printf_scanf_) {
+    fn_ptrs_lock_.unlock();
+    return 0;
+  }
+
   if (rw_engine_)
```
mauriciovasquezbernal: What happens if `rw_engine_` is null? Shouldn't it return at this point?

src/cc/bpf_module.cc
yonghong-song: good suggestion. I tried scoped_lock which is not available for c++11, but I forgot to try lock_guard.

src/cc/bpf_module.cc
yonghong-song: right.

yonghong-song:
@drzaeus77 I hit "Some checks haven't completed yet" for over a day now. Not enough info to see what is the problem? Could you take a look?

drzaeus77:
[buildbot, add to whitelist]

yonghong-song:
test_libbcc uses table scanf/snprintf directly through Table interface. Will add an interface in ebpf::BPF to permit late binding as well.

yonghong-song:
@drzaeus77  Could you take a look?

yonghong-song:
@drzaeus77 just a reminder. Did you get a chance to look at the patch?

drzaeus77:
Yes, I am looking at an alternative.

yonghong-song:
Thanks. I guess you are thinking to reduce scanf/snprintf overhead!

drzaeus77:
Ok, please take a look at #1167 as an alternative. I had several reasons to take this approach: 1. I didn't like the extra std::maps that were used in the code (prior to this commit) 2. llvm::Module already keeps sufficient data structures around, so makes sense to use them (and it has built-in lock guards).

PR #1164, opened 2017-05-10
Add ability for users to control symboling behavior

palmtenor:
This set of changes mainly adds `bcc_symbol_option`, and use it across stack to control how we want to do symboling.

The main motivation for this is that we often want different symboling behavior depending on the environment. For example, reading debug file and verifying its CRC checksum could be very expensive for a large binary, and we may not want to do that in some production systems. However we prefer doing so for some other use cases, such as in the script tools. So hard-code any behavior would not fit the need of all users.

`bcc_symbol_option` holds various configurations of what to do for symboling. It is passed from user API all the way down to ELF helpers. This also makes future addition to more different options easy and possible.

This set of changes touches many parts components of BCC. To make it easier to review, I have break it down to multiple small commits, and added explanation in both code comments and commit message. However, not all commits will compile.

This set of changes keeps all current behaviors unchanged, and allows pass `NULL` / `nullptr` option parameter to indicate use default, which is current behaviors. I will put up follow up PRs to make the options configurable in C++ and Python API so that users could configure the behavior they want.

cc @markdrayton @goldshtn

### DIFF
--- a/src/cc/bcc_elf.c
+++ b/src/cc/bcc_elf.c
@@ -26,7 +26,10 @@
 
 #include <gelf.h>
 #include "bcc_elf.h"
+#include "bcc_syms.h"
+
 #define NT_STAPSDT 3
+#define ELF_ST_TYPE(x) (((uint32_t) x) & 0xf)
 
 static int openelf(const char *path, Elf **elf_out, int *fd_out) {
   if (elf_version(EV_CURRENT) == EV_NONE)
@@ -150,6 +153,7 @@ int bcc_elf_foreach_usdt(const char *path, bcc_elf_probecb callback,
 }
 
 static int list_in_scn(Elf *e, Elf_Scn *section, size_t stridx, size_t symsize,
+                       struct bcc_symbol_option *option,
                        bcc_elf_symcb callback, void *payload) {
   Elf_Data *data = NULL;
 
@@ -168,6 +172,15 @@ static int list_in_scn(Elf *e, Elf_Scn *section, size_t stridx, size_t symsize,
 
       if ((name = elf_strptr(e, stridx, sym.st_name)) == NULL)
         continue;
+      if (name[0] == 0)
+        continue;
+
+      if (sym.st_value == 0)
+        continue;
+
+      uint32_t flag = 1 << ELF_ST_TYPE(sym.st_info);
+      if (!(option->use_symbol_type & flag))
+        continue;
 
       if (callback(name, sym.st_value, sym.st_size, sym.st_info, payload) < 0)
         return 1;      // signal termination to caller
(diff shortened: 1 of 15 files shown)


src/cc/bcc_elf.c
```diff
@@ -169,6 +172,15 @@ static int list_in_scn(Elf *e, Elf_Scn *section, size_t stridx, size_t symsize,
       if ((name = elf_strptr(e, stridx, sym.st_name)) == NULL)
         continue;
 
+      if (sym.st_value == 0)
```
palmtenor: This will cover #1137

src/cc/usdt_args.cc
```diff
@@ -35,14 +35,21 @@ std::string Argument::ctype() const {
...
+      .use_debug_file = 1,
+      .check_debug_file_crc = 1,
+      .use_symbol_type = {0, 1, 1, 0, 0, 0, 1} // STT_OBJECT, STT_FUNC and STT_TLS
+    };
+    return ProcSyms(*pid, &default_option)
         .resolve_name(binpath.c_str(), deref_ident_->c_str(), address);
   }
 
   if (!bcc_elf_is_shared_obj(binpath.c_str())) {
-    struct bcc_symbol sym = {deref_ident_->c_str(), binpath.c_str(), 0x0};
```
palmtenor: This was previously broken after #638. This PR also fixes that.

palmtenor:
Talked with @markdrayton offline, use bitmask flag instead of array

palmtenor:
@goldshtn: Could you take a look? Thanks a lot:)

goldshtn:
I'm sorry for the delay, it's a very busy couple of weeks - I am mostly out of the office at conferences. The general idea sounds good, I just wonder if there are any additional customizations we will need. What is the rationale behind the specific set of three flags that you introduced in this patch? What other flags will we likely need?

palmtenor:
@goldshtn No worries! Thanks for the comment.
- **The two debug file options** In the past few months we changed to prefer using debug file rather than binary itself when reading symbols. This is generally a good idea especially for script tools. However for our production services the debug files are huge and expensive (both CPU and memory) to read / calculate CRC, and most of them has good symbols in the binary already. We want to opt-out for using debug files for most of our binaries, but also don't want to have custom patch and lose the ability to opt back in if needed. Hence the options.
- **Symbol types** This is mostly for #1137. This way we can configure `ProcSyms` (currently reading all types of symbols) and some function -> address helpers (currently using customized logic to filter out non-function, but missing `STT_GNU_IFUNC`) using this option, for efficiency and code simplicity, meanwhile for the USDT path (as you have pointed on in #1137) we can still keep the behavior of look up all types of symbols.

markdrayton:
LGTM. I can't think of other options we'd want soon. If/when we do think of something it'll at least be easy to add to the struct here.

4ast:

4ast:
@palmtenor forgot how to retrigger the test. can you rebase and force push one more time? Thx!

drzaeus77:
It would make sense to rebase/merge anyway since the workaround for test_brb failure needs to be picked up.

palmtenor:
Rebased. Ready to merge it build passes. Thank you all!

PR #1167, opened 2017-05-10
Use late-binding to finalize snprintf/sscanf

drzaeus77:
Not all users require the snprintf/sscanf helpers, and in some types the
finalization can take a long time. Defer the finalization until the
first use. Wrap the invocation with a bound function such that the
BPFTable and other users don't need to keep an explicit BPFModule
reference. Use-after-free will be avoided by resetting the function to
an error-returning stub when/if the BPFModule is freed.

### DIFF
--- a/src/cc/bpf_module.cc
+++ b/src/cc/bpf_module.cc
@@ -116,16 +116,23 @@ BPFModule::BPFModule(unsigned flags, TableStorage *ts)
   }
 }
 
+static StatusTuple unimplemented_sscanf(const char *, void *) {
+  return StatusTuple(-1, "sscanf unimplemented");
+}
+static StatusTuple unimplemented_snprintf(char *, size_t, const void *) {
+  return StatusTuple(-1, "snprintf unimplemented");
+}
+
 BPFModule::~BPFModule() {
   engine_.reset();
   rw_engine_.reset();
   ctx_.reset();
 
   for (auto &v : tables_) {
-    v->key_sscanf = nullptr;
-    v->leaf_sscanf = nullptr;
-    v->key_snprintf = nullptr;
-    v->leaf_snprintf = nullptr;
+    v->key_sscanf = unimplemented_sscanf;
+    v->leaf_sscanf = unimplemented_sscanf;
+    v->key_snprintf = unimplemented_snprintf;
+    v->leaf_snprintf = unimplemented_snprintf;
   }
 
   ts_->DeletePrefix(Path({id_}));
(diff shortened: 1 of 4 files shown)


4ast:
if i'm reading it correctly clang/llvm will be invoked at sprintf/scanf time?
Like for example:
```
        b = BPF(text=text, debug=0)
        fn = b.load_func("foo", BPF.KPROBE)
        t = b.get_table("stats")
        s1 = t.key_sprintf(t.Key(2))
        self.assertEqual(s1, b"0x2")
        s2 = t.leaf_sprintf(t.Leaf(2, 3, 4, 1, (5, 6)))
        l = t.leaf_scanf(s2)
```
before:
bpf and x86 bytecode generated at 'b.load_func' time.

after:
'b.load_func' compiles bpf bits only.
't.key_sprintf' compiles sprintf x86 code
't.leaf_scanf' compiles scanf x86 code

right?

drzaeus77:
Yes. Making finalize_annotate public and required seemed like unnecessary complication from api point of view.

drzaeus77:
Just to clarify, we will still run the C++-IR generation phase, so the horrible code that is taking forever to optimize will still exist in a `Function *` stored in the `llvm::Module *`, but no x86 code exists yet.

Only once something calls `rw_engine_->getFunctionAddress` will llvm internally do `finalizeLoadedModules()` (see http://llvm.org/docs/doxygen/html/MCJIT_8cpp_source.html#l00370)

4ast:
> Only once something calls rw_engine_->getFunctionAddress will llvm internally do finalizeLoadedModules()

so on the first scanf/printf access all of them will be compiled or just one accessor?

drzaeus77:
Everything in the module, per LLVM's implementation. The module exclusively contains the reader/writer functions for the key/leaf types, nothing else.

src/cc/bpf_module.cc
```diff
@@ -355,15 +364,6 @@ int BPFModule::annotate() {
...
   auto m = ebpf::make_unique<Module>("sscanf", *ctx_);
 
-  struct llvmfnpointers {
-    llvm::Function *key_sscanf;
-    llvm::Function *leaf_sscanf;
-    llvm::Function *key_snprintf;
-    llvm::Function *leaf_snprintf;
-  };
-
-  std::map<TableDesc *, llvmfnpointers> ptrs_map;
```
4ast: removing this map as part of the fix is indeed nice

yonghong-song:
The change looks good to me. This is indeed a better fix than my previous patch to turn on/off based on whether any of functions is used. The fix here is really fine grained to each table key/value sscanf/snprintf functions.

4ast:
would be great to add a test for it

PR #1168, opened 2017-05-10
Fix bug when finding symbols for shared libraries

palmtenor:
A shared library can be loaded into multiple non-consecutive memory sections. Currently `Module::find_addr` always look up the address against the first section by subtracting the address with `start()`.

This commit fixes the issue by calculating the binary offset when calling `contains()` and respect to the memory section actually containing the address.

### DIFF
--- a/src/cc/bcc_syms.cc
+++ b/src/cc/bcc_syms.cc
@@ -201,9 +201,10 @@ bool ProcSyms::resolve_addr(uint64_t addr, struct bcc_symbol *sym,
   sym->offset = 0x0;
 
   const char *original_module = nullptr;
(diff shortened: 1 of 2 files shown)


src/cc/bcc_syms.cc
```diff
@@ -315,9 +317,7 @@ bool ProcSyms::Module::find_name(const char *symname, uint64_t *addr) {
   return false;
```
palmtenor: This part (`find_name`) is probably broken as well. But I have no idea how to fix. Given the binary offset, is it even possible to map it back to the correct memory load section?

PR #1169, opened 2017-05-10
llvm5.0: use input kind helper to avoid breakage

drzaeus77:
Upstream llvm renamed an enum, so use a helper function that is
available in all versions to fix compatibility.


yonghong-song:
Look good to me. I have a similar workaround in my private llvm repo for llvm 5.0.

Issue #1171, opened 2017-05-11
Public tables are never deleted


mauriciovasquezbernal:
I was looking at the code that handles the table creation and destruction, I cannot find any place where public tables are destroyed. In fact the code at [1] works, note that after the first program has been destroyed, the second one is still able to use the public table.

Is that behavior excepted?
@drzaeus77 

Thanks,
Mauricio

[1] https://gist.github.com/mauriciovasquezbernal/844ceca4a3ba48941c1a28fffc161dc5

drzaeus77:
No, I would consider this a bug, but one that I was aware of when pushing the initial implementation.

PR #1172, opened 2017-05-11
[RFC] add support for shared tables

mauriciovasquezbernal:
Currently tables can be created in two modes:
1. private to the ebpf program
2. shared among all the ebpf programs created by the same user-space
process

This commit extends bcc allowing to create a table that is shared among
a set of specific ebpf programs.
This is useful when creating network functions that are composed
by more than 1 ebpf program, a map is shared among all the programs of
the network function but isolated from different instances of that
function.

This code is just a RFC for now, shared tables are not destroyed as the same behavior applies on public tables [1].

The code itself is quire simple, most of it is just to pass around the `maps_ns` parameter, any suggestion about how to implement it without adding an extra argument is welcome.

[1] https://github.com/iovisor/bcc/issues/1171

### DIFF
--- a/src/cc/BPF.h
+++ b/src/cc/BPF.h
@@ -44,8 +44,9 @@ class BPF {
 public:
   static const int BPF_MAX_STACK_DEPTH = 127;
 
-  explicit BPF(unsigned int flag = 0, TableStorage* ts = nullptr)
-      : bpf_module_(new BPFModule(flag, ts)) {}
+  explicit BPF(unsigned int flag = 0, TableStorage* ts = nullptr,
+               const std::string &maps_ns = "")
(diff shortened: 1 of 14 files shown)


mauriciovasquezbernal:
@drzaeus77 do you have any comments about it?
Thanks!

goldshtn:
@drzaeus77 Your opinion would be appreciated; this has gone a bit stale, and would need a refresh in any case before merging.

mauriciovasquezbernal:
Sorry for the long delay from my side.
I had already talked to Brenden about this internally, but then we stopped using this feature and concentrate in other things. I'll back to you when I have some news.

4ast:

PR #1180, opened 2017-05-15
Enable recursive scanf support for char[] as string

drzaeus77:
When a bpf table contains i8[] in one of its keys/leaves, use "" to
enclose the value, rather than [ %i %i %i ... ] format. This simplifies
the code that is generated for cases such as #1154, and brings it back
under ~200ms code generation, instead of >30s. This change of format is
not particularly robust (it doesn't handle escaping the doublequote
character itself), but it should make more sense for the common case,
such as tracing files and pathnames.

The test case included tests both the functionality of the format string
handling as well as the compile time, since test_clang already has an
implicit 10second timeout limit.

Fixes: #1154

### DIFF
--- a/src/cc/bpf_module.cc
+++ b/src/cc/bpf_module.cc
@@ -214,11 +214,40 @@ static void parse_type(IRBuilder<> &B, vector<Value *> *args, string *fmt,
         *fmt += "\"%s\"";
         args->push_back(out);
       } else {
-        // Scan a single "" enclosed string. Passing multiple %[^"] arguments
-        // doesn't work because scanf stops parsing the string when an empty
-        // string is encountered, so here we individually call scanf and mask
-        // the empty string case. A scan failure (e.g. no enclosing "") should
-        // still return an error.
+        // When reading strings, scanf doesn't support empty "", so we need to
+        // break this up into multiple scanf calls. To understand it, let's take
+        // an example:
+        // struct Event {
+        //   u32 a;
+        //   struct {
+        //     char x[64];
+        //     int y;
+        //   } b[2];
+        //   u32 c;
+        // };
+        // The writer string would look like:
+        //  "{ 0x%x [ { \"%s\" 0x%x } { \"%s\" 0x%x } ] 0x%x }"
+        // But the reader string needs to restart at each \"\".
+        //  reader0(const char *s, struct Event *val) {
+        //    int nread, rc;
+        //    nread = 0;
+        //    rc = sscanf(s, "{ %i [ { \"%n", &val->a, &nread);
+        //    if (rc != 1) return -1;
+        //    s += nread; nread = 0;
+        //    rc = sscanf(s, "%[^\"]%n", &val->b[0].x, &nread);
+        //    if (rc < 0) return -1;
+        //    s += nread; nread = 0;
+        //    rc = sscanf(s, "\" %i } { \"%n", &val->b[0].y, &nread);
+        //    if (rc != 1) return -1;
+        //    s += nread; nread = 0;
+        //    rc = sscanf(s, "%[^\"]%n", &val->b[1].x, &nread);
+        //    if (rc < 0) return -1;
+        //    s += nread; nread = 0;
+        //    rc = sscanf(s, "\" %i } ] %i }%n", &val->b[1].y, &val->c, &nread);
+        //    if (rc != 2) return -1;
+        //    s += nread; nread = 0;
+        //    return 0;
+        //  }
         *fmt += "\"";
         finish_sscanf(B, args, fmt, locals, true);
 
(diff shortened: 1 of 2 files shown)


4ast:
magic. cc @yonghong-song

drzaeus77:
Hacks on top of hacks. Wish we had CTF.

src/cc/bpf_module.cc
```diff
@@ -153,24 +153,90 @@ static void debug_printf(Module *mod, IRBuilder<> &B, const string &fmt, vector<
   B.CreateCall(fprintf_fn, args);
 }
 
+static void finish_sscanf(IRBuilder<> &B, vector<Value *> *args, string *fmt,
```
yonghong-song: This is a genius hack! I just learned there exists %n for sscanf today! But the implementation is indeed very convoluted. Do you want to give a little more information like below about how reader and writer are generated.
The comments to illustrate how below (or similar) is generated will be good:
typedef struct {
  char stack1[STACK_MAX_LEN];
} Aux;

typedef struct {
  uint32_t a;
  Aux stack1[STACK_MAX_LEN];
  uint32_t b;
} Event;
Event type is used as key or value type.
 
fmt = "{ %i [ { \"%n"; sscanf(...)

/* the following two lines repeated a few times */
fmt = "%[^\"]%n"; sscanf(...)
fmt = "\" } { \"%n"; sscanf(...)

fmt = ""\" } ] %i }%n"; sscanf(...)

hopefully I did not make any mistake.

src/cc/bpf_module.cc
```diff
@@ -153,24 +153,90 @@ static void debug_printf(Module *mod, IRBuilder<> &B, const string &fmt, vector<
...
+  args->push_back(nread);
+  CallInst *call = B.CreateCall(sscanf_fn, *args);
+  call->setTailCall(true);
+
+  BasicBlock *label_true = BasicBlock::Create(B.getContext(), "", cur_fn);
+  BasicBlock *label_false = BasicBlock::Create(B.getContext(), "", cur_fn);
+
+  // exact_args means fail if don't consume exact number of "%" inputs
+  // exact_args is disabled for string parsing (empty case)
+  Value *cond = exact_args ? B.CreateICmpNE(call, B.getInt32(args->size() - 3))
```
yonghong-song: You could check B.getInt32(args->size() - 3) and bail out earlier.

src/cc/bpf_module.cc
```diff
@@ -153,24 +153,90 @@ static void debug_printf(Module *mod, IRBuilder<> &B, const string &fmt, vector<
...
       *fmt += " ";
     }
     *fmt += "}";
   } else if (ArrayType *at = dyn_cast<ArrayType>(type)) {
-    *fmt += "[ ";
-    for (size_t i = 0; i < at->getNumElements(); ++i) {
-      parse_type(B, args, fmt, at->getElementType(), B.CreateStructGEP(type, out, i), is_writer);
-      *fmt += " ";
+    if (at->getElementType() == B.getInt8Ty()) {
+      // treat i8[] as a char string instead of as an array of u8's
```
yonghong-song: Current implementation will match "\"" so users are forced to use string to encode u8 arrays.
How about test cases where users want to define a boolean array as a u8 array? Do we force
users to encode it with a string?

src/cc/bpf_module.cc
```diff
@@ -153,24 +153,90 @@ static void debug_printf(Module *mod, IRBuilder<> &B, const string &fmt, vector<
...
+        // the empty string case. A scan failure (e.g. no enclosing "") should
+        // still return an error.
+        *fmt += "\"";
+        finish_sscanf(B, args, fmt, locals, true);
+
+        *fmt = "%[^\"]";
+        args->push_back(out);
+        finish_sscanf(B, args, fmt, locals, false);
+
+        *fmt += "\"";
```
yonghong-song: finish_sscanf always clears *fmt, so you can just do an assignment here.

src/cc/bpf_module.cc
drzaeus77: Yes that would force a string, unfortunately. So far I see char[] a lot more often in the tools, so I would consider deferring until we actually have a use case (and so far very few people use these helpers anyway).

src/cc/bpf_module.cc
drzaeus77: I don't follow. At this point in the code, args->size() is always >= 3, since we always do resize(2) followed by push_back(nread). The IR that is generated can't be in any other location either, since it needs to immediately follow the sscanf call.

src/cc/bpf_module.cc
drzaeus77: Done.

src/cc/bpf_module.cc
yonghong-song: You are right. Awful mistake from me.

PR #1183, opened 2017-05-16
Use new Kernel functionality to get first key of map

palmtenor:
In https://github.com/torvalds/linux/commit/8fe45924387be6b5c1be59a7eb330790c61d5d10 we changed `BPF_MAP_GET_NEXT_KEY` so that it will return the first key of the map is given `NULL`. This helps us to walk the map without having to guess a non-existing key.

This change adds a helper `bpf_get_first_key` that automatically tries to use this new way to get first key of a map, and falls back to current behavior of trying a few different values if running on older Kernels. Also changes to use this new helper in C++ and  Python API.

### DIFF
--- a/src/cc/BPFTable.h
+++ b/src/cc/BPFTable.h
@@ -70,6 +70,11 @@ class BPFTableBase {
                            static_cast<void*>(value)) >= 0;
   }
 
+  bool first(KeyType* key) {
+    return bpf_get_first_key(desc.fd, static_cast<void*>(key),
+                             desc.key_size) >= 0;
+  }
+
   bool next(KeyType* key, KeyType* next_key) {
     return bpf_get_next_key(desc.fd, static_cast<void*>(key),
                             static_cast<void*>(next_key)) >= 0;
@@ -174,17 +179,18 @@ class BPFHashTable : public BPFTableBase<KeyType, ValueType> {
 
   std::vector<std::pair<KeyType, ValueType>> get_table_offline() {
     std::vector<std::pair<KeyType, ValueType>> res;
-
-    KeyType cur, nxt;
+    KeyType cur;
     ValueType value;
 
+    if (!this->first(&cur))
+      return res;
+
     while (true) {
-      if (!this->next(&cur, &nxt))
+      if (!this->lookup(&cur, &value))
         break;
-      if (!this->lookup(&nxt, &value))
+      res.emplace_back(cur, value);
+      if (!this->next(&cur, &cur))
         break;
-      res.emplace_back(nxt, value);
-      std::swap(cur, nxt);
     }
 
     return res;
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -133,6 +133,44 @@ int bpf_delete_elem(int fd, void *key)
   return syscall(__NR_bpf, BPF_MAP_DELETE_ELEM, &attr, sizeof(attr));
 }
 
+int bpf_get_first_key(int fd, void *key, size_t key_size)
+{
+  union bpf_attr attr;
+  int i, res;
+
+  memset(&attr, 0, sizeof(attr));
+  attr.map_fd = fd;
+  attr.key = 0;
+  attr.next_key = ptr_to_u64(key);
+
+  // 4.12 and above kernel supports passing NULL to BPF_MAP_GET_NEXT_KEY
+  // to get first key of the map. For older kernels, the call will fail.
+  res = syscall(__NR_bpf, BPF_MAP_GET_NEXT_KEY, &attr, sizeof(attr));
+  if (res < 0 && errno == EFAULT) {
+    // Fall back to try to find a non-existing key.
+    static unsigned char try_values[3] = {0, 0xff, 0x55};
+    attr.key = ptr_to_u64(key);
+    for (i = 0; i < 3; i++) {
+      memset(key, try_values[i], key_size);
+      // We want to check the existence of the key but we don't know the size
+      // of map's value. So we pass an invalid pointer for value, expect
+      // the call to fail and check if the error is ENOENT indicating the
+      // key doesn't exist. If we use NULL for the invalid pointer, it might
+      // trigger a page fault in kernel and affect performence. Hence we use
+      // ~0 which will fail and return fast.
+      // This should fail since we pass an invalid pointer for value.
+      if (bpf_lookup_elem(fd, key, ~0) >= 0)
+        return -1;
+      // This means the key doesn't exist.
+      if (errno == ENOENT)
+        return syscall(__NR_bpf, BPF_MAP_GET_NEXT_KEY, &attr, sizeof(attr));
+    }
+    return -1;
+  } else {
+    return res;
+  }
+}
+
 int bpf_get_next_key(int fd, void *key, void *next_key)
 {
   union bpf_attr attr;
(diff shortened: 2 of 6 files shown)


src/cc/libbpf.c
```diff
@@ -133,6 +133,42 @@ int bpf_delete_elem(int fd, void *key)
...
+  res = syscall(__NR_bpf, BPF_MAP_GET_NEXT_KEY, &attr, sizeof(attr));
+  if (res < 0 && errno == EFAULT) {
+    // Fall back to try to find a non-existing key.
+    static unsigned char try_values[3] = {0, 0xff, 0x55};
+    value = malloc(value_size);
+    for (i = 0; i < 3; i++) {
+      memset(key, try_values[i], key_size);
+      if (bpf_lookup_elem(fd, key, value) == 0)
+        continue;
+      memset(&attr, 0, sizeof(attr));
```
palmtenor: @4ast Can I safely assume the first syscall didn't modify `attr` (currently yes) so that I don't need to re-initialize it here?

src/cc/libbpf.c
```diff
@@ -133,6 +133,42 @@ int bpf_delete_elem(int fd, void *key)
...
+  attr.key = 0;
+  attr.next_key = ptr_to_u64(key);
+
+  // 4.12 and above kernel supports passing NULL to BPF_MAP_GET_NEXT_KEY
+  // to get first key of the map. For older kernels, the call will fail.
+  res = syscall(__NR_bpf, BPF_MAP_GET_NEXT_KEY, &attr, sizeof(attr));
+  if (res < 0 && errno == EFAULT) {
+    // Fall back to try to find a non-existing key.
+    static unsigned char try_values[3] = {0, 0xff, 0x55};
+    value = malloc(value_size);
```
4ast: you can avoid this alloc as well.
if you're only looking for existence then pass value=null then -ENOENT will indicate that element is not there
and EFAULT that it's present.

src/cc/libbpf.c
4ast: yes. drop this memset()

src/cc/libbpf.c
```diff
@@ -133,6 +133,42 @@ int bpf_delete_elem(int fd, void *key)
...
+      if (bpf_lookup_elem(fd, key, value) == 0)
+        continue;
+      memset(&attr, 0, sizeof(attr));
+      attr.map_fd = fd;
+      attr.key = ptr_to_u64(key);
+      attr.next_key = ptr_to_u64(key);
+      free(value);
+      return syscall(__NR_bpf, BPF_MAP_GET_NEXT_KEY, &attr, sizeof(attr));
+    }
+    free(value);
```
4ast: this double free looks wrong too.

palmtenor: This isn't double free. The previous one is before returning. But we will drop this.

src/cc/libbpf.c
palmtenor: cool!

palmtenor:
Is there something flaky going on for the Fedora build host? The failure looks like something related to the NIC situation of the host...

src/cc/libbpf.c
```diff
@@ -133,6 +133,38 @@ int bpf_delete_elem(int fd, void *key)
...
+  // to get first key of the map. For older kernels, the call will fail.
+  res = syscall(__NR_bpf, BPF_MAP_GET_NEXT_KEY, &attr, sizeof(attr));
+  if (res < 0 && errno == EFAULT) {
+    // Fall back to try to find a non-existing key.
+    static unsigned char try_values[3] = {0, 0xff, 0x55};
+    attr.key = ptr_to_u64(key);
+    for (i = 0; i < 3; i++) {
+      memset(key, try_values[i], key_size);
+      // This should fail since we pass NULL for value pointer.
+      if (bpf_lookup_elem(fd, key, NULL) >= 0)
```
yonghong-song: since this should fail, we probably should enforce (using an assert) instead of just return.

palmtenor: Hmm... I don't think we have any `assert` in the core C library. Plus aborting on issues reading the table feels a bit aggressive to me.

yonghong-song: I just checked. Yes, we do not any use of `assert` in libbpf.c etc. Looks like it tries to propagate error code back to user and let user to decide what to do. How about adding a stderr message to communicate this unusual error. I would be really surprised this error will show up which indicates some libbpf or kernel bugs.

4ast: i have second thoughts on this. Can we benchmark the speed of bpf_map_lookup_elem() with value==null and with valid pointer?
i thought that kernel should not be taking a page fault in such case, but now i'm not so sure.
If you do the experiment, try with value == ~0 as well.
Thanks!

drzaeus77:
```
16:   File "/home/fedora/jenkins/workspace/bcc-pr/label/fc24/tests/python/simulation.py", line 69, in _ns_add_ifc
16:     in_ifc = ns_ipdb.interfaces[in_ifname]
16: KeyError: 'ns2b'
16: 
16: ----------------------------------------------------------------------
16: Ran 1 test in 0.676s
16: 
16: FAILED (errors=1)
16: Failed
16/32 Test #16: py_test_brb2 .....................***Failed    0.80 sec
```

This is an intermittent failure, it should go away next time you push a change.

src/cc/libbpf.c
palmtenor: You are actually right:

palmtenor: You are actually right. I'm changing the pointer to use `~0`
```
Running 8192 lookups with value pointer 0x7efebe42c440
Spent 3708235 ns
Running 8192 lookups with value pointer 0x7efebe42c440
Spent 3699266 ns
Running 8192 lookups with value pointer 0x7efebe42c440
Spent 3703722 ns

Running 8192 lookups with value pointer (nil)
Spent 10260908 ns
Running 8192 lookups with value pointer (nil)
Spent 10142762 ns
Running 8192 lookups with value pointer (nil)
Spent 10079773 ns

Running 8192 lookups with value pointer 0xffffffffffffffff
Spent 3403328 ns
Running 8192 lookups with value pointer 0xffffffffffffffff
Spent 3418311 ns
Running 8192 lookups with value pointer 0xffffffffffffffff
Spent 3404907 ns
```

src/cc/libbpf.c
```diff
@@ -133,6 +133,38 @@ int bpf_delete_elem(int fd, void *key)
...
+  // to get first key of the map. For older kernels, the call will fail.
+  res = syscall(__NR_bpf, BPF_MAP_GET_NEXT_KEY, &attr, sizeof(attr));
+  if (res < 0 && errno == EFAULT) {
+    // Fall back to try to find a non-existing key.
+    static unsigned char try_values[3] = {0, 0xff, 0x55};
+    attr.key = ptr_to_u64(key);
+    for (i = 0; i < 3; i++) {
+      memset(key, try_values[i], key_size);
+      // This should fail since we pass an invalid pointer for value.
+      if (bpf_lookup_elem(fd, key, ~0) >= 0)
```
4ast: lgtm. could you add a comment to explain what is happening here?

src/cc/BPFTable.h
```diff
@@ -174,17 +179,18 @@ class BPFHashTable : public BPFTableBase<KeyType, ValueType> {
...
-    KeyType cur, nxt;
+    KeyType cur;
     ValueType value;
 
+    if (!this->first(&cur))
+      return res;
+
     while (true) {
-      if (!this->next(&cur, &nxt))
+      if (!this->lookup(&cur, &value))
```
drzaeus77: Since you're touching these lines, care to take out the `this->`?

palmtenor: @drzaeus77 I would love too. But because the base class is templated and how symbol lookup works, I had to do either `this->lookup()` or `BPFHashTable<KeyType, ValueType>::lookup()`, or `BPFTableBase<KeyType, ValueType>::lookup()`. Do we prefer the later?

drzaeus77:
Fair explanation, current alternative is preferred.

PR #1186, opened 2017-05-17
Implement free_compiler_resource interface method

yonghong-song:
o Many bcc use cases have a pattern like
    step_1: compile and load the program
    step_2: runtime monitoring by read/write to maps

    step_1 will allocate some memory for compilation, and
    right now they are not freed until process got killed
    or module unloaded.

    step_2 still uses the module for map-related functionality.
    step_2 will not use compilation context except for
    sscanf/snprintf functions.

  o This patch provides a new interface function
    free_compiler_resource.
    Application can call after step 1 to
    free compiler-incurred memory consumption.

    Optionally, free_compiler_resource can call
    malloc_trim to return top freed heap memory back
    to system. This is useful for glibc malloc.
    For some other malloc libraries like jemalloc,
    explicit malloc_trim is not needed.

  o A new python test case is added. For a real simple test, I have
    ('Before freeing and returning memory:', ['RssAnon:', '16788', 'kB'])
    ('After freeing and returning memory:', ['RssAnon:', '8196', 'kB'])

### DIFF
--- a/src/cc/usdt.cc
+++ b/src/cc/usdt.cc
@@ -152,7 +152,7 @@ bool Probe::usdt_getarg(std::ostream &stream) {
 
   for (size_t arg_n = 0; arg_n < arg_count; ++arg_n) {
     std::string ctype = largest_arg_type(arg_n);
-    std::string cptr = tfm::format("*((volatile %s *)dest)", ctype);
+    std::string cptr = tfm::format("dest", ctype);
 
     tfm::format(stream,
                 "static __always_inline int _bpf_readarg_%s_%d("
--- a/src/cc/usdt_args.cc
+++ b/src/cc/usdt_args.cc
@@ -67,17 +67,17 @@ bool Argument::assign_to_local(std::ostream &stream,
   }
 
   if (!deref_offset_) {
-    tfm::format(stream, "%s = (%s)ctx->%s;", local_name, ctype(),
+    tfm::format(stream, "%s = *(volatile %s *)&ctx->%s;", local_name, ctype(),
                 *base_register_name_);
     return true;
   }
 
   if (deref_offset_ && !deref_ident_) {
-    tfm::format(stream, "{ u64 __addr = ctx->%s + (%d)",
+    tfm::format(stream, "{ u64 __addr = (*(volatile u64 *)&ctx->%s) + (%d)",
                 *base_register_name_, *deref_offset_);
     if (index_register_name_) {
       int scale = scale_.value_or(1);
-      tfm::format(stream, " + (ctx->%s * %d);", *index_register_name_, scale);
+      tfm::format(stream, " + ((*(volatile u64 *)&ctx->%s) * %d);", *index_register_name_, scale);
     } else {
       tfm::format(stream, ";");
     }


yonghong-song:
Test failure:
>> The following tests FAILED:
>>	 15 - py_test_brb (Failed)

Details:
15: Test command: /home/fedora/jenkins/workspace/bcc-pr/label/fc25/build/tests/wrapper.sh "py_brb_c" "sudo" "/home/fedora/jenkins/workspace/bcc-pr/label/fc25/tests/python/test_brb.py" "test_brb.c"
15: Test timeout computed to be: 9.99988e+06
15: E
15: ======================================================================
15: ERROR: test_brb (__main__.TestBPFSocket)
15: ----------------------------------------------------------------------
15: Traceback (most recent call last):
15:   File "/home/fedora/jenkins/workspace/bcc-pr/label/fc25/tests/python/test_brb.py", line 172, in test_brb
15:     disable_ipv6=True)
15:   File "/home/fedora/jenkins/workspace/bcc-pr/label/fc25/tests/python/simulation.py", line 93, in _create_ns
15:     disable_ipv6)
15:   File "/home/fedora/jenkins/workspace/bcc-pr/label/fc25/tests/python/simulation.py", line 69, in _ns_add_ifc
15:     in_ifc = ns_ipdb.interfaces[in_ifname]
15: KeyError: 'ns_routerb'
15: 
15: ----------------------------------------------------------------------
15: Ran 1 test in 1.446s
15: 
15: FAILED (errors=1)
15: Failed
15/33 Test #15: py_test_brb ......................***Failed    1.64 sec

Passed on my local box.

Looks like it is flaky.

src/cc/bpf_module.cc
```diff
@@ -134,7 +135,14 @@ BPFModule::~BPFModule() {
...
   rw_engine_.reset();
   ctx_.reset();
+  if (do_trim) {
+    // return the freed top heap memory back to the system.
+    malloc_trim(0);
+  }
+}
 
+BPFModule::~BPFModule() {
+  free_compiler_resource(0);
```
4ast: overall lgtm.
looks like 2+ calls free_compiler_resource() will work fine,
but what about other BPFModule methods?
will they start crashing after free_compiler_resource() ?

drzaeus77:
I will review this sometime today.

yonghong-song:
Once compiler resources are freed. You can reload the program (e.g., calling load_c, which involving the whole compilation and loading process) since llvm context is gone. module sscanf and snprintf will fail since rw_engine_ is gone.

I did not test but it should not crash for sscanf and snprintf as I check the code it should return
-1.

4ast:
I mean BPFModule::function_start() will return the pointer that came from llvm SectionMemoryManager and when we do engine_.reset() and rw_engine_.reset(), is it still a valid pointer or use-after-free is about to happen?

drzaeus77:
>  use-after-free is about to happen

it will happen, but at least for the bpf instructions they are contiguous and can be copied for instance into an external vector<u8>. The same is not necessarily true for the reader/writer helpers since they may have global string pointers and such.

yonghong-song:
regarding to
> BPFModule::function_start() 
it will have use-after-free issue as @drzaeus77 mentioned since BPFModule private variable sections_ not reset.

BPFModule::function_start() is used by load_func API call in python. It is mandatory that free_compiler_resource is called after load_func. Otherwise, it won't work.

src/cc/bpf_module.cc
```diff
@@ -134,7 +135,14 @@ BPFModule::~BPFModule() {
   engine_.reset();
   rw_engine_.reset();
   ctx_.reset();
+  if (do_trim) {
+    // return the freed top heap memory back to the system.
+    malloc_trim(0);
```
yonghong-song: > if we can always immediately free the execution engine once we are done using it.
One particular issue related to this is sscanf/snprintf functionality. User could still use it.
so rw_engine_ may never be able to freed without explicit permission from users.

Let me do an experiment first: only reset engine_, but leave rw_engine_ and llvm ctx_ to
see how much saving. If most of memory is in engine_ itself (for the case no actual sscanf/snprintf usage from user), I will go with suggested approach.

For the saving, malloc_trim is only used for glibc. I can hack to find the real saving. In our production system with jemalloc (which will do trimming automatically), the saving is more than 2MB for one moderate size bpf program.

yonghong-song:
@drzaeus77 I did some additional measurement. See above comments. Any opinion?

PR #1190, opened 2017-05-19
Use BPF_F_CURRENT_CPU in perf_submit when possible

palmtenor:
https://github.com/torvalds/linux/commit/1e33759c788c78f31d4d6f65bac647b23624734c introduced the `BPF_F_CURRENT_CPU` to be used in `bpf_perf_event_output`. Using it would improve efficiency by saving one helper call.

This commit adds a `CUR_CPU_IDENTIFIER` from `CMake` file that automatically decides whether to use the new flag based on the running Kernel version.

Updated rewriter. Tested with both configs.

### DIFF
--- a/src/cc/export/helpers.h
+++ b/src/cc/export/helpers.h
@@ -65,6 +65,17 @@ BPF_TABLE(_table_type, _key_type, _leaf_type, _name, _max_entries); \
 __attribute__((section("maps/export"))) \
 struct _name##_table_t __##_name
 
+// Identifier for current CPU used in perf_submit and perf_read
+// Prefer BPF_F_CURRENT_CPU flag, falls back to call helper for older kernel
+// Can be overridden from BCC
+#ifndef CUR_CPU_IDENTIFIER
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 8, 0)
+#define CUR_CPU_IDENTIFIER BPF_F_CURRENT_CPU
+#else
+#define CUR_CPU_IDENTIFIER bpf_get_smp_processor_id()
+#endif
+#endif
+
 // Table for pushing custom events to userspace via ring buffer
 #define BPF_PERF_OUTPUT(_name) \
 struct _name##_table_t { \
--- a/src/cc/frontends/clang/CMakeLists.txt
+++ b/src/cc/frontends/clang/CMakeLists.txt
@@ -2,4 +2,8 @@
 # Licensed under the Apache License, Version 2.0 (the "License")
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DKERNEL_MODULES_DIR='\"${BCC_KERNEL_MODULES_DIR}\"'")
+if(DEFINED BCC_CUR_CPU_IDENTIFIER)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DCUR_CPU_IDENTIFIER='\"${BCC_CUR_CPU_IDENTIFIER}\"'")
+endif()
+
 add_library(clang_frontend STATIC loader.cc b_frontend_action.cc tp_frontend_action.cc kbuild_helper.cc ../../common.cc)
--- a/tests/python/test_perf_event.py
+++ b/tests/python/test_perf_event.py
@@ -17,7 +17,7 @@ BPF_ARRAY(prev, u64, NUM_CPUS);
 BPF_HISTOGRAM(dist);
 int kprobe__sys_getuid(void *ctx) {
     u32 cpu = bpf_get_smp_processor_id();
-    u64 val = cnt1.perf_read(cpu);
+    u64 val = cnt1.perf_read(CUR_CPU_IDENTIFIER);
 
     if (((s64)val < 0) && ((s64)val > -256))
         return 0;
@@ -27,7 +27,7 @@ int kprobe__sys_getuid(void *ctx) {
 }
 int kretprobe__sys_getuid(void *ctx) {
     u32 cpu = bpf_get_smp_processor_id();
-    u64 val = cnt1.perf_read(cpu);
+    u64 val = cnt1.perf_read(CUR_CPU_IDENTIFIER);
 
     if (((s64)val < 0) && ((s64)val > -256))
         return 0;
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -341,7 +341,7 @@ bool BTypeVisitor::VisitCallExpr(CallExpr *Call) {
           string args_other = rewriter_.getRewrittenText(expansionRange(SourceRange(Call->getArg(1)->getLocStart(),
                                                            Call->getArg(2)->getLocEnd())));
           txt = "bpf_perf_event_output(" + arg0 + ", bpf_pseudo_fd(1, " + fd + ")";
-          txt += ", bpf_get_smp_processor_id(), " + args_other + ")";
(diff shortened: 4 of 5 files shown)


src/cc/export/helpers.h
```diff
@@ -90,6 +90,14 @@ struct _name##_table_t { \
 __attribute__((section("maps/perf_array"))) \
 struct _name##_table_t _name
 
+// Identifier for current CPU used in perf_submit and perf_read
+// Prefer use BPF_F_CURRENT_CPU flag, falls back to call helper for older kernel
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 8, 0)
```
4ast: we got rid of all KERNEL_VERSION() checks, since they're not reliable due to backports.
imo we need a different way to check.
Unfortunately no good suggestions.

tests/python/test_perf_event.py
```diff
@@ -17,13 +17,13 @@ def test_cycles(self):
 BPF_HISTOGRAM(dist);
 int kprobe__sys_getuid(void *ctx) {
     u32 cpu = bpf_get_smp_processor_id();
-    u64 val = cnt1.perf_read(cpu);
+    u64 val = cnt1.perf_read(CUR_CPU_IDENTIFIER);
```
4ast: that anti-opimization, since llvm will still emit call into bpf_get_smp_processor_id() in the line 19.

src/cc/export/helpers.h
palmtenor: I see:(
We **would** be able to check `#ifdef BPF_F_CURRENT_CPU`. However we keep our copy of `bpf.h` and always keep it update to the latest, so no way of checking what user's Kernel actually has.....

tests/python/test_perf_event.py
palmtenor: Right, this is just to verify the Macro works correctly, because it's a test:)

src/cc/export/helpers.h
palmtenor: BTW, does anyone actually do "forward port"? i.e. they have higher Kernel version number but don't have all the corresponding patches? If people normally just back port newer changes to older version, I think this particular change is fine because they are just missing the optimization, nothing will break.

4ast: yeah. was thinking to do:
grep BPF_F_CURRENT_CPU /lib/modules/`uname -r`/build/include/uapi/linux/bpf.h
we already have directory hacks in KBuildHelper,
but not sure it's worth it.

4ast: not quite a 'forward port', but there are cases where features available in the older kernels will be disabled in the newer ones via .config

palmtenor:
@4ast What if we make this a build option, default to call the helper, and user can configure to build their BCC to use the `BPF_F_CURRENT_CPU` flag if their kernel supports it?

4ast:
@palmtenor yeah. build flags makes sense. ideally cmake configure step will pick it up automatically and user can override if necessary

palmtenor:
- Determine the flag from `CMake` file and allow user to override.
- Not doing this for `perf_read` since that's from within BPF program and harder to implement with current setup.

4ast:
not pretty. but get the job done. @drzaeus77 what do you think?

src/cc/frontends/clang/CMakeLists.txt
```diff
@@ -1,5 +1,18 @@
...
+                  COMMAND sort -V COMMAND tail -1
+                  OUTPUT_VARIABLE HIGHER_KERNEL_VERSION)
+  if(HIGHER_KERNEL_VERSION STREQUAL CUR_KERNEL_VERSION)
+    set(BCC_CUR_CPU_IDENTIFIER "BPF_F_CURRENT_CPU")
+  else()
+    set(BCC_CUR_CPU_IDENTIFIER "bpf_get_smp_processor_id()")
+  endif()
+endif()
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DKERNEL_MODULES_DIR='\"${BCC_KERNEL_MODULES_DIR}\"' -DCUR_CPU_IDENTIFIER='\"${BCC_CUR_CPU_IDENTIFIER}\"'")
```
yonghong-song: We could automate this completely inside the rewriter.
Basically, rewriter can generate code like below:
```
#ifdef BPF_F_CURRENT_CPU
  bpf_perf_event_output(..., BPF_F_CURRENT_CPU, ...)
#else 
  bpf_perf_event_output(..., bpf_get_smp_processor_id(), ...)
#endif 
```

This change will be automatically picked up and compiled during the transformation from rewritten src to llvm IR.

palmtenor:
@yonghong-song I wish we could:( See above discussion. `bpf.h` used in BPF program is from `src/cc/compat/linux/virtual_bpf.h`, which we always keep updated to newest features. So having `BPF_F_CURRENT_CPU` defined (from the BPF program's view) doesn't mean the system (Kernel) has this feature.

drzaeus77:
How about leaving the command line override in the cmake file, but move the kernel version check to helpers.h?

yonghong-song:
I see. Currently, we always take a snapshot from compat headers. However we can change the remapping based availability of valid kernel build tree. That is, if a valid build directory available, we should use $kernel_build_dir/include/uapi/linux/{bpf.h, bpf_common.h} instead of the one under
bcc/src/cc/compat/linux directory. With this plus rewriter change, we should have proper support, any issues here?

If a valid kernel build directory not available, we should just use bpf_smp_processor_id().

In most properly installed systems, valid kernel build tree should available.

palmtenor:
@drzaeus77 @yonghong-song I think I basically tried these ideas in previous versions. The forced push may have confused you, sorry for that!

> How about leaving the command line override in the cmake file, but move the kernel version check to helpers.h?

We can definitely do that. But we'll need to pass the override to the `helper.h`(in the BPF program) through CFlags (in `loader.cc`). Is that OK?

> However we can change the remapping based availability of valid kernel build tree

I don't know the background behind pulling in a copy of `bpf.h`. Currently we need valid Kernel build directory anyways (for other Kernel headers).

drzaeus77:
> But we'll need to pass the override to the helper.h(in the BPF program) through CFlags (in loader.cc). Is that OK?

Yes

palmtenor:
Updated according to @drzaeus77
- Select flag based on Kernel version from `helpers.h`.
- Allow override from build flag through the Clang loader.

yonghong-song:
I hacked a rewriter change. It is not too hard. The core part:

@@ -148,6 +149,16 @@ int ClangLoader::parse(unique_ptr<llvm::Module> *mod, TableStorage &ts, const st
   vector<string> kflags;
   if (kbuild_helper.get_flags(un.machine, &kflags))
     return -1;
+  // use the authentic bpf.h if it is available
+  if (kernel_path_info.first) {
+    auto it = ExportedFiles::headers().find("/virtual/include/bcc/bpf.h");
+    if (it == ExportedFiles::headers().end())
+      return -1;
+    auto buffer = llvm::MemoryBuffer::getFile(kdir + "/build/include/uapi/linux/bpf.h");
+    if (buffer) {
+      remapped_files_[it->first] = std::move(buffer.get());
+    }
+  }
   kflags.push_back("-include");

Later on in b_frontend_action, based on whether the source source is available or not,
we can generate #ifdef ....

The problem is that just including real kernel header file build/include/uapi/linu/bpf.h
is not enough. bcc helpers.h assumes the latest header file in compat/linux/bpf.h.
So with host kernel bpf.h, bpf program won't compile because of incompability between 
bpf.h and helpers.h, unless we want to tweak helpers.h, which seems way more
complicated.

Therefore, the current solution seems okay, let user decide what feature to use.
Users already need to be careful about what features to use when they write bpf program
anyway.

PR #1191, opened 2017-05-19
Fix usage of CMAKE_BUILD_TYPE setting

GBuella:
The build type was set to Release in the top level cmake file.
This was a rather confusing behaviour, as one would expect to
be able to easily debug the code after using `cmake -DCMAKE_BUILD_TYPE=Debug`.


drzaeus77:
[buildbot, test this please]

PR #1192, opened 2017-05-20
Unify Perf Event related Enums and checks

palmtenor:
Currently we have both helpers for reading Perf Event counters, and attaching to Perf Event counter overflow events. We have separated and duplicated value checking and Enums for them in Python API and C library. This change tries to unify them.

We are working on adding `HW_CACHE` event / read support on Kernel side, so this will make future addition of that into BCC easier.

In addition, make the BPF program of `test_perf_event.py` check for potential errno returned by `bpf_perf_event_read`, so that it would catch the cases when the Perf Event Array is not opened correctly. There is a small chance that a valid (`u64`) return value could coincide with an errno. But since this is a test case I think we should still add the check.

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -604,9 +604,31 @@ error:
   return NULL;
 }
 
+int invalid_perf_config(uint32_t type, uint64_t config) {
+  switch (type) {
+    case PERF_TYPE_HARDWARE:
+      return config >= PERF_COUNT_HW_MAX;
+    case PERF_TYPE_SOFTWARE:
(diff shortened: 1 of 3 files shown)


brendangregg:
LGTM

palmtenor:
@yonghong-song `bpf_open_perf_event` is for `bpf_perf_event_read`, which currently supports `PERF_TYPE_HARDWARE` and `PERF_TYPE_RAW` event counters

yonghong-song:
@palmtenor Thanks. Make sense. I mixed two types of usage.

PR #1195, opened 2017-05-21
IPDB.initdb() accepts parameter nl.

zejn:


drzaeus77:
Can you please provide a description of why this change is needed? Does it fix something caused by the recent commit on this line?

zejn:
Hi, after further observations I see that this change will break on newer pyroute2 releases. In pyroute2 releases before 0.4.2, IPDB initdb method signature is ``initdb(nl=None)`` and default is to ignore current namespace, in newer it is ``initdb()`` and default is to keep current namespace and this makes it somewhat hard to support both versions.

I was trying to build a package on debian and it has version 0.3.16 in jessie backports and testing (what is to become stretch) and there's pyroute2 0.4.12 in unstable, see https://packages.debian.org/search?keywords=python-pyroute2&searchon=names&suite=all&section=all

Ubuntu xenial (LTS) has package python-pyroute2 version 0.3.5, ubuntu yakkety (upcoming release) has 0.3.16, see http://packages.ubuntu.com/search?suite=all&searchon=names&keywords=python-pyroute2

PR #1196, opened 2017-05-22
Support symbol option in C++ API stack table

palmtenor:
Add parameter to control the usage of debug files on symboling in C++ `BPFStackTable` API, using the newly added `bcc_symbol_option` support.

Default behavior the the same as current behavior.


palmtenor:
Force push try to kick buildbot

PR #1200, opened 2017-05-24
Improve perf event clean up

palmtenor:
Recently we noticed that when all references of a perf event FD are closed, the free-ing of the perf event happens asynchronously and could delay for as much as over 100ms. If we are opening and closing perf events frequently, this delayed free could cause exhaustion of the physical PMU resource and failures. This is my rough understanding and @4ast could probably explain the situation much better than I do.

This set of changes:
- Adds `bpf_close_perf_event_fd` helper and use it when detaching BPF program from perf event in C++ and Python API.
- Disable the perf event fd in `perf_reader_free` before closing.

Will address `bpf_perf_event_read` and the perf ring buffer use cases in separated commits.

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -808,10 +808,21 @@ int bpf_attach_perf_event(int progfd, uint32_t ev_type, uint32_t ev_config,
   return fd;
 }
 
-int bpf_detach_perf_event(uint32_t ev_type, uint32_t ev_config) {
-  // Right now, there is nothing to do, but it's a good idea to encourage
-  // callers to detach anything they attach.
-  return 0;
+int bpf_close_perf_event_fd(int fd) {
+  int res, error = 0;
+  if (fd >= 0) {
+    res = ioctl(fd, PERF_EVENT_IOC_DISABLE, 0);
+    if (res != 0) {
+      perror("ioctl(PERF_EVENT_IOC_DISABLE) failed");
+      error = res;
+    }
+    res = close(fd);
+    if (res != 0) {
+      perror("close perf event FD failed");
+      error = (res && !error) ? res : error;
+    }
+  }
+  return error;
 }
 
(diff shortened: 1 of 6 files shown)


4ast:
thanks! Great fix. Please rebase

yonghong-song:
@4ast Just to double check. When you close a perf. event file descriptor, some resource associated with the file descriptor may be freed async later on, which causes the issue. So close really has aggressive return.

If after close, you will immediately open for another perf event, you will get a different FD if the previous close is not really finished yet? If you try to require access to certain non-shared resources, you may get EAGAIN so that you could try again?

4ast:
the only thing that close(fd) does it flush+atomic_dec. The resource may be freed soon or may be freed hours later, since somebody else might be holding it.
In case of perf_event anon file when it refcnt goes to zero it still takes a jiffy or so to cleanup the hw resources. It looks like when there are no hw resources next perf_event_open() will succeed, but the hw counter won't be ready, so reads from valid fd/perf counter will be failing until pmu programming catches up. It seems disabling perf_event frees up hw pmu synchronously. The rest of perf_event kernel datastructures are still around, but they don't interfere with new perf_event_open.

src/cc/libbpf.c
```diff
@@ -798,10 +798,19 @@ int bpf_attach_perf_event(int progfd, uint32_t ev_type, uint32_t ev_config,
...
-  // callers to detach anything they attach.
-  return 0;
+int bpf_close_perf_event_fd(int fd) {
+  int res, error = 0;
+
+  if (fd >= 0) {
+    res = ioctl(fd, PERF_EVENT_IOC_DISABLE, 0);
+    if (res != 0)
+      error = res;
+    res = close(fd);
```
yonghong-song: For various libc call related failures, do you want to print out an error message with perror, that will be good.

palmtenor:
Rebased~

PR #1201, opened 2017-05-24
funclatency/sslsniff: fix the TypeError when enabling pid filter on u…

htbegin:
…probe

attach_uprobe() and attach_uretprobe() expect the type of pid argument
is "int" instead of "str". Fix it by specifying the type of the "-p"/"--pid"
option as "int", and updating the format string of args.pid accordingly.

Fixes: d73c58f0c8c754bd591264205ac2ddc83c1d2c7f

### DIFF
--- a/tools/funclatency.py
+++ b/tools/funclatency.py
@@ -45,7 +45,7 @@ parser = argparse.ArgumentParser(
     description="Time functions and print latency as a histogram",
     formatter_class=argparse.RawDescriptionHelpFormatter,
     epilog=examples)
-parser.add_argument("-p", "--pid",
+parser.add_argument("-p", "--pid", type=int,
     help="trace this PID only")
 parser.add_argument("-i", "--interval", default=99999999,
     help="summary interval, seconds")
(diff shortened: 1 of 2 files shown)


drzaeus77:
[buildbot, ok to test]

brendangregg:
thanks, I'm fine with the change, I just couldn't reproduce the error. Is this for python3? (I was testing python2).

htbegin:
hi @brendangregg,

The fix has nothing to do with the python version. The TypeError can be reproduced by using the following command:
```bash
[htbegin@fbox bcc]$ sudo ./tools/funclatency.py -p 1 'c:*printf'
  File "./tools/funclatency.py", line 204, in <module>
    pid=args.pid or -1)
  File "/home/htbegin/sw/x86/lib/python2.7/site-packages/bcc/__init__.py", line 815, in attach_uprobe
    group_fd=group_fd)
  File "/home/htbegin/sw/x86/lib/python2.7/site-packages/bcc/__init__.py", line 818, in attach_uprobe
    (path, addr) = BPF._check_path_symbol(name, sym, addr, pid)
  File "/home/htbegin/sw/x86/lib/python2.7/site-packages/bcc/__init__.py", line 615, in _check_path_symbol
    psym,
ctypes.ArgumentError: argument 4: <type 'exceptions.TypeError'>: wrong type
```

Issue #1197, opened 2017-05-23
Separate perf event buffer readers from kprobe readers in Python API


palmtenor:
I'm planning to improve Perf Event clean-up to make BCC more reliable.

I noticed in Python API, the `perf_reader` from `PerfEventArray.open_perf_buffer` are put into `BPF.open_kprobes`, along with `perf_reader` from actual kprobes. Also, in `BPF.kprobe_poll`, they are also being polled together.

Is this intentional? I would like to have a separated dict for the Perf Event buffer readers, are there any concerns for that?

cc @drzaeus77 @4ast

drzaeus77:
They are treated the same because both regular kprobes and perf buffers can be polled, and hence have the same underlying perf_reader struct. See the test in tests/python/test_callchain.py, or the original commit #255

PR #1203, opened 2017-05-25
Improve PerfEventArray clean up

palmtenor:
This is a follow-up to #1200. More context are discussed there.

This commit changes:
- BPF `PERF_EVENT_ARRAY ` supports delete operation unlike regular array maps. Use that on clean-up instead of the "fake" delete from `ArrayBase`.
- Python `PerfEventArray` class is used for both `perf_submit / bpf_perf_event_output` and `perf_read / bpf_perf_event_read`. Differentiate them on clean-up and do proper clean-up respectively.
- Clean-up `PerfEventArray` instances on `BPF` instance exit.

For most of our script tools, this won't make a big difference since they just exit after finish running. But for more complicated and continuous use cases this would be helpful.

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -25,7 +25,7 @@ import sys
 basestring = (unicode if sys.version_info[0] < 3 else str)
 
 from .libbcc import lib, _CB_TYPE, bcc_symbol, bcc_symbol_option, _SYM_CB_TYPE
-from .table import Table
+from .table import Table, PerfEventArray
 from .perf import Perf
 from .utils import get_online_cpus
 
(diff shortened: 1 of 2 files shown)


src/python/bcc/__init__.py
```diff
@@ -1093,7 +1093,12 @@ def cleanup(self):
             # non-string keys here include the perf_events reader
             if isinstance(k, str):
```
palmtenor: I would love to see we separate storage for kprobes and perf ring buffers. But probably would do it in a separated change and after #1197 is discussed:)

drzaeus77: I'm not entirely convinced, unless we deprecate (:() the feature that a kprobe event can be polled the same as a perf buffer. Just as a historical point, polling kprobes used to be the only way to get callchains into userspace, all the other infra (perf_submit, stackid table) came several major versions later.

palmtenor: I see, that makes sense. I was wondering before why would people use the polling to get callchain:)
Maybe we should have `kprobe_poll` vs `perf_buffer_poll` then?

4ast: while looking into it recently I noticed that we waste 8 kernel pages for every kprobe due to this mmap and there are a bunch of kprobes, so make it optional would be good memory savings.

drzaeus77: I doubt anyone is using the `return 1` style, the perf_submit functionality was added early enough and all of our tools have adopted it if they could. @brendangregg does that sound accurate to you?

brendangregg: I used "return 1" twice, in dcsnoop and execsnoop, I think only by accident. I don't think those tools need it.

```
# grep "return 1" *.py
dcsnoop.py:    return 1;
execsnoop.py:    return 1;
tcptracer.py:  return 1;
tcptracer.py:  return 1;
```

drzaeus77: Oops, dcsnoop should probably be converted as a performance optimization. Returning 1 there causes more work on the kernel side, and I can tell from the python code that you don't need the perf event in userspace. The others are fine since they are just returning 1 from an internal static inline helper.

PR #1204, opened 2017-05-25
Some cleanups to libbpf and bcc_syms

drzaeus77:
This is some cleanup work geared towards making bcc somewhat more organized. Someday maybe we can break up the monolith. Please see individual commits for description.

### DIFF
--- a/src/cc/CMakeLists.txt
+++ b/src/cc/CMakeLists.txt
@@ -35,12 +35,18 @@ if (CMAKE_COMPILER_IS_GNUCC AND LIBCLANG_ISSTATIC)
   endif()
 endif()
 
-add_library(bcc-shared SHARED bpf_common.cc bpf_module.cc libbpf.c perf_reader.c table_storage.cc shared_table.cc bpffs_table.cc json_map_decl_visitor.cc exported_files.cc bcc_elf.c bcc_perf_map.c bcc_proc.c bcc_syms.cc usdt_args.cc usdt.cc common.cc BPF.cc BPFTable.cc)
+add_library(bcc-shared SHARED bpf_common.cc bpf_module.cc libbpf.c perf_reader.c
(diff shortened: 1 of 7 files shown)


4ast:
so long term idea is to make core library into c++ code with c interface?
it's fine for us, since we build it with multiple c++ compilers at once,
but it won't fly for typical distro, since c++ .o will be compiler dependent.

drzaeus77:
Can you be more specific on what you think will break?

4ast:
c++ compiler will produce .o that can only work with that compiler of that specific version,
so distros would need to somehow make sure that libbcc.so pulls all c++ dependencies and doesn't use any exceptions, frame unwinds and such.

drzaeus77:
Then I assume you would have the same concern about bpf_common.cc where we already do this externing?

4ast:
bpf_common.cc is different, since it's a C wrapper for existing C++ classes,
whereas libbpf.c is pure C and could have been pulled into separate pure C library
that could have been used out of C only projects.
I just don't see what this c->c++ buys us for libbpf.c

drzaeus77:
This argument makes more sense, but then we should clean out the circular dependency on ProcMountNS (c++ or c++-wrapper, doesn't matter) instead.

IMO libbpf.c is a mess, and a lot of the implementation is very specific to how python/c++ uses it. Besides the bpf syscalls, I don't see a lot of these functions being useful outside of this project. My inclination is to write this piece of code in a style that is more consistent with the rest of bcc, else there should be a standalone iovisor/libbpf github repo with _just_ the dedicated functions. I think the number of people that are using bcc and not using the c++/python/lua api/tools is 0 or close to it. otherwise they will need the adopt the elf loader approach and hence already have forked from here or are writing it inside the kernel tree.

PR #1210, opened 2017-05-31
Remove c++ dependencies from libbpf

drzaeus77:
This moves ProcMountNS to its own file for cleanliness and then reimplements the same API in C purely for libbpf to use. The first change isn't strictly necessary but I felt like making the cleanup anyway. This change will probably supersede #1204

### DIFF
--- a/src/cc/CMakeLists.txt
+++ b/src/cc/CMakeLists.txt
@@ -14,8 +14,6 @@ include_directories(${CMAKE_CURRENT_SOURCE_DIR}/compat)
 add_definitions(${LLVM_DEFINITIONS})
 configure_file(libbcc.pc.in ${CMAKE_CURRENT_BINARY_DIR}/libbcc.pc @ONLY)
 
-# prune unused llvm static library stuff when linking into the new .so
-set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--exclude-libs=ALL")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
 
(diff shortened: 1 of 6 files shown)


yonghong-song:
I looked at the code as well. Look good to me too.

qmonnet:
Hi, I get the following warning and error when trying to compile from master branch, and it seems related to this PR. Should I file a separate issue?

```
/root/bcc/src/cc/libbpf.c: In function ‘enter_mount_ns’:
/root/bcc/src/cc/libbpf.c:481:7: warning: implicit declaration of function ‘setns’ [-Wimplicit-function-declaration]
   if (setns(target_fd, CLONE_NEWNS)) {
       ^
/root/bcc/src/cc/libbpf.c:481:24: error: ‘CLONE_NEWNS’ undeclared (first use in this function)
   if (setns(target_fd, CLONE_NEWNS)) {
                        ^
/root/bcc/src/cc/libbpf.c:481:24: note: each undeclared identifier is reported only once for each function it appears in
/root/bcc/src/cc/libbpf.c: In function ‘exit_mount_ns’:
/root/bcc/src/cc/libbpf.c:501:17: error: ‘CLONE_NEWNS’ undeclared (first use in this function)
   if (setns(fd, CLONE_NEWNS))
                 ^
```

(Ubuntu 16.04, Linux 4.4.0, clang-3.7)

drzaeus77:
I can add that. In newer systems the setns warning should go away (hopefully). On my arch-linux host it is defined in:

```
/usr/include/bits/sched.h:

/* Switch process to namespace of type NSTYPE indicated by FD.  */
extern int setns (int __fd, int __nstype) __THROW;
```

qmonnet:
I investigated a little further. I appear to have a similar file on Ubuntu, `/usr/include/x86_64-linux-gnu/bits/sched.h` on my machine. It is called by `/usr/include/sched.h` and defines both `setns` and `CLONE_NEWNS`.

The reason I had the error, it seems, is that those declarations are under a `#ifdef __USE_GNU` condition that is not met (for me) by default. A possible fix is to add `#define _GNU_SOURCE` at the top of `src/cc/libbpf.c` (before all `#include`s), which triggers the definition of `__USE_GNU` in `/usr/include/features.h` and then of `setns` and `CLONE_NEWNS` in `/usr/include/x86_64-linux-gnu/bits/sched.h`.

Do you know if the buildbot could somehow define `_GNU_SOURCE` somewhere, which would explain the difference in behavior we observe? Would adding `#define _GNU_SOURCE` at the top of `libbpf.c` make sense, or could it [break](https://stackoverflow.com/questions/5582211/what-does-define-gnu-source-imply) stuff?

qmonnet:
Sorry, just noticed that you already added `<linux/sched.h>` in PR #1211. That fixes my issue. Thanks!

PR #1212, opened 2017-05-31
Force udst ctx->#reg load to be volatile

yonghong-song:
This is related to issue #1133. Compiler sometimes
generates code patterns likes:
     r1 = ctx + 96
     goto next
   here:
     r1 = ctx + 48
   next:
     r3 = load (r1 + 0)
Verifier will fail for such cases as r1 is marked
as "unknown" at the time of load.

The previous workaround is to add volatile attribute
to the store like
   *(volatile u64 *)&dest = ctx->bx
The hope is to force ctx related load in-place since
its value is needed for store.

Unfortunately, this does not always work and compiler
still has freedom to merge different ctx loads at the
same time honoring the volatile &dest. In USDT generated
code, different branches of &dest are the same.

This patch directly make ctx->bx itself as a volatile load:
  *(volatile u64 *)&ctx->bx
This seems working as compiler stops playing around
the address pointing to a volatile data.

### DIFF
--- a/src/cc/usdt.cc
+++ b/src/cc/usdt.cc
@@ -152,7 +152,7 @@ bool Probe::usdt_getarg(std::ostream &stream) {
 
   for (size_t arg_n = 0; arg_n < arg_count; ++arg_n) {
     std::string ctype = largest_arg_type(arg_n);
-    std::string cptr = tfm::format("*((volatile %s *)dest)", ctype);
+    std::string cptr("dest");
 
     tfm::format(stream,
                 "static __always_inline int _bpf_readarg_%s_%d("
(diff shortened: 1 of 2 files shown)


4ast:
that's cleaner workaround indeed.
I don't see the patch removing old volatile though?

yonghong-song:
It is removed in usdt.cc:
   for (size_t arg_n = 0; arg_n < arg_count; ++arg_n) {
     std::string ctype = largest_arg_type(arg_n);
-    std::string cptr = tfm::format("*((volatile %s *)dest)", ctype);
+    std::string cptr = tfm::format("dest", ctype);

4ast:
right. misread it.

yonghong-song:
@drzaeus77 @4ast Do you know how to initiate a test run? I just pushed a  change to fix a test failure (due to compilation error).

drzaeus77:
You are on the whitelist already. The buildbot should run automatically if you push the change to the remote branch.

PR #1213, opened 2017-05-31
Add a python test for usdt

yonghong-song:
o The sample application covers different aspects
    of usdt rewriter code generation:
    - const
    - *(ctx + offset)
    - use bpf_probe_read to get trace data
    - the switch in ctx->ip if the same probe
      is triggered in multiple .text locations.
  o Additional package required, systemtap-sdt-dev (ubuntu) or
    systemtap-sdt-devel (debian) package.

### DIFF
--- a/tests/python/test_usdt.py
+++ b/tests/python/test_usdt.py
@@ -1,0 +1,79 @@
+#!/usr/bin/python
+#
+# USAGE: test_usdt.py
+#
+# Copyright 2017 Facebook, Inc
+# Licensed under the Apache License, Version 2.0 (the "License")
+
+from __future__ import print_function
+from bcc import BPF, USDT
+from unittest import main, TestCase
+from subprocess import Popen, PIPE
+from tempfile import NamedTemporaryFile
+import os
+import signal
+
+class TestUDST(TestCase):
+    def setUp(self):
+        # Application, minimum, to define three trace points
+        app_text = b"""
+#include <unistd.h>
+#include <folly/tracing/StaticTracepoint.h>
+
+int main() {
+  char s[100];
+  int i, a = 20, b = 40;
+  for (i = 0; i < 100; i++) s[i] = (i & 7) + (i & 6);
+  while (1) {
+    FOLLY_SDT(test, probe_point_1, s[7], b);
+    FOLLY_SDT(test, probe_point_3, a, b);
+    sleep(3);
+    a++; b++;
+    FOLLY_SDT(test, probe_point_1, s[4], a);
+    FOLLY_SDT(test, probe_point_2, 5, s[10]);
+    FOLLY_SDT(test, probe_point_3, s[4], s[7]);
+  }
+  return 1;
+}
+"""
+        # BPF program
+        self.bpf_text = """
+#include <uapi/linux/ptrace.h>
+int do_trace(struct pt_regs *ctx) {
+    return 0;
+};
+int do_trace2(struct pt_regs *ctx) {
+    return 0;
+}
+int do_trace3(struct pt_regs *ctx) {
+    return 0;
+}
+"""
+
+        # Compile and run the application
+        self.ftemp = NamedTemporaryFile(delete=False)
+        self.ftemp.close()
+        comp = Popen(["gcc", "-I", "%s/include" % os.getcwd(),
+                      "-x", "c", "-o", self.ftemp.name, "-"],
+                     stdin=PIPE)
+        comp.stdin.write(app_text)
+        comp.stdin.close()
+        self.assertEqual(comp.wait(), 0)
+        self.app = Popen([self.ftemp.name])
+
+    def test_attach1(self):
+        # enable USDT probe from given PID and verifier generated BPF programs
+        u = USDT(pid=int(self.app.pid))
+        u.enable_probe(probe="probe_point_1", fn_name="do_trace")
+        u.enable_probe(probe="probe_point_2", fn_name="do_trace2")
+        u.enable_probe(probe="probe_point_3", fn_name="do_trace3")
+        b = BPF(text=self.bpf_text, usdt_contexts=[u])
+
+    def tearDown(self):
+        # kill the subprocess, clean the environment
+        self.app.kill()
+        self.app.wait()
+        os.unlink(self.ftemp.name)
+
+if __name__ == "__main__":
+    main()
--- a/tests/python/CMakeLists.txt
+++ b/tests/python/CMakeLists.txt
@@ -66,3 +66,5 @@ add_test(NAME py_test_dump_func WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
   COMMAND ${TEST_WRAPPER} py_dump_func simple ${CMAKE_CURRENT_SOURCE_DIR}/test_dump_func.py)
 add_test(NAME py_test_tools_smoke WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
   COMMAND ${TEST_WRAPPER} py_test_tools_smoke sudo ${CMAKE_CURRENT_SOURCE_DIR}/test_tools_smoke.py)
+add_test(NAME py_test_usdt WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+  COMMAND ${TEST_WRAPPER} py_test_usdt sudo ${CMAKE_CURRENT_SOURCE_DIR}/test_usdt.py)
--- a/tests/python/include/folly/tracing/StaticTracepoint-ELFx86.h
+++ b/tests/python/include/folly/tracing/StaticTracepoint-ELFx86.h
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2017 Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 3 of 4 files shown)


yonghong-song:
@drzaeus77 What do you think? We need to install extra packages on our test machines.

palmtenor:
Wouldn't it be better to just add a test binary have USDTs with different
argument formats? This way it's more stable and we can easily test
different setup?

drzaeus77:
Quote:
```
/* <sys/sdt.h> - Systemtap static probe definition macros.
   Copyright (C) 2010-2011 Red Hat Inc.

   This file is part of systemtap, and is free software in the public domain.
*/
```

If that is all you are looking for, include it in bcc or maybe pull out the relevant bits for your test program, i.e. expand the DTRACE_PROBE2() macro? Personally, I don't want to add a build dependency for this one test, if it can be avoided.

yonghong-song:
@palmtenor The reason I am using a source code instead of binary is to make it easy to maintain and change.
@drzaeus77 Okay. I pulled in the necessary header files in tests/python/include directory. This will not add additional build dependency.

tests/python/test_usdt.py
```diff
@@ -0,0 +1,69 @@
...
+  }
+  return 1;
+}
+"""
+
+# Compile and run the application
+ftemp = tempfile.NamedTemporaryFile(delete=False, suffix=".c")
+ftemp.write(app_text)
+ftemp.close()
+subprocess.check_call(["gcc", "-I", "%s/include" % os.getcwd(), "-o", "%s.out" % ftemp.name, ftemp.name])
```
drzaeus77: Can you use a pipe instead of a named temp file? I expect you'll still want it for the output binary, but 1 is better than 2.

See https://stackoverflow.com/questions/16552794/piping-a-string-to-gcc-with-pythons-subprocess-popen for example.

tests/python/include/sys/sdt.h
```diff
@@ -0,0 +1,426 @@
+/* <sys/sdt.h> - Systemtap static probe definition macros.
```
4ast: please copy paste folly's statictracepoint instead

yonghong-song: Done in the new revision.

tests/python/test_usdt.py
yonghong-song: Done in the new revision.

tests/python/test_usdt.py
```diff
@@ -0,0 +1,78 @@
...
+from unittest import main, TestCase
+from subprocess import Popen, PIPE
+from tempfile import NamedTemporaryFile
+import os
+import signal
+
+class TestUDST(TestCase):
+    def setUp(self):
+        # Application, minimum, to define three trace points
+        app_text = """
```
drzaeus77: Please prefix the string with `b"""`

tests/python/test_usdt.py
```diff
@@ -0,0 +1,78 @@
...
+                     stdin=PIPE, stdout=PIPE, stderr=PIPE)
+        comp.stdin.write(app_text)
+        comp.stdin.close()
+        self.assertEqual(comp.wait(), 0)
+        self.app = Popen([self.ftemp.name])
+
+    def test_attach1(self):
+        # enable USDT probe from given PID and verifier generated BPF programs
+        u = USDT(pid=int(self.app.pid))
+        u.enable_probe(probe="probe_point_1", fn_name="do_trace")
```
drzaeus77: This is probably preexisting, but I see a bunch of warnings from the generated code:

```
29: /virtual/main.c:5:26: warning: incompatible integer to pointer conversion assigning to 'void *' from 'volatile int8_t' (aka 'volatile signed char') [-Wint-conversion]
29:   case 0x40051fULL: dest = *(volatile int8_t *)&ctx->ax; return 0;
29:                          ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
29: /virtual/main.c:6:26: warning: incompatible integer to pointer conversion assigning to 'void *' from 'volatile int8_t' (aka 'volatile signed char') [-Wint-conversion]
29:   case 0x400537ULL: dest = *(volatile int8_t *)&ctx->ax; return 0;
29:                          ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
29: /virtual/main.c:13:153: warning: incompatible integer to pointer conversion assigning to 'void *' from 'int32_t' (aka 'int') [-Wint-conversion]
29:   case 0x40051fULL: { u64 __addr = (*(volatile u64 *)&ctx->bp) + (-12);int32_t __res = 0x0; bpf_probe_read(&__res, sizeof(__res), (void *)__addr); dest = __res; } return 0;
29:                                                                                                                                                         ^ ~~~~~
29: /virtual/main.c:14:152: warning: incompatible integer to pointer conversion assigning to 'void *' from 'int32_t' (aka 'int') [-Wint-conversion]
29:   case 0x400537ULL: { u64 __addr = (*(volatile u64 *)&ctx->bp) + (-8);int32_t __res = 0x0; bpf_probe_read(&__res, sizeof(__res), (void *)__addr); dest = __res; } return 0;
29:                                                                                                                                                        ^ ~~~~~
29: /virtual/main.c:21:152: warning: incompatible integer to pointer conversion assigning to 'void *' from 'int32_t' (aka 'int') [-Wint-conversion]
29:   case 0x400520ULL: { u64 __addr = (*(volatile u64 *)&ctx->bp) + (-8);int32_t __res = 0x0; bpf_probe_read(&__res, sizeof(__res), (void *)__addr); dest = __res; } return 0;
29:                                                                                                                                                        ^ ~~~~~
29: /virtual/main.c:22:26: warning: incompatible integer to pointer conversion assigning to 'void *' from 'volatile int8_t' (aka 'volatile signed char') [-Wint-conversion]
29:   case 0x400545ULL: dest = *(volatile int8_t *)&ctx->ax; return 0;
29:                          ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
29: /virtual/main.c:29:153: warning: incompatible integer to pointer conversion assigning to 'void *' from 'int32_t' (aka 'int') [-Wint-conversion]
29:   case 0x400520ULL: { u64 __addr = (*(volatile u64 *)&ctx->bp) + (-12);int32_t __res = 0x0; bpf_probe_read(&__res, sizeof(__res), (void *)__addr); dest = __res; } return 0;
29:                                                                                                                                                         ^ ~~~~~
29: /virtual/main.c:30:26: warning: incompatible integer to pointer conversion assigning to 'void *' from 'volatile int8_t' (aka 'volatile signed char') [-Wint-conversion]
29:   case 0x400545ULL: dest = *(volatile int8_t *)&ctx->dx; return 0;
29:                          ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
29: /virtual/main.c:36:8: warning: incompatible integer to pointer conversion assigning to 'void *' from 'int' [-Wint-conversion]
29:   dest = 5;
29:        ^ ~
29: /virtual/main.c:41:8: warning: incompatible integer to pointer conversion assigning to 'void *' from 'volatile int8_t' (aka 'volatile signed char') [-Wint-conversion]
29:   dest = *(volatile int8_t *)&ctx->ax;
29:        ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

tests/python/test_usdt.py
```diff
@@ -0,0 +1,78 @@
...
+};
+int do_trace2(struct pt_regs *ctx) {
+    return 0;
+}
+int do_trace3(struct pt_regs *ctx) {
+    return 0;
+}
+"""
+
+        # Compile and run the application
```
drzaeus77: There are a few dangling file descriptors in this test. On python3 I get this:

```
29: /usr/lib/python3.6/unittest/case.py:597: ResourceWarning: unclosed file <_io.BufferedReader name=5>
29:   self.setUp()
29: /usr/lib/python3.6/unittest/case.py:597: ResourceWarning: unclosed file <_io.BufferedReader name=7>
29:   self.setUp()
29: ./usr/lib/python3.6/subprocess.py:761: ResourceWarning: subprocess 22720 is still running
29:   ResourceWarning, source=self)
```

yonghong-song: Fixed two python3 related issues.
Can we add some python3 tests in buildbot?

tests/python/test_usdt.py
yonghong-song: Yes, this kind of warnings pre-exist, and
this is due to the single probe function to serve different types.
Currently, all generated probe has a proto type like:
 static __always_inline int _bpf_readarg_do_trace_2(struct pt_regs *ctx, void *dest, size_t len)

here, "dest" is defined as a "void *" to cover different types (from int8 to int64, char *, etc.)
The value read from register or probing will need to first convert to origin type, and
then currently explicitly converted to "void *".

Such warning probably hard to remove unless we have more explicit "dest" type, which
is sometimes impossible since it may have multiple types need to cover.

yonghong-song:
@drzaeus77 Comments addressed. Could you take a look again?

tests/python/test_usdt.py
```diff
@@ -0,0 +1,78 @@
...
+        self.ftemp = NamedTemporaryFile(delete=False)
+        self.ftemp.close()
+        comp = Popen(["gcc", "-I", "%s/include" % os.getcwd(),
+                      "-x", "c", "-o", self.ftemp.name, "-"],
+                     stdin=PIPE)
+        comp.stdin.write(app_text)
+        comp.stdin.close()
```
drzaeus77: I still get:

```
29: ./usr/lib/python3.6/subprocess.py:761: ResourceWarning: subprocess 29570 is still running
```

The way to close this is with:
```
comp.kill()
comp.wait()
```
You should put those in the tearDown() function.

yonghong-song:
Thanks. Will fix this based on your suggestion.
It is weird that I did not observe this on my FC25 Python3.5 environment. Any opinion on why this may not show up sometimes?

drzaeus77:
The code that is producing the warning is new in 3.6 it seems.

https://github.com/python/cpython/commit/5a48e21ff163107a6f1788050b2f1ffc8bce2b6d

yonghong-song:
I see. Need higher python version to reproduce. I am using Popen.poll() to test which subprocess is open and found it is self.app, which is expected.  I then close it in the teardown using the procedure you provided. Hopefully it fixed the issue.

Issue #1109, opened 2017-04-12
print_log2_hist to strip leading zeros


brendangregg:
Instead of this:

```
# ./funclatency vfs_write
Tracing 1 functions for "vfs_write"... Hit Ctrl-C to end.
^C
     nsecs               : count     distribution
         0 -> 1          : 0        |                                        |
         2 -> 3          : 0        |                                        |
         4 -> 7          : 0        |                                        |
         8 -> 15         : 0        |                                        |
        16 -> 31         : 0        |                                        |
        32 -> 63         : 0        |                                        |
        64 -> 127        : 0        |                                        |
       128 -> 255        : 0        |                                        |
       256 -> 511        : 0        |                                        |
       512 -> 1023       : 0        |                                        |
      1024 -> 2047       : 1        |                                        |
      2048 -> 4095       : 400      |*                                       |
      4096 -> 8191       : 4250     |*******************                     |
      8192 -> 16383      : 8652     |****************************************|
     16384 -> 32767      : 4663     |*********************                   |
     32768 -> 65535      : 453      |**                                      |
     65536 -> 131071     : 17       |                                        |
    131072 -> 262143     : 0        |                                        |
    262144 -> 524287     : 0        |                                        |
    524288 -> 1048575    : 0        |                                        |
   1048576 -> 2097151    : 0        |                                        |
   2097152 -> 4194303    : 1        |                                        |
Detaching...
```

Print this:

```
# ./funclatency vfs_write
Tracing 1 functions for "vfs_write"... Hit Ctrl-C to end.
^C
     nsecs               : count     distribution
      1024 -> 2047       : 1        |                                        |
      2048 -> 4095       : 400      |*                                       |
      4096 -> 8191       : 4250     |*******************                     |
      8192 -> 16383      : 8652     |****************************************|
     16384 -> 32767      : 4663     |*********************                   |
     32768 -> 65535      : 453      |**                                      |
     65536 -> 131071     : 17       |                                        |
    131072 -> 262143     : 0        |                                        |
    262144 -> 524287     : 0        |                                        |
    524288 -> 1048575    : 0        |                                        |
   1048576 -> 2097151    : 0        |                                        |
   2097152 -> 4194303    : 1        |                                        |
Detaching...
```

To keep the output more concise. This is especially noticeable by tools that print multiple histograms.

frisso:
Personally I like the original format, which makes more clear which intervals are taken into consideration and it's easy to compare two histograms together.

[...]

PR #1226, opened 2017-06-20
Supports to strip leading zero on print_log2_hist function

thnam91:
I found #1109 issue about `print_log2_hist` function and read its comments.

I implemented a option that doesn't show the leading zeros.

I added argument `strip_leading_zero` to `print_log2_hist` function and its default value is `None`.

And I applied it to `funclatency.py`, `funclatency_example.txt`, `biolatency.py`, and `biolatency_example.txt`.

Instead of this:
```
# ./funclatency.py do_sys_open
Tracing 1 functions for "do_sys_open"... Hit Ctrl-C to end.
^C
     nsecs               : count     distribution
      1024 -> 2047       : 189      |**                                      |
      2048 -> 4095       : 2646     |****************************************|
      4096 -> 8191       : 2563     |**************************************  |
      8192 -> 16383      : 2107     |*******************************         |
     16384 -> 32767      : 302      |****                                    |
     32768 -> 65535      : 79       |*                                       |
     65536 -> 131071     : 14       |                                        |
Detaching...
```

Print this using -s option:
```
# ./funclatency.py -s do_sys_open
Tracing 1 functions for "do_sys_open"... Hit Ctrl-C to end.
^C
     nsecs               : count     distribution
       512 -> 1023       : 9        |                                        |
      1024 -> 2047       : 925      |*******                                 |
      2048 -> 4095       : 4781     |****************************************|
      4096 -> 8191       : 3532     |*****************************           |
      8192 -> 16383      : 625      |*****                                   |
     16384 -> 32767      : 163      |*                                       |
     32768 -> 65535      : 26       |                                        |
     65536 -> 131071     : 3        |                                        |
    131072 -> 262143     : 0        |                                        |
    262144 -> 524287     : 0        |                                        |
    524288 -> 1048575    : 0        |                                        |
   1048576 -> 2097151    : 0        |                                        |
   2097152 -> 4194303    : 0        |                                        |
   4194304 -> 8388607    : 0        |                                        |
   8388608 -> 16777215   : 0        |                                        |
  16777216 -> 33554431   : 0        |                                        |
  33554432 -> 67108863   : 1        |                                        |
Detaching...
```

And Instead of this:
```
# ./biolatency.py 
...
Tracing block device I/O... Hit Ctrl-C to end.
^C
        32 -> 63         : 5        |****                                    |
        64 -> 127        : 6        |*****                                   |
       128 -> 255        : 2        |*                                       |
      2048 -> 4095       : 41       |****************************************|
      4096 -> 8191       : 2        |*                                       |
```

Print this using -s option:
```
# ./biolatency.py -s
...
Tracing block device I/O... Hit Ctrl-C to end.
^C
        32 -> 63         : 2        |***********                             |
        64 -> 127        : 1        |*****                                   |
       128 -> 255        : 1        |*****                                   |
       256 -> 511        : 1        |*****                                   |
      2048 -> 4095       : 7        |****************************************|
```

Please review my code and comments about it. :D

If it can cause any other problems, please let me know how can I fix it.

Thank you.

### DIFF
--- a/tools/biolatency.py
+++ b/tools/biolatency.py
@@ -132,7 +132,7 @@ while (1):
     if args.timestamp:
         print("%-8s\n" % strftime("%H:%M:%S"), end="")
 
-    dist.print_log2_hist(label, "disk", strip_leading_zero=None)
+    dist.print_log2_hist(label, "disk")
     dist.clear()
 
     countdown -= 1
--- a/tools/biolatency.py
+++ b/tools/biolatency.py
@@ -10,7 +10,6 @@
 # Licensed under the Apache License, Version 2.0 (the "License")
 #
 # 20-Sep-2015   Brendan Gregg   Created this.
-# 28-Jun-2017   Taekho Nam      Not stripped leading zero.
 
 from __future__ import print_function
 from bcc import BPF
--- a/tools/biolatency.py
+++ b/tools/biolatency.py
@@ -10,7 +10,7 @@
 # Licensed under the Apache License, Version 2.0 (the "License")
 #
 # 20-Sep-2015   Brendan Gregg   Created this.
-# 23-Jun-2017   Taekho Nam      Added strip leading zero option.
+# 28-Jun-2017   Taekho Nam      Stripped leading zero.
 
 from __future__ import print_function
 from bcc import BPF
@@ -132,7 +132,7 @@ while (1):
     if args.timestamp:
         print("%-8s\n" % strftime("%H:%M:%S"), end="")
 
-    dist.print_log2_hist(label, "disk")
+    dist.print_log2_hist(label, "disk", strip_leading_zero=True)
     dist.clear()
 
     countdown -= 1
--- a/tools/biolatency.py
+++ b/tools/biolatency.py
@@ -4,13 +4,13 @@
 # biolatency    Summarize block device I/O latency as a histogram.
 #       For Linux, uses BCC, eBPF.
 #
-# USAGE: biolatency [-h] [-T] [-Q] [-m] [-D] [-s] [interval] [count]
+# USAGE: biolatency [-h] [-T] [-Q] [-m] [-D] [interval] [count]
 #
 # Copyright (c) 2015 Brendan Gregg.
 # Licensed under the Apache License, Version 2.0 (the "License")
 #
 # 20-Sep-2015   Brendan Gregg   Created this.
-# 21-Jun-2017   Taekho Nam      Added strip leading zero option.
+# 23-Jun-2017   Taekho Nam      Added strip leading zero option.
 
 from __future__ import print_function
 from bcc import BPF
@@ -37,8 +37,6 @@ parser.add_argument("-m", "--milliseconds", action="store_true",
     help="millisecond histogram")
 parser.add_argument("-D", "--disks", action="store_true",
     help="print a histogram per disk device")
-parser.add_argument("-s", "--strip", action="store_true",
-    help="strip leading zero on output")
 parser.add_argument("interval", nargs="?", default=99999999,
     help="output interval, in seconds")
 parser.add_argument("count", nargs="?", default=99999999,
@@ -134,7 +132,7 @@ while (1):
     if args.timestamp:
         print("%-8s\n" % strftime("%H:%M:%S"), end="")
 
-    dist.print_log2_hist(label, "disk", strip_leading_zero=args.strip)
+    dist.print_log2_hist(label, "disk")
     dist.clear()
 
     countdown -= 1
--- a/tools/biolatency.py
+++ b/tools/biolatency.py
@@ -4,12 +4,13 @@
 # biolatency    Summarize block device I/O latency as a histogram.
 #       For Linux, uses BCC, eBPF.
 #
-# USAGE: biolatency [-h] [-T] [-Q] [-m] [-D] [interval] [count]
+# USAGE: biolatency [-h] [-T] [-Q] [-m] [-D] [-s] [interval] [count]
 #
 # Copyright (c) 2015 Brendan Gregg.
 # Licensed under the Apache License, Version 2.0 (the "License")
 #
 # 20-Sep-2015   Brendan Gregg   Created this.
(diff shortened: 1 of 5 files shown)


goldshtn:
[buildbot, ok to test]

tools/biolatency.py
```diff
@@ -36,6 +37,8 @@
     help="millisecond histogram")
 parser.add_argument("-D", "--disks", action="store_true",
     help="print a histogram per disk device")
+parser.add_argument("-s", "--strip", action="store_true",
+    help="strip leading zero on output")
```
goldshtn: I don't think this makes a lot of sense as an additional option for any tool. This is a decision the tool author should make. And in fact, I don't see why we wouldn't want this behavior by default -- so probably in most cases, the choice should be to strip leading zeros. @brendangregg

brendangregg:
Yes, I think it should be up to the tool author. Some tools may want this as the default, some not. Adding an -s option should be an exception, only if really needed, and not on everything as it would clutter up their usage.

This was discussed briefly once before on #1109, as I'd suggested dropping the zeros by default, but someone pointed out that it can be a good visual clue.

Maybe... if the metric is latency, the zeros could be kept for that visual clue: as higher is bad. But for other metrics, like size, I'm not sure printing zeros matter. Hence, again, the tool author can pick the default that makes sense.

TaekhoNam:
I really agree that the default value is up to the tool author. Keeping zeros is currently the default option on both `funclatency.py` and `biolatency.py' because I also agree @frisso's opinion in #1109.

Actually, I'm not a initial tool author of `funclatency.py` and `biolatency.py`. So I'm very careful to change tools because I don't want to spoil the intention of the original author for `funclatency.py` and `biolatency.py` tools.

Is there anything I need to correct or impove before `-s` option is merged?

goldshtn:
Yes - I think Brendan (the author of these tools) and I are saying that we don't need the `-s` switch.

TaekhoNam:
Thanks to reply. Because I'm not good at English, I'm confused something. As you and Brendan discussed the default value, isn't it necessary to add `-s` option for only latency metric? And the metric both `funclatency.py` and `biolatency.py` is the latency. So I understood that `-s` switch is necessary, isn't it right?

If we don't need `-s` option anymore, do I close this pull request?

goldshtn:
We are saying that the tool should decide whether to use the new stripping behavior or not. It should not be exposed to the user as a `-s` switch. So you can make funclatency and biolatency not strip the leading zeros, per Brendan's suggestion, but you don't need the `-s` switch for that. 

Also, why close the PR? You can push a new commit that makes the changes you want and the PR will be updated automatically.

TaekhoNam:
Thanks to reply again. It's hard to understand contribution processes because I'm not familiar with code contribution. But I'm being able to get a great experience in this time. Also, I'm learning how to communicate with contributors and how can my code be applied on open source project.

As a result, `funclatency.py` and `biolatency.py` don't need `-s` switch and strip leading zeros, is it right?

I'm really sorry to bother you, but I'm curious about something. We already decided that `strip leading zero function` is not necessary. So #1109 issue is solved as our decision. But this PR is for #1109. Therefore if I push a new commit about another issue, I should make new PR. Isn't it?

goldshtn:
What makes this PR "about #1109"? You can edit the description and the commits to whatever you want. I think at this point we should make funclatency and biolatency strip zeros, and have the library function as is. No need for `-s`.

thnam91:
Due to your advices, I was able to understand what you and he said. And I realized that the tool author should decide to strip leading zeros or not. 

So I removed the python code and fixed the description for `-s` switch in `funclatency` and `biolatency`. Also, I only added an argument, 'strip_leading_zero', in the print_log2_hist function and its helper function in `table.py`.

- If author use `print_log2_hist` function using new argument like `strip_leading_zero=1`, print a histogram without leading zeros.

- If author use `print_log2_hist` function without `strip_leading_zero` argument, print a histogram including consecutive zeros from beginning.

I really hope that I understood properly what you mean. Could you give me any other advice or comment?

goldshtn:
> I think at this point we should make funclatency and biolatency strip zeros, and have the library function as is

Your latest patchset doesn't do this. You added the `strip_leading_zero` option to the Python API, but currently none of the scripts use it. We agreed that funclatency and biolatency should pass `True` to strip leading zeros.

thnam91:
Because of my English skill, I completely misunderstood something. I'm so sorry.
Finally, I updated funclatency and biolatency tools that should pass `True` to strip leading zeros.

I think that Eng comprehension is harder than code comprehension T_T.....
Anyway, thank you so much to reply.

thnam91:
Thank you so much :D. Finally I understood this issue clearly. I hope that this PR will be my final pull request.

tools/biolatency.py
```diff
@@ -131,7 +132,7 @@
     if args.timestamp:
         print("%-8s\n" % strftime("%H:%M:%S"), end="")
 
-    dist.print_log2_hist(label, "disk")
+    dist.print_log2_hist(label, "disk", strip_leading_zero=None)
```
goldshtn: I don't think this is needed because `None` is the default value.

tools/funclatency.py
```diff
@@ -237,9 +238,9 @@ def print_section(key):
 
     if need_key:
         dist.print_log2_hist(label, "Function", section_print_fn=print_section,
-            bucket_fn=lambda k: (k.ip, k.pid))
+            bucket_fn=lambda k: (k.ip, k.pid), strip_leading_zero=None)
```
goldshtn: Same here and below.

tools/biolatency.py
```diff
@@ -10,6 +10,7 @@
 # Licensed under the Apache License, Version 2.0 (the "License")
 #
 # 20-Sep-2015   Brendan Gregg   Created this.
+# 28-Jun-2017   Taekho Nam      Not stripped leading zero.
```
goldshtn: I'm sorry, but there are no changes in biolatency and funclatency any more, so these comments also seem unnecessary.

thnam91:
You are right. It is not unnecessary any more. Thank you. I missed it.

goldshtn:
Thanks, merged.

PR #1229, opened 2017-06-23
Added usdt_sample.

bveldhoen:
This sample contains:
- A library with an operation that uses usdt probes.
- A console application that calls the operation.
- Scripts to trace the latency of the operation.
- Corresponding cmake files.

N.B. There is still a problem with the response string, see lib1.cpp.


tests/usdt_sample/scripts/latency.c
```diff
@@ -0,0 +1,127 @@
+#include <linux/blkdev.h>
+#include <uapi/linux/ptrace.h>
+
+// The following methods were taken from the verbose output of argdist.py.
+// Function names were adjusted to match the BPF function name expectations.
+static inline int _bpf_readarg_trace_operation_start_1(struct pt_regs* ctx, void* dest, size_t len)
```
bveldhoen: I could not find documentation on the naming conventions and how this works, is there any documentation available?

tests/usdt_sample/scripts/latency.c
```diff
@@ -0,0 +1,127 @@
...
+        bpf_probe_read(&__res, sizeof(__res), (void*)__addr);
+        *((uint64_t*)dest) = __res;
+    }
+    return 0;
+}
+static inline int _bpf_readarg_trace_operation_start_2(struct pt_regs* ctx, void* dest, size_t len)
+{
+    if (len != sizeof(uint64_t))
+        return -1;
+    *((uint64_t*)dest) = (uint64_t)ctx->ax;
```
yonghong-song: The latest code base has "*(volatile uint64_t *)&ctx->ax" instead of "(uint64_t)ctx->ax" to prevent compiler optimization resulting programs being rejected by verifier. Could you update it?

tests/usdt_sample/scripts/latency.c
yonghong-song: I agree that we probably should add comments to describe the transformation.

The best non-code doc is probably from the following long commit message:

```
commit 3e39a08a4fd5eec7fac6b8910b190087e76805b4
Author: Sasha Goldshtein <goldshtn@gmail.com>
Date:   Thu Mar 24 08:39:47 2016 -0700

    argdist, trace, and tplist support for USDT probes

    ......
```
Basic idea:
    When you enable a tracepoint probe like:
    usdt_ctx.enable_probe(probe="operation_start", fn_name="trace_operation_start")
    The library will read the binary .note.stapsdt section to find the structure for the probe
    "operation_start", and generate functions to read its structure fields.
    In the above example, the first argument can be read through function:
    _bpf_readarg_trace_operation_start_1, and the second argument can be read from
    _bpf_readarg_trace_operation_start_2, etc.

in the BPF program, if you specify:
    bpf_usdt_readarg(2, ctx, &input)
It means you try to read the second field of the attached probe. The call be translated into
_bpf_readarg_trace_operation_start_2.

yonghong-song: I think you should not define function _bpf_readarg_trace_operation_start_1 like this in your bpf program. The bcc should generate this for you. You can use debug=4 option in BPF class constructor to dump out the processed file.

tests/usdt_sample/usdt_sample.md
```diff
@@ -0,0 +1,28 @@
+Tested on Fedora25 4.11.3-200.fc25.x86_64, gcc (GCC) 6.3.1 20161221 (Red Hat 6.3.1-1)
+
+Next to the bcc requirements, the following package(s) are required to build this sample:
+$ sudo dnf install systemtap-sdt-devel  # For Fedora25, other distro's might have differently named packages.
```
yonghong-song: This will be require an additional package. Also here, the application prints out the latency of the operation. There is no checking whether the test is passed or not. In my opinion, this probably should belong to "example" instead of "test".

We have a unit test test_usdt.py, which you may take a look. It uses folly usdt header file and avoid to use the systemtap-sdt-devel package. I suggest you take a look. This way, we can avoid all these "dtrace" stuff.

tests/usdt_sample/scripts/latency.c
goldshtn: Indeed, these functions should be generated by the USDT code in the bcc module.

goldshtn:
[buildbot, add to whitelist]

tests/usdt_sample/scripts/latency.c
bveldhoen: Thanks for the feedback, I'll get back to it shortly with the following changes (summary of feedback so far):
- Move usdt_sample to examples (instead of tests).
- Use the generated code for the _bpf_readarg_xxx methods (using the suggested debug=4 option).
- Use folly usdt headers to avoid systemtap-sdt-devel dependency. I'll add some documentation explaining that systemtap-sdt-devel can be used as an alternative to the usdt headers.

goldshtn: @bveldhoen: you don't need to pass debug=4 to get the generated code. @yonghong-song suggested using this flag to _see_ the generated code so it's easier to understand what's going on. If you don't get the generated code by default, that's a bug.

examples/usdt_sample/scripts/latency.py
```diff
@@ -0,0 +1,143 @@
...
+usdt_ctx.enable_probe(probe="operation_start", fn_name="trace_operation_start")
+usdt_ctx.enable_probe(probe="operation_end", fn_name="trace_operation_end")
+
+# Create BPF context, load BPF program
+bpf_ctx = BPF(text=latency_bpf_text, usdt_contexts=[usdt_ctx], debug=debugLevel)
+
+# Potential issue: when using a separate file (i.e. latency.c), the following error occurs:
+# LLVM ERROR: Program used external function '_bpf_readarg_trace_operation_end_1' which could not be resolved!
+#latency_src_file = "%s/latency.c" % os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))
+#bpf_ctx = BPF(src_file=latency_src_file, usdt_contexts=[usdt_ctx], debug=debugLevel)
```
bveldhoen: As the comments describe, using src_file for the BPF ctor didn't work, so I used the latency_bpf_text variable instead.

tests/python/test_usdt.py
```diff
@@ -19,25 +20,44 @@ def setUp(self):
...
     FOLLY_SDT(test, probe_point_1, s[7], b);
     FOLLY_SDT(test, probe_point_3, a, b);
     FOLLY_SDT(test, probe_point_1, s[4], a);
     FOLLY_SDT(test, probe_point_2, 5, s[10]);
     FOLLY_SDT(test, probe_point_3, s[4], s[7]);
+
+    memset(&s1, '\0', sizeof(s1));
+    strncpy(s1, str, len);
+    snprintf(s1 + len, sizeof(s1) - len, "%d", j);
+    FOLLY_SDT(test, probe_point_4, j++, &s1);
```
bveldhoen: Is there a better way to communicate strings/char array?
Is it possible to avoid a string copy altogether?

examples/usdt_sample/scripts/latency.c
```diff
@@ -0,0 +1,74 @@
+#include <linux/blkdev.h>
```
goldshtn: Do we still need this separate file if you ended up embedding it in the Python script?

examples/usdt_sample/usdt_sample.md
```diff
@@ -0,0 +1,30 @@
+Tested on Fedora25 4.11.3-200.fc25.x86_64, gcc (GCC) 6.3.1 20161221 (Red Hat 6.3.1-1)
+
+As an alternative to use ...bcc/tests/python/include/folly/tracing/StaticTracepoint.h,
+it's possible to use systemtap-sdt-devel.
+However, this is *not* required for this sample.
+$ sudo dnf install systemtap-sdt-devel  # For Fedora25, other distro's might have differently named packages.
```
goldshtn: I'd put the command-line examples in ```syntax-highlighted``` blocks.

tests/python/test_usdt.py
goldshtn: Not sure what the issue is. Most USDT probes for strings report either the pointer directly (a null-terminated string), or a pointer and a length as two arguments.

bveldhoen: Actually, these questions would have been more appropriate in the context of lib1.cpp.

I ran into some shenanigans when I provided the FOLLY_SDT macro with a hardcoded string, such as:
    FOLLY_SDT(test, probe_point_4, "bla123456", j++);
BPF comes back with a compilation error:
/virtual/main.c:55:21: error: use of undeclared identifier 'int80_t'
  if (len != sizeof(int80_t)) return -1;

Note that the erroneous identifier depends on the length of the hardcoded string, so if you specify "bla", it compiles (int32_t). :)

One can avoid this by creating a temporary const char* variable.

However, the above was reason for me to panic :) and start copying strings to a known-size stack allocated buffer before passing it along to BPF. That's a bit unnecessary, so I updated lib1.cpp, see my next commit. All's good now.

examples/usdt_sample/scripts/latency.c
bveldhoen: For this sample, not really. I'm adding more python files which perform similar tasks, and could have reused some of the code in latency.c. I could see a solution, where multiple python files use the same external .c file. No biggie, though, I'll remove the .c file.

examples/usdt_sample/usdt_sample.md
bveldhoen: Done!

bveldhoen:
I think BPF/bcc is a really powerful toolset! I had fun putting this sample together, and I'm hoping to apply some of it in the project I'm currently working on.

During the time I was putting this sample together, I ran into some questions. I'll post them here, if I need to post them somewhere else, pls. let me know.
* I'm copying a string using bpf_probe_read (i.e. see latency.py, line 94). I also saw the usage of __builtin_memcpy in other tools. What is the preferred usage?
* I had an attempt to implement a filter using a strcmp (in the code that's fed to BPF/bcc), but that failed. My guess is that the reason for this is that strcmp uses a loop construct, which is not allowed. Does this mean that this kind of filtering should be done from the using (python) script? If so, this would eliminate some possibilities to only trace a subset. Any hints?
* Calculations using doubles failed, are these restricted as well? Is there a list (or other documentation) of these restrictions?
* I looked at funclatency.py and argdist.py. I think not all functionality in these tools is available for usdt probes, is that correct? (I don't think funclatency works with usdt probes, and neither does the exit probe $latency filter in argdist?)

Thanks!

yonghong-song:
On Fri, Jul 7, 2017 at 12:02 AM, bveldhoen <notifications@github.com> wrote:

> I think BPF/bcc is a really powerful toolset! I had fun putting this
> sample together, and I'm hoping to apply some of it in the project I'm
> currently working on.
>
> During the time I was putting this sample together, I ran into some
> questions. I'll post them here, if I need to post them somewhere else, pls.
> let me know.
>
>    - I'm copying a string using bpf_probe_read (i.e. see latency.py, line
>    94). I also saw the usage of __builtin_memcpy in other tools. What is the
>    preferred usage?
>
> bcc does not have a path (-I) to include string.h, hence if you are using
memcpy,
the compiler will complain that is is undefined. So yes, please use
__builtin_memcpy as other examples.

>
>    - I had an attempt to implement a filter using a strcmp (in the code
>    that's fed to BPF/bcc), but that failed. My guess is that the reason for
>    this is that strcmp uses a loop construct, which is not allowed. Does this
>    mean that this kind of filtering should be done from the using (python)
>    script? If so, this would eliminate some possibilities to only trace a
>    subset. Any hints?
>
> Yes. strcmp cannot be inclined (plus it does have a loop), so it cannot be
used in
bpf program. What is you exact use case? We may propose a helper if we have
a good
use case.

>
>    - Calculations using doubles failed, are these restricted as well? Is
>    there a list (or other documentation) of these restrictions?
>
> doubles are not supported. There are no floating point instructions. You
can look at
kernel documentation filter.txt.

>
>    - I looked at funclatency.py and argdist.py. I think not all
>    functionality in these tools is available for usdt probes, is that correct?
>    (I don't think funclatency works with usdt probes, and neither does the
>    exit probe $latency filter in argdist?)
>
> funclatency.py based on kprobe/uprobe. Looks like argdist.py tries to
check kprobe/uprobe/usdt depending on input parameters.

goldshtn:
String operations can be done if you know the string you're comparing with at compile time. See trace and argdist - they have a helper that accepts a command line string and tests equality from the BPF program using a loop with a fixed number of iterations.

examples/usdt_sample/scripts/latency.py
yonghong-song: If there are code duplication, you can still put into a common C file and read it into a string in python code. If there are minor differences between different C files, you can use macro's to differentiate them and use -D to control each individual case.

tests/python/test_usdt.py
yonghong-song: I tried a little on this (passing a const string directly in FOLLY_SDT macro).
Yes, folly will generate a record based on size of the parameter, in which case
is the length of the string and the value is the string address itself.
 
If you use a "const char *" variable in the FOLLY_SDT
macro, the record size will be the pointer size and the value is stack+off
where you can get actual str pointer.

Yes, something can be improved here.

examples/usdt_sample/scripts/latency.py
bveldhoen: Done, moved the shared bpf code into a separate .c file, which is loaded as text in the python code.

tests/python/test_usdt.py
bveldhoen: Although a little unexpected, it's not a blocking issue.

bveldhoen:
Thanks for your feedback, I've added a new commit with the discussed changes.

yonghong-song:
The change generally looks good now. I only tested the default setup (i.e., using folly).

Could you squash 4 commits into one? Could you check the usdt_sample.md file to make sure its output conforms to the actual output from the program?

goldshtn:
We can squash when merging :)

bveldhoen:
Squashed, fixed up some parameter checking in the python scripts, and some changes to the .md file.

goldshtn:
Thank you for this, and for carefully fixing and polishing this PR!

bveldhoen:
Thank you for your help (and patience :) ). This was a great way to learn about BPF and bcc, apologies for some of my beginner questions. BPF/bcc is already very awesome, and it has great potential for further development!

PR #1230, opened 2017-06-26
Added helpers for BPF_PERCPU_ARRAY

romain-intel:
Tiny patch to add helpers for PERCPU_ARRAY.

### DIFF
--- a/src/cc/export/helpers.h
+++ b/src/cc/export/helpers.h
@@ -133,6 +133,23 @@ struct _name##_table_t _name
 #define BPF_ARRAY(...) \
   BPF_ARRAYX(__VA_ARGS__, BPF_ARRAY3, BPF_ARRAY2, BPF_ARRAY1)(__VA_ARGS__)
 
+#define BPF_PERCPU_ARRAY1(_name)                        \
+    BPF_TABLE("percpu_array", int, u64, _name, 10240)
+#define BPF_PERCPU_ARRAY2(_name, _leaf_type) \
+    BPF_TABLE("percpu_array", int, _leaf_type, _name, 10240)
+#define BPF_PERCPU_ARRAY3(_name, _leaf_type, _size) \
+    BPF_TABLE("percpu_array", int, _leaf_type, _name, _size)
+
+// helper for default-variable macro function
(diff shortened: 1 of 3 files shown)


drzaeus77:
Thanks @romain-intel 

Before moving forward, I have a question to the general users: do you find these shorthand macros helpful? My gut reaction at seeing another of these added to the helpers is that the cost of reading through the helpers.h is starting to negate the usefulness of having a slightly shorter table definition syntax the c/bpf code. Thoughts?

goldshtn:
Why would anyone be reading helpers.h? :blush: 
I'd say the helpers are useful if there are scripts and examples using them. I don't know if this one specifically is the most common case (it probably isn't), but it's nice to have.

goldshtn:
[buildbot, ok to test]

palmtenor:
I think these helpers are useful. We basically defined a Section Name -> Map Type mapping [in `b_frontend_action.cc`](https://github.com/iovisor/bcc/blob/master/src/cc/frontends/clang/b_frontend_action.cc#L651), which is not easy to learn by a new user.

Meanwhile @romain-intel can you also update `docs/reference_guide.md` to make this more visible? Thanks!

src/cc/export/helpers.h
```diff
@@ -133,6 +133,23 @@ struct _name##_table_t _name
...
+    BPF_TABLE("percpu_array", int, u64, _name, 10240)
+#define BPF_PERCPU_ARRAY2(_name, _leaf_type) \
+    BPF_TABLE("percpu_array", int, _leaf_type, _name, 10240)
+#define BPF_PERCPU_ARRAY3(_name, _leaf_type, _size) \
+    BPF_TABLE("percpu_array", int, _leaf_type, _name, _size)
+
+// helper for default-variable macro function
+#define BPF_PERCPU_ARRAYX(_1, _2, _3, NAME, ...) NAME
+
+// Define an array function (per CPU), some arguments optiona
```
drzaeus77: missing 'l' at the end of the comment :)

romain-intel:
Addressed comments. Let me know if there is anything else you would like me to do for this one. Thx.

PR #1232, opened 2017-06-29
Add open_perf_event to the C/C++ API

romain-intel:
Another change to bring the C/C++ API more in line with the Python one.

### DIFF
--- a/src/cc/BPF.cc
+++ b/src/cc/BPF.cc
@@ -15,6 +15,7 @@
  */
 
 #include <linux/bpf.h>
+#include <linux/perf_event.h>
 #include <unistd.h>
 #include <cstdio>
 #include <cstring>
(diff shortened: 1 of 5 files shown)


4ast:
cc @palmtenor

palmtenor:
Already chatted with @romain-intel offline. This looks good in general.

My only concern is naming are a bit confusing. I suggest rename the table class to `BPFPerfEventArray` (more aligned with name in Python API), and in `BPF` class rename `perf_event_buffers_` also to `perf_event_arrays_` (they are not really buffers).

Let's also maybe add some tests reading counter values to make sure this would always work in the future.

goldshtn:
[buildbot, add to whitelist]

romain-intel:
Addressed renaming comments. Will do a future update with a few tests. Should I add a new file to tests/cc/ for these tests or is there a better place?

PR #1239, opened 2017-07-03
MySQL tracing without USDT

WIgor:
What do you think about tracing MySQL without recompilation with USDT support? 

Actually, there is some downside, at least temporary, this script supports only 5.6 version and derivatives (MariaDB, Percona Sever 5.6, etc.), may be lower version also - but it wasn't tested.

From another side, starting from 8.0 MySQL, as I understand documentation, DTrace tracepoints were removed.

This script isn't a final version, just example for discussion.


goldshtn:
Hi Igor, it's definitley a good idea. In fact, it can be a good idea for any runtime or product that doesn't have USDT probes enabled by default -- which is, unfortunately, many of them. As you noted, using uprobes for this is a bit fragile -- functions are prone to moving and changing without notice -- but it's the next best thing if there is no static tracing available.

If you want to follow this through, it might be nice to actually consolidate this into the existing dbslower tool, so that it tries to use USDT probes and if they are not available, falls back to uprobes. How's that sound?

WIgor:
Yes. A movement to dbslower tool looks good, a code is almost same.

WIgor:
I updated dbslower, added support for MySQL 5.7. 

But for PostgreSQL same functionality might be not easy at all. Most interesting function (that contains tracing probes) exec_simple_query - is static.

tools/dbslower.py
```diff
@@ -101,17 +174,61 @@
...
 
-usdts = map(lambda pid: USDT(pid=pid), args.pids)
-for usdt in usdts:
-    usdt.enable_probe("query__start", "probe_start")
-    usdt.enable_probe("query__done", "probe_end")
+if args.path and not args.pids:
+    # Uprobes mode
+    program = program_uprobe
+
+    symbols = subprocess.check_output(["nm", "-aD", args.path])
```
goldshtn: We have built-in symbol support in the BPF module. Can you see if that can be used here, instead of invoking nm directly?

tools/dbslower.py
```diff
@@ -39,21 +46,87 @@
...
+};
+
+BPF_HASH(temp, u64, struct temp_t);
+BPF_PERF_OUTPUT(events);
+
+int mysql56_dispatch_start(struct pt_regs *ctx) {
+    u64 pid = bpf_get_current_pid_tgid();
+    u64 command  = (u64) PT_REGS_PARM1(ctx);
+
+    if (command == 3) {
```
goldshtn: It might be nice to add some documentation around these magic numbers.

tools/dbslower.py
```diff
@@ -39,21 +46,87 @@
...
+        bpf_probe_read(&data.query, sizeof(data.query), tempp->query);
+        events.perf_submit(ctx, &data, sizeof(data));
+    }
+    temp.delete(&pid);
+    return 0;
+};
+
+"""
+
+program_udst = """
```
goldshtn: There's quite a bit of code duplication, especially here around the result reporting. Perhaps instead of having two completely separate programs, you could use conditional compilation or text replacement to have as much shared code as possible? When the scripts grow, any duplication becomes a maintenance burden.

tools/dbslower.py
```diff
@@ -2,14 +2,19 @@
 #
 # dbslower      Trace MySQL and PostgreSQL queries slower than a threshold.
 #
-# USAGE: dbslower [-v] [-p PID [PID ...]] [-m THRESHOLD] {mysql,postgres}
+# USAGE: dbslower [-v] [-p PID [PID ...]] [-b PATH_TO_BINARY] [-m THRESHOLD] {mysql,postgres}
```
goldshtn: I think I'd go with `-x` for executable, or `-l` for library, and not `-b`, which I think we used for buffer sizes in other tools. It's not very important, but it's nice to have consistency across different tools.

tools/dbslower.py
```diff
@@ -128,8 +245,8 @@ def print_event(cpu, data, size):
         float(event.timestamp - start) / 1000000000,
         event.pid, float(event.delta) / 1000000, event.query))
 
-print("Tracing database queries for pids %s slower than %d ms..." %
-      (', '.join(map(str, args.pids)), args.threshold))
+# print("Tracing database queries for pids %s slower than %d ms..." %
```
goldshtn: Was this intentionally commented out?

WIgor: Accidentally, forgot to update this line.

tools/dbslower.py
WIgor: Can you point me an example? Initially, I tried to find more "pythonian" way to do this, but haven't found yet.

goldshtn: I think you want [`BPF.get_user_functions_and_addresses`](https://github.com/iovisor/bcc/blob/master/src/python/bcc/__init__.py#L770). It's not Pythonian, it's a wrapper on top of libbcc, which uses libelf for this.

tools/dbslower.py
```diff
@@ -101,17 +174,61 @@
...
+
+        if len(dispatch_fname) == 0:
+            print("Cant find function 'dispatch_command' in %s" % (args.path))
+            exit(1)
+
+        m = re.search("\\w+dispatch_command\\w+", dispatch_fname[0])
+        if m:
+            func_name = m.group(0)
+        else:
+            print("Cant extract real 'dispatch_command' function name from %s" % (dispatch_fname[0]))
```
goldshtn: This should be "Can't".

WIgor:
I updated script according to comments. Tested it in uprobe modes, but haven't tested it with USDT, yet. As I understand - USDT works only with Oracle Linux or Solaris.

goldshtn:
USDT is available on Linux. You need to build MySQL with the appropriate configure flag to enable DTrace probes.

WIgor:
Am I right that for this I need https://github.com/dtrace4linux/linux?

goldshtn:
No. Just build MySQL with the right configure switch. I'm on mobile right now but it should be googleable. You'd need the systemtap-sdt-devel package or similar for your distri.

WIgor:
cmake ../mysql-server -DENABLE_DTRACE=1 -DDOWNLOAD_BOOST=1 -DWITH_BOOST=~/boost

Gives an error during next make.

goldshtn:
Which distribution are you building on? You need the systemtap-sdt-dev package on Ubuntu, or the systemtap-sdt-devel package on Fedora, to get the dtrace command-line tool.

WIgor:
Yes, with systemtap-sdt-dev, compiles. And USDT mode works.

goldshtn:
[buildbot, ok to test]

goldshtn:
@WIgor Thanks for this! Merged.

sillyousu:
>But for PostgreSQL same functionality might be not easy at all. Most interesting function (that contains tracing probes) exec_simple_query - is static.
Are there any limitations when using uprobe with static function?

goldshtn:
It needs to be a symbol visible in the binary so we can find its address and place a probe on it.

sillyousu:
gcc generate symbol for non-inlined static function [1] , so uprobe can work with `exec_simple_query` in some case (at least work for me, pg10+gcc5.4+default build config ).
But I don't know is it common or not.

[1] https://gcc.gnu.org/ml/gcc-help/2015-08/msg00142.html

PR #1247, opened 2017-07-07
docs: update features list

qmonnet:
And fix a typo: direct_action -> direct-action

I could not find a tree with ARM32 JIT applied (just the [mail stating](https://www.spinics.net/lists/netdev/msg440291.html) it was applied). If someone has a link, please do not hesitate to replace mine.


yonghong-song:
Looks good to me. Thanks!

PR #1252, opened 2017-07-11
Fix trace.py for library filenames containing colons

vkhromov:
`trace.py` parses a probe using the colon as a separator.  As a result, it
fails to create a uprobe for binary/library with a filename containing colons.

This diff fixes that issue with `trace.py`.  It requires a kernel with
https://lkml.org/lkml/2017/1/13/585 merged to work properly, otherwise
`trace.py` fails to create uprobes.


goldshtn:
[buildbot, ok to test]

goldshtn:
Thanks for the fix!

PR #1262, opened 2017-07-21
Improve kprobe instance handling

palmtenor:
This change improves instance logic and clean-up for failure of kprobe attaching. Specifically:
- Move instance creation to a separated helper with clear logic of error handling and fallback.
- Improves check on all `snprintf` return results for possible buffer overflow.
- Make sure the event is deleted if the attaching (i.e.`bpf_attach_tracing_event`) failed.
- Currently, if we attach `kprobe1` and `kprobe2` in a BCC Process, both of them would be under the instance `bcc_PID`. However when we detach say `kprobe1`, we would just delete the instance. `kprobe2` would still be running since the event is still references by the attached probe, but user would not be able to see the instance under `/sys/kernel/debug/tracing/instances` for debug. Moreover, if now we attach a `kprobe3`, we will re-create an instance `bcc_PID` but that would actually be a different instance from the first one we created, which is confusing and consumes more Kernel memory. Fixed in this change by keeping a probe count and only delete when it's 0.

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -13,7 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#ifndef _GNU_SOURCE
 #define _GNU_SOURCE
+#endif
 
 #include <arpa/inet.h>
 #include <errno.h>
@@ -70,11 +72,61 @@
 #define PERF_FLAG_FD_CLOEXEC (1UL << 3)
 #endif
 
+#define BCC_INSTANCE_FAIL -1
+#define BCC_INSTANCE_NOT_INIT -2
+#define BCC_INSTANCE_USE_DEFAULT -3
+static int bcc_instance_probe_cnt = BCC_INSTANCE_NOT_INIT;
+
 static int probe_perf_reader_page_cnt = 8;
 
 static uint64_t ptr_to_u64(void *ptr)
 {
   return (uint64_t) (unsigned long) ptr;
+}
+
+static int bcc_get_instance(char *buf, size_t buf_size) {
+  int res;
+  // An instance has already been created.
+  if (bcc_instance_probe_cnt >= 0) {
+    res = snprintf(buf, buf_size, "/sys/kernel/debug/tracing/instances/bcc_%d/", getpid());
+    if (res <= 0 || res >= buf_size)
+      return BCC_INSTANCE_FAIL;
+    return 0;
+  }
+  // Already failed creating an instance. Don't try again and use default.
+  if (bcc_instance_probe_cnt == BCC_INSTANCE_USE_DEFAULT)
+    goto default_instance;
+  // Try create an instance.
+  if (access("/sys/kernel/debug/tracing/instances/", F_OK) != 0) {
+    // Fail if instance folder does not exist.
+    bcc_instance_probe_cnt = BCC_INSTANCE_USE_DEFAULT;
+    goto default_instance;
+  }
+  res = snprintf(buf, buf_size, "/sys/kernel/debug/tracing/instances/bcc_%d/", getpid());
+  if (res <= 0 || res >= buf_size)
+    return BCC_INSTANCE_FAIL;
+  if (mkdir(buf, 0755) != 0) {
+    bcc_instance_probe_cnt = BCC_INSTANCE_USE_DEFAULT;
+    goto default_instance;
+  }
+  // Instance for this BCC Process created.
+  bcc_instance_probe_cnt = 0;
+  return 0;
+
+default_instance:
+  res = snprintf(buf, buf_size, "/sys/kernel/debug/tracing/");
+  if (res <= 0 || res >= buf_size)
+    return BCC_INSTANCE_FAIL;
+  return BCC_INSTANCE_USE_DEFAULT;
+}
+
+static void bcc_check_clean_instance() {
+  if (bcc_instance_probe_cnt == 0) {
+    char buf[256];
+    snprintf(buf, sizeof(buf), "/sys/kernel/debug/tracing/instances/bcc_%d", getpid());
+    rmdir(buf);
+    bcc_instance_probe_cnt = BCC_INSTANCE_NOT_INIT;
+  }
 }
 
 int bpf_create_map(enum bpf_map_type map_type, int key_size, int value_size, int max_entries, int map_flags)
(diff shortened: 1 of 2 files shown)


palmtenor:
@4ast: It seems that the event id of Kernel Tracepoint events in different instances are always the same. Would there be a way to be able to attach multiple probes from different Processes to the same Kernel Tracepoint events? Or it is simply not supported by Kernel?

yonghong-song:
I guess you may try to change event name in attach_kprobe/detach_kprobe in python
__init__.py by adding pid (if not -1) to it. This way, attach/detach to the same kprobe
attachment point from different pid's should not impact each other.

yonghong-song:
There are no restrictions. You need to use different event name though.

As an example,
[root@localhost tracing]# cat kprobe_events
p:kprobes/myprobe do_sys_open dfd=%ax filename=%dx flags=%cx mode=+4($stack)
p:kprobes/myprobe2 do_sys_open dfd=%ax filename=%dx flags=%cx
mode=+4($stack)
[root@localhost tracing]#

palmtenor:
@yonghong-song I might have confused you, let me better explain:
- The creation of event instance is to be able to attach kprobe to the same Kernel function from **different** BCC process. It was implemented in #918 and this PR just improves the implementation.
- My question was about the event instance doesn't seem to help Kernel Tracepoint (not kprobe) events, since they have same id even in different instances. Thus we won't be able to have multiple BCC processes attached to the same Kernel Tracepoint.
- Your suggestion is a good way to support attaching multiple k/uprobes to the same Kernel / User function from **same** BCC process. That's a good idea and I will probably implement it later separately.

yonghong-song:
On Sat, Jul 22, 2017 at 1:47 AM, Teng Qin <notifications@github.com> wrote:
>
> @yonghong-song I might have confused you, let me better explain:
>
> The creation of event instance is to be able to attach kprobe to the same Kernel function from different BCC process. It was implemented in #918 and this PR just improves the implementation.
> My question was about the event instance doesn't seem to help Kernel Tracepoint (not kprobe) events, since they have same id even in different instances. Thus we won't be able to have multiple BCC processes attached to the same Kernel Tracepoint.

I see. tracepoint has predefined event name, id, format etc. For
uprobe/kprobe, users can define their own events with different names,
ids, etc., even for the same k/u probe function.

So yes, for a particular kernel tracepoint, it can only be enabled once.

src/cc/libbpf.c
```diff
@@ -70,13 +72,62 @@
 #define PERF_FLAG_FD_CLOEXEC (1UL << 3)
 #endif
 
+#define BCC_INSTANCE_FAIL -1
+#define BCC_INSTANCE_NOT_INIT -2
+#define BCC_INSTANCE_USE_DEFAULT -3
+static int bcc_instance_probe_cnt = BCC_INSTANCE_NOT_INIT;
```
yonghong-song: Does this need to be per event?

palmtenor: No just number of probes in the instance. I can add some test case for sure.

palmtenor:
@yonghong-song Added Unit Tests:)

src/cc/libbpf.c
```diff
@@ -70,13 +72,63 @@
...
   return (uint64_t) (unsigned long) ptr;
 }
 
+static int bcc_get_instance(char *buf, size_t buf_size) {
+  int res;
+  // An instance has already been created.
+  if (bcc_instance_probe_cnt >= 0) {
+    res = snprintf(buf, buf_size, "/sys/kernel/debug/tracing/instances/bcc_%d/", getpid());
+    if (res <= 0 || res >= buf_size)
+      return BCC_INSTANCE_FAIL;
```
yonghong-song: Can this happen at all?

src/cc/libbpf.c
```diff
@@ -70,13 +72,63 @@
...
 
 static uint64_t ptr_to_u64(void *ptr)
 {
   return (uint64_t) (unsigned long) ptr;
 }
 
+static int bcc_get_instance(char *buf, size_t buf_size) {
+  int res;
+  // An instance has already been created.
+  if (bcc_instance_probe_cnt >= 0) {
```
yonghong-song: This is a little misleading. bcc_instance_probe_cnt of 0 actually means there is one probe already.

src/cc/libbpf.c
```diff
@@ -70,13 +72,63 @@
...
+  if (bcc_instance_probe_cnt == BCC_INSTANCE_USE_DEFAULT)
+    goto default_instance;
+  // Try create an instance.
+  if (access("/sys/kernel/debug/tracing/instances/", F_OK) != 0) {
+    // Fail if instance folder does not exist.
+    bcc_instance_probe_cnt = BCC_INSTANCE_USE_DEFAULT;
+    goto default_instance;
+  }
+  res = snprintf(buf, buf_size, "/sys/kernel/debug/tracing/instances/bcc_%d/", getpid());
+  if (res <= 0 || res >= buf_size)
```
yonghong-song: We have numerous cases like 
  res = snprintf(..)
  if (res <= 0 || res >= buf_size) {
    ...
  }

Can we abstract this out as a macro or static function? This way, code is more concise and less distraction?

src/cc/libbpf.c
palmtenor: Not really... `bcc_instance_probe_cnt` means either:
- The instance is just created, still in the process of attaching a probe. If the attaching succeeded the count would increase to 1.  
- The last probe from the instance was detached and the instance is in the process of being deleted.

src/cc/libbpf.c
palmtenor: Good point, fixed

src/cc/libbpf.c
yonghong-song: Okay, I see. Thanks for explanation.

drzaeus77:
The change makes sense, but I feel that the probe_cnt variable is a little bit convoluted. First, though, one question: why do you special case the `instances` handling. From what I can tell, that code will be reliably enabled in all kernels that bpf is available, and so failure to create an instance should be treated as fatal within libbpf.c. That would simplify some of the logic here. In that case, how do you feel about using the `instances/bcc_$PID/events/[ku]probes` directories within the instance subdirectory as a hint about whether to cleanup or not. For instance, consider the following diff and let me know what you think:


```patch
diff --git a/src/cc/libbpf.c b/src/cc/libbpf.c
index 40bd836..776b1cd 100644
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -18,6 +18,7 @@
 #endif
 
 #include <arpa/inet.h>
+#include <dirent.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <limits.h>
@@ -72,10 +73,28 @@
 #define PERF_FLAG_FD_CLOEXEC (1UL << 3)
 #endif
 
-#define BCC_INSTANCE_FAIL -1
-#define BCC_INSTANCE_NOT_INIT -2
-#define BCC_INSTANCE_USE_DEFAULT -3
-static int bcc_instance_probe_cnt = BCC_INSTANCE_NOT_INIT;
+#ifndef BCC_TRACING_PATH
+#define BCC_TRACING_PATH "/sys/kernel/debug/tracing"
+#endif
+
+static char bcc_instance_path[256];
+static char bcc_instance_kprobes_path[256];
+static char bcc_instance_uprobes_path[256];
+
+__attribute__((constructor))
+static void instance_init() {
+  snprintf(bcc_instance_path, sizeof(bcc_instance_path),
+      BCC_TRACING_PATH "/instances/bcc_%d", getpid());
+  snprintf(bcc_instance_kprobes_path, sizeof(bcc_instance_kprobes_path),
+      "%s/events/kprobes", bcc_instance_path);
+  snprintf(bcc_instance_uprobes_path, sizeof(bcc_instance_uprobes_path),
+      "%s/events/uprobes", bcc_instance_path);
+}
+
+__attribute__((destructor))
+static void instance_cleanup() {
+  rmdir(bcc_instance_path);
+}
 
 static int probe_perf_reader_page_cnt = 8;
 
@@ -84,47 +103,51 @@ static uint64_t ptr_to_u64(void *ptr)
   return (uint64_t) (unsigned long) ptr;
 }
 
-static int bcc_get_instance(char *buf, size_t buf_size) {
-  int res;
-  // An instance has already been created.
-  if (bcc_instance_probe_cnt >= 0) {
-    snprintf(buf, buf_size, "/sys/kernel/debug/tracing/instances/bcc_%d/", getpid());
-    return 0;
-  }
-  // Already failed creating an instance. Don't try again and use default.
-  if (bcc_instance_probe_cnt == BCC_INSTANCE_USE_DEFAULT)
-    goto default_instance;
-  // Try create an instance.
-  if (access("/sys/kernel/debug/tracing/instances/", F_OK) != 0) {
-    // Fail if instance folder does not exist.
-    bcc_instance_probe_cnt = BCC_INSTANCE_USE_DEFAULT;
-    goto default_instance;
-  }
-  res = snprintf(buf, buf_size, "/sys/kernel/debug/tracing/instances/bcc_%d/", getpid());
-  if (res <= 0 || res >= buf_size)
-    return BCC_INSTANCE_FAIL;
-  if (mkdir(buf, 0755) != 0) {
-    bcc_instance_probe_cnt = BCC_INSTANCE_USE_DEFAULT;
-    goto default_instance;
-  }
-  // Instance for this BCC Process created.
-  bcc_instance_probe_cnt = 0;
-  return 0;
+static int instance_probe_count() {
+  struct dirent *dirent;
+  int rc = 0;
+  DIR *dir;
+
+  dir = opendir(bcc_instance_kprobes_path);
+  if (dir) {
+    for (dirent = readdir(dir); dirent; dirent = readdir(dir)) {
+      if (dirent->d_type != DT_DIR)
+        continue;
+      if (!strcmp(dirent->d_name, "..") || !strcmp(dirent->d_name, "."))
+        continue;
+      ++rc;
+    }
+    closedir(dir);
+  }
+
+  dir = opendir(bcc_instance_uprobes_path);
+  if (dir) {
+    for (dirent = readdir(dir); dirent; dirent = readdir(dir)) {
+      if (dirent->d_type != DT_DIR)
+        continue;
+      if (!strcmp(dirent->d_name, "..") || !strcmp(dirent->d_name, "."))
+        continue;
+      ++rc;
+    }
+    closedir(dir);
+  }
 
-default_instance:
-  res = snprintf(buf, buf_size, "/sys/kernel/debug/tracing/");
-  if (res <= 0 || res >= buf_size)
-    return BCC_INSTANCE_FAIL;
-  return BCC_INSTANCE_USE_DEFAULT;
+  return rc;
 }
 
-static void bcc_check_clean_instance() {
-  if (bcc_instance_probe_cnt == 0) {
-    char buf[256];
-    snprintf(buf, sizeof(buf), "/sys/kernel/debug/tracing/instances/bcc_%d", getpid());
-    rmdir(buf);
-    bcc_instance_probe_cnt = BCC_INSTANCE_NOT_INIT;
+static int bcc_get_instance() {
+  if (access(bcc_instance_path, F_OK) != 0) {
+    if (mkdir(bcc_instance_path, 0755) != 0) {
+      fprintf(stderr, "mkdir(%s): %s\n", bcc_instance_path, strerror(errno));
+      return -1;
+    }
   }
+  return 0;
+}
+
+static void bcc_check_clean_instance() {
+  if (instance_probe_count() == 0)
+    instance_cleanup();
 }
 
 int bpf_create_map(enum bpf_map_type map_type, int key_size, int value_size, int max_entries, int map_flags)
@@ -442,8 +465,7 @@ void * bpf_attach_kprobe(int progfd, enum bpf_probe_attach_type attach_type, con
   char buf[256];
   char new_name[128];
   struct perf_reader *reader = NULL;
-  int res, instance_res;
-  size_t buf_len;
+  int res;
 
   reader = perf_reader_new(cb, NULL, NULL, cb_cookie, probe_perf_reader_page_cnt);
   if (!reader) {
@@ -475,21 +497,17 @@ void * bpf_attach_kprobe(int progfd, enum bpf_probe_attach_type attach_type, con
     goto error;
   }
 
-  instance_res = bcc_get_instance(buf, sizeof(buf));
-  if (instance_res == BCC_INSTANCE_FAIL) {
-    fprintf(stderr, "Failed to get an event instance\n");
+  if (bcc_get_instance() < 0)
     goto clean_event_error;
-  }
-  buf_len = strlen(buf);
-  res = snprintf(&buf[buf_len], sizeof(buf) - buf_len, "events/kprobes/%s", new_name);
-  if (res <= 0 || (res + buf_len) >= sizeof(buf)) {
+
+  res = snprintf(buf, sizeof(buf), "%s/%s", bcc_instance_kprobes_path, new_name);
+  if (res <= 0 || res >= sizeof(buf)) {
     fprintf(stderr, "Event file path too long\n");
     goto clean_event_error;
   }
+
   if (bpf_attach_tracing_event(progfd, buf, reader, pid, cpu, group_fd) == 0) {
     close(kfd);
-    if (instance_res != BCC_INSTANCE_USE_DEFAULT)
-      bcc_instance_probe_cnt++;
     return reader;
   }
 
@@ -648,8 +666,6 @@ int bpf_detach_kprobe(const char *ev_name)
 {
   int res = bpf_detach_probe(ev_name, "kprobe");
 
-  if (bcc_instance_probe_cnt != BCC_INSTANCE_USE_DEFAULT)
-    bcc_instance_probe_cnt--;
   bcc_check_clean_instance();
 
   return res;
```

PR #1263, opened 2017-07-22
permit multiple pids attaching to the same probe

yonghong-song:
Currently, if more than one pid-associated USDT attaching to
the same probe, usdt readarg code will be generated twice and
the compiler will complain.

This patch solves issue by preventing code duplication if
a previous context with the same mnt point and exec binary
has generated the code. The event name is also changed
to have pid embedded in so different pid-associated
uprobe event will have different names.

This patch won't work for the case where the first pid
does not generate code for all subsequent probes like:
   # pid1 and pid2 in the same namespace and execute the
   # same binary
   u = USDT(pid=pid1)
   u.enable_probe(probe="probe_point_1", fn_name="do_trace1")
   u2 = USDT(pid=pid2)
   u2.enable_probe(probe="probe_point_2", fn_name="do_trace2")
   u2.enable_probe(probe="probe_point_3", fn_name="do_trace3")
   b = BPF(text=bpf_text, usdt_contexts=[u, u2])
The compilation error will still appear. But this use case
should be rare in practical.

### DIFF
--- a/src/cc/ns_guard.cc
+++ b/src/cc/ns_guard.cc
@@ -21,7 +21,7 @@
 
 #include "ns_guard.h"
 
-ProcMountNS::ProcMountNS(int pid) {
+ProcMountNS::ProcMountNS(int pid) : target_ino_(0) {
   if (pid < 0)
     return;
 
@@ -38,6 +38,7 @@ ProcMountNS::ProcMountNS(int pid) {
   if (fstat(target_fd, &target_stat) != 0)
     return;
 
+  target_ino_ = target_stat.st_ino;
   if (self_stat.st_ino == target_stat.st_ino)
     // Both current and target Process are in same mount namespace
     return;
--- a/src/cc/usdt.cc
+++ b/src/cc/usdt.cc
@@ -15,6 +15,7 @@
  */
 #include <cstring>
 #include <sstream>
+#include <unordered_set>
 
 #include <fcntl.h>
 #include <sys/types.h>
@@ -250,15 +251,6 @@ Probe *Context::get(const std::string &probe_name) {
   return nullptr;
 }
 
-bool Context::generate_usdt_args(std::ostream &stream) {
-  stream << USDT_PROGRAM_HEADER;
-  for (auto &p : probes_) {
-    if (p->enabled() && !p->usdt_getarg(stream))
-      return false;
-  }
-  return true;
-}
-
 bool Context::enable_probe(const std::string &probe_name,
                            const std::string &fn_name) {
   if (pid_stat_ && pid_stat_->is_stale())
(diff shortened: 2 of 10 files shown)


drzaeus77:
[buildbot, add to whitelist]

drzaeus77:
[buildbot, test this please]

drzaeus77:
@yonghong-song The failure might be related to this change.

yonghong-song:
@drzaeus77 thanks for checking the result. Yes, forgot to change event name for detaching uprobes. Fixed and test results look fine now.

goldshtn:
Sorry, why can't we combine the enabled probes from all USDT contexts passed in, and generate code for their union rather than just the first one?

yonghong-song:
This can be done although I have not done it yet. My original thinking is
to
avoid over-engineering. Any known use case? Or I just assume somebody
will find it necessary and go ahead implementing ☺

palmtenor:
One advantage of @goldshtn's approach is that we would be able to move the logic to `usdt.cc` library and it would be easily shared between C++ and Python interface, so that when we add the PID based USDT interface in C++ (which @romain-intel is working on), we don't have to duplicate the `mnt_exe_set` logic again.

yonghong-song:
Make sure. Will have a revision to push most of helpers into C++ soon.

goldshtn:
@yonghong-song I don't have a specific use case for this within my tools, but I imagine it would be extremely misleading if someone runs into a situation where the first USDT context enables just one probe and the second USDT context enables another probe but doesn't get code generation for its arguments. The consolidation seems more natural and fairly easy to implement.

yonghong-song:
@goldshtn I agreed with you. I have implemented to allow flexible probe enabling for different pid's. Could you take a look at the pull request? Thanks!

src/cc/usdt.cc
```diff
@@ -345,13 +352,32 @@ int bcc_usdt_enable_probe(void *usdt, const char *probe_name,
...
   std::ostringstream stream;
-  if (!ctx->generate_usdt_args(stream))
-    return nullptr;
+
+  stream << USDT::USDT_PROGRAM_HEADER;
+  // Generate genargs codes for an array of USDT Contexts.
+  //
+  // Each mnt_point + cmd_bin_path + probe_provider + probe_name
+  // uniquely identifies a probe.
+  std::set<std::string> generated_probes;
```
goldshtn: This is a bit nitpicky but why not `unordered_set`? I don't think we need the order.

src/cc/usdt.h
```diff
@@ -218,12 +219,13 @@ class Context {
   optional<int> pid() const { return pid_; }
   bool loaded() const { return loaded_; }
   size_t num_probes() const { return probes_.size(); }
+  std::string cmd_bin_path() { return cmd_bin_path_; }
```
goldshtn: `const std::string &cmd_bin_path() const { return cmd_bin_path_; }`?

src/cc/usdt.h
```diff
@@ -218,12 +219,13 @@ class Context {
...
+  ino_t inode() { return mount_ns_instance_->target_ino(); }
```
goldshtn: const?

yonghong-song:
@goldshtn comments addressed. Could you take a look again?

src/cc/usdt.h
```diff
@@ -218,12 +219,13 @@ class Context {
   optional<int> pid() const { return pid_; }
   bool loaded() const { return loaded_; }
   size_t num_probes() const { return probes_.size(); }
+  const std::string & cmd_bin_path() { return cmd_bin_path_; }
+  const ino_t inode() { return mount_ns_instance_->target_ino(); }
```
goldshtn: Sorry, but I think I suggested these functions can be `const` - the rest of the API follows this convention.

tests/python/test_usdt2.py
```diff
@@ -0,0 +1,171 @@
...
+        b["event6"].open_perf_buffer(print_event6)
+
+        # three iterations to make sure we get some probes and have time to process them
+        for i in range(3):
+            b.kprobe_poll()
+
+        # note that event1 and event4 do not really fire, so their state should be 0
+        self.assertTrue(self.evt_st_1 == 1 and self.evt_st_2 == 1 and
+                        self.evt_st_3 == 0 and self.evt_st_4 == 0 and
+                        self.evt_st_5 == 1 and self.evt_st_6 == 1)
```
goldshtn: IMHO it's a bit cleaner to do separate assertions for each condition with `assertEqual`, so that if the assertion fails in a test, we know which part failed.

src/cc/ns_guard.cc
```diff
@@ -22,6 +22,7 @@
 #include "ns_guard.h"
 
 ProcMountNS::ProcMountNS(int pid) {
+  target_ino_ = 0;
```
goldshtn: Initializer list?

goldshtn:
Thanks; I missed a couple of nits in my first review, so added comments. Thanks for your patience.

src/cc/usdt.h
yonghong-song: Oh. Now I know what you mean, to mark the function as the const, not just return value. That makes sense. Will update.

yonghong-song:
@goldshtn No problem. Your comments make sense and just addressed all of them. Please take a look. Thanks!

PR #1268, opened 2017-07-28
Simple futex contention bcc script

gdankel:
This is the starting point for a more comprehensive lock analysis
tool ala Solaris' lockstat.

This initial version only traces the futex system call and measures
the time blocked or acquiring futex related kernel spinlocks due
to user-space lock contention.

Report results per process and per lock.

Several shortcomings of this approach:
* Only captures contention in locking constructs using futexes.
  Applications and libraries might only use futexes as a last resort.
* Does not distinguish between mutexes, conditional variables and
  semaphores etc.
* Does not capture user-space contention or performance impact due
  to cacheline bouncing.
* Etc etc. You get the idea.

### DIFF
--- a/tools/lockstat.py
+++ b/tools/lockstat.py
@@ -1,0 +1,232 @@
+#!/usr/bin/env bcc-py
+#
+# lockstats Trace and display lock contention stats
+#
+# USAGE: lockstats
+
+# Licensed under the Apache License, Version 2.0 (the "License")
+# 28-Jul-2017   Gisle Dankel   Created this.
+
+from bcc import BPF
+from ctypes import c_int
+from time import sleep
+from datetime import datetime
+import argparse
+import subprocess
+import os
+
+class Lock(object):
+    def __init__(self):
+        self.contention_count = 0
+        self.elapsed_blocked = 0
+        self.thread_count = 0
+
+    def update(self, count, block_time):
+        self.contention_count += count
+        self.elapsed_blocked += block_time
+        self.thread_count += 1
+
+
+examples = """
+EXAMPLES:
+
+./lockstats -p $(pidof allocs)
+        Trace allocations and display a summary of "leaked" (outstanding)
+        allocations every 5 seconds
+./lockstats -p $(pidof allocs) -t
+        Trace allocations and display each individual allocator function call
+./lockstats -ap $(pidof allocs) 10
+        Trace allocations and display allocated addresses, sizes, and stacks
+        every 10 seconds for outstanding allocations
+./lockstats -c "./allocs"
+        Run the specified command and trace its allocations
+./lockstats
+        Trace allocations in kernel mode and display a summary of outstanding
+        allocations every 5 seconds
+./lockstats -o 60000
+        Trace allocations in kernel mode and display a summary of outstanding
+        allocations that are at least one minute (60 seconds) old
+./lockstats -s 5
+        Trace roughly every 5th allocation, to reduce overhead
+"""
+
+description = """
+Trace kernel futex events.
+These often occur because of lock contention, e.g. involving a pthread_mutex.
+This script resemblers the following SystemTap example:
+https://sourceware.org/systemtap/SystemTap_Beginners_Guide/futexcontentionsect.html
+"""
+
+parser = argparse.ArgumentParser(description=description,
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        epilog=examples)
+parser.add_argument("-p", "--pid", type=int, default=-1,
+        help="the PID to trace; if not specified, trace all")
+parser.add_argument("-t", "--trace", action="store_true",
+        help="print trace messages for each futex enter/exit")
+parser.add_argument("interval", nargs="?", default=5, type=int,
+        help="interval in seconds to print summary")
+parser.add_argument("count", nargs="?", type=int,
+        help="number of times to print the report before exiting")
+parser.add_argument("-c", "--command",
+        help="execute and trace the specified command")
+
+args = parser.parse_args()
+
+pid = args.pid
+command = args.command
+interval = args.interval
+num_prints = args.count
+trace_all = args.trace
+
+if command is not None:
+        print("Executing '%s' and tracing the resulting process." % command)
+        pid = run_command_get_pid(command)
+
+bpf_source = """
+#include <uapi/linux/futex.h>
+#include <uapi/linux/ptrace.h>
+#include <linux/sched.h>
+#include <linux/time.h>
+
+struct comm_t {
+    char name[TASK_COMM_LEN];
+};
+
+struct lock_key_t {
+        u64 uaddr;
+        u32 pid;
+        u32 tgid;
+};
+
+struct lock_info_t {
+        u64 elapsed_blocked;
+        u64 contention_count;
+};
+
+
+BPF_HASH(pid_lock, u32, u64);
+BPF_HASH(pid_blocktime, u32, u64);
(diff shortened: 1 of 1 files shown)


tools/lockstat.py
```diff
@@ -0,0 +1,228 @@
...
+
+// FIXME: Should attach to sys_enter_futex and sys_exit_futex tracepoints here,
+// but that does not currently work
+int sys_futex_enter(struct pt_regs *ctx, u32 *uaddr, int op, u32 val,
+                    struct timespec *utime, u32 *uaddr2, u32 val3) {
+        int cmd = op & FUTEX_CMD_MASK;
+        if (cmd != FUTEX_WAIT)
+                return 0;
+
+        u32 pid = bpf_get_current_pid_tgid();
```
yonghong-song: For kprobe, passing pid to kernel does not guarantee that the bpf program only gets called with that pid. This is because kprobe is tricky, sometimes you do not have a user process associated at all. The same could be true for kernel tracepoints. 

Therefore, if user specifies a pid, the best practice is to place a check in the beginning of function
   if (pid != PREDEFINED_PID)
     return 0;

gdankel: OK curious... I can understand why you wouldn't have a pid at all points (e.g. h/w interrupt, network rx), but what's the reason it's not available sometimes in this case? It's a system call, so has to involve a process?

yonghong-song: I think kernel probably just want to be uniform for kprobes and regular kernel tracepoints.

syscalls (sys_enter_* and sys_exit_* tracepoints) are special processed in kernel. Not sure whether it honors pid or not. I am working on a patch to enable sys_enter/exit_* bpf tracepoints and will comment later once I found it.

palmtenor: I think these are two different topics we are discussing here:
- For filtering on PID (more context: #1278), we'll need to do so from BPF program. i.e., as @yonghong-song mentioned:
```
if (pid != PREDEFINED_PID)
    return 0;
```
- For availability of PID and COMM in kprobes, from my experience (please correct me if I'm wrong) I would always get valid PID number (might be 0) and COMM name (might be `kswapper` / `kworker`) even for those Kernel function possibly running not from a user Process.

gdankel: Right, makes sense. I obviously hadn't tested that case - there's no code in there to deal with -p option :/
I will push an updated version.

gdankel:
Yonghong: run_command_get_pid is used in the latest commit :) I've now tested the different options a fair amount.

gdankel:
Note to reviewers: I posted this mainly to get both high-level feedback and detailed feedback (this is my first attempt at using bcc / bpf).
I have a good idea about which direction to take this tool, and this is only the first small warm-up step to get used to the process.
If anyone has comments / suggestions / ideas, now is a good time to chime in.
One obvious question is, why has such a tool not yet been developed? Maybe there's not much perceived value on top of what already exists?

yonghong-song:
Sorry, I may miss the usage of run_command_get_pid....

The code looks good to me. All tools in tools directories have corresponding example.txt to show how the tool should be used. You mentioned that there will be additional patches down the road to make the tool more useful. I guess that at some point, once you feel the tool is ready for production run, you should add the example file.

gdankel:
The call was in the first commit, I added the function in the third commit, so that's why you didn't see it :)

brendangregg:
Yes, we need a lockstat tool. In general, please try to follow https://github.com/iovisor/bcc/blob/master/CONTRIBUTING-SCRIPTS.md . 

If you think so far it makes a good programming example of bcc/BPF, then I'd put it under /examples/tracing.

If you feel it's useful and safe for people to run, as root, on production Netflix/Facebook instances (Netflix installs all of these by default on all our instances), then it can be in /tools, but it also needs an _examples.txt and man page. And the man page should pull no punches regarding overhead: have an "OVERHEAD" section that gives the end user an expectation before they run it. Also list caveats in the man page (ilke you did above).

If it's just unfinished and in development, it may not need to be in bcc yet. You can file an issue as a RFC and link to where it is, to get our feedback that way. (I've put my in-development stuff here https://github.com/brendangregg/BPF-tools , which reminds me I should upload what I'm currently working on...)

brendangregg:
To answer why lockstat hasn't been done yet. No particular reason. I've been writing/adding tools as I hit issues.

gdankel:
Brendan, thanks - CONTRIBUTING-SCRIPTS.md is very useful.
The aim is to make this a production tool, but only after a couple of more commits to flesh it out.
I might follow your in-development approach...

goldshtn:
[buildbot, ok to test]

goldshtn:
@gdankel Are you planning to add more commits?

gdankel:
Yes, planning on more commits. Down my list of priorities atm. I will post a new patch when I have something solid and slightly more interesting.

joelagnel:
This is quite interesting. In Android we use futexes for conditional vars quite a bit. It then becomes quite important that something waiting isn't falsely considered waiting on a lock. I see you mentioned this in the caveats. One (non-generic) way is to look for signatures in the stack to find usage. Does the equivalent Solaris tool address this issue?

gdankel:
@joelagnel Yes filtering by signatures in the stack is the way I do it now. I am thinking of progressing this in the following way:

1) Create a bcc script that does minimal user-space processing and just dumps one or more output files with varying level of detail (depending on what is requested). This output will contain a bunch of stats per lock and per stack (including requested number of stack frame symbols). I will also likely separate out the BPF C code into its own file so it can be used by non-bcc code.

2) Create a couple of tools to analyze the data. One will be similar to lockstat. Another might generate output for other visualization tools. These will have many more options to allow processing and viewing the data in different ways.

3) Add user-space profiling capabilities. Tracing user-level locking functions is painful but I've been experimenting with sampling locked loads using the DataLA facility to identify locked loads and then also sampling cycles (precise mode) to assign costs to these.

joelagnel:
@gdankel Cool, looking forward to seeing it and also contributing / applying to Android kernels. `DataLA` I think is Intel specific. Can you not just use perf_events for sampling? Perhaps I don't understand point 3. enough without more background on the profiling you're doing.

gdankel:
@joelagnel Yes unfortunately DataLA is Intel specific, I'm not aware of a similar feature for ARM, but I'm not too familiar with ARM PMUs. Something similar can probably be done on ARM, but without per-lock breakdown. DataLA gives you the the address of the locks when sampling, so it gives you an idea of which particular locks contribute most to synchronization costs. But in many cases, the call stack is sufficient.

And yes, sampling is done using perf_event_open, and using BPF to read the values. There is some advantage to doing it all in the BPF program, but it can also be done separately and combined in post-processing.

joelagnel:
Sounds great. For address of locks, can the `uaddr` and `PID` combination not suffice to identify a lock? I guess it may not for inter-process futexes but I'm not sure how common that is in practice.
I was talking about stack signatures more for the purpose of identifying **how** a futex is used (cond.var vs lock, etc).

In any case, if you have some early code for checking common stack signatures in-order to identify locking scenarios, I would love to see it. A tool like this is high on my priority list, so I will start hacking on it soon enough. I could also just start using this PR as a starting point.
thanks!

gdankel:
@joelagnel sure, uaddr and pid is fine. Uaddr is recorded from sys_futex tracing but then also combined with the DataLA samples for a more complete picture.

I think we're talking about the same thing for stack signatures. I also use them to filter out conditional variables etc. But the code I have for that is pretty rudimentary right now, not much more than looking for "lock" and a couple of other things in the first five or so user frames. It should be customizable via an argument or config file.

I'll try to drop some code next week.

PR #1269, opened 2017-07-28
tcptop: Cleanup and fix

benthaman:


brendangregg:
thanks; so there's similar code in the offcputime.py program (which I happen to be editing):

```Python
# arg validation
def positive_int(val):
    try:
        ival = int(val)
    except ValueError:
        raise argparse.ArgumentTypeError("must be an integer")

    if ival < 0:
        raise argparse.ArgumentTypeError("must be positive")
    return ival

def positive_nonzero_int(val):
    ival = positive_int(val)
    if ival == 0:
        raise argparse.ArgumentTypeError("must be nonzero")
    return ival
```

I guess it doesn't matter for this patch, but maybe in the future we want to put this in bcc so people don't have to keep defining them.

brendangregg:
[buildbot, ok to test]

PR #1270, opened 2017-07-28
have uniform uprobe event names for python and C++

yonghong-song:
This is a followup change for previous commit 0ba15075fc5f
(permit multiple pids attaching to the same probe).
That commit changes the event name for python uprobe API
in order to permit multiple processes attaching to the
same uprobe point. The C++ uprobe event name remains
unchanged and thus a descrepancy.

This patch add changes to C++ side and also make python uretprobe
having the same naming convention.

Originally I experimented to put the common code to generate
event names in libbpf.c. But doing this seems more klunky
than simplicity of C++ and Python (mostly due to languages). 
So I stick to the current design.

No need to add pid to kprobe event names as kprobe bpf invocation
will ignore pid anyway.

### DIFF
--- a/src/cc/BPF.cc
+++ b/src/cc/BPF.cc
@@ -199,7 +199,7 @@ StatusTuple BPF::attach_uprobe(const std::string& binary_path,
   uint64_t offset;
   TRY2(check_binary_symbol(binary_path, symbol, symbol_addr, module, offset));
 
-  std::string probe_event = get_uprobe_event(module, offset, attach_type);
+  std::string probe_event = get_uprobe_event(module, offset, attach_type, pid);
   if (uprobes_.find(probe_event) != uprobes_.end())
     return StatusTuple(-1, "uprobe %s already attached", probe_event.c_str());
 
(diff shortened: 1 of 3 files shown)


yonghong-song:
cc @palmtenor

src/cc/BPF.h
```diff
@@ -142,7 +143,7 @@ class BPF {
   std::string get_kprobe_event(const std::string& kernel_func,
                                bpf_probe_attach_type type);
   std::string get_uprobe_event(const std::string& binary_path, uint64_t offset,
-                               bpf_probe_attach_type type);
+                               bpf_probe_attach_type type, pid_t pid);
```
palmtenor: Also add a default value here?

palmtenor:
Thanks for the patch! Looks good in general~

yonghong-song:
> Also add a default value here?

Look at the typical code path:
StatusTuple BPF::attach_uprobe(const std::string& binary_path,
                               const std::string& symbol,
                               const std::string& probe_func,
                               uint64_t symbol_addr,
                               bpf_probe_attach_type attach_type,
                               pid_t pid, int cpu, int group_fd,
                               perf_reader_cb cb, void* cb_cookie) {
  std::string module;
  uint64_t offset;
  TRY2(check_binary_symbol(binary_path, symbol, symbol_addr, module, offset));

  std::string probe_event = get_uprobe_event(module, offset, attach_type);

attach_uprobe has "pid" as one of parameter although its default value is -1.
get_uprobe_event should just have non-default pid to avoid potential mistake that
users passing in non -1 pid in attach_uprobe, but use the default pid
in get_uprobe_event.

src/cc/BPF.cc
```diff
@@ -551,10 +552,13 @@ BPFStackTable BPF::get_stack_table(const std::string& name,
...
 
 std::string BPF::get_uprobe_event(const std::string& binary_path,
-                                  uint64_t offset, bpf_probe_attach_type type) {
+                                  uint64_t offset, bpf_probe_attach_type type,
+                                  pid_t pid) {
   std::string res = attach_type_prefix(type) + "_";
   res += sanitize_str(binary_path, &BPF::uprobe_path_validator);
   res += "_0x" + uint_to_hex(offset);
+  if (pid != -1)
+    res += "_" + pid;
```
drzaeus77: This must be `std::to_string(pid)`.

yonghong-song:
@drzaeus77 made the change as suggested. Please take a look again. Thanks!

PR #1273, opened 2017-07-30
add MEMBER_(ADDR|READ) helper macros

yadutaf:
These helper macros are handy when bcc is not yet able to expand ``dst = src->member;`` constructs automatically. With these macro, such constructs can be written as ``MEMBER_READ(&dst, src, member)`` which preserve readability and functionality. It is also generic enough to fit a large number of use cases.

This is an adaptation of the macro introduced in https://blog.yadutaf.fr/2017/07/28/tracing-a-packet-journey-using-linux-tracepoints-perf-ebpf/. They may be of use to other users of bcc, hence this PR.

If this sounds good to you, I'll add some documentation to the reference guide before merging.

### DIFF
--- a/src/cc/export/helpers.h
+++ b/src/cc/export/helpers.h
@@ -588,5 +588,22 @@ int tracepoint__##category##__##event(struct tracepoint__##category##__##event *
             bpf_probe_read((void *)dst, __length, (char *)args + __offset); \
         } while (0);
 
+#define MEMBER_ADDR(src, field)                                             \
+        ({                                                                  \
+            void* __ret;                                                    \
+            __ret = (void*) (((char*)src) + offsetof(typeof(*src), field)); \
+            __ret;                                                          \
+        })
+
+#define MEMBER_READ(dst, src, field)    \
+        do{                             \
+            bpf_probe_read(             \
+                dst,                    \
+                sizeof(src->field),     \
+                MEMBER_ADDR(src, field) \
+            );                          \
+        } while(0)
+
+
 #endif
 )********"
--- a/tests/python/test_tracepoint.py
+++ b/tests/python/test_tracepoint.py
@@ -64,5 +64,44 @@ class TestTracepointDataLoc(unittest.TestCase):
         self.assertTrue("/bin/ls" in [v.filename.decode()
                                       for v in b["execs"].values()])
 
+@unittest.skipUnless(kernel_version_ge(4,7), "requires kernel >= 4.7")
+class TestTracepointDataLoc(unittest.TestCase):
+    def test_tracepoint_data_loc(self):
+        text = """
+        #include <net/inet_sock.h>
+
+        BPF_HASH(pkts, u32, u64);
+        TRACEPOINT_PROBE(net, net_dev_queue) {
+            u64 count = 0;
+
+            // Get sk_buff fro args
+            struct sk_buff* skb = (struct sk_buff*)args->skbaddr;
+
+            // Get skb->dev
+            struct net_device *dev;
+            MEMBER_READ(&dev, skb, dev);
+
+            // Get dev->nd_net.net
+            struct net* net;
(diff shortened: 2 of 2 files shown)


yadutaf:
Any news on this PR ?

yadutaf:
Hi !  Thanks for the review ! I've added a test case in ``tests/python/test_tracepoint.py`` based on the linked blog post. It's not as simple as I'd like but it demonstrates that both macros work as expected.

I've run the tests using ``ctest -R test_tracepoint -VV`` and manually made sure that the netns was consistent with the value in ``/proc/self/ns/net``.

src/cc/export/helpers.h
```diff
@@ -588,5 +588,22 @@ int tracepoint__##category##__##event(struct tracepoint__##category##__##event *
             bpf_probe_read((void *)dst, __length, (char *)args + __offset); \
         } while (0);
 
+#define MEMBER_ADDR(src, field)                                             \
+        ({                                                                  \
+            void* __ret;                                                    \
+            __ret = (void*) (((char*)src) + offsetof(typeof(*src), field)); \
+            __ret;                                                          \
+        })
```
yonghong-song: Is this macro simpler? ({ (void *)&src->field }). If this is the case, do we really need this macro?

src/cc/export/helpers.h
```diff
@@ -588,5 +588,22 @@ int tracepoint__##category##__##event(struct tracepoint__##category##__##event *
...
+
+#define MEMBER_READ(dst, src, field)    \
+        do{                             \
+            bpf_probe_read(             \
+                dst,                    \
+                sizeof(src->field),     \
+                MEMBER_ADDR(src, field) \
+            );                          \
+        } while(0)
+
```
yonghong-song: If my assessment about MEMBER_ADDR is correct, a direct bpf_probe_read is not complicated and using a macro is less justified.

tests/python/test_tracepoint.py
```diff
@@ -64,5 +64,44 @@ def test_tracepoint_data_loc(self):
         self.assertTrue("/bin/ls" in [v.filename.decode()
                                       for v in b["execs"].values()])
 
+@unittest.skipUnless(kernel_version_ge(4,7), "requires kernel >= 4.7")
+class TestTracepointDataLoc(unittest.TestCase):
+    def test_tracepoint_data_loc(self):
```
yonghong-song: class name TestTracepointDataLoc has been used before and you need to choose a different name.

yadutaf:
Here is the test case without the macros. It does work and has the same number of lines:

```c
        #include <net/inet_sock.h>

        BPF_HASH(pkts, u32, u64);
        TRACEPOINT_PROBE(net, net_dev_queue) {
            u64 count = 0;

            // Get sk_buff fro args
            struct sk_buff* skb = (struct sk_buff*)args->skbaddr;

            // Get skb->dev
            struct net_device *dev;
            bpf_probe_read(&dev, sizeof(skb->dev), (void *)&skb->dev);

            // Get dev->nd_net.net
            struct net* net;
            possible_net_t *skc_net = &dev->nd_net;
            bpf_probe_read(&net, sizeof(skc_net->net), (void *)&skc_net->net);

            // Get net->ns.inum
            u32 netns;
            struct ns_common* ns = (struct ns_common*)&net->ns;
            bpf_probe_read(&netns, sizeof(ns->inum), (void *)&ns->inum);

            // Increment packet count for netns
            u64 *existing = pkts.lookup_or_init(&netns, &count);
            (*existing)++;

            return 0;
        }
```

I have to admit I feel frustrated. I'm pretty sure it did not work when I initially tested it. But I may as well just have missed it. Or the rebase did help. Dunno.

Anyway, thanks again for you time. I still find the intent more obvious using the macros, and I'll keep using them in my code. That said, I'd given the smaller benefit than expected, I'd understand if the PR was closed :)

Have a nice day.

4ast:

PR #1275, opened 2017-07-31
Fix netlink pid check

tmakita:
The first patch fixes bogus check for netlink pid in bpf_attach_xdp, by which bpf_attach_xdp incorrectly ended up with failure.
The second patch improves printf format of pid.


drzaeus77:
[buildbot, ok to test]

PR #1277, opened 2017-08-01
add --state to offcputime

brendangregg:
Useful for a number of reasons: isolate TASK_UNINTERRUPTIBLE sleeps, measure __TASK_TRACED overhead, etc.

### DIFF
--- a/tools/offcputime.py
+++ b/tools/offcputime.py
@@ -88,11 +88,16 @@ parser.add_argument("-M", "--max-block-time", default=(1 << 64) - 1,
     type=positive_nonzero_int,
     help="the amount of time in microseconds under which we " +
          "store traces (default U64_MAX)")
+parser.add_argument("--state", default=999,
+    type=positive_int,
+    help="filter on this thread state bitmask (eg, 2 == TASK_UNINTERRUPTIBLE" +
+         ") see include/linux/sched.h")
 args = parser.parse_args()
 if args.pid and args.tgid:
     parser.error("specify only one of -p and -t")
 folded = args.folded
 duration = int(args.duration)
+debug = 0
 
 # signal handler
 def signal_ignore(signal, frame):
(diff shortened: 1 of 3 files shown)


tools/offcputime.py
```diff
@@ -177,7 +182,15 @@ def signal_ignore(signal, frame):
 else:
     thread_context = "all threads"
     thread_filter = '1'
+if args.state == 0:
+    state_filter = 'prev->state == 0'
+elif args.state != 999:
```
yonghong-song: in linux/sched.h, we have
#define TASK_RUNNING                    0
#define TASK_INTERRUPTIBLE              1
#define TASK_UNINTERRUPTIBLE            2
#define __TASK_STOPPED                  4
#define __TASK_TRACED                   8
/* Used in tsk->exit_state: */
#define EXIT_DEAD                       16
#define EXIT_ZOMBIE                     32
#define EXIT_TRACE                      (EXIT_ZOMBIE | EXIT_DEAD)
/* Used in tsk->state again: */
#define TASK_DEAD                       64
#define TASK_WAKEKILL                   128
#define TASK_WAKING                     256
#define TASK_PARKED                     512
#define TASK_NOLOAD                     1024
#define TASK_NEW                        2048
#define TASK_STATE_MAX                  4096

why you choose 999 as the magic number? Maybe a big number like (unsigned int)-1?

markdrayton: We don't need a special value here — if we drop the default `state` will either be a positive integer or be `None`. This code can then be:

```python
if args.state == 0:
    state_filter = 'prev->state == 0'
elif args.state:
    # these states are sometimes bitmask checked
    state_filter = 'prev->state & %d' % args.state
else:
    state_filter = '1'
```

brendangregg: ah, thanks @markdrayton @yonghong-song !

brendangregg:
Yes, I wanted to allow --state=0, which matches TASK_RUNNING (since this is offcputime, that should measure largely involuntary context switches).

PR #1279, opened 2017-08-02
Remove obsolete LUA_GLOBALSINDEX

NanXiao:
Since `LUA_GLOBALSINDEX` is obsolete from `Lua 5.2`, use `lua_getglobal` function instead.


drzaeus77:
[buildbot, ok to test]

PR #1280, opened 2017-08-02
Remove unused header file in pidpersec.py

NanXiao:


drzaeus77:
Hmm, the inclusion of ptrace.h was not meant to be automatic, but since https://github.com/iovisor/bcc/commit/0ba15075fc5f621e21fc68634816468d97bda5f1 it has been implicitly added as part of `__init__.py`:

```
usdt_text = lib.bcc_usdt_genargs(ctx_array, len(usdt_contexts)).decode()
text = usdt_text + text
```

In the empty usdt_contexts case (typical), this results in the single line of `#include <uapi/include/ptrace.h>`. I would rather that behavior be reverted and the explicit inclusion be restored. The reason being that in the case of non-tracing (e.g. networking) compilation, extraneous header files should not be included implicitly.

I would prefer that this PR be closed, but I'll leave it open for a small time in case there is further discussion.

NanXiao:
@drzaeus77 Thanks for your explanation! You can close it, thx!

NanXiao:
BTW, I am not a nitpicker, but find following code in [pidpersec.py](https://github.com/iovisor/bcc/blob/8e26b4916da5efd500eaed0fe5159c8cc8f981bf/tools/pidpersec.py#L29):    

`BPF_ARRAY(stats, u64, S_MAXSTAT + 1);`

Shouldn't 

`BPF_ARRAY(stats, u64, S_MAXSTAT);` is enough?

Thanks!

drzaeus77:
@NanXiao Yeah it should be enough.

NanXiao:
@drzaeus77 Thanks very much for your response! I will close this issue.

yonghong-song:
There are several files like this. Looks like they have the same issue.
-bash-4.3$ egrep -r S_MAXSTAT
dcstat.py:    S_MAXSTAT
dcstat.py:BPF_ARRAY(stats, u64, S_MAXSTAT + 1);
pidpersec.py:    S_MAXSTAT
pidpersec.py:BPF_ARRAY(stats, u64, S_MAXSTAT + 1);
vfsstat.py:    S_MAXSTAT
vfsstat.py:BPF_ARRAY(stats, u64, S_MAXSTAT + 1);
-bash-4.3$

Issue #1283, opened 2017-08-04
New Enhancement Request - tcptracer tool


shehzadkhawar:
Currently tcptracer output is like this. 
# ./tcptracer
Tracing TCP established connections. Ctrl-C to end.
T  PID    COMM             IP SADDR            DADDR            SPORT  DPORT
C  28943  telnet           4  192.168.1.2      192.168.1.1      59306  23
C  28818  curl             6  [::1]            [::1]            55758  80
X  28943  telnet           4  192.168.1.2      192.168.1.1      59306  23
A  28817  nc               6  [::1]            [::1]            80     55758
X  28818  curl             6  [::1]            [::1]            55758  80
X  28817  nc               6  [::1]            [::1]            80     55758
A  28978  nc               4  10.202.210.1     10.202.109.12    8080   59160
X  28978  nc               4  10.202.210.1     10.202.109.12    8080   59160

# Question (Enhancement Request): 
Is it possible that we can add a timestamp on each line as well? This way some analytics on a time-based scale can be extracted out of it. 

Cheers,
K.

iaguis:
I filed a PR implementing this: https://github.com/iovisor/bcc/pull/1284

shehzadkhawar:
Just checked, this is great. There is a traffic based DDoS scenario (or any other high load situation), in which millisecond granularity isn't much of a help. Is it possible that we can get it as in micro seconds or perhaps nanoseconds as options?

iaguis:
> Just checked, this is great. There is a traffic based DDoS scenario (or any other high load situation), in which millisecond granularity isn't much of a help. Is it possible that we can get it as in micro seconds or perhaps nanoseconds as options?

I added a commit that shows nanoseconds when `-v` is passed.

PR #1284, opened 2017-08-04
tools/tcptracer: add timestamp option

iaguis:
Similar to other tools from the tcp* family.

Fixes https://github.com/iovisor/bcc/issues/1283


drzaeus77:
[buildbot, ok to test]

PR #1285, opened 2017-08-06
Trace external pointers through maps

### LINKED ISSUE #253: invalid mem access when navigating struct sock
Here is a program that I thought should work:

``` Python
#!/usr/bin/python
from bcc import BPF
bpf_text = """
#include <net/sock.h>
#include <bcc/proto.h>

BPF_HASH(currsock, u32, struct sock *);

int kprobe__tcp_v4_connect(struct pt_regs *ctx, struct sock *sk,
    struct sockaddr *uaddr, int addr_len)
{
    currsock.update(&pid, &sk);
    return 0;
};

int kretprobe__tcp_v4_connect(struct pt_regs *ctx)
{
    struct sock **skpp;
    skpp = currsock.lookup(&pid);
    if (skpp) {
        struct sock *skp = *skpp;
        bpf_trace_printk("addr: %lx\\n", skp->__sk_common.skc_daddr);
        currsock.delete(&pid);
    }
    return 0;
}
"""
b = BPF(text=bpf_text)
print("running"); b.trace_print(); exit()
```

But won't run:

``` Shell
# ./tcpv4connect
bpf: Permission denied
0: (85) call 14
1: (63) *(u32 *)(r10 -4) = r0
2: (18) r1 = 0xe2f31180
4: (bf) r2 = r10
5: (07) r2 += -4
6: (85) call 1
7: (15) if r0 == 0x0 goto pc+18
 R0=map_value(ks=4,vs=8) R10=fp
8: (79) r1 = *(u64 *)(r0 +0)
9: (b7) r2 = 0
10: (73) *(u8 *)(r10 -12) = r2
11: (b7) r2 = 175664165
12: (63) *(u32 *)(r10 -16) = r2
13: (18) r2 = 0x75746572
15: (7b) *(u64 *)(r10 -24) = r2
16: (61) r3 = *(u32 *)(r1 +0)
R1 invalid mem access 'inv'

  File "./tcpv4connect.2", line 30, in <module>
    b = BPF(text=bpf_text)
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 383, in __init__
    self._trace_autoload()
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 610, in _trace_autoload
    fn = self.load_func(func_name, BPF.KPROBE)
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 418, in load_func
    raise Exception("Failed to load BPF program %s" % func_name)
Exception: Failed to load BPF program kretprobe__tcp_v4_connect
```

It works if I use this instead, in the kretprobe:

``` Python
[...]
        if (skpp) {
                struct sock *skp = *skpp;
                u32 daddr = 0;
                bpf_probe_read(&daddr, sizeof(daddr), &skp->__sk_common.skc_daddr);
                bpf_trace_printk("addr: %x\\n", daddr);
[...]
```

Is needing an explicit bpf_probe_read() a bug or is it supposed to be this way? If it is supposed to be this way, then we'll need to document the logic to make it easier to follow...

pchaigno:
The bcc rewriter currently traces external pointers using
`ProbeVisitor` in order to replace dereferences with calls to
`bpf_probe_read`. It is, however, unable to trace them through maps.

This pull request fixes this for simple (yet common) cases. Through a
first traversal of the Clang AST, `MapVisitor` looks for calls to
update (and insert) to find maps with an external pointer as value.
When `ProbeVisitor` runs, in a second time, it looks for calls to
lookup (and lookup_or_init). If the map was registered as having an
external pointer as value, the l-value of the lookup assignment is
marked as being an external pointer.

Two traversals of the Clang AST are needed because the update of a
map may happen after the lookup in the AST. Therefore, the first
traversal makes sure we inspect all updates before acting on lookups.
To implement this two-stage traversal without parsing the AST twice,
`ProbeConsumer` and `BTypeConsumer` now implement
`HandleTranslationUnit`, which is called after a whole translation
unit has been parsed.

This change should probably be considered a breaking API change. In
bcc, it breaks a few tools that were relying on the rewriter not
being able to replace dereferences. For instance, the
`bpf_probe_read(&saddr, sizeof(saddr), &skp->__sk_common.skc_rcv_saddr)`
expression in `examples/tracing/tcpv4connect.py` now returns an error.

Fixes #253.

/cc @drzaeus77

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -108,7 +108,36 @@ class ProbeSetter : public RecursiveASTVisitor<ProbeSetter> {
   set<Decl *> *ptregs_;
 };
 
-ProbeVisitor::ProbeVisitor(ASTContext &C, Rewriter &rewriter) : C(C), rewriter_(rewriter) {}
+// Traces maps with external pointers as values. 
+class MapVisitor : public RecursiveASTVisitor<MapVisitor> {
+ public:
+  explicit MapVisitor(set<Decl *> &m) : m_(m) {}
+  bool VisitCallExpr(CallExpr *Call) {
+    if (MemberExpr *Memb = dyn_cast<MemberExpr>(Call->getCallee()->IgnoreImplicit())) {
+      StringRef memb_name = Memb->getMemberDecl()->getName();
+      if (DeclRefExpr *Ref = dyn_cast<DeclRefExpr>(Memb->getBase())) {
+        if (SectionAttr *A = Ref->getDecl()->getAttr<SectionAttr>()) {
+          if (!A->getName().startswith("maps"))
+            return true;
+
+          if (memb_name == "update" || memb_name == "insert") {
+            if (ProbeChecker(Call->getArg(1), ptregs_).needs_probe()) {
+              m_.insert(Ref->getDecl());
+            }
+          }
+        }
+      }
+    }
+    return true;
+  }
+  void set_ptreg(Decl *D) { ptregs_.insert(D); }
+ private:
+  set<Decl *> &m_;
+  set<clang::Decl *> ptregs_;
+};
+
+ProbeVisitor::ProbeVisitor(ASTContext &C, Rewriter &rewriter, set<Decl *> &m) :
+  C(C), rewriter_(rewriter), m_(m) {}
 
 bool ProbeVisitor::VisitVarDecl(VarDecl *Decl) {
   if (Expr *E = Decl->getInit()) {
@@ -141,6 +170,25 @@ bool ProbeVisitor::VisitBinaryOperator(BinaryOperator *E) {
   if (ProbeChecker(E->getRHS(), ptregs_).is_transitive()) {
     ProbeSetter setter(&ptregs_);
     setter.TraverseStmt(E->getLHS());
+  } else if (E->isAssignmentOp() && E->getRHS()->getStmtClass() == Stmt::CallExprClass) {
+    CallExpr *Call = dyn_cast<CallExpr>(E->getRHS());
+    if (MemberExpr *Memb = dyn_cast<MemberExpr>(Call->getCallee()->IgnoreImplicit())) {
+      StringRef memb_name = Memb->getMemberDecl()->getName();
+      if (DeclRefExpr *Ref = dyn_cast<DeclRefExpr>(Memb->getBase())) {
+        if (SectionAttr *A = Ref->getDecl()->getAttr<SectionAttr>()) {
+          if (!A->getName().startswith("maps"))
+            return true;
+
+          if (memb_name == "lookup" || memb_name == "lookup_or_init") {
+            if (m_.find(Ref->getDecl()) != m_.end()) {
+            // Retrieved an external pointer from a map, mark LHS as external pointer.
+              ProbeSetter setter(&ptregs_);
+              setter.TraverseStmt(E->getLHS());
+            }
+          }
+        }
+      }
+    }
   }
   return true;
 }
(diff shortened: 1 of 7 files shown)


brendangregg:
Looks like a good improvement.

src/cc/frontends/clang/b_frontend_action.cc
```diff
@@ -107,7 +107,36 @@ class ProbeSetter : public RecursiveASTVisitor<ProbeSetter> {
...
+          if (!A->getName().startswith("maps"))
+            return true;
+
+          if (memb_name == "update" || memb_name == "insert") {
+            if (ProbeChecker(Call->getArg(1), ptregs_).needs_probe()) {
+              m_.insert(Ref->getDecl());
+            }
+          }
+        }
+      }
```
yonghong-song: I do not fully understand why we need to track map.update or map.insert here?
It is totally possible that map.update and map.insert not in the same compilation unit, e.g., populated by user space or shared with another bpf program. Then we will not able to trace map value pointers below?

src/cc/frontends/clang/b_frontend_action.cc
```diff
@@ -140,6 +169,25 @@ bool ProbeVisitor::VisitBinaryOperator(BinaryOperator *E) {
...
+    CallExpr *Call = dyn_cast<CallExpr>(E->getRHS());
+    if (MemberExpr *Memb = dyn_cast<MemberExpr>(Call->getCallee()->IgnoreImplicit())) {
+      StringRef memb_name = Memb->getMemberDecl()->getName();
+      if (DeclRefExpr *Ref = dyn_cast<DeclRefExpr>(Memb->getBase())) {
+        if (SectionAttr *A = Ref->getDecl()->getAttr<SectionAttr>()) {
+          if (!A->getName().startswith("maps"))
+            return true;
+
+          if (memb_name == "lookup" || memb_name == "lookup_or_init") {
+            if (m_.find(Ref->getDecl()) != m_.end()) {
```
yonghong-song: Again not sure whether we need this m_ check or not. What if we do not check?

src/cc/frontends/clang/b_frontend_action.cc
```diff
@@ -773,7 +852,8 @@ void BFrontendAction::EndSourceFileAction() {
 unique_ptr<ASTConsumer> BFrontendAction::CreateASTConsumer(CompilerInstance &Compiler, llvm::StringRef InFile) {
   rewriter_->setSourceMgr(Compiler.getSourceManager(), Compiler.getLangOpts());
   vector<unique_ptr<ASTConsumer>> consumers;
-  consumers.push_back(unique_ptr<ASTConsumer>(new ProbeConsumer(Compiler.getASTContext(), *rewriter_)));
+  consumers.push_back(unique_ptr<ASTConsumer>(new MapConsumer(m_)));
+  consumers.push_back(unique_ptr<ASTConsumer>(new ProbeConsumer(Compiler.getASTContext(), *rewriter_, m_)));
```
yonghong-song: If we use handleTranslationUnit to post process ProbeConsumer and BTypeConsumer, I am wondering whether we need ProbeConsumer and BTypeConsumer registered as the BFrontendAction consumer. We can do post process at MapConsumer handleTranslationUnit?

Note that this implementation changes how each construct is processed by ProbeConsumer and BTypeConsumer.  The old implementation will call the processing function for each consumer for each construct at a time before moving to the next construct. The new implementation will process all constructs for one consumer before moving to the next consumer. 

Look at the ProbeConsumer and BTypeConsumer code, they are largely independent so it should be okay.

src/cc/frontends/clang/b_frontend_action.cc
pchaigno: Where would you track `map.update` and `map.insert` if not here?

Yes, if the updates/inserts are not in the same compilation unit, we can't trace map value pointers. I don't think there's much we can do in that case, at least without hints from developers.

src/cc/frontends/clang/b_frontend_action.cc
pchaigno: The m_ check makes sure the map has an external pointer as value. AFAIK, the only way to know that is to track them from the function arguments (what's already implemented) to the map.inserts/updates, remember that, and use it when we meet the map.lookups (what this pull request implements).

src/cc/frontends/clang/b_frontend_action.cc
pchaigno: > If we use handleTranslationUnit to post process ProbeConsumer and BTypeConsumer, I am wondering whether we need ProbeConsumer and BTypeConsumer registered as the BFrontendAction consumer. We can do post process at MapConsumer handleTranslationUnit?

So we'd have a single `Consumer` for all visitors? Sure :+1:

pchaigno:
@yonghong-song Thanks for the review and sorry for the delay to answer.

4ast:
Looks great. Probably needs rebase before merge?

src/cc/frontends/clang/b_frontend_action.cc
pchaigno: Done in the last commit.

pchaigno:
> Probably needs rebase before merge?

Done (with a new commit to address @yonghong-song's comment above).

PR #1287, opened 2017-08-06
Trace external pointers from helpers

pchaigno:
This pull request extends `ProbeChecker` to trace external pointers
from helpers. At this time, a single helper can return a kernel
pointer, `bpf_get_current_task`.

This change should probably be considered a breaking API change. In
bcc, it breaks a few tools that were relying on the rewriter not
being able to replace dereferences. For instance, the
`bpf_probe_read(&nsproxy, sizeof(nsproxy), &task->nsproxy)`
expression in `tools/mountsnoop.py` now returns an error.

/cc @drzaeus77

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -77,6 +77,9 @@ class ProbeChecker : public RecursiveASTVisitor<ProbeChecker> {
   }
   bool VisitCallExpr(CallExpr *E) {
     needs_probe_ = false;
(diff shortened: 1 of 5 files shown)


pchaigno:
The new test case is failing on the Ubuntu 16.04 Jenkins because `bpf_get_current_task` requires a kernel > 4.4. Should I remove the test case, add an Exception for this specific error, or try and detect the kernel version beforehand?

goldshtn:
I think our usual strategy is to mark the test case as requiring that specific kernel version. See the tools smoke tests for example.

pchaigno:
Ah! of cource, the smoke tests! Thanks @goldshtn!

PR #1288, opened 2017-08-06
Trace external pointers in parenthesized expressions

pchaigno:
Partially reverts #290, keeping the test case. With #1005, only
dereference operations are rewritten, removing the need for the
previous fix.

Reverting #290 allows bcc to rewrite more dereferences, as
highlighted in the new test case.

/cc @drzaeus77


pchaigno:
@drzaeus77 Anything I can do to help with the review? (add comments/tests?)

PR #1292, opened 2017-08-16
python - set attach_xdp's default flag value to 0

GabrielGanne:
Hi,

Following commit 9f3cab70d290d62f663d36369a441b0ee359e8d1 attach_xdp() requires a 3rd flag argument.
It caused me an error when I tried to launch some old code I had.
This just sets default flag value to 0 so that it is not explicitly required.

BR,


drzaeus77:
[buildbot, ok to test]

Issue #1291, opened 2017-08-14
RSS of BPF_STACK_TRACE programs


markdrayton:
Is it expected that increasing the size of a `BPF_STACK_TRACE` map would increase the RSS of the userspace component of a BPF program? E.g., given the following test:

```c
#include <stdio.h>
#include "bpf_common.h"

int main(int argc, char **argv) {
  char text[64];
  sprintf(text, "BPF_STACK_TRACE(test, %d);\n", atoi(argv[1]));
  void *mod = bpf_module_create_c_from_string(text, 4, NULL, 0);
  return !(mod != NULL);
}
```

I see:

```
# env time tests/cc/test_map_size 10 # output omitted
0.08user 0.00system 0:00.09elapsed 96%CPU (0avgtext+0avgdata 39056maxresident)k

# env time tests/cc/test_map_size 100
0.07user 0.00system 0:00.09elapsed 81%CPU (0avgtext+0avgdata 39084maxresident)k

# env time tests/cc/test_map_size 1000
0.07user 0.00system 0:00.07elapsed 97%CPU (0avgtext+0avgdata 40128maxresident)k

# env time tests/cc/test_map_size 10000
0.10user 0.00system 0:00.12elapsed 91%CPU (0avgtext+0avgdata 69552maxresident)k

# env time tests/cc/test_map_size 100000
0.41user 0.10system 0:00.51elapsed 99%CPU (0avgtext+0avgdata 336784maxresident)k
```

So around 3KB per map entry.

A similar test with `BPF_HASH(test, u64, u64, N)` shows ~constant RSS regardless of the map size.

I had assumed that the memory use of the map would be in the kernel, not userspace. Am I misunderstanding how the stack trace maps work, or is this a leak/unreclaimed memory? I tried running this test with jemalloc's heap profiling and the stack that leads to the bulk of these allocations is:

```
llvm::SmallVectorBase::grow_pod(void*, unsigned long, unsigned long)
llvm::raw_svector_ostream::write_impl(char const*, unsigned long)
llvm::raw_ostream::write(unsigned char)
llvm::MCAssembler::writeSectionData(llvm::MCSection const*, llvm::MCAsmLayout const&) const
(anonymous namespace)::ELFObjectWriter::writeObject(llvm::MCAssembler&, llvm::MCAsmLayout const&)
llvm::MCAssembler::Finish()
llvm::AsmPrinter::doFinalization(llvm::Module&)
[...]

PR #1294, opened 2017-08-16
avoid large map memory allocation in userspace

yonghong-song:
In bcc, internal BPF_F_TABLE defines a structure to
contain all the table information for later easy
extraction. A global structure will be defined
with this type. Note that this structure will be
allocated by LLVM during compilation.

In the table structure, one of field is:
   _leaf_type data[_max_entries]

If the _leaf_type and _max_entries are big,
significant memory will be consumed. A big
_leaf_type size example is for BPF_STACK_TRACE map
with 127*8=1016 bytes. If max_entries is bigger
as well, significant amount of memory will be
consumed by LLVM.

This patch replaces
  _leaf_type data[_max_entries]
to
  u32 max_entries

The detail of a test example can be found in issue #1291.
For the example in #1291, without this patch, for a
BPF_STACK_TRACE map with 1M entries, the RSS is roughly
3GB (roughly 3KB per entry, looks like significant LLVM overhead).
With this patch, it is 5.8MB.

### DIFF
--- a/src/cc/export/helpers.h
+++ b/src/cc/export/helpers.h
@@ -50,11 +50,11 @@ struct _name##_table_t { \
   void (*call) (void *, int index); \
   void (*increment) (_key_type); \
   int (*get_stackid) (void *, u64); \
-  _leaf_type data[_max_entries]; \
+  u32 max_entries; \
   int flags; \
 }; \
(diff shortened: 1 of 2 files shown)


4ast:
Wow. Nice!

PR #1295, opened 2017-08-17
Use correct data type for class Data

NanXiao:


brendangregg:
[buildbot, ok to test]

Thanks, looks like this one was an accident. Others were done on purpose, due to compilation/alignment issues with u32's, and I've tried to put comments on them like in xfsslower.py:

```
// XXX: switch some to u32's when supported
```

I'm not sure if the u32 alignment/padding issue is fixed yet to switch those other ones over.

NanXiao:
I modify code to: 

```
class Data(ct.Structure):
    _fields_ = [
        ("pid", ct.c_int),
        ("delta", ct.c_ulonglong),
        ("comm", ct.c_char * TASK_COMM_LEN),
        ("fname", ct.c_char * DNAME_INLINE_LEN)
    ]
print("sizeof(Data) is %d" % ct.sizeof(Data))
```

and test it on my `X86_64` platform:  

`sizeof(Data) is 288`

It seems has processed byte-alignment. BTW, in this [example](https://github.com/iovisor/bcc/blob/e85aa7f934593c4836866e6dee248d6fd7c00b27/examples/tracing/hello_perf_output.py#L40), you use `c_uint` type for `u32 pid`. :-)

PR #1298, opened 2017-08-18
Fix an error message when detach a generic XDP program

lcp:
When detaching a generic XDP program to an interface without native XDP
support, remove_xdp() showed the following message:
    
   bpf: nlmsg error Operation not supported
    
Add the flags parameter to notify the kernel that it's a generic XDP
program.


drzaeus77:
[buildbot, ok to test]

PR #1303, opened 2017-08-19
Remove unnecessary bpf_probe_reads

pchaigno:
While working on improving bcc's rewriting of dereferences, I found a few tools that use `bpf_probe_read`s where bcc already support deferences.

I tested that the three tools still work as expected on both IPv4 and IPv6 connections.

Note: The same applies to several tools in `tools/old/`, but I wasn't sure we're supposed to change those.


brendangregg:
tested, works, thanks. Much better.

PR #1309, opened 2017-08-23
Reorganize some src/cc files

drzaeus77:
This moves around some of the c/c++ source files to (hopefully) provide some organization. This is not a complete effort, but does work so using it as a stopping point. One new feature is that it is possible to build the library with only the libbpf code enabled (without any clang). This might be useful for some external dependencies that want a bpf loader but don't want to build inside the kernel tree. Enabled with `-DENABLE_CLANG_JIT=Off` (defaults to On).


src/cc/link_all.cc
```diff
@@ -0,0 +1,21 @@
...
+      // getenv never returns -1, but compiler doesn't know!
+      if (::getenv("bar") != (char *)-1)
+        return;
+
+      (void)bcc_usdt_new_frompid(-1);
+      (void)bcc_usdt_new_frompath(nullptr);
+      (void)bcc_usdt_close(nullptr);
+    }
+  } LinkAll;  // declare one instance to invoke the constructor
+}
```
yonghong-song: Could you explain why you need to link_all.cc here? Since usdt is on, is it possible that we can just include usdt files and we will be fine?

Also, anybody use libbcc.a? libbcc.a does not have this trick, so that means if user use libbcc.a instead of libbcc.so, it won't work, right?

yonghong-song: Looked at the Cmakefile again. It does look like that usdt static library is added to the shared library and the static library. But the reference may be dropped since there is no reference. link_all.cc is a way to keep references and definitions available in the shared library. That makes sense.

We may need to add such tricks to static library as well as for sure static library may drop the references as well. This may be a lower priority if static library is not really used by anybody.

CMakeLists.txt
```diff
@@ -11,16 +11,22 @@ enable_testing()
 
 include(cmake/GetGitRevisionDescription.cmake)
 include(cmake/version.cmake)
+include(cmake/CMakeDependentOption.cmake)
```
drzaeus77: I actually didn't check for it, I will prune this extra file, thanks!

CMakeLists.txt
```diff
@@ -11,16 +11,22 @@ enable_testing()
...
 include(GNUInstallDirs)
 include(CheckCXXCompilerFlag)
 
+option(ENABLE_CLANG_JIT "Enable Loading BPF through Clang Frontend" ON)
+option(ENABLE_USDT "Enable User-level Statically Defined Tracing" ON)
+CMAKE_DEPENDENT_OPTION(ENABLE_CPP_API "Enable C++ API" ON "ENABLE_USDT" OFF)
```
drzaeus77: The idea is mostly to compartmentalize the code, and if this reduces binary size then its a plus. I am aware that the tests fail, for now my goal was just to capture the code dependencies in the library and move the directories around. I don't expect that our buildbots or upstream distros will disable these flags.

src/cc/link_all.cc
drzaeus77: Well, we're only installing the shared library in the final deliverable, so any linker errors with the .a should be visible in our local compiles and buildbots, so I don't worry too much about that. If we start shipping the .a as well, then we can double check this.

CMakeLists.txt
drzaeus77: Removed!

CMakeLists.txt
drzaeus77: I've added guards for the failing compiles.

PR #36, opened 2015-06-10
Make installation prefixes more realistic

drzaeus77:
* Add/document cmake variables that control various installation path
  options. See README.md for examples.
* Updated README a bit
* Hide helpers.h from include requirements
* Install things to real paths in a proper way. Header files will go
  into <prefix>/share/bcc/include.
* Move the kickstart script readme to its own directory.

Signed-off-by: Brenden Blanco <bblanco@plumgrid.com>

PR #1311, opened 2017-08-23
fix lua-bcc build issue with cmake try_compile

yonghong-song:
Certain versions of recent gcc (e.g., gcc 6.3.0 on ubuntu17.04)
has pie enabled by default at linker (collect2) stage.
The compilation flag "-no-pie" is available to negate
this linker option.

Add -no-pie to compilation flag only if it is available.
Earlier gcc compiler may not have this option but it does
not have linker pie on-by-default either.

Tested with ubuntu 17.04 and my local gcc 4.8.5 (not accepting -no-pie).


yonghong-song:
cc @prakashsurya @markdrayton Please help verify as well.

prakashsurya:
This works for me. Thank you!

markdrayton:
I'll verify this first thing this morning.

markdrayton:
The patch works for me. Thanks for coming up with a robust solution!

Nit: is the PR description inaccurate? As far as I can tell we only add `-no-pie` if the compiler supports it. As far as I can see we don't add `--no-pie` in any case.

yonghong-song:
The patch is in its 2nd revision and I have changed description. Goto commit page you will see the new commit message. Unfortunately, it is not automatically updated here. I will manually update here.

drzaeus77:
I think this patch deserves to be separated out into a file in the cmake/ directory, with an implementation as a macro that fills in the proper string for nopie.

yonghong-song:
Okay. Will give another spin.

yonghong-song:
I have reorged and simplified patch a little. @prakashsurya @markdrayton  Could you help take a look again?

cirocosta:
Working fine here as well :+1: I'm on `Linux version 4.10.0-32-generic (buildd@lcy01-05) (gcc version 6.3.0 20170406 (Ubuntu 6.3.0-12ubuntu2) ) #36-Ubuntu SMP Tue Aug 8 12:10:06 UTC 2017
`

markdrayton:
Looks good to me.

PR #1313, opened 2017-08-23
User space monitoring of postgres DB:

ggaurav10:
The example file monitors latency in mdwrite path and draws histogram.

### DIFF
--- a/examples/tracing/pgstat.py
+++ b/examples/tracing/pgstat.py
@@ -1,0 +1,100 @@
+#!/usr/bin/python
+#
+# dbstat        Display a histogram of PostgreSQL query latencies.
+#
+# USAGE: dbstat [-v] [-p PID [PID ...]] [-m THRESHOLD] [-u]
+#               [-i INTERVAL] {postgres} PATH_TO_LIBRARY
+#
+# Written as a basic example of getting latency of mdwrite path in postgres
+# This can be extended to other DataBases and functions
+
+from bcc import BPF
+import argparse
+import subprocess
+from time import sleep, strftime
+from ctypes import c_int
+
+examples = """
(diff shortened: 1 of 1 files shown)


drzaeus77:
[buildbot, ok to test]

brendangregg:
This shouldn't be called dbstat -- we already have something called that, and it's also tracing mdwrite. So it'd have called it pgmdwstat (which follows the convention of other tools) or something very similar and short.

The comments in the tool are inaccurate and haven't been updated since this was copied from dbstat. The output is likely inaccurate as well ("query latency" instead of "mdwrite latency").

For the examples/tracing/ directory, I'd want this to be as short as possible. Imagine it's a text book example that must be to the point and short to fit in as few lines as possible. Remove as much of the argument processing as is possible. See examples/tracing/mysqld_query.py, or examples/tracing/disksnoop.py vs tools/biosnoop.py.

If this is meant for the tools/ directory, then it would need a man page and examples_*.txt file.

As for what the tool does: Why mdwrite? Why only mdwrite and not other functions as well? To be clear about this -- if we're going to ask our end-users to spend time studying this example or learning this tool, are we making the best use of their time? Would it be better or more complete with a few more uprobes? I would not want to integrate this without very confident answers to these.

If I added a example or tool to bcc every time I found a function who'se latency was interesting, we'd have hundreds of examples and tools, and the end-user experience to dig through them all would be much worse. We must be selective.

Currently it's just printing a histogram of uprobe mdwrite latency. If that was very important to do, then I'd consider making this a shell wrapper to argdist and using its $latency functionality.

goldshtn:
Yep. I'm a bit worried that all this does is basically wrap argdist with $latency ...

goldshtn:
@ggaurav10: I'm inclined to close this PR, unless you think otherwise. As-is, this does not warrant a stand-alone tools, considering the remarks above.

PR #1315, opened 2017-08-23
funccount: add -d for duration

brendangregg:


tools/funccount.py
```diff
@@ -257,13 +264,17 @@ def run(self):
         print("Tracing %d functions for \"%s\"... Hit Ctrl-C to end." %
               (self.probe.matched, self.args.pattern))
         exiting = 0 if self.args.interval else 1
+        seconds = 0
         while True:
             try:
                 sleep(int(self.args.interval))
+                seconds += int(self.args.interval)
```
goldshtn: I know it's a nitpick, but this is assuming that the amount of time actually elapsed is equal to the sleep interval. On a loaded system, it might not necessarily be the case. Don't know if we want to do anything about it, but it kind of stuck out.

goldshtn:
LGTM.

Issue #1314, opened 2017-08-23
map.increment() is not performance efficient.


4ast:
currently map.increment(&key); does
update(zero, noexist);
if (leaf) *(leaf)++;
so it's hitting spinlocks() in the hashmap for every increment,
we should change it to
leaf = lookup(key)
if (leaf) *(leaf)++; else update(zero);
to avoid spinlocks for all increments.

yonghong-song:
> leaf = lookup(key)
> if (leaf) *(leaf)++; else update(zero); 

I guess the above "update(zero)" should be "update(one)?

4ast:
ahh. yes :)

drzaeus77:
I have a patch for this ready, unless @yonghong-song wants to take care of it in a particular way.

yonghong-song:
@drzaeus77  go ahead to proceed with your patch. What I will do is exactly like the above @4ast suggested change.

PR #1316, opened 2017-08-23
Update after lookup in map.increment for HASH types

drzaeus77:
Performing the update before lookup incurs an extra spinlock in the
kernel, which hurts performance. Reorder the code and increment to 1.
The memset is there I believe due to some type mismatch warnings, so I'm
leaving as is rather than doing a direct assign to 1. The resulting code
is optimized away anyway.

Fixes: #1314

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -332,12 +332,14 @@ bool BTypeVisitor::VisitCallExpr(CallExpr *Call) {
           string lookup = "bpf_map_lookup_elem_(bpf_pseudo_fd(1, " + fd + ")";
           string update = "bpf_map_update_elem_(bpf_pseudo_fd(1, " + fd + ")";
           txt  = "({ typeof(" + name + ".key) _key = " + arg0 + "; ";
-          if (desc->second.type == BPF_MAP_TYPE_HASH) {
-            txt += "typeof(" + name + ".leaf) _zleaf; memset(&_zleaf, 0, sizeof(_zleaf)); ";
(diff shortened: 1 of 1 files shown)


src/cc/frontends/clang/b_frontend_action.cc
```diff
@@ -332,12 +332,14 @@ bool BTypeVisitor::VisitCallExpr(CallExpr *Call) {
           string lookup = "bpf_map_lookup_elem_(bpf_pseudo_fd(1, " + fd + ")";
           string update = "bpf_map_update_elem_(bpf_pseudo_fd(1, " + fd + ")";
           txt  = "({ typeof(" + name + ".key) _key = " + arg0 + "; ";
+          txt += "typeof(" + name + ".leaf) *_leaf = " + lookup + ", &_key); ";
+          txt += "if (_leaf) (*_leaf)++; ";
           if (desc->second.type == BPF_MAP_TYPE_HASH) {
-            txt += "typeof(" + name + ".leaf) _zleaf; memset(&_zleaf, 0, sizeof(_zleaf)); ";
-            txt += update + ", &_key, &_zleaf, BPF_NOEXIST); ";
+            txt += "else { typeof(" + name + ".leaf) _zleaf; memset(&_zleaf, 0, sizeof(_zleaf)); ";
```
yonghong-song: Looks good. Only one minor comment, I guess originally, "_zleaf" is to indicate "zero leaf"? Maybe just use a different name like "_tleaf"?

drzaeus77:
Well, it starts as a zero leaf :) I think it's fine as-is.

PR #1319, opened 2017-08-24
fix a bug introduced by previous lua-bcc build fix commit

yonghong-song:
The bug does not show up in my previous test probably because
I did not clean up the build directory and did not have a clean cmake.

In gcc 4.8.5 environment where -no-pie is not supported,
a clean cmake will trigger the issue as subsequent other
cxx compile test will fail due to additional -no-pie flag.

The fix is to reset the CMAKE_REQUIRED_FLAGS to its previous value
after the specific test.

### DIFF
--- a/cmake/FindCompilerFlag.cmake
+++ b/cmake/FindCompilerFlag.cmake
@@ -4,6 +4,7 @@
 if (CMAKE_C_COMPILER_ID MATCHES "Clang")
 	set(COMPILER_NOPIE_FLAG "-nopie")
 else()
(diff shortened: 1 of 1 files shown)


cmake/FindCompilerFlag.cmake
```diff
@@ -12,4 +12,5 @@ else()
 	else()
 		set(COMPILER_NOPIE_FLAG "")
 	endif()
+	set(CMAKE_REQUIRED_FLAGS "")
```
drzaeus77: Probably you should cache the previous value and restore it here, just in case.

PR #1327, opened 2017-08-30
Support PID only USDT for C++

romain-intel:
These two patches add support for specifying USDT with just PID. They also attempt to add some checks when the same USDT is present in multiple places as well as a bit more error checking and messages in case something goes wrong (to help the user debug). I am not 100% sure on the checks that I do (mostly in terms of argument sizes, etc). Feedback welcome.


romain-intel:
The error counting is not needed (really just a bool). I don't mind reporting the error through the payload if that is the preferred method. I just need a way to trickle it back-up.
Re: argument format check, I think it should work but definitely take a close look :).

src/cc/api/BPF.cc
```diff
@@ -233,22 +233,39 @@ StatusTuple BPF::attach_usdt(const USDT& usdt, pid_t pid, int cpu,
...
-      for (auto addr : u.addresses_) {
-        auto res =
-            attach_uprobe(u.binary_path_, std::string(), u.probe_func_, addr);
-        if (res.code() != 0) {
-          failed = true;
-          err_msg += "USDT " + u.print_name() + " at " + std::to_string(addr);
-          err_msg += ": " + res.msg() + "\n";
-          break;
+      for (auto bin_addr : u.bin_addresses_) {
+        bool break_out = false;
```
yonghong-song: Looks like bin_addresses_ only for debug purpose, right?

src/cc/api/BPF.cc
```diff
@@ -619,24 +638,55 @@ StatusTuple BPF::detach_perf_event_all_cpu(open_probe_t& attr) {
...
+  int count = 0;
+  for (auto probe : probes) {
+    if (!probe->enable(probe_func_))
+      return StatusTuple(-1, "Failed to enable USDT " + print_name());
+    // We check if the format is compatible between them
+    uint64_t tmpfmt = probe->usdt_getargfmt();
+    if (!::USDT::Probe::encloses_argfmt(argfmt, tmpfmt)) {
+      // Check if tmpfmt encloses argfmt
+      if (::USDT::Probe::encloses_argfmt(tmpfmt, argfmt)) {
+        maxfmtidx = count;
```
yonghong-song: Could you add some comments here? It is hard to understand what is the intention here.

src/cc/usdt/usdt.cc
```diff
@@ -193,40 +212,87 @@ bool Probe::usdt_getarg(std::ostream &stream) {
...
 }
 
+uint64_t Probe::usdt_getargfmt() {
+  assert(!locations_.empty());
+  if (argfmt_ == (uint64_t)-1) {
+    argfmt_ = 0;
+    const size_t arg_count = locations_[0].arguments_.size();
+
+    if (arg_count > 0) {
+      assert(arg_count <= 16);
```
yonghong-song: why 16? derived from some headers?

src/cc/usdt/usdt.cc
```diff
@@ -193,40 +212,87 @@ bool Probe::usdt_getarg(std::ostream &stream) {
...
+        assert(t > -5 && t < 5);
+        if (t < 0) {
+          argfmt_ |= (1 | (-t));
+        } else {
+          argfmt_ |= t;
+        }
+      }
+    }
+  }
+  return argfmt_;
```
yonghong-song: Some comments are needed to understand the algorithm ...

src/cc/api/BPF.cc
romain-intel: Not entirely; it's used to keep track of the addresses to attach to. Basically instead of just an address, it now has a binary and an address.

src/cc/api/BPF.cc
romain-intel: Will do. The intention, and I am not sure if this is 100% correct, is to pick the representation of the generated functions that is the most generic. Basically, the one that matches in signedness and has the largest size argument (prefer u64 over u32 for example).

src/cc/usdt/usdt.cc
romain-intel: No, it's my encoding of argument type that takes 4 bits so in a 64 bit variable, I can only encode up to 16 arguments. I suppose that should be enough. It's not the prettiest but is faster when comparing things. I can remove this and just use bigger structures too.

src/cc/usdt/usdt.cc
romain-intel: OK. Ya, it's basically an encoding of the argument types in a single 64 bit value. I'll clarify.

4ast:

PR #1331, opened 2017-09-01
Nfsslower tool

samuelnair:
This is similar to the {xfs,zfs,ext4}slower scripts that we have. Most of the code was copied from ext4slower. I have tested the script against NFSv3 and NFSv4 which should cover a vast majority of use cases. Anything older is probably irrelevant since eBPF would probably not be supported on older distributions or kernels.

I have followed the CONTRIBUTING-SCRIPTS.md guidelines as closely as possible and since this is fairly similar to the other tools that we have, it should be safe to merge.

### DIFF
--- a/man/man8/nfsslower.8
+++ b/man/man8/nfsslower.8
@@ -1,0 +1,113 @@
+.TH nfsslower 8  "2017-09-01" "USER COMMANDS"
+.SH NAME
+nfsslower \- Trace slow nfs file operations, with per-event details.
+.SH SYNOPSIS
+.B nfsslower [\-h] [\-j] [\-p PID] [min_ms]
+.SH DESCRIPTION
+This tool traces common NFSv3 & NFSv4 file operations: reads, writes, opens, and
+getattrs. It measures the time spent in these operations, and prints details
+for each that exceeded a threshold.
+
+WARNING: See the OVERHEAD section.
+
+By default, a minimum millisecond threshold of 10 is used. If a threshold of 0
+is used, all events are printed (warning: verbose).
+
+Since this works by tracing the nfs_file_operations interface functions, it
+will need updating to match any changes to these functions.
+
+Since this uses BPF, only the root user can use this tool.
+.SH REQUIREMENTS
+CONFIG_BPF and bcc.
+.SH OPTIONS
+\-p PID
+Trace this PID only.
+.TP
+min_ms
+Minimum I/O latency (duration) to trace, in milliseconds. Default is 10 ms.
+.SH EXAMPLES
+.TP
+Trace synchronous file reads and writes slower than 10 ms:
+#
+.B nfsslower
+.TP
+Trace slower than 1 ms:
+#
+.B nfsslower 1
+.TP
+Trace slower than 1 ms, and output just the fields in parsable format (csv):
+#
+.B nfsslower \-j 1
+.TP
+Trace all file reads and writes (warning: the output will be verbose):
+#
+.B nfsslower 0
+.TP
+Trace slower than 1 ms, for PID 181 only:
+#
+.B nfsslower \-p 181 1
+.SH FIELDS
+.TP
+TIME(s)
+Time of I/O completion since the first I/O seen, in seconds.
+.TP
+COMM
+Process name.
+.TP
+PID
+Process ID.
+.TP
+T
+Type of operation. R == read, W == write, O == open, G == getattr.
+.TP
+OFF_KB
+File offset for the I/O, in Kbytes.
+.TP
+BYTES
+Size of I/O, in bytes.
+.TP
+LAT(ms)
+Latency (duration) of I/O, measured from when it was issued by VFS to the
+filesystem, to when it completed. This time is inclusive of RPC latency,
+Network latency, cache lookup, remote fileserver processing latency, etc. 
+Its a more accurate measure of the latency suffered by applications performing
+NFS read/write calls to a fileserver.
+.TP
+FILENAME
+A cached kernel file name (comes from dentry->d_iname).
+.TP
+ENDTIME_us
+Completion timestamp, microseconds (\-j only).
+.TP
+OFFSET_b
+File offset, bytes (\-j only).
+.TP
+LATENCY_us
+Latency (duration) of the I/O, in microseconds (\-j only).
+.SH OVERHEAD
+This adds low-overhead instrumentation to NFS operations,
+including reads and writes from the file system cache. Such read, writes and
+particularly getattrs can be very frequent (depending on the workload; eg, 1M/sec),
+at which point the overhead of this tool (even if it prints no "slower" events) can
+begin to become significant. Measure and quantify before use. If this
+continues to be a problem, consider switching to a tool that prints in-kernel
+summaries only. This tool has been tested with NFSv3 & NVSv4, but it might work
+with NFSv{1,2}, since it is tracing the generic functions from nfs_file_operations.
+.PP
+Note that the overhead of this tool should be less than fileslower(8), as
+this tool targets nfs functions only, and not all file read/write paths.
+.SH SOURCE
+This is from bcc.
+.IP
+https://github.com/iovisor/bcc
+.PP
+Also look in the bcc distribution for a companion _examples.txt file containing
+example usage, output, and commentary for this tool.
+.SH OS
+Linux
+.SH STABILITY
+Unstable - in development.
+.SH AUTHOR
+Samuel Nair
+.SH SEE ALSO
+biosnoop(8), funccount(8), fileslower(8)
--- a/man/man8/nfsslower.8
+++ b/man/man8/nfsslower.8
@@ -1,6 +1,6 @@
 .TH nfsslower 8  "2017-09-01" "USER COMMANDS"
 .SH NAME
-nfsslower \- Trace slow nfs file operations, with per-event details.
+nfsslower \- Trace slow NFS file operations, with per-event details.
 .SH SYNOPSIS
 .B nfsslower [\-h] [\-j] [\-p PID] [min_ms]
 .SH DESCRIPTION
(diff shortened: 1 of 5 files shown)


brendangregg:
Some minor things.

I see it uses kprobes. I did that in other tools for a couple of reasons: 1) BPF tracepoint support didn't exist yet, and 2) some file systems don't export tracepoints.

NFS has tracepoints, which can be used by BPF on 4.7 onwards:

```
# perf list | grep -i nfs
  nfs:nfs_access_enter                               [Tracepoint event]
  nfs:nfs_access_exit                                [Tracepoint event]
  nfs:nfs_atomic_open_enter                          [Tracepoint event]
  nfs:nfs_atomic_open_exit                           [Tracepoint event]
  nfs:nfs_create_enter                               [Tracepoint event]
  nfs:nfs_create_exit                                [Tracepoint event]
  nfs:nfs_fsync_enter                                [Tracepoint event]
  nfs:nfs_fsync_exit                                 [Tracepoint event]
[...]
```

I'm not sure if they can be used in this case (I don't see read or write!), but I'd at least put a comment in nfsslower to say something like: "this currently uses kprobes; a future version may use NFS tracepoints if and when possible." (Maybe the other slower tools need that comment as well, to remind future programmers to check for tracepoints first.)

Code comment says XFS:

```
# WARNING: This adds low-overhead instrumentation to these XFS operations,
```

There are a number of places in the code that have double new lines (eg, lines 198 and 199), which I find distracting, and looks accidental. I rarely use a double new line. I'd make them all single new lines.

Other minor whitespace changes:

```
pep8 --show-source --ignore=E123,E125,E126,E127,E128,E302 nfsslower.py 
nfsslower.py:274:58: W503 line break before binary operator
                                                         / 1024,
                                                         ^
nfsslower.py:276:58: W503 line break before binary operator
                                                         / 1000,
                                                         ^
```

The addition in README.md puts it at the end. I'd insert nfsslower in alphabetical order.

That's it. Code, example file, and man page look good.

samuelnair:
Hey @brendangregg 

I chose kprobes for the same reasons you listed. There aren't any tracepoints for the functions i needed to trace, with the exception of nfs_getattr which does have entry and exit tracepoints, but unfortunately its arguments don't provide a file name, which is fairly critical for this tool to be useful. 

I have added in the note to the doc's indicating that in the future we will probably want to move to tracepoints and fixed some of the pep8 and whitespace/newline issues

brendangregg:
[buildbot, ok to test]

man/man8/nfsslower.8
```diff
@@ -0,0 +1,122 @@
...
+
+By default, a minimum millisecond threshold of 10 is used. If a threshold of 0
+is used, all events are printed (warning: verbose).
+
+Since this works by tracing the nfs_file_operations interface functions, it
+will need updating to match any changes to these functions.
+
+This tool uses kprobes to instrument the kernel for entry and exit
+information, in the future a preferred way would be to use tracepoints.
+Currently there are'nt any tracepoints available for nfs_read_file,
```
goldshtn: Typo, should be "aren't".

man/man8/nfsslower.8
```diff
@@ -0,0 +1,122 @@
+.TH nfsslower 8  "2017-09-01" "USER COMMANDS"
+.SH NAME
+nfsslower \- Trace slow nfs file operations, with per-event details.
```
goldshtn: Capitalize NFS? A matter of taste I suppose.

man/man8/nfsslower.8
```diff
@@ -0,0 +1,122 @@
...
+
+Since this uses BPF, only the root user can use this tool.
+.SH REQUIREMENTS
+CONFIG_BPF and bcc.
+.SH OPTIONS
+\-p PID
+Trace this PID only.
+.TP
+\-j
+Trace output in CSV format
```
goldshtn: Add trailing period for consistency with other options?

man/man8/nfsslower.8
```diff
@@ -0,0 +1,122 @@
...
+.TP
+Trace synchronous file reads and writes slower than 10 ms:
+#
+.B nfsslower
+.TP
+Trace slower than 1 ms:
+#
+.B nfsslower 1
+.TP
+Trace slower than 1 ms, and output just the fields in parsable format (csv):
```
goldshtn: Capitalize CSV?

man/man8/nfsslower.8
```diff
@@ -0,0 +1,122 @@
...
+OFF_KB
+File offset for the I/O, in Kbytes.
+.TP
+BYTES
+Size of I/O, in bytes.
+.TP
+LAT(ms)
+Latency (duration) of I/O, measured from when it was issued by VFS to the
+filesystem, to when it completed. This time is inclusive of RPC latency,
+Network latency, cache lookup, remote fileserver processing latency, etc.
```
goldshtn: Lowercase "network"?

man/man8/nfsslower.8
```diff
@@ -0,0 +1,122 @@
...
+including reads and writes from the file system cache. Such read, writes and
+particularly getattrs can be very frequent (depending on the workload; eg, 1M/sec),
+at which point the overhead of this tool (even if it prints no "slower" events) can
+begin to become significant. Measure and quantify before use. If this
+continues to be a problem, consider switching to a tool that prints in-kernel
+summaries only. This tool has been tested with NFSv3 & NVSv4, but it might work
+with NFSv{1,2}, since it is tracing the generic functions from nfs_file_operations.
+.PP
+Note that the overhead of this tool should be less than fileslower(8), as
+this tool targets nfs functions only, and not all file read/write paths.
```
goldshtn: NFS?

tools/nfsslower_example.txt
```diff
@@ -0,0 +1,158 @@
...
+11:25:16 dd             21295  W 1048576 18432      22.93 1.test
+11:25:16 dd             21295  W 1048576 19456      14.65 1.test
+11:25:16 dd             21295  W 1048576 20480      12.58 1.test
+11:25:16 dd             21297  W 1048576 6144       10.50 1.test.w
+11:25:16 dd             21297  W 1048576 7168       16.65 1.test.w
+11:25:16 dd             21297  W 1048576 8192       13.01 1.test.w
+11:25:16 dd             21297  W 1048576 9216       14.06 1.test.w
+
+This shows NFS writes from dd each 1MB in size to 2 different files. The
+writes all had latency higher than 10ms
```
goldshtn: Trailing period.

tools/nfsslower_example.txt
```diff
@@ -0,0 +1,158 @@
...
+
+This "latency" is measured from when the operation was issued from the VFS
+interface to the file system, to when it completed. This spans everything:
+RPC latency, Network latency, file system CPU cycles, file system locks, run
```
goldshtn: Lowercase "network"?

tools/nfsslower_example.txt
```diff
@@ -0,0 +1,158 @@
...
+queue latency, etc. This is a better measure of the latency suffered by
+applications reading from a NFS share and can better expose problems
+experienced by NFS clients.
+
+Note that this only traces the common NFS operations (read,write,open and
+getattr). I chose to include getattr as a significant percentage of NFS
+traffic end up being getattr calls and are a good indicator of problems
+with an NFS server
```
goldshtn: Trailing period.

tools/nfsslower_example.txt
```diff
@@ -0,0 +1,158 @@
...
+
+The threshold can be provided as an argument. Eg, I/O slower than 1 ms:
```
goldshtn: "E.g."

tools/nfsslower.py
```diff
@@ -0,0 +1,322 @@
...
+b.attach_kretprobe(event="nfs_file_write", fn_name="trace_write_return")
+b.attach_kretprobe(event="nfs4_file_open", fn_name="trace_file_open_return")
+b.attach_kretprobe(event="nfs_file_open", fn_name="trace_file_open_return")
+b.attach_kretprobe(event="nfs_getattr", fn_name="trace_getattr_return")
+
+if(csv):
+    print("ENDTIME_us,TASK,PID,TYPE,BYTES,OFFSET_b,LATENCY_us,FILE")
+else:
+    if min_ms == 0:
+        print("Tracing NFS operations")
```
goldshtn: We usually have tools say something like "Tracing NFS operations... Ctrl+C to quit".

tools/nfsslower.py
```diff
@@ -0,0 +1,322 @@
...
+    else:
+        print("Tracing NFS operations that are slower than %d ms" % min_ms)
```
goldshtn: "Tracing NFS operations slower than %d ms... Ctrl+C to quit"

samuelnair:
Thanks for detailed review, have addressed the comments in the latest commit.

goldshtn:
LGTM. Merging. Thank you for this!

goldshtn:
Oh no, sorry ☹️  Forgot one more thing: could you please add a smoke test to https://github.com/iovisor/bcc/blob/master/tests/python/test_tools_smoke.py -- we tend to forget about this one. @samuelnair

goldshtn:
Opened #1339 so we don't forget to add smoke tests in the future...

samuelnair:
Good catch.
But im not sure if nfsslower  will work in the smoke test scirpt. Since depending on where you execute this, it may not have the NFS kernel module loaded. It typically gets loaded when a new NFS mount request is made or if someone manually insmod the module.

Looking at the script i see a number of other tests i.e. xfsslower & zfsslower have their tests commented out with a pass being returned by default. I could do the same or put in additional logic to check if the module is loaded and only then run the test, which do you prefer ?

goldshtn:
If the check is easy, I think it's best to do the check and then run the test -- so that even if we don't have NFS loaded on the test machines, it would at least allow contributors to run the tests on their machine, if they have it.

samuelnair:
Have added in a simple check to see if the NFS module is loaded. Let me know if this is good enough

goldshtn:
Thanks!

PR #1335, opened 2017-09-03
stackcount: add -K, -U, -d, -D, and -f

### LINKED ISSUE #1219: stackcount to support both kernel and user stacks
Currently stackcount will count kernel stacks for kernel probes, and user stacks for user probes. I was just tracing the __alloc_fd() kernel function, and wanted to see the user stacks, and there isn't currently a mode for that.

Here's how I think it could be enhanced (low priority)

- by default: as it currently is: kernel stacks for kernel probes, user stacks for user probes
- -K: kernel stacks only
- -U: user stacks only
- -KU: kernel and user stacks

The -K and -U options follow trace.py.

brendangregg:
This fixes #1219 so that stackcount can trace both user and kernel stacks (code based on Sasha's trace.py code), and also adds a duration option and folded output modes for flame graph generation.

### DIFF
--- a/man/man8/stackcount.8
+++ b/man/man8/stackcount.8
@@ -2,7 +2,7 @@
 .SH NAME
 stackcount \- Count function calls and their stack traces. Uses Linux eBPF/bcc.
 .SH SYNOPSIS
-.B stackcount [\-h] [\-p PID] [\-i INTERVAL] [\-T] [\-r] [\-s]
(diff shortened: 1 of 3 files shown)


goldshtn:
Thanks, this is very useful. I've also run across situations where this could have been useful, once or twice. The -D and -f also make sense.

PR #1336, opened 2017-09-04
Do not create instance for kprobe

palmtenor:
Logic to create instances when attach kprobes was introduced in #918 to address #872. However it is not necessary to do so, as creating separated kprobe events with different names would already allow multiple programs attached to same Kernel function.

On the other hand, creating instances leads to additional overhead on kprobe tracing, such as each instance would consume tens of megabytes of memories, and more complicated logic as we tried to address in #1262.

Therefore this commit removes instance logic from BCC. Also made a minor variable name change (`new_name` -> `event_alias`) to make the code more readable.

Tested by running multiple `opensnoop` and make sure they all work simoutanously.

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -392,12 +392,10 @@ void * bpf_attach_kprobe(int progfd, enum bpf_probe_attach_type attach_type, con
 {
   int kfd;
   char buf[256];
-  char new_name[128];
+  char event_alias[128];
   struct perf_reader *reader = NULL;
   static char *event_type = "kprobe";
-  int n;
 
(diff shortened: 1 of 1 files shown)


goldshtn:
@yonghong-song We still have direct access to the trace pipe from BCC, but it was never a really great idea because other ftrace users could clobber the global trace pipe anyway. Perhaps we should even deprecate that API or get rid of it entirely.

yonghong-song:
Right. Look at the "trace_pipe" usage from bcc. They are all from /sys/kernel/debug/tracing/trace_pipe, mostly probably for bpf_trace_printk purpose. Therefore, this should not interfere
with this patch.

I do agree that trace_print API is probably only for debugging purpose...

PR #1338, opened 2017-09-04
Improve string buffer checking on uprobe attach and detach

palmtenor:
BCC encodes target binary's path in uprobe event's name and the attach command also includes target binary's path, which could be very long. Error would happen if the long event name or binary path overflows our internal string buffer.

This commit improves buffer size checking, on top of #1328:
- Increase buffer size of event alias in uprobe to `PATH_MAX`.
- Add check of `snprintf` return value and output debug message when overflow happens.
- Expand buffer size and add check for `bpf_detach_probe` as well.

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -606,15 +606,12 @@ void * bpf_attach_uprobe(int progfd, enum bpf_probe_attach_type attach_type, con
                         pid_t pid, int cpu, int group_fd,
                         perf_reader_cb cb, void *cb_cookie)
 {
-  int kfd;
   char buf[PATH_MAX];
-  char new_name[256];
+  char event_alias[PATH_MAX];
   struct perf_reader *reader = NULL;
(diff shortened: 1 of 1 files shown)


src/cc/libbpf.c
```diff
@@ -513,13 +513,11 @@ void * bpf_attach_uprobe(int progfd, enum bpf_probe_attach_type attach_type, con
...
   struct perf_reader *reader = NULL;
   static char *event_type = "uprobe";
-  int ns_fd = -1;
-  int n;
+  int res, ns_fd = -1;
 
-  snprintf(new_name, sizeof(new_name), "%s_bcc_%d", ev_name, getpid());
   reader = perf_reader_new(cb, NULL, NULL, cb_cookie, probe_perf_reader_page_cnt);
   if (!reader)
     goto error;
```
yonghong-song: If returning error here, kfd will be uninitialized. So you may want to initialize kfd = -1.

PR #1341, opened 2017-09-05
examples: fixed http_filter example

pbhole:
Loop unrolling wasn't happening because of variable upper
bound. Making it const fixes the problem.

Tested with 4.13.0-rc6 on fedora 26

### DIFF
--- a/examples/networking/http_filter/http-parse-complete.c
+++ b/examples/networking/http_filter/http-parse-complete.c
@@ -92,7 +92,8 @@ int http_filter(struct __sk_buff *skb) {
 	unsigned long p[7];
 	int i = 0;
 	int j = 0;
-	for (i = payload_offset ; i < (payload_offset + 7) ; i++) {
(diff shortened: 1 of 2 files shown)


examples/networking/http_filter/http-parse-complete.c
```diff
@@ -92,7 +92,8 @@ int http_filter(struct __sk_buff *skb) {
 	unsigned long p[7];
 	int i = 0;
 	int j = 0;
-	for (i = payload_offset ; i < (payload_offset + 7) ; i++) {
+	const int last_index = payload_offset + 7;
```
yonghong-song: This is certainly working before. But llvm compiler loop unrolling has evolved not liking the expression in the for loop conditions. I did not track down why.

Just with "int last_index = payload_offset + 7" also works fine. The point is to have a variable instead of expression.

yonghong-song:
[buildbot, ok to test]

PR #1344, opened 2017-09-09
Fix for incorrect tool arguments in the man pages

samuelnair:
While writing up the man page for a new nfsdist tool that i hope to commit shortly. I noticed that the man pages for the {ext4,xfs,zfs}dist tools had a small error in their man pages. This PR fixes that.


drzaeus77:
[buildbot, ok to test]

brendangregg:
thanks!

PR #1348, opened 2017-09-12
Fix freezing of 'test_brb' and 'test_brb2' if netperf is not found

sandip4n:
If netperf is not installed or installed at a location that is not in PATH as recognized by Python, then 'test_brb' and 'test_brb2' freeze after an OSError is raised. To avoid this, we proactively check if the 'iperf', 'netserver' and 'netperf' binaries are available before making the corresponding NSPopen() calls.

### DIFF
--- a/tests/python/test_brb.py
+++ b/tests/python/test_brb.py
@@ -65,6 +65,7 @@ from ctypes import c_uint
 from netaddr import IPAddress, EUI
 from bcc import BPF
 from pyroute2 import IPRoute, NetNS, IPDB, NSPopen
+from utils import NSPopenWithCheck
 import sys
 from time import sleep
 from unittest import main, TestCase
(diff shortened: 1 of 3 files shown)


sandip4n:
Makes sense. I have added a new commit with the suggested changes.

yonghong-song:
[buildbot, ok to test]

sandip4n:
Can this be considered for a merge? If so, I can go ahead and remove the RFC tag.

yonghong-song:
@sandip4n please go ahead to remove RFC tag. Thanks.

sandip4n:
@yonghong-song Done. Thanks.

drzaeus77:
I think the last step is to add a `@skipUnless` check, similar to the version check in test_tools_smoke for example. I'll merge this one, but can you please add this in a follow-up?

sandip4n:
@drzaeus77 Sure. I can do that.

PR #1349, opened 2017-09-12
Add missing nfsdist and nfsslower to snap

ColinIanKing:
Signed-off-by: Colin Ian King <colin.king@canonical.com>

### DIFF
--- a/snapcraft/snapcraft.yaml
+++ b/snapcraft/snapcraft.yaml
@@ -156,6 +156,12 @@ apps:
     mysqld-qslower:
         command: wrapper mysqld_qslower
         aliases: [mysqld-qslower]
+    nfsdist:
+        command: wrapper nfsdist
+        aliases: [nfsdist]
+    nfsslower:
+        command: wrapper nfsslower
+        aliases: [nfsslower]
     nodegc:
         command: wrapper nodegc
         aliases: [nodegc]


goldshtn:
Thanks! Perhaps it might make sense to add this to the CONTRIBUTING-SCRIPTS document?

In fact, while we're at it, maybe it could be nice to have an automatic script for new tool contributions that:

- Checks spelling
- Runs lint checks
- Makes sure there's a smoke test
- Makes sure there's a man page, examples page, README entry for the tool
- Makes sure the tool is added to the snap file
- etc.

Thoughts? @brendangregg @drzaeus77

goldshtn:
[buildbot, ok to test]

pchaigno:
> Runs lint checks

@goldshtn There's already a [.travis.yml file](https://github.com/iovisor/bcc/blob/master/.travis.yml) to run link checks. It's just not enabled yet.

brendangregg:
@goldshtn we could write a tool checker, but OTOH, I'm not sure how often we'd need to run it. I think once we get to about 150 tools, we may have pretty much everything we want for tackling low-hanging fruit and common problems, and additional tools would be more niche/esoteric, and can live in other locations/repositories. We have to draw the line somewhere, otherwise we're maintaining too much for too little additional value. I published over 500 different DTrace tools, and wrote more that I didn't publish, and I don't think they'd all belong in bcc. So far what I've been including is pretty generic: ext4 latency, runq latency, etc.

goldshtn:
Sure, that makes sense.

PR #1356, opened 2017-09-20
Fix 'test_debuginfo' from failing if a kernel symbol has multiple aliases

sandip4n:
The first symbol from `/proc/kallsyms` is read by `test_debuginfo` and the name obtained here is compared against the name obtained from the `BPF.ksym()` library call. In some architectures such as `powerpc64le`, a kernel symbol may have multiple aliases that refer to the same address. So, to avoid the test from failing, we need to compare the name returned by `BPF.ksym()` against all possible aliases for the given address in `/proc/kallsyms` and look for a match.

### DIFF
--- a/tests/python/test_debuginfo.py
+++ b/tests/python/test_debuginfo.py
@@ -9,12 +9,27 @@ from unittest import main, TestCase
 
 class TestKSyms(TestCase):
     def grab_sym(self):
-        # Grab the first symbol in kallsyms that has type 't'.
+        address = ""
+        aliases = []
+
+        # Grab the first symbol in kallsyms that has type 't' or 'T'.
+        # Also, find all aliases of this symbol which are identifiable
+        # by the same address.
         with open("/proc/kallsyms") as f:
             for line in f:
-                (addr, t, name) = line.strip().split()
-                if t == "t":
-                    return (addr, name)
+
+                # Extract the first 3 columns only. The 4th column
+                # containing the module name may not exist for all
+                # symbols.
+                (addr, t, name) = line.strip().split()[:3]
+                if t == "t" or t == "T":
+                    if not address:
+                        address = addr
+                    if addr == address:
+                        aliases.append(name)
+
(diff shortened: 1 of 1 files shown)


tests/python/test_debuginfo.py
```diff
@@ -9,22 +9,38 @@
...
 
     def test_ksymname(self):
         sym = BPF.ksymname("__kmalloc")
         self.assertIsNotNone(sym)
         self.assertNotEqual(sym, 0)
 
     def test_ksym(self):
-        (addr, name) = self.grab_sym()
+        (addr, aliases) = self.grab_sym()
         sym = BPF.ksym(int(addr, 16))
```
yonghong-song: This is not related to your change. The above uses "int(addr, 16)". It looks like "long(addr, 16)" may be better for addr? The reason that "int(addr, 16)" works could be due to later sign extension. The "addr" c binding has a type of unsigned long long.

lib.bcc_symcache_resolve_no_demangle.argtypes = [ct.c_void_p, ct.c_ulonglong, ct.POINTER(bcc_symbol)]

sandip4n: I tried changing it to `long(addr, 16)` and testing on both `x86_64` and `ppc64le` systems. Works fine on both.

However, from what I have read [here](http://deeplearning.net/software/theano/tutorial/python-memory-management.html), integers in Python have a minimum size of 24 bytes, so the current code should be doing fine. Also, when the size of an integer exceeds `sys.maxint` (i.e. 2<sup>63</sup> - 1 for 64-bit systems), the minimum size becomes 36 bytes and can also continue to grow linearly in the logarithm of the integer represented.

yonghong-song: Thanks for explanation. I thought that "int" in python is also 32bit. Obviously I am wrong. Your original patch is good then.

yonghong-song:
[buildbot, ok to test]

PR #1357, opened 2017-09-21
Fix edge case when doing symbol name -> address resolution

palmtenor:
When resolving symbol name to file address, we would need to look at the ELF's load sections for executable (`ET_EXEC`) binaries. Our current logic:
- Looks at the first load section without checking executable flag.
- Only looks at first load section.
Therefore for certain special binaries it would fail or calculate incorrect file offset.

This PR:
- Adds `bcc_elf_foreach_load_section` to iterate over all executable load section of a ELF file.
- Makes change to `bcc_resolve_symname`.
- Makes some documentation and style improvements.

### DIFF
--- a/src/cc/bcc_syms.cc
+++ b/src/cc/bcc_syms.cc
@@ -416,11 +416,24 @@ static int _find_sym(const char *symname, uint64_t addr, uint64_t,
   return 0;
 }
 
+struct load_addr_t {
+  uint64_t target_addr;
+  uint64_t binary_addr;
+};
+int _find_load(uint64_t v_addr, uint64_t mem_sz, uint64_t file_offset,
+                       void *payload) {
+  struct load_addr_t *addr = static_cast<load_addr_t *>(payload);
+  if (addr->target_addr >= v_addr && addr->target_addr < (v_addr + mem_sz)) {
+    addr->binary_addr = addr->target_addr - v_addr + file_offset;
+    return -1;
+  }
+  return 0;
+}
+
 int bcc_resolve_symname(const char *module, const char *symname,
                         const uint64_t addr, int pid,
                         struct bcc_symbol_option *option,
                         struct bcc_symbol *sym) {
-  uint64_t load_addr;
   static struct bcc_symbol_option default_option = {
     .use_debug_file = 1,
     .check_debug_file_crc = 1,
@@ -437,29 +450,37 @@ int bcc_resolve_symname(const char *module, const char *symname,
   } else {
     sym->module = bcc_procutils_which_so(module, pid);
   }
-
   if (sym->module == NULL)
     return -1;
 
   ProcMountNSGuard g(pid);
 
-  if (bcc_elf_loadaddr(sym->module, &load_addr) < 0)
-    goto invalid_module;
-
   sym->name = symname;
   sym->offset = addr;
-
   if (option == NULL)
     option = &default_option;
 
   if (sym->name && sym->offset == 0x0)
     if (bcc_elf_foreach_sym(sym->module, _find_sym, option, sym) < 0)
       goto invalid_module;
-
   if (sym->offset == 0x0)
     goto invalid_module;
 
-  sym->offset = (sym->offset - load_addr);
+  // For executable (ET_EXEC) binaries, translate the virtual address
+  // to physical address in the binary file.
+  // For shared object binaries (ET_DYN), the address from symbol table should
+  // already be physical address in the binary file.
+  if (bcc_elf_get_type(sym->module) == ET_EXEC) {
+    struct load_addr_t addr = {
+      .target_addr = sym->offset,
+      .binary_addr = 0x0,
+    };
+    if (bcc_elf_foreach_load_section(sym->module, &_find_load, &addr) < 0)
+      goto invalid_module;
+    if (!addr.binary_addr)
+      goto invalid_module;
+    sym->offset = addr.binary_addr;
+  }
   return 0;
 
 invalid_module:
--- a/src/cc/bcc_elf.c
+++ b/src/cc/bcc_elf.c
@@ -470,6 +470,41 @@ int bcc_elf_foreach_sym(const char *path, bcc_elf_symcb callback,
       path, callback, (struct bcc_symbol_option*)option, payload, 0);
 }
 
+int bcc_elf_foreach_load_section(const char *path,
+                                 bcc_elf_load_sectioncb callback,
+                                 void *payload) {
+  Elf *e = NULL;
+  int fd = -1, err = -1, res;
+  size_t nhdrs, i;
+
+  if (openelf(path, &e, &fd) < 0)
+    goto exit;
+
+  if (elf_getphdrnum(e, &nhdrs) != 0)
+    goto exit;
+
+  GElf_Phdr header;
+  for (i = 0; i < nhdrs; i++) {
+    if (!gelf_getphdr(e, (int)i, &header))
+      continue;
+    if (header.p_type != PT_LOAD || !(header.p_flags & PF_X))
+      continue;
+    res = callback(header.p_vaddr, header.p_memsz, header.p_offset, payload);
+    if (res < 0) {
+      err = 1;
+      goto exit;
+    }
+  }
+  err = 0;
+
+exit:
+  if (e)
+    elf_end(e);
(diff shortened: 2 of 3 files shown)


goldshtn:
Wow, I have just run into the same issue last week with libcoreclr.so (part of .NET Core) symbols, and was working on a PR to fix this.

LGTM, although I haven't tested because I'm traveling this week.

src/cc/bcc_syms.cc
```diff
@@ -416,50 +416,68 @@ static int _find_sym(const char *symname, uint64_t addr, uint64_t,
...
     .use_symbol_type = BCC_SYM_ALL_TYPES,
   };
 
-  if (module == NULL)
-    return -1;
-
   memset(sym, 0, sizeof(bcc_symbol));
 
+  if (module == NULL)
+    return -1;
```
yonghong-song: Any particular reason why you move if condition here?

src/cc/bcc_syms.cc
```diff
@@ -416,50 +416,68 @@ static int _find_sym(const char *symname, uint64_t addr, uint64_t,
...
     if (bcc_elf_foreach_sym(sym->module, _find_sym, option, sym) < 0)
       goto invalid_module;
-
   if (sym->offset == 0x0)
     goto invalid_module;
 
-  sym->offset = (sym->offset - load_addr);
+  // For executable binaries, translate the virtual address
+  // to physical address in the binary file
+  if (bcc_elf_get_type(sym->module) == ET_EXEC) {
```
yonghong-song: Any possibility sym->module type is not ET_EXEC? What if it is the case?

src/cc/bcc_syms.cc
palmtenor: Well, I think it would be good to do `memset` on result before returning from error

src/cc/bcc_syms.cc
palmtenor: It could be shared library (`ET_DYN`) where the address from symbol table is already binary file address

src/cc/bcc_syms.cc
yonghong-song: The return value -1 should already indicate the error. The user should not count on any particular "sym" values. BTW, this is per user space, there is no security concerns.

src/cc/bcc_syms.cc
yonghong-song: Could you add a comment like this? This will show all cases are addressed.

palmtenor:
Address comments

goldshtn:
Just tested this PR and it doesn't work in my case. This is the memory map for libcoreclr.so:

```
cat /proc/$(pidof Simple)/maps | grep libcoreclr.so
7f22a8fe6000-7f22a923d000 r-xp 00000000 ca:01 547279                     /home/fedora/Simple/out/libcoreclr.so
7f22a923d000-7f22a923e000 rwxp 00257000 ca:01 547279                     /home/fedora/Simple/out/libcoreclr.so
7f22a923e000-7f22a95dc000 r-xp 00258000 ca:01 547279                     /home/fedora/Simple/out/libcoreclr.so
7f22a95dc000-7f22a95dd000 rwxp 005f6000 ca:01 547279                     /home/fedora/Simple/out/libcoreclr.so
7f22a95dd000-7f22a9692000 r-xp 005f7000 ca:01 547279                     /home/fedora/Simple/out/libcoreclr.so
7f22a9692000-7f22a96cd000 rw-p 006ac000 ca:01 547279                     /home/fedora/Simple/out/libcoreclr.so
```

Symbols in the first r-xp range are resolved fine, symbols in the second r-xp range are not resolved successfully. Here's a sample broken stack as reported by `trace`:

```
19398  19398  Simple       _ZN3WKS6GCHeap24GarbageCollectGenerationEj9gc_reason
7f22a9354f80 [unknown] [perf-19398.map]
7f22a9163b3c GCInterface::Collect(int, int)+0x8c [libcoreclr.so]
7f2230070b89 void [Simple] Simple.Program::Main(string[])+0xa9 [perf-19398.map]
7f22a923cf07 CallDescrWorkerInternal+0x7c [libcoreclr.so]
7f22a914dce0 MethodDescCallSite::CallTargetWorker(unsigned long const*, unsigned long*, int)+0x4f0 [libcoreclr.so]
7f22a925ec03 [unknown] [perf-19398.map]
7f22a925eea3 [unknown] [perf-19398.map]
7f22a90903fb CorHost2::ExecuteAssembly(unsigned int, char16_t const*, int, char16_t const**, unsigned int*)+0x1db [libcoreclr.so]
7f22a906add6 coreclr_execute_assembly+0x106 [libcoreclr.so]
7f22a9788433 coreclr::execute_assembly(void*, unsigned int, int, char const**, char const*, unsigned int*)+0x83 [libhostpolicy.so]
7f22a977d0d8 run(arguments_t const&)+0x1758 [libhostpolicy.so]
7f22a977d772 corehost_main+0x262 [libhostpolicy.so]
7f22a9a3d8f4 execute_app(std::string const&, corehost_init_t*, int, char const**)+0xd4 [libhostfxr.so]
7f22a9a47978 fx_muxer_t::read_config_and_execute(std::string const&, std::string const&, std::unordered_map<std::string, std::vector<std::string, std::allocator<std::string> >, std::hash<std::string>, std::equal_to<std::string>, std::allocator<std::pair<std::string const, std::vector<std::string, std::allocator<std::string> > > > > const&, int, char const**, host_mode_t)+0xfb8 [libhostfxr.so]
7f22a9a468f7 fx_muxer_t::parse_args_and_execute(std::string const&, std::string const&, int, int, char const**, bool, host_mode_t, bool*)+0x7b7 [libhostfxr.so]
7f22a9a47eb9 fx_muxer_t::execute(int, char const**)+0x2c9 [libhostfxr.so]
7f22a9a3d975 hostfxr_main+0x35 [libhostfxr.so]
40a2bf run(int, char const**)+0x26f [Simple]
40a407 main+0xa7 [Simple]
7f22a9c94431 __libc_start_main+0xf1 [libc-2.24.so]
```

Note how the 7f22a925eea3 address is not resolved correctly to a symbol; it is in the second r-xp section in the /proc/PID/maps output. I can even find it with objdump by looking at the offset from the start of the binary. 7f22a925eea3 (the symbol we're looking for) - 7f22a8fe6000 (the start load address of libcoreclr.so) = 278ea3, and that falls within the following method, which also makes sense as it is called by `CorHost2::ExecuteAssembly`:

```
$ objdump -tT out/libcoreclr.so | grep ExecuteMainMethod | c++filt
0000000000278dc0 l     F .text  00000000000003e8            Assembly::ExecuteMainMethod(PtrArray**, int)
```

To reproduce, create the following C# program:

```C#
using System;
using System.IO;
using System.Threading;

namespace Simple
{
    class Program
    {
        static void Main(string[] args)
        {                                                                                           
                while (true)                                                                         
                {
                        File.WriteAllText("hello.txt", "Hello, World!");                                                                               
                        Console.Write("."); 
                        Thread.Sleep(1000); 
                        GC.Collect();                                                               
                }                                                                                   
        }                                                                                           
    }                                                                                               
}  
```

And then run it as follows:

```
$ sudo docker run --rm -it -v $PWD:/app -e "COMPlus_PerfMapEnabled=1" microsoft/dotnet \
    sh -c 'cd /app && dotnet publish -c Release -o out --self-contained -r linux-x64 && out/Simple'
```

In another shell, you can try `trace -p $(pidof Simple) /app/out/libcoreclr.so:_ZN3WKS6GCHeap24GarbageCollectGenerationEj9gc_reason -U`, which is my repro for seeing the bad symbol behavior.

palmtenor:
@goldshtn Thanks for the good example. I see, so your issue is on resolving addr to symbol in dynamic library loaded into multiple sections. We (me and @markdrayton) kind of know the issue (see #1168) but didn't fix it completely by adding file offset to the `Range` struct. I actually already have a fix on top of #1358 and I will put it up shortly.

@yonghong-song This PR addresses issue with memory address to binary offset resolution and provides util for #1358. It is related to but not completely the issue @goldshtn described:)

yonghong-song:
Looks the issue mentioned by @goldshtn is not covered by this patch and it will be addressed separately. Merge the patch for now and anticipating additional patches to fix the multiple dynamic lib loading sym resolution issue later.

PR #1361, opened 2017-09-25
hardirqs, softirqs: Fix distribution mode units handling

navytux:
Even if units was μs and thus factor=1000 `hardirqs -d` and `softirqs
-d` were printing actual data in nanoseconds but with distribution unit
labeled as "usecs", e.g.:

```
    softirq = sched
          1024 -> 2047       : 6        |*****                                   |
          2048 -> 4095       : 3        |**                                      |
          4096 -> 8191       : 8        |*******                                 |
          8192 -> 16383      : 31       |*****************************           |
         16384 -> 32767      : 42       |****************************************|
         32768 -> 65535      : 18       |*****************                       |
```

Fix it.

NOTE: not putting "/ FACTOR" into common code because for counting mode
we do not want to round intermediate results as e.g. there could be lots
of say 0.5μs interrupts that should be all accounted as N*0.5, not N*0.0.

### DIFF
--- a/tools/hardirqs.py
+++ b/tools/hardirqs.py
@@ -109,7 +109,7 @@ int trace_completion(struct pt_regs *ctx)
 # code substitutions
 if args.dist:
     bpf_text = bpf_text.replace('STORE',
-        'irq_key_t key = {.slot = bpf_log2l(delta)};' +
+        'irq_key_t key = {.slot = bpf_log2l(delta / %d)};' % factor +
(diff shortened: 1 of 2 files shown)


goldshtn:
[buildbot, ok to test]

goldshtn:
@brendangregg Could you please take a look?

navytux:
ping @brendangregg, @goldshtn

brendangregg:
Good catch, thanks!

navytux:
Thanks for merging.

PR #1363, opened 2017-09-25
bpf_probe_read*: src argument should be const void *.

navytux:
For the following program:

    #include <linux/interrupt.h>

    // remember t(last-interrupt) on interface
    int kprobe__handle_irq_event_percpu(struct pt_regs *ctx, struct irq_desc *desc) {
        const char *irqname = desc->action->name;

        char c;

        bpf_probe_read(&c, 1, &irqname[0]);
        if (c != 'e') return 0;

        bpf_probe_read(&c, 1, &irqname[1]);
        if (c != 't') return 0;

        ...

LLVM gives warnings because irqaction->name is `const char *`:

    /virtual/main.c:10:27: warning: passing 'const char *' to parameter of type 'void *' discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
        bpf_probe_read(&c, 1, &irqname[0]);
                              ^~~~~~~~~~~
    /virtual/main.c:13:27: warning: passing 'const char *' to parameter of type 'void *' discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
        bpf_probe_read(&c, 1, &irqname[1]);
                              ^~~~~~~~~~~
    ...

Instead of adding casts in source everywhere fix bpf_probe_read* signature to
indicate the memory referenced by src won't be modified, as it should be.

P.S.

bpf_probe_read_str was in fact already marked so in several places in comments
but not in actual signature.

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -234,7 +234,7 @@ Examples in situ:
 
 ### 1. bpf_probe_read()
 
-Syntax: ```int bpf_probe_read(void *dst, int size, void *src)```
+Syntax: ```int bpf_probe_read(void *dst, int size, const void *src)```
 
 Return: 0 on success
 
@@ -246,7 +246,7 @@ Examples in situ:
 
 ### 2. bpf_probe_read_str()
 
-Syntax: ```int bpf_probe_read_str(void *dst, int size, void *src)```
+Syntax: ```int bpf_probe_read_str(void *dst, int size, const void *src)```
 
 Return:
   - \> 0 length of the string including the trailing NUL on success
--- a/src/cc/compat/linux/bpf.h
+++ b/src/cc/compat/linux/bpf.h
@@ -244,7 +244,7 @@ union bpf_attr {
  * int bpf_map_delete_elem(&map, &key)
  *     Return: 0 on success or negative error
  *
- * int bpf_probe_read(void *dst, int size, void *src)
+ * int bpf_probe_read(void *dst, int size, const void *src)
  *     Return: 0 on success or negative error
  *
  * u64 bpf_ktime_get_ns(void)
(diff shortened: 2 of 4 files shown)


yonghong-song:
[buildbot, ok to test]

docs/reference_guide.md
```diff
@@ -234,7 +234,7 @@ Examples in situ:
 
 ### 1. bpf_probe_read()
 
-Syntax: ```int bpf_probe_read(void *dst, int size, void *src)```
+Syntax: ```int bpf_probe_read(void *dst, int size, const void *src)```
```
yonghong-song: The below is kernel uapi/linux/bpf.h definitions:

linux/bpf.h: * int bpf_probe_read(void *dst, int size, void *src)
linux/bpf.h: * int bpf_probe_read_str(void *dst, int size, const void *unsafe_ptr)

I know it is a little bit annoying so it is better to keep the func signature in line with
the kernel tree.

I will check whether we could add "const" modifier to "void *src" in kernel:include/uapi/linux/bpf.h.

navytux: Thanks for feedback. BPF is new to me and I too noticed after sending the patch that part of the changes should go first to the linux.git tree as `src/cc/compat/linux/bpf.h` comes copied from there.

Please let me know the outcome of your study. (imho it is ok and should be done there and not only for 2 functions touched in my patch but for others as well - e.g. for `bpf_probe_write_user()` for `void *src` too).

yonghong-song:
cc @4ast

navytux:
@yonghong-song, @4ast thanks for applying the patch. Kirill.

P.S. for topics with only 1 patch maybe it makes more sense just to apply the patch without creating merge commit. Github allows this in web UI nowdays.

PR #1365, opened 2017-09-26
Add basic USDT support for powerpc64

sandip4n:
Add support for parsing and reading USDT arguments in formats compatible with powerpc64. This also updates some of the USDT related tests such as `test_usdt_args`, `test_usdt` and `test_usdt2` so that they pass on powerpc64 systems.

### DIFF
--- a/tests/python/include/folly/tracing/StaticTracepoint-ELF.h
+++ b/tests/python/include/folly/tracing/StaticTracepoint-ELF.h
@@ -18,7 +18,11 @@
 
 // Default constraint for the probe arguments as operands.
 #ifndef FOLLY_SDT_ARG_CONSTRAINT
+#if defined(__powerpc64__) || defined(__powerpc__)
+#define FOLLY_SDT_ARG_CONSTRAINT      "nQr"
+#elif defined(__x86_64__) || defined(__i386__)
 #define FOLLY_SDT_ARG_CONSTRAINT      "nor"
+#endif
 #endif
 
 // Instruction to emit for the probe.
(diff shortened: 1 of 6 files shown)


goldshtn:
[buildbot, ok to test]

tests/python/include/folly/tracing/StaticTracepoint-ELF.h
```diff
@@ -18,8 +18,12 @@
 
 // Default constraint for the probe arguments as operands.
```
palmtenor: Well, this file is not `StaticTracepoint-ELF.h` per say:) It's just x86 and PowerPC. How about we make a `StaticTracepoint-Impl.h` and have x86 and Power PC has their own header and define the `FOLLY_SDT_ARG_CONSTRAINT` and `FOLLY_SDT_ARGTMPL` there?
Also, we we are done, it would be great if you could contribute this change to [folly](https://github.com/facebook/folly) as well!

sandip4n: I modeled this after SystemTap's `sys/sdt.h` where `STAP_SDT_ARG_CONSTRAINT` is set to `nor` for all architectures. However, for use with `bcc` running on `powerpc64`, we would prefer using `nQr` instead. So, would it be okay if I just removed `#elif defined(__x86_64__) || defined(__i386__)` from line 23 of `StaticTracepoint-ELF.h` and replace it with an `#else`?

As for contributing the changes to [folly](https://github.com/facebook/folly), I will definitely look into it.

sandip4n: @palmtenor ping

PR #1368, opened 2017-09-28
examples: fix dns_matching

pbhole:
Fixed major bugs in this example.

Please note that this example still doesn't work as intended because of its design.
AFAIK original author expected it to drop dns packets having dns names which are also present in map. (see description in .c file).
I will create another PR to make it work as a DNS sniffer.

This will still remain an example of loop unrolling.


goldshtn:
[buildbot, ok to test]

PR #1369, opened 2017-09-29
execsnoop: argument to change the number of arguments parsed

pchaigno:
New argument to change the maximum number of arguments parsed and
displayed.

~The maximum number is currently limited to 188 due to the maximum
number of instructions authorized by the verifier. The new smoke test
ensures we get a failing test if the actual maximum decreases.~

~Note: I proactively limited the number of arguments in argparse to
avoid exposing the user to the verifier's error message.~

/cc @brendangregg

### DIFF
--- a/man/man8/execsnoop.8
+++ b/man/man8/execsnoop.8
@@ -35,6 +35,9 @@ Only print command lines matching this name (regex)
 .TP
 \-l LINE
 Only print commands where arg contains this line (regex)
+.TP
+\--max-args MAXARGS
+Maximum number of arguments parsed and displayed, defaults to 20
 .SH EXAMPLES
 .TP
 Trace all exec() syscalls:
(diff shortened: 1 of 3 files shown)


yonghong-song:
ubuntu 16 failure:
> 28: Exception: Failed to load BPF program kprobe__sys_execve: Invalid argument
I guess some version of llvm compiler is smart enough to unroll the loop.

pchaigno:
Should be fixed.

I removed the limit on the maximum number of arguments since we cannot know what the actual limit (from the verifier) will be. It depends on how the loop content is compiled. Don't know what I was thinking.

yonghong-song:
@brendangregg @goldshtn could you check the patch? This involves an user interface change.

brendangregg:
Thanks. It needs an entry on the man page.

I also tried it with "--max-args 1000", and I think the error message is good enough for now:

```
bpf: Invalid argument. Program too large (7741 insns), at most 4096 insns
```

pchaigno:
> It needs an entry on the man page.

Fixed.

> I also tried it with "--max-args 1000", and I think the error message is good enough for now:

Hm, I guess it's not so bad given the option. We could do better if the bcc API exposed the error messages returned by the verifier.

drzaeus77:
[buildbot, test this please]

brendangregg:
Thanks!

PR #1372, opened 2017-09-29
add debug option to dump asm insns embedded with source

yonghong-song:
The patch adds a new debug option "DEBUG_SOURCE = 8" to
dump insns embedded with source. In C++ API, users
can change BPF constructor "flag" value to enable debug output.
In Python API, users can change "debug" value to enable
debug output. For example, for python test program test_usdt.py,
the debug output looks like below:

......
Disassembly of section .bpf.fn.do_trace1:
do_trace1:
; int do_trace1(struct pt_regs *ctx) { // Line 110
   0:   bf 16 00 00 00 00 00 00         r6 = r1
   1:   b7 01 00 00 00 00 00 00         r1 = 0
; struct probe_result_t1 result = {}; // Line 111
   2:   7b 1a f0 ff 00 00 00 00         *(u64 *)(r10 - 16) = r1
; switch(ctx->ip) { // Line   5
   3:   79 61 80 00 00 00 00 00         r1 = *(u64 *)(r6 + 128)
   4:   15 01 04 00 d7 06 40 00         if r1 == 4196055 goto 4
   5:   55 01 06 00 ce 06 40 00         if r1 != 4196046 goto 6
; case 0x4006ceULL: *((int8_t *)dest) = ctx->ax; __asm__ __volatile__("": : :"memory"); return 0; // Line   6
   6:   79 61 50 00 00 00 00 00         r1 = *(u64 *)(r6 + 80)
......

For asm insns, byte code is also dumped out (similar to objdump).
For source codes, only lines in the module file are printed (as expected).
The line number is added at the end of source code, which is
especially helpful for inlined functions.

Currently, the debug output only goes to stderr.
A subsequent patch will dump the per-function output into
<BCC_PROG_TAG_DIR>/bpf_prog_<tag>/ if it is available.

### DIFF
--- a/src/cc/CMakeLists.txt
+++ b/src/cc/CMakeLists.txt
@@ -17,6 +17,9 @@ configure_file(libbcc.pc.in ${CMAKE_CURRENT_BINARY_DIR}/libbcc.pc @ONLY)
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC -DBCC_PROG_TAG_DIR='\"${BCC_PROG_TAG_DIR}\"'")
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
 
+string(REGEX MATCH "^([0-9]+).*" _ ${LLVM_PACKAGE_VERSION})
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DLLVM_MAJOR_VERSION=${CMAKE_MATCH_1}")
+
 include(static_libstdc++)
 
 add_library(bpf-static STATIC libbpf.c perf_reader.c)
@@ -25,6 +28,10 @@ add_library(bpf-shared SHARED libbpf.c perf_reader.c)
 set_target_properties(bpf-shared PROPERTIES OUTPUT_NAME bpf)
 
 set(bcc_common_sources bpf_common.cc bpf_module.cc exported_files.cc)
+if (${LLVM_PACKAGE_VERSION} VERSION_GREATER "5")
+  set(bcc_common_sources ${bcc_common_sources} bcc_debug.cc)
+endif()
+
 set(bcc_table_sources table_storage.cc shared_table.cc bpffs_table.cc json_map_decl_visitor.cc)
 set(bcc_util_sources ns_guard.cc common.cc)
 set(bcc_sym_sources bcc_syms.cc bcc_elf.c bcc_perf_map.c bcc_proc.c)
--- a/src/cc/bcc_debug.cc
+++ b/src/cc/bcc_debug.cc
@@ -1,0 +1,213 @@
+/*
+ * Copyright (c) 2017 Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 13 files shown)


4ast:
awesome! super useful.

yonghong-song:
Need to check the builder failure. looks like bpfdisassembler library is not available on the test machine. Will try to reproduce locally.

yonghong-song:
bpfdisassembler library is not available for 3.7/3.8/3.9 and only available on 4.0 and later.
need to find a way in cmake to differentiate different llvm/clang-dev versions.

yonghong-song:
cc @drzaeus77 Could you take a look?

drzaeus77:
@yonghong-song What are you asking me to look at? Do you want a buildbot with a newer clang? That will take some time.

As a side note, please look at the style-check test, there are some clang-format complaints.

src/cc/CMakeLists.txt
```diff
@@ -17,6 +17,13 @@ configure_file(libbcc.pc.in ${CMAKE_CURRENT_BINARY_DIR}/libbcc.pc @ONLY)
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC -DBCC_PROG_TAG_DIR='\"${BCC_PROG_TAG_DIR}\"'")
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
 
+# set LLVM major version number, only care about version 6 and above
```
drzaeus77: This will break when 7 comes out? This check seems hacky, why does it need to be 0?

src/cc/CMakeLists.txt
```diff
@@ -25,12 +32,16 @@ add_library(bpf-shared SHARED libbpf.c perf_reader.c)
...
+  set(bcc_common_sources ${bcc_common_sources} bcc_debug.cc)
+endif()
+
 set(bcc_table_sources table_storage.cc shared_table.cc bpffs_table.cc json_map_decl_visitor.cc)
 set(bcc_util_sources ns_guard.cc common.cc)
 set(bcc_sym_sources bcc_syms.cc bcc_elf.c bcc_perf_map.c bcc_proc.c)
 set(bcc_common_headers libbpf.h perf_reader.h)
 set(bcc_table_headers file_desc.h table_desc.h table_storage.h)
-set(bcc_api_headers bpf_common.h bpf_module.h bcc_exception.h bcc_syms.h)
+set(bcc_api_headers bpf_common.h bpf_module.h bcc_exception.h bcc_syms.h bcc_debug.h)
```
drzaeus77: Why is bcc_debug.h an external API?

src/cc/bcc_debug.cc
```diff
@@ -0,0 +1,208 @@
...
+    if (mod_src_[end] == '\n' || end == FileBufSize - 1 ||
+        (mod_src_[end] == '\r' && mod_src_[end + 1] == '\n')) {
+      // Not including the endline
+      LineCache.push_back(string(mod_src_.substr(start, end - start)));
+      if (mod_src_[end] == '\r')
+        end++;
+      start = end + 1;
+    }
+
+  return std::move(LineCache);
```
drzaeus77: Shouldn't return value optimization take care of this without std::move?

src/cc/CMakeLists.txt
yonghong-song: LLVM version 7 is a good point. Will use the precise the version number.

src/cc/CMakeLists.txt
yonghong-song: Will remove it.

src/cc/bcc_debug.cc
yonghong-song: Could be. Using std::move just for safety. Will remove it.

Oh, yes, I am asking for a review 👍  Will look at clang format as well.

yonghong-song:
@drzaeus77 patch has been updated to address review comments. Please take a look!

PR #1374, opened 2017-10-02
 examples:dns_matching: make it work as a DNS sniffer

pbhole:
Also mentioned about this in my previous PR.

### DIFF
--- a/examples/networking/dns_matching/dns_matching.py
+++ b/examples/networking/dns_matching/dns_matching.py
@@ -9,6 +9,7 @@ import socket
 import os
 import struct
 import dnslib
+import argparse
 
 
 def encode_dns(name):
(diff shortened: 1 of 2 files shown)


yonghong-song:
[buildbot, ok to test]

yonghong-song:
The current example is not self-contained. For example, it is not clear how to set up the environment to have some meaningful output for the test. Could you help fix it? A complete example will be much more useful. You can either use pyroute2 python library or you can use ip utilities.

examples/networking/dns_matching/dns_matching.py
```diff
@@ -46,12 +52,34 @@ def encode_dns(name):
...
+  UDP_HLEN = 8
+
+  #IP HEADER
+  #calculate ip header length
+  ip_header_length = packet_bytearray[ETH_HLEN]               #load Byte
+  ip_header_length = ip_header_length & 0x0F                  #mask bits 0..3
+  ip_header_length = ip_header_length << 2                    #shift to obtain length
+
+  #calculate payload offset
+  payload_offset = ETH_HLEN + ip_header_length + UDP_HLEN
```
yonghong-song: How do you know it must be UDP packet?

pbhole: Because bpf program is filtering-in only UDP packets (only those containing DNS question).

pbhole:
@yonghong-song Thank you for reviewing. I will make it a complete example. However I did not get the context of your suggestion regarding use of pyroute2/ip utilites. Can you please elaborate?

examples/networking/dns_matching/dns_matching.py
yonghong-song: Oh, right. the bpf program drops everything except dns packets.

yonghong-song:
For pyroute2/ip utilities, I mean that you can use them to build a complete test case. But you are free to use whatever utilities to build your example. Thanks!

pbhole:
PR updated. Please review.

Issue #1054, opened 2017-03-19
ksymname doesn't work


tejom:
I started trying out this library recently. I ran into a problem when trying to the python function `ksymname`. 

The error I get is:

        ("kernel address: %x" % b.ksymname("vfs_read"))
      File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 1037, in ksymname
        return BPF._sym_cache(-1).resolve_name(None, name)
      File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 75, in resolve_name
        self.cache, module.encode("ascii"),
    AttributeError: 'NoneType' object has no attribute 'encode'

resolve_name function is called with the parameters None and name. resolve_name tries to use a function encode on module_name. I could create a pull request that checks the type before using encode on the parameters passed to resolve_name?

While I am making a post I might as well ask... What I was trying to do is call some kernel function that already exists. Including the header gave me an llvm error saying the function was external. I thought I might able to get the address of the function and use bpf_probe_read with a function pointer and the symbol address. That doesn't seem to work. Is what I am trying to do sound possible?

goldshtn:
Yes, this is a bug -- I think it was introduced in 7b35436a when adding the `.encode`, but on my current Internet connection I can barely run `git pull` so I'll check later. 

Regarding your other question, you can't call arbitrary kernel functions from your BPF program. You can only use a closed set of kernel helpers like `bpf_ktime_get_ns`. Unless I'm misunderstanding something about your question?

tejom:
Yeah that looks like it could be where it started.

And no I think you understood my question. I had a feeling that was true but thanks for confirming that.

goldshtn:
OK, so this will be fixed in #1055. In the meantime, could you share a bit more detail on which function you're trying to trace and extract arguments from? Did you take a look at `fileslower`? It has some code for inspecting `file->f_path.dentry->d_name`.

tejom:
Cool and sure I can.

I was trying something similar to that with ext4_file_open. Some of the variables I tried to print previously out gave me errors needing to use bpf_probe_read. I started using that to access any variable. When I used bpf_probe_read with f_path,dentry or d_name I got an error about lvalue(or rvalue? I don't have the original code I tried anymore) 

I took a look at the tools again and saw that one you mentioned and accessing file->f_path.dentry->d_name without using bpf_probe_read does work for me for getting the file name.

[...]

PR #1376, opened 2017-10-05
Fix 'tools/statsnoop' from failing to attach kprobes

sandip4n:
This fixes `tools/statsnoop` from failing to attach probes when the expected entry point for a system call cannot be found. This script uses the `stat`, `statfs` and `newstat` system calls, all of which must be implemented to be POSIX compliant. However, the names of the actual entry points for their respective implementations in the kernel might vary across architectures. For example, a `powerpc64` kernel does not define `sys_stat` but still provides the `stat` system call via `sys_newstat`. This causes the script to fail if it tries to attach a probe at `sys_stat`. We avoid this by performing some extra checks to see if these entry points exist.

### DIFF
--- a/tools/statsnoop.py
+++ b/tools/statsnoop.py
@@ -105,12 +105,22 @@ if debug:
 
 # initialize BPF
 b = BPF(text=bpf_text)
-b.attach_kprobe(event="sys_stat", fn_name="trace_entry")
-b.attach_kprobe(event="sys_statfs", fn_name="trace_entry")
-b.attach_kprobe(event="sys_newstat", fn_name="trace_entry")
-b.attach_kretprobe(event="sys_stat", fn_name="trace_return")
-b.attach_kretprobe(event="sys_statfs", fn_name="trace_return")
-b.attach_kretprobe(event="sys_newstat", fn_name="trace_return")
+
+# for POSIX compliance, all architectures implement these
+# system calls but the name of the actual entry point may
+# be different for which we must check if the entry points
+# actually exist before attaching the probes
+if BPF.ksymname("sys_stat") != -1:
+    b.attach_kprobe(event="sys_stat", fn_name="trace_entry")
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, ok to test]

brendangregg:
thanks!

nathanleclaire:
Hi @sandip4n, FYI, this causes an error on my machine (Linux 4.9.0 Debian) which does not happen without the change:

```
$ sudo ./tools/statsnoop.py                                              
  File "./tools/statsnoop.py", line 113, in <module>                        
    if BPF.ksymname("sys_stat") != -1:                                                 
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 1035, in ksymname      
    return BPF._sym_cache(-1).resolve_name(None, name)                                 
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 74, in resolve_name 
    self.cache, module.encode("ascii"),                                              
AttributeError: 'NoneType' object has no attribute 'encode'    
```

I'm not sure if it's a bug or a problem on my end (e.g., needing to re-install BCC libraries), but I wanted to bring it to your awareness.

pchaigno:
@nathanleclaire This bug was reported in #1054 and should be fixed with #1055 (which, according to your traceback, you don't have).

sandip4n:
Hi @nathanleclaire,
If you look at the latest code for `src/python/bcc/__init__.py`, you will notice that there is already a check in line 85 to test if `module` is `None` introduced in #1055 as pointed out by @pchaigno.
```
if lib.bcc_symcache_resolve_name(
            self.cache, module.encode("ascii") if module else None,
            name.encode("ascii"), ct.pointer(addr)) < 0:
    return -1
```
According to the traceback you provided, this is exactly where the script crashes on your system even though the line numbers might not match because of later changes to this file. So, checking out the latest code and re-installing might fix your problem.

nathanleclaire:
ok great, thanks y'all!

Issue #1373, opened 2017-10-01
offcputime segfault


diamond-lizard:
offcputime segfaults for me when I run it on Gentoo Linux amd64 with python3.4:

```
# offcputime -f 30
kworker/0:3;return_from_SYSCALL_64;do_syscall_64;sys_execve;do_execveat_common.isra.14;search_binary_handler;load_elf_binary;kfree;load_elf_binary;load_elf_binary;retint_kernel;preempt_schedule_irq;load_elf_binary;__schedule;finish_task_switch 12
bash;[unknown];[unknown];retint_user;prepare_exit_to_usermode;exit_to_usermode_loop;schedule;__schedule;finish_task_switch 13
zsh: segmentation fault  /usr/share/bcc/tools/offcputime -f 30
```
Here's a backtrace:
```
#0  0x00007ffff5c62b64 in KSyms::resolve_addr(unsigned long, bcc_symbol*, bool) () from /usr/lib64/libbcc.so.0
#1  0x00007ffff5c6319b in bcc_symcache_resolve_no_demangle () from /usr/lib64/libbcc.so.0
#2  0x00007ffff65df6f6 in ffi_call_unix64 () from /usr/lib64/libffi.so.6
#3  0x00007ffff65df01d in ffi_call () from /usr/lib64/libffi.so.6
#4  0x00007ffff683e93e in _ctypes_callproc () from /usr/lib64/python3.4/lib-dynload/_ctypes.cpython-34m.so
#5  0x00007ffff68388cf in ?? () from /usr/lib64/python3.4/lib-dynload/_ctypes.cpython-34m.so
#6  0x00007ffff799573a in PyObject_Call () from /usr/lib64/libpython3.4m.so.1.0
#7  0x00007ffff7a526da in PyEval_EvalFrameEx () from /usr/lib64/libpython3.4m.so.1.0
#8  0x00007ffff7a54670 in PyEval_EvalFrameEx () from /usr/lib64/libpython3.4m.so.1.0
#9  0x00007ffff7a55c3b in PyEval_EvalCodeEx () from /usr/lib64/libpython3.4m.so.1.0
#10 0x00007ffff7a533a4 in PyEval_EvalFrameEx () from /usr/lib64/libpython3.4m.so.1.0
#11 0x00007ffff7a55c3b in PyEval_EvalCodeEx () from /usr/lib64/libpython3.4m.so.1.0
[...]

PR #1377, opened 2017-10-05
Avoid potential SEGFAULT when resolving Kernel symbols

palmtenor:
This is an attempt to address #1373. I'm not sure why it would happen or if it's indeed the reason of that `SEGFAULT` but it definitely could be.

@diamond-lizard If you could try if this patch solves your issue?


diamond-lizard:
I just pulled from master, rebuilt, and was able to run `offcputime` without segfault.  Thank you.

PR #1378, opened 2017-10-05
Improve Kernel symbols loading

palmtenor:
- Current logic basically ignores the first line. That could be related to the reason of #1373 / #1377.
- Add some checks.
- Ignore data symbols as we only care about the relationship between Kernel function addresses and function names, for stack-trace symbol resolution and kprobe attaching. Reduces memory and CPU cost. The ignored types come from `man nm`:
```
           "B"
           "b" The symbol is in the uninitialized data section (known as BSS).

           "D"
           "d" The symbol is in the initialized data section.

           "R"
           "r" The symbol is in a read only data section.
```

### DIFF
--- a/src/cc/bcc_proc.c
+++ b/src/cc/bcc_proc.c
@@ -17,19 +17,27 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <sys/mman.h>
-#include <fcntl.h>
-#include <unistd.h>
-#include <stdlib.h>
-#include <stdbool.h>
-#include <string.h>
-#include <stdint.h>
 #include <ctype.h>
-#include <stdio.h>
+#include <fcntl.h>
+#include <limits.h>
 #include <math.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
 
 #include "bcc_perf_map.h"
 #include "bcc_proc.h"
 #include "bcc_elf.h"
+
+#ifdef __x86_64__
+// https://www.kernel.org/doc/Documentation/x86/x86_64/mm.txt
+const unsigned long long kernelAddrSpace = 0x00ffffffffffffff;
+#else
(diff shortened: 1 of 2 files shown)


src/cc/bcc_proc.c
```diff
@@ -149,21 +154,23 @@ int bcc_procutils_each_ksym(bcc_procutils_ksymcb callback, void *payload) {
   if (!kallsyms)
     return -1;
 
-  if (!fgets(line, sizeof(line), kallsyms)) {
```
palmtenor: This line is gone......

src/cc/bcc_proc.c
```diff
@@ -17,20 +17,24 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
 
 #include "bcc_perf_map.h"
 #include "bcc_proc.h"
 #include "bcc_elf.h"
 
+// https://www.kernel.org/doc/Documentation/x86/x86_64/mm.txt
+const unsigned long long kernelAddrSpace = 0x00ffffffffffffff;
```
yonghong-song: How about other architectures?

src/cc/bcc_proc.c
yonghong-song: What the first line shows us? There must be a reason to ignore it in the first place?

src/cc/bcc_proc.c
palmtenor: Good point. I couldn't find any similar file in other architecture's folder, so I will just gate this for x64

src/cc/bcc_proc.c
palmtenor: I don't know. It looks like totally normal symbol on my machine.....

yonghong-song: It looks like it depends on CONFIG_KALLSYMS_ALL.
If the config is on, all init, text and data symbols will be included.
Otherwise, data symbols are executed (except a few special symbols in the linker script file).

PR #1379, opened 2017-10-06
 libbpf: return error when if_nametoindex returns 0 

pbhole:


yonghong-song:
[buildbot, ok to test]

Issue #1199, opened 2017-05-23
Support parsing vDSO symbols


palmtenor:
BCC doesn't understand vDSO addresses for symboling now. This is causing unnecessary unknown symbols in production runs. We should support that.

cc @markdrayton @goldshtn

palmtenor:
Resolved with #1382

PR #1382, opened 2017-10-07
Support resolve vDSO symbols

palmtenor:
This PR adds the ability for BCC to resolve `vDSO` symbol. We write the content of the `vDSO` memory to a temp file once per BCC Process, and use it to resolve symbols later. Also does some clean-up on proc map parsing.

Tested by making a program that does `vDSO` function calls and run `profile.py` on it
Before
```
    [unknown]
    do_read()
    main
    __libc_start_main
    -                testvdso (4154018)
        3463
```
After
```
    gettimeofday
    do_read()
    main
    __libc_start_main
    -                testvdso (4154018)
        30
```

Addresses #1199 

cc @myreg @markdrayton @brendangregg

### DIFF
--- a/src/cc/bcc_elf.c
+++ b/src/cc/bcc_elf.c
@@ -16,6 +16,7 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <sys/mman.h>
+#include <errno.h>
 #include <fcntl.h>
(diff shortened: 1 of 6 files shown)


goldshtn:
LGTM.

PR #987, opened 2017-02-19
Travis CI build to check compliance with PEP8

pchaigno:
This Travis CI build only checks the compliance with PEP8 (ignoring E123, E125, E126, E127, E128, and E302).

It only runs `pep8` against Python files in `tools/`. Should I extend it to all Python files in the repository?

I installed `pep8` using pip to get the last version; the version in Ubuntu 12.04 packages misses a few errors.

/cc @brendangregg

PR #920, opened 2017-01-23
gethostlatency: Filter by process ID

pchaigno:
This pull request adds an option to `gethostlatency.py` to only trace processes with a given PID.

/cc @brendangregg

PR #1383, opened 2017-10-07
Linter cleanups

pchaigno:
Would it be possible to enable Travis CI for the repo **before** merging this pull request?

### DIFF
--- a/tools/bashreadline.py
+++ b/tools/bashreadline.py
@@ -56,7 +56,8 @@ print("%-9s %-6s %s" % ("TIME", "PID", "COMMAND"))
 
 def print_event(cpu, data, size):
     event = ct.cast(data, ct.POINTER(Data)).contents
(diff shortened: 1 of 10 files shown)


goldshtn:
Paddin my ignorance, what would we gain by having Travis CI build this repository? We already have three Jenkins build bots.

goldshtn:
Paddin = pardon, silly phone :)

pchaigno:
I'm :+1: with having this as part of the Jenkins' pipeline.

I initially went for Travis CI (proposed in #920, added in #987) because, contrary to the other tests, lint checks should not depend on the underlying OS version and we thus don't need to start a VM; a Travis container is faster. I also really appreciate the integration with GitHub. In particular, the configuration is part of the source code, which doesn't seem to be the case for the current Jenkins build bots.

4ast:
LGTM. Merge ready? cc @drzaeus77

pchaigno:
Without Travis-CI (or some other CI) enabled, we'll probably have to do this again in a few weeks...

PR #1384, opened 2017-10-08
mdflush: fix for Linux 4.14

pchaigno:
[Starting Linux 4.14](https://github.com/torvalds/linux/commit/74d46992e0d9dee7f1f376de0d56d31614c8a17a), there's one less indirection layer to get the
disk name.

### DIFF
--- a/tools/mdflush.py
+++ b/tools/mdflush.py
@@ -36,8 +36,19 @@ int kprobe__md_flush_request(struct pt_regs *ctx, void *mddev, struct bio *bio)
     u32 pid = bpf_get_current_pid_tgid();
     data.pid = pid;
     bpf_get_current_comm(&data.comm, sizeof(data.comm));
+/*
+ * The following deals with a kernel version change (in mainline 4.14, although
+ * it may be backported to earlier kernels) with how the disk name is accessed.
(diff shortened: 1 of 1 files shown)


yonghong-song:
This probably will not work with backported distribution (fb internal, netflix internal, certain redhat dist.) I am not sure what is the standard way to deal with this.

cc @brendangregg

brendangregg:
So far we've gotten away with checking if symbols exist rather than kernel versions. But in this case it's a structure difference. I wonder if there's a symbol change from the same patchset we could use as an indicator?

In any case, this should be commented to say that it's temporary code. "# workaround for the 4.14 change." That way, we can find it in the distant future and delete it when it's no longer needed.

pchaigno:
I hadn't though about backported distribution.
I've updated the pull request to use the `bio-dev` macro as a discriminator to detect the structure difference. That macro was introduced in the [same commit](https://github.com/torvalds/linux/commit/74d46992e0d9dee7f1f376de0d56d31614c8a17a) as the structure change.

@drzaeus77 Could you relaunch the build? Looks like it failed due to the certificate expiration.

yonghong-song:
LGTM. I think this addressed Brendan's concern.

PR #1388, opened 2017-10-11
Fix 'tools/syscount' from using incorrect fallback values

sandip4n:
This prevents `tools/syscount` from using incorrect system call names if the 'ausyscall' command is not found. This is because it does not come pre-installed in all distros. In this case, it uses a lookup table as a fallback but this table is usable for `x86_64` only. For now, we fix this by raising an exception and exiting if the `ausyscall` executable is not available for non-`x86_64` systems instead of having additional lookup tables for other architectures.

### DIFF
--- a/tools/syscount.py
+++ b/tools/syscount.py
@@ -15,6 +15,7 @@ import argparse
 import itertools
 import subprocess
 import sys
+import platform
 
 if sys.version_info.major < 3:
     izip_longest = itertools.izip_longest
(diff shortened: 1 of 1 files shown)


brendangregg:
ok, thanks, makes sense for now.

brendangregg:
[buildbot, ok to test]

yonghong-song:
[buildbot, ok to test]

yonghong-song:
[buildbot, ok to test]

yonghong-song:
@drzaeus77 could you help check buildbot issue? it seems not responding (at least publicly)?

drzaeus77:
Taking a look now...a couple days ago I updated packages and jenkins plugins, as well as rebooted it. It was stuck in some weird state where the bcc-pr job was not spawning slaves.

PR #1390, opened 2017-10-13
attempt to compile with system bpf.h if default compile failed

yonghong-song:
Currently, bcc uses its own version of bpf.h which tries to
sync with upstream header regularly. If the host bpf.h version
is lower, bcc can still compile as some bcc codes may requires
a higher version of bpf.h.

Such an approach does have a drawback. Suppose service A,
statically linked with bcc, runs on kernel version X.
Now, the kernel upgrades to version Y. After kernel upgrade/reboot,
service A may not be able to compile since old bcc bpf.h
may not align with the new kernel headers.
For such cases, new version of service A needs rollout.

This patch addresses this issue by attempting a second
compilation using system bpf.h instead.

### DIFF
--- a/src/cc/frontends/clang/loader.cc
+++ b/src/cc/frontends/clang/loader.cc
@@ -108,8 +108,6 @@ int ClangLoader::parse(unique_ptr<llvm::Module> *mod, TableStorage &ts,
                        const string &file, bool in_memory, const char *cflags[],
                        int ncflags, const std::string &id, FuncSource &func_src,
                        std::string &mod_src) {
-  using namespace clang;
-
   string main_path = "/virtual/main.c";
   unique_ptr<llvm::MemoryBuffer> main_buf;
   struct utsname un;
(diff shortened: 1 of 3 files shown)


yonghong-song:
py_test_clang failed on all three platforms. Will need to debug a little bit more.

yonghong-song:
@drzaeus77  could you help review this patch as well?

src/cc/frontends/clang/loader.cc
```diff
@@ -155,23 +153,58 @@ int ClangLoader::parse(unique_ptr<llvm::Module> *mod, TableStorage &ts,
...
 
+  // SharedTableStorage is global per process and ts may contain tables
+  // from previous modules
+  if (do_compile(mod, ts, in_memory, flags_cstr, flags_cstr_rem, main_path,
+                 main_buf, id, func_src, mod_src, true)) {
+    // try one more time to compile with system bpf.h
+    llvm::errs() << "WARNING: compilation failure, trying use system bpf.h\n";
+
+    func_src.clear();
+    mod_src.clear();
```
drzaeus77: Hmm, this may have other repercussions that may need to be rolled back or stashed temporarily. Would it be possible to get a compile error after the frontend action has run, resulting in some tables being opened twice?

src/cc/frontends/clang/loader.cc
```diff
@@ -155,23 +153,58 @@ int ClangLoader::parse(unique_ptr<llvm::Module> *mod, TableStorage &ts,
...
-  flags_cstr.push_back(cur_cpu_flag.c_str());
+  flags_cstr_rem.push_back(cur_cpu_flag.c_str());
 #endif
 
+  // SharedTableStorage is global per process and ts may contain tables
+  // from previous modules
+  if (do_compile(mod, ts, in_memory, flags_cstr, flags_cstr_rem, main_path,
+                 main_buf, id, func_src, mod_src, true)) {
+    // try one more time to compile with system bpf.h
+    llvm::errs() << "WARNING: compilation failure, trying use system bpf.h\n";
```
drzaeus77: This warning is grammatically incorrect, please rephrase to "WARNING: compilation failure, trying with system bpf.h" or similar.

yonghong-song: Sure.

src/cc/frontends/clang/loader.cc
yonghong-song: Just checked it is possible that fd is created twice for a table. Will try to roll back before the second compile.

I actually thought this is no harm earlier since we may just waste a fd. But think twice, this gets reference counted in kernel and may not be a good idea.

yonghong-song:
This is for a specific type of use case. Yes, I can add a cmake flag for this.

yonghong-song:
test_usdt.py failed on fc25 could be due to ring buffer record lost.

yonghong-song:
[buildbot, ok to test]

yonghong-song:
@drzaeus77 Addressed your comments, could you take a look? Currently, I am using cmake option -DBCC_BACKUP_COMPILE=1 to enable the backup compile (the compile with system bpf.h header).
I did not use BCC_ALT_COMPILE or BCC_COMPILE_WITH_SYSTEM_BPFH since it may sound like that we will change current primary compilation mechanism. Please suggest a new name if you think it is better.

yonghong-song:
[buildbot, ok to test]

yonghong-song:
@drzaeus77 could you help take a look at this. The testing state seems not in sound state.
I accidentally pushed a different patch to the same branch and then I realized that this patch is still under review and then pushed this patch again. But then the test is not running. Thanks!

drzaeus77:
It has to go through each one in the queue, I will monitor it but it seems to be making progress so far...

yonghong-song:
Oh, I see. Thanks! I thought I botched the system.

PR #1393, opened 2017-10-14
Exit on EPIPE

### LINKED ISSUE #1010: handle SIGPIPE for "| head"
Didn't this used to work? I might be thinking of my ftrace version...

```
# ./opensnoop.py | head
2218   snmp-pass           4   0 /proc/cpuinfo
2218   snmp-pass           4   0 /proc/stat
1968   redis-server        5   0 /proc/1968/stat
1651   redis-server        5   0 /proc/1651/stat
1968   redis-server        5   0 /proc/1968/stat
1651   redis-server        5   0 /proc/1651/stat
1968   redis-server        5   0 /proc/1968/stat
1651   redis-server        5   0 /proc/1651/stat
1968   redis-server        5   0 /proc/1968/stat
  File "/usr/lib/python2.7/dist-packages/bcc/table.py", line 522, in <lambda>
    fn = _RAW_CB_TYPE(lambda _, data, size: callback(cpu, data, size))
  File "./opensnoop.py", line 178, in print_event
    event.comm, fd_s, err, event.fname))
IOError: [Errno 32] Broken pipe
  File "/usr/lib/python2.7/dist-packages/bcc/table.py", line 522, in <lambda>
    fn = _RAW_CB_TYPE(lambda _, data, size: callback(cpu, data, size))
  File "./opensnoop.py", line 178, in print_event
    event.comm, fd_s, err, event.fname))
IOError: [Errno 32] Broken pipe
^Cclose failed in file object destructor:
sys.excepthook is missing
lost sys.stderr
```

SIGPIPE should trigger a clean exit. Maybe this should be in libbcc.py as a default?

pchaigno:
Fixes #1010.

/cc @brendangregg

### DIFF
--- a/src/python/bcc/table.py
+++ b/src/python/bcc/table.py
@@ -17,6 +17,7 @@ import ctypes as ct
 from functools import reduce
 import multiprocessing
 import os
+import errno
 
 from .libbcc import lib, _RAW_CB_TYPE, _LOST_CB_TYPE
 from .perf import Perf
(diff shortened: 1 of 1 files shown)


src/python/bcc/table.py
```diff
@@ -523,8 +524,24 @@ def open_perf_buffer(self, callback, page_cnt=8, lost_cb=None):
...
-        fn = _RAW_CB_TYPE(lambda _, data, size: callback(cpu, data, size))
-        lost_fn = _LOST_CB_TYPE(lambda lost: lost_cb(lost)) if lost_cb else ct.cast(None, _LOST_CB_TYPE)
+        def raw_cb_(_, data, size):
+            try:
+                callback(cpu, data, size)
+            except IOError, e:
+                if e.errno == errno.EPIPE:
+                    exit()
+                else:
+                    raise e
```
yonghong-song: The callback function can be arbitrary user function here. Why you single out EPIPE errno from
another other exceptions?

pchaigno: If the developer makes a mistake in his callback function (say, write to a file on which she doesn't have permissions), we want to display the usual traceback, no?

yonghong-song: Any exception should already have a traceback here?

pchaigno: Yes. If I exit on every exception (or ignore some), these tracebacks won't be displayed.

yonghong-song: So why the exit for errno.EPIPE? If you have an example of callback function and/or lost_cb function, probably it is easier to try and understand your point?

pchaigno: The goal is to properly exit when receiving EPIPE or SIGPIPE as described in #1010.
If you take the example from #1010, when running `./opensnoop.py | head`, it will display 10 lines then exception tracebacks for each new line without exiting because of the EPIPE Exception. With the present pull request it will properly exit (call cleanup and not display a traceback) on the first EPIPE Exception.

Issue #462, opened 2016-03-30
add /docs/INSTALL-CENTOS.md


brendangregg:
Should work, provided the kernel is new enough (at least 4.1). If you've figured out the install steps, please drop comments here or a pull request.

hydandata:
http://hydandata.org/installing-ebpf-tools-bcc-and-ply-on-centos-7 but no Lua support. I will get a PR together when I have a bit more time.

dimovnike:
hello, are there any up-to-date docs how to install this on centos 7?  The http://hydandata.org/installing-ebpf-tools-bcc-and-ply-on-centos-7 is outdated, i get this error when following that tutorial:

dimovnike:
Also, would it be possible to use it from a fedora based docker container or a chroot on centos7?

yonghong-song:
@dimovnike If you can show the detailed error messages on your centos 7 installation, folks here may be able to help. Yes, it is possible to use it inside a container. However, some tools esp. PID based filtering may not work properly inside the container.

dimovnike:
i get this error when compiling clang:

> /usr/local/bin/llvm-config: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by /usr/local/bin/llvm-config)


is there a way to use another version that is compatible with centos7 glibc?

hydandata:
@dimovnike looks like issue with LD_LIBRARY_PATH, make sure it contains the path from the warning before you compile. If not, export it with the path added, ldconfig.

samofatov:
I submitted pull request to merge little changes needed for CentOS 6 support.
Here are the few easy steps to install on CentOS 6:
```bash
# Pre-requisites:
# 1) Configure SUDO
# 2) Set up http_proxy and https_proxy for build user and for root (via .bashrc)
#    if there is no direct internet connection

# Install GIT
sudo yum install git

# Create build folder:
cd
mkdir build
cd build

# Detect number of threads
export THREADS=$(grep -c ^processor /proc/cpuinfo)

# Install kernel 4.9.71 from source
wget https://cdn.kernel.org/pub/linux/kernel/v4.x/linux-4.9.71.tar.xz
tar -xf linux-4.9.71.tar.xz
cd linux-4.9.71

# MANUAL copy .config 

# build kernel

make -j $THREADS
sudo make modules_install
sudo make install
sudo dracut --force /boot/initramfs-4.9.71.img 4.9.71
cd ..

# Install GCC
[...]

PR #1395, opened 2017-10-16
adding dockerfile install for centos7

fr0stbyte:
- adds a dockerfile to build the package on centos:7 using the bcc+clang.spec file
- adds instructions on how to do it
- adding a script to build the bcc+clang.spec
- small fix in the bcc+clang spec file to pick up the right version of tarball


alexandrnikitin:
JFYI, worked for me 😉

fr0stbyte:
@brendangregg @goldshtn can you take a look at this please ?

goldshtn:
[buildbot, ok to test]

fr0stbyte:
@goldshtn looks like it's passing build

h1z1:
Curious, why on earth would you run this in a container??

alexandrnikitin:
@h1z1 Say, you are interested in end result, not in installing dependencies, syncing their versions, fixing weird error messages, etc.

fr0stbyte:
@goldshtn this is just for build, not for running. Theoretically, you could also run it from within a container if so inclined, but that's not what this PR is doing.

h1z1:
@fr0stbyte While that is of course correct, problem I have is when projects _rely_ on containers.  It's quite easy to build an environment that technically works yet does nothing but pacify the build system.  #462  is another example of this.

fr0stbyte:
@h1z1 I don't think I understand your apprehension on using containers for build purposes. Can you please detail ?

h1z1:
@fr0stbyte Containers are fine so long as it remains possible to build outside them too.  It's possible for the container to be out of sync with the host or even majority of installs, thus not compatible.

fr0stbyte:
One of the main advantages of the container image is to provide an immutable, reproducible environment. I don't understand what you mean by out-of-sync. 
Which host are you referring to ? The target host on which this rpm gets installed ? 
Isn't that the point of versioned dependencies in the RPM ?

I am sorry, maybe I am missing something obvious, but this doesn't make much sense to me and I am not sure what you're suggesting in terms of changes.

If building in a Docker container is not the right approach, I am totally fine with closing this PR. I thought it would be acceptable, given that fedora / ubuntu are already being built in Docker containers. 🤷‍♂️

h1z1:
No worries.  

> One of the main advantages of the container image is to provide an immutable, reproducible 
 environment. I don't understand what you mean by out-of-sync.
 Which host are you referring to ? The target host on which this rpm gets installed ?
Isn't that the point of versioned dependencies in the RPM ?

In theory yes, docker doesn't really have nor care about that though.  Your install steps for example do not specify any version dependencies within RPM which means yum will pull whatever is the latest.  That may not match the host you're installing on/to.    Running update-alternatives can also lead to unexpected consequences.  In this case the system version of cmake for example is in fact 2.8.  Though 3 should be backward compatible it's not guaranteed.   The correct way to call cmake v3 is usually cmake3.   The current bcc.spec specifies cmake 2.8.7 minimum but runs cmake, no idea what the real requirement is.
 
> If building in a Docker container is not the right approach, I am totally fine with closing this PR. I thought it would be acceptable, given that fedora / ubuntu are already being built in Docker containers.

Not everyone runs Fedora or Ubuntu, certainly not for something like this :)     Docker turns a simple rpmbuild --rebuild foo.src.rpm in a bit more of an ordeal  (docker itself is not small).

Just to be clear, I don't speak for anyone in particular, certainly not Brendan.

f0:
@brendangregg @goldshtn any chance to get this merged? 
Tested this branch, works good

bothra90:
This works like a charm. Can this be merged to master please?

PR #1397, opened 2017-10-16
bpf: print out the src debug info to a temporary file

yonghong-song:
Currently, for C++ API and for each func, the original src
and the rewritten source has been stored in
BCC_PROG_TAG_DIR directory. This patch allows the
source debug info (bytecode embedded by source code)
also stored in the same BCC_PROG_TAG_DIR directory.

This feature is not turned on by default now.
It requires non-zero (debug_flag & DEBUG_SOURCE).
The DEBUG_SOURCE enables "-g", with which a lot of more
llvm insns executed and it may increase application
RSS overhead by 4M (in my test).

As an example, if you modify examples/cpp/RandomRead.cc
to enable DEBUG_SOURCE, as below
-  bpf = new ebpf::BPF();
+  bpf = new ebpf::BPF(8);

After running the application, you can see:
-bash-4.3$ ls /var/tmp/bcc/bpf_prog_7f01346289a53cc3/
on_urandom_read.c  on_urandom_read.dis.txt  on_urandom_read.rewritten.c
-bash-4.3$ cat /var/tmp/bcc/bpf_prog_7f01346289a53cc3/on_urandom_read.dis.txt
; int on_urandom_read(struct urandom_read_args* attr) { // Line  23
   0:	bf 16 00 00 00 00 00 00 	r6 = r1
   1:	b7 01 00 00 00 00 00 00 	r1 = 0
; struct event_t event = {}; // Line  24
   2:	63 1a f8 ff 00 00 00 00 	*(u32 *)(r10 - 8) = r1
   3:	63 1a f4 ff 00 00 00 00 	*(u32 *)(r10 - 12) = r1
.....

### DIFF
--- a/src/cc/bcc_debug.cc
+++ b/src/cc/bcc_debug.cc
@@ -77,14 +77,15 @@ vector<string> SourceDebugger::buildLineCache() {
 
 void SourceDebugger::dumpSrcLine(const vector<string> &LineCache,
                                  const string &FileName, uint32_t Line,
-                                 uint32_t &CurrentSrcLine) {
+                                 uint32_t &CurrentSrcLine,
+                                 llvm::raw_ostream &os) {
   if (Line != 0 && Line != CurrentSrcLine && Line < LineCache.size() &&
       FileName == mod_->getSourceFileName()) {
-    errs() << "; " << StringRef(LineCache[Line - 1]).ltrim()
-           << format(
-                  " // Line"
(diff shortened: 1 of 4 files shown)


palmtenor:
This would be helpful! LGTM!

yonghong-song:
> RSS overhead by 4M
We somehow need to reduce this. Shared llvm library may be an option so that we can dlclose it once we are done ...

PR #529, opened 2016-05-05
Introduce helpers to access pt_regs in an arch-independent manner

rnav:
This addresses some of the review comments on the RFC patch #522 from @goldshtn, @drzaeus77 and @aaronpuchert.

The additional white-space changes in examples/lua/bashreadline.c are for satisfying style-check.

This should help address #225 and #336. Most of the tests/tools now work on powerpc except for those affected by #525.

PR #1398, opened 2017-10-18
Fix helper to access stack pointer for powerpc

sandip4n:
This fixes the definition of `PT_REGS_SP()` for `powerpc`. There is no member called `sp` in `pt_regs` for `powerpc`. According to ELF ABI v2, `GPR1` should be used as the stack pointer.

Fixes: #529


rnav:
Thanks @yonghong-song.
And oops... nasty typo.
The change LGTM. Good catch, @sandip4n!

yonghong-song:
[buildbot, ok to test]

PR #1405, opened 2017-10-20
Introduce Bpf Program Snapshot (bps)

iamkafai:
This patch set introduces a new binary command/tool to list all loaded BPF programs of a system.  The first two patches are prep work.  The last patch introduces the 'bps' binary and its commit message has some sample output.


yonghong-song:
[buildbot, ok to test]

yonghong-song:
@drzaeus77 could you help check buildbot? Seems not working now?

introspection/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) Facebook, Inc.
```
yonghong-song: Year "2017"?

introspection/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+# Licensed under the Apache License, Version 2.0 (the "License")
+
+include_directories(${CMAKE_SOURCE_DIR}/src/cc)
+include_directories(${CMAKE_SOURCE_DIR}/src/cc/api)
+
+option(INSTALL_INTROSPECTION "Install BPF introspection tools. Those binaries are statically linked and can take plenty of disk space" ON)
```
yonghong-song: It is static linking. However, all llvm related functions are not called and hence not included in binary. So currently, the size is actually fair small.

-bash-4.3$ ls -l /usr/share/bcc/introspection/bps 
-rwxr-xr-x. 1 root root 37872 Oct 21 03:44 /usr/share/bcc/introspection/bps
-bash-4.3$ 

So I suggest to remove the sentence "Those binaries are statically linked and can take plenty of space."

yonghong-song:
[buildbot, ok to test]

palmtenor:
Do you think we should add `deprecated` attribute to the old `bpf_map_create` and `bpf_prog_load` without name? This would encourage people to move to the new helper and provide better information for introspection

iamkafai:
@palmtenor , how about I remove bpf_prog_load_name and bpf_create_map_name? and just add a 'const char *name' argument to the existing bpf_prog_name() and bpf_create_map()?  That we can force people to change.

iamkafai:
Last set of changes:
1. Fix rpmbuild error
2. Fix tests/python/test_trace4.py test case to accommodate the extra retry-call to syscall(__NR_bpf, BPF_PROG_LOAD) when the kernel does not support prog_name
3. After discussing with @yonghong-song and @palmtenor , my understand is that it is acceptable to change the libbpf's ABI in this stage.  The preference is to keep a clean API by having one bpf_prog_load() instead of adding another _name() variants.  This patch made this change by adding an extra 'const char *name' argument to bpf_prog_load() and bpf_create_map().
4. Due to (3), the lua side has to change also.  I have tested ./bcc-probe ../../examples/lua/task_switch.lua

yonghong-song:
[buildbot, ok to test]

PR #1414, opened 2017-10-26
Add usdt support for ARM64

yonghong-song:
Specifically the following argument patterns will be
supported:
```
   [-]<size>@<value>
   [-]<size>@<reg>
   [-]<size>@[<reg>]
   [-]<size>@[<reg>,<offset>]
```

I did not use std regex library as the old gcc (e.g. 4.8.5)
does not have a good support for it.

### DIFF
--- a/tests/cc/test_usdt_args.cc
+++ b/tests/cc/test_usdt_args.cc
@@ -54,7 +54,9 @@ static void verify_register(USDT::ArgumentParser &parser, int arg_size,
 
 TEST_CASE("test usdt argument parsing", "[usdt]") {
   SECTION("parse failure") {
-#ifdef __powerpc64__
+#ifdef __aarch64__
+    USDT::ArgumentParser_aarch64 parser("4@[x32,200]");
+#elif __powerpc64__
     USDT::ArgumentParser_powerpc64 parser("4@-12(42)");
 #elif defined(__x86_64__)
     USDT::ArgumentParser_x64 parser("4@i%ra+1r");
@@ -69,7 +71,13 @@ TEST_CASE("test usdt argument parsing", "[usdt]") {
     REQUIRE(i < 10);
   }
   SECTION("argument examples from the Python implementation") {
-#ifdef __powerpc64__
+#ifdef __aarch64__
+    USDT::ArgumentParser_aarch64 parser("-1@x0 4@5 8@[x12] -4@[x31,-40]");
+    verify_register(parser, -1, "regs[0]");
+    verify_register(parser, 4, 5);
+    verify_register(parser, 8, "regs[12]", 0);
+    verify_register(parser, -4, "regs[31]", -40);
+#elif __powerpc64__
     USDT::ArgumentParser_powerpc64 parser(
         "-4@0 8@%r0 8@i0 4@0(%r0) -2@0(0) "
         "1@0 -2@%r3 -8@i9 -1@0(%r4) -4@16(6) "
--- a/src/cc/usdt.h
+++ b/src/cc/usdt.h
@@ -74,24 +74,49 @@ public:
   const optional<int> deref_offset() const { return deref_offset_; }
 
   friend class ArgumentParser;
+  friend class ArgumentParser_aarch64;
   friend class ArgumentParser_powerpc64;
   friend class ArgumentParser_x64;
 };
 
 class ArgumentParser {
-protected:
+ protected:
   const char *arg_;
   ssize_t cur_pos_;
 
   void skip_whitespace_from(size_t pos);
   void skip_until_whitespace_from(size_t pos);
   void print_error(ssize_t pos);
+  ssize_t parse_number(ssize_t pos, optional<int> *result) {
+    char *endp;
+    int number = strtol(arg_ + pos, &endp, 0);
+    if (endp > arg_ + pos)
+      *result = number;
+    return endp - arg_;
+  }
+  bool error_return(ssize_t error_start, ssize_t skip_start) {
+    print_error(error_start);
+    skip_until_whitespace_from(skip_start);
+    return false;
+  }
 
-public:
+ public:
   virtual bool parse(Argument *dest) = 0;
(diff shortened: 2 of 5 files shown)


yonghong-song:
@palmtenor

tests/cc/test_usdt_args.cc
```diff
@@ -113,7 +121,7 @@ TEST_CASE("test usdt argument parsing", "[usdt]") {
     verify_register(parser, 2, 1097);
     verify_register(parser, 4, "gpr[30]", 108);
     verify_register(parser, -2, "gpr[31]", -4);
-#elif defined(__x86_64__)
```
4ast: why remove defined(__x86_64__) ?
Other than this nit it all looks good to me.
cc @goldshtn

yonghong-song: This is for consistency. In all other places, we have x86_64 as the default. So I changed a couple of places here to have x86_64 as the default. 
 
The x86_64 as the default in the core source files makes build easier for not-fully-supported architectures. Although some tests/tools may not work as expected, at least build can go through without change.

tests/cc/test_usdt_args.cc
```diff
@@ -54,9 +54,11 @@ static void verify_register(USDT::ArgumentParser &parser, int arg_size,
 
 TEST_CASE("test usdt argument parsing", "[usdt]") {
   SECTION("parse failure") {
-#ifdef __powerpc64__
+#ifdef __aarch64__
```
palmtenor: Theoritically these tests doesn't rely on architecture. We just put in fake argument format strings and see if the parser outputs correct results. It would be nice to always test all architectures all the time, right?

yonghong-song: Okay. I will add the arch guard back to the test.

yonghong-song:
updated the patch to add "#elif __x86_64__" back to test_usdt_args.c.

tests/cc/test_usdt_args.cc
yonghong-song: > Theoritically these tests doesn't rely on architecture. We just put in fake argument format strings, call specific parser, and see if it outputs correct results. It would be nice to always test all architectures all the time, right?

The test _input_ needs to be different for different architecture. That is why we have different "ifdef"s.

yonghong-song:
Just a little more information. The following is the output of
`readelf -n /lib64/libc.so.6`
```
......
  stapsdt              0x00000048       NT_STAPSDT (SystemTap probe descriptors)
    Provider: libc
    Name: memory_mallopt_arena_max
    Location: 0x000000000007e720, Base: 0x00000000001366b8, Semaphore: 0x0000000000000000
    Arguments: -4@x20 8@[x22,32]
  stapsdt              0x00000054       NT_STAPSDT (SystemTap probe descriptors)
    Provider: libc
    Name: memory_mallopt_mmap_max
    Location: 0x000000000007e740, Base: 0x00000000001366b8, Semaphore: 0x0000000000000000
    Arguments: -4@x20 -4@[x22,44] -4@[x22,52]
  stapsdt              0x00000037       NT_STAPSDT (SystemTap probe descriptors)
    Provider: libc
    Name: memory_malloc_retry
    Location: 0x000000000007ec18, Base: 0x00000000001366b8, Semaphore: 0x0000000000000000
    Arguments: 8@x20
  stapsdt              0x0000003e       NT_STAPSDT (SystemTap probe descriptors)
    Provider: libc
    Name: memory_realloc_retry
    Location: 0x000000000007efa0, Base: 0x00000000001366b8, Semaphore: 0x0000000000000000
    Arguments: 8@x23 8@x19
  stapsdt              0x0000003f       NT_STAPSDT (SystemTap probe descriptors)
    Provider: libc
    Name: memory_memalign_retry
    Location: 0x000000000007f120, Base: 0x00000000001366b8, Semaphore: 0x0000000000000000
    Arguments: 8@x21 8@x20
  stapsdt              0x00000037       NT_STAPSDT (SystemTap probe descriptors)
    Provider: libc
    Name: memory_valloc_retry
    Location: 0x000000000007f7cc, Base: 0x00000000001366b8, Semaphore: 0x0000000000000000
    Arguments: 8@x20
  stapsdt              0x00000038       NT_STAPSDT (SystemTap probe descriptors)
    Provider: libc
    Name: memory_pvalloc_retry
    Location: 0x000000000007f9d8, Base: 0x00000000001366b8, Semaphore: 0x0000000000000000
    Arguments: 8@x21
......
```

tests/cc/test_usdt_args.cc
palmtenor: It feels to me that, say, the code `ArgumentParser_aarch64` itself, can also run on x86, is that right?

In that case, we can always invoke `ArgumentParser_aarch64` on arm64 test inputs, `USDT::ArgumentParser_powerpc64` on Power PC inputs, etc.?

Otherwise, since our testbot only runs on x86, we won't be able to catch regressions and breakage for other architectures at PR time?

yonghong-song: I still prefer that it is guarded by different architecture. The reason is usdt parsing code is arch specific. Even if we run some other arch inputs on x86, this still won't catch regression
on powerpc since powerpc parsing code is not exercised on x86.

There is a chance that illegal input in one arch is legal in a different one.
For example 4@4 is illegal on x86 but it is legal in aarch64. Mixing together
will prevent put 4@4 in negative test since it is legal in aarch64.

yonghong-song: After the second thought, I know what you mean. Yes, we could exercise non x86 parsing code on x86 host. I will make a revision to reflect that.

yonghong-song: Sorry, I cannot make the change because powerpc64 uses the regex library which old compiler likes gcc 4.8.5 does not have a good support (as I mentioned in my commit message as well).

....
 REQUIRE( !parser.parse(&arg) )
due to unexpected exception with message:
  regex_error
....

Therefore, let us stick to arch. specific usdt parser tests for now. Supposedly, each arch parser change should have run this test on their respective arch to catch all possible errors.

PR #1418, opened 2017-10-29
Consistently use spaces in C in Python tutorial

zoidbergwill:
Sorry, nitpick, but it was confusing me a bit while going through it.

### DIFF
--- a/docs/tutorial_bcc_python_developer.md
+++ b/docs/tutorial_bcc_python_developer.md
@@ -67,8 +67,8 @@ from bcc import BPF
 # define BPF program
 prog = """
 int hello(void *ctx) {
-    bpf_trace_printk("Hello, World!\\n");
-    return 0;
+	bpf_trace_printk("Hello, World!\\n");
+	return 0;
 }
 """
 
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, ok to test]

brendangregg:
Much of the C that we are tracing originates from the Linux kernel, which uses tabs as indentation. Switching that C code to look like Python, with 4 character indents, I do not think looks right to a kernel engineer.

The bcc repository has examples of both: some use tabs in their kernel code, and some uses 4 character indents. I'm fine with having examples of both, to appeal to different audiences. If I had to pick one, I'd switch the C code to tabs to be consistent with the kernel code that it is tracing, even though the remainder of the tool will be Python with 4 character indents. But I don't want to insist on one convention for all authors.

I'd accept this change as it makes one file consistent -- the tutorial. But I would not like to see all the other tools in bcc changed. If a kernel engineer uses tabs in their C code, we should let them. If a python developer writes a bcc tool and wants 4 character indents everywhere, we should let them too.

zoidbergwill:
Sounds good to me. That makes sense. Will just change it to be standardised as kernel style.

I think following the kernel's style guide then is probably the best solution. Some of the snippets just had both spaces and tabs in a single string of C which got me started down the rabbit hole.

It might confuse people's IDEs if we have tabs and spaces together in Python, but the reasoning is sane and I'm happy with whichever cuz I'm a noob to C, as long as it's consistent within snippets, and ideally a file, but the mixing of C and Python in a single file is confusing and only for simple cases a lot of the time anyway.

docs/tutorial_bcc_python_developer.md
```diff
@@ -598,30 +598,30 @@ This program instruments a user-defined static tracing (USDT) probe, which is th
 
 ```
 # ./nodejs_http_server.py 24728
-TIME(s)            COMM             PID    ARGS
-24653324.561322998 node             24728  path:/index.html
-24653335.343401998 node             24728  path:/images/welcome.png
-24653340.510164998 node             24728  path:/images/favicon.png
+TIME(s)			COMM			 PID	ARGS
```
palmtenor: This is what the tool actually outputs, we should change this, right?

zoidbergwill: Whoops, I didin't mean to edit any tool outputs.

palmtenor:
LGTM. Could you collapse the commits to one?

zoidbergwill:
Awesome, thanks. Done

yonghong-song:
@4ast could you take a look?

PR #1419, opened 2017-10-31
Handle unknown user in cachetop

runejuhl:
If running on a system that uses cgroups the process owner uid may not exist on
the host system and `pwd.getpwuid` throws a KeyError.

This patch defaults to a uid of `UNKNOWN(id)` when we're unable to look up the
username.

...and I just noticed that columns are truncated to 8 characters meaning that the `id` won't be shown -- should I leave it like it is and wait for variable column width (if that's on the drawing board?), or would you prefer another format?

### DIFF
--- a/tools/cachetop.py
+++ b/tools/cachetop.py
@@ -222,11 +222,20 @@ def handle_loop(stdscr, args):
         )
         (height, width) = stdscr.getmaxyx()
         for i, stat in enumerate(process_stats):
+            uid = int(stat[1])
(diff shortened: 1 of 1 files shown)


goldshtn:
[buildbot, ok to test]

goldshtn:
I think it would make sense to make the id column a bit wider in this case.

brendangregg:
LGTM, thanks. We can make that column wider later.

Issue #1420, opened 2017-10-31
In funcslower.py and other similar tools, functions called from same thread are not traced correctly


ggaurav10:
In tools like:
https://github.com/iovisor/bcc/blob/master/tools/funcslower.py

If the function list provided has functions which will be called by the same thread such as:
`#./funcslower.py vfs_read __vfs_read`

then the latencies calculated will be wrong, because the tool is only filtering based on the tid/pid and not on the function name/id.
I made below changes to correct this. Basically after hash lookup for tid/pid in trace_return, proceed only if the function name/id also matches. 

```diff --git a/tools/funcslower.py b/tools/funcslower.py
index b8ef230..cc74a4b 100755
--- a/tools/funcslower.py
+++ b/tools/funcslower.py
@@ -115,13 +115,13 @@ static int trace_entry(struct pt_regs *ctx, int id)
     return 0;
 }
 
-int trace_return(struct pt_regs *ctx)
+int trace_return(struct pt_regs *ctx, int id)
 {
     struct entry_t *entryp;
     u64 tgid_pid = bpf_get_current_pid_tgid();
 
     entryp = entryinfo.lookup(&tgid_pid);
-    if (entryp == 0) {
+    if (entryp == 0 || entryp->id != id) {
         return 0;
     }
 
@@ -160,7 +160,10 @@ for i in range(len(args.functions)):
 int trace_%d(struct pt_regs *ctx) {
     return trace_entry(ctx, %d);
 }
-""" % (i, i)
+int rtrace_%d(struct pt_regs *ctx) {
+    return trace_return(ctx, %d);
+}
+""" % (i, i, i, i)
 
 if args.verbose:
     print(bpf_text)
@@ -171,10 +174,10 @@ for i, function in enumerate(args.functions):
     if ":" in function:
         library, func = function.split(":")
         b.attach_uprobe(name=library, sym=func, fn_name="trace_%d" % i)
-        b.attach_uretprobe(name=library, sym=func, fn_name="trace_return")
+        b.attach_uretprobe(name=library, sym=func, fn_name="rtrace_%d" % i)
     else:
         b.attach_kprobe(event=function, fn_name="trace_%d" % i)
-        b.attach_kretprobe(event=function, fn_name="trace_return")
+        b.attach_kretprobe(event=function, fn_name="rtrace_%d" % i)
 
 TASK_COMM_LEN = 16  # linux/sched.h
 
```

However, I am getting below error when I run: 

```
[...]

PR #1422, opened 2017-10-31
Functions called from same thread are not traced correctly

ggaurav10:
The current implementation uses only tgid_pid for hashing. When 2 functions called by a same thread are given as input (eg. vfs_read and __vfs_read), then the tool calculates the difference from when the thread entered the first function to when it exited second function.
This fix uses the function name (or id) along with the tgid_pid as key to hash the structure that has time details.
Fix for issue: #1420

### DIFF
--- a/tools/funcslower.py
+++ b/tools/funcslower.py
@@ -67,11 +67,15 @@ bpf_text = """
 #include <linux/sched.h>    // for TASK_COMM_LEN
 
 struct entry_t {
-    u64 id;
     u64 start_ns;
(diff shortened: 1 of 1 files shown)


brendangregg:
My only hesitation is the additional performance overhead: what is it, say, for a frequent function? It might mean adjusting the man page overhead section. If the overhead becomes prohibitive (I don't really see why in this case), then the fix should be changing the man page to document this tool's caveats.

goldshtn:
[buildbot, ok to test]

PR #1424, opened 2017-10-31
Fix minor compilation issue in bps

palmtenor:
The re-declaration of `info_len` triggers `-Werror=shadow-compatible-local` on some compilers.


palmtenor:
cc @iamkafai

PR #1425, opened 2017-11-01
Fix Module display for unreadable Modules

palmtenor:
Currently when BCC symbolizing addresses, if it could not read the file backing up a memory module (due to deleted or any other reason), it won't add the module to `ProcSyms`. This might improve efficiency a little bit. However, due to this, we won't be able to report the module name correctly for addresses belong to these modules, which is often very useful when debugging why the stack trace contains `[unknown]`. This PR fixes the issue by always add module even if it's not readable.

Also, BCC would always add a module for the `perf-PID.map` with address range `(0, INT_MAX)`. This causes that we would always fall back to the perf map module in the end and not report the module name correctly. This PR also fixes the issue by correctly handle `perf-PID.map` and the saved module name.

I locally hacked `profile.py` to always display the module for testing.
Before:
```
    folly::FunctionScheduler::runOneFunction(std::unique_lock<std::mutex>&, std::chrono::time_point<std::chrono::_V2::steady_clock, std::chrono::duration<long, std::ratio<1l, 1000000000l> > >) [test_bin]
    folly::FunctionScheduler::run() [test_bin]
    [unknown]
    [unknown]
    [unknown]
    -                servicedata-pub (2003)
        1
```
After:
```
    folly::EventBase::loopForever() [test_bin]
    folly::IOThreadPoolExecutor::threadRun(std::shared_ptr<folly::ThreadPoolExecutor::Thread>) [test_bin]
    void folly::detail::function::FunctionTraits<void ()>::callBig<std::_Bind<std::_Mem_fn<void (folly::ThreadPoolExecutor::*)(std::shared_ptr<folly::ThreadPoolExecutor::Thread>)> (folly::ThreadPoolExecutor*, std::shared_ptr<folly::ThreadPoolExecutor::Thread>)> >(folly::detail::function::Data&) [test_bin]
    [unknown] [libstdc++.so.6.0.21 (deleted)]
    [unknown] [libpthread-2.23.so (deleted)]
    [unknown] [libc-2.23.so (deleted)]
    -                IOThreadPool7 (2003)
        1
```

### DIFF
--- a/src/cc/bcc_perf_map.c
+++ b/src/cc/bcc_perf_map.c
@@ -25,9 +25,11 @@
 bool bcc_is_perf_map(const char *path) {
   char* pos = strstr(path, ".map");
   // Path ends with ".map"
-  if (pos == NULL || *(pos + 4) != 0)
-    return false;
-  return access(path, R_OK) == 0;
+  return (pos != NULL) && (*(pos + 4)== 0);
+}
+
+bool bcc_is_valid_perf_map(const char *path) {
+  return bcc_is_perf_map(path) && (access(path, R_OK) == 0);
 }
 
 int bcc_perf_map_nstgid(int pid) {
--- a/src/cc/bcc_perf_map.h
+++ b/src/cc/bcc_perf_map.h
@@ -28,6 +28,7 @@ extern "C" {
 typedef int (*bcc_perf_map_symcb)(const char *, uint64_t, uint64_t, void *);
 
 bool bcc_is_perf_map(const char *path);
+bool bcc_is_valid_perf_map(const char *path);
 
 int bcc_perf_map_nstgid(int pid);
 bool bcc_perf_map_path(char *map_path, size_t map_len, int pid);
--- a/src/cc/bcc_syms.cc
+++ b/src/cc/bcc_syms.cc
@@ -125,8 +125,6 @@ void ProcSyms::load_exe() {
   std::string exe = ebpf::get_pid_exe(pid_);
   Module module(exe.c_str(), mount_ns_instance_.get(), &symbol_option_);
 
-  if (!module.init())
-    return;
   if (module.type_ != ModuleType::EXEC)
     return;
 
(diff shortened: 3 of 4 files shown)


palmtenor:
cc @goldshtn

yonghong-song:
When:
  (1). the symbol address falls into one of deleted module, and
  (2). the /tmp/perf-<pid>.map can resolve this address, say to saymbol "A".
the the display
  A  "deleted_module"

This could be confusing and may not correct since it is possible that the address may resolve into
a different symbol if the module is not deleted?

Should we already use "deleted_module" since typically /tmp/perf-<pid>.map is not
for a file-based module?

palmtenor:
> This could be confusing and may not correct since it is possible that the address may resolve into
a different symbol if the module is not deleted?

This concern makes sense. How about we do this:
- If we successfully resolved the symbol from `perf-PID.map`, then we honestly return the name of `perf-PID.map` and module name in the result.
- If we can't find the symbol anywhere, but was able to map the address to a file-based module (most likely deleted or something else), then we return the name of the module instead of the `perf-PID.map` which is just a fallback and not actually have the symbol resolve?

> Should we already use "deleted_module" since typically /tmp/perf-.map is not
for a file-based module?

No, currently `perf-PID.map` is a totally normal file-based module:(

yonghong-song:
> 
>    If we successfully resolved the symbol from perf-PID.map, then we honestly return the name of perf-PID.map and module name in the result.
 >   If we can't find the symbol anywhere, but was able to map the address to a file-based module (most likely deleted or something else), then we return the name of the module instead of the perf-PID.map which is just a fallback and not actually have the symbol resolve?
>

This sounds good to me.

palmtenor:
Update according to @yonghong-song 's comment.
Example of stack trace:
```
    PHP::Asio::enterAsyncEntryPoint [perf-2698931.map]
    PHP::__pseudoMain [perf-2698931.map]
    HHVM::callToExit [perf-2698931.map]
    HPHP::jit::enterTCImpl(unsigned char*, HPHP::ActRec*) [hhvm]
    HPHP::AsyncFuncImpl::ThreadFunc(void*) [hhvm]
    HPHP::start_routine_wrapper(void*) [hhvm]
    [unknown] [libpthread-2.23.so (deleted)]
    [unknown] [libc-2.23.so (deleted)]
```
We can see that:
- File-backed symbol correctly resolved with module name.
- Symbols found in perf map has name of the perf map as module name.
- Symbols failed to resolve anywhere has original module's name as module name.

PR #1436, opened 2017-11-08
Add flag to enable verifier log_level 2

palmtenor:
Current `DEBUG_BPF` flag only enables register state information on branch. Add an option to enable verifier `log_level` 2 that would have register state on every instruction, which should be helpful for some debugging.

On top of #1433


palmtenor:
Rebased

PR #1445, opened 2017-11-17
Fix 'test_tools_smoke' from failing due to timeout

sandip4n:
This fixes `test_tools_smoke` from failing if `ucalls` and `vfscount` need more time to terminate than the specified timeout duration when running on certain architectures such as `powerpc`. To avoid this, we increase the timeout duration.


yonghong-song:
[buildbot, ok to test]

PR #1447, opened 2017-11-18
solisten.py maintenance

yadutaf:
When using a program like:

```python
import socket
sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
sock.bind(('127.0.1.1', 4242))
sock.listen(42)
sock.close()
```

``solisten.py`` displays:

```
PID    COMM         PROTO  BACKLOG  PORT  ADDR                                   
2346   python       TCPv4  42       4242  1.1.0.127
```

Additionally, clang complains about enum casts in ``include/net/sock.h``:

```
include/linux/memcontrol.h:580:31: warning: implicit conversion from enumeration type 'enum node_stat_item' to different enumeration type 'enum memcg_stat_item' [-Wenum-conversion]
        __mod_memcg_state(pn->memcg, idx, val);
        ~~~~~~~~~~~~~~~~~            ^~~
```

Both are fixed in this PR.

### DIFF
--- a/tools/solisten.py
+++ b/tools/solisten.py
@@ -108,7 +108,6 @@ int kprobe__inet_listen(struct pt_regs *ctx, struct socket *sock, int backlog)
         // Get IP
         if (family == AF_INET) {
             evt.laddr[0] = inet->inet_rcv_saddr;
-            evt.laddr[0] = be32_to_cpu(evt.laddr[0]);
         } else if (family == AF_INET6) {
             bpf_probe_read(evt.laddr, sizeof(evt.laddr),
                            sk->__sk_common.skc_v6_rcv_saddr.in6_u.u6_addr32);
--- a/tools/solisten.py
+++ b/tools/solisten.py
@@ -46,10 +46,12 @@ parser.add_argument("-n", "--netns", default=0, type=int,
 
 # BPF Program
 bpf_text = """
-#include <net/sock.h>
-#include <net/inet_sock.h>
 #include <net/net_namespace.h>
 #include <bcc/proto.h>
+#pragma clang diagnostic push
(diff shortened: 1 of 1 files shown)


goldshtn:
[buildbot, ok to test]

yadutaf:
I believe this is related to this commit: https://github.com/iovisor/bcc/commit/83320685d39e54d5af95001d948f6cdba6eb2482#diff-9d1b58294603e86e349c97249d99c3e0. This commit moved to ``inet_ntop`` in the Python side to decode the addresses. As I understand it, it removed to need to call ``be64_to_cpu`` for both IPv4 and IPv6 but only IPv6 was patched in this commit.

yonghong-song:
Tried latest net-next and the warning is gone. Not sure whether it is worthwhile or not to fix the warning issue as (1). the warning is legitimate although a little annoying, and if people dig into it, it can be properly fixed upstream, and (2). the warning will be gone in the newer linux versions.

yadutaf:
Regarding (1), as a user, it seems a bit odd to me to get warnings related to kernel headers when using a tool to monitor listening sockets as this is something we can not really act on. Regarding (2): Thanks :)

A better approach that would allow to keep relevant warnings like this one without polluting user's output, could be to pre-compile the

goldshtn:
I agree that as a user, I wouldn't want the kernel warning to appear at all, on any version. I'm OK with the fix.

yonghong-song:
Maybe I'm a little bit selfish to use bcc to discover kernel warnings -:). We can add this workaround for now. In the future, when major distro's all have later kernels, we should remove it. I am aware that we have a similar workaround in tcptracer.py.

yadutaf:
Thanks!

PR #1448, opened 2017-11-20
Update perf event type and config checks

palmtenor:
After https://github.com/torvalds/linux/commit/f91840a32deef5cb1bf73338bc5010f843b01426, BPF support all perf events including `HW_CACHE`, `RAW` and dynamic PMU events etc. This commit updates the check in BCC that only allows `HARDWARE` and `SOFTWARE` events. Also updates error message to be more informative from user feedback.

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -20,6 +20,7 @@
 #include <arpa/inet.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <inttypes.h>
 #include <limits.h>
 #include <linux/bpf.h>
 #include <linux/bpf_common.h>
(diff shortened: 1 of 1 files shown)


src/cc/libbpf.c
```diff
@@ -846,27 +847,48 @@ void * bpf_open_perf_buffer(perf_reader_raw_cb raw_cb,
...
+      goto is_invalid;
+    } else if (config == 10 /* PERF_COUNT_SW_BPF_OUTPUT */) {
+      fprintf(stderr, "Unable to open or attach perf event for BPF_OUTPUT\n");
+      goto is_invalid;
+    }
+    return 0;
+  case PERF_TYPE_HW_CACHE:
+    if (((config >> 16) >= PERF_COUNT_HW_CACHE_RESULT_MAX) ||
+        (((config >> 8) & 255) >= PERF_COUNT_HW_CACHE_OP_MAX) ||
+        ((config & 255) >= PERF_COUNT_HW_CACHE_MAX)) {
```
yonghong-song: It is better to replace 255 with 0xff, since it is a bit operation...

palmtenor:
Rebase, address comment

PR #1450, opened 2017-11-21
Fix 'test_libbcc' from failing in certain scenarios

sandip4n:
This fixes `test_libbcc` from failing because of the following scenarios:

1. For `test_c_api`, if the address of `strtok` as obtained from `dlsym` is passed on to `bcc_symcache_resolve`, then the name of the symbol returned by this function may not always be `strtok` but something else like `__GI_strtok` even though it is still a valid alias. This causes the test to fail.

2. For `test_usdt_probes`, the test fails if the static markers are present in `libruby` rather than the `ruby` executable. Also, the number of locations at which the probe `array__create` is present can vary across `ruby` versions. In the code, it is hardcoded to `7` but for recent versions such as `ruby-2.4`, this is now `2`. This can also cause the test to fail.

Both these problems are fixed by verifying values with the output from `nm` and `readelf` respectively. Since external tools are executed, an utility to capture output from an external command has been added. This is placed in a newly introduced utilities file for `cc` tests.

### DIFF
--- a/tests/cc/CMakeLists.txt
+++ b/tests/cc/CMakeLists.txt
@@ -17,7 +17,8 @@ add_executable(test_libbcc
 	test_bpf_table.cc
 	test_hash_table.cc
 	test_usdt_args.cc
-	test_usdt_probes.cc)
+	test_usdt_probes.cc
+	utils.cc)
 
 target_link_libraries(test_libbcc bcc-shared dl)
 add_test(NAME test_libbcc COMMAND ${TEST_WRAPPER} c_test_all sudo ${CMAKE_CURRENT_BINARY_DIR}/test_libbcc)
--- a/tests/cc/test_c_api.cc
+++ b/tests/cc/test_c_api.cc
@@ -193,6 +193,8 @@ static int mntns_func(void *arg) {
   return 0;
 }
 
+extern int cmd_scanf(const char *cmd, const char *fmt, ...);
+
 TEST_CASE("resolve symbol addresses for a given PID", "[c_api]") {
   struct bcc_symbol sym;
   void *resolver = bcc_symcache_new(getpid(), nullptr);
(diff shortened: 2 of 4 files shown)


palmtenor:
Thanks a lot for taking care of the tests! A couple thoughts:
- For the symbol resolution test, the handling of function alias feels like an overkill... Could we just pick a non-aliased function? For example `atoi`......
- For the USDT test, the `probe_num_locations` and `probe_num_arguments` helpers are another and non-trivial way that we introduced to parse the ELF header, and we are testing our original implementation with this another implementation. Feels a bit like "circular testing" :P

What do you think?

sandip4n:
@palmtenor Thanks for your inputs. However, I have a few concerns.

- For the symbol resolution test, if we lookup `atoi` instead of `strtok`, the test will pass for now but would it be guaranteed `atoi` does not end up having such aliases in the future?
- For the USDT probe test, the only other alternative that I can think of is checking if the number of locations is greater than zero. Do you have any other alternatives in mind?

yonghong-song:
For USDT probe test, I think it is okay to use external tool to help verify the result.

For symbol resolution test, I am thinking whether we need to solve the aliasing issue in general, e.g., user could provide a function "alias_f1" which is aliased to "original_f1" for trace.py. Later on, when stack frame pointers back from kernel, we may end up with referencing "original_f1" or any other aliased function in the stack trace. Should we worry about this at all?
Typically, alias is due to weak symbols, which is pretty common in libc.

yonghong-song:
[buildbot, ok to test]

sandip4n:
@yonghong-song I ran a simple program that uses `strtok` with `trace.py -U -K 'u:c:strtok'` on a `powerpc` system and this is what I saw:
```
PID    TID    COMM         FUNC
31157  31157  test         strtok
        -14
        __GI_strtok+0x0 [libc-2.25.so]
        main+0x70 [test]
        generic_start_main.isra.0+0x140 [libc-2.25.so]
        __libc_start_main+0xb8 [libc-2.25.so]
```
So, yes, as you had pointed out, we might see aliases in stack traces.

sandip4n:
@yonghong-song Fixing the symbol resolution issue by filtering out aliases sounds like a proper long-term solution but for now, can we use cross-checking with `nm` output?

yonghong-song:
@sandip4n could you help find a separate issue with alias symbol resolution problem? Thanks!

sandip4n:
@yonghong-song Sure, I can help with that.

PR #1451, opened 2017-11-21
libbpf: Enable the creation of unbound raw socket

oliviertilmans:
By checking for a NULL or empty char string in bpf_open_raw_sock(), we can detect whether
we should try to bind on an interface or not.
The dns_matching example has been updated to illustrate this behavior.


4ast:
[buildbot ok to test]
lgtm

4ast:
[buildbot, ok to test]

PR #1457, opened 2017-11-23
Do not keep Loader instances around

palmtenor:
After we finish loading the BPF program, we don't need to keep the `ClangLoader` instance around. This commit removes it from `BPFModule`.
Also, for the header files included with `ClangLoader`, have it destruct with the class instance instead of keeping around as static map.


palmtenor:
Yeah, this would be very tiny change, just doing some clean up before attempting to make more improvement on memory usage

PR #1458, opened 2017-11-24
python: make _decode_table_types aware of __int128

oliviertilmans:
There are no ctypes types ready for int128. However, recognizing this
type and representing it as an array is better than simply erroring out.

Triggering such crash can easily be reproduced in current tools if one
tries to access the shared maps from python, for example, accessing 
`b["tuplepid_ipv6"]` in [tcptracer](https://github.com/iovisor/bcc/blob/master/tools/tcptracer.py) right before the final
loop crashes with:
```
Tracing TCP established connections. Ctrl-C to end.
T  PID    COMM             IP SADDR            DADDR            SPORT  DPORT
  File "/usr/share/bcc/tools/tcptracer", line 666, in <module>
    b["tuplepid_ipv6"]
  File "/usr/lib/python3.6/site-packages/bcc/__init__.py", line 437, in __getitem__
    self.tables[key] = self.get_table(key)
  File "/usr/lib/python3.6/site-packages/bcc/__init__.py", line 427, in get_table
    keytype = BPF._decode_table_type(json.loads(key_desc.decode()))
  File "/usr/lib/python3.6/site-packages/bcc/__init__.py", line 391, in _decode_table_type
    fields.append((t[0], BPF._decode_table_type(t[1])))
  File "/usr/lib/python3.6/site-packages/bcc/__init__.py", line 386, in _decode_table_type
    return BPF.str2ctype[desc]
KeyError: 'unsigned __int128'
```

The commit:
* Adds mappings between (unsigned) __int128 C types declaration and ctypes
  parsing
* Adds a simple test case in test_clang to recognize and use such types
* Fixes deprecation warnings about self.assertEquals in 2 other test
  cases.

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -378,7 +378,9 @@ class BPF(object):
         u"unsigned long long": ct.c_ulonglong,
         u"float": ct.c_float,
         u"double": ct.c_double,
-        u"long double": ct.c_longdouble
+        u"long double": ct.c_longdouble,
+        u"__int128": ct.c_int64 * 2,
+        u"unsigned __int128": ct.c_uint64 * 2,
     }
     @staticmethod
     def _decode_table_type(desc):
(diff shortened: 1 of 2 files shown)


yonghong-song:
[buildbot, ok to test]

PR #1464, opened 2017-11-28
python: Avoid unnecessary pointer object creations

oliviertilmans:
Use ctypes.byref() instead of ctypes.pointer() in cases where we do not
need the actual pointer object itself as it is much lighter.

Quoting the [ctypes documentation §15.17.1.9](https://docs.python.org/2/library/ctypes.html#passing-pointers-or-passing-parameters-by-reference):
> ctypes exports the byref() function which is used to pass parameters by reference. The same effect can be achieved with the pointer() function, although pointer() does a lot more work since it constructs a real pointer object, so it is faster to use byref() if you don’t need the pointer object in Python itself:

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -68,11 +68,11 @@ class SymbolCache(object):
         address as the offset.
         """
         sym = bcc_symbol()
-        psym = ct.pointer(sym)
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, ok to test]

PR #1468, opened 2017-12-01
cross-compiling: link against target's codegen library

SJLC:
Trying to link in host's codegen lib instead of target's codegen lib
causes failure during linking stage when cross-compiling:

  aarch64-poky-linux/5.3.0/ld: cannot find -lLLVMX86CodeGen

Signed-off-by: S. Lockwood-Childs <sjl@vctlabs.com>


drzaeus77:
[buildbot, ok to test]

Issue #606, opened 2016-07-11
BCC cannot handle alignment in struct used in HashMap


Eichhoernchen:
Hi,

I had the following struct as an item in a hash map:

```
struct Leaf {
  uint16_t len;
  unsigned char p[32];
};
```

I needed to copy the contents of p to the stack and I though uint64_t\* copies would be faster and less instructions so I did:

```
char dst[32];
((uint64_t*)dst)[0] = ((uint64_t*)leaf->p)[0];
((uint64_t*)dst)[1] = ((uint64_t*)leaf->p)[1];
((uint64_t*)dst)[2] = ((uint64_t*)leaf->p)[2];
((uint64_t*)dst)[3] = ((uint64_t*)leaf->p)[3];
```

However, this caused the compiler to restore a register like this `r1 = (uint64_t*)(r5 + 2)` (I don't remember the actual error line) which of cause caused a misalignment error. I came to the conclusion ( even though I think the compiler should not have done that uint64_t\* restoring) that this does not work as the struct is misaligned... So I changed the struct to:

```
struct Leaf {
  uint16_t len;
    unsigned char p[32] __attribute__ ((aligned (8)));
};
```

which works as expected... however, now when I fill the map from userspace... the data that I put into p is off by 6 byte, it seems that the userspace is unable to detect the actual alignment of the members in the struct and starts to put data for p right after len. The alignment should cause the compiler to add 6 bytes of padding between len and p.

Of course I can change everything in a way that it works however I find this rather difficult to debug...

Using `__builtin_memcpy` does not produce this behavior, there I can leave the alignment out or I can also switch p and len in the struct to make everything work... 

I was just wondering is this a conceptual limitation or just a bug in BCC?

4ast:
@drzaeus77 I think __attribute__ ((aligned (8))) is not recognized by python side, so it's a bug.
so it's working fine with __builtin_memcpy and no extra align directives, right?
Changing order of p and len makes sense to avoid the hole regardless.
Also memcpy generated code will be more efficient. Same rules as in normal C applies:
llvm will use the largest load/store quantities depending on the alignment it can detect.

Eichhoernchen:
Yep, it works with the __builtin_memcpy without extra alignment directives.
[...]

PR #1474, opened 2017-12-05
provide padded structure for table_{key|leaf}_desc API

yonghong-song:
This patch intends to fix issue #606.

Currently, the key/value type information is passed from
C++ to Python through a JSON interface. The JSON is
constructed by traversing the struct/field's through clang
AST interface. Once Python gets the JSON, it will
reconstruct the C structure through ctype module.

There are two known issues where Python reconstructed
C structure may not be the same as the original C structure:
  . if user explicitly use "__attribute__ ((align))" to alter
    field alignment and such information is not passed to
    Python.
  . the "__int128" type is a u64[2] array in python.
    So in C, __int128 needs to align on 16 bytes boundary, and
    in Python, it aligns with 8 bytes boundary.

To solve this issue, this patch provided the structure
with added padding fields to Python. For example,
  struct {
    char a;
    __int128 b;
  };
Python will receive
  struct {
    char a;
    char __pad_1[15];
    __int128 b;
  };

### DIFF
--- a/src/cc/json_map_decl_visitor.cc
+++ b/src/cc/json_map_decl_visitor.cc
@@ -18,6 +18,7 @@
 #include <string>
 
 #include <clang/AST/ASTContext.h>
+#include <clang/AST/RecordLayout.h>
 #include <clang/AST/RecursiveASTVisitor.h>
 #include "common.h"
 #include "table_desc.h"
@@ -44,12 +45,46 @@ class BMapDeclVisitor : public clang::RecursiveASTVisitor<BMapDeclVisitor> {
   bool VisitEnumDecl(clang::EnumDecl *D);
 
  private:
+  bool shouldSkipPadding(const RecordDecl *D);
+  void genJSONForField(FieldDecl *F);
+
+ private:
   clang::ASTContext &C;
   std::string &result_;
 };
 
 // Encode the struct layout as a json description
 BMapDeclVisitor::BMapDeclVisitor(ASTContext &C, string &result) : C(C), result_(result) {}
+
+bool BMapDeclVisitor::shouldSkipPadding(const RecordDecl *D) {
+  if (D->isUnion() || D->field_empty())
+    return true;
+  for (auto F : D->getDefinition()->fields()) {
+    if (F->isBitField())
+      return true;
+    QualType Ty = F->getType();
+    if (Ty->isIncompleteArrayType())
+      return true;
+  }
+  return false;
+}
+
+void BMapDeclVisitor::genJSONForField(FieldDecl *F) {
+  if (F->isAnonymousStructOrUnion()) {
+    if (const RecordType *R = dyn_cast<RecordType>(F->getType()))
+      TraverseDecl(R->getDecl());
+    result_ += ", ";
+    return;
+  }
+  result_ += "[";
+  TraverseDecl(F);
+  if (const ConstantArrayType *T = dyn_cast<ConstantArrayType>(F->getType()))
+    result_ += ", [" + T->getSize().toString(10, false) + "]";
+  if (F->isBitField())
+    result_ += ", " + to_string(F->getBitWidthValue(C));
+  result_ += "], ";
+}
+
 bool BMapDeclVisitor::VisitFieldDecl(FieldDecl *D) {
   result_ += "\"";
   result_ += D->getName();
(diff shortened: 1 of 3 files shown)


yonghong-song:
cc @oliviertilmans  @goldshtn  @drzaeus77 @4ast

oliviertilmans:
This looks nice! Will try to apply asap on my code.

As you take care of any offset that would occur in the struct, shouldn't you also force _pack_ = 1 in the ctypes Structure instantiation ? i.e., change
```python
src/python/bcc/__init__.py +416
def _decode_table_type(desc):
[...]
        cls = type(str(desc[0]), (base,), dict(_anonymous_=anon,
            _fields_=fields))
```
to
```python
def _decode_table_type(desc):
[...]
        extra_args = dict(_pack_=1) if base == ct.Structure else {}
        cls = type(str(desc[0]), (base,), dict(_anonymous_=anon,
            _fields_=fields, **extra_args))
```
(not so neat-looking as unions don't support _pack_ ...)

yonghong-song:
Thanks @oliviertilmans for reviewing. I did not use _pack_ = 1 in python side mainly because 
I did not handle all cases.

I skipped bitfield and incomplete array type (e.g., int a[]).
bitfield should be rare, and I am not sure how to deal with incomplete array type with padding.
Since I did not handle all cases, forcing _pack_=1 in python side may risk the wrong result
for unsupported structures.

Note that map key/leaf type will reject the structure with "packed" attribute since python interface
is not able to handle it at all.

ctypes suppose to follow C padding convention, so extra padding from C++ should satisfy alignment requirement so even without _pack_, ctypes should not do anything about it. Otherwise, it is either a C++ side bug or a ctypes bug.

Will push another commit to modify test_clang test case a little bit to use aligned attribute.

yonghong-song:
@oliviertilmans does my explanation make sense? Or I still missed something cases?

oliviertilmans:
> Thanks @oliviertilmans for reviewing. I did not use _pack_ = 1 in python
> side mainly because I did not handle all cases.
> 
> I skipped bitfield and incomplete array type (e.g., int a[]).
> bitfield should be rare, and I am not sure how to deal with incomplete
> array type with padding. Since I did not handle all cases, forcing
> _pack_=1 in python side may risk the wrong result for unsupported
> structures.

These sounds like fair limitations, bitfields can be worked around using
unions to refer to their storage unit or the individual fields, and the
incomplete arrays would require to be handled by hand on the python side
anyway (although I guess they should start with the same alignment
constraints than their storage unit?).

> 
> Note that map key/leaf type will reject the structure with "packed"
> attribute since python interface is not able to handle it at all.

I am not sure I understood that point. What part of bcc will reject this
? The clang frontend seems to handle it fine on my end, and the python
part 'works' as long as the fields are aligned without padding (i.e. I
used packed mainly to play around the int128 alignment).

> ctypes suppose to follow C padding convention, so extra padding from C++
> should satisfy alignment requirement so even without _pack_, ctypes should
> not do anything about it. Otherwise, it is either a C++ side bug or a
> ctypes bug.

I agree that as you're inserting explicit padding through chars, ctypes
should/has to interpret their alignment requirements properly hence
_pack_ should not be needed. I did not consider that aspect in the above
suggestion.

> 
> Will push another commit to modify test_clang test case a little bit to use
> aligned attribute.

> @oliviertilmans does my explanation make sense? 
> Or I still missed something cases?


The patch removed my needs for __attribute__((packed)) in the C/BPF side (I 
do not use bitfields or incomplete arrays). So it solves my issues, thanks!

Is there any place listing all the features of the clang frontend ?
E.g., the automatic bpf_probe_read insertion conditions and limitations
(e.g., macro support), the bpf_dext_pkt/bpf_dins_pkt insertions and effect
(which did surprise me when I first encountered unexpected change of 
endianness), or this explicit padding.

yonghong-song:
On Mon, Dec 11, 2017 at 12:40 AM, Olivier Tilmans <notifications@github.com>
wrote:

> > Thanks @oliviertilmans for reviewing. I did not use _pack_ = 1 in python
> > side mainly because I did not handle all cases.
> >
> > I skipped bitfield and incomplete array type (e.g., int a[]).
> > bitfield should be rare, and I am not sure how to deal with incomplete
> > array type with padding. Since I did not handle all cases, forcing
> > _pack_=1 in python side may risk the wrong result for unsupported
> > structures.
>
> These sounds like fair limitations, bitfields can be worked around using
> unions to refer to their storage unit or the individual fields, and the
> incomplete arrays would require to be handled by hand on the python side
> anyway (although I guess they should start with the same alignment
> constraints than their storage unit?).
>

I tried a simple example. This seems the case. My implementation is based on
clang:lib/StaticAnalyzer/Checkers/PaddingChecker.cpp
which doesn't handle incomplete arrays citing "tricky". So I skipped it as
well.


> >
> > Note that map key/leaf type will reject the structure with "packed"
> > attribute since python interface is not able to handle it at all.
>
> I am not sure I understood that point. What part of bcc will reject this
> ? The clang frontend seems to handle it fine on my end, and the python
> part 'works' as long as the fields are aligned without padding (i.e. I
> used packed mainly to play around the int128 alignment).
>

Sorry about this. I think previously I tried an example with packed
attribute and
it failed with clang so I thought it was not working. I could not find my
previous example
any more. Just tried another example and it was working fine.


>
> > ctypes suppose to follow C padding convention, so extra padding from C++
> > should satisfy alignment requirement so even without _pack_, ctypes
> should
> > not do anything about it. Otherwise, it is either a C++ side bug or a
> > ctypes bug.
>
> I agree that as you're inserting explicit padding through chars, ctypes
> should/has to interpret their alignment requirements properly hence
> _pack_ should not be needed. I did not consider that aspect in the above
> suggestion.
>
> >
> > Will push another commit to modify test_clang test case a little bit to
> use
> > aligned attribute.
>
> > @oliviertilmans does my explanation make sense?
> > Or I still missed something cases?
>
>
> The patch removed my needs for __attribute__((packed)) in the C/BPF side
> (I
> do not use bitfields or incomplete arrays). So it solves my issues, thanks!
>
> Is there any place listing all the features of the clang frontend ?
> E.g., the automatic bpf_probe_read insertion conditions and limitations
> (e.g., macro support), the bpf_dext_pkt/bpf_dins_pkt insertions and effect
> (which did surprise me when I first encountered unexpected change of
> endianness), or this explicit padding.
>

Currently we did not have documentation about this. We tried to hide this
from
the users. Maybe at the least we should have documentation in the code.

Issue #1472, opened 2017-12-04
bpf_prog_load prints verifier messages to stderr always


iaguis:
It seems there was a change in v0.5.0's `bpf_prog_load()`. It now prints verifier messages to stderr always.

Doesn't it make more sense to skip writing to stderr if the user provided a `log_buf`?

yonghong-song:
cc @palmtenor

palmtenor:
That is true, apologize for the unexpected change.

Will put up a change to skip printing to stderr if user provided log buffer. Do we want to skip the printed message on error as well?

iaguis:
It seems v0.4.0 did print the log to stderr on failure.

However, as a user of libbcc, I'd rather have control over printing logs than have the library print things to stderr, even when there're errors.

palmtenor:
The `stderr` message we print has some "hints", i.e. interpreting common errors to actionable instructions, which I feel would be useful for beginner users. Probably will go for keeping the original behavior, that is, only skip printing when user provided log buffer in non-error cases.

PR #1475, opened 2017-12-06
Do not print to stderr if user specified log buffer

palmtenor:
Addresses #1472

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -474,10 +474,12 @@ int bpf_prog_load(enum bpf_prog_type prog_type, const char *name,
     }
   }
 
-  // If log_level is not 0, either speficied by user or set due to error,
-  // print the log message.
+  // Check if we should print the log message if log_level is not 0,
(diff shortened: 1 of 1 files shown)


yonghong-song:
cc @iaguis

src/cc/libbpf.c
```diff
@@ -474,10 +474,12 @@ int bpf_prog_load(enum bpf_prog_type prog_type, const char *name,
     }
   }
 
-  // If log_level is not 0, either speficied by user or set due to error,
-  // print the log message.
+  // Check if we should print the log message if log_level is not 0,
+  // either speficied by user or set due to error.
```
iaguis: speficied -> specified

shenghaoyuan:
The latest version doesn't support this feature, am I right? 
I tested `tests/python/test_bpf_log.py`, but no output.
Here is a [temporary solution](https://github.com/shenghaoyuan/bcc/blob/master/src/python/bcc/__init__.py#L503) for printing verifier log to user-defined log_buffer

PR #1480, opened 2017-12-09
histogram: support sorting bucket before dumping table

htbegin:
When there are multiple fields in the secondary key of a histogram
(e.g. output the bio latency for each combination of disk and blk cgroup),
if the buckets are not sorted before the iteration, the output order
of tables often does not meet our expectations.

So provide a way to sort them.

### DIFF
--- a/src/python/bcc/table.py
+++ b/src/python/bcc/table.py
@@ -275,9 +275,11 @@ class TableBase(MutableMapping):
         return next_key
 
     def print_log2_hist(self, val_type="value", section_header="Bucket ptr",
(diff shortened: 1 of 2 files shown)


yonghong-song:
Could you add a test or example to use the new interface?

htbegin:
Hi @yonghong-song,
Could you please check and review the commit 33a5a4b ?
I'm also working on adding blk cgroup support to biolatency tools, when it's ready, I will initiate a new PR.

yonghong-song:
[buildbot, ok to test]

PR #1481, opened 2017-12-09
execsnoop: fix the prototype for kprobe__sys_execve

htbegin:
For sys_execve() the type of filename parameter should be
"const char __user *" instead of "struct filename *".

The fix doesn't imply any functional change.


yonghong-song:
[buildbot, ok to test]

yonghong-song:
LGTM.

brendangregg:
thanks!

PR #1486, opened 2017-12-14
Add C++ helper to clear BPF tables

palmtenor:
In production we often want to clear BPF tables during continuous profiling.

To make the use case easier, this PR adds a wrapped helper that iterates through Hash and Stack tables and clear all values.

This PR also adds some test for C++ API's Hash and Stack table, for both existing and this new feature.

### DIFF
--- a/src/cc/api/BPFTable.cc
+++ b/src/cc/api/BPFTable.cc
@@ -106,6 +106,12 @@ BPFStackTable::~BPFStackTable() {
     bcc_free_symcache(it.second, it.first);
 }
 
+void BPFStackTable::clear_table_non_atomic() {
+  for (int i = 0; i < capacity(); i++) {
+    remove(&i);
+  }
+}
+
 std::vector<uintptr_t> BPFStackTable::get_stack_addr(int stack_id) {
   std::vector<uintptr_t> res;
   stacktrace_t stack;
--- a/tests/cc/test_bpf_table.cc
+++ b/tests/cc/test_bpf_table.cc
@@ -14,6 +14,10 @@
  * limitations under the License.
  */
 
+#include <linux/version.h>
+#include <unistd.h>
+#include <string>
+
 #include "BPF.h"
 #include "catch.hpp"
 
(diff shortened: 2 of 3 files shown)


palmtenor:
Looks like Ubuntu 1604 only has 4.4 Kernel, so gate the stack trace table test. Any chance we can upgrade the Kernel version of Jenkins test machines?

src/cc/api/BPFTable.cc
```diff
@@ -106,6 +106,12 @@ BPFStackTable::~BPFStackTable() {
     bcc_free_symcache(it.second, it.first);
 }
 
+void BPFStackTable::clear_table_non_atomic() {
+  for (int i = 0; i < capacity(); i++) {
+    remove(&i);
+  }
```
yonghong-song: stack table essentially is a hash table with a key of 32-bit int. There are two ways to clear the table. One is the above, and the other is to do similar to hash table, using bpf_get_first_key and bpf_get_next_key API. Do you have particular reason to choose this one?

tests/cc/test_bpf_table.cc
```diff
@@ -62,3 +66,101 @@ TEST_CASE("test bpf table", "[bpf_table]") {
...
   REQUIRE(res.code() != 0);
 }
+
+TEST_CASE("test bpf hash table", "[bpf_hash_table]") {
+  const std::string BPF_PROGRAM = R"(
+    BPF_HASH(myhash, int, int, 128);
+  )";
+
+  ebpf::BPF *bpf(new ebpf::BPF);
+  ebpf::StatusTuple res(0);
```
yonghong-song: I guess you can just do `ebpf::BPF *bpf = new ebpf::BPF()`?

tests/cc/test_bpf_table.cc
```diff
@@ -62,3 +66,101 @@ TEST_CASE("test bpf table", "[bpf_table]") {
...
+  REQUIRE(res.code() == 0);
+  values = t.get_table_offline();
+  REQUIRE(values.size() == 0);
+}
+
+TEST_CASE("test bpf stack table", "[bpf_stack_table]") {
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 6, 0)
+  const std::string BPF_PROGRAM = R"(
+    BPF_HASH(id, int, int, 8);
+    BPF_STACK_TRACE(stack_traces, 8);
```
yonghong-song: Since you already use `zero` as the key below, you can have hash table size 1.

tests/cc/test_bpf_table.cc
```diff
@@ -62,3 +66,101 @@ TEST_CASE("test bpf table", "[bpf_table]") {
...
+    int on_sys_getuid(void *ctx) {
+      int stack_id = stack_traces.get_stackid(ctx, BPF_F_REUSE_STACKID);
+      int zero = 0, *val;
+      val = id.lookup_or_init(&zero, &stack_id);
+      (*val) = stack_id;
+    }
+  )";
+
+  ebpf::BPF *bpf(new ebpf::BPF);
+  ebpf::StatusTuple res(0);
```
yonghong-song: bpf = new ebpf::BPF?

src/cc/api/BPFTable.cc
palmtenor: @yonghong-song I thought so too, but: https://github.com/torvalds/linux/blob/master/kernel/bpf/stackmap.c#L231

palmtenor:
Address comments

src/cc/api/BPFTable.cc
yonghong-song: Okay, then the change looks good. I will check the kernel side to see whether we can implement get_next_key API for stacktrace map.

PR #1489, opened 2017-12-15
ucalls: convert bytes to str

r4f4:
This fixes the following error with python3

Tracing calls in process 2577 (language: python)... Ctrl-C to quit.
^C
  File "/usr/share/bcc/tools/lib/ucalls", line 303, in <module>
    data = get_data()   # [(function, (num calls, latency in ns))]
  File "/usr/share/bcc/tools/lib/ucalls", line 264, in get_data
    bpf["counts"].items()))
  File "/usr/share/bcc/tools/lib/ucalls", line 262, in <lambda>
    data = list(map(lambda kv: (kv[0].clazz + "." + kv[0].method,
TypeError: can't concat str to bytes

Signed-off-by: Rafael Fonseca <r4f4rfs@gmail.com>

### DIFF
--- a/tools/lib/ucalls.py
+++ b/tools/lib/ucalls.py
@@ -255,11 +255,13 @@ if args.syscalls:
 def get_data():
     # Will be empty when no language was specified for tracing
     if args.latency:
-        data = list(map(lambda kv: (kv[0].clazz + "." + kv[0].method,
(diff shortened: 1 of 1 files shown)


yonghong-song:
Another python3 decode() related issue. LGTM, but cc @drzaeus77 to see whether we have a broader strategy to address the issue.

PR #1490, opened 2017-12-16
Clean-up unused parameter in tracing APIs

palmtenor:
BPF tracing events are system-wide and does not support CPU filter. Also among them only uprobes can have PID filter. We are also not using group_fd for BPF programs at all. However, all kprobe, uprobe and Kernel Tracepoint APIs, including C, C++, Python and Lua, all has `cpu`, `pid` and `group_fd` parameters. This causes the APIs to be more complicated and users to be confused because BCC "doesn't" work when they specify `cpu` for uprobes or `pid` for kprobes etc.

This PR removes all the unused parameters from all interfaces.

### DIFF
--- a/src/cc/api/BPF.cc
+++ b/src/cc/api/BPF.cc
@@ -172,7 +172,7 @@ StatusTuple BPF::attach_kprobe(const std::string& kernel_func,
 
   void* res =
       bpf_attach_kprobe(probe_fd, attach_type, probe_event.c_str(), kernel_func.c_str(),
-                        pid, cpu, group_fd, cb, cb_cookie);
+                        cb, cb_cookie);
 
   if (!res) {
     TRY2(unload_func(probe_func));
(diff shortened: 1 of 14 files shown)


4ast:
Mixing removal of arguments with formatting changes made the patch harder to review.
In the future please separate style cleanups from the main changes into separate commits.
Thanks

goldshtn:
Did you confirm that none of our tools were using these options? I’m quite sure this would break some of them.

palmtenor:
@goldshtn Some of them have (like `stackcount` changed in this PR). But these options would have no effect anyways now~

goldshtn:
Sorry, I didn’t notice there were tools included in the list of changes. And yes, they wouldn’t have worked as designed, but at least wouldn’t generate errors if the API parameter was missing. Thanks!

palmtenor:
@goldshtn Yeah, I did some grepping on all tools and examples and updated all the call sites that would have had argument mismatch. If you see any file that I missed please let me know!

PR #1491, opened 2017-12-16
Improve trace.py formatting and printing options

palmtenor:
This PR:
- Improves `trace.py` formatting. PID / TGID could easily be 7-digits long, increase the column width. Process COMM could be 15 characters long, increase the column width.
- Add `-C` / `--print_cpu` option that would optionally print the CPU ID of the event, which is useful when debugging CPU affinity.

Tested with no options, `-t`, `-T`, `-t -C`, `-T -C` and `-C`. Example output:
```
$ trace.py c:strchr -t -C
TIME     CPU PID     TID     COMM            FUNC
0.997841 15  533191  533191  sshd            strchr
0.998253 15  533191  533191  sshd            strchr
1.008683 18  533193  533193  unix_chkpwd     strchr
1.009422 19  533191  533191  sshd            strchr
1.010921 19  533191  533191  sshd            strchr
1.029421 23  533194  533194  (systemd)       strchr
1.092944 17  533196  533196  sh              strchr
```

### DIFF
--- a/man/man8/trace.8
+++ b/man/man8/trace.8
@@ -2,8 +2,8 @@
 .SH NAME
 trace \- Trace a function and print its arguments or return value, optionally evaluating a filter. Uses Linux eBPF/bcc.
 .SH SYNOPSIS
-.B trace [-h] [-p PID] [-L TID] [-v] [-Z STRING_SIZE] [-S]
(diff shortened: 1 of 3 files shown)


palmtenor:
Update an example:)

PR #1493, opened 2017-12-18
xdp_drop_count.py: removed duplicate h_proto check

spacewander:


drzaeus77:
The code is there for parsing double-vlan packets. It is intentionally duplicated.

spacewander:
@drzaeus77 
Got it.

PR #1496, opened 2017-12-20
Fix and improvement for perf_reader's lost_cb

palmtenor:
From `uapi/linux/perf_event.h`, the `PERF_RECORD_LOST` sample format looks like:
```
/*
 * struct {
 *	struct perf_event_header	header;
 *	u64				id;
 *	u64				lost;
 * 	struct sample_id		sample_id;
 * };
 */
```
However, when we add the ability to consume such sample in https://github.com/iovisor/bcc/commit/8207d10e99f0b6bd2aa13d7f44adc4af5c803fe4, we incorrect calculated the offset to read the field `id` where we actually want to read the field `lost` for the count of samples lost. The same bug was inherited in https://github.com/iovisor/bcc/commit/4b764de6bcb3655b29b2fb6df6c9c65b7a31338c.

This PR fixes the issue by using the correct offset.

Also, user feedback shows that the `lost_cb` is not very useful unless it could take a `cb_cookie` as the `raw_cb` does, so that necessary context could be added when the lost event happens. Therefore, this PR also adds `cb_cookie` to `lost_cb`'s parameter. It will share the same `cb_cookie` from the `perf_reader` as `raw_cb` uses. This functionality is only added in #1092 and no tool is using it yet.  

Tested both Python and C++ tools. Example of lost samples by tracing high-frequency events.
Before:
```
$ trace.py t:sched:sched_switch -t -C
......
1.505114 11  0       0       swapper/11      sched_switch
Possibly lost 13576517 samples
1.505150 11  0       0       swapper/11      sched_switch
Possibly lost 13576517 samples
1.505185 11  0       0       swapper/11      sched_switch
1.505233 11  997252  997252  sshd            sched_switch
......
```
After:
```
$ trace.py t:sched:sched_switch -t -C
......
1.494558 11  0       0       swapper/11      sched_switch
Possibly lost 1 samples
1.494607 11  997252  997252  sshd            sched_switch
Possibly lost 2 samples
1.494650 11  0       0       swapper/11      sched_switch
Possibly lost 1 samples
......
```

### DIFF
--- a/src/cc/libbpf.h
+++ b/src/cc/libbpf.h
@@ -68,7 +68,7 @@ int bpf_open_raw_sock(const char *name);
 typedef void (*perf_reader_cb)(void *cb_cookie, int pid, uint64_t callchain_num,
                                void *callchain);
 typedef void (*perf_reader_raw_cb)(void *cb_cookie, void *raw, int raw_size);
-typedef void (*perf_reader_lost_cb)(uint64_t lost);
+typedef void (*perf_reader_lost_cb)(void *cb_cookie, uint64_t lost);
 
 void *bpf_attach_kprobe(int progfd, enum bpf_probe_attach_type attach_type,
                         const char *ev_name, const char *fn_name,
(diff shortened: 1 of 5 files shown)


palmtenor:
Added another commit and updated PR title and description. @yonghong-song could you take another look? Sorry for the inconvenience.

PR #1497, opened 2017-12-21
Fix incorrect lost_cb type in Python

palmtenor:
This is a bug fix for #1496, sorry for the mistake.
Added `lost_cb` in test cases so this won't happen in the future.

### DIFF
--- a/tests/python/test_array.py
+++ b/tests/python/test_array.py
@@ -49,6 +49,9 @@ class TestArray(TestCase):
             event = ct.cast(data, ct.POINTER(Data)).contents
             self.counter += 1
 
+        def lost_cb(lost):
+            self.assertGreater(lost, 0)
+
         text = """
 BPF_PERF_OUTPUT(events);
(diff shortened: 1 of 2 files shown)


tests/python/test_array.py
```diff
@@ -49,6 +49,9 @@ def cb(cpu, data, size):
             event = ct.cast(data, ct.POINTER(Data)).contents
             self.counter += 1
 
+        def lost_cb(lost):
+            self.assertGreater(lost, 0)
+
```
yonghong-song: How can you be sure there must be >0 lost?

palmtenor: I assume when the callback is called there must be >0 lost count? Or would Kernel emit `PERF_RECORD_LOST` even for 0 lost samples?

If you think it's better I could just change this to `pass`

yonghong-song: Just checked the kernel side, yes when LOST_SAMPLE type is recorded, the lost count is >0. This kind of makes sense as well. Thanks!

PR #1501, opened 2017-12-25
Wrap long paragraphs in README.md

pclouds:
Markdown is supposed to be human friendly even in raw format. Since most paragraphs are line wrapped, these three stand out and make it a bit hard to read. Wrap them.


brendangregg:
No thanks.

PR #1504, opened 2017-12-28
CentOS 6 support

samofatov:
With these changes BCC can be built on CentOS 6/RHEL 6/GosLinux IC4/etc


brendangregg:
[buildbot, ok to test]

PR #1508, opened 2017-12-30
tcplife: add tcp:tcp_set_state tracepoint support

brendangregg:


tools/tcplife.py
```diff
@@ -221,6 +235,131 @@
...
+
+    if (mep != 0)
+        whoami.delete(&sk);
+
+    return 0;
+}
+"""
+
+if (os.path.exists(TRACEFS + "/events/tcp/tcp_set_state")):
+    bpf_text += bpf_text_tracepoint
```
goldshtn: I think we have an API in bcc that enumerates tracepoints, so you could use that instead of the file check. If it doesn’t support checking for one specific tracepoint, maybe it should be added there — I think other tools might need it in the future.

brendangregg:
@goldshtn Good idea, pushed. I made it match one (and not a regexp), since I think that'll be the common case.

PR #1512, opened 2018-01-03
update INSTALL.md with notes from arm64 install

vielmetti:
Four things of note for the arm64 install of bcc on Ubuntu 16.04 (based on the Packet Type 2A Cavium ThunderX server).

1. clang dependency on unreleased code; this should get simpler by end of February when there's a 6.0.0 LLVM release. clang takes a bunch of memory to build, so a released version will help build times and ease of install.
2. LuaJIT dependency on unpackaged code for Xenial, which looks like it's addressed in Bionic as that release will have the 2.1.0-beta3 release in it.
3. A few stray dependencies for test tools, I don't know if I have them all, especially all of the Python dependencies.
4. The `cmake3` package is just `cmake` on Xenial and beyond.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -234,10 +234,13 @@ sudo dpkg -i *bcc*.deb
 ## Ubuntu - Source
 
 To build the toolchain from source, one needs:
-* LLVM 3.7.1 or newer, compiled with BPF support (default=on)
+* LLVM 
+** For Intel systems, 3.7.1 or newer, compiled with BPF support (default=on)
+** For arm64 systems, 6.0.0 or newer, compiled with BPF support (default=on)
 * Clang, built from the same tree as LLVM
 * cmake (>=3.1), gcc (>=4.7), flex, bison
 * LuaJIT, if you want Lua support
+** for arm64 systems, LuaJIT 2.1.0-beta3 or newer
 
 ### Install build dependencies
 ```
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, ok to test]

INSTALL.md
```diff
@@ -249,12 +252,28 @@ deb-src http://llvm.org/apt/$VER/ llvm-toolchain-$VER-3.7 main" | \
...
+  
+# For Lua support (Intel)
 sudo apt-get -y install luajit luajit-5.1-dev
+
+# Xenial 16.04 LTS on arm64
+#
+# The arm64 system llvm and clang are not new enough to build this, and at the
+# moment LLVM 6.0.0 is not yet released. Until it is, build it from source
+# according to the supplied directions at http://llvm.org/
+#
```
yonghong-song: You can add the following to the above "or using the instruction at below subsection `Build LLVM and Clang development libs` with `-DLLVM_TARGETS_TO_BUILD="BPF;AArch64"`. My format is probably not
correct. Please adjust properly.

yonghong-song:
hi, @vielmetti, any update on this pull request?

vielmetti:
@yonghong-song 

I'm going to update this for Clang 6.0.0 as well as for getting the current state of LuaJit (which seems to be complicated)

bobrik:
I think this part needs to change as well:

* https://github.com/iovisor/bcc/blob/eb819caf39b399dece89/debian/control#L15

vielmetti:
@bobrik is right, updates are needed to generate Debian images. 

In three months, the LLVM situation on arm64 is much better (a binary distribution is available from the http://llvm.org web site), so that's easier. LuaJIT is not improved.

bobrik:
@vielmetti, I don't see 6.0.0 built for arm64 here:

* http://apt.llvm.org/stretch/dists/llvm-toolchain-stretch-6.0/main/

Do you mean some other distribution?

vielmetti:
@bobrik - I know that some LLVM is being bulit by Linaro here

http://lab.llvm.org:8011/buildslaves/linaro-armv8-01-aarch64-full

that said I don't know where the binary artifacts are released that correspond to that build cycle.

vielmetti:
@bobrik I found pre-built binaries here

http://releases.llvm.org/download.html#6.0.0
http://releases.llvm.org/6.0.0/clang+llvm-6.0.0-aarch64-linux-gnu.tar.xz

bobrik:
I filed a ticket to provide llvm arm64 apt builds: https://bugs.llvm.org/show_bug.cgi?id=37116

bobrik:
With llvm 6.0 and bcc tools from tip built on qemu, running on Qualcomm Centriq hardware:

image

Still hoping to get those upstream builds for LLVM.

yonghong-song:
Cool! Thanks for sharing!

PR #1518, opened 2018-01-09
Try use new API to create [k,u]probe with perf_event_open

liu-song-6:
New kernel API allows creating [k,u]probe with perf_event_open.
This patch tries to use the new API. If the new API doesn't work,
we fall back to old API.

The kernel part of this work is available for test at:
git://git.kernel.org/pub/scm/linux/kernel/git/peterz/queue.git perf/core

Thanks,
Song

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -642,44 +642,48 @@ static int bpf_try_perf_event_open_with_probe(const char *name, uint64_t offs,
 }
 
 static int bpf_attach_tracing_event(int progfd, const char *event_path,
-                                    struct perf_reader *reader, int pid) {
-  int efd, pfd, cpu = 0;
+                                    struct perf_reader *reader, int pid,
+                                    int pfd)
+{
+  int efd, cpu = 0;
   ssize_t bytes;
   char buf[PATH_MAX];
   struct perf_event_attr attr = {};
 
-  snprintf(buf, sizeof(buf), "%s/id", event_path);
-  efd = open(buf, O_RDONLY, 0);
-  if (efd < 0) {
-    fprintf(stderr, "open(%s): %s\n", buf, strerror(errno));
-    return -1;
-  }
-
-  bytes = read(efd, buf, sizeof(buf));
-  if (bytes <= 0 || bytes >= sizeof(buf)) {
-    fprintf(stderr, "read(%s): %s\n", buf, strerror(errno));
-    close(efd);
-    return -1;
-  }
-  close(efd);
-  buf[bytes] = '\0';
-  attr.config = strtol(buf, NULL, 0);
-  attr.type = PERF_TYPE_TRACEPOINT;
-  attr.sample_type = PERF_SAMPLE_RAW | PERF_SAMPLE_CALLCHAIN;
-  attr.sample_period = 1;
-  attr.wakeup_events = 1;
-  // PID filter is only possible for uprobe events.
-  if (pid < 0)
-    pid = -1;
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, ok to test]

src/cc/libbpf.c
```diff
@@ -536,6 +536,106 @@ int bpf_attach_socket(int sock, int prog) {
...
 
+#define PMU_TYPE_FILE "/sys/bus/event_source/devices/%s/type"
+static int bpf_find_probe_type(const char *event_type)
+{
+  int fd;
+  int ret;
+  char buf[64];
+
+  snprintf(buf, sizeof(buf), PMU_TYPE_FILE, event_type);
+
```
yonghong-song: Do you need to check the return value of `snprintf`? Is it possible that it may fail in the future?

src/cc/libbpf.c
```diff
@@ -536,6 +536,106 @@ int bpf_attach_socket(int sock, int prog) {
...
+}
+
+#define PMU_RETPROBE_FILE "/sys/bus/event_source/devices/%s/format/retprobe"
+static int bpf_get_retprobe_bit(const char *event_type)
+{
+  int fd;
+  int ret;
+  char buf[64];
+
+  snprintf(buf, sizeof(buf), PMU_RETPROBE_FILE, event_type);
```
yonghong-song: The same here, check `snprintf` return status?

src/cc/libbpf.c
```diff
@@ -637,44 +637,48 @@ static int bpf_try_perf_event_open_with_probe(const char *name, uint64_t offs,
...
 
 static int bpf_attach_tracing_event(int progfd, const char *event_path,
-                                    struct perf_reader *reader, int pid) {
-  int efd, pfd, cpu = 0;
+                                    struct perf_reader *reader, int pid,
+                                    int pfd)
+{
+  int efd, cpu = 0;
   ssize_t bytes;
   char buf[256];
```
yonghong-song: Not related to your patch, but for long file path, the buffer size may not be enough. Could you help change it to PATH_MAX?

src/cc/libbpf.c
```diff
@@ -808,42 +817,47 @@ void *bpf_attach_uprobe(int progfd, enum bpf_probe_attach_type attach_type,
...
-    goto error;
-  }
-  res = snprintf(buf, sizeof(buf), "%c:%ss/%s %s:0x%lx", attach_type==BPF_PROBE_ENTRY ? 'p' : 'r',
-			event_type, event_alias, binary_path, offset);
-  if (res < 0 || res >= sizeof(buf)) {
-    fprintf(stderr, "Event alias (%s) too long for buffer\n", event_alias);
-    goto error;
-  }
+    res = snprintf(event_alias, sizeof(event_alias), "%s_bcc_%d", ev_name, getpid());
+    if (res < 0 || res >= sizeof(event_alias)) {
```
yonghong-song: Not sure what these solid green block is? Is this due to using TAB? In bcc non-kernel code, we use spaces instead of TABs.

yonghong-song:
Please rebase on top of trunk as there is a conflict for ready-to-merge.

src/cc/libbpf.c
liu-song-6: Teng just fixed it :)

src/cc/libbpf.c
liu-song-6: This is because we moved many lines into if (pfd < 0). Diff broke it into several sections, but it is just adding 2 more spaces for the whole chunk (

PR #1522, opened 2018-01-10
Increase bpf_attach_tracing_event buffer size

palmtenor:
It is awesome that the new probe APIs (#1518) is on the horizon. However most machines with older Kernels would still be using the file-based APIs. In `debugfs`, we put the full binary name as uprobe's event name, hence it would be really long and the current 256 buffer size won't be enough.

This commit increase it to `PATH_MAX`.


yonghong-song:
Looks good. Thanks!

PR #1524, opened 2018-01-10
usdt: better handling for probes existing in multiple binary files

yonghong-song:


yonghong-song:
cc @goldshtn @brendangregg @palmtenor 
Please help take a look and give a try. I broke the patch into three commits for easy review. Thanks!

src/cc/usdt.h
```diff
@@ -205,6 +207,7 @@ class Probe {
   uint64_t semaphore()   const { return semaphore_; }
 
   uint64_t address(size_t n = 0) const { return locations_[n].address_; }
+  const char *loc_bin_path(size_t n = 0) const { return locations_[n].bin_path_.c_str(); }
```
goldshtn: `location_bin_path`?

src/cc/usdt/usdt.cc
```diff
@@ -31,7 +31,9 @@
 
 namespace USDT {
 
-Location::Location(uint64_t addr, const char *arg_fmt) : address_(addr) {
+Location::Location(uint64_t addr, std::string bin_path, const char *arg_fmt)
+    : address_(addr),
+      bin_path_(bin_path) {
```
goldshtn: If we're not doing `std::move` from `bin_path`, it's wasteful to pass it by value -- there will be two copies for lvalue arguments.

src/cc/usdt/usdt.cc
```diff
@@ -201,18 +205,18 @@ bool Probe::usdt_getarg(std::ostream &stream) {
   return true;
 }
 
-void Probe::add_location(uint64_t addr, const char *fmt) {
-  locations_.emplace_back(addr, fmt);
+void Probe::add_location(uint64_t addr, std::string bin_path, const char *fmt) {
+  locations_.emplace_back(addr, bin_path, fmt);
```
goldshtn: This will not move `bin_path` because the parameter is an lvalue.

goldshtn:
Perhaps rename the PR to something more indicative?

goldshtn:
LGTM, left some minor comments.

yonghong-song:
@goldshtn Changed the title and will make bin_path related changes tomorrow. Thanks for reviewing!

yonghong-song:
@goldshtn Fixed the problems you commented, could you take a look again?

src/cc/usdt.h
```diff
@@ -255,6 +259,7 @@ class Context {
   ino_t inode() const { return mount_ns_instance_->target_ino(); }
 
   Probe *get(const std::string &probe_name);
+  Probe *get(const std::string &provider_name, const std::string &probe_name);
```
palmtenor: Can you also update C++ API to use the provider-and-name version of `get`? It currently only has the binary code path but we'll need to add PID code path soon:) And in that case maybe we don't need to keep the (kind of incorrect) `get(name)` anymore?

src/cc/usdt/usdt.cc
```diff
@@ -272,13 +277,41 @@ Probe *Context::get(const std::string &probe_name) {
...
+}
+
 bool Context::enable_probe(const std::string &probe_name,
                            const std::string &fn_name) {
   if (pid_stat_ && pid_stat_->is_stale())
     return false;
 
-  auto p = get(probe_name);
-  return p && p->enable(fn_name);
+  // FIXME: we may have issues here if the context has two same probes's
```
palmtenor: Can we also have a `provider_name` **and** `probe_name` version of `enable_probe`?

src/cc/usdt.h
yonghong-song: Yes, for C++ API, USDT constructor needs provider + probe_name, so we can use `get(provider_, name_)` instead of `get(name_)`. Just updated the patch to reflect this.

src/cc/usdt/usdt.cc
yonghong-song: For C++, we do not need `enable_probe(provider_name, probe_name)` yet as no C++ API will use it. For Python API, currently, all our major users, `trace.py`, `argdist.py` and `funccount.py` all use `u:<binary_path>:<probe_name>` to specify usdt, and hence, `enable_probe(probe_name)` is used.

We could have `enable_probe(provider_name, probe_name)` for some specific tools which uses usdt (e.g., tracing java, ruby and python etc.) where no user interface change is needed. This is nice to have but not required as it is a controlled environment, `enable_probe(probe_name)` should just work.

But we cannot get rid of `enable_probe(probe_name)` due to python API with `trace.py` etc. Do we want to change `trace.py` etc. to have `provider_name` name to handle the case where the same binary contains both `provider_a:probe_name` and `provider_b:probe_name`? We can discuss this separately.

palmtenor: I see. For correctness probably yes, but I'd like to know what @goldshtn and @brendangregg thinks......

Issue #1528, opened 2018-01-12
invalid usdt may be generated by linker ld.gold for probe inside templates


yonghong-song:
gcc/g++ (clang as well in most cases if lld is not used) provides two options for linker:
   - ld.bfd
   - ld.gold

`ld.bfd` is traditional linker implementation and `ld.gold` is elf only and is faster as well.
However, `ld.gold` may generate usdt probes with incorrect offset for certain cases as demonstrated below.

The code
```
-bash-4.3$ cat template.h
#include <iostream>
#include "folly/tracing/StaticTracepoint.h"

// Template Declaration
template<typename T>
class Queue
{
public:
    Queue();
    ~Queue();
    void push(T e);
    T pop();
private:
    struct node
    {
	T data;
	node* next;
    };
    typedef node NODE;
    NODE* mHead;
};

// template definition
template<typename T>
Queue<T>::Queue() 
{ 
    mHead = NULL;
}

template<typename T>
Queue<T>::~Queue() 
{
    NODE *tmp;
    while(mHead) {
	tmp = mHead;
	mHead = mHead->next;
	delete tmp;
    }
}

template<typename T>
void Queue<T>::push(T e)
{
    NODE *ptr = new node;
    FOLLY_SDT(test, probe, 1);
    ptr->data = e;
    ptr->next = NULL;
    if(mHead == NULL) {
	mHead = ptr;
	return;
    }
    NODE *cur = mHead;
    while(cur) {
	if(cur->next == NULL) {
	    cur->next = ptr;
	    return;
	}
	cur = cur->next;
    }
}

template<typename T>
T Queue<T>::pop()
{
    if(mHead == NULL) return (T)0;
    NODE *tmp = mHead;
    T d = mHead->data;
    mHead = mHead->next;
    delete tmp;
    return d;
}
-bash-4.3$ cat a.cpp
#include "template.h"
void test_a()
{
    Queue<int> *i = new Queue<int>();
    i->push(10);
    i->push(20);
    i->pop();
    i->pop();
    delete i;
}
-bash-4.3$ cat b.cpp
#include "template.h"
void test_b()
{
    Queue<int> *i = new Queue<int>();
    i->push(10);
    i->pop();
    delete i;
}
-bash-4.3$ cat main.cpp
extern void test_a();
extern void test_b();
int main() {
    test_a();
    test_b();
    return 0;
}
```
The compilation script
```
-bash-4.3$ cat run.sh
# proper include path to folly USDT header files is needed
[...]

PR #1529, opened 2018-01-12
usdt: ignore location with incorrect probe virtual address

yonghong-song:
The patch fixed issue #1528. The linker ld.gold may generate
a probe with invalid virtual address if the section which the probe
relocates against is discarded.

Instead of failing, a simple checking is added such that
if the probe virtual address is less than the address of the first
instruction, print out a warning and ignore this probe.

### DIFF
--- a/src/cc/bcc_elf.c
+++ b/src/cc/bcc_elf.c
@@ -85,7 +85,7 @@ static const char *parse_stapsdt_note(struct bcc_elf_usdt *probe,
 
 static int do_note_segment(Elf_Scn *section, int elf_class,
                            bcc_elf_probecb callback, const char *binpath,
-                           void *payload) {
+                           uint64_t first_inst_offset, void *payload) {
   Elf_Data *data = NULL;
 
   while ((data = elf_getdata(section, data)) != 0) {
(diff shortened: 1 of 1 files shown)


src/cc/bcc_elf.c
```diff
@@ -122,10 +128,26 @@ static int listprobes(Elf *e, bcc_elf_probecb callback, const char *binpath,
   Elf_Scn *section = NULL;
   size_t stridx;
   int elf_class = gelf_getclass(e);
+  uint64_t first_inst_offset = 0;
 
   if (elf_getshdrstrndx(e, &stridx) != 0)
     return -1;
 
+  // Get the offset to the first instruction
+  while ((section = elf_nextscn(e, section)) != 0) {
```
palmtenor: Nit: Maybe you can update `bcc_elf_foreach_load_section` to be more general, so that it could take different filtering condition (note section rather than load section) and reuse the same code.
Up to your call:))

yonghong-song: I will leave it now. We have a couple of other while loops for sections as well. Maybe somebody else can volunteer to do this in the future?

PR #1538, opened 2018-01-17
Adding minimum latency filter to tcpconnlat

dpayne:
I found it useful to filter out all the fast tcp connects so that I can only see slow ones that were causing issues. To solve this I added `-m` and `-u` options to tcpconnlat. The latency filter is only added when the min latency is set to something greater than zero.

I also added the `-v` option to print out the bpf program. I know this was already controlled with the debug setting in the python script, but this makes it a bit more consistent with some of the other scripts here.

### DIFF
--- a/man/man8/tcpconnlat.8
+++ b/man/man8/tcpconnlat.8
@@ -2,7 +2,7 @@
 .SH NAME
 tcpconnlat \- Trace TCP active connection latency. Uses Linux eBPF/bcc.
 .SH SYNOPSIS
-.B tcpconnlat [\-h] [\-t] [\-p PID] [-m MIN_MS] [-u MIN_US] [-v]
+.B tcpconnlat [\-h] [\-t] [\-p PID] [-v] [min_ms]
 .SH DESCRIPTION
 This tool traces active TCP connections
 (eg, via a connect() syscall), and shows the latency (time) for the connection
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, ok to test]

yonghong-song:
Sounds useful to me. You may need to update the tcpconnlat_example.txt file and the man page.
cc @brendangregg

dpayne:
I updated the tcpconnlat_example.txt file and the man page. When I was updating the man page I also noticed that the naming seemed to still have the tcpconnect name instead of tcpconnlat, so I updated that as well.

yonghong-song:
LGTM. Thanks.

brendangregg:
I like it, but I worry a bit about consistency with, say:

```
    ./ext4slower 1           # trace operations slower than 1 ms
[...]
    ./offcputime -m 1000     # trace only events that last more than 1000 usec
    ./offcputime -M 10000    # trace only events that last less than 10000 usec
```
(The long form of those offcputime options are: --min-block-time and --max-block-time.)

Would it be ok to do either?:

```
    ./tcpconnlat 1           # trace connection latency slower than 1 ms
    ./tcpconnlat -m 1000    # trace connection latency slower than 1 ms
```

The first could accept fractions: 0.001 for 1 us. I think the first might be acceptable since it's the primary function of the tool. For things other than the primary function, I put them behind option flags.

dpayne:
Updated with the suggested changes. Also added a check for number below 0.

brendangregg:
Great, thanks, just missing [min_ms] on line 37 of the _example file.

dpayne:
Updated tcpconnlat_example.txt. Good catch, I missed that.

brendangregg:
thanks!

PR #1539, opened 2018-01-17
Add bpf_override_return() helper definition

hMcLauchlan:
This is currently in bpf-next; just adding the helper definition. I tested it with a simple program:

```python
#!/usr/bin/env python

from bcc import BPF

BPF(text='int kprobe__io_ctl_init(void *ctx) { bpf_override_return(ctx,-12); return 0;}').trace_print() 
```
which seems to work as expected.

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -22,6 +22,8 @@ This guide is incomplete. If something feels missing, check the bcc and kernel s
         - [5. bpf_get_current_uid_gid()](#5-bpf_get_current_uid_gid)
         - [6. bpf_get_current_comm()](#6-bpf_get_current_comm)
         - [7. bpf_log2l()](#7-bpflog2l)
+    - [Debugging](#debugging)
+        - [1. bpf_override_return()](#1-bpf_override_return)
     - [Output](#output)
         - [1. bpf_trace_printk()](#1-bpf_trace_printk)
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, ok to test]

goldshtn:
It would be great to also update the reference guide mentioning this helper and what it can do. Thanks!

palmtenor:
Sorry for the late comment, but can you also update `kernel-versions.md` for commit including the support on Kernel side? I assume it would be in next Kernel release, right?

docs/reference_guide.md
```diff
@@ -324,6 +326,28 @@ Examples in situ:
...
+Syntax: ```int bpf_override_return(struct pt_regs *, unsigned long rc)```
+
+Return: 0 on success
+
+When used in a program attached to a function entry kprobe, causes the
+execution of the function to be skipped, immediately returning `rc` instead.
+This is used for targeted error injection. 
+
+Note: bpf_override_return will only work when the kprobed function is
+whitelisted to allow error injections.
```
goldshtn: Sorry, for the non-initiated, it would be very useful to explain how to discover whether a function is whitelisted for error injection, and what kind of error to expect if it is not whitelisted. Thanks!

hMcLauchlan:
@goldshtn @palmtenor

yonghong-song:
[buildbot, ok to test]

PR #1541, opened 2018-01-19
Fix folding option

linuxraptor:
Using the "folding" option produces an error.

/usr/share/bcc/tools/offwaketime -f 3 > out.stacks

  File "/usr/share/bcc/tools/offwaketime", line 305, in <module>
    print("%s %d" % (str.join(";", line), v.value))
TypeError: sequence item 0: expected str instance, bytes found

It seems that k.target and k.waker are always type = bytes.  This edit resolves this error.
Thank you for your time.


yonghong-song:
[buildbot, ok to test]

brendangregg:
thanks!

PR #1544, opened 2018-01-20
Small fixes to some man pages for bcc tools

dpayne:
Hi,

I noticed a couple of small naming mismatches when I was looking through the man pages. This patch fixes the ones I noticed.


yonghong-song:
[buildbot, ok to test]

PR #1546, opened 2018-01-23
execsnoop: Retrieve PPID on kernel side with bpf_get_current_task

pchaigno:
Retrieve the PPID on kernel side through the `bpf_get_current_task` helper as discussed in the comments of `execsnoop.py`:

https://github.com/iovisor/bcc/blob/a44d26ed3e5e40eaad8e69d630562fb389266bee/tools/execsnoop.py#L158-L161

This new version of execsnoop requires Linux 4.8 minimum.

/cc @brendangregg

### DIFF
--- a/tools/old/execsnoop.py
+++ b/tools/old/execsnoop.py
@@ -63,7 +63,6 @@ enum event_type {
 
 struct data_t {
     u32 pid;  // PID as in the userspace term (i.e. task->tgid in kernel)
-    int ppid;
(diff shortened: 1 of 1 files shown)


pchaigno:
> Maybe just adding another test in bpf_print_hints() in cc/libbpf.c, that searches for "invalid func 35" then prints out a HINT paragraph?

I've made a new hint for `invalid func` errors in #1557. I'll hold the present PR until #1557 (or something equivalent) is merged.

PR #1547, opened 2018-01-23
doc: bpf_get_current_task helper

pchaigno:
Add an entry for `bpf_get_current_task` in the reference guide.

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -21,7 +21,8 @@ This guide is incomplete. If something feels missing, check the bcc and kernel s
         - [4. bpf_get_current_pid_tgid()](#4-bpf_get_current_pid_tgid)
         - [5. bpf_get_current_uid_gid()](#5-bpf_get_current_uid_gid)
         - [6. bpf_get_current_comm()](#6-bpf_get_current_comm)
-        - [7. bpf_log2l()](#7-bpflog2l)
+        - [7. bpf_get_current_task()](#7-bpf_get_current_task)
+        - [8. bpf_log2l()](#8-bpflog2l)
     - [Debugging](#debugging)
         - [1. bpf_override_return()](#1-bpf_override_return)
     - [Output](#output)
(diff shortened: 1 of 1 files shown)


yonghong-song:
> wasn’t there some issue with field randomization in recent kernel versions?

4.13 has issues and you will need to put the below before your header files:
```
#define randomized_struct_fields_start  struct {
#define randomized_struct_fields_end    };
```

Other kernel versions should be okay.

pchaigno:
> It would be nice to specify what you can get from the task_struct.

@goldshtn I've added a sentence with a few examples. If you have other more relevant examples, please tell me!

> 4.13 has issues and you will need to put the below before your header files:

Since the `randomized_struct_fields_xxx` directives are already defined in `linux/compiler.h`, which itself is included by `linux/sched.h`, do we really need to make a note about it? It looks like a very specific corner case (?)

yonghong-song:
At `linux:include/linux` directory, for 4.13,
```
-bash-4.2$ egrep -r randomized_struct_fields_start                                                                                   
compiler-gcc.h:#define randomized_struct_fields_start   struct {
compiler.h:#ifndef randomized_struct_fields_start
compiler.h:# define randomized_struct_fields_start
sched.h:        randomized_struct_fields_start
-bash-4.2$
```
4.15
```
-bash-4.2$ egrep -r randomized_struct_fields_start
compiler-clang.h:#define randomized_struct_fields_start struct {
compiler-gcc.h:#define randomized_struct_fields_start   struct {
sched.h:        randomized_struct_fields_start
compiler_types.h:#ifndef randomized_struct_fields_start
compiler_types.h:# define randomized_struct_fields_start
-bash-4.2$ 
```

for clang, `randomized_struct_fields_start` is not defined in 4.13, it is defined in 4.14 stable and 4.15.  So people use task_struct with 4.13 in their bpf programs may have issues. It would be good to make a note of this, so we can refer to it when people have issues.

pchaigno:
I see. I though it was only necessary when the kernel headers and the kernel used differed because I couldn't reproduce with any bcc tool under 4.13. I'll add a note.

pchaigno:
Done.

Issue #1533, opened 2018-01-16
Python crash when running as PCP PMDA plugin


myllynen:
While testing PCP BCC PMDA (see https://github.com/performancecopilot/pcp/pull/409) we occasionally see Python that runs the BCC module crashing on pmcd (the PCP daemon running plugins) shutdown. PCP has other Python plugins as well and we've never seen this issue with them so this and the stack trace make me think the issue is originating from BCC. I realize this might be a hard-to-reproduce issue especially for those not familiar with PCP but the initial GitHub page at https://github.com/myllynen/pcp-bcc-pmda provides step-by-step instructions how to set this is up and I'd be happy to test any possible patch proposals if someone sees what might be causing this.

```
Process 6993 (python3) of user 0 dumped core.

Stack trace of thread 7008:
#0  0x00007f4c142916cd perf_reader_event_read (libbcc.so.0)
#1  0x00007f4c14291b03 perf_reader_poll (libbcc.so.0)
#2  0x00007f4c18979d1e ffi_call_unix64 (libffi.so.6)
#3  0x00007f4c1897968f ffi_call (libffi.so.6)
#4  0x00007f4c18b8da29 _call_function_pointer (_ctypes.cpython-36m-x86_64-linux-gnu.so)
#5  0x00007f4c18b8dc14 PyCFuncPtr_call (_ctypes.cpython-36m-x86_64-linux-gnu.so)
#6  0x00007f4c22058204 _PyObject_FastCallDict (libpython3.6m.so.1.0)
#7  0x00007f4c220c73ad call_function (libpython3.6m.so.1.0)
#8  0x00007f4c220f0eca _PyEval_EvalFrameDefault (libpython3.6m.so.1.0)
#9  0x00007f4c220570b3 _PyEval_EvalCodeWithName (libpython3.6m.so.1.0)
#10 0x00007f4c22089e11 fast_function (libpython3.6m.so.1.0)
#11 0x00007f4c220c733e call_function (libpython3.6m.so.1.0)
#12 0x00007f4c220f0eca _PyEval_EvalFrameDefault (libpython3.6m.so.1.0)
#13 0x00007f4c22057d9a _PyFunction_FastCall (libpython3.6m.so.1.0)
#14 0x00007f4c220583de _PyObject_FastCallDict (libpython3.6m.so.1.0)
#15 0x00007f4c22062261 _PyObject_Call_Prepend (libpython3.6m.so.1.0)
[...]

PR #1549, opened 2018-01-24
fix a race condition between perf_reader munmap and read

yonghong-song:
Fix issue #1533.

Currently, there exist a race condition between perf_reader
buffer munmap and read if they are happening in two different
threads, crash is possible as in issue #1533.

          thread 1                    thread 2
          perf_reader_event_read      ...
                                      detach_probe
                                      munmap
          access ring buffer

detach_probe may happen as part of bpf object cleanup process at which point
thread 1 is still alive. In this case, accessing ring buffer may
cause segfault since the original mmap'ed memory is not available
any more.

It is hard to fix such races outside bcc since user
calls kprobe_poll which has valid BPF object when it is called,
but race happens inside the kprobe_poll.

This patch adds a state of the ring buffer and the read will
not happen once the state comes to the munmap.

### DIFF
--- a/src/cc/perf_reader.c
+++ b/src/cc/perf_reader.c
@@ -28,6 +28,12 @@
 #include "libbpf.h"
 #include "perf_reader.h"
 
+enum {
+  RB_NOT_USED = 0, // ring buffer not usd
+  RB_USED_IN_MUNMAP = 1, // used in munmap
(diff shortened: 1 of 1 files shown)


yonghong-song:
cc @myllynen @palmtenor

PR #1550, opened 2018-01-24
adding filtering example to trace.py help

tehnerd:
there is no straight forward example of how to use filtering for situation when you are using custom function's signature. this diff is adding it


yonghong-song:
[buildbot, ok to test]

yonghong-song:
Looks like py_test_usdt3 a little bit flaky. Will check this separately. At the same time, rerun the test.

yonghong-song:
[buildbot, ok to test]

PR #1558, opened 2018-01-28
cc: missing bpf helpers in export/helpers.h

pchaigno:
This pull request adds several missing helpers to `src/cc/export/helpers.h`.

### DIFF
--- a/src/cc/export/helpers.h
+++ b/src/cc/export/helpers.h
@@ -250,6 +250,30 @@ static int (*bpf_perf_event_read_value)(void *map, u64 flags, void *buf, u32 buf
   (void *) BPF_FUNC_perf_event_read_value;
 static int (*bpf_perf_prog_read_value)(void *ctx, void *buf, u32 buf_size) =
   (void *) BPF_FUNC_perf_prog_read_value;
+static int (*bpf_current_task_under_cgroup)(void *map, int index) =
+  (void *) BPF_FUNC_current_task_under_cgroup;
+static u32 (*bpf_get_socket_cookie)(void *ctx) =
+  (void *) BPF_FUNC_get_socket_cookie;
+static u64 (*bpf_get_socket_uid)(void *ctx) =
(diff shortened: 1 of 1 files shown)


src/cc/export/helpers.h
```diff
@@ -250,6 +250,30 @@ static int (*bpf_perf_event_read_value)(void *map, u64 flags, void *buf, u32 buf
...
+static int (*bpf_setsockopt)(void *ctx, int level, int optname, void *optval, int optlen) =
+  (void *) BPF_FUNC_setsockopt;
+static int (*bpf_skb_adjust_room)(void *ctx, int len_diff, u32 mode, u64 flags) =
+  (void *) BPF_FUNC_skb_adjust_room;
+static int (*bpf_skb_under_cgroup)(void *ctx, void *map, int index) =
+  (void *) BPF_FUNC_skb_under_cgroup;
+static int (*bpf_sk_redirect_map)(void *ctx, void *map, int key, int flags) =
+  (void *) BPF_FUNC_sk_redirect_map;
+static int (*bpf_sock_map_update)(void *map, void *key, void *value, unsigned long long flags) =
+  (void *) BPF_FUNC_sock_map_update;
```
yonghong-song: For all other map_update helpers, bcc has a map.update(key, val), maybe we could also use this shortcut for bpf_sock_map_update as well? All other helper definitions are fine.

pchaigno: > For all other map_update helpers, bcc has a map.update(key, val), maybe we could also use this shortcut for bpf_sock_map_update as well?

Sure that makes sense! 👍 
It will require some changes to the rewriter to distinguish sock maps from other maps and to the Python side to add a new sock map class. We still need to define it in `src/cc/export/helpers.h`. If you don't mind, I'd prefer to work on this in a separate PR; I don't know when I'll have time to send a new set of patches.

yonghong-song:
Okay, merged and filed issue #1564 to track the potential improvement.

Issue #1033, opened 2017-03-08
Reduce CPU overhead of per-event tools by allowing more buffering


brendangregg:
Not a big problem yet, but at some point we should look into this...

The normal perf_output-consuming loop we've been using is:

```
while 1:
    b.kprobe_poll()
```

This wakes up as quickly as it can, and can consume noticeable CPU. I had to tune it for my cpuunclaimed tool:

```
while 1:
    # allow some buffering by calling sleep(), to reduce the context switch
    # rate and lower overhead.
    try:
        if not exiting:
            sleep(wakeup_s)
    except KeyboardInterrupt:
        exiting = 1
    b.kprobe_poll()
    slept += wakeup_s
```

This ticket is to explore doing this for more tools. Should b.kprobe_poll() have a built-in sleep, that can be tuned? Eg, 100ms by default?

This is more practical now we have #997, and can increase the buffer size.

brendangregg:
A problem of a built-in sleep to kprobe_poll() would be Ctrl-C behavior. Maybe it's fine if it just lets it quit early, but the calling program would need to know Ctrl-C was hit so it can begin the exit code path.

vitalyvch:
In my opinion, the best way is to implement it in kernel and to configure through ioctl(), like ttys and sockets do :-)

> Should b.kprobe_poll() have a built-in sleep, that can be tuned? 

Most likely no. I believe, much better is an example of this good practice in the documentation.

> 100ms by default?

I prefer something about 30 ms.

> A problem of a built-in sleep to kprobe_poll() would be Ctrl-C behaviour.

I don't like the current handling of Ctrl-C at all. Probably it is a good idea to redevelop or to refactor it.

brendangregg:
I'd allow a 30 ms sleep through an argument:

`b.kprobe_poll(30)`

It's about finding the balance: too short and the tool wakes up too often and consumes CPU. Too long and the tool may overflow its buffer. I'm guessing a 100 ms default for all tools would work, then we tune some as needed.

yonghong-song:
Not sure whether `b.kprobe_poll(30)` would reduce the CPU usage or not as the `poll` will still wake up immediately if the file descriptor is ready. We may have to implement the `sleep` functionality in the library somehow to really enforce the waiting period between subsequent `poll`s.

@brendangregg is this what you want?

brendangregg:
[...]

PR #1560, opened 2018-01-28
opensnoop: -d option for duration

pchaigno:


yonghong-song:
[buildbot, ok to test]

yonghong-song:
Looks like `-d` starts to appear in more and more tools. Maybe it is worthwhile to standardize this option to all tools (even including trace.py, argdist.py, etc.)?

yonghong-song:
cc @brendangregg  @goldshtn

brendangregg:
Yes, we'll likely add a "-d duration_secs" to more tools that don't already have a duration, as it's useful for scripting.

Note that this code only works because b.kprobe_poll() doesn't sleep, but we might want to change that to a 100 ms default. See #1033. In this case, having a 100 ms rounding error on the sleep duration shouldn't be a big problem.

PR #1561, opened 2018-01-28
Miscellaneous BCC patches

joelagnel:
Here are 3 miscellaneous patches while I prepare the broader set of changes for making BCC work with BPFd.

### DIFF
--- a/src/python/bcc/table.py
+++ b/src/python/bcc/table.py
@@ -205,6 +205,11 @@ class TableBase(MutableMapping):
             errstr = os.strerror(ct.get_errno())
             raise Exception("Could not update table: %s" % errstr)
 
+    def __delitem__(self, key):
+        res = lib.bpf_delete_elem(self.map_fd, ct.byref(key))
+        if res < 0:
+            raise KeyError
+
     # override the MutableMapping's implementation of these since they
     # don't handle KeyError nicely
     def itervalues(self):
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, ok to test]

src/python/bcc/table.py
```diff
@@ -430,11 +430,11 @@ def __setitem__(self, key, leaf):
 
     def __delitem__(self, key):
         key = self._normalize_key(key)
-        # Deleting from array type maps does not have an effect, so
-        # zero out the entry instead.
+        super(ArrayBase, self).__delitem__(key)
+
+    def clearitem(self, key):
         leaf = self.Leaf()
```
yonghong-song: should we do _normalize_key here?

joelagnel: Yes, The user's of `ArrayBase`'s __delitem__ after this patch are, `PerfEventArray` and `ProgArray`. Both these were calling `_normalize_key` in the previous code (before this patch). 

However I screwed up the normalization for __delitem__ in `PerCpuArray` and `Array`. Those need to normalize too before calling `clearitem`. Or `clear_item` itself should normalize within. I'll fix that in the next rev. thanks

src/python/bcc/table.py
```diff
@@ -665,6 +661,10 @@ def __getitem__(self, key):
     def __setitem__(self, key, leaf):
         super(PerCpuArray, self).__setitem__(key, leaf)
 
+    def __delitem__(self, key):
+        # Delete in this type does not have an effect, so zero out instead
```
joelagnel: Need to add a `_normalize_key` here, on in `clearitem`

src/python/bcc/table.py
joelagnel: Fixed this.

PR #1569, opened 2018-02-02
BCC cross compilation support

joelagnel:
These patches make it possible to build eBPF programs from a custom kernel source directory and a custom user-specified ARCH. This support is one of the pieces for the effort to make BCC work across systems using BPFd.

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -17,6 +17,7 @@
 #include <linux/version.h>
 #include <sys/utsname.h>
 #include <unistd.h>
+#include <stdlib.h>
 
 #include <clang/AST/ASTConsumer.h>
 #include <clang/AST/ASTContext.h>
@@ -30,6 +31,7 @@
 #include "common.h"
 #include "loader.h"
 #include "table_storage.h"
+#include "arch_helper.h"
 
 #include "libbpf.h"
 
@@ -47,16 +49,35 @@ const char *calling_conv_regs_s390x[] = {"gprs[2]", "gprs[3]", "gprs[4]",
 
 const char *calling_conv_regs_arm64[] = {"regs[0]", "regs[1]", "regs[2]",
                                        "regs[3]", "regs[4]", "regs[5]"};
-// todo: support more archs
-#if defined(__powerpc__)
-const char **calling_conv_regs = calling_conv_regs_ppc;
-#elif defined(__s390x__)
-const char **calling_conv_regs = calling_conv_regs_s390x;
-#elif defined(__aarch64__)
-const char **calling_conv_regs = calling_conv_regs_arm64;
-#else
-const char **calling_conv_regs = calling_conv_regs_x86;
-#endif
+
+void *get_call_conv_cb(bcc_arch_t arch)
+{
+  const char **ret;
+
+  switch(arch) {
+    case BCC_ARCH_PPC:
+    case BCC_ARCH_PPC_LE:
+      ret = calling_conv_regs_ppc;
+      break;
+    case BCC_ARCH_S390X:
+      ret = calling_conv_regs_s390x;
+      break;
+    case BCC_ARCH_ARM64:
+      ret = calling_conv_regs_arm64;
+      break;
+    default:
+      ret = calling_conv_regs_x86;
+  }
+
+  return (void *)ret;
+}
+
+const char **get_call_conv(void) {
+  const char **ret;
+
+  ret = (const char **)run_arch_callback(get_call_conv_cb);
+  return ret;
+}
 
 using std::map;
 using std::move;
@@ -256,6 +277,8 @@ BTypeVisitor::BTypeVisitor(ASTContext &C, BFrontendAction &fe)
     : C(C), diag_(C.getDiagnostics()), fe_(fe), rewriter_(fe.rewriter()), out_(llvm::errs()) {}
 
 bool BTypeVisitor::VisitFunctionDecl(FunctionDecl *D) {
+  const char **calling_conv_regs = get_call_conv();
+
   // put each non-static non-inline function decl in its own section, to be
   // extracted by the MemoryManager
   auto real_start_loc = rewriter_.getSourceMgr().getFileLoc(D->getLocStart());
--- a/src/cc/frontends/clang/kbuild_helper.cc
+++ b/src/cc/frontends/clang/kbuild_helper.cc
@@ -14,7 +14,9 @@
  * limitations under the License.
  */
 #include <fcntl.h>
+#include <stdlib.h>
 #include <ftw.h>
+#include <iostream>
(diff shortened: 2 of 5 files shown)


src/cc/frontends/clang/b_frontend_action.cc
```diff
@@ -47,16 +48,35 @@ const char *calling_conv_regs_s390x[] = {"gprs[2]", "gprs[3]", "gprs[4]",
 
 const char *calling_conv_regs_arm64[] = {"regs[0]", "regs[1]", "regs[2]",
                                        "regs[3]", "regs[4]", "regs[5]"};
+
+const char **get_call_conv(void) {
+  const char *archenv = getenv("ARCH");
+  if (!archenv)
+    archenv = getenv("BCC_ARCH");
+
```
yonghong-song: what is the difference between ARCH and BCC_ARCH? Should we use an environment variable or a cmake macro definition?

src/cc/frontends/clang/kbuild_helper.cc
```diff
@@ -56,6 +60,17 @@ int KBuildHelper::get_flags(const char *uname_machine, vector<string> *cflags) {
     arch = "arm64";
   }
 
+  // If ARCH env is defined, use it over uname
+  archenv = getenv("ARCH");
+  if (!archenv)
+    archenv = getenv("BCC_ARCH");
+
+  if (archenv) {
+    if (getenv("BCC_KBUILD_DEBUG"))
```
yonghong-song: Can we use one of existing debug flags for output here? For example, debug=4 for preprocessor result can be used since this impacts the clang commands?

src/cc/frontends/clang/loader.cc
```diff
@@ -108,11 +109,24 @@ int ClangLoader::parse(unique_ptr<llvm::Module> *mod, TableStorage &ts,
   unique_ptr<llvm::MemoryBuffer> main_buf;
   struct utsname un;
   uname(&un);
-  string kdir = string(KERNEL_MODULES_DIR) + "/" + un.release;
-  auto kernel_path_info = get_kernel_path_info (kdir);
+  string kdir, kpath;
+  const char *kpath_env = ::getenv("BCC_KERNEL_SOURCE");
```
yonghong-song: Again about env variable vs. cmake macro? I would prefer cmake macro to make the code less dependent on external factors. If use cmake macro, maybe KERNEL_SOURCE_DIR as the macro name?

src/cc/frontends/clang/b_frontend_action.cc
joelagnel: CMake macro definition would make it static then right? We want it to be dynamic. This is indeed how the kernel is built in a cross compiler environment as well. One does something like this in kernel build directory, for example:
```
ARCH=arm64 make zImage
```
Similarly, we need BCC to adapt to the user's cross compile environment like the kernel does.

src/cc/frontends/clang/kbuild_helper.cc
joelagnel: I'll look into that, thanks

src/cc/frontends/clang/loader.cc
joelagnel: Same reason as above, BCC should adapt to different kernel sources that the same BCC installation can point to dynamically. We shouldn't restrict this to a statically known path.

src/cc/frontends/clang/kbuild_helper.cc
joelagnel: Sure, I'll reuse `DEBUG_PREPROCESSOR` to print the chose ARCH. Can I also reuse `DEBUG_PREPROCESSOR` to print the kernel source directory path or does another flag make more sense?

src/cc/frontends/clang/b_frontend_action.cc
yonghong-song: okay, env seems fine here. What is the different between ARCH and BCC_ARCH?

src/cc/frontends/clang/kbuild_helper.cc
yonghong-song: You can just reuse `DEBUG_PREPROCESSOR` to print the kernel source directoryy path.

yonghong-song: Could you think of a way to put the code dealing different architectures in one place (header is okay), so the main.c file will be clean?

src/cc/frontends/clang/b_frontend_action.cc
joelagnel: I just thought of giving `BCC_ARCH` as an additional option since most (all?) env vars or macros are prefixed with `BCC_`. There's no difference between them.

src/cc/frontends/clang/kbuild_helper.cc
joelagnel: @yonghong-song didn't follow your comment. When you say `main.c`, do you mean the `/virtual/main.c` file?

Also the code dealing with different architectures is already in a separate header (`helpers.h`). Do you mean the code shared between `loader.cc` and `b_frontend_action.cc` should be made common? I think we can do that using function pointers and enums. So we could define one enum per arch, and pass a function pointer callback to generic code that handles different arches, which would call then call the callback after detecting the ARCH. That would keep the "arch detection logic" separate from the code that "uses the arch and does different things".

joelagnel: Note that I would keep `helpers.h` as is in the original patch though since that code isn't shared by the CC code (unless I missed something).

src/cc/frontends/clang/b_frontend_action.cc
yonghong-song: Can we just one, e.g., `BCC_TARGET_ARCH`? Having two is not really necessary here, in my opinion.

src/cc/frontends/clang/kbuild_helper.cc
yonghong-song: Sorry for confusion. yes, I do `mean the code shared between loader.cc and b_frontend_action.cc should be made common`.

src/cc/frontends/clang/loader.cc
yonghong-song: Okay with me for this.

src/cc/frontends/clang/b_frontend_action.cc
joelagnel: Yes I agree we can have one. Hope you agree that lets just keep `ARCH` because this what the kernel uses as well so it will simplify the kernel developer's workflow here.

src/cc/frontends/clang/kbuild_helper.cc
joelagnel: Got it.

joelagnel:
Simplified the patch, its much cleaner. Let me know comments if any.

joelagnel:
Yes, I agree. thanks

PR #1570, opened 2018-02-02
continue adding --ebpf option to the python tools/ scripts

natoscott:
Several python tools allow their eBPF code to be printed to
stdout for debugging.  There are other projects that would
like to share these program definitions however, instead of
duplicating code.  We previously agreed on an --ebpf option
and we now continue adding it to more tools.

Signed-off-by: Nathan Scott <nathans@redhat.com>

### DIFF
--- a/tools/execsnoop.py
+++ b/tools/execsnoop.py
@@ -46,6 +46,8 @@ parser.add_argument("-l", "--line",
     help="only print commands where arg contains this line (regex)")
 parser.add_argument("--max-args", default="20",
     help="maximum number of arguments parsed and displayed, defaults to 20")
+parser.add_argument("--ebpf", action="store_true",
+    help=argparse.SUPPRESS)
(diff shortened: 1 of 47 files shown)


yonghong-song:
[buildbot, ok to test]

tools/execsnoop.py
```diff
@@ -128,8 +130,13 @@
 }
 """
 
+bpf_text.replace("MAXARG", args.max_args)
```
yonghong-song: You need `bpf_text = bpf_text.replace(...)`.  Otherwise, the statement has no effect and the tool will fail.

brendangregg:
other than @yonghong-song 's comment, LGTM.

natoscott:
@yonghong-song good catch - I've updated the PR with that fix.

yonghong-song:
Looks good. Thanks.

PR #1574, opened 2018-02-05
tcptracer: Fix argparse is not defined error

GalPressman:
argparse is imported as ap, using argparse explicitly results in a name
not defined error.

Signed-off-by: Gal Pressman <galp@mellanox.com>


yonghong-song:
[buildbot, ok to test]

yonghong-song:
LGTM.

PR #1575, opened 2018-02-05
Updated the FAQ with the error produced if python[2-3]-bcc isn't inst…

Matty9191:
Updated the FAQ with the error produced when python[2-3]-bcc isn't installed.

- Ryan


yonghong-song:
[buildbot, ok to test]

PR #1573, opened 2018-02-04
Add missing includes to fix build against libc++

springmeyer:
This fixes the compile with `-stdlib=libc++`. Tested on ubuntu precise and LLVM 5.0.0.

PR #1577, opened 2018-02-05
Fix missing include

springmeyer:
This is a followup to #1573. I found one more instance of a missing header that (must) only break the compile on linux with libc++:

This PR fixes this issue for me:

```
In file included from ../tests/cc/test_array_table.cc:17:
In file included from ../src/cc/api/BPF.h:24:
../src/cc/api/BPFTable.h:118:71: error: use of undeclared identifier 'errno'
      return StatusTuple(-1, "Error getting value: %s", std::strerror(errno));
```

### DIFF
--- a/src/cc/api/BPFTable.h
+++ b/src/cc/api/BPFTable.h
@@ -17,6 +17,7 @@
 #pragma once
 
 #include <cstring>
+#include <errno.h>
 #include <sys/epoll.h>
 #include <exception>
 #include <map>


yonghong-song:
[buildbot, ok to test]

yonghong-song:
Just comment on why we have suddenly have this missing header issues. I checked one of instances. Looks like `errno.h` is included through C++ `cstdio` in my environment. It is possible that different C++ compiler or C++ compiler configuration may prevent `cstdio` including `errno.h`, resulting in an error in certain cases.

To make it less dependent on compilers, adding an explicit include is totally fine to me.

PR #1578, opened 2018-02-06
Document bpf_tail_call()

lcp:
These commits are mainly to document bpf_tail_call(). For convenience, a new macro, BPF_PROG_ARRAY, is defined to replace BPF_TABLE("prog"), and a new section is created to introduce BPF_PROG_ARRAY.


goldshtn:
[buildbot, ok to test]

PR #1583, opened 2018-02-09
Fix two issues in examples

liubogithub:
Two cosmetic fixes here, they may mislead new users about how to write a tracing script.


yonghong-song:
[buildbot, ok to test]

PR #1587, opened 2018-02-12
opensnoop: convert bytes to str

lcp:
This commit fixes the following error with python3:

  File "/usr/lib/python3.6/site-packages/bcc/table.py", line 508, in raw_cb_
    callback(cpu, data, size)
  File "./opensnoop.py", line 169, in print_event
    if args.name and args.name not in event.comm:


yonghong-song:
[buildbot, ok to test]

lcp:
Ok, I updated the commit to parse "-n" with ArgString.

lcp:
Close this PR since I just sent a V2 patch.

PR #1589, opened 2018-02-12
filesystem scripts: check filesystem by superblock's magic number

liubogithub:
Scripts of btrfs and ext4 get updated to use superblock's magic number.

### DIFF
--- a/tools/btrfsdist.py
+++ b/tools/btrfsdist.py
@@ -16,9 +16,6 @@ from bcc import BPF
 from time import sleep, strftime
 import argparse
 
-# symbols
-kallsyms = "/proc/kallsyms"
-
 # arguments
 examples = """examples:
     ./btrfsdist            # show operation latency as a histogram
@@ -61,6 +58,7 @@ bpf_text = """
 #include <uapi/linux/ptrace.h>
 #include <linux/fs.h>
 #include <linux/sched.h>
+#include <uapi/linux/magic.h>
 
 #define OP_NAME_LEN 8
 typedef struct dist_key {
@@ -83,16 +81,15 @@ int trace_entry(struct pt_regs *ctx)
 
 // The current btrfs (Linux 4.5) uses generic_file_read_iter() instead of it's
 // own read function. So we need to trace that and then filter on btrfs, which
-// I do by checking file->f_op.
+// I do by checking file->f_mapping->host->i_sb->s_magic.
 int trace_read_entry(struct pt_regs *ctx, struct kiocb *iocb)
 {
     u32 pid = bpf_get_current_pid_tgid();
     if (FILTER_PID)
         return 0;
 
-    // btrfs filter on file->f_op == btrfs_file_operations
     struct file *fp = iocb->ki_filp;
-    if ((u64)fp->f_op != BTRFS_FILE_OPERATIONS)
+    if (fp->f_mapping->host->i_sb->s_magic != BTRFS_SUPER_MAGIC)
         return 0;
 
     u64 ts = bpf_ktime_get_ns();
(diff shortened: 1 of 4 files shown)


brendangregg:
What's the reasons?

Pros:
- avoid CPU cost of the /proc/kallsyms read
- anything else?

Cons:
- Is "file->f_mapping->host->i_sb->s_magic" more brittle than "fp->f_op"? Might people be more likely to get "struct X undefined" errors on some distros? Or is that totally fine?

liubogithub:
Simply thought sb's magic number is easier to get than parsing kallsyms as it's defined in <uapi/linux/magic.h>.

Not sure about the 'undefined' error though, Fedora is fine at least.  But I think it'd be fine since struct file, struct address_space, struct inode and struct super_block have been defined in <linux/fs.h>, which has been included in the bpf_text.

brendangregg:
Ok, works for me. If this turns out to be a problem somewhere, we can roll back -- it's a small change.

[buildbot, ok to test]

yonghong-song:
The test failed due to something related to terminal setting in fc{25,26,27}.
For example, in fc27, the error output
```
28: ....clear: terminal attributes: No such device or address
```

I tried the patch in my local fc27 system and it works fine. Not exactly sure what is going on.
Rerun the test to see whether it will be okay or not.

[buildbot, ok to test]

drzaeus77:
I think that is just a warning, and the error is something else.

yonghong-song:
Thanks @drzaeus77, I should have carefully looked at more results before running the test again. Yes, it does look like real failure from test output.

brendangregg:
Is it just failing because the test system doesn't have btrfs, therefore no btrfs symbols to trace?

yonghong-song:
The log says
```
28: ..Traceback (most recent call last):
28:   File "../../tools/btrfsdist.py", line 180, in <module>
28:     b.attach_kprobe(event="btrfs_file_write_iter", fn_name="trace_entry")
28:   File "/home/fedora/jenkins/workspace/bcc-pr/label/fc25/src/python/bcc/__init__.py", line 536, in attach_kprobe
28:     raise Exception("Failed to attach BPF to kprobe")
28: Exception: Failed to attach BPF to kprobe
```

The symbol should be there since without this patch, attach_kprobe is successful. Maybe @drzaeus77 can help try on the test VM?

liubogithub:
The patch also removed the early abort part when btrfs_file_operation is not in kallsyms, is it possible that the previous tests treated that early abort as a 'pass'?

brendangregg:
This works on my system with btrfs. I'm guessing the failure is due to btrfs missing on the test target, because the console output shows it failing on the kprobe:

```
28:   File "../../tools/btrfsdist.py", line 180, in <module>
28:     b.attach_kprobe(event="btrfs_file_write_iter", fn_name="trace_entry")
28:   File "/home/fedora/jenkins/workspace/bcc-pr/label/fc27/src/python/bcc/__init__.py", line 536, in attach_kprobe
28:     raise Exception("Failed to attach BPF to kprobe")
28: Exception: Failed to attach BPF to kprobe
```

So the test is from tests/python/test_tools_smoke.py:

```Python
    def test_btrfsdist(self):
        # Will attempt to do anything meaningful only when btrfs is installed.
        self.run_with_duration("btrfsdist.py 1 1")

    @skipUnless(kernel_version_ge(4,4), "requires kernel >= 4.4")
    def test_btrfsslower(self):
        # Will attempt to do anything meaningful only when btrfs is installed.
        self.run_with_int("btrfsslower.py", allow_early=True)
```

I think this needs another skip rule to skip these tests if btrfs is not installed (eg, no btrfs in ```lsmod``` output).

PR #1593, opened 2018-02-14
Allow to delete elements from prog tables

mauriciovasquezbernal:


yonghong-song:
Could you add a test case somewhere for this new method?

mauriciovasquezbernal:
Sure, I pushed an extra commit with that.

yonghong-song:
[buildbot, ok to test]

PR #1595, opened 2018-02-15
fix hang with "trace.py --max-events #"

yonghong-song:
Currently running "trace.py --max-events #" (shortname "trace.py -M #")
will hang like below:
```
  -bash-4.2$ sudo ./trace.py -M 2 'SyS_futex'
  137727  138229  IOThreadPool0   SyS_futex
  138250  151288  OBC Dispatcher  SyS_futex
  ^C^Z
  [1]+  Stopped                 sudo ./trace.py -M 2 'SyS_futex'
  -bash-4.2$
```
The hang happens in perf_reader_free.

Commit cd5d4a6ce519 ("fix a race condition between perf_reader
munmap and read") fixed a race condition by adding some coordination
between perf_reader_event_read and perf_reader_free.

In this case, however, the callback function inside perf_reader_event_read
never returns and actually calling exit(). This is because
the maximum number of perf_events have been received. The exit()
is calling BPF object cleanup() which calls perf_reader_free to
free the ring buffer. perf_reader_free got stuck since it thinks
perf_reader_event_reader did not finish yet.

The issue exists in C++ as well. If the callback function calls
the BPF object destructor, the program will hang with the same reason
explained in the above.

To fix this issue, we introduce a state pointer to the
raw_cb function. Right before the exit() or destructor,
the callback function can set the perf_reader state to
PERF_READER_CB_EXITING. This will notify later perf_reader_free
function to ignore lock acquisition and go ahead with munmap
and closing the reader file descriptor.
    
Python BPF module API open_perf_buffer adds one parameter
possible_callback_exit so the tools can tell BPF module
whether the event handler may exit or not. If it is,
the callback function will get another parameter which
can be set to PERF_READER_CB_EXITING right before exiting.

After the fix,
```
  -bash-4.2$ sudo ./trace.py -M 2 'SyS_futex'
  137727  138178  load-monitor    SyS_futex
  6359    6440    load-monitor    SyS_futex
  -bash-4.2$
```
Reported-by: Teng Qin <qinteng@fb.com>

### DIFF
--- a/src/cc/perf_reader.c
+++ b/src/cc/perf_reader.c
@@ -20,8 +20,10 @@
 #include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
+#include <syscall.h>
 #include <sys/ioctl.h>
 #include <sys/mman.h>
+#include <sys/types.h>
 #include <unistd.h>
 #include <linux/perf_event.h>
 
@@ -43,6 +45,7 @@ struct perf_reader {
   size_t buf_size;
   void *base;
   int rb_use_state;
+  pid_t rb_read_tid;
   int page_size;
   int page_cnt;
   int fd;
@@ -70,7 +73,12 @@ struct perf_reader * perf_reader_new(perf_reader_cb cb,
 void perf_reader_free(void *ptr) {
   if (ptr) {
     struct perf_reader *reader = ptr;
-    while (!__sync_bool_compare_and_swap(&reader->rb_use_state, RB_NOT_USED, RB_USED_IN_MUNMAP));
+    pid_t tid = syscall(__NR_gettid);
+    while (!__sync_bool_compare_and_swap(&reader->rb_use_state, RB_NOT_USED, RB_USED_IN_MUNMAP)) {
+      // If the same thread, it is called from call back handler, no locking needed
+      if (tid == reader->rb_read_tid)
+        break;
+    }
     munmap(reader->base, reader->page_size * (reader->page_cnt + 1));
     if (reader->fd >= 0) {
       ioctl(reader->fd, PERF_EVENT_IOC_DISABLE, 0);
@@ -228,6 +236,7 @@ void perf_reader_event_read(struct perf_reader *reader) {
   uint8_t *sentinel = (uint8_t *)reader->base + buffer_size + reader->page_size;
   uint8_t *begin, *end;
 
+  reader->rb_read_tid = syscall(__NR_gettid);
   if (!__sync_bool_compare_and_swap(&reader->rb_use_state, RB_NOT_USED, RB_USED_IN_READ))
     return;
 
@@ -280,6 +289,8 @@ void perf_reader_event_read(struct perf_reader *reader) {
     write_data_tail(perf_header, perf_header->data_tail + e->size);
   }
   reader->rb_use_state = RB_NOT_USED;
+  __sync_synchronize();
+  reader->rb_read_tid = 0;
 }
 
 int perf_reader_poll(int num_readers, struct perf_reader **readers, int timeout) {


src/cc/api/BPF.cc
```diff
@@ -559,7 +559,7 @@ std::string BPF::get_uprobe_event(const std::string& binary_path,
 StatusTuple BPF::detach_kprobe_event(const std::string& event,
                                      open_probe_t& attr) {
   if (attr.reader_ptr) {
-    perf_reader_free(attr.reader_ptr);
+    perf_reader_free(attr.reader_ptr, false);
```
palmtenor: For C++, maybe we have these individual detach helpers to `perf_reader_free(..., false)` but the one called in destructor to do `perf_reader_free(..., true)`?

yonghong-song: Right. We do not need to do munmap if we know for sure it is exiting!

Originally, I thought about whether we could return from callback function with indication it wants to exit. There are the following concern:
   . user program may catch Ctrl^C. If this is caught in the middle of perf_read, and the python keyboard exception call exit(), then the hang is not resolved. Is this possible? Should we take care of this?

For the question to call perf_reader_free(..., true) in destructor, we still need to call perf_reader_free(..., false) since the program is not exiting. This still needed for proper
sync between reader and free in normal cases.

yonghong-song: For
> Actually, in that case, we don't even need to do the munmap since the Process is going away, right?

Just checked. We can skip munmap, but perf_reader_free have to be called to close reader_fd. Otherwise, we are not able to clean the entry in `/sys/kernel/debug/tracing/kprobe_events`.

With debugfs still in use for k/uprobes, we may still need to preserve the current execution path, i.e., during atexit, perf_reader_free needs to be properly called.

palmtenor:
Thanks, I think current version is actually a lot nicer. However, it still feels strange having the callback to manipulate the perf reader's internal state, and I think it would be hard for users to comprehend how to make it work correctly without looking at these PRs. I have two ideas on this:
- Instead of passing a pointer to state, we pass `const void* reader_context`. Note that `perf_reader` internals are not exposed to users (in `perf_reader.h`) so it would be just a black-box to users. Then we add another helper, `perf_reader_prepare_exit`, and we clearly document that if you want to do stuffs **inside the callback** that leads the destruction of your `BPF` class, you must call `perf_reader_prepare_exit(reader_context)` in the callback first.
- Otherwise, we could have the callback return a function pointer `void*(void)`. That pointer will do normally be 0 (`NULL`) and the perf reader will do nothing. But when it's not null, the perf reader will just call it, which would execute `exit` or some othero d things (like C++ destruction, etc.).

What do you think?

yonghong-song:
Agreed that I did not particularly like it as well. Just pushed a new approach, please take a look.

src/cc/perf_reader.c
```diff
@@ -70,7 +73,12 @@ struct perf_reader * perf_reader_new(perf_reader_cb cb,
 void perf_reader_free(void *ptr) {
   if (ptr) {
     struct perf_reader *reader = ptr;
-    while (!__sync_bool_compare_and_swap(&reader->rb_use_state, RB_NOT_USED, RB_USED_IN_MUNMAP));
+    pid_t tid = syscall(__NR_gettid);
+    while (!__sync_bool_compare_and_swap(&reader->rb_use_state, RB_NOT_USED, RB_USED_IN_MUNMAP)) {
+      // If the same thread, it is called from call back handler, no locking needed
+      if (tid == reader->rb_read_tid)
```
palmtenor: We'll read a read barrier here, right?

yonghong-song: __sync_bool_compare_and_swap will act as a memory barrier.
https://gcc.gnu.org/onlinedocs/gcc/_005f_005fsync-Builtins.html
Quote
```
In most cases, these built-in functions are considered a full barrier. That is, no memory operand is moved across the operation, either forward or backward. Further, instructions are issued as necessary to prevent the processor from speculating loads across the operation and from queuing stores after the operation. 
```
`__sync_bool_compare_and_swap` is a full barrier.

PR #1596, opened 2018-02-17
scripts: avoid check-helpers.sh's writes to disk

pchaigno:
As a follow up to #1592, @qmonnet suggested the `diff <(ls dir1) <(ls dir2)` syntax to avoid writing to disk to compare the lists of helpers. This pull request implements this and also replaces backticks for command substitution with the recommended parenthesis syntax (`$(diff ...)`).

/cc @yonghong-song @qmonnet

### DIFF
--- a/scripts/check-helpers.sh
+++ b/scripts/check-helpers.sh
@@ -1,25 +1,25 @@
 #!/bin/bash
 ret=0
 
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, ok to test]

PR #1599, opened 2018-02-20
Do not rely on RTLD_DI_ORIGIN

gmile:
This will allow the bcc to be built on Alpine without additional
patches to musl.

### DIFF
--- a/tests/cc/test_c_api.cc
+++ b/tests/cc/test_c_api.cc
@@ -17,6 +17,7 @@
 #include <dlfcn.h>
 #include <stdint.h>
 #include <string.h>
+#include <link.h>
 #include <sys/mman.h>
 #include <sys/mount.h>
 #include <sys/stat.h>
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, ok to test]

yonghong-song:
The testtest_libbcc  failed.
>> 3: Unable to open /lib64/libz.so.1/libz.so.1: Not a directory

Looks like RTLD_DI_ORIGIN gives a library path (not including lib name) while
link_map->l_name includes the path + lib_name.

Could you correct this and resubmit the patch?

Thanks.

gmile:
@yonghong-song thank you for pointing out what the problem was. I was looking at the build logs but could not figure out what was the problem.

I've amended the original commit, the build passes now. Please consider taking another look at the patch.

yonghong-song:
Looks good. Thanks!

PR #1602, opened 2018-02-22
add a BPFModule API to disable rw_engine sscanf/snprintf functions

yonghong-song:
Currently, for every table of a module, corresponding
key/value sscanf/snprintf functions will be generated.
These functions are mostly used for python API to
facilitate passing keys between python and C++.

It is a known issue that these sscanf/snprintf functions
can consume a lot of system resources esp. memory for
large arrays. Commit 22eae8c6b29d ("Use late-binding to
finalize snprintf/sscanf") avoids unnecessary code
generation for snprintf/sscanf until they are called.
Even with this commit, however, the overhead can still
be significant for large arrays.

For example, the following large array,
  #define TCP6_RXMIT_BUCKET_BITS 18
  struct tcp6_rxmit_tbl {
    __u64 buckets[1 << TCP6_RXMIT_BUCKET_BITS];
  };
  BPF_ARRAY(rxmit_marking_map, struct tcp6_rxmit_tbl, 1);
is used inside Facebook. The reason to use such a large array
is for performance.

If it is added to examples/cpp/HelloWorld.cpp,
the HelloWorld RSS memory will increase from 7MB to
370MB.

This patch add a BPFModule API to disable rw_engine sscanf/snprintf
function. This is only for advanced user which directly interects
with BPFModule. Most C++ API users may not use sscanf/snprintf
functionalities and if there is a need, we can add a corresponding
C++ API.

### DIFF
--- a/src/cc/bpf_module.cc
+++ b/src/cc/bpf_module.cc
@@ -99,8 +99,9 @@ class MyMemoryManager : public SectionMemoryManager {
   map<string, tuple<uint8_t *, uintptr_t>> *sections_;
 };
 
-BPFModule::BPFModule(unsigned flags, TableStorage *ts)
+BPFModule::BPFModule(unsigned flags, TableStorage *ts, bool rw_engine_enabled)
     : flags_(flags),
+      rw_engine_enabled_(rw_engine_enabled),
       used_b_loader_(false),
       ctx_(new LLVMContext),
       id_(std::to_string((uintptr_t)this)),
@@ -530,6 +531,8 @@ int BPFModule::annotate() {
 }
 
 StatusTuple BPFModule::sscanf(string fn_name, const char *str, void *val) {
+  if (!rw_engine_enabled_)
+    return StatusTuple(-1, "rw_engine not enabled");
   auto fn =
       (int (*)(const char *, void *))rw_engine_->getFunctionAddress(fn_name);
   if (!fn)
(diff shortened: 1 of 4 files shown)


src/cc/bpf_module.cc
```diff
@@ -101,6 +101,7 @@ class MyMemoryManager : public SectionMemoryManager {
 
 BPFModule::BPFModule(unsigned flags, TableStorage *ts)
     : flags_(flags),
+      rw_engine_enabled_(true),
```
palmtenor: Can we do this via `flags`? like `rw_engine_enabled_(!(flags & BPF_DISABLE_RW_ENGINE))`? This way it would be easy to control the behavior through Python / C++ API (who already has `flags` as constructor argument of the `BPF` class) without need to change the interface?

palmtenor: Can we do this via flags? like `rw_engine_enabled_(!(flags & BPF_DISABLE_RW_ENGINE))`? This way it would be easy to control the behavior through Python / C++ API (who already has flags as constructor argument of the `BPF` class) without need to change the interface, or make extra call to enable / disable is later?

yonghong-song: The flags is used for debugging purpose. We probably do not want to mix it with other non-debugging thing.

Also, we probably do not want to expose this rw_engine_disabled thing to python API, it is just confusing to python users and unnecessary exposing internal details.

src/cc/bpf_module.h
```diff
@@ -117,9 +117,11 @@ class BPFModule {
   char * license() const;
   unsigned kern_version() const;
   TableStorage &table_storage() { return *ts_; }
+  void disable_rw_engine() { rw_engine_enabled_ = false; }
```
palmtenor: This is a bit tricky. For Python (even if we expose it), the loading is already done at construction time and if user call this `disable_rw_engine` later there would be no effect / saving on memory. For C++, it must be called before `BPF::init()` otherwise same thing. And this gives people the impression that it could be turned on and off on-demand, rather than a one-off switch at load time. How about we just keep it at the constructor?

yonghong-song: Adding to constructor making sense. Will add C++ API as well.

PR #1603, opened 2018-02-23
fix percpu table support in c++ api

mauriciovasquezbernal:
Current table C++ API doesn't take into consideration the special structure of percpu maps. Hence all operations performed on those maps have only effect on CPU 0.

### DIFF
--- a/src/cc/api/BPFTable.h
+++ b/src/cc/api/BPFTable.h
@@ -29,6 +29,7 @@
 #include "bcc_exception.h"
 #include "bcc_syms.h"
 #include "bpf_module.h"
+#include "common.h"
 #include "libbpf.h"
 #include "perf_reader.h"
 #include "table_desc.h"
@@ -67,28 +68,24 @@ class BPFTableBase {
  protected:
   explicit BPFTableBase(const TableDesc& desc) : desc(desc) {}
 
-  bool lookup(KeyType* key, ValueType* value) {
-    return bpf_lookup_elem(desc.fd, static_cast<void*>(key),
-                           static_cast<void*>(value)) >= 0;
+  bool lookup(void* key, void* value) {
+    return bpf_lookup_elem(desc.fd, key, value) >= 0;
   }
 
-  bool first(KeyType* key) {
-    return bpf_get_first_key(desc.fd, static_cast<void*>(key),
-                             desc.key_size) >= 0;
+  bool first(void* key) {
+    return bpf_get_first_key(desc.fd, key, desc.key_size) >= 0;
   }
 
-  bool next(KeyType* key, KeyType* next_key) {
-    return bpf_get_next_key(desc.fd, static_cast<void*>(key),
-                            static_cast<void*>(next_key)) >= 0;
+  bool next(void* key, void* next_key) {
+    return bpf_get_next_key(desc.fd, key, next_key) >= 0;
   }
 
-  bool update(KeyType* key, ValueType* value) {
-    return bpf_update_elem(desc.fd, static_cast<void*>(key),
-                           static_cast<void*>(value), 0) >= 0;
+  bool update(void* key, void* value) {
+    return bpf_update_elem(desc.fd, key, value, 0) >= 0;
   }
 
-  bool remove(KeyType* key) {
-    return bpf_delete_elem(desc.fd, static_cast<void*>(key)) >= 0;
+  bool remove(void* key) {
+    return bpf_delete_elem(desc.fd, key) >= 0;
   }
 
   const TableDesc& desc;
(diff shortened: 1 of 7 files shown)


yonghong-song:
[buildbot, ok to test]

palmtenor:
LGTM overall.

General question: Could the per-CPU tables just use the "normal" tables templated on `vector<ValueType>` so that we could reduce duplicated code? Maybe I missed something:)

mauriciovasquezbernal:
Unfortunately it won't work. The `BPFTable` performs a casting and pass that value to the syscall, if `vector<ValueType>` is used, the address of the vector object will be passed instead of the elements.

palmtenor:
Ah, OK, makes sense. So... can we also pass an adapter (as template argument) that is normally just taking address (`&`) but for per-CPU we would use `.data()` instead?
It is indeed a bit more complicated... what do you think? I'm just concerned about the (mostly) duplicated code that if in the future we want to change things about tables, we would have so many call sites to change would easily miss some...

tests/cc/test_bpf_table.cc
```diff
@@ -67,6 +67,37 @@ TEST_CASE("test bpf table", "[bpf_table]") {
   REQUIRE(res.code() != 0);
 }
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 6, 0)
+TEST_CASE("test bpf percpu tables", "[bpf_percpu_table]") {
+  const std::string BPF_PROGRAM = R"(
+    BPF_TABLE("percpu_hash", int, u64, myhash, 128);
+  )";
+
+  ebpf::BPF *bpf(new ebpf::BPF);
```
yonghong-song: Do you want to use `ebpf::BPF bpf` instead of the above `new` operator? This way, you won't need to call `delete`.

mauriciovasquezbernal:
I pushed a possible solution.
I had to modify the signatures of `BPFTableBase` to receive `void *`, otherwise I'd have had to add an extra template parameter to avoid type conversion errors. However this methods are for internal use, it should not be a problem. There are some issues with `const void *` and `void *` usage all around, I plan to fix those in a following PR.

I specialized the `get_value` and `update_value` in the percpu implementations in order to resize and check the vector size, I feel this is important.

Do you know of a more elegant way to declare the adapters? As far as I know lambdas cannot be used as template parameters in C++11.

yonghong-song:
@palmtenor Any further comment on this?

src/cc/api/BPFTable.h
```diff
@@ -99,12 +96,25 @@ class BPFTable : public BPFTableBase<void, void> {
...
+  StatusTuple update_value(const std::string& key_str,
+                           const std::vector<std::string>& value_str);
+
   StatusTuple remove_value(const std::string& key_str);
 };
 
 template <class ValueType>
+void * get_element_addr(ValueType& t) { return &t; }
+
+template <class ValueType>
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Thanks for the recommendation, now code looks a lot cleaner.

src/cc/api/BPFTable.h
```diff
@@ -143,7 +153,33 @@ class BPFArrayTable : public BPFTableBase<int, ValueType> {
...
+      : BPFArrayTable<std::vector<ValueType>, get_vector_addr>(desc) {
+    if (desc.type != BPF_MAP_TYPE_PERCPU_ARRAY)
+      throw std::invalid_argument("Table '" + desc.name + "' is not a percpu array table");
+    // leaf structures have to be aligned to 8 bytes as hardcoded in the linux kernel.
+    if (sizeof(ValueType) % 8)
+      throw std::invalid_argument("leaf must be aligned to 8 bytes");
+  }
+
+  StatusTuple get_value(const int& index, std::vector<ValueType>& value) {
+    size_t ncpus = get_possible_cpus().size();
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I don't know exactly what could be the performance penalty for it, for sure it is faster to save this value in a class member. Already done in last pushed version.

src/cc/api/BPFTable.h
```diff
@@ -114,14 +123,14 @@ class BPFArrayTable : public BPFTableBase<int, ValueType> {
       throw std::invalid_argument("Table '" + desc.name + "' is not an array table");
   }
 
-  StatusTuple get_value(const int& index, ValueType& value) {
-    if (!this->lookup(const_cast<int*>(&index), &value))
+  virtual StatusTuple get_value(const int& index, ValueType& value) {
```
palmtenor: @mauriciovasquezbernal Sorry I missed this during review, but why do we need `virtual` for these methods? We will be having the table instances in correct class types (like `BPFPercpuArrayTable`) and we don't need virtual methods? Did I miss something?

mauriciovasquezbernal: Those methods are called from the `operator[]` and `get_table_offline` methods on the base classes (`BPFArrayTable` and `BPFHashTable`)

For instance: https://github.com/iovisor/bcc/blob/master/src/cc/api/BPFTable.h#L140

palmtenor: Makes sense. Thanks for the explanation!

Issue #1601, opened 2018-02-22
Can we remove perf reader for TRACEPOINT events?


palmtenor:
If I understand correctly, the `perf_reader` in TRACEPOINT (kprobe / uprobe / Kernel Tracepoint) events logic is mostly in order to get the Callchain / IP for these events. However we already have the in-Kernel Stack-trace table and it was added quite early in BPF's development process. A quick check shows all our current BCC tools are using the in-Kernel Stack-trace table, and I would imagine most other users are probably also doing that as well.

Would it be possible for us to remove the `PERF_SAMPLE_RAW` / `PERF_SAMPLE_CALLCHAIN` / `perf_reader` logic from BCC's kprobe / uprobe / Kernel Tracepoint API? This would make the code simpler and more efficient, reduce possible issues like #1533 and make fixes like #1595 easier.

cc @4ast @drzaeus77 @yonghong-song

4ast:
makes sense to me

yonghong-song:
Indeed, looks like 
```
class BPF(object):
...
    def __init__(self, src_file=b"", hdr_file=b"", text=None, cb=None, debug=0,
            cflags=[], usdt_contexts=[]):
```

No tools really uses `cb`, so I think it is safe to remove it.

But the issue with #1533 and #1595 could be different, they are from RAW output due to bpf_perf_event_output, so we may have to address them somehow.

PR #1607, opened 2018-02-27
Remove PERF_SAMPLE_CALLCHAIN logic from TRACEPOINT events

palmtenor:
As discussed in #1601, since we had the stack-trace table early-on, and no tools still uses the old `PERF_SAMPLE_CALLCHAIN`, this PR removes the creation of `perf_reader` for `TRACEPOINT` events (kprobe / uprobe / Kernel Tracepoint), and parsing logic in `perf_reader`.

Also added some comments in `libbpf.c`


src/cc/libbpf.c
```diff
@@ -762,47 +766,38 @@ static int bpf_attach_tracing_event(int progfd, const char *event_path,
...
       fprintf(stderr, "perf_event_open(%s/id): %s\n", event_path, strerror(errno));
       return -1;
     }
   }
-  perf_reader_set_fd(reader, pfd);
 
-  if (perf_reader_mmap(reader, attr.type, attr.sample_type) < 0)
-    return -1;
-
-  if (ioctl(pfd, PERF_EVENT_IOC_SET_BPF, progfd) < 0) {
```
palmtenor: Previously, if the `pfd` was created here (falling back to `debugfs`) and the`ioctl` fails, no would would clean them up. This commit fixes the issue and documented that caller should do the clean-up.

src/cc/libbpf.c
```diff
@@ -823,15 +817,16 @@ void *bpf_attach_kprobe(int progfd, enum bpf_probe_attach_type attach_type,
...
+  // Otherwise it will read the event ID from the path in buf, create the
+  // Perf Event event using that ID, and updated value of pfd.
+  if (bpf_attach_tracing_event(progfd, buf, -1 /* PID */, &pfd) == 0)
+    return pfd;
 
 error:
-  perf_reader_free(reader);
-  return NULL;
-
+  bpf_close_perf_event_fd(pfd);
```
yonghong-song: Do you need to check whether pfd >= 0 here? Is it possible pfd still equals to initial value -1?
The same for below a few places.

src/lua/bcc/bpf.lua
```diff
@@ -55,6 +56,12 @@ function Bpf.static.cleanup()
 
   detach_all("kprobes", Bpf.static.open_kprobes)
   detach_all("uprobes", Bpf.static.open_uprobes)
+
+  for key, perf_buffer in pairs(Bpf.static.perf_buffers) do
+    libbcc.perf_reader_free(perf_buffer)
+    Bpf.static.perf_buffers[key] = nil
+  end
```
yonghong-song: Not a lua expert. Maybe after all perf_reader_free, just clear Bpf.static.perf_buffers?

src/cc/libbpf.c
palmtenor: `bpf_close_perf_event_fd` already checks the FD inside:)

src/lua/bcc/bpf.lua
palmtenor: Yeah, me neither:( I basically mimicked existing code, it compiled and tests pass, but that's all I know......

Issue #1586, opened 2018-02-10
cachestat is confusing regarding write misses.


joelagnel:
In cachestat, we have some code like so:
```
        misses = (apcl + apd)

        # rtaccess is the read hit % during the sample period.
        # wtaccess is the write hit % during the smaple period.
        if mpa > 0:
            rtaccess = float(mpa) / (access + misses)
        if apcl > 0:
            wtaccess = float(apcl) / (access + misses)
```
As shown, `apcl` is a part of the misses equation, however later it is also used to calculate "write hits" in `wtaccess`. Does `apcl` constitute hits or misses?

Also I tried a simple experiment, I dd'd a file 500MB file, and read it back. Then I set /proc/sys/kernel/drop_caches and read it back again. What I see is Read hit % is around 50%. I expected that to be 0%. What's the expectation with regards to the counters when one sets `drop_caches` and reads a dd'd file back? Thanks.

joelagnel:
@brendangregg 's original cache stat script has this which makes a bit more sense to me:
```
		mpa = a["mark_page_accessed"]
		mbd = a["mark_buffer_dirty"]
		apcl = a["add_to_page_cache_lru"]
		apd = a["account_page_dirtied"]
		total = mpa - mbd
		misses = apcl - apd
		if (misses < 0)
			misses = 0
		hits = total - misses
```
@brendangregg any thoughts? :)

PR #1609, opened 2018-02-28
tools/cachestat: Rewrite it so it makes more sense

joelagnel:
cachstat separation between read and writes doesn't make much sense. During tests
its seen that a workload that is supposed to be completely in the missed category
appears to be 50% hit.

This patch rewrites the BCC tool to be more inline with Brendan Gregg's original
cachestat perf tool which makes more sense:
https://github.com/brendangregg/perf-tools/blob/master/fs/cachestat
http://www.brendangregg.com/blog/2014-12-31/linux-page-cache-hit-ratio.html

Fixes Issue: https://github.com/iovisor/bcc/issues/1586

### DIFF
--- a/tools/cachestat.py
+++ b/tools/cachestat.py
@@ -39,16 +39,13 @@ def get_meminfo():
     return result
 
 # set global variables
-rtaccess = 0
-wtaccess = 0
 mpa = 0
 mbd = 0
 apcl = 0
 apd = 0
-access = 0
+total = 0
 misses = 0
-rhits = 0
-whits = 0
+hits = 0
 debug = 0
 
 # args
@@ -116,9 +113,8 @@ b.attach_kprobe(event="mark_buffer_dirty", fn_name="do_count")
 # header
 if tstamp:
     print("%-8s " % "TIME", end="")
-print("%8s %8s %8s %10s %10s %12s %10s" %
-     ("HITS", "MISSES", "DIRTIES",
-     "READ_HIT%", "WRITE_HIT%", "BUFFERS_MB", "CACHED_MB"))
+print("%8s %8s %8s %8s %12s %10s" %
+     ("TOTAL", "MISSES", "HITS", "DIRTIES", "BUFFERS_MB", "CACHED_MB"))
 
 loop = 0
 exiting = 0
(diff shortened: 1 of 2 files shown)


yonghong-song:
[buildbot, ok to test]

tools/cachestat.py
```diff
@@ -153,25 +149,27 @@ def usage():
         # access = total cache access incl. reads(mpa) and writes(mbd)
```
yonghong-song: change comment here?

tools/cachestat_example.txt
```diff
@@ -15,61 +15,42 @@ examples:
...
+  152032   152032        0        0            9       1028
+  157408   157405        3        0            9       1707
+  150432   150432        0        0            9       2331
+       0        0        0        0            9       2331
+       1        1        0        1            9       2331
+       0        0        0        0            9       2331
+       0        0        0        0            9       2331
+       0        0        0        0            9       2331
+
+The misses counter reflects a 2GB file being read and almost everything being
```
yonghong-song: I actually run the command with a 20MB file on latest bpf-next. With debug = 1, I got:
```
7549 0 508 700 7549 0 7549
mpa, mbd, apcl, apd, total, misses, hits
```
Here, `apcl` is less then `apd`.
Look at the kernel code, it is not clear to me what is the relationship between
`add_to_page_cache_lru` and `account_page_dirtied`. `

tools/cachestat.py
joelagnel: sure

tools/cachestat_example.txt
joelagnel: `add_to_page_cache_lru` counts pages added to the page cache which included both pages added because of read misses and pages added because of writes. IMO a "write miss" isn't meaningful since the page will be lazy written back to the disk and isn't a harmful miss, so such "adds" to the page cache should not be counted, and this is the reason why  `account_page_dirtied` is subtracted.

Here's a run of funccount counting all these, during some dd's:
```
# DD commands:
dd if=/dev/urandom of=tmp2 bs=4k count=$((256 * 20))
dd if=tmp1 of=/dev/null bs=4k
```
```
# ./tools/funccount.py -r  '(add_to_page_cache_lru|account_page_dirtied|mark_buffer_dirty|mark_page_accessed)'
Tracing 5 functions for "(add_to_page_cache_lru|account_page_dirtied|mark_buffer_dirty|mark_page_accessed)"... Hit Ctrl-C to end.
^C
account_page_dirtied                     5120
mark_buffer_dirty                        5120
add_to_page_cache_lru                   13153
mark_page_accessed                      16103
```

apcl - abd = ~8k pages which is the 30MB file that was added to the page cache as a part of the "read" dd and these are the hurtful misses.

Unfortunately, I think its possible a page already exists in the page cache and is being dirtied, so these `account_page_dirtied` are the extra ones which don't have a corresponding `add_to_page_cache_lru`. 

Another thing to note is Brendan's script also assumed `apcl` can be < `apd`: https://github.com/brendangregg/perf-tools/blob/master/fs/cachestat#L150

joelagnel: Just to clarify, the `tmp1` file above is the 30MB I had dd'd in advance.

yonghong-song: Do you think it makes sense to have finer control on add_to_page_cache_lru/account_page_dirtied? For example, have a map shared between these two functions and account_page_dirtied event is counted if only page has been added to lru?

joelagnel: Yeah I think so, but I think it will also complicate the script more. I am wondering how often this negative value happens for you? I believe the script is expected to provide "rough estimate" than accurate results. And also that, in the future the meanings and semantics can change anyway so I am wondering if its worth handling this case (unless it happens way too often)

brendangregg:
Thanks @joelagnel , I tested on 4.4 and 4.9 and it looks like it works (and the old tool overcounted misses). This is a pretty important tool: there's no other way I found to get file system cache statistics.

When I wrote this with ftrace function counting, I had to figure out cache activity from function counts alone (not easy to figure out). Maybe one day we'll revisit this and make it better, now we can access arguments and return values, or maybe it's fine as is. ... Anyone working on this area does needs to be careful and take their time to investigate different workloads: the mm code is tricky.

Later on we might want to add a percent column back, eg, a "HIT%" = HITS * 100 / (HITS + MISSES). Just to make reading the output a bit easier.

joelagnel:
Thanks a lot Brendan. I will address the one comment related to code comment change from  Yonghong then and repost.

yonghong-song:
I hacked to add a map to communicate between `add_to_page_cache_lru` and `account_page_dirtied`.  `account_page_dirtied` only counts if the `page` has
been recorded by `add_to_page_lru`. Now they are the same. This seems implying
that every page added to page cache lru is `account_page_dirtied`, and yes
there are some account_page_dirtied pages not added by page cache lru.

My file system is `btrfs`. Not sure this is the cause or not.

Anyway, this patch is fixing the existing issues. I can investigate the accounting issue separately.

PR #1616, opened 2018-03-01
Rename kprobe_poll to perf_buffer_poll

palmtenor:
After #1607, it makes sense for the interface to be renamed to `perf_buffer_poll`, since it would be more informative for users that the perf buffer is not only for kprobes, but for all other kinds of events.

Still kept the `kprobe_poll` interface in Python and Lua to avoid breaking existing tools


yonghong-song:
[buildbot, ok to test]

Issue #1571, opened 2018-02-03
Gauging interest for a tool contribution


rmanyari:
Hi,

I wrote a script some months ago that is based of Brendan Gregg's tcptop. It essentially categorizes outgoing IPv4 TCP traffic by subnets (the list of subnets is passed as an argument) and reports the aggregates at some interval.

The use case here is to figure out how much cross-zone, cross-region and cloud-to-premise traffic is generated from individual VMs. I'm aware that cloud providers have flow analysis tools but I still find benefit in having something closer to the VM. It opens doors to do more interesting things such as filtering by PID and eventually by UID.

I'd like to know if there is interest from the maintainers of BCC in having this script. I will take the time to follow the contributing scripts guidelines and make it production ready.

Script: https://github.com/rmanyari/bcc-scripts/blob/master/tcpdest.py

goldshtn:
I think it would be useful!

brendangregg:
Could you post example output with an explanation of what it is showing?
Any submission to bcc/tools needs this in _example.txt files anyway.

Brendan

rmanyari:
Here I bucket the traffic in 3 categories:

- Local network (192.168.0.0/24)
- github.com (192.30.253.0/24 - I got this from tcpdump, just an approximation for demonstration purposes)
- Everything else

```bash
root@rodrigo-XPS-13-9360:~/dev/bcc-scripts# ./tcpdest.py -S 192.168.0.0/24,192.30.253.0/24
0.0.0.0/0                 35
0.0.0.0/0               2106
192.30.253.0/24         1813
192.30.253.0/24         2093
0.0.0.0/0               1836
192.30.253.0/24        18068
0.0.0.0/0               5876
192.168.0.0/24             1
0.0.0.0/0               2242
192.168.0.0/24             8
192.168.0.0/24            12
0.0.0.0/0                 35
````

Output in JSON format:

```bash
root@rodrigo-XPS-13-9360:~/dev/bcc-scripts# ./tcpdest.py -S 192.168.0.0/24,192.30.253.0/24 -J
{}
{"192.168.0.0/24": 9.0, "192.30.253.0/24": 35.0, "0.0.0.0/0": 35.0}
{}
{"0.0.0.0/0": 6406.0}
{"0.0.0.0/0": 342.0}
{"192.30.253.0/24": 1813.0, "0.0.0.0/0": 2813.0}
{"0.0.0.0/0": 892.0}
[...]

PR #1618, opened 2018-03-03
Tool tcpsubnet

rmanyari:
PR for [1571](https://github.com/iovisor/bcc/issues/1571)

I updated the tool based on Brendan's feedback, added some validation as well.

Signed-off-by: Rodrigo Manyari rodrigo@twosigma.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ pair of .c and .py files, and some are directories of files.
 #### Tools:
 <center><a href="images/bcc_tracing_tools_2017.png"><img src="images/bcc_tracing_tools_2017.png" border=0 width=700></a></center>
 
+
 - tools/[argdist](tools/argdist.py): Display function parameter values as a histogram or frequency count. [Examples](tools/argdist_example.txt).
 - tools/[bashreadline](tools/bashreadline.py): Print entered bash commands system wide. [Examples](tools/bashreadline_example.txt).
 - tools/[biolatency](tools/biolatency.py): Summarize block device I/O latency as a histogram. [Examples](tools/biolatency_example.txt).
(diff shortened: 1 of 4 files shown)


yonghong-song:
[buildbot, ok to test]

yonghong-song:
Tested. Looks good.

cc @brendangregg

brendangregg:
Looks good. A couple of things.

I'd add 0.0.0.0/0 to the default, as a catch-all.

And there should be a newline after the output, or a timestamp line before the output (like argparse's "[18:21:00]"), because without these, we can't really understand the output. Eg:

```
# ./tcpsubnet.py
Tracing... Output every 1 secs. Hit Ctrl-C to end
127.0.0.1/32            2734
127.0.0.1/32             183
^C
```

Was that 2734 segments in the first second, and then 183 segments in the second? No, it wasn't, they were several seconds apart, but tcpsubnet prints nothing if there was no activity. It should look like this, at least:

```
# ./tcpsubnet.py
Tracing... Output every 1 secs. Hit Ctrl-C to end
127.0.0.1/32            2734





127.0.0.1/32             183

^C
```

Or:

```
# ./tcpsubnet.py
Tracing... Output every 1 secs. Hit Ctrl-C to end
[18:21:03]
127.0.0.1/32            2734
[18:21:04]
[18:21:05]
[18:21:06]
[18:21:07]
[18:21:08]
127.0.0.1/32             183
^C
```

rmanyari:
Good catch, the behavior was obvious to me but only because I wrote the tool... I made date and time keys in the JSON object, putting it as a header would defeat the point of having JSON in the first place.

I also updated the examples, man page and brought back 0.0.0.0/0 as the last default

brendangregg:
Looks good, thanks!

PR #1620, opened 2018-03-05
C++ API: Allow obtaining BPFPerfBuffer pointer for polling

palmtenor:
Perf Buffer polling is a high frequency event and normally runs in a loop for continuous profiling. Doing a map lookup every time is not efficient. This commit adds an interface to get a pointer of the `BPFPerfbuffer` instance for polling. Also added some comments on the header file to explain these APIs.

Updated  `FollyRequestContextSwitch` example to use this. The `RandomRead` example still uses `perf_buffer_poll`, just to demonstrate both interfaces.

### DIFF
--- a/examples/cpp/FollyRequestContextSwitch.cc
+++ b/examples/cpp/FollyRequestContextSwitch.cc
@@ -98,8 +98,11 @@ int main(int argc, char** argv) {
 
   signal(SIGINT, signal_handler);
   std::cout << "Started tracing, hit Ctrl-C to terminate." << std::endl;
-  while (true)
-    bpf->poll_perf_buffer("events");
+  auto perf_buffer = bpf->get_perf_buffer("events");
+  if (perf_buffer)
+    while (true)
+      // 100ms timeout
+      perf_buffer->poll(100);
 
   return 0;
 }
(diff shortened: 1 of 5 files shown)


brendangregg:
Is this part of addressing #1033?

palmtenor:
I was not aware of #1033, but this would definitely help:)

yonghong-song:
@palmtenor  Do you know how much we save for cpu resource?

palmtenor:
I think it highly depends on the sample rate and polling setup. I did a simple experiment:
- A CPU profiler that generates around ~20 samples per second.
- The Perf Buffer callback simply copy it to a queue and do no other processing.
- Use `perf` to profile the tool only during this period of time (no compilation / symbolization / etc.)

The `poll_perf_buffer` is about ~2.5% of all overhead during profiling (including Kernel PMU costs), and mostly attributed to the cost of map lookup. Here is the [`perf annotate` result](https://paste.ubuntu.com/p/JhNjM9DZ7N/). Everything is heavily inlined so hard to see, but the `jmpq` at line 80 is going into the actual `epoll` / parsing buffer logic.

So not too much a big deal, but a good small improvement.

@brendangregg I believe for Python, construction of the reader FD array every loop might be also considerably costly. In C++ we use `epoll` to avoid that. After recent clean-up of the kprobe logic from Perf Buffer polling, this should also be possible in Python.

PR #1625, opened 2018-03-09
docs: fix commit hash and link for bpf_redirect_map()

qmonnet:
The link wrongly pointed to the same commit as for
bpf_sk_redirect_map().


yonghong-song:
Looks good. Thanks.

PR #1626, opened 2018-03-09
Show negative groups correctly in killsnoop

frioux:
killsnoop currently shows giant numbers when the target pid was negative.  This takes the two's compliment before showing it.  It is hardcoded to assume 64 bit, which may be wrong, but I don't know for sure.

### DIFF
--- a/tools/killsnoop.py
+++ b/tools/killsnoop.py
@@ -133,6 +133,10 @@ def print_event(cpu, data, size):
 
     if (args.failed and (event.ret >= 0)):
         return
+    
+    tpid = event.tpid
+    if (tpid & (1 << (64 - 1 ))) != 0:
+        tpid = tpid - (1<<64)
 
     print("%-9s %-6d %-16s %-4d %-6d %d" % (strftime("%H:%M:%S"),
         event.pid, event.comm.decode(), event.sig, event.tpid, event.ret))


yonghong-song:
Do you have a use case this is really happening?
pid_t is a `int` type which is expressed as the sys_kill argument.

Based on `man proc`,
```
       /proc/sys/kernel/pid_max (since Linux 2.5.34)
              This file specifies the value at which PIDs wrap around (i.e., the value in this file is
              one greater than the maximum PID).  PIDs greater than  this  value  are  not  allocated;
              thus,  the  value  in  this file also acts as a system-wide limit on the total number of
              processes and threads.  The default value for this file,  32768,  results  in  the  same
              range  of  PIDs  as on earlier kernels.  On 32-bit platforms, 32768 is the maximum value
              for pid_max.   On  64-bit  systems,  pid_max  can  be  set  to  any  value  up  to  2^22
              (PID_MAX_LIMIT, approximately 4 million).
```
The maximum pid on 64bit system is `2^22` with which `pid` never goes to negative.
What is your system's configuration which caused negative?

brendangregg:
@yonghong-song this is just a weird shortcut that kill(1) implements: if you do a kill to a negative pid, it treats it as a group ID.

brendangregg:
... and yes, I think killsnoop should be modified to handle this correctly.

yonghong-song:
@brendangregg, that is right. Thanks for clarification. Just checked the manpage and the source code. Indeed, 
  positive: pid
  -1: every process for which the calling process has permission to send signals (except process id 1)
  non -1 negative value: process group id
  0: all processes in the process group of the calling process.

Then, if the `pid = -1`, we probably should still display `-1` so it is clear it is not a pgrp id?

frioux:
It is the pgrp and it is negative.

--

frioux:
Note that this is arguably easier to just try out than discuss😉

--

yonghong-song:
Just tried with negative numbers:
```
09:33:34  1885862 bash             0    -1     0
09:34:32  1885862 bash             0    -2157301 -3
```
It seems working fine with me.
@frioux, do you have a concrete example to show the issue?

tools/killsnoop.py
```diff
@@ -133,6 +133,10 @@ def print_event(cpu, data, size):
 
     if (args.failed and (event.ret >= 0)):
         return
+    
+    tpid = event.tpid
+    if (tpid & (1 << (64 - 1 ))) != 0:
+        tpid = tpid - (1<<64)
 
     print("%-9s %-6d %-16s %-4d %-6d %d" % (strftime("%H:%M:%S"),
         event.pid, event.comm.decode(), event.sig, event.tpid, event.ret))
```
yonghong-song: We still use event.tpid which won't change the output.

frioux:
Yeah I clearly fumbled my (in browser submitted) patch, sorry about that.  Here was the example, which I'll try to repro quickly:

```
14:01:26  1      init             15   18446744073709533932 0
```

yonghong-song:
What is you `kill` command?

frioux:
I can reproduce the above with `sudo kill 15 -13615`; the kernel is `4.4.0-111-generic #134~14.04.1-Ubuntu SMP`, Ubuntu 14.04 otherwise.

frioux:
(the giant number varies based on the group, ofc)

frioux:
Note also that I *cannot* reproduce this on my laptop, 17.10, so clearly there's some library or other version interaction.

frioux:
For what it's worth, checking for a pid greater than the max and doing the bit shifting dance *does* fix the issue on the affected host.

PR #1627, opened 2018-03-10
Perf Event interface with raw perf_event_attr argument

palmtenor:
Attaching BPF programs to Perf Events is a commonly used profiling and tracing tools. However, Kernel Perf Events system interface (i.e., `sys_perf_event_open`) has so many arguments and config combinations. We have recently experimented `pinned` events, events with different `sample_type`, events with difference `pricise_ip`, etc. and those are only small portion of all possible configs.

It is not possible for us to add all those options to either C++ API or `libbpf` interface. But those configs are definitely useful for experts and complex production systems.

Therefore, this PR adds `bpf_attach_perf_event_raw` / `attach_perf_event_raw` interface to `libbpf` and C++ API, that user would just provide a pointer to the raw `perf_event_attr` struct, and our libraries would handle rest of the attaching logic (loading program, opening and enabling, etc.)

Unfortunately on C++ API, the attached Perf Events are still uniquely keyed on `type` / `config` pair, which is not ideal. I am planning to change and improve that in separated PRs.

### DIFF
--- a/src/cc/api/BPF.cc
+++ b/src/cc/api/BPF.cc
@@ -15,6 +15,7 @@
  */
 
 #include <linux/bpf.h>
+#include <linux/perf_event.h>
 #include <unistd.h>
 #include <cstdio>
 #include <cstring>
@@ -292,12 +293,13 @@ StatusTuple BPF::attach_perf_event(uint32_t ev_type, uint32_t ev_config,
   int probe_fd;
   TRY2(load_func(probe_func, BPF_PROG_TYPE_PERF_EVENT, probe_fd));
 
-  auto fds = new std::map<int, int>();
   std::vector<int> cpus;
   if (cpu >= 0)
     cpus.push_back(cpu);
   else
     cpus = get_online_cpus();
+  auto fds = new std::vector<std::pair<int, int>>();
+  fds->reserve(cpus.size());
   for (int i : cpus) {
     int fd = bpf_attach_perf_event(probe_fd, ev_type, ev_config, sample_period,
                                    sample_freq, pid, i, group_fd);
@@ -309,7 +311,7 @@ StatusTuple BPF::attach_perf_event(uint32_t ev_type, uint32_t ev_config,
       return StatusTuple(-1, "Failed to attach perf event type %d config %d",
                          ev_type, ev_config);
     }
-    fds->emplace(i, fd);
+    fds->emplace_back(i, fd);
   }
 
   open_probe_t p = {};
@@ -317,6 +319,46 @@ StatusTuple BPF::attach_perf_event(uint32_t ev_type, uint32_t ev_config,
   p.per_cpu_fd = fds;
   perf_events_[ev_pair] = std::move(p);
   return StatusTuple(0);
+}
+
+StatusTuple BPF::attach_perf_event_raw(void* perf_event_attr,
+                                       const std::string& probe_func,
+                                       pid_t pid, int cpu, int group_fd) {
+  auto attr = static_cast<struct perf_event_attr*>(perf_event_attr);
+  auto ev_pair = std::make_pair(attr->type, attr->config);
+  if (perf_events_.find(ev_pair) != perf_events_.end())
+    return StatusTuple(-1, "Perf event type %d config %d already attached",
+                       attr->type, attr->config);
+
+  int probe_fd;
+  TRY2(load_func(probe_func, BPF_PROG_TYPE_PERF_EVENT, probe_fd));
+
+  std::vector<int> cpus;
+  if (cpu >= 0)
+    cpus.push_back(cpu);
+  else
+    cpus = get_online_cpus();
+  auto fds = new std::vector<std::pair<int, int>>();
+  fds->reserve(cpus.size());
+  for (int i : cpus) {
+    int fd = bpf_attach_perf_event_raw(probe_fd, attr, pid, i, group_fd);
+    if (fd < 0) {
+      for (const auto& it : *fds)
+        close(it.second);
+      delete fds;
+      TRY2(unload_func(probe_func));
+      return StatusTuple(-1, "Failed to attach perf event type %d config %d",
+                         attr->type, attr->config);
+    }
+    fds->emplace_back(i, fd);
+  }
+
+  open_probe_t p = {};
+  p.func = probe_func;
+  p.per_cpu_fd = fds;
+  perf_events_[ev_pair] = std::move(p);
+  return StatusTuple(0);
+
 }
 
 StatusTuple BPF::detach_kprobe(const std::string& kernel_func,
@@ -392,6 +434,11 @@ StatusTuple BPF::detach_perf_event(uint32_t ev_type, uint32_t ev_config) {
   TRY2(detach_perf_event_all_cpu(it->second));
   perf_events_.erase(it);
   return StatusTuple(0);
+}
+
+StatusTuple BPF::detach_perf_event_raw(void* perf_event_attr) {
+  auto attr = static_cast<struct perf_event_attr*>(perf_event_attr);
+  return detach_perf_event(attr->type, attr->config);
 }
 
 StatusTuple BPF::open_perf_event(const std::string& name, uint32_t type,
--- a/src/cc/api/BPF.h
+++ b/src/cc/api/BPF.h
@@ -36,7 +36,7 @@ namespace ebpf {
 struct open_probe_t {
   int perf_event_fd;
   std::string func;
-  std::map<int, int>* per_cpu_fd;
+  std::vector<std::pair<int, int>>* per_cpu_fd;
 };
 
 class USDT;
@@ -84,7 +84,12 @@ class BPF {
                                 uint64_t sample_period, uint64_t sample_freq,
                                 pid_t pid = -1, int cpu = -1,
                                 int group_fd = -1);
+  StatusTuple attach_perf_event_raw(void* perf_event_attr,
+                                    const std::string& probe_func,
+                                    pid_t pid = -1, int cpu = -1,
+                                    int group_fd = -1);
   StatusTuple detach_perf_event(uint32_t ev_type, uint32_t ev_config);
+  StatusTuple detach_perf_event_raw(void* perf_event_attr);
 
   BPFTable get_table(const std::string& name) {
     TableStorage::iterator it;
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -1278,6 +1278,28 @@ cleanup:
     return ret;
 }
 
+int bpf_attach_perf_event_raw(int progfd, void *perf_event_attr, pid_t pid,
+                              int cpu, int group_fd) {
+  int fd = syscall(__NR_perf_event_open, perf_event_attr, pid, cpu, group_fd,
+                   PERF_FLAG_FD_CLOEXEC);
+  if (fd < 0) {
+    perror("perf_event_open failed");
+    return -1;
+  }
+  if (ioctl(fd, PERF_EVENT_IOC_SET_BPF, progfd) != 0) {
+    perror("ioctl(PERF_EVENT_IOC_SET_BPF) failed");
+    close(fd);
+    return -1;
+  }
+  if (ioctl(fd, PERF_EVENT_IOC_ENABLE, 0) != 0) {
+    perror("ioctl(PERF_EVENT_IOC_ENABLE) failed");
+    close(fd);
+    return -1;
+  }
+
+  return fd;
+}
+
 int bpf_attach_perf_event(int progfd, uint32_t ev_type, uint32_t ev_config,
                           uint64_t sample_period, uint64_t sample_freq,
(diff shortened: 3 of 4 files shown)


yonghong-song:
Agree with the rationale to add this new API to take raw config to cover user arbitrary use cases.
Could you add a test for this new API? Optionally, you can add an example for this new API if it represents real use case and it is quite different from the test.

Thanks!

palmtenor:
I actually plan to do a header update + test case once https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git/commit/?id=95da0cdb723260362fc126a563285ac66a193da7 hits upstream, which would use the raw perf event interface as well. How does that sound?

yonghong-song:
You can start to implement a test/example once it hits net-next. Thanks!

drandynisbet:
Would it be possible to read the LBR on supprted x86's such as Skylake, using the bpf_attach_perf_event_raw,  or is it still a deal breaker in that we need access to the rdmsr instructions?

yonghong-song:
`bpf_attach_perf_event_raw` does not read lbr. The current kernel stack walking does not support lbr yet.

drandynisbet:
Hi,
sorry I was not clear, I'm not seeking to do full stack/capture walking at the moment. I'm interested in purely reading the LBR record, which on SkyLake is 32 entries long and storing this in a HASHMAP using an appropriate struct. I was wondering if the raw interface and then some use of bpf_perf_event_read_value or similar might be feasible. I'm interested in trying to develop prototype functionality for this supported by BPF. I haven't scoped what is required in terms or implementing additional kernel helpers versus BPF library code, but I guess it would be feasible to try this out in a VM development environment with modified kernels etc.

yonghong-song:
Did not check. bpf_perf_event_read_value gets the value from perf core. I guess its value could be from lbr if the lbr sampling method is used. I could be wrong though.

drandynisbet:
Hi,
I've got a few questions on how to use/test bpf_perf_event_read_value ...
1. How do I create an appropriate map type for the first argument of bpf_perf_event_read_value?  I cannot create a simple BPF_PERF_ARRAY as I need 16/32 LBR entries - see struct sampleLBR 

2 The structure for the LBR information with 32 entries is over 512bytes long - so it cannot be passed on the stack - how can I get round this limitation. 

3. Should I be using bpf_perf_prog_read_value instead?

Note: I am trying to create a BPF test example that reads LBR based on tests/cc/test_perf_event.cc and a standalone C++ program that uses perf_event_open to sample the LBR performance counter information. See below for the skeleton outline of the BPF code. Ive omitted the C++ that does the attach_perf_event_raw ... 
 
Using bpf_perf_prog_read_value the BPF stack limit of 512 bytes is exceeded. Reducing the LBR entries buffer from 32 to 16 got round this problem, but the call to bpf_prog_read_value returns the error code of -22 - potentially related to an incorrect map type.

Using bpf_perf_event_read_value ... I get various errors from the JIT compiler ... typically related to loads dependent on what type of MAP ~(HASH, ARRAY etc) is supplied as an argument to bpf_perf_event_read_value ....
  
e.g ...
BPF_LD_[ABS|IND] instructions not allowed for this program type
and 
65: BUG_ld_88
invalid BPF_LD mode



TEST_CASE("test attach perf event RAW", "[bpf_perf_event]") {
#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 9, 0)
  const std::string BPF_PROGRAM = R"( 
#include <linux/perf_event.h>
    BPF_HASH(pid, int, u64, 1);
    BPF_HASH(ret, int, int, 1);
    struct sampleLBR
        {
            struct perf_event_header header;
            size_t size;
            struct perf_branch_entry values[16]; // 32 entries in reality for Skylake
        };
    // tried the combinations below ...
    BPF_HASH(counter, int, struct sampleLBR, 4);
    //BPF_PERCPU_ARRAY(counter,  struct sampleLBR, 4);
    //BPF_ARRAY(counter, struct sampleLBR, 4);


  
    int on_event(void *ctx) { 
      int zero = 0;
      struct sampleLBR d = {0};
      
      u64 p = bpf_get_current_pid_tgid();
      pid.update(&zero, &p);
    #if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 15, 0)
/**** TRIED the combinations below ...  what type should counter be 
*****/
      //int r = bpf_perf_prog_read_value(ctx, &d, sizeof(d));
      int r = bpf_perf_event_read_value(&counter, BPF_F_CURRENT_CPU, &d, sizeof(d));
      ret.update(&zero, &r); // use this to get the 
      counter.update(&zero, &d);
    #endif
      return 0;
    }
  )";

yonghong-song:
Please look at the example 
https://github.com/torvalds/linux/blob/master/samples/bpf/tracex6_kern.c
in kernel/samples/bpf.

`bpf_perf_event_read_value` is used to read the counter periodically by bpf program. Its counter value is associated with a particular perf event. This only applies to a single 64bit counter as configured by perf event.

One usage of lbr is to do stack unwinding if the application is compiled without framepointer enabled. This needs kernel support though.

drandynisbet:
Many thanks, but do you mean that reading LBR is not possible using bpf_perf_event_read_value as LBR is not a single 64bit counter?  In my example, (as configured), the intention is that an interrupt calls the BPF program that will read the LBR. 

Also, I must still be doing something stupid .... still cannot define the appropriate map type as I get the issue below during JIT/verification - the type of the struct such as sampleLBR in the values2 map is irrelevant. 

/virtual/main.c:23:1: error: unsupported map type: maps
struct bpf_map_def SEC("maps") values2 =  {
^
1 error generated.
 From the BPF c ....

struct sampleLBR
        {
            struct perf_event_header header;
            size_t size;
            struct perf_branch_entry values[32];
        };

struct bpf_map_def SEC("maps") values2 = {
        .type = BPF_MAP_TYPE_HASH,
        .key_size = sizeof(int),
        .value_size = sizeof(struct sampleLBR),
        .max_entries = 64,
};

palmtenor:
As @yonghong-song said, `bpf_perf_event_read_value` is to read counter values, like `perf stat`. Recording LBR (or any other things you'd expect to get in the `PERF_RECORD_SAMPLE` struct) requires Kernel copying them into the actual `bpf_perf_event_data` context. We currently only `sample_period` and `addr`. There were indeed talks to add LBR recording as well. If you can submit a Kernel patch it would be great (like https://github.com/torvalds/linux/commit/95da0cdb723260362fc126a563285ac66a193da7), but as you said the LBR records are kind of large and I'm not sure if it's going to create issues with verifier.

drandynisbet:
@palmtenor  and @yonghong-song are there any pointers to relevant information such as BPF design overview, email threads (BPF interactions with the kernel perspectives), plus any information on interactions with perf/PMUs that you might be able to direct me to that are relevant to improving access to PMUs for BPF? Hopefully it is not just a case of UTSL?  

This would very much need to be a background out of work time task for me.  From my perspective there is a need to enable bpf.attach_perf_event_raw and perf_event_read_value to support accessing arbitrary PMU counter information.  
Cheers,
Andy

yonghong-song:
cilium is great source of bpf documentation although it slightly tilts to networking.
https://cilium.readthedocs.io/en/v1.3/bpf/

The documentation of bpf on tracing side has been scarce. The best probably is checking
git log for kernel file `git log kernel/trace/bpf_trace.c` which should give most information
about bpf supported related to tracing.

drandynisbet:
Are there any known issues with perf_read on a BPF_PERF_ARRAY? I sometimes (with reasonably high probability) get identical values for CPU_CYCLES inside a RAW_TRACEPOINT_PROBE in sched_switch ...  I have configured 6 performance counters using open_perf_event (on a kaby lake with hyper threading turned off - 4 cores), kernel  5.0.0-37-generic 

Pseudo code.... 
RAW_TRACEPOINT_PROBE(sched_switch) {
read counters into currentVALUES
if(prevPIDstate == TASK_RUNNING) {
      if(prevPID == myPID) {
             get/initialise/update struct from a MAP on a per thread basis for myPID, increment counter values  associated with lastVALUES  and currentVALUES
            Added debugging to count how many times the CPU_CYCLES  (currentVALUES -lastVALUES) was zero ...
            if 100 quanta have passed do a perf submit
      }
}
lastVALUES =  counter values read into currentVALUES
}
Have I misunderstood something basic regarding sched_switch, or the rate at which counters from the PMU can be read/are updated ... the probability of  logging 1 or more zero CPU_CYCLE changes across 100 quanta is approximately 0.6 per thread ....  I'm running a java workload ... not that it matters.

drandynisbet:
BPF_PERF_ARRAY(cpucounters, NUM_CPUS);

And if I try in  my sched_switch attacked tracepoint ... 
struct bpf_perf_event_value cyclebuf;
error = bpf_perf_event_read_value((void *) &cpucounters,cpu,&cyclebuf,sizeof(cyclebuf));

I get unknown opcode ... unknown opcode a0 .... obviously I have done something wrong or missed nuances of the linux kernel example ... I configured the counters using open_perf_event ...

yonghong-song:
@drandynisbet I suggest you put a reproducible code here so people can help. It is hard to check whether this is due to program logic or it may be just what it should be.

drandynisbet:
@yonghong-song  Apologies, ... I've now double checked with perf and the counters are indeed multiplexed for that specific configuration of events, even if the watchdog timer NMI is disabled - so it is entirely reasonable to get zero cpu cycle counter changes between timeslices as the kernel multiplexes the events. I will produce a python/C file for the bpf_perf_event_read_value  issue shortly

drandynisbet:
bpf_perf_event_read_value issue - apologies in advance for the hacky code, and also if I've done something dumb ...  run as sudo ./examplebpf.txt -p PID ... 
Failed to load program: Invalid argument
unknown opcode d8

HINT: The 'unknown opcode' can happen if you reference a global or static variable, or data in read-only section. For example, 'char *p = "hello"' will result in p referencing a read-only section, and 'char p[] = "hello"' will have "hello" stored on the stack.

[examplebpf.txt](https://github.com/iovisor/bcc/files/3939819/examplebpf.txt)

yonghong-song:
The linux kernel bpf helper bpf_perf_event_read_value has the following signature:
```
int bpf_perf_event_read_value(struct bpf_map *map, u64 flags, struct bpf_perf_event_value *buf, u32 buf_size)
```
The first parameter is a `map` pointer. For kernel libbpf, the loader will replaces it to a map_fd.
In bcc, bcc will replace with first pseudo_map_fd and then a real map_fd (see b_frontend_action.cc and bpf_module.cc for details).

Since bcc does not do relocation, you will hit the issue like
```
Failed to load program: Invalid argument
unknown opcode d8
```

To fix the issue, we need to introduce a bcc intrinsic function like
```
map.perf_read_value(flag, buf, buf_size)
```
will internally be transformed to
```
bpf_perf_event_read_value(map_pseudo_id, flag, buf, buf_size).
```
The implementation should be done in b_frontend_actions.cc, you can look at other bcc intrinsics for examples.

Do you want to help make a contribution for this?

drandynisbet:
@yonghong-song I will take a look first and try to follow how it works (next week), and then try to ask some sensible questions If I get stuck. I was also planning on adding some documentation and examples for configuring and accessing performance counters. It will help me as I do plan to look at reading LBR.

yonghong-song:
@drandynisbet  sounds good! Thank you.

drandynisbet:
Sorry I have not made progress with LBR - but one of our students has an issue the following code to access  a perf counter - unc_cbo_cache_lookup.read_i 

cnt1.open_perf_event(11,0x1834)

We obtain the configuration values  of 11, 0x1834 using 
sudo perf stat -vv -e unc_cbo_cache_lookup.read_i /bin/ls

The perf stat command works fine, but for the BPF - it prints an perf_event_open: Invalid argument, and all reads of the counter are -2 (I guess it's not being configured correctly).

Have we missed something obvious, concerning the uncore counter, my understanding was that any simple 64 bit counters should work in this way?  

This is  on    Ubuntu 20.04 LTS,5.4.0-33-generic, built from source release of  bcc release v0.14.0
Again - apologies in advance if we've missed something obvious.

Andy

drandynisbet:
Ill take a look at strace of perf stat and the BCC python and see what's up with the arguments etc ..

drandynisbet:
The failing BCC results in a call to perf_event_open({type=0xb /* PERF_TYPE_??? */, size=0 /* PERF_ATTR_SIZE_??? */, config=0x1834, ...}, -1, 0, -1, PERF_FLAG_FD_CLOEXEC) = -1 EINVAL (Invalid argument)

Whereas the perf version (stat) sys_perf_event_open: pid -1  cpu 0  group_fd -1  flags 0x8perf_event_open({type=0xb /* PERF_TYPE_??? */, size=:q, config=0x1834, ...}, -1, 0, -1, PERF_FLAG_FD_CLOEXEC) = -1 EINVAL (Invalid argument)

Although this then goes on to succesfully open/work with perf stat. I guess the size being set to zero is an issue ...

PR #1629, opened 2018-03-14
fix fd leakage

nirmoy:
bpf_prog_compute_tag() isn't closing socket fds
before successful return


4ast:
thanks. not sure how I missed that.

yonghong-song:
[buildbot, ok to test]

PR #1630, opened 2018-03-14
tools/slabratetop: detect the current memory allocator

lcp:
Currently, slabratetop always included slub_def.h. For the system uses
SLAB, the definition of 'struct kmem_cache' is different and the eBPF
program would fetch the wrong field and print garbages.


yonghong-song:
[buildbot, ok to test]

PR #1632, opened 2018-03-15
cmake: search for iperf3 if there is no iperf

nirmoy:
Some destro just provide iperf3 like openSUSE Tumbleweed


yonghong-song:
[buildbot, ok to test]

yonghong-song:
LGTM

PR #1636, opened 2018-03-19
tcplife: switch to the new sock:inet_sock_set_state tracepoint

brendangregg:
The tcp:tcp_set_state tracepoint appeared in 4.15, and has been disabled in 4.16, replaced by sock:inet_sock_set_state. See:

https://github.com/torvalds/linux/commit/563e0bb0dc74b3ca888e24f8c08f0239fe4016b0

By supporting sock:inet_sock_set_state, tcplife on 4.15 will now revert back to kprobes of tcp_set_state(), which should be fine.


yonghong-song:
Tested on latest net-next. Looks good. Thanks!

PR #1638, opened 2018-03-20
debian: add python3-bcc package

andihit:
This patch extends the PYTHON_CMD cmake variable to allow a list of
Python binaries and installs the bcc Python bindings for each Python
installation. The Debian control files are updated to generate bindings
for Python 2 and Python 3. Compiled Python files are excluded from the
Debian packages.


yonghong-song:
[buildbot, ok to test]

yonghong-song:
cc @drzaeus77

PR #1639, opened 2018-03-21
add the tcpstates tool

brendangregg:
making use of the new sock:inet_sock_set_state tracepoint in 4.16.

### DIFF
--- a/tools/tcpstates.py
+++ b/tools/tcpstates.py
@@ -1,0 +1,333 @@
+#!/usr/bin/python
+# @lint-avoid-python-3-compatibility-imports
+#
+# tcpstates   Trace the TCP session state changes with durations.
+#             For Linux, uses BCC, BPF. Embedded C.
+#
+# USAGE: tcpstates [-h] [-C] [-S] [interval [count]]
+#
+# This uses the sock:inet_sock_set_state tracepoint, added to Linux 4.16.
+# Linux 4.16 also adds more state transitions so that they can be traced.
+#
+# Copyright 2018 Netflix, Inc.
+# Licensed under the Apache License, Version 2.0 (the "License")
+#
+# 20-Mar-2018   Brendan Gregg   Created this.
+
+from __future__ import print_function
+from bcc import BPF
+import argparse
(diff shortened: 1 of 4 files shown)


tools/tcpstates.py
```diff
@@ -0,0 +1,333 @@
...
+#define KBUILD_MODNAME "foo"
+#include <linux/tcp.h>
+#include <net/sock.h>
+#include <bcc/proto.h>
+
+BPF_HASH(last, struct sock *, u64);
+
+// separate data structs for ipv4 and ipv6
+struct ipv4_data_t {
+    // XXX: switch some to u32's when supported
```
yonghong-song: Do you want to give a try to use proper types instead of uniformly u64?
bcc/ctype support u64 and lower-width type pretty well.

tools/tcpstates.py
```diff
@@ -0,0 +1,333 @@
...
+    u64 newstate;
+    char task[TASK_COMM_LEN];
+};
+BPF_PERF_OUTPUT(ipv4_events);
+
+struct ipv6_data_t {
+    u64 ts_us;
+    u64 skaddr;
+    u64 pid;
+    unsigned __int128 saddr;
```
yonghong-song: This won't work. __int128 will requires an alignment 128bit and the whole structure will have an alignment of 128bit as well. So there will be a padding between `pid` and  `saddr`.
For python, `unsigned __int128` becomes `u64[2]` which has an an alignment of `u64`. So no padding in python side.

tools/tcpstates.py
brendangregg: I tried, but it still doesn't work. I submitted #1642 as a simple test case. Maybe I'm doing it wrong?

tools/tcpstates.py
brendangregg: ok, thanks, pushing the fix.

tools/tcpstates.py
brendangregg: Well... I know what works now: "c_uint" instead of "c_ulong". I've submitted the fix.

PR #1645, opened 2018-03-22
sync bpf compat headers with latest net-next

qmonnet:
Sync from net-next at commit 454bfe97837a.

I directly copied include/uapi/linux/bpf.h from kernel to src/cc/compat/linux/bpf.h in bcc (save for `#include "bpf_common.h"`), so there is a number of whitespace changes. I don't believe this is a problem since this file is only updated from the kernel header file?


yonghong-song:
LGTM. Thanks!

yonghong-song:
@qmonnet , right, the white space is not a problem.

PR #1646, opened 2018-03-23
scripts: check virtual_list.h's list of helpers

pchaigno:
The script to verify all lists of helper in bcc contain the same entries did not check `src/cc/compat/linux/virtual_bpf.h`. This pull request fixes it.

/cc @qmonnet


qmonnet:
Looks good to me as well, thanks for the update :)

PR #1648, opened 2018-03-23
Fix spelling mistake in raised Exception

neosilky:


yonghong-song:
[buildbot, ok to test]

PR #1654, opened 2018-03-27
tools/syscount: convert ausyscall output to string

lcp:
In python3, subprocess.check_output() returns a byte array and it
failed the following split(). Convert the output to string to avoid
the error.

### DIFF
--- a/tools/syscount.py
+++ b/tools/syscount.py
@@ -362,7 +362,7 @@ try:
     # Skip the first line, which is a header. The rest of the lines are simply
     # SYSCALL_NUM\tSYSCALL_NAME pairs.
     out = subprocess.check_output('ausyscall --dump | tail -n +2', shell=True)
-    syscalls = dict(map(parse_syscall, out.strip().split('\n')))
+    syscalls = dict(map(parse_syscall, out.decode().strip().split('\n')))
 except Exception as e:
     if platform.machine() == "x86_64":
         pass


yonghong-song:
[buildbot, ok to test]

tools/syscount.py
```diff
@@ -362,7 +362,7 @@ def parse_syscall(line):
     # Skip the first line, which is a header. The rest of the lines are simply
     # SYSCALL_NUM\tSYSCALL_NAME pairs.
     out = subprocess.check_output('ausyscall --dump | tail -n +2', shell=True)
-    syscalls = dict(map(parse_syscall, out.strip().split('\n')))
+    syscalls = dict(map(parse_syscall, out.decode().strip().split('\n')))
```
drzaeus77: We should be avoiding decode(). Does split(b'\n') work instead?

lcp: split(b'\n') works but every syscall in the output will be prepended a 'b'. Just feel it prettier without 'b'.

yonghong-song:
cc @drzaeus77

tools/syscount.py
drzaeus77: Can the tool use instead the printb function that was added to bcc.utils to handle this type of case? It directly uses `stdout.write()` instead of `print`, which will preserve whichever special characters were there in the process name.

lcp: Ok, I'll send a v2 to use printb.

PR #1656, opened 2018-03-28
Add strict LLVM 3.7.1 version check to CMakeLists.txt

pbn4:
Newer LLVM versions will cause build failure due to headers layout changes. Reproduced with llvm-4.0 on Ubuntu 16.04:
```
./bcc/src/cc/frontends/clang/loader.cc:35:37: fatal error: clang/Basic/FileManager.h: No such file or directory
```


yonghong-song:
This is not the right way to fix. We want bcc works all released llvm versions. I just tested with llvm 4.0 (on centos7 with locally built 4.0) and it works fine for me. Not sure how llvm 4.0 is built on ubuntu 16.04. What is the issue?

palmtenor:
This is definitely not the right fix. BCC can and should build with latest LLVM version and in fact many new features request newer LLVM version. If you are experiencing issue with the build please feel free to file an issue with more details:)

PR #1658, opened 2018-04-02
tools/syscount: convert syscall and process names to bytes array

lcp:
subprocess.check_output() returns a bytes array in python3 while it
returns str in python2. To improve the compatibility, convert the
related fields to bytes arrays and print them with printb().


drzaeus77:
[buildbot, ok to test]

PR #1659, opened 2018-04-02
tools/cachetop: match bpf.ksym() with bytes arrays

lcp:
bpf.ksym() now returns a bytes array, and python3 would issue TypeError
when matching the string.


yonghong-song:
[buildbot, ok to test]

yonghong-song:
[buildbot, ok to test]

PR #1660, opened 2018-04-02
Return number of CPUs with data for BPFPerfBufferTable::poll()

anakryiko:
Currently `BPFPerfBuffer::poll` doesn't return anything. In some cases, though, it's important to know whether poll returned due to reaching timeout or due to new data being available.

For instance, when doing profiling/tracing for some predefined period of time, it's useful to keep reading/processing data until timeout, then detach kprobe/tracepoint to stop tracing, but after that there might be still some data left in perf buffer, so one needs to drain buffer until no more data remain. For the latter part we need to know when to stop, which might not be simple to do right now.

This change passes result of `epoll_wait` through, so it's possible to know when there was an error (-1 will be returned), when there was no data received (0), and some data was read (result >0).

### DIFF
--- a/src/cc/api/BPFTable.cc
+++ b/src/cc/api/BPFTable.cc
@@ -306,14 +306,13 @@ StatusTuple BPFPerfBuffer::close_all_cpu() {
   return StatusTuple(0);
 }
 
-void BPFPerfBuffer::poll(int timeout_ms) {
+int BPFPerfBuffer::poll(int timeout_ms) {
   if (epfd_ < 0)
-    return;
+    return -1;
   int cnt = epoll_wait(epfd_, ep_events_.get(), cpu_readers_.size(), timeout_ms);
-  if (cnt <= 0)
-    return;
   for (int i = 0; i < cnt; i++)
     perf_reader_event_read(static_cast<perf_reader*>(ep_events_[i].data.ptr));
+  return cnt;
 }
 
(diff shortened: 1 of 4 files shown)


src/cc/api/BPFTable.cc
```diff
@@ -306,14 +306,13 @@ StatusTuple BPFPerfBuffer::close_all_cpu() {
   return StatusTuple(0);
 }
 
-void BPFPerfBuffer::poll(int timeout_ms) {
+int BPFPerfBuffer::poll(int timeout_ms) {
   if (epfd_ < 0)
     return;
```
yonghong-song: Missing return value here?

anakryiko: yep, fixing, thanks

yonghong-song:
[buildbot, ok to test]

palmtenor:
Do we want the `BPF` class's `poll_perf_buffer` also return this?

anakryiko:
Good point, just updated BPF::poll_perf_buffer as well. 
I'm wondering if it's good idea to just return -1 if no perf buffer exists or `epfd_ < 0`? Do we want to distinguish these cases by returning -2 and -3, for instance?

yonghong-song:
Returning -1 should be fine. If epfd_ < 0, the application should have got an error earlier.

PR #1665, opened 2018-04-05
fix "dereferencing type-punned pointer" warning

nirmoy:
fix folowing warning:

bcc/src/cc/perf_reader.c:149:3: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
   uint64_t data_head = *((volatile uint64_t *)&perf_header->data_head);
   ^~~~~~~~

### DIFF
--- a/src/cc/perf_reader.c
+++ b/src/cc/perf_reader.c
@@ -146,7 +146,7 @@ static void parse_sw(struct perf_reader *reader, void *data, int size) {
 }
 
 static uint64_t read_data_head(struct perf_event_mmap_page *perf_header) {
-  uint64_t data_head = *((volatile uint64_t *)&perf_header->data_head);
+  volatile uint64_t data_head = *(&perf_header->data_head);
   asm volatile("" ::: "memory");
   return data_head;
 }


nirmoy:
I am not expert in compiler optimization but I think this change wont have any consequences.

yonghong-song:
[buildbot, ok to test]

yonghong-song:
@nirmoy Your patch works because compiler does not know whether perf_header->data_head changed or not due to callback functions. it will fetch perf_header->data_head even without
volatile marking.

How about the following patch?

```
diff --git a/src/cc/perf_reader.c b/src/cc/perf_reader.c
index afed162..958c00a 100644
--- a/src/cc/perf_reader.c
+++ b/src/cc/perf_reader.c
@@ -145,19 +145,19 @@ static void parse_sw(struct perf_reader *reader, void *data, int size) {
     reader->raw_cb(reader->cb_cookie, raw->data, raw->size);
 }
 
-static uint64_t read_data_head(struct perf_event_mmap_page *perf_header) {
-  uint64_t data_head = *((volatile uint64_t *)&perf_header->data_head);
+static uint64_t read_data_head(volatile struct perf_event_mmap_page * perf_header) {
+  uint64_t data_head = perf_header->data_head;
   asm volatile("" ::: "memory");
   return data_head;
 }
 
-static void write_data_tail(struct perf_event_mmap_page *perf_header, uint64_t data_tail) {
+static void write_data_tail(volatile struct perf_event_mmap_page * perf_header, uint64_t data_tail) {
   asm volatile("" ::: "memory");
   perf_header->data_tail = data_tail;
 }
 
 void perf_reader_event_read(struct perf_reader *reader) {
-  struct perf_event_mmap_page *perf_header = reader->base;
+  volatile struct perf_event_mmap_page * perf_header = reader->base;
   uint64_t buffer_size = (uint64_t)reader->page_size * reader->page_cnt;
   uint64_t data_head;
   uint8_t *base = (uint8_t *)reader->base + reader->page_size;
```
It explicitly says that the pointee of perf_header may change and then the compiler will do the right thing and fetch perf_header fields.
Kernel samples/bpf/trace_output_user.c uses similar mechanism to declare the perf_event_mmap_page as volatile pointer.

nirmoy:
@yonghong-song I like your patch better.

nirmoy:
ref: https://github.com/iovisor/bcc/pull/1669

Issue #1668, opened 2018-04-05
execsnoop: Arguments containing spaces are indistinguishable from multiple arguments


brtmr:
When running execsnoop, it's not possible to determine wether a process was created with one argument containing spaces, or with multiple arguments, so opening a shell and running
`cat "hello world"`
and 
`cat hello world`
will show up just the same, namely as 
`cat              3752   3668     0 /usr/bin/cat hello world`
Would it be possible to include argument quoting, so that these two are different? Preferably so that 
`cat "hello world"`
`cat hello world`
and
`cat \"hello world\"` are all different. 
I thought of adding a flag that would null - byte separate multiple arguments on the same line (not very human-readable, but useful for piping to other tools). If that sounds acceptable, i will create a pull-request.

Quick edit: Simply escaping spaces as ' ' -> '\ ' before calling printb would be sufficient as well.

yonghong-song:
Do you have a real world example to show this is needed?
How about we detect there is a space in the string and add "" then?

brendangregg:
I'd make it an option, like -q which quoted all arguments (whether they need it or not).

brtmr:
@yonghong-song 
I was using execsnoop to test and debug a tool that was creating subprocesses, and was trying to figure out wether the argument splitting / passing worked correctly.

PR #1672, opened 2018-04-08
Added -q option to execsnoop to quote individual arguments. 

brtmr:
This helps when working with arguments that contain spaces. It also enables to trace calls that contain empty arguments. As discussed in issue #1668


yonghong-song:
Sounds good. Can you add this option to the example file and manpage as well? Thanks!

brtmr:
@yonghong-song Done.

yonghong-song:
[buildbot, ok to test]

PR #1673, opened 2018-04-09
Improve clear table API in C++

palmtenor:
- In current `BPFHashTable`'s `clear_table_non_atomic`, we remove the key then do `next` on the same key, which effectively gives us the first key table. This commit changes it to always just find and remove the first key to clear the table.
- Add `clear_table_non_atomic` for the generic `BPFTable` since it doesn't need type information.
- Tests

### DIFF
--- a/src/cc/api/BPFTable.cc
+++ b/src/cc/api/BPFTable.cc
@@ -142,6 +142,45 @@ StatusTuple BPFTable::remove_value(const std::string& key_str) {
   return StatusTuple(0);
 }
 
+StatusTuple BPFTable::clear_table_non_atomic() {
+  if (desc.type == BPF_MAP_TYPE_HASH ||
+      desc.type == BPF_MAP_TYPE_PERCPU_HASH ||
+      desc.type == BPF_MAP_TYPE_LRU_HASH ||
+      desc.type == BPF_MAP_TYPE_PERCPU_HASH ||
+      desc.type == BPF_MAP_TYPE_HASH_OF_MAPS) {
+    // For hash maps, use the first() interface (which uses get_next_key) to
+    // iterate through the map and clear elements
+    auto key = std::unique_ptr<void, decltype(::free)*>(
+      ::malloc(desc.key_size),
+      ::free);
+
+    while (this->first(key.get()))
+      if (!this->remove(key.get())) {
+        return StatusTuple(
+          -1,
+          "Failed to delete element when clearing table %s",
+          desc.name.c_str());
+      }
+  } else if (desc.type == BPF_MAP_TYPE_ARRAY ||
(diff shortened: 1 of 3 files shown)


src/cc/api/BPFTable.cc
```diff
@@ -142,26 +142,37 @@ StatusTuple BPFTable::remove_value(const std::string& key_str) {
   return StatusTuple(0);
 }
 
+StatusTuple BPFTable::clear_table_non_atomic() {
```
yonghong-song: The implementation here takes advantage of kernel internal implementation that if a key is not found the kernel will search from the start. I think we should still stick to the public interface get_next and not relying on this particular kernel behavior.

Also, if implemented this function at BPFTable level, it should work for certain types of array maps as well.
For example, prog_array does support get_next and delete_elem. Using get_first and remove mechanism won't work for prog_array map. The same holds true for cgroup_fd_array and array_of_maps.

palmtenor: Good point for the table types, will add check.

For the implementation, even if we use get_next, we still need to somehow find the first key to "get next for", right? If that's the case we could just keep re-using it anyways...

yonghong-song: Yes, for non-array map type, if the key cannot be found, it will find the first key, the same as you pass an invalid key or NULL pointer key.

palmtenor:
Added type check

src/cc/api/BPFTable.cc
```diff
@@ -142,26 +142,69 @@ StatusTuple BPFTable::remove_value(const std::string& key_str) {
...
+    return StatusTuple(
+      -1, "Array map %s do not support clearing elements", desc.name.c_str());
+  } else if (desc.type == BPF_MAP_TYPE_PROG_ARRAY ||
+             desc.type == BPF_MAP_TYPE_PERF_EVENT_ARRAY ||
+             desc.type == BPF_MAP_TYPE_STACK_TRACE ||
+             desc.type == BPF_MAP_TYPE_ARRAY_OF_MAPS) {
+    // For Stack-trace and FD arrays, just iterate over all indices
+    for (size_t i = 0; i < desc.max_entries; i++)
+      if (!this->remove(&i)) {
+        return StatusTuple(
```
yonghong-song: When you deleted an element from an array, it is possible that that array index does not have a valid entry, so it will return an -ENOENT. One example is STACK_TRACE, not every array index will result in valid stack trace. Another is ARRAY_OF_MAPS, it may not be true every array index has corresponding maps.

So we should just call this->remove(&i) for all array elements and there is not need to care about the return value.

palmtenor:
Makes sense, updated!

PR #1675, opened 2018-04-10
Add support for remote target devices via BPFd

jcanseco:
### Overview
These patches add features that allow BCC to support cross-development workflows where the development machine and the target machine running the BPF program are different.

This is achieved by integrating the BPFd (BPF Daemon) project into BCC. BPFd is a standalone executable that can be loaded onto a remote target device, and which then can act as a proxy for BCC for whenever BCC wishes to perform an operation on the system (e.g. load BPF programs, read /proc/kallsyms, attach kprobes, etc.).

Through this arrangement, BCC can be used to profile a remote target device while it mostly runs on a separate host machine. The advantage of this arrangement is that it removes the need to have kernel sources and the LLVM stack on the target machine. These can, instead, be kept on the host. This arrangement therefore reduces the space required on a target for BCC tools to run, which is a key benefit for devices that have a more limited disk space (e.g. embedded devices).

In addition, the above set-up also allows embedded developers who use a cross-compiler in their workflow to run clang on a different architecture than the target's architecture, thus facilitating cross-compilation development.

For more information, please check out this [LWN article](https://lwn.net/Articles/744522/) which explains the purpose of BPFd and how it works in more detail.

### Integration of BPFd sources into BCC

These patches add the sources for the BPFd executable into the BCC source tree. This is done to ensure that BCC and BPFd remain compatible with each other.

BPFd depends on existing BCC components such as `libbpf.c` and `bcc_syms.cc` to function. However, the converse is also true: BCC makes calls to BPFd via BCC's Python interface. The Python interface is the main way by which communication happens between BCC and BPFd, and so any changes there could break interoperability. As a result, it is not sufficient for BPFd to just use `libbcc` and remain independent like other projects (e.g. bpftrace).

Therefore, to ensure that BCC and BPFd are always compatible with each other, it is more feasible to keep them in the same tree instead of keeping them separate. This is also why these patches come with smoke tests which ensure that the interoperability between BCC and BPFd isn't broken silently when changes are made to either.

### Tools

The tools that currently work for remote devices with these patches are as follows:

- Biolatency
- Biosnoop
- Biotop
- Cachestat
- Cachetop
- Filetop
- Hardirqs
- Offcputime
- Opensnoop
- Profile
- Runqlen
- Stackcount
- Syscount
- Trace

Reviewed-by: Joel Fernandes (joelaf@google.com)

### DIFF
--- a/src/cc/CMakeLists.txt
+++ b/src/cc/CMakeLists.txt
@@ -81,6 +81,7 @@ if(ENABLE_USDT)
 endif()
 
 add_subdirectory(frontends)
+add_subdirectory(bpfd)
 
 # Link against LLVM libraries
 target_link_libraries(bcc-shared ${bcc_common_libs_for_s})
--- a/src/cc/bpf_common.cc
+++ b/src/cc/bpf_common.cc
@@ -17,6 +17,14 @@
 #include "bpf_module.h"
 
 extern "C" {
+
+// This stores the appropriate callback function to invoke when there is a
+// need to create a map. This is done so that the Python interface can pass
+// custom callbacks to create maps for situations wherein the map creation
+// functions provided by libbbc are not sufficient (e.g. when there is a need
+// to create maps on remote target devices).
+bpf_create_map_cb_t bpf_create_map_cb;
+
 void * bpf_module_create_b(const char *filename, const char *proto_filename, unsigned flags) {
   auto mod = new ebpf::BPFModule(flags);
   if (mod->load_b(filename, proto_filename) != 0) {
@@ -26,7 +34,9 @@ void * bpf_module_create_b(const char *filename, const char *proto_filename, uns
   return mod;
 }
 
-void * bpf_module_create_c(const char *filename, unsigned flags, const char *cflags[], int ncflags) {
+void * bpf_module_create_c(const char *filename, unsigned flags, const char *cflags[], int ncflags, bpf_create_map_cb_t map_cb) {
+  bpf_create_map_cb = map_cb;
+
   auto mod = new ebpf::BPFModule(flags);
   if (mod->load_c(filename, cflags, ncflags) != 0) {
     delete mod;
@@ -35,7 +45,9 @@ void * bpf_module_create_c(const char *filename, unsigned flags, const char *cfl
   return mod;
 }
 
-void * bpf_module_create_c_from_string(const char *text, unsigned flags, const char *cflags[], int ncflags) {
+void * bpf_module_create_c_from_string(const char *text, unsigned flags, const char *cflags[], int ncflags, bpf_create_map_cb_t map_cb) {
+  bpf_create_map_cb = map_cb;
+
   auto mod = new ebpf::BPFModule(flags);
   if (mod->load_string(text, cflags, ncflags) != 0) {
     delete mod;
--- a/src/cc/bpf_common.h
+++ b/src/cc/bpf_common.h
@@ -24,9 +24,10 @@
 extern "C" {
 #endif
 
+typedef void (*bpf_create_map_cb_t)(void *data);
 void * bpf_module_create_b(const char *filename, const char *proto_filename, unsigned flags);
-void * bpf_module_create_c(const char *filename, unsigned flags, const char *cflags[], int ncflags);
-void * bpf_module_create_c_from_string(const char *text, unsigned flags, const char *cflags[], int ncflags);
+void * bpf_module_create_c(const char *filename, unsigned flags, const char *cflags[], int ncflags, bpf_create_map_cb_t map_cb);
+void * bpf_module_create_c_from_string(const char *text, unsigned flags, const char *cflags[], int ncflags, bpf_create_map_cb_t map_cb);
 void bpf_module_destroy(void *program);
 char * bpf_module_license(void *program);
 unsigned bpf_module_kern_version(void *program);
--- a/src/cc/bpfd/CMakeLists.txt
+++ b/src/cc/bpfd/CMakeLists.txt
@@ -1,0 +1,6 @@
+# Copyright (c) Jazel Canseco, 2018
+# Licensed under the Apache License, Version 2.0 (the "License")
+
+add_executable(bpfd bpfd.c base64.c remote_perf_reader.c utils.c cmd_parsers.c)
+target_link_libraries(bpfd bpf-shared)
+target_link_libraries(bpfd bcc-shared)
--- a/src/cc/bpfd/base64.c
+++ b/src/cc/bpfd/base64.c
@@ -1,0 +1,273 @@
+/*
+ * Base64 Encoding/Decoding Library
+ *
+ * Author: freecode http://freecode-freecode.blogspot.com/
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+
+/**
+* characters used for Base64 encoding
+*/
+const char *BASE64_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+
+/**
+* encode three bytes using base64 (RFC 3548)
+*
+* @param triple three bytes that should be encoded
+* @param result buffer of four characters where the result is stored
+*/
+void _base64_encode_triple(unsigned char triple[3], char result[4]) {
+  int tripleValue, i;
+
+  tripleValue = triple[0] ;
+  tripleValue *= 256;
+  tripleValue += triple[1];
+  tripleValue *= 256;
+  tripleValue += triple[2];
+
+  for (i = 0; i < 4; i++) {
+    result[3 - i] = BASE64_CHARS[tripleValue % 64];
+    tripleValue /= 64;
+  }
+}
+
+/**
+* encode an array of bytes using Base64 (RFC 3548)
+*
+* @param source the source buffer
+* @param sourcelen the length of the source buffer
+* @param target the target buffer
+* @param targetlen the length of the target buffer
+* @return 1 on success, 0 otherwise
+*/
+int base64_encode(unsigned char *source, size_t sourcelen, char *target, size_t targetlen) {
+  /* check if the result will fit in the target buffer */
+  if ((sourcelen+2)/3*4 > targetlen-1)
+    return 0;
+
+  /* encode all full triples */
+  while (sourcelen >= 3)
+  {
+    _base64_encode_triple(source, target);
+    sourcelen -= 3;
+    source += 3;
+    target += 4;
+  }
+
+  /* encode the last one or two characters */
+  if (sourcelen > 0)
+  {
+    unsigned char temp[3];
+    memset(temp, 0, sizeof(temp));
+    memcpy(temp, source, sourcelen);
+    _base64_encode_triple(temp, target);
+    target[3] = '=';
+    if (sourcelen == 1)
+      target[2] = '=';
+
+    target += 4;
+  }
+
+  /* terminate the string */
+  target[0] = 0;
+
+  return 1;
+}
+
+/**
+* determine the value of a base64 encoding character
+*
+* @param base64char the character of which the value is searched
+* @return the value in case of success (0-63), -1 on failure
+*/
+int _base64_char_value(char base64char) {
+  if (base64char >= 'A' && base64char <= 'Z')
+    return base64char-'A';
+  if (base64char >= 'a' && base64char <= 'z')
+    return base64char-'a'+26;
+  if (base64char >= '0' && base64char <= '9')
+    return base64char-'0'+2*26;
+  if (base64char == '+')
+    return 2*26+10;
+  if (base64char == '/')
+    return 2*26+11;
+  return -1;
+}
+
+/**
+* decode a 4 char base64 encoded byte triple
+*
+* @param quadruple the 4 characters that should be decoded
+* @param result the decoded data
+* @return lenth of the result (1, 2 or 3), 0 on failure
+*/
+int _base64_decode_triple(char quadruple[4], unsigned char *result) {
+  int i, triple_value, bytes_to_decode = 3, only_equals_yet = 1;
+  int char_value[4];
+
+  for (i = 0; i < 4; i++)
+    char_value[i] = _base64_char_value(quadruple[i]);
+
+  /* check if the characters are valid */
+  for (i = 3; i >= 0; i--)
+  {
+    if (char_value[i] < 0)
+    {
+      if (only_equals_yet && quadruple[i] == '=')
+      {
+        /* we will ignore this character anyway, make it something
+        * that does not break our calculations */
+        char_value[i] = 0;
+        bytes_to_decode--;
+        continue;
+      }
+      return 0;
+    }
+    /* after we got a real character, no other '=' are allowed anymore */
+    only_equals_yet = 0;
+  }
+
+  /* if we got "====" as input, bytes_to_decode is -1 */
+  if (bytes_to_decode < 0)
+    bytes_to_decode = 0;
+
+  /* make one big value out of the partial values */
+  triple_value = char_value[0];
+  triple_value *= 64;
+  triple_value += char_value[1] ;
+  triple_value *= 64;
+  triple_value += char_value[2];
+  triple_value *= 64;
+  triple_value += char_value[3];
+
+  /* break the big value into bytes */
+  for (i = bytes_to_decode; i < 3; i++)
+    triple_value /= 256;
+  for (i = bytes_to_decode - 1; i >= 0; i--)
+  {
+    result[i] = triple_value % 256;
+    triple_value /= 256;
+  }
+
+  return bytes_to_decode;
+}
+
+/**
+* decode base64 encoded data
+*
+* @param source the encoded data (zero terminated)
+* @param target pointer to the target buffer
+* @param targetlen length of the target buffer
+* @return length of converted data on success, -1 otherwise
+*/
+size_t base64_decode(char *source, unsigned char *target, size_t targetlen) {
+  char *src, *tmpptr;
+  char quadruple[4], tmpresult[3];
+  int i;
+  int tmplen = 3;
+  size_t converted = 0;
+
+  /* concatinate '===' to the source to handle unpadded base64 data */
+  src = (char *)malloc(strlen(source) + 5);
+  if (src == NULL)
+    return -1;
+  strcpy(src, source);
+  strcat(src, "====");
+  tmpptr = src;
+
+  /* convert as long as we get a full result */
+  while (tmplen == 3)
+  {
+    /* get 4 characters to convert */
+    for (i = 0; i < 4; i++)
+    {
+      /* skip invalid characters - we won't reach the end */
+      while (*tmpptr != '=' && _base64_char_value(*tmpptr) < 0)
+        tmpptr++;
+
+      quadruple[i] = *(tmpptr++);
+    }
+
+    /* convert the characters */
+    tmplen = _base64_decode_triple(quadruple, (unsigned char *)tmpresult);
+
+    /* check if the fit in the result buffer */
+    if (targetlen < tmplen)
+    {
+      free(src);
+      return -1;
+    }
+
+    /* put the partial result in the result buffer */
+    memcpy(target, tmpresult, tmplen);
+    target += tmplen;
+    targetlen -= tmplen;
+    converted += tmplen;
+  }
+
+  free(src);
+  return converted;
+}
+
+void test_base64(char *file) {
+  struct stat st;
+  char *fileout, *encoded, *filebuf;
+  size_t size;
+  int ret;
+  FILE *fp;
+  char *target;
+
+  stat(file, &st);
+  size = st.st_size;
+
+  fileout = (char *)malloc(strlen(file) + 1 + 4);
+  fileout[0] = 0;
+  strcat(fileout, file);
+  strcat(fileout, ".b64dec");
+
+  printf("Encoding and then decoding %s into %s, filesize is %d\n", file, fileout, (int) size);
+
+  encoded = (char *)malloc((size * 4) + 1);
+  encoded[(size * 4)] = 0;
+
+  filebuf = (char *)malloc(size);
+  fp = fopen(file, "rb");
+  fread(filebuf, size, 1, fp);
+
+  ret = base64_encode(filebuf, size, encoded, size * 4);
+
+  printf("encoded len: %d\n", (int)strlen(encoded));
+
+  printf("encoded stat: %s\n", encoded);
+
+  target = (char *)malloc(size);
+
+  ret = base64_decode(encoded, target, size);
+
+  fp = fopen(fileout, "wb");
+
+  printf("fp=%p ret=%d fileout=%s\n", (void *)fp, ret, fileout);
+
+  fwrite(target, size, 1, fp);
+
+  fclose(fp);
+}
--- a/src/cc/bpfd/base64.h
+++ b/src/cc/bpfd/base64.h
@@ -1,0 +1,3 @@
+int base64_encode(unsigned char *source, size_t sourcelen, char *target, size_t targetlen);
+size_t base64_decode( char *source, unsigned char *target, size_t targetlen);
+void test_base64(char *file);
--- a/src/cc/bpfd/bpfd.c
+++ b/src/cc/bpfd/bpfd.c
@@ -1,0 +1,777 @@
+/*
+ * BPFd (Berkeley Packet Filter daemon)
+ *
+ * Copyright (C) 2017 Joel Fernandes <agnel.joel@gmail.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <linux/bpf.h>
+#include <arpa/inet.h>
+#include <inttypes.h>
+
+#include "bpfd.h"
+#include "bcc_syms.h"
+
+#define LINEBUF_SIZE  2000000
+
+#define DEFAULT_MAX_PID  32768
+
+struct usym_cache {
+  int pid;
+  char *exe_path;
+  void *cache;
+};
+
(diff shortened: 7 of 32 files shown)


yonghong-song:
[buildbot, ok to test]

jcanseco:
Update: applied style changes suggested by the style-check.sh script

yonghong-song:
@drzaeus77 Could you add python module `pexpect` on our unit test VMs? Looks like tests failed due to this missing module.

drzaeus77:
Looking into it.

drzaeus77:
[buildbot, test this please]

This is just a fc27 test, if that passes I will apply the update to the other vm images.

jcanseco:
Hi @drzaeus77, thanks for taking the time. Is `pexpect` still installed for the fc27 machine? And from what I can see, everything else is passing again except `test_libbcc` and `test_tools_on_remote` (the one added by this PR). 

I'll take a look the moment I get into the office on Monday, but I think I know what the issue may be.

jcanseco:
Hi @drzaeus77, I think I figured out why `test_tools_on_remote` was failing. The VMs are not configured to build the `bpfd` executable, which the test depends on. This shouldn't be a problem if the VMs invoke `make all` before invoking `make test` (which was my original assumption), but I believe the VMs are actually configured to build specific targets instead of just the `all` target.

For instance, I believe that the VMs build the `bcc-static`, `bpf-static`, `usdt-static` targets (and many others), but not the `LLCStat`, `TCPSendStack`, and `CPUDistribution` targets. I determined this by printing the directory structure during test execution and verified the presence (or lack of presence) of binaries associated with these targets.

Is my understanding here correct? If so, what am I supposed to do to have the VMs build the `bpfd` target as well (i.e. are there any config files I am supposed to change, or can this change only be done from your end?)

drzaeus77:
We `make test` to build just the test binaries, the rest of the libraries/binaries are built through the rpm/deb build steps and then installed (rpm -i, dpkg -i). So, if you are missing a target, that means it needs to be added to a package spec/debian rules.

jcanseco:
Thank you @drzaeus77, you were completely right. I updated the rpm and deb build steps. The tests for fc27 are passing now.

Can you please now install the `pexpect` module for the other VMs as well?

drzaeus77:
I've updated the images, though the Ubuntu variants didn't compile, seemingly due to an issue during bpfd linking. The fc27 test also failed, though that image is unchanged so maybe the test is a little bit flaky?

jcanseco:
I resolved the issues, but the ubuntu1604 tests failed due a `java.io.EOFException`. The issue this time seems to be unrelated to this PR, since the ubuntu1604 [tests for another PR](https://buildbot.iovisor.org/jenkins/job/bcc-pr/1044/) also failed for the same reason.

Also, I'd like to apologize for the test problems this PR has been causing. I hope I'm not causing you too much of an inconvenience, @drzaeus77. I deeply appreciate all the help you're providing!

drzaeus77:
[buildbot, test this please]

No apologies necessary btw. Thanks for working through these tests.

yonghong-song:
@drzaeus77 what is possible reason for this failure? My local ubuntu1604 with `4.4.0-87-generic` works fine. Looks like the ubuntu1604 in the test infra is `4.4.0-119-generic`. Maybe some other updated softwares along the way caused this failure?

drzaeus77:
@yonghong-song The java error is probably the jenkins master losing connection to the jenkins slave. Seems that something during that traffic test is interrupting the connection, perhaps a crash? I will take a look tomorrow if I get a chance...it will probably involve having access to the GCE/VM serial console.

yonghong-song:
Sounds good. Thanks!

drzaeus77:
The failure is weird, posting my observations so far even though they are unrelated to this PR.

The ubuntu1604 VM, if upgraded with latest packages, is producing an empty libbcc.so.0.5.0 in the build directory when `make -C tests/cc` is run. The deb package has a perfectly normal build.

```
iovisor@ubuntu1604-slave-917:~/jenkins/workspace/bcc-pr/label/ubuntu1604/build$ ls -l src/cc/                                                                                                                   
total 980                                                                                                                                                                                                            
drwxrwxr-x 3 iovisor iovisor   4096 Apr 18 17:19 api                                                                                                                                                                       
drwxrwxr-x 8 iovisor iovisor   4096 Apr 18 17:18 CMakeFiles                                                                                                                                               
-rw-rw-r-- 1 iovisor iovisor   5738 Apr 18 17:18 cmake_install.cmake                                                                                                                               
-rw-rw-r-- 1 iovisor iovisor    389 Apr 18 17:18 CTestTestfile.cmake                                                                                                                                                        
drwxrwxr-x 5 iovisor iovisor   4096 Apr 18 17:18 frontends                                                                                                                                                            
-rw-rw-r-- 1 iovisor iovisor 702700 Apr 18 17:20 libbcc.a                                                                                                                                                     
-rw-rw-r-- 1 iovisor iovisor 171282 Apr 18 17:18 libbcc-loader-static.a                                                                                                                                               
-rw-rw-r-- 1 iovisor iovisor    303 Apr 18 17:18 libbcc.pc                                                                                                                                                                  
lrwxrwxrwx 1 iovisor iovisor     11 Apr 18 17:20 libbcc.so -> libbcc.so.0                                                                                                                                 
lrwxrwxrwx 1 iovisor iovisor     15 Apr 18 17:20 libbcc.so.0 -> libbcc.so.0.5.0                                                                                                                                      
-rwxrwxr-x 1 iovisor iovisor      0 Apr 18 17:20 libbcc.so.0.5.0                                                                                                                                          
-rw-rw-r-- 1 iovisor iovisor  49108 Apr 18 17:18 libbpf.a                                                                                                                                                               
-rw-rw-r-- 1 iovisor iovisor  42359 Apr 18 17:18 Makefile                                                                                                                                                     
drwxrwxr-x 3 iovisor iovisor   4096 Apr 18 17:19 usdt                                                                                                                                                                       
```
That 0-sized libbcc.so.0.5.0 causes the test to fail with:
```
4: Traceback (most recent call last):                                                                                                                                                                     
4:   File "/home/iovisor/jenkins/workspace/bcc-nightly-test/label/ubuntu1604/tests/python/test_stat1.py", line 10, in <module>                                                                     
4:     from bcc import BPF                                                                                                                                                                                                  
4:   File "/home/iovisor/jenkins/workspace/bcc-nightly-test/label/ubuntu1604/src/python/bcc/__init__.py", line 27, in <module>                                                                                        
4:     from .libbcc import lib, bcc_symbol, bcc_symbol_option, _SYM_CB_TYPE                                                                                                                                   
4:   File "/home/iovisor/jenkins/workspace/bcc-nightly-test/label/ubuntu1604/src/python/bcc/libbcc.py", line 17, in <module>                                                                                          
4:     lib = ct.CDLL("libbcc.so.0", use_errno=True)                                                                                                                                                                         
4:   File "/usr/lib/python2.7/ctypes/__init__.py", line 362, in __init__                                                                                                                                                    
4:     self._handle = _dlopen(self._name, mode)                                                                                                                                                                      
4: OSError: /home/iovisor/jenkins/workspace/bcc-nightly-test/label/ubuntu1604/build/src/cc/libbcc.so.0: file too short                                                                                    
4: Failed                                                                                                                                                                                                               
1/1 Test #4: py_test_stat1_b ..................***Failed    1.60 sec                                                                                                                                          
```
For some reason, the ssh session is interrupted when this failure occurs.

yonghong-song:
[buildbot, ok to test]

jcanseco:
Hello, please note that the ubuntu1604 tests failed again since it was lacking the `pexpect` module.

drzaeus77:
Yes, please sit tight. I had to revert to unblock other failing PRs.

jcanseco:
Okay, sounds good, no worries!

yonghong-song:
Started to checkout and look at the code, I got a bunch of warnings with gcc 7.3,
```
[ 94%] Building C object src/cc/bpfd/CMakeFiles/bpfd.dir/utils.c.o
/home/yhs/work/bcc/src/cc/bpfd/base64.c: In function ‘test_base64’:
/home/yhs/work/bcc/src/cc/bpfd/base64.c:250:23: warning: pointer targets in passing argument 1 of ‘base64_encode’ differ in signedness [-Wpointer-sign]
   ret = base64_encode(filebuf, size, encoded, size * 4);
                       ^~~~~~~
/home/yhs/work/bcc/src/cc/bpfd/base64.c:63:5: note: expected ‘unsigned char *’ but argument is of type ‘char *’
 int base64_encode(unsigned char *source, size_t sourcelen, char *target,
     ^~~~~~~~~~~~~
/home/yhs/work/bcc/src/cc/bpfd/base64.c:258:32: warning: pointer targets in passing argument 2 of ‘base64_decode’ differ in signedness [-Wpointer-sign]
   ret = base64_decode(encoded, target, size);
                                ^~~~~~
/home/yhs/work/bcc/src/cc/bpfd/base64.c:178:8: note: expected ‘unsigned char *’ but argument is of type ‘char *’
 size_t base64_decode(char *source, unsigned char *target, size_t targetlen) {
        ^~~~~~~~~~~~~
/home/yhs/work/bcc/src/cc/bpfd/bpfd.c: In function ‘bpf_prog_load_handle’:
/home/yhs/work/bcc/src/cc/bpfd/bpfd.c:53:31: warning: pointer targets in passing argument 2 of ‘base64_decode’ differ in signedness [-Wpointer-sign]
   if (!base64_decode(bin_b64, bin_buf, bin_len))
                               ^~~~~~~
In file included from /home/yhs/work/bcc/src/cc/bpfd/bpfd.h:20:0,
                 from /home/yhs/work/bcc/src/cc/bpfd/bpfd.c:32:
/home/yhs/work/bcc/src/cc/bpfd/base64.h:3:8: note: expected ‘unsigned char *’ but argument is of type ‘char *’
 size_t base64_decode(char *source, unsigned char *target, size_t targetlen);
        ^~~~~~~~~~~~~
/home/yhs/work/bcc/src/cc/bpfd/bpfd.c: In function ‘bpf_remote_update_elem’:
/home/yhs/work/bcc/src/cc/bpfd/bpfd.c:117:6: warning: ‘lbin’ may be used uninitialized in this function [-Wmaybe-uninitialized]
   if (lbin)
      ^
/home/yhs/work/bcc/src/cc/bpfd/remote_perf_reader.c: In function ‘remote_perf_reader_poll’:
/home/yhs/work/bcc/src/cc/bpfd/remote_perf_reader.c:72:14: warning: unused variable ‘ret’ [-Wunused-variable]
   int i, fd, ret;
              ^~~
/home/yhs/work/bcc/src/cc/bpfd/bpfd.c: In function ‘bpf_remote_lookup_elem’:
/home/yhs/work/bcc/src/cc/bpfd/bpfd.c:124:9: warning: ‘lstr’ may be used uninitialized in this function [-Wmaybe-uninitialized]
   char *lstr, *rets = NULL;
         ^~~~
/home/yhs/work/bcc/src/cc/bpfd/bpfd.c:144:6: warning: ‘lbin’ may be used uninitialized in this function [-Wmaybe-uninitialized]
   if (lbin)
      ^
/home/yhs/work/bcc/src/cc/bpfd/bpfd.c: In function ‘bpf_remote_get_first_key’:
/home/yhs/work/bcc/src/cc/bpfd/bpfd.c:247:6: warning: ‘kstr’ may be used uninitialized in this function [-Wmaybe-uninitialized]
   if (!rets && kstr)
      ^
/home/yhs/work/bcc/src/cc/bpfd/bpfd.c: In function ‘bpf_remote_get_next_key’:
/home/yhs/work/bcc/src/cc/bpfd/bpfd.c:254:9: warning: ‘next_kstr’ may be used uninitialized in this function [-Wmaybe-uninitialized]
   char *next_kstr, *rets = NULL;
         ^~~~~~~~~
/home/yhs/work/bcc/src/cc/bpfd/bpfd.c:276:6: warning: ‘next_kbin’ may be used uninitialized in this function [-Wmaybe-uninitialized]
   if (next_kbin)
      ^
```
Could you try to fix these warnings?

Second, when I build and run test locally
```
make -j10
sudo make install
make test
```
Most tests are failed. I checked one of them `py_test_usdt`
```
[yhs@localhost python]$ sudo ./test_usdt.py
  File "./test_usdt.py", line 9, in <module>
    from bcc import BPF, USDT
  File "/usr/lib/python3.6/site-packages/bcc/__init__.py", line 25, in <module>
    from remote import libremote
ModuleNotFoundError: No module named 'remote'
[yhs@localhost python]$ 
```
Make the above change
```
diff --git a/src/python/bcc/__init__.py b/src/python/bcc/__init__.py
index 2aa8741..ef62926 100644
@@ -22,7 +22,7 @@ import re
 import struct
 import errno
 import sys
-from remote import libremote
+from bcc.remote import libremote
 basestring = (unicode if sys.version_info[0] < 3 else str)
 
 from .libbcc import lib, bcc_symbol, bcc_symbol_option, _SYM_CB_TYPE, _MAP_CB_TYPE
```
It proceeded further but hit another similar issue:
```
  File "./test_usdt.py", line 9, in <module>
    from bcc import BPF, USDT
  File "/usr/lib/python3.6/site-packages/bcc/__init__.py", line 25, in <module>
    from bcc.remote import libremote
  File "/usr/lib/python3.6/site-packages/bcc/remote/libremote.py", line 21, in <module>
    from shell import ShellRemote
ModuleNotFoundError: No module named 'shell'
```

Did I miss any steps for python path?

jcanseco:
I fixed the warnings.

I'm looking into the Python import issue right now, but I can't seem to reproduce it on my end, and the test VMs don't seem to have an issue either. I'll keep investigating.

jcanseco:
Could you perhaps be facing this [issue](https://github.com/nest/nest-simulator/issues/412)?

jcanseco:
Can you try now? I applied a modified variant of your patches and fixed other Python 3 support issues that I identified.

yonghong-song:
much better now. The test `test_tools_on_remote.py` still failed though (with python3 of course).
I have the following diff to fix a couple of things:
```
[yhs@localhost remote]$ git diff
diff --git a/src/python/bcc/remote/libremote.py b/src/python/bcc/remote/libremote.py
index 012d50d..3fcfeb2 100644
--- a/src/python/bcc/remote/libremote.py
+++ b/src/python/bcc/remote/libremote.py
@@ -114,12 +114,12 @@ class LibRemote(object):
         return ret[0]
 
     def bpf_attach_kprobe(self, fd, t, evname, fnname):
-        cmd = "BPF_ATTACH_KPROBE {} {} {} {}".format(fd, t, evname, fnname)
+        cmd = "BPF_ATTACH_KPROBE {} {} {} {}".format(fd, t, evname.decode(), fnname.decode())
         ret = self._remote_send_command(cmd)
         return ret[0]
 
     def bpf_detach_kprobe(self, evname):
-        cmd = "BPF_DETACH_KPROBE {}".format(evname)
+        cmd = "BPF_DETACH_KPROBE {}".format(evname.decode())
         ret = self._remote_send_command(cmd)
         return 0
 
@@ -141,7 +141,7 @@ class LibRemote(object):
 
     def bpf_create_map(self, map_type, name, key_size, leaf_size, max_entries,
                        flags):
-        cmd = "BPF_CREATE_MAP {} {} {} {} {} {}".format(map_type, name, key_size,
+        cmd = "BPF_CREATE_MAP {} {} {} {} {} {}".format(map_type, name.decode(), key_size,
                                     leaf_size, max_entries, flags)
         ret = self._remote_send_command(cmd)
 
diff --git a/src/python/bcc/remote/shell.py b/src/python/bcc/remote/shell.py
index e5e4d10..396c3be 100644
--- a/src/python/bcc/remote/shell.py
+++ b/src/python/bcc/remote/shell.py
@@ -45,7 +45,7 @@ class ShellRemote(BccRemote):
         except pe.exceptions.EOF:
             return ['Command not recognized (timeout)']
 
-        ret = c.before.split('\n')
+        ret = c.before.decode().split('\n')
 
         # Sanitize command output
         ret = [r.rstrip() for r in ret if r]
[yhs@localhost remote]$ 
```

The `shell.py` is necessary to avoid a lot of errors for bytes vs. str.

The `libremote.py` change is to fix the map-creation failure and attaching kprobe failure.
Without this change, the name becomes `b'abc'` instead just `abc`. And of course the
name `b'abc'` will cause the test failure.

We still have some failures like below, which I suspect they all related with this byte vs. str thing.
```
[root@localhost python]# ./test_tools_on_remote.py 
Error in atexit._run_exitfuncs:
  File "/usr/lib/python3.6/site-packages/bcc/__init__.py", line 1305, in cleanup
    self.detach_kprobe_event(k)
  File "/usr/lib/python3.6/site-packages/bcc/__init__.py", line 673, in detach_kprobe_event
    raise Exception("Failed to close kprobe FD")
Exception: Failed to close kprobe FD
.Traceback (most recent call last):
  File "../../tools/biosnoop.py", line 186, in <module>
    b["events"].open_perf_buffer(print_event, page_cnt=64)
  File "/usr/lib/python3.6/site-packages/bcc/table.py", line 598, in open_perf_buffer
    self._open_perf_buffer(i, callback, page_cnt, lost_cb)
  File "/usr/lib/python3.6/site-packages/bcc/table.py", line 631, in _open_perf_buffer
    self[self.Key(cpu)] = self.Leaf(fd)
  File "/usr/lib/python3.6/site-packages/bcc/table.py", line 502, in __setitem__
    super(ArrayBase, self).__setitem__(key, leaf)
  File "/usr/lib/python3.6/site-packages/bcc/table.py", line 233, in __setitem__
    raise Exception("Could not update table")
Exception: Could not update table
Error in atexit._run_exitfuncs:
...
```

How do you debug bpfd? I could add codes to have debug output to a special file. But I am wondering you have a better way to do this. Or the bpfd logs can go to a special location for easy debugging.

yonghong-song:
Thanks for reminding. These will be just workaround to pinpoint where is the issue. I will make sure final code will not use decode().

drzaeus77:
[buildbot, test this please]

yonghong-song:
I did some experiments on the patch. With the following change,
```
diff --git a/src/python/bcc/__init__.py b/src/python/bcc/__init__.py
index 46bfe85..1b0ad6c 100644
@@ -1268,7 +1268,7 @@ class BPF(object):
         try:
             if BPF._libremote:
                 fd_callbacks = []
-                for k, v in self.perf_buffers.iteritems():
+                for k, v in self.perf_buffers.items():
                     # Only support polling for per-cpu perf buffers
                     # { (PerfEventArray-Obj, cpu) -> fd }
                     if v and type(k) == tuple:
diff --git a/src/python/bcc/remote/libremote.py b/src/python/bcc/remote/libremote.py
index 3fcfeb2..9969aa3 100644
--- a/src/python/bcc/remote/libremote.py
+++ b/src/python/bcc/remote/libremote.py
@@ -104,12 +104,12 @@ class LibRemote(object):
             return comm
 
     def bpf_attach_tracepoint(self, fd, cat, tp_name):
-        cmd = "BPF_ATTACH_TRACEPOINT {} {} {}".format(fd, cat, tp_name)
+        cmd = "BPF_ATTACH_TRACEPOINT {} {} {}".format(fd, cat.decode(), tp_name.decode())
         ret = self._remote_send_command(cmd)
         return ret[0]
 
     def bpf_detach_tracepoint(self, tp_category, tp_name):
-        cmd = "BPF_DETACH_TRACEPOINT {} {}".format(tp_category, tp_name)
+        cmd = "BPF_DETACH_TRACEPOINT {} {}".format(tp_category.decode(), tp_name.decode())
         ret = self._remote_send_command(cmd)
         return ret[0]
 
@@ -124,18 +124,19 @@ class LibRemote(object):
         return 0
 
     def bpf_attach_uprobe(self, fd, t, evname, binpath, offset, pid):
-        cmd = "BPF_ATTACH_UPROBE {} {} {} {} {} {}".format(fd, t, evname, binpath, offset, pid)
+        cmd = "BPF_ATTACH_UPROBE {} {} {} {} {} {}".format(fd, t, evname.decode(), binpath.decode(), offset, pid)
         ret = self._remote_send_command(cmd)
         return ret[0]
 
     def bpf_detach_uprobe(self, evname):
-        cmd = "BPF_DETACH_UPROBE {}".format(evname)
+        cmd = "BPF_DETACH_UPROBE {}".format(evname.decode())
         ret = self._remote_send_command(cmd)
         return 0
 
     def bpf_prog_load(self, prog_type, name, func_str, license_str, kern_version):
-        cmd = "BPF_PROG_LOAD {} {} {} {} {} {}".format(prog_type, name, len(func_str),
-              license_str, kern_version, base64.b64encode(func_str))
+        func_bstr = base64.b64encode(func_str)
+        cmd = "BPF_PROG_LOAD {} {} {} {} {} {}".format(prog_type, name.decode(), len(func_str),
+              license_str.decode(), kern_version, func_bstr.decode())
         ret = self._remote_send_command(cmd)
         return ret[0]
 
@@ -151,8 +152,8 @@ class LibRemote(object):
         return ret[0]
 
     def bpf_update_elem(self, map_fd, kstr, klen, lstr, llen, flags):
-        cmd = "BPF_UPDATE_ELEM {} {} {} {} {} {}".format(map_fd, kstr, klen,
-                                                         lstr, llen, flags)
+        cmd = "BPF_UPDATE_ELEM {} {} {} {} {} {}".format(map_fd, kstr.decode(), klen,
+                                                         lstr.decode(), llen, flags)
         ret = self._remote_send_command(cmd)
         return ret[0]
 
@@ -167,7 +168,7 @@ class LibRemote(object):
         if map_fd not in self.map_dumped or self.map_dumped[map_fd] == False:
             self.bpf_get_first_key(map_fd, klen, llen, dump_all=True)
 
-        cmd = "BPF_LOOKUP_ELEM {} {} {} {}".format(map_fd, kstr, klen, llen)
+        cmd = "BPF_LOOKUP_ELEM {} {} {} {}".format(map_fd, kstr.decode(), klen, llen)
         ret = self._remote_send_command(cmd)
         return ret
 
@@ -209,12 +210,12 @@ class LibRemote(object):
             if kstr in self.nkey_cache[map_fd]:
                 return (0, [self.nkey_cache[map_fd][kstr]])
 
-        cmd = "BPF_GET_NEXT_KEY {} {} {}".format(map_fd, kstr, klen)
+        cmd = "BPF_GET_NEXT_KEY {} {} {}".format(map_fd, kstr.decode(), klen)
         ret = self._remote_send_command(cmd)
         return ret
 
     def bpf_delete_elem(self, map_fd, kstr, klen):
-        cmd = "BPF_DELETE_ELEM {} {} {}".format(map_fd, kstr, klen)
+        cmd = "BPF_DELETE_ELEM {} {} {}".format(map_fd, kstr.decode(), klen)
         ret = self._remote_send_command(cmd)
         self._invalidate_map_cache(map_fd)
         return ret[0]
@@ -271,10 +272,14 @@ class LibRemote(object):
             return None, addr, None
         else:
             name, offset, module = ret[1][1].split(";")
-            return name, offset, module
+            # the type of name could be "str" or "bytes"
+            if type(name) is str:
+                return bytearray(name, 'utf-8'), offset, module
+            else:
+                return name, offset, module
 
     def ksymname(self, name):
-        cmd = "GET_KSYM_ADDR {}".format(name)
+        cmd = "GET_KSYM_ADDR {}".format(name.decode())
         ret = self._remote_send_command(cmd)
 
         ret_code = ret[0]
@@ -293,7 +298,11 @@ class LibRemote(object):
             return None, addr, None
         else:
             name, offset, module = ret[1][1].split(";")
-            return name, offset, module
+            # the type of name could be "str" or "bytes"
+            if type(name) is str:
+                return bytearray(name, 'utf-8'), offset, module
+            else:
+                return name, offset, module
 
     def usymname(self, pid, name, module):
         cmd = "GET_USYM_ADDR {} {} {}".format(pid, name, module)
```
python3 test_tools_on_remote can pass. Most decode() here might be justified since we need to base64 conversion and send the commands to bpfd.  I prefer to do `decode()` in python side
so that C side of bpfd should just expect strings (not b'strings').

One annoying thing is related to `ksym` and `usym`. Sometimes, a string is returned and some other time is a byte array. I did not do further investigation though.

jcanseco:
@yonghong-song: thanks for investigating. I've been trying to fix this problem today, and I came up with a patch somewhat similar to the one you just posted. I got the tests passing, but there were other problems that required investigating. We also need to do a thorough test of BCC and BPFd on our end on both Python-2 and Python-3, since this is a non-trivial code change. So please sit tight, this'll probably take some time.

Also, to answer your question about debugging from before: I usually run the individual tools that failed if the test fails. You can set the following environment variables to hookup BCC to BPFd. This will make BCC think that there is a remote target device (in the form of a shell process on the same machine):

```
export ARCH=x86
export BCC_REMOTE=shell
```

And the following variable to turn on debugging logs:
```
export BCC_REMOTE_DEBUG=1
```

This should show you what BCC is sending to BPFd and what BPFd is sending back.

yonghong-song:
Thanks, @jcanseco. Additional testing is definitely helpful. Thanks for the tip for run on the shell with env variable. I indeed used it this morning by looking at the test. One thing I noticed is that the error log inside the bpfd is not visible now. But I see a comment in the bpfd.c which says for later work. Indeed, this can be improved after initial patch.

jcanseco:
Hello everyone, I talked to Joel recently, and we both agreed that we should close this PR for now. We realize that the patch is probably not yet fully ready. There are a few issues that we need to fix on our end first which then need to be thoroughly tested. We don't wish to take up too much of everyone's time during this process, so we've decided to close this for now and reopen it later.

Thank you @yonghong-song and @drzaeus77 for the time you've spent and all the help you've provided! We really appreciate it, and we hope that you look forward to a more refined version of this patch.

@nirmoy If you still wish to try out BPFd in action, you can read the README and INSTALL docs in the original [BPFd repository](https://github.com/joelagnel/bpfd). If you need help, you can contact Joel (joelaf@google.com). Although I recommend waiting before we resubmit this PR so that you'll be able to try out a more refined version of BPFd.

yonghong-song:
@jcanseco  Sure. Just post when you are ready. Thanks!

joelagnel:
Thanks a lot @yonghong-song @drzaeus77 and @jcanseco .

Currently I am working on improving BCC tools on Android and I found another way to run tools without needing BPFd called androdeb : http://tinyurl.com/androdeb

So, I am moving my own focus slightly away from BPFd since I use `androdeb` now for running BCC on Android, however several folks pinged me about BPFd for non-Android so I am sure if there's enough interest and volunteers, then we can do something here. The BCC side changes (this PR) need to be fixed by someone who needs BPFd for their usecases and @jcanseco 's internship at Android which was to get BCC changes completed for BPFd has come to an end (and he's going back to school and said wont get any time).

The other issue with BCC changes for BPFd is that only python frontend will be supported (based on @jcanseco 's PR here) where as with `androdeb`, all front ends and even tools like `bpftrace` will work on Android. So that's why I have been using `androdeb` more now.

About the BPFd project: Several nice things and side-effects like BCC fixes came out of the `BPFd` project, and thanks a lot to @jcanseco and @yonghong-song for that - so I would still look at BPFd as a success and a nice experiment - certainly I learned so much about BCC internals from it.

@nirmoy If you want to continue with BPFd work and have usecases for it, please find the latest work in the following branches:

BCC tree (BCC master with changes to make it work with BPFd):
https://github.com/joelagnel/bcc/tree/bcc-bpfd-upstream

Stand-alone BPFd which you can cross-compile and push as a proxy to targets:
https://github.com/joelagnel/bpfd

Thanks a lot.

Kullu14:
Is this PR opened yet ? We are really looking run bcc on host target environment. Any leads would be a great help.

Kullu14:
@jcanseco  were you trying to use bpfd on host and target environment (2 VMs, non-android platforms). You tried it on same machine (remote process).

Kullu14:
@joelagnel I have already tried some of bcc examples locally on the same host. I would really like to test it on host target environment  (non-android). Is there anything specific do we need to do to just try some of the examples on host target environment?

Kullu14:
@joelagnel, Sorry to bother you again. I could run BPFd on host target environment (non-android). But I need root permission on host to run any tools(filetop.py). Is there any way to run those tools (filetop.py) as normal user ?

yonghong-song:
unfortunately, filetop.py uses kprobe which requires CAP_SYS_ADMIN.

joelagnel:
Thanks Yonghong.

Kullu14:
@yonghong-song  Thanks for the information.
I just have one question, my target has root permission and host doesn't have root permission. As per BPFd implementation, kprobes will be running on target. So Why do I need root permission for target?

yonghong-song:
Could you help debug this a little bit since you have a test case here?

joelagnel:
The target loads the BPF program. If the program is of type KPROBE, then you need to running as CAP_SYS_ADMIN.  I actually hit this too. Only the networking related program types like SOCK_FILTER donot need CAP_SYS_ADMIN

Kullu14:
@joel I have root permission available on target. But I dont have root permission on host while I am running a script there. why do I need to have root permission on host any idea ?

joelagnel:
Your username or shell may not have the cap_sys_admin needed for kprobes. Many bcc tools use kprobes under the hood

PR #1681, opened 2018-04-13
filetop: probe vfs_read and vfs_write instead to fix inlining issue

### LINKED ISSUE #1667: filetop.py does not show any WRITES
Daniel (@danobi) found an issue with filetop.py. 
For example, in window 1,
```
while true; do cat testfile > /dev/null; echo "asdfasdf" >> testfile; done
```
in another window, do
```
sudo filetop.py 5 1 --sort writes
```
you got
```
TID    COMM             READS  WRITES R_Kb    W_Kb    T FILE
1848616 bash             3      0      0       0       R cat
1849361 bash             2      0      0       0       R ld-2.17.so
1848738 chef-client      2      0      16      0       R set.rb
1849110 bash             3      0      0       0       R cat
1849684 bash             3      0      0       0       R cat
1848738 chef-client      2      0      16      0       R aix.rb
1849219 servicerouter    1      0      7       0       R type
1849579 cat              1      0      0       0       R libc-2.17.so
1848730 bash             3      0      0       0       R cat
1847273 python3.6        2      0      5       0       R protocols.cpython-36.pyc
1848547 bash             3      0      0       0       R cat
1849162 scribe_cat       1      0      7       0       R shared_cpu_list
1848577 bash             3      0      0       0       R cat
1848805 cat              7      0      448     0       R testfile
1849598 cat              7      0      448     0       R testfile
1848329 cat              1      0      0       0       R libc-2.17.so
1849247 bash             3      0      0       0       R cat
1848441 cat              7      0      448     0       R testfile
1850087 bash             3      0      0       0       R cat
1849725 bash             3      0      0       0       R cat
```

But I got zero write. 

I did a little bit debug and the following code is the culprit:
```
b.attach_kprobe(event="__vfs_read", fn_name="trace_read_entry")
try:
    b.attach_kprobe(event="__vfs_write", fn_name="trace_write_entry")
except:
    # older kernels don't have __vfs_write so try vfs_write instead
    b.attach_kprobe(event="vfs_write", fn_name="trace_write_entry")
```
In my backported 4.11 fb kernel, both `vfs_write` and `__vfs_write` symbol exists, and in
`fs/read_write.c`, we have
```
...
ssize_t __vfs_write(struct file *file, const char __user *p, size_t count,
                    loff_t *pos)
{       
        if (file->f_op->write)
                return file->f_op->write(file, p, count, pos);
        else if (file->f_op->write_iter)
                return new_sync_write(file, p, count, pos);
        else
                return -EINVAL;
}
...
ssize_t vfs_write(struct file *file, const char __user *buf, size_t count, loff_t *pos)
{
        ssize_t ret;

        if (!(file->f_mode & FMODE_WRITE))
                return -EBADF;
        if (!(file->f_mode & FMODE_CAN_WRITE))
                return -EINVAL;
        if (unlikely(!access_ok(VERIFY_READ, buf, count)))
                return -EFAULT;

         ...
             ret = __vfs_write(file, buf, count, pos);
        ...
}
```

`vfs_write` is calling `__vfs_write`, and tracing `__vfs_write` is more precise since
some error edge cases are filtered out by `vfs_write`.

Based on the implementation, kprobe is attached to `__vfs_write`.
However, by looking at vmlinux disassemble file for `vfs_write`,
```
ffffffff81232610 <vfs_write>:
ffffffff81232610:       e8 bb 2a 73 00          callq  ffffffff819650d0 <__fentry__> 
ffffffff81232615:       55                      push   %rbp
ffffffff81232616:       48 89 e5                mov    %rsp,%rbp
ffffffff81232619:       41 57                   push   %r15
ffffffff8123261b:       41 56                   push   %r14
ffffffff8123261d:       41 55                   push   %r13
ffffffff8123261f:       41 54                   push   %r12
ffffffff81232621:       53                      push   %rbx
ffffffff81232622:       44 8b 67 44             mov    0x44(%rdi),%r12d
...
```
I did not find a call to `__vfs_write`. This means `__vfs_write` is inlined into `vfs_write`.
This perfectly explains why I got 0 write.

With kprobe attaching to vfs_write, the above test can get
```
TID    COMM             READS  WRITES R_Kb    W_Kb    T FILE
2283584 bash             0      6259   0       55      R testfile
1293   rs:main Q:Reg    0      103    0       19      R smc_proxy.log
2944622 cppfbagentd      0      44     0       9       R fbagent_sys_stats_dbg.log
1293   rs:main Q:Reg    0      41     0       7       R configerator_proxy2.log
2944639 cppfbagentd      0      28     0       4       R fbagent_sys_stats_dbg.log
5989   EntityMgr        0      23     0       644     R 371103.log
5205   configerator_fu  1      22     32      644     R 396478.log
2944616 cppfbagentd      0      14     0       2       R fbagent_sys_stats_dbg.log
2944635 cppfbagentd      0      11     0       1       R fbagent_sys_stats_dbg.log
2944626 cppfbagentd      0      10     0       1       R fbagent_sys_stats_dbg.log
979264 TmCheckStatus    0      6      0       0       R 000849.log
2944636 cppfbagentd      0      5      0       8       R 20180405
2944636 cppfbagentd      0      5      0       0       R fbagent_sys_stats_dbg.log
6623   CanaryMgr        0      5      0       0       R 371103.log
1293   rs:main Q:Reg    0      5      0       0       R configerator_fuse.log
1815483 collect_mismatc  0      2      0       0       R storage.stage_nfsmount_fix
546553 fbflow           0      1      0       0       R fbflow.devbig474.prn1.faceboo...
1815990 cat              4      0      256     0       R testfile
1816391 cat              1      0      0       0       R libc-2.17.so
1815327 scribe_cat       1      0      0       0       R libdl-2.23.so
```

File an issue here, so people are aware of this potential issue when they are using kprobe-based tools.

To fix this kind of issue, probably we can have some bcc helper to check whether one function is inlined in another, and the tool can use this information to decide whether which function to attach.

Or we can just attach to the conservative choice, which is `vfs_write` in this case.

NOTE kprobe attachment to `__vfs_read` in this tool may potentially have issue as well if it is
inlined into `vfs_read`.

cc @brendangregg

jcanseco:
Filetop was originally designed to probe __vfs_read and __vfs_write,
which are called by vfs_read and vfs_write, respectively. However,
__vfs_read and __vfs_write are at times inlined, leading to these
particular functions never actually being called. Since these functions
are never called, the probes are never hit, and so filetop ends up
registering zero READs and WRITEs.

This patch fixes the issue by attaching the probes to vfs_read and
vfs_write instead, which are guaranteed to be called during file reads
and writes.

Signed-off-by: Jazel Canseco <jcanseco@google.com>

Fixes #1667

### DIFF
--- a/tools/filetop.py
+++ b/tools/filetop.py
@@ -158,12 +158,8 @@ if debug or args.ebpf:
 
 # initialize BPF
 b = BPF(text=bpf_text)
-b.attach_kprobe(event="__vfs_read", fn_name="trace_read_entry")
-try:
-    b.attach_kprobe(event="__vfs_write", fn_name="trace_write_entry")
-except:
-    # older kernels don't have __vfs_write so try vfs_write instead
-    b.attach_kprobe(event="vfs_write", fn_name="trace_write_entry")
+b.attach_kprobe(event="vfs_read", fn_name="trace_read_entry")
+b.attach_kprobe(event="vfs_write", fn_name="trace_write_entry")
 
 DNAME_INLINE_LEN = 32  # linux/dcache.h
 


yonghong-song:
[buildbot, ok to test]

PR #1688, opened 2018-04-17
fix profile.py with latest net-next

yonghong-song:
The profile.py tries to validate whether the "ip" from the
context register is a kernel ip address or not by comparing
"ip" to the kernel PAGE_OFFSET.

On non-x86_64 architectures, PAGE_OFFSET can be directly used.
On x86_64 architecture, things become more complex.
With 4.17 (latest linus tree), __PAGE_OFFSET_BASE is gone
and replaced with __PAGE_OFFSET_BASE_L5 and __PAGE_OFFSET_BASE_L4
for 5 and 4 level page table respectively.
Running profile.py will have the following error:
    /virtual/main.c:41:18: error: use of undeclared identifier '__PAGE_OFFSET_BASE'
            if (ip > __PAGE_OFFSET_BASE) {
                     ^
    1 error generated.

So, for 4.17, __PAGE_OFFSET_BASE_{L4, L5} should be used.
For 4.16 and some old kernels (e.g., 4.11), which I did not
check earlier one, __PAGE_OFFSET_BASE should be used.
For even older kernels like 4.6, PAGE_OFFSET should be used.

### DIFF
--- a/tools/profile.py
+++ b/tools/profile.py
@@ -151,13 +151,26 @@ int do_perf_event(struct bpf_perf_event_data *ctx) {
         struct pt_regs regs = {};
         bpf_probe_read(&regs, sizeof(regs), (void *)&ctx->regs);
         u64 ip = PT_REGS_IP(&regs);
+        u64 page_offset;
 
         // if ip isn't sane, leave key ips as zero for later checking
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, ok to test]

PR #1568, opened 2018-02-01
Small fixes

liu-song-6:
small fixes in examples/cpp/TCPSendStack.cc and tools/profile.py

PR #1692, opened 2018-04-19
Refactor offcputime stack id error handling

palmtenor:
`offcputime.py` profiles the system for context switches for a period of time, aggregate result in a hash map and output them aggregated after profiling. This means we will only read the stack trace addresses using stack id in the end of run. However, the tool currently uses `BPF_F_REUSE_STACKID`, that means if two different stack traces has the same hash signature, they would have same stack id (i.e., bucket number in the stack map) and the content stored in the stack map would be the later stack trace. This  would lead to incorrect or misleading result.

This PR changes the tool to not use `BPF_F_REUSE_STACKID`. Unfortunately this means a lot of stacks would be missing due to hash collision (`-EEXIST`, current hash function performance is poor, @4ast, @yonghong-song and I are working on it from Kernel side already). Hence, this PR also adds placeholder for these errors to make it clear to users about the existence of those sample / stacks, and just that the tool missed them.

### DIFF
--- a/tools/offcputime.py
+++ b/tools/offcputime.py
@@ -35,6 +35,11 @@ def positive_nonzero_int(val):
         raise argparse.ArgumentTypeError("must be nonzero")
     return ival
 
+def stack_id_err(stack_id):
+    # -EFAULT in get_stackid normally means the stack-trace is not availible,
+    # Such as getting kernel stack trace in userspace code
+    return (stack_id < 0) and (stack_id != -errno.EFAULT)
+
 # arguments
 examples = """examples:
     ./offcputime             # trace off-CPU stack time until Ctrl-C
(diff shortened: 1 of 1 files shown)


brendangregg:
Thanks, this looks like the right approach.

One small thing to fix, probably for older llvm versions:

```
bgregg-xenial-bpf-v000@~> ./offcputime.py
/virtual/main.c:17:36: error: expected ';' after top level declarator
BPF_STACK_TRACE(stack_traces, 1024)
                                   ^
                                   ;
1 error generated.
```

palmtenor:
@brendangregg I tried both LLVM 4.0.1 and 5.0.1 but couldn't repro the issue. I think we changed the semicolon uniformly in #1568. Which LLVM version are you using?

brendangregg:
That was an old system with 3.7.

yonghong-song:
LGTM. Thanks.

palmtenor:
@brendangregg I think you might be using a pre-#1568 runtime but new script, which would make pre-processed code has duplicated semicolon. Can you check?

Issue #1679, opened 2018-04-12
Stacks seem incorrect with offwaketime on Android


joelagnel:
`offwaketime` is giving me some weird waker/sleep stacks on Android/ARM64.

Here's one of the stacks it found:
```
    waker:           find 0
    el0_svc_naked
    sys_newfstatat
    vfs_fstatat
    vfs_getattr_nosec
    kernfs_iop_getattr
    mutex_unlock
    __mutex_unlock_slowpath
    --               --
    try_to_wake_up
    kthread_queue_work
    sugov_irq_work
    irq_work_run_list
    irq_work_run
    handle_IPI
    gic_handle_irq
    el1_irq
    _raw_spin_unlock_irqrestore
    __wake_up_sync_key
    pipe_write
    new_sync_write
    vfs_write
    sys_write
    el0_svc_naked
    target:          ActivityManager 1163
        69
```
Basically the test was running `find /data | cksum` is running in the background with `offwaketime -K`.

What I see above is 3 unrelated stacks. ActivityManager isn't really involved in the find/cksum activity, yet its identified as a sleeper. Lets assume it did do a pipe_write but further down in the sleeper stack, it doesn't show schedule or any such call that can cause a sleep. All it shows is `try_to_wake_up` which is actually a wakeup function not a sleep function.

Also the `try_to_wake_up` is coming from an interrupt stack, and its a path that is handling an IPI related to cpufreq.  This IRQ stack isn't related to the pipe write or `vfs_fsstatat` at above. Its just pure IPI/cpufreq stuff.

Could it be that the IRQ stack (for the IPI) is messing up the offwaketime logic?

@brendangregg , how are we handling IRQ stacks with offwaketime? Like what happens if an IRQ wakes up a task while some other totally unrelated task is running? Is there a false waker/wakee link established there somehow? Any other ideas on the above stack? Thanks!

yonghong-song:
On Thu, Apr 12, 2018 at 1:18 PM, Joel <notifications@github.com> wrote:
> offwaketime is giving me some weird waker/sleep stacks on Android/ARM64.
>
> Here's one of the stacks it found:
>
>     waker:           find 0
>     el0_svc_naked
>     sys_newfstatat
>     vfs_fstatat
>     vfs_getattr_nosec
>     kernfs_iop_getattr
>     mutex_unlock
>     __mutex_unlock_slowpath
>     --               --
>     try_to_wake_up
>     kthread_queue_work
>     sugov_irq_work
[...]

PR #1693, opened 2018-04-19
Fix offwaketime PID / TGID handling

palmtenor:
Currently, `offwaketime.py` uses the same TGID for symbol resolution. But I think the waker and target theoretically could belong to different TGID and we would get wrong symbols. I wonder if this would help for #1679 (@joelagnel can you try?), but it's the right thing to do anyways.

Also improve documentation a bit.

### DIFF
--- a/tools/offwaketime.py
+++ b/tools/offwaketime.py
@@ -114,22 +114,31 @@ struct key_t {
     int t_k_stack_id;
     int t_u_stack_id;
     u32 t_pid;
+    u32 t_tgid;
     u32 w_pid;
-    u32 tgid;
+    u32 w_tgid;
 };
 BPF_HASH(counts, struct key_t);
+
+// Key of this hash is PID of waiting Process,
+// value is timestamp when it went into waiting
 BPF_HASH(start, u32);
+
 struct wokeby_t {
     char name[TASK_COMM_LEN];
     int k_stack_id;
     int u_stack_id;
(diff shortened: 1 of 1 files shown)


yonghong-song:
cc @joelagnel

palmtenor:
I will apply same change of #1692 after this lands. That might also cause incorrect stacks being displayed

joelagnel:
Hi,  Sorry for the delays in replying. I am going through conference
traveling right now, will reply ASAP once I am back (which is 1-2 days).

palmtenor:
@joelagnel No worries. Just want to let you know. We should fix this regardless of if it's root cause of your issue or not:))

tools/offwaketime.py
```diff
@@ -114,58 +114,73 @@ def signal_ignore(signal, frame):
...
+
     wokeby.update(&pid, &woke);
     return 0;
 }
 
 int oncpu(struct pt_regs *ctx, struct task_struct *p) {
-    u32 pid = p->pid, t_pid=pid;
+    // PID and TGID of the previous Process (Process going into waiting)
+    u32 pid = p->pid;
     u32 tgid = p->tgid;
```
yonghong-song: looks like we do not need tgid assignment here.

palmtenor: We don’t use it directly here, but might be used in THREAD_FILTER

PR #1696, opened 2018-04-21
Add a shell.nix

nyarly:
The shell.nix config file captures the build dependencies for BCC and exposes
them on Nix based systems, so that developers can get up and running quickly.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -28,8 +28,10 @@ in
       "-DENABLE_CPP_API=ON"
     ];
 
+    dontFixCmake = true;
+
     shellHook = ''
-      ( cd build; cmakeConfigurePhase 2>&1 > /dev/null )
+      cmakeConfigurePhase 2>&1 > /dev/null
       export LD_LIBRARY_PATH="$PWD/build/src/cc:$LD_LIBRARY_PATH"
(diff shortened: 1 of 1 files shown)


nyarly:
In NixOS or Nix-based systems (c.f. https://nixos.org/), there's a utility provided called `nix-shell` which will start a bash shell (by default) in which a declared set of dependencies are made available. This makes it easy set up a reproducible build environment, as well as making it easier to integrate into the Nix package distribution.

The `shell.nix` file describes that build environment. The buildInputs and nativeBuildInputs declared in the file enumerate the required build tools.

The audience that will benefit is admittedly rather small, but for folks that are using NixOS, or have set up the Nix repository on their Unixy machine, it can been really useful.

That said I can understand if the feeling is that the audience size is one and the maintenance isn't warranted.

4ast:
Considering nix being such a niche offering I don't see how this addition will not instantly rot, since buildbots don't support it. Therefore I don't think it's worth merging it.

lheckemann:
Using the package expression (i.e. `nix-shell '<nixpkgs>' -A linuxPackages.bcc`) might be better?

PR #1697, opened 2018-04-21
locktrace bcc script

gdankel:
Add a script for recording per-lock statistics and write the result
to a csv file. See example file. The output is intended to be processed
by another script / tool.
One such script is lockstat.py, used to view the data in various
forms. For usage see examples file.

I've separated out the .c file so that it can be used in a C++ tool as
well.

This is the first in a series of patches.
Future patches will add:
* Tracking long hold times and location/thread where lock was released.
* Tracking threads that sleep with a contended lock (including nested lock cases)
* Better user-space overhead reporting if h/w counters are available (incl. DataLA support on Haswell and above)
* Special support for adaptive locks that use spin and yield before blocking

Maybe we can also add uprobe support for targeted cases, but it's very fragile and too high overhead for general use.

### DIFF
--- a/tools/lockstat.py
+++ b/tools/lockstat.py
@@ -512,13 +512,10 @@ def print_stats_for_pid(pid, stats, full_stacks, syms,
             key=lambda x: x.sum_all()[order_by], reverse=True):
         sub_stats = sub_stats.sort(order_by)
         subtotal = sub_stats.sum_all()
-        sort_key_us = subtotal[order_by]
-        sort_key_ms = us_to_ms(sort_key_us)
-        if sort_key_ms == 0:
-            break
         add_derived_stats(subtotal, sub_stats, group_by, order_by)
-        total_so_far += sort_key_us
-        subtotal['pct_this'] = round(100.0 * sort_key_us / total[order_by])
+        order_by_val = subtotal[order_by]
+        total_so_far += order_by_val
+        subtotal['pct_this'] = round(100.0 * order_by_val / total[order_by])
         subtotal['pct_cuml'] = round(100.0 * total_so_far / total[order_by])
         str_values = [formatters[col].fmt(subtotal[col]) for col in columns]
         padded = zip(widths, str_values)
@@ -529,6 +526,11 @@ def print_stats_for_pid(pid, stats, full_stacks, syms,
             stack_indent = " " * (len(header) - formatters[syms].width)
             for sym in stack_syms(subtotal[syms]):
                 print(indent + stack_indent + shorten_symbol(sym))
+        # Don't print long tails
+        # TODO: Add flag to print everything?
+        if subtotal['pct_cuml'] == 100 or\
+           formatters[order_by].fmt(order_by_val) == "0":
+            break
 
 
 def to_literal(s):


gdankel:
@brendangregg Yes you're right, I definitely wanted to add a straightforward way to run it.

By the way - I view lockstat.py as a post-processing analysis tool, similar to creating flamegraphs or what have you. locktrace.py seems to be in line with the rest of the bcc scripts in that it's simple and just dumps the collected data. I wasn't even sure about including lockstat.py for that reason.

I can add the interface you describe by simply making lockstat invoke locktrace by default, and expose tracing options to lockstat. Note that I think the current lockstat.py requires python3, which I prefer to use whenever possible. Let me know if that will be a problem.

There is the question of whether lockstat should write a file by default or not, so that you can use a subsequent invocations of lockstat -r to view the data in different ways. Opinion?

gdankel:
Updated diff to make invoking lockstat more user-friendly, as well as pulled in changes from future patches to increase readability and robustness.

gdankel:
Ping ?

brendangregg:
Thanks, but how many files is this tool now? 

- lockstat.py
- locktrace.c
- locktrace.h
- locktrace.py
- locktrace_dbg.c

You see that this doesn't follow the convention in bcc/tools at all, right?

If it's not technically possible to have this as lockstat.py -- or two files: lockstat.py and lockstat.c -- I don't think it fits into bcc/tools as is.

If we had to put this in bcc, it's looking like it would belong in a tools/lockstat subdirectory, but we'll need to adjust packaging so that it builds correctly in /usr/share/bcc/tools correctly, etc.

At some point people are going to write large tool "projects" that are deserve their own repository. This is already about 2,000 lines...

gdankel:
@brendangregg Yes and it will grow... the more interesting stuff is not part of this patch. Part of the reason for the multiple files is that there's a C++ version of locktrace.py, which shares locktrace.h, locktrace.c and locktrace_dbg.inc.

I *could* reduce the number of files down to one or two of course. Won't reduce the number of lines though, and will make maintenance a pain.

To adhere to the spirit of bcc, I could also provide a stripped-down version that covers the basics, but not sure how much value that adds on top of offcputime.py.

I'll look into the implications of making it a separate repository.

gdankel:
@brendangregg Do you have an example of a separate repository, or a blueprint for how one should be created? This would seem to involve managing dependencies via git submodules or some such?

If there's not an appropriate place in bcc, then probably we will keep it as an internal tool for the time being. I might still push the rest of the patches here in case someone can benefit.

4ast:
I think all bcc scripts have a place in bcc repo.
The only question whether they should be in tools/ and installed by distros or in examples/ and keep maturing there until there is a demand for such tool.

brendangregg:
examples/ is where we ideally have short code examples for newbies. This is a fleshed out tool.

tools/ is for things we want installed by distros, and which has tool documentation.

This is a tool, but it's unlike anything else in tools/ as it has multiple source files without different names. It has what's starting to be a source tree, and should have its own directory. I'd put it in a tools/lockstat subdirectory, or a new projects/lockstat directory.

Yes, I'm tempted to keep asking for a shorter bcc-tools/-like version for tools/, but (so long as it can be maintained -- longer is harder) I don't think most of end users will care about that, so long as we have a final /usr/share/bcc/tools/lockstat that just works like the other ones, and has a man page and examples file.

palmtenor:
@brendangregg We have other tools with stand-alone `.c` file like `deadlock_detector.py`. Is it generally preferable to have Python script in `tools/` to have BPF program rather than stand-alone source code file? Like @gdankel said we can definitely put the `.c` / `.h` / `.inc` files into the Python script altogether, it would definitely make the code harder to read / maintain.

I do agree with you though that end-user probably doesn't care if the script is uglier as long as the tool works.

brendangregg:
I still don't get why this is two different tools, which inflates the file count. I still don't get why this is the "hello world" usage:

```
$ locktrace -p 591905 3 > locks.csv
$ lockstat.py locks.csv
```

and not:

```
lockstat -p 591905 3
```

With -r and -w for file reads and writes if and when people want it. Like tcpdump. Or perf.

I don't think having a toolname.py and toolname.c pair is a problem. Easily understandable: they have the same basename. But when we start to add other files with different basenames (and this can of course get worse in other cases: imagine wanting to add a Makefile, main.c, etc) it'll make a mess of tools/ and should be its own source tree directory. This example has locktrace_dbg.inc -- this would be better as locktrace.inc, since "locktrace_dbg" could be the name of a different tool. But I still don't get why we want two tools in the first place when one should do.

gdankel:
Interesting discussion.

Re "hello world" usage:
The latest update addressed this. It's no longer necessary to do it in two steps, you can just do:

  $ lockstat.py -p 591905 3

Re reason for separate scripts:
1) lockstat.py is only one consumer of the data output by locktrace, even if it's the only one in this patch.
2) Typically I run locktrace once and then lockstat.py multiple times (now using the -i option in the latest update) to specify additional columns, filters etc. to dig around.
3) Doing the analysis on a different machine, comparing traces from different points in time / before and after patches etc. becomes easier, as you've got all the raw data as opposed to one view of it (depending on which options were passed to lockstat.py)
4) locktrace.py is only one of the producers of this data. We also have a C++ version internally, which is the one we actually use here.

gdankel:
Slight correction, the update that addressed the two-step process is here: https://github.com/iovisor/bcc/pull/1697/commits/6f9ace4594fffc7b3b43e5828c1005b902e39410.

gdankel:
How about having lockstats.py in tools/ and the rest of the files in tools/lockstat/ ?

gdankel:
I'm also happy to place it in examples for the time being! Especially since it's not a tool that will work for everyone, e.g. Java and Go users may be disappointed.

brendangregg:
Thanks, but how is this going to work?:

```
# ./lockstat.py -p 9799
  File "./lockstat.py", line 580, in <module>
    p = subprocess.Popen(cmd, stdout=outfile)
  File "/usr/lib/python3.5/subprocess.py", line 947, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.5/subprocess.py", line 1551, in _execute_child
    raise child_exception_type(errno_num, err_msg)
FileNotFoundError: [Errno 2] No such file or directory: '/root/locktrace.py'
```

So it has a dependency on another file? This is sounding like a library, and we have tools/lib/ for that, which becomes /usr/share/bcc/tools/lib when packaged.

A) Is this a possible solution?

- tools/lockstat.py
- tools/lockstat_example.txt
- tools/lib/locktrace.c
- tools/lib/locktrace.h
- tools/lib/locktrace.py
- tools/lib/locktrace.inc

Although currently the tools/lib/ files have the same basename as tools/.

B) Another possible solution?

- tools/lockstat.py
- tools/lockstat_example.txt
- tools/lockstat/locktrace.c
- tools/lockstat/locktrace.h
- tools/lockstat/locktrace.py
- tools/lockstat/locktrace.inc
- tools/lockstat/README.md

Where README.md says:

```
These are libraries for lockstat.

Optionally,  tools/lib/locktrace and tools/lockstat can be run as separate dump & post process steps. For example...
```

...

C) But I'm still wondering if we can have a perf/tcpdump front-end that does it all. What about:

- tools/lockstat.py
- tools/lockstat_example.txt
- tools/lib/lockstat.py
- tools/lib/locktrace.c
- tools/lib/locktrace.h
- tools/lib/locktrace.py
- tools/lib/locktrace.inc

In this case, lockstat.py goes into tools/lib/, and tools/lockstat.py is a new front-end that just does argument processing (including a optional -o outfile, -i infile) and invokes the lib tools as desired. Like a 30 line python program, or shell script. Then you just put the current lockstat/locktrace code into tools/lib, and write a new wrapper that gives us the unified interface.

gdankel:
Yes, lockstat.py still has a dependency on the rest of the files, all I did was add the ability to invoke the tracing script from lockstat.py.

Whether it is a library or not depends on your perspective. For lockstat.py it's a library yes.
But if you e.g.  want to create a flamegraph or similar of longest holders, you'd want the output of locktrace.py.

Therefore I'm leaning towards your option B), or B) combined with the wrapper idea from C).

gdankel:
Thought about a bit more, and if the C) option has the ability to only invoke locktrace on its own then I think this is a good solution. The lib directory seems fine in that case.

OK, so any other opinions? @palmtenor or @4ast?

I vote for this (option C from Brendan):

tools/lockstat.py  -- wrapper for lib/locktrace.py and lib/lockstat.py
tools/lockstat_example.txt
tools/lib/lockstat.py  -- maybe rename this? e.g. locktrace_report.py
tools/lib/locktrace.c
tools/lib/locktrace.h
tools/lib/locktrace.py
tools/lib/locktrace.inc

Alternatively, I'm also OK with examples for now. Please let me know your opinions.

brendangregg:
I think either (C) is good, or (B) with the wrapper.

gdankel:
I've been thinking about how to best do a wrapper.
There are a few options:

1) Perf style
lockstat.py record <options>
lockstat.py report <options>

Pros: Simple wrapper that invokes the respective script and forwards all options. Running perf report multiple times with different options is intuitive. Python versions don't matter (lockstat.py is currently python3).
Cons: Two commands needed to get the report.

2) Combine options using click (http://click.pocoo.org/5/) or similar.

Pros: Can easily group options from different python sub-modules. Sub-modules can also be run as independent tools, and as sub-commands as with 1).
Cons: External dependency. Python versions matter (everything must be executed by the bcc python interpreter).

3) Leave the current lockstat.py in bcc/tools and optionally invoke bcc/tools/lib/locktrace.py, similar to how it works now.

Pros: Simple change. Can get report with single command.
Cons: Some duplication of help messages and argument parsing. Less extensible if we want to add new sub-commands in the future.

1) and 3) are simplest. 2) is a more comprehensive and elegant solution for supporting complex use cases. Maybe something to consider for the future, or for more advanced tools.

joelagnel:
So there's also a 'perf script' for futex contention. I saw this when doing `perf script -l`. Has anyone tried that? Thought of giving that a shot as well.

yonghong-song:
[buildbot, ok to test]

gdankel:
@joelagnel Re perf script, yes I used that in the past. It's very basic, but may still be useful. It's pretty much identical to (probably lifted from) an similar systemtap script.

gdankel:
@joelagnel All good question.

re complexity - I think for many use cases, a tool like offcputime will probably get the job done (with some regexes to look for lock signatures, and filtering out cond_waits etc).
This tool is aimed at in-depth analysis and is probably more similar to Solaris lockstat that way. There are some patches I didn't add here that add to the complexity quite a bit more, which may be bad news if you'd like it to go in the other direction ;)

Re two steps. Yes, I agree, I prefer one step also. If you have an idea of how to achieve that cleanly then please let me know.

Re CSV files - I felt that was a natural fit as the output is basically a table / data frame. But hey, feel free to suggest alternatives ;)

Re sys_us / max_sys_us - two cases that I've seen come immediately to mind: 1) High contention cases with short hold times, e.g. in a key-value store such as memcached. Often the mutex is free by the time the kernel is ready to queue and block the thread, and it'll just return straight out. Other times it'll block for a very short amount of time. If you look at the relationship between sys time and block time it's quite easy to detect this pattern. Error counts are also useful here.
2) A more extreme case of the above, where there's a lot of contention on a lock, maybe with longer hold times. In this case, you can get a lot of time spent in kernel spinlocks since threads are contending over futex queues. This will show up as high sys time and block times could be either low or high.

There are probably other situations where sys time gives you some useful indicator. Also, if there is significant sys_time, and you measure time from futex_enter to futex_exit, then you will over-measure. Maybe that's not a problem most of the time.

You can of course use other tools to drill down into sys time (sampling is the simplest), and combine info from multiple tools to get a more complete picture. The aim of lockstat is to give you that complete picture right away.

Anyway, sys_time is not the main reason the sched_switch handler is there. It is to catch more interesting cases of threads holding locks while going to sleep, tracking nested locks, and help estimating hold times (later patches).

Maybe there should be a simplified version of this tool, but I feel that you can get a long way with offcputime, so it's not clear to me how much value add there would be.

gdankel:
@joelagnel Just to mention also that I started out very simple. It grew in complexity as I encountered more cases in the field. It's an enormously complex area, especially with all the fancy adaptive lock strategies in something like Folly or the Java concurrency library.

joelagnel:
> re complexity - I think for many use cases, a tool like offcputime will probably get the job done (with some regexes to look for lock signatures, and filtering out cond_waits etc).
> This tool is aimed at in-depth analysis and is probably more similar to Solaris lockstat that way. There are some patches I didn't add here that add to the complexity quite a bit more, which may be bad news if you'd like it to go in the other direction ;)
> 
Yes, I was thinking of writing a simpler one called `lockstat` and we get that merged first (probably based on your initial PR or `perf script futex-contention`). But also add lock-owner-blaming to it (which I don't think your PR is doing).

> Re sys_us / max_sys_us - two cases that I've seen come immediately to mind: 1) High contention cases with short hold times, e.g. in a key-value store such as memcached. Often the mutex is free by the time the kernel is ready to queue and block the thread, and it'll just return straight out. Other times it'll block for a very short amount of time. If you look at the relationship between sys time and block time it's quite easy to detect this pattern. Error counts are also useful here.
> 2) A more extreme case of the above, where there's a lot of contention on a lock, maybe with longer hold times. In this case, you can get a lot of time spent in kernel spinlocks since threads are contending over futex queues. This will show up as high sys time and block times could be either low or high.

Ok, that's fair enough. It'll be good if you can document the need for these counters in the examples (if you haven't already)

> Anyway, sys_time is not the main reason the sched_switch handler is there. It is to catch more interesting cases of threads holding locks while going to sleep, tracking nested locks, and help estimating hold times (later patches).

I don't understand this usecase. If a thread is already holding a lock, its in userspace. After that if it sleeps, it sleeps outside of the futex. How do you know if a task is holding a lock when it went to sleep?

Also, what if there was no contention and it went to sleep while holding a lock? How do you know such locks were held?

> Maybe there should be a simplified version of this tool, but I feel that you can get a long way with offcputime, so it's not clear to me how much value add there would be.

I'd think having a dedicated tool that's simple and shows the most common case of contention is still useful without needing the user to post-process output of offcputime. Maybe @brendangregg can comment more on that.

gdankel:
@joelagnel Been very busy, but finally have time to respond to your comments:

Re simpler lockstat - Yes, I've been thinking about the complexity issue and it's definitely more in line with the current bcc scripts to have a very simple tool. If you have use cases in mind and can get good value out of such a tool then who can argue with that? As far as perf script futex-contention goes, that's pretty much as simple as it gets. If you chose to go for something like this then I'd stay clear of tracking max blocking time, as it can be very misleading (context switches can happen during sys_futex, so the max times will frequently be because of that - it will also affect total block time in many cases). Also, it will count all the time not actually blocked as blocked (e.g. high contention, low hold time cases, where most calls to sys_futex just returns without blocking). There are other issues too.

Also keep in mind (and this is separate of complexity, true for any lockstat tool), that I believe a pretty recent kernel is needed, one that supports attaching to sys_futex tracepoints. Using kprobes won't work well (due to the concurrency problem with kretprobes), unless you can find a stable kprobe location for the return path as well (so you can avoid kretprobe).

Re documentation - I'm working on a use-case driven document with real examples that we've been dealing with here. Hopefully I can adapt that to something that can be publicly shared.

Re tracking threads sleeping with a lock. The way this works is as follows: When a thread releases a lock and calls sys_futex(wake), we know whether there are threads waiting on that particular lock right now. If such a thread entered sys_futex *before* the last context switch of the thread that released the lock, and this is the first call to futex wake since then, it follows that the thread holding the lock performed a context switch while holding the lock. Once we know this, we can then record the callstack of that context switch in the stats for the lock.
As you point out, if there is no contention then there is no way of knowing. So there might be some lurking problems that are not captured. However if it's causing actual issues then it's just a matter of collecting enough samples / tracing for long enough to catch it.
Estimating hold times uses some similar logic but is more complex. The great thing about BPF however is that this logic can be applied on-line and drastically reduce the amount of data logged (which is required to keep it low-impact, but also makes post-processing 100x faster).

Re a simple lockstat vs offcputime - I think all that's needed here is some specific real-world examples where you can show clear added value.

gdankel:
To elaborate a bit on the complexity issue: I think the tool I'm working on will follow the classic phases simplistic -> complex -> simple. Right now I'm still in the complex phase as there are plenty of new use cases requiring additions. But over time I see two approaches for simplifying:
1) Remove or simplify logic that turned out to not add much value in most use cases.
2) Extract simpler tools that focus in on one particular use case. One example would be the threads sleeping with lock case mentioned above. That will be more familiar to BCC users as you can create your usual flame graphs etc.

Thanks also to @yonghong-song for offline input.

gdankel:
> Yes, I was thinking of writing a simpler one called lockstat and we get that merged first (probably based on your initial PR or perf script futex-contention). But also add lock-owner-blaming to it (which I don't think your PR is doing) - basically show who is causing the contention, not just who is suffering.

By the way, how do you intend to add lock-owner-blaming? The current patch does it to some extent - you can look at the waker stacks instead of the waiter stack. It doesn't really work that well when you have multiple threads waiting simultaneously for a lock however. I added another mechanism (not in this patch) for tracking that (estimated hold time), without resorting to uprobes. If you have some ideas, please share.

joelagnel:
I agree with you a documentation/usecase driven approach would be nice since your tool is quite complex. I agree it may be exactly what's needed for a comprehensive analysis. But I'm not sure if I need so much detail, as long as I can catch most issues related to poor locking design or lock contention. About looking at waker stacks for lock blaming, could you clarify for me why you feel this isn't sufficient if there are multiple waiters? thanks!

joelagnel:
@gdankel regarding using kprobes in your initial simple code (lockstat-1 branch in your tree), you did this:
```
 bpf_program = BPF(text=bpf_source)
bpf_program.attach_kprobe(event="SyS_futex", fn_name="sys_futex_enter")
bpf_program.attach_kretprobe(event="SyS_futex", fn_name="sys_futex_exit")
```
Now you're saying that kretprobe doesn't work, so does that mean your tool initial didn't work?

gdankel:
@joelagnel Yes, the initial simple code was broken in high contention cases, in that it would miss events. I uploaded it for general comments. There was no way at the time to make it work in a generic way. Before the kernel fix for tracepoints, I used two kprobes, one on entry and one just before the ret instruction in sys_futex, i.e. sys_futes+offset. That works correctly, but the offset changes between kernel versions so that's an extremely hacky and fragile approach.

gdankel:
@joelagnel Re waker stack for lock blaming.
You might have found the problem already. Basically, if you're associating block time of waiting thread T1 with waking thread X then when you have multiple waiters T1,T2,T3, the last waiting will accumulate the most block time and whoever wakes that thread up gets blamed for the the whole amount even when not the culprit. When sorted, the culprit will still appear somewhere near the top but now you don't know which one it is, just a list of possible candidates.

PR #1699, opened 2018-04-23
Add -c option for profile.py

palmtenor:
This gives better feature parity with `perf record`, and it's useful when people want to do fixed-rate sampling.

### DIFF
--- a/tools/profile.py
+++ b/tools/profile.py
@@ -63,6 +63,7 @@ def positive_nonzero_int(val):
 examples = """examples:
     ./profile             # profile stack traces at 49 Hertz until Ctrl-C
     ./profile -F 99       # profile stack traces at 99 Hertz
+    ./profile -c 1000000  # profile stack traces every 1 in a million events
     ./profile 5           # profile at 49 Hertz for 5 seconds only
     ./profile -f 5        # output in folded format for flame graphs
     ./profile -p 185      # only profile threads for PID 185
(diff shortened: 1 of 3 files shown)


tools/profile.py
```diff
@@ -214,10 +218,22 @@ def positive_nonzero_int(val):
 bpf_text = bpf_text.replace('USER_STACK_GET', user_stack_get)
 bpf_text = bpf_text.replace('KERNEL_STACK_GET', kernel_stack_get)
 
+sample_freq = 0
+sample_period = 0
+if args.frequency:
+    sample_freq = args.frequency
+elif args.count:
+    sample_period = args.count
+else:
```
yonghong-song: If both frequency and count are specified, do you know how perf handle it?

brendangregg:
I'm ok with this. I think it will make more sense when profile.py supports other PMC events, and not just CPU cycles. I never want -c count for CPU cycles, but I do want it for other events.

tools/profile.py
palmtenor: The `argparser` makes sure that user can only specify at most one (by using `add_mutually_exclusive_group`). We also have check in `libbpf` (in `bpf_attach_perf_event`) to check only one is specified.

palmtenor:
@brendangregg Agreed. This is probably more useful for lower frequency events such as cache misses etc. I was planning to add other events later, and I just created #1700 so we can discuss there:)

PR #1704, opened 2018-04-24
Refactor offwaketime stack id error handling

palmtenor:
Same change as #1692 but for `offwaketime.py`

### DIFF
--- a/tools/offwaketime.py
+++ b/tools/offwaketime.py
@@ -35,6 +35,11 @@ def positive_nonzero_int(val):
         raise argparse.ArgumentTypeError("must be nonzero")
     return ival
 
+def stack_id_err(stack_id):
+    # -EFAULT in get_stackid normally means the stack-trace is not availible,
+    # Such as getting kernel stack trace in userspace code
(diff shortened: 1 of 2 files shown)


tools/offwaketime.py
```diff
@@ -288,15 +292,20 @@ def signal_ignore(signal, frame):
...
+        has_enomem = has_enomem or (k.w_k_stack_id == -errno.ENOMEM) or \
+                     (k.t_k_stack_id == -errno.ENOMEM)
+    if not args.kernel_stacks_only:
+        missing_stacks += int(stack_id_err(k.w_u_stack_id))
+        missing_stacks += int(stack_id_err(k.t_u_stack_id))
+        has_enomem = has_enomem or (k.w_u_stack_id == -errno.ENOMEM) or \
+                     (k.t_u_stack_id == -errno.ENOMEM)
 
     waker_user_stack = [] if k.w_u_stack_id < 1 else \
         reversed(list(stack_traces.walk(k.w_u_stack_id))[1:])
```
palmtenor: @brendangregg Do you still remember what's reason for doing the `[1:]` (and we don't do it in non-folded case)?

PR #1718, opened 2018-04-29
introduce {attach|detach}_raw_tracepoint API

yonghong-song:
The motivation comes from pull request #1689.
It attached a kprobe bpf program to kernel function
`ttwu_do_wakeup` for more accurate tracing.
Unfortunately, it broke `runqlat.py` in my
4.17 environment since `ttwu_do_wakeup` function
is inlined in my kernel with gcc 7.3.1.

4.17 introduced raw_tracepoint and this patch
added the relevant API to bcc. With this,
we can use tracepoints
`sched:{sched_wakeup, sched_wakeup_new, sched_switch}`
to measure runq latency more reliably.


yonghong-song:
cc @4ast @congwang @palmtenor @brendangregg

bobrik:
I noticed that there's no detach in C API. This is somewhat inconsistent with regular tracepoints:

```c
int bpf_detach_tracepoint(const char *tp_category, const char *tp_name) {
  // Right now, there is nothing to do, but it's a good idea to encourage
  // callers to detach anything they attach.
  return 0;
}
```

Is this intentional?

yonghong-song:
This is due to compatibility with debugfs kprobe/uprobe handling. kprobe/uprobe detachment needs to clean up the `/sys/kernel/debug/tracing/{k|u}probe_events` which is exactly libbpf `bpf_detach_{k|u}probe` is doing. There is nothing to do for tracepoint but the API is there
for API compatibility reasons.

bobrik:
I get why `bpf_detach_tracepoint` exists and it's empty. What I don't get is why there is no empty equivalent for raw tracepoints.

yonghong-song:
This probably comes from different implementation. You can check bcc/src/python/bcc/__init__.py `detach_tracepoint` and `detach_raw_tracepoint`. `detach_tracepoint`, similar to kuprobe, calls into C/C++ code; while `detach_raw_tracepoint` closes the fd in the python code directly.

bobrik:
I see that Python API is similar, but I use C api in Go and that looks different. That seems a bit inconsistent and that's why I was wondering.

Issue #934, opened 2017-02-02
trace.py support for args with Go's gc compiler


brendangregg:
trace's arg# aliases work with Go when compiled with gccgo ("gccgo -o hello hello.go"). But Go's gc compiler ("go build hello.go") uses a different function calling convention, based on the Plan 9 compiler.

Is it important to fix this? Does everyone run gccgo binaries (where trace already works), because they are expected to be faster, and not Go's compiler? I don't know. I've added a prio:low tag for now.

I wrote some quick docs on the issues here: http://www.brendangregg.com/blog/2017-01-31/golang-bcc-bpf-function-tracing.html

And included a proof of concept:

```
# trace '/home/bgregg/functions:main.add "%d %d" go1, go2'
PID    TID    COMM         FUNC             -
17555  17555  functions    main.add         42 13
```

The change I made was a dirty hack, and needs to be rewritten properly. I'd done this:

```diff
# diff -u trace.py trace-go.py 
--- trace.py	2017-02-01 19:34:01.477727396 +0000
+++ trace-go.py	2017-02-02 09:28:23.796606244 +0000
@@ -209,6 +209,8 @@
 
         aliases = {
                 "retval": "PT_REGS_RC(ctx)",
+                "go1": "1;bpf_probe_read(&__data.v0,sizeof(__data.v0),(void *)ctx->sp+8)",
+                "go2": "1;bpf_probe_read(&__data.v1,sizeof(__data.v1),(void *)ctx->sp+16)",
                 "arg1": "PT_REGS_PARM1(ctx)",
                 "arg2": "PT_REGS_PARM2(ctx)",
                 "arg3": "PT_REGS_PARM3(ctx)",
```

They should be "goarg1", "goarg1", etc. I suspect this also needs to be done differently so that they can be used in tests, eg, "(goarg1 == 42)".

I'd also check the following resources to confirm that this approach is valid:

- https://golang.org/doc/asm
- https://9p.io/sys/doc/asm.html

drzaeus77:
If you're looking for datapoints, I used to (go 1.0) rely on gccgo when I needed to link in external C libraries, but nowadays native go is my go-to.

goldshtn:
[...]

Issue #1320, opened 2017-08-25
Go crash with uretprobe


ggaurav10:
I am attempting to get function latencies from a module written in 'Go'. I am using uprobe and uretprobe in a custom eBPF program to track the function entry and exit respectively.
Below is the system config I am working on:
```
4.4.0-81-generic #104~14.04.1-Ubuntu SMP Wed Jun 14 12:45:52 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

The issue is that the Go process crashes because of the conflict in the way return probes work and Go manages stack.
I am interested in knowing if there is any solution to the issue or a workaround for this. Does uretprobe work with Go in any of the later versions, or is this issue fixed in later versions.

4ast:
are you saying that setting a uprobe somewhere in go code makes it fail?
That is certainly a kernel bug. Please describe the steps to reproduce.
If uprobe is responsible it should fail without bcc. Just adding a uprobe
as described here:
https://www.kernel.org/doc/Documentation/trace/uprobetracer.txt
should make it fail.

shashankmjain:
If we replace uretprobe with uprobe in the bcc code, it works. If uretprobe is used it is crashing the process.

brendangregg:
Here's a comment on my go tracing blog post[1] by Suresh Kumar:

> Another problem I ran into: the uretprobe seems to place the return probes by modifying the stack, which is in conflict with how Go manages stack (stacks in Go can grow/shrink at anytime, it does so by copying entire stack to a new larger area, adjusting the pointers in the stack to point to new area etc). So if we are doing a uretprobe, and stack happens to grow (or shrink) at that time, it can lead Go runtime panics. Please see here for an example panic message: https://github.com/surki/misc/blob/master/go.stp#L32-L58

[1] http://www.brendangregg.com/blog/2017-01-31/golang-bcc-bpf-function-tracing.html

4ast:
@brendangregg thanks!
Indeed that's the case.
uprobe works whereas uretprobe makes c++ exception handling fail
```
$ cat throw.cxx
#include <stdexcept>
int foo(int x) { if (x == 1) throw std::runtime_error("oops!"); return x; }
int main(int ac) {
    try { foo(ac); } catch (std::runtime_error&) {}
    return 0;
}
$ clang++ throw.cxx
$ sudo /usr/share/bcc/tools/trace './a.out:_Z3fooi "%d" arg1' &
$ ./a.out
[...]

PR #1719, opened 2018-04-30
Gotrace

nyarly:
This touches on #934 - it adds a rough `gotrace.py` tool (which still would require documentation etc), but handles Go `gc` compiled code by using stack offsets, and manages Go's Pascal-style strings. I'm still not sure how to handle slices or concrete structs, but it seemed reasonable to present this for comment now that it will work for some cases.

The appropriate comparison is to `trace.py` - reviewing what's changed, I _think_ it would be possible to merge most of the changes back into `trace.py` with a few caveats -

* to format Go strings properly, we need to capture the length, and then trim them in Python (the BPF validator doesn't like variable length string copies.) That behavior obviously isn't consistent with C-style null-terminated strings.
* future work should address slices, which also transmit a capacity argument; if anything, they'll need to be formatted differently
* I'm not certain, but I suspect that signatures in the probe definition won't work with Go programs at all.

I think this all could be handled by a flag and conditionals within `trace.py`


nyarly:
Does this make sense, @brendangregg @goldshtn I'd like to make progress on this, but I'd like some sense of what y'all think about this direction. 

If it were to go back into `trace.py`, it occurred to me that, basically, it would need an expansion to the trace definition language - since afaict it doesn't make sense to allow signatures for Go gc traces.

nyarly:
Just a reminder on this PR - @goldshtn @brendangregg

nyarly:
Renewing my request for review, @goldshtn & @brendangregg - I got totally distracted myself, but I'd like to make progress on this

bobrik:
FYI: https://github.com/golang/go/issues/27077

nyarly:
@bobrik I don't follow that issue completely. How does unwinding the stack relate? Is it peculiar to Go?

bobrik:
I was playing with `funclatency` and it broke with my test go program. Tracing of `uretprobes` can get into the same issues.

nyarly:
Do you understand the issue with unwinding the stack? Are uretprobes just fundamentally incompatible with Go programs - seems to me that they sometimes work, but I'm unclear on when or why. At the least,  should we consider if BCC tools can determine if a particular probe will be incompatible?

bobrik:
I don't have enough knowledge to understand the issue with stack unwinding, sorry. My hope is that Go developers can figure this out.

nyarly:
@bobrik This comment seems to summarize the whole thing: https://github.com/iovisor/bcc/issues/1320#issuecomment-407927542

I'm not entirely sure what to do about this for this PR, which is kind of languishing anyway.

Julio-Guerra:
What about making BCC tools ABI-aware?

nyarly:
@Julio-Guerra I'm not sure I understand what you mean?

Julio-Guerra:
> @Julio-Guerra I'm not sure I understand what you mean?

Since, I guess, `trace.py` expects C conventions that are for example false with Go, one way to merge your work into `trace.py` would be to abstract this out so that `trace.py` no longer makes assumptions on the binary interfaces and can therefore be ported to other languages.

nyarly:
The C conventions are pretty universal in Linux, though, as I understand it. Go's "everything goes on the stack directly" is a significant outlier, which also foils uretprobes, unfortunately. It's natural to want to say "report the 3rd argument when called", but that means something totally different to Go than it does to _everything else._ It's different enough that the Go trace version needs not only needs to handle it differently, but also present a different user interface.

brendangregg:
How much delta is there between gotrace and trace? If it's somewhat significant, then I'd go ahead with accepting gotrace into bcc.

I was just rereading Gianluca Borello's experiments with using uprobes on the return points (https://github.com/iovisor/bcc/issues/1320#issuecomment-407927542), and that's something that could be added to gotrace. With such a change, it'd definitely be a significant divergence from trace.

bobek:
This may be of interest -- https://github.com/golang/go/issues/40724 -- they are proposing to switch to register-based calling for go 1.16.

PR #1722, opened 2018-04-30
profile: update --stack-storage-size default value in help message

cippaciong:
While looking at the code of `profile.py` I noticed that the help message for `--stack-storage-size` was not in line with the current default value.

### DIFF
--- a/man/man8/profile.8
+++ b/man/man8/profile.8
@@ -52,7 +52,7 @@ Show stacks from kernel space only (no user space stacks).
 .TP
 \-\-stack-storage-size COUNT
 The maximum number of unique stack traces that the kernel will count (default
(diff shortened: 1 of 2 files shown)


goldshtn:
[buildbot, ok to test]

cippaciong:
Thanks, I wasn't aware of the existence of those magic keywords. I changed the commit to use `%(default)s` instead.

yonghong-song:
One more request, since you are changing this file, could you change the default to 16384 (power-of-2)? We did this implicitly at
https://github.com/iovisor/bcc/blob/master/src/cc/export/helpers.h#L199
since the kernel will always round up to power-or-2.

But it would be good we make default at least power-of-2.

Thanks.

cippaciong:
@yonghong-song done :+1:

PR #1724, opened 2018-05-01
Detect external pointers from context argument

### LINKED ISSUE #1683: Tracepoint tcp_retransmit_skb invalid memory access even with bpf_probe_read
Hi,

I am trying to hook into the tcp/tcp_retransmit_skb tracepoint and extract the sequence number of the retransmitted segment.

```
cat /sys/kernel/debug/tracing/events/tcp/tcp_retransmit_skb/format
name: tcp_retransmit_skb
ID: 1123
format:
	field:unsigned short common_type;	offset:0;	size:2;	signed:0;
	field:unsigned char common_flags;	offset:2;	size:1;	signed:0;
	field:unsigned char common_preempt_count;	offset:3;	size:1;	signed:0;
	field:int common_pid;	offset:4;	size:4;	signed:1;

	field:const void * skbaddr;	offset:8;	size:8;	signed:0;
	field:const void * skaddr;	offset:16;	size:8;	signed:0;
	field:__u16 sport;	offset:24;	size:2;	signed:0;
	field:__u16 dport;	offset:26;	size:2;	signed:0;
	field:__u8 saddr[4];	offset:28;	size:4;	signed:0;
	field:__u8 daddr[4];	offset:32;	size:4;	signed:0;
	field:__u8 saddr_v6[16];	offset:36;	size:16;	signed:0;
	field:__u8 daddr_v6[16];	offset:52;	size:16;	signed:0;

print fmt: "sport=%hu dport=%hu saddr=%pI4 daddr=%pI4 saddrv6=%pI6c daddrv6=%pI6c", REC->sport, REC->dport, REC->saddr, REC->daddr, REC->saddr_v6, REC->daddr_v6
```

Hooking into the tracepoint and access to all variables provided directly by the tracepoint works just fine, but when i try to access the sequence number of the skb i get 

```
0: (79) r6 = *(u64 *)(r1 +8)
1: (79) r3 = *(u64 *)(r1 +16)
2: (07) r3 += 12
3: (bf) r1 = r10
4: (07) r1 += -2
5: (b7) r2 = 2
6: (85) call bpf_probe_read#4
7: (b7) r1 = 2677
8: (6b) *(u16 *)(r10 -8) = r1
9: (18) r1 = 0x25203a74726f7064
11: (7b) *(u64 *)(r10 -16) = r1
12: (b7) r7 = 0
13: (73) *(u8 *)(r10 -6) = r7
14: (69) r3 = *(u16 *)(r10 -2)
15: (bf) r1 = r10
16: (07) r1 += -16
17: (b7) r2 = 11
18: (85) call bpf_trace_printk#6
19: (69) r1 = *(u16 *)(r6 +194)
R6 invalid mem access 'inv'

HINT: The invalid mem access 'inv' error can happen if you try to dereference memory without first using bpf_probe_read() to copy it to the BPF stack. Sometimes the bpf_probe_read is automatic by the bcc rewriter, other times you'll need to be explicit.
```

I have made multiple attempts to solve this using bpf_probe_read, but it does not work.
Here is a minimal (not) working example, including three of the things i have tried. 

example.py
```
#!/usr/bin/python3

from bcc import BPF
import time

def main():
    bpf = None
    with open("example.c", "r") as f:
        bpf = BPF(text=f.read())
    while True:
        time.sleep(1)

if __name__ == '__main__':
    main()
```

example.c

```
// Fix build error because of missing "KBUILD_MODNAME"
#define KBUILD_MODNAME "FOO"

#include <linux/skbuff.h>
#include <linux/tcp.h>
#include <linux/bpf.h>
#include <linux/types.h>
#include <linux/sched.h>
#include <net/tcp.h>

/*
 * From /include/linux/skbuff.h
 */
static inline unsigned char *my_skb_transport_header(const struct sk_buff *skb)
{
    return skb->head + skb->transport_header;
}

/*
 * From /include/linux/tcp.h
 */ 
static inline struct tcphdr *my_tcp_hdr(const struct sk_buff *skb)
{
    return (struct tcphdr *)my_skb_transport_header(skb);
}

TRACEPOINT_PROBE(tcp, tcp_retransmit_skb) {
    struct sk_buff *skb = (struct sk_buff *) args->skbaddr;
    struct sock *sk = (struct sock *) args->skaddr;
    
    // Access to members of the socket struct works
    __u16 dport;
    bpf_probe_read(&dport, sizeof(sk->sk_dport), &sk->sk_dport);
    bpf_trace_printk("dport: %u\n", dport);
    
    // Try #1 - without bpf_probe_read
    ///*
    struct tcphdr *th = my_tcp_hdr(skb);
    bpf_trace_printk("seq: %u\n", th->seq);
    //*/
    
    // Try #2 - with bpf_probe_read for the sequence number
    /*
    u32 seq;
    struct tcphdr *th = my_tcp_hdr(skb);
    bpf_probe_read(&seq, sizeof(th->seq), &th->seq);
    bpf_trace_printk("seq: %u\n", seq);
    */
    
    // Try #3 - with bpf_probe_read for both, the sequence number and the
    // TCP-Header-Pointer itself 
    /*
    struct tcphdr *th;
    u32 seq;
    struct tcphdr *th_tmp = my_tcp_hdr(skb);
    bpf_probe_read(&th, sizeof(th_tmp), &th_tmp);
    bpf_probe_read(&seq, sizeof(th->seq), &th->seq);
    bpf_trace_printk("%u\n", seq);
    */

    return 0;
}
```

Do you know, why bpf_probe_read does not work here? Any help would be appreciated.

Thanks,

lumru

pchaigno:
The bcc rewriter is currently unable to detect external pointers (i.e., to a memory address that requires calls to `bpf_probe_read`) if they are not declared as arguments, e.g., if they are retrieved through the context argument.
For example, although the two following examples translate to the same C code in the end (the bcc rewriter translates the first into the second), the `sk` pointer is recognized as an external pointer only in the first example.
```c
int test1(struct pt_regs *ctx, struct sock *sk) {
    // sk is correctly recognized as an external pointer.
}
int test2(struct pt_regs *ctx) {
    struct sock *sk = (struct sock *)ctx->di;
    // sk is not recognized as an external pointer.
}
```

This pull request fixes that by detecting member dereferences of the context argument (i.e., the first argument of externally visible functions). It also works for the `TRACEPOINT_PROBE` macro and helps fix #1683.

I committed the tests first because I expect both commits to be squashed, and it makes it easier the see the tests fail without the second commit (see CI builds).

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -156,8 +156,9 @@ ProbeVisitor::ProbeVisitor(ASTContext &C, Rewriter &rewriter, set<Decl *> &m) :
 
 bool ProbeVisitor::VisitVarDecl(VarDecl *Decl) {
   if (Expr *E = Decl->getInit()) {
-    if (ProbeChecker(E, ptregs_).is_transitive())
+    if (ProbeChecker(E, ptregs_).is_transitive() || IsContextMemberExpr(E)) {
       set_ptreg(Decl);
+    }
   }
   return true;
 }
@@ -185,7 +186,7 @@ bool ProbeVisitor::VisitBinaryOperator(BinaryOperator *E) {
   if (ProbeChecker(E->getRHS(), ptregs_).is_transitive()) {
     ProbeSetter setter(&ptregs_);
     setter.TraverseStmt(E->getLHS());
-  } else if (E->isAssignmentOp() && E->getRHS()->getStmtClass() == Stmt::CallExprClass) {
+  } else if (E->getRHS()->getStmtClass() == Stmt::CallExprClass) {
     CallExpr *Call = dyn_cast<CallExpr>(E->getRHS());
     if (MemberExpr *Memb = dyn_cast<MemberExpr>(Call->getCallee()->IgnoreImplicit())) {
       StringRef memb_name = Memb->getMemberDecl()->getName();
@@ -204,6 +205,9 @@ bool ProbeVisitor::VisitBinaryOperator(BinaryOperator *E) {
         }
       }
     }
+  } else if (IsContextMemberExpr(E->getRHS())) {
+    ProbeSetter setter(&ptregs_);
+    setter.TraverseStmt(E->getLHS());
   }
   return true;
 }
@@ -260,6 +264,40 @@ bool ProbeVisitor::VisitMemberExpr(MemberExpr *E) {
   rewriter_.InsertText(E->getLocStart(), pre);
   rewriter_.ReplaceText(expansionRange(SourceRange(member, E->getLocEnd())), post);
   return true;
+}
+
+bool ProbeVisitor::IsContextMemberExpr(Expr *E) {
+  if (!E->getType()->isPointerType())
+    return false;
+
+  MemberExpr *Memb = dyn_cast<MemberExpr>(E->IgnoreParenCasts());
+  Expr *base;
+  SourceLocation rhs_start, member;
+  bool found = false;
+  MemberExpr *M;
+  for (M = Memb; M; M = dyn_cast<MemberExpr>(M->getBase())) {
+    memb_visited_.insert(M);
+    rhs_start = M->getLocEnd();
+    base = M->getBase();
+    member = M->getMemberLoc();
+    if (M->isArrow()) {
+      found = true;
+      break;
+    }
+  }
+  if (!found) {
+    return false;
+  }
+  if (member.isInvalid()) {
+    return false;
+  }
+
+  if (DeclRefExpr *base_expr = dyn_cast<DeclRefExpr>(base->IgnoreImplicit())) {
+    if (base_expr->getDecl() == ctx_) {
+      return true;
+    }
+  }
+  return false;
 }
 
 SourceRange
@@ -862,8 +900,11 @@ void BTypeConsumer::HandleTranslationUnit(ASTContext &Context) {
     if (FunctionDecl *F = dyn_cast<FunctionDecl>(D)) {
       if (fe_.is_rewritable_ext_func(F)) {
         for (auto arg : F->parameters()) {
-          if (arg != F->getParamDecl(0) && !arg->getType()->isFundamentalType())
+          if (arg == F->getParamDecl(0)) {
+            probe_visitor_.set_ctx(arg);
+          } else if (!arg->getType()->isFundamentalType()) {
             probe_visitor_.set_ptreg(arg);
+          }
         }
         probe_visitor_.TraverseDecl(D);
       }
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, ok to test]

src/cc/frontends/clang/b_frontend_action.cc
```diff
@@ -185,7 +186,7 @@ bool ProbeVisitor::VisitBinaryOperator(BinaryOperator *E) {
   if (ProbeChecker(E->getRHS(), ptregs_).is_transitive()) {
     ProbeSetter setter(&ptregs_);
     setter.TraverseStmt(E->getLHS());
-  } else if (E->isAssignmentOp() && E->getRHS()->getStmtClass() == Stmt::CallExprClass) {
+  } else if (E->getRHS()->getStmtClass() == Stmt::CallExprClass) {
```
yonghong-song: Could you explain why you remove `E->isAssignmentOp()` here?

src/cc/frontends/clang/b_frontend_action.cc
```diff
@@ -204,6 +205,9 @@ bool ProbeVisitor::VisitBinaryOperator(BinaryOperator *E) {
         }
       }
     }
+  } else if (IsContextMemberExpr(E->getRHS())) {
+    ProbeSetter setter(&ptregs_);
+    setter.TraverseStmt(E->getLHS());
```
yonghong-song: Could you have an example when this piece of code is executed at runtime?

src/cc/frontends/clang/b_frontend_action.cc
pchaigno: Because we're sure `E->isAssignmentOp()` is true when we reach this point (see 6 lines above). I should have commented on that. Sorry.

src/cc/frontends/clang/b_frontend_action.cc
pchaigno: The second example in `test_clang.py` (kprobe example) goes through that path.
```c
// Detected in VisitVarDecl:
struct sock *sk = (struct sock *)args->skaddr;

// Detected in VisitBinaryOp:
struct sock *sk;
sk = (struct sock *)ctx->di;
```

src/cc/frontends/clang/b_frontend_action.cc
yonghong-song: Oh, yes. It is an apparent from the whole source code. Sorry I did not check in the context of the whole file.

src/cc/frontends/clang/b_frontend_action.cc
yonghong-song: Right, this is an assignment binary and the right is a ctx access!

Issue #1723, opened 2018-04-30
bpf_trace_printk() prints garbled text on powerpc64


sandip4n:
As reported by @chandanr, upon running a simple script like the following on a `powerpc64` (big endian) system, the text printed by `bpf_trace_printk()` is garbled.

```
#!/usr/bin/env python

from bcc import BPF

program = r"""
int hello(void *ctx)
{
        bpf_trace_printk("Hello, World!\\n");
        return 0;
}
"""

b = BPF(text=program)
b.attach_kprobe(event="sys_write", fn_name="hello")
print("%-18s %-16s %-6s %s" % ("TIME(s)", "COMM", "PID", "MESSAGE"))

while True:
    try:
        (task, pid, cpu, flags, ts, msg) = b.trace_fields()
    except ValueError:
        continue
    print "hello: %-18.9f %-16s %-6d %s" % (ts, task, pid, msg)
```

The output from this script is:
```
TIME(s)            COMM             PID    MESSAGE
hello: 60275.886506000    bash             9907   W ,olleHdlro
hello: 60368.518992000    !         system -1     W ,olleHdlro
hello: 60421.502999000    !            tmu -9617  W ,olleHdlro
...
```
Also, if either `,` or `!` is removed from the format string, the probe handler is not even invoked.
If similar `bpf_trace_printk()` calls are added to any of the kernel samples (`samples/bpf/*_kern.c`), the output is as expected.
The test system is running Fedora 27 with upstream `net-next` kernel and `bcc`.
There were no problems with running this on `powerpc64le` though.

yonghong-song:
So the indicates, as expected, the issue is in bcc for `powerpc64be`. Maybe you could debug this a little more? `trace_fields` is defined in `src/python/bcc/__init__.py`. It takes the lines directly from
`/sys/kernel/debug/tracing/trace_pipe`. So, you can
 - first verify `/sys/kernel/debug/tracing/trace_pipe` contents are good, and then
 - verify `trace_fields` does right parseing for `powerpc64be`.

Thanks!

sandip4n:
Hi @yonghong-song 

Sorry I forgot to mention this earlier but the `trace_pipe` content is itself garbled.
From what I am seeing in the `bpftool dump xlated` output for the `bcc` script and the modified kernel sample, the string `Hello, World!\n` is pushed on to the stack by a series of stores that write constant values. These constant values seem to be have one or more bytes of the string.

This is the dump from the `bcc` script:
```
[...]

PR #1729, opened 2018-05-03
Fix bpf_trace_printk() for big-endian targets

sandip4n:
The format string for `bpf_trace_printk()` gets garbled during the conversion of the LLVM IR code to the corresponding BPF object. This happens because bcc explicitly sets the data layout of the module to `e-m:e-p:64:64-i64:64-n32:64-S128` which is valid for little-endian targets only. If the arch is set to `bpf`, LLVM selects the correct data layout automatically and there is no need to set this explicitly.

Reported-by: @chandanr
Fixes: #1723


yonghong-song:
[buildbot, ok to test]

yonghong-song:
LGTM. Thanks for the fixing!

PR #1733, opened 2018-05-05
Fix dereference replacements for structure members

pchaigno:
Currently, if a structure member is assigned an external pointer (pointer to kernel address), the pointer to the structure is marked as external instead of the structure member itself. This issue affects all uses of structures with pointers to external addresses: `lp->ptr` is incorrectly replaced with a call to `bpf_probe_read`, while `lp->ptr->val` should be and is not. 

This pull request fixes it by marking the structure member as external.


src/cc/frontends/clang/b_frontend_action.cc
```diff
@@ -255,6 +261,12 @@ bool ProbeVisitor::VisitMemberExpr(MemberExpr *E) {
     error(base->getLocEnd(), "internal error: MemberLoc is invalid while preparing probe rewrite");
     return false;
   }
+
+  // Checks to see if the expression references something that needs to be run
+  // through bpf_probe_read.
+  if (!ProbeChecker(base, ptregs_).needs_probe())
```
pchaigno: Moving this check after we retrieve `base` is necessary the base points to an external address and not the structure member. For example, we want to replace `lp->ext_ptr` with a call to `bpf_probe_read` only if `lp` points to an external address.

tests/python/test_clang.py
```diff
@@ -343,6 +343,34 @@ def test_probe_simple_assign(self):
     return 0;
 }""")
 
+    def test_probe_simple_member_assign(self):
+        b = BPF(text="""
+#include <uapi/linux/ptrace.h>
+#include <linux/netdevice.h>
+struct leaf { void *ptr; };
+int test(struct pt_regs *ctx, struct sk_buff *skb) {
+    u32 pid = bpf_get_current_pid_tgid();
```
yonghong-song: This `u32 pid = ...` is not needed here.

tests/python/test_clang.py
```diff
@@ -343,6 +343,34 @@ def test_probe_simple_assign(self):
...
+}""")
+        b.load_func("test", BPF.KPROBE)
+
+    def test_probe_member_expr(self):
+        b = BPF(text="""
+#include <uapi/linux/ptrace.h>
+#include <linux/netdevice.h>
+struct leaf { struct sk_buff *ptr; };
+int test(struct pt_regs *ctx, struct sk_buff *skb) {
+    u32 pid = bpf_get_current_pid_tgid();
```
yonghong-song: The same as above, `u32 pid = ...` is not needed.

pchaigno:
@yonghong-song Thanks for the review! I fixed the two issues.

PR #1736, opened 2018-05-07
Fix USDT probes arguments' encoding in Python3

javierhonduco:
Running `trace` on a binary's USDT while fetching some arguments (`$ sudo python3 trace.py -p $(pidof ruby) 'u:ruby:array__create "%d", arg1'`) fails with `argument 2: <class 'TypeError'>: wrong type`.

This PR fixes the encoding of the USDT probe name to the `get_probe_arg_ctype`, by encoding it to a bytestring and decoding it afterwards. I've tested this works in Python2 too 😄 

Cheers!

### DIFF
--- a/tools/trace.py
+++ b/tools/trace.py
@@ -342,10 +342,10 @@ BPF_PERF_OUTPUT(%s);
                 if self.probe_type == "u" and expr[0:3] == "arg":
                         arg_index = int(expr[3])
                         arg_ctype = self.usdt.get_probe_arg_ctype(
-                                self.usdt_name.encode('ascii'), arg_index - 1)
+                                self.usdt_name, arg_index - 1)
                         text = ("        %s %s = 0;\n" +
                                 "        bpf_usdt_readarg(%s, ctx, &%s);\n") \
-                                % (arg_ctype.decode(), expr, expr[3], expr)
+                                % (arg_ctype, expr, expr[3], expr)
 
                 if field_type == "s":
                         return text + """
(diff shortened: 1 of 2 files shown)


tools/trace.py
```diff
@@ -342,10 +342,10 @@ def _generate_field_assign(self, idx):
                 if self.probe_type == "u" and expr[0:3] == "arg":
                         arg_index = int(expr[3])
                         arg_ctype = self.usdt.get_probe_arg_ctype(
-                                self.usdt_name, arg_index - 1)
+                                self.usdt_name.encode('ascii'), arg_index - 1)
                         text = ("        %s %s = 0;\n" +
                                 "        bpf_usdt_readarg(%s, ctx, &%s);\n") \
-                                % (arg_ctype, expr, expr[3], expr)
+                                % (arg_ctype.decode(), expr, expr[3], expr)
```
yonghong-song: `encode()` and `decode()` mechanism is not really recommented. See details in #1582.
I think you need to change src/python/bcc/usdt.py and remove `encode()` and `decode()` there
and plus some changes in `trace.py` and you may get it working. Thanks!

javierhonduco: Yeah, makes sense to do it in that file better, thanks!! 

I added a commit that moves the `encode()` and `decode()` there. This way other users of `usdt.py` don't have to worry about the encoding. I've also decided to do it this way because it's how other encoding conversions are handled there. What do you think? (I would squash the commits if this gets approved :smile:)

yonghong-song:
[buildbot, ok to test]

yonghong-song:
LGTM. Thanks! No need to squash on your side, I will squash and merge.

javierhonduco:
Thanks, @yonghong-song!! 💞

PR #1737, opened 2018-05-07
Trace all external pointers passed through a first map

pchaigno:
This pull request removes a limitation of #1285, which added initial support for tracking external pointers through maps. There are still limitations with the tracking of external pointers through maps though; the rewriter is still unable to track external pointers if they are passed as members of a structure itself used as a map value or if they are passed through two successive maps (although the latter should be fairly uncommon).

Currently, `MapVisitor` only detects maps with external pointers as values if the value was directly passed from a function's argument. For example, in the following, the rewriter is currently unable to detect `currsock` has an external pointer as value because an intermediate variable is used instead of passing directly `sk` as the map's value.

```c
int test(struct pt_regs *ctx, struct sock *sk) {
    struct sock **skp = &sk;
    currsock.update(&pid, skp);
    return 0;
};
```

With this pull request, `MapVisitor` is able to trace any external pointer derived from the function's argument and used as a map value. This pull request breaks the `ProbeVisitor` traversal in two distinct traversals. The first rewrites dereferences of external pointers originating from function's arguments and helpers, while the second rewrites only dereferences of external pointers passed through maps. Maps with external pointers as values are identified between the two `ProbeVisitor` traversals.

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -90,8 +90,9 @@ using namespace clang;
 
 class ProbeChecker : public RecursiveASTVisitor<ProbeChecker> {
  public:
-  explicit ProbeChecker(Expr *arg, const set<Decl *> &ptregs)
-      : needs_probe_(false), is_transitive_(false), ptregs_(ptregs) {
+  explicit ProbeChecker(Expr *arg, const set<Decl *> &ptregs, bool track_helpers)
+      : needs_probe_(false), is_transitive_(false), ptregs_(ptregs),
+        track_helpers_(track_helpers) {
     if (arg) {
       TraverseStmt(arg);
       if (arg->getType()->isPointerType())
@@ -100,9 +101,10 @@ class ProbeChecker : public RecursiveASTVisitor<ProbeChecker> {
   }
   bool VisitCallExpr(CallExpr *E) {
     needs_probe_ = false;
-    if (VarDecl *V = dyn_cast<VarDecl>(E->getCalleeDecl())) {
+    if (!track_helpers_)
+      return false;
+    if (VarDecl *V = dyn_cast<VarDecl>(E->getCalleeDecl()))
       needs_probe_ = V->getName() == "bpf_get_current_task";
-    }
     return false;
   }
   bool VisitMemberExpr(MemberExpr *M) {
@@ -123,6 +125,7 @@ class ProbeChecker : public RecursiveASTVisitor<ProbeChecker> {
   bool needs_probe_;
   bool is_transitive_;
   const set<Decl *> &ptregs_;
+  bool track_helpers_;
 };
 
 // Visit a piece of the AST and mark it as needing probe reads
@@ -152,7 +155,7 @@ bool MapVisitor::VisitCallExpr(CallExpr *Call) {
           return true;
 
         if (memb_name == "update" || memb_name == "insert") {
-          if (ProbeChecker(Call->getArg(1), ptregs_).needs_probe()) {
+          if (ProbeChecker(Call->getArg(1), ptregs_, true).needs_probe()) {
             m_.insert(Ref->getDecl());
           }
         }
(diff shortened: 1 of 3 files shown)


yonghong-song:
@pchaigno I did not really understand how useful the below
```
With this pull request, MapVisitor is able to trace any external pointer derived from the function's argument and used as a map value. 
```
It does not seem this helped to reduce the number of manual bpf_probe_read.
Note that in kernel we have
```
        if (arg_type == ARG_PTR_TO_MAP_KEY ||
            arg_type == ARG_PTR_TO_MAP_VALUE) {
                expected_type = PTR_TO_STACK;
                if (!type_is_pkt_pointer(type) && type != PTR_TO_MAP_VALUE &&
                    type != expected_type)
                        goto err_type;
```
So for map key/value,  the data must be on the stack. So there is no bpf_probe_read is needed
for map key/values.

Maybe I missed something?

pchaigno:
According to the kernel lines you cited, for map values/keys, data must be on the stack, from the packet, or from a map value. It used to throw an error for anything else than data coming from the stack, but it changed recently (I added the case for [map values used as map values](https://patchwork.ozlabs.org/patch/903419/)).

In any case, this is unrelated to this pull request. Currently, the bcc rewriter can only identify maps with external pointers if they come directly from the function's arguments. For example, in the following, `currsock` is correctly identified as carrying an external pointer:
```c
int test(struct pt_regs *ctx, struct sock *sk) {
    currsock.update(&pid, &sk);
    return 0;
};
```
However, if we use an intermediate variable, because `ProbeVisitor` runs after `MapVisitor`, external pointers are not tracked beyond function arguments, and the `skp` variable in the following is not identified as an external pointer. Therefore, `currsock` is not identified as a map carrying an external value either.
```c
int test(struct pt_regs *ctx, struct sock *sk) {
    struct sock **skp = &sk;
    currsock.update(&pid, skp);
    return 0;
};
```
Is that clearer? (Sorry, it's complicated enough that I'm having a hard time explaining it in English.)

The sentence you cited on `MapVisitor` is unclear; I'll try to reword it. Anything else that could be improved in the comments?

yonghong-song:
> stack, but it changed recently (I added the case for map values used as map values).

This makes much sense now. I did not pay attention to `type != PTR_TO_MAP_VALUE &&` and still with my old knowledge of key/value must be on stack. Your implementation now makes sense now.
Adding this information (valid map value type) in the comment will be more useful.

I tried the newly added tests and found the number of bpf_probe_read is the same with or without your implementation. It would be good to demonstrate that. If a newer kernel is needed the test case can be annotated with `skipUnless` annotation.

pchaigno:
> I tried the newly added tests and found the number of bpf_probe_read is the same with or without your implementation.

The new `test_ext_ptr_maps_indirect` test should fail without this implementation. I ran it on my system after recompiling/installing the last `master` version of bcc:
```
$ sudo python tmp.py 
Running from kernel directory at: /lib/modules/4.16.0-041600-generic/build
[...]

#include <net/sock.h>
#include <bcc/proto.h>

BPF_HASH(currsock, u32, struct sock *);

__attribute__((section(".bpf.fn.trace_entry")))
int trace_entry(struct pt_regs *ctx) { struct sock *sk = ctx->di;
    struct sock **skp = &sk;
    bpf_map_update_elem((void *)bpf_pseudo_fd(1, 3), &pid, skp, BPF_ANY);
    return 0;
};

__attribute__((section(".bpf.fn.trace_exit")))
int trace_exit(struct pt_regs *ctx) {
    struct sock **skpp;
    skpp = bpf_map_lookup_elem((void *)bpf_pseudo_fd(1, 3), &pid);
    if (skpp) {
        struct sock *skp = *skpp;
        return skp->__sk_common.skc_dport;
    }
    return 0;
}
0: (85) call bpf_get_current_pid_tgid#14
1: (63) *(u32 *)(r10 -4) = r0
2: (18) r1 = 0xffff9796250c2200
4: (bf) r2 = r10
5: (07) r2 += -4
6: (85) call bpf_map_lookup_elem#1
7: (bf) r1 = r0
8: (b7) r0 = 0
9: (15) if r1 == 0x0 goto pc+2
 R0=inv0 R1=map_value(id=0,off=0,ks=4,vs=8,imm=0) R10=fp0,call_-1
10: (79) r1 = *(u64 *)(r1 +0)
 R0=inv0 R1_w=map_value(id=0,off=0,ks=4,vs=8,imm=0) R10=fp0,call_-1
11: (69) r0 = *(u16 *)(r1 +12)
R1 invalid mem access 'inv'

HINT: The invalid mem access 'inv' error can happen if you try to dereference memory without first using bpf_probe_read() to copy it to the BPF stack. Sometimes the bpf_probe_read is automatic by the bcc rewriter, other times you'll need to be explicit.
```

The bcc rewriter fails to replace `skp->__sk_common.skc_dport` and therefore the BPF verifier throws an `invalid mem access`.

> It would be good to demonstrate that. If a newer kernel is needed the test case can be annotated with skipUnless annotation.

This change does not depend on any new kernel version. As I said above, it's unrelated to the checks made by the verifier for map values.

yonghong-song:
@pchaigno Thanks for explanation. Not exactly sure what I did last time for testing. As you mentioned, the test `test_ext_ptr_maps_indirect` demonstrated the new functionality. The code change looks good to me.

PR #1741, opened 2018-05-09
Skip idle swapper threads, closes #1732

### LINKED ISSUE #1732: Exclude swapper kthreads from runqlat
I'm gaining some understanding from working on #1728 and it seems that we should exclude `swapper` kthreads from `runqlat` calculations.

On my idle system I have:

```
         0 -> 1          : 273      |*                                       |
         2 -> 3          : 445      |***                                     |
         4 -> 7          : 2289     |***************                         |
         8 -> 15         : 5908     |****************************************|
        16 -> 31         : 5520     |*************************************   |
        32 -> 63         : 2617     |*****************                       |
        64 -> 127        : 317      |**                                      |
       128 -> 255        : 120      |                                        |
       256 -> 511        : 64       |                                        |
       512 -> 1023       : 27       |                                        |
      1024 -> 2047       : 6        |                                        |
      2048 -> 4095       : 3        |                                        |
      4096 -> 8191       : 1        |                                        |
```

If I filter out `pid == 0 && tgid == 0`:

```
         0 -> 1          : 1        |                                        |
         2 -> 3          : 164      |*                                       |
         4 -> 7          : 1842     |******************                      |
         8 -> 15         : 3843     |**************************************  |
        16 -> 31         : 3946     |****************************************|
        32 -> 63         : 1250     |************                            |
        64 -> 127        : 87       |                                        |
       128 -> 255        : 10       |                                        |
       256 -> 511        : 10       |                                        |
       512 -> 1023       : 8        |                                        |
      1024 -> 2047       : 3        |                                        |
      2048 -> 4095       : 4        |                                        |
```

Total number of events is half of what it was before, since each running task is accompanied by idle state afterwards (with some exceptions when run queue is non-zero).

If this makes sense, I can send a PR.

bobrik:
Run queue latency does not make much sense for idle `swapper` threads.

The same happens in `perf sched`:

* https://github.com/torvalds/linux/blob/v4.14/tools/perf/builtin-sched.c

```c
    /*
     * Ignore idle threads:
     */
    if (!strcmp(thread__comm_str(work_list->thread), "swapper"))
        return;
```

```c
static bool is_idle_sample(struct perf_sample *sample,
	       struct perf_evsel *evsel)
{
    /* pid 0 == swapper == idle task */
    if (strcmp(perf_evsel__name(evsel), "sched:sched_switch") == 0)
        return perf_evsel__intval(evsel, sample, "prev_pid") == 0;

    return sample->pid == 0;
}
```

### DIFF
--- a/tools/runqlat.py
+++ b/tools/runqlat.py
@@ -90,7 +90,7 @@ struct rq;
 // record enqueue timestamp
 static int trace_enqueue(u32 tgid, u32 pid)
 {
-    if (FILTER)
+    if (FILTER || pid == 0)
         return 0;
     u64 ts = bpf_ktime_get_ns();
     start.update(&pid, &ts);
@@ -119,7 +119,7 @@ int trace_run(struct pt_regs *ctx, struct task_struct *prev)
     if (prev->state == TASK_RUNNING) {
         tgid = prev->tgid;
         pid = prev->pid;
-        if (!(FILTER)) {
+        if (!(FILTER || pid == 0)) {
             u64 ts = bpf_ktime_get_ns();
             start.update(&pid, &ts);
         }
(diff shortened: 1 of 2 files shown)


yonghong-song:
Could you also change the places in raw_tracepoint bpf programs in both tools?

bobrik:
@yonghong-song, sure, done.

Issue #1744, opened 2018-05-09
bcc crashes when trying to open a perf buffer in Python 3.5


lumru:
Hi,

The bcc-Binding for Python 3 crashes with Python 3.5 when trying to open a perf buffer.

Here is the StackTrace:

```
BPF._decode_table_type(json.loads(key_desc))

Traceback (most recent call last):
  File "./test.py", line 59, in <module>
    s = subcommand(args)
  File "/home/user/test.py", line 31, in __init__
    print_fn=self.print_event
  File "/home/user/test.py", line 71, in register_handler
    self._bpf[perf_buffer].open_perf_buffer(fn)
  File "/usr/lib64/python3.5/site-packages/bcc/__init__.py", line 461, in __getitem__
    self.tables[key] = self.get_table(key)
  File "/usr/lib64/python3.5/site-packages/bcc/__init__.py", line 451, in get_table
    keytype = BPF._decode_table_type(json.loads(key_desc))
  File "/usr/lib64/python3.5/json/__init__.py", line 312, in loads
    s.__class__.__name__))
TypeError: the JSON object must be str, not 'bytes'
```

In /usr/lib64/python3.5/site-packages/bcc/__init__.py" line 451 json.loads()
tries to load key_desc, which contains bytes. This dosn't work, because json.loads() cannot load
bytes in Python < 3.6, see also:

- https://bugs.python.org/issue10976
- https://bugs.python.org/issue17909

I suggest do add a .decode("utf-8") or .decode("ascii") to fix this, but unfortunately i am not familiar enough with the code to find the right spot to do this, since the same problem might occur in other locations.

Regards,

lumru

yonghong-song:
So the change should be in `__init__.py` function `get_table` right? I think it is okay to have `decode()` here since json.loads expects a string parameter. Could you send a pull request?

lumru:
Done - https://github.com/iovisor/bcc/pull/1746

PR #1746, opened 2018-05-09
Fix issue #1744.

lumru:
- Fix a crash in the python binding when trying to open a perf buffer
    in python < 3.6
- See https://github.com/iovisor/bcc/issues/1744

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -445,12 +445,12 @@ class BPF(object):
         if map_fd < 0:
             raise KeyError
         if not keytype:
-            key_desc = lib.bpf_table_key_desc(self.module, name)
+            key_desc = lib.bpf_table_key_desc(self.module, name).decode("utf-8")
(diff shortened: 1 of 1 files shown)


src/python/bcc/__init__.py
```diff
@@ -445,12 +445,12 @@ def get_table(self, name, keytype=None, leaftype=None, reducer=None):
         if map_fd < 0:
             raise KeyError
         if not keytype:
-            key_desc = lib.bpf_table_key_desc(self.module, name)
+            key_desc = lib.bpf_table_key_desc(self.module, name).decode("utf-8")
```
palmtenor: Does the JSON library always return UTF-8 or it depends on the system's encoding?

lumru:
@palmtenor: The data is not coming from the json library: Its coming from libbcc and going to the json library.

yonghong-song:
LGTM. Thanks!

PR #1748, opened 2018-05-10
Use print function for Python3 compatibility

javierhonduco:
Hi!

After bumping into some minor Python3 incompatibilities that are already fixed upstream (thanks a bunch!! 😄), I decided to have a look to see if there were more! 

This PR replaces the `print` statement with the Python3 printing function. It also adds `from __future__ import print_function` where needed to have the same print semantics in Python2.

Thanks! :)

PS: I also checked scripts that don't have the import, but didn't submit a PR because these cases work fine in Python3, although the semantics are different and it might be needed in the future. These changes are [here](https://github.com/javierhonduco/bcc/commit/cd5fad2454680a853b632afa0f7514d7dad0b49e), let me know if you think it would be worth merging :).

### DIFF
--- a/examples/tracing/bitehist.py
+++ b/examples/tracing/bitehist.py
@@ -13,6 +13,7 @@
 #
 # 15-Aug-2015	Brendan Gregg	Created this.
 
+from __future__ import print_function
 from bcc import BPF
 from time import sleep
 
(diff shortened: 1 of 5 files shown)


yonghong-song:
[buildbot, ok to test]

yonghong-song:
LGTM. Thanks!

Issue #1747, opened 2018-05-09
get_kprobe_functions() omits module functions


markdrayton:
I noticed funccount and friends were unable to attach to NFS functions on my systems. E.g.:

```
$ sudo funccount nfs_opendir
No functions matched by pattern ^nfs_opendir$
```

There are two reasons for this:

1) Module functions are listed in `kallsyms` with an additional `[module_name]` column. The parsing code needs to be updated to handle the extra column (e.g. https://github.com/markdrayton/bcc/commit/9f0ab450e73ae299557ee0abfaeffa912dc57c79)

2) #1647 added excluding of symbols outside of `[_stext, _etext]`. It seems module functions are outside of this range:

```
$ sudo egrep '(_[es]text|nfs_opendir)' /proc/kallsyms 
ffffffff81000000 T _stext
ffffffff8196a69e T _etext
ffffffffa01f9df0 t nfs_opendir  [nfs]
```

I'm not sure how the second issue should be solved. @yonghong-song, any suggestions?

yonghong-song:
Thanks for reporting. I think we can just exclude functions in init sections (__init_begin, __init_end). This should catch most of cases.
https://github.com/torvalds/linux/blob/4c3579f6cadd5eb8250a36e789e6df66f660237a/kernel/extable.c#L122

Will submit a patch soon.

PR #1750, opened 2018-05-10
fix get_kprobe_functions

yonghong-song:
Fix issue #1747.

In commit #1647, we excluded all symbols outside [_stext, _etext].
This is incorrect as it excluded module symbols as well.

This patch changed the algorithm to only skip symbols
in init sections [__init_begin, __init_end].


markdrayton:
Looks good to me. Thanks!

```
$ sudo funccount nfs_opendir
Tracing 1 functions for "nfs_opendir"... Hit Ctrl-C to end.
```

PR #1752, opened 2018-05-11
Fix dereference replacements for pointers to pointers

pchaigno:
This is essentially the same fix as #1733, but for external pointers with several levels of indirection instead of structure member dereferencing.

Currently, the bcc rewriter is unable to track external pointers if there is more than a single level of indirection (e.g., pointer to external pointer).  For example, in the following, the rewriter is unable to detect that `ptr2` doesn't need a call to bpf_probe_read, only `*ptr2` do.
```c
int test(struct pt_regs *ctx, struct sock *sk) {
    struct sock *ptr1;
    struct sock **ptr2 = &ptr1;
    *ptr2 = sk;
    return ((struct sock *)(*ptr2))->sk_daddr;
}
```
This pull request fixes this issue by tracking the levels of indirections in addition to the variable declarations (identifies each variable).  When traversing dereferences, the level of indirections is used to decide whether the base expression is an external pointer.  The level of indirections is inherited when a pointer is assigned to a new variable (assignments and function calls).

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -82,6 +82,8 @@ const char **get_call_conv(void) {
 using std::map;
 using std::move;
 using std::set;
+using std::tuple;
+using std::make_tuple;
 using std::string;
 using std::to_string;
 using std::unique_ptr;
@@ -90,15 +92,19 @@ using namespace clang;
 
 class ProbeChecker : public RecursiveASTVisitor<ProbeChecker> {
  public:
-  explicit ProbeChecker(Expr *arg, const set<Decl *> &ptregs, bool track_helpers)
+  explicit ProbeChecker(Expr *arg, const set<tuple<Decl *, int>> &ptregs,
+                        bool track_helpers, bool is_assign)
       : needs_probe_(false), is_transitive_(false), ptregs_(ptregs),
-        track_helpers_(track_helpers) {
+        track_helpers_(track_helpers), nb_derefs_(0), is_assign_(is_assign) {
     if (arg) {
       TraverseStmt(arg);
       if (arg->getType()->isPointerType())
         is_transitive_ = needs_probe_;
     }
   }
+  explicit ProbeChecker(Expr *arg, const set<tuple<Decl *, int>> &ptregs,
+                        bool is_transitive)
+      : ProbeChecker(arg, ptregs, is_transitive, false) {}
   bool VisitCallExpr(CallExpr *E) {
     needs_probe_ = false;
     if (!track_helpers_)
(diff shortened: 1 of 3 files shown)


yonghong-song:
@pchaigno The rewriter is already very complex and I feel for the problem we try to address in this pull request, maybe it is reasonable to require users to write a better code?

For example, for the example code,
```
int test(struct pt_regs *ctx, struct sock *sk) {
    struct sock *ptr1;
    struct sock **ptr2 = &ptr1;
    *ptr2 = sk;
    return ((struct sock *)(*ptr2))->sk_daddr;
}
```
User can just write
```
int test(struct pt_regs *ctx, struct sock *sk) {
     return sk->sk_daddr;
}
```

yonghong-song:
I would expect for normal bpf programming, such pointer reference/dereference manipulation will be really rare.

pchaigno:
Given the current test cases, I understand why you might feel that way. Correctly identifying pointers is however needed to move forward: I opened this pull request in preparation for a larger work (#1733 and #1737 are the two other pieces I needed), which should enable the bcc rewriter to fully support cases such as #1683.

That larger work allows the rewriter to track external pointer through pointer arithmetics and arbitrary casts. In particular, the rewriter will track all external values and not only external pointers anymore.
The issue is that this next pull request currently fails to correctly rewrite some of our tools (e.g., [tcptop.py](https://github.com/iovisor/bcc/blob/55de66b5859176de256759dae303ced9683a838d/tools/tcptop.py#L116)). For example, in the following, if `size` is an external value, `val` is detected as an external pointer, and the rewriter attempts to replace `*val` with a call to `bpf_probe_read`. That doesn't make sense and fails, obviously :-)
```c
(*val) += size;
```
With the current pull request, the rewriter correctly identifies that `size` is assigned to the value pointed to by `val` and that only `**val` would require a call to `bpf_probe_read`.

I'm trying to break my work into smaller, easier to review patches, but I can resubmit this as part of the larger work if that makes more sense.

yonghong-song:
@pchaigno could you add more comments in the code? The algorithm is pretty complex and it is easy to be confused. For example, 
  - `nb_derefs_` is increased in ProbeChecker for UO_Deref, and `nb_deref_` is decreased in ProbeSetter, I guess this is the reason when you tried to check `need_probe()`, you use `-nb_derefs_`.

Comments with examples to show how values nb_derefs_/nb_deref_ change will make the algorithm easy to understand and easy for people to follow the code later.

Thanks!

pchaigno:
@yonghong-song Thanks for the review!
I've pushed an update with a few additional comments. I also reworked `ProbeSetter` to try and make it more natural and renamed a couple variables. If there's anything that still looks obscure or could benefit from more comments, don't hesitate!

yonghong-song:
LGTM.

But we did not solve the previous xfsslower.py failure issue due to innocent user bpf_probe_read. The current rewriter approach is really based on the single function when it comes to compare declarations for probechecker.

In case of multiple functions, declaration and statement traversal from different functions may mix together. Since comparison is based on declaration, some false rewriting may happen. Maybe we need to include some function identifier in ptregs?

PR #1755, opened 2018-05-11
Do not calculate syscall prefix proactively in C++ API

palmtenor:
Currently do calculate the syscall prefix in `BPF::init`, which requires loading `kallsyms` etc. But a lot of times the functionality will not be used. This commit changes that we only calculate the syscall prefix the first time we call `get_syscall_fnname`

Also change to use the `KSym` class directly for better destruct production

### DIFF
--- a/src/cc/api/BPF.cc
+++ b/src/cc/api/BPF.cc
@@ -32,17 +32,13 @@
 #include "common.h"
 #include "libbpf.h"
 #include "perf_reader.h"
+#include "syms.h"
 #include "table_storage.h"
 #include "usdt.h"
 
 #include "BPF.h"
 
 namespace ebpf {
-
-static const char* syscall_prefix[] = {
-    "sys_",
-    "__x64_sys_",
-};
 
 std::string uint_to_hex(uint64_t value) {
   std::stringstream ss;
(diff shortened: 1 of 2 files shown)


src/cc/api/BPF.cc
```diff
@@ -567,8 +549,19 @@ StatusTuple BPF::unload_func(const std::string& func_name) {
 }
 
 std::string BPF::get_syscall_fnname(const std::string& name) {
-  std::string fn_name = syscall_prefix[syscall_prefix_idx_] + name;
-  return std::move(fn_name);
+  if (syscall_prefix_ == nullptr) {
+    KSyms ksym;
+    uint64_t addr;
+
+    if (ksym.resolve_name(nullptr, "sys_bpf", &addr))
```
palmtenor: This is just poorman's `folly::Optional`...

yonghong-song:
LGTM. Thanks!

yonghong-song:
@palmtenor Could you add some commit summary? Current commit has a empty summary and github by default refuses to merge.

palmtenor:
@yonghong-song Done. Sorry for that!

PR #1759, opened 2018-05-15
Add introspection to bcc-tools deb package

bobrik:
cc @yonghong-song

### DIFF
--- a/debian/bcc-tools.install
+++ b/debian/bcc-tools.install
@@ -1,2 +1,3 @@
+usr/share/bcc/introspection/*
 usr/share/bcc/tools/*
 usr/share/bcc/man/*


bobrik:
Local build:

```
# dpkg -c pkg/bcc-tools_0.5.0-2_all.deb | fgrep introspection
drwxr-xr-x root/root         0 2018-05-11 00:06 ./usr/share/bcc/introspection/
-rwxr-xr-x root/root  41104864 2018-05-11 00:06 ./usr/share/bcc/introspection/bps
```

It's weird to see 40MiB binary. Should it be that big?

yonghong-song:
LGTM. Thanks!

yonghong-song:
> It's weird to see 40MiB binary. Should it be that big?

No . It should not be that big. In the beginning, it is only a couple of Mega's. Probably a Makefile issue. Will take a look soon.

PR #1762, opened 2018-05-15
usdt: fail when binary doesn't exist. Fixes #1749

### LINKED ISSUE #1749: trace.py with bad binary path and correct pid
Hi,

While tracing some programs I usually add the program path and the pid: 
```shell
$ sudo /usr/share/bcc/tools/trace -p $(pidof ruby) 'u:ruby:array__create'
```
Unfortunately, if the path is not correct, like in this case, but the PID is correct, this fails with:
```
error in probe 'u:ruby:array__create': unrecognized USDT probe array__create
```

Do you think it would be interesting to check [here](https://github.com/iovisor/bcc/blob/master/src/python/bcc/usdt.py#L131) if the provided `path` actually exists and in that case throwing an error? informing somehow the user?, or checking the pid to verify if its valid and raise an exception if none of them are valid?

Let me know if checking this kind of stuff is out of the scope of the Python library and by default we assume this kind of data is correct!

Thanks!! :) 

cc/ @yonghong-song

javierhonduco:
And add error message to hint if the problem is that the passed binary
path is not absolute or if the binary doesn't exist.

In case the PID is correct:

* but the binary couldn't be found, it will print:

  ```
  HINT: Specified binary doesn't exist.
  [...]
  ```

* but the binary is not absolute:

  ```
  HINT: Binary path should be absolute.
  [...]
  ```

Otherwise, it should keep behaving as before.

Let me know if there are better approaches to do this (first steps in C++ here 😄)!

### DIFF
--- a/src/cc/usdt/usdt.cc
+++ b/src/cc/usdt/usdt.cc
@@ -19,6 +19,7 @@
 #include <unordered_set>
 
 #include <fcntl.h>
+#include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
 
@@ -393,10 +394,19 @@ extern "C" {
 void *bcc_usdt_new_frompid(int pid, const char *path) {
   USDT::Context *ctx;
 
-  if (!path)
+  if (!path) {
     ctx = new USDT::Context(pid);
-  else
+  } else {
+    struct stat buffer;
+    if (strlen(path) >= 1 && path[0] != '/') {
+      fprintf(stderr, "HINT: Binary path should be absolute.\n\n");
+      return nullptr;
+    } else if (stat(path, &buffer) == -1) {
+      fprintf(stderr, "HINT: Specified binary doesn't exist.\n\n");
+      return nullptr;
+    }
     ctx = new USDT::Context(pid, path);
+  }
   if (!ctx->loaded()) {
     delete ctx;
     return nullptr;


yonghong-song:
[buildbot, ok to test]

src/cc/usdt/usdt.cc
```diff
@@ -393,10 +394,15 @@ extern "C" {
 void *bcc_usdt_new_frompid(int pid, const char *path) {
   USDT::Context *ctx;
 
-  if (!path)
+  if (!path) {
     ctx = new USDT::Context(pid);
-  else
+  } else {
+    struct stat buffer;
+    if ((strlen(path) >= 1 && path[0] != '/' ) || stat(path, &buffer) == -1) {
```
yonghong-song: Could you put the error messages here? You can see from `libbpf.c` and `usdt.cc` where a lot of errors are printed out simply with `fprintf(stderr, ...)`. This way, in python, you do not need to check `startwith('/'` or `open(path)` any more, you can just throw an exception since the detailed errors and hints have been printed out.

If you added the error messages in `bcc_usdt_new_frompid`, C++ and lua will be automatically covered.

src/python/bcc/usdt.py
```diff
@@ -129,10 +129,28 @@ def __init__(self, pid=None, path=None):
             self.pid = pid
             if path:
                 self.context = lib.bcc_usdt_new_frompid(pid, path.encode('ascii'))
+                if self.context == None:
+                    if not path.startswith('/'):
```
yonghong-song: The whole `if ... else ...` can be done in `usdt.cc`.

javierhonduco:
@yonghong-song I think it's ready for a second review 😄 

BTW, didn't add the checks to `bcc_usdt_new_frompath`, because I couldn't make `tools/trace.py` work only with the path without a pid. Once I manage to make that work I would create another PR to add the check to that function as well :)

yonghong-song:
LGTM. Thanks!

PR #1763, opened 2018-05-15
link with bpf-static library for bps

yonghong-song:
the issue is reported at #1759.

bps does not need any C++ library functions in bcc.
It only needs libbpf. So link it with bpf-static instead
of bcc-static. This avoids pulling in any C++ module/symbolization/usdt
functions and llvm libraries.

On my local box, the binary size is reduced from ~60MB to 44KB.


yonghong-song:
cc @bobrik Could you help check whether it reduced the bps footprint in your environment?
cc @iamkafai

PR #1765, opened 2018-05-16
Add extra_flag option to bpf_attach_perf_event_raw

palmtenor:
The bpf_attach_perf_event_raw API is designed to provide maximum flexibility for people to use advanced features of Kernel Perf Events with BPF. Some times specifying flags is necessary, such as if we want to use `PERF_FLAG_PID_CGROUP` to profile a container. This commit adds `extra_flag` option to C and C++ interface

I fee bad for the interface change after introducing the interface. However it was not so long time ago so I guess it should be find...:)

### DIFF
--- a/src/cc/api/BPF.cc
+++ b/src/cc/api/BPF.cc
@@ -324,7 +324,8 @@ StatusTuple BPF::attach_perf_event(uint32_t ev_type, uint32_t ev_config,
 
 StatusTuple BPF::attach_perf_event_raw(void* perf_event_attr,
                                        const std::string& probe_func, pid_t pid,
-                                       int cpu, int group_fd) {
+                                       int cpu, int group_fd,
+                                       unsigned long extra_flags) {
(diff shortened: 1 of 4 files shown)


yonghong-song:
@palmtenor do you think this `extra_flags` will be also useful for python scripts using raw_tracepoints? Adding this to python API as well to keep API parity between C++ and python?

palmtenor:
I think this is mainly targeting on the (hardware) Perf Event, not Raw Tracepoint (although it also uses Perf Event interface). We discussed that it might be harder for Python users to fine-tune the `perf_event_attr` struct, and that's not the normal use case in a Python script tool. I still feel the same way, what do you think?

yonghong-song:
Sorry. I messed `raw_tracepoints` with `attach_perf_event`. Okay, let us only add the extra flags in C++ now. We can add to python API later if there is ever a need. Thanks!

PR #1767, opened 2018-05-17
xfsslower: Fix compilation error due to rewriter update

### LINKED ISSUE #1766: xfsslower is broken in the newest bcc
```
$ sudo /usr/share/bcc/tools/xfsslower 10
/virtual/main.c:136:37: error: cannot take the address of an rvalue of type 'typeof(struct dentry *)' (aka 'struct dentry *')
    bpf_probe_read(&de, sizeof(de), &({ typeof(struct dentry *) _val; __builtin_memset(&_val, 0, sizeof(_val)); bpf_probe_read(&_val, sizeof(_val), (u64)&valp->fp->f_path.dentry); _val; }));
                                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This seems related to #622, specifically https://github.com/iovisor/bcc/issues/622#issuecomment-386856427.

There may be more places where this doesn't work. Should we detect this and print a warning instead of breaking existing usage?

cc @pchaigno

pchaigno:
Since #1737, the bcc rewriter is able to track more external pointers going through maps.  xfsslower was relying on the rewriter not being able to replace some dereferences.  This commit takes this
into account and removes two unnecessary calls to `bpf_probe_read`.

Fixes #1766.

/cc @bobrik

### DIFF
--- a/tools/xfsslower.py
+++ b/tools/xfsslower.py
@@ -187,10 +187,7 @@ static int trace_return(struct pt_regs *ctx, int type)
     bpf_get_current_comm(&data.task, sizeof(data.task));
 
     // workaround (rewriter should handle file to d_name in one step):
-    struct dentry *de = NULL;
-    struct qstr qs = {};
-    bpf_probe_read(&de, sizeof(de), &valp->fp->f_path.dentry);
-    bpf_probe_read(&qs, sizeof(qs), (void *)&de->d_name);
+    struct qstr qs = valp->fp->f_path.dentry->d_name;
     if (qs.len == 0)
         return 0;
     bpf_probe_read(&data.file, sizeof(data.file), (void *)qs.name);
--- a/tools/zfsslower.py
+++ b/tools/zfsslower.py
@@ -190,11 +190,7 @@ static int trace_return(struct pt_regs *ctx, int type)
     data.offset = valp->offset;
     bpf_get_current_comm(&data.task, sizeof(data.task));
 
-    // workaround (rewriter should handle file to d_name in one step):
-    struct dentry *de = NULL;
-    struct qstr qs = {};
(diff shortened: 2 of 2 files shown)


bobrik:
I'm concerned that we have other tools with the same issue. Any thoughts on this:

> Should we detect this and print a warning instead of breaking existing usage?

pchaigno:
@bobrik I'm answering in the original issue. Give me a minute to track the appropriate commits I need to reference. Could you double check that this does fix the issue for you too in the meantime?

tools/xfsslower.py
```diff
@@ -187,10 +187,8 @@
     bpf_get_current_comm(&data.task, sizeof(data.task));
 
     // workaround (rewriter should handle file to d_name in one step):
-    struct dentry *de = NULL;
-    struct qstr qs = {};
-    bpf_probe_read(&de, sizeof(de), &valp->fp->f_path.dentry);
-    bpf_probe_read(&qs, sizeof(qs), (void *)&de->d_name);
+    struct dentry *de = valp->fp->f_path.dentry;
+    struct qstr qs = de->d_name;
```
bobrik: You can do just this:

```
struct qstr qs = valp->fp->f_path.dentry->d_name;
```

pchaigno: Oh, yes. Looks like we can :-)
That addresses the above comment too. Thanks!

pchaigno:
Looks like zfsslower has the exact same issue (could have guessed :/). Other tools look fine. I'll try to improve the smoke_tool tests to catch this kind of issues tomorrow.

PR #1768, opened 2018-05-17
Add -d (duration) option to argdist, funclatency and syscount

akilesh88:
Adding "-d" (duration) trace option to argdist, funclatency and syscount functionalities. This could be helpful when tracing hot-paths in kernel for specific duration.

### DIFF
--- a/tools/argdist.py
+++ b/tools/argdist.py
@@ -611,6 +611,8 @@ argdist -I 'kernel/sched/sched.h' \\
                   help="maximum string size to read from char* arguments")
                 parser.add_argument("-i", "--interval", default=1, type=int,
                   help="output interval, in seconds")
+                parser.add_argument("-d", "--duration",
+                  help="total duration of trace, seconds")
                 parser.add_argument("-n", "--number", type=int, dest="count",
                   help="number of outputs")
                 parser.add_argument("-v", "--verbose", action="store_true",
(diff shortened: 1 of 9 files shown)


yonghong-song:
[buildbot, ok to test]

yonghong-song:
Thanks for adding this. `-d` definitely useful for these tools. Could you add `-d` to the manpage as well?

goldshtn:
... and the _example.txt files, please :)

tools/argdist.py
```diff
@@ -611,6 +611,8 @@ def __init__(self):
                   help="maximum string size to read from char* arguments")
                 parser.add_argument("-i", "--interval", default=1, type=int,
                   help="output interval, in seconds")
+                parser.add_argument("-d", "--duration",
+                  help="total duration of trace, seconds")
```
javierhonduco: What do you think on adding here the type, so we don't have to cast it below? (also applies to the other files :) )

tools/argdist.py
```diff
@@ -684,9 +686,11 @@ def _attach(self):
 
         def _main_loop(self):
                 count_so_far = 0
+                seconds = 0
                 while True:
                         try:
-                                sleep(self.args.interval)
+                                sleep(int(self.args.interval))
```
javierhonduco: I think this `int` casting might not be necessary (same for the ones below). argparse should handle that for us already here: https://github.com/iovisor/bcc/pull/1768/files#diff-080a37c8911a2d34cbff11a2f1ca1255R612 😄

tools/argdist.py
akilesh88: Thanks. I have addressed it in new commit.

PR #1772, opened 2018-05-19
Limit dereference rewriter to tracing contexts

pchaigno:
We should only track and rewrite external pointers from the context pointer for tracing programs.  Other types of context pointers point to e.g. packets and do not require a rewrite to a `bpf_probe_read` call.

I should have done this before, in #1724.

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -911,7 +911,16 @@ void BTypeConsumer::HandleTranslationUnit(ASTContext &Context) {
       if (fe_.is_rewritable_ext_func(F)) {
         for (auto arg : F->parameters()) {
           if (arg == F->getParamDecl(0)) {
-            probe_visitor1_.set_ctx(arg);
+            /**
+             * Limit tracing of pointers from context to tracing contexts.
+             * We're whitelisting instead of blacklisting to avoid issues with
+             * existing programs if new context types are added in the future.
+             */
+            string type = arg->getType().getAsString();
+            if (type == "struct pt_regs *" ||
+                type == "struct bpf_raw_tracepoint_args *" ||
+                type.substr(0, 19) == "struct tracepoint__")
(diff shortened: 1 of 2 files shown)


src/cc/frontends/clang/b_frontend_action.cc
```diff
@@ -911,7 +911,11 @@ void BTypeConsumer::HandleTranslationUnit(ASTContext &Context) {
       if (fe_.is_rewritable_ext_func(F)) {
         for (auto arg : F->parameters()) {
           if (arg == F->getParamDecl(0)) {
-            probe_visitor1_.set_ctx(arg);
+            string type = arg->getType().getAsString();
+            if (type == "struct pt_regs *" ||
+                type == "struct bpf_raw_tracepoint_args *" ||
+                type.substr(0, 19) == "struct tracepoint__")
```
pchaigno: This seems a bit fragile. Is there a better way to check the type of the context argument?

yonghong-song: This indeed fragile. The best way is to know the attachment of the external function so based on what kind of attachment it is (`raw_tracepoint`, `tracepoint`, `kuprobe`, or others), we will be able to know whether the first parameter is a ctx or not. Unfortunately, we did not have such information in program load time.

So I think this change sounds good. It may miss some cases if user have `void *ctx` as the argument and later on cast `ctx` to something else, but it won't rewrite incorrectly.

pchaigno: > It may miss some cases if user have void *ctx as the argument and later on cast ctx to something else, but it won't rewrite incorrectly.

I was tempted to blacklist instead of whitelisting for a moment (i.e., do not add ctx pointer if it points to a packet), but that seemed even more fragile, especially when new types of context will be added...

pchaigno:
I updated the branch to fix the failure (I had forgotten that tc program didn't have direct access to packets before 4.7) and add a comment.

PR #1780, opened 2018-05-25
Fix smoke test for tcplife

pchaigno:


yonghong-song:
[buildbot, ok to test]

PR #1782, opened 2018-05-25
 Track all external values of 64 bits instead of pointers only

pchaigno:
The last pull request for the bcc rewriter in my recent series. I've submitted #1724, #1733, and #1752 to be able to submit this one :tada: 

Currently, the bcc rewriter only tracks external pointers if they are actually of type pointer. External pointers may however have other types before being casted to pointers and dereferenced.
For example, the rewriter is unable to recognize that th is a pointer to a kernel address in the following code snippet because `skb->head` and `skb->transport_header` are of types `unsigned char` and `u16` respectively.
```c
int test(struct pt_regs *ctx, struct sock *sk, struct sk_buff *skb) {
  struct tcphdr *th = (struct tcphdr *)(skb->head + skb->transport_header);
  return th->seq;
}
```
This pull request fixes it by verifying variables are size 64 bits instead of verifying their type.

There is a large number of changed lines because I renamed `ptregs` into `extregs`. This pull request initially tracked all external values regardless of their size; that approach works too, but I preferred a more conservative approach. I can separate the actual changes from the rest or even remove the renaming if you think that makes more sense.

With this, #1683 is fully fixed and shouldn't require any manual call to `bpf_probe_read` anymore.

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -92,19 +92,17 @@ using namespace clang;
 
 class ProbeChecker : public RecursiveASTVisitor<ProbeChecker> {
  public:
-  explicit ProbeChecker(Expr *arg, const set<tuple<Decl *, int>> &ptregs,
+  explicit ProbeChecker(Expr *arg, const set<tuple<Decl *, int>> &extregs,
                         bool track_helpers, bool is_assign)
-      : needs_probe_(false), is_transitive_(false), ptregs_(ptregs),
+      : needs_probe_(false), extregs_(extregs),
         track_helpers_(track_helpers), nb_derefs_(0), is_assign_(is_assign) {
     if (arg) {
       TraverseStmt(arg);
-      if (arg->getType()->isPointerType())
-        is_transitive_ = needs_probe_;
     }
   }
   explicit ProbeChecker(Expr *arg, const set<tuple<Decl *, int>> &ptregs,
-                        bool is_transitive)
-      : ProbeChecker(arg, ptregs, is_transitive, false) {}
+                        bool track_helpers)
+      : ProbeChecker(arg, ptregs, track_helpers, false) {}
   bool VisitCallExpr(CallExpr *E) {
     needs_probe_ = false;
     if (!track_helpers_)
@@ -114,8 +112,8 @@ class ProbeChecker : public RecursiveASTVisitor<ProbeChecker> {
     return false;
   }
   bool VisitMemberExpr(MemberExpr *M) {
-    tuple<Decl *, int> pt = make_tuple(M->getMemberDecl(), nb_derefs_);
-    if (ptregs_.find(pt) != ptregs_.end()) {
+    tuple<Decl *, int> ext = make_tuple(M->getMemberDecl(), nb_derefs_);
+    if (extregs_.find(ext) != extregs_.end()) {
       needs_probe_ = true;
       return false;
     }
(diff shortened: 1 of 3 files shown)


src/cc/frontends/clang/b_frontend_action.cc
```diff
@@ -92,19 +92,17 @@ using namespace clang;
...
-  explicit ProbeChecker(Expr *arg, const set<tuple<Decl *, int>> &ptregs,
+  explicit ProbeChecker(Expr *arg, const set<tuple<Decl *, int>> &extregs,
                         bool track_helpers, bool is_assign)
-      : needs_probe_(false), is_transitive_(false), ptregs_(ptregs),
+      : needs_probe_(false), extregs_(extregs),
         track_helpers_(track_helpers), nb_derefs_(0), is_assign_(is_assign) {
     if (arg) {
       TraverseStmt(arg);
-      if (arg->getType()->isPointerType())
-        is_transitive_ = needs_probe_;
```
pchaigno: This is one of the actual changes (i.e., not renaming). Instead of checking the expression is of pointer type here, we now check the result from that expression has size 64 bits, in the code below (where `is_transitive` used to be called).

src/cc/frontends/clang/b_frontend_action.cc
```diff
@@ -216,10 +212,11 @@ ProbeVisitor::ProbeVisitor(ASTContext &C, Rewriter &rewriter,
 
 bool ProbeVisitor::VisitVarDecl(VarDecl *D) {
   if (Expr *E = D->getInit()) {
-    ProbeChecker checker = ProbeChecker(E, ptregs_, track_helpers_, true);
-    if (checker.is_transitive() || IsContextMemberExpr(E)) {
-      tuple<Decl *, int> pt = make_tuple(D, checker.get_nb_derefs());
-      set_ptreg(pt);
+    ProbeChecker checker = ProbeChecker(E, extregs_, track_helpers_, true);
+    if ((checker.needs_probe() && C.getTypeSize(E->getType()) == 64) ||
```
pchaigno: What I mentioned above. Instead of calling `is_transitive()` we check that the expression has size 64 bits. It's easier to do here instead of inside of `ProbeChecker` since it requires a reference to ASTContext and is only necessary in a couple places.

yonghong-song:
> because skb->head and skb->transport_header are of types unsigned char and u16 respectively.

The skb->head type is `unsigned char *`. In most cases, if variable is to be used as a pointer, the type should reflect it. In certain cases, people use `u64` for pointers and later convert it to pointer type and this is esp true for some uapi's. 

But from bcc perspective, if the variable is defined by user and later used as a pointer, we should encourage the variable type to be pointer type for better programming purpose?

Maybe you want to avoid directly comparing to `64`? Instead you could define a macro/const with `sizeof(void *)`?

pchaigno:
> The `skb->head` type is unsigned char *. In most cases, if variable is to be used as a pointer, the type should reflect it.

Oh, right. Only `skb->transport_header` is not a pointer. My bad.

> But from bcc perspective, if the variable is defined by user and later used as a pointer, we should encourage the variable type to be pointer type for better programming purpose?

I'm all for that, but how should we encourage it? Currently, the rewriter just fails to replace dereferences and there's no way for the user to know why.

> Maybe you want to avoid directly comparing to 64? Instead you could define a macro/const with sizeof(void *)?

:+1:

pchaigno:
Hm, actually, the #1683 example fails because of another issue. So it could be that this pull request is only needed if the user explicitly casts to `u64`, in which case we can probably consider it a corner case (I guess that was your point, sorry I'm only seeing this now). I'll look into it and close the pull request if that's the case.

pchaigno:
Closing this as, as far as I can see, it is only needed if the user explicitly casts pointers to u64.

The last call to `bpf_probe_read` in #1683's example is required because the rewriter is still unable to track external pointers through function returns. I'm working on a fix for that.

PR #1783, opened 2018-05-28
Add missing kernel options to Gentoo instructions

ryao:
CONFIG_FUNCTION_TRACER, CONFIG_DEBUG_INFO and CONFIG_KALLSYMS_ALL are
needed by various ${FS}dist tools.

Using zfsdist does not need CONFIG_KALLSYMS_ALL, but it does need the
modules to be unstripped, so special instructions are included as a note
for ZFS users.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -127,12 +127,17 @@ CONFIG_NET_CLS_BPF=m
 CONFIG_NET_ACT_BPF=m
 CONFIG_BPF_JIT=y
 CONFIG_BPF_EVENTS=y
+CONFIG_FUNCTION_TRACER=y
+CONFIG_DEBUG_INFO=y
+CONFIG_KALLSYMS_ALL=y
 ```
 Finally, you can install bcc with:
 ```
 emerge dev-util/bcc
 ```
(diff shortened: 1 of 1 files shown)


pchaigno:
Shouldn't this be documented in the tool rather than in the installation instructions for bcc?

ryao:
@pchaigno By this, do you mean the kernel config options or the Gentoo tip? Checking the kernel config options on an arbitrary system is a pain. We can check if the module is stripped or not, but unless we start doing distribution specific checks, we aren't going to be able to tell them how to fix it. I'd rather not volunteer myself to write those right now.

yonghong-song:
[buildbot, ok to test]

yonghong-song:
@ryao  Any update on this? Is `CONFIG_DEBUG_INFO` really needed?

pchaigno:
> By this, do you mean the kernel config options or the Gentoo tip? Checking the kernel config options on an arbitrary system is a pain.

Sorry, that was unclear. I meant adding these additional requirements to the [tool's documentation](https://github.com/iovisor/bcc/tree/master/man/man8). That's where I'd expect to find such instructions, if there are specific to a couple of tools and not required for all tools running on bcc.

Issue #1730, opened 2018-05-04
tools/wakeuptime.py says "UnicodeDecodeError: 'ascii' codec can't decode byte 0x95 in position 2"


luohao-brian:
[root@ecs-blog2 tools]# ./wakeuptime.py
Tracing blocked time (us) by kernel stack... Hit Ctrl-C to end.
^C
    target:
    ffffffff81a00205 ret_from_fork
    ffffffff810a6a15 kthread
    ffffffff810a1821 worker_thread
    ffffffff8109f700 create_worker
    waker:           kworker/0:1
        5

Traceback (most recent call last):
  File "./wakeuptime.py", line 216, in <module>
    print("    %-16s %s" % ("target:", k.target.decode())
UnicodeDecodeError: 'ascii' codec can't decode byte 0x95 in position 2: ordinal not in range(128)

it seems bpf_probe_read(&key.target, sizeof(key.target), p->comm) gets corrupted string，so decode() reports an error.

My enviornment:

CentOS 7.4 + elrepo kernel-ml 4.16 +  bcc(cf44278)

yonghong-song:
Could you print the byte sequence so that we may get some clue whether this is just a strange process name set by somebody or it is a kernel bug or it is a memory corruption somewhere?

luohao-brian:
Thanks for your respond.

I tried to print out the byte code as well as the string,  and below is my code change and part of the output. I'm pretty sure there is no process with a strange name.


214         else:
215             # print default multi-line stack output
216             #print("    %-16s %s" % ("target:", k.target.decode()))
217             print("    %-16s %s (%s)" % ("target:", k.target, ":".join("{:02x}".format(ord(c)) for c in k.    target)))
218             #for addr in waker_kernel_stack:
219             #    print("    %-16x %s" % (addr, b.ksym(addr)))
220             print("    %-16s %s" % ("waker:", k.waker.decode()))
221             print("        %d\n" % v.value)

    target:          �q�q���kworker/ (80:71:e2:71:04:88:ff:ff:6b:77:6f:72:6b:65:72:2f)
    waker:           swapper/5
        575986

    target:           ()
    waker:           swapper/7
        704986

    target:          ��/����pool (c0:d6:2f:a6:04:88:ff:ff:70:6f:6f:6c)
    waker:           swapper/2
        800744

    target:          @�<����pcscd (40:c5:3c:94:04:88:ff:ff:70:63:73:63:64)
    waker:           pool
        801235

[...]

PR #1784, opened 2018-05-29
let rewriter add code to define CONFIG_CC_STACKPROTECTOR

yonghong-song:
Fix issue #1730

Linux kernel commit 2bc2f688fdf8 ("Makefile: move stack-protector
availability out of Kconfig") moved CONFIG_CC_STACKPROTECTOR
from Kconfig to Makefile. Commit 44c6dc940b19 ("Makefile: introduce
CONFIG_CC_STACKPROTECTOR_AUTO") introduced CONFIG_CC_STACKPROTECTOR_AUTO.

Whether CONFIG_CC_STACKPROTECTOR is defined depends on
CONFIG_CC_STACKPROTECTOR_{AUTO,REGULAR,STRONG}. Since the clang supports
stack-protector, CONFIG_CC_STACKPROTECTOR_AUTO will imply
CONFIG_CC_STACKPROTECTOR for gcc/clang based compilation.

Such changes are introduced in 4.16. For example, the following code
is defined in linux/include/linux/sched.h,
```
        pid_t                           pid;
        pid_t                           tgid;

        /* Canary value for the -fstack-protector GCC feature: */
        unsigned long                   stack_canary;
        /*
         * Pointers to the (original) parent process, youngest child, younger sibling,
         * older sibling, respectively.  (p->father can be replaced with
         * p->real_parent->pid)
         */

        /* Real parent process: */
        struct task_struct __rcu        *real_parent;
```
If kernel config has CONFIG_CC_STACKPROTECTOR_{STRONG,REGULAR,AUTO} defined,
CONFIG_CC_STACKPROTECTOR will be defined in compilation flags by kernel toplevel Makefile.
But since CONFIG_CC_STACKPROTECTOR is not defined in configuration file autoconf.h,
bcc will consider it is not defined. This will cause bcc to access wrong data
in task_struct for any fields after the above stack_canary.

Instead to fix any individual tool, in this patch the bcc rewriter added necessary
macro definition for CONFIG_CC_STACKPROTECTOR in the source code, depending on
CONFIG_CC_STACKPROTECTOR_{AUTO,REGULAR,STRONG}.

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -1057,7 +1057,28 @@ bool BFrontendAction::is_rewritable_ext_func(FunctionDecl *D) {
           (file_name.empty() || file_name == main_path_));
 }
 
+void BFrontendAction::DoMiscWorkAround() {
+  // In 4.16 and later, CONFIG_CC_STACKPROTECTOR is moved out of Kconfig and into
+  // Makefile. It will be set depending on CONFIG_CC_STACKPROTECTOR_{AUTO|REGULAR|STRONG}.
+  // CONFIG_CC_STACKPROTECTOR is still used in various places, e.g., struct task_struct,
+  // to guard certain fields. The workaround here intends to define
+  // CONFIG_CC_STACKPROTECTOR properly based on other configs, so it relieved any bpf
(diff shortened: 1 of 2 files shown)


yonghong-song:
cc @luohao-brian could you try this patch whether it fixed your problem?

PR #1787, opened 2018-05-30
Add support for attaching kprobes at custom offsets

sandip4n:
Currently, `attach_kprobe()` only allows a `kprobe` to be attached to an arch-dependent default location usually in the prologue of the function corresponding to the event. With these changes, one can attach a `kprobe` at a custom offset from the start of the function.

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -794,7 +794,7 @@ static int bpf_attach_tracing_event(int progfd, const char *event_path, int pid,
 }
 
 int bpf_attach_kprobe(int progfd, enum bpf_probe_attach_type attach_type,
-                      const char *ev_name, const char *fn_name)
+                      const char *ev_name, const char *fn_name, uint64_t fn_offset)
 {
   int kfd, pfd = -1;
   char buf[256];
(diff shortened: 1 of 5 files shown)


yonghong-song:
[buildbot, ok to test]

src/cc/libbpf.c
```diff
@@ -794,15 +794,15 @@ static int bpf_attach_tracing_event(int progfd, const char *event_path, int pid,
 }
 
 int bpf_attach_kprobe(int progfd, enum bpf_probe_attach_type attach_type,
-                      const char *ev_name, const char *fn_name)
+                      const char *ev_name, const char *fn_name, uint64_t fn_offset)
```
palmtenor: Can you maybe check **either** function or offset should be provided, not both?

sandip4n: Hi @palmtenor 
Are you saying that for `perf_event_attr`, if `kprobe_addr` is set then `kprobe_func` should be `NULL`? If I am not mistaken, this applies to `kprobe_addr` and we should still be able to use `kprobe_func` and `probe_offset` together.

yonghong-song: Yes, we can either sym+sym_offset or kaddr for kprobe.

palmtenor: Thanks for the explanation! I always thought kprobe is either `kprobe_func` or `kprobe_addr` and `probe_offset` is for uprobe only.

But after reading the implementation, I wonder how would that work. In `trace_event_perf.c`, we pass the same value (i.e., `config2`) as both `addr` and `offset` to `alloc_trace_kprobe` (See https://github.com/torvalds/linux/blob/master/kernel/trace/trace_event_perf.c#L267, via `create_local_trace_kprobe`). Does that mean if we pass `0x1234` it's actually tracing `0x1234 + 0x1234`?

palmtenor: I see, so when name is provided, `config2` is additional offset, otherwise if `config1` is empty, `config2` would be just the address.

Although I understand the possible reasons behind such Kernel side API, I wonder if we could do it a bit better on BCC side. Like maybe we can explicitly separate `addr` and `offset`, although they would still both go to `config2`...

I'm OK with current change as well if you wish.

yonghong-song:
Looks good. Thanks!

PR #1792, opened 2018-06-01
Refactor external pointer assignments

pchaigno:
The code to track assignments of external pointers was scattered between `VisitVarDecl` and `VisitBinaryOperator`. This commit defines a shared `assignsExtPtr` method.

Doing so also fixes a bug as `VisitVarDecl` was missing a case for external pointers retrieved from maps.

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -214,11 +214,56 @@ ProbeVisitor::ProbeVisitor(ASTContext &C, Rewriter &rewriter,
                            set<Decl *> &m, bool track_helpers) :
   C(C), rewriter_(rewriter), m_(m), track_helpers_(track_helpers) {}
 
+bool ProbeVisitor::assignsExtPtr(Expr *E, int *nbAddrOf) {
+  if (IsContextMemberExpr(E)) {
+    *nbAddrOf = 0;
+    return true;
+  }
+
+  ProbeChecker checker = ProbeChecker(E, ptregs_, track_helpers_,
+                                      true);
+  if (checker.is_transitive()) {
+    // The negative of the number of dereferences is the number of addrof.  In
+    // an assignment, if we went through n addrof before getting the external
+    // pointer, then we'll need n dereferences on the left-hand side variable
+    // to get to the external pointer.
+    *nbAddrOf = -checker.get_nb_derefs();
+    return true;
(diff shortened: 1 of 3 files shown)


src/cc/frontends/clang/b_frontend_action.cc
```diff
@@ -214,11 +214,56 @@ ProbeVisitor::ProbeVisitor(ASTContext &C, Rewriter &rewriter,
                            set<Decl *> &m, bool track_helpers) :
   C(C), rewriter_(rewriter), m_(m), track_helpers_(track_helpers) {}
 
+bool ProbeVisitor::assignsExtPtr(Expr *E, int *nbAddrOf) {
```
pchaigno: I'm not sure this is the best API to retrieve the number of addrof. Should we instead return the number with a special value if the assignment isn't that of an external pointer?

pchaigno:
I can't reproduce Jenkins' failure with Ubuntu 16.04 locally and it doesn't look related to this PR's changes.

yonghong-song:
[buildbot, ok to test]

src/cc/frontends/clang/b_frontend_action.cc
yonghong-song: I think this interface is okay.

src/cc/frontends/clang/b_frontend_action.cc
```diff
@@ -214,11 +214,56 @@ ProbeVisitor::ProbeVisitor(ASTContext &C, Rewriter &rewriter,
...
+    if (MemberExpr *Memb = dyn_cast<MemberExpr>(Call->getCallee()->IgnoreImplicit())) {
+      StringRef memb_name = Memb->getMemberDecl()->getName();
+      if (DeclRefExpr *Ref = dyn_cast<DeclRefExpr>(Memb->getBase())) {
+        if (SectionAttr *A = Ref->getDecl()->getAttr<SectionAttr>()) {
+          if (!A->getName().startswith("maps"))
+            return false;
+
+          if (memb_name == "lookup" || memb_name == "lookup_or_init") {
+            if (m_.find(Ref->getDecl()) != m_.end()) {
+            // Retrieved an ext. pointer from a map, mark LHS as ext. pointer.
```
yonghong-song: The comment is off by 2 bytes. But I think we can fix it when the file is touched next time.

PR #1794, opened 2018-06-01
Add critical section tracer

ErickReyesR:
This tool detects code sections where IRQs or preemption are disabled for a user-specified amount of time.

Requires a kernel built with CONFIG_DEBUG_PREEMPT and CONFIG_PREEMPTIRQ_EVENTS

### DIFF
--- a/tools/criticalstat.py
+++ b/tools/criticalstat.py
@@ -1,0 +1,335 @@
+#!/usr/bin/python
+# @lint-avoid-python-3-compatibility-imports
+#
+# criticalstat  Trace long critical sections (IRQs or preemption disabled)
+#               For Linux, uses BCC, eBPF. Requires kernel built with
+#               CONFIG_DEBUG_PREEMPT and CONFIG_PREEMPTIRQ_EVENTS
+#
+# USAGE: criticalstat [-h] [-p] [-i] [-d DURATION]
+#
+# Copyright (c) 2018, Google, Inc.
+# Licensed under the Apache License, Version 2.0 (the "License")
+#
+# By Joel Fernandes <joel@linuxinternals.org>
+#
+
+from __future__ import print_function
+from bcc import BPF
+import argparse
+import ctypes as ct
+import sys
+import subprocess
+import os.path
+
+examples=""
+
+parser = argparse.ArgumentParser(
+    description="Trace long critical sections",
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    epilog=examples)
+
+parser.add_argument("-p", "--preemptoff", action="store_true",
+                    help="Find long sections where preemption was off")
+
+parser.add_argument("-i", "--irqoff", action="store_true",
+                    help="Find long sections where IRQ was off")
+
+parser.add_argument("-d", "--duration", default=100,
+                    help="Duration in uS (microseconds) below which we filter")
+
+args = parser.parse_args()
+
+preemptoff = False
+irqoff = False
+
+if args.irqoff:
+    preemptoff = False
+    irqoff = True
+elif args.preemptoff:
+    preemptoff = True
+    irqoff = False
+
+
+debugfs_path = subprocess.Popen ("cat /proc/mounts | grep -w debugfs" + 
+    " | awk '{print $2}'",
+    shell=True,
+    stdout=subprocess.PIPE).stdout.read().split("\n")[0]
+
+if debugfs_path == "":
+    print("ERROR: Unable to find debugfs mount point");
+    sys.exit(0);
+
+trace_path = debugfs_path + "/tracing/events/preemptirq/";
+
+if (not os.path.exists(trace_path + "irq_disable") or
+   not os.path.exists(trace_path + "irq_enable") or
+   not os.path.exists(trace_path + "preempt_disable") or
+   not os.path.exists(trace_path + "preempt_enable")):
+    print("ERROR: required tracing events are not available\n" + 
+        "Make sure the kernel is built with CONFIG_DEBUG_PREEMPT " + 
+        "and CONFIG_PREEMPTIRQ_EVENTS enabled")
+    sys.exit(0)
+
+bpf_text = """
+#include <uapi/linux/ptrace.h>
+#include <uapi/linux/limits.h>
+#include <linux/sched.h>
+
+extern char _stext[];
+
+enum addr_offs {
+    start_caller_off,
+    start_parent_off,
+    end_caller_off,
+    end_parent_off
+};
+
+struct start_data {
+    u32 addr_offs[2];
+    u64 ts;
+    int idle_skip;
+    int active;
+};
+
+struct data_t {
+    u64 time;
+    u64 stack_id;
+    u32 cpu;
+    u64 id;
+    u32 addrs[4];   /* indexed by addr_offs */
+    char comm[TASK_COMM_LEN];
+};
+
+BPF_STACK_TRACE(stack_traces, 1024);
+BPF_PERCPU_ARRAY(sts, struct start_data, 1);
+BPF_PERCPU_ARRAY(isidle, u64, 1);
+BPF_PERF_OUTPUT(events);
+
+TRACEPOINT_PROBE(power, cpu_idle)
+{
+    int idx = 0;
+    u64 val;
+    struct start_data *stdp, std;
+
+    stdp = sts.lookup(&idx);
+    bpf_probe_read(&std, sizeof(struct start_data), stdp);
+
+    // Mark active sections as that they should be skipped
+
+    // Handle the case CSenter, Ienter, CSexit, Iexit
+    // Handle the case CSenter, Ienter, Iexit, CSexit
+    if (std.active) {
+        std.idle_skip = 1;
+        sts.update(&idx, &std);
+    }
+
+    // Mark CPU as actively within idle or not.
+    if (args->state < 100) {
+        bpf_trace_printk(\"Setting idle\\n\");
+        val = 1;
+        isidle.update(&idx, &val);
+    } else {
+        bpf_trace_printk(\"Resetting idle\\n\");
+        val = 0;
+        isidle.update(&idx, &val);
+    }
+    return 0;
+}
+
+static int in_idle(void)
+{
+     u64 *idlep, idle;
+     int idx = 0;
+
+    // Skip event if we're in idle loop
+    idlep = isidle.lookup(&idx);
+    bpf_probe_read(&idle, sizeof(u64), idlep);
+
+    // Skip if we're in idle loop
+    if (idle)
+        return 1;
+    else
+        return 0;
+}
+
+static void reset_state(void)
+{
+    int idx = 0;
+    struct start_data s = {};
+
+    sts.update(&idx, &s);
+}
+
+TRACEPOINT_PROBE(preemptirq, TYPE_disable)
+{
+    int idx = 0;
+    struct start_data s;
+
+    // Handle the case Ienter, CSenter, CSexit, Iexit
+    // Handle the case Ienter, CSenter, Iexit, CSexit
+    if (in_idle()) {
+        bpf_trace_printk(\"disable: In idle, resetting\\n\");
+        reset_state();
+        return 0;
+    }
+
+    u64 ts = bpf_ktime_get_ns();
+
+    bpf_trace_printk(\"Entered new section, setting time to %lu\\n\",
+        (unsigned long)ts);
+    s.idle_skip = 0;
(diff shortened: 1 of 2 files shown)


ErickReyesR:
@joelagnel

tools/criticalstat.py
```diff
@@ -0,0 +1,335 @@
...
+   not os.path.exists(trace_path + "preempt_disable") or
+   not os.path.exists(trace_path + "preempt_enable")):
+    print("ERROR: required tracing events are not available\n" + 
+        "Make sure the kernel is built with CONFIG_DEBUG_PREEMPT " + 
+        "and CONFIG_PREEMPTIRQ_EVENTS enabled")
+    sys.exit(0)
+
+bpf_text = """
+#include <uapi/linux/ptrace.h>
+#include <uapi/linux/limits.h>
```
yonghong-song: The include of `limits.h` is not needed. It has been included in below `linux/sched.h`.

tools/criticalstat.py
```diff
@@ -0,0 +1,335 @@
...
+#include <linux/sched.h>
+
+extern char _stext[];
```
yonghong-song: This `extern` is not needed.

tools/criticalstat.py
```diff
@@ -0,0 +1,335 @@
...
+
+enum addr_offs {
+    start_caller_off,
+    start_parent_off,
+    end_caller_off,
+    end_parent_off
```
yonghong-song: Do we want capital letters here for enum members? This will make it clear that it is not a variable?

tools/criticalstat.py
```diff
@@ -0,0 +1,335 @@
...
+    u64 val;
+    struct start_data *stdp, std;
+
+    stdp = sts.lookup(&idx);
+    bpf_probe_read(&std, sizeof(struct start_data), stdp);
+
+    // Mark active sections as that they should be skipped
+
+    // Handle the case CSenter, Ienter, CSexit, Iexit
+    // Handle the case CSenter, Ienter, Iexit, CSexit
```
yonghong-song: The comment here is not really helpful. I guess we need better overall comments to describe the approach and the comments here for the specific event transition.

tools/criticalstat.py
```diff
@@ -0,0 +1,335 @@
...
+BPF_PERF_OUTPUT(events);
+
+TRACEPOINT_PROBE(power, cpu_idle)
+{
+    int idx = 0;
+    u64 val;
+    struct start_data *stdp, std;
+
+    stdp = sts.lookup(&idx);
+    bpf_probe_read(&std, sizeof(struct start_data), stdp);
```
yonghong-song: bpf_probe_read may not be needed, you can directly access/reference it values. See below another example.

tools/criticalstat.py
```diff
@@ -0,0 +1,335 @@
...
+
+    // Skip event if we're in idle loop
+    idlep = isidle.lookup(&idx);
+    bpf_probe_read(&idle, sizeof(u64), idlep);
+
+    // Skip if we're in idle loop
+    if (idle)
+        return 1;
+    else
+        return 0;
```
yonghong-song: The above function body can be simplified as:
```
    // Skip event if we're in idle loop
    idlep = isidle.lookup(&idx);
    if (!idlep && *idlep)
        return 1;
    return 0;
```

tools/criticalstat.py
```diff
@@ -0,0 +1,335 @@
...
+    sts.update(&idx, &s);
+}
+
+TRACEPOINT_PROBE(preemptirq, TYPE_disable)
+{
+    int idx = 0;
+    struct start_data s;
+
+    // Handle the case Ienter, CSenter, CSexit, Iexit
+    // Handle the case Ienter, CSenter, Iexit, CSexit
```
yonghong-song: Can we have better comments here?

tools/criticalstat.py
```diff
@@ -0,0 +1,335 @@
...
+    // Handle the case CSenter, Ienter, CSexit, Iexit
+    // Handle the case CSenter, Ienter, Iexit, CSexit
+    if (std.active) {
+        std.idle_skip = 1;
+        sts.update(&idx, &std);
+    }
+
+    // Mark CPU as actively within idle or not.
+    if (args->state < 100) {
+        bpf_trace_printk(\"Setting idle\\n\");
```
yonghong-song: These bpf_trace_printk must be for debugging purpose. They should be removed for the tool.

tools/criticalstat.py
```diff
@@ -0,0 +1,335 @@
...
+    struct data_t data = {};
+
+    if (bpf_get_current_comm(&data.comm, sizeof(data.comm)) == 0) {
+        data.addrs[start_caller_off] = std.addr_offs[start_caller_off];
+        data.addrs[start_parent_off] = std.addr_offs[start_parent_off];
+        data.addrs[end_caller_off] = args->caller_offs;
+        data.addrs[end_parent_off] = args->parent_offs;
+
+        data.id = id;
+        data.stack_id = stack_traces.get_stackid(args, BPF_F_REUSE_STACKID);
```
yonghong-song: Please do not use BPF_F_REUSE_STACKID. just use 0. Reusing stack id may have issues since you have pid_tgid associated with this stack id.

brendangregg:
This is really neat. Needs a man page, and line in README.md.

Just because I won't be the last person to ask... How does it differ to ftrace's irqsoff and preemptoff tracers? Eg:

https://github.com/torvalds/linux/blob/master/Documentation/trace/ftrace.rst#irqsoff
https://github.com/torvalds/linux/blob/master/Documentation/trace/ftrace.rst#preemptoff

I'm sure it can extend capabilities in this area, if it doesn't already.

tools/criticalstat.py
joelagnel: Sorry, I'll explain the algorithm better in comments.

tools/criticalstat.py
joelagnel: AFAIR, not doing `bpf_probe_read` crashed the tool. This was several weeks ago so let me try direct access again and see what happens.

tools/criticalstat.py
joelagnel: Didn't follow, did you mean `if (idlep && *idlep)` ? Also this assumes the `bpf_probe_read` is not needed I think. I will try it and simplify it, thank you for the suggestion

tools/criticalstat.py
joelagnel: Yes that's a stray, I'll get rid of it, sorry and thanks.

tools/criticalstat.py
joelagnel: Yes, it may give wrong results if there's collision :(.

But with your suggestion, what will happen if there is collision? Are we supposed to discard the sample then and not store it in the perf buffer?

tools/criticalstat.py
joelagnel: Sure

tools/criticalstat.py
joelagnel: Ok

joelagnel: Yes, this was a spurious one. In a previous algorithm I was using `_stext` directly. But you're right that its not needed here.

tools/criticalstat.py
joelagnel: sure, makes sense

tools/criticalstat.py
yonghong-song: If there is a collision, setting flag 0 will give a negative stack id (after casting to int type).
User space will then now that the stack is not valid.

An alternative is to use 4.18 introduced `bpf_get_stack()` helper which will give you a list of ip's. and then you can send to user space directly. But since this is only available at 4.18, I guess you have to leave with old mechanism for a while.

joelagnel:
Glad you like it @brendangregg . Thanks for your suggestions and we will work on implementing them and send you another PR. Also thanks @yonghong-song for the review.

About irqsoff/preemptoff tracer in the kernel, the problem with those is that it shows you a "max latency" trace. That means all the older critical sections will be skipped. With criticalstat, it shows you all critical sections below a threshold, not just the maximum one.

The other issue is the irqsoff/preemptoff tracer is susceptible to breakage due to different cpuidle implementations. If the `stop_critical_timings` kernel function isn't called correctly, then it can lead to long critical section timers. However in the `criticalstat` BCC tracer we're using cpuidle trace events to ignore the time spent in idle.

I guess both these problems can be solved eventually in the preemptoff/irqsoff tracer by some redesign.

tools/criticalstat.py
joelagnel: Changed it to:
```
    // Skip event if we're in idle loop
    idlep = isidle.lookup(&idx);
    if (idlep && *idlep)
        return 1;
```
Let me know if that doesn't look right to you.

joelagnel: Err, and ofcourse the return 0; in the idle.
The function now looks like:
```
static int in_idle(void)
{
     u64 *idlep;
     int idx = 0;

    // Skip event if we're in idle loop
    idlep = isidle.lookup(&idx);
    if (idlep && *idlep)
        return 1;
    return 0;
}
```

tools/criticalstat.py
joelagnel: So if I replace that `if (std.active)` with `if (stdp->active)`, I get:
```
R7 invalid mem access 'map_value_or_null'
```
So I need `bpf_probe_read` here. But I agree I do need it in `in_idle` and I changed it based on your suggestion below. I guess accessing a compound structure map value directly is forbidden, but a simpler u64 map value is Ok?

pchaigno: `if (stdp && stdp->active)` should work fine. Note `bpf_probe_read` would return a non-null value if `stdp` is null.

tools/criticalstat.py
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    bpf_trace_printk(\"enable: start lookup\\n\");
+
+    stdp = sts.lookup(&idx);
+    bpf_trace_printk(\"enable: start read\\n\");
+    bpf_probe_read(&std, sizeof(struct start_data), stdp);
+
+    // Handle the case Ienter, Csenter, Iexit, Csexit
+    if (!std.active) {
```
pchaigno: Same as above: the call to `bpf_probe_read` is not needed here with the latest bcc. This can be simplified to `if (stdp && stdp->active)`, which also handles the case when `stdp` is null.

joelagnel: @pchaigno  I wish that worked. The failure I am now seeing is with with the `sts.update` as well in the if clause.
So if I do:
```
    if (stdp && stdp->active) {
        stdp->idle_skip = 1;
        sts.update(&idx, stdp);
    }
```
I get:
```
./criticalstat.py 
0: (bf) r6 = r1
1: (b7) r1 = 0
2: (63) *(u32 *)(r10 -4) = r1
3: (18) r1 = 0xffffffc0439ea000
5: (bf) r2 = r10
6: (07) r2 += -4
7: (85) call 1
8: (bf) r7 = r0
9: (bf) r1 = r10
10: (07) r1 += -32
11: (b7) r2 = 24
12: (bf) r3 = r7
13: (85) call 4
14: (15) if r7 == 0x0 goto pc+11
 R0=inv,min_value=0,max_value=0 R6=ctx R7=map_value(ks=4,vs=24,id=0),min_value=0,max_value=0 R10=fp
15: (61) r1 = *(u32 *)(r7 +20)
16: (15) if r1 == 0x0 goto pc+9
 R0=inv,min_value=0,max_value=0 R1=inv R6=ctx R7=map_value(ks=4,vs=24,id=0),min_value=0,max_value=0 R10=fp
17: (b7) r1 = 1
18: (63) *(u32 *)(r7 +16) = r1
19: (18) r1 = 0xffffffc0439ea000
21: (bf) r2 = r10
22: (07) r2 += -4
23: (bf) r3 = r7
24: (b7) r4 = 0
25: (85) call 2
R3 type=map_value expected=fp

  File "./criticalstat.py", line 252, in <module>
    b = BPF(text=bpf_text)
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 328, in __init__
    self._trace_autoload()
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 1014, in _trace_autoload
    fn = self.load_func(func_name, BPF.TRACEPOINT)
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 368, in load_func
    (func_name, errstr))
Exception: Failed to load BPF program tracepoint__power__cpu_idle: Permission denied
```

joelagnel: Tell you what. I'll finish the docs and post another PR. I would like wrap up the docs this week. Then we can discuss this issue there. Thanks!

PR #1796, opened 2018-06-02
skip probe rewriter for bpf_probe_read()

yonghong-song:
`bpf_probe_read()` is often used to access pointees in bpf programs.
Recent rewriter has become smarter so a lot of `bpf_probe_read()`
can be replaced with simple pointer/member access.

In certain cases, `bpf_probe_read()` is still preferred though.
For example, kernel `net/tcp.h` defined `TCP_SKB_CB` as below
```
  #define TCP_SKB_CB(__skb)	((struct tcp_skb_cb *)&((__skb)->cb[0]))
```
User can use below to access `tcp_gso_size` of a skb data structure.
```
  TCP_SKB_CB(skb)->tcp_gso_size
```
The rewriter will fail as it attempts to rewrite `(__skb)->cb[0]`.

Instead of chasing down to prevent exactly the above pattern,
this patch detects function `bpf_probe_read()` in `ProbeVisitor` and
will skip it so `bpf_probe_read()`'s parameters will not be
rewritten. This can also help other cases where rewriter is not
capable and user used `bpf_probe_read()` as the workaround.

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -242,7 +242,7 @@ bool ProbeVisitor::assignsExtPtr(Expr *E, int *nbAddrOf) {
 
           if (memb_name == "lookup" || memb_name == "lookup_or_init") {
             if (m_.find(Ref->getDecl()) != m_.end()) {
-            // Retrieved an ext. pointer from a map, mark LHS as ext. pointer.
+              // Retrieved an ext. pointer from a map, mark LHS as ext. pointer.
               // Pointers from maps always need a single dereference to get the
               // actual value.  The value may be an external pointer but cannot
               // be a pointer to an external pointer as the verifier prohibits
@@ -269,7 +269,20 @@ bool ProbeVisitor::VisitVarDecl(VarDecl *D) {
   }
   return true;
 }
+
 bool ProbeVisitor::VisitCallExpr(CallExpr *Call) {
+  // Skip bpf_probe_read for the third argument if it is an AddrOf.
+  if (VarDecl *V = dyn_cast<VarDecl>(Call->getCalleeDecl())) {
+    if (V->getName() == "bpf_probe_read" && Call->getNumArgs() >= 3) {
+      const Expr *E = Call->getArg(2)->IgnoreParenCasts();
+      if (const UnaryOperator *UnaryExpr = dyn_cast<UnaryOperator>(E)) {
+        if (UnaryExpr->getOpcode() == UO_AddrOf)
+          return false;
+      }
+      return true;
+    }
+  }
+
   if (FunctionDecl *F = dyn_cast<FunctionDecl>(Call->getCalleeDecl())) {
     if (F->hasBody()) {
       unsigned i = 0;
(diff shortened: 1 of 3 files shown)


src/cc/frontends/clang/b_frontend_action.cc
```diff
@@ -269,7 +269,37 @@ bool ProbeVisitor::VisitVarDecl(VarDecl *D) {
...
+    if (UnaryExpr->getOpcode() == UO_AddrOf)
+      return true;
+  } else if (ExplicitCastExpr *ExplicitExpr = dyn_cast<ExplicitCastExpr>(E)) {
+    return IsAddrOfExpr(ExplicitExpr->getSubExpr());
+  } else if (ImplicitCastExpr *ImplicitExpr = dyn_cast<ImplicitCastExpr>(E)) {
+    return IsAddrOfExpr(ImplicitExpr->getSubExpr());
+  } else if (CStyleCastExpr *CStyleExpr = dyn_cast<CStyleCastExpr>(E)) {
+    return IsAddrOfExpr(CStyleExpr->getSubExpr());
+  }
+  return false;
```
pchaigno: Are the 3 `else if` different from `E->IgnoreParenCasts()`?

src/cc/frontends/clang/b_frontend_action.cc
```diff
@@ -269,7 +269,37 @@ bool ProbeVisitor::VisitVarDecl(VarDecl *D) {
   }
   return true;
 }
+
+bool ProbeVisitor::IsAddrOfExpr(Expr *E) {
```
pchaigno: I had the impression we were going for the `isAddrOfExpr` case (lower case for first letter) for private methods. Is it a coincidence that all private, not-inherited methods are lower case? (I did use an upper case first letter for `IsContextMemberExpr` by mistake.)

src/cc/frontends/clang/b_frontend_action.cc
```diff
@@ -269,7 +269,37 @@ bool ProbeVisitor::VisitVarDecl(VarDecl *D) {
...
+  // Skip bpf_probe_read for the third argument if it is an AddrOf.
+  if (VarDecl *V = dyn_cast<VarDecl>(Call->getCalleeDecl())) {
+    if (V->getName() == "bpf_probe_read") {
+      unsigned i = 0;
+      // check the third argument which is the src pointer.
+      for (auto arg : Call->arguments()) {
+        i++;
+        if (i != 3)
+          continue;
+        return !IsAddrOfExpr(arg);
```
pchaigno: Wouldn't `return !isAddrOfExpr(Call->getArg(2))` be the same (after a check on `Call->getNumArgs()` maybe)?

src/cc/frontends/clang/b_frontend_action.cc
yonghong-song: I actually followed `IsContextMemberExpr`. Looks all other private method indeed starting with lower case. After simplification, I guess I do not need extra function `isAddrOfExpr` since its body becomes so small.

src/cc/frontends/clang/b_frontend_action.cc
yonghong-song: This works too. Thanks for all suggestions. This simplified the code a lot!

PR #1797, opened 2018-06-05
Feat/add stack frames to funcslower

dpayne:
This change adds options to funcslower to print out the user/kernel stacks from slow functions. It also adds an option for outputting the frames in folded format. Most of the code is lifted from the profile script.

I found this script useful when debugging a production issue to see what was the source of a slow database write.

### DIFF
--- a/man/man8/funcslower.8
+++ b/man/man8/funcslower.8
@@ -1,8 +1,8 @@
 .TH funcslower 8  "2017-03-30" "USER COMMANDS"
 .SH NAME
-fileslower \- Trace slow kernel or user function calls.
+funcslower \- Trace slow kernel or user function calls.
 .SH SYNOPSIS
-.B fileslower [\-hf] [\-p PID] [\-U | \-K] [-m MIN_MS] [-u MIN_US] [-a ARGUMENTS] [-T] [-t] [-v] function [function ...]
+.B funcslower [\-hf] [\-p PID] [\-U | \-K] [-m MIN_MS] [-u MIN_US] [-a ARGUMENTS] [-T] [-t] [-v] function [function ...]
 .SH DESCRIPTION
 This script traces a kernel or user function's entry and return points, and
 prints a message when the function's latency exceeded the specified threshold.
--- a/man/man8/funcslower.8
+++ b/man/man8/funcslower.8
@@ -2,7 +2,7 @@
 .SH NAME
 fileslower \- Trace slow kernel or user function calls.
 .SH SYNOPSIS
-.B fileslower [\-h] [\-p PID] [-m MIN_MS] [-u MIN_US] [-a ARGUMENTS] [-T] [-t] [-v] function [function ...]
+.B fileslower [\-hf] [\-p PID] [\-U | \-K] [-m MIN_MS] [-u MIN_US] [-a ARGUMENTS] [-T] [-t] [-v] function [function ...]
 .SH DESCRIPTION
 This script traces a kernel or user function's entry and return points, and
 prints a message when the function's latency exceeded the specified threshold.
@@ -36,6 +36,15 @@ Print a HH:MM:SS timestamp with each entry.
 \-t
 Print a seconds timestamp with each entry, at microsecond resolution.
 .TP
+\-f
+Print output in folded stack format.
+.TP
+\-U
+Show stacks from user space only (no kernel space stacks).
+.TP
+\-K
+Show stacks from kernel space only (no user space stacks).
+.TP
 \-v
 Print the resulting BPF program, for debugging purposes.
 .TP
@@ -61,6 +70,10 @@ Trace both malloc() and free() slower than 10us, in pid 135 only:
 Trace the write syscall and print its first 4 arguments:
 #
 .B funcslower -a 4 SyS_write
+.TP
+Trace opens from libc and print the user and kernel stack frames:
+#
+.B funcslower -UK c:open
 .SH FIELDS
 .TP
 TIME
--- a/tools/funcslower.py
+++ b/tools/funcslower.py
@@ -109,12 +109,8 @@ struct data_t {
 BPF_HASH(entryinfo, u64, struct entry_t);
 BPF_PERF_OUTPUT(events);
 
-#ifdef USER_STACKS
+#if defined(USER_STACKS) || defined(KERNEL_STACKS)
 BPF_STACK_TRACE(stacks, 2048);
-#else
-#ifdef KERNEL_STACKS
-BPF_STACK_TRACE(stacks, 2048);
-#endif
 #endif
 
 static int trace_entry(struct pt_regs *ctx, int id)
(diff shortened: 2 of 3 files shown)


tools/funcslower.py
```diff
@@ -142,6 +165,31 @@
...
+    data.user_stack_id = stacks.get_stackid(ctx, BPF_F_REUSE_STACKID | BPF_F_USER_STACK);
+#endif
+
+#ifdef KERNEL_STACKS
+    data.kernel_stack_id = stacks.get_stackid(ctx, 0 | BPF_F_REUSE_STACKID);
+
+    if (data.kernel_stack_id >= 0) {
+        // populate extras to fix the kernel stack
+        struct pt_regs regs = {};
+        bpf_probe_read(&regs, sizeof(regs), (void *)ctx);
```
pchaigno: Why is this required? Could you use `ctx` directly instead of `regs`?

dpayne: I switched it to use ctx directly. There was no good reason to use it, it was just a copy paste issue.

tools/funcslower.py
```diff
@@ -94,6 +109,14 @@
 BPF_HASH(entryinfo, u64, struct entry_t);
 BPF_PERF_OUTPUT(events);
 
+#ifdef USER_STACKS
+BPF_STACK_TRACE(stacks, 2048);
+#else
+#ifdef KERNEL_STACKS
+BPF_STACK_TRACE(stacks, 2048);
+#endif
+#endif
```
yonghong-song: In the above, you can use `#if defined(USER_STACKS) || defined(KERNEL_STACKS)`.

tools/funcslower.py
```diff
@@ -142,6 +165,28 @@
     data.start_ns = entryp->start_ns;
     data.duration_ns = delta_ns;
     data.retval = PT_REGS_RC(ctx);
+
+#ifdef USER_STACKS
+    data.user_stack_id = stacks.get_stackid(ctx, BPF_F_REUSE_STACKID | BPF_F_USER_STACK);
```
yonghong-song: if user specified pid based filtering, `BPF_F_REUSE_STACKID` can be used. Otherwise, it may cause issues as reusing stack id may cause the stack id associating with a different pid.

tools/funcslower.py
```diff
@@ -142,6 +165,28 @@
...
+        // if ip isn't sane, leave key ips as zero for later checking
+#ifdef CONFIG_RANDOMIZE_MEMORY
+        if (ip > __PAGE_OFFSET_BASE) {
+#else
+        if (ip > PAGE_OFFSET) {
+#endif
+            data.kernel_ip = ip;
+        }
+    }
+#endif
```
yonghong-song: This may not be correct. You can take a look at profile.py.

tools/funcslower.py
```diff
@@ -222,12 +285,51 @@ def args_str(event):
...
+    kernel_stack = []
+    if args.kernel_stack and event.kernel_stack_id > 0:
+        kernel_tmp = stack_traces.walk(event.kernel_stack_id)
+
+        # fix kernel stack
+        for addr in kernel_tmp:
+            kernel_stack.append(addr)
+        # the later IP checking
+        if event.kernel_ip:
+            kernel_stack.insert(0, event.kernel_ip)
```
yonghong-song: I am not sure whether kernel_ip here is really useful or not. It will give you ip inside the last frame. Most other fools do not print out this extra ip. What is your particular concern here?

tools/funcslower_example.txt
```diff
@@ -75,9 +75,38 @@ This shows the first two arguments to __kmalloc -- the first one is the size
...
 would indicate a failure).
 
+# ./funcslower.py -U -u 100 'c:open'
+Tracing function calls slower than 100 us... Ctrl+C to quit.
+COMM           PID    LAT(us)             RVAL FUNC
+ranlib         23857   910.01                6 c:open
+    b'llvm::sys::fs::openFileForRead(llvm::Twine const&, int&, llvm::SmallVectorImpl<char>*)'
+    b'[unknown]'
+    b'[unknown]'
+ar             24259  17460.29                8 c:open
```
yonghong-song: All these examples have `[unknown]`. Could you produce some examples without `[unknown]`? Thanks!

tools/funcslower.py
dpayne: The only reason was to better match what the profile script was doing. Originally I was working off a somewhat old version of profile, which is why it's a little out of sync now. For now I just removed the line, if you think it is good to do I can also port over all the newer changes from the profile script.

tools/funcslower_example.txt
dpayne: I updated the example to show a more real use case. Originally I went with llvm and c:open because I wanted an example everyone could reproduce, but unfortunately most projects are not compiled with framepointers : /

man/man8/funcslower.8
```diff
@@ -2,7 +2,7 @@
 .SH NAME
 fileslower \- Trace slow kernel or user function calls.
```
yonghong-song: since you are touching this file, could you change here `fileslower` to `funcslower`?

man/man8/funcslower.8
```diff
@@ -2,7 +2,7 @@
...
 .SH SYNOPSIS
-.B fileslower [\-h] [\-p PID] [-m MIN_MS] [-u MIN_US] [-a ARGUMENTS] [-T] [-t] [-v] function [function ...]
+.B fileslower [\-hf] [\-p PID] [\-U | \-K] [-m MIN_MS] [-u MIN_US] [-a ARGUMENTS] [-T] [-t] [-v] function [function ...]
```
yonghong-song: The same here, `fileslower` -> `funcslower`?

yonghong-song:
[buildbot, ok to test]

brendangregg:
Thanks @yonghong-song @pchaigno for the review!

Thanks @dpayne, looks useful and looks like it might work. Eg:

```
(root) ~ # ./funcslower.py -U -p 6307 sys_pselect6
Tracing function calls slower than 1 ms... Ctrl+C to quit.
COMM           PID    LAT(ms)             RVAL FUNC
bash-fp        6307    167.43                1 sys_pselect6
    pselect
    rl_read_key
    readline_internal_char
    readline_internal_charloop
    readline_internal
    readline
    yy_readline_get
    yy_readline_get
    yy_getc
    shell_getc
    read_token
    yylex
    yyparse
    parse_command
    read_command
    reader_loop
    main
    __libc_start_main
    [unknown]
bash-fp        6307   1984.37                1 sys_pselect6
    pselect
    rl_read_key
    readline_internal_char
    readline_internal_charloop
    readline_internal
    readline
    yy_readline_get
    yy_readline_get
    yy_getc
    shell_getc
    read_token
    yylex
    yyparse
    parse_command
    read_command
    reader_loop
    main
    __libc_start_main
    [unknown]
[...]
```

(That's a bash shell I compiled with frame pointers.)

I say "might work" since I was expecting do_select() to block, but it doesn't seem to do anything:

```
(root) ~ # ./funcslower.py -U -p 6307 -u 0.001 do_select
Tracing function calls slower than 0.001 us... Ctrl+C to quit.
COMM           PID    LAT(us)             RVAL FUNC
^C
```

Here's the stack trace (via perf record -e cs -g ...):

```
bash-fp  6307 1536602.955635:          1 cs:
            7fffa092e0ec schedule ([kernel.kallsyms])
            7fffa0932671 schedule_hrtimeout_range_clock ([kernel.kallsyms])
            7fffa0932693 schedule_hrtimeout_range ([kernel.kallsyms])
            7fffa028b905 poll_schedule_timeout ([kernel.kallsyms])
            7fffa028c332 do_select ([kernel.kallsyms])
            7fffa028d0c7 core_sys_select ([kernel.kallsyms])
            7fffa028d454 sys_pselect6 ([kernel.kallsyms])
            7fffa0006ae3 do_syscall_64 ([kernel.kallsyms])
            7fffa0a03081 entry_SYSCALL_64_after_hwframe ([kernel.kallsyms])
                  1170e9 pselect (/lib/x86_64-linux-gnu/libc-2.27.so)
                   f2240 rl_read_key (/usr/bin/bash-fp)
                   d5092 readline_internal_char (/usr/bin/bash-fp)
                   d522c readline_internal_charloop (/usr/bin/bash-fp)
                   d5255 readline_internal (/usr/bin/bash-fp)
                   d4c7e readline (/usr/bin/bash-fp)
                   339bd yy_readline_get (/usr/bin/bash-fp)
                   33ad2 yy_readline_get (/usr/bin/bash-fp)
                   33908 yy_getc (/usr/bin/bash-fp)
                   347fc shell_getc (/usr/bin/bash-fp)
                   35d67 read_token (/usr/bin/bash-fp)
                   352ff yylex (/usr/bin/bash-fp)
                   3045b yyparse (/usr/bin/bash-fp)
                   3000e parse_command (/usr/bin/bash-fp)
                   300f4 read_command (/usr/bin/bash-fp)
                   2fd24 reader_loop (/usr/bin/bash-fp)
                   2d7f3 main (/usr/bin/bash-fp)
                   21b97 __libc_start_main (/lib/x86_64-linux-gnu/libc-2.27.so)
```

Ok, I see a clue -- the kretprobe isn't firing:

```
(root) ~ # /usr/share/bcc/tools/trace -p 6274 p::do_select
6274    6274    bash            do_select
6274    6274    bash            do_select
6274    6274    bash            do_select
^C
(root) ~ # /usr/share/bcc/tools/trace -p 6274 r::do_select
^C
```

So this might not be a problem with funcslower, but rather the kernel's kretprobe instrumentation.  I'm testing on Linux 4.15.

You should be able to test this on a stock bash shell -- you just won't get the nice user stacks. Eg:

```
(root) ~ # ./funcslower.py -U -p 6274 sys_pselect6
Tracing function calls slower than 1 ms... Ctrl+C to quit.
COMM           PID    LAT(ms)             RVAL FUNC
bash           6274    158.79                1 sys_pselect6
    pselect
bash           6274   1520.87                1 sys_pselect6
    pselect
bash           6274    149.93                1 sys_pselect6
    pselect
```

If we agree that this is a kernel issue, then funcslower is ok as is (modulus the remaining fixes), and we can chase up the kretprobe issue separately.

@dpayne thanks for the change, and btw, please don't get too carried away adding too many features to these tools. It helps to keep them simple (like Unix -- do one thing and do it well). This particular change looks suitable.

brendangregg:
(oh, looks like the remaining fixes were pushed as I was typing this. thanks!)

yonghong-song:
Thanks for everyone!

PR #1800, opened 2018-06-07
Add the map types for bpf_redirect_map()

lcp:
This patch series adds the new map types, DEVMAP and CPUMAP, for bpf_redirect_map() and updates the reference guide.


yonghong-song:
[buildbot, ok to test]

src/python/bcc/table.py
```diff
@@ -766,3 +777,11 @@ def __len__(self):
 
     def clear(self):
         pass
+
+class DevMap(HashTable):
```
yonghong-song: Should DevMap be an Array instead of HashTable? The same for below CpuMap?

lcp: Right. They should use ArrayBase.

lcp:
I'll update table.py and add an example for CPUMAP in v2.

PR #1810, opened 2018-06-11
[V2] Add two map types for bpf_redirect_map()

lcp:
This patch series adds DEVMAP and CPUMAP which are necessary for bpf_redirect_map().

v2:
  + Use ArrayBase as the base class of DEVMAP and CPUMAP
  + Add an example for CPUMAP and update the reference guide


yonghong-song:
[buildbot, ok to test]

yonghong-song:
LGTM. Thanks!

PR #1811, opened 2018-06-11
Smoke test for tcpdrop

pchaigno:
If I'm not mistaken, this tool requires a kernel >= 4.7 for the `tcp_drop` function (?).

/cc @brendangregg


yonghong-song:
Thanks!

PR #1812, opened 2018-06-11
Fix #1796 and skip probe rewriter for bpf_probe_read

pchaigno:
#1796 stops the whole AST traversal if it meets a `bpf_probe_read` call.  I think the original intent was to simply not rewrite the third argument, so this commit fixes it by remembering the third argument on `bpf_probe_read` call traversals and overriding `dataTraverseStmtPre` to skip the traversal of that
argument when we meet it later.

/cc @yonghong-song

### DIFF
--- a/tests/python/test_clang.py
+++ b/tests/python/test_clang.py
@@ -76,7 +76,7 @@ int count_foo(struct pt_regs *ctx, unsigned long a, unsigned long b) {
         b = BPF(text=text, debug=0)
         fn = b.load_func("count_foo", BPF.KPROBE)
 
-    def test_probe_read3(self):
+    def test_probe_read_whitelist1(self):
         text = """
 #define KBUILD_MODNAME "foo"
 #include <net/tcp.h>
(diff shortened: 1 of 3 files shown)


tests/python/test_clang.py
```diff
@@ -90,6 +90,24 @@ def test_probe_read3(self):
...
+#include <net/tcp.h>
+int count_tcp(struct pt_regs *ctx, struct sk_buff *skb) {
+    // The below define is in net/tcp.h:
+    //    #define TCP_SKB_CB(__skb) ((struct tcp_skb_cb *)&((__skb)->cb[0]))
+    // Note that it has AddrOf in the macro, which will cause current rewriter
+    // failing below statement
+    // return TCP_SKB_CB(skb)->tcp_gso_size;
+    u16 val = 0;
+    bpf_probe_read(&val, sizeof(val), &(TCP_SKB_CB(skb)->tcp_gso_size));
+    return val + skb->protocol;
```
pchaigno: Without this pull request, this second test fails because the AST traversal is stopped after meeting the `bpf_probe_read` call, even though a rewrite is needed for the later expression `skb->protocol`.

pchaigno:
I don't understand the Jenkins' failures. The tests passed locally (Ubuntu 16.04 and 17.10), and they're passing for the Fedora builds...

yonghong-song:
The function `dataTraverseStmtPre` is defined in clang starting from 3.9. So effectively, 3.7 and 3.8 does not have this function defined in clang hence the `ProbeVisitor::dataTraverseStmtPre` will not execute for 3.7/3.8.

Currently, we already track LLVM major numbers in the cmake file. We can also track minor number, and guard reference of  `ProbeVisitor::dataTraverseStmtPre` unless compiler version is >= 3.9.
For 3.7 and 3.8, the compiler is pretty old, I guess it is OK to fall into conservative side that the whole bpf_probe_read specialization is not needed and users are responsible to rewrite their code.

What do you think?

pchaigno:
> For 3.7 and 3.8, the compiler is pretty old, I guess it is OK to fall into conservative side that the whole bpf_probe_read specialization is not needed and users are responsible to rewrite their code.

I agree, though here, we can actually implement the same behavior without `dataTraverseStmtPre`, by overriding `TraverseStmt`. `RecursiveASTVisitor::dataTraverseStmtPre` is just syntactic sugar. I'll take a stab at it this evening or tomorrow.

yonghong-song:
LGTM. Thanks!

PR #1813, opened 2018-06-11
tcplife: fix dport byte order

andihit:
commit 95b3d8c fixed the dport filtering of the kprobes variant by
moving the network byte order to host byte order conversation before the
filtering. Before submitting the perf event the byte order of the dport
was again converted - this commit removes this double conversion.


yonghong-song:
[buildbot, ok to test]

yonghong-song:
[buildbot, ok to test]

Issue #1804, opened 2018-06-08
Question: why bpf_helpers.h declares _license="GPL"?


oriolarcas:
**Disclaimer:** totally in favor of GPL, just curious, but I didn't find any documentation or issues about this.

I noticed that [`bpf_helpers.h`](https://github.com/iovisor/bcc/blob/master/src/cc/export/helpers.h#L208) declares `char _license[4] SEC("license") = "GPL"`. I stumbled upon this because it conflicted with my own identical declaration of `_license`.

**My question is: why GPL?**

One possible answer would be "because eBPF helpers need it". Well, _some_ need a GPL-compatible license, for instance `bpf_probe_read()`. But some not, for instance `bpf_get_prandom_u32()` or `bpf_get_smp_processor_id()` (which is supported in BCC but not documented, I think). Thus, using helpers not necessarily implies requiring GPL compatibility. The [`gpl_only`](https://elixir.bootlin.com/linux/latest/ident/gpl_only) field discriminates which eBPF helpers require it.

A second issue is the license itself. The kernel doesn't require GPL, but GPL-compatibility, which [currently](https://elixir.bootlin.com/linux/latest/source/include/linux/license.h#L4) is defined as GPL, BSD/GPL, MIT/GPL or MPL/GPL. I am not a lawyer, maybe I am wrong here and "GPL" is just fine...

Finally, the helpers header is incompatible with programs that declare their own license, even if it already is GPL.

I'm not sure if it can be inconvenient to license code without a proper warning. BCC itself is not GPL, it is Apache 2.0, so its C headers don't force GPL. Is there some warning in the documentation that I missed? *Is it possible that some users will discover that their eBPF program was licensed under GPL without their knowledge?*

So, tentative suggestions:

- Technically, BCC should allow the eBPF program to select if it needs a license, and in that case which one.
- It would be interesting to document which eBPF helpers require GPL-compatibility, like the awesome [BPF features by Linux kernel version](https://github.com/iovisor/bcc/blob/master/docs/kernel-versions.md).

Anyway, I understand that this is a low priority discussion now, and that it is much more convenient to have eBPF programs that just work.

Thanks!

Oriol

oriolarcas:
**Update:** this is becoming an issue, in my opinion.

I detected the following issues, correct me if I'm wrong:
* It should be an issue that BCC automatically licenses the code under GPL. There are plentiful of eBPF programs that can run without GPL compatibility (48/63 helpers don't require GPL).
[...]

PR #1815, opened 2018-06-12
New macro EBPF_LICENSE to select program license (#1804)

oriolarcas:
The helpers.h header specifies the eBPF program license as 'GPL'. However,
other GPL-compatible licenses are possible (e.g., Dual BSD/GPL) or even
proprietary licenses (e.g., cachetop can run with a proprietary license).

With this commit, the user can specify an EBPF_LICENSE macro:

  b = BPF(text=bpf_text, cflags=['-DEBPF_LICENSE=Custom license'])

Note it supports multiple words and the absence of quotes. If the eBPF
doesn't have a GPL-compatible license and it uses any GPL-only helpers,
the kernel will reject it with:

See all the GPL-compatible license strings at:

https://github.com/torvalds/linux/blob/v4.17/include/linux/module.h#L172

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -270,6 +270,12 @@ bool ProbeVisitor::VisitVarDecl(VarDecl *D) {
   return true;
 }
 
+bool ProbeVisitor::TraverseStmt(Stmt *S) {
+  if (whitelist_.find(S) != whitelist_.end())
+    return true;
+  return RecursiveASTVisitor<ProbeVisitor>::TraverseStmt(S);
+}
+
 bool ProbeVisitor::VisitCallExpr(CallExpr *Call) {
   // Skip bpf_probe_read for the third argument if it is an AddrOf.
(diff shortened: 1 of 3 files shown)


src/cc/export/helpers.h
```diff
@@ -222,7 +222,15 @@ struct _name##_table_t _name = { .max_entries = (_max_entries) }
 #define cursor_advance(_cursor, _len) \
   ({ void *_tmp = _cursor; _cursor += _len; _tmp; })
 
-char _license[4] SEC("license") = "GPL";
+#ifndef EBPF_LICENSE
```
4ast: could you drop letter E, please ?
And please add a testcase.

oriolarcas: Ok.

oriolarcas: Should the compiler warn the user about adding GPL with a #warning? On one hand, it is not nice that the compiler relicenses the code silently. On the other hand, without the warning the behavior is backwards compatible (won't break -Werror=all, for instance). In this case, I think there should be a big notice in the documentation.

yonghong-song: yes, you can put something in documentation about the new mechanism to define a different bpf program license.

Issue #1802, opened 2018-06-07
Weid characters in execsnoop with wrong encoding


SPYFF:
Just installed the latest bcc from source as described in the installation guide Ubuntu section.
So far everything works fine except `execsnoop`: some weird characters with wrong encoding just show up. 
**Example execsnoop (bad)**
![execsnoop](https://i.imgur.com/kvBoHrq.png)

**Example opensnoop (good)**
![opensnoop](https://i.imgur.com/W62MoWo.png)

Ubuntu 18.04 server, LLVM 6.0

yonghong-song:
It could that the regs content coming to bpf program is incorrect for newer kernel versions or some kinds of kernel quirks. Will take a look.

spyff:
Yes, thats it: with 4.14 works as expected

yonghong-song:
Yes, this is a 4.17 x64 specific issue. The following hack would work.
```
diff --git a/tools/execsnoop.py b/tools/execsnoop.py
index 9a66b39..deae241 100755
--- a/tools/execsnoop.py
+++ b/tools/execsnoop.py
@@ -98,11 +98,17 @@ static int submit_arg(struct pt_regs *ctx, void *ptr, struct data_t *data)
     return 0;
 }
 
-int do_sys_execve(struct pt_regs *ctx,
-    const char __user *filename,
-    const char __user *const __user *__argv,
-    const char __user *const __user *__envp)
+int do_sys_execve(struct pt_regs *ctx)
 {
+    struct pt_regs *__ctx = (struct pt_regs *)PT_REGS_PARM1(ctx);
+    const char __user *filename;
+    const char __user *const __user *__argv;
+    const char __user *const __user *__envp;
+
+    bpf_probe_read(&filename, sizeof(filename), &(PT_REGS_PARM1(__ctx)));
+    bpf_probe_read(&__argv, sizeof(__argv), &(PT_REGS_PARM2(__ctx)));
+    bpf_probe_read(&__envp, sizeof(__envp), &(PT_REGS_PARM3(__ctx)));
+
     // create data here and pass to submit_arg to save stack space (#555)
     struct data_t data = {};
     data.pid = bpf_get_current_pid_tgid() >> 32;
```
The hack will not work with earlier kernel versions or other architectures.

The following kernel patch changed the syscall function input parameter definition.
```
commit fa697140f9a20119a9ec8fd7460cc4314fbdaff3
Author: Dominik Brodowski <linux@dominikbrodowski.net>
[...]

PR #1816, opened 2018-06-13
generate indirect parameter assignment if arch uses syscall wrapper

yonghong-song:
Fix issue #1802.

On x64, the following commit (in 4.17) changed the raw parameter passed to
the syscall entry function from a list of parameters supplied in user space
to a single `pt_regs *` parameter. Also in 4.17, x64 syscall entry function
is changed from `sys_<name>` to `__x64_sys_<name>`.

```
commit fa697140f9a20119a9ec8fd7460cc4314fbdaff3
Author: Dominik Brodowski <linux@dominikbrodowski.net>
Date:   Thu Apr 5 11:53:02 2018 +0200

    syscalls/x86: Use 'struct pt_regs' based syscall calling convention for 64-bit syscalls

    Let's make use of ARCH_HAS_SYSCALL_WRAPPER=y on pure 64-bit x86-64 systems:

    Each syscall defines a stub which takes struct pt_regs as its only
    argument. It decodes just those parameters it needs, e.g:

            asmlinkage long sys_xyzzy(const struct pt_regs *regs)
            {
                    return SyS_xyzzy(regs->di, regs->si, regs->dx);
            }

    This approach avoids leaking random user-provided register content down
    the call chain.

    ...
```

In bcc, we support kprobe function signatures in the bpf program.
The rewriter will automatically generate proper assignment to
these parameters. With the above function signature change, the
original method does not work any more.

This patch enhanced rewriter to generate two version codes guarded
with CONFIG_ARCH_HAS_SYSCALL_WRAPPER. But we need to identify
whether a function will be attached to syscall entry function
or not during prog load time at which time the program has not
attached to any event.

The prefix `kprobe__` is used for kprobe autoload, we can use
`kprobe____x64_sys_` as the prefix to identify x64 syscall entry
functions. To support other architecture or not-autoloading program,
the prefix `syscall__` is introduced to signal it is a syscall
entry function.

trace.py and other tools which uses kprobe syscall entry functions
are also modified with the new interface so that they can
work properly with 4.17.

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -429,9 +429,83 @@ DiagnosticBuilder ProbeVisitor::error(SourceLocation loc, const char (&fmt)[N]) 
 BTypeVisitor::BTypeVisitor(ASTContext &C, BFrontendAction &fe)
     : C(C), diag_(C.getDiagnostics()), fe_(fe), rewriter_(fe.rewriter()), out_(llvm::errs()) {}
 
-bool BTypeVisitor::VisitFunctionDecl(FunctionDecl *D) {
+void BTypeVisitor::genParamDirectAssign(FunctionDecl *D, string& preamble,
+                                        const char **calling_conv_regs) {
+  for (size_t idx = 0; idx < fn_args_.size(); idx++) {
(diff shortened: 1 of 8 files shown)


yonghong-song:
cc @spyff

PR #1818, opened 2018-06-13
Modify trace function from generic_file_read_iter() to ext4_file_read_iter()

kernel-z:
The new Linux version (since form 4.10) uses ext4_file_read_iter(), And if the 'CONFIG_FS_DAX' 
 is not set ,then ext4_file_read_iter() will call generic_file_read_iter(), else it will call 
 ext4_dax_read_iter(), and trace generic_file_read_iter() will fail.

### DIFF
--- a/tools/ext4slower.py
+++ b/tools/ext4slower.py
@@ -23,6 +23,7 @@
 #
 # 11-Feb-2016   Brendan Gregg   Created this.
 # 15-Oct-2016   Dina Goldshtein -p to filter by process ID.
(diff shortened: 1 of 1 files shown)


kernel-z:
There are too many scripts, and will be more. 
I think one script for one function is better.
If two scripts for one function, when the kernel evolving ,it may be  need another one.

yonghong-song:
[buildbot, ok to test]

kernel-z:
Thank you ,@yonghong-song,@pchaigno, I have changed the code as zfsdist.py type.

yonghong-song:
LGTM. Thanks!

PR #1821, opened 2018-06-13
Trace external pointers through function returns

pchaigno:
Surprisingly, the rewriter wasn't able to trace external pointers returned by inlined functions until now.  This commit fixes it by adding functions that return an external pointer to `ProbeVisitor`'s set of external pointers, along with the levels of indirection.

This change requires reversing a few traversals to visit called functions before they are called. Then, we check the presence of an external pointer on return statements and retrieve that information at the call expression.

With this, #1683 is finally fixed and shouldn't require any manual call to `bpf_probe_read` anymore.


yonghong-song:
LGTM. Thanks!

PR #1822, opened 2018-06-13
profile: remove unnecessary bpf_probe_read call

pchaigno:
Remove an unnecessary call to `bpf_probe_read` in `tools/profile.py`. Since `PT_REGS_IP(&ctx->regs)` is still a direct access to context, using `bpf_probe_read` is not required.

@brendangregg Do you remember if you had a particular reason for the `bpf_probe_read` call? Maybe some issue I missed with older kernels?


yonghong-song:
LGTM. Thanks!

PR #1823, opened 2018-06-13
Recognize context member dereferences despite array accesses 

pchaigno:
This pull request updates the rewriter's `IsContextMemberExpr` method to recognize context member dereferences despite array accesses. For example, the rewriter should recognize, in the following, that `prev` is an external pointer retrieved from the context pointer, despite the access to the second element of the `args` array.
```c
struct task_struct *prev = (struct task_struct *)ctx->args[1];
```
The same could be done for the translation of member dereferences to `bpf_probe_read` calls (cf. `ProbeVisitor::VisitMemberExpr`), but that would be a little bit more complex (to retrieve the correct base) and there's currently no tool I could see that would benefit from it, so I'm not sure it's worth it.

### DIFF
--- a/tools/tcpaccept.py
+++ b/tools/tcpaccept.py
@@ -88,16 +88,14 @@ int kretprobe__inet_csk_accept(struct pt_regs *ctx)
 
     // pull in details
     u16 family = 0, lport = 0;
-    bpf_probe_read(&family, sizeof(family), &newsk->__sk_common.skc_family);
-    bpf_probe_read(&lport, sizeof(lport), &newsk->__sk_common.skc_num);
+    family = newsk->__sk_common.skc_family;
+    lport = newsk->__sk_common.skc_num;
 
     if (family == AF_INET) {
         struct ipv4_data_t data4 = {.pid = pid, .ip = 4};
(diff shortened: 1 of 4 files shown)


tools/tcpaccept.py
```diff
@@ -88,16 +88,14 @@
```
pchaigno: I haven't found a way to replace the above call to `bpf_probe_read` with a dereference, so the Jenkins' builds are going to fail. Pull request # "fixes" this by skipping all dereferences inside `bpf_probe_read` calls, regardless of their type (the last argument is not a `UnaryOperatorExpr` of type `AddrOf` here).

yonghong-song:
@pchaigno could you submit a separate pull request for this change if you still wants to pursue? Looks like this is specifically dealing with raw_tracepoint input argument?

pchaigno:
Reopened at #1828. It was closed by mistake because of #1824's OP.

> Looks like this is specifically dealing with raw_tracepoint input argument?

It's definitely going to be useful for raw tracepoints, but it also is for other tracepoints and for kprobes (cf. the new Clang test case).

PR #1824, opened 2018-06-13
Skip dereferences inside bpf_probe_reads calls

pchaigno:
This pull request extends the scope of #1796 and #1812. If the user decides to rely on a manual call to `bpf_probe_read`, we don't try to rewrite its last argument. This is needed as the rewriter starts to recognize and rewrite more and more dereferences. In particular, it fixes #1823's builds.


yonghong-song:
LGTM.  I had an instance that `bpf_probe_read(&data, data_len, (&src_data + 1))`. We should prevent `&src_data + 1` as well since it is also kind of `addrof`. But let us go ahead to avoid the third argument all together. That makes sense and simplifies things a lot.

PR #1826, opened 2018-06-14
python: add __version__ attribute to bcc module

andihit:
Uses cmake to create a `version.py` with the `REVISION` from cmake

If the python module should be executable without cmake (at the moment the `__init__.py` imports `version.py` which is created on the fly by cmake), I could rename `version.py.in` to `version.py` and let cmake replace the file in the build process.


yonghong-song:
[buildbot, ok to test]

yonghong-song:
[buildbot, ok to test]

Issue #1830, opened 2018-06-14
rewriter may rewrite p->b in &p->b for bpf_probe_read


yonghong-song:
BTW, find another issue with rewriter, for the following example,
```
static __always_inline void unaligned_u32_set(unsigned char *to, __u8 *from)
{
  to[0] = from[0];
  to[1] = from[1];
  to[2] = from[2];
  to[3] = from[3];
}

static __always_inline void conn_id_ipv4_init(
  struct tcp_estats_conn_id *conn_id,
  __be32 *saddr, __be32 *daddr)
{
  conn_id->localaddressType = TCP_ESTATS_ADDRTYPE_IPV4;

  unaligned_u32_set(conn_id->remaddress.data, (__u8 *)(daddr));
  unaligned_u32_set(conn_id->localaddress.data, (__u8 *)(saddr));
}

....
    conn_id_ipv4_init(conn_id,
          &inet_sk(sk)->inet_saddr,
          &inet_sk(sk)->inet_daddr);
```

The rewriter tries to transform inet_sk(sk)->inet_saddr and inet_sk(sk)->inet_daddr)
with bpf_probe_read. The reason is that we did not do anything for
Uninary operator & and at the time for member access inet_sk(sk)->inet_saddr,
we do not know there exists an outside AddrOf and goes ahead to perform
the transformation.

The compilation will failure since the addrof & cannot take a rvalue after
bpf_probe_read. Should we implement some thing like
. Whenever we see an AddrOf, remember that children expr (not sure what is the easy
to do this yet.)
. Any member access expression, check whether it is already is children expr
of an addrof, if it is, skip the transformation

PR #1831, opened 2018-06-15
avoid probe rewriting of p->m in &(p->m)

yonghong-song:
Fix issue #1830.

After the rewrite, the code approximately becomes
```
  &(({type _val; bpf_probe_read(&_val, sizeof(_val), &(p->m)); _val)
```
Firstly the rewriting is really unnecessary, and secondly
the compilation will fail since the addressOf cannot take address
of the rvalue _val.

This patch intends to fix the problem by avoiding
the rewriting in the first place. The implementation
assumes no nested addressOf operations which should be
true in most cases if not all.


pchaigno:
@yonghong-song I think I know why the Jenkins' builds are failing. I forgot to handle a case for #1752 in `ProbeChecker`. I can fix it this evening (afternoon for you) if this is not too urgent.

yonghong-song:
@pchaigno Just checked the result. It is not that urgent. I will test my patch again after your patch. Thanks!

Issue #613, opened 2016-07-13
Bitfield access


markdrayton:
BCC doesn't currently support accessing bitfields. We can work around this by accessing a nearby member and moving pointers appropriately (e.g. #270) but these tricks are very brittle and will likely be upset by changes across kernel versions.

Any idea what would need to be done to make accessing bitfields by name work?

4ast:
plain C doesn't have a way to figure out the bit position of the bitfield, but clang should have this data. Can we teach clang rewriter to replace bitfield access within probe_read to normal load plus mask? Other ideas?
tcpaccept.py is a primary candidate for this cleanup to access sk_protocol in kernel independent way.
@goldshtn @drzaeus77

krkini16:
@4ast The workaround in #270 seems very specific--I've tried something similar in order to access
icsk_ca_state as below
`u8 ca_state = 0;
 bpf_probe_read(&ca_state, 1, (void *)((long)&icsk->icsk_ca_state) - 3);` 

yet unfortunately the 
`error: address of bit-field requested
    bpf_probe_read(&ca_state, 1, (void *)((long)&icsk->icsk_ca_state) - 3);
`

Still occurs. has any progress been made or are there any suggestions?

Eichhoernchen:
You need to read a non bitfield member and calculate the offset towards the member (here -3) to the byte representing the bitfield. Afterwards you can use bitmasks to select the appropriate byte..

I got here as I also wanted to read the icsk_ca_state field, yet even with this trick bcc has a problem with as it seems to again rewrite the the bpf_probe_read...

The following statement:
``` 
void my_rcv_established(struct pt_regs *ctx, struct sock *sk, struct sk_buff *skb, const struct tcphdr *th, unsigned int len)
{
    struct tcp_sock *tp = (struct tcp_sock *)sk;
    struct inet_connection_sock *inet = (struct inet_connection_sock *)sk;
    //...
    u8 ca_state = 0;
    bpf_probe_read(&ca_state, 1, (void *)((long)&inet->icsk_retransmits) - 1);
```
the last line is transformed into:
```
bpf_probe_read(&ca_state, 1, (void *)((long)&({ typeof(__u8) _val; memset(&_val, 0, sizeof(_val)); bpf_probe_read(&_val, sizeof(_val), (u64)inet + offsetof(struct inet_connection_sock, icsk_retransmits)); _val; })) - 1);
```
[...]

Issue #270, opened 2015-10-13
accessing bitfield for sk_protocol


brendangregg:
I'm not sure bitfields work yet. sk_protocol is a bitfield. Trying to test it:

```C
int kretprobe__inet_csk_accept(struct pt_regs *ctx)
{
        struct sock *newsk = (struct sock *)ctx->ax;
        u32 pid = bpf_get_current_pid_tgid();

        if (newsk == NULL)
                return 0;

        // check this is TCP
        if (newsk->sk_protocol != IPPROTO_TCP)
                return 0;
[...]
```

gets:

```Shell
bpf: Permission denied
0: (79) r6 = *(u64 *)(r1 +80)
1: (85) call 14
2: (15) if r6 == 0x0 goto pc+71
 R0=inv R6=inv R10=fp
3: (61) r1 = *(u32 *)(r6 +328)
R6 invalid mem access 'inv'
```

I also tried:

```C
        u8 protocol = 0;
        bpf_probe_read(&protocol, sizeof(protocol), &newsk->sk_protocol);
```

which gets:

```Shell
<bcc-memory-buffer>:16:46: error: address of bit-field requested
        bpf_probe_read(&protocol, sizeof(protocol), &newsk->sk_protocol);
                                                    ^~~~~~~~~~~~~~~~~~~
1 error generated.
Traceback (most recent call last):
  File "./tcpaccept", line 101, in <module>
    b = BPF(text=bpf_text)
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 349, in __init__
    raise Exception("Failed to compile BPF module %s" % src_file)
Exception: Failed to compile BPF module 
```

4ast:
In general bitfields are quite difficult for bpf_probe_read.
Eventually we can add some sophistication to bcc to get offsetof of nearby
fields,
do the endian math and apply the mask.
For now the easiest is to call bpf_probe_read manually like
bpf_probe_read(&protocol, 1, ((long)&newsk->sk_wmem_queued) - 3);
the above will work because sk_protocol bitfield is 8 bits and nicely
aligned.

brendangregg:
Thanks, I see what you're doing, but it doesn't like it:

```C
        // check this is TCP
        u8 protocol = 0;
        // workaround for reading the sk_protocol bitfield:
        bpf_probe_read(&protocol, 1, (void *)((long)&newsk->sk_wmem_queued) - 3);
[...]

PR #1834, opened 2018-06-16
Patch for tools/tcpaccept.py

kernel-z:
The script  can not work since Linux 4.10,
because the sk_wmem_queued is not following the bitfield of sk_protocol.
And the following member now is sk_gso_max_segs.
So, we can use this:
bpf_probe_read(&protocol, 1, (void *)((u64)&newsk->sk_gso_max_segs) - 3);


tools/tcpaccept.py
```diff
@@ -83,9 +83,22 @@
...
+    // because the sk_wmem_queued is not following the bitfield of sk_protocol.
+    // And the following member is sk_gso_max_segs.
+    // So, we can use this:
+    // bpf_probe_read(&protocol, 1, (void *)((u64)&newsk->sk_gso_max_segs) - 3);
+    // And it works from 4.10 to the latest 4.17.1, but maybe one day will fail
+    // again. 
 
+    if (protocol != IPPROTO_TCP)
+    {    
+        bpf_probe_read(&protocol, 1, (void *)((u64)&newsk->sk_gso_max_segs) - 3);
```
yonghong-song: You need to consider endianness here.

tools/tcpaccept.py
```diff
@@ -83,9 +83,22 @@
...
-    if (protocol != IPPROTO_TCP)
-        return 0;
+    
+    // Following comments add by Joe Yin:
+    // Unfortunately,it can not work since Linux 4.10,
+    // because the sk_wmem_queued is not following the bitfield of sk_protocol.
+    // And the following member is sk_gso_max_segs.
+    // So, we can use this:
+    // bpf_probe_read(&protocol, 1, (void *)((u64)&newsk->sk_gso_max_segs) - 3);
+    // And it works from 4.10 to the latest 4.17.1, but maybe one day will fail
```
yonghong-song: So the tool here won't work for 4.9 and earlier, right? Can we do some dynamic check here? You can use linux `offsetof` to find member offset of a structure. By comparing two different member offsets and you may be able to accommodate both pre-4.10 and 4.10+.

pchaigno: I agree that a dynamic check would be better here, rather than trying both fields. In 4.10+, the byte retrieved from `sk_wmem_queued` might actually equal `IPPROTO_TCP` by chance.

That structure is likely to be changed often, so you'll want to compare two close fields (so that the chance of something being removed between them is lower). Comparing the offsets of `__sk_flags_offset` and `sk_gso_max_segs` might do the trick.

pchaigno:
@kernel-z Thanks for the fix!

For future reference, commit [`9115e8c`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=9115e8cd2a0c6eaaa900c462721f12e1d45f326c) changed this. There doesn't appear to be any symbol change in the same patch (or even in the same patchset) we could use to detect this at compile time :-/

tools/tcpaccept.py
kernel-z: It can work both pre-4.10 and 4.10+. As @pchaigno said " In 4.10+, the byte retrieved from sk_wmem_queued might actually equal IPPROTO_TCP by chance." , but the probability is very low. 
As the issue "accessing bitfield for sk_protocol #270" proposaled by @brendangregg ,and the "Bitfield access #613 "  , BCC doesn't currently support accessing bitfields.
@pchaigno __sk_flags_offset is not exist at pre-4.10, so it hard to work well.

kernel-z: the offsetof function is very nice.
I have modified the script. @yonghong-song

yonghong-song:
[buildbot, ok to test]

yonghong-song:
[buildbot, ok to test]

yonghong-song:
LGTM. Thanks!

brendangregg:
Thanks for the fixes. A future change: in Linux 4.16, the sock:inet_sock_set_state tracepoint was added, and I think this tool can be enhanced to use it. That will put an end to such kernel version checks.

As it's a clean check to see if that tracepoint exists, in tcplife I implemented it as a whole different block of C code. The intent is to keep it simple, and in the distant future delete the kprobe versions and just leave the tracepoint version.

kernel-z:
@brendangregg thank you. Using tracepoint is better and stable. I have added the tracepoint check as you did in the tcplife script. #1838

PR #1835, opened 2018-06-17
Fix nested dereference rewrites

pchaigno:
When the rewriter meets a dereference of a member dereference it fails to properly rewrite them into calls to `bpf_probe_read`. The reason is that Clang is unable to track the position of rewritten text, but we can accommodate this by inserting text around the dereference instead of completely rewriting it. We are already doing that for member dereference, but not for simple dereference.


yonghong-song:
LGTM. Thanks!

PR #1836, opened 2018-06-17
Fix external pointer propagation in nested dereferences

pchaigno:
This pull request fixes external pointer propagation in nested dereferences and the count of indirections for `addrof`s of member dereferences (`&A->b`).

In nested dereferences, a dereference of an external pointer may give a new external pointer.  For example, if `A` is an external pointer, then `*A` and `A->b` should also be considered as external pointers when appropriate (e.g., in `**A` or `*(A->b)`).

In addition, a member dereference is a dereference, so we need to count it when counting the number of indirections in `ProbeChecker`. If we don't, `*(&A->b)` won't be rewritten correctly as `&A->b` will be
considered a pointer to an external pointer.

@yonghong-song This is a different fix than the one you implemented in #1831. I think it's a more general fix, as it isn't specific to the case of array accesses (I had to revert your fix, `ProbeVisitor::VisitArraySubscriptExpr`, for this to work). It also produces a more appropriate rewrite for tcptop's dereferences:
```c
// Original statement:
ipv6_key.saddr0 = *(u64 *)&sk->__sk_common.skc_v6_rcv_saddr.in6_u.u6_addr32[0];
// Rewrote to the following with the ProbeVisitor::VisitArraySubscriptExpr fix:
ipv6_key.saddr0 = *(u64 *)&({ typeof(__be32 [4]) _val; __builtin_memset(&_val, 0, sizeof(_val)); bpf_probe_read(&_val, sizeof(_val), (u64)&sk->__sk_common.skc_v6_rcv_saddr.in6_u.u6_addr32); _val; })[0];
// Rewrote to the following with the current pull request:
ipv6_key.saddr0 = ({ typeof(u64) _val; __builtin_memset(&_val, 0, sizeof(_val)); bpf_probe_read(&_val, sizeof(_val), (u64)(u64 *)&sk->__sk_common.skc_v6_rcv_saddr.in6_u.u6_addr32[0]); _val; });
```

I've opened this as a pull request against #1831, but it might just be easier to directly apply the diff to #1831.

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -136,13 +136,37 @@ class ProbeChecker : public RecursiveASTVisitor<ProbeChecker> {
       needs_probe_ = true;
       return false;
     }
+    if (M->isArrow()) {
+      /* In A->b, if A is an external pointer, then A->b should be considered
+       * one too.  However, if we're taking the address of A->b
+       * (nb_derefs_ < 0), we should take it into account for the number of
+       * indirections; &A->b is a pointer to A with an offset. */
+      if (nb_derefs_ >= 0) {
+        ProbeChecker checker = ProbeChecker(M->getBase(), ptregs_,
+                                            track_helpers_, is_assign_);
+        if (checker.needs_probe() && checker.get_nb_derefs() == 0) {
+          needs_probe_ = true;
+          return false;
+        }
+      }
+      nb_derefs_++;
+    }
     return true;
   }
   bool VisitUnaryOperator(UnaryOperator *E) {
-    if (E->getOpcode() == UO_Deref)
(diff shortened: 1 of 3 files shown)


yonghong-song:
Right, my array subscript visitor is mostly for tcptop.py case 
```
        ipv6_key.saddr0 = *(u64 *)&sk->__sk_common.skc_v6_rcv_saddr.in6_u.u6_addr32[0];
```
Where the first deref is not recognized as needs probe and the member access `sk->__sk_common.skc_v6_rcv_saddr.in6_u.u6_addr32` is recognized as need_probe().

With your implementation, the whole `*(u64 *)&sk->__sk_common.skc_v6_rcv_saddr.in6_u.u6_addr32[0]` is recognized as need_probe().
So yes, subscript visitor is not needed.

PR #1837, opened 2018-06-17
 Fix external pointer propagation in nested dereferences

yonghong-song:
Somehow github did not merge pull request #1836 properly which is stacked on top of #1831. Maybe stacked pull request does not work? Here, I created a separate pull request for the remaining code of #1836.


pchaigno:
> Somehow github did not merge pull request #1836 properly which is stacked on top of #1831. Maybe stacked pull request does not work?

I used them before, but it's possible they don't work with all types of merge (i.e., merge, rebase, and squash).

PR #1840, opened 2018-06-20
Added new footer.h header where BPF_LICENSE is set if not defined

oriolarcas:
The helpers.h header specifies the BPF program license as 'GPL'.
However, other GPL-compatible licenses are possible (e.g., Dual BSD/GPL)
or even proprietary licenses (e.g., cachetop can run with a proprietary
license).

With this commit, the user can specify a BPF_LICENSE macro in the source
code:

  #define BPF_LICENSE Custom license

Note it supports multiple words and the absence of quotes. If the BPF
doesn't have a GPL-compatible license and it uses any GPL-only helpers,
the kernel will reject it with:

If no license is specified, it will fall back to GPL (the current
behavior before this commit) so that the BCC tools and examples remain
usable.

Updated the documentation with BPF_LICENSE description, licensing error
description, and licenses for each helper.

Signed-off-by: Oriol Arcas <oriol@starflownetworks.com>


yonghong-song:
[buildbot, ok to test]

yonghong-song:
Look good. But ubuntu 16.04 test failed. ubuntu 16.04 is linux 4.4 which does not have bpf_get_stackid support. I think you can just remove it since bpf_ktime_get_ns also requires gpl and is available on 4.4.

docs/reference_guide.md
```diff
@@ -1521,3 +1545,16 @@ Traceback (most recent call last):
     raise Exception("Failed to load BPF program %s" % func_name)
 Exception: Failed to load BPF program kretprobe__inet_csk_accept
 ```
+
+## 2. Cannot call GPL only function from proprietary program
+
+This error happens when a GPL-only helper is called from a non-GPL BPF program. To fix this error, do not use GPL-only helpers from a proprietary BPF program, or relicense the BPF program under a GPL-compatible license. Check which [BPF helpers](https://github.com/iovisor/bcc/blob/master/docs/kernel-versions.md#helpers) are GPL-only, and what licenses are considered GPL-compatible.
+
+Example calling `bpf_get_stackid()`, a GPL-only BPF helper, from a proprietary program (`cflags=['-DBPF_LICENSE=Proprietary']`):
```
4ast: please remove the ability to specify the license via flags. I don't think it was ever done by any open source project. The license should only be explicitly set in the source code (either .h or .c)

yonghong-song:
FYI, the following change
```
iff --git a/src/cc/frontends/clang/b_frontend_action.cc b/src/cc/frontends/clang/b_frontend_action.cc
index 593859a..d07f09e 100644
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -1279,6 +1279,9 @@ void BFrontendAction::DoMiscWorkAround() {
   // CONFIG_CC_STACKPROTECTOR properly based on other configs, so it relieved any bpf
   // program (using task_struct, etc.) of patching the below code.
   rewriter_->getEditBuffer(rewriter_->getSourceMgr().getMainFileID()).InsertText(0,
+    "#if defined(BPF_LICENSE)\n"
+    "#error BPF_LICENSE cannot be specified through cflags\n"
+    "#endif\n"
     "#if !defined(CONFIG_CC_STACKPROTECTOR)\n"
     "#if defined(CONFIG_CC_STACKPROTECTOR_AUTO) \\\n"
     "    || defined(CONFIG_CC_STACKPROTECTOR_REGULAR) \\\n"
-bash-4.2$ 
```
can prevent user defined macro BPF_LICENSE

src/cc/export/footer.h
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef BPF_LICENSE
+/* No license defined, using GPL
+ * Use cflags to define your own BPF_LICENSE */
```
4ast: this part of comment is obsolete?
Other than this bit the patch looks good to me.

oriolarcas: Thanks for the feedback, I changed it in 639217f. Any other fixes?

PR #1841, opened 2018-06-21
Unify and improve C++'s USDT implementation

palmtenor:
This PR makes C++ API uses `USDT::Context` and `USDT::Probe` directly instead of customized list-of-addresses logic. This has numbers of benefits:
- C++ API now automatically support USDT-for-PID
- C++ API now supports "same USDT from different binary in same PID"
- C++ API now support semaphore when attaching and detaching
- Reduces duplicated code and logic

### DIFF
--- a/examples/cpp/FollyRequestContextSwitch.cc
+++ b/examples/cpp/FollyRequestContextSwitch.cc
@@ -12,6 +12,7 @@
  */
 
 #include <signal.h>
+#include <functional>
 #include <iostream>
 #include <vector>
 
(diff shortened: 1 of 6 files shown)


yonghong-song:
Could you add a test case? The change is nontrivial and a test case will help people review/understand. 
What is the impact on Python APIs?

palmtenor:
Add test case, simplify implementation

yonghong-song:
LGTM. Thanks!

PR #1843, opened 2018-06-21
Remove handling of KeyboardInterrupt exception

banh-gao:
The KeyboardInterrupt exception should not be handled by the library since it is not related to bpf but rather to an interactive user interaction.
Moreover, handing it by calling the exit function impedes the caller from properly handling the exception depending on the context.


yonghong-song:
[buildbot, ok to test]

yonghong-song:
LGTM. Thanks!

PR #1847, opened 2018-06-23
Decode C strings into Python strings

juergenhoetzel:
Another leftover from #986.
`profile` failes on my `python3` setup without properly decoding:

```
juergen@samson:~/cxx/bcc → sudo /usr/share/bcc/tools/profile  -f -U 3 
  File "/usr/share/bcc/tools/profile", line 321, in <module>
    print("%s %d" % (";".join(line), v.value))
```

### DIFF
--- a/tools/profile.py
+++ b/tools/profile.py
@@ -304,7 +304,7 @@ for k, v in sorted(counts.items(), key=lambda counts: counts[1].value):
         # print folded stack output
         user_stack = list(user_stack)
         kernel_stack = list(kernel_stack)
-        line = [k.name.decode()]
+        line = [k.name]
         # if we failed to get the stack is, such as due to no space (-ENOMEM) or
         # hash collision (-EEXIST), we still print a placeholder for consistency
         if not args.kernel_stacks_only:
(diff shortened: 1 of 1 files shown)


yonghong-song:
Maybe the following is better?
```
diff --git a/tools/profile.py b/tools/profile.py
index ec5533f..240355c 100755
--- a/tools/profile.py
+++ b/tools/profile.py
@@ -304,7 +304,7 @@ for k, v in sorted(counts.items(), key=lambda counts: counts[1].value):
         # print folded stack output
         user_stack = list(user_stack)
         kernel_stack = list(kernel_stack)
-        line = [k.name.decode()]
+        line = [k.name]
         # if we failed to get the stack is, such as due to no space (-ENOMEM) or
         # hash collision (-EEXIST), we still print a placeholder for consistency
         if not args.kernel_stacks_only:
@@ -318,7 +318,7 @@ for k, v in sorted(counts.items(), key=lambda counts: counts[1].value):
                 line.append("[Missed Kernel Stack]")
             else:
                 line.extend([b.ksym(addr) for addr in reversed(kernel_stack)])
-        print("%s %d" % (";".join(line), v.value))
+        print("%s %d" % (b";".join(line).decode(), v.value))
     else:
         # print default multi-line stack output
         if not args.user_stacks_only:
[yhs@localhost tools]$
```
We should try to remove all intermediate `decode()` except the last one in `print`. The last one is used so the not-so-pleasant prefix `b` can be removed.

juergenhoetzel:
> We should try to remove all intermediate decode() except the last one in print. The last one is used so the not-so-pleasant prefix b can be removed.

I see :+1:  Amended.

yonghong-song:
[buildbot, ok to test]

yonghong-song:
LGTM. Thanks!

PR #1848, opened 2018-06-23
Strict version dependencies for debian packages, closes #1770

### LINKED ISSUE #1770: Strict version dependencies for debian packages
Currently we have:

```
Package: python-bcc
Architecture: all
Depends: libbcc, python, binutils
Description: Python wrappers for BPF Compiler Collection (BCC)
```

```
Package: bcc-tools
Architecture: all
Depends: python-bcc
Description: Command line tools for BPF Compiler Collection (BCC)
```

We upgraded `bcc-tools` (it's the package end users care about) and it broke, because APIs changed in `libbcc`. I think we should require the same package version for all packages to avoid this.

bobrik:


yonghong-song:
LGTM. Thanks!

PR #1855, opened 2018-06-26
Modify tools/ext4dist.py for 4.10+

kernel-z:
From Linux 4.10, the function .read_iter at the ext4_file_operations has 
 changed to ext4_file_read_iter.
 So, I add the try and except,it will first to attach ext4_file_read_iter,
 if fails and will attach the generic_file_read_iter which used to pre-4.10.

### DIFF
--- a/tools/ext4dist.py
+++ b/tools/ext4dist.py
@@ -181,7 +181,7 @@ b = BPF(text=bpf_text)
 # if fails and will attach the generic_file_read_iter which used to pre-4.10.
 
 try:
-	b.attach_kprobe(event="ext4_file_read_iter", fn_name="trace_read_entry")
+	b.attach_kprobe(event="ext4_file_read_iter", fn_name="trace_entry")
 except:
 	b.attach_kprobe(event="generic_file_read_iter", fn_name="trace_read_entry")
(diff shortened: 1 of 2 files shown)


kernel-z:
Oh,I forget to add the function to replace the "trace_read_entry"

tools/ext4dist.py
```diff
@@ -174,7 +174,16 @@
...
 # Common file functions. See earlier comment about generic_file_read_iter().
-b.attach_kprobe(event="generic_file_read_iter", fn_name="trace_read_entry")
+# Comment by Joe Yin 
+# From Linux 4.10, the function .read_iter at the ext4_file_operations has 
+# changed to ext4_file_read_iter.
+# So, I add the try and except,it will first to attach ext4_file_read_iter,
+# if fails and will attach the generic_file_read_iter which used to pre-4.10.
+
+try:
+	b.attach_kprobe(event="ext4_file_read_iter", fn_name="trace_entry")
```
yonghong-song: Maybe you can use BPF helper `get_kprobe_functions`? This way, we do not need try-and-error approach. Several other tools uses `get_kprobe_functions` as well.

yonghong-song:
[buildbot, ok to test]

Issue #1853, opened 2018-06-25
ioctl(PERF_EVENT_IOC_SET_BPF): Permission denied in BCC 0.6.0


bobrik:
I have the following minimal reproduction:

```c
BPF_HASH(foo, u64);

TRACEPOINT_PROBE(block, block_rq_complete) {
  u64 key = 0;

  if (args->rwbs[0] == 'W' || args->rwbs[1] == 'W') {
    key = 1;
  }

  foo.increment(key);

  return 0;
}
```

Usable python wrapper:

```python
from bcc import BPF
from time import sleep

program="""
BPF_HASH(foo, u64);

TRACEPOINT_PROBE(block, block_rq_complete) {
  u64 key = 0;

  if (args->rwbs[0] == 'W' || args->rwbs[1] == 'W') {
    key = 1;
  }

  foo.increment(key);

  return 0;
}
"""

BPF(text=program, debug=8)
```

On BCC 0.6.0 with either LLVM 3.8 or LLVM 6.0 (I haven't tried others) this fails with:

```
Disassembly of section .bpf.fn.tracepoint__block__block_rq_complete:
tracepoint__block__block_rq_complete:
; TRACEPOINT_PROBE(block, block_rq_complete) { // Line  21
   0:	79 12 20 00 00 00 00 00 	r2 = *(u64 *)(r1 + 32)
; if (args->rwbs[0] == 'W' || args->rwbs[1] == 'W') { // Line  25
   1:	15 02 03 00 57 00 00 00 	if r2 == 87 goto +3
   2:	b7 02 00 00 00 00 00 00 	r2 = 0
   3:	79 11 28 00 00 00 00 00 	r1 = *(u64 *)(r1 + 40)
   4:	55 01 01 00 57 00 00 00 	if r1 != 87 goto +1
   5:	b7 02 00 00 01 00 00 00 	r2 = 1
; ({ typeof(foo.key) _key = key; typeof(foo.leaf) *_leaf = bpf_map_lookup_elem_(bpf_pseudo_fd(1, 3), &_key); if (_leaf) (*_leaf)++; else { typeof(foo.leaf) _zleaf; __builtin_memset(&_zleaf, 0, sizeof(_zleaf)); _zleaf++; bpf_map_update_elem_(bpf_pseudo_fd(1, 3), &_key, &_zleaf, BPF_NOEXIST); } }); // Line  29
[...]

PR #1856, opened 2018-06-26
fix a bug in tracepoint struct rewriter

yonghong-song:
Fix issue #1853.

Commit 7c489469a7dc ("adjust tracepoint field type
based on size") tried to fix the tracepoint format
descrepancy between declared type and actual size is 8.
The type has to be promoted to match the size.

The commit introduced a bug if the field is an array.
For exmaple, `block:block_rq_complete` tracepoint has
field rwbs:
```
  field:char rwbs[8];	offset:32;	size:8;	signed:1;
```
The current implementation will incorrectly translate it
into
```
  s64 rwbs[8];
```
since it considers the type is "char".

This patch fixed this issue by checking the field name
and if it is an array, rewriting will be skipped.

### DIFF
--- a/src/cc/frontends/clang/tp_frontend_action.cc
+++ b/src/cc/frontends/clang/tp_frontend_action.cc
@@ -98,6 +98,9 @@ static inline field_kind_t _get_field_kind(string const& line,
     return field_kind_t::data_loc;
   if (field_name.find("common_") == 0)
     return field_kind_t::common;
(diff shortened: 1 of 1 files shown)


yonghong-song:
cc @bobrik

bobrik:
I would still love to get a better error than just `-EPERM`, but the fix LGTM. Thanks!

PR #1863, opened 2018-06-29
tools: remove unnecessary calls to bpf_probe_read

pchaigno:
Most of these calls have been rendered useless by #1828 ("Recognize context member dereferences despite array accesses").


pchaigno:
Actually, we can remove three more calls to `bpf_probe_read` in tcptracer.py. I updated the branch.

brendangregg:
Thanks, looks like these work. Gee, bcc is getting easier.

PR #1868, opened 2018-07-03
examples/http_filter: Add pointer offset accounting for variable ip header len

ChaosData:
This fix adds additional accounting logic to the http_filter examples ("simple" and "complete") that make sure to shift the pointer for the IP body/payload to the correct offset before accessing TCP header fields. This is done by taking into account the IP header length field. Previously, the IP header length field was used, but it was done later in processing, after TCP header values were extracted using the size of the BCC `proto.h` `struct ip_t` a static offset.

Prior to this, it was possible to evade HTTP detection by injecting IP options data into the IP header that (may spoof parts of the TCP header and) shift the real TCP header one down, as done in the scapy snippet below:

```Python
IP(
  dst=target[0],
  options=struct.pack(">BBHHHB",130,11,8080,0,0,0),
)/TCP(
  ...
)
```

I'm also not 100% sure on the model of how packets get into this subsystem and to what degree the networking stack has already validated/processed them, so I added in a minimum length sanity check as well. It may be completely unnecessary, but I don't know enough to conclusively say that (it was also unclear if there was an API to get the full packet length out-of-band from the packet data itself or in an otherwise secure manner).

### DIFF
--- a/examples/networking/http_filter/http-parse-complete.c
+++ b/examples/networking/http_filter/http-parse-complete.c
@@ -56,6 +56,19 @@ int http_filter(struct __sk_buff *skb) {
 	struct Key 	key;
 	struct Leaf zero = {0};
 
+        //calculate ip header length
+        //value to multiply * 4
+        //e.g. ip->hlen = 5 ; IP Header Length = 5 x 4 byte = 20 byte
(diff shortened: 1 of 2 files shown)


yonghong-song:
[buildbot, ok to test]

yonghong-song:
[buildbot, ok to test]

Issue #1777, opened 2018-05-22
Python tutorial examples broken


myllynen:
The examples using sys_clone in https://github.com/iovisor/bcc/blob/master/docs/tutorial_bcc_python_developer.md seem to be broken on 4.17-rc6 / latest bcc nightly, changing sys_clone to __x64_sys_clone would be a way to fix this that's of course platform dependent.

pchaigno:
Since it's only for the tutorial, we could switch to some other, easy-to-trigger syscall. Do you want to open a pull request?

fche:
Do you have access to the general sys_enter tracepoint?

palmtenor:
Actually, can we have Python's `BPF._trace_autoload` to check if the kprobe function name is starting with `sys` or `SyS` and if yes, call `BPF.get_syscall_fnname`?

yonghong-song:
`_trace_autoload` already implemented the checking.

dustin-decker:
I think #1869 is related.

PR #1872, opened 2018-07-04
Improve Python autoload syscall name handling

palmtenor:
Tries to resolve issues like #1777. Check if the autoloaded kprobe name looks like a syscall and then replace with correct prefix.

cc @yonghong-song @pchaigno


palmtenor:
Good points. Addressed those comments

PR #1873, opened 2018-07-06
tcptop: fix display of received bytes, reduce syscalls (fixes #1871)

### LINKED ISSUE #1871: tcptop: possibly wrong display of received bytes
Looking at tcptop, in line
https://github.com/iovisor/bcc/blob/4ac63076a2ceb8b348894f6e025366bfef43768f/tools/tcptop.py#L227
`keys` now points to the `ipv4_recv_bytes` table, and all keys from `ipv4_send_bytes` which are not yet in `ipv4_recv_bytes` will be copied (referenced).

For the case that there is a key in `ipv4_send_bytes` but not in `ipv4_recv_bytes`, this line will always evaluate to true anyway:
https://github.com/iovisor/bcc/blob/4ac63076a2ceb8b348894f6e025366bfef43768f/tools/tcptop.py#L242
because this table contains *all* the keys now (because `keys` is a reference to `ipv4_recv_bytes` and not a copy of this table)

So in this case it will display an equal amount of received and sent bytes, but in reality received bytes should be zero.
Did I miss some magic here, or is this a bug?

My probably not so elegant solution would be to create new dicts of the tables where the key is a tuple of pid, laddr, lport, daddr, dport, and use a new dict for merging the keys of the two tables.

andihit:
the keys variable was a reference to ipv4_recv_bytes, and after merging
the keys of ipv4_send_bytes the original ipv4_recv_bytes BPF table also
contained all keys, therefore the `if k in ipv4_recv_bytes` check a few
lines below always evaluates to true

with this commit all BPF tables are copied to userspace as dicts once
(to reduce the number of syscalls) and the keys are merged in a new dict


palmtenor:
[buildbot, ok to test]

tools/tcptop.py
```diff
@@ -223,60 +240,60 @@ def pid_to_comm(pid):
...
-    # IPv4:  build dict of all seen keys
-    keys = ipv4_recv_bytes
-    for k, v in ipv4_send_bytes.items():
-        if k not in keys:
-            keys[k] = v
+    ipv4_send_bytes_dict = ipv4_table_to_dict(ipv4_send_bytes)
+    ipv4_recv_bytes_dict = ipv4_table_to_dict(ipv4_recv_bytes)
+
+    # IPv4: build dict of all seen keys
+    keys = ipv4_send_bytes_dict.copy()
```
palmtenor: Maybe the `dict` can be from `TCPSessionKey` keys to a pair of send / received bytes, and we just go through both send the received table to update the values? You can use Python's `defaultdict` to initialize the values to 0 if the key is not seen before.

tools/tcptop.py
```diff
@@ -223,60 +240,60 @@ def pid_to_comm(pid):
...
 
         print("%-6d %-12.12s %-21s %-21s %6d %6d" % (k.pid,
             pid_to_comm(k.pid),
-            inet_ntop(AF_INET, pack("I", k.saddr)) + ":" + str(k.lport),
-            inet_ntop(AF_INET, pack("I", k.daddr)) + ":" + str(k.dport),
+            k.laddr + ":" + str(k.lport),
+            k.daddr + ":" + str(k.dport),
             recv_kbytes, send_kbytes))
 
     ipv4_send_bytes.clear()
```
palmtenor: Let's now clear right after we read. The clearing is also an iteration through the tables and there are definitely race between clearing them from user space and BPF program still writing things to it.

andihit:
Updated my commit. I'm sorting by `sum(sent+recv)` now instead of `recv ? recv : sent`.

PR #1880, opened 2018-07-09
clang: Allow user to override kernel version

joelagnel:
BCC currently requires exactly matching headers. Sometimes this is quite
inconvenient especially if the kernel version is only very slightly
different such as updates in a stable kernel. This patch gives the user
the flexibility to override the the LINUX_VERSION_CODE provided in the
linux kernel headers, so that the eBPF program may load. We also print a
message when this is done, so that the user is warned about the override
happening and that results may be unpredictable.

Signed-off-by: Joel Fernandes <joel@joelfernandes.org>


pchaigno:
Should this be documented somewhere? Not sure where's the appropriate place though :-/

joelagnel:
Yes, not sure either. There's also env variables like `BCC_KERNEL_MODULES_DIR` and `BCC_KERNEL_SOURCE` that could use docs I guess. Perhaps it should be a separate PR.

palmtenor:
[buildbot, ok to test]

palmtenor:
[buildbot, ok to test]

pchaigno:
@joelagnel What about a new section in [the reference guide](https://github.com/iovisor/bcc/blob/master/docs/reference_guide.md)?

joelagnel:
Ok, I'll update the docs.

But any idea what happened to the checks?
https://buildbot.iovisor.org/jenkins/job/bcc-pr/1250/console

pchaigno:
Looks like a [failure](https://buildbot.iovisor.org/jenkins/job/bcc-pr/1250/label=ubuntu1710/console) of [`test_usdt3.py`](https://github.com/iovisor/bcc/blob/7a582320b77dee0af6ac6a7c43f1e5706a5051c7/tests/python/test_usdt3.py#L3), but I don't see the relation to your changes. @drzaeus77 ?

drzaeus77:
[buildbot, test this please]

yonghong-song:
@joelagnel There is a conflict, could rebase on top of master branch?
Also, can we change the environment variable to BCC_LINUX_VERSION_CODE? This way, the
environment naming conversion will be consistent with other BCC env variables (except ARCH).

docs/reference_guide.md
```diff
@@ -1558,3 +1562,26 @@ bpf: Failed to load program: Invalid argument
 8: (85) call bpf_get_stackid#27
 cannot call GPL only function from proprietary program
 ```
+
+# Environment Variables
+
+## 1. Kernel source directory
+
+eBPF program compilation needs kernel sources or kernel headers with headers
+compiled.  Incase your kernel sources are at a non-standard location where BCC
```
palmtenor: Incase -> In case (and the extra space)

joelagnel:
This got closed because I rebased I think. I'll send another PR.

Issue #1860, opened 2018-06-27
Linux kernel 4.11, get_table_offline gives a segmentation fault for percpu maps


mohankku:
I have a pinned map of type BPF_MAP_TYPE_LRU_PERCPU_HASH,  which gives std::bad_alloc when I invoke get_table_offline for that map.  

Is someone looking at this issue, or some one faced the same problem ?

palmtenor:
Do you have a stack trace? Can you show your code for repro?

mohankku:
I analyzed the issue, and it seems that BPFPercpuHashTable passes vector of ValueType to BPFHashTable. BPFHashTable passes this vector pointer to bpf_lookup_elem, which considers this as a pointer to the value structure. And, when I looked at the vector, it seems to be corrupt after bpf_lookup_elem. I used a structure array of ncpus, instead of a vector. And, that seems to solve the problem.

mauriciovasquezbernal:
@mohankku could you please try this?

```
diff --git a/src/cc/api/BPFTable.h b/src/cc/api/BPFTable.h
index 3c83b25..e1fbeb6 100644
--- a/src/cc/api/BPFTable.h
+++ b/src/cc/api/BPFTable.h
@@ -236,7 +236,7 @@ class BPFHashTable : public BPFTableBase<KeyType, ValueType> {
       return res;
 
     while (true) {
-      if (!this->lookup(&cur, &value))
+      if (!this->lookup(&cur, get_value_addr(value)))
         break;
       res.emplace_back(cur, value);
       if (!this->next(&cur, &cur))
```

If this works I'll open a PR with the fix and some tests case for it.

mohankku:
Mauricio, the above fix solved the issue. I can send a PR with the fix tomorrow.

PR #1884, opened 2018-07-11
fix get_table_offline for percpu hash table

mauriciovasquezbernal:
Use get_value() instead of lookup() as the value as in the case of
percpu tables the ValueType is an std::vector that has to be resized
before performing the lookup function.

Add also some testing for it.

Solves https://github.com/iovisor/bcc/issues/1860

Signed-off-by: Mauricio Vasquez B <mauricio.vasquez@polito.it>


4ast:
thanks for the fix!

palmtenor:
[buildbot, ok to test]

Issue #1883, opened 2018-07-10
Unable to get PID from real parent


calavera:
I was working on porting execnoop to Go and I wanted to get the parent PID from the current task. Unfortunately, the TGID for the task's real parent is always 0.

I used the code from this closed issue: https://github.com/iovisor/bcc/issues/799

```
  struct task_struct *task;
  struct task_struct *real_parent_task;
  u64 ppid;

  task = (struct task_struct *)bpf_get_current_task();
  bpf_probe_read(&real_parent_task, sizeof(real_parent_task), &task->real_parent);
  bpf_probe_read(&ppid, sizeof(ppid), &real_parent_task->tgid);
```

I also used `bpf_trace_printk` to print the result from calling `bpf_probe_read` and it always returns 0. You can see my bpf program in this PR: https://github.com/iovisor/gobpf/pull/133/files#diff-3ef2d414e040c89dc6f8a62e402b33a9R92

pchaigno:
You wouldn't happen to run a 4.13 Linux kernel by any chance?

calavera:
Yes! Is that a known issue? I couldn't find anything related anywhere:

```
Linux c 4.13.0-46-generic #51-Ubuntu SMP Tue Jun 12 12:36:29 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

palmtenor:
@calavera Can you try the fix mentioned here: https://github.com/iovisor/bcc/issues/1548#issuecomment-359964600?

calavera:
@palmtenor that doesn't seem to work in my case. I applied this patch to execsnoop:

```diff
diff --git i/examples/bcc/execsnoop/execsnoop.go w/examples/bcc/execsnoop/execsnoop.go
index 1bb3f20..50687c9 100644
--- i/examples/bcc/execsnoop/execsnoop.go
+++ w/examples/bcc/execsnoop/execsnoop.go
@@ -37,6 +37,9 @@ const (
 )
 
 const source string = `
+#define randomized_struct_fields_start  struct {
+#define randomized_struct_fields_end    };
+
 #include <uapi/linux/ptrace.h>
 #include <linux/sched.h>
 #include <linux/fs.h>
@@ -272,9 +275,9 @@ func main() {
                                        RetVal: event.RetVal,
                                }
 
-                               if event.Ppid == 0 {
-                                       event.Ppid = getPpid(event.Pid)
-                               }
+                               //                              if event.Ppid == 0 {
[...]

PR #1885, opened 2018-07-11
[tools/execsnoop] Try to get parent PID from current task's real parent.

calavera:
Fallback to read the PPid from /proc if the real parent's TGID is 0.

Related to #1883

Signed-off-by: David Calavera <david.calavera@gmail.com>


palmtenor:
Given what you said in #1883 , have you tested in another Kernel to make sure the BPF part works?

Also, please update the `TODO` comment at ` def get_ppid(pid):` (my inline comment somehow doesn't work...)

palmtenor:
[buildbot, ok to test]

tools/execsnoop.py
```diff
@@ -105,7 +106,17 @@
...
+    struct task_struct *task;
+    struct task_struct *real_parent_task;
+    u32 ppid;
+
     data.pid = bpf_get_current_pid_tgid() >> 32;
+
+    task = (struct task_struct *)bpf_get_current_task();
+    bpf_probe_read(&real_parent_task, sizeof(real_parent_task), &task->real_parent);
+    bpf_probe_read(&ppid, sizeof(ppid), &real_parent_task->tgid);
+    data.ppid = ppid;
```
pchaigno: You shouldn't need explicit calls to `bpf_probe_read` here:
```c
daat.ppid = task->real_parent->tgid;
```
bcc will handle the rewrite.

tools/execsnoop.py
```diff
@@ -128,7 +139,16 @@
...
```
pchaigno: Same as above.

tools/execsnoop.py
calavera: ok, good to know, I tried that too and got similar results.

pchaigno: Yep, that wouldn't change the issue with the parent PID. It translates to the same bytecode.

brendangregg:
Thanks for getting this done. Any further info on the kernel versions where task->real_parent->tgip sometimes returns zero? Just want to understand that caveat as I'm sure I'll run into it in the future too.

calavera:
I know for sure that 4.13.0-46-generic returns 0. I just tested 4.17.5 and it returns the real parent PID

calavera:
If you want to wait before merging this, I can test 4.14, 4.15 and 4.16 to have the minimum version that works. It might be good to have that information in the docs. I can also open an new PR with docs if you point me to the best place to have that kind of information.

brendangregg:
Since you're the first person to do ppid properly, I think people will copy-n-paste your code everywhere. So I'd put a comment with the code at least. :-)

palmtenor:
@calavera Thanks! I think it is not a minimum version, only 4.13 (and maybe 4.14) should have the problem and earlier Kernels should be fine as well. Is your Kernel the Ubuntu mainline 4.13? I will try debug see why the temporary fix didn't work~

calavera:
@brendangregg I've added an inline comment.

calavera:
It looks like the kernel that Jenkins use for tests in Xenial is too old and doesn't include `bpf_get_current_task`

```
27: 
27: HINT: bpf_get_current_task missing (added in Linux 4.8).
```

pchaigno:
> It looks like the kernel that Jenkins use for tests in Xenial is too old and doesn't include `bpf_get_current_task`

@calavera You'll have to add a precondition to the test in the same way as `test_mountsnoop`.

calavera:
Updated, thanks @pchaigno

PR #1888, opened 2018-07-16
ucalls: use replace error handler

jeromemarchand:
Prevents the following error when tracing a java program that contains
non-ascii method name:
```
  File "/usr/share/bcc/tools/lib/ucalls", line 305, in <module>
    data = get_data()   # [(function, (num calls, latency in ns))]
  File "/usr/share/bcc/tools/lib/ucalls", line 266, in get_data
    bpf["counts"].items()))
  File "/usr/share/bcc/tools/lib/ucalls", line 264, in <lambda>
    kv[0].method.decode(),
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc4 in position 11: ordinal not in range(128)
```
Signed-off-by: Jerome Marchand <jmarchan@redhat.com>

### DIFF
--- a/tools/lib/ucalls.py
+++ b/tools/lib/ucalls.py
@@ -255,13 +255,15 @@ if args.syscalls:
 def get_data():
     # Will be empty when no language was specified for tracing
     if args.latency:
-        data = list(map(lambda kv: (kv[0].clazz.decode() + "." + \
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, ok to test]

PR #1890, opened 2018-07-18
add usdt calls to libbcc

brendangregg:
I'm using libbcc in another project, and would like to use bcc_usdt_new_frompid() etc like it currently uses bcc_resolve_symname(), instead of having to rewrite the USDT stuff.

### DIFF
--- a/src/cc/CMakeLists.txt
+++ b/src/cc/CMakeLists.txt
@@ -47,10 +47,15 @@ add_library(bcc-shared SHARED
 set_target_properties(bcc-shared PROPERTIES VERSION ${REVISION_LAST} SOVERSION 0)
 set_target_properties(bcc-shared PROPERTIES OUTPUT_NAME bcc)
 
+if(ENABLE_USDT)
+  set(bcc_usdt_sources usdt/usdt.cc usdt/usdt_args.cc)
+  # else undefined
+endif()
+
 add_library(bcc-loader-static STATIC ${bcc_sym_sources} ${bcc_util_sources})
 target_link_libraries(bcc-loader-static elf)
 add_library(bcc-static STATIC
-  ${bcc_common_sources} ${bcc_table_sources} ${bcc_util_sources})
+  ${bcc_common_sources} ${bcc_table_sources} ${bcc_util_sources} ${bcc_usdt_sources})
 set_target_properties(bcc-static PROPERTIES OUTPUT_NAME bcc)
 set(bcc-lua-static
   ${bcc_common_sources} ${bcc_table_sources} ${bcc_sym_sources} ${bcc_util_sources})
@@ -76,6 +81,7 @@ if(ENABLE_CPP_API)
 endif()
 
 if(ENABLE_USDT)
+  list(APPEND bcc_api_headers bcc_usdt.h)
   add_subdirectory(usdt)
   list(APPEND bcc_common_libs_for_a usdt-static)
   list(APPEND bcc_common_libs_for_s usdt-static)


palmtenor:
Does it work when `ENABLE_USDT` build option is turned off?

brendangregg:
Thanks, I'll include the if(ENABLE_USDT) checks...

src/cc/CMakeLists.txt
```diff
@@ -76,6 +81,7 @@ if(ENABLE_CPP_API)
 endif()
 
 if(ENABLE_USDT)
+  list(APPEND bcc_api_headers bcc_usdt.h)
   add_subdirectory(usdt)
   list(APPEND bcc_common_libs_for_a usdt-static)
   list(APPEND bcc_common_libs_for_s usdt-static)
```
palmtenor: `bcc-static` has this (`bcc_common_libs_for_s`) linked below which already has `usdt-static` (from `src/cc/usdt/CMakeLists.txt`). Why do we need to have another `bcc_usdt_sources` rule? Maybe we just need the extra header?

yonghong-song:
Looks like bcc_usdt_new_frompid already in the shared library.
```
[yhs@localhost cc]$ nm libbcc.so.0.6.0 | grep bcc_usdt
0000000002b1e660 T bcc_usdt_close
0000000002b1e020 T bcc_usdt_enable_probe
0000000002b1c640 T bcc_usdt_foreach
0000000002b1c9a0 T bcc_usdt_foreach_uprobe
0000000002b1f150 T bcc_usdt_genargs
0000000002b1c790 T bcc_usdt_get_argument
0000000002b1c650 T bcc_usdt_get_location
0000000002b1c560 T bcc_usdt_get_probe_argctype
0000000002b1e590 T bcc_usdt_new_frompath
0000000002b1e400 T bcc_usdt_new_frompid
00000000038f4100 b _ZGVZ16bcc_usdt_genargsE8storage_
0000000002b1c410 T _ZN4USDT7Context4eachEPFvP8bcc_usdtE
00000000038f4120 b _ZZ16bcc_usdt_genargsE8storage_
[yhs@localhost cc]$
```
But it is not in static library `libbcc.a`. Do you need the symbol in the static library as well?

If there is no need for the static library, simply exposing the header file should be enough.

src/cc/CMakeLists.txt
brendangregg: I tried, but without the bcc_usdt_sources rule, I don't get the bcc_usdt_* symbols in libbcc-loader-static.a and libbcc.a, and my other project doesn't compile.

brendangregg:
Ok, thanks, yes, this project is built using the static library.

src/cc/CMakeLists.txt
```diff
@@ -47,10 +47,15 @@ add_library(bcc-shared SHARED
 set_target_properties(bcc-shared PROPERTIES VERSION ${REVISION_LAST} SOVERSION 0)
 set_target_properties(bcc-shared PROPERTIES OUTPUT_NAME bcc)
 
-add_library(bcc-loader-static STATIC ${bcc_sym_sources} ${bcc_util_sources})
+if(ENABLE_USDT)
+  set(bcc_usdt_sources usdt/usdt.cc usdt/usdt_args.cc)
+  # else undefined
+endif()
+
+add_library(bcc-loader-static STATIC ${bcc_sym_sources} ${bcc_util_sources} ${bcc_usdt_sources})
```
yonghong-song: bcc_usdt_sources has been added below directly to bcc-static and it should work. So the bcc_usdt_sources at bcc-loader-static can be dropped. Could you give a try?

brendangregg: You're right, thanks, I've made the change and it works. Pushed.

brendangregg:
It might have lost the comment when I merged the commits, but I just made the change and it worked, thanks.

yonghong-song:
LGTM. Thanks!

PR #1891, opened 2018-07-18
[profile.py]: adding support to collect profile only from specified CPU

tehnerd:
Summary:
sometime it is usefull to collect stack only from single cpu
for example you have single core saturated while others dont and you
want to know whats going on there. in this diff i'm adding this ability
(network related code could be example of when single core is saturated
as usually you have 1 to 1 mappng between rx queue and cpu)

example of generated code w/ CPU specified:
```
./tools/profile.py -C 14 2 --ebpf
Sampling at 49 Hertz of all threads by user + kernel stack for 2 secs.

struct key_t {
    u32 pid;
    u64 kernel_ip;
    u64 kernel_ret_ip;
    int user_stack_id;
    int kernel_stack_id;
    char name[TASK_COMM_LEN];
};
BPF_HASH(counts, struct key_t);
BPF_STACK_TRACE(stack_traces, 16384);

// This code gets a bit complex. Probably not suitable for casual hacking.

int do_perf_event(struct bpf_perf_event_data *ctx) {

    if (bpf_get_smp_processor_id() != 14)
        return 0;

    u32 pid = bpf_get_current_pid_tgid() >> 32;
...
```
and w/o
```
./tools/profile.py  2 --ebpf
Sampling at 49 Hertz of all threads by user + kernel stack for 2 secs.

struct key_t {
    u32 pid;
    u64 kernel_ip;
    u64 kernel_ret_ip;
    int user_stack_id;
    int kernel_stack_id;
    char name[TASK_COMM_LEN];
};
BPF_HASH(counts, struct key_t);
BPF_STACK_TRACE(stack_traces, 16384);

// This code gets a bit complex. Probably not suitable for casual hacking.

int do_perf_event(struct bpf_perf_event_data *ctx) {

    u32 pid = bpf_get_current_pid_tgid() >> 32;
    if (!(1))
        return 0;
...
```

### DIFF
--- a/tools/profile.py
+++ b/tools/profile.py
@@ -140,7 +140,6 @@ BPF_STACK_TRACE(stack_traces, STACK_STORAGE_SIZE);
 // This code gets a bit complex. Probably not suitable for casual hacking.
 
 int do_perf_event(struct bpf_perf_event_data *ctx) {
-    CPU_FILTER
     u32 pid = bpf_get_current_pid_tgid() >> 32;
     if (!(THREAD_FILTER))
         return 0;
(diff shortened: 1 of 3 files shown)


tehnerd:
actually mb it should be done other way. let me rethink this

tehnerd:
ok. it works as i expected. just checked src. each core will run sep instance of bpf prog so we should be good here

tools/profile.py
```diff
@@ -138,6 +140,7 @@ def stack_id_err(stack_id):
 // This code gets a bit complex. Probably not suitable for casual hacking.
 
 int do_perf_event(struct bpf_perf_event_data *ctx) {
+    CPU_FILTER
```
palmtenor: You don't need to do it in the BPF program. Just call `attach_perf_event` with specific CPU.

palmtenor:
[buildbot, ok to test]

PR #1895, opened 2018-07-24
clang: loader: Allow user to override kernel version

joelagnel:
BCC currently requires exactly matching headers. Sometimes this is quite
inconvenient especially if the kernel version is only very slightly
different such as updates in a stable kernel. This patch gives the user
the flexibility to override the the LINUX_VERSION_CODE provided in the
linux kernel headers, so that the eBPF program may load. We also print a
message when this is done, so that the user is warned about the override
happening and that results may be unpredictable.

Also updated the docs.

Signed-off-by: Joel Fernandes <joel@joelfernandes.org>

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -90,6 +90,10 @@ This guide is incomplete. If something feels missing, check the bcc and kernel s
     - [1. Invalid mem access](#1-invalid-mem-access)
     - [2. Cannot call GPL only function from proprietary program](#2-cannot-call-gpl-only-function-from-proprietary-program)
 
+- [Environment Variables](#envvars)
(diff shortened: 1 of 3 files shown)


joelagnel:
This same as the previous PR https://github.com/iovisor/bcc/pull/1880 which got closed.

It has got LGTMs with minor nits corrected, all checks were passed in the old PR so safe to merge.

yonghong-song:
[buildbot, ok to test]

yonghong-song:
Thanks @joelagnel  will merge once the test passed. The github won't allow to merge without passing tests.

joelagnel:
I understand, thanks!

PR #1897, opened 2018-07-25
Allow arbitrary hashtable increments. Fixes #1742

### LINKED ISSUE #1742: Arbitrary increments for maps
Currently `map.increment()` only takes the key, but not the increment. If you want to increment some specific amount (number of bytes, for example), you need to write boilerplate:

```c
u64 zero = 0 , *val;
val = counts.lookup_or_init(&key, &zero);
(*val) += inc;
```

When we can provide `map.increment(key, inc)` instead and do:

```
counts.increment(&key, inc);
```

There are a few places in existing tools that will benefit:

```
$ fgrep '(*val) +=' -r tools
tools/wakeuptime.py:    (*val) += delta;
tools/tcptop.py:        (*val) += size;
tools/tcptop.py:        (*val) += size;
tools/tcptop.py:        (*val) += copied;
tools/tcptop.py:        (*val) += copied;
tools/offwaketime.py:    (*val) += delta;
tools/old/wakeuptime.py:    (*val) += delta;
tools/old/offwaketime.py:    (*val) += delta;
tools/old/offcputime.py:    (*val) += delta;
tools/llcstat.py:    (*val) += ctx->sample_period;
tools/llcstat.py:    (*val) += ctx->sample_period;
tools/tcpsubnet.py:          (*val) += size;
tools/offcputime.py:    (*val) += delta;
```

Some places can switch to `map.increment()` already:

```
$ fgrep '(*val)++' -r tools
tools/tcpretrans.py:        (*val)++;
tools/cachestat.py:    (*val)++;
tools/vfscount.py:    (*val)++;
tools/profile.py:    (*val)++;
tools/old/profile.py:    (*val)++;
tools/old/stackcount.py:    (*val)++;
tools/funccount.py:    (*val)++;
tools/cachetop.py:        (*val)++;
tools/stackcount.py:    (*val)++;
```

javierhonduco:
Right now incrementing some datastructure's values like maps or histograms can
be done with some boilerplate[1] or with `increment` which increments a value
by 1.

This patch allows a second optional parameter to use as the increment.

- [1]:
```
u64 zero = 0, *val;
val = map.lookup_or_init(&key, &zero);
(*val) += inc;
```

Notes:
- Some lines in the documentation where changed because of trailing spaces
deletion
- The test is quite simple right now
- Will update the tools to use `increment` in another PR

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -413,7 +413,7 @@ Return: 0 on success
 
 When used in a program attached to a function entry kprobe, causes the
 execution of the function to be skipped, immediately returning `rc` instead.
-This is used for targeted error injection. 
+This is used for targeted error injection.
 
 bpf_override_return will only work when the kprobed function is whitelisted to
 allow error injections. Whitelisting entails tagging a function with
@@ -756,9 +756,9 @@ Examples in situ:
 
 ### 17. map.increment()
 
-Syntax: ```map.increment(key)```
+Syntax: ```map.increment(key[, increment_amount])```
 
-Increments the key's value by one. Used for histograms.
+Increments the key's value by `increment_amount`, which defaults to 1. Used for histograms.
 
 Examples in situ:
 [search /examples](https://github.com/iovisor/bcc/search?q=increment+path%3Aexamples&type=Code),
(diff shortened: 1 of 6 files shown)


yonghong-song:
The implementation looks good. Could you move the test to the test_clang.py. The current test is not triggered for `ctest` or `make test`, and it is also too small in functionality to have a separate test file. Thanks!

yonghong-song:
[buildbot, ok to test]

PR #1900, opened 2018-07-27
[trace.py]: allow to use STRCMP helper with binary values

tehnerd:
Summary:
sometimes in probe you want to compare char* w/ some predefined value
which is not a string. e.g. setsockopt syscall has signature like this:
sys_setsockopt(int fd, int level, int optname, char* optval, int optlen)
and if you want to catch where/who is setting up specific value you are
forced to compare optval against some predefined array. it's not
possible today w/ trace.py and in this diff i'm adding such ability

Test Plan:
as example: we want to catch setsockopt when someone is setting up
IP_TOS equal to 108
```
trace.py 'sys_setsockopt(int fd, int level, int optname, char* optval,
int optlen)(level==0 && optname == 1 && STRCMP("{0x6C,0x00, 0x00,
0x00}", optval))' -U -M 1 --bin_cmp -v

```
without this new modifier:
```
static inline bool streq_0(char const *ignored, uintptr_t str) {
        char needle[] = "{0x6C,0x00, 0x00, 0x00}";
        char haystack[sizeof(needle)];
        bpf_probe_read(&haystack, sizeof(haystack), (void *)str);
        for (int i = 0; i < sizeof(needle) - 1; ++i) {
                if (needle[i] != haystack[i]) {
                        return false;
                }
        }
        return true;
}

```
// see needle is qouted above

with:

```
static inline bool streq_0(char const *ignored, uintptr_t str) {
        char needle[] = {0x6C,0x00, 0x00, 0x00};
        char haystack[sizeof(needle)];
        bpf_probe_read(&haystack, sizeof(haystack), (void *)str);
        for (int i = 0; i < sizeof(needle) - 1; ++i) {
                if (needle[i] != haystack[i]) {
                        return false;
                }
        }
        return true;
}

...
1855611 1863183 worker          sys_setsockopt   found
```

### DIFF
--- a/man/man8/trace.8
+++ b/man/man8/trace.8
@@ -48,6 +48,9 @@ Print the time column.
 \-C
 Print CPU id.
 .TP
+\-B
+Treat argument of STRCMP helper as a binary value
+.TP
 \-K
 Print the kernel stack for each event.
 .TP
--- a/tools/trace.py
+++ b/tools/trace.py
@@ -48,6 +48,7 @@ class Probe(object):
                 cls.tgid = args.tgid or -1
                 cls.pid = args.pid or -1
                 cls.page_cnt = args.buffer_pages
+                cls.bin_cmp = args.bin_cmp
 
         def __init__(self, probe, string_size, kernel_stack, user_stack):
                 self.usdt = None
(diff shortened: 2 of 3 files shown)


yonghong-song:
[buildbot, ok to test]

palmtenor:
Would be nice to add your test call to `trace_example.txt`:)

tehnerd:
oh. this one wasn't pulled yet. so will add example in this one

PR #1901, opened 2018-07-27
added new helper bpf_get_current_pidns_info and an example 

cneira:
To fix issue https://github.com/iovisor/bcc/issues/1329 , I have created a new bpf helper, patch is included in the that issue. 
This change is for bcc to be able to use that helper, also I created an example for this helper in tracing/examples.

### DIFF
--- a/libbpf-tools/.gitignore
+++ b/libbpf-tools/.gitignore
@@ -1,4 +1,5 @@
 /.output
+/biolatency
 /bitesize
 /cpudist
 /drsnoop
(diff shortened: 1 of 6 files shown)


yonghong-song:
[buildbot, ok to test]

pchaigno:
@cneira Could you run `./script/check-helpers.sh`? I think you're missing a few changes in the documentation.

cneira:
@yonghong-song, to be able to use this helper first the kernel patch in #1329 should be integrated in upstream, could you take a look at the code and example provided in #1329?.

@pchaigno thanks, I'll do that.

PR #1907, opened 2018-08-01
Minor correction to Ubuntu build documentation

maestretti:
`cmake3` was throwing an error that the package did not exist. `cmake` exists and the build was successful using it.


palmtenor:
[buildbot, ok to test]

PR #1908, opened 2018-08-02
sslsniff: add NSS support

jeromemarchand:
It allows sslsniff to monitor tools such as curl that uses NSS library for its SSL/TLS operations.


yonghong-song:
Could you also update the example file (tools/sslsniff_example.txt) and the man page (man/man8/sslsniff.8)?

jeromemarchand:
Good point.

yonghong-song:
[buildbot, ok to test]

PR #1917, opened 2018-08-08
Add name to program too large error message

palmtenor:
When loading multiple programs, it's hard to find out which program triggered the error. This commit adds the function name to the error message.


yonghong-song:
could you rebase on top of trunk? A lot of fc28 failures and Brenden's patch should have fixed all of them.

yonghong-song:
[buildbot, ok to test]

PR #1918, opened 2018-08-09
llcstat: print a nicer error message when hardware events are missing

jeromemarchand:
Hardware events such as CACHE_MISSES and CACHE_REFERENCES are usually
not available on virtual machine. Print a more useful message when
this happen.


yonghong-song:
[buildbot, ok to test]

PR #1930, opened 2018-08-15
provide a printb parameter to suppress printing a new line at the end

ncultra:
This is to support printing output by one of the tools as a json object (not part of this PR). 

Provide an optional parameter, nl=1, that controls the printing of the final new line. To suppress the new line, call as in:
printb(b, nl=0)

calling as in:
printb(b)

will preserve the final new line


yonghong-song:
[buildbot, ok to test]

yonghong-song:
looks okay to me. Thanks!

PR #1931, opened 2018-08-16
[tools/syscount.py]: fix total_ns wrong count

muahao:
It's useful to know each syscall's total latency
in that period, but not the latency of the last time
in that period.

Signed-off-by: Ahao Mu <muahao@linux.alibaba.com>


palmtenor:
[buildbot, ok to test]

yonghong-song:
Thanks for the fix!

PR #1936, opened 2018-08-19
tools.py: Allow args dereference

prasad-joshi:
libc functions like for example eventfd_read()

    int eventfd_read(int fd, eventfd_t *value);

may be passed with pointer argument.

Applying star ('*') to such argument must safely copy a memory location
to the BPF stack.

Signed-off-by: Prasad Joshi <prasadjoshi.linux@gmail.com>

### DIFF
--- a/tools/trace.py
+++ b/tools/trace.py
@@ -393,6 +393,13 @@ BPF_PERF_OUTPUT(%s);
                 bpf_probe_read(&__data.v%d, sizeof(__data.v%d), (void *)%s);
         }
                 """ % (expr, idx, idx, expr)
+                if expr.startswith('*'):
+                    ptr_expr = expr[1:]
+                    return text + """
+        if (%s != 0) {
+            bpf_probe_read(&__data.v%d, sizeof(__data.v%d), (void *)%s);
+        }
+            """ % (ptr_expr, idx, idx, ptr_expr)
                 if field_type in Probe.fmt_types:
                         return text + "        __data.v%d = (%s)%s;\n" % \
                                         (idx, Probe.c_type[field_type], expr)


prasad-joshi:
This is the smallest (all error handling removed) eventfd_read program I tested with.

```
$ cat eventfd.c 
#include <sys/eventfd.h>
#include <stdint.h>
#include <unistd.h>

int main() {
	int fd;
	eventfd_t count;

	fd = eventfd(0, EFD_CLOEXEC);
	while (1) {
		eventfd_write(fd, 1);
		eventfd_read(fd, &count);
		sleep(1);
	}

	return 0;
}

```

Here is the output of trace.py

```
$ sudo ./trace.py 'r:c:eventfd_read (arg1 == 3) "fd = %d, count = %llu", arg1, *arg2'
1225    1225    eventfd         eventfd_read     fd = 3, count = 1
1225    1225    eventfd         eventfd_read     fd = 3, count = 1
1225    1225    eventfd         eventfd_read     fd = 3, count = 1
1225    1225    eventfd         eventfd_read     fd = 3, count = 1
1225    1225    eventfd         eventfd_read     fd = 3, count = 1
```

- Tested the code by passing NULL pointer to eventfd_read()
- On my VM, 
> python test_tools_smoke.py

**reported 7 failures with and without my patches**

Uploading smoke test results.

[WithoutPatchFailures.txt](https://github.com/iovisor/bcc/files/2300596/WithoutPatchFailures.txt)
[WithPatchFailures.txt](https://github.com/iovisor/bcc/files/2300597/WithPatchFailures.txt)

tools/trace.py
```diff
@@ -393,6 +393,13 @@ def _generate_field_assign(self, idx):
                 bpf_probe_read(&__data.v%d, sizeof(__data.v%d), (void *)%s);
         }
                 """ % (expr, idx, idx, expr)
+                if expr.startswith('*'):
+                    ptr_expr = expr[1:]
+                    return text + """
+        if (%s != 0) {
+            bpf_probe_read(&__data.v%d, sizeof(__data.v%d), (void *)%s);
+        }
```
yonghong-song: I think we do not need the `if` condition in the above. User has a dereference and we should just honor it. bpf_probe_read won't crash the kernel. This will be consistent with other rewriter changes related to
bpf_probe_read as we do not really check return value for bpf_probe_read.

Typically, rewriter should handle `*` expressions properly. The exception is for anonymous function parameters like arg1, arg2, etc, which are replaced with PT_REGS_PARM1, etc. where `*PT_REGS_PARM1(ctx)` won't work due to type mismatch. One way could be to only handle `arg1/arg2/arg3/arg4/arg5/arg/retval` here and leave others to be handled by rewriter. But I am okay to handle all cases of `*<...>` here for simplicity as done in the above.

Could you put an example in trace.py and trace_example.txt? For example, the command line for your example in this pull request?

PR #1939, opened 2018-08-21
[tools/argdist.py] Fix a wrong place of help info

muahao:
Signed-off-by: Ahao Mu <muahao@linux.alibaba.com>


palmtenor:
[buildbot, ok to test]

muahao:
argdist -H 'r::__vfs_read(void *file, void *buf, size_t count):size_t
            $entry(count):$latency > 1000000'
is a wrong example which cann't excute success, because lack of ":" and cann't split field correctly. 

So, the right command is: 

argdist -H 'r::__vfs_read(void *file, void *buf, size_t count):size_t:
            $entry(count):$latency > 1000000'

yonghong-song:
[buildbot, test this please]

PR #1940, opened 2018-08-21
Scripts for lock contention measurement in PostgreSQL

erthalion:
Hi,

As part of my small research I've created a few scripts as POC to measure the lock contention from PostgreSQL point of view (to be more precise, they time SpinLocks, LWLocks and PGSemaphores). I was using [lockstat example](https://github.com/goldshtn/linux-tracing-workshop/blob/master/lockstat-solution.py) as a basis, plus I've noticed that there is [locktrace PR](https://github.com/iovisor/bcc/pull/1697).

Of course, probably these scripts are too specific to be included into the repository, but since there are already some DB specific tools here, I wonder if it could be interesting. Anyway I would appreciate any feedback and review. Obviously, I see that there are a lot of repeating parts and no example files, but I just want to be sure that I didn't miss anything important.


yonghong-song:
cc @brendangregg @joelagnel @gdankel

gdankel:
@erthalion Do you notice any perf degradation when tracing? Are any of the spin locks adaptive?

gdankel:
How do you have confidence that you're not changing the behavior by enabling the measurement? If you see that level of slowdown application-wise, it can skew the locking stats. Do you have a separate low-overhead mode where you can just count events? I imagine the uprobe mechanism constitutes the bulk of the overhead.

brendangregg:
thanks, see https://github.com/iovisor/bcc/blob/master/CONTRIBUTING-SCRIPTS.md

missing _examples.txt, man pages, README.md links, test coverage.

To follow the style of other tools, I'd rename pg_spin_locks.py to pg_spinlocks.py or better just pg_spin.py. I was going to suggest this:

- pg_lwlocks.py -> pglwlocks.py
- pg_sema.py -> pgsema.py
- pg_spin_locks.py -> pgspin.py

But I see I used an underscore in mysqld_qslower.py. Hrm. I think I did that because it was becoming alphabet soup (mysqldqslower.py) and hard to spot the "d" (daemon) and "q" (query), which will become more important to see once we add other mysql tools (client ones that lack the "d", and command ones that lack the "q"). pgsema.py isn't so bad, and more in line with other tools.

Looks like you're always using uprobes. Does PostgreSQL not have any USDT probes for lock operations?

brendangregg:
Should also mention: this is already three tools. If you have in mind creating, say, ten or twenty PostgreSQL tracing tools, I'd consider making a separate repository (PostgreSQL bcc/BPF tracing) that you can go to town in, and include the top few most useful ones in bcc plus docs to point people to the full set.

erthalion:
@gdankel
> How do you have confidence that you're not changing the behavior by enabling the measurement? If you see that level of slowdown application-wise, it can skew the locking stats. Do you have a separate low-overhead mode where you can just count events? I imagine the uprobe mechanism constitutes the bulk of the overhead.

I don't, but my assumption is that in most cases this slowdown will be distributed evenly between all backends that try to acquire a lock, which will in turn just increase level of contention, but not skew the stats significantly. You want to say that this assumption can be wrong more often that I expect?

@brendangregg 
> missing _examples.txt, man pages, README.md links, test coverage.

Yep, I know - my idea was to get some feedback and put everything in place when I'll be confident that these scripts are correct from functionality point of view.

> To follow the style of other tools, I'd rename pg_spin_locks.py to pg_spinlocks.py or better just pg_spin.py. I was going to suggest this:

Yeah, sounds good, will rename them.

> Looks like you're always using uprobes. Does PostgreSQL not have any USDT probes for lock operations?

Looks like it does, but only for LWLocks, and I find spin locks and semaphores interesting enough to take a look at them too. But originally I just never though about USDT - what would be the advantage of using it for measuring LWLocks instead of uprobes, is it better in terms of performance?

> Should also mention: this is already three tools. If you have in mind creating, say, ten or twenty PostgreSQL tracing tools, I'd consider making a separate repository (PostgreSQL bcc/BPF tracing) that you can go to town in, and include the top few most useful ones in bcc plus docs to point people to the full set.

Actually what I had in mind is to merge all of them into one script, since there are a lot of duplication in the code. But the idea of a separate repo sounds good, thanks.

erthalion:
Few more questions:

* How to cover this kind of scripts with automated tests? Does it mean just to add some tests to [test_tools_smoke.py](https://github.com/iovisor/bcc/blob/master/tests/python/test_tools_smoke.py), or something more?

* I tried to use USDT, but it turns out that to use them I need to provide a pid of the target process (so far I'm mostly passing path to the binary). PostgreSQL adopts process-based model, which means in this case I need to track all the children, that were forked out - I don't see examples of something similar in this repository, is it possible?

gdankel:
@erthalion Well it depends on the stat and lock type. Let's take your spinlock tracing as an example. Clearly your histogram might now include some outliers directly caused by the tracing. In my understanding, each stat collection involves a context switch and presumably 1000+ cycles, so if the hold time is small (which it should be for spinlocks), then there might be a very significant skew depending on level of contention.

gdankel:
@erthalion But I'm more worried about dynamic locks, e.g. spin-locks with exponential back-off that uses yield and futex when timing out. Then a heavy-handed approach might change the behavior, sometimes dramatically. I've even seen cases where the the application detects a contention problem and tries to self-correct or report a problem. Hopefully you don't have anything like that to worry about.

erthalion:
@gdankel Thanks for the reply

> Well it depends on the stat and lock type. Let's take your spinlock tracing as an example. Clearly your histogram might now include some outliers directly caused by the tracing. 

Yes, but then I want to use this mostly for comparison between patched/unpatched PostgreSQL, so those outliers are going to be the same in both cases and there is a chance that I'll still be able to see the difference, introduced by the modification.

>  In my understanding, each stat collection involves a context switch and presumably 1000+ cycles

Just for me to understand, a switch between which contexts?

> But I'm more worried about dynamic locks, e.g. spin-locks with exponential back-off that uses yield and futex when timing out. Then a heavy-handed approach might change the behavior, sometimes dramatically. I've even seen cases where the the application detects a contention problem and tries to self-correct or report a problem. Hopefully you don't have anything like that to worry about.

I also though about it, but my first experiments show that in the benchmark with the high level of contention without the tracing I've got about 7 out of ~10000 spin-locks, that were backed-off in this way, with the tracing it's about 12 (and it's quite hard to reach the point when PostgreSQL will try to act upon such a spin-lock, thinking that it's stuck).

At the end of the day I see your point, and of course there still may be the cases when the tracing will have more significant impact, but I believe it's like that will almost any tool and I just need to keep this in mind while interpreting a results.

gdankel:
> @gdankel Thanks for the reply
> 
> > Well it depends on the stat and lock type. Let's take your spinlock tracing as an example. Clearly your histogram might now include some outliers directly caused by the tracing.
> 
> Yes, but then I want to use this mostly for comparison between patched/unpatched PostgreSQL, so those outliers are going to be the same in both cases and there is a chance that I'll still be able to see the difference, introduced by the modification.
> 

OK, all good when you're aware of the potential issue.

> > In my understanding, each stat collection involves a context switch and presumably 1000+ cycles
> 
> Just for me to understand, a switch between which contexts?
> 

BPF code is executed in kernel context, so each time a probe is hit a switch to and from kernel switch is needed. It'd be great if we had a user-space library that could run BPF code in user space and share data with the kernel. But it seems like a big project :)

> > But I'm more worried about dynamic locks, e.g. spin-locks with exponential back-off that uses yield and futex when timing out. Then a heavy-handed approach might change the behavior, sometimes dramatically. I've even seen cases where the the application detects a contention problem and tries to self-correct or report a problem. Hopefully you don't have anything like that to worry about.
> 
> I also thought about it, but my first experiments show that in the benchmark with the high level of contention without the tracing I've got about 7 out of ~10000 spin-locks, that were backed-off in this way, and with the tracing it's about 12 (and it's quite hard to reach the point when PostgreSQL will try to act upon such a spin-lock, thinking that it's stuck).

OK cool!

> 
> At the end of the day I see your point, and of course there still may be the cases when the tracing will have more significant impact. I believe it's like that for almost any tool and I just need to keep this in mind while interpreting a results. But maybe you have some suggestions how to make it better in this case?

I think just being aware that the switch to kernel happens is the important thing, and then potentially using a different tracing mechanism like LTTng for those cases that are sensitive. As I stated earlier it would be awesome if BCC could handle these cases with high performance also.

erthalion:
Due to lack of feedback, I'll close this PR, and, as advised, will accumulate all the scripts I'm developing in a separate repo [postgres-bcc](https://github.com/erthalion/postgres-bcc). At the same time I've opened a few issues with questions that were not answered here (plus one extra question) - #2026, #2027, #2028. I would appreciate if someone can answer them.

PR #1941, opened 2018-08-21
Update oomkill.py

smita-koralahalli:
The bcc tool available till date supports from kernel version 4.4 to 4.8 and gives anamolous output as given below for OOM kill of PID and total_pages consumed

17:39:52 Triggered by PID 10505 ("a.out"), OOM kill of PID 1835561760 ("lukB high:%lukB "), 2361183241434822607 pages, loadavg: 1.88 1.70 1.24 3/479 10508.

As mentioned iovisor#1911 function oom_kill_process has been changed after kernel version 4.8 and these few changes in the bcc tool works fine for the kernel till date.

Output after modification has been given below!
23:38:03 Triggered by PID 22426 ("a.out"), OOM kill of PID 22426 ("a.out"), 4085393 pages, loadavg: 6.01 5.09 4.22 1/715 23199

PR #1943, opened 2018-08-22
Update oomkill.py

smita-koralahalli:


yonghong-song:
Could you merge this patch with #1941 ? Otherwise, it will become difficulty to apply cleanly. Thanks!

PR #1950, opened 2018-08-27
Fix bytes/str issue

myllynen:
Looks like commit c0ca99a missed a spot, seen with Python 3.


palmtenor:
Sorry if you have encountered any issue. But I'm still confused with this. Is it not true that `name` is bytes and `prefix`(from `_syscall_prefixes`) is also bytes?

myllynen:
You're right, I thought I checked latest git master around this but apparently I did not, I checked 0.6.1 instead which did you yet have `_syscall_prefixes_` a bytes. Sorry for the confusion, let's disregard this PR. Thanks.

PR #1951, opened 2018-08-28
Serialize python builds

### LINKED ISSUE #1934: parallel makes can fail when building both python2 and python3 support
When attempting to do a RPM scratch build from a recent checkout of bcc from the git repository.  The rpmbuild failed for some targets:

https://koji.fedoraproject.org/koji/taskinfo?taskID=29144905

The ppc64 and ppc64le built fine but the x86_64 failed.  The x86_64 build.log shows it not being able to find bcc-0.6.1:

https://kojipkgs.fedoraproject.org//work/tasks/4909/29144909/build.log

It looks like the python2 and python3 related builds are being built at the same time and interfere with each other.  One quick and dirty fix would be to remove %{?_smp_mflags} from the bcc.spec:

make %{?_smp_mflags}

It would probably be preferable to change the src/python/CMakeLists.txt to serialize the steps for building the python2/python3 support, but I don't know how to implement that in cmake.

philipgian:
Generating sdist for both python2 and python3 at the same time can fail
as they both use the same temporary files.

Make sure that we generate sdist for each `PYTHON_CMD` sequentially.

Fixes #1934

### DIFF
--- a/src/python/CMakeLists.txt
+++ b/src/python/CMakeLists.txt
@@ -20,16 +20,23 @@ endif()
 foreach(PY_CMD ${PYTHON_CMD})
   string(REPLACE "/" "-" PY_CMD_ESCAPED ${PY_CMD})
 
+  set(PY_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/bcc/__init__.py ${CMAKE_CURRENT_BINARY_DIR}/setup.py)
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
cc @drzaeus77 could you take a look?

PR #1955, opened 2018-08-29
tools: Skip backward time entries

olsajiri:
On RHEL7 we are facing timing issues in some bcc-tools
scripts. The time subsystem might return backward timestamp
via bpf_ktime_get_ns and thus screw up latency computation.

This seems to be RHEL7 kernel issue and needs to be fixed,
however meanwhile it'd be nice if affected scripts display
sane values despite the kernel issue.

This patchset harden 2 scripts (xfsslower and ext4dist)
we saw this behaviour so far, to ensure the latency is
always > 0.

### DIFF
--- a/tools/ext4dist.py
+++ b/tools/ext4dist.py
@@ -110,14 +110,20 @@ static int trace_return(struct pt_regs *ctx, const char *op)
     if (tsp == 0) {
         return 0;   // missed start or filtered
     }
-    u64 delta = (bpf_ktime_get_ns() - *tsp) / FACTOR;
+    s64 delta = bpf_ktime_get_ns() - *tsp;
+    start.delete(&pid);
+
+    // Skip entries with backwards time: temp workaround for #728
+    if (delta < 0)
+        return 0;
+
+    delta /= FACTOR;
 
     // store as histogram
     dist_key_t key = {.slot = bpf_log2l(delta)};
     __builtin_memcpy(&key.op, op, sizeof(key.op));
     dist.increment(key);
 
-    start.delete(&pid);
     return 0;
 }
 


yonghong-song:
[buildbot, test this please]

yonghong-song:
the tools smoke test failed.
```
27: ======================================================================
27: FAIL: test_ext4dist (__main__.SmokeTests)
27: ----------------------------------------------------------------------
27:   File "/home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1604/tests/python/test_tools_smoke.py", line 152, in test_ext4dist
27:     self.run_with_duration("ext4dist.py 1 1")
27:   File "/home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1604/tests/python/test_tools_smoke.py", line 32, in run_with_duration
27:     (timeout, full_command), shell=True))
27: AssertionError: 0 != 1
27: 
```
could you take a look?

olsajiri:
hum, I can't see that on my setup.. could you please share what
kernel/config you use and the way you run the smoke test?

thanks

yonghong-song:
When I ran the patch on ubuntu16.04, I got the following error:

```
$ sudo ./ext4dist.py 1 1
LLVM ERROR: Cannot select: 0x4144510: i64 = sdiv 0x40eeef0, 0x37457c8 [ORD=2] [ID=11]
  0x40eeef0: i64,ch = CopyFromReg 0x403ab90, 0x4144d28 [ORD=2] [ID=10]
    0x4144d28: i64 = Register %vreg1 [ID=1]
  0x37457c8: i64 = Constant<1000> [ID=2]
In function: trace_read_return
yhs@ubuntu:~/work/bcc/tools$
```

The BPF does not support signed division and the LLVM will flag a fatal error if it is used.
This may depend on the compiler version installed in the system and compiler optimization. I did not check but it seems some compiler version does not generate sdiv?

Anyway, could you change `delta` type to u64? and you can use `if ((s64)delta < 0)` for comparison.

olsajiri:
nice ;-) will make that change and resubmit, thanks

PR #1959, opened 2018-09-03
Add Perl support for ucalls / uflow / ustat

myllynen:
Sort language entries while at it.

### DIFF
--- a/man/man8/ucalls.8
+++ b/man/man8/ucalls.8
@@ -1,27 +1,29 @@
 .TH ucalls 8  "2016-11-07" "USER COMMANDS"
 .SH NAME
-ucalls, javacalls, pythoncalls, rubycalls, phpcalls \- Summarize method calls
+ucalls, javacalls, perlcalls, phpcalls, pythoncalls, rubycalls \- Summarize method calls
 from high-level languages and Linux syscalls.
 .SH SYNOPSIS
 .B javacalls [-h] [-T TOP] [-L] [-S] [-v] [-m] pid [interval]
+.br
+.B perlcalls [-h] [-T TOP] [-L] [-S] [-v] [-m] pid [interval]
+.br
+.B phpcalls [-h] [-T TOP] [-L] [-S] [-v] [-m] pid [interval]
 .br
 .B pythoncalls [-h] [-T TOP] [-L] [-S] [-v] [-m] pid [interval]
 .br
 .B rubycalls [-h] [-T TOP] [-L] [-S] [-v] [-m] pid [interval]
 .br
-.B phpcalls [-h] [-T TOP] [-L] [-S] [-v] [-m] pid [interval]
-.br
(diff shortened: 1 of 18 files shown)


yonghong-song:
[buildbot, test this please]

myllynen:
I don't think anything special is needed for Perl; however the reason I added them is because similar wrappers exists for other languages already, for example:

```
$ ls bcc.git/tools/python{calls,flow,stat}.sh                            
bcc.git/tools/pythoncalls.sh  bcc.git/tools/pythonflow.sh  bcc.git/tools/pythonstat.sh
```

Thanks.

yonghong-song:
Ok, sounds good then!

brendangregg:
The only thing I don't like is the lack of specific examples; feel free to fix this in a follow up PR. Eg:

```
ls -lh perlcalls_example.txt
lrwxr-xr-x  1 bgregg  2119906183    22B Sep  4 11:36 perlcalls_example.txt@ -> lib/ucalls_example.txt
```

That's kinda a hack. Better than nothing, but as a Perl user I'm reading an example that's actually demonstrating Java. At some point I'd delete this symlink and replace it with a real perlcalls_example.txt file of Perl examples, and the same for all the others.

myllynen:
I agree that the it's not ideal at the moment but I was following the earlier convention with the other supported languages (Node, PHP, Python, Ruby), they all link to the generic u*txt versions:

```
$ ls -1l bcc.git/tools/*stat_example.txt
lrwxrwxrwx. 1 testuser testuser   21 2017-12-06 22:10 bcc.git/tools/javastat_example.txt -> lib/ustat_example.txt
lrwxrwxrwx. 1 testuser testuser   21 2017-12-06 22:10 bcc.git/tools/nodestat_example.txt -> lib/ustat_example.txt
lrwxrwxrwx. 1 testuser testuser   21 2018-09-04 20:40 bcc.git/tools/perlstat_example.txt -> lib/ustat_example.txt
lrwxrwxrwx. 1 testuser testuser   21 2017-12-06 22:10 bcc.git/tools/phpstat_example.txt -> lib/ustat_example.txt
lrwxrwxrwx. 1 testuser testuser   21 2017-12-06 22:10 bcc.git/tools/pythonstat_example.txt -> lib/ustat_example.txt
lrwxrwxrwx. 1 testuser testuser   21 2017-12-06 22:10 bcc.git/tools/rubystat_example.txt -> lib/ustat_example.txt
```

myllynen:
Glad you like it! FWIW, Performance Co-Pilot (PCP, https://pcp.io) now has BCC plugin (https://github.com/performancecopilot/pcp/tree/master/src/pmdas/bcc) which supports ucalls and ustat modules so in case you'd like to leave things running in the background for a longer period of time or collect also OS/DB/etc metrics at the same time, that might be something worth considering.

Issue #269, opened 2015-10-13
mismatch structure size between bpf syscall and bpf program in bcc


yonghong-song:
For structure like:
struct routing_key_1 {
    u32 key_field_1;
    char key_field_0[6];
};

LLVM IR:
%struct.routing_key_1 = type { i32, [6 x i8] }

The bcc compiler infers size 12 and uses it for bpf create_map syscall.
The bpf program itself, however, uses the size 10. This mismatch
causes verifier failure.

To reproduce:
yhs@ubuntu:~/work/bcc/examples$ cat t.c
# include <uapi/linux/bpf.h>
# include <uapi/linux/if_ether.h>
# include <uapi/linux/if_packet.h>
# include <uapi/linux/ip.h>
# include <linux/skbuff.h>
# include <linux/netdevice.h>
# include <linux/pkt_cls.h>

enum ErrorCode {
    p4_pe_no_error,
    p4_pe_index_out_of_bounds,
    p4_pe_out_of_packet,
    p4_pe_header_too_long,
    p4_pe_header_too_short,
    p4_pe_unhandled_select,
    p4_pe_checksum,
};
# define EBPF_MASK(t, w) ((((t)(1)) << (w)) - (t)1)
# define BYTES(w) ((w + 7) / 8)

struct ethernet_t {
    char dstAddr[6]; /\* 48 bits _/
    u8 valid; /_ 1 bits _/
};
struct standard_metadata_t {
    u16 ingress_port; /_ 9 bits _/
    u32 packet_length; /_ 32 bits _/
    u16 egress_spec; /_ 9 bits _/
    u16 egress_port; /_ 9 bits _/
    u32 egress_instance; /_ 32 bits _/
    u32 instance_type; /_ 32 bits _/
    u32 clone_spec; /_ 32 bits _/
    u8 _padding; /_ 5 bits _/
};
struct ipv4_t {
    u32 srcAddr; /_ 32 bits _/
    u8 valid; /_ 1 bits */
};
struct ebpf_headers_t {
    struct ethernet_t ethernet;
    struct ipv4_t ipv4;
};
struct ebpf_metadata_t {
    struct standard_metadata_t standard_metadata;
};
struct routing_key_1 {
    u32 key_field_1;
    char key_field_0[6];
};
enum routing_actions_0 {
    routing_nop,
};
struct routing_value_2 {
    u32 action;
    union {
        struct {
        } nop;
    } u;
};
[...]

PR #1961, opened 2018-09-06
tcp tools: finalize data type conversion

myllynen:
Finish the conversion started with commit 36ce1124.

Minor tweaks to make the tools consistent within themselves.

### DIFF
--- a/tools/tcpaccept.py
+++ b/tools/tcpaccept.py
@@ -49,7 +49,6 @@ bpf_text = """
 
 // separate data structs for ipv4 and ipv6
 struct ipv4_data_t {
-    // XXX: switch some to u32's when supported
     u64 ts_us;
     u32 pid;
     u32 saddr;
--- a/tools/tcpconnect.py
+++ b/tools/tcpconnect.py
@@ -57,7 +57,6 @@ BPF_HASH(currsock, u32, struct sock *);
 
 // separate data structs for ipv4 and ipv6
 struct ipv4_data_t {
-    // XXX: switch some to u32's when supported
     u64 ts_us;
     u32 pid;
     u32 saddr;
--- a/tools/tcpconnlat.py
+++ b/tools/tcpconnlat.py
@@ -81,7 +81,6 @@ BPF_HASH(start, struct sock *, struct info_t);
 
 // separate data structs for ipv4 and ipv6
 struct ipv4_data_t {
-    // XXX: switch some to u32's when supported
     u64 ts_us;
     u32 pid;
     u32 saddr;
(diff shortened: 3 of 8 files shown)


myllynen:
Please double check this, I'm not 100% __builtin_memcpy() is the right choice here but it seems to work. Thanks.

myllynen:
(Btw, after this only the *slower tools are in the need of this sort of conversion.)

yonghong-song:
[buildbot, test this please]

yonghong-song:
@myllynen some of changes may be problematic. For a structure with pads, The C compiler will not initialize these pads and verifier may complain. In such cases, it is still necessary to initialize the whole structure first and then assign individual fields.

myllynen:
@yonghong-song It would be nice to know exactly which cases could be problematic, as mentioned here such changes were to make the tools consistent within themselves, more precisely:

* tcpdrop, tcpretrans: use the same initialization with IPv4 as is already used with IPv6
* tcplife: use the same initialization with IPv6 as is already used with IPv4

Thanks.

pchaigno:
@yonghong-song is referring to the issue documented in #269. Not all objects are initialized the same way because of that issue. If you look at the Jenkins' logs, you'll see that the tests are failing because the kernel verifier is rejecting the programs. You should be able to reproduce locally easily if you run the tools.

yonghong-song:
[buildbot, test this please]

yonghong-song:
There is a missing change for `tcptop.py` for ipv6 (probably you did not test it),
```
 def get_ipv6_session_key(k):
     return TCPSessionKey(pid=k.pid,
-                         laddr=inet_ntop(AF_INET6, pack("QQ", k.saddr0, k.saddr1)),
+                         laddr=inet_ntop(AF_INET6, k.saddr),
                          lport=k.lport,
-                         daddr=inet_ntop(AF_INET6, pack("QQ", k.daddr0, k.daddr1)),
+                         daddr=inet_ntop(AF_INET6, k.daddr),
                          dport=k.dport)
```
Could you add this change and resubmit?
Thanks!

myllynen:
The current PR actually would do:

```
def get_ipv6_session_key(k):
    return TCPSessionKey(pid=k.pid,
                         laddr=inet_ntop(AF_INET6, pack("QQ", k.saddr)),
                         lport=k.lport,
                         daddr=inet_ntop(AF_INET6, pack("QQ", k.daddr)),
                         dport=k.dport)
```

Which is like for IPv4 except uses "QQ" instead of "I". If you still prefer, I'll change this to your proposal? Thanks.

yonghong-song:
`pack("QQ", k.saddr)` probably won't work. The following is what I got:
```
  File "./tcptop.py", line 266, in <module>
    key = get_ipv6_session_key(k)
  File "./tcptop.py", line 204, in get_ipv6_session_key
    laddr=inet_ntop(AF_INET6, pack("QQ", k.saddr)),
struct.error: pack expected 2 items for packing (got 1)
```
Unfortunately, I did not see pack has an encoding letter for
16-byte integer.
Maybe just change to my proposed change is easier.

myllynen:
Ok, I've updated the PR to match your suggestion. Thanks.

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

PR #1962, opened 2018-09-07
Add move constructor to C++ USDT instance

palmtenor:
This way, it's easier to move initialized USDT instance around without having to re-read everything.

Piggy-back a small logging fix

### DIFF
--- a/src/cc/api/BPF.cc
+++ b/src/cc/api/BPF.cc
@@ -59,10 +59,13 @@ StatusTuple BPF::init(const std::string& bpf_program,
                       const std::vector<USDT>& usdt) {
   std::string all_bpf_program;
 
+  usdt_.reserve(usdt.size());
   for (const auto& u : usdt) {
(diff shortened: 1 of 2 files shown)


src/cc/api/BPF.cc
```diff
@@ -733,6 +736,18 @@ USDT::USDT(const USDT& usdt)
       name_(usdt.name_),
       probe_func_(usdt.probe_func_) {}
 
+USDT::USDT(USDT&& usdt) noexcept
+    : initialized_(usdt.initialized_),
```
yonghong-song: Since `usdt.initialized_` is set below, do we need this initialization?

palmtenor: This is setting the target's value.Below is to reset the value of the origin (that is being moved away)

yonghong-song: Oh, that is true. Thanks for reminding me of this!

PR #1965, opened 2018-09-10
tcpretrans: use user-space PID displayed to the user

myllynen:
PID collected in tcpretrans' trace_event() is displayed under the
label of "PID" to the user so it would seem more appropriate to use
TGID as PID not kernel's PID.


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

PR #1974, opened 2018-09-18
add vfsrwcount and corresponding example.txt

yupeng0921:
I wanted to find a method to know how many IOs each process generates on every disks, the google search leaded me to the vfsstat.py script in bcc project, but I found vfsstat.py couldn't show the read/write bytes for each process, so I wrote the vfsrwcount.py, which could print "[pid] [comm] [file_name] [byte_count]" for vfs_read and vfs_write functions. Hope it could be useful for other people.

### DIFF
--- a/tools/vfsrwcount.py
+++ b/tools/vfsrwcount.py
@@ -1,0 +1,106 @@
+#!/usr/bin/env python
+# @lint-avoid-python-3-compatibility-imports
+#
+# vfsrwcount  Count vfs_read, vfs_write bytes
+#             per process per file
+
+from __future__ import (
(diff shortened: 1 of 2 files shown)


yonghong-song:
cc @brendangregg

brendangregg:
Tracing VFS reads & writes covers not only some disk I/O (with the exception of major faults), but also file system cache I/O, and other I/O types entirely like pipes and likely networking. This script does not do what it states it does.

It's also missing a man page, and it's so similar to vfscount I would have just added a per-process option to vfscount.

To solve disk I/O by process, use the bio* tools like biotop.

PR #1981, opened 2018-09-21
Remove unused ts field in val_t in opensnoop.py.

bolinfest:
Note this allows us to eliminate the `bpf_ktime_get_ns()`
call from `trace_entry()`.

### DIFF
--- a/tools/opensnoop.py
+++ b/tools/opensnoop.py
@@ -61,7 +61,6 @@ bpf_text = """
 
 struct val_t {
     u64 id;
-    u64 ts;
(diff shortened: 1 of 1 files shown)


bolinfest:
While I'm here, why is `data.ts = tsp / 1000;` done inside `trace_return()` and then the division by `1000000` done inside `print_event()`? That is, why not just divide by 1e9 inside `print_event()`? Is it an issue that the Python cannot represent numbers that large so we have to divide by 1000 while we have it as a `u64` in C?

palmtenor:
[buildbot, test this please]

PR #1991, opened 2018-09-29
lua: codegen tests, tracking inverse form of null guards, IPv6 header builtins

vavrusa:
All kinds of fixes for the Lua compiler to sync with my branch.

Added:
* direct skb->data access (in addition to BPF_LDABS, BPF_LDIND)
* loads and stores from skb->data and map value backed memory
* unified variable source tracking (ptr_to_ctx, ptr_to_map[_or_null], ptr_to_skb, ptr_to_pkt, ptr_to_stack)
* BPF constants introduced between 4.10-4.15
* bpf.dump_string() to dump generated assembly (text) to string
* add builtins for parsing variable size IPv6 headers

Fixes:
* inverse pointer null check tracking (e.g. `if x ~= nil`)
* dissectors for map value backed memory
* ljsyscall extensions when the version is too old
* KPRI nil variables used in conditions
* wrongly elided constant materialization on condition-less jumps
* loads/stores from stack memory using variable offset
* shadowing of variables
* interpreting integer type cdata as integers (e.g. `uint64_t`)
* fixes `ntohl()` with NET_OFF reads


yonghong-song:
[buildbot, test this please]

yonghong-song:
Any lua expert here could help review this patch? Thanks!

yonghong-song:
I will merge this. We can fix if any issues come up later.

vavrusa:
(If it's any consolation, I don't think anyone besides me is using it and it's fairly disconnected from the rest of the bcc)

Issue #1998, opened 2018-10-05
Tracing USDTs in executables compiled with -pie


zilder:
Hi,

I have a problems tracing tracepoints in postgres built with `-pie` flag. This is how postgres is configured:

```
./configure --enable-dtrace --enable-cassert --enable-debug CFLAGS='-ggdb -g3 -O0 -pie' --prefix=$HOME/pg10
```

I can see tracepoints embedded using `readelf`. Let's say I want to trace transaction start event.
```
readelf -n ~/pg10/bin/postgres
...
  stapsdt              0x0000003d	NT_STAPSDT (SystemTap probe descriptors)
    Provider: postgresql
    Name: transaction__start
    Location: 0x000000000017e272, Base: 0x00000000008d3911, Semaphore: 0x0000000000998f68
    Arguments: 4@%eax
...
```

There is a simple script I use to trace event:

```
#!/usr/bin/env python

from bcc import BPF, USDT
import sys

if len(sys.argv) < 2:
    print("USAGE: {0} PID".format(sys.argv[0]))
    exit()
pid = sys.argv[1]

prog = """
#include <uapi/linux/ptrace.h>
int do_trace(struct pt_regs *ctx) {
    bpf_trace_printk("Hello, transaction__start\\n");
    return 0;
};
"""

u = USDT(pid=int(pid))
u.enable_probe(probe="transaction__start", fn_name="do_trace")

for p in u.enumerate_probes():
    print('{}\t0x{:02x}'.format(p.name, p.semaphore))
print("Active probes: ", u.enumerate_active_probes())

b = BPF(text=prog, usdt_contexts=[u])
b.trace_print()
```
When I run script I get this (output was truncated):
```
sudo ./pg.py 32737
...
b'transaction__start'	0x998f68
b'transaction__commit'	0x998f6a
b'transaction__abort'	0x998f6c
...
Active probes:  []
```
32737 here is the PID of postgres backend that handles connection from `psql` in a separate terminal. So `bcc` can see those probes in executable, but none of transactions were actually traced. And note that enumerate_active_probes() doesn't return any probes.
I inspected `bcc` with `gdb` and found two things in `enable_probe` call.
1. `in_shared_object()` returns true even though it is an executable binary file. `readelf` shows the same thing:

```
readelf -h ~/pg10/bin/postgres
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00 
  Class:                             ELF64
[...]

PR #2006, opened 2018-10-09
USDT: take into account the result of Probe::enable()

zilder:
Hi,

Currently Context::enable_probe() discards the result of Probe::enable() even if it fails. This is something that I've already mentioned in #1998 (it didn't show any warning when tried to read semaphore value from unreadable memory location). Today I hit another case when I tried to setup USDT context by passing only path to binary without specifying PID. Here is a tiny patch that fixes this.


palmtenor:
[buildbot, test this please]

palmtenor:
LGTM

Issue #2002, opened 2018-10-09
killsnoop: str/bytes TypeError on output


r4f4:
Steps to Reproduce:
1. $ sudo /usr/share/bcc/tools/killsnoop
2. open new terminal emulator 

Actual results:

$ sudo /usr/share/bcc/tools/killsnoop
TIME      PID    COMM             SIG  TPID   RESULT

Traceback (most recent call last):
  File "_ctypes/callbacks.c", line 234, in 'calling callback function'
  File "/usr/lib/python3.6/site-packages/bcc/table.py", line 572, in raw_cb_
    callback(cpu, data, size)
  File "./killsnoop", line 143, in print_event
    event.pid, event.comm, event.sig, event.tpid, event.ret))
TypeError: %b requires a bytes-like object, or an object that implements __bytes__, not 'str'

(and this traceback repeats)

palmtenor:
Fixed by #2007

PR #2007, opened 2018-10-10
fix killsnoop.py srr/bytes type error

yonghong-song:
Fix issue #2002


yonghong-song:
cc @r4f4 could you verify whether this change fixed your issue or not?

r4f4:
@yonghong-song Yes, it solves the issue. Thank you!

Issue #2017, opened 2018-10-26
Reading from unaligned structs (class) is causing invalid values.


chaitu-tk:
When the below structure is used in the BPF Code, packet_num works fine, but if the same is retrieved using python it doesn't work.
struct reorder_info {
        u16 seq_num;   
        u32 packet_num; 
}__packed;

In the above structs, even though its packed, python inserts "holes" to try and align the struct because BCC doesn't understand the "__packed" macro as its from Linux Kernel. While accessing "packet_num" the python code starts at the hole offset causing invalid values.

BPF should understand packed from and refrain from aligning the struct.

yonghong-song:
Do you have a concrete reproducible test case? You mean in python with a ctype structure definition like above, right? Python C type can specify packed (attribute `_pack_ = 1`), maybe you want to try?

chaitu-tk:
Yes, I do have a reproducible case. The above struct is defined in the BPF C code with __packed attribute and stored as a BPF_TABLE("hash", tp).
In python i am using bpf_handle["tp"][0]  (Type is 'class bcc.reorder_info') to fetch the contents and thats where the problem comes, I don't have control over BCC which is converting the BPF_TABLE as a class.

chaitu-tk:
[packed_structs_bpf.py.zip](https://github.com/iovisor/bcc/files/2521583/packed_structs_bpf.py.zip)
 This is an example to show the issue. To get it working just swap the a, b definitions and it works.

yonghong-song:
Thanks for the test case. I can reproduce the issue. The following change can fix the issue.

```
diff --git a/src/cc/json_map_decl_visitor.cc b/src/cc/json_map_decl_visitor.cc
index d54a71c..c7fe9b8 100644
--- a/src/cc/json_map_decl_visitor.cc
+++ b/src/cc/json_map_decl_visitor.cc
@@ -159,8 +159,12 @@ bool BMapDeclVisitor::VisitRecordDecl(RecordDecl *D) {
   result_ += "]";
   if (D->isUnion())
     result_ += ", \"union\"";
-  else if (D->isStruct())
-    result_ += ", \"struct\"";
+  else if (D->isStruct()) {
+    if (SkipPadding)
+      result_ += ", \"struct\"";
+    else
+      result_ += ", \"struct_packed\"";
+  }
   result_ += "]";
   return true;
 }
diff --git a/src/python/bcc/__init__.py b/src/python/bcc/__init__.py
[...]

PR #2020, opened 2018-10-30
allow packed structure in bpf program in python API

yonghong-song:
Fix issue #2017.

For python programs, the map data passed from C++ library
is parsed through the key/value desc types which are
generated by C++ json map declaration visitor.

The map declaration visitor visits the map key/value
declaration types and generate a string to represent
the type, which is later used by python to reconstruct
the cttype.

The map declaration already tries to add all the padding
to the type string in order to make sure C++ and python
see the same layout.

This patch further added packed support such that if
C++ json map visitor has applied padding, the python
type reconstructor is free to add _pack_=1 for structure
type since the structure is already packed.

For example, for a type,
```
  struct t { char a; int b; }
```
the structure after json visitor will look like
```
  struct t { char a; char __pad[3]; int b; }
```

If the type is
```
  struct t { char a; int b; } __packed;
```
the structure after json visitor will look like
```
  struct t { char a; int b; }
```

In either case, it will be okay to add __packed attribute
for the type generated by json map visitor in order to
match the original declaration.

Thanks Chaitanya for filing the issue and providing the test case!

### DIFF
--- a/src/cc/json_map_decl_visitor.cc
+++ b/src/cc/json_map_decl_visitor.cc
@@ -159,8 +159,12 @@ bool BMapDeclVisitor::VisitRecordDecl(RecordDecl *D) {
   result_ += "]";
   if (D->isUnion())
     result_ += ", \"union\"";
-  else if (D->isStruct())
-    result_ += ", \"struct\"";
+  else if (D->isStruct()) {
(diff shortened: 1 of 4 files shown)


yonghong-song:
cc @chaitu-tk

PR #2024, opened 2018-11-02
Update bindings for custom offset kprobes

sandip4n:
This updates the python and lua bindings for libbcc to be able to support attaching kprobes at custom offsets.

Update 03/11: Change title and description. Remove `uprobe` related changes.

~~This adds support for custom offset `uprobes`. The changes made previously for custom offset `kprobes` were also missing a few bits for the Python and Lua APIs. This also provides fixes for `powerpc64` to ensure `uprobe` events work as expected.~~

### DIFF
--- a/src/lua/bcc/bpf.lua
+++ b/src/lua/bcc/bpf.lua
@@ -211,9 +211,10 @@ function Bpf:attach_kprobe(args)
   local event = args.event or ""
   local ptype = args.retprobe and "r" or "p"
   local ev_name = string.format("%s_%s", ptype, event:gsub("[%+%.]", "_"))
+  local offset = args.fn_offset or 0
   local retprobe = args.retprobe and 1 or 0
 
-  local res = libbcc.bpf_attach_kprobe(fn.fd, retprobe, ev_name, event)
+  local res = libbcc.bpf_attach_kprobe(fn.fd, retprobe, ev_name, event, offset)
 
   assert(res >= 0, "failed to attach BPF to kprobe")
   self:probe_store("kprobe", ev_name, res)
--- a/src/lua/bcc/libbcc.lua
+++ b/src/lua/bcc/libbcc.lua
@@ -43,7 +43,7 @@ typedef void (*perf_reader_raw_cb)(void *cb_cookie, void *raw, int raw_size);
 typedef void (*perf_reader_lost_cb)(void *cb_cookie, uint64_t lost);
 
 int bpf_attach_kprobe(int progfd, int attach_type, const char *ev_name,
-                      const char *fn_name);
+                      const char *fn_name, uint64_t fn_offset);
 
 int bpf_detach_kprobe(const char *ev_name);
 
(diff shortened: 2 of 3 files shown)


src/cc/bcc_elf.c
```diff
@@ -184,11 +184,57 @@ int bcc_elf_foreach_usdt(const char *path, bcc_elf_probecb callback,
...
+      if (section_idx)
+        *section_idx = index;
+      return section;
+    }
+  }
+
+  return NULL;
+}
+
+static Elf_Data * get_section_elf_data(Elf *e, const char *section_name) {
```
palmtenor: If you don't mind, maybe still keep the functions at where they were before so it's easier to see what's changed?

src/cc/bcc_syms.cc
```diff
@@ -397,6 +397,38 @@ void bcc_symbol_free_demangle_name(struct bcc_symbol *sym) {
...
+    return;
+
+#if defined(__powerpc64__) && __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+  /*
+   * The PowerPC 64-bit ELF v2 ABI says that the 3 most significant bits
+   * in the st_other field of the symbol table specifies the number of
+   * instructions between a function's Global Entry Point (GEP) and Local
+   * Entry Point (LEP).
+   */
+  switch (sym->other >> 5) {
```
palmtenor: If this is all what this `other` thing does (adjusting offset), maybe do it in the iteration function so that we could still keep the (potentially user-facing) callback unchanged...

src/cc/bcc_elf.c
sandip4n: Sure

src/cc/bcc_syms.cc
sandip4n: There is a slight problem I am seeing if I do it in the iteration function. Upon printing the user stacks, the symbol offsets are all being reduced by the factor deduced from `other`.

For example, when we should be seeing a stack trace like
```
__GI___inet_pton+0x8 [libc-2.27.so]
gaih_inet.constprop.7+0x1040 [libc-2.27.so]
getaddrinfo+0x164 [libc-2.27.so]
[unknown] [ping]
generic_start_main.isra.0+0x138 [libc-2.27.so]
__libc_start_main+0xc4 [libc-2.27.so]
```
we instead end up with
```
__GI___inet_pton+0x0 [libc-2.27.so]
gaih_inet.constprop.7+0x1038 [libc-2.27.so]
getaddrinfo+0x15c [libc-2.27.so]
[unknown] [ping]
generic_start_main.isra.0+0x130 [libc-2.27.so]
__libc_start_main+0xbc [libc-2.27.so]
```
So, the symbol offsets are all off by 8. One way to prevent this would be to do the offset adjustment in the user-facing API like it was done previously as that is the only place we know whether to use a symbol name or a direct binary offset. This way, we can keep the original symbol offsets (aka `st_value`) intact and not mess up the stack trace entries.

I am wondering if there is a simpler way to do this. Your thoughts?

yonghong-song:
The patch (kprobe only) looks good to me. Thanks!

yonghong-song:
[buildbot, test this please]

Issue #2029, opened 2018-11-03
BCC fails to resolve symbols of executable running in different mount namespaces. 


vijunag:
Hi BCC, 
BCC fails to resolve symbols of exe when it is running on a different mount namespace with the below error. 
WARNING: Couldn't find .text section in XXX
WARNING: BCC can't handle sym look ups for XXX.

vijunag:
Issue is fixed.

Issue #1990, opened 2018-09-29
WARNING: Couldn't find .text section in /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java


tekumara:
When running `trace` inside the `zlim/bcc` container, to debug a java application running inside another container, I get a bunch of `WARNING: Couldn't find .text section` and `WARNING: BCC can't handle sym look ups` errors and a large number of `[unknown]` in the stack trace:

```
$ sudo docker run -it --rm   --privileged   -v /lib/modules:/lib/modules:ro   -v /usr/src:/usr/src:ro   -v /etc/localtime:/etc/localtime:ro   --workdir /usr/share/bcc/tools   --pid=host   zlim/bcc
root@8f2d7d06add3:/usr/share/bcc/tools# /usr/share/bcc/tools/trace 'SyS_write (arg1==1) "%s", arg2' -U -p `pidof java`
PID     TID     COMM            FUNC             -
4472    13660   java            SyS_write        2018-09-29 10:47:23,925 [info] application-akka.actor.default-dispatche
WARNING: Couldn't find .text section in /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java
WARNING: BCC can't handle sym look ups for /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/javaWARNING: Couldn't find .text section in /lib/x86_64-linux-gnu/libresolv-2.24.so
WARNING: BCC can't handle sym look ups for /lib/x86_64-linux-gnu/libresolv-2.24.soWARNING: Couldn't find .text section in /lib/x86_64-linux-gnu/libnss_dns-2.24.so
WARNING: BCC can't handle sym look ups for /lib/x86_64-linux-gnu/libnss_dns-2.24.soWARNING: Couldn't find .text section in /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/libsunec.so
WARNING: BCC can't handle sym look ups for /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/libsunec.soWARNING: Couldn't find .text section in /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/libnio.so
WARNING: BCC can't handle sym look ups for /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/libnio.soWARNING: Couldn't find .text section in /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/libnet.so
[...]

PR #2030, opened 2018-11-03
Fix symbol resolution across mount namespaces (#2029)

vijunag:
* Symbol lookup fails when the process is running on a different namespace. 
 * ProcSyms already has the capability to switch mount namespaces to resolve symbols. 
 * Identified bug where the BPF doesn't migrate to process mount namespace before opening the symbol files and fails to locate the file. 
* Patch makes sure that 'ProcMountNSGuard g(ps->mount_ns_instance_.get())' is created every single time a text section for the process is loaded. 
* Patch was tested by  1) creating docker container and running the process 2) running "bcc/tools/profile" script on the global PID namespace of the process 

Test - Performed a test by running bcc/tools/profile on a PID running in a container and found the resolution to be working fine. 
`#define  _GNU_SOURCE
#include <link.h>
#include <stdio.h>
#include <unistd.h>
#include <sys/auxv.h>
#include <string.h>

int len;
int baz()
{
   int i=0;
   char a[8192], b[8192];

label:
   for(i=0;i<10000;++i) {
     memcpy(a,b,len);
   }

   goto label;
}

int foo()
{
	baz();
}

int bar()
{
	foo();
}

int main(int argc, char **argv)
{
	printf("Start profiling\n");
         len=8192;
	bar();
	return 1;
}`

python profile -U -p 26966 1
Sampling at 49 Hertz of PID 26966 by user stack for 1 secs.
warning: JITed object file architecture unknown is not compatible with target architecture i386:x86-64.

    [unknown]
    foo
    bar
    main
    __libc_start_main
    -                sampleme (26966)
        1

    baz
    foo
    bar
    main
    __libc_start_main
    -                sampleme (26966)

### DIFF
--- a/src/cc/bcc_syms.cc
+++ b/src/cc/bcc_syms.cc
@@ -122,13 +122,13 @@ int ProcSyms::_add_load_sections(uint64_t v_addr, uint64_t mem_sz,
 }
 
 void ProcSyms::load_exe() {
+  ProcMountNSGuard g(mount_ns_instance_.get());
   std::string exe = ebpf::get_pid_exe(pid_);
   Module module(exe.c_str(), mount_ns_instance_.get(), &symbol_option_);
 
   if (module.type_ != ModuleType::EXEC)
     return;
 
-  ProcMountNSGuard g(mount_ns_instance_.get());
 
   bcc_elf_foreach_load_section(exe.c_str(), &_add_load_sections, &module);
 
@@ -163,6 +163,7 @@ int ProcSyms::_add_module(const char *modname, uint64_t start, uint64_t end,
     // It only gives the mmap offset. We need the real offset for symbol
     // lookup.
     if (module.type_ == ModuleType::SO) {
+      ProcMountNSGuard g(ps->mount_ns_instance_.get());
       if (bcc_elf_get_text_scn_info(modname, &module.elf_so_addr_,
                                     &module.elf_so_offset_) < 0) {
         fprintf(stderr, "WARNING: Couldn't find .text section in %s\n", modname);


src/cc/bcc_syms.cc
```diff
@@ -122,13 +122,13 @@ int ProcSyms::_add_load_sections(uint64_t v_addr, uint64_t mem_sz,
...
+  ProcMountNSGuard g(mount_ns_instance_.get());
   std::string exe = ebpf::get_pid_exe(pid_);
   Module module(exe.c_str(), mount_ns_instance_.get(), &symbol_option_);
 
   if (module.type_ != ModuleType::EXEC)
     return;
 
-  ProcMountNSGuard g(mount_ns_instance_.get());
 
   bcc_elf_foreach_load_section(exe.c_str(), &_add_load_sections, &module);
```
vijunag: We have to move to mount namespace before accessing the file /proc/pid/exe file.

src/cc/bcc_syms.cc
```diff
@@ -163,6 +163,7 @@ int ProcSyms::_add_module(const char *modname, uint64_t start, uint64_t end,
     // It only gives the mmap offset. We need the real offset for symbol
     // lookup.
     if (module.type_ == ModuleType::SO) {
+      ProcMountNSGuard g(ps->mount_ns_instance_.get());
```
vijunag: Ditto. Migrate to task's mount namespace before accessing the symbol file.

yonghong-song:
@tekumara could you check whether this fixed your issue (#1990)?
@palmtenor could you comment on the change as well?

tekumara:
Using this PR, I run `bcc` inside a container to debug another container, and  I don't get the warnings! Thank you! 🙇
However I still get `[unknown]`s in the stack trace (see below). Should these be resolved?

```
host $ sudo docker run -it --privileged   -v /sys/kernel/debug:/sys/kernel/debug:rw -v /lib/modules:/lib/modules:ro   -v /usr/src:/usr/src:ro   -v /etc/localtime:/etc/localtime:ro   --workdir /usr/share/bcc/tools   --pid=host   bcc
root@22de24248ccc:/usr/share/bcc/tools# /usr/share/bcc/tools/trace 'SyS_write (arg1==1) "%s", arg2' -U -p `pidof java`
5573    27047   java            SyS_write        2018-11-05 12:33:05,094 [info] application-akka.actor.default-dispatche
        __GI___libc_write+0x2d [libpthread-2.24.so]
        [unknown] [libjava.so]
        Java_java_io_FileOutputStream_writeBytes+0x1a [libjava.so]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown]
        [unknown] [libjvm.so]
        [unknown] [libjvm.so]
        [unknown] [libjvm.so]
        [unknown] [libjvm.so]
        [unknown] [libjvm.so]
        [unknown] [libjvm.so]
        [unknown] [libjvm.so]
        start_thread+0xc4 [libpthread-2.24.so]
```

vijunag:
@tekumara - You don't have to run bcc inside of a container. Can you try profiling the process using the default namespace pid and check if it works ?

tekumara:
@vijunag do you mean running bcc from the host? I've tried running bcc (compiled with this PR) on the host and it produces the same output as above.

tekumara:
@vijunag I've installed the jdk8 debug symbols into the container and now I'm not getting `[unknown] [libjvm.so]` or `[unknown] [libjava.so]` lines.

I've also created a java perf map and am now getting `Interpreter+0x24f0 [perf-8.map]` instead of `[unknown]`.

yonghong-song:
[buildbot, test this please]

palmtenor:
Yeah this LGTM. Thanks for the fix!

PR #2032, opened 2018-11-05
Fix uprobes on powerpc64

sandip4n:
This makes sure that the correct entry point addresses are used when attaching `uprobes` to functions on `powerpc64` (both big and little endian).

### DIFF
--- a/src/cc/bcc_syms.h
+++ b/src/cc/bcc_syms.h
@@ -34,6 +34,13 @@ typedef int (*SYM_CB)(const char *symname, uint64_t addr);
 #ifndef STT_GNU_IFUNC
 #define STT_GNU_IFUNC 10
 #endif
+
+#if defined(__powerpc64__) && __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+// Indicate if the Local Entry Point (LEP) should be used as a symbol's
+// start address
+#define STT_PPC64LE_SYM_LEP 31
+#endif
+
 static const uint32_t BCC_SYM_ALL_TYPES = 65535;
 struct bcc_symbol_option {
   int use_debug_file;
--- a/src/cc/bcc_elf.c
+++ b/src/cc/bcc_elf.c
@@ -184,10 +184,49 @@ int bcc_elf_foreach_usdt(const char *path, bcc_elf_probecb callback,
   return res;
 }
 
+static Elf_Scn * get_section(Elf *e, const char *section_name,
+                             GElf_Shdr *section_hdr, size_t *section_idx) {
+  Elf_Scn *section = NULL;
+  GElf_Shdr header;
+  char *name;
+
+  size_t stridx;
+  if (elf_getshdrstrndx(e, &stridx) != 0)
+    return NULL;
+
+  size_t index;
+  for (index = 1; (section = elf_nextscn(e, section)) != 0; index++) {
+    if (!gelf_getshdr(section, &header))
+      continue;
+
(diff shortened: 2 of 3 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
The current implementation looks okay to me. Another way is to implement a new variant of `bcc_elf_foreach_sym` and `bcc_elf_symcb` to accommodate the original callback function signature. I think it is not worth it and one unused argument in `find_sym` and `_sym_cb_wrapper` is okay to me.
@palmtenor any more comments on this?

yonghong-song:
[buildbot, test this please]

palmtenor:
Right, I mentioned maybe doing the adjustment inside each_sym, mainly because that way we would be able to keep the callback interface unchanged, without having to change all the call sites and bindings...

sandip4n:
@palmtenor That can simplify things but I see a problem with that when it comes to printing the stack traces. For example, if the local entry point is at offset 8 from the global entry point, the symbol offsets in the call stacks would all be off by 8. So, when we expect to see a stack trace like this:
```
__GI___inet_pton+0x8 [libc-2.27.so]
gaih_inet.constprop.7+0x1040 [libc-2.27.so]
getaddrinfo+0x164 [libc-2.27.so]
[unknown] [ping]
generic_start_main.isra.0+0x138 [libc-2.27.so]
__libc_start_main+0xc4 [libc-2.27.so]
```
we instead see something like this:
```
__GI___inet_pton+0x0 [libc-2.27.so]
gaih_inet.constprop.7+0x1038 [libc-2.27.so]
getaddrinfo+0x15c [libc-2.27.so]
[unknown] [ping]
generic_start_main.isra.0+0x130 [libc-2.27.so]
__libc_start_main+0xbc [libc-2.27.so]
```
This is why we need to keep the `bcc_symbol`'s `offset` field intact (i.e. keep the global entry point address) when using `bcc_elf_foreach_sym()` and wait till later on to add the local entry point offset. While it will probably be okay to modify `offset` in the iteration function in case of the `_find_sym()` callback, if the same is done when using the `_add_symbol()` callback it will affect `bcc_symcache_resolve()` and hence the stack traces.

yonghong-song:
All these are guarded by powerpc macro define, so it should be fine. Please go ahead update the patch. Thanks!

sandip4n:
@palmtenor @yonghong-song Thanks for the reviews. I have pushed the updates that do not modify the callbacks.

src/cc/bcc_syms.h
```diff
@@ -34,6 +34,13 @@ typedef int (*SYM_CB)(const char *symname, uint64_t addr);
 #ifndef STT_GNU_IFUNC
 #define STT_GNU_IFUNC 10
 #endif
+
+#if defined(__powerpc64__) && __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
```
palmtenor: Also make sure `STT_PPC64LE_SYM_LEP` is not defined?

palmtenor:
[buildbot, ok to test]

src/cc/bcc_syms.h
sandip4n: Unlike `STT_GNU_IFUNC`, `STT_PPC64LE_SYM_LEP` is not really a valid Elf symbol type. I just made up a name for it. It is not defined in `libelf`.

palmtenor: Haha I see, makes sense.

PR #2034, opened 2018-11-08
add kernel stack trace option to capable

arighi:
Signed-off-by: Andrea Righi <andrea@betterlinux.com>

### DIFF
--- a/tools/capable.py
+++ b/tools/capable.py
@@ -4,9 +4,7 @@
 # capable   Trace security capabilitiy checks (cap_capable()).
 #           For Linux, uses BCC, eBPF. Embedded C.
 #
-# USAGE: capable [-h] [-v] [-p PID]
-#
-# ToDo: add -s for kernel stacks.
+# USAGE: capable [-h] [-v] [-p PID] [-K]
 #
 # Copyright 2016 Netflix, Inc.
 # Licensed under the Apache License, Version 2.0 (the "License")
@@ -14,6 +12,7 @@
 # 13-Sep-2016   Brendan Gregg   Created this.
 
 from __future__ import print_function
+from functools import partial
 from bcc import BPF
 import argparse
 from time import strftime
@@ -24,6 +23,7 @@ examples = """examples:
     ./capable             # trace capability checks
     ./capable -v          # verbose: include non-audit checks
     ./capable -p 181      # only trace PID 181
+    ./capable -K          # add kernel stacks to trace
 """
 parser = argparse.ArgumentParser(
     description="Trace security capability checks",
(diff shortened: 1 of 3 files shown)


tools/capable.py
```diff
@@ -150,8 +176,11 @@ def print_event(cpu, data, size):
     print("%-9s %-6d %-6d %-16s %-4d %-20s %d" % (strftime("%H:%M:%S"),
         event.uid, event.pid, event.comm.decode('utf-8', 'replace'),
         event.cap, name, event.audit))
+    if args.kernel_stack:
+        print_stack(bpf, event.kernel_stack_id, -1)
```
yonghong-song: You always pass -1 with tgid for print_stack. Any issues here?

tools/capable.py
```diff
@@ -86,25 +88,37 @@
...
+BPF_STACK_TRACE(stacks, 1024);
+#endif
+
 int kprobe__cap_capable(struct pt_regs *ctx, const struct cred *cred,
     struct user_namespace *targ_ns, int cap, int audit)
 {
-    u32 pid = bpf_get_current_pid_tgid();
+    u64 __pid_tgid = bpf_get_current_pid_tgid();
+    u32 __tgid = __pid_tgid >> 32;
+    u32 __pid = __pid_tgid;
```
yonghong-song: I suspect the original intention is for tgid, so __pid here may not really be needed. Could you double check?

tools/capable.py
```diff
@@ -86,25 +88,37 @@
...
     FILTER1
     FILTER2
 
     u32 uid = bpf_get_current_uid_gid();
-    struct data_t data = {.pid = pid, .uid = uid, .cap = cap, .audit = audit};
+    struct data_t data = {.tgid = __tgid, .pid = __pid, .uid = uid, .cap = cap, .audit = audit};
+#ifdef KERNEL_STACKS
+    data.kernel_stack_id = stacks.get_stackid(ctx, BPF_F_REUSE_STACKID);
```
yonghong-song: Please use `get_stackid(ctx, 0)` to avoid kernel overwrite of stack trace in the kernel for the same stack_id.

tools/capable.py
arighi: I was going to add also the user-space stack, so print_stack() is already prepared to resolve both kernel stack and user stack. I'll add the option to print the user stack soon in the next change.

tools/capable.py
arighi: True. What about printing both, PID and TID, like we do with other tools? I can add that.

tools/capable.py
arighi: OK. Thanks!

tools/capable.py
yonghong-song: Yes, we can print both.

tools/capable.py
yonghong-song: Yes, we should print user stack as well in parity with kernel stack print. Then in print_stack
```
print_stack(bpf, stack_id, tgid):
```
You will need to change to
```
print_stack(bpf, kern_stack_id, user_stack_id, tgid):
```
In this case, tgid should be real tgid.

arighi:
Closing this pull request and opening a new one with the changes suggested by @yonghong-song

PR #2036, opened 2018-11-08
Print name of failing program

oriolarcas:
When kprobe/kreprobe/tracepoint attachment fails, print the name of the failing function and its target.


palmtenor:
[buildbot, ok to test]

PR #2037, opened 2018-11-09
Fix the lua standalone test

sandip4n:
This fixes the subtest that checks if `bcc-lua` depends on `libluajit`.


yonghong-song:
[buildbot, test this please]

PR #2038, opened 2018-11-10
INSTALL.md - on ubuntu bionic use llvm 6.0 

tekumara:
because it doesn't have llvm 3.7 packages


yonghong-song:
[buildbot, test this please]

Issue #2035, opened 2018-11-08
http_filter not working in kernel 4.18


cippaciong:
Hello, I'm trying to run [http_filter](https://github.com/iovisor/bcc/tree/master/examples/networking/http_filter) on my laptop where I have Archlinux installed with kernel: `4.18.16-arch1-1-ARCH` but I get this error which I guess is coming from the verifier (?)
```
binding socket to 'wlp3s0'
bpf: Failed to load program: Permission denied
0: (bf) r6 = r1
1: (28) r0 = *(u16 *)skb[12]
2: (55) if r0 != 0x800 goto pc+96
 R0=inv2048 R6=ctx(id=0,off=0,imm=0) R10=fp0,call_-1
3: (30) r0 = *(u8 *)skb[23]
4: (55) if r0 != 0x6 goto pc+94
 R0=inv6 R6=ctx(id=0,off=0,imm=0) R10=fp0,call_-1
5: (30) r0 = *(u8 *)skb[14]
6: (bf) r8 = r0
7: (67) r8 <<= 2
8: (57) r8 &= 60
9: (b7) r1 = 20
10: (2d) if r1 > r8 goto pc+88
 R0=inv(id=0) R1=inv20 R6=ctx(id=0,off=0,imm=0) R8=inv(id=0,umin_value=20,umax_value=60,var_off=(0x0; 0x3c)) R10=fp0,call_-1
11: (bf) r7 = r8
12: (07) r7 += 26
13: (50) r0 = *(u8 *)skb[r7 + 0]
14: (bf) r7 = r0
15: (28) r0 = *(u16 *)skb[16]
16: (1f) r0 -= r8
17: (77) r7 >>= 2
18: (57) r7 &= 60
19: (1f) r0 -= r7
20: (67) r0 <<= 32
21: (77) r0 >>= 32
22: (b7) r1 = 7
23: (2d) if r1 > r0 goto pc+75
 R0=inv(id=0,umin_value=7,umax_value=4294967295,var_off=(0x0; 0xffffffff)) R1=inv7 R6=ctx(id=0,off=0,imm=0) R7=inv(id=0,umax_value=60,var_off=(0x0; 0x3c)) R8=inv(id=0,umin_value=20,umax_value=60,var_off=(0x0; 0x3c)) R10=fp0,call_-1
24: (0f) r7 += r8
25: (bf) r9 = r7
26: (07) r9 += 12
27: (07) r7 += 20
28: (bf) r8 = r9
[...]

PR #2039, opened 2018-11-10
fix verifier errors in http_filter example

yonghong-song:
Fix issue #2035.

For code like
```
    int i;
    int j = 0;
    const int last_index = payload_offset + 7;
    for (i = payload_offset ; i < last_index ; i++) {
               p[j] = load_byte(skb , i);
```
Here, the payload_offset is unknown. llvm 7.0 and trunk
compiler seems generating code like
```
   p[0] = load_byte(skb, payload_offset)
   if (payload_offset + 1 < last_index) {
      p[1] = ...
      ...
      p[6] = ...
   } else {
      /* do nothing */
   }
   /* accessing p[0], p[1], ..., p[6] */
```
The compiler did the above transformation because the potential
overflow for `last_index` and/or `payload_offset + 1` in which case
compiler preserved both branches.

This caused a problem for verifier as in the else branch
p[1] is not assigned and the verifier will reject the program.

Changing the loop to simply iterate from 0 to 6 fixed the problem.

### DIFF
--- a/examples/networking/http_filter/http-parse-complete.c
+++ b/examples/networking/http_filter/http-parse-complete.c
@@ -99,11 +99,8 @@ int http_filter(struct __sk_buff *skb) {
 	//direct access to skb not allowed
 	unsigned long p[7];
 	int i = 0;
-	int j = 0;
-	const int last_index = payload_offset + 7;
-	for (i = payload_offset ; i < last_index ; i++) {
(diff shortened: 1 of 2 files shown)


yonghong-song:
cc @cippaciong could you check whether this patch fixed your issue or not?

cippaciong:
Yes, it's working with this patch. Thanks!

PR #2043, opened 2018-11-14
Add shmsnoop/sofdsnoop tools

olsajiri:
The sofdsnoop traces FDs passed through unix sockets:

  # ./sofdsnoop
  ACTION TID    COMM             SOCKET                    FD    NAME
  SEND   2576   Web Content      24:socket:[39763]         51    /dev/shm/org.mozilla.ipc.2576.23874
  RECV   2576   Web Content      49:socket:[809997]        51
  SEND   2576   Web Content      24:socket:[39763]         58    N/A
  ...

The shmsnoop tool to trace System V shared memory syscalls:

  # ./shmsnoop.py
  PID    COMM                SYS              RET ARGs
  19813  server           SHMGET            10000 key: 0x78020001, size: 20, shmflg: 0x3b6 (IPC_CREAT|0666)
  19813  server            SHMAT     7f1cf8b1f000 shmid: 0x10000, shmaddr: 0x0, shmflg: 0x0
  19816  client           SHMGET            10000 key: 0x78020001, size: 20, shmflg: 0x1b6 (0666)
  ...

### DIFF
--- a/tools/shmsnoop.py
+++ b/tools/shmsnoop.py
@@ -1,0 +1,335 @@
+#!/usr/bin/python
+# @lint-avoid-python-3-compatibility-imports
+#
+# shmsnoop Trace shm*() syscalls.
+#           For Linux, uses BCC, eBPF. Embedded C.
+#
+# USAGE: shmsnoop [-h] [-T] [-x] [-p PID] [-d DURATION] [-t TID] [-n NAME]
+#
+# Copyright (c) 2018 Jiri Olsa.
+# Licensed under the Apache License, Version 2.0 (the "License")
+#
+# 08-Oct-2018   Jiri Olsa   Created this.
+
+from __future__ import print_function
+from bcc import ArgString, BPF
+import argparse
+import ctypes as ct
+from datetime import datetime, timedelta
+
+# arguments
+examples = """examples:
+    ./shmsnoop           # trace all shm*() syscalls
+    ./shmsnoop -T        # include timestamps
+    ./shmsnoop -p 181    # only trace PID 181
+    ./shmsnoop -t 123    # only trace TID 123
+    ./shmsnoop -d 10     # trace for 10 seconds only
+    ./shmsnoop -n main   # only print process names containing "main"
+"""
+parser = argparse.ArgumentParser(
+    description="Trace shm*() syscalls",
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    epilog=examples)
+parser.add_argument("-T", "--timestamp", action="store_true",
+    help="include timestamp on output")
+parser.add_argument("-p", "--pid",
+    help="trace this PID only")
+parser.add_argument("-t", "--tid",
+    help="trace this TID only")
+parser.add_argument("-d", "--duration",
+    help="total duration of trace in seconds")
+parser.add_argument("-n", "--name",
+    type=ArgString,
+    help="only print process names containing this name")
+parser.add_argument("--ebpf", action="store_true",
+    help=argparse.SUPPRESS)
+args = parser.parse_args()
+debug = 0
+if args.duration:
+    args.duration = timedelta(seconds=int(args.duration))
+
+# define BPF program
+bpf_text = """
+#include <uapi/linux/ptrace.h>
+#include <uapi/linux/limits.h>
+#include <linux/sched.h>
+
+struct val_t {
+    u64            id;
+    u64            ts;
+    int            sys;
+    unsigned long  key;
+    unsigned long  size;
+    unsigned long  shmflg;
+    unsigned long  shmid;
+    unsigned long  cmd;
+    unsigned long  buf;
+    unsigned long  shmaddr;
+    unsigned long  ret;
+    char           comm[TASK_COMM_LEN];
+};
+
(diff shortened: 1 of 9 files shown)


tools/shmsnoop.py
```diff
@@ -0,0 +1,335 @@
...
+
+int trace_shmget(struct pt_regs *ctx, struct pt_regs *_p)
+{
+    struct val_t val = {
+        .sys = SYS_SHMGET,
+    };
+    struct pt_regs p;
+
+    if (bpf_probe_read(&p, sizeof(p), (void *) _p))
+        return 0;
```
yonghong-song: You do not need bpf_probe_read here, you can directly access registers through _p. e.g.,
`val.key = PT_REGS_PARM1(_p);`. The same for all other accessing kprobe pt_regs in this patch set.

tools/shmsnoop.py
```diff
@@ -0,0 +1,335 @@
...
+SYS_SHMCTL = 3
+
+initial_ts = 0
+
+class Data(ct.Structure):
+    _fields_ = [
+        ("id",      ct.c_ulonglong),
+        ("ts",      ct.c_ulonglong),
+        ("sys",     ct.c_int),
+        ("key",     ct.c_ulonglong),
```
yonghong-song: Could you make the type consistent? The C structure is type long and here is ulonglong.

yonghong-song:
[buildbot, test this please]

yonghong-song:
cc @brendangregg

olsajiri:
On Wed, Nov 14, 2018 at 08:05:08AM -0800, yonghong-song wrote:
> yonghong-song commented on this pull request.
> 
> 
> 
> > +    events.perf_submit(ctx, val, sizeof(*val));
> +
> +out:
> +    infotmp.delete(&id);
> +    return 0;
> +}
> +
> +int trace_shmget(struct pt_regs *ctx, struct pt_regs *_p)
> +{
> +    struct val_t val = {
> +        .sys = SYS_SHMGET,
> +    };
> +    struct pt_regs p;
> +
> +    if (bpf_probe_read(&p, sizeof(p), (void *) _p))
> +        return 0;
> 
> You do not need bpf_probe_read here, you can directly access registers through _p. e.g.,
> `val.key = PT_REGS_PARM1(_p);`. The same for all other accessing kprobe pt_regs in this patch set.

hum, if I do that I get the access violation:

...
13: (7b) *(u64 *)(r10 -104) = r2
14: (79) r2 = *(u64 *)(r1 +112)
R1 invalid mem access 'inv'

we need to do that, because __x64_sys_shmget is called with regs argument:

#define __SYSCALL_DEFINEx(x, name, ...)                                 \
        asmlinkage long __x64_sys##name(const struct pt_regs *regs);    \
	...

and AFAICS we get that pointer and need to get its data via bpf_probe_read

> 
> > +
> +TASK_COMM_LEN = 16    # linux/sched.h
> +
> +SYS_SHMGET = 0
> +SYS_SHMAT  = 1
> +SYS_SHMDT  = 2
> +SYS_SHMCTL = 3
> +
> +initial_ts = 0
> +
> +class Data(ct.Structure):
> +    _fields_ = [
> +        ("id",      ct.c_ulonglong),
> +        ("ts",      ct.c_ulonglong),
> +        ("sys",     ct.c_int),
> +        ("key",     ct.c_ulonglong),
> 
> Could you make the type consistent? The C structure is type long and here is ulonglong.

right, will change:w

thanks,
jirka

yonghong-song:
Oh, I forget  syscall kprobe interface is changed in recent kernel. The following patch tried to fix the issue:
```
commit 2da34267fcae4485f4e05a17521214749f6f0edd
Author: yonghong-song <ys114321@gmail.com>
Date:   Wed Jun 13 06:12:22 2018 -0700

    generate indirect parameter assignment if arch uses syscall wrapper (#1816)
```
Alternatively, you could use something like
```
int syscall_shmget(struct pt_regs *ctx, u64 key, u64 size, u64 shmflg)
```
Note that if you have `syscall__`, compiler is able to generate correct code.
This way the compiler generated code will hide all details between different
kernel versions. You can use `execsnoop.py` as an example.

olsajiri:
On Wed, Nov 14, 2018 at 01:02:52PM -0800, yonghong-song wrote:
> Oh, I forget  syscall kprobe interface is changed in recent kernel. The following patch tried to fix the issue:
> ```
> commit 2da34267fcae4485f4e05a17521214749f6f0edd
> Author: yonghong-song <ys114321@gmail.com>
> Date:   Wed Jun 13 06:12:22 2018 -0700
> 
>     generate indirect parameter assignment if arch uses syscall wrapper (#1816)
> ```
> Alternatively, you could use something like
> ```
> int syscall_shmget(struct pt_regs *ctx, u64 key, u64 size, u64 shmflg)
> ```
> Note that if you have `syscall__`, compiler is able to generate correct code.
> This way the compiler generated code will hide all details between different
> kernel versions. You can use `execsnoop.py` as an example.

aaah the syscall__ prefix.. that explains why that prototype
did not work for me before ;-) will change

thanks,
jirka

yonghong-song:
[buildbot, test this please]

yonghong-song:
the revision of this pull request has been merged.

PR #2051, opened 2018-11-20
offwaketime: properly format "[Missed User Stack]" with folded output

arighi:
In folded output format when both kernel and user stacks are missing,
the message "[Missing User Stack]" is formatted incorrectly, for example:

chrome;entry_SYSCALL_64_after_hwframe;do_syscall_64;SyS_poll;do_sys_poll;poll_schedule_timeout;schedule_hrtimeout_range;schedule_hrtimeout_range_clock;schedule;--;[Missed Kernel Stack];[;M;i;s;s;e;d; ;U;s;e;r; ;S;t;a;c;k;];fio 4955989

This is because we're incorrectly appending the missing stack message in
the list of stack entries. Fix by appending the message as a single item.

Signed-off-by: Andrea Righi <righi.andrea@gmail.com>


yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks for the fix!

PR #2054, opened 2018-11-21
set proper layout for the module before finalizing the object

yonghong-song:
Otherwise, if llvm is built with assertion on,
we will hit the following error for a simple trace.py script like
```
  -bash-4.4$ sudo ./trace.py 'SyS_nanosleep'
  python: /home/yhs/work/llvm/lib/ExecutionEngine/MCJIT/MCJIT.cpp:204:
          virtual void llvm::MCJIT::generateCodeForModule(llvm::Module*):
          Assertion `M->getDataLayout() == getDataLayout() && "DataLayout Mismatch"'
          failed.
  Aborted (core dumped)
  -bash-4.4$
```

### DIFF
--- a/src/cc/bpf_module.cc
+++ b/src/cc/bpf_module.cc
@@ -618,6 +618,11 @@ int BPFModule::finalize() {
       *sections_p;
 
   mod->setTargetTriple("bpf-pc-linux");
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

PR #2061, opened 2018-11-27
Update the tutorial to match what is currently in the examples

wcohen:
There have been some updates the code in the examples.  The tutorial
text should match up with the current example code. In particular we
want to make ensure that the probe for the syscall functions match up
with what is currently needed for newer kernels that do not have
syscall functions uniformly start with "sys_".

### DIFF
--- a/docs/tutorial_bcc_python_developer.md
+++ b/docs/tutorial_bcc_python_developer.md
@@ -67,14 +67,14 @@ from bcc import BPF
 # define BPF program
 prog = """
 int hello(void *ctx) {
-	bpf_trace_printk("Hello, World!\\n");
-	return 0;
+    bpf_trace_printk("Hello, World!\\n");
+    return 0;
 }
 """
 
 # load BPF program
 b = BPF(text=prog)
-b.attach_kprobe(event="sys_clone", fn_name="hello")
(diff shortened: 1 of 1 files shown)


docs/tutorial_bcc_python_developer.md
```diff
@@ -67,14 +67,14 @@ from bcc import BPF
 # define BPF program
 prog = """
 int hello(void *ctx) {
-	bpf_trace_printk("Hello, World!\\n");
-	return 0;
+    bpf_trace_printk("Hello, World!\\n");
+    return 0;
```
pchaigno: Any reason you changed all indentations?

wcohen: The patch copied over what the example codes currently have and that indentation is what the examples currently have.  However, commit 44ad8a1ececed09e41dea5f8622759b9f772aa61 for docs/tutorial_bcc_python_developer.md the comment says:

    Always use spaces in C in the tutorial

There are still tabs in tutorial_bcc_python_developer.md.  Should the remaining tabs in the patch be converted to spaces to be consistent?

yonghong-song:
[buildbot, test this please]

yonghong-song:
The change looks good. Thanks for the change!

PR #2062, opened 2018-11-29
RFC: add kernel tools/lib/bpf to bcc

yonghong-song:
The kernel tools/lib/bpf implements some core bpf related functionality and already implements the latest kernel features. The kernel/lib/bpf code has been mirrored at `https://github.com/libbpf/libbpf`.

This patch incorporates the libbpf github source codes into bcc. This way, for the future btf support in bcc, we do not need to re-implement all the parsing, dumping functions etc since kernel `tools/lib/bpf` has all of them.

The current approach is to copy the code to bcc. My previous idea is to use `git submodule` but I did not find a way to track a particular commit. The `git submodule` allows you to track a particular branch, but I am not sure whether we should create a lot of even just release branches is a good idea for libbpf repo.

Another incompatible change is for two API functions `bpf_create_map` and `bpf_prog_load`. Unfortunately, the kernel `tools/lib/bpf/{bpf.c, libbpf.c}` defined these two functions with different signatures. So in this patch, I renamed them to `bpf_create_map2` and `bpf_prog_load2`. Please share if you have better ideas.

There are 3 separate commits in order to simplify review. Commit #1 add kernel libbpf code, removed `bcc/src/cc/compat/linux/{bpf.h, bpf_common.h}` and use the kernel uapi headers.
Commit #2 incorporate kernel libbpf source codes into build process.
Commit #3 reimplement `bpf_map_create2` to use the newly added functions as a demonstration.

Please do take a look and comment if this has any impact on you and you have any suggestions.


yonghong-song:
the master branch works. I implemented this way just want to get opinions. Using gitsubmodule branch approach, for every bcc release, we will need to create a branch in libbpf repo. I guess this should be okay.

If anybody knows better to use `git submodule` to track commits, please comment and make suggestions.

dcbeckett:
git submodule can link to a particular commit

For example see https://github.com/Netronome/bpf-samples/ - it is linked to libbpf commit 319ff2f. This was the latest commit when initialised the submodule but is now out of date.

I believe you can go enter into the submodule repo and checkout the particular commit you wish, then commit that change in the parent repo. You would still need to patch the submodule with a script if this is required.

yonghong-song:
Thanks @dcbeckett for the tip! I will use git submodule similar to your netronome approach in the next revision.

torgil:
Since both repos are on github I suggest cloning with relative path. This way it'll follow whatever style of clone (https, ssh) the user has done and can even work offline (if placed in the same folders).

```
$ git submodule add ../../libbpf/libbpf.git
...
$ cat .gitmodules 
[submodule "libbpf"]
	path = libbpf
	url = ../../libbpf/libbpf.git
```

yonghong-song:
close this rfc patch. Will submit a formal patch soon.

PR #2064, opened 2018-12-03
usdt: use ProcMountNSGuard for pid and path attach

tjfontaine:
This adds a couple of lexical scopes for usage with `ProcMountNSGuard` so the hint'ing works for the C interface when using `bcc_usdt_new_frompid()`. Also, add a guard for the `Context::Context(pid, path)` constructor for resolving the probes.

From the commit message:

Since the probe may be in a descendent namespace, ensure that when `stat()`ing for HINTs and resolving linked elf locations, be sure to always perform them relative to any observed mount namespace.

However, usage of the mount namespace must be surgical, since we're observing the pid with its outter identity, make sure we're not in the mount namespace when attempting to `readlink` the `/proc/<pid>/exe`

### DIFF
--- a/src/cc/usdt/usdt.cc
+++ b/src/cc/usdt/usdt.cc
@@ -373,7 +373,12 @@ Context::Context(int pid, const std::string &bin_path)
       mount_ns_instance_(new ProcMountNS(pid)), loaded_(false) {
   std::string full_path = resolve_bin_path(bin_path);
   if (!full_path.empty()) {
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

dalehamel:
What is still outstanding to merge this? Ran into the same issues as @tjfontaine when looking at usdt support in bpftrace, and discovered that bcc itself doesn't properly handle mount namespaces for its own path check, which this seems to fix.

yonghong-song:
Let us merge this then.

PR #2066, opened 2018-12-04
Add test script for XDP_TX action

gospo:
Mostly a cut and paste of the xdp_drop_count.py with needed changes for
doing a source and destination MAC swap.

Signed-off-by: Andy Gospodarek <gospo@broadcom.com>


gospo:
Rather than continuing to hack on xdp_drop_count.py, I thought I'd add something to the repo so others can use it too.

yonghong-song:
Looks good to me. Thanks!

PR #2070, opened 2018-12-06
tools: prepare block tools for the removing of legacy I/O path

jeromemarchand:
Recent -next kernels don't have blk_start_request() function
anymore. It has been removed in a recent cleanup. bio* tools should be
able to handle the lack of this probe.

### DIFF
--- a/tools/biolatency.py
+++ b/tools/biolatency.py
@@ -116,7 +116,8 @@ b = BPF(text=bpf_text)
 if args.queued:
     b.attach_kprobe(event="blk_account_io_start", fn_name="trace_req_start")
 else:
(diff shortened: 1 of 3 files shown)


yonghong-song:
@jeromemarchand could you share the cleanup commit here?  So `blk_mq_start_request` will cover all start_request, right?

jeromemarchand:
The commit that actually removed blk_start_request is commit bd93676fa69faf ("block: remove dead elevator code"). But it's only a small part of a big series that removed all of the legacy request code and converted its remaining users to blk-mq.
And yes, from now on, blk_mq_start_request will cover all request that were started by blk_start_request,

yonghong-song:
@jeromemarchand Thanks for clarifying! The change makes sense.

PR #2072, opened 2018-12-10
Fix syscall map generation via ausyscall

sandip4n:
When `ausyscall` is used as a fallback plan to generate the syscall map, the `subprocess.check_output()` call fails as the program name and arguments must be passed separately.

Fixes: 218f7482f ("Wcohen/efficiency (#2063)")


yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks for the fix!

PR #2074, opened 2018-12-11
Add option to build with rtti

torgil:
Enables linking bcc-static in binaries compiled with rtti

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,7 @@ include(GNUInstallDirs)
 include(CheckCXXCompilerFlag)
 include(cmake/FindCompilerFlag.cmake)
 
+option(ENABLE_RTTI "Enable compiling with real time type information" OFF)
 option(ENABLE_LLVM_SHARED "Enable linking LLVM as a shared library" OFF)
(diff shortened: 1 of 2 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
Could you specify your use case when `ENABLE_RTTI` will be enabled?

torgil:
> So you mean if the application compiled with -rtti, and wants to link with bcc-static, the bcc-static needs to compile with -rtti as well? Just want to confirm.

Yes. I get link errors otherwise. Last time i needed this was to compile bpftrace on arm64. I needed to recompile llvm with -rtti and did recompile bcc with -rtti as well, I didn't try to compile only LLVM with -rtti though.

Searching the net, it seems like mixing RTTI is a common source for link errors.

yonghong-song:
The change itself looks good. Do you know why bpftrace/arm64 needs llvm rtti? Because of certain clang/llvm functions?

torgil:
> The change itself looks good. Do you know why bpftrace/arm64 needs llvm rtti? Because of certain clang/llvm functions?

bpftrace defaulted rtti for me but trying to compile with --no-rtti resulted in errors due to "typeid" function requiring rtti in code generated by bison.

PR #2075, opened 2018-12-11
Remove dependency on LLVM native target

torgil:
Not needed for BCC to work. It somehow brought in the interpreter and
executionengine which is needed. Those features are added instead.

### DIFF
--- a/cmake/clang_libs.cmake
+++ b/cmake/clang_libs.cmake
@@ -1,8 +1,8 @@
 if(ENABLE_LLVM_SHARED)
 set(llvm_libs "LLVM")
 else()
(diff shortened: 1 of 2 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

palmtenor:
This could actually be a good thing to reduce both binary size and RSS consumed by text during run time. Feel free to explorer options:)

torgil:
Yes. That was the purpose. I did a new patch to make native optional but managed to force-push it and github don't like that. Opened a new pull-request instead.

PR #2076, opened 2018-12-12
[RFC] Use best name for symbols with aliases

sandip4n:
A symbol can have multiple aliases. A lot of symbols in popular libraries like glibc have this trait. So, when building a cache that helps in resolving symbol names, order symbols having the same address based on a few basic heuristics such that symbols with better aliases are picked first. This helps in avoiding an obscure symbol name when printing a stack trace entry.

This has been verified as shown below.

```
  $ readelf -s /usr/lib64/libc-2.27.so | grep "inet_pton$"
    3548: 0000000000118db0    40 FUNC    LOCAL  DEFAULT   12 __inet_pton
    4168: 0000000000118db0    40 FUNC    LOCAL  DEFAULT   12 __GI___inet_pton
    5455: 0000000000118db0    40 FUNC    LOCAL  DEFAULT   12 __GI_inet_pton
    6807: 0000000000118db0    40 FUNC    WEAK   DEFAULT   12 inet_pton

  $ readelf -s /usr/lib64/libc-2.27.so | grep "getaddrinfo$"
    3499: 00000000000e4720  3223 FUNC    LOCAL  DEFAULT   12 __GI_getaddrinfo
    8162: 00000000000e4720  3223 FUNC    GLOBAL DEFAULT   12 getaddrinfo

  $ ping -6 ::1
  PING ::1(::1) 56 data bytes
  64 bytes from ::1: icmp_seq=1 ttl=64 time=0.101 ms
  64 bytes from ::1: icmp_seq=2 ttl=64 time=0.102 ms
  ^C
  --- ::1 ping statistics ---
  2 packets transmitted, 2 received, 0% packet loss, time 1021ms
  rtt min/avg/max/mdev = 0.101/0.101/0.102/0.010 ms
```
Before:
```
  $ sudo trace "c:inet_pton" -U
  28410   28410   ping            inet_pton
          __inet_pton+0x0 [libc-2.27.so]
          __GI_getaddrinfo+0x124 [libc-2.27.so]
          [unknown] [ping]
```
After:
```
  $ sudo trace "c:inet_pton" -U
  28449   28449   ping            inet_pton
          inet_pton+0x0 [libc-2.27.so]
          getaddrinfo+0x124 [libc-2.27.so]
          [unknown] [ping]
```

### DIFF
--- a/src/cc/bcc_syms.cc
+++ b/src/cc/bcc_syms.cc
@@ -46,6 +46,41 @@ bool ProcStat::is_stale() {
 ProcStat::ProcStat(int pid)
     : procfs_(tfm::format("/proc/%d/exe", pid)), inode_(getinode_()) {}
 
+static int _sym_count_prefix_underscores(const std::string symname) {
+  std::string::const_iterator it;
+  for (it = symname.begin(); it != symname.end() && (*it) == '_'; it++);
+  return it - symname.begin();
+}
+
+static bool _sym_has_better_name(const std::string symnamea,
+                                 const std::string symnameb) {
+  int a, b;
+
+  // Prefer the symbol with less underscores in prefix
+  a = _sym_count_prefix_underscores(symnamea);
+  b = _sym_count_prefix_underscores(symnameb);
+  if (a < b)
+    return true;
+  else if (a > b)
+    return false;
+
+  // Prefer the symbol with less underscores overall
+  a += std::count(symnamea.begin() + a, symnamea.end(), '_');
+  b += std::count(symnameb.begin() + b, symnameb.end(), '_');
+  if (a < b)
+    return true;
+  else if (a > b)
+    return false;
+
+  // Prefer the symbol with a longer name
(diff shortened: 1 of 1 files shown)


palmtenor:
This would definitely be nice as an user experience. However I'm a bit worried about doing it in `SymCache` (`KSym` / `ProcSym`), for a few reasons:
  - They are often the performance bottleneck on a lot of tracing use cases. Adding expensive string operation may make things worse
  - The current symbols isn't too bad to pay that cost. When users see `__GI_getaddrinfo` in the Stack-trace it's still quite clear it's doing `getaddrinfo`.
  - `SymCache` is supposed to just just faithfully load and accept queries for the symbol table, but not making heuristic choices for users?

I feel this would be better be as a printing feature to just beautify things if needed?

sandip4n:
@palmtenor Fair enough. Let me re-evaluate this. Closing this PR for now.

PR #2078, opened 2018-12-12
Support compiling BPF module without clang jit

torgil:
This patch separates BPFModule clang- and runtime functionality. The
    purpose is to be make it possible to precompile- or on-the-fly remote
    compile BPF code and use the runtime functionality on systems without
    resources for the clang toolchain (embedded) or cases where the startup
    compile overhead is not wanted.

    * Re-purpose ENABLE_CLANG_JIT from "building libbpf only" to "enable
      clang jit" (eg do as it says).
    * Add LIBBPF_ONLY option
      - Replaces current ENABLE_CLANG_JIT usage which according to git log was
        to build "JUST the libbpf library"
      - Name chosen to follow the same concept as "PYTHON_ONLY"
    * BPFModule: Separate out all clang functionality to bcc_module_clang.cc
    * BPFModule: Implement bcc_module_clang_dummy.cc
      - module can now be built without need to link to LLVM
      - constructor/destructor moved to h-file as they are dependent on LLVM
        implementation. This file is included fron "bcc_module_clang.cc" and
        "bcc_module_clang_dummy.cc" after LLVM classes is defined.
      - Clang constructor/destructor code moved to init_clang/cleanup_clang
        functions to allow dummy implementation.
      - Module constructor/destructor code moved to init_module/cleanup_module
        functions to reduce dependencies on constructor h-file making both
        clang versions easier to compile.
    * Dummy version of json_map_visitor implemented
      - Makes shared_table.cc compile without LLVM


torgil:
I have not yet any code that "pre-compiles" or "on-the-fly remote compile". I wanted to push this early as to get comments about the refactorization and ideas on the topic.

Also, these kind of patches is usually prone to merge conflicts and saves time to have upstream as early as possible.

yonghong-song:
[buildbot, test this please]

PR #2079, opened 2018-12-12
manpage: remove non-existent -p option from the biotop manpage

jeromemarchand:
The biotop manpage references a -p option that the tools doesn't have,
and AFAICT, never had. It's only referenced in the manpage option, not
in the synopsis, in "biotop -h" output not biotop_example.txt.


palmtenor:
[buildbot, test this please]

PR #2080, opened 2018-12-13
Make dependency on LLVM native target optional

torgil:
Adds an option ENABLE_LLVM_NATIVECODEGEN with default value ON.
If set to off the "nativecodegen" llvm will not be enabled, thus
reducing dependencies on needed libraries (reduced text size when
building with statically linked libraries).

Code that uses native target will not be compiled reducing text size.
Currently this affects the rw_engine which needs the native target.

BPF api "rw_engine_enabled" will have default value "true" if
ENABLE_LLVM_NATIVECODEGEN="ON" and "false" if
ENABLE_LLVM_NATIVECODEGEN="OFF"

Not needed for BCC to work. It somehow brought in the interpreter and
executionengine which is needed. Those features are added instead.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,7 @@ include(GNUInstallDirs)
 include(CheckCXXCompilerFlag)
 include(cmake/FindCompilerFlag.cmake)
 
+option(ENABLE_LLVM_NATIVECODEGEN "Enable use of llvm nativecodegen module (needed by rw-engine)" ON)
 option(ENABLE_RTTI "Enable compiling with real time type information" OFF)
 option(ENABLE_LLVM_SHARED "Enable linking LLVM as a shared library" OFF)
 option(ENABLE_CLANG_JIT "Enable Loading BPF through Clang Frontend" ON)
--- a/cmake/clang_libs.cmake
+++ b/cmake/clang_libs.cmake
@@ -1,8 +1,11 @@
 if(ENABLE_LLVM_SHARED)
 set(llvm_libs "LLVM")
 else()
-set(llvm_raw_libs bitwriter bpfcodegen debuginfodwarf irreader linker
-  mcjit objcarcopts option passes nativecodegen lto)
+set(llvm_raw_libs bitwriter bpfcodegen executionengine debuginfodwarf irreader linker
+  interpreter mcjit objcarcopts option passes lto)
+if(ENABLE_LLVM_NATIVECODEGEN)
+set(llvm_raw_libs ${llvm_raw_libs} nativecodegen)
+endif()
 list(FIND LLVM_AVAILABLE_LIBS "LLVMCoverage" _llvm_coverage)
 if (${_llvm_coverage} GREATER -1)
   list(APPEND llvm_raw_libs coverage)
--- a/cmake/clang_libs.cmake
+++ b/cmake/clang_libs.cmake
@@ -1,8 +1,8 @@
 if(ENABLE_LLVM_SHARED)
 set(llvm_libs "LLVM")
 else()
-set(llvm_raw_libs bitwriter bpfcodegen executionengine debuginfodwarf irreader linker
-  interpreter mcjit objcarcopts option passes lto)
+set(llvm_raw_libs bitwriter bpfcodegen debuginfodwarf irreader linker
+  mcjit objcarcopts option passes lto)
 if(ENABLE_LLVM_NATIVECODEGEN)
 set(llvm_raw_libs ${llvm_raw_libs} nativecodegen)
 endif()
--- a/src/cc/CMakeLists.txt
+++ b/src/cc/CMakeLists.txt
@@ -34,6 +34,12 @@ if (${LLVM_PACKAGE_VERSION} VERSION_EQUAL 6 OR ${LLVM_PACKAGE_VERSION} VERSION_G
   set(bcc_common_sources ${bcc_common_sources} bcc_debug.cc)
 endif()
 
+if(ENABLE_LLVM_NATIVECODEGEN)
+set(bcc_common_sources ${bcc_common_sources} bpf_module_rw_engine.cc)
+else()
+set(bcc_common_sources ${bcc_common_sources} bpf_module_rw_engine_disabled.cc)
+endif()
+
 set(bcc_table_sources table_storage.cc shared_table.cc bpffs_table.cc json_map_decl_visitor.cc)
 set(bcc_util_sources ns_guard.cc common.cc)
 set(bcc_sym_sources bcc_syms.cc bcc_elf.c bcc_perf_map.c bcc_proc.c)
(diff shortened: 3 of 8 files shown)


palmtenor:
[buildbot, test this please]

palmtenor:
This is great, thank you!

My only concern is that we are getting a lot more files (along with #2078) for `BPFModule`, it's a bit hard to track for for future development. How do you feel if we gate the different implementation with `ifdef` still in a single `bpf_module.cc`? So it would be easier to see and compare them side-by-side in the same file.

yonghong-song:
@torgil Thanks for the change. There are some compilation errors, could you fix them? It will be good to check the examples with C++ API to see how much saving (binary size and runtime RSS memory) we can get.

torgil:
> How do you feel if we gate the different implementation with `ifdef` still in a single `bpf_module.cc`? 

I think we should refactor logic into reusable components rather than adding ifdefs. ifdefs doesn't scale and is a pain to deal with when they stack up, altering code inside these ifdefs (probably more features needs to be switched on/off in the future). Since the future is hard to predict, I'd argue for the solution that keeps the most available options open. For instance, I will need to compile bcc library both with- and without clang support (for different tools). Without ifdef, the buildsystem can compile both at once (all variants), with ifdef I will need two full set of target builds. If the ifdef is system-wide (like different compilers, architectures), it's no problem since we cannot reuse anything anyway.

Maybe this is the time to brainstorm refactoring that removes the need to have "everything in one class". In this case, could refactor out the generation of sscanf/sprintf to a separate component with two implementations?

torgil:
> @torgil Thanks for the change. There are some compilation errors, could you fix them? It will be good to check the examples with C++ API to see how much saving (binary size and runtime RSS memory) we can get.

No problem. There was some garbage entered between my tests and push. Fixed. Size for HelloWorld on Ubuntu 18.04:
$ mkdir build && cd build
$ cmake -G Ninja .. 
$ ninja
examples/cpp/HelloWorld  is 70108920 bytes
$ rm -rf *
$ cmake -G Ninja .. -DENABLE_LLVM_NATIVECODEGEN=OFF
$ ninja
examples/cpp/HelloWorld is 62514808 bytes

yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks for the data, 8MB saving for disk space. I suspect we may not get much for RSS memory as rw_engine functions are not called anyway at runtime. Your current approach certainly is an acceptable one, although I agree, refactoring into a separate class with different implementation may be even better from code quality/clarity perspective. But since bcc module handling does not really need a lot of extensibility, a separate class could overkill. I will look at the implementation soon.

torgil:
> I will look at the implementation soon.

Thanks. Should I squash the commits to keep history clean?
Is that feasible in github without starting a new pull-request?

pchaigno:
Yes it's possible if you simply force push to the branch. @yonghong-song can also do it from the Web UI when he merges the PR.

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

yonghong-song:
The change looks good. I also saw libbcc.so binary is reduced from 67MB to 60MB if rw_engine is disabled through cmake option.

fc25/26/27 all failed with the following error:
```
[ 84%] Building C object src/cc/CMakeFiles/bcc-shared.dir/bcc_proc.c.o
[ 85%] Building CXX object src/cc/CMakeFiles/bcc-shared.dir/ns_guard.cc.o
make[2]: *** No rule to make target '/usr/lib64/libffi.so', needed by 'src/cc/libbcc.so.0.7.0'.  Stop.
make[2]: *** Waiting for unfinished jobs....
[ 85%] Building CXX object src/cc/CMakeFiles/bcc-shared.dir/common.cc.o
make[1]: *** [CMakeFiles/Makefile2:241: src/cc/CMakeFiles/bcc-shared.dir/all] Error 2
```

Do you know why? The code without this patch seems not require it. We can certainly install this library on the host if required.

yonghong-song:
On my local centos based system, I did not see libffi dependency.

```
-bash-4.4$ ldd libbcc.so.0.7.0 
        linux-vdso.so.1 =>  (0x00007ffcead09000)
        libelf.so.1 => /lib64/libelf.so.1 (0x00007fcec5fc1000)
        libz.so.1 => /lib64/libz.so.1 (0x00007fcec5dab000)
        librt.so.1 => /lib64/librt.so.1 (0x00007fcec5ba2000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007fcec599e000)
        libtinfo.so.5 => /lib64/libtinfo.so.5 (0x00007fcec5774000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007fcec5557000)
        libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007fcec5250000)
        libm.so.6 => /lib64/libm.so.6 (0x00007fcec4f4e000)
        libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007fcec4d37000)
        libc.so.6 => /lib64/libc.so.6 (0x00007fcec496a000)
        /lib64/ld-linux-x86-64.so.2 (0x0000562c60c2d000)
-bash-4.4$ 
```

torgil:
Yes.

I installed a Ubuntu18.04 here now and LLVM "interpreter" adds a dependency on ffi (removed ffi from ninja build file and got the below errors). I added "interpreter" since HelloWorld (I guess, might have been bpftrace) complained about this on ARM64. I also added "executionengine" since we use that alot.

On this Ubuntu 18.04 system it can run without both "interpreter" and "executionengine". I don't know why my ARM64 platform required either "nativetarget" or "interpreter" but we can fix that as a separate pull-request later on (It's x-mas here so I will not be able to troubleshoot that for a week).

(.text._ZL10ffiTypeForPN4llvm4TypeE+0x3a): undefined reference to `ffi_type_float'
(.text._ZL10ffiTypeForPN4llvm4TypeE+0x43): undefined reference to `ffi_type_void'
(.text._ZL10ffiTypeForPN4llvm4TypeE+0x53): undefined reference to `ffi_type_pointer'
(.text._ZL10ffiTypeForPN4llvm4TypeE+0x63): undefined reference to `ffi_type_double'
(.text._ZL10ffiTypeForPN4llvm4TypeE+0x78): undefined reference to `ffi_type_sint8'
(.text._ZL10ffiTypeForPN4llvm4TypeE+0x83): undefined reference to `ffi_type_sint16'
(.text._ZL10ffiTypeForPN4llvm4TypeE+0x93): undefined reference to `ffi_type_sint64'
(.text._ZL10ffiTypeForPN4llvm4TypeE+0xb3): undefined reference to `ffi_type_sint32'
/usr/lib/llvm-6.0/lib/libLLVMInterpreter.a

Pushed a patch to remove interpreter and executionengine again.

Why you cannot link ffi is another story. Which CentOS version do you run? Maybe I can bring up an install and check. Maybe latest stable?

It looks odd to search in "lib64", is that common for CentOS?

yonghong-song:
[buildbot, test this please]

yonghong-song:
I am using a facebook internel development server, which is centos 7 based, may not exactly the same.

yonghong-song:
Looks good to me. Thanks!

PR #2083, opened 2018-12-17
Tools: properly handle the keyboard interrupt

jeromemarchand:
Many tools rely on the user to type Ctrl-C to end, but don't actually
catch the keyboard interrupt and thus show an ugly backtrace when it
happens. Let's catch the interrupt.

### DIFF
--- a/tools/bashreadline.py
+++ b/tools/bashreadline.py
@@ -61,4 +61,7 @@ def print_event(cpu, data, size):
 
 b["events"].open_perf_buffer(print_event)
 while 1:
-    b.perf_buffer_poll()
+    try:
+        b.perf_buffer_poll()
+    except KeyboardInterrupt:
+        exit()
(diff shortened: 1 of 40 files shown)


palmtenor:
[buildbot, ok to test]

yonghong-song:
Could you share the `ugly backtrace` you have seen? I am curious what it is.
It would be interesting to know in what situations you could see these kinds of backtraces.

I have seen situations where there are so many events (e.g., through trace.py tracing `t:raw_syscall:sys_enter`), it seems stuck inside the loop in function `perf_reader_event_read()`
and Ctrl-C even fails and I have to do Ctrl-Z. I guess this is probably different from your situation.
I tried a few applications it seems all generate `sane` tracebacks.

jeromemarchand:
By ugly backtrace, I mean **any** backtrace.  The backtraces don't show anything wrong and the program usually exit right away (the only exception I've seen is in funcslower where I often need to hit Ctrl-C several times to quit - but that's another story). It just that users don't like to see error message when they quit a program.

yonghong-song:
Yes. We may have a few other places, e.g., `trace.py`, we also need to add this change. But this can be done separately.

PR #2084, opened 2018-12-17
dbslower: fix a python3 bytes/string issue int the -x option

jeromemarchand:
In python3, the find method requires a bytes-like object. It fixes the
following error:

$ dbslower mysql -x $(which mysqld)
  File "/usr/share/bcc/tools/dbslower", line 72, in <module>
    if mysql_func_name.find("COM_DATA") >= 0:

Also the -x option is currently undocumented in the man page and the
example file. So let's ix that too.


yonghong-song:
[buildbot, test this please]

yonghong-song:
The fix looks good to me. Thanks!

PR #2094, opened 2018-12-27
trace.py: fix compiler warning

pbhole:
Compiler shows warning "incompatible integer to pointer conversion
initializing" while compiling bpf program.
This patch adds necessary typecast when assigning PT_REGS_PARAM vaules
to struct pt_regs pointer


yonghong-song:
Looks good to me. Thanks!

PR #2096, opened 2018-12-28
opensnoop: print flags, enable filtering

timdoug:
This PR adds printing of the `flags` argument to `open(2)` and allows for filtering, e.g., to only print out files for writing.

```
$ sudo ./opensnoop.py -f O_WRONLY -f O_RDWR
PID    COMM               FD ERR FLAGS    PATH
15286  postgres            3   0 00101101 /var/run/postgresql/9.6-main.pg_stat_tmp/global.tmp
15286  postgres            6   0 00101101 /var/run/postgresql/9.6-main.pg_stat_tmp/db_0.tmp
23562  postgres           11   0 00100001 /proc/self/oom_score_adj
23562  postgres            3   0 00100002 base/16385/2601
...
```

Thanks!

### DIFF
--- a/tools/opensnoop_example.txt
+++ b/tools/opensnoop_example.txt
@@ -135,7 +135,7 @@ to the '-n' option.
 The -e option prints out extra columns; for example, the following output
 contains the flags passed to open(2), in octal:
 
-# ./opensnoop.py -e
+# ./opensnoop -e
 PID    COMM               FD ERR FLAGS    PATH
 28512  sshd               10   0 00101101 /proc/self/oom_score_adj
 28512  sshd                3   0 02100000 /etc/ld.so.cache
@@ -150,7 +150,7 @@ PID    COMM               FD ERR FLAGS    PATH
 
 The -f option filters based on flags to the open(2) call, for example:
 
-$ sudo ./opensnoop.py -e -f O_WRONLY -f O_RDWR
+# ./opensnoop -e -f O_WRONLY -f O_RDWR
 PID    COMM               FD ERR FLAGS    PATH
 28084  clear_console       3   0 00100002 /dev/tty
 28084  clear_console      -1  13 00100002 /dev/tty0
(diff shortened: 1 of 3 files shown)


timdoug:
Thanks @brendangregg for the review. I've made the following changes:

- The `FLAGS` field is now conditionally printed based on the `-e` flag.
- The structs only have the flag member when it's used. The Python string mangling feels a bit hacky, but it's comparable to the technique used for `pid`/`tid` filtering. The `ctypes` `Structure` still has the `flags` entry even when it's not used, but I don't know how to nicely conditionally provide it, and it empirically seems safe this way.
- I moved the flags filtering from Python to eBPF.
- I've updated the man page and the `_examples.txt` documentation.

brendangregg:
Thanks! The man page is missing FLAGS under the FIELDS section, but I think
that's all.

Yes, the string substitution is hacky, but that's what we've done so far in
bcc. I'll create a ticket when I'm back at a computer: bcc should provide a
cleaner helper for all this. Maybe python already has an m4-like library we
can use, or maybe we should have a custom one that suits what we are doing
better.

brendangregg:
[buildbot, ok to test]

tools/opensnoop_example.txt
```diff
@@ -148,6 +178,10 @@ optional arguments:
   -d DURATION, --duration DURATION
                         total duration of trace in seconds
   -n NAME, --name NAME  only print process names containing this name
+  -e, --extended_fields
```
pchaigno: I think we've used the `--extended-fields` syntax in the past. See [`trace_example.txt`](https://github.com/timdoug/bcc/blob/7a1c85eec5395c020be61f24efbe8c15b464f3bb/tools/trace_example.txt) for instance.

brendangregg:
Thanks, merging.

brendangregg:
Correction, I'm unable to merge until the buildbot is happy.

[buildbot, ok to retest]

brendangregg:

yonghong-song:
fc28 build failure. looks like an infra failure.

yonghong-song:

yonghong-song:
[buildbot, test this please]

PR #2097, opened 2018-12-29
implement free_bcc_memory() API

yonghong-song:
The main purpose of this API is to proactively release llvm/clang
.text memory which is brought in during compilation.
bcc .text memory for some other functions, e.g., attach_tracepoint,
bpf_prog_load, etc. can also be freed after all these tasks are done.

Note that such memory is reclaimable in kernel since it has
file backup. But certain applicaiton may want to reduce this
memory immediately to satisfy constraints imposed by sysadmin, etc.

The implementation uses madvise with MADV_DONTNEED.
For the case where bcc is static linked into the binary,
we do not really know the start and the end of memory regions
used by bcc, so the implementation here bluntly returned
all .text memory back to kernel. This will incur some performance
overhead as later on executed instructions will need to bring
back to memory again.

For static linked library, instrumented RandomRead example,
without this patch, the RSS memory before load is:
```
  VmRSS:     63644 kB
  RssAnon:           23876 kB
  RssFile:           39768 kB
  RssShmem:              0 kB
```
After this patch,
```
  VmRSS:     34264 kB
  RssAnon:           23880 kB
  RssFile:           10384 kB
  RssShmem:              0 kB
```
For shared library, a python unit test, test_free_llvm_memory.py, is
added, which shows for a do-nothing bpf program, we have
```
  Before freeing llvm memory: RssFile:  43000 kB
  After  freeing llvm memory: RssFile:  11992 kB
```
The RssFile reduction on Facebook internal applications
also ranges in 30-40MB.

### DIFF
--- a/examples/cpp/RandomRead.cc
+++ b/examples/cpp/RandomRead.cc
@@ -117,6 +117,12 @@ int main(int argc, char** argv) {
     return 1;
   }
 
+  // done with all initial work, free bcc memory
+  if (bpf->free_bcc_memory()) {
+    std::cerr << "Failed to free llvm/clang memory" << std::endl;
+    return 1;
+  }
+
   signal(SIGINT, signal_handler);
   std::cout << "Started tracing, hit Ctrl-C to terminate." << std::endl;
   while (true)
--- a/src/cc/CMakeLists.txt
+++ b/src/cc/CMakeLists.txt
@@ -45,7 +45,7 @@ set(bcc_util_sources ns_guard.cc common.cc)
 set(bcc_sym_sources bcc_syms.cc bcc_elf.c bcc_perf_map.c bcc_proc.c)
 set(bcc_common_headers libbpf.h perf_reader.h)
 set(bcc_table_headers file_desc.h table_desc.h table_storage.h)
-set(bcc_api_headers bpf_common.h bpf_module.h bcc_exception.h bcc_syms.h)
+set(bcc_api_headers bpf_common.h bpf_module.h bcc_exception.h bcc_syms.h bcc_elf.h)
 
 if(ENABLE_CLANG_JIT)
 add_library(bcc-shared SHARED
--- a/src/cc/api/BPF.cc
+++ b/src/cc/api/BPF.cc
@@ -27,6 +27,7 @@
 #include <vector>
 
 #include "bcc_exception.h"
+#include "bcc_elf.h"
 #include "bcc_syms.h"
 #include "bpf_module.h"
 #include "common.h"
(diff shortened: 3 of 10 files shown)


4ast:
great stuff!
Is it possible to determine start/end of text without parsing the elf?
May be some linker magic (similar to kernel) to add two special markers?

yonghong-song:
@4ast This heavily depends on linker script. The default gcc linker script likes below:
```
-bash-4.4$ gcc main.c -Wl,-verbose                                                                                                           
...
ENTRY(_start)
...
  .plt            : { *(.plt) *(.iplt) }
.plt.got        : { *(.plt.got) }
.plt.bnd        : { *(.plt.bnd) }
  .text           :
  {
    *(.text.unlikely .text.*_unlikely .text.unlikely.*)
    *(.text.exit .text.exit.*)
    *(.text.startup .text.startup.*)
    *(.text.hot .text.hot.*)
    *(.text .stub .text.* .gnu.linkonce.t.*)
    /* .gnu.warning sections are handled specially by elf32.em.  */
    *(.gnu.warning)
  }
  .fini           :
  {
    KEEP (*(SORT_NONE(.fini)))
  }
```
So there is no begin/end marker of the .text section. For end, we could use `_fini` symbol which defined in glibc for approximation purpose, but if applications has a lot of fini functions even that can be screwed. There is no easy way to guess the beginning of .text section. The known `_start` symbol is not guaranteed sitting in the very beginning of the .text section.

The linux linker script for x86 provides the following markers
```
        /* Text and read-only data */
        .text :  AT(ADDR(.text) - LOAD_OFFSET) {
                _text = .;
                _stext = .;
                /* bootstrapping code */
                HEAD_TEXT
                TEXT_TEXT
                SCHED_TEXT
                CPUIDLE_TEXT
                LOCK_TEXT
                KPROBES_TEXT
                ALIGN_ENTRY_TEXT_BEGIN
                ENTRY_TEXT
                IRQENTRY_TEXT
                ALIGN_ENTRY_TEXT_END
                SOFTIRQENTRY_TEXT
                *(.fixup)
                *(.gnu.warning)
                
#ifdef CONFIG_RETPOLINE
                __indirect_thunk_start = .;
                *(.text.__x86.indirect_thunk)
                __indirect_thunk_end = .;
#endif

                /* End of text section */
                _etext = .;
        } :text = 0x9090
```
so `_text` and `_stext` can be used to mark the beginning of .text section and `_etext` can be used to mark the end of .text section.

yonghong-song:
The reliable marker for .text section is to change the linker script for the application, but that is probably too restricted...

Issue #226, opened 2015-09-17
string assignment invalid BPF_LD_IMM insn


brendangregg:
When dealing with strings, I often want to switch to a default if a given char \* is null.

Test program:

``` Python
from __future__ import print_function
from bcc import BPF

# load BPF program
b = BPF(text = """
void kprobe__sys_sync(void *ctx) {
        char *out = NULL, *hello = "world";
        if (out == NULL) {
                out = hello;
        }
        bpf_trace_printk("sync() %s\\n", out);
};
""")

# header
print("%-18s %s" % ("TIME(s)", "CALL"))

# format output
while 1:
        (task, pid, cpu, flags, ts, msg) = b.trace_fields()
        print("%-18.9f %s" % (ts, msg))
```

``` Shell
# ./teststr 
bpf: Invalid argument
0: (b7) r1 = 0
1: (73) *(u8 *)(r10 -6) = r1
2: (b7) r1 = 2675
3: (6b) *(u16 *)(r10 -8) = r1
4: (18) r1 = 0x636e7973
6: (7b) *(u64 *)(r10 -16) = r1
7: (bf) r1 = r10
8: (07) r1 += -16
9: (b7) r2 = 11
10: (00) r0 = 0x0
invalid BPF_LD_IMM insn

Traceback (most recent call last):
  File "./teststr", line 26, in <module>
    """)
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 353, in __init__
    self._trace_autoload()
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 573, in _trace_autoload
    fn = self.load_func(func_name, BPF.KPROBE)
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 388, in load_func
    raise Exception("Failed to load BPF program %s" % func_name)
Exception: Failed to load BPF program kprobe__sys_sync
```

drzaeus77:
Global strings won't work. Try changing it to `hello[] = "world"`

brendangregg:
Yes, that worked, thanks. We'll need to include an example of that somewhere at least.

PR #2101, opened 2018-12-31
better error meessage for error "unknown opcode"

yonghong-song:
fix issue #226

The unknown opcode typically happens if the bpf
program has an external reference which does not
get resolved. Note bcc does not even preform
relocations for maps as map_id is directly
used in bpf problem through bpf_pseudo_fd()
intrinsic.

Instead of the error:
  bpf: Failed to load program: Invalid argument
  unknown opcode 00

A little explanation is added like the below:
  HINT: The 'unknown opcode' can happen if you referencea global
  or static variable, or data in read only section.
  For example,'char *p = "hello"' will result in p referencing a
  read only section,and 'char p[] = "hello"' will have "hello"
  stored on the stack.

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -344,6 +344,14 @@ static void bpf_print_hints(int ret, char *log)
       "you'll need to be explicit.\n\n");
   }
 
+  // referencing global/static variables or read only data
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

PR #2102, opened 2019-01-02
Unify python executables to use `#!/usr/bin/env python` shebang

SaveTheRbtz:
Currently some python executables either do not have shebang or have it hardcoded to `#!/usr/bin/python`. This is causing some problems on our production systems. 

RPM specs currently workaround that problem by `sed`ing the resulting files:
https://github.com/iovisor/bcc/blob/dccc4f28b7404b2f77f8ad8c7ad1ea741db589f2/SPECS/bcc.spec#L91-L94

This diff:
* rewrites shebangs across all executable python files to: `#!/usr/bin/env python`.
* removes the mangling code from RPM specs.
* adds a travis-ci test to verify that all python executables continue to have proper shabangs.

PS. It seems like the travis build is not enabled for bcc: https://travis-ci.org/iovisor/bcc

PR #2103, opened 2019-01-03
ustat: added basic race condition protection

SaveTheRbtz:
In #2102 build there were 4 test failures on Fedora.

3 of them were due to the following traceback:
```
27:   File "../../tools/lib/ustat.py", line 294, in <module>
27:     Tool().run()
27:   File "../../tools/lib/ustat.py", line 286, in run
27:     self._loop_iter()
27:   File "../../tools/lib/ustat.py", line 255, in _loop_iter
27:     counts.update(probe.get_counts(self.bpf))
27:   File "../../tools/lib/ustat.py", line 114, in get_counts
27:     result[pid.value][category] = count.value
27: KeyError: 20335L
```

Seems like there is a possible inconsistency between `self.targets` and `bpf[*_counts]`:
https://github.com/iovisor/bcc/blob/dccc4f28b7404b2f77f8ad8c7ad1ea741db589f2/tools/lib/ustat.py#L107-L116

Workaround it by skipping PIDs that are not in `result` dict.

Also while here, fix a race condition on the `USDT` instrumentation init and PID going away, which happens on the boxes with high process churn.

### DIFF
--- a/tools/lib/ustat.py
+++ b/tools/lib/ustat.py
@@ -20,8 +20,9 @@
 
 from __future__ import print_function
 import argparse
-from bcc import BPF, USDT
+from bcc import BPF, USDT, USDTException
 import os
+import sys
 from subprocess import call
 from time import sleep, strftime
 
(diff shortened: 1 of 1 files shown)


SaveTheRbtz:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

yonghong-song:
changes look good to me. Thanks for the fix!

PR #2049, opened 2018-11-20
Remove duplicate code from xdp_drop_count.py

sevagh:
Also, fix an incorrect indentation in the same block.

Fixes https://github.com/iovisor/bcc/issues/2048

PR #2107, opened 2019-01-03
Comment on duplicate code for parsing double vlans

pchaigno:
I also changed it into a short loop, which I don't think improves (or degrades) the code, but at least makes it clear that the double parsing is intentional.

See #1493 and #2049.


yonghong-song:
[buildbot, test this please]

PR #2109, opened 2019-01-04
biolatency.py: use blk_account_io_done instead of blk_account_io_comp…

gurugio:
…letion

The biolatency tool attaches kprobe at blk_account_io_completion() function
to check when a request is finished. But sometimes request can be finished
partially and blk_account_io_completion() is called again.
The blk_account_io_done() is called when a request is finally
finished and biolatency tool should attach blk_account_io_done(), not
blk_account_io_completion().

Signed-off-by: Gioh Kim <gi-oh.kim@cloud.ionos.com>

### DIFF
--- a/tools/biolatency.py
+++ b/tools/biolatency.py
@@ -67,7 +67,7 @@ int trace_req_start(struct pt_regs *ctx, struct request *req)
 }
 
 // output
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

PR #2110, opened 2019-01-04
Add stdout flushing to some tools

japroc:
I use next tools in one of my projects:
- opensnoop
- solisten
- tcpaccept
- tcpconnect
But i dont get output without flushing output.
Saw in execsnoop tool that bcc.utils.printb is used for flushing. Replaced standart python print function with this one.


yonghong-song:
[buildbot, test this please]

yonghong-song:
Looks good. Thanks! Probably some other tools needs to be updated as well for similar print issues. But we can do it later.

PR #2111, opened 2019-01-04
cmake/python: fix build race condition

SaveTheRbtz:
A followup for the build flakes from https://github.com/iovisor/bcc/pull/2102#issuecomment-451304517

Rewritten `CMakeLists.txt` for the python code. Now it:
* autodiscovers .py and .py.in files, so there is no need to hardcode them there.
* copies everything to a separate directory for each python provided.
* does not have `PREVIOUS_PY` logic.
* runs in parallel.

Also while here removed the MANIFEST file since:
* it is outdated.
* `distutils` regenerates it anyway.

Tested locally with:
```
cmake ../src/python -DPYTHON_CMD="python;python2;python3;python2.7;python3.5"
while :; do make clean; make -j || break; done; echo $?
^C
make clean; make -j; make install
```

Please note that I have no idea what i'm doing, since this is the first time I'm touching CMake, not to mention using it to build python extensions. It would be nice if someone who actually knows cmake could take a look at this diff.
Main known issue is why do we build `sdist` in the first place, since I do not see it being used anywhere and maybe we should use `build` instead?

cc: @yonghong-song 
ref: #1951  #2102


yonghong-song:
[buildbot, test this please]

SaveTheRbtz:

Did you consider enabling build automatically? If there is a concern about the Jenkins abuse you can consider switching it from a whitelist to blacklist.

yonghong-song:
make sense. Let us discuss in the iovisor bi-weekly meeting to find an appropriate approach.

yonghong-song:
I checked and tested the patch for full build with multiple python commands. It works fine. The newly changed makefile has separate build and install directory for each python command. The old way has one build directory and multiple install directories.  The new way is cleaner and safer as well.

PR #2113, opened 2019-01-06
opensnoop: supported UID options

takumakume:
Hi. Nice to meet you!

UID related options supported.

I think that it is use case of a multi-tenant environment.

- The -U option include UID on output:
  ```sh
  # ./opensnoop -U
  UID   PID    COMM               FD ERR PATH
  0     27063  vminfo              5   0 /var/run/utmp
  103   628    dbus-daemon        -1   2 /usr/local/share/dbus-1/system-services
  103   628    dbus-daemon        18   0 /usr/share/dbus-1/system-services
  103   628    dbus-daemon        -1   2 /lib/dbus-1/system-services
  ```

The reason for the print extension by the `-U` option is to maintain backward compatibility.

- The -u option filtering UID:
  ```sh
  # ./opensnoop -Uu 1000
  UID   PID    COMM               FD ERR PATH
  1000  30240  ls                  3   0 /etc/ld.so.cache
  1000  30240  ls                  3   0 /lib/x86_64-linux-gnu/libselinux.so.1
  1000  30240  ls                  3   0 /lib/x86_64-linux-gnu/libc.so.6
  1000  30240  ls                  3   0 /lib/x86_64-linux-gnu/libpcre.so.3
  1000  30240  ls                  3   0 /lib/x86_64-linux-gnu/libdl.so.2
  1000  30240  ls                  3   0 /lib/x86_64-linux-gnu/libpthread.so.0
  ```

### DIFF
--- a/man/man8/opensnoop.8
+++ b/man/man8/opensnoop.8
@@ -27,6 +27,9 @@ Print usage message.
 \-T
 Include a timestamp column.
 .TP
+\-U
+Show UID.
+.TP
 \-x
 Only print failed opens.
 .TP
@@ -35,6 +38,9 @@ Trace this process ID only (filtered in-kernel).
 .TP
 \-t TID
 Trace this thread ID only (filtered in-kernel).
+.TP
+\-u UID
+Trace this UID only (filtered in-kernel).
 .TP
 \-d DURATION
 Total duration of trace in seconds.
@@ -61,6 +67,10 @@ Trace all open() syscalls, and include timestamps:
 #
 .B opensnoop \-T
 .TP
+Show UID:
+#
+.B opensnoop \-U
+.TP
 Trace only open() syscalls that failed:
 #
 .B opensnoop \-x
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
@brendangregg what is your opinion on new options? Any better alternative? The `-u` and `-U` in various other tools have been used as an option to get user stack. But for `opensnoop`, `-u` and `-U` seem fine as stacks are not dumped in this tool.

brendangregg:
Thanks for the addition @takumakume . @yonghong-song  I'm fine with it. There's not much consistency with -u at the moment anyway (microseconds, stacks). I'd be more worried about flags like -p and -v, which would have stronger expectations, and I'd want to always be used for PID and verbose.

takumakume:
Thank you for checking !
I'd like to add the same implementation with `tcpconnect` .

PR #2099, opened 2018-12-31
Need support for BUILD_ID support for BPF stackmap

vijunag:
Added build_id support for BPF stackmap.
The scope of the changes is as below
1) Create a new stackmap with the required BUILD_ID flag. It is achieved with the help of a macro. 
2) Support for adding builds to build sym cache. A mapping is setup between build_id to resolve symbols within the file. 
3) Changes to libbcc python module to support the stack trace reported in the new format. 
4) Made sure the changes are backwards compatible. 
5) An example tool is also available in tools/stack_buildid_example.py.

PR #2115, opened 2019-01-07
Fix #2098

### LINKED ISSUE #2098: Need support for BUILD_ID support for BPF stackmap
The latest kernel also reports build_id in stacktrace. BCC should also report and resolve stack trace containing build_id.

vijunag:
Added build_id support for BPF stackmap.
A separate build_id stackmap can be created with the help of new macro BPF_STACK_TRACE_BUILDID
The kernel BPF reports stacktrace in the structure bpf_stack_build_id.
Changes have been made to BPF modules to support resolving symbols mentioned in the above format.
An example tool is also available in tools/stack_buildid_example.py

Fix #2098
Added build_id support for BPF stackmap.
A separate build_id stackmap can be created with the help of new macro BPF_STACK_TRACE_BUILDID
The kernel BPF reports stacktrace in the structure bpf_stack_build_id.
Changes have been made to BPF modules to support resolving symbols mentioned in the above format.
An example tool is also available in tools/stack_buildid_example.py

Fix #2098

Moved stack_buildid_example to examples directory

Moved to examples/tracing

Fix #2099 - Added eBPF APIs and test cases related to build_id stackmap

Fix #2098, #2099 - Fixed review comments

### DIFF
--- a/src/cc/api/BPF.cc
+++ b/src/cc/api/BPF.cc
@@ -86,6 +86,8 @@ BPF::~BPF() {
   if (res.code() != 0)
     std::cerr << "Failed to detach all probes on destruction: " << std::endl
               << res.msg() << std::endl;
+  bcc_free_buildsymcache(bsymcache_);
+  bsymcache_ = NULL;
 }
 
 StatusTuple BPF::detach_all() {
@@ -648,6 +650,21 @@ BPFStackTable BPF::get_stack_table(const std::string& name, bool use_debug_file,
   if (bpf_module_->table_storage().Find(Path({bpf_module_->id(), name}), it))
     return BPFStackTable(it->second, use_debug_file, check_debug_file_crc);
   return BPFStackTable({}, use_debug_file, check_debug_file_crc);
+}
+
+BPFStackBuildIdTable BPF::get_stackbuildid_table(const std::string &name, bool use_debug_file,
+                                                 bool check_debug_file_crc) {
+  TableStorage::iterator it;
+
+  if (bpf_module_->table_storage().Find(Path({bpf_module_->id(), name}), it))
+    return BPFStackBuildIdTable(it->second, use_debug_file, check_debug_file_crc, get_bsymcache());
+  return BPFStackBuildIdTable({}, use_debug_file, check_debug_file_crc, get_bsymcache());
+}
+
+bool BPF::add_module(std::string module)
+{
+  return bcc_buildsymcache_add_module(get_bsymcache(), module.c_str()) != 0 ?
+    false : true;
 }
 
 std::string BPF::get_uprobe_event(const std::string& binary_path,
--- a/src/cc/api/BPF.h
+++ b/src/cc/api/BPF.h
@@ -49,6 +49,7 @@ class BPF {
   explicit BPF(unsigned int flag = 0, TableStorage* ts = nullptr,
                bool rw_engine_enabled = bpf_module_rw_engine_enabled(), const std::string &maps_ns = "")
       : flag_(flag),
+        bsymcache_(NULL),
       bpf_module_(new BPFModule(flag, ts, rw_engine_enabled, maps_ns)) {}
   StatusTuple init(const std::string& bpf_program,
                    const std::vector<std::string>& cflags = {},
@@ -137,6 +138,13 @@ class BPF {
     return BPFPercpuHashTable<KeyType, ValueType>({});
   }
 
+  void* get_bsymcache(void) {
+    if (bsymcache_ == NULL) {
+      bsymcache_ = bcc_buildsymcache_new();
+    }
+    return bsymcache_;
+  }
+
   BPFProgTable get_prog_table(const std::string& name);
 
   BPFCgroupArray get_cgroup_array(const std::string& name);
@@ -146,6 +154,12 @@ class BPF {
   BPFStackTable get_stack_table(const std::string& name,
                                 bool use_debug_file = true,
                                 bool check_debug_file_crc = true);
+
+  BPFStackBuildIdTable get_stackbuildid_table(const std::string &name,
+                                              bool use_debug_file = true,
+                                              bool check_debug_file_crc = true);
+
+  bool add_module(std::string module);
 
   StatusTuple open_perf_event(const std::string& name, uint32_t type,
                               uint64_t config);
@@ -224,6 +238,8 @@ class BPF {
                                   uint64_t& offset_res);
 
   int flag_;
+
+  void *bsymcache_;
 
   std::unique_ptr<std::string> syscall_prefix_;
 
--- a/src/cc/bcc_syms.h
+++ b/src/cc/bcc_syms.h
@@ -21,6 +21,7 @@ extern "C" {
 #endif
 
 #include <stdint.h>
+#include <compat/linux/bpf.h>
 
 struct bcc_symbol {
   const char *name;
@@ -67,6 +68,13 @@ void bcc_symcache_refresh(void *resolver);
 int bcc_resolve_global_addr(int pid, const char *module, const uint64_t address,
                             uint64_t *global);
 
+/*bcc APIs for build_id stackmap support*/
+void *bcc_buildsymcache_new(void);
+void bcc_free_buildsymcache(void *symcache);
+int  bcc_buildsymcache_add_module(void *resolver, const char *module_name);
+int bcc_buildsymcache_resolve(void *resolver,
+                              struct bpf_stack_build_id *trace,
+                              struct bcc_symbol *sym);
 // Call cb on every function symbol in the specified module. Uses simpler
 // SYM_CB callback mainly for easier to use in Python API.
 // Will prefer use debug file and check debug file CRC when reading the module.
--- a/src/cc/syms.h
+++ b/src/cc/syms.h
@@ -147,3 +147,42 @@ public:
   virtual bool resolve_name(const char *module, const char *name,
                             uint64_t *addr);
 };
+
+class BuildSyms {
+  struct Symbol {
+    Symbol(const std::string *name, uint64_t start, uint64_t size)
+      :name(name), start(start), size(size) {}
+    const std::string *name;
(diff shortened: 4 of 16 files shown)


yonghong-song:
[buildbot, test this please]

vijunag:
Fixed build test errors. Can you please request test-bot to give it a spin again ?

yonghong-song:
[buildbot, test this please]

vijunag:
@yonghong-song - Test case failed on one of the test beds i.e fc28 with the below BT. The test expects the symbol to be resolved as __getuid and it returned unknown instead. Can you please check why "test_stackid.py" failed ? May be because Get_libc_path() didn't return libc path properly ?

How to get access to this box or how to simulate this error ?

21:   File "/home/fedora/jenkins/workspace/bcc-pr/label/fc28/tests/python/test_stackid.py", line 90, in test_simple
21:     self.assertEqual(b.sym(stack.trace[0], -1), b"__getuid")
21: AssertionError: b'[unknown]' != b'__getuid'

src/python/bcc/__init__.py
```diff
@@ -1189,7 +1192,31 @@ def sym(addr, pid, show_module=False, show_offset=False, demangle=True):
         Example output when both show_module and show_offset are False:
             "start_thread"
         """
-        name, offset, module = BPF._sym_cache(pid).resolve(addr, demangle)
+
+        #addr is of type stacktrace_build_id
+        #so invoke the bsym address resolver
+        typeofaddr = str(type(addr))
+        if typeofaddr.find('bpf_stack_build_id') > -1:
```
palmtenor: Why can't we just do equality here?

src/cc/api/BPF.h
```diff
@@ -137,6 +137,10 @@ class BPF {
     return BPFPercpuHashTable<KeyType, ValueType>({});
   }
 
+  void* get_bsymcache(void) {
```
palmtenor: How does this work? I imagine user would get this (bsymcache) to add module. But how is it initialized?

If you could also add a C++ example that would be great:))!

src/cc/bcc_syms.h
```diff
@@ -29,6 +29,27 @@ struct bcc_symbol {
   uint64_t offset;
 };
 
+enum bcc_stack_build_id_status {
```
palmtenor: Could we just use enum and definition from the Kernel header (`"compat/linux/bpf.h"`) instead of redefining them again all over the place?

src/cc/syms.h
```diff
@@ -147,3 +147,47 @@ class ProcSyms : SymbolCache {
...
+  };
+
+  std::unordered_map<std::string, Module*> buildmap_;
+
+public:
+  BuildSyms() {}
+  virtual ~BuildSyms() {
+    std::unordered_map<std::string, Module*>::iterator it;
+    for (it=buildmap_.begin();it!=buildmap_.end();++it) {
+      delete it->second;
```
palmtenor: Why not just `unique_ptr`?

yonghong-song:
For the fc28 testing environment:
  - you cannot access the VM in the jenkins. But you can just download a fc28 iso image and set up by yourself
  - the libc location is /lib64/libc.so.6
  - both python version is tested on fc28, python2.7 and python3.6.

src/python/bcc/__init__.py
vijunag: Ok.

src/cc/api/BPF.h
vijunag: test_bpf_table.cc has the required example as a test-case. Also, it is modeled similar to BPF python object. As suggested in another comment, I moved the bsymcache_ at the BPF level.

src/cc/bcc_syms.h
vijunag: Idea was to be backwards compatible with the kernel that doesn't support this type of reporting. The functionality may break, but build doesn't break at-least.

src/cc/syms.h
vijunag: Did you mean buildmap_ can be an unique_ptr ?

vijunag:
@yonghong-song - Can you request a fresh spin on the test bed ? I fixed the test cases that were failing.

yonghong-song:
[buildbot, test this please]

vijunag:
Build failure and not related to the checkin. 
+ cd build
+ [[ fc28 fedora28-slave-251 = *\f\c\2\8* ]]
+ EXTRA_CMAKE_FLAGS=-DENABLE_LLVM_SHARED=1
+ sudo ln -s /usr/bin/python3 /usr/local/bin/python
ln: failed to create symbolic link '/usr/local/bin/python': File exists
Build step 'Execute shell' marked build as failure
INFO: Starting to record.

src/cc/bcc_syms.h
palmtenor: This include with use the (latest) copy and we pull into BCC. See how `libbpf.h` uses the header and how we sync it with upstream periodically (Example #2100)

src/cc/syms.h
palmtenor: I mean the map values could be, so you don't have to remember manually delete them?

yonghong-song:
Seems an infra issue for fc28. Let us try again.

yonghong-song:
[buildbot, test this please]

src/cc/bcc_syms.h
vijunag: Now I remember why I had to use an auxiliary type. These structures are passed between python and c++ using ctypes abstraction. Since these are defined in a kernel header, the type is never defined and hence it was not possible the define the signatures in libbcc.bcc_buildsymc apis. This worked for the normal stackmap because it was a POD type. I have removed redundant definitions in libbcc libraries as suggested.

vijunag:
@yonghong-song @palmtenor - I have addressed all the review comments. While I have fixed the python test case related failures, there is a SIGSEGV in the newly added cxx test case. I'm unable to reproduce this on my local fc28. Can you please provide the BT if possible ?

yonghong-song:
[buildbot, test this please]

tests/cc/test_bpf_table.cc
```diff
@@ -222,3 +222,59 @@ TEST_CASE("test bpf stack table", "[bpf_stack_table]") {
...
+  REQUIRE(res.code() == 0);
+  REQUIRE(getuid() >= 0);
+  res = bpf.detach_kprobe(getuid_fnname);
+  REQUIRE(res.code() == 0);
+
+  auto id = bpf.get_hash_table<int, int>("id");
+  auto stack_traces = bpf.get_stackbuildid_table("stack_traces");
+
+  /* libc locations on different distributions are added below*/
+  stack_traces.add_module("/lib/x86_64-linux-gnu/libc.so.6"); //Location of libc in ubuntu
```
yonghong-song: I remembered I suggested instead of `stack_traces.add_module(...)`, we should have `bpf.add_module(...)`? This will be consistent with python interface. Any concern with this?

@palmtenor what do you think?

yonghong-song:
For the question:
>> there is a SIGSEGV in the newly added cxx test case. I'm unable to reproduce this on my local fc28. Can you please provide the BT if possible ?

Yes, I can reproduce the seg fault. Try to do the following a few things:
  - rebase on top of bcc trunk. This will make testing on local branch easier as there is a Makefile change recently. I suggest you push you rebased change to github as well.
  - FYI, my fc28 kernel version is: 4.18.12-200.fc28.x86_64

The failure happens for statement:
```
  stack_traces.add_module("/lib/x86_64-linux-gnu/libc.so.6"); //Location of libc in ubuntu
```
Traced a little more and looks like in the following routine, the `resolver` (or `bsym`) could be invalid/garbage.
```
int  bcc_buildsymcache_add_module(void *resolver, const char *module_name)
{
  BuildSyms *bsym = static_cast<BuildSyms *>(resolver);
  int ret = bsym->add_module(module_name);
  return  ret ? 0 : -1;
}
```

Maybe you want to debug along this line.

vijunag:
@yonghong-song  - Thank you so much for the tip. I had installed a custom kernel whose version wasn't matching with the version.h file and hence the test-case never executed on my kernel. I had missed adding bsymcache_ to constructor initialization list. I guess all the test cases should pass now. Please let me know if the changes look good now.

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

vijunag:
I guess we can move it to BPF class for the sake of consistency. In-fact it makes more sense to be at BPF level.

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

src/cc/api/BPF.cc
```diff
@@ -650,6 +652,25 @@ BPFStackTable BPF::get_stack_table(const std::string& name, bool use_debug_file,
...
+
+  if (bpf_module_->table_storage().Find(Path({bpf_module_->id(), name}), it))
+    return BPFStackBuildIdTable(it->second, use_debug_file, check_debug_file_crc, bsymcache_);
+  return BPFStackBuildIdTable({}, use_debug_file, check_debug_file_crc, bsymcache_);
+}
+
+bool BPF::add_module(std::string module)
+{
+  return bcc_buildsymcache_add_module(bsymcache_, module.c_str()) != 0 ?
+    false : true;
```
yonghong-song: Currently, C++ application needs to do get_stackbuildid_table first before calling add_module. Do you think in BPF::add_module() call, we should also call
```
  if (bsymcache_ == NULL) {
    bsymcache_ = bcc_buildsymcache_new();
  }
```
so that user can add modules before calling `get_stackbuildid_table`?

I agree with your current implementation that a bsymcache is built only if it may be used.

vijunag: Point taken. Made the necessary change.

yonghong-song:
[buildbot, test this please]

yonghong-song:
looks good. Thanks!

PR #2131, opened 2019-01-10
Fix BPFModule error when RW engine is disabled

palmtenor:
In #2080 we added the option to build completely without RW Engine code. This works out mostly, except when:
- `ENABLE_LLVM_NATIVECODEGEN=OFF` in build config
- `BPFModule` is constructed with default constructor arguments, which has `rw_engine_enabled = true` by default.

In such case, `BPFModule` would still call the (non-existent) RW Engine logic, which would either error or segfault. Specifically, the Python interface is using `BPFModule` with default constructor arguments, meaning all Python tool would break if `ENABLE_LLVM_NATIVECODEGEN` is turned off.

This commit fixes that by checking for the `bpf_module_rw_engine_enabled()` in `BPFModule` constructor.


yonghong-song:
The fc28 failure seems like a testinfra failure. Let us do another round of test.

yonghong-song:
[buildbot, test this please]

palmtenor:
On the other hand, would it be possible for buildbot to test on a few different build arguments? It currently only build and test the default argument so similar things like this could happen again in the future. While we don't have to have it build all `2^n` combinations, try a few important variations would be great.

@drzaeus77 would this be possible?

yonghong-song:
[buildbot, test this please]

torgil:
Nice. I think we should keep the function in the default arguments though and calls to keep things explicit. 
Add it here instead of "true" ?
BPFModule(unsigned flags, TableStorage *ts = nullptr, bool rw_engine_enabled = true,

... and to all places that makes a BPFModule instance that sets this argument. Then maybe the constructor could fail instead of implicitly changing user wishes to be even more explicit.

palmtenor:
@torgil That makes sense. I would ideally wish to make `bpf_module_rw_engine_enabled()` a constexpr function but I'm not sure if some of our older compilers (the FC25 buildbot seems still on GCC4.8) support that.

palmtenor:
@torgil Actually, I just realized it might not be easy to do. The constructor's default argument is defined in the header file, so it could not use a function that's only in the source code file (i.e. `bpf_module_rw_engine(_disabled).cc`.

PR #2136, opened 2019-01-11
Misc fixes on Python tools

palmtenor:
Mostly things caught by PEP Linter. Some of them are stylistic (like "don't catch bare exception"), but others are syntactically valid:
- `signal` defined both as module name and function argument
- Function default argument has function call
- Undefined local variable `pid` should be `args.pid`
- `uflow_example.txt` has non-ASCII character


yonghong-song:
Looks good to me. Thanks!

PR #2138, opened 2019-01-13
trace.py: fix syscall parameter rewriting on x64

boat0:
On x64 from kernel v4.17 onwards, an indirect table is used to
rewrite syscall parameters in trace.py. However, it only works
for arg1. For arg[2-6], the results are wrong. This patch fixes
that.

Fixes: 2da34267fcae ("generate indirect parameter assignment if arch uses syscall wrapper (#1816)")
Signed-off-by: Xiaozhou Liu <liuxiaozhou@bytedance.com>


yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks for the fix!

PR #2139, opened 2019-01-14
kernel_cache_alloc_free.py monitors kmem_cache

jugurthab:
Demo for tracking **kmem_cache_alloc** and **kmem_cache_free** using TRACEPOINTS

### DIFF
--- a/examples/tracing/kernel_cache_alloc_free.py
+++ b/examples/tracing/kernel_cache_alloc_free.py
@@ -1,0 +1,98 @@
+#!/usr/bin/env python
+#
+# kernel_cache_alloc_free.py	Kernel memory (alloc-free) monitoring tool.
+#		
+#
+# Written as a basic tool example of using ePBF to track of kernel cache allocation and deallocation process.
+#
+#
+# Copyright (c) 2019 Jugurtha BELKALEM.
+# Licensed under the Apache License, Version 2.0 (the "License")
+#
+# 14-Jan-2019 Jugurtha BELKALEM Created this.
+
+from bcc import BPF
+import ctypes as ct
+# prog will store the eBPF C program
+counter = 0;
+prog = """
+struct memory_allocations {
+     u64 timestamp;
+     u64 call_site;
+     u64 ptr;
+     u64 bytes_req;
+     u64 bytes_alloc;
+     u32 gfp_flags;
+     char command[30];
+};
+
+/* Remember : When using tracepoints we need to make use of /sys/kernel/debug/tracing/events directory.
+
+For example : the parameters of the function kmem_cache_alloc are defined in /sys/kernel/debug/tracing/events/kmem/kmem_cache_alloc/format
+ */
+
+
+BPF_PERF_OUTPUT(events_alloc);
+
+BPF_PERF_OUTPUT(events_free);
+
+	TRACEPOINT_PROBE(kmem, kmem_cache_alloc)
+	{
+	      struct memory_allocations memoryAllocationsInstance = {};
+	      memoryAllocationsInstance.timestamp = bpf_ktime_get_ns();
+	      memoryAllocationsInstance.call_site = args->call_site;
+	      memoryAllocationsInstance.ptr = (u64)args->ptr;
+	      memoryAllocationsInstance.bytes_req = args->bytes_req;
+	      memoryAllocationsInstance.bytes_alloc = args->bytes_alloc;
+	      memoryAllocationsInstance.gfp_flags = args->gfp_flags;
+          sprintf(memoryAllocationsInstance.command,"%s", "kmem_cache_alloc");
+    	  events_alloc.perf_submit(args, &memoryAllocationsInstance, sizeof(memoryAllocationsInstance));
+	      return 0; 
+	}
+
+    TRACEPOINT_PROBE(kmem, kmem_cache_free)
+	{
+	      struct memory_allocations memoryFreeInstance = {};
+	      memoryFreeInstance.timestamp = bpf_ktime_get_ns();
+	      memoryFreeInstance.call_site = args->call_site;
+	      memoryFreeInstance.ptr = (u64)args->ptr;
+	      memoryFreeInstance.bytes_req = 0;
+	      memoryFreeInstance.bytes_alloc = 0;
+	      memoryFreeInstance.gfp_flags = 0;
+	      sprintf(memoryFreeInstance.command,"%s", "kmem_cache_free");
+
+	      events_free.perf_submit(args, &memoryFreeInstance, sizeof(memoryFreeInstance));
+	      
+	      return 0;
+	}
+
+"""
+
+# Loads eBPF program
+b = BPF(text=prog)
+
+
+
+
+class Data(ct.Structure):
+    _fields_ = [("timestamp", ct.c_ulonglong), ("call_site", ct.c_ulonglong),("ptr", ct.c_ulonglong),("bytes_req", ct.c_ulonglong),("bytes_alloc", ct.c_ulonglong),("gfp_flags", ct.c_ulonglong), ("command", ct.c_char * 30)]
+
+
+# Show message when ePBF stats
+print("Kernel cache allocation started ... Hit Ctrl-C to end!")
+
+print("%-18s %-20s %-24s %-15s %-12s %-28s %-30s" % ("TIME(S)", "Call_site", "ptr", "bytes_req", "bytes_alloc", "gfp_flags", "Command"))
+
+
+def display_kernel_alloc_free(cpu, data, size):
+    evenement = ct.cast(data, ct.POINTER(Data)).contents
+    print("%-18.0f %-20.0f 0x%-22.0f %-15.0f %-12.0f 0x%-28d %-30s" % (evenement.timestamp, evenement.call_site, evenement.ptr, evenement.bytes_req, evenement.bytes_alloc, evenement.gfp_flags, evenement.command))
+
+b["events_alloc"].open_perf_buffer(display_kernel_alloc_free)
+b["events_free"].open_perf_buffer(display_kernel_alloc_free)
+
+
+
+while 1:
+    # read messages from perf buffer 
+	b.perf_buffer_poll()


yonghong-song:
[buildbot, test this please]

examples/tracing/kernel_cache_alloc_free.py
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python
+#
+# kernel_cache_alloc_free.py	Kernel memory (alloc-free) monitoring tool.
+#		
+#
```
yonghong-song: One empty line is enough in the above.

examples/tracing/kernel_cache_alloc_free.py
```diff
@@ -0,0 +1,98 @@
...
+# Written as a basic tool example of using ePBF to track of kernel cache allocation and deallocation process.
+#
+#
```
yonghong-song: One empty line?

examples/tracing/kernel_cache_alloc_free.py
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python
+#
+# kernel_cache_alloc_free.py	Kernel memory (alloc-free) monitoring tool.
+#		
+#
+# Written as a basic tool example of using ePBF to track of kernel cache allocation and deallocation process.
```
yonghong-song: The same here and below, the line is too long, could you break it into multiple lines. Although sometimes we do have exceptions, it would be good if you can constrain each line to maximum 80 characters.

examples/tracing/kernel_cache_alloc_free.py
```diff
@@ -0,0 +1,98 @@
...
+     u64 ptr;
+     u64 bytes_req;
+     u64 bytes_alloc;
+     u32 gfp_flags;
+     char command[30];
+};
+
+/* Remember : When using tracepoints we need to make use of /sys/kernel/debug/tracing/events directory.
+
+For example : the parameters of the function kmem_cache_alloc are defined in /sys/kernel/debug/tracing/events/kmem/kmem_cache_alloc/format
```
yonghong-song: No need for the empty line in the above. Actually the above comments are not needed. But if you really want, you can just say "the tracepoint kmem/kmem_cache_{alloc,free} format is defined at /sys/kernel/debug/tracing/events/kmem/kmem_cache_{alloc,free}/format and they are automatically retrieved by bcc."

examples/tracing/kernel_cache_alloc_free.py
```diff
@@ -0,0 +1,98 @@
...
+ */
+
+
+BPF_PERF_OUTPUT(events_alloc);
+
```
yonghong-song: no need for empty line.

examples/tracing/kernel_cache_alloc_free.py
```diff
@@ -0,0 +1,98 @@
...
+     u32 gfp_flags;
+     char command[30];
+};
+
+/* Remember : When using tracepoints we need to make use of /sys/kernel/debug/tracing/events directory.
+
+For example : the parameters of the function kmem_cache_alloc are defined in /sys/kernel/debug/tracing/events/kmem/kmem_cache_alloc/format
+ */
+
+
```
yonghong-song: extra empty line.

examples/tracing/kernel_cache_alloc_free.py
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python
+#
+# kernel_cache_alloc_free.py	Kernel memory (alloc-free) monitoring tool.
+#
```
yonghong-song: extra characters after `#`?

examples/tracing/kernel_cache_alloc_free.py
```diff
@@ -0,0 +1,98 @@
...
+	{
+	      struct memory_allocations memoryAllocationsInstance = {};
+	      memoryAllocationsInstance.timestamp = bpf_ktime_get_ns();
+	      memoryAllocationsInstance.call_site = args->call_site;
+	      memoryAllocationsInstance.ptr = (u64)args->ptr;
+	      memoryAllocationsInstance.bytes_req = args->bytes_req;
+	      memoryAllocationsInstance.bytes_alloc = args->bytes_alloc;
+	      memoryAllocationsInstance.gfp_flags = args->gfp_flags;
+          sprintf(memoryAllocationsInstance.command,"%s", "kmem_cache_alloc");
+    	  events_alloc.perf_submit(args, &memoryAllocationsInstance, sizeof(memoryAllocationsInstance));
```
yonghong-song: code alignment issue in the above?

examples/tracing/kernel_cache_alloc_free.py
```diff
@@ -0,0 +1,98 @@
...
+	      return 0;
```
yonghong-song: extra character after `return 0;`

examples/tracing/kernel_cache_alloc_free.py
```diff
@@ -0,0 +1,98 @@
...
+	      memoryFreeInstance.timestamp = bpf_ktime_get_ns();
+	      memoryFreeInstance.call_site = args->call_site;
+	      memoryFreeInstance.ptr = (u64)args->ptr;
+	      memoryFreeInstance.bytes_req = 0;
+	      memoryFreeInstance.bytes_alloc = 0;
+	      memoryFreeInstance.gfp_flags = 0;
+	      sprintf(memoryFreeInstance.command,"%s", "kmem_cache_free");
+
+	      events_free.perf_submit(args, &memoryFreeInstance, sizeof(memoryFreeInstance));
+
```
yonghong-song: extra character at above empty line and the line is not really needed.

examples/tracing/kernel_cache_alloc_free.py
```diff
@@ -0,0 +1,98 @@
...
+"""
+
+# Loads eBPF program
+b = BPF(text=prog)
+
+
+
+
+class Data(ct.Structure):
+    _fields_ = [("timestamp", ct.c_ulonglong), ("call_site", ct.c_ulonglong),("ptr", ct.c_ulonglong),("bytes_req", ct.c_ulonglong),("bytes_alloc", ct.c_ulonglong),("gfp_flags", ct.c_ulonglong), ("command", ct.c_char * 30)]
```
yonghong-song: The `("gfp_flags", ct.c_ulonglong)` is not correct. This will cause python layout different C layout and hence incorrect results will be printed out.

jugurthab: Thank you for your comments, they were really beneficial, I have corrected the issues and pushed the new changes.

examples/tracing/kernel_cache_alloc_free.py
```diff
@@ -0,0 +1,80 @@
...
+    events_free.perf_submit(args, &memoryFreeInstance, sizeof(memoryFreeInstance));
+    return 0;
+}
+"""
+
+# Loads eBPF program
+b = BPF(text=prog)
+
+class Data(ct.Structure):
+    _fields_ = [("timestamp", ct.c_ulonglong), ("call_site", ct.c_ulonglong),("ptr", ct.c_ulonglong),("bytes_req", ct.c_ulonglong),("bytes_alloc", ct.c_ulonglong),("gfp_flags", ct.c_uint), ("command", ct.c_char * 30)]
```
yonghong-song: line too long, could you have shorter multiple lines?

examples/tracing/kernel_cache_alloc_free.py
```diff
@@ -0,0 +1,80 @@
...
+
+# Show message when ePBF stats
+print("Kernel cache allocation started ... Hit Ctrl-C to end!")
+
+print("%-18s %-22s %-24s %-15s %-12s %-32s %-30s" % ("TIME(S)", "Call_site", "ptr", "bytes_req", "bytes_alloc", "gfp_flags", "Command"))
```
yonghong-song: multiple lines?

examples/tracing/kernel_cache_alloc_free.py
```diff
@@ -0,0 +1,80 @@
...
+
+def display_kernel_alloc_free(cpu, data, size):
+    evenement = ct.cast(data, ct.POINTER(Data)).contents
+    print("%-18.0f %-22.0f 0x%-22.0f %-15.0f %-12.0f 0x%-30d %-29s" % (evenement.timestamp, evenement.call_site, evenement.ptr, evenement.bytes_req, evenement.bytes_alloc, evenement.gfp_flags, evenement.command))
```
yonghong-song: Multiple lines?
The print format is weird. Could you try the kernel tracepoint and output in a similar way? This is also a good way to verify your example does work as expected.

examples/tracing/kernel_cache_alloc_free_example.txt
```diff
@@ -0,0 +1,23 @@
+Demonstrations of kernel_cache_alloc_free.py, the Linux eBPF/bcc version.
+
+This tracks kmem_cache_alloc and kmem_cache_free routines (using tracepoints).
+
+
+# ./kernel_cache_alloc_free
+Kernel cache allocation started ... Hit Ctrl-C to end!
+TIME(S)            Call_site              ptr                      bytes_req       bytes_alloc  gfp_flags                      Command                       
+8308825716036      18446744072166397952   0x18446629912970690560   4096            4096         0x7882827029131821248          _cache_alloc
```
yonghong-song: This output is incorrect. `_cache_alloc` is printed out instead of `kmem_cache_alloc`. See my previous comments. Please update the output format to be similar to kernel tracepoint output.

examples/tracing/kernel_cache_alloc_free.py
```diff
@@ -0,0 +1,80 @@
...
+# kernel_cache_alloc_free.py - Kernel memory (alloc-free) monitoring tool.
+#
+# Written as a basic tool example of using ePBF
+# to track of kernel cache allocation and deallocation process.
+#
+# Copyright (c) 2019 Jugurtha BELKALEM.
+# Licensed under the Apache License, Version 2.0 (the "License")
+#
+# 14-Jan-2019 Jugurtha BELKALEM Created this.
+
```
yonghong-song: Please add additional comments like:
This program shows to dump trace point kmem/kmem_cache_alloc and kmem/kmem_cache_free results using bcc python APIs. The same/similar result can be achieved though directly use kernel debugfs system.

examples/tracing/kernel_cache_alloc_free.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python
+#
+# kernel_cache_alloc_free.py - Kernel memory (alloc-free) monitoring tool.
```
yonghong-song: Please do a `chmod +x kernel_cache_alloc_free.py` on your file.

brendangregg:
I know this is an /examples and not a /tool, but I'd still be worried about random people running it as though it was a tool -- then causing overhead due to tracing the kmem functions. I'd put a warning in the header of the tool to warn about the overhead of tracing frequent functions like this.

Yikes -- this uses BPF_PERF_OUTPUT? I was assuming it was a summary tool. Have you tested on a busy system? I worry this is an example that could cause harm if people run it. I'm reading this correctly, right, you're basically doing a printf() on kmem alloc and free?? I would never do this without trying a hundred other ways to solve the problem first.

PR #2143, opened 2019-01-16
Add udp_rcvbuf_errors.py to tools to help debug UDP RcvbufErrors packet drops

th0m:
`udp_rcvbuf_errors` prints information about UDP packets that were dropped by the
kernel due to the receive buffer being full.
This is useful for debugging high rates of UDP drops due to RcvbufErrors and finding which socket is responsible for them in real time.

### DIFF
--- a/tools/udp_rcvbuf_errors.py
+++ b/tools/udp_rcvbuf_errors.py
@@ -1,0 +1,163 @@
+#!/usr/bin/env python
+#
+# udp_rcvbuf_errors.py	UDP RcvbufErrors analysis tool.
+#
+# Prints out information for UDP packets which were dropped
+# a result of the socket receive buffer being full.
+#
+# USAGE: udp_rcvbuf_errors
+
+# Copyright (c) 2019 Cloudflare, Inc.
+# Licensed under the Apache License, Version 2.0 (the "License")
+#
+# 15-Jan-2019 Thomas Lefebvre Created this.
+
+from bcc import BPF
+import ctypes as ct
+import argparse
+from time import strftime
+from socket import inet_ntop, AF_INET, AF_INET6
+from struct import pack
+
+examples = """examples:
+    ./udp_rcvbuf_errors           # trace UDP RcvbufErrors kernel drops
+"""
+parser = argparse.ArgumentParser(
+    description="Trace UDP RcvbufErrors drops by the kernel",
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    epilog=examples)
+args = parser.parse_args()
+debug = 0
+
+# define BPF program
+bpf_text = """
+#include <uapi/linux/udp.h>
+#include <uapi/linux/ip.h>
+#include <net/sock.h>
+
+struct ipv4_data_t {
+    u32 saddr;
+    u32 daddr;
+    u16 sport;
+    u16 dport;
+    u32 rmem_alloc;
+    u32 rcvbuf;
+};
+BPF_PERF_OUTPUT(ipv4_events);
+
+struct ipv6_data_t {
+    unsigned __int128 saddr;
+    unsigned __int128 daddr;
+    u16 sport;
+    u16 dport;
+    u32 rmem_alloc;
+    u32 rcvbuf;
+};
+BPF_PERF_OUTPUT(ipv6_events);
+
+static struct udphdr *skb_to_udphdr(const struct sk_buff *skb)
+{
+    // unstable API. verify logic in udp_hdr() -> skb_transport_header().
+    return (struct udphdr *)(skb->head + skb->transport_header);
+}
+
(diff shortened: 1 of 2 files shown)


yonghong-song:
[buildbot, test this please]

tools/udp_rcvbuf_errors.py
```diff
@@ -0,0 +1,163 @@
+#!/usr/bin/env python
```
yonghong-song: could you change the shebang to `#!/usr/bin/python`?

tools/udp_rcvbuf_errors.py
```diff
@@ -0,0 +1,163 @@
...
+
+static inline struct iphdr *skb_to_iphdr(const struct sk_buff *skb)
+{
+    // unstable API. verify logic in ip_hdr() -> skb_network_header().
+    return (struct iphdr *)(skb->head + skb->network_header);
+}
+
+int kprobe__udp_queue_rcv_skb(struct pt_regs *ctx, struct sock *sk, struct sk_buff *skb) {
+    if (sk->sk_rmem_alloc.counter <= sk->sk_rcvbuf)
+    {
```
yonghong-song: put `{` in previous linee?
Also, could you give a little more explanation how do you come up with the above condition? Any caveat here (e.g., any possible missed counting?) Some comments will help people understand the approach applied here? Also it seems `udp_queue_rcv_skb` is a static function and it could get inlined. It it does get inlined, the counting will not be accurate. Maybe some warning in comments as well.

tools/udp_rcvbuf_errors_example.txt
```diff
@@ -0,0 +1,37 @@
+Demonstrations of udp_rcvbuf_errors, the Linux eBPF/bcc version.
```
yonghong-song: Could you also add a man page?

yonghong-song:
Somehow, I feel this tool is too narrow to be in bcc/tools directory... Maybe we can change the tool name to be `udperrors.py` or just `udpdrop.py`? Note that we `tcpdrop.py`. Even today you just trace a particular drop reason, maybe the tool can be extended to cover other udp drop scenarios in the future?

brendangregg:
@yonghong-song looks like our messages crossed in flight, but we are in agreement!

dannyk81:
Apologies in advance for piggy-backing on this PR :pray: but I came across it while investigating an illusive issue with UDP, our problem though is on the transmit path for which I'm not able to find any relevant tracepoints to try and narrow it down (so I'm trying to play with the code in this PR to make it work on the send path - not much luck so far...).

We have a Java application that transmits relatively small (100-150 bytes) datagrams in a containerized deployment (Kubernetes), we are testing the same application on a newer kernel (moved from 4.14.96 -> 4.19.28) and hitting send failures for which we can't seem to find any explanation.

The application uses `java.nio.channels.DatagramChannel` to transmit the datagrams (using a non-blocking channel and 4096 bytes `SO_SNDBUF`), we also limit the `ByteBuffer` to 1500 so that we never try to send something that might exceed the send buffer size.

When the load on the system increases (around 2~3k sends per second) the `send` method seem to be failing to send some percentage of the datagrams and returns 0 which (based on the docs) suggests: 
>if this channel is non-blocking, may be zero if there was insufficient room for the datagram in the underlying output buffer

https://docs.oracle.com/javase/7/docs/api/java/nio/channels/DatagramChannel.html#send(java.nio.ByteBuffer,%20java.net.SocketAddress)

for example: 
```Failed to send stat java.nio.HeapByteBuffer[pos=0 lim=1500 cap=1500] to host <host:port>. Only sent 0 bytes out of 107 bytes```

the buffer to be sent has a capacity of 1500 bytes, and current position in the buffer is 0 as expected since we send each buffer independently.

in this case, it tried to transmit 107 bytes and `send` returned 0, this datagram doesn't exceed the `ByteBuffer` limit, nor does it exceed the SO_SNDBUF limit, so the only thing I could think of is that it failed to transmit this datagram somewhere down the line.

Based on everything I understand on how this works, the only reason that a 0 would be returned here is in case the various underlying buffers in the Kernel/NIC are saturated.

We do see increased counter in the container for `UdpSndbufErrors` stat which seem to roughly correlate with the number of application exceptions we see, but I can't seem to find a clear way to trace this more accurately.

We have compared all the possible tuning points between the two systems running the different kernel versions, and as far as we can tell - they are identical, specifically:
```
$ sysctl net.ipv4.udp_mem
net.ipv4.udp_mem = 768285	1024383	1536570
```
which I think the one that plays the biggest role here.

Kernel 4.19 introduced 2 additional namespaced tunables:
```
udp_rmem_min
udp_rmem_min
```
however, these have the same value on both deployments (4096).

@th0m maybe a pointer on how we could use similar approach for the send path?
@brendangregg any ideas or suggestions would be greatly appreciated

I again apologize for jumping on here with our issue.

Issue #2014, opened 2018-10-16
Support symbol resolution for short-lived processes


sandip4n:
In some cases, symbols in stack traces are not resolved at all. This typically happens when the program being traced exits quickly. This can be observed using the `trace` script as shown below.

```
trace 'p:c:inet_pton' -U
```
when tracing
```
ping -6 -c 1 ::1
```
produces a stack trace like
```
PID     TID     COMM            FUNC             
27444   27444   ping            inet_pton        
        [unknown]
        [unknown]
        [unknown]
```
However, if `ping` keeps running e.g. `ping -6 ::1`, a proper stack trace is seen.
```
PID     TID     COMM            FUNC             
27415   27415   ping            inet_pton        
        __inet_pton+0x0 [libc-2.27.so]
        __GI_getaddrinfo+0x124 [libc-2.27.so]
        [unknown] [ping]
```
This was seen on both `x86_64` and `powerpc64` systems. There are no problems when doing something similar with `perf` though.

sandip4n:
@palmtenor Ah, I see. Thanks for the explanation.
So, can I keep this open for now as more of a feature request?
That is, unless this has been already captured in some other issue.

palmtenor:
Sure, feel free to leave it open. You should change the description to something more generic like "Support symbol resolution on short-lived Processes" or what you find descriptive.

sandip4n:
@palmtenor Done. Thanks.

vijunag:
Part of the problem is that proc entries for the short-lived process disappear. I tried using BUILD_ID in bpf_get_stackid() and it does report the build id. However for a system wide profile, scanning through all the ELF files on the box to match the build-id is too much of a overkill. The patch that added this support here (http://patchwork.ozlabs.org/cover/885955/ ) clearly says the user-space applications will have to maintain mapping between build-id and the corresponding ELF file so that /proc/XXX/smaps doesn't have to be looked up every single time. So, there are two parts to this problem 
1) We definitely need BUILD_ID support so that we don't have to scan through VMAs from smaps
2) The tool has to separately maintain BUILD_ID to ELF mappings. 
I have a patch that has support for BUILD_ID for stackmap and will send it shortly, but if only BPF reported file name instead of BUILD_ID, this problem could've been solved very easily.

yonghong-song:
The idea for the build id is for sysadmin or some super user to build such a database so everybody else can use it to query. If a binary is added by user, the build_id->filename mapping can be
[...]

PR #2144, opened 2019-01-16
support symbol resolution of short-lived process. 

vijunag:
Support symbol resolution for short-lived processes (#2014)
New command line options have been added to tools/trace.py to support the new BUILD_ID stackmap. List of symbol files can be added to the script to resolve symbols from build id as reported by the kernel in the stack trace. A comma separated list of symbol files can be passed to the script as shown below. 

Eg invocation to trace inet_pton(as mentioned in #2014)
python tools/trace.py -s /lib/x86_64-linux-gnu/libc.so.6,/bin/ping  'p:c:inet_pton' -U

### DIFF
--- a/tools/trace.py
+++ b/tools/trace.py
@@ -4,7 +4,7 @@
 #               parameters, with an optional filter.
 #
 # usage: trace [-h] [-p PID] [-L TID] [-v] [-Z STRING_SIZE] [-S]
-#              [-M MAX_EVENTS] [-T] [-t] [-K] [-U] [-a] [-I header]
+#              [-M MAX_EVENTS] [-s SYMBOLFILES] [-T] [-t] [-K] [-U] [-a] [-I header]
 #              probe [probe ...]
 #
 # Licensed under the Apache License, Version 2.0 (the "License")
@@ -35,6 +35,7 @@ class Probe(object):
         tgid = -1
         pid = -1
         page_cnt = None
+        build_id_enabled = False
 
         @classmethod
         def configure(cls, args):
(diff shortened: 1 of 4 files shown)


yonghong-song:
@brendangregg Now bcc core has stackmap buildid support. This patch tries to amend `trace.py` for short lived processes. What do you think for `trace.py`? What else in bcc we may utilize buildid support? `profile.py`?

vijunag:
> Since a new flag is introduced, the example file and man page needs changes as well.

Updated the man file.

yonghong-song:
[buildbot, test this please]

tools/trace.py
```diff
@@ -693,6 +697,8 @@ def __init__(self):
                   help="print CPU id")
                 parser.add_argument("-B", "--bin_cmp", action="store_true",
                   help="allow to use STRCMP with binary values")
+                parser.add_argument('-s', "--sym_file_list", type=str, metavar="SYM_FILE_LIST",
```
yonghong-song: Maybe break the line so that it will remain  maximum 80 characters per line?

tools/trace.py
```diff
@@ -693,6 +697,8 @@ def __init__(self):
...
+                  dest="sym_file_list", help="coma separated list of symbol files to use for symbol resolution")
```
yonghong-song: break into two lines?

vijunag:
Please note that test bed failures are not due to this change.

yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks. The change looks good to me.

Issue #2145, opened 2019-01-16
bpflist broken after use of #!/usr/bin/env


brendangregg:
After #2102 replaced the interpreter line with #!/usr/bin/env, we can no longer see the tool names in bpflist:

```
~/Build/bcc/tools/bpflist.py -vv
open kprobes:
p:kprobes/p_do_sys_open_bcc_30516 do_sys_open
r:kprobes/r_do_sys_open_bcc_30516 do_sys_open

open uprobes:

PID    COMM             TYPE     COUNT
1      systemd          prog     6   
1      systemd          map      6   
30516  python           map      2   
30516  python           kprobe   2   
30516  python           prog     2   
```

It used to look like this:

```
bpflist.py -vv
open kprobes:
p:kprobes/p_do_sys_open_bcc_30542 do_sys_open
r:kprobes/r_do_sys_open_bcc_30542 do_sys_open

open uprobes:

PID    COMM             TYPE     COUNT
1      systemd          map      6   
1      systemd          prog     6   
30542  opensnoop        kprobe   2   
30542  opensnoop        map      2   
30542  opensnoop        prog     2   
```

This particular issue shouldn't be too hard to fix, but I'm concern about others, and I've discussed env problems in the past. I'm happy to go back to #!/usr/bin/python and have people fix their production systems in a different way.

@SaveTheRbtz

yonghong-song:
@brendangregg thanks for bringing this up. @SaveTheRbtz any comments on this? Having the default /usr/bin/python and provide a way to a customer shebang in bcc?

SaveTheRbtz:
Let me work on this later today.  I definitely understand the pain it brings to the operator when everything is called `python`.

Let's then standardize on `#!/usr/bin/python` as long as it is the same everywhere this should not cause problems on the packaging side. I'll also go over fedora[1] and debian[2] packaging guidelines and see what is the best way of fixing this on that side. Quickly glancing over them -- they are also based on the shebang mangling.

[0] https://fedoraproject.org/wiki/Packaging:Python
[1] https://www.debian.org/doc/packaging-manuals/python-policy/ch-programs.html

PR #2146, opened 2019-01-16
Unify shebangs under `#!/usr/bin/python`

SaveTheRbtz:
This is a partial rollback of #2102.

Previously we've consolidate shebangs on `#!/usr/bin/env python`, which brought some unintended consequences in #2145 like lack of observability, since all proctitles are all `python` now.

Now all shebangs are unified as `#!/usr/bin/python` and it's up to package maintainers to mangle shebangs:
* SPEC is doing it's own explicit shebang rewrite with `sed`.
  * newer Fedora has `pathfix.py`.
* `dh_python2`/`dh_python3` also fixes shebangs by default.

Fixes: #2145

### DIFF
--- a/SPECS/bcc.spec
+++ b/SPECS/bcc.spec
@@ -88,6 +88,10 @@ popd
 %install
 pushd build
 make install/strip DESTDIR=%{buildroot}
+# mangle shebangs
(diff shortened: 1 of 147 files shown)


yonghong-song:
[buildbot, test this please]

PR #2147, opened 2019-01-16
Add free_bcc_memory to BPFModule

palmtenor:
Some users uses `BPFModule` directly instead of C++ / Python API, and they would like to have similar interface of free BCC `.text` memory

### DIFF
--- a/src/cc/bpf_module.cc
+++ b/src/cc/bpf_module.cc
@@ -35,6 +35,7 @@
 
 #include "common.h"
 #include "bcc_debug.h"
+#include "bcc_elf.h"
 #include "frontends/b/loader.h"
 #include "frontends/clang/loader.h"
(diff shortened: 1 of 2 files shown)


yonghong-song:
Compilation error:
```
/tmp/rpmbuild.gB0owo/BUILD/bcc/src/cc/bpf_module.cc: In member function 'int ebpf::BPFModule::free_bcc_memory()':
/tmp/rpmbuild.gB0owo/BUILD/bcc/src/cc/bpf_module.cc:145:26: error: 'bcc_free_memory' was not declared in this scope
   return bcc_free_memory();
```

palmtenor:
Ah, sorry for that. The header is slightly different in our build environment, forgot to include in public build:)

palmtenor:
[buildbot, test this please]

PR #2151, opened 2019-01-17
Fix process map parsing when freeing bcc memory

sandip4n:
This fixes the format string used to parse the major and minor device fields in `/proc/self/maps`. These fields have hexadecimal values and hence cannot be parsed as unsigned integers.

Fixes: 51480d05 ("implement free_bcc_memory() API (#2097)")


yonghong-song:
Oh, it is my mistake. Thanks for the fix!

yonghong-song:
[buildbot, test this please]

sandip4n:
> Oh, it is my mistake. Thanks for the fix!

No problem. Happy to help :)

PR #2155, opened 2019-01-20
print_log2_hist(): check and skip possible paddings

boat0:
When a struct S is used as key to a BPF_HISTOGRAM, it is assumed that the second
    member of S holds the slot. But when S is converted to python from bpf C,
    a padding may be inserted as a second member. This breaks print_log2_hist().

        root@debian:~/bcc/tools# ./softirqs.py -d
        ^C
          File "./softirqs.py", line 144, in <module>
            dist.print_log2_hist(label, "softirq", section_print_fn=vec_to_name)
          File "/usr/local/lib/python2.7/dist-packages/bcc/table.py", line 326, in print_log2_hist
            vals[slot] = v.value
        TypeError: list indices must be integers, not str

    Fix it by skipping the possible padding. Future work would be fixing/working
    around in the library where the padding is introduced.

### DIFF
--- a/src/python/bcc/table.py
+++ b/src/python/bcc/table.py
@@ -317,6 +317,15 @@ class TableBase(MutableMapping):
             tmp = {}
             f1 = self.Key._fields_[0][0]
             f2 = self.Key._fields_[1][0]
+
+            # The above code assumes that self.Key._fields_[1][0] holds the
+            # slot. But a padding member may have been inserted here, which
+            # breaks the assumption and leads to chaos.
+            # TODO: this is a quick fix. Fixing/working around in the BCC
+            # internal library is the right thing to do.
+            if f2 == '__pad_1' and len(self.Key._fields_) == 3:
+                f2 = self.Key._fields_[2][0]
+
             for k, v in self.items():
                 bucket = getattr(k, f1)
                 if bucket_fn:


yonghong-song:
[buildbot, test this please]

jeromemarchand:
Shouldn't be this issue still open given the patch is explicitely a temporary one?
This patch seems very unreliable to me. I'm not going to dive into C standards, but  for instance, what guaranty do we have that the padding field is always going to be called '__pad_1'?

Before we come to a proper solution, the format of the key structure should be clearly defined. I couldn't find anything about that in the reference guide. For instance, if the slot field was to be always named 'slot', which seems to be already the case in the tools and examples, then print_log2_hist() could just look for that field.

jeromemarchand:
> Fix it by skipping the possible padding. Future work would be fixing/working
> around in the library where the padding is introduced.

The padding is added by the compiler for alignment reason. There is nothing to fix there.

boat0:
> This patch seems very unreliable to me. I'm not going to dive into C standards, but for instance, what guaranty do we have that the padding field is always going to be called '__pad_1'?

https://github.com/iovisor/bcc/blob/51d62d36bd072530a238ac147a61b631fdc44659/src/cc/json_map_decl_visitor.cc#L140

> Before we come to a proper solution, the format of the key structure should be clearly defined. I couldn't find anything about that in the reference guide. 

I believe help to complete the docs is always welcome :)

boat0:
> > Fix it by skipping the possible padding. Future work would be fixing/working
> > around in the library where the padding is introduced.
> 
> The padding is added by the compiler for alignment reason. There is nothing to fix there.

As a matter of fact, the padding is added by libbcc, the C part of BCC itself.

yonghong-song:
I agree. Do you want to add some documentation about what C=>python key/value structure padding? Although most people should not care about this, but such information is still useful if users reply on particular fields and padding may change their relative position in the structure.

jeromemarchand:
> As a matter of fact, the padding is added by libbcc, the C part of BCC itself.

OK, my bad!


> I agree. Do you want to add some documentation about what C=>python key/value structure padding? Although most people should not care about this, but such information is still useful if users reply on particular fields and padding may change their relative position in the structure.

That might be useful too, but I was thinking about documenting the format of the key_type of an histogram when said type is a structure.

yonghong-song:
> That might be useful too, but I was thinking about documenting the format of the key_type of an histogram when said type is a structure.

This is definitely needed. Thanks!

PR #2156, opened 2019-01-21
use libbpf APIs in libbcc

pbhole:
Recently kernel libbpf was imported in bcc. It makes some of the
libbcc code redundant. This patch modifies libbcc functions to use
libbpf APIs.

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -227,54 +227,29 @@ int bcc_create_map(enum bpf_map_type map_type, const char *name,
 
 int bpf_update_elem(int fd, void *key, void *value, unsigned long long flags)
 {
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
> I am planning to do the same with bpf_attach_xdp(). But while we remove redundant code, some error logging code will be removed. Similar error checking isn't present in libbpf's bpf_set_link_xdp_fd(). Let me know if that is ok

Thanks for looking into this. Uing libbpf's bpf_set_link_xdp_fd() is a good idea. I think printing out the return value associated string should be enough. Just to be sure the return value may be an error code like `-LIBBPF_ERRNO__...`.

PR #2164, opened 2019-01-24
fix cpuunclaimed.py with cfs_rq structure change

yonghong-song:
Similar to runqlen.py, make proper adjustment for
cfs_rq_partial structure so it can align with
what the kernel expects.


brendangregg:
Thanks, works for me!

We really need BTF for this in the future, right?

yonghong-song:
Right. With BTF which is able to resolve to correct offset, we do not need to define `cfs_rq_partial`...

yonghong-song:
[buildbot, test this please]

PR #2166, opened 2019-01-27
profile: exclude CPU idle stacks by default

brendangregg:
I should have excluded these a long time ago. -I exists if people want them back. They are excluded by simply checking PID == 0. I can also excluding them by digging idle status from the task_struct, but that involves adding a bpf_get_current_task(), and we already have the pid there, so just use the pid.


yonghong-song:
Thanks!

PR #2174, opened 2019-02-01
Fix debuginfo search on Ubuntu

antonblanchard:
bcc on Ubuntu picks up the glibc binary instead of the debuginfo file
because find_debug_via_debuglink() doesn't handle symbolic or hard
links when checking if two paths point to the same file.

Add a helper, same_file() which uses the device and inode to check if
the two paths do point to the same file.

Signed-off-by: Anton Blanchard <anton@ozlabs.org>

### DIFF
--- a/src/cc/bcc_elf.c
+++ b/src/cc/bcc_elf.c
@@ -453,6 +453,21 @@ static int verify_checksum(const char *file, unsigned int crc) {
   return actual == crc;
 }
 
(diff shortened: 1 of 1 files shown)


antonblanchard:
The specific issue on Ubuntu is with a soft link, so the other option is to use realpath()

yonghong-song:
[buildbot, test this please]

PR #2180, opened 2019-02-02
Fix tools/syscount -l

Trundle:
In #2063, syscount's syscall mapping was moved into its own module.
Unfortunately, that broke the "print list of recognized syscalls and
exit" usage of syscount. Fix it by importing the syscall mapping from
the new module.


yonghong-song:
[buildbot, test this please]

PR #2185, opened 2019-02-05
Fix misc file permissions

jeromemarchand:
A few files that are obviously not executable have the exec bit set.
Most those files were created with faulty mode to begin with, but
src/cc/export/helpers.h and src/cc/frontends/clang/loader.cc have had
their exec bit (unintendedly I assume) set by commit bfecc243fc8e
("clang: loader: Allow user to override kernel version (#1895)").


palmtenor:
[buildbot, test this please]

yonghong-song:
LGTM. Thanks for the fix!

PR #2191, opened 2019-02-09
add btf func_info and line_info support

yonghong-song:
The libbpf API functions are used to parse/load
    .BTF and .BTF.ext ELF sections, and retrieve
    func_info and line_info for prog loading.
    
    The LLVM cannot get remapped file source for
    line_info. So a postprocessing after compilation
    fixed the line_off (offset to the string table
    for the line_info) in .BTF.ext and
    adjusted string section in .BTF to accommodate
    new source lines.
    
    For users using python and C++ API, no changes
    to their codes. If .BTF and .BTF.ext are supported
    in both compiler and kernel, they will automatically
    get benefit, e.g., verifier log annotated with source
    code, bpftool showing source annotated jit codes, etc.

    For example, with latest bpf-next and llvm trunk,
    running `tcpconnect.py`, bpftool (from kernel) is able
    to show the following:
```
      -bash-4.4$ sudo bpftool p d jited id 165
      int trace_connect_v6_return(struct pt_regs * ctx):
      bpf_prog_8937c3924bd93e28_trace_connect_v6_return:
      ; int trace_connect_v6_return(struct pt_regs *ctx)
         0:   push   %rbp
         1:   mov    %rsp,%rbp
         4:   sub    $0x88,%rsp
         b:   sub    $0x28,%rbp
         f:   mov    %rbx,0x0(%rbp)
        13:   mov    %r13,0x8(%rbp)
        17:   mov    %r14,0x10(%rbp)
        1b:   mov    %r15,0x18(%rbp)
        1f:   xor    %eax,%eax
        21:   mov    %rax,0x20(%rbp)
        25:   mov    %rdi,%rbx
      ; int ret = PT_REGS_RC(ctx);
        28:   mov    0x50(%rbx),%r13
      ; u32 pid = bpf_get_current_pid_tgid();
        2c:   callq  0xffffffffe0e4ebfe
      ; u32 pid = bpf_get_current_pid_tgid();
        31:   mov    %eax,-0x4(%rbp)
      ...
      -bash-4.4$ sudo bpftool p d xlated id 165
      int trace_connect_v6_return(struct pt_regs * ctx):
      ; int trace_connect_v6_return(struct pt_regs *ctx)
         0: (bf) r6 = r1
      ; int ret = PT_REGS_RC(ctx);
         1: (79) r7 = *(u64 *)(r6 +80)
      ; u32 pid = bpf_get_current_pid_tgid();
         2: (85) call bpf_get_current_pid_tgid#76208
      ; u32 pid = bpf_get_current_pid_tgid();
         3: (63) *(u32 *)(r10 -4) = r0
      ; skpp = bpf_map_lookup_elem((void *)bpf_pseudo_fd(1, 3), &pid);
         4: (18) r1 = map[id:298]
         6: (bf) r2 = r10
```

### DIFF
--- a/src/cc/bcc_btf.h
+++ b/src/cc/bcc_btf.h
@@ -1,0 +1,70 @@
+/*
+ * Copyright (c) 2019 Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef BCC_BTF_H
+#define BCC_BTF_H
+
+#include <stdbool.h>
+#include <stdint.h>
+#include <string>
+#include <map>
+#include <vector>
+
+struct btf;
+struct btf_ext;
+
+namespace ebpf {
+
+class BTFStringTable {
+ private:
+  uint32_t Size;
+  std::map<uint32_t, uint32_t> OffsetToIdMap;
+  std::vector<std::string> Table;
+
+ public:
+  BTFStringTable(): Size(0) {}
+  uint32_t getSize() { return Size; }
+  std::vector<std::string> &getTable() { return Table; }
+  uint32_t addString(std::string Str);
+};
+
+class BTF {
+ public:
+  ~BTF();
+  int load(uint8_t *btf_sec, uintptr_t btf_sec_size,
+           uint8_t *btf_ext_sec, uintptr_t btf_ext_sec_size,
+           std::map<std::string, std::string> &remapped_sources);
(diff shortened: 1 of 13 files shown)


src/cc/bcc_btf.h
```diff
@@ -0,0 +1,70 @@
...
+                   unsigned *linfo_rec_size);
+
+ private:
+  void adjust(uint8_t *btf_sec, uintptr_t btf_sec_size,
+              uint8_t *btf_ext_sec, uintptr_t btf_ext_sec_size,
+              std::map<std::string, std::string> &remapped_sources,
+              uint8_t **new_btf_sec, uintptr_t *new_btf_sec_size);
+
+ private:
+  struct btf *btf_;
```
4ast: I think would have been cleaner to create pure c++ wrapper for btf_* where btf__new() would have been a constructor and btf__free() called in destructor. Another pure c++ wrapper for btf_ext__*(). Than this class would be called BCC_BTF, since it has this 'adjust()' workaround and private members at this line would be unique_ptr<BTF> and unique_ptr<BTF_EXT>.
May be such c++ wrappers should be in github/libbpf/ repo.
Such refactoring can be done later.
Overall great stuff!

yonghong-song: Agree. I think some extensible C++ wrapper will be good for common use. Implementing in libbpf repo certainly is a great way for reusability. As you suggested, will try to refactor later.

PR #2198, opened 2019-02-12
generate perf event data structure in Python automatically

boat0:
When ring buffers are opened to receive custom perf event, we have
to define the event data structure twice: once in BPF C and once
in Python. It is redundant and error-prone.

This patch implements the automatic generation of Python data structure
from the C declaration, thus making the redundant definition in Python
unnecessary. Example:

    // define output data structure in C
    struct data_t {
        u32 pid;
        u64 ts;
        char comm[TASK_COMM_LEN];
    };
    BPF_PERF_OUTPUT(events);
    ...

Old way:

    # define output data structure in Python
    TASK_COMM_LEN = 16    # linux/sched.h
    class Data(ct.Structure):
        _fields_ = [("pid", ct.c_ulonglong),
                    ("ts", ct.c_ulonglong),
                    ("comm", ct.c_char * TASK_COMM_LEN)]

    def print_event(cpu, data, size):
        event = ct.cast(data, ct.POINTER(Data)).contents
    ...

New way:

    def print_event(cpu, data, size):
        event = b["events"].event(data)
    ...

### DIFF
--- a/src/python/bcc/table.py
+++ b/src/python/bcc/table.py
@@ -18,6 +18,7 @@ from functools import reduce
 import multiprocessing
 import os
 import errno
+import re
 
 from .libbcc import lib, _RAW_CB_TYPE, _LOST_CB_TYPE
 from .perf import Perf
@@ -122,7 +123,7 @@ def _print_linear_hist(vals, val_type):
                       _stars(val, val_max, stars)))
 
 
-def Table(bpf, map_id, map_fd, keytype, leaftype, **kwargs):
+def Table(bpf, map_id, map_fd, keytype, leaftype, name, **kwargs):
     """Table(bpf, map_id, map_fd, keytype, leaftype, **kwargs)
 
     Create a python object out of a reference to a bpf table handle"""
@@ -136,7 +137,7 @@ def Table(bpf, map_id, map_fd, keytype, leaftype, **kwargs):
     elif ttype == BPF_MAP_TYPE_PROG_ARRAY:
         t = ProgArray(bpf, map_id, map_fd, keytype, leaftype)
     elif ttype == BPF_MAP_TYPE_PERF_EVENT_ARRAY:
-        t = PerfEventArray(bpf, map_id, map_fd, keytype, leaftype)
+        t = PerfEventArray(bpf, map_id, map_fd, keytype, leaftype, name)
     elif ttype == BPF_MAP_TYPE_PERCPU_HASH:
         t = PerCpuHash(bpf, map_id, map_fd, keytype, leaftype, **kwargs)
     elif ttype == BPF_MAP_TYPE_PERCPU_ARRAY:
@@ -162,7 +163,7 @@ def Table(bpf, map_id, map_fd, keytype, leaftype, **kwargs):
 
 class TableBase(MutableMapping):
 
-    def __init__(self, bpf, map_id, map_fd, keytype, leaftype):
+    def __init__(self, bpf, map_id, map_fd, keytype, leaftype, name=None):
         self.bpf = bpf
         self.map_id = map_id
         self.map_fd = map_fd
@@ -171,6 +172,7 @@ class TableBase(MutableMapping):
         self.ttype = lib.bpf_table_type_id(self.bpf.module, self.map_id)
         self.flags = lib.bpf_table_flags_id(self.bpf.module, self.map_id)
         self._cbs = {}
+        self._name = name
 
     def key_sprintf(self, key):
         buf = ct.create_string_buffer(ct.sizeof(self.Key) * 8)
@@ -537,6 +539,7 @@ class PerfEventArray(ArrayBase):
     def __init__(self, *args, **kwargs):
         super(PerfEventArray, self).__init__(*args, **kwargs)
         self._open_key_fds = {}
+        self._event_class = None
 
     def __del__(self):
         keys = list(self._open_key_fds.keys())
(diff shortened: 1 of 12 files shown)


yonghong-song:
[buildbot, test this please]

boat0:
It seems that buildbot skips 
https://github.com/iovisor/bcc/pull/2198/commits/60073efe9bcbc143a8560420867baa1f021c018e upon which 
https://github.com/iovisor/bcc/pull/2198/commits/78ea22fe415257502f3c4a0df34551597197f9de is based.

src/python/bcc/table.py
```diff
@@ -559,6 +561,58 @@ def __delitem__(self, key):
...
+                       'u16': ct.c_ushort,
+                       'int': ct.c_int,
+                       'unsigned int': ct.c_uint,
+                       'u32': ct.c_uint,
+                       'long': ct.c_long,
+                       'unsigned long': ct.c_ulong,
+                       'long long': ct.c_longlong,
+                       'unsigned long long': ct.c_ulonglong,
+                       'u64': ct.c_ulonglong,
+                       'void *': ct.c_void_p }
```
yonghong-song: Maybe add s8, s16, s32, s64 as well?
How do we handle `__int128` and `unsigned __int128`? I guess in such cases, the old way is needed where `__int128` is coded as `s64[2]`?

src/python/bcc/table.py
```diff
@@ -559,6 +561,58 @@ def __delitem__(self, key):
...
+        fields = []
+        prog = re.compile(r"(.+) \[([0-9]+)\]$")
+
+        num_fields = lib.bpf_perf_event_fields(self.bpf.module)
+        i = 0
+        while i < num_fields:
+            field_name = lib.bpf_perf_event_field_name(self.bpf.module, i)
+            field_type = lib.bpf_perf_event_field_type(self.bpf.module, i)
```
yonghong-song: Is it possible to have just one libbcc call like `lib.bpf_perf_event_field` to get field name and type in one shot?

yonghong-song:
@boat0 buildbot is okay. I see the error with llvm 3.8.0 and llvm 3.8.1 on  ubuntu.
```
CMakeFiles/usdt-static.dir/usdt_args.cc.o -c /tmp/debuild.DTQhKj/bcc/src/cc/usdt/usdt_args.cc
/tmp/debuild.DTQhKj/bcc/src/cc/frontends/clang/b_frontend_action.cc: In member function ‘bool ebpf::BTypeVisitor::VisitCallExpr(clang::CallExpr*)’:
/tmp/debuild.DTQhKj/bcc/src/cc/frontends/clang/b_frontend_action.cc:818:46: error: ‘const class clang::Type’ has no member named ‘getPointeeOrArrayElementType’; did you mean ‘getPointeeType’?
                                            ->getPointeeOrArrayElementType()->getAsTagDecl();
                                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                              getPointeeType
```
Could you check whether you can find an alternative implementation? It would be good to find an implementation working for all llvm's. But if you cannot, using `#ifdef` is okay.

Also, could you add some tests to test some of your supported types including array types?
This patch can greatly simplify how to write bcc python scripts. Thanks!

src/python/bcc/table.py
boat0: Ok. I can concatenate type and name in libbcc and then split them in Python.

boat0:
> @boat0 buildbot is okay. I see the error with llvm 3.8.0 and llvm 3.8.1 on ubuntu.
> 
> ```
> CMakeFiles/usdt-static.dir/usdt_args.cc.o -c /tmp/debuild.DTQhKj/bcc/src/cc/usdt/usdt_args.cc
> /tmp/debuild.DTQhKj/bcc/src/cc/frontends/clang/b_frontend_action.cc: In member function ‘bool ebpf::BTypeVisitor::VisitCallExpr(clang::CallExpr*)’:
> /tmp/debuild.DTQhKj/bcc/src/cc/frontends/clang/b_frontend_action.cc:818:46: error: ‘const class clang::Type’ has no member named ‘getPointeeOrArrayElementType’; did you mean ‘getPointeeType’?
>                                             ->getPointeeOrArrayElementType()->getAsTagDecl();
>                                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
>                                               getPointeeType
> ```
> Could you check whether you can find an alternative implementation? It would be good to find an implementation working for all llvm's. But if you cannot, using `#ifdef` is okay.

now use  getPointeeType().getTypePtr() instead of getPointeeOrArrayElementType()

src/python/bcc/table.py
boat0: s8/16/32/64 added. All other types not listed in this map will cause script
to stop with message "Type: '%s' not recognized. Please use ctypes manually."
As _int128 is rare (I assume), how about coming to this issue later sometime?

boat0: yes, the old way works great for __int128:

boat0:
> Also, could you add some tests to test some of your supported types including array types?

OK. I will do it in a subsequent PR.
I am using tools/tcpconnect.py as a test and everything is OK.

boat0:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

boat0:
Last build failed at test_usdt2.py in which event type is defined as a simple int type.
It is supposed to do nothing in this case, but the checking turns out to be not working.
In this update I make it working by 

` auto type_arg1 = Call->getArg(1)->IgnoreCasts()->getType().getTypePtr()->getPointeeType().getTypePtr();
 if (type_arg1->isStructureType()) {...}`

yonghong-song:
[buildbot, test this please]

yonghong-song:
The following `__int128` should be the most common usage:
```
                      '__int128'          : (ct.c_longlong * 2),
                       'unsigned __int128' : (ct.c_ulonglong * 2),
```
If users try to map __int128 differently in python, they should try to explicit ct_type definitions in python instead.

pchaigno:
This is great! Thanks a lot @boat0!

[The developer tutorial](https://github.com/iovisor/bcc/blob/master/docs/tutorial_bcc_python_developer.md) would also need to be updated to take this pull request into account.

boat0:
> This is great! Thanks a lot @boat0!
> 
> [The developer tutorial](https://github.com/iovisor/bcc/blob/master/docs/tutorial_bcc_python_developer.md) would also need to be updated to take this pull request into account.

pr submitted #2226. Thanks!

PR #2116, opened 2019-01-08
INSTALL.md: add new Ubuntu bpfcc-tools package name

brendangregg:
Fixes #2085

Issue #2085, opened 2018-12-17
Ubuntu package rename to bpfcc-tools


brendangregg:
The Ubuntu packages may have been renamed, from bcc-tools to bpfcc-tools, etc. I'm making a note here so we don't forget to update the INSTALL.md.

How it used to look:

```
# apt-cache search bcc | grep BCC
libbcc - Shared Library for BPF Compiler Collection (BCC)
python-bcc - Python wrappers for BPF Compiler Collection (BCC)
bcc-tools - Command line tools for BPF Compiler Collection (BCC)
python3-bpfcc - Python 3 wrappers for BPF Compiler Collection (BCC)
libbpfcc - shared library for BPF Compiler Collection (BCC)
```

how it now looks:

```
# apt-cache search bcc | grep BCC
bpfcc-lua - Standalone tool to run BCC tracers written in Lua
bpfcc-tools - tools for BPF Compiler Collection (BCC)
libbpfcc - shared library for BPF Compiler Collection (BCC)
libbpfcc-dev - shared library for BPF Compiler Collection (BCC)
python-bpfcc - Python wrappers for BPF Compiler Collection (BCC)
python3-bpfcc - Python 3 wrappers for BPF Compiler Collection (BCC)
```

PR #2203, opened 2019-02-15
Declare provides+conflicts on Debian/Ubuntu packages

bmoyles:
See  #2085 and #2116 for previous work:

While the Debian/Ubuntu packages have different package names, aside from `bpfcc-tools`/`bcc-tools`, the contents of the packages conflict and cannot be simultaneously installed or mixed with upstream packages.

Further, should one install a package declaring a dependency on `bpfcc-tools`, `libbpfcc`, etc, while the upstream packages are installed, given there's no relationship between the packages, `apt` will gladly pull down and try to install the Debian/Ubuntu packages and explode in a half-configured mess :)

This PR adds a Provides relationship between the iovisor and corresponding Debian/Ubuntu packages so the iovisor packages can satisfy dependencies (PCP is a good example, although they've since adjusted their dependencies to allow `python3-bpfcc | python3-bcc`) and also adds a Conflicts relationship to try to prevent users from mixing the installs.

cc @brendangregg

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -55,7 +55,32 @@ Kernel compile flags can usually be checked by looking at `/proc/config.gz` or
 
 The stable and the nightly packages are built for Ubuntu Xenial (16.04), Ubuntu Artful (17.10) and Ubuntu Bionic (18.04). The steps are very straightforward, no need to upgrade the kernel or compile from source!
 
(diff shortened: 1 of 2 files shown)


yonghong-song:
[buildbot, test this please]

brendangregg:
thanks!

PR #2205, opened 2019-02-15
tools: fix some python3 bytes vs strings issues

jeromemarchand:
It fixes the following errors:

$ execsnoop.py -q
  File "/usr/lib/python3.6/site-packages/bcc/table.py", line 572, in raw_cb_
    callback(cpu, data, size)
  File "tools/execsnoop.py", line 229, in print_event
    for arg in argv[event.pid]
  File "tools/execsnoop.py", line 229, in <listcomp>
    for arg in argv[event.pid]

$ offcputime.py -K -f 5
  File "./tools/offcputime.py", line 298, in <module>
    print("%s %d" % (";".join(line), v.value))

$ offwaketime.py -f 5
  File "./tools/offwaketime.py", line 350, in <module>
    print("%s %d" % (";".join(line), v.value))

### DIFF
--- a/tools/execsnoop.py
+++ b/tools/execsnoop.py
@@ -225,7 +225,7 @@ def print_event(cpu, data, size):
             skip = True
         if args.quote:
             argv[event.pid] = [
-                "\"" + arg.replace("\"", "\\\"") + "\""
+                b"\"" + arg.replace(b"\"", b"\\\"") + b"\""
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks for the fix!

PR #2212, opened 2019-02-16
Fix #1851 for Arch Linux users

### LINKED ISSUE #1851: `bashreadline` requires statically linked `libreadline`
I suspect most Linux distributions use a dynamic linked `bash/readline`:

```
juergen@samson:~/cxx/bcc → ldd /usr/bin/bash
	linux-vdso.so.1 (0x00007ffdefa81000)
	libreadline.so.7 => /usr/lib/libreadline.so.7 (0x00007f4059494000)
	libdl.so.2 => /usr/lib/libdl.so.2 (0x00007f4059290000)
	libc.so.6 => /usr/lib/libc.so.6 (0x00007f4058ed4000)
	libncursesw.so.6 => /usr/lib/libncursesw.so.6 (0x00007f4058c67000)
	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007f40599c2000)
```

Can't resolve symbol:
```
juergen@samson:~ → sudo /usr/share/bcc/tools/bashreadline 
  File "/usr/share/bcc/tools/bashreadline", line 52, in <module>
    b.attach_uretprobe(name="/bin/bash", sym="readline", fn_name="printret")
  File "/usr/lib/python3.6/site-packages/bcc/__init__.py", line 957, in attach_uretprobe
    (path, addr) = BPF._check_path_symbol(name, sym, addr, pid)
  File "/usr/lib/python3.6/site-packages/bcc/__init__.py", line 671, in _check_path_symbol
    raise Exception("could not determine address of symbol %s" % symname)
Exception: could not determine address of symbol b'readline'
```

Should `bashreadline` check for dynamic linking and attach the probe to symbol in `libreadline.so`?

JayceCao:
The fix for **Arch Linux** came from #1851. 

Have tested that the script can run on `Ubuntu 18.04.2` without any modification.
Haven't tested for other distributions.

### DIFF
--- a/tools/bashreadline.py
+++ b/tools/bashreadline.py
@@ -18,9 +18,12 @@ from time import strftime
 # load BPF program
 bpf_text = """
 #include <uapi/linux/ptrace.h>
+// include <linux/sched.h>
 
 struct str_t {
     u64 pid;
+ // If you use Arch Linux, you may need this field to filter info
+ // char comm[TASK_COMM_LEN];
     char str[80];
 };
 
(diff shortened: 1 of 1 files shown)


yonghong-song:
Maybe implement this alternative as an option `-s` will be better? `-s` is for shared library. This way, user does not need to change anything and just run on Arch Linux.

PR #2213, opened 2019-02-17
Fix BCC on arm64 by allowing missing ausyscall

joelagnel:
ausyscall is not available on ARM64 systems I am testing. The system is
running debian buster. All BCC tools fail as a result. Let us not fail
if ausyscall is missing.


joelagnel:
CC @wcohen

yonghong-song:
[buildbot, test this please]

PR #2217, opened 2019-02-18
fix a bug related to syscall.py

yonghong-song:
Commit 218f7482f8ae refactored syscall number=>name
mapping into a separate file src/python/bcc/syscall.py.
The commit added a reference in python __init__.py,
and this will cause virtually all non x64 arch python
tools failure.

Commit bc0d472ec9d7 attempted to fix the issue by
removing the failure in syscall.py but this is not
the correct fix for arm64 as the syscall numbers
won't match.

Removing the syscall.py reference in __init__.py
should be enough to restore the previous
working behavior.

Fixes: bc0d472ec9d7 ("Fix BCC on arm64 by allowing missing ausyscall")
Fixes: 218f7482f8ae ("Wcohen/efficiency (#2063)")


joelagnel:
LGTM, thanks!

PR #2204, opened 2019-02-15
tools: remove redundant Python event data structure definitions

boat0:
Simplify code following #2198 (https://github.com/iovisor/bcc/pull/2198).

Some tools are not touched: mountsnoop.py, trace.py, lib/*.py, old/*.py.

PR #2220, opened 2019-02-19
add a new bcc tool drsnoop

ethercflow:
Trace direct reclaim and print details including issuing PID.

### DIFF
--- a/tools/drsnoop.py
+++ b/tools/drsnoop.py
@@ -16,7 +16,6 @@
 from __future__ import print_function
 from bcc import ArgString, BPF
 import argparse
-import ctypes as ct
 from datetime import datetime, timedelta
 
 # arguments
@@ -156,20 +155,6 @@ b.attach_tracepoint(tp="vmscan:mm_vmscan_direct_reclaim_begin",
 b.attach_tracepoint(tp="vmscan:mm_vmscan_direct_reclaim_end",
                     fn_name="trace_direct_reclaim_end")
 
-TASK_COMM_LEN = 16  # linux/sched.h
-
-
-class Data(ct.Structure):
-    _fields_ = [
-        ("id", ct.c_ulonglong),
-        ("uid", ct.c_uint32),
-        ("nr_reclaimed", ct.c_ulonglong),
-        ("delta", ct.c_ulonglong),
-        ("ts", ct.c_ulonglong),
-        ("name", ct.c_char * TASK_COMM_LEN)
-    ]
-
-
 initial_ts = 0
 
 # header
@@ -184,7 +169,7 @@ print("%-14s %-8s %-14s %7s" %
 
 
 def print_event(cpu, data, size):
-    event = ct.cast(data, ct.POINTER(Data)).contents
+    event = b["events"].event(data)
 
     global initial_ts
 
--- a/tools/drsnoop.py
+++ b/tools/drsnoop.py
@@ -18,6 +18,8 @@ from __future__ import print_function
 from bcc import ArgString, BPF
 import argparse
 from datetime import datetime, timedelta
+import os
+import math
 
 # symbols
 kallsyms = "/proc/kallsyms"
@@ -63,7 +65,6 @@ if args.duration:
 
 
 # vm_stat
-
 vm_stat_addr = ''
 with open(kallsyms) as syms:
     for line in syms:
@@ -82,7 +83,8 @@ with open(kallsyms) as syms:
 
 NR_FREE_PAGES = 0
 
-PAGE_SHIFT = 12    # defined in page_types.h
+PAGE_SIZE = os.sysconf("SC_PAGE_SIZE")
+PAGE_SHIFT = int(math.log(PAGE_SIZE) / math.log(2))
 
 def K(x):
     return x << (PAGE_SHIFT - 10)
--- a/tools/drsnoop.py
+++ b/tools/drsnoop.py
@@ -1,0 +1,217 @@
+#!/usr/bin/python
+# @lint-avoid-python-3-compatibility-imports
+#
+# drsnoop  Trace direct reclaim and print details including issuing PID.
+#       For Linux, uses BCC, eBPF.
+#
+# This uses in-kernel eBPF maps to cache process details (PID and comm) by
+# direct reclaim begin, as well as a starting timestamp for calculating
+# latency.
+#
+# Copyright (c) 2019 Ethercflow
+# Licensed under the Apache License, Version 2.0 (the "License")
+#
+# 20-Feb-2019   Ethercflow   Created this.
+
+from __future__ import print_function
+from bcc import ArgString, BPF
+import argparse
+import ctypes as ct
+from datetime import datetime, timedelta
+
+# arguments
+examples = """examples:
+    ./drsnoop           # trace all direct reclaim
+    ./drsnoop -T        # include timestamps
+    ./drsnoop -U        # include UID
+    ./drsnoop -P 181    # only trace PID 181
+    ./drsnoop -t 123    # only trace TID 123
+    ./drsnoop -u 1000   # only trace UID 1000
+    ./drsnoop -d 10     # trace for 10 seconds only
+    ./drsnoop -n main   # only print process names containing "main"
+"""
+parser = argparse.ArgumentParser(
+    description="Trace direct reclaim",
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    epilog=examples)
+parser.add_argument("-T", "--timestamp", action="store_true",
+                    help="include timestamp on output")
+parser.add_argument("-U", "--print-uid", action="store_true",
+                    help="print UID column")
+parser.add_argument("-p", "--pid",
+                    help="trace this PID only")
+parser.add_argument("-t", "--tid",
+                    help="trace this TID only")
+parser.add_argument("-u", "--uid",
+                    help="trace this UID only")
+parser.add_argument("-d", "--duration",
+                    help="total duration of trace in seconds")
+parser.add_argument("-n", "--name",
+                    type=ArgString,
+                    help="only print process names containing this name")
+parser.add_argument("--ebpf", action="store_true",
+                    help=argparse.SUPPRESS)
+args = parser.parse_args()
+debug = 0
+if args.duration:
+    args.duration = timedelta(seconds=int(args.duration))
+
+# load BPF program
+bpf_text = """
+#include <uapi/linux/ptrace.h>
+#include <linux/sched.h>
+
+struct val_t {
+    u64 id;
+    u64 ts; // start time
+    char name[TASK_COMM_LEN];
+};
+
+struct data_t {
+    u64 id;
+    u32 uid;
+    u64 nr_reclaimed;
+    u64 delta;
+    u64 ts;    // end time
+    char name[TASK_COMM_LEN];
+};
+
+BPF_HASH(start, u64, struct val_t);
+BPF_PERF_OUTPUT(events);
+
+int trace_direct_reclaim_begin(void)
+{
+    struct val_t val = {};
+    u64 id = bpf_get_current_pid_tgid();
+    u32 pid = id >> 32; // PID is higher part
+    u32 tid = id;       // Cast and get the lower part
+    u32 uid = bpf_get_current_uid_gid();
+    u64 ts;
+
+    PID_TID_FILTER
+    UID_FILTER
+    if (bpf_get_current_comm(&val.name, sizeof(val.name)) == 0) {
+        val.id = id;
+        val.ts = bpf_ktime_get_ns();
+        start.update(&id, &val);
+    }
+    return 0;
+}
+
+struct reclaimed_args {
+    // from /sys/kernel/debug/tracing/events/vmscan/mm_vmscan_direct_reclaim_-
+    // end/format
+    u64 __unused__;
+    u64 nr_reclaimed;
+};
+
+int trace_direct_reclaim_end(struct reclaimed_args *args)
+{
+    u64 id = bpf_get_current_pid_tgid();
+    struct val_t *valp;
+    struct data_t data = {};
+    u64 ts = bpf_ktime_get_ns();
+
+    valp = start.lookup(&id);
+    if (valp == NULL) {
+        // missed entry
+        return 0;
+    }
+
+    data.delta = ts - valp->ts;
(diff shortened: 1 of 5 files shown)


pchaigno:
Please see [`CONTRIBUTING-SCRIPTS.md`](https://github.com/iovisor/bcc/blob/master/CONTRIBUTING-SCRIPTS.md).

ethercflow:
Thanks a lot. I'll read it and make adjustments as soon as I can.

yonghong-song:
@brendangregg could you take a look?

tools/drsnoop.py
```diff
@@ -0,0 +1,217 @@
...
+b = BPF(text=bpf_text)
+b.attach_tracepoint(tp="vmscan:mm_vmscan_direct_reclaim_begin",
+                    fn_name="trace_direct_reclaim_begin")
+b.attach_tracepoint(tp="vmscan:mm_vmscan_direct_reclaim_end",
+                    fn_name="trace_direct_reclaim_end")
+
+TASK_COMM_LEN = 16  # linux/sched.h
+
+
+class Data(ct.Structure):
```
boat0: Please take a look at #2204 and avoid redundant definition in Python.

brendangregg:
This tool makes sense to me, as I already trace direct reclaim latency to quantify memory saturation. But I wonder if the name won't make sense to others who aren't so familiar with the internals.

```
# ./drsnoop
COMM           PID      LAT(ms)          PAGES
summond        17678    0.196455           143
summond        17669    0.559810           313
summond        17669    0.154161           145
summond        17669    0.277360           237
```

Look off. I don't know if it's in CONTRIBUTING_SCRIPTS.md yet, but the convention I've used for years is:
- input fields: left justified
- result fields: right justified

PAGES and LAT(ms) is the result of the direct reclaim, so I'd made them both right justified, and LAT(ms) last. Also, we'll be adding more fields later with other options, so I'd make the columns narrower. PAGES can be %5d and LAT(ms) could be %8.2f.

It would be nice to have a verbose mode (-v) that added system memory state columns: free memory, at least. It might need to be fetched from the BPF kernel program to reflect the free memory at the time that direct reclaim was used.

Needs a man page.

ethercflow:
Thanks a lot, I'll fix them at the weekend.

ethercflow:
@brendangregg I think we can get men info from [vm_stat](https://elixir.bootlin.com/linux/v3.10.108/source/mm/vmstat.c#L80)(kernel version 3.x)  or [vm_zone_stat](https://elixir.bootlin.com/linux/v4.20.13/source/mm/vmstat.c#L162) (kernel version 4.x), so I tried to get `vm_zone_stat`'s addr with `kallsyms_lookup_name`, but got error:
> LLVM ERROR: Program used external function 'kallsyms_lookup_name' which could not be resolved!


The code is below:
``` python
from bcc import BPF

# This may not work for 4.17 on x64, you need replace kprobe__sys_clone with kprobe____x64_sys_clone
bpf_text='''
#include <linux/kallsyms.h>

int kprobe__sys_clone(void *ctx)
{
    u64 *mem = (void*)kallsyms_lookup_name("vm_zone_stat");
    if (mem != NULL)
        bpf_trace_printk("mem: %p\\n", mem);
    else
        bpf_trace_printk("Hello, World!\\n");
    return 0;
}
'''
BPF(text=bpf_text).trace_print()
```

I make sure I've `kernel-headers` installed, and `tcpdrop.py` can be compiled correctly.  So is there any limit with bcc/ebpf? Is there a place I can get more info about this?


**Sys info**: Ubuntu 18.04
**Kernel info**:  4.15.0-45-generic
**Kernel header**: ./linux-headers-4.15.0-45/include/linux/kallsyms.h:21:unsigned long kallsyms_lookup_name(const char *name);


Thank you.

tools/drsnoop.py
ethercflow: @boat0 Thanks a lot, I've fixed this, PTAL.

ethercflow:
> But I wonder if the name won't make sense to others who aren't so familiar with the internals.

@yonghong-song @brendangregg 

I don't have good ideas right now. What about you?

yonghong-song:
@ethercflow regarding to your latest issue with `kallsyms_lookup_name`. The function is defined in kernel but not a bpf helper so it cannot be used as in bpf program. The kernel global variable vm_zone_stat is not available to bpf either as bpf accesses all kernel functionalities through helpers and `context` argument.
I will also try to look at your `drsnoop` tool soon.

ethercflow:
@yonghong-song Thanks a lot, I'll explore the bpf internal implementation to have a deeper understanding of bpf.

tools/drsnoop.py
```diff
@@ -0,0 +1,200 @@
...
+}
+
+struct reclaimed_args {
+    // from /sys/kernel/debug/tracing/events/vmscan/mm_vmscan_direct_reclaim_-
+    // end/format
+    u64 __unused__;
+    u64 nr_reclaimed;
+};
+
+int trace_direct_reclaim_end(struct reclaimed_args *args)
```
yonghong-song: could you usee `TRACEPOINT_PROBE` instead of explicit func name `trace_direct_reclaim_end`? The structure `reclaimed_args` also not needed.

tools/drsnoop.py
```diff
@@ -0,0 +1,200 @@
...
+    u64 nr_reclaimed;
+    u64 delta;
+    u64 ts;    // end time
+    char name[TASK_COMM_LEN];
+};
+
+BPF_HASH(start, u64, struct val_t);
+BPF_PERF_OUTPUT(events);
+
+int trace_direct_reclaim_begin(void)
```
yonghong-song: even if the argument not used, still using `TRACEPOINT_PROBE`?

tools/drsnoop.py
```diff
@@ -0,0 +1,200 @@
...
+else:
+    bpf_text = bpf_text.replace('UID_FILTER', '')
+if debug or args.ebpf:
+    print(bpf_text)
+    if args.ebpf:
+        exit()
+
+# initialize BPF
+b = BPF(text=bpf_text)
+b.attach_tracepoint(tp="vmscan:mm_vmscan_direct_reclaim_begin",
```
yonghong-song: If you use tracepoint macros, these explicit attach_tracepoint is not needed any more.

yonghong-song:
[buildbot, test this please]

yonghong-song:
@brendangregg suggested to print out `free` memory in verbose mode. Looks like not easy. One way is to check whether you could reach from `current` task structure.

yonghong-song:
the ubuntu 16.04 failed
```
29: Traceback (most recent call last):
29:   File "../../tools/drsnoop.py", line 154, in <module>
29:     fn_name="trace_direct_reclaim_begin")
29:   File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 810, in attach_tracepoint
29:     fn = self.load_func(fn_name, BPF.TRACEPOINT)
29:   File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 384, in load_func
29:     (func_name, errstr))
29: Exception: Failed to load BPF program trace_direct_reclaim_begin: Invalid argument
29: Fs....'unknown': I need something more specific.
29: ......Killed
29: .s.ss...sss..s.ss'unknown': I need something more specific.
29: .ss..s..s...s..'unknown': I need something more specific.
29: ...'unknown': I need something more specific.
29: .s...'unknown': I need something more specific.
29: ....s....
29: ======================================================================
29: FAIL: test_drsnoop (__main__.SmokeTests)
29: ----------------------------------------------------------------------
29: Traceback (most recent call last):
29:   File "/home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1604/tests/python/test_tools_smoke.py", line 148, in test_drsnoop
29:     self.run_with_int("drsnoop.py")
29:   File "/home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1604/tests/python/test_tools_smoke.py", line 52, in run_with_int
29:     or (rc == 137 and kill), "rc was %d" % rc)
29: AssertionError: rc was 1
```
You may need to guard the test with a higher kernel version.

yonghong-song:
Not 100% sure `vm_zone_stat`, `vm_numa_stat` or `vm_node_stat` are right counters to use for free memory. I guess the best way could be tracing how `/proc/meminfo` Free memory is calculated inside the kernel and to see whether we can reach the same `Free` memory result from `currect` task_struct.

tools/drsnoop.py
ethercflow: refactored, PTAL.

tools/drsnoop.py
ethercflow: refactored, PTAL.

tools/drsnoop.py
ethercflow: refactored, PTAL.

ethercflow:
> the ubuntu 16.04 failed
> 
> ```
> 29: Traceback (most recent call last):
> 29:   File "../../tools/drsnoop.py", line 154, in <module>
> 29:     fn_name="trace_direct_reclaim_begin")
> 29:   File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 810, in attach_tracepoint
> 29:     fn = self.load_func(fn_name, BPF.TRACEPOINT)
> 29:   File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 384, in load_func
> 29:     (func_name, errstr))
> 29: Exception: Failed to load BPF program trace_direct_reclaim_begin: Invalid argument
> 29: Fs....'unknown': I need something more specific.
> 29: ......Killed
> 29: .s.ss...sss..s.ss'unknown': I need something more specific.
> 29: .ss..s..s...s..'unknown': I need something more specific.
> 29: ...'unknown': I need something more specific.
> 29: .s...'unknown': I need something more specific.
> 29: ....s....
> 29: ======================================================================
> 29: FAIL: test_drsnoop (__main__.SmokeTests)
> 29: ----------------------------------------------------------------------
> 29: Traceback (most recent call last):
> 29:   File "/home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1604/tests/python/test_tools_smoke.py", line 148, in test_drsnoop
> 29:     self.run_with_int("drsnoop.py")
> 29:   File "/home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1604/tests/python/test_tools_smoke.py", line 52, in run_with_int
> 29:     or (rc == 137 and kill), "rc was %d" % rc)
> 29: AssertionError: rc was 1
> ```
> You may need to guard the test with a higher kernel version.

Right, I find It's because of the [enum bpf_prog_type BPF_PROG_TYPE_TRACEPOINT](https://elixir.bootlin.com/linux/v4.7-rc1/source/include/uapi/linux/bpf.h#L95) first introduced in the Linux Kernel 4.7, so I add a `skipUnless` to fix it.

yonghong-song:
[buildbot, test this please]

yonghong-song:
@ethercflow You are right in the earlier comment that `vm_zone_stat` is the correct place to get free memory. I confirmed that /proc/meminfo `MemFree` is from there. But currently, this info is not available to bpf programs.

@brendangregg what do you think?

yonghong-song:
cc @brendangregg

brendangregg:
If vm_zone_stat is the right thing to print (I haven't checked) its address can be read from /proc/kallsyms from the python program (is there a bcc helper for this already?) and passed to the BPF program. 

ext4dist does this (or used to) with code like:

```Python
# code replacements
with open(kallsyms) as syms:
    ops = ''
    for line in syms:
        (addr, size, name) = line.rstrip().split(" ", 2)
        name = name.split("\t")[0]
        if name == "ext4_file_operations":
            ops = "0x" + addr
            break
    if ops == '':
        print("ERROR: no ext4_file_operations in /proc/kallsyms. Exiting.")
        print("HINT: the kernel should be built with CONFIG_KALLSYMS_ALL.")
        exit()
    bpf_text = bpf_text.replace('EXT4_FILE_OPERATIONS', ops)
```

This was to add a free memory column, right?

yonghong-song:
I forgot this! Yes, we can just get the address of vm_zone_stat and do a bpf_probe_read on this.

ethercflow:
Get it, I'll add it soon.

tools/drsnoop.py
```diff
@@ -48,22 +52,52 @@
...
 
+
+# vm_stat
+
+vm_stat_addr = ''
+with open(kallsyms) as syms:
+    for line in syms:
+        (addr, size, name) = line.rstrip().split(" ", 2)
+        name = name.split("\t")[0]
+        if name == "vm_stat":
```
yonghong-song: I remember you mentioned `vm_stat` is defined in 3.x kernel? The minimum bpf support is 4.4. Maybe we do not need this?

tools/drsnoop.py
```diff
@@ -48,22 +52,52 @@
...
+            vm_stat_addr = "0x" + addr
+            break
+    if vm_stat_addr == '':
+        print("ERROR: no vm_stat or vm_zone_stat in /proc/kallsyms. Exiting.")
+        print("HINT: the kernel should be built with CONFIG_KALLSYMS_ALL.")
+        exit()
+
+NR_FREE_PAGES = 0
+
+PAGE_SHIFT = 12    # defined in page_types.h
```
yonghong-song: different systems may have differeent PAGE_SHIFT. Maybe you need to get from sysconf and replace it with proper values?

tools/drsnoop.py
```diff
@@ -48,22 +52,52 @@
 parser.add_argument("-n", "--name",
                     type=ArgString,
                     help="only print process names containing this name")
+parser.add_argument("-v", "--verbose", action="store_true",
```
yonghong-song: could you add this flag description in man page?

tools/drsnoop.py
ethercflow: Centos 7.6 with kernel version 3.10.0-957.el7.x86_64 support bpf, which still use `vm_stat`, so I think we need this.

tools/drsnoop.py
ethercflow: Oh, sorry, I forgot this.

tools/drsnoop.py
ethercflow: Agree! It seems no way to get `PAGE_SHIFT` directly,  maybe we can get `PAGE_SIZE` to cal `PAGE_SHIFT`. I'll make sure, and fix this.

yonghong-song:
[buildbot, test this please]

Issue #1250, opened 2017-07-10
Execsnoop can miss some events in parallel environments


anisse:
It turns out even execsnoop can miss some short-lived processes, when those run in a highly-parallel environment. I've ran into this while building a tree version of execsnoop (for flamegraphs). I have a build system that has quite a few short-lived processes, and sometimes I'd miss a few process, completely breaking the tree chain (which I've rewritten to get the ppid in-kernel).

I've built a simple program to illustrate the issue. It will run a given number of execve, over a given number of processors (it does this by forking).
```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>


#define min(a, b) ( (a) < (b) ? (a) : (b))
#define max(a, b) ( (a) > (b) ? (a) : (b))

void execbomb(int children, int start, int end, const char *filename) {
        int span = end - start;
        int per_child = 0, cstart, cend;
        char s_start[32], s_end[32];
        int parent = getpid();
        if (span <= 0)
                return;

        //printf("PID %d Got c=%d s=%d e=%d %s\n", parent, children, start, end, filename);
        if (children > 0) {
                // split the range
                children = min(span, children);
                per_child = max(span/children, 1);
                cstart = start;

                while (children > 0 && cstart <= end) {
                        cend = cstart + per_child;
                        if(!fork()) {
                                // child only works on its slice
                                start = cstart;
                                end = cend;
                                break; //stop forking
                        }
                        children--;
                        cstart += per_child;
                }
                if (parent == getpid()) {
                        if(cend < end)
                                start = cend; // parent gets to finish the job no child had
                        else
                                return;
                }
        }
        sprintf(s_start, "%d", start + 1);
        sprintf(s_end, "%d", end);

        // Fork to generate a new pid every time
        if (!fork())
                execl(filename, filename, s_start, s_end, NULL);
}

int main(int argc, char *argv[]) {
        int children = 0;
        if(argc < 3 || argc > 4) {
                printf("Usage: %s <start> <end> [forks]\n", argv[0]);
                return -1;
        }

        if (argc == 4)
                children = atoi(argv[3]);

        execbomb(children, atoi(argv[1]), atoi(argv[2]), argv[0]);

        return 0;
}

```
Say you want to run this in a single CPU, and generate 4000 execve():
`./execbomb 0 4000 0`

In this case, it seems `execsnoop.py  -n execbomb` catches all event. Now, run 17 parallel processes at a given time:
[...]

Issue #1920, opened 2018-08-12
http_filter missing packets near end of HTTP conversation


Birch-san:
I've followed the http_filter example, and am trying to count the bytes transferred to/from my webserver when files are downloaded via HTTP.

I successfully catch the packets at the start of the HTTP conversation (entirety of inbound HTTP request, and most packets comprising my outbound HTTP response). But frequently there are packets toward the end of my outbound HTTP response, that my packet filter doesn't log.

I used [`http-parse-complete.c`](https://github.com/iovisor/bcc/blob/master/examples/networking/http_filter/http-parse-complete.c) @ 3d9b687  
And I made my own, simple [`http-parse-seq.py`](https://gist.github.com/Birch-san/e9f21b2dc996c4405cd423527a633b2d), which does just one thing: print the sequence number of every packet that it sees.

Expected: Every sequence number I see in Wireshark/tcpdump should be logged by my bcc program also.  
Actual: later packets are often absent from my bcc program's logs, yet are visible in Wireshark/tcpdump.

Repro:

```bash
# make a directory which our webserver will mount
mkdir -p htdocs

# let's get a large photo for our webserver to serve
curl -o htdocs/docker.jpg 'https://reterwebber.files.wordpress.com/2014/09/docker-wallpaper-grey.jpg'

# launch apache webserver serving our image at http://localhost:8080/docker.jpg
docker run -i --rm --name httpd -p 8080:80 -v "$PWD"/htdocs:/usr/local/apache2/htdocs:ro,delegated


#== (meanwhile, in another terminal)
# download the bcc script
curl -o http-parse-complete.c 'https://raw.githubusercontent.com/iovisor/bcc/3d9b687/examples/networking/http_filter/http-parse-complete.c'
curl -o http-parse-seq.py 'https://gist.githubusercontent.com/Birch-san/e9f21b2dc996c4405cd423527a633b2d/raw/b91b16c5ae0705d4bc098d342c10bd0ed84f6e18/http-parse-seq.py'

# run the bcc script (logs sequence numbers of all HTTP packets inbound/outbound on docker0 interface)
sudo ./http-parse-seq.py -i docker0


#== (meanwhile, in another terminal)
# watch packets on docker0 interface
sudo tcpdump ip -Snnli docker0 -B 65536


#== (meanwhile, in another terminal)
# download photo from our website
curl -Is 'http://localhost:8080/docker.jpg'
```

I observed that my bcc script output the following sequence numbers:

```
2002486161
[...]

PR #2224, opened 2019-02-21
Expose kprobe's maxactive parameter in bcc

### LINKED ISSUE #1072: tools/tcpaccept.py's kretprobes not triggered due to low maxactive
When a kretprobe is installed on a kernel function, there is a maximum limit of how many calls in parallel it can catch (aka "maxactive"). In the case of a eBPF kretprobe, the maxactive is let to the default as defined in kernel/kprobes.c:

```
        /* Pre-allocate memory for max kretprobe instances */
        if (rp->maxactive <= 0) {
#ifdef CONFIG_PREEMPT
                rp->maxactive = max_t(unsigned int, 10, 2*num_possible_cpus());
#else
                rp->maxactive = num_possible_cpus();
#endif
        }
```

The maxactive can be as low as 1 on single-cores on kernel without CONFIG_PREEMPT.

tools/tcpaccept.py installs a kretprobe on `inet_csk_accept()` but that function can sleep for a long time (waiting for an incoming connection).

How to reproduce:

1. Start the tracer: `sudo ./tools/tcpaccept.py`
2. In another shell, start a bunch of processes waiting on the `accept()` system call: `for i in $(seq 1 25) ; do (busybox nc -l -p $((8081 + i)) &) ; done`
3. Start nginx locally: `sudo docker run -d nginx`
4. Connect to nginx: `curl 172.17.0.2`
5. Notice that there is no "accept" event

Note: use `nc` from `busybox` to reproduce this issue. `nc` from https://nmap.org/ncat (the default on Fedora) will not reproduce the issue because it does not block on the accept() system call (it waits on the `select()` system call to know when there is an incoming connection and only start the `accept()` system call at that time).

Similar issue also reported on https://github.com/weaveworks/tcptracer-bpf/issues/24 & https://github.com/weaveworks/tcptracer-bpf/issues/34

pchaigno:
When a kretprobe is installed on a kernel function, there is a limit on how many parallel calls it can catch. You can change that limit with the `maxactive` parameter. Since [commit `696ced4` ("tracing/kprobes: expose maxactive for kretprobe in kprobe_events")](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=696ced4fb1d76802f864d8848aa4716633f83c17) in the Linux kernel, debugfs exposes that parameter.

This pull request exposes that parameter in bcc as well, through `bpf_attach_kprobe` in libbcc and `BPF.attach_kretprobe` at the Python level. If a `maxactive` value is given, we fallback to using debugfs to install the kprobe event, instead of the usual `perf_event_open`.

For kernels without `maxactive` supports in debugfs, the error resolution is not straightforward. The kernel will still install a probe, but under a different name. We therefore need to check for our expected name, delete the probe with the different name, and try to attach the probe again, without the `maxactive` value.

Fixes #1072 #1250

/cc @alban @anisse

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -856,47 +856,100 @@ static int bpf_attach_tracing_event(int progfd, const char *event_path, int pid,
   return 0;
 }
 
+static int create_kprobe_event(char *buf, const char *ev_name,
+                               enum bpf_probe_attach_type attach_type,
+                               const char *fn_name, uint64_t fn_offset,
+                               int maxactive)
+{
+  int kfd;
+  char ev_alias[128];
+  static unsigned int buf_size = 256;
+
+  kfd = open("/sys/kernel/debug/tracing/kprobe_events", O_WRONLY | O_APPEND, 0);
+  if (kfd < 0) {
+    fprintf(stderr, "open(/sys/kernel/debug/tracing/kprobe_events): %s\n", buf,
+            strerror(errno));
+    return -1;
+  }
+
+  snprintf(ev_alias, sizeof(ev_alias), "%s_bcc_%d", ev_name, getpid());
+
+  if (fn_offset > 0 && attach_type == BPF_PROBE_ENTRY)
+    snprintf(buf, buf_size, "p:kprobes/%s %s+%"PRIu64,
+             ev_alias, fn_name, fn_offset);
+  else if (maxactive > 0 && attach_type == BPF_PROBE_RETURN)
+    snprintf(buf, buf_size, "r%d:kprobes/%s %s",
+             maxactive, ev_alias, fn_name);
+  else
+    snprintf(buf, buf_size, "%c:kprobes/%s %s",
+             attach_type == BPF_PROBE_ENTRY ? 'p' : 'r',
+             ev_alias, fn_name);
+
+  if (write(kfd, buf, strlen(buf)) < 0) {
+    if (errno == ENOENT)
+      fprintf(stderr, "cannot attach kprobe, probe entry may not exist\n");
+    else
+      fprintf(stderr, "cannot attach kprobe, %s\n", strerror(errno));
+    close(kfd);
+    return -1;
+  }
+  close(kfd);
+  snprintf(buf, buf_size, "/sys/kernel/debug/tracing/events/kprobes/%s",
+           ev_alias);
+  return 0;
+}
+
 int bpf_attach_kprobe(int progfd, enum bpf_probe_attach_type attach_type,
-                      const char *ev_name, const char *fn_name, uint64_t fn_offset)
+                      const char *ev_name, const char *fn_name, uint64_t fn_offset,
+                      int maxactive)
 {
   int kfd, pfd = -1;
-  char buf[256];
-  char event_alias[128];
-  static char *event_type = "kprobe";
+  char buf[256], fname[256];
 
-  // Try create the kprobe Perf Event with perf_event_open API.
-  pfd = bpf_try_perf_event_open_with_probe(fn_name, fn_offset, -1, event_type,
-                                           attach_type != BPF_PROBE_ENTRY);
+  if (maxactive <= 0)
+    // Try create the kprobe Perf Event with perf_event_open API.
+    pfd = bpf_try_perf_event_open_with_probe(fn_name, fn_offset, -1, "kprobe",
+                                             attach_type != BPF_PROBE_ENTRY);
+
   // If failed, most likely Kernel doesn't support the new perf_event_open API
   // yet. Try create the event using debugfs.
   if (pfd < 0) {
-    snprintf(buf, sizeof(buf), "/sys/kernel/debug/tracing/%s_events", event_type);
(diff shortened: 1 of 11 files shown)


abc123abd:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

src/cc/libbpf.c
```diff
@@ -856,47 +856,100 @@ static int bpf_attach_tracing_event(int progfd, const char *event_path, int pid,
...
+      if (access(fname, F_OK) == -1) {
+        // Deleting kprobe event with incorrect name.
+        kfd = open("/sys/kernel/debug/tracing/kprobe_events",
+                   O_WRONLY | O_APPEND, 0);
+        if (kfd < 0) {
+          fprintf(stderr, "open(/sys/kernel/debug/tracing/kprobe_events): %s\n",
+                  strerror(errno));
+          return -1;
+        }
+        snprintf(fname, sizeof(fname), "-:kprobes/%s_0", ev_name);
```
yonghong-song: It took a while for me to understand why a different name is used here. I did not check the kernel, but looks like kernel is pretty consistent with this name regardless of what probe name you gave to the kernel.

pchaigno: Oh, right! I'll add a comment.

pchaigno: Uh. I didn't get time to add that comment... I guess it's in the commit message at least.

yonghong-song: it should be okay. It is in the comments as well.

src/cc/libbpf.c
```diff
@@ -856,47 +856,100 @@ static int bpf_attach_tracing_event(int progfd, const char *event_path, int pid,
...
+
+  snprintf(ev_alias, sizeof(ev_alias), "%s_bcc_%d", ev_name, getpid());
+
+  if (fn_offset > 0 && attach_type == BPF_PROBE_ENTRY)
+    snprintf(buf, buf_size, "p:kprobes/%s %s+%"PRIu64,
+             ev_alias, fn_name, fn_offset);
+  else if (maxactive > 0 && attach_type == BPF_PROBE_RETURN)
+    snprintf(buf, buf_size, "r%d:kprobes/%s %s",
+             maxactive, ev_alias, fn_name);
+  else
```
palmtenor: Does `maxactive` do anything in k(entry)probe? If not, maybe we should check that the value should be 0?

src/cc/libbpf.c
palmtenor: Is it possible to re-use detach kprobe logic here?

src/cc/libbpf.c
```diff
@@ -856,47 +856,100 @@ static int bpf_attach_tracing_event(int progfd, const char *event_path, int pid,
...
+            fprintf(stderr, "cannot detach kprobe, probe entry may not exist\n");
+          else
+            fprintf(stderr, "cannot detach kprobe, %s\n", strerror(errno));
+          close(kfd);
+          goto error;
+        }
+        close(kfd);
+
+        // Re-creating kprobe event without maxactive.
+        if (create_kprobe_event(buf, ev_name, attach_type, fn_name,
```
palmtenor: So... ideally here we should go back and try use Perf Event based approach?

yonghong-song: probably not needed. pref event based approach is supported in 4.17 and maxactive is supported in 4.12...

src/cc/libbpf.c
pchaigno: `maxactive` doesn't do anything for kprobes. Do you mean check that the value is 0 at the beginning of `bpf_attach_kprobe`?

src/cc/libbpf.c
pchaigno: I considered it, but `bpf_detach_probe` also contains some code to handle the case where we attached the probe using `perf_event_open` (it iterates on the list of events). We could break `bpf_detach_probe` down to refactor, but I want to try and add support for `maxactive` in `perf_event_open`, so I thought I'd refactor after that.

src/cc/libbpf.c
pchaigno: Hm. I missed that. I should've called `bpf_attach_kprobe` with `maxactive=0` instead. I could refactor if you think it's worth it.

Also, I might be missing something, but is there any reason the `perf_event_open` API is preferred to the debugfs'?

src/cc/libbpf.c
yzhao1012: Sorry to bug a closed issue.

But where is the documentation for maxactive as referred in "See the kprobes documentation for its default value."?

We run into issue similar to https://github.com/iovisor/bcc/issues/1920 where bpf_trace_printk() states that the perf_submit() succeeded, but user-space did not read the events from the perf buffer.

pchaigno: The default value is discussed in [the kprobes documentation in the kernel](https://www.kernel.org/doc/Documentation/kprobes.txt):

> If maxactive <= 0, it is
set to a default value.  If CONFIG_PREEMPT is enabled, the default
is max(10, 2*NR_CPUS).  Otherwise, the default is NR_CPUS.

PR #2228, opened 2019-02-25
support gcc 8's cold subfunctions

arighi:
GCC 8 can split functions bodies into hot and cold regions, causing
extra symbols with a ".cold.N" suffix to be emitted. Exclude these
extra symbols and only allow the parent function to be traced.

Signed-off-by: Andrea Righi <righi.andrea@gmail.com>


yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks for the fix!

PR #2232, opened 2019-02-26
tcpaccept: only show TCP accept event

boat0:
pr #1842 forgot to add filtering in the new tracepoint code so it's
incorrectly showing every sock:inet_sock_set_state event.

This patch fixes that.


yonghong-song:
[buildbot, test this please]

yonghong-song:
This change makes sense. For the serve side, the `accept` call should TCP_SYNC_RECV followed by TCP_ESTABLISHED. For the client side, TCP_SYNC_SEND followed by TCP_ESTABLISHED.

PR #2234, opened 2019-02-27
Fix the output of example scripts with python3 and handle KeyboardInterrupt

lcp:
Switch to printb to get rid of the prefix 'b' of the byte arrays.
Also handle the KeyboardInterrupt exception for some scripts to avoid the code tracing when stopping the program.

### DIFF
--- a/examples/tracing/disksnoop.py
+++ b/examples/tracing/disksnoop.py
@@ -12,6 +12,7 @@
 
 from __future__ import print_function
 from bcc import BPF
+from bcc.utils import printb
 
 REQ_WRITE = 1		# from include/linux/blk_types.h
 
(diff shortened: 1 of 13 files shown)


brendangregg:
I would not change hello world from a 1 line to a 4 line program. It needs to be the simplest possible program -- that's its whole purpose.

yonghong-song:
[buildbot, test this please]

yonghong-song:
@lcp could you revert the hello_world.py change as commented above by @brendangregg?

lcp:
I see. Will update the PR.

lcp:
The change of hello_world.py has been removed.

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

PR #2236, opened 2019-02-27
added DNAME_INLINE_LEN to be 32 as per filetop

mcaleavya:
fileslower was failing on newest version of bcc/tools unsure as to why right now however filetop has explicit declaration of DNAME_INLINE_LEN , adding resolves immediate issue.


boat0:
This bug is due to #2204 where we cleaned up definitions in python.
'DNAME_INLINE_LEN = 32' should NOT have been deleted.
Thanks for the fix!

yonghong-song:
[buildbot, test this please]

Issue #2231, opened 2019-02-26
Attaching uprobes to fixed addresses


mephi42:
I'm trying to trace an asm instruction at the known offset relative to module start. The Python doc for `attach_uprobe` says: "The real address addr may be supplied in place of sym." So I tried:

```
b.attach_uprobe(name=obj, addr=insn, fn_name='<redacted>')
```

which results in

```
Exception: could not determine address of symbol b''
```

I also tried

```
b.attach_uprobe(name=obj, sym=None, addr=insn, fn_name='<redacted>')
```

but then I get

```
Exception: could not determine address of symbol None
```

Is there some other way to do this? Is this supported at all?

mephi42:
It turned out that the problem was that the module in question was non-PIE executable, and in that case the address must have been the absolute one, as opposed to the offset relative to the start of the module.

Maybe this could be added to the `attach_uprobe` documentation?

palmtenor:
In the underlying logic, when you have provided a symbol name, we still resolve it to an address for attaching, so this should work. Can you repro this on some common library like `libc` etc. so that I could debug?

mephi42:
This is the minimal repro of what I struggled with:

```
$ echo 'int main() {}' >test.c
$ gcc test.c
$ objdump -x a.out | grep -w -e "LOAD off" -e main
    LOAD off    0x0000000000000000 vaddr 0x0000000001000000 paddr 0x0000000001000000 align 2**12
    LOAD off    0x0000000000000e00 vaddr 0x0000000001001e00 paddr 0x0000000001001e00 align 2**12
00000000010004f8 g     F .text	000000000000001a              main

# offset = 0x10004f8 - 0x1000000 = 0x4f8

$ python3 -c 'import bcc; b = bcc.BPF(text="int foo() { return 0; }"); b.attach_uprobe(name="./a.out", addr=0x4f8, fn_name="foo")'
Exception: could not determine address of symbol b''
```

My fix/workaround:

```
$ python3 -c 'import bcc; b = bcc.BPF(text="int foo() { return 0; }"); b.attach_uprobe(name="./a.out", addr=0x10004f8, fn_name="foo")'
```

yonghong-song:
In the above, the virtual address is needed, not the offset from the start of the section.

mephi42:
Ok, but for shared libraries one needs an offset:

```
$ objdump -x /lib64/libc.so.6 | grep -w -e "LOAD off" -e __libc_main
[...]

PR #2243, opened 2019-02-28
Fix uprobe behavior when user provided an address

palmtenor:
See #2231 for context. Updated the behavior, and updated documentation for better explanation.


palmtenor:
@mephi42 Can you test on this patch?

PR #2248, opened 2019-03-02
Add source installation instructions for Centos 7

ethercflow:


palmtenor:
[buildbot, ok to test]

yonghong-song:
[buildbot, test this please]

yonghong-song:
The change looks good. It gives complete detailed information about Centos installation. I am wondering whether in the future we can abstract out the common installation instructions so they do not need to be repeated for every instruction.

PR #2254, opened 2019-03-06
enhance tcpaccept

boat0:
- add option `-T': include time column on output (HH:MM:SS)
- add option `-P PORT': only trace port specified
- add RPORT colume on output

### DIFF
--- a/tools/tcpaccept.py
+++ b/tools/tcpaccept.py
@@ -4,7 +4,7 @@
 # tcpaccept Trace TCP accept()s.
 #           For Linux, uses BCC, eBPF. Embedded C.
 #
-# USAGE: tcpaccept [-h] [-t] [-p PID]
+# USAGE: tcpaccept [-h] [-T] [-t] [-p PID] [-P PORTS]
 #
 # This uses dynamic tracing of the kernel inet_csk_accept() socket function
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, test this please]

tools/tcpaccept.py
```diff
@@ -21,21 +21,27 @@
...
+    ./tcpaccept -P 80     # only trace port 80
     ./tcpaccept -p 181    # only trace PID 181
 """
 parser = argparse.ArgumentParser(
     description="Trace TCP accepts",
     formatter_class=argparse.RawDescriptionHelpFormatter,
     epilog=examples)
+parser.add_argument("-T", "--time", action="store_true",
+    help="include time column on output (HH:MM:SS)")
 parser.add_argument("-t", "--timestamp", action="store_true",
```
yonghong-song: Should we make -t and -T exclusive to each other? It probably does make sense to have both in outputs?

boat0: With `tcplife` we can:
    `./tcplife -stT      # csv output, with times & timestamps`
So I guess it is ok to make -t and -T non-exclusive.

boat0:
> Could you add new flags to man page as well? Thanks!
man pages updated.

yonghong-song:
[buildbot, test this please]

Issue #2252, opened 2019-03-05
filetop still shows no writes


joelagnel:
Similar to: https://github.com/iovisor/bcc/issues/1667

Just tried the command in Android x86 emulator running 4.14 kernel:
```
while true; do cat testfile > /dev/null; echo "asdfasdf" >> testfile; done
```

And I still see the issue. First issue is wrong filename is displayed (I was expecting to see `testfile`) second issue is write column is 0. This is using latest BCC
```
TID    COMM             READS  WRITES R_Kb    W_Kb    T FILE
2463   clear            2      0      60      0       R screen-256color
2296   cat              35     0      35      0       R ld.config.Q.txt
2363   cat              35     0      35      0       R ld.config.Q.txt
2357   cat              35     0      35      0       R ld.config.Q.txt
2170   cat              35     0      35      0       R ld.config.Q.txt
2373   cat              35     0      35      0       R ld.config.Q.txt
2280   cat              35     0      35      0       R ld.config.Q.txt
2299   cat              35     0      35      0       R ld.config.Q.txt
2443   cat              35     0      35      0       R ld.config.Q.txt
2347   cat              35     0      35      0       R ld.config.Q.txt
2381   cat              35     0      35      0       R ld.config.Q.txt
2257   cat              35     0      35      0       R ld.config.Q.txt
2469   cat              35     0      35      0       R ld.config.Q.txt
2288   cat              35     0      35      0       R ld.config.Q.txt
2175   cat              35     0      35      0       R ld.config.Q.txt
2440   cat              35     0      35      0       R ld.config.Q.txt
2256   cat              35     0      35      0       R ld.config.Q.txt
2374   cat              35     0      35      0       R ld.config.Q.txt
2472   cat              35     0      35      0       R ld.config.Q.txt
2453   cat              35     0      35      0       R ld.config.Q.txt
```

joelagnel:
The `ld.config.Q.txt` read is actually a real file being read on Android by its linker. So that's not an issue here. The real issue is fixed by about PR.

PR #2263, opened 2019-03-08
filetop: Avoid missing important entries

joelagnel:
While testing, it is noticed that sometimes filetop can miss
certain heavy-weight activity. This is because it sorts by only a single
column. So for example, if writes are flooding, then it will be way down
in the filetop list and not displayed at all for the default maxrows of
20 and default sort of rbytes. It is better instead to sort by all
columns, by default. This will help catch them.

Test:
Start a bash and do: while true; do echo "asdfasdf" >> testfile; done
Run filetop. Without patch no activity is displayed due to other small
reads on the system. With the path they are.

Fixes issue: https://github.com/iovisor/bcc/issues/2252

Signed-off-by: Joel Fernandes (Google) <joel@joelfernandes.org>

### DIFF
--- a/tools/filetop.py
+++ b/tools/filetop.py
@@ -38,8 +38,8 @@ parser.add_argument("-C", "--noclear", action="store_true",
     help="don't clear the screen")
 parser.add_argument("-r", "--maxrows", default=20,
     help="maximum rows to print, default 20")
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

brendangregg:
makes sense, thanks!

PR #2266, opened 2019-03-11
Add basic USDT support for s390x

mephi42:
Approach and code shamelessly borrowed from "422db709: Add basic USDT
support for powerpc64".


palmtenor:
[buildbot, ok to test]

PR #2279, opened 2019-03-19
biosnoop: add -Q for queued time

brendangregg:
I also cleaned up the output logic.


brendangregg:
I see more things to fix; hang on...

brendangregg:
Should be ok now. I might add a -p PID later...

brendangregg:
added a biotop(8) pid fix as well.

PR #2282, opened 2019-03-20
fix string re being used on bytes for Python 3

lilydjwg:
When using killsnoop with Python 3, I get the following errors:

```pybt
  File "_ctypes/callbacks.c", line 232, in 'calling callback function'                                                                 
  File "/usr/lib/python3.7/site-packages/bcc/table.py", line 676, in raw_cb_                                                           
    callback(cpu, data, size)
  File "/usr/share/bcc/tools/killsnoop", line 124, in print_event                                                                      
    event = b["events"].event(data)                                                                                                    
  File "/usr/lib/python3.7/site-packages/bcc/table.py", line 654, in event                                                             
    self._event_class = self._get_event_class()
  File "/usr/lib/python3.7/site-packages/bcc/table.py", line 625, in _get_event_class                                                  
    m = re.match(r"(.*)#(.*)", field)                                                                                                  
  File "/usr/lib/python3.7/re.py", line 173, in match
    return _compile(pattern, flags).match(string)                                                                                      
TypeError: cannot use a string pattern on a bytes-like object
```

This change fixes this. It works with Python 2 too.

### DIFF
--- a/src/python/bcc/table.py
+++ b/src/python/bcc/table.py
@@ -621,7 +621,7 @@ class PerfEventArray(ArrayBase):
         num_fields = lib.bpf_perf_event_fields(self.bpf.module, self._name)
         i = 0
         while i < num_fields:
-            field = lib.bpf_perf_event_field(self.bpf.module, self._name, i)
+            field = lib.bpf_perf_event_field(self.bpf.module, self._name, i).decode()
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

PR #2283, opened 2019-03-20
bitesize: switch to insert tracepoint

brendangregg:
Sorry, should have used issue. insert rarely fires.


yonghong-song:
Indeed, trace_block_rq_issue corresponds to blk_mq_start_request or blk_start_request and is a better choice.

PR #2285, opened 2019-03-21
delete allocated array properly

yonghong-song:
In bcc_btf.cc, new_btf_sec is allocated with something like
```
   new_btf_sec = new uint8_t[tmp_sec_size]
```
Since the allocation is an array, the deletion should be
```
   delete[] new_btf_sec
```
instead of
```
   delete new_btf_sec
```
This patch fixed the problem.

### DIFF
--- a/src/cc/bcc_btf.cc
+++ b/src/cc/bcc_btf.cc
@@ -179,7 +179,7 @@ int BTF::load(uint8_t *btf_sec, uintptr_t btf_sec_size,
 
   if (new_btf_sec) {
     btf = btf__new(new_btf_sec, new_btf_sec_size);
-    delete new_btf_sec;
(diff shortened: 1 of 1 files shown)


palmtenor:
I think we've had similar issues before. Do you think we can have Travis run an ASAN build as well?

yonghong-song:
Yes, this happens before for one in bpf_module.cc. I do agree that we should increase coverage for this. Maybe just add additional round of testing with different compilation flags? I am not familiar with Travis. Any details of how to do this?

pchaigno:
Note: Despite the presence of a [`.travis.yml` file](https://github.com/iovisor/bcc/blob/master/.travis.yml), Travis CI is not enabled at the moment on the repository.

PR #2287, opened 2019-03-25
tools/cachestat: fix python 3 str/bytes type error

10ne1:
TypeError: argument should be integer or bytes-like object, not 'str'

Signed-off-by: Adrian Ratiu <adrian.ratiu@collabora.com>


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

PR #2122, opened 2019-01-09
Translate arch into source directory when ARCH is set

jeromemarchand:
When ARCH is not defined, bcc get the architecture from uname. It then
modifies it to get the name of arch directory in linux source.

When ARCH is defined however, it just copy it as is, without the
translation to the arch directory. If for instance ARCH is set to
x86_64, it tries to look into the include directory
build/arch/x86_64/, which doesn't exist.

It fixes the following issue:
$ echo $ARCH
x86_64
$ /usr/share/bcc/tools/bashreadline
In file included from <built-in>:2:
In file included from /virtual/include/bcc/bpf.h:12:
In file included from /lib/modules/4.18.0-49.el8.x86_64/build/include/linux/types.h:6:
/lib/modules/4.18.0-49.el8.x86_64/build/include/uapi/linux/types.h:5:10: fatal error: 'asm/types.h' file
      not found
         ^~~~~~~~~~~~~
1 error generated.
Traceback (most recent call last):
  File "/usr/share/bcc/tools/bashreadline", line 51, in <module>
    b = BPF(text=bpf_text)
  File "/usr/lib/python3.6/site-packages/bcc/__init__.py", line 318, in __init__
    raise Exception("Failed to compile BPF text")
Exception: Failed to compile BPF text

PR #2288, opened 2019-03-26
clang/kbuild_helper: fix arm64 cross compilation

10ne1:
Commit 28949f17e0 ("Translate arch into source directory when ARCH is set")
moved $ARCH processing earlier in the function so $ARCH gets processed
by the the same logic which parses the uname_machine value.

This introduced two bugs breaking ARCH=arm64 cross-compilation:

1. The arch.compare(0, 3, "arm") test matches both $ARCH=arm and $ARCH=arm64
leading to builds which always include from "arch/arm/include" instead
of "arch/arm64/include".

2. The only way arch == arm64 is if $ARCH=aarch64. uname returns aarch64
but the rest of the compiler logic expects $ARCH=arm64.

This fixes the above bugs by moving the "aarch64" test earlier than
the "arm" test and also accepting ARCH=arm64.

Fixes: 28949f17e0 ("Translate arch into source directory when ARCH is set (#2122)")
Signed-off-by: Adrian Ratiu <adrian.ratiu@collabora.com>

### DIFF
--- a/src/cc/frontends/clang/kbuild_helper.cc
+++ b/src/cc/frontends/clang/kbuild_helper.cc
@@ -46,6 +46,8 @@ int KBuildHelper::get_flags(const char *uname_machine, vector<string> *cflags) {
     arch = "x86";
   } else if (arch[0] == 'i' && !arch.compare(2, 2, "86")) {
     arch = "x86";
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

PR #2289, opened 2019-03-26
doc: explain how to use perf with skb programs

oriolarcas:
There are no examples of how to use perf_submit in skb programs. This commit explains that a skb can be used as a ctx.

Signed-off-by: Oriol Arcas <oriol@starflownetworks.com>


yonghong-song:
[buildbot, test this please]

PR #2293, opened 2019-03-29
cmake: expose bcc_elf.h in the API headers

mmarchini:
Ref: https://github.com/iovisor/bpftrace/pull/410


yonghong-song:
@palmtenor You used to have some concerns to expose this header. Could you elaborate?
@mmarchini Which exact functions you try to use from bcc_elf.h?

mmarchini:
`bcc_elf_foreach_sym`, we want to use it on bpftrace to get available uprobes from a binary without using `objdump`.

palmtenor:
[buildbot, ok to test]

palmtenor:
I remember last time it was `syms.h` which was explicitly designed not to be exposed. I think `bcc_elf.h` is OK...

PR #2294, opened 2019-03-29
Add bcc_usdt_enable_fully_specified_probe to avoid usdt provider collisions

### LINKED ISSUE #2275: USDT probe provider collisions
I believe this is mentioned in https://github.com/iovisor/bcc/issues/1527, so perhaps this is a duplicate of that. Either way, I want to write up this problem specifically as I would like to make an attempt to solve it.

In the [current usdt implementation](https://github.com/iovisor/bcc/blob/master/src/cc/usdt/usdt.cc#L296-L320) if you have two tracepoints with the same name, but different provider, there will be a collision and the one found first will be used, hiding the other one.

This will result in unexpected and probably undesired behavior, where the same name may exist in different libraries included by an executable. 

The interface is currently broken, as the [interface doesn't require or accept a provider name](https://github.com/iovisor/bcc/blob/master/src/cc/usdt/usdt.cc#L445), so the collision is inevitable in certain circumstances. 

I propose that the existing signature of  to include the provider name. So the current signature

```
bcc_usdt_enable_probe(void *usdt, const char *probe_name, const char *fn_name)
```

would be changed to:

```
bcc_usdt_enable_probe(void *usdt, const char* provide_name, const char *probe_name, const char *fn_name)
```

Then, the enable_probe function would just need to be modified to search by namespace which should be pretty easy to implement, as this provider name just needs to be propagated. The probe object seems to [already support and store the provider name](https://github.com/iovisor/bcc/blob/master/src/cc/usdt/usdt.cc#L56-L59), so this seems like a small change, aside from it being a potentially breaking API change.

For compatibility, we may want to support signature requiring only 3 arguments at least until the next point release, preserving the current behavior but throwing a deprecation warning. This would give some time for consumers of the API to update to avoid the error, and then I would suggest eventually removing the 3 argument interface.

dalehamel:
fixes #2275

I aimed for the most minimal changeset I could in order to avoid provider collisions.

I have successfully used this code in https://github.com/iovisor/bpftrace/pull/508 to be able to probe a function that would otherwise have a collision with another provider:

As you can see, `testprobe2` is defined on both the `tracetest` and `tracetest2` providers:

```
bpftrace -l 'usdt:./tracetest'
usdt:./tracetest:tracetest:testprobe
usdt:./tracetest:tracetest:testprobe2
usdt:./tracetest:tracetest2:testprobe2
```

Using the old API (`bcc_usdt_enable_probe`):

```
bpftrace -e 'usdt::tracetest2:testprobe2 { @[probe]++ }' -p ${PID}
Attaching 1 probe...
Found duplicate provider (tracetest2) for underspecified probe (testprobe2)
Found 2 duplicate providers for underpecified probe (probe_testprobe2_1)
Error finding or enabling probe: usdt:tracetest2:testprobe2
```

And the new API (`bcc_usdt_enable_fully_specified_probe`):

```
bpftrace -e 'usdt::tracetest2:testprobe2 { @[probe]++ }' -p ${PID}
Attaching 1 probe...
^C

@[usdt:tracetest2:testprobe2]: 2
```

To maintain compatibility:

* I added a check `BCC_USDT_HAS_FULLY_SPECIFIED_PROBE`, which can be useful for tools running against a previous release of libbcc to maintain their current (degraded) functionality.
* I defined a new function to enable probes, specifying the provider: `bcc_usdt_enable_fully_specified_probe`

Things I'd like comments on, in addition to the review:
* We may want to have a deprecation strategy for the old API, or perhaps it should remain in place. If we decide to deprecate, we will need to make an issue to update bcc tools to use the new API
* A better (shorter) name for `bcc_usdt_enable_fully_specified_probe`
* Tests to add to existing suite to verify this functionality
* Thoughts on what, if anything, should hold up merging this.

### DIFF
--- a/src/cc/usdt/usdt.cc
+++ b/src/cc/usdt/usdt.cc
@@ -304,35 +304,17 @@ bool Context::enable_probe(const std::string &provider_name,
   if (pid_stat_ && pid_stat_->is_stale())
     return false;
 
-  unsigned int matches = 0;
   Probe *found_probe = nullptr;
   for (auto &p : probes_) {
-    if (p->name_ == probe_name) {
-      if (found_probe == nullptr && provider_name == "")
-      {
-        found_probe = p.get();
-        matches++;
+    if (p->name_ == probe_name &&
+        (provider_name.empty() || p->provider() == provider_name)) {
+      if (found_probe != nullptr) {
+        fprintf(stderr, "Two same-name probes (%s) but different providers\n",
+                probe_name.c_str());
+        return false;
       }
-      else if (found_probe != nullptr && provider_name == "")
-      {
-        fprintf(stderr, "Found duplicate provider (%s) for underspecified probe (%s)\n",
-                p->provider().c_str(), p->name().c_str());
-        matches++;
-      } else if (provider_name != "" && p->provider() == provider_name)
-      {
-        found_probe = p.get();
-        matches++;
-      }
+      found_probe = p.get();
     }
-  }
-
-  if (matches > 1) {
-    fprintf(stderr, "Found %i duplicate providers for underpecified probe (%s)\n",
-                matches, fn_name.c_str());
-    return false;
-  } else if(matches < 1) {
-    fprintf(stderr, "No matches found for probe (%s)\n", fn_name.c_str());
-    return false;
   }
 
   if (found_probe != nullptr)
@@ -470,8 +452,7 @@ int bcc_usdt_enable_probe(void *usdt, const char *probe_name,
   return ctx->enable_probe(probe_name, fn_name) ? 0 : -1;
 }
 
-int bcc_usdt_enable_fully_specified_probe(void *usdt,
-                                          const char *provider_name,
+int bcc_usdt_enable_fully_specified_probe(void *usdt, const char *provider_name,
                                           const char *probe_name,
                                           const char *fn_name) {
   USDT::Context *ctx = static_cast<USDT::Context *>(usdt);
--- a/src/cc/usdt/usdt.cc
+++ b/src/cc/usdt/usdt.cc
@@ -295,22 +295,44 @@ Probe *Context::get(const std::string &provider_name,
 
 bool Context::enable_probe(const std::string &probe_name,
                            const std::string &fn_name) {
+  return enable_probe("", probe_name, fn_name);
+}
+
+bool Context::enable_probe(const std::string &provider_name,
+                           const std::string &probe_name,
+                           const std::string &fn_name) {
   if (pid_stat_ && pid_stat_->is_stale())
     return false;
 
-  // FIXME: we may have issues here if the context has two same probes's
-  // but different providers. For example, libc:setjmp and rtld:setjmp,
-  // libc:lll_futex_wait and rtld:lll_futex_wait.
+  unsigned int matches = 0;
   Probe *found_probe = nullptr;
   for (auto &p : probes_) {
     if (p->name_ == probe_name) {
-      if (found_probe != nullptr) {
-         fprintf(stderr, "Two same-name probes (%s) but different providers\n",
-                 probe_name.c_str());
-         return false;
+      if (found_probe == nullptr && provider_name == "")
+      {
+        found_probe = p.get();
+        matches++;
       }
-      found_probe = p.get();
+      else if (found_probe != nullptr && provider_name == "")
+      {
+        fprintf(stderr, "Found duplicate provider (%s) for underspecified probe (%s)\n",
+                p->provider().c_str(), p->name().c_str());
+        matches++;
+      } else if (provider_name != "" && p->provider() == provider_name)
+      {
+        found_probe = p.get();
+        matches++;
+      }
     }
+  }
+
+  if (matches > 1) {
+    fprintf(stderr, "Found %i duplicate providers for underpecified probe (%s)\n",
+                matches, fn_name.c_str());
+    return false;
+  } else if(matches < 1) {
+    fprintf(stderr, "No matches found for probe (%s)\n", fn_name.c_str());
+    return false;
   }
 
   if (found_probe != nullptr)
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
cc @palmtenor

src/cc/usdt/usdt.cc
```diff
@@ -295,24 +295,46 @@ Probe *Context::get(const std::string &provider_name,
...
+  unsigned int matches = 0;
   Probe *found_probe = nullptr;
   for (auto &p : probes_) {
     if (p->name_ == probe_name) {
-      if (found_probe != nullptr) {
-         fprintf(stderr, "Two same-name probes (%s) but different providers\n",
-                 probe_name.c_str());
-         return false;
+      if (found_probe == nullptr && provider_name == "")
+      {
```
palmtenor: Please use consistent formatting as rest of the file, or you can just do `clang-format -i`.

src/cc/usdt/usdt.cc
```diff
@@ -295,24 +295,46 @@ Probe *Context::get(const std::string &provider_name,
...
-  // libc:lll_futex_wait and rtld:lll_futex_wait.
+  unsigned int matches = 0;
   Probe *found_probe = nullptr;
   for (auto &p : probes_) {
     if (p->name_ == probe_name) {
-      if (found_probe != nullptr) {
-         fprintf(stderr, "Two same-name probes (%s) but different providers\n",
-                 probe_name.c_str());
-         return false;
+      if (found_probe == nullptr && provider_name == "")
```
palmtenor: `provider_name.empty()`, and rest call sites below

src/cc/usdt/usdt.cc
```diff
@@ -295,24 +295,46 @@ Probe *Context::get(const std::string &provider_name,
...
+      {
+        found_probe = p.get();
+        matches++;
+      }
+      else if (found_probe != nullptr && provider_name == "")
+      {
+        fprintf(stderr, "Found duplicate provider (%s) for underspecified probe (%s)\n",
```
palmtenor: Why do we need these logic? I mean it's totally normal for Tracepoints from two different providers have the same name (which is exactly the problem we are solving).

Also these `if` conditions are quite twisted... How about:
```
if ((p->name_ == probe_name) && (provider_name.empty() || (provider_name == p->provider()))) {
  do_normal_found_probe_work;
  if (provider_name.empty()) {
    do_some_work_to_warn_people;
  }
}
```

src/cc/usdt/usdt.cc
```diff
@@ -295,24 +295,46 @@ Probe *Context::get(const std::string &provider_name,
...
       }
-      found_probe = p.get();
     }
   }
 
+  if (matches > 1) {
+    fprintf(stderr, "Found %i duplicate providers for underpecified probe (%s)\n",
+                matches, fn_name.c_str());
+    return false;
+  } else if(matches < 1) {
```
palmtenor: I think simply return `false` would do. Caller are already checking the not found failure result.

src/cc/usdt/usdt.cc
dalehamel: Will do! I wasn't aware of that tool. I found other issues in this file, I'll ignore those and just fix the ones related to my changes.

src/cc/usdt/usdt.cc
dalehamel: Hmm yes I agree it can be simplified. I thought it would be helpful to indicate where the duplicate providers are if they didn't specify the provider name, but we can just leave it up to the caller to figure out.

src/cc/usdt/usdt.cc
dalehamel: Yes I'll clean this up, your above suggestion negates the need for this.

dalehamel:
thanks for the review @palmtenor, I hope to get the revised version up later today or tomorrow.

src/cc/usdt/usdt.cc
dalehamel: I've run the tool, and found a helpful script that was able to do it on just the diff:

`git diff -U0 --no-color HEAD^ | clang-format-diff -p1`

Is there an existing document with contributor guidelines that I missed where this could be added?

src/cc/usdt/usdt.cc
dalehamel: I've replaced this one occurrence with the idiomatic approach, and removed the others.

src/cc/usdt/usdt.cc
dalehamel: I followed this approach of checking all of the conditions, and setting the pointer if it's unset, otherwise logging an error.

This is the behavior it does now, and the enhancement that I've added is that if a provider is specified it will use it and avoid the collision.

src/cc/usdt/usdt.cc
dalehamel: I just outright got rid of this, as I ended up putting the same thing in my bpftrace PR anyways.

I would expect other users (such as scripts in this repo) should take the same approach of handling this. I'd be happy to help make those changes as well.

dalehamel:
@palmtenor I believe I've addressed your suggestions, please give it another look when you have the chance.

note that in my environment, the following test fails, but also fail on master. 
https://gist.github.com/dalehamel/e46f1b8a380231722e9c1bf68236a5a5:

It doesn't appear that there are any existing tests that cover enabling probes, so to validate my code I have used the test program specified in the PR description:

```
#include <sys/sdt.h>
 #include <sys/time.h>
 #include <unistd.h>
 #include <stdio.h>

static long
myclock() {
    struct timeval tv;
    gettimeofday(&tv, NULL);
    DTRACE_PROBE1(tracetest, testprobe, tv.tv_sec);
    DTRACE_PROBE1(tracetest, testprobe2, tv.tv_sec);
    DTRACE_PROBE1(tracetest2, testprobe2, tv.tv_sec);
    return tv.tv_sec;
}

int
main(int argc, char **argv) {
    while (1) {
        myclock();
        sleep(1);
    }
    return 0;
}
```

And then I've used a branch of bpftrace that i've been working on to test this. The behavior that I observe is that by using the new function and specifying a provider, I'm able to successfully attach to a probe where there previously would have been a provider collision.

dalehamel:
@palmtenor @yonghong-song sorry to ping again, would appreciate another review / if one of you could kick off the tests for this

yonghong-song:
[buildbot, test this please]

palmtenor:
Sorry @dalehamel, on the road this week, will take a closer look during the weekend:)

dalehamel:
Rebased-in master

yonghong-song:
[buildbot, test this please]

yonghong-song:
I do not particularly like BCC_USDT_HAS_FULLY_SPECIFIED_PROBE, but I guess we can live with it for now and later retire it with a few more releases. Otherwise, the patch looks good to me. In the future, we need to tweak python/C++ (esp. python) usdt user interfaces to include provider name.

PR #2296, opened 2019-04-01
Add libbcc-loader-static.a symbols into libbcc.a

danobi:
It's useful to have all the bcc symbols in one place when statically
linking against bcc. This patch adds all the symbols from
libbcc-loader-static into libbcc. This is in line with how
libbcc-lua-static does it as well.


yonghong-song:
[buildbot, test this please]

PR #2303, opened 2019-04-07
tcpaccept/tcpconnect: use PID not TID

brendangregg:


brendangregg:
hang on, fixing something...

brendangregg:
ok, that's better. tcpconnect uses both the pid for printing and filtering, and the tid as a thread-ID.

PR #2305, opened 2019-04-09
tcpaccept: remove sock:inet_sock_set_state tracepoint code

### LINKED ISSUE #2304: tcpaccept PID/COMM wrong
Since #1838 and #2232 etc, tcpaccept will use the sock:inet_sock_set_state tracepoint, if available. However this breaks the PID and COMM columns:

```
# /usr/share/bcc/tools/tcpaccept
PID     COMM         IP RADDR            RPORT LADDR            LPORT
26565   ssh          4  127.0.0.1        59978 127.0.0.1        22   
```

ssh was not accepting. It was sshd. Tracing inet_csk_accept() matches this correctly.

I don't think sock:inet_sock_set_state can be used, unfortunately. Can someone please remove the tracepoint code, unless they have a solution? // @kernel-z @boat0

boat0:
Fixes #2304.

On Linux 4.16 and later, sock:inet_sock_set_state tracepoint was
used for efficency, but it may output wrong PIDs. This is because
sock:inet_sock_set_state may run outside of process context.
Hence, we stick to kprobes until we find a proper solution.

### DIFF
--- a/tools/tcpaccept.py
+++ b/tools/tcpaccept.py
@@ -80,10 +80,11 @@ BPF_PERF_OUTPUT(ipv6_events);
 """
 
 #
-# The following is the code for older kernels(Linux pre-4.16).
-# It uses kprobes to instrument inet_csk_accept(). On Linux 4.16 and
-# later, the sock:inet_sock_set_state tracepoint should be used instead, as
-# is done by the code that follows this. 
+# The following code uses kprobes to instrument inet_csk_accept().
+# On Linux 4.16 and later, we could use sock:inet_sock_set_state
+# tracepoint for efficency, but it may output wrong PIDs. This is
+# because sock:inet_sock_set_state may run outside of process context.
+# Hence, we stick to kprobes until we find a proper solution.
 #
 bpf_text_kprobe = """
 int kretprobe__inet_csk_accept(struct pt_regs *ctx)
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
@boat0 Are you referring the cases when retransmit happens?
@brendangregg what do you think?

boat0:
> @boat0 Are you referring the cases when retransmit happens?

TCP state transition may happen in irq context under which
tracepoint sock:inet_sock_set_state could not handle everything correctly.
For example, below is what I captured with `stackcount inet_sk_state_store`:
``` 
  inet_sk_state_store
  tcp_set_state
  tcp_rcv_state_process
  tcp_child_process
  tcp_v4_rcv
  ip_protocol_deliver_rcu
  ip_local_deliver_finish
  ip_local_deliver
  ip_rcv
  __netif_receive_skb_one_core
  netif_receive_skb_internal
  napi_gro_receive
  ixgbe_poll
  net_rx_action
  __softirqentry_text_start
  irq_exit
  do_IRQ
  ret_from_intr
  mwait_idle
  do_idle
  cpu_startup_entry
  start_secondary
  secondary_startup_64
  swapper/10 [0]

yonghong-song:
cc @kernel-z You introduced the tracepoint based mechanism into tcpaccept. Any comment on this?

brendangregg:
Looks good, thanks. This also proves the virtue of doing the two bpf_text_kprobe/bpf_text_tracepoint strings: makes it easy to maintain -- in this case, to remove one of them.

PR #2311, opened 2019-04-12
man: remove duplicated description of '-f' option from offcputime.8

ao2:
In commit 66bf2e8e (offcputime: one symbol cache per process, improve
pid/tid handling, 2016-07-31) the '-f' option was documented again
probably with the intent of following the order of options in the
synopsis, but it was already documented.

Remove the fist instance and keep the new one to follow the same order
of the synopsis, but use the original description.


yonghong-song:
[buildbot, test this please]

PR #2312, opened 2019-04-12
loader: Add support for reading kernel headers from /proc

joelagnel:
BCC right now relies on the filesystem to have kernel headers however
this is quite a hindrance to embedded and Android systems, and even
distros that may not have the kernel headers for the running kernel.
This patch makes use of the new /proc/kheaders.tar.xz archive that is
proposed upstream: https://lore.kernel.org/patchwork/patch/1059427/

The approach involves creating a temporary directory containing the
headers and compiling from there. It is used as a last resort if headers
could not be found by other means.

Testing shows this adds around 400ms to start up time of BCC, however
the cost is 0 if the headers already exists in /lib. For
Android/embedded systems, the cost of 400ms is acceptable considering
otherwise the tools just remain broken. A clang virtual filesystem
approach was suggested on LKML, however such an approach does not add
benefit since most of the time goes in tarball extraction and
decompression which would have to be done in either way.

Signed-off-by: Joel Fernandes (Google) <joel@joelfernandes.org>

### DIFF
--- a/src/cc/frontends/clang/kbuild_helper.cc
+++ b/src/cc/frontends/clang/kbuild_helper.cc
@@ -13,9 +13,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#include <fstream>
+#include <iostream>
+
+#include <dirent.h>
 #include <fcntl.h>
 #include <stdlib.h>
-#include <iostream>
+#include <sys/utsname.h>
+#include <unistd.h>
+
 #include "kbuild_helper.h"
 
 namespace ebpf {
@@ -105,6 +111,84 @@ int KBuildHelper::get_flags(const char *uname_machine, vector<string> *cflags) {
   cflags->push_back("-fno-stack-protector");
 
   return 0;
+}
+
+static inline int file_exists(const char *f)
+{
+  struct stat buffer;
+  return (stat(f, &buffer) == 0);
+}
+
+static inline int proc_kheaders_exists(void)
+{
+  return file_exists(PROC_KHEADERS_PATH);
+}
+
+static inline int extract_kheaders(const std::string &dirpath,
+                                   const struct utsname &uname_data)
+{
+  char tar_cmd[256], dirpath_tmp[256];
+  int ret;
+  bool module = false;
+
+  if (!proc_kheaders_exists()) {
+    ret = system("modprobe kheaders");
+    if (ret)
+      return ret;
+    module = true;
+    if (!proc_kheaders_exists()) {
+      ret = -1;
+      goto cleanup;
+    }
+  }
+
+  snprintf(dirpath_tmp, 256, "/tmp/kheaders-%s-XXXXXX", uname_data.release);
+  if (mkdtemp(dirpath_tmp) == NULL) {
+    ret = -1;
+    goto cleanup;
+  }
+
+  snprintf(tar_cmd, 256, "tar -xf %s -C %s", PROC_KHEADERS_PATH, dirpath_tmp);
+  ret = system(tar_cmd);
+  if (ret) {
+    system(("rm -rf " + std::string(dirpath_tmp)).c_str());
+    goto cleanup;
+  }
+
+  /*
+   * If the new directory exists, it could have raced with a parallel
+   * extraction, in this case just delete the old directory and ignore.
+   */
+  ret = rename(dirpath_tmp, dirpath.c_str());
+  if (ret)
+    ret = system(("rm -rf " + std::string(dirpath_tmp)).c_str());
+
+cleanup:
+  if (module) {
+    int ret1 = system("rmmod kheaders");
+    if (ret1)
+      return ret1;
+  }
+
+  return ret;
+}
+
+int get_proc_kheaders(std::string &dirpath)
+{
+  struct utsname uname_data;
+  char dirpath_tmp[256];
+
+  if (uname(&uname_data))
+    return -errno;
+
+  snprintf(dirpath_tmp, 256, "/tmp/kheaders-%s", uname_data.release);
+  dirpath = std::string(dirpath_tmp);
+
+  if (file_exists(dirpath_tmp))
+    return 0;
+
+  // First time so extract it
+  return extract_kheaders(dirpath, uname_data);
 }
 
 }  // namespace ebpf
--- a/src/cc/frontends/clang/kbuild_helper.h
+++ b/src/cc/frontends/clang/kbuild_helper.h
@@ -21,6 +21,8 @@
 #include <errno.h>
 #include <ftw.h>
 
+#define PROC_KHEADERS_PATH "/proc/kheaders.tar.xz"
+
 namespace ebpf {
 
 struct FileDeleter {
@@ -101,4 +103,5 @@ class KBuildHelper {
   bool has_source_dir_;
 };
 
+int get_proc_kheaders(std::string &dir);
 }  // namespace ebpf
--- a/src/cc/frontends/clang/loader.cc
+++ b/src/cc/frontends/clang/loader.cc
@@ -120,6 +120,7 @@ int ClangLoader::parse(unique_ptr<llvm::Module> *mod, TableStorage &ts,
   const char *version_override = ::getenv("BCC_LINUX_VERSION_CODE");
   bool has_kpath_source = false;
   string vmacro;
+  std::string tmpdir;
 
   if (kpath_env) {
     kpath = string(kpath_env);
@@ -128,6 +129,13 @@ int ClangLoader::parse(unique_ptr<llvm::Module> *mod, TableStorage &ts,
     auto kernel_path_info = get_kernel_path_info(kdir);
     has_kpath_source = kernel_path_info.first;
     kpath = kdir + "/" + kernel_path_info.second;
+  }
+
+  // If all attempts to obtain kheaders fail, check for /proc/kheaders.tar.xz
+  if (!is_dir(kpath)) {
+    int ret = get_proc_kheaders(tmpdir);
+    if (!ret)
+      kpath = tmpdir;
   }
 
   if (flags_ & DEBUG_PREPROCESSOR)


yonghong-song:
[buildbot, test this please]

src/cc/frontends/clang/kbuild_helper.cc
```diff
@@ -107,4 +107,54 @@ int KBuildHelper::get_flags(const char *uname_machine, vector<string> *cflags) {
...
+  if (!proc_kheaders_exists()) {
+    ret = system("modprobe kheaders");
+    if (ret)
+      return ret;
+    if (!proc_kheaders_exists())
+      return -1;
+    module = true;
+  }
+
+  ret = system("rm -rf /tmp/proc-headers-*");
```
4ast: what happens when multiple bcc scripts run at the same time? a race to extract/rm ?
shouldn't all bcc scripts reuse single dir?

joelagnel: yes you're right, I missed that. They could race. I will try such scenarios and fix, thanks

joelagnel:
Sorry about the libbpf submodule change in my PR, will fix that.

src/cc/frontends/clang/kbuild_helper.cc
joelagnel: I will come with an approach that tries to reuse a single directory, but in the worst case could have  different directories if there are multiple processes trying to do extraction at the same time. In the most common case, only a single directory will be used even for racing tasks. Since its only a single /tmp directory now, clean up will be left to be handled automatically by the system (such as reboot).  Does that sound Ok to you?

joelagnel:
@4ast I address all the comments you had, please take a look, thanks

joelagnel:
Corresponding kernel patch is here:
https://lkml.org/lkml/2019/4/12/847
Masahiro Yamada provided a Reviewed-by tag for the latest version ( v7 ).

src/cc/frontends/clang/kbuild_helper.cc
```diff
@@ -107,4 +109,78 @@ int KBuildHelper::get_flags(const char *uname_machine, vector<string> *cflags) {
...
+  }
+  return ret;
+}
+
+int get_proc_kheaders(std::string &dirpath)
+{
+  // Check if kheaders already exist and reuse it
+  DIR *dir;
+  struct dirent *ent;
+  if ((dir = opendir("/tmp/")) != NULL) {
```
4ast: that doesn't work for systems with multiple kernels and corresponding kernel headers installed.
The host may reboot into a different kernel and wrong headers will be picked up.
I suggest to use a fingerprint of /proc/kheaders.tar.xz and extract it into fixed location like:
/tmp/proc-kheaders-sha1(kheaders.txz)/

joelagnel: I can do that, but my thought was the /tmp directory is cleaned up after a reboot anyway and if a new kernel is installed, then a reboot is needed. The clearing may not be the default though on all systems...

I can do the md5 approach you suggested.. Thanks.

4ast: bcc is already using sha1 for such fingerprinting purpose. See bpf_prog_compute_tag.
Another alternative is to use uname -r plus some other property of kheaders file as part of the dir.

joelagnel: When trying to use the code in `bpf_prog_compute_tag`, I get:
```
sha1 socket not available Address family not supported by protocol
```
I am worried this is due to missing kernel support and will break on Android.  We do have md5sum binary available on the filesystem, so I will just call that binary. md5sum is pretty universal in my experience.

joelagnel:
Updated PR with using md5sum to calculate hash of /proc/kheaders. Verified that booting into a new kernel created a new extraction, and not booting into new one reused the previously extracted one.

I did not post a v8 of the [kernel patch](https://lkml.org/lkml/2019/4/12/847) because the only thing for v8 is to add Masahiro's Reviewed-by tag which he previous gave.
If you want me to repost it with his tag, please let me know. (I am assuming you are going to be pulling it). Thanks a lot.

src/cc/frontends/clang/kbuild_helper.cc
4ast: and how subsequent bcc run will know the directory name? every time it tries to compile bpf prog it will do md5sum of /proc/kheader.tar.xz ? I think we need a cheaper way to construct dirname.

joelagnel: Subsequent run does a strstr search of kheaders-<hash> in /tmp. The file being hashed is in memory, md5sum is very fast in my tests - just 14-15ms ; which is a negligible cost.

4ast: > file being hashed is in memory

don't quite follow.
when /tmp/kheaders-TBD is present bcc shouldn't be loading kmod, reading /proc/kheader and doing md5sum on it. modprobe is not a cheap operation.
May be just uname -r is enough.
Or uname -r plus crc of /proc/config.gz That should give stable dir without modprobe of kheaders

joelagnel: Ok. I agree that's better. I will rework patches accordingly and resend.

anakryiko: Instead of relying on marker file ("/extracted") and doing reading /tmp (which can be pretty big, my devmachine has 8500 files right now, and it's not the max), why not atomically rename "kheaders-<sha>-<rndsuffix>" into a "canonical" "kheaders-<sha>" folder. Then the presence of this well-defined directory would be a marker of success. And then even in case of race, we'll end up still having just one directory with good contents. Do you think it will work?

joelagnel: I don't see a benefit of such approach , tbh, my current approach works so I'm happy with it. You could post a follow up patch if you want with your approach to discuss your idea, once my patch is merged.

src/cc/frontends/clang/kbuild_helper.cc
```diff
@@ -107,4 +113,107 @@ int KBuildHelper::get_flags(const char *uname_machine, vector<string> *cflags) {
...
+    if (!proc_kheaders_exists())
+      return -1;
+    module = true;
+  }
+
+  snprintf(dirpath1, 256, "/tmp/kheaders-%s-XXXXXX", hash.c_str());
+  dirpath = std::string(dirpath1);
+
+  if (mkdtemp((char *)dirpath.data()) == NULL)
+    return -1;
```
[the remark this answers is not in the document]
joelagnel: thanks, fixed

joelagnel:
@4ast implemented with uname suggestion, all looks good with my testing. Please take a look, thanks

src/cc/frontends/clang/kbuild_helper.cc
anakryiko: As I see it, your current approach has two downsides:
1. It requires iterating over all contents of /tmp directory, which might be slow, depending on the state of /tmp on user's host. It's also unnecessarily pollutes kernel's dentry/inode cache. And it's just a bunch of code that can be eliminated.
2. I know you are aware, as you mentioned this possibility above, but it still has races and can leave multiple directories with uncompressed headers. By doing atomic rename, there is still race, but it won't leave extra directories lying around.
Both seem to be pretty clear disadvantages, which could be easily avoided.

joelagnel: Ok I think I know what you mean. But it still has the disadvantage that during a race, there may be 2 rename attempts and the failure of one of them will have to be special-handled. I revised these patches 4-5 times and I am currently approaching a burn out :) I prefer any future such extensions as you propose, can be worked on top of this patch, since I don't see the drawbacks you mentioned as anything major. The race is quite rare, and in the such an event, the only cost is an extra extraction. So I don't see a big benefit. I don't mind seeing if I can work on that as a follow-up patch (unless you can), but I don't see the benefit of doing it in this patch other than being for bike-shedding.

@4ast are you Ok with this patch now as-is ?

4ast: @joelagnel I'm confused why mktemp and readdir are still in the patch.
I thought we agreed on stable name:
/tmp/kheaders-`uname -r`-`crc of /proc/config.gz`

joelagnel: @4ast because if 2 instances of a tool are running, then they may try to extract at the same time to same directory and can cause issues. The parallel extractions need to happen in their own tmp directories. The only way around that is to use a lock file or something.. does that make sense?

Also I prefer to just stick to /tmp/kheaders-uname and not to take CRC of /proc/config.gz because that may break if IKCONFIG is disabled in the .config. Let me know if you have any objection to anything I said.

4ast: to solve two parallel untars how about let it extract into temp directory and then rename into a dir with stable name? If rename fails we just rm that untared dir that lost the race.
if we assume that /tmp is cleaned on every reboot then /tmp/kheaders-uname is fine.

joelagnel: Yes, ok, I will do that. So the plan is mktemp into a directory, and then rename it to a stable one. If the rename fails, then just delete the directory. I will do that then and update the PR, thanks.

joelagnel:
@anakryiko @4ast Implemented based on ideas we discussed in this thread. Looks ok now?

src/cc/frontends/clang/kbuild_helper.cc
```diff
@@ -107,4 +113,81 @@ int KBuildHelper::get_flags(const char *uname_machine, vector<string> *cflags) {
...
+  struct stat buffer;
+  return (stat(f, &buffer) == 0);
+}
+
+static inline int proc_kheaders_exists(void)
+{
+  return file_exists(PROC_KHEADERS_PATH);
+}
+
+static inline int extract_kheaders(std::string &dirpath, struct utsname &uname_data)
```
anakryiko: nit: const std::string& ? same for uname_data

src/cc/frontends/clang/kbuild_helper.cc
```diff
@@ -107,4 +113,81 @@ int KBuildHelper::get_flags(const char *uname_machine, vector<string> *cflags) {
...
+    system(("rm -rf " + std::string(dirpath_tmp)).c_str());
+    goto cleanup;
+  }
+
+  /*
+   * If the new directory exists, it could have raced with a parallel
+   * extraction, in this case just delete the old directory and ignore.
+   */
+  ret = rename(dirpath_tmp, dirpath.c_str());
+  if (ret)
```
anakryiko: if ret is ENOTEMPTY or EEXIST then we shouldn't return error from extract_kheaders, because kernel headers were actually successfully extracted, right? Otherwise loader won't use those headers in loader.cc:137 below.

src/cc/frontends/clang/kbuild_helper.h
```diff
@@ -101,4 +103,5 @@ class KBuildHelper {
   bool has_source_dir_;
 };
 
+int get_proc_kheaders(std::string &dir);
```
anakryiko: const?

src/cc/frontends/clang/kbuild_helper.cc
joelagnel: No. ret is overwritten by the following rm, see below

src/cc/frontends/clang/kbuild_helper.cc
joelagnel: ok

src/cc/frontends/clang/kbuild_helper.h
joelagnel: no, this function writes to dir

joelagnel:
Update extract_kheaders arguments to const

src/cc/frontends/clang/kbuild_helper.cc
anakryiko: Ah, you are right, sorry, I missed that.

src/cc/frontends/clang/kbuild_helper.h
anakryiko: oh, it's output parameter, ignore me :)

src/cc/frontends/clang/kbuild_helper.h
```diff
@@ -21,6 +21,8 @@
 #include <errno.h>
 #include <ftw.h>
 
+#define PROC_KHEADERS_PATH "/proc/kheaders.tar.xz"
```
4ast: I thought lkml requested /sys/kernel/tracing/ instead?
The rest looks good.
let's land it when that bit is certain.

joelagnel: I think there was no agreement on that and any location seemed ok. How do we resolve this decision?

yonghong-song: From what I read, to relieve Olof's concerns, /sys/ directory (not sure about /sys/kernel/tracing) is preferred? It looks a a few other folks like Steve is also on board on this.

joelagnel: Ok. It's a confusing discussion. I guess will target sys/kernel/ and resend then.. wdyt? Greg kh was on board with that. thanks.

joelagnel: Since Greg gave a reviewed by tag, I think he's onboard with proc as well.  What do you guys prefer? If proc, is ok please give Ack or Reviewed-by tag on lkml posting. Thanks!

yonghong-song:
Sounds good. Let me do a test and then if everything is okay, will merge it.

yonghong-song:
[buildbot, test this please]

PR #2313, opened 2019-04-12
stackcount: do per-pid statistics only if -P is provided

### LINKED ISSUE #2308: stackcount default output shouldn't do -P
This is annoying -- stackcount by default does per-process breakdowns:

```
# stackcount kmem_cache_alloc
Tracing 1 functions for "kmem_cache_alloc"... Hit Ctrl-C to end.
[...]

  kmem_cache_alloc
  user_path_at_empty
  vfs_statx
  SYSC_newstat
  SyS_newstat
  do_syscall_64
  __xstat
  [unknown]
    python3 [15243]
    1026
```

See the "python3 [15243]" line? Most of the time I don't want this -- I'd rather just group on the stack trace, to reduce the output. Sometimes I want per-process, but only sometimes, and stackcount already has this:

```
  -P, --perpid          display stacks separately for each process
```

My guess is that -P is broken, and has become the default behavior by mistake.

Could someone please look into fixing this: eliminate the per-process breakdown by default, and only do it if -P is used.

boat0:
Fixes #2308.

### DIFF
--- a/man/man8/stackcount.8
+++ b/man/man8/stackcount.8
@@ -2,8 +2,8 @@
 .SH NAME
 stackcount \- Count function calls and their stack traces. Uses Linux eBPF/bcc.
 .SH SYNOPSIS
-.B stackcount [\-h] [\-p PID] [\-c CPU] [\-i INTERVAL] [\-D DURATION] [\-T] [\-r]
-              [\-s] [\-P] [\-K] [\-U] [\-v] [\-d] [\-f] pattern
+.B stackcount [\-h] [\-p PID] [\-c CPU] [\-i INTERVAL] [\-D DURATION] [\-T]
+              [\-r] [\-s] [\-P] [\-K] [\-U] [\-v] [\-d] [\-f] [\-\-debug] pattern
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
Do we need to adjust examples and man pages?

brendangregg:
Thanks! The quickest fix for the examples is to add -P to all the existing examples, so we don't have to change the output.

If you look at the git history for stackcount_example.txt, my first commit has examples that can be included again: examples without -P.

boat0:
Yes. Examples and man pages need some updates too.

boat0:
> Thanks! The quickest fix for the examples is to add -P to all the existing examples, so we don't have to change the output.

-P added to all but two examples with "-K".

> If you look at the git history for stackcount_example.txt, my first commit has examples that can be included again: examples without -P.

Done. That example is restored to be 1st one.

yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks!

PR #2314, opened 2019-04-14
Support loading elf notes from memfd backed map

dalehamel:
This allows for elf notes to be read from a memfd_backed "file".

If the inode number of a file descriptor matches exactly the inode number in the memory map, we know that it is a memory backed file descriptor (this can also be verified against `name` in this context, using `strstr(name, '/memfd')`, but this is less performant than comparing the inode number).

For such file descriptors, we can return a path that is relative to the pid and use that, as `/memfd:NAME` is not a file that can be checked later on. In such cases, we will return a path like `/proc/PID/fd/FDNUM`, where FDNUM corresponds to the file descriptor with that matches the inode number.

The practical application for this is that probes that are dynamically created via a library such as libstapsdt can be scoped to the lifetime of the process. This ensures that no `.so` files which were dlopen'd to load probes into the execution context will be dangling. It also has the benefit of not ever needing to touch even a temporary filesystem, as the file descriptor will always point to memory.

This allows for applications like bpftrace to have a behavior more similar to dtrace, where the elf notes can be mapped in with valloc and made accessible to dtrace.

Eg, tplist output for probes discovered this way:

```
/proc/4732/fd/7 provider:name
```

I have an example of a use-case for this up on https://github.com/sthima/libstapsdt/pull/24

### DIFF
--- a/src/cc/bcc_proc.c
+++ b/src/cc/bcc_proc.c
@@ -14,10 +14,8 @@
  * limitations under the License.
  */
 
-#include <sys/types.h>
-#include <sys/stat.h>
-#include <sys/mman.h>
 #include <ctype.h>
+#include <dirent.h>
 #include <fcntl.h>
 #include <limits.h>
 #include <math.h>
@@ -26,6 +24,9 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <sys/types.h>
 #include <unistd.h>
 
 #include "bcc_perf_map.h"
@@ -82,6 +83,40 @@ int bcc_mapping_is_file_backed(const char *mapname) {
     STARTS_WITH(mapname, "[vsyscall]"));
 }
 
+/*
+Finds a file descriptor for a given inode if it's a memory-backed fd.
+*/
+static char *_procutils_memfd_path(int pid, uint64_t inum) {
+  char path_buffer[PATH_MAX + 1];
+  char *path = NULL;
+  char *dirstr;
+  DIR *dirstream;
+  struct stat sb;
+  struct dirent *dent;
+
+  snprintf(path_buffer, (PATH_MAX + 1), "/proc/%d/fd", pid);
+  dirstr = malloc(sizeof(char) * (strlen(path_buffer) + 1));
+  strcpy(dirstr, path_buffer);
+  dirstream = opendir(dirstr);
+
+  while (path == NULL && dirstream != NULL &&
+         (dent = readdir(dirstream)) != NULL) {
+    snprintf(path_buffer, (PATH_MAX + 1), "%s/%s", dirstr, dent->d_name);
+    if (stat(path_buffer, &sb) == -1)
+      continue;
+
+    if (sb.st_ino == inum) {
+      char *pid_fd_path = malloc(sizeof(char) * (strlen(path_buffer) + 1));
+      strcpy(pid_fd_path, path_buffer);
+      path = pid_fd_path;
+    }
+  }
+  closedir(dirstream);
+  free(dirstr);
+
+  return path;
+}
+
 int bcc_procutils_each_module(int pid, bcc_procutils_modulecb callback,
                               void *payload) {
   char procmap_filename[128];
@@ -111,6 +146,15 @@ int bcc_procutils_each_module(int pid, bcc_procutils_modulecb callback,
       name++;
     if (!bcc_mapping_is_file_backed(name))
       continue;
+
+    if (strstr(name, "/memfd:")) {
+      char *memfd_name = _procutils_memfd_path(pid, inode);
+      if (memfd_name != NULL) {
+        strcpy(buf, memfd_name);
+        free(memfd_name);
+        name = buf;
+      }
+    }
 
     if (callback(name, begin, end, (uint64_t)offset, true, payload) < 0)
       break;
--- a/src/cc/bcc_proc.c
+++ b/src/cc/bcc_proc.c
@@ -86,7 +86,7 @@ int bcc_mapping_is_file_backed(const char *mapname) {
 /*
 Finds a file descriptor for a given inode if it's a memory-backed fd.
 */
-static char *_procutils_memfd_path(int pid, uint64_t inum) {
+static char *_procutils_memfd_path(const int pid, const uint64_t inum) {
   char path_buffer[PATH_MAX + 1];
   char *path = NULL;
   char *dirstr;
--- a/src/cc/bcc_proc.c
+++ b/src/cc/bcc_proc.c
@@ -95,18 +95,20 @@ static char *_procutils_memfd_path(const int pid, const uint64_t inum) {
   struct dirent *dent;
 
   snprintf(path_buffer, (PATH_MAX + 1), "/proc/%d/fd", pid);
-  dirstr = malloc(sizeof(char) * (strlen(path_buffer) + 1));
+  dirstr = malloc(strlen(path_buffer) + 1);
   strcpy(dirstr, path_buffer);
   dirstream = opendir(dirstr);
 
-  while (path == NULL && dirstream != NULL &&
-         (dent = readdir(dirstream)) != NULL) {
+  if (dirstream == NULL)
+    return NULL;
+
+  while (path == NULL && (dent = readdir(dirstream)) != NULL) {
     snprintf(path_buffer, (PATH_MAX + 1), "%s/%s", dirstr, dent->d_name);
     if (stat(path_buffer, &sb) == -1)
       continue;
 
     if (sb.st_ino == inum) {
-      char *pid_fd_path = malloc(sizeof(char) * (strlen(path_buffer) + 1));
+      char *pid_fd_path = malloc(strlen(path_buffer) + 1);
       strcpy(pid_fd_path, path_buffer);
       path = pid_fd_path;
     }


src/cc/bcc_proc.c
```diff
@@ -82,6 +83,43 @@ int bcc_mapping_is_file_backed(const char *mapname) {
...
+  char glob_buffer[1024];  // a ridiculous buffer size
+  int glob_size, i;
+  glob_t globbuf;
+  struct stat sb;
+
+  snprintf(glob_buffer, 1024, "/proc/%d/fd/*", pid);
+  glob_size = strlen(glob_buffer);
+
+  globstr = malloc(sizeof(char) * (glob_size + 1));
+  strcpy(globstr, glob_buffer);
```
palmtenor: Can we use `opendir` / `readdir` instead of a glob search?

src/cc/bcc_proc.c
```diff
@@ -82,6 +83,43 @@ int bcc_mapping_is_file_backed(const char *mapname) {
...
+  glob(globstr, GLOB_NOSORT, NULL, &globbuf);
+  free(globstr);
+
+  for ((i = 0 && path == NULL); i < globbuf.gl_pathc; i++) {
+    char *pid_fd_path = globbuf.gl_pathv[i];
+    if (stat(pid_fd_path, &sb) == -1) {
+      printf("Error calling lstat on %s\n", pid_fd_path);
+    }
+
+    printf("inode: %ld %ld\n", sb.st_ino, inum);
```
palmtenor: For error we should output to stderr and this informative info shouldn’t be printed by default

dalehamel: Yeah i've already removed this, accidentally pushed it to the draft PR

src/cc/bcc_proc.c
dalehamel: I'll take a look :+1:

dalehamel: done

src/cc/bcc_proc.c
```diff
@@ -82,6 +84,48 @@ int bcc_mapping_is_file_backed(const char *mapname) {
...
+         (dent = readdir(dirstream)) != NULL) {
+    char *pid_fd_path;
+    pid_fd_path =
+        calloc(sizeof(char), strlen(dirstr) + strlen(dent->d_name) + 1);
+    strcat(pid_fd_path, dirstr);
+    strcat(pid_fd_path, dent->d_name);
+
+    if (stat(pid_fd_path, &sb) == -1) {
+      free(pid_fd_path);
+      continue;
```
dalehamel: I decided there is no point in printing an error/warning if we can't stat an inode, we'll just return NULL and let the caller handle it i we don't end up finding what we need. No need to print errors for fds that don't matter.

dalehamel:
@palmtenor if you have a moment, I think this is ready for a review now (sorry for having this up earlier when it wasn't ready, I had hoped it'd be ignored as a draft - just wanted to use the github diff viewer).

I'm not sure what to do for automated tests here, as I think I'd need to pull in libstapsdt to write the elf notes for me in the test suite, as the current usdt tests are all based on macro-defined (and not dlopen'd) probes.

As for manual testing, I have verified this myself using a branch of libstapsdt that I'm still working on, I can push that if you'd like to run a test yourself.

src/cc/bcc_proc.c
```diff
@@ -14,18 +14,20 @@
...
 #include <fcntl.h>
+#include <glob.h>
 #include <limits.h>
 #include <math.h>
 #include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/mman.h>
```
dalehamel: clang-format-diff decided to move these, so I let it.

dalehamel:
I pushed up a PR to libstapsdt that shows how this functionality can be used. https://github.com/sthima/libstapsdt/pull/24

yonghong-song:
[buildbot, test this please]

dalehamel:
```
bcc-pr » ubuntu1804 completed with result SUCCESS
bcc-pr » ubuntu1604 completed with result SUCCESS
bcc-pr » ubuntu1710 completed with result SUCCESS
bcc-pr » fc25 completed with result SUCCESS
bcc-pr » fc26 completed with result FAILURE
bcc-pr » fc28 completed with result SUCCESS
```

Is the build flakey? I can't seem to see the build output further than this, but I'm not sure what reason there could be for this selectively failing on fedora 26.

yonghong-song:
[buildbot, test this please]

yonghong-song:
just started another test. the test itself could be flaky. It is a networking related test depends on udp packets and udp packets may get lost for whatever reason.

src/cc/bcc_proc.c
```diff
@@ -82,6 +83,40 @@ int bcc_mapping_is_file_backed(const char *mapname) {
...
+static char *_procutils_memfd_path(const int pid, const uint64_t inum) {
+  char path_buffer[PATH_MAX + 1];
+  char *path = NULL;
+  char *dirstr;
+  DIR *dirstream;
+  struct stat sb;
+  struct dirent *dent;
+
+  snprintf(path_buffer, (PATH_MAX + 1), "/proc/%d/fd", pid);
+  dirstr = malloc(sizeof(char) * (strlen(path_buffer) + 1));
```
yonghong-song: nit: sizeof(char) seems unnecessary. The same as below.

src/cc/bcc_proc.c
```diff
@@ -82,6 +83,40 @@ int bcc_mapping_is_file_backed(const char *mapname) {
...
+  strcpy(dirstr, path_buffer);
+  dirstream = opendir(dirstr);
```
yonghong-song: maybe you want to test dirstream is NULL or not here? If it is NULL, you can just return NULL. This seems more cleaner then below trying to test `dirstream != NULL` in every while loop iteration.

dalehamel: I agree

src/cc/bcc_proc.c
dalehamel: Will do

dalehamel:
@yonghong-song I believe I've addressed your comments

yonghong-song:
[buildbot, test this please]

palmtenor:
@dalehamel I have a question for you. Do you know if these files (`memfd:...`) are OK to read from the `procfs` symlink, i.e., `/proc/PID/map_files/xxxxx-yyyyy`? I'm planning to change the PID-only USDT path to use these symlinks for easier handling of mount namespace and I'm WIP. But I don't have an environment to test these `memfd` files?

src/cc/bcc_proc.c
```diff
@@ -82,6 +83,42 @@ int bcc_mapping_is_file_backed(const char *mapname) {
...
+  struct stat sb;
+  struct dirent *dent;
+
+  snprintf(path_buffer, (PATH_MAX + 1), "/proc/%d/fd", pid);
+  dirstr = malloc(strlen(path_buffer) + 1);
+  strcpy(dirstr, path_buffer);
+  dirstream = opendir(dirstr);
+
+  if (dirstream == NULL)
+    return NULL;
```
palmtenor: We didn't free `dirstr` in this error path right? Just furious why do we need copying of the string:)

dalehamel: nice catch - this is definitely a memory leak. I missed it when I added the handling that @yonghong-song suggested of returning early. I'll get a quickfix up for that.

> why do we need copying of the string:)
The string is copied here so that it can be used to build the path that we return a heap-allocated reference to. Would it be better to use a local buffer to construct the directory path, and only malloc for the string we want to return? If so, I can include this change in my PR. I had considered it, but decided to just reuse the buffer - which is why we are copying here.

dalehamel:
> I have a question for you. Do you know if these files (memfd:...) are OK to read from the procfs symlink, i.e., /proc/PID/map_files/xxxxx-yyyyy? I'm planning to change the PID-only USDT path to use these symlinks for easier handling of mount namespace and I'm WIP. But I don't have an environment to test these memfd files?

@palmtenor if by OK you mean possible, then the answer appears to be yes, as I am able to read the elf-notes that I created in my example from there:

```
7f0df7579000-7f0df757a000 r-xp 00000000 00:05 19991634                   /memfd:libstapsdt:global (deleted)
7f0df757a000-7f0df7779000 ---p 00001000 00:05 19991634                   /memfd:libstapsdt:global (deleted)
7f0df7779000-7f0df777a000 rw-p 00000000 00:05 19991634                   /memfd:libstapsdt:global (deleted)
```

These all succeed, and produce the same output:

```
readelf --notes /proc/23729/map_files/7f0df7579000-7f0df757a000
readelf --notes /proc/23729/map_files/7f0df757a000-7f0df7779000
readelf --notes /proc/23729/map_files/7f0df7779000-7f0df777a000
```

```

Displaying notes found in: .note.stapsdt
  Owner                 Data size       Description
  stapsdt              0x00000039       NT_STAPSDT (SystemTap probe descriptors)
    Provider: global
    Name: hello_nsec
    Location: 0x0000000000000280, Base: 0x0000000000000340, Semaphore: 0x0000000000000000
    Arguments: 8@%rdi -8@%rsi
  stapsdt              0x0000002e       NT_STAPSDT (SystemTap probe descriptors)
    Provider: global
    Name: enabled
    Location: 0x0000000000000285, Base: 0x0000000000000340, Semaphore: 0x0000000000000000
    Arguments: 8@%rdi
```

If by OK you mean safe, I can't think of anything unsafe here, the only potential race I can think of is around the state of this file descriptor. Unlike normal file descriptors, closing a `memfd` one will result in the file being lost, as there is no reference to it on the filesystem. IE, closing has the same behavior as unlinking.

> But I don't have an environment to test these memfd files?

If I understand what you mean here I suspect you could probably set one up easily, the only requirements I'm aware of are kernel 3.17+ and glibc 2.27+. The patch I linked (https://github.com/sthima/libstapsdt/pull/24) shows the library I have been testing with.

PR #2319, opened 2019-04-17
Fixed "TypeError: cannot use a string pattern on a bytes-like object"…

palao:
… in some utilities

### DIFF
--- a/src/python/bcc/table.py
+++ b/src/python/bcc/table.py
@@ -622,7 +622,7 @@ class PerfEventArray(ArrayBase):
         i = 0
         while i < num_fields:
             field = lib.bpf_perf_event_field(self.bpf.module, self._name, i).decode()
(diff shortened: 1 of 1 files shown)


src/python/bcc/table.py
```diff
@@ -622,7 +622,7 @@ def _get_event_class(self):
         i = 0
         while i < num_fields:
             field = lib.bpf_perf_event_field(self.bpf.module, self._name, i).decode()
```
yonghong-song: looks like field has been decoded here?

palao: Sorry. My bad.
I thought I had checked against the current repo, but I was testing against Gentoo's stable...

palmtenor:
[buidlbot, ok to test]

yonghong-song:
The issue has been fixed in latest bcc.

PR #2320, opened 2019-04-18
Fix possible memory leak in error path when elf notes are backed by a memfd path

dalehamel:
As @palmtenor points out, https://github.com/iovisor/bcc/pull/2314#discussion_r276495408, the original PR misses freeing the memory that it allocates for a string on the error path (where the directory listing fails) resulting in a potential memory leak.

I had considered just using a separate buffer for the directory string, but decided to malloc it so I could reuse the buffer for building the full path while still retaining access to the string. If there's a reason to prefer that approach I can modify this PR according, which would eliminate the need to bother with freeing this string buffer as it would be stack-allocated instead.

### DIFF
--- a/src/cc/bcc_proc.c
+++ b/src/cc/bcc_proc.c
@@ -99,8 +99,10 @@ static char *_procutils_memfd_path(const int pid, const uint64_t inum) {
   strcpy(dirstr, path_buffer);
   dirstream = opendir(dirstr);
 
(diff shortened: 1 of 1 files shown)


palmtenor:
[buildbot, ok to test]

yonghong-song:
[buildbot, test this please]

PR #2322, opened 2019-04-18
tools: don't mix print(end="") with printb()

lcp:
While mixing print(end="") with printb(), some messages may miss due to
the underlying buffer handling in python 3.

For example:

  # python3 opensnoop.py
  /proc/18849/cmdline
  4109   tmux: server       67   0 /proc/18849/cmdline
  4109   tmux: server       67   0 /proc/18849/cmdline
  4109   tmux: server       67   0 /proc/18849/cmdline

The PID, COMM, FD, and ERR are printed with print(end=""), and those of
the first instance was eaten by printb() which outputs PATH.

The following scripts mix print(end="") and printb() for the same line:

tools/execsnoop.py
tools/opensnoop.py
tools/tcpaccept.py
tools/tcpconnect.py

Those scripts work fine with python 2 but some messages may miss while
using python 3.

This commit converts print(end="") to printb(nl="") to avoid the
inconsistent outputs.

### DIFF
--- a/tools/execsnoop.py
+++ b/tools/execsnoop.py
@@ -216,7 +216,7 @@ def print_event(cpu, data, size):
 
         if not skip:
             if args.timestamp:
-                print("%-8.3f" % (time.time() - start_ts), end="")
+                printb(b"%-8.3f" % (time.time() - start_ts), nl="")
             ppid = event.ppid if event.ppid > 0 else get_ppid(event.pid)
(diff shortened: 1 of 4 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

PR #2324, opened 2019-04-19
Use /proc/[pid]/root when looking at /proc/[pid]/maps paths

davemarchevsky:
I recently ran into a problem when trying to attach to a USDT in a shared library. Specifically, the shared library was loaded into the address space of a chrooted process. If I just provide the `pid` of this process, initializing `USDT` object fails because all paths in `/proc/[pid]/maps` for the target are relative to its chroot, while the process I'm trying to attach from is not in a chroot and therefore the paths don't make sense and `USDT` initialization can't look through shared lib ELFs to find the right probe.

Similarly, if I try to initialize the `USDT` object with `pid` and `bin_path` to the shared lib, initialization succeeds but actually attaching fails: address lookup for the usdt's semaphore in `pid`'s address space fails because `/proc/[pid]/maps` shared lib path is expected to match provided `bin_path` exactly, but maps paths are chroot-relative so this doesn't work.

Implementation evolved significantly since this PR was created, look at [this comment](https://github.com/iovisor/bcc/pull/2324#issuecomment-494169782) for details about impl.


palmtenor:
[buildbot, ok to test]

yonghong-song:
the test failed due to an infra error. Let us try again.

yonghong-song:
[buildbot, test this please]

yonghong-song:
@palmtenor any comment for this pull request?

yonghong-song:
@vijunag could you take a look as well?

src/cc/usdt/usdt.cc
```diff
@@ -238,12 +238,18 @@ void Context::_each_probe(const char *binpath, const struct bcc_elf_usdt *probe,
...
+  if (ctx->pid_) {
+    path = tfm::format("/proc/%d/root%s", *ctx->pid_, path);
+  }
+
   // Modules may be reported multiple times if they contain more than one
   // executable region. We are going to parse the ELF on disk anyway, so we
   // don't need these duplicates.
-  if (ctx->modules_.insert(modpath).second /*inserted new?*/) {
+  if (ctx->modules_.insert(path).second /*inserted new?*/) {
     ProcMountNSGuard g(ctx->mount_ns_instance_.get());
```
vijunag: PID namespaces corresponding to the prefixed paths are now different since you've moved to the process namespace ?

davemarchevsky: I'm not sure I understand this comment. Are saying that this NSGuard will alter the PID namespace of the process running this code to be this PID namespace of the target pid? If so, I don't think this will happen, since we're [`setns`ing with `CLONE_NEWNS` here](https://github.com/iovisor/bcc/blob/master/src/cc/ns_guard.cc#L70). 

That being said, I plan on removing all `ProcMountNSGuard`s from `usdt.cc` as part of this PR, just haven't pushed that up yet.

davemarchevsky: Ah, I see what you mean. Different mount namespace may have different `/proc` mounted. Guess that makes removing NSGuards a necessity.

yonghong-song:
@palmtenor what do you think about this approach?

yonghong-song:
@davemarchevsky related to MATCH_FILENAME, could you have false positives? Or because this is shared library and we should just proceed with library-name-only comparision even with potentially incorrect user full binary path. 

I do not see any issues with MATCH_FILENAME if user only specifies library name like abbr. `c` or `m` in command line. But if user specify a full name, we could still compare name only but after this do a partial comparison to the original path and print an error if it does not partial match based on chroot convention?

yonghong-song:
Agree with @vijunag, but changing path, with NSGuard, we do risk going into a different mount namespace then intended, so this change may require NSGuard becomes noop, right? I have not done any experiments to show this is possible or not.

yonghong-song:
I did a little bit experiment. For USDT, suppose that we have
   . process A is running inside a container with process ID 456 with local path
     /tmp/a.out and chroot path /var/local/containers/container1/.
   . process A is also visible outside the container as process ID 123456.

If we supply the process ID 123456 and shared binary path displayed by /proc/123456/maps,
e.g., /var/local/containers/container1/usr/lib64/libc.so.6, we should be okay for USDT, right?
All the probes and semaphore addresses can be obtained with process ID 123456 and binary path (also in /proc/123456/maps) /var/local/containers/container1/. Traversing /proc/123456/maps will get the shared library address range. Together with semaphore address in ELF file (/var/local/containers/container1/usr/lib64/libc.so.6), we can derive the runtime address.

So we do not need NSGuard at all in such cases? Did I miss anything?

@palmtenor

davemarchevsky:
@yonghong-song I think you are correct in saying that if we have `pid` to access `/proc/pid/maps` and know which shared library path we're trying to match against we should be able to get semaphore address. I'm pushing up a removal of NSGuard later today.

Speaking of "match against": 

> related to MATCH_FILENAME, could you have false positives? 

Definitely. I think the default should be `MATCH_EXACT` for this reason. One way of making this less likely that I think you're mentioning here:

> But if user specify a full name, we could still compare name only but after this do a partial comparison to the original path and print an error if it does not partial match based on chroot convention?

Is that we could add a match like `MATCH_SUBPATH` (not best name) such that, if you have a library at `/var/local/containers/container1/usr/lib64/libwhatever.so` and the process you're trying to attach USDT to has chroot at `/var/local/containers/container1`, you could try to do match like `/usr/lib64/libwhatever.so, MATCH_SUBPATH` and since the chrooted path is substr of full path match would succeed. I would imagine that this would reduce false positives relative to `MATCH_EXACT`.

Another thought I had while doing first pass was matching based on `(device, inode)` but that feels too expensive. Could use this as a backup w/ `MATCH_FILENAME` to reduce false positive, perhaps.

yonghong-song:
Regarding to MATCH_SUBPATH, I am still thinking we can either to full path comparison if user provides one, or bcc just derives the path if user does not supply one. This seems cleaner and error prone. But I will see your patch first. Could you add some examples to show how it is tested so we can check usability as well? Thanks!

vijunag:
@davemarchevsky @yonghong-song - Now is the time to eliminate ProcNsGuard() and any other fix is incomplete. What we need now is PrefixGuard() which will prefix the paths according to the process namespace that way we can keep this infra same across USDT/Symlookups.

yonghong-song:
Just realized that `/tmp/perf-<pid>.map` may need choot path prefix, '/proc/<pid>/root/tmp/perf-<pid>.map` or `<chroot path>/tmp/perf-<pid>.map`.

yonghong-song:
@davemarchevsky Any update on this?

davemarchevsky:
@yonghong-song @vijunag @palmtenor 

I made some changes to incorporate feedback and added some tests. It might be best to consider this an entirely new PR, solving the same problems outlined in my first post. I will edit the first post to make this more clear. 

Here's a summary of the changes:

* Remove `ns_guard` everywhere as suggested
* Modify `bcc_resolve_global_addr` to compare `(inode, device_{major,minor})` instead of filename.
    * `stat` the module path passed in to `bcc_resolve_global_addr` and use inode, device info from stat struct to match against same data from `/proc/PID/maps`
        * Comparing `stat` device nums w/ `/proc/PID/maps` device numbers doesn't work on btrfs - major/minor numbers from stat differ from those in proc maps. An `inode_match_only` kludge was added to work around this. Added `USDT::set_probe_matching_kludge` function to public API, kludge flag needs to propagate `USDT -> Context -> Probe` unfortunately.
    * Functionality to do filename comparison as a backup is retained: if inode `0` is used to search with `_bcc_syms_find_module`, it'll fallback to doing name comparison. Can remove this fallback if desired
* `bcc_procutils_each_module`'s callback changed to add `enter_ns`, which is 1 when iterating `/proc/PID/maps` and 0 in the case where perf map might be outside of chroot. This allows callbacks to know whether they should do anything special to paths that `bcc_procutils_each_module` gives them.
* Since `enter_ns` is added, `Context::_each_module` can do the right thing and prepend `/proc/PID/root` to paths where `enter_ns` is set and leave them alone otherwise. This obviates the need for `ns_guard`. 
* Some refactoring for clarity / testability
* Added some tests to validate "iterate through /proc/PID/maps" functionality, one test to resolve addr in a dynamically linked lib, and one test to init USDT on a probe defined in a shared lib
   * This required making some minor additions to cmake files, which I was entirely unfamiliar with before starting this work, and thus likely messed something up here.

yonghong-song:
[buildbot, test this please]

yonghong-song:
@davemarchevsky there are two test failures, 
```
	 21 - py_uprobes (Failed)
	 32 - py_test_usdt2 (Failed)
```
They are legitimate failures. could you take a look.
Also, could you rebase on top of the master. There is a conflict when applying the change to the trunk.

@palmtenor if you get some time, could you take a look as well?

davemarchevsky:
(rebased, some small additional changes to address rebase conflict, integrating memfd addition)

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

src/cc/CMakeLists.txt
```diff
@@ -47,11 +47,11 @@ set(bcc_common_sources ${bcc_common_sources} bpf_module_rw_engine_disabled.cc)
 endif()
 
 set(bcc_table_sources table_storage.cc shared_table.cc bpffs_table.cc json_map_decl_visitor.cc)
-set(bcc_util_sources ns_guard.cc common.cc)
-set(bcc_sym_sources bcc_syms.cc bcc_elf.c bcc_perf_map.c bcc_proc.c)
+set(bcc_util_sources common.cc)
+set(bcc_sym_sources bcc_syms.cc bcc_proc.h bcc_elf.c bcc_perf_map.c bcc_proc.c)
```
yonghong-song: We don't put header files here (`bcc_sym_sources`).

src/cc/api/BPF.cc
```diff
@@ -787,14 +792,22 @@ bool USDT::operator==(const USDT& other) const {
          (probe_func_ == other.probe_func_);
 }
 
+int USDT::set_probe_matching_kludge(uint8_t kludge) {
```
yonghong-song: this function is not used?

src/cc/api/BPF.h
```diff
@@ -287,6 +287,8 @@ class USDT {
                << usdt.probe_func_;
   }
 
+  int set_probe_matching_kludge(uint8_t kludge);
```
yonghong-song: this can be removed?

src/cc/bcc_proc.c
```diff
@@ -148,39 +143,66 @@ int bcc_procutils_each_module(int pid, bcc_procutils_modulecb callback,
...
 
     if (strstr(name, "/memfd:")) {
-      char *memfd_name = _procutils_memfd_path(pid, inode);
+      char *memfd_name = _procutils_memfd_path(pid, mod.inode);
       if (memfd_name != NULL) {
         strcpy(buf, memfd_name);
         free(memfd_name);
-        name = buf;
+        mod.name = buf;
+        enter_ns = 0;
```
yonghong-song: The logic for enter_ns is not quite right here. If `enter_ns` here is set to 0, in the next iteration, it will remain 0, is this what you want. Also, for maps in a particular module, should be all files either enter_ns or not enter_ns?

src/cc/bcc_proc.c
```diff
@@ -148,39 +143,66 @@ int bcc_procutils_each_module(int pid, bcc_procutils_modulecb callback,
...
+  char procmap_filename[128];
+  FILE *procmap;
+  snprintf(procmap_filename, sizeof(procmap_filename), "/proc/%ld/maps",
+           (long)pid);
+  procmap = fopen(procmap_filename, "r");
+  if (!procmap)
+    return -1;
+
+  if(!_procfs_maps_each_module(procmap, pid, callback, payload))
+    goto done;
```
yonghong-song: this is different from original implementation. Previously, regardless return value, the below perf-<PID>.map is searched. This is not the case any more?

src/cc/bcc_proc.c
```diff
@@ -148,39 +143,66 @@ int bcc_procutils_each_module(int pid, bcc_procutils_modulecb callback,
...
   // in case it is generated by the process itself.
-  if (bcc_perf_map_path(map_path, sizeof(map_path), pid))
-    if (callback(map_path, 0, -1, 0, true, payload) < 0)
-      return 0;
+  mod_info mod;
+  memset(&mod, 0, sizeof(mod_info));
+  if (bcc_perf_map_path(map_path, sizeof(map_path), pid)) {
+    mod.name = map_path;
+    mod.end_addr = -1;
+    if (callback(&mod, 1, payload) < 0)
```
yonghong-song: Here, `bcc_perf_map_path` may already deal with `/proc/<pid>/root`? So callback function does not need `enter_ns`?

src/cc/bcc_syms.cc
```diff
@@ -547,27 +542,45 @@ int bcc_buildsymcache_resolve(void *resolver,
...
 
-struct mod_st {
-  const char *name;
-  uint64_t start;
-  uint64_t file_offset;
-};
+int _bcc_syms_find_module(mod_info *info, int enter_ns, void *p) {
+  struct mod_search *mod = (struct mod_search *)p;
+  // use inode + dev to determine match if inode set
+  if (mod->inode) {
```
yonghong-song: Agree, if inode information is available, we should try inode comparison first. This will be more robust.

src/cc/usdt/usdt.cc
```diff
@@ -273,6 +279,10 @@ std::string Context::resolve_bin_path(const std::string &bin_path) {
     ::free(which_so);
   }
 
+  if (!result.empty() && pid_) {
+    result = tfm::format("/proc/%d/root%s", *pid_, result);
```
yonghong-song: is this true all the time? Or it is just another try.

src/cc/bcc_proc.c
davemarchevsky: This is definitely not right - intention was to set `enter_ns` just for the iteration where memfd logic is triggered.

Before I dealt w/ rebase conflict due to the memfd logic here, `enter_ns` was true for all maps. When the memfd logic is triggered (`if (strstr(name, "/memfd:"))`), the name is replaced with something like `/proc/PID/fd/10`. I think we don't want to `enter_ns` here because `/proc/PID/fd/10` may not be same file as `/proc/PID/root/proc/PID/fd/10` due to pid namespace?

davemarchevsky: Will modify to only set `enter_ns` to 0 for memfd iteration.

src/cc/api/BPF.cc
davemarchevsky: This is intended to be called by the program after creating a `USDT` object, before the usdt object(s) are passed to `BPF::init`. This way all `USDT::Context`s and `USDT::Probe`s created by the USDT have their `mod_match_inode_only_` members set and the "match inode only, ignore dev" kludge is successfully enabled.

For example, my program which runs on btrfs hosts does the following:
```cpp
std::vector<USDT> usdts;
auto u = USDT(
      pid,
      Provider.str(),
      Name.str(),
      ProbeFuncName);
u.set_probe_matching_kludge(1);
usdts.push_back(u);
bpf_.init(bpfProgramToRun, cflags, usdts);
```

I debated whether it made sense to put this set-kludge option in the `USDT` constructor but decided against it because:
  * Hopefully sometime in future btrfs fixes device id disparity between `stat` and `procfs` and I can remove this logic.
  * This is sufficiently obscure of a corner case that this kludgey setter function felt appropriate versus adding another constructor arg

I don't feel particularly strong about setter vs constructor thing here, so happy to change if you'd like.

src/cc/api/BPF.h
davemarchevsky: addressed this on your other `set_probe_matching_kludge` comment

src/cc/CMakeLists.txt
davemarchevsky: will figure out why i thought this was necessary and remove / adjust

src/cc/api/BPF.cc
yonghong-song: Then could you add some comments for the function. Since this will be part of API, we need to document when to use it.

tests/cc/test_c_api.cc
```diff
@@ -439,6 +439,119 @@ TEST_CASE("resolve symbols using /tmp/perf-pid.map", "[c_api]") {
...
+  uint64_t local_addr = 0x15;
+  uint64_t global_addr;
+
+  struct mod_search search;
+  memset(&search, 0, sizeof(struct mod_search));
+  search.name = sopath;
+
+  int res = bcc_procutils_each_module(pid, _bcc_syms_find_module,
+                                      &search);
+  REQUIRE(search.start != 0);
```
yonghong-song: Should you check `res == 0` here?

tests/cc/usdt_test_lib.c
```diff
@@ -0,0 +1,10 @@
+#ifdef HAVE_SDT_HEADER
+#include <sys/sdt.h>
+
```
yonghong-song: Some of our test machines may not have sys/sdt.h. Maybe you can use the folly headers? Currently, it is in bcc/tests/python/include. But I think you can add a symbolic link at bcc/tests/cc to use it.

src/cc/bcc_proc.c
davemarchevsky: I thought this was the case as well, but it's doing readlink on `/proc/PID/root` and prepending the path of the symlink to the perf map path. Some processes on the hosts I'm working on have `/proc/PID/root` symlink resolve to `/`, but the inode is not the same as the actual root inode.

When I tried modifying this line to do `callback(&mod, 0, payload)`, one of the symcache tests in `test_c_api` broke. Think it's necessary to do `enter_ns` here.

yonghong-song: Okay, the original code indeed entered mount ns.
```
int Context::_each_module(const char *modpath, uint64_t, uint64_t, uint64_t,
                          bool, void *p) {
  Context *ctx = static_cast<Context *>(p);
  // Modules may be reported multiple times if they contain more than one
  // executable region. We are going to parse the ELF on disk anyway, so we
  // don't need these duplicates.
  if (ctx->modules_.insert(modpath).second /*inserted new?*/) {
    ProcMountNSGuard g(ctx->mount_ns_instance_.get());
    bcc_elf_foreach_usdt(modpath, _each_probe, p);
  }
  return 0;
}
```

If you got some time, could you dig a little more to find out why this is needed? It is puzzling.

yonghong-song: cc @palmtenor who may know the answer as well.

palmtenor: This is actually tricky. You don't always want to go with proc root for reading perf map, since they are often only in root /tmp. I remember we try both (with either global PID or NS PID), but I need to take a better look at the code.

davemarchevsky: @palmtenor I think [this code](https://github.com/iovisor/bcc/blob/master/src/cc/bcc_proc.c#L169-L182) is what you are referring to. This PR preserves that functionality.

palmtenor:
The "inode-only" logic is kind of annoying. I guess it's for `btrfs`'s device number issues, am I right?

davemarchevsky:
@palmtenor Indeed, search for 'kludge' in [this comment](https://github.com/iovisor/bcc/pull/2324#issuecomment-494169782) for explanation of the issue. But you know what the issue is 😄 . I tried to avoid making the kludge look like it belongs as a permanent API feature for USDTs and their helper classes with expectation that btrfs issue is fixed eventually.

palmtenor:
Right right... But I'm wondering for this particular case (USDT), maybe we can do better and avoid it. The only place we use it is for resolving argument and semaphore address, which should be just in the initial module that we've already determined at load time. I guess we wouldn't need to do "match" of the module against maps again to look up global address of those symbols? Essentially we just need an offset...

davemarchevsky:
@palmtenor Are you referring to something like:

* When doing `USDT::init`, `Context::init` is called
* This initializes `Probe`s and `Location`s for each `Probe`
* If the `Probe` is in a shared object AND we have a pid, we can figure out all these global addresses at `Context::init` time instead of later, in `BPF::attach_usdt`

If so, I see what you mean, but what about the case where `Probe` doesn't have pid information? I see that probe's `pid_` is `optional`, but that `add_to_semaphore` does an `assert(pid_)`, so ostensibly there's some case where you might want to make a `Probe` without a pid and then set it later? But I don't see a place in current code where this is done.

src/cc/bcc_proc.c
davemarchevsky: this was a mistake on my part, changed to function as before

tests/cc/test_c_api.cc
davemarchevsky: addressed

src/cc/api/BPF.cc
davemarchevsky: comments added

davemarchevsky:
latest commit addresses some comments:
* `enter_ns` logic in `_procfs_maps_each_module` changed to set `enter_ns = 1` every time a new map is looked at, which was the original intent
* comments added explaining the kludge
* missing `res == 0` check added in a test
* `bcc_procutils_each_module` changed back to look at perf maps after looking at proc maps, which is how it worked before

tests/cc/usdt_test_lib.c
davemarchevsky: addressed, and removed `sys/sdt.h` requirement from another test

src/cc/usdt/usdt.cc
davemarchevsky: `Context` can be initialized without a `pid`, and separately from that both `bcc_procutils_which` and `bcc_procutils_which_so` can fail due to diabolical input. Functions that call `Context::resolve_bin_path` rely on its return string to be empty as a signal of "I couldn't find a bin path" so we must only prepend `/proc/PID/root` when there's something to prepend to.

davemarchevsky:
I think the most recent commit brings the tests to a passing state:

```
Total Test time (real) = 452.79 sec
++ head -n 1 Testing/TAG
+ cp Testing/20190612-1828/Test.xml ./CTestResults.xml
INFO: Starting to record.
INFO: Processing CTest-Version 3.x (default)
INFO: [CTest-Version 3.x (default)] - 1 test report file(s) were found with the pattern 'build/CTestResults.xml' relative to '/home/fedora/jenkins/workspace/bcc-pr/label/fc27' for the testing framework 'CTest-Version 3.x (default)'.
INFO: Check 'Failed Tests' threshold.
INFO: Check 'Skipped Tests' threshold.
INFO: Setting the build status to SUCCESS
INFO: Stopping recording.
Finished: SUCCESS
```

@palmtenor How do you feel about me attempting the smarter way of doing this in a followup PR? I think your idea will work but it requires a refactoring pass to clarify when `Probe` needs to know about `pid_`.

I will be submitting some further (and hopefully much smaller) PRs enhancements to USDT stuff to further support our usecase so should be able to do refactoring pass in next twoish months.

yonghong-song:
[buildbot, test this please]

yonghong-song:
@davemarchevsky usdt test in test_libbcc still failed in a few platforms.
For example,
```
3: Event name (p__proc__1_root_proc_6586_root_home_iovisor_jenkins_workspace_bcc_pr_label_ubuntu1710_build_tests_cc_test_libbcc_0x5defc3) is too long for buffer
```
or
```
perf_event_open(/sys/kernel/debug/tracing/events/uprobes/p__proc__1_root_proc_10687_root_home_fedora_jenkins_workspace_bcc_pr_label_fc28_build_tests_cc_test_libbcc_0x4a0ec1_bcc_10687/id): Invalid argument
```
May be internal buffer length not large enough? Could you take a look?

davemarchevsky:
@yonghong-song Pushed up a commit trying to address the "path too long" issue. Will look into the other one as well, just wanted to re-run tests on test infra.

src/cc/bcc_proc.c
```diff
@@ -121,25 +121,21 @@ static char *_procutils_memfd_path(const int pid, const uint64_t inum) {
...
-  char procmap_filename[128];
-  FILE *procmap;
-  snprintf(procmap_filename, sizeof(procmap_filename), "/proc/%ld/maps",
-           (long)pid);
-  procmap = fopen(procmap_filename, "r");
-  if (!procmap)
-    return -1;
-
-  char buf[PATH_MAX + 1], perm[5], dev[8];
+// return: 0 -> stopped iterating over modules, no need to look at perf maps
```
palmtenor: Is this comment still valid? I don't see the return value checked in `bcc_procutils_each_module`

src/cc/bcc_proc.c
```diff
@@ -148,39 +144,65 @@ int bcc_procutils_each_module(int pid, bcc_procutils_modulecb callback,
...
+      return 0;
   }
 
-  fclose(procmap);
+  return -1;
+}
+
+int bcc_procutils_each_module(int pid, bcc_procutils_modulecb callback,
+                              void *payload) {
+  char procmap_filename[128];
```
palmtenor: Why most the opening of `procmap` file handle here and pass in as argument? Is that mostly for Unit Test?

src/cc/usdt/usdt.cc
palmtenor: I think this is a bit debatable. Imagine this case. User wants to trace an USDT in `libc`, so user provided the path to `libc` (in root fs), as well as a PID. Now that PID has a different mount namespace that `libc` may live in a different path. This would actually cause that to fail.

@yonghong-song what do you think?

src/cc/bcc_syms.h
```diff
@@ -65,8 +66,21 @@ int bcc_symcache_resolve_name(void *resolver, const char *module,
                               const char *name, uint64_t *addr);
 void bcc_symcache_refresh(void *resolver);
 
+struct mod_search {
```
palmtenor: I feel this should probably be an internal struct (`bcc_syms.h` is user-facing header). You can just re-define it in the Unit Test.

yonghong-song:
[buildbot, test this please]

yonghong-song:
latest code still failed with `test_libbcc`. One error is
```
perf_event_open(/sys/kernel/debug/tracing/events/uprobes/p__proc__1_root_proc_self_exe_0x7096e1_bcc_5554/id): Unknown error 524
```
error code 524 is UNSUPPORTED.

another is
```
3: perf_event_open(/sys/kernel/debug/tracing/events/uprobes/p__proc__1_root_proc_10692_root_home_fedora_jenkins_workspace_bcc_pr_label_fc28_build_tests_cc_test_libbcc_0x4a4701_bcc_10692/id): Invalid argument
```

on ubuntu
```
3: /home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1710/tests/cc/test_usdt_probes.cc:37
3: 
3: /home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1710/tests/cc/test_usdt_probes.cc:45: FAILED:
3:   REQUIRE( probe->in_shared_object(probe->bin_path()) == false )
3: with expansion:
3:   true == false
3: 
3: -------------------------------------------------------------------------------
3: test find a probe in our process' shared libs with c++ API
3: -------------------------------------------------------------------------------
3: /home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1710/tests/cc/test_usdt_probes.cc:86
3: 
3: /home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1710/tests/cc/test_usdt_probes.cc:91: FAILED:
3:   REQUIRE( res.msg() == "" )
3: with expansion:
3:   "Unable to find USDT libbcc_test:sample_lib_probe_1 from binary  PID 6735 for
3:   probe on_event"
3:   ==
3:   ""
```

src/cc/bcc_proc.c
davemarchevsky: fixed in latest commit

src/cc/bcc_proc.c
davemarchevsky: yep, I wanted to be able to easily test the `/proc/PID/maps` iteration

src/cc/bcc_syms.h
davemarchevsky: moved it to `bcc_syms.cc` and redefined in unit tests

src/cc/usdt/usdt.cc
davemarchevsky: I think this would be broken in the `setns` world as well? But not 100% sure I understand the scenario you're describing. Let me try to restate it.

This code assumes that absolute paths provided to `Context::resolve_bin_path` are relative to the root of the process being examined. If I want to trace USDTs in `/some/path/libc` in process `PID`, it's assumed that I'm talking about `/proc/PID/root/some/path/libc`, but I may actually be talking about a different `libc` that, as far as `PID` knows, lives at `/proc/PID/root/other/path/libc`, even though in the root ns it was at `/some/path/libc`.

If I'm understanding this right I would expect this to be broken in `setns` world. Not sure how to unbreak it going forward, but the fact that we're looking at `(device_id, inode)` starting with this diff is a good start.

yonghong-song:
[buildbot, test this please]

src/cc/usdt/usdt.cc
yonghong-song: For usdt, if users provides both a pid and binary path, and the pid is the one outside the namespace, what should the binary path look look? In my opinion, user needs to provide
the binary path accessible outside the name space. Is this the right thing to do?

For example,
    inside namespace: pid = 4, path = /usr/lib/libc.so.6
    outside namespac: pid = 1234, path = /opt/name1/usr/lib/libc.so.6
If bpf tracing with usdt is done outside namespace, user parameters should be
    pid = 1234, path = /opt/name1/usr/lib/libc.so.6
                      or path = /proc/1234/root/usr/lib/libc.so.6
Do this sound right? 

We may not handle the following case properly if user says
  pid = 1234, library = c (libc)
In this case, bcc has to find out libc path inside the namespace and prefix it with
/proc/1234/root. Do we have a reliable way to get this without increasing complexity
too much? setns may not work if it is multi-threaded.
Otherwise, I am okay with forcing user to have a full path to avoid potential surprise.

yonghong-song:
@davemarchevsky the last commit `Try excluding usdt tests when no sdt header again` is not needed. Could you remove it? No sdt.h should not block the test since you use folly headers.

davemarchevsky:
All ubuntu failures are same:

```
3: ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
3: test_libbcc is a Catch v1.4.0 host application.
3: Run with -? for options
3: 
3: -------------------------------------------------------------------------------
3: test find a probe in our process' shared libs with c++ API
3: -------------------------------------------------------------------------------
3: /home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1710/tests/cc/test_usdt_probes.cc:87
3: 
3: /home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1710/tests/cc/test_usdt_probes.cc:92: FAILED:
3:   REQUIRE( res.msg() == "" )
3: with expansion:
3:   "Unable to find USDT libbcc_test:sample_lib_probe_1 from binary  PID 6709 for
3:   probe on_event"
3:   ==
3:   ""
3: 
3: ===============================================================================
3: test cases:  33 |  32 passed | 1 failed
3: assertions: 562 | 561 passed | 1 failed
```

fc28 looks more complicated, so going to focus on this first.

yonghong-song:
For
```
3: with expansion:
3:   "Unable to find USDT libbcc_test:sample_lib_probe_1 from binary  PID 6709 for
3:   probe on_event"
3:   ==
3:   ""
```
This is a test case bug. The probe name should be `sample_probe_1`, not `sample_lib_probe_1`.

Could you fix and resubmit the pull request?

davemarchevsky:
@yonghong-song The `sample_lib_probe_1` is a tracepoint I'm adding in this PR [here](https://github.com/iovisor/bcc/pull/2324/files#diff-ceb8e778e6086e007ed72cd763b4e89e) (and associated test to confirm that it works). Want to add this test in because looking for usdt in shared lib is different codepath that was previously untested.

yonghong-song:
[buildbot, test this please]

davemarchevsky:
Only `fc28` tests remain broken. Addressing.

Ubuntu tests were broken because the shared library (w/ USDT) that I added to tests was never actually used in `test_libbcc`. Ubuntu's `gcc` decided that it could remove it from the binary's list of shared libs because of this, which is why the test was unable to locate the USDT in the shared lib's dummy function.

yonghong-song:
[buildbot, test this please]

davemarchevsky:
@yonghong-song It's finally green. The remaining `fc28` error was actually indicative of a bug: there are a few places where I prepend `/proc/PID/root` to the path if the pid is set - by default the pid is set to `-1` (e.g. `attach_usdt`), in this case we shouldn't be prepending `/proc/PID/root`. `perf_event_open` was failing for this reason.

yonghong-song:
Cools. Thanks! Will do another round of review.
@palmtenor  could you also take a look?

davemarchevsky:
@palmtenor I squashed down to one commit and pushed. Since the PR got big and messy it was hard to find logical multi-commit split.

FYI there were two `REVERT ME` commits that added debug prints which I removed.

I'd like to validate that this still fixes the production issue that prompted the PR (since there were lots of minor test-related bugfixes), will update when that's done.

> Can you catch me what's with the -1? I got a bit lost there.

A few places in this PR I do a check: "If we have an actual PID, prepend `/proc/PID/root` to path, otherwise do nothing". The "have an actual PID" check was initially checking if `optional` wrapper had a value. However, elsewhere in the code `-1` is used as a sentinel value for PID when one isn't provided, so we were erroneously prepending `/proc/-1/root`, which caused a test to break.

yonghong-song:
[buildbot, test this please]

yonghong-song:
@davemarchevsky sounds good to me for first verifying in production environment. Thanks!

davemarchevsky:
New CI system seems broken: 

Last log line in recent Travis CI build says: `/home/travis/.travis/functions: line 107: secure: unbound variable`, which doesn't look like an issue with bcc code to me.

yonghong-song:
@davemarchevsky do not need to worry Travis CI run for now. There are a different commit trying to address all failures. For your commit, maybe you can rename `.travis.yml` to `.travis.yml.org` to
temporarily disable it since it looks travis CI failure will prevent merging?

yonghong-song:
Looks like builtbot has some issues. @drzaeus77 is in the middle of fixing them.

yonghong-song:
[buildbot, test this please]

palmtenor:

yonghong-song:
[buildbot, test this please]

PR #2328, opened 2019-04-26
tutorial_bcc_python_developer: Finish auto-generated update

chrahunt:
A sentence was leftover from before we could auto-generate the Python data structure.


palmtenor:
[buildbot, ok to test]

PR #2334, opened 2019-04-30
type error on tools/dbslower.py

201341:
there is no member called delta on struct data_t, which can cause print errors when using dbslower.py


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

PR #2344, opened 2019-05-09
vfscount.py: add args time

201341:
Just like vfsstat.py. I think the args time is good for us to analysis.If we need to know the number of vfs call during a period of time, the previous implement can't tell the details.


yonghong-song:
[buildbot, test this please]

201341:
I have updated them. @yonghong-song

yonghong-song:
[buildbot, test this please]

PR #2349, opened 2019-05-13
tools: Added exitsnoop: Trace all process termination (exit, signal)

amdn:


brendangregg:
I'd suggest two changes:

- Add a column for the lifespan of the process in milliseconds, or seconds with a couple of decimal places (and keep the output < 80 char width).
- Instead of "signal: 2" I'd have it say "signal: 2 (INT)". I don't know if there's a library function for signal translation (you can see them in `kill -l`). (Note: I probably would not have killsnoop do the translations as well, as it's one of the tools I was keeping as simple as possible for use in tutorials; extra note: I need to add a comment to such tools otherwise people will add stuff to them not knowing I was trying to keep them short).

amdn:
@brendangregg Thank you for the review.  I have made the changes you suggested.

brendangregg:
Thanks, however, one of the changes is not right -- maybe I did not explain it properly. Here's a mock-up of what I was suggesting:

```
TIME     COMM             PID    PPID   AGE(s)   EXIT_CODE
18:10:40 bash             19110  18676  0.12     FAIL: exit_code: 65
18:10:40 bash             19113  18676  0.11     FAIL: exit_code: 66
18:10:40 bash             19115  18676  0.12     FAIL: exit_code: 67
18:10:40 bash             19117  18676  0.10     FAIL: exit_code: 68
18:10:41 bash             19119  18676  0.09     FAIL: exit_code: 69
18:10:41 bash             19121  18676  0.11     FAIL: exit_code: 70
18:10:41 sleep            19125  19124  11.41    KILL: signal: 2 (INT)
18:10:41 bash             19124  18676  0.13     KILL: signal: 2 (INT)
```

The AGE(s) column shows the duration of the process, from start to exit. As an existing example, see filelife:

```
# ./filelife 
TIME     PID    COMM             AGE(s)  FILE
05:57:59 8556   gcc              0.04    ccCB5EDe.s
05:57:59 8560   rm               0.02    .entry_64.o.d
05:57:59 8563   gcc              0.02    cc5UFHXf.s
05:57:59 8567   rm               0.01    .thunk_64.o.d
05:57:59 8578   rm               0.02    .syscall_64.o.d
```

If we're instrumenting process exit, you'll frequently see short-lived processes in the output, and the end-user will want to see their duration.

It can be instrumented at least two ways: via sched:sched_process_fork and recording the start time in a map; or, via the task_struct and digging out the times as used by rusage(). The first way will be easy but adds a second probe (at process execution rates, it shouldn't matter); the second way is slightly more efficient, but will be unstable as it'll involve digging things out of task_struct.

If you use the extra tracepoint, then there's the situation where you will see an exit but have not seen its start. I'd either not print a line of output, or do so with "?" as the AGE(s) -- documenting in the man page that it began before the tool was executed, so its birth time was unknown.

amdn:
@brendangregg That makes sense, I'll get to it today.

amdn:
@brendangregg
Good news: `struct task_struct` does have the starting time of the task in nanos, `u64 start_time` and there's an example where the age of the task is computed, just like you suggested, in `kernel/taskstats.c:fill_stats_for_tgid()`
Bad news: on Red Hat 7.6 (the version that supports eBPF) the kernel's task_struct uses a different type, `struct timespec` for start_time... the base kernel is 3.10.
So, working on getting it to work either way, but it takes some work, which you predicted.
Also, a colleague has helped refactor the code so it can be imported as a python module or run as a script.  Our plan is to abandon this PR and resubmit from an instana/bcc fork.  Any objections?

brendangregg:
The base kernel is 3.10 still?? ugh...

But I'm not too concerned about stashing the time away during the fork tracepoint. For more-frequent events, I'd be more insistent that we don't add another probe if it can be avoided, but process execution is infrequent.

Yes, you can close this PR and resend from another if that works.

amdn:
@brendangregg I got the AGE feature working, exitsnoop will use a simple heuristic to auto-detect Centos/Red Hat with the older kernel and convert the task->start_time from struct timespec to nanoseconds since epoch, so the conversion is done in eBPF and no other probe was needed (or use of a map).  Added an option --timespec to explicitly request the conversion.  Also added a simple checker if clang complains about struct timespec, exitsnoop retries automatically with the --timespec flag once to see if it will compile.  I will close this PR and resubmit from the instana fork. Thanks.

amdn:
Resubmitted from Instana fork as #2382

PR #2359, opened 2019-05-17
cap_capable: Display options correctly, add filter for INSETID

pothos:
The audit filter value has been changed from simply checking
for 0 to checking the LSB against 0.
Since Linux 5.1 kernels got a second bit defined (in
https://github.com/torvalds/linux/blob/v5.1/include/linux/security.h#L65)
the simple check filtered this as well.

This new bit is shown in an own column and can be filtered out by a flag.

### DIFF
--- a/tools/capable.py
+++ b/tools/capable.py
@@ -99,6 +99,7 @@ StackType = Enum(("Kernel", "User",))
 bpf_text = """
 #include <uapi/linux/ptrace.h>
 #include <linux/sched.h>
+#include <linux/security.h>
 
 struct data_t {
    u32 tgid;
@@ -106,6 +107,7 @@ struct data_t {
    u32 uid;
    int cap;
    int audit;
+   int insetid;
    char comm[TASK_COMM_LEN];
 #ifdef KERNEL_STACKS
    int kernel_stack_id;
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, test this please]

tools/capable.py
```diff
@@ -149,12 +153,15 @@ def __getattr__(self, name):
     bpf_text = bpf_text.replace('FILTER1',
         'if (pid != %s) { return 0; }' % args.pid)
 if not args.verbose:
-    bpf_text = bpf_text.replace('FILTER2', 'if (audit == 0) { return 0; }')
+    bpf_text = bpf_text.replace('FILTER2', 'if ((cap_opt & 0b1) == 0) { return 0; }')
+if args.no_insetid:
+    bpf_text = bpf_text.replace('FILTER_INSETID', 'if ((cap_opt & 0b10) != 0) { return 0; }')
```
yonghong-song: In old kernel, we have
```
#define SECURITY_CAP_NOAUDIT 0
#define SECURITY_CAP_AUDIT 1
```
In the new kernel, we have
```
/* Default (no) options for the capable function */
#define CAP_OPT_NONE 0x0        
/* If capable should audit the security request */
#define CAP_OPT_NOAUDIT BIT(1) 
/* If capable is being called by a setid function */
#define CAP_OPT_INSETID BIT(2)
```

Looks like the above `cap_opt & ...` may not be correct, right? Testing for audit/noaudit is different for old and new kernels.

Can we include `linux/security.h` and using macros to differentiate old and new kernels, this way, we do not need option `no_insetid`?

In print out, if INSETID is not supported, we just print out N/A. In bpf program. The above macro should indicate whether INSETID is supported or not.

pothos: Thanks for the review!
Yes, I didn't expect the logic to change that much and should have checked again.
I've fixed it and now it also makes sense to remove the `-n` option.

palmtenor:
[buildbot, ok to test]

yonghong-song:
Looks good to me. Is it possible for you to add an example with valid INSETID?

pothos:
I have updated the example there. Is it ok like this?

yonghong-song:
@pothos Sorry, if my previous comment is not clear. I mean to have two examples:
one example with
```
TIME      UID    PID    TID    COMM             CAP  NAME                 AUDIT  INSETID
08:22:36  0      12869  12869  chown            0    CAP_CHOWN            1      N/A
...
```

and other example with
```
TIME      UID    PID    TID    COMM             CAP  NAME                 AUDIT  INSETID
08:22:36  0      12869  12869  chown            0    CAP_CHOWN            1      0/1
...
```

pothos:
I've added it as second example.

yonghong-song:
Looks good! Thanks!

pothos:
Thank you :)

PR #2369, opened 2019-05-23
Add option to print unix timestamp in trace tool

letmaik:
Currently, the trace tool only supports accurate timestamps from trace start (through the monotonic clock within BPF directly), and local time HH:ss (through calls to Python's datetime when printing the record). I had a need to get accurate real-time timestamps which this PR implements.

I realize that this can only ever be approximate since you need to relate monotonic time to real time somehow, and I do that by querying the real time at trace start directly after the same is done for getting the monotonic time at trace start. The hope is that the error is small enough. An improvement would be to query both times within C/C++ to avoid any Python overhead between both calls, but that could be done in the future if there is need as that needs more effort.

Sample when using `--unix-timestamp`:
```
TIME              PID     TID     COMM            FUNC             -
1558631890.211117 21369   21369   ksgxswapd       sgx_ewb          139895674765314
```

Currently the new flag is an addon to the `--timestamp` flag, which may be weird. I'm happy to change that, but it required more code changes, hence I first wanted to see if this PR is accepted or not.

### DIFF
--- a/man/man8/trace.8
+++ b/man/man8/trace.8
@@ -3,7 +3,7 @@
 trace \- Trace a function and print its arguments or return value, optionally evaluating a filter. Uses Linux eBPF/bcc.
 .SH SYNOPSIS
 .B trace [-h] [-b BUFFER_PAGES] [-p PID] [-L TID] [-v] [-Z STRING_SIZE] [-S] [-s SYM_FILE_LIST]
-         [-M MAX_EVENTS] [-t] [-T] [-C] [-K] [-U] [-a] [-I header]
+         [-M MAX_EVENTS] [-t] [-u] [-T] [-C] [-K] [-U] [-a] [-I header]
          probe [probe ...]
 .SH DESCRIPTION
 trace probes functions you specify and displays trace messages if a particular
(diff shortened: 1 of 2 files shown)


yonghong-song:
@letmaik  could you provide a complete dump than just below
```
TIME              PID     TID     COMM            FUNC             -
1558630422.366075           ...
```
People may get better sense whether the information is still concise and presentable with 80 char limit per line.

cc @brendangregg

yonghong-song:
[buildbot, test this please]

letmaik:
@yonghong-song I updated the original post. The last entry is a memory address, but that's not relevant here.

yonghong-song:
Thanks. Looks good to me. Could you make a change to man page and example file as well?

brendangregg:
This is pretty niche. Probably ok to add to trace, since it's complex as it is, but I would not add --unix-timestamp to anything else (people can hack it in when needed in their own custom versions).

letmaik:
@brendangregg I agree it's niche, though for me essential (to be able to compare with other logs). The reason for contributing the feature is so that I don't have to maintain my own copy of this already quite complex script.

brendangregg:
ok; can you add this to the man page?

letmaik:
Done. Is that OK?

yonghong-song:
[buildbot, test this please]

PR #2375, opened 2019-05-27
fix proto compare bug in example

ChenLingPeng:
Signed-off-by: forrestchen <forrestchen@tencent.com>


ChenLingPeng:
`h_proto` here is l3 protocol, what we what to compare is l4 protocol I think.

yonghong-song:
[buildbot, test this please]

PR #2378, opened 2019-05-28
Support pinned maps through BPF_TABLE_PINNED macro

pothos:
Define a map to be loaded from a pinned path through
the new macro `BPF_TABLE_PINNED(..., "/sys/fs/bpf/xyz");`
instead of creating a new map.

Internally the path is appended to the section attribute
with a delimiter. The approach of adding a new pinned_path
pointer field to the table struct seemed promising but
caused more troubles than the percieved cleanless may be worth.

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.h
+++ b/src/cc/frontends/clang/b_frontend_action.h
@@ -174,8 +174,8 @@ class BFrontendAction : public clang::ASTFrontendAction {
   void DoMiscWorkAround();
   // negative fake_fd to be different from real fd in bpf_pseudo_fd.
   int get_next_fake_fd() { return next_fake_fd_--; }
-  void add_map_def(int fd, std::tuple<int, std::string, int, int, int, int> map_def) {
-    fake_fd_map_[fd] = map_def;
+  void add_map_def(int fd, std::tuple<int, std::string, int, int, int, int, unsigned int> map_def) {
+    fake_fd_map_[fd] = move(map_def);
   }
 
  private:
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -543,6 +543,17 @@ Examples in situ:
 [search /examples](https://github.com/iovisor/bcc/search?q=BPF_TABLE+path%3Aexamples&type=Code),
 [search /tools](https://github.com/iovisor/bcc/search?q=BPF_TABLE+path%3Atools&type=Code)
 
+#### Pinned Maps
+
+Maps that were pinned to the BPF filesystem can be accessed through an extended syntax: ```BPF_TABLE_PINNED(_table_type, _key_type, _leaf_type, _name, _max_entries, "/sys/fs/bpf/xyz")```
+The type information is not enforced and the actual map type depends on the map that got pinned to the location.
+
+For example:
+
+```C
+BPF_TABLE_PINNED("hash", u64, u64, ids, 1024, "/sys/fs/bpf/ids");
+```
+
 ### 2. BPF_HASH
 
 Syntax: ```BPF_HASH(name [, key_type [, leaf_type [, size]]])```
--- a/src/cc/api/BPFTable.h
+++ b/src/cc/api/BPFTable.h
@@ -64,6 +64,10 @@ class BPFTableBase {
     return rc;
   }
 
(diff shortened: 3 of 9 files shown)


yonghong-song:
[buildbot, test this please]

src/cc/frontends/clang/b_frontend_action.h
```diff
@@ -174,7 +174,7 @@ class BFrontendAction : public clang::ASTFrontendAction {
   void DoMiscWorkAround();
   // negative fake_fd to be different from real fd in bpf_pseudo_fd.
   int get_next_fake_fd() { return next_fake_fd_--; }
-  void add_map_def(int fd, std::tuple<int, std::string, int, int, int, int> map_def) {
+  void add_map_def(int fd, std::tuple<int, std::string, int, int, int, int, std::string> map_def) {
     fake_fd_map_[fd] = map_def;
```
palmtenor: `move` here as well?

pothos: Hi, do you mean `fake_fd_map_[fd] = move(map_def);`?

pothos:
Does the CI runs have `bpftool` available? Then I could add a python test that pins a map and accesses it.

yonghong-song:
No. CI probably does not have bpftool, which is not widely distributed yet.
Travis CI is not working for bcc. It would be nice to make it work.

src/cc/frontends/clang/b_frontend_action.h
yonghong-song: General approach looks okay. Could you change the above last element in map_def from `std::string` to `int`? In any case, it is expected to be an int to represent the pinned map id.

pothos:
Thanks for the comments. I did the map ID roundtrip.

I have also added a test case (which needed `table.get_fd()`) but count not run it because `make test` does not seem to execute it despite `ENABLE_USDT:BOOL=on` in `CMakeCache.txt`. What's missing?

yonghong-song:
[buildbot, test this please]

pothos:
Ah, actually the tests can't run on my machine due to `CommandLine Error: Option 'no-x86-call-frame-opt' registered more than once!`… I'll try to fix the test failure.

pothos:
The pinning test succeeds locally. Why could it fail in Jenkins?

yonghong-song:
Error: 3:   REQUIRE( bpf_obj_pin(bpf.get_hash_table<int, int>("ids").get_fd(), "/sys/fs/bpf/test_pinned_table") == 0 )

the test machine may not have /sys/fs/bpf, you may need to do a check and mount by yourself.

pothos:
Thanks for the tip. If this works now, I'll update it on Friday to make a proper check and unmounting it again if it was mounted by the test.

src/cc/frontends/clang/b_frontend_action.h
palmtenor: @pothos Yeah, just a nit, since we already the tuple constructed:)

palmtenor: Yeah, just a nit, since we already constructed the tuple~

pothos: Ok, done.

palmtenor:
[buildbot, ok to test]

pothos:
Oh, so only Ubuntu 16.04 failed. It should run Linux 4.4 were I thought pinning got introduced. Anyway, I've not tried yet to run the test on a local Ubuntu 16.04.

pothos:
Hm, initially Ubuntu 16.04 had the 4.4 kernel where it fails. But as soon as you update, it's a newer kernel (4.15) and the problem disappears. What should we do? Can you use an up-to-date Ubuntu 16.04 in the CI?

yonghong-song:
The 16.04 failure mostly due to the 4.4 kernel not supporting `bpf_map_get_fd_by_id` and `bpf_obj_get_info_by_fd`, which is introduced later. I think it is okay to mark the test to skip kernel version lower than 4.12.

pothos:
@yonghong-song Thanks, done.

yonghong-song:
[buildbot, test this please]

PR #2379, opened 2019-05-28
Fix small things from the python developer tutorial

DoctorSher:
disksnoop does not execute properly on kernels 5.0+ because `blk_start_request` no longer exists in `block/blk-core.c`. I added the check found in the biolatency tool that checks the existence of the kprobe before it attaches.

For the other commit, the text read "user-defined static tracing" which does not match the acronym USDT, so I changed it to the standard "user statically-defined tracing".


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

Issue #2380, opened 2019-05-28
Python tutorial's BPF_TABLE example does not use BPF_TABLE


DoctorSher:
Lesson 16 in the python developer tutorial (`task_switch`) discusses `BPF_TABLE` at length, but the program ends up using `BPF_HASH`. There is no usage of `BPF_TABLE` in the code. Although, oddly enough, the comment above `BPF_HASH` is the argument list for `BPF_TABLE`.

palmtenor:
That probably just got bulk-replaced to the `BPF_HASH` alias at certain point. As the lesson says
```
This is an older tutorial included as a bonus lesson. Use this for recap and to reinforce what you've already learned.
```
It may came before the time we have the more specific Macros like `BPF_HASH`, `BPF_ARRAY`, etc.

Feel free to update the tutorial:)

DoctorSher:
I certainly wouldn't mind updating the tutorial! I'd like to double check that the direction makes sense though. Essentially, my impression is that given all of the helper macros mentioned above, there's not an obvious reason for people to be calling `BPF_TABLE`. So I would be taking out the text describing `BPF_TABLE` and instead frame it as using `BPF_HASH`. Does this sound reasonable?

yonghong-song:
@DoctorSher Yes, in general BPF_HASH is preferred. Your suggestions above sound reasonable. Thanks for helping with the tutorial!

PR #2383, opened 2019-05-28
Revise python tutorial lesson 16 to make sure tutorial text and code match

DoctorSher:
PR addressing Issue #2380 

Lesson 16 spent a lot of text discussing `BPF_TABLE`, but the code doesn't use `BPF_TABLE` at all. Rather, it uses `BPF_HASH`, which users should already be quite familiar with by the end of the tutorial. As a result, I cut out the text explaining `BPF_TABLE`.

I also separated the BPF C code into a separate file, since the tutorial was already written as though that was the case (it discusses `task_switch.c` but that file did not exist).


yonghong-song:
[buildbot, test this please]

Issue #2392, opened 2019-05-31
capable: move TID and INSETID columns to -x


brendangregg:
People have added extra columns to capable, and the default output is now > 80 chars. I can no longer use capable:

- in slide decks
- in blog posts
- in books
- in emails
- in ticketing systems

In each case, things go wrong with output > 80 chars. (e.g., I just copy-n-pasted some capable output in the book I'm writing, and it's wrapped horribly, and the publisher does not allow small font sizes to accommodate).

It's ok if you really really need to go beyond 80 chars, but I don't think TID and INSETID columns must be in the default output. I'd move them to a -x option (for extra fields).

I did mention this in https://github.com/iovisor/bcc/blob/master/CONTRIBUTING-SCRIPTS.md , but maybe that isn't enough. I'll add a ticket for doing a test suite for checking default output of these tools is < 80 chars, and if a PR exceeds 80 chars, it will get rejected and require an "is this really necessary?" override to continue.

CC @pothos @arighi

pothos:
Sure, I've submitted a change.

PR #2394, opened 2019-06-03
capable: Hide TID and INSETID columns by default

pothos:
The TID and the INSETID bit are the less interesting
fields and keeping them out allows to reduce the line
length below 80 characters.

Closes https://github.com/iovisor/bcc/issues/2392

I have updated the example description since I anyway found out that one had the TID column and the other not.

### DIFF
--- a/man/man8/capable.8
+++ b/man/man8/capable.8
@@ -25,6 +25,9 @@ Include kernel stack traces to the output.
 .TP
 \-U
 Include user-space stack traces to the output.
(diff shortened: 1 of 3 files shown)


palmtenor:
[buildbot, ok to test]

palmtenor:
@brendangregg Does this look good to you?

brendangregg:
Looks good, thanks!

PR #2397, opened 2019-06-04
Fix compiler warnings

tobiasblass:
These commits fix two compiler warnings about silent string truncations when building up pathnames using snprintf.

In the case of `bpf_attach_probe` I chose the minimal fix. However, I'm somewhat confused why `fname` and `buf` have different sizes in the first place. `create_probe_event` assumes that it generates paths of up to _PATH_MAX_ characters, but then `bpf_attach_probe` assumes that the result (including the "/id" suffix) fits into 256 characters. Maybe it might make sense to either shrink `buf`, enlarge `fname` to _PATH_MAX_ or allocate `fname` with malloc based on the runtime length of the `buf` path.

### DIFF
--- a/src/cc/frontends/clang/kbuild_helper.cc
+++ b/src/cc/frontends/clang/kbuild_helper.cc
@@ -142,13 +142,16 @@ static inline int extract_kheaders(const std::string &dirpath,
     }
   }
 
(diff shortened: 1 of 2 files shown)


palmtenor:
[buildbot, ok to test]

PR #2403, opened 2019-06-07
tools/execsnoop: add -T option

boat0:
Add a -T option to include a time column on output (HH:MM:SS) and update manpage
and example file accordingly.


brendangregg:
[buildbot, ok to test]

brendangregg:
thanks!

PR #2411, opened 2019-06-17
tools/tcpstates: fix local/remote port output

boat0:


yonghong-song:
[buildbot, test this please]

yonghong-song:
The code change looks okay. But I think this won't change end result as the original implementation should be still correct, right?

boat0:
> The code change looks okay. But I think this won't change end result as the original implementation should be still correct, right?

The original implementation
`data4.ports = dport + ((0ULL + lport) << 32);`
has a bug here: since data4.ports is u32, lport gets lost.

yonghong-song:
Okay, I see. Thanks!

yonghong-song:
[buildbot, test this please]

PR #2412, opened 2019-06-17
update tools diagram for 2019

brendangregg:
completely redrawn and updated; now wider to better fit 16:9 presentations, which have become the norm.


brendangregg:
[buildbot, ok to test]

marvelm:
@brendangregg How are you generating this diagram?

Issue #2385, opened 2019-05-29
USDT bpf_readarg_* functions are not generated when compiling BPF programs from a src_file


tobiasblass:
To access the arguments of an USDT probe, bcc generates a bunch of helper functions that parse the `.note.stapsd` ELF section and take the arguments from the correct registers. In the python implementation, this is done  [here](https://github.com/iovisor/bcc/blob/master/src/python/bcc/__init__.py#L309-L319)

As you can see by the if statement, this is only done if the BPF program is passed as a string (using the `text` argument), but not if the BPF program is passed as a file (using the `src_file` argument). Using `bpf_usdt_readarg` in such a file therefore fails with a _function not defined_ error.

Is this an oversight, or is there a reason for this asymmetry?

----
Minimal example:
Running `test.py` fails with
```
bpf: Failed to load program: Invalid argument
0: (bf) r2 = r10
1: (07) r2 += -8
2: (b7) r3 = 8
3: (85) call unknown#-1
BPF_CALL uses reserved fields
```
unless `_bpf_readarg_test_1` is defined by hand in the BPF program.

test.py:
```Python
#!/usr/bin/env python3
import bcc

usdt = bcc.USDT(path='./test')
usdt.enable_probe(probe='main', fn_name='test')
bpf = bcc.BPF(src_file='test.bpf', usdt_contexts=[usdt])
bpf.trace_print()
```

test.c:
```C
#include<sys/sdt.h>
int main(int argc, char **argv) {
	DTRACE_PROBE1(test,main,argv[0]);
	return 0;
}
```

test.bpf:
```C
#include <uapi/linux/ptrace.h>
// SPDX-License-Identifier: GPL
#define BPF_LICENSE GPL

/* Uncomment this to define the readarg-function by hand. This fixes the issue,
   which proves that this function is indeed the culprit. */
//#define define_readarg_by_hand

#ifdef define_readarg_by_hand
static int _bpf_readarg_test_1(struct pt_regs *ctx, void *dest, size_t len) {
	if (len != sizeof(char*)) return -1;
        /* Adjust the register depending on the output of readelf -n test */
	*(void**)dest = (void*)ctx->ax;
	return 0;
}
#endif

int test(struct pt_regs *ctx) {
    void *ptr;
    char argv0[64];
    bpf_usdt_readarg(1, ctx, &ptr);
    bpf_probe_read_str(argv0, sizeof argv0, ptr);
    bpf_trace_printk("%s\n", argv0);
    return 0;
[...]

PR #2417, opened 2019-06-22
Handle the text and src_file the same way in the BPF constructor

tobiasblass:
Instead of having two different code paths for code passed in as a
string (via the text parameter) and code passed in via file (via the
src_file parameter), load the contents of src_file into the text
variable and handle both in the same code path.
This fixes issue #2385.

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -291,6 +291,8 @@ class BPF(object):
         hdr_file = _assert_is_bytes(hdr_file)
         text = _assert_is_bytes(text)
 
+        assert not (text and src_file)
+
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
How about C++ APIs? Do you think we could change clang frontends for both C++ and python?

palmtenor:
@yonghong-song I don’t think C++ API has by-file interface

yonghong-song:
Looks good to me. Currently, the `b` language still uses src_file. In the future, when `b` language is retired, we can remove those related codes which deal with src_file.

PR #2420, opened 2019-06-24
Tag overridden functions with override

danobi:
This is pretty standard practice these days as it ensures you don't
accidentally create a new method that doesn't override the base class
method. It also helps document the fact you're overriding methods.


palmtenor:
[buildbot, ok to test]

PR #2427, opened 2019-06-27
[Don't merge] Testing whether removal of HAVE_SDT_HEADER ifdefs causes tests on some platforms to break on its own 

davemarchevsky:
Yonghong and I were talking offline about [my other PR](https://github.com/iovisor/bcc/pull/2324). I'm starting to think that some of the weird test failures we're seeing on that PR are due to the removal of the `HAVE_SDT_HEADER` gate instead of the numerous other code changes.

Just pushing this up so that it can run on iovisor test infra. Nothing to see here.

/cc @yonghong-song

### DIFF
--- a/src/cc/bcc_elf.c
+++ b/src/cc/bcc_elf.c
@@ -672,6 +672,7 @@ int bcc_elf_is_exe(const char *path) {
 }
 
 int bcc_elf_is_shared_obj(const char *path) {
+  printf("DAVEMARCHEVSKY: elf type of %s is %d\n", path, bcc_elf_get_type(path)); 
   return bcc_elf_get_type(path) == ET_DYN;
 }
 
--- a/src/cc/bcc_elf.c
+++ b/src/cc/bcc_elf.c
@@ -672,7 +672,6 @@ int bcc_elf_is_exe(const char *path) {
 }
 
 int bcc_elf_is_shared_obj(const char *path) {
-  printf("DAVEMARCHEVSKY: elf type of %s is %d\n", path, bcc_elf_get_type(path)); 
   return bcc_elf_get_type(path) == ET_DYN;
 }
 
--- a/tests/cc/CMakeLists.txt
+++ b/tests/cc/CMakeLists.txt
@@ -4,6 +4,7 @@
 include_directories(${CMAKE_SOURCE_DIR}/src/cc)
 include_directories(${CMAKE_SOURCE_DIR}/src/cc/api)
 include_directories(${CMAKE_SOURCE_DIR}/src/cc/libbpf/include/uapi)
+include_directories(${CMAKE_SOURCE_DIR}/tests/python/include)
 
 add_executable(test_static test_static.c)
 target_link_libraries(test_static bcc-static)
@@ -29,9 +30,4 @@ add_executable(test_libbcc
 
 target_link_libraries(test_libbcc bcc-shared dl)
 add_test(NAME test_libbcc COMMAND ${TEST_WRAPPER} c_test_all sudo ${CMAKE_CURRENT_BINARY_DIR}/test_libbcc)
-
-find_path(SDT_HEADER NAMES "sys/sdt.h")
-if (SDT_HEADER)
-	target_compile_definitions(test_libbcc PRIVATE HAVE_SDT_HEADER=1)
-endif()
 endif(ENABLE_USDT)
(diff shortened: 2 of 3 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
ubuntu 17.10 and 18.04 failed with the same reason
```
3: /home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1710/tests/cc/test_usdt_probes.cc:37
3: 
3: /home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1710/tests/cc/test_usdt_probes.cc:45: FAILED:
3:   REQUIRE( probe->in_shared_object(probe->bin_path()) == false )
3: with expansion:
3:   true == false
```

davemarchevsky:
I think this might be because `in_shared_object` ends up [checking whether elf type == `ET_DYN`](https://github.com/iovisor/bcc/blob/master/src/cc/bcc_elf.c#L674-L676) and perhaps the relatively newer Ubuntus [compile binaries as PIEs by default](https://wiki.ubuntu.com/SecurityTeam/PIE), so we end up having `ET_DYN` binaries that aren't meant to be used as shared libs?

palmtenor:
Well, I think PIE executable is more or less the same as shared object, that you can't really use USDT in it without PID right?

yonghong-song:
Tried a simple example,
```
-bash-4.4$ cat t.c
int g;
int main()  { return g; }
-bash-4.4$ gcc -pie t.c
-bash-4.4$ readelf -a a.out                                                                                                          
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00 
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              DYN (Shared object file)
  Machine:                           Advanced Micro Devices X86-64
  Version:                           0x1
  Entry point address:               0x550
  Start of program headers:          64 (bytes into file)
  Start of section headers:          6664 (bytes into file)
....
```
it does look like if compiled with -pie, the binary did have
```
  Type:                              DYN (Shared object file)
```

With the type DYN, bcc really does not know this is a truely binary or a shared library. 
We could check function `_start` since typically it is the starting point of the function, but it still looks like hacky.

We will need to find a solution for this. Otherwise, usdt won't work properly on ubuntu 17.04 and later.

For this particular test, we can temporarily fix it by returning success if the binary is considered as a shared library.

yonghong-song:
okay, let us run the test with some debug info.

yonghong-song:
[buildbot, test this please]

yonghong-song:
I just installed ubuntu 18.04 and compiled with test_libcc.
The following is `readelf -a test_libbcc` header
```
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 03 00 00 00 00 00 00 00 00
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - GNU
  ABI Version:                       0
  Type:                              DYN (Shared object file)
  Machine:                           Advanced Micro Devices X86-64
  Version:                           0x1
  Entry point address:               0x948490
  Start of program headers:          64 (bytes into file)
  Start of section headers:          81326576 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           56 (bytes)
  Number of program headers:         10
  Size of section headers:           64 (bytes)
  Number of section headers:         38
  Section header string table index: 37
```
So, yes, it is marked as a DYN.

davemarchevsky:
> For this particular test, we can temporarily fix it by returning success if the binary is considered as a shared library.

I will modify the test to do this (and remove debug print). If this results in a green test run, I'll make similar changes on the other PR and see what state that brings it to.

davemarchevsky:
@yonghong-song You already confirmed this two ways, but here's a 3rd from test output:

```
3: DAVEMARCHEVSKY: elf type of /home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1804/build/tests/cc/test_libbcc is 3
```

Glad we got to the bottom of this!

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

PR #2428, opened 2019-06-27
Add cli argument to limit run time

williamlw999:
If a user is granted sudo access only to btrfsslower.py and not kill or timeout, then user is unable to easily limit the runtime of btrfsslower.py. The optional cli argument stops execution of program after specified number of seconds passes.

### DIFF
--- a/man/man8/btrfsslower.8
+++ b/man/man8/btrfsslower.8
@@ -2,7 +2,7 @@
 .SH NAME
 btrfsslower \- Trace slow btrfs file operations, with per-event details.
 .SH SYNOPSIS
-.B btrfsslower [\-h] [\-j] [\-p PID] [min_ms]
+.B btrfsslower [\-h] [\-j] [\-p PID] [min_ms] [\-d DURATION]
 .SH DESCRIPTION
 This tool traces common btrfs file operations: reads, writes, opens, and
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
Maybe change the option from "-s" to "-d". We have a few other tools to use "-d" or "-D" to indicate the duration. Not sure whether signal handler is needed here. For example, `shmsnoop.py` exits after the duration like
```
start_time = datetime.now()
while not args.duration or datetime.now() - start_time < args.duration:
    try:
        b.perf_buffer_poll(timeout=1000)
    except KeyboardInterrupt:
        exit()
```
Not 100% precise, but should be close.

Please also modify man page and example file.

williamlw999:
I've made changes to the implementation so that it's identical to shmsnoop and I have updated the exampls and man page. I'm not familiar with buildbot, but I'll start looking into how to run the tests.

yonghong-song:
[buildbot, test this please]

williamlw999:
Please let me know if there's anything else you'd like me to add!

yonghong-song:
[buildbot, test this please]

yonghong-song:
Looks good. Thanks!

PR #2431, opened 2019-06-28
Travis CI: Lint for Python syntax errors and undefined names

cclauss:


yonghong-song:
[buildbot, test this please]

.travis.yml
```diff
@@ -1,4 +1,20 @@
 language: generic
+matrix:
+  include:
+    - name: "generic tests"
+    - name: "flake8 lint on Python 2.7"
+      language: python
+      python: 2.7
+      install: pip install flake8
+      script: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
```
pchaigno: Couldn't these be checked with the existing PEP8 script?

.travis.yml
```diff
@@ -1,4 +1,20 @@
...
+    - name: "flake8 lint on Python 3.7"
+      dist: xenial  # required for Python >= 3.7
+      language: python
+      python: 3.7
+      install: pip install flake8
+      script: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
```
pchaigno: Any reason the `install` and `script` lines can't be moved outside the matrix, to the `install` and `script` below?

.travis.yml
```diff
@@ -1,4 +1,20 @@
...
+  allow_failures:
+    - python: 3.7
```
pchaigno: Why allow failures for Python 3.7?

.travis.yml
cclauss: 1. __pep8__ was deprecated a long time ago in favor of __pycodestyle__.
2. __flake8__ is a superset of __pycodestyle__ so the Fxxx errors are in __pycodestyle__ but the Exxx errors are not.
3. __flake8__ provides different results when run on Python 2 vs. Python 3.

.travis.yml
cclauss: 1. We would lose the advantage of parallelism.
2. We would only run tests on legacy Python and not on Python.

.travis.yml
cclauss: The current code contains Python 3 syntax errors and Travis CI [__allow failures__ _lets you add in experimental and preparatory builds to test against versions or configurations that you are not ready to officially support_](https://docs.travis-ci.com/user/build-matrix/#rows-that-are-allowed-to-fail).

.travis.yml
pchaigno: I don't think that's how it works. Travis CI matrix configuration is often used only to set environment variables (see [strace's](https://github.com/strace/strace/blob/master/.travis.yml) for example), without preventing parallel builds.

.travis.yml
pchaigno: Shouldn't we fix those syntax errors instead? /cc @yonghong-song

.travis.yml
pchaigno: Ah, I see. Thanks for the explanation.
In that case, it would be better to update [the existing script](https://github.com/iovisor/bcc/blob/master/scripts/py-style-check.sh) rather than add new separate tests.

yonghong-song:
@cclauss @pchaigno I would like to take this as an opportunity to enable travis CI for bcc repo.
What are necessary steps for me to help enable that?

cclauss:
Go to https://travis-ci.org/iovisor/bcc and enable the repo.

yonghong-song:
thanks. in the process to enable this.

cclauss:
@yonghong-song No luck so far?  The default CI tool seems to remain on vacation.

yonghong-song:
> Other repos seem to be working on Travis CI: https://travis-ci.org/iovisor
Looks like I do not have enough permission to enable travis ci for bcc. I will try to resolve it soon.

yonghong-song:
@cclauss the iovisor/bcc travis CI is enabled. please check and experiment.

cclauss:
Whoever has permissions might need to go to  https://travis-ci.org/iovisor/bcc and force the first build to prime the pump.

cclauss:
@yonghong-song Can you please kick off a retest?

drzaeus77:
@cclauss Jenkins is currently down for maintenance, I hope to have it restored in a couple hours.

cclauss:
No problems...  Enjoy the 4th.

yonghong-song:
[buildbot, test this please]

cclauss:
Can you please try to rerun the tests for some of the open pull requests to see if they pass or not?

PR #2433, opened 2019-07-01
Lazily symbolize regular ELF binaries

### LINKED ISSUE #2421: bcc_symcache_resolve uses too much memory
More specifically, `ProcSyms` uses a pretty crazy amount of resources on big C++ binaries. I think this happens because ProcSyms reads every single symbol in the binary into `std::vector<Symbol> syms_`, where `Symbol` contains a string of the symbol name. For large C++ binaries, symbol names can get quite long.

In practice, we've seen bpftrace consume well over 1GB of memory just to resolve a single symbol. I have a couple ideas:

* Store the elf string index to the symbol name instead of storing the full string. Not sure if the string idx is valid across multiple `elf_begin`s.
* Don't read all symbols in when loading the symcache. Maybe do a streaming resolution (instead of the current method of reading all symbols + addrs in, sorting by addr, and walking the vector). Downside of this is added complexity.

I'm still trying to hack up a solution but any input would be appreciated. cc @yonghong-song  who I mentioned this in person to a while ago.

danobi:
This patch adds a new API to bcc_elf.h, bcc_elf_foreach_sym_lazy. This
helper avoids storing symbol names in string format, as for large
binaries this data can get quite large.

Instead we store the location in the ELF binary where we can later find
the string. Later on, we can load these strings on demand and cache them
in case they're accessed again.

This patch also makes lazy resolution the default for regular ELF
binaries, where regular means not perfmap or VDSO.

This closes #2421 .

### DIFF
--- a/src/cc/bcc_elf.c
+++ b/src/cc/bcc_elf.c
@@ -931,26 +931,53 @@ int bcc_elf_get_buildid(const char *path, char *buildid)
 int bcc_elf_symbol_str(const char *path, size_t section_idx,
                        size_t str_table_idx, char *out, size_t len)
 {
-  Elf *e;
-  int fd, err = 0;
+  Elf *e = NULL, *d = NULL;
+  int fd = -1, dfd = -1, err = 0;
   const char *name;
+  char *debug_file = NULL;
 
-  if (!out || !len)
+  if (!out)
     return -1;
 
   if (openelf(path, &e, &fd) < 0)
     return -1;
 
   if ((name = elf_strptr(e, section_idx, str_table_idx)) == NULL) {
-    err = -1;
-    goto exit;
+    // The binary did not have the symbol information, try the debuginfo
+    // file as backup.
+    debug_file = find_debug_via_buildid(e);
+    if (!debug_file)
+      debug_file = find_debug_via_debuglink(e, path, 0); // No crc for speed
+
+    if (!debug_file) {
+      err = -1;
+      goto exit;
+    }
+
+    if (openelf(debug_file, &d, &dfd) < 0) {
+      err = -1;
+      goto exit;
+    }
+
+    if ((name = elf_strptr(d, section_idx, str_table_idx)) == NULL) {
+      err = -1;
+      goto exit;
+    }
   }
 
   strncpy(out, name, len);
 
 exit:
-  elf_end(e);
-  close(fd);
+  if (debug_file)
+    free(debug_file);
+  if (e)
+    elf_end(e);
+  if (d)
+    elf_end(d);
+  if (fd >= 0)
+    close(fd);
+  if (dfd >= 0)
+    close(dfd);
   return err;
 }
 
--- a/src/cc/bcc_elf.c
+++ b/src/cc/bcc_elf.c
@@ -291,7 +291,7 @@ static int list_in_scn(Elf *e, Elf_Scn *section, size_t stridx, size_t symsize,
 #endif
 
       int ret;
-      if (callback_lazy)
+      if (option->lazy_symbolize)
         ret = callback_lazy(stridx, sym.st_name, name_len, sym.st_value,
                             sym.st_size, debugfile, payload);
       else
@@ -585,14 +585,16 @@ static int foreach_sym_core(const char *path, bcc_elf_symcb callback,
 
 int bcc_elf_foreach_sym(const char *path, bcc_elf_symcb callback,
                         void *option, void *payload) {
-  return foreach_sym_core(
-      path, callback, NULL, (struct bcc_symbol_option*)option, payload, 0);
+  struct bcc_symbol_option *o = option;
+  o->lazy_symbolize = 0;
+  return foreach_sym_core(path, callback, NULL, o, payload, 0);
 }
 
 int bcc_elf_foreach_sym_lazy(const char *path, bcc_elf_symcb_lazy callback,
                         void *option, void *payload) {
-  return foreach_sym_core(path, NULL, callback,
-      (struct bcc_symbol_option*)option, payload, 0);
+  struct bcc_symbol_option *o = option;
+  o->lazy_symbolize = 1;
+  return foreach_sym_core(path, NULL, callback, o, payload, 0);
 }
 
 int bcc_elf_get_text_scn_info(const char *path, uint64_t *addr,
--- a/src/cc/bcc_elf.c
+++ b/src/cc/bcc_elf.c
@@ -214,7 +214,8 @@ static Elf_Scn * get_section(Elf *e, const char *section_name,
 
 static int list_in_scn(Elf *e, Elf_Scn *section, size_t stridx, size_t symsize,
                        struct bcc_symbol_option *option,
-                       bcc_elf_symcb callback, void *payload) {
+                       bcc_elf_symcb callback, bcc_elf_symcb_lazy callback_lazy,
+                       void *payload) {
   Elf_Data *data = NULL;
 
 #if defined(__powerpc64__) && __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
@@ -237,6 +238,7 @@ static int list_in_scn(Elf *e, Elf_Scn *section, size_t stridx, size_t symsize,
     for (i = 0; i < symcount; ++i) {
       GElf_Sym sym;
       const char *name;
+      size_t name_len;
 
       if (!gelf_getsym(data, (int)i, &sym))
         continue;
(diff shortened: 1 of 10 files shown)


danobi:
Couldn't get local tests running against master so I'll let buildbot figure it out.

yonghong-song:
[buildbot, test this please]

yonghong-song:
The following two tests
```
         3 - test_libbcc (Failed)
	 15 - py_test_debuginfo (Failed)
```
both seg faults, you may want to investigate further.

src/cc/bcc_elf.c
```diff
@@ -287,15 +290,22 @@ static int list_in_scn(Elf *e, Elf_Scn *section, size_t stridx, size_t symsize,
 #endif
 #endif
 
-      if (callback(name, sym.st_value, sym.st_size, payload) < 0)
+      int ret;
+      if (callback_lazy)
```
palmtenor: Can we put this in `bcc_symbol_option` so that we could reliably select which option to use, avoid people get confused and specifying both etc.?

src/cc/syms.h
```diff
@@ -69,10 +69,24 @@ class KSyms : SymbolCache {
...
         : name(name), start(start), size(size) {}
-    const std::string *name;
+    Symbol(size_t section_idx, size_t str_table_idx, size_t str_len, uint64_t start,
+           uint64_t size)
+        : start(start), size(size) {
+      name_idx.section_idx = section_idx;
+      name_idx.str_table_idx = str_table_idx;
+      name_idx.str_len = str_len;
+    }
+    struct NameIdx name_idx;
```
palmtenor: Maybe we can do
```
struct Symbol {
  uint32_t isset;
  union {
    std::string* name;
    struct NameIdx name_idx;
  } data;
};
So that we save little memory in either case?

File: src/cc/bcc_syms.cc [hunk 3]
```diff
@@ -291,7 +299,7 @@ void ProcSyms::Module::load_sym_table() {
   if (type_ == ModuleType::PERF_MAP)
     bcc_perf_map_foreach_sym(name_.c_str(), _add_symbol, this);
   if (type_ == ModuleType::EXEC || type_ == ModuleType::SO)
-    bcc_elf_foreach_sym(name_.c_str(), _add_symbol, symbol_option_, this);
+    bcc_elf_foreach_sym_lazy(name_.c_str(), _add_symbol_lazy, symbol_option_, this);
```
palmtenor: Do you meant to just change the default behavior? Would be great if it is configurable and default behavior kept the same?

danobi: I originally had the extra option and kept the default the same. @yonghong-song suggested to change the default. I can do either -- just let me know.

yonghong-song: I suggest to change the default as it uses less memory and generally good. But if we are afraid of any breakage, we can keep the original as default, make the new implementation configurable, and after some baking and testing, and then flip.

palmtenor: I wish we could keep the default behavior. There are a lot of possibilities of where an executable is (like it could be in NFS etc.) and going to the fs for every symbol may not be a very good default thing to do?

src/cc/bcc_elf.c
```diff
@@ -911,6 +928,32 @@ int bcc_elf_get_buildid(const char *path, char *buildid)
...
+                       size_t str_table_idx, char *out, size_t len)
+{
+  Elf *e;
+  int fd, err = 0;
+  const char *name;
+
+  if (!out || !len)
+    return -1;
+
+  if (openelf(path, &e, &fd) < 0)
```
palmtenor: Can we have the `Module` just keep an opened ELF instance

src/cc/bcc_elf.c
```diff
@@ -911,6 +928,32 @@ int bcc_elf_get_buildid(const char *path, char *buildid)
...
+    return -1;
+
+  if ((name = elf_strptr(e, section_idx, str_table_idx)) == NULL) {
+    err = -1;
+    goto exit;
+  }
+
+  strncpy(out, name, len);
```
palmtenor: So that we can just refer to the string pointer here directly?

yonghong-song:
[buildbot, test this please]

danobi:
@palmtenor , sure I will add an option. Still trying to make all the tests pass first. Will address comments after.

yonghong-song:
[buildbot, test this please]

danobi:
[buildbot, test this please]

drzaeus77:
[buildbot, ok to test]
Sorry if the buildbot commands are confusing:
"ok to test" means this PR is ok to test, and subsequent pushes will trigger new builds
"test this please" means this PR will be tested once, with no implication for subsequent pushes.

src/cc/bcc_elf.c
danobi: Unless I misunderstand, not with the new changes.

src/cc/bcc_elf.c
danobi: Yes, but I think that would make the bcc_elf API inconsistent. It seems like all the other functions take a path to the binary. Do you think that's a good enough reason to keep it slightly inefficient?

yonghong-song:
fc28 passed and other systems failed. some compilation errors, probably due to old compilers.

danobi:
Yes, it seems designated initializer support is pretty bad for C++.

danobi:
I'm seeing RSS reduction from
```
Maximum resident set size (kbytes): 291780
```
to 
```
Maximum resident set size (kbytes): 189076
```

with this change:
```
diff --git a/trunk/src/bpftrace.git-trunk/src/bpftrace.cpp b/trunk/src/bpftrace.git-trunk/src/bpftrace.cpp
index 6cb3d49..bac2e5f 100644
--- a/trunk/src/bpftrace.git-trunk/src/bpftrace.cpp
+++ b/trunk/src/bpftrace.git-trunk/src/bpftrace.cpp
@@ -1835,6 +1835,7 @@ std::string BPFtrace::resolve_usym(uintptr_t addr, int pid, bool show_offset, bo
   // TODO: deal with these:
   symopts = {.use_debug_file = true,
              .check_debug_file_crc = true,
+             .lazy_symbolize = true,
              .use_symbol_type = BCC_SYM_ALL_TYPES};
 
   if (pid_sym_.find(pid) == pid_sym_.end())
```

An improvement, but it doesn't look like it's good enough to solve my initial problem with really large binaries.

yonghong-song:
That is still a good improvement. Thanks!
Please do a rebase and resubmit. The `ProcMountNSGuard` is gone. You need to remove it and test whether it still works or not.

palmtenor:
@danobi Thanks for the updates! Will take a more detailed look at the implementation soon. On the other hand, can you share the RSS improvement without debug info and/or debug info CRC?

danobi:
```diff
diff --git a/trunk/src/bpftrace.git-trunk/src/bpftrace.cpp b/trunk/src/bpftrace.git-trunk/src/bpftrace.cpp
index 6cb3d49..2481a0d 100644
--- a/trunk/src/bpftrace.git-trunk/src/bpftrace.cpp
+++ b/trunk/src/bpftrace.git-trunk/src/bpftrace.cpp
@@ -1833,8 +1833,9 @@ std::string BPFtrace::resolve_usym(uintptr_t addr, int pid, bool show_offset, bo
   void *psyms;
 
   // TODO: deal with these:
-  symopts = {.use_debug_file = true,
-             .check_debug_file_crc = true,
+  symopts = {.use_debug_file = false,
+             .check_debug_file_crc = false,
+             .lazy_symbolize = false,
              .use_symbol_type = BCC_SYM_ALL_TYPES};
 
   if (pid_sym_.find(pid) == pid_sym_.end())
```
```
Maximum resident set size (kbytes): 247316
```
---
```diff
diff --git a/trunk/src/bpftrace.git-trunk/src/bpftrace.cpp b/trunk/src/bpftrace.git-trunk/src/bpftrace.cpp
index 6cb3d49..8f277b5 100644
--- a/trunk/src/bpftrace.git-trunk/src/bpftrace.cpp
+++ b/trunk/src/bpftrace.git-trunk/src/bpftrace.cpp
@@ -1833,8 +1833,9 @@ std::string BPFtrace::resolve_usym(uintptr_t addr, int pid, bool show_offset, bo
   void *psyms;
 
   // TODO: deal with these:
-  symopts = {.use_debug_file = true,
-             .check_debug_file_crc = true,
+  symopts = {.use_debug_file = false,
+             .check_debug_file_crc = false,
+             .lazy_symbolize = true,
              .use_symbol_type = BCC_SYM_ALL_TYPES};
 
   if (pid_sym_.find(pid) == pid_sym_.end())

```
```
Maximum resident set size (kbytes): 177644
```

danobi:
@yonghong-song rebased and looks like tests still pass.

yonghong-song:
LGTM. Maybe further optimization is possible for `find_addr` where loading the whole symbol table still happens. This will be mostly useful for uprobe to find address of a particular user function. We can do this optimization later.

PR #2437, opened 2019-07-03
Python linting: Upgrade from pep8 to flake8

cclauss:
As suggested at https://github.com/iovisor/bcc/pull/2431#discussion_r298824303

There are a _ton_ (61) of error codes to eliminate from the __.flake8__ file but many (40) of them could be autocorrected by running the codebase through [__python/black__](https://github.com/python/black).

Tests are failing because of #2445

PR #2438, opened 2019-07-03
Upgrade from pep8 to flake8 and black

cclauss:
A far more invasive alternative to #2437 which touches more that 200 files by running the [__python/black__](https://github.com/python/black) formatter on all Python files in the codebase.  The __.flake8__ file has much fewer issues (61 --> 21) to resolve than #2437 but this PR is harder to review because of the large number of files modified.

Tests are failing because of #2445


examples/hello_world.py
```diff
@@ -9,4 +9,6 @@
 from bcc import BPF
 
 # This may not work for 4.17 on x64, you need replace kprobe__sys_clone with kprobe____x64_sys_clone
-BPF(text='int kprobe__sys_clone(void *ctx) { bpf_trace_printk("Hello, World!\\n"); return 0; }').trace_print()
+BPF(
+    text='int kprobe__sys_clone(void *ctx) { bpf_trace_printk("Hello, World!\\n"); return 0; }'
+).trace_print()
```
yonghong-song: This intentionally to be one-liner. Could you restore to one line and add necessary comments so that flake8 will ignore this one?

yonghong-song:
[buildbot, test this please]

CONTRIBUTING-SCRIPTS.md
```diff
@@ -34,7 +34,7 @@ A checklist for bcc tool development:
...
 1. **Default output <80 chars wide**. Try hard to keep the output less than 80 characters wide, especially the default output of the tool. That way, the output not only fits on the smallest reasonable terminal, it also fits well in slide decks, blog posts, articles, and printed material, all of which help education and adoption. Publishers of technical books often have templates they require books to conform to: it may not be an option to shrink or narrow the font to fit your output.
 1. **Short tool name**. Follow the style of the other tools, which follow the style of other /usr/bin utilities. They are short and easy to type. No underscores.
-1. **Use pep8 to check Python style**: pep8 --show-source --ignore=E123,E125,E126,E127,E128,E302 filename . Note that it misses some things, like consistent usage, so you'll still need to double check your script.
+1. **Use flake8 to check Python style**: flake8 filename(s) . Note that it misses some things, like consistent usage, so you'll still need to double check your script.
```
pchaigno: Does this need to be executed from the root of the repository to use the `.flake8` configuration?

examples/networking/distributed_bridge/main.py
```diff
@@ -74,19 +85,39 @@ def start(self):
...
+                        "netns",
+                        "exec",
+                        "host%d" % i,
+                        "ping",
+                        host0_br_ips[j],
+                        "-c",
+                        "3",
+                        "-i",
+                        "0.2",
+                        "-q",
```
pchaigno: Maybe we should ignore that one as well? It's really not helping making the code clearer.

examples/networking/distributed_bridge/tunnel.py
```diff
@@ -32,21 +32,37 @@
...
     conf[c_int(1)] = c_int(vx.index)
 
     ipr.tc("add", "ingress", vx.index, "ffff:")
-    ipr.tc("add-filter", "bpf", vx.index, ":1", fd=ingress_fn.fd,
-           name=ingress_fn.name, parent="ffff:", action="drop", classid=1)
+    ipr.tc(
+        "add-filter",
+        "bpf",
+        vx.index,
+        ":1",
```
pchaigno: Same for this one and any other that reads like a command line.

examples/networking/distributed_bridge/tunnel_mesh.py
```diff
@@ -116,24 +157,36 @@ def run():
...
+                "/tmp/dhcp_%d_%d/dhclient.lease" % (host_id, j),
+            ]
             d_client.append(Popen(cmd, stdout=PIPE, stderr=PIPE))
 
             # make sure we get address for eth0
             retry = -1
             while retry < 0:
-                check = Popen(["ip", "addr", "show", "br%d" % j], stdout=PIPE, stderr=PIPE)
+                check = Popen(
+                    ["ip", "addr", "show", "br%d" % j], stdout=PIPE, stderr=PIPE
```
pchaigno: Uh. What's the rule it's following here?

examples/networking/http_filter/http-parse-complete.py
```diff
@@ -93,206 +97,221 @@ def help():
...
-#dictionary containing association <key(ipsrc,ipdst,portsrc,portdst),payload_string>
-#if url is not entirely contained in only one packet, save the firt part of it in this local dict
-#when I find \r\n in a next pkt, append and print all the url
+# dictionary containing association <key(ipsrc,ipdst,portsrc,portdst),payload_string>
+# if url is not entirely contained in only one packet, save the firt part of it in this local dict
+# when I find \r\n in a next pkt, append and print all the url
 local_dictionary = {}
 
 while 1:
-  #retrieve raw packet from socket
```
pchaigno: We might want to disable the rule on indentations too, at least in some subdirectories. Otherwise, the whole file is changed and it's just messing with git blame. It probably matters less for examples anyway.

src/cc/frontends/p4/compiler/ebpfAction.py
```diff
@@ -60,24 +70,27 @@ def __init__(self, p4action, program):
 
         self.builtin = False
         self.invalid = False  # a leaf action which is never
-                              # called from a table can be invalid.
+        # called from a table can be invalid.
```
pchaigno: This should either be ignored or the comment should be moved atop the variable assignment.

.flake8
```diff
@@ -0,0 +1,31 @@
+[flake8]
+count = True
+# max-complexity should be 10
+max-complexity = 30
+# max-line-length should be 79 (pep8) or 88 (python/black) or 127 (GitHub editor)
+max-line-length = 370
```
pchaigno: Why such a high limit? I'd expect it to be 80.

src/cc/frontends/p4/compiler/ebpfTable.py
```diff
@@ -84,11 +89,12 @@ def __init__(self, match_fields, program):
...
-            if ((matchType is p4_match_type.P4_MATCH_TERNARY) or
-                (matchType is p4_match_type.P4_MATCH_LPM) or
-                (matchType is p4_match_type.P4_MATCH_RANGE)):
-                raise NotSupportedException(
-                    False, "Match type {0}", matchType)
+            if (
+                (matchType is p4_match_type.P4_MATCH_TERNARY)
+                or (matchType is p4_match_type.P4_MATCH_LPM)
+                or (matchType is p4_match_type.P4_MATCH_RANGE)
+            ):
```
pchaigno: I really think we should not do that.

tools/old/biosnoop.py
```diff
@@ -16,7 +16,8 @@
 from bcc import BPF
 
 # load BPF program
-b = BPF(text="""
+b = BPF(
```
pchaigno: We might want to ignore files under `tools/old/`. Those are just there for posterity.

cclauss:
Perhaps you should read thru the docs for __python/black__.  If you want to hand tune the formatting then __black__ is not the right choice.  This is why I kept #2437 open.

On __max-line-length__ there is a line that long in the current codebase that __black__ did not wrap so I did not either.  __max-complexity__ worries me more.

pchaigno:
> Perhaps you should read thru the docs for **python/black**. If you want to hand tune the formatting then **black** is not the right choice. This is why I kept #2437 open.

Can't we avoid the above mentioned error without fine tuning? You already added a list of ignored errors; I imagine the errors I pointed to above correspond to a number too, no?

> On **max-line-length** there is a line that long in the current codebase that **black** did not wrap so I did not either.

If it's a single ligne, maybe we could just ignore it?

> **max-complexity** worries me more.

Is that the cyclomatic complexity? If it is, we probably shouldn't feel obligated to limit ourselves to 10. 9 predicates is not necessarily a big number for a single function. It's also a different issue than what the rest of this pull request addresses (not that it doesn't matter).

CONTRIBUTING-SCRIPTS.md
cclauss: http://flake8.pycqa.org/en/latest/user/configuration.html#configuration-locations

cclauss:
__black__ reformats the code completely ignoring all of those error codes because they are __flake8__ error codes.  The [__black --exclude__ option](https://black.readthedocs.io/en/stable/installation_and_usage.html) might be of interest.  Baring that, __black__ really means it when it calls itself _uncompromising__ so let's shift our focus to #2437 because running __black__ and than hand tuning the output is not the way to go.

CONTRIBUTING-SCRIPTS.md
pchaigno: Ok. Then, it might be worth mentioning the command needs to be executed within bcc (peeps might be developing their new tools outside the repo. first). What do you think?

Could you also remove the white space before the dot in that sentence?

pchaigno:
> **black** reformats the code completely ignoring all of those error codes because they are **flake8** error codes. The [**black --exclude** option](https://black.readthedocs.io/en/stable/installation_and_usage.html) might be of interest. Baring that, **black** really means it when it calls itself _uncompromising_ so let's shift our focus back to #2437 because running **black** and than hand tuning the output is not the way to go.

I agree. Thanks for the explanation!

yonghong-song:
The command line `./scripts/check-helpers.sh` failed. Sorry that we have not run this script for a while and it is current broken due to introducing git submodule libbpf. If you get cycles, maybe you could fix this scripts? Otherwise, it is okay to disable this script run at this patch and I will fix it later. Thanks!

cclauss:
Makes sense...  I added a __TODO__ comment above the command.

yonghong-song:
[buildbot, test this please]

drzaeus77:
Sorry, buildbot is still down, I'm having to re-install and re-configure from scratch.

yonghong-song:
okay.

yonghong-song:
[buildbot, test this please]

yonghong-song:
@cclauss Looks like travis ci succeeded for this. Can we break this into multiple patches? For example, first for tools directory. We can do other directories later.

.flake8
cclauss: This is the current state of the code.  Lines could be wrapped and this number lowered in future pull requests.

cclauss:
Are you saying separate PRs?

yonghong-song:
yes, separate and successive PRs.

yonghong-song:
@cclauss did you get some time to work on this? Since travis ci is enabled, all tests failed with travis ci :-(

cclauss:
@yonghong-song The first PR will be #2431.  Once that PR is landed then all the Travis tests will pass and we can add these changes piece-by-piece.

yonghong-song:
Done. Thanks.

PR #2441, opened 2019-07-04
Resolve global object symbols

jasonkeene:
I am trying to resolve a symbol for a global variable in a Go program. Something like this:

```go
package main

var counter int64

func main() {}
```

I want to look up where `main.counter` is. When I dump the symbol table using readelf it is there:

```
$ readelf -s /tmp/counter | grep main.counter
   156: 00000000005ec228     8 OBJECT  GLOBAL DEFAULT   11 main.counter
```

When I call into `bcc_resolve_symname` it finds the symbol but it gets caught up calling into `bcc_elf_foreach_load_section`. When it calls into the callback `_find_load` this is the state:

```
addr->target_addr: 6210088
v_addr: 4194304
mem_sz: 763136
file_offset: 0
```

This condition evaluates to:

```
addr->target_addr >= v_addr && addr->target_addr < (v_addr + mem_sz)
6210088 >= 4194304 && 6210088 < (4194304 + 763136)
6210088 >= 4194304 && 6210088 < 4957440
true && false
false
```

Which means `binary_addr` does not get set.

You can see the `addr->target_addr` is the value I want (6210088 == 0x5ec228). I made this one line change to set the `binary_addr` to the `target_addr` if it was not set by `bcc_elf_foreach_load_section`. This fixed the issue for me. To be frank, I have very little understanding of ELF so I am not confident that this change is correct. I am opening this as a PR since I have a working change but I mainly want to start a conversation about this since I'm out of my depth a bit.

### DIFF
--- a/src/cc/bcc_elf.c
+++ b/src/cc/bcc_elf.c
@@ -630,7 +630,7 @@ int bcc_elf_foreach_load_section(const char *path,
   for (i = 0; i < nhdrs; i++) {
     if (!gelf_getphdr(e, (int)i, &header))
       continue;
-    if (header.p_type != PT_LOAD || !(header.p_flags & PF_X))
+    if (header.p_type != PT_LOAD)
       continue;
     res = callback(header.p_vaddr, header.p_memsz, header.p_offset, payload);
     if (res < 0) {


palmtenor:
I don’t think this is the right fix since it means we are going to resolve address against module regardless of address range:)
Would you mind sharing the `readelf -Wl` result of your binary and `/proc/PID/maps` of the Process to see what’s actually going on?

jasonkeene:
Cool, I was pretty sure it wasn't the right fix but didn't know where to go from here. Here is the output of readelf: 

```
$ readelf -Wl /tmp/counter

Elf file type is EXEC (Executable file)
Entry point 0x454380
There are 7 program headers, starting at offset 64

Program Headers:
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
  PHDR           0x000040 0x0000000000400040 0x0000000000400040 0x000188 0x000188 R   0x1000
  NOTE           0x000f9c 0x0000000000400f9c 0x0000000000400f9c 0x000064 0x000064 R   0x4
  LOAD           0x000000 0x0000000000400000 0x0000000000400000 0x0ba500 0x0ba500 R E 0x1000
  LOAD           0x0bb000 0x00000000004bb000 0x00000000004bb000 0x0f6ca6 0x0f6ca6 R   0x1000
  LOAD           0x1b2000 0x00000000005b2000 0x00000000005b2000 0x01dda0 0x03cb38 RW  0x1000
  GNU_STACK      0x000000 0x0000000000000000 0x0000000000000000 0x000000 0x000000 RW  0x8
  LOOS+0x5041580 0x000000 0x0000000000000000 0x0000000000000000 0x000000 0x000000     0x8

 Section to Segment mapping:
  Segment Sections...
   00
   01     .note.go.buildid
   02     .text .note.go.buildid
   03     .rodata .typelink .itablink .gosymtab .gopclntab
   04     .noptrdata .data .bss .noptrbss
   05
   06
```

and maps when it is running:

```
$ cat /proc/$(pgrep counter)/maps
00400000-004bb000 r-xp 00000000 08:01 526004                             /tmp/counter
004bb000-005b2000 r--p 000bb000 08:01 526004                             /tmp/counter
005b2000-005d0000 rw-p 001b2000 08:01 526004                             /tmp/counter
005d0000-005ef000 rw-p 00000000 00:00 0
c000000000-c004000000 rw-p 00000000 00:00 0
7f63a3906000-7f63a5bb7000 rw-p 00000000 00:00 0
7ffc1279a000-7ffc127bb000 rw-p 00000000 00:00 0                          [stack]
7ffc127f6000-7ffc127f9000 r--p 00000000 00:00 0                          [vvar]
7ffc127f9000-7ffc127fb000 r-xp 00000000 00:00 0                          [vdso]
ffffffffff600000-ffffffffff601000 r-xp 00000000 00:00 0                  [vsyscall]
```

jasonkeene:
It looks like the `_find_load` callback only gets called once with this section:

```
LOAD           0x000000 0x0000000000400000 0x0000000000400000 0x0ba500 0x0ba500 R E 0x1000
```

Which the global is not within. It is actually in this section:

```
LOAD           0x1b2000 0x00000000005b2000 0x00000000005b2000 0x01dda0 0x03cb38 RW  0x1000
```

Which makes sense since it is marked as RW. And the range it is located in maps is here:

```
005d0000-005ef000 rw-p 00000000 00:00 0
```

I'm going to do some more reading to see why `_find_load` only gets called once.

jasonkeene:
Ah, it looks like there is a check to make sure the section is executable:

https://github.com/iovisor/bcc/blob/6836168f3894d5dc559ab167dc18e9bdea6107db/src/cc/bcc_elf.c#L633-L634

Since this global variable is in a RW section and not an X section this makes sense. It looks like `bcc_elf_foreach_load_section` is only called in two locations within the library, the `bcc_resolve_symname` function and here:

https://github.com/iovisor/bcc/blob/aa1b904e2aecdfceb838c416dc095f9a7e9cedf7/src/cc/bcc_syms.cc#L124-L137

I am just going to remove this check and update the PR, however, since this behavior could be relied upon by callers it might cause problems for folks. An alternative would be to introduce a new function with this new behavior, or a global flag that could be set to skip the executable check.

Let me know what you think.

jasonkeene:
So, doing this still gives me the wrong address for my use case. `bcc_elf_foreach_load_section` processes all the sections now but the math done in the `_find_load` callback gives me the wrong address:

https://github.com/iovisor/bcc/blob/6836168f3894d5dc559ab167dc18e9bdea6107db/src/cc/bcc_syms.cc#L616

It resolves the symbol to the address `0x1ec228` when it needs to be `0x5ec228`.

```
addr->binary_addr = addr->target_addr - v_addr + file_offset;
addr->binary_addr = 0x5ec228 - 0x5b2000 + 0x1b2000;
addr->binary_addr = 0x1ec228;
```

palmtenor:
I'm a little bit confused here. Where did you get the address from? As you can see it's not in an executable section and not loaded as an executable mapping, so it's very unlikely you get it from a stack-trace, which is what the BCC symbol resolution helpers designed for. Are you just using `bcc_resolve_symname` to resolve your own address? Can you explain a bit what you are trying to do here?

The `0x1ec228` result you get is expected behavior. It is the actual, physical offset of that symbol in the file, which symbol tables for executable (`ET_EXEC`) files records virtual loaded address. We do this to because we needed it in various places (uprobe, USDT, etc.) and in the symbol resolution path we compensate for it correctly (to symbolize stack-traces).

As you can see these helpers are tailored toward BCC's specific use cases (like filtering on executable load section, binary offset, etc.), and probably are not friendly to use as a generic-purpose symbol resolution library. "Fixing" them would just creating more trouble on both side. If you have very specific usage of just one-off resolving things maybe you can just call `bcc_elf_foreach_sym` from `bcc_elf.h` which just helps you iterate over the symbol table.

jasonkeene:
> Where did you get the address from?

I got it from `readelf -s` which shows the symbol as present and the addr. I also used gdb and observed the same addr value.

> As you can see it's not in an executable section and not loaded as an executable mapping, so it's very unlikely you get it from a stack-trace, which is what the BCC symbol resolution helpers designed for.

I see. I came upon this via bpftrace's use of `bcc_resolve_symname` here:

https://github.com/iovisor/bpftrace/blob/2f7dc75569a155af7dfc5966c2ce7d135c23c11a/src/bpftrace.cpp#L1744-L1748

This is used to implement their `uaddr` builtin function that allows for user space address resolution:

https://github.com/iovisor/bpftrace/blob/43c4a66a2c6399d22d7c5d86de72f24d77bcf6d8/docs/reference_guide.md#9-uaddr-address-resolution-user-level

> Are you just using bcc_resolve_symname to resolve your own address? Can you explain a bit what you are trying to do here?

Yes, I am trying to resolve a symbol in a program that I authored. The symbol is a global variable, a simple counter. I want my BPF program to copy the contents at this memory location into a table.

palmtenor:
@jasonkeene Thanks for the explanation. While the changes in the PR may not be the best fit, I think it is definitely a valid use case to resolve a global object symbol. In fact, we do exactly the same thing in one of our example tools. You can see here: https://github.com/iovisor/bcc/blob/master/examples/cpp/pyperf/PyPerfUtil.cc#L141 .

jasonkeene:
Thanks for the example and all the help! I'm closing this since https://github.com/iovisor/bpftrace/pull/811 should fix my problem.

PR #2475, opened 2019-07-30
tools/tcpconnect: add option -c to count connects

boat0:
Add -c to count all active connections per dest ip/port so we can
easily spot the heavy outbound connection attempts.

    # ./tcpconnect.py -c
    Tracing connect ... Hit Ctrl-C to end
    ^C
    LADDR                 RADDR:RPORT             CONNECTS
    192.168.10.50    <-> 172.217.21.194:443         70
    192.168.10.50    <-> 172.213.11.195:443         34
    192.168.10.50    <-> 172.212.22.194:443         21
    [...]

### DIFF
--- a/man/man8/tcpconnect.8
+++ b/man/man8/tcpconnect.8
@@ -2,7 +2,7 @@
 .SH NAME
 tcpconnect \- Trace TCP active connections (connect()). Uses Linux eBPF/bcc.
 .SH SYNOPSIS
-.B tcpconnect [\-h] [\-t] [\-x] [\-p PID] [-P PORT]
+.B tcpconnect [\-h] [\-c] [\-t] [\-x] [\-p PID] [-P PORT]
 .SH DESCRIPTION
 This tool traces active TCP connections (eg, via a connect() syscall;
 accept() are passive connections). This can be useful for general
@@ -24,6 +24,9 @@ Print usage message.
 .TP
 \-t
 Include a timestamp column.
+.TP
+\-c
+Count connects per src ip and dest ip/port.
 .TP
 \-p PID
 Trace this process ID only (filtered in-kernel).
@@ -53,16 +56,18 @@ Trace PID 181 only:
 Trace ports 80 and 81 only:
 #
 .B tcpconnect \-P 80,81
-.SH FIELDS
 .TP
 Trace all TCP connects, and include UID:
 #
 .B tcpconnect \-U
-.SH FIELDS
 .TP
 Trace UID 1000 only:
 #
 .B tcpconnect \-u 1000
+.TP
+Count connects per src ip and dest ip/port:
+#
+.B tcpconnect \-c
 .SH FIELDS
 .TP
 TIME(s)
@@ -88,11 +93,14 @@ Destination IP address.
 .TP
 DPORT
 Destination port
+.TP
+CONNECTS
+Accumulated active connections since start.
 .SH OVERHEAD
 This traces the kernel tcp_v[46]_connect functions and prints output for each
 event. As the rate of this is generally expected to be low (< 1000/s), the
 overhead is also expected to be negligible. If you have an application that
-is calling a high rate of connects()s, such as a proxy server, then test and
+is calling a high rate of connect()s, such as a proxy server, then test and
 understand this overhead before use.
 .SH SOURCE
 This is from bcc.
--- a/tools/tcpconnect.py
+++ b/tools/tcpconnect.py
@@ -4,7 +4,7 @@
 # tcpconnect    Trace TCP connect()s.
 #               For Linux, uses BCC, eBPF. Embedded C.
 #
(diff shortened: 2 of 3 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

man/man8/tcpconnect.8
```diff
@@ -25,6 +25,9 @@ Print usage message.
 \-t
 Include a timestamp column.
 .TP
+\-c
+Count connects per dest ip/port.
```
yonghong-song: Could you use "connections" here, which is more common usage, instead of "connects"? Same for below several other cases.

man/man8/tcpconnect.8
```diff
@@ -53,16 +56,18 @@ Trace PID 181 only:
...
 Trace all TCP connects, and include UID:
 #
 .B tcpconnect \-U
-.SH FIELDS
 .TP
 Trace UID 1000 only:
 #
 .B tcpconnect \-u 1000
+.TP
+Count connects per dest ip/port:
```
yonghong-song: It should be source ip + dest ip/port?

tools/tcpconnect.py
```diff
@@ -168,7 +171,56 @@
...
+               data4.dport = ntohs(dport);
+               bpf_get_current_comm(&data4.task, sizeof(data4.task));
+               ipv4_events.perf_submit(ctx, &data4, sizeof(data4));"""
+               },
+        'ipv6':
+        { 'count' :
+               """
+               struct ipv6_flow_key_t flow_key = {};
+               flow_key.saddr = skp->__sk_common.skc_rcv_saddr;
+               flow_key.daddr = skp->__sk_common.skc_daddr;
```
yonghong-song: The above to calculate ipv6 src/dest addresses are not correct. Do you have environment to test it?

man/man8/tcpconnect.8
boat0: yes. will fix it

man/man8/tcpconnect.8
boat0: "connects" may be more accurate than connections in this context? because we inspect the connect actions, not any already established connections.

tools/tcpconnect.py
boat0: copy and paste error. will fix it.

man/man8/tcpconnect.8
yonghong-song: We could use "new connections" or "connects". I guess "connect" okay which is short and also captures the essence.

yonghong-song:
Thanks for the fix. I have one more request for this.
On an ipv6 machine, I got:
```
-bash-4.4$ sudo ./tcpconnect.py -c
Tracing connect ... Hit Ctrl-C to end
^C
LADDR                     RADDR:RPORT               CONNECTS  
127.0.0.1            <-> 127.0.0.1:1456                5
127.0.0.1            <-> 127.0.0.1:2377                1
2401:db00:1120:405d:face:0:1d:0 <-> 2401:db00:1120:4190:face:0:3d:0:6969         22
2401:db00:1120:405d:face:0:1d:0 <-> 2401:db00:1120:6067:face:0:39:0:6969         18
2401:db00:1120:405d:face:0:1d:0 <-> 2401:db00:1120:405d:face:0:1d:0:4330          2
::1                  <-> ::1:16262                     2
...
```

You can see ipv4 and ipv6 addresses are printed differently.
ip4 address/port like `127.0.0.1:1456` looks okay with only one `:` as the separator.
ipv6 address/port separation is not really clear.
Could you print RADDR and RPORT separately just like plain `tccconnect.py` does?

brendangregg:
FWIW, I would not want to see this behavior added to all the tools. That the tools are (or were) simple and have a single function is a benefit that aids learning, and makes them double as code examples. I'd add switches like this only when necessary.

boat0:
> Could you print RADDR and RPORT separately just like plain `tccconnect.py` does?
ok.

boat0:
> FWIW, I would not want to see this behavior added to all the tools. That the tools are (or were) simple and have a single function is a benefit that aids learning, and makes them double as code examples. I'd add switches like this only when necessary.

Got your point. Plan is to integrate various BCC tools into production monitoring systems. So '-c' will make tcpconnect(maybe other tools as well) more efficient and more succinct. Personally I think this feature will be useful to others.

boat0:
> You can see ipv4 and ipv6 addresses are printed differently.
> ip4 address/port like `127.0.0.1:1456` looks okay with only one `:` as the separator.
> ipv6 address/port separation is not really clear.
> Could you print RADDR and RPORT separately just like plain `tccconnect.py` does?

OK. Resubmitted.

yonghong-song:
Looks good. Thanks!

PR #2476, opened 2019-07-30
Add allow_partial_usdt_init param so partial init of USDT vector is possible

davemarchevsky:
Looking for early feedback here before I write tests / flesh this out more.

Consider this scenario: I'm hoping to attach USDTs to many short-lived processes, let's say 30. I want to gather data from USDT handlers on these processes for 5 seconds. It's guaranteed that at least one of the processes that was alive when I constructed the vector of USDTs to pass into `BPF::init` will not be alive by the time the 5 seconds are up. I'm OK with this and would still like to gather as much data as possible.

If the process dies after `USDT::init()` is successful but before `attach_usdt` is called, we're already OK because we can handle each `StatusTuple` independently outside of `bcc`. Similarly, if process dies between `attach_usdt` and `detach_usdt` calls, we're fine for same reason.

But if it dies between construction of `USDT` object and `USDT::init()` being called by `BPF::init()`, we currently consider failure to init any `USDT` critical enough to stop the whole `BPF::init()`. 

This PR adds a `allow_partial_usdt_init` param to `BPF::init` to let users of `bcc` signal that they're OK with `USDT::init()` failures. `USDT::initialized()` getter is added so `bcc` user can check which `USDT`s succeeded / failed.

One downside of this initial pass is that the info in `StatusTuple`s for `USDT::init` is lost. So if we have some `USDT::init` failures when `allow_partial_usdt_init` is `true` it'll be hard to get visibility into them. 

/cc @yonghong-song

### DIFF
--- a/src/cc/api/BPF.cc
+++ b/src/cc/api/BPF.cc
@@ -55,18 +55,33 @@ std::string sanitize_str(std::string str, bool (*validator)(char),
   return str;
 }
 
+StatusTuple BPF::init_usdt(const USDT& usdt) {
+  USDT u(usdt);
+  StatusTuple init_stp = u.init();
+  if (init_stp.code() != 0) {
+    return init_stp;
+  }
+
+  usdt_.push_back(std::move(u));
+  all_bpf_program_ += usdt_.back().program_text_;
+  return StatusTuple(0);
+}
+
+void BPF::init_fail_reset() {
+  usdt_.clear();
+  all_bpf_program_ = "";
+}
+
 StatusTuple BPF::init(const std::string& bpf_program,
                       const std::vector<std::string>& cflags,
                       const std::vector<USDT>& usdt) {
-  std::string all_bpf_program;
-
   usdt_.reserve(usdt.size());
   for (const auto& u : usdt) {
-    usdt_.emplace_back(u);
-  }
-  for (auto& u : usdt_) {
-    TRY2(u.init());
-    all_bpf_program += u.program_text_;
+    StatusTuple init_stp = init_usdt(u);
+    if (init_stp.code() != 0) {
+      init_fail_reset();
+      return init_stp;
+    }
   }
 
   auto flags_len = cflags.size();
@@ -74,9 +89,11 @@ StatusTuple BPF::init(const std::string& bpf_program,
   for (size_t i = 0; i < flags_len; i++)
     flags[i] = cflags[i].c_str();
 
-  all_bpf_program += bpf_program;
-  if (bpf_module_->load_string(all_bpf_program, flags, flags_len) != 0)
+  all_bpf_program_ += bpf_program;
+  if (bpf_module_->load_string(all_bpf_program_, flags, flags_len) != 0) {
+    init_fail_reset();
     return StatusTuple(-1, "Unable to initialize BPF program");
+  }
 
   return StatusTuple(0);
 };
--- a/tests/cc/test_usdt_probes.cc
+++ b/tests/cc/test_usdt_probes.cc
@@ -99,6 +99,32 @@ TEST_CASE("test find a probe in our process' shared libs with c++ API", "[usdt]"
   REQUIRE(res.code() == 0);
 }
 
+TEST_CASE("test usdt partial init w/ fail init_usdt", "[usdt]") {
+  ebpf::BPF bpf;
+  ebpf::USDT u(::getpid(), "libbcc_test", "sample_lib_probe_nonexistent", "on_event");
+  ebpf::USDT p(::getpid(), "libbcc_test", "sample_lib_probe_1", "on_event");
+
+  // We should be able to fail initialization and subsequently do bpf.init w/o USDT
+  // successfully
+  auto res = bpf.init_usdt(u);
+  REQUIRE(res.msg() != "");
+  REQUIRE(res.code() != 0);
+
+  // Shouldn't be necessary to re-init bpf object either after failure to init w/
+  // bad USDT
+  res = bpf.init("int on_event() { return 0; }", {}, {u});
+  REQUIRE(res.msg() != "");
+  REQUIRE(res.code() != 0);
+
+  res = bpf.init_usdt(p);
+  REQUIRE(res.msg() == "");
+  REQUIRE(res.code() == 0);
+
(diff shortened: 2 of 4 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

anakryiko:
Would it be better to have BPF::init and BPF::init_usdt (called for each USDT separately) better and more flexible solution? That way application can decide for itself which USDTs are critical and which can be ignored, etc.

davemarchevsky:
@anakryiko This sounds better to me, I was optimizing for "least change to API surface". Some clarification questions:
* Are you proposing that `BPF::init` no longer accept a vector of USDTs? And therefore that its functionality be split out into `BPF::init_usdt`?
   * Alternatively, we can have `BPF::init` accept USDTs but consider those required to init successfully but allow you to call `BPF::init_usdt` before calling `BPF::init` for non-success-required USDT initialization. This would preserve current functionality while enabling my use-case.

I'm going to take a stab at the approach proposed in "Alternatively," unless anyone's opposed.

anakryiko:
@davemarchevsky, yeah, I was thinking about second approach, no need to break existing behavior. Thanks!

palmtenor:
The only reason we do the `USDT::init` in `BPF::init` is that we couldn't copy the initialized `probe_` pointer, otherwise user would just initialize the `USDT` instances outside, pick those succeeded and pass (move) them into `BPF::init`.

I'm OK with Andrii's idea. We can also add a `BPF::add_usdt_info` call which need to happen before `BPF::init` but can fail individually so that users can just add USDTs one by one and choose to ignore the failures.

davemarchevsky:
New commit implements @anakryiko's suggestion.

* Remove no-longer-needed `USDT::initialized()` getter
* Add `BPF::init_usdt`: tries to init USDT while being tolerant of failure. Should be called before `BPF::init`
* Add  `BPF::init_fail_reset`, contains common logic for "We had a failure somewhere in `BPF::init`, reset state that we've modified while `init`ing so we can try again.
  * This fixes an additional issue that's maybe a bug. I have a few programs that try doing `BPF::init` with some `USDT`s and, if there's an issue `init`ing them, try doing `BPF::init` again without the `usdts` arg. Apparently this never worked because `usdt_` vector is never cleared so next attempt to do `BPF::init` would result in `USDT::init` eventually being called on the `USDT` which failed to init in the first place (and the program text `all_bpf_program` being repopulated with duplicate bpf helper functions, which would additionally break things if compilation step was ever reached). Anyways, if we expect folks to be able to do a second `BPF::init` call after first one fails we must clean some things up to make it possible. 

TODO:
- [ ] maybe better name for `BPF::init_fail_reset`
- [ ] add comments explaining what `BPF::init_usdt` does to `BPF.h`, appropriate docs
- [ ] add test
- [ ] validate for my usecase

yonghong-song:
I looked at the code. Looks good to me. Function name init_fail_reset() is okay to me. We do need some doc about what is the typical way to use the new API and a test case to actually show how the new API is used.

src/cc/api/BPF.cc
```diff
@@ -55,28 +55,44 @@ std::string sanitize_str(std::string str, bool (*validator)(char),
   return str;
 }
 
+StatusTuple BPF::init_usdt(const USDT& usdt) {
+  usdt_.emplace_back(usdt);
+  StatusTuple init_stp = usdt_.back().init();
+  if (init_stp.code() != 0) {
+    usdt_.pop_back();
```
anakryiko: is there some internal mutability inside USDT? If not, it seems weird to

1. emplace usdt
2. use back() to get it back
3. do init
4. if it fails - pop from vector

Instead of:
1. usdt.init()
2. if it succeeds - add to usdt_

Thoughts?

src/cc/api/BPF.cc
```diff
@@ -55,28 +55,52 @@ std::string sanitize_str(std::string str, bool (*validator)(char),
...
+  }
+  usdt_.push_back(std::move(u));
+
+  /*
+  usdt_.emplace_back(usdt);
+  StatusTuple init_stp = usdt_.back().init();
+  if (init_stp.code() != 0) {
+    usdt_.pop_back();
+    return init_stp;
+  }*/
```
yonghong-song: The change looks good. Could you remove the above commented code? We do not need them any more.
After this, it might be clear to make the change to have
```
   all_bpf_program_ += u.program_text_;
   usdt_.push_back(std::move(u));
```

davemarchevsky: will do! just pushed up this latest set of changes so i could do some final validation before adding some docs and poking you to do final pass :)

src/cc/api/BPF.cc
davemarchevsky: i thought there was some weirdness with the move constructor when the usdt had already been `init`ted, but that is not the case. Agreed with your suggestions and resolved.

davemarchevsky:
Addressed comments. Added a test and some docs. Re-validated against my usecase successfully. @yonghong-song ready for final pass. Will rebase after you give it the go-ahead.

tests/cc/test_usdt_probes.cc
```diff
@@ -99,6 +99,32 @@ TEST_CASE("test find a probe in our process' shared libs with c++ API", "[usdt]"
...
+  ebpf::USDT u(::getpid(), "libbcc_test", "sample_lib_probe_nonexistent", "on_event");
+  ebpf::USDT p(::getpid(), "libbcc_test", "sample_lib_probe_1", "on_event");
+
+  // We should be able to fail initialization and subsequently do bpf.init w/o USDT
+  // successfully
+  auto res = bpf.init_usdt(u);
+  REQUIRE(res.msg() != "");
+  REQUIRE(res.code() != 0);
+
+  // Shouldn't ne necessary to re-init bpf object either after failure to init w/
```
yonghong-song: ne => be

yonghong-song:
LGTM. Only one minor typo.

davemarchevsky:
@yonghong-song fixed typo, squashed, added descriptive commit msg

yonghong-song:
LGTM. Thanks!

PR #2481, opened 2019-08-04
CMake: Allow to disable building man pages

RomainNaour:
Avoid building man pages when it's not necessary
(i.e when building for an embedded system).

Signed-off-by: Romain Naour <romain.naour@gmail.com>


yonghong-song:
[buildbot, test this please]

PR #2482, opened 2019-08-05
Support perf's symfs mechanism for debuginfo lookup.

michalgr:
Symfs is a file path prefix that needs to be prepended to full path of a library
to compute location of corresponding debuginfo. It will allow bcc users to
provide debug symbols if only they can write to at least one directory. Nothing
else needs to be assumed regarding the system bcc runs on.

My intention is to make debuginfo lookup work on Android in one form or another.
None of the gdb rules works out of the box:
- native libraries are often managed by Android apps themselves and uncompressed
  by apps into private directories. Sideloading debuginfo into those directories
  might not be feasible. What's more it's desirable to remove data like
  debuglink from release builds of mobile apps. At the same time those builds
  are the most interesting ones to investigate with bcc.
- there is no /usr directory on Android, so there is no /usr/lib/debug/. It
  could be created with some effort, but symfs is much more convenient and
  already used when profiling with simpleperf (fork of perf for Android).

### DIFF
--- a/src/cc/bcc_elf.c
+++ b/src/cc/bcc_elf.c
@@ -546,6 +546,54 @@ static char *find_debug_via_buildid(Elf *e) {
   return NULL;
 }
 
+static char *find_debug_via_symfs(Elf *e, const char* path) {
+  char fullpath[PATH_MAX];
+  char buildid[128];
+  char symfs_buildid[128];
+  int check_build_id;
+  char *symfs;
+  Elf *symfs_e = NULL;
+  int symfs_fd = -1;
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

yonghong-song:
[buildbot, test this please]

yonghong-song:
The patch looks good to me. There are some build infra failures. Just restarted the buildbot.

michalgr:
thanks !

PR #2483, opened 2019-08-05
tools: fix vfscount output when no duration is specified

jeromemarchand:
Since commit a2e71a9eb71a ("vfscount.py: add args time (#2344)"),
vfscount does not show any output when it is interrupted by a key
press, which is the only way out when no time limit is specified. I
assume the exit() that has been added in the keyboard interrupt
handler is there so that no output is displayed when the program has
been interrupted early when a time limit has been specified. But since
the tool still invite the user to use Ctrl-C to end the tracing in
that case, it seems more consistent to show an output in that case
too.

This patch removes the exit() and the tools always show a result at
the end. It also adds the duration argument to the synopsis and the
option section in the man page.

### DIFF
--- a/man/man8/vfscount.8
+++ b/man/man8/vfscount.8
@@ -2,7 +2,7 @@
 .SH NAME
 vfscount \- Count VFS calls ("vfs_*"). Uses Linux eBPF/bcc.
 .SH SYNOPSIS
-.B vfscount
+.B vfscount [duration]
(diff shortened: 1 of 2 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
LGTM. Thanks for the fix!

PR #2494, opened 2019-08-15
Update mysqld_query.py

SuperSix0:
store the query in query[128] and printk("%s\n",query) instead of printk("%s\n",addr)


palmtenor:
[buildbot, ok to test]

PR #2496, opened 2019-08-16
Update reference_guide.md

SuperSix0:
I think it should be attach_uretprobe rather than attach_uprobe.


yonghong-song:
[buildbot, test this please]

PR #2500, opened 2019-08-22
Make number of stack traces configurable from command line in example…

johnaohara:
…s/tracing/mallocstacks.py

By default examles/tracing/mallocstacks.py captures 1024 stack traces before the script exits. This PR retains the default, but allows for the number of stack traces to be configured from the command line.

I am not sure if passing an unbounded/non-validated value to BPF_STACK_TRACE is sensible, or if the implementation of BPF_STACK_TRACE validates _max_entries

### DIFF
--- a/examples/tracing/mallocstacks.py
+++ b/examples/tracing/mallocstacks.py
@@ -21,6 +21,12 @@ if len(sys.argv) < 2:
     exit()
 pid = int(sys.argv[1])
 if len(sys.argv) == 3:
+    try:
+        assert int(sys.argv[2]) > 0, ""
+    except (ValueError, AssertionError) as e:
+        print("USAGE: mallocstacks PID [NUM_STACKS=1024]")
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
No, the implementation of BPF_STACK_TRACE does not validate _max_entries. Non number stack size argument will cause compilation errors. negative number may survive and will be used as a big number (need to double check). Could you check at least the number should be a positive nonzero interage? If it is a positive integer, we do not need to check the range.

johnaohara:
Great, thanks for your comments. I will update the PR

johnaohara:
I have pushed another commit that includes validation, do you prefer to have all PR commits squashed, or to keep commits separate?

yonghong-song:
[buildbot, test this please]

yonghong-song:
separate commits are okay and I can do squash and then commit.

PR #2502, opened 2019-08-29
Support for hardware offload

pchaigno:
This pull request implements hardware offload support in bcc.

To offload maps and programs to the device, `map_ifindex` (for `bpf$BPF_MAP_CREATE`) and `prog_ifindex` (`bpf$BPF_PROG_LOAD`) need to be set to the index of the network interface at creation time. This support requires a number of changes, some **breaking to the bcc APIs**.

Changes to the Python methods `load_func` and `BPF` are backward-compatible. Changes to `bpf_module_create_b`, `bpf_module_create_c`, `bpf_module_create_c_from_string`, and `bcc_func_load`, and `BPFModule::BPFModule` are not. I'm not sure which of these we want to avoid or there's a better way; I'm opening this pull request to start the discussion.

I've also updated the `xdp_drop_count.py` example to support hardware offload with the new `-H device` option.

### DIFF
--- a/src/cc/bcc_common.h
+++ b/src/cc/bcc_common.h
@@ -25,11 +25,13 @@
 extern "C" {
 #endif
 
-void * bpf_module_create_b(const char *filename, const char *proto_filename, unsigned flags);
+void * bpf_module_create_b(const char *filename, const char *proto_filename, unsigned flags,
+                           const char *dev_name);
 void * bpf_module_create_c(const char *filename, unsigned flags, const char *cflags[], int ncflags,
-                           bool allow_rlimit);
+                           bool allow_rlimit, const char *dev_name);
 void * bpf_module_create_c_from_string(const char *text, unsigned flags, const char *cflags[],
-                                       int ncflags, bool allow_rlimit);
+                                       int ncflags, bool allow_rlimit,
+                                       const char *dev_name);
 void bpf_module_destroy(void *program);
 char * bpf_module_license(void *program);
 unsigned bpf_module_kern_version(void *program);
@@ -69,7 +71,8 @@ struct bpf_insn;
 int bcc_func_load(void *program, int prog_type, const char *name,
                   const struct bpf_insn *insns, int prog_len,
                   const char *license, unsigned kern_version,
-                  int log_level, char *log_buf, unsigned log_buf_size);
+                  int log_level, char *log_buf, unsigned log_buf_size,
+                  const char *dev_name);
 
 #ifdef __cplusplus
 }
--- a/examples/networking/xdp/xdp_drop_count.py
+++ b/examples/networking/xdp/xdp_drop_count.py
@@ -16,24 +16,29 @@ flags = 0
 def usage():
     print("Usage: {0} [-S] <ifdev>".format(sys.argv[0]))
     print("       -S: use skb mode\n")
+    print("       -H: use hardware offload mode\n")
     print("e.g.: {0} eth0\n".format(sys.argv[0]))
     exit(1)
 
 if len(sys.argv) < 2 or len(sys.argv) > 3:
     usage()
 
+offload_device = None
 if len(sys.argv) == 2:
     device = sys.argv[1]
+elif len(sys.argv) == 3:
+    device = sys.argv[2]
 
+maptype = "percpu_array"
 if len(sys.argv) == 3:
     if "-S" in sys.argv:
         # XDP_FLAGS_SKB_MODE
-        flags |= 2 << 0
-
-    if "-S" == sys.argv[1]:
-        device = sys.argv[2]
-    else:
-        device = sys.argv[1]
+        flags |= (1 << 1)
+    if "-H" in sys.argv:
+        # XDP_FLAGS_HW_MODE
+        maptype = "array"
+        offload_device = device
+        flags |= (1 << 3)
 
 mode = BPF.XDP
 #mode = BPF.SCHED_CLS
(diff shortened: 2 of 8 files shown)


yonghong-song:
Since hardware offload happens rarely (at least for now), to avoid api change, maybe we can set devname can be a private variable or module, and we can have a new api to set the device name.
 
BTW, how do I test hardware offload here? I assume I do need something like nfp hardware?

pchaigno:
> Since hardware offload happens rarely (at least for now), to avoid api change, maybe we can set devname can be a private variable or module, and we can have a new api to set the device name.

Then, should we create two new functions in the C API for the creating and loading of Modules? Right now, the `bpf_module_create_xxx` functions do both at once. With two separate functions, we could set private fields (e.g., `dev_name`) in between the creation and the loading of Modules. Each new field would be a single backward-compatible change to the API.

<br>

> BTW, how do I test hardware offload here? I assume I do need something like nfp hardware?

Yes, ideally you would need nfp hardware.

Since all that is needed is an argument value change at the syscall layer, you can also check with strace. If your device doesn't support BPF offload, you should get the following (notice the `map_ifindex=if_nametoindex("eth0")`):
```
$ sudo strace -ebpf -e signal=none python examples/networking/xdp/xdp_drop_count.py -H eth0
bpf(BPF_MAP_CREATE, {map_type=BPF_MAP_TYPE_ARRAY, key_size=4, value_size=8, max_entries=256, map_flags=0, inner_map_fd=0, map_name="dropcnt", map_ifindex=if_nametoindex("eth0"), btf_fd=0, btf_key_type_id=0, btf_value_type_id=0}, 112) = -1 EOPNOTSUPP (Operation not supported)
could not open bpf map: dropcnt, error: Operation not supported
  File "examples/networking/xdp/xdp_drop_count.py", line 132, in <module>
    device=offload_device)
  File "/usr/local/lib/python2.7/dist-packages/bcc/__init__.py", line 347, in __init__
+++ exited with 1 +++
```
To check the map loading, you can remove `device=offload_device` from the `BPF()` call in the Python script and relaunch:
```
$ sudo strace -ebpf -e signal=none python examples/networking/xdp/xdp_drop_count.py -H eth0
bpf(BPF_MAP_CREATE, {map_type=BPF_MAP_TYPE_ARRAY, key_size=4, value_size=8, max_entries=256, map_flags=0, inner_map_fd=0, map_name="dropcnt", map_ifindex=0, btf_fd=0, btf_key_type_id=0, btf_value_type_id=0}, 112) = 3
bpf(BPF_PROG_LOAD, {prog_type=BPF_PROG_TYPE_XDP, insn_cnt=63, insns=0x7f2aa54b17d0, license="GPL", log_level=0, log_size=0, log_buf=NULL, kern_version=KERNEL_VERSION(5, 2, 2), prog_flags=0, prog_name="xdp_prog1", prog_ifindex=if_nametoindex("eth0"), expected_attach_type=BPF_CGROUP_INET_INGRESS, prog_btf_fd=0, func_info_rec_size=0, func_info=NULL, func_info_cnt=0, line_info_rec_size=0, line_info=NULL, line_info_cnt=0}, 112) = -1 EOPNOTSUPP (Operation not supported)
bpf(BPF_PROG_LOAD, {prog_type=BPF_PROG_TYPE_XDP, insn_cnt=63, insns=0x7f2aa54b17d0, license="GPL", log_level=1, log_size=65536, log_buf="", kern_version=KERNEL_VERSION(5, 2, 2), prog_flags=0, prog_name="xdp_prog1", prog_ifindex=if_nametoindex("eth0"), expected_attach_type=BPF_CGROUP_INET_INGRESS, prog_btf_fd=0, func_info_rec_size=0, func_info=NULL, func_info_cnt=0, line_info_rec_size=0, line_info=NULL, line_info_cnt=0}, 112) = -1 EOPNOTSUPP (Operation not supported)
bpf: Failed to load program: Operation not supported

  File "examples/networking/xdp/xdp_drop_count.py", line 134, in <module>
    fn = b.load_func("xdp_prog1", mode, offload_device)
  File "/usr/local/lib/python2.7/dist-packages/bcc/__init__.py", line 395, in load_func
    (func_name, errstr))
Exception: Failed to load BPF program xdp_prog1: Operation not supported
+++ exited with 1 +++
```
If you'd prefer, I can also try to find someone else with nfp hardware to do a second review.

qmonnet:
Hi, there's also the `netdevsim` test module which can probably be used to test offload with a dummy interface. It won't be able to run the program, but it should support loading it.

Thanks a ton Paul for taking care of this!

qmonnet:
Please disregard my comment on `netdevsim`, it cannot load maps or JIT-compile the program, making it too limited to check bcc tools.

I tested the PR, though, and `xdp_drop_count.py` works for me after the latest changes, both with driver and offload XDP (nfp).

src/cc/bcc_common.h
```diff
@@ -69,7 +70,8 @@ struct bpf_insn;
 int bcc_func_load(void *program, int prog_type, const char *name,
                   const struct bpf_insn *insns, int prog_len,
                   const char *license, unsigned kern_version,
-                  int log_level, char *log_buf, unsigned log_buf_size);
+                  int log_level, char *log_buf, unsigned log_buf_size,
+                  const char *dev_name);
```
yonghong-song: could you make dev_name with default value nullptr?

yonghong-song:
@qmonnet thanks for the testing. @pchaigno The change looks good. The API impact is minimum. Just a minor comment. Thanks for adding hw offload support to bcc!

src/cc/bcc_common.h
pchaigno: I don't think I can for these functions because they have C linkage. I can---and just did---for method `BPFModule::bcc_func_load` however.

yonghong-song: Sorry. I actually indeed mean the change in bpf_module.h and it looks like you just did. Great!

PR #2504, opened 2019-09-02
Support XSKMAP

sevagh:
This PR may not be comprehensive (and I'll take any suggestions to make it better), but it helped me use xskmaps successfully with the `BPF_TABLE` macro.

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -1220,6 +1220,8 @@ bool BTypeVisitor::VisitVarDecl(VarDecl *Decl) {
       table.max_entries = numcpu;
     } else if (section_attr == "maps/perf_array") {
       map_type = BPF_MAP_TYPE_PERF_EVENT_ARRAY;
+    } else if (section_attr == "maps/xskmap") {
+      map_type = BPF_MAP_TYPE_XSKMAP;
     } else if (section_attr == "maps/cgroup_array") {
       map_type = BPF_MAP_TYPE_CGROUP_ARRAY;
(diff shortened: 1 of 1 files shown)


yonghong-song:
This should work if you want to use generic BTF_TABLE interface. 

I am not sure whether you have special need for a subclass of BTF_TABLE. One example is devmap, see commit https://github.com/iovisor/bcc/commit/6ce918bd7030241f0598ee6b1107940bf8480085.
This way, you do not need to go through this str <-> key/value conversions.

For xsk map, which should be very similar to BPF_MAP_TYPE_PROG_ARRAY and BPF_MAP_TYPE_DEVMAP as the values are all fd's. Maybe you can help refactor
the code to have a common super class to handle all these three fd-value maps
so you will only minimum code for XSKMAP.

sevagh:
So a `devmap` will work for AF_XDP? My goal is to transmit packets from my XDP program to userspace with AF_XDP.

yonghong-song:
No. devmap won't work for AF_XDP. What I mean is to have native support like BPFXSKmapTable might be better since then you do not need to do this str<->value conversion. If Indeed you want to do BPFXSKmapTable, the following way is preferred
```
class BPFFDArray : public BPFTableBase<int, int> { ...}
class BPFProgTable : public BPFFDArray {...}
class BPFDevmapTable: public BPFFDArray { ...}
class BPFXSKmapTable: public BPFFDArray { ... }
```

sevagh:
I'll give it a try, thanks.

sevagh:
I'll close this PR for now, and re-open if I have something more substantial.

PR #2511, opened 2019-09-11
support gcc 8's cold subfunctions

mwesolowski:
GCC 8 can split functions bodies into hot and cold regions, causing extra symbols with a ".cold" or ".cold.N" suffix to be emitted. Extra symbols with suffix ".cold.N" were already excluded. This PR excludes also symbols with the suffix ".cold".


palmtenor:
[buildbot, test this please]

palmtenor:
Thanks for the fix!

Issue #2512, opened 2019-09-11
funccount  error: Can't mix strings and bytes in path components


mwesolowski:
Hi, I try to trace function calls to go library and unfortunately get an error:
```
$ sudo /usr/share/bcc/tools/funccount -D 'go:os.*'
Can't mix strings and bytes in path components 
```
It looks like python type mismatch error. More detailed output:
```
$ sudo /usr/share/bcc/tools/funccount -D 'go:os.*'
Traceback (most recent call last):
  File "/usr/share/bcc/tools/funccount", line 299, in <module>
    Tool().run()
  File "/usr/share/bcc/tools/funccount", line 250, in __init__
    self.probe = Probe(self.args.pattern, self.args.regexp, self.args.pid)
  File "/usr/share/bcc/tools/funccount", line 76, in __init__
    libpath = BPF.find_exe(self.library)
  File "/usr/lib/python3.7/site-packages/bcc/__init__.py", line 269, in find_exe
    exe_file = os.path.join(path, bin_path)
  File "/usr/lib64/python3.7/posixpath.py", line 94, in join
    genericpath._check_arg_types('join', a, *p)
  File "/usr/lib64/python3.7/genericpath.py", line 151, in _check_arg_types
    raise TypeError("Can't mix strings and bytes in path components") from None
TypeError: Can't mix strings and bytes in path components
```

mwesolowski:
Actually the problem is more general. For example when I try to generate flame graph:
```
$ sudo ./stackcount -f -D 1 --debug ktime_get
```
there is another type mismatch error:
```
Traceback (most recent call last):
  File "./stackcount", line 374, in <module>
    Tool().run()
  File "./stackcount", line 353, in run
    print("%s %d" % (";".join(line), v.value))
TypeError: sequence item 1: expected str instance, bytes found
```

martinvonwittich:
I'm having the same problem with `trace.py`, while using the Debian buster-backports packaged downloaded from https://packages.debian.org/buster-backports/bpfcc-tools:

```
host ~ # trace-bpfcc -v 'r:bash:readline "%s", retval'
[...]
Traceback (most recent call last):
  File "/usr/sbin/trace-bpfcc", line 886, in run
    self._attach_probes()
  File "/usr/sbin/trace-bpfcc", line 862, in _attach_probes
    probe.attach(self.bpf, self.args.verbose)
  File "/usr/sbin/trace-bpfcc", line 637, in attach
    self._attach_u(bpf)
[...]

PR #2514, opened 2019-09-11
stackcount: fix TypeError

mwesolowski:
fix for one of the problems mentioned in https://github.com/iovisor/bcc/issues/2512

### DIFF
--- a/tools/stackcount.py
+++ b/tools/stackcount.py
@@ -346,10 +346,10 @@ class Tool(object):
                     user_stack = list(user_stack)
                     kernel_stack = list(kernel_stack)
                     line = [k.name.decode('utf-8', 'replace')] + \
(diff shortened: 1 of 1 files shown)


palmtenor:
[buildbot, test this please]

palmtenor:
[buildbot, ok to test]

palmtenor:
@yonghong-song Is it OK for us to hard-code `utf-8` as the encoding?

yonghong-song:
@palmtenor I think it should be okay in this case as it is already used 'utf-8' encoding in the same file. Typically, we do not want to do encoding for parameters of bcc python library calls. In this case, it is for print and sometimes for python 3, byte array may print with prefix 'b' which may be undesirable in some cases.

PR #2515, opened 2019-09-12
add a new bcc tool iothrottle

ethercflow:
Trace the buffer io throttle's latency system-wide

### DIFF
--- a/libbpf-tools/.gitignore
+++ b/libbpf-tools/.gitignore
@@ -1,3 +1,4 @@
 /.output
-/runqslower
 /drsnoop
+/runqslower
+/xfsslower
--- a/libbpf-tools/Makefile
+++ b/libbpf-tools/Makefile
@@ -8,7 +8,7 @@ LIBBPF_OBJ := $(abspath $(OUTPUT)/libbpf.a)
 INCLUDES := -I$(OUTPUT)
 CFLAGS := -g -O2 -Wall
 
-APPS = drsnoop runqslower
+APPS = drsnoop runqslower xfsslower
 
 .PHONY: all
 all: $(APPS)
--- a/libbpf-tools/xfsslower.bpf.c
+++ b/libbpf-tools/xfsslower.bpf.c
@@ -1,0 +1,163 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2020 Wenbo Zhang
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include "xfsslower.h"
+
+#define BPF_F_INDEX_MASK		0xffffffffULL
+#define BPF_F_CURRENT_CPU		BPF_F_INDEX_MASK
+
+const volatile pid_t targ_tgid = 0;
+const volatile __u64 min_lat = 0;
+
+struct piddata {
+	u64 ts;
+	loff_t start;
+	loff_t end;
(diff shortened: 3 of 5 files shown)


tools/iothrottle.py
```diff
@@ -0,0 +1,162 @@
...
+b.attach_kretprobe(event="vfs_write", fn_name="trace_vfs_write_return")
+
+initial_ts = 0
+
+# header
+if args.timestamp:
+    print("%-14s" % ("TIME(s)"), end="")
+print("%-14s %-6s %-6s %-20s %-12s %-12s %-6s %-8s %-16s %12s %12s"
+      % ("COMM", "PID", "TID", "FNAME", "COUNT", "POS",
+         "BDI", "DIRTIED", "RATELIMIT(KBps)", "PAUSED(ms)", "LAT(ms)"))
```
yonghong-song: This is exceeding 80 columns. We should try hard to have <= 80 columns with only essential output and have an option (e..g, `-x`) to allow to dump more if needed. Could you take a look at this?

tools/iothrottle_example.txt
```diff
@@ -0,0 +1,60 @@
...
+^C
+
+While tracing, the processes write buffer IO, when there are too many dirty pages, balance_dirty_pages will be called
+to wake up bdi writeback thread and give up cpu by io_schedule_timeout(paused), which will increase the latency of
+vfs_write.
+
+iothrottle can be useful for discovering this situation and give some judge:
+- the buffer IO's performance jitter
+- the bdi's performance
+- whether we should make /proc/sys/vm/dirty_background_ratio lower
```
yonghong-song: Based on your above data, could you give more explanation on the above three listed items? For example, what data indicates buffer io performance jitter, what is the bdi's performance and in this particular case, should we make /proc/sys/vm/dirty_background_ratio lower or not and why.

ethercflow:
@yonghong-song Sorry to be late. I've updated this, PTAL.

yonghong-song:
@ethercflow No problem. Will take a look soon.

brendangregg:
Looks interesting; here's a few things I want you to think about and discuss, if you haven't thought about already:

A) Does this track all block throttle events?

B) I have a kprobe:blk_throtl_bio based tool in the upcoming book for showing block throttles by container. Why not use kprobe:blk_throtl_bio as well? I notice the tool instruments vfs_write, which is going to make me hesitate on production workloads due to the overhead - vfs_write is very frequent.

C) The output is similar to biosnoop. Why not add another option to biosnoop, like it's -Q option, for this block throttle time? I'd especially encourage this if vfs_write can be avoided (I don't want to add vfs_write instrumentation to biosnoop), such as by using other kernel functions.

These are the things I investigate about before I get anywhere near a PR.

brendangregg:
I should mention: I'm aware that these are likely different throttle events. blk_throtl_bio is for cgroup throttles, whereas this tool is doing writeback-based throttles. But that makes a stronger case for (A): if there are multiple ways a block I/O can be throttled, then a tool should show them all.

yonghong-song:
@ethercflow could you look into @brendangregg 's suggestions? One thing is whether we can fold this into `biosnoop` with additional flags and use b`ioisnoop` kprobe functions instead of `vfs_wriite` which is quite expensive since it is for all kinds of file writes?

@brendangregg could you give more details for `kprobe:blk_throtl_bio` so @ethercflow can do more study?

Issue #2510, opened 2019-09-09
Profiling tasks in cgroup containers


jxiong:
I'm planning to work on a project to profile jobs on computer nodes. Those jobs are executed inside containers, where the resources are constrained by cgroups. To make this work, there will be a job-profiler container running on computer host all the time, and whenever a new job is scheduled, which will cause new container to be created, a notification is sent to the job-profiler so that the resources used by job, including CPU, memory, block IO, and network will be profiled and reported.

I'm thinking about using BCC and ebpf to implement this, since it is well discussed in general. However, I figure out that BCC is not supporting cgroup well at this point. Kernel seems to already have cgroup support. Taking a look at the implementation of `perf_event_open()` where it accepts `cgroup_fd` from `pid` if flag `PERF_FLAG_PID_CGROUP` is set. It can only be set on a specific CPU(I don't know the reason behind this). Please let me know if I miss something here.

It doesn't seem to be a lot of work to add this support to BCC. I can work on this under some technical guidance.

palmtenor:
`PERF_FLAG_PID_CGROUP` works as intended at least for perf events from my experience.

Another way is to just use `bpf_get_current_cgroup_id` in your BPF program and classify the cost from different cgroups yourself. Probably easier this way if you run the profiler on the host, since you won't need to do BPF program attaching and processing for each container individually.

jxiong:
Thanks for replying. `perf_event` is the only API that can take extra flags for setting CGROUP from the implementation, specifically `bpf_attach_perf_event()`  function. I wonder how easy to use this to do whatever `PERF_KPROBE` does. Or do you have any example code to show me?

I thought about using `bpf_get_current_cgroup_id()` too. And I could define a map to keep track of cgroups that needs to profile; this map can be modified by job profiler from user space when containers come and go. However, I'm a little worrying about the atomic access to this map.

yonghong-song:
@jxiong I do not particularly worry about racing between user and kernel accessing to the map in the above method. kernel has proper mutex lock and rcu to protect write and read properly.

Another option is to extend bcc python API `attach_perf_event`:
```
    def attach_perf_event(self, ev_type=-1, ev_config=-1, fn_name=b"",
            sample_period=0, sample_freq=0, pid=-1, cpu=-1, group_fd=-1)
```

Maybe we can add an extra_flags which can have PERF_FLAG_PID_CGROUP.
[...]

PR #2516, opened 2019-09-15
Add cgroup support for attach_kprobe()

jxiong:
Close issue #2510

Test cases:
1. taskset -c 1,2 stress -c 2
2. mkdir /sys/fs/cgroup/perf_event/test && echo `pidof stress` > /sys/fs/cgroup/perf_event/test/tasks
3. `./examples/cpp/CPUDistribution 10 test` should only see activities on one CPU

### DIFF
--- a/src/cc/api/BPF.cc
+++ b/src/cc/api/BPF.cc
@@ -26,6 +26,10 @@
 #include <utility>
 #include <vector>
 
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+
 #include "bcc_exception.h"
 #include "bcc_elf.h"
 #include "bcc_syms.h"
@@ -164,11 +168,12 @@ StatusTuple BPF::detach_all() {
     return StatusTuple(0);
 }
 
-StatusTuple BPF::attach_kprobe(const std::string& kernel_func,
-                               const std::string& probe_func,
-                               uint64_t kernel_func_offset,
-                               bpf_probe_attach_type attach_type,
-                               int maxactive) {
+StatusTuple BPF::attach_kprobe_(const std::string& kernel_func,
+                                const std::string& probe_func,
+                                const std::string& cgroup_name,
+                                uint64_t kernel_func_offset,
+                                bpf_probe_attach_type attach_type,
+                                int maxactive) {
   std::string probe_event = get_kprobe_event(kernel_func, attach_type);
   if (kprobes_.find(probe_event) != kprobes_.end())
     return StatusTuple(-1, "kprobe %s already attached", probe_event.c_str());
@@ -176,10 +181,24 @@ StatusTuple BPF::attach_kprobe(const std::string& kernel_func,
   int probe_fd;
   TRY2(load_func(probe_func, BPF_PROG_TYPE_KPROBE, probe_fd));
 
+  int cgroup_fd = -1;
+  if (!cgroup_name.empty()) {
+    // Compose a full path of cgroup
+    std::string cgroup_path = "/sys/fs/cgroup/perf_event/" + cgroup_name;
+
+    cgroup_fd = open(cgroup_path.c_str(), O_RDONLY | O_DIRECTORY);
+    if (cgroup_fd < 0) {
+      TRY2(unload_func(probe_func));
+      return StatusTuple(-1, "Wrong cgroup path %s specified", cgroup_name.c_str());
+    }
+  }
+
   int res_fd = bpf_attach_kprobe(probe_fd, attach_type, probe_event.c_str(),
                                  kernel_func.c_str(), kernel_func_offset,
-                                 maxactive);
+                                 maxactive, cgroup_fd);
 
+  if (cgroup_fd >= 0)
+    close(cgroup_fd);
   if (res_fd < 0) {
     TRY2(unload_func(probe_func));
     return StatusTuple(-1, "Unable to attach %skprobe for %s using %s",
--- a/src/cc/api/BPF.h
+++ b/src/cc/api/BPF.h
@@ -64,8 +64,19 @@ class BPF {
   StatusTuple attach_kprobe(const std::string& kernel_func,
                             const std::string& probe_func,
                             uint64_t kernel_func_offset = 0,
-                            bpf_probe_attach_type = BPF_PROBE_ENTRY,
-                            int maxactive = 0);
+                            bpf_probe_attach_type type = BPF_PROBE_ENTRY,
+                            int maxactive = 0) {
+    return attach_kprobe_(kernel_func, probe_func, "", kernel_func_offset, type, maxactive);
+  }
+
+  StatusTuple attach_kprobe_cgroup(const std::string& kernel_func,
+                            const std::string& probe_func,
+                            const std::string& cgroup_name,
+                            uint64_t kernel_func_offset = 0,
+                            bpf_probe_attach_type type = BPF_PROBE_ENTRY,
+                            int maxactive = 0) {
+    return attach_kprobe_(kernel_func, probe_func, cgroup_name, kernel_func_offset, type, maxactive);
+  }
   StatusTuple detach_kprobe(
       const std::string& kernel_func,
       bpf_probe_attach_type attach_type = BPF_PROBE_ENTRY);
@@ -197,6 +208,12 @@ class BPF {
   int free_bcc_memory();
 
  private:
+  StatusTuple attach_kprobe_(const std::string& kernel_func,
+                             const std::string& probe_func,
+                             const std::string& cgroup_name,
+                             uint64_t kernel_func_offset = 0,
+                             bpf_probe_attach_type = BPF_PROBE_ENTRY,
+                             int maxactive = 0);
   std::string get_kprobe_event(const std::string& kernel_func,
                                bpf_probe_attach_type type);
   std::string get_uprobe_event(const std::string& binary_path, uint64_t offset,
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -757,12 +757,13 @@ static int bpf_get_retprobe_bit(const char *event_type)
  * the [k,u]probe. This function tries to create pfd with the perf_kprobe PMU.
  */
 static int bpf_try_perf_event_open_with_probe(const char *name, uint64_t offs,
-             int pid, const char *event_type, int is_return)
+             int pid, const char *event_type, int is_return, int cgroup_fd)
 {
   struct perf_event_attr attr = {};
   int type = bpf_find_probe_type(event_type);
   int is_return_bit = bpf_get_retprobe_bit(event_type);
   int cpu = 0;
+  int flags = PERF_FLAG_FD_CLOEXEC;
 
   if (type < 0 || is_return_bit < 0)
     return -1;
@@ -796,15 +797,18 @@ static int bpf_try_perf_event_open_with_probe(const char *name, uint64_t offs,
   /* config1 here is kprobe_func or  uprobe_path */
   attr.config1 = ptr_to_u64((void *)name);
   // PID filter is only possible for uprobe events.
-  if (pid < 0)
+  if (cgroup_fd >= 0) {
+    flags |= PERF_FLAG_PID_CGROUP;
+    pid = cgroup_fd;
+  } else if (pid < 0)
     pid = -1;
   // perf_event_open API doesn't allow both pid and cpu to be -1.
   // So only set it to -1 when PID is not -1.
   // Tracing events do not do CPU filtering in any cases.
-  if (pid != -1)
+  if (pid != -1 && !(flags & PERF_FLAG_PID_CGROUP))
     cpu = -1;
   return syscall(__NR_perf_event_open, &attr, pid, cpu, -1 /* group_fd */,
-                 PERF_FLAG_FD_CLOEXEC);
+		 flags);
 }
 
 // When a valid Perf Event FD provided through pfd, it will be used to enable
@@ -1005,7 +1009,7 @@ error:
 // see bpf_try_perf_event_open_with_probe().
 static int bpf_attach_probe(int progfd, enum bpf_probe_attach_type attach_type,
                             const char *ev_name, const char *config1, const char* event_type,
-                            uint64_t offset, pid_t pid, int maxactive)
+                            uint64_t offset, pid_t pid, int maxactive, int cgroup_fd)
 {
   int kfd, pfd = -1;
   char buf[PATH_MAX], fname[256];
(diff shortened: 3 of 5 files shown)


src/cc/libbpf.c
```diff
@@ -796,15 +797,18 @@ static int bpf_try_perf_event_open_with_probe(const char *name, uint64_t offs,
...
+  if (cgroup_fd >= 0) {
+    flags |= PERF_FLAG_PID_CGROUP;
+    pid = cgroup_fd;
+  } else if (pid < 0)
     pid = -1;
   // perf_event_open API doesn't allow both pid and cpu to be -1.
   // So only set it to -1 when PID is not -1.
   // Tracing events do not do CPU filtering in any cases.
-  if (pid != -1)
+  if (pid != -1 && !(flags & PERF_FLAG_PID_CGROUP))
```
palmtenor: Why `!(flags & PERF_FLAG_PID_CGROUP)`?

src/cc/api/BPF.cc
```diff
@@ -164,22 +168,37 @@ StatusTuple BPF::detach_all() {
...
   if (kprobes_.find(probe_event) != kprobes_.end())
     return StatusTuple(-1, "kprobe %s already attached", probe_event.c_str());
 
   int probe_fd;
   TRY2(load_func(probe_func, BPF_PROG_TYPE_KPROBE, probe_fd));
 
+  int cgroup_fd = -1;
+  if (!cgroup_name.empty()) {
+    // Compose a full path of cgroup
+    std::string cgroup_path = "/sys/fs/cgroup/perf_event/" + cgroup_name;
```
palmtenor: Let's not get into speculating where is user's `cgroupfs` mounted. Just pass in the user-opened FD. If you open the FDs here we'll also need to properly close them, which is complicated.

src/cc/libbpf.c
jxiong: > I feel we shouldn't change the general kprobe API for this specific case. There are many more controls in the `perf_event_open` and we probably couldn't add all of them to the kprobe API.

As long as we keep compatibility to existing code, I don't see there would be problems to extend kprobe API. Profiling performance in container wise is a very important feature for cluster.

jxiong: > Why `!(flags & PERF_FLAG_PID_CGROUP)`?

Kernel doesn't allow coexistence of `cpu` to be `-1` and this flag.

src/cc/api/BPF.cc
jxiong: That works for me too if this is what you want. My concern is that it would be awkward to get a file descriptor for some languages, but certainly all languages should be able to handle string well ;-)

I know `cgroupfs` could be mounted anywhere before, but almost of distribution have it mounted to `/sys/fs/cgroup`. If you're concerning about this, we can read and parse `/proc/mounts` to get the exact location of cgroup mount, which is probably better than passing file descriptor.

jxiong:
I'm still working on a kernel patch to make the whole thing work properly.

src/cc/libbpf.c
jxiong: Does anyone know why kernel doesn't allow to have cpu and pid setting to `-1` at the same time? Perf tools have to work this around by creating many per-cpu task for system wide tracing.
 
For kprobe, kernel doesn't seems to have implemented CPU filtering at this point, so bcc tools are taking this advantage. I wonder if kernel has got this `fixed`, would it break all existing tools? Please let me know if I understood it wrong.

src/cc/api/BPF.cc
yonghong-song: Considering sometimes cgroup could have some hierarchies like `/sys/fs/cgroup/system.slice/system-mysqld.slice/mysqld@3300.service`. You want to trace `mysqld@3300.service` which has a few processes in it. Just cgroup name `mysqld@3300.service` as it is possible to have the same name in other hierarchy.

Also, let us consider the python tools interface. The tools could accept a complete path like `/sys/fs/cgroup/system.slice/system-mysqld.slice/mysqld@3300.service` or a relative path `system.slice/system-mysqld.slice/mysqld@3300.service`. We should not bother tools writer with managing cgroup_fd. So bcc python library should open/close cgroup_fd as appropriate. 

So in my opinion, the C/C++ API should be cgroup_fd based and python API, which should be cgroup_path based. python library should handle open/close cgroup_fd and then using C++ binding. This way, we keep C/C++ library a little bit concise.

src/cc/libbpf.c
yonghong-song: Maybe some comments are better here.

yonghong-song:
> Does anyone know why kernel doesn't allow to have cpu and pid setting to -1 at the same time? Perf tools have to work this around by creating many per-cpu task for system wide tracing.

Not exactly sure what is the reaon. My guess is that if you do this (cpu = -1, pid = -1), kenrel needs to create per-cpu ring buffers with number of cpu's perf_event's. This may make kernel more complex. Just let user does this.

> For kprobe, kernel doesn't seems to have implemented CPU filtering at this point, so bcc tools are taking this advantage. I wonder if kernel has got this fixed, would it break all existing tools? Please let me know if I understood it wrong.

This is by design. bpf program is running before cpu filtering kicks in.
```
static int
kprobe_perf_func(struct trace_kprobe *tk, struct pt_regs *regs)
{       
        struct trace_event_call *call = trace_probe_event_call(&tk->tp);
        struct kprobe_trace_entry_head *entry;
        struct hlist_head *head;
        int size, __size, dsize;
        int rctx;
        
        if (bpf_prog_array_valid(call)) {
                unsigned long orig_ip = instruction_pointer(regs);
                int ret;
                
                ret = trace_call_bpf(call, regs);
                
                /*
                 * We need to check and see if we modified the pc of the
                 * pt_regs, and if so return 1 so that we don't do the
                 * single stepping.
                 */
                if (orig_ip != instruction_pointer(regs))
                        return 1;
                if (!ret)
                        return 0;
        }
        
        head = this_cpu_ptr(call->perf_events);
        if (hlist_empty(head))
                return 0;
```
If needed, bpf program itself can do filtering.

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

yonghong-song:
@jxiong Agree with you. tracing at cgroup level is very important. Thanks for working on this. Let us do C++ API first and then do python API/tools changes.

yonghong-song:
> I'm still working on a kernel patch to make the whole thing work properly.
Not sure what issue you try to address. But please cc me with your kernel patch!

jxiong:
> > Does anyone know why kernel doesn't allow to have cpu and pid setting to -1 at the same time? ..
> If needed, bpf program itself can do filtering.

This is exactly what I'm talking about. For the context of the problem, please check this out `https://github.com/iovisor/bcc/issues/2510#issuecomment-531586404`.

What I'm working on is to filter the data by `cgroup` inside `kprobe_perf_func()`. For know, probably it's a good idea to filter cgroup only because it would break existing BCC tools if I filter CPU too. Something like this:
```
diff --git a/kernel/trace/trace_kprobe.c b/kernel/trace/trace_kprobe.c
index 7d736248a070..673949b621bc 100644
--- a/kernel/trace/trace_kprobe.c
+++ b/kernel/trace/trace_kprobe.c
@@ -1155,11 +1155,20 @@ kprobe_perf_func(struct trace_kprobe *tk, struct pt_regs *regs)
        int size, __size, dsize;
        int rctx;

+       head = this_cpu_ptr(call->perf_events);
        if (bpf_prog_array_valid(call)) {
                unsigned long orig_ip = instruction_pointer(regs);
+               struct perf_event *event;
                int ret;

-               ret = trace_call_bpf(call, regs);
+               WARN_ONCE(hlist_empty(head), "head is empty\n");
+               hlist_for_each_entry_rcu(event, head, hlist_entry) {
+                       if (perf_event_match(event, regs)) {
+                               ret = trace_call_bpf(call, regs);
+                               if (!ret)
+                                       break;
+                       }
+               }

                /*
                 * We need to check and see if we modified the pc of the
```

src/cc/api/BPF.cc
jxiong: which C/C++ API is used by python? Is it `src/cc/api/BPF.[cc,h]` or `src/cc/libbpf.c`?

yonghong-song: See bcc/src/python/bcc/libbcc.py.

yonghong-song:
@jxiong the above change won't work. As I mentioned earlier, the bpf program gets to run regardless of whether a perf_event is opened on that cpu or not. This is by design, which gives bpf program some flexibility about the event. If necessary, bpf program can implement filtering by itself.

yonghong-song:
The following is another way to filter inside the bpf program based on cgroup id.
https://github.com/torvalds/linux/blob/master/tools/testing/selftests/bpf/get_cgroup_id_user.c
https://github.com/torvalds/linux/blob/master/tools/testing/selftests/bpf/cgroup_helpers.c#L239-L246
https://github.com/torvalds/linux/blob/master/tools/testing/selftests/bpf/progs/get_cgroup_id_kern.c

If similar to cpu's, kernel does not filter based on configured cgroup for bpf programs, bpf program itself doing filtering will become necessary.

src/cc/api/BPF.cc
```diff
@@ -164,22 +168,37 @@ StatusTuple BPF::detach_all() {
     return StatusTuple(0);
 }
 
-StatusTuple BPF::attach_kprobe(const std::string& kernel_func,
-                               const std::string& probe_func,
-                               uint64_t kernel_func_offset,
-                               bpf_probe_attach_type attach_type,
-                               int maxactive) {
+StatusTuple BPF::attach_kprobe_(const std::string& kernel_func,
```
palmtenor: We never name methods this way. Maybe `attach_kprobe_internal` and make it private

src/cc/api/BPF.cc
palmtenor: Agreed with @yonghong-song . Let's keep the C / C++ API take cgroup FD and we can figure out Python interface later.

src/cc/libbpf.c
```diff
@@ -796,15 +797,18 @@ static int bpf_try_perf_event_open_with_probe(const char *name, uint64_t offs,
   /* config1 here is kprobe_func or  uprobe_path */
   attr.config1 = ptr_to_u64((void *)name);
   // PID filter is only possible for uprobe events.
-  if (pid < 0)
+  if (cgroup_fd >= 0) {
```
palmtenor: I think we should also warn (or even error) if user passed in both PID and cgroup FD, which is possible in current interface.

src/cc/libbpf.c
palmtenor: I think a lot of these are just general parameter checks from the `perf_event_open` syscall and it would be hard to make exception for BPF events. I don't think CPU filtering matter with kprobe / uprobe / Kernel Tracepoints, so let's just keep it this way.

src/cc/api/BPF.h
```diff
@@ -197,6 +208,12 @@ class BPF {
   int free_bcc_memory();
 
  private:
+  StatusTuple attach_kprobe_(const std::string& kernel_func,
+                             const std::string& probe_func,
+                             const std::string& cgroup_name,
+                             uint64_t kernel_func_offset = 0,
```
palmtenor: The internal method doesn't need default argument

jxiong:
> The following is another way to filter inside the bpf program based on cgroup id.
...
> If similar to cpu's, kernel does not filter based on configured cgroup for bpf programs, bpf program itself doing filtering will become necessary.

Yeah it's obvious to set and match a single cgroup in BPF program. However, it would be difficult if we're handling the cases where a cgroup has descendants, and also the cgroup in profiling would spawn more descendants even after BPF program is installed.

I tend to think that doing kernel filtering is the best way of handling such an issue. Can you please reveal more details about why it's designed like this?

yonghong-song:
> However, it would be difficult if we're handling the cases where a cgroup has descendants, and also the cgroup in profiling would spawn more descendants even after BPF program is installed.

Yes, this is indeed the case. Have you tried whether perf handled this properly or not? That is, suppose we have
    cgroup1
        cgroup2
           process1
        cgroup3
           process2

If perf_event_open is for cgroup `cgroup1', only process1 and process2 is traced for bpf program?

> I tend to think that doing kernel filtering is the best way of handling such an issue. Can you please reveal more details about why it's designed like this?

I am only aware of cpu handling (i.e., bpf prog to see events for all cpus). But we did not do anything special related to cgroup. Hopefully perf cgroup handling just fine.

I will do some experiments as well later.

jxiong:
> Yes, this is indeed the case. Have you tried whether perf handled this properly or not? That is, suppose we have
> cgroup1
> cgroup2
> process1
> cgroup3
> process2
> 
> If perf_event_open is for cgroup `cgroup1', only process1 and process2 is traced for bpf program?

This is my expectation, but it's not working that way for now.

> I will do some experiments as well later.

You could verify how it works for `perf_event_open()` with `perf record ... -G`. We should make `perf event` for kprobe to work the same way. I don't know all the history about this though.

jxiong:
With little [changes](https://paste.ofcode.org/zhBcNAeEnWrqL8cYqFScjJ) made to `examples/cpp/LLCStat.cc`, I have verified that `perf_event_open()` works well with cgroup. It passed the test cases that i mentioned before, which only profiles the processes if they are within specified cgroup.

I think we should do the same thing for kprobe of perf event. We can do it in two ways: 1. add filtering as I said in `https://github.com/iovisor/bcc/pull/2516#issuecomment-531930948`, or 2. implement `pmu_disable` and `pmu_enable` for kprobe pmu, so that the BPF program will be run only when interesting processes are running.

The first way is simpler; the 2nd way is more in line with the rest of PMUs and can take advantage of existing framework. I tend to move forward with the 2nd way.

@yonghong-song comments?

yonghong-song:
@jxiong Right. From perf_event_open man page
```
       PERF_FLAG_PID_CGROUP (since Linux 2.6.39)
              This flag activates per-container system-wide monitoring.  A container is an abstraction that isolates a set of
              resources  for  finer-grained  control  (CPUs,  memory, etc.).  In this mode, the event is measured only if the
              thread running on the monitored CPU belongs to the designated container (cgroup).  The cgroup is identified  by
              passing  a  file descriptor opened on its directory in the cgroupfs filesystem.  For instance, if the cgroup to
              monitor is called test, then a file descriptor opened on /dev/cgroup/test  (assuming  cgroupfs  is  mounted  on
              /dev/cgroup)  must  be passed as the pid parameter.  cgroup monitoring is available only for system-wide events
              and may therefore require extra permissions.
```
the cgroup based filtering only available for system-wide events, which means it is not available for k/u probes and tracepoints.

I agree with you, it would be good if cgroup based filtering works for events created through perf_event_open(). We should at least attempt first. Please feel free to take a look at the issue first. Let me know if you need help or do not have cycles.

jxiong:
I will make a patch and make sure you will be in cc list. It should be a short patch.

jxiong:
As Alexei Starovoitov mentioned, `bpf_current_task_under_cgroup` should be able to filter tasks by cgroup. I'm closing this request and create a new request with sample code.

yonghong-song:
Thanks, @jxiong, do you think we could implement cgroup support in one of tools, e.g., trace.py or another tool you are frequently using and wish to have cgroup support?

PR #2520, opened 2019-09-20
Fixes #2518 -- weird behaviour of lookup_or_init

### LINKED ISSUE #2518: Weird side effect of map.lookup_or_init()
I discovered today that `map.lookup_or_init()` has bizarre behavior if the insert into the map fails (e.g. the map is full). It is either a documentation error, or an implementation error.

The actual behavior is that `lookup_or_init()` actually returns (with a return value of 0) from the function in which it is executed!! I would think that the more reasonable approach would be to make the `lookup_or_init` method itself return 0 if the insert fails....

Thoughts?

pjsg:
This changes the behaviour of `map.lookup_or_init` so that it returns 0 if the entry is not found *and* it cannot be inserted (e.g. the map is full). #2518 

Documentation updated. Also, all tests and sample code updated.

During testing, found two errors in test setup -- one was a mismatch between a function prototype and the libbcc python definition, and the other was an incorrect python path in the test wrapper. 

I can't actually get all the tests to pass on my platform -- but the same set of tests fail on master, so I'm not unduly worried. I didn't understand the test failures -- but they look like a failure to setup the test environment.

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -767,7 +767,7 @@ Examples in situ:
 
 Syntax: ```*val map.lookup_or_init(&key, &zero)```
 
-Lookup the key in the map, and return a pointer to its value if it exists, else initialize the key's value to the second argument. This is often used to initialize values to zero.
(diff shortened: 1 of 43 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

yonghong-song:
Thanks! The change looks good to me.

PR #2522, opened 2019-09-20
Add sampe code for cgoup test

jxiong:
How to run this test:
0. Make sure you have cgroup2 mounted:
  $ mount -t cgroup2
  cgroup on /sys/fs/cgroup/unified type cgroup2 (rw,nosuid,nodev,noexec,relatime)
1. Create a test cgroup in cgroup2 fs:
  $ sudo mkdir /sys/fs/cgroup/unified/test
2. Run this program in a bash:
  $ sudo CGroupTest /sys/fs/cgroup/unified/test
3. In another bash window, do the following:
  $ echo $$ > /sys/fs/cgroup/unified/test/cgroup.procs
  $ run some other commands
4. You should be able to see lots of prints like:
  'file ... was opened!'

### DIFF
--- a/examples/cpp/CGroupTest.cc
+++ b/examples/cpp/CGroupTest.cc
@@ -1,0 +1,85 @@
+/*
+ * CGroupTest Demonstrate how to use BPF cgroup API to collect file open event
+ *
+ * Basic example of cgroup and BPF kprobes.
+ *
+ * USAGE: CGroupTest cgroup2_path
+ *
+ * EXAMPLES:
+ * 1. Create a directory under cgroup2 mountpoint:
+ *    $ sudo mkdir /sys/fs/cgroup/unified/test
+ * 2. Add current bash into the testing cgroup:
+ *    $ sudo echo $$ | sudo tee -a /sys/fs/cgroup/unified/test/cgroup.procs
+ * 3. Open another bash window, and start CGroupTest as:
+ *    $ sudo ./examples/cpp/CGroupTest /sys/fs/cgroup/unified/test
+ * 4. Run file open activity from previous bash window should be printed.
+ *
+ * Copyright (c) Jinshan Xiong
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ */
+
+#include <unistd.h>
+#include <fstream>
+#include <cstdlib>
+#include <iomanip>
+#include <iostream>
+#include <string>
+
+#include "BPF.h"
+
+const std::string BPF_PROGRAM = R"(
+#include <linux/sched.h>
+#include <linux/path.h>
+#include <linux/dcache.h>
+
+BPF_CGROUP_ARRAY(cgroup, 1);
+
+int on_vfs_open(struct pt_regs *ctx, struct path *path) {
+  if (cgroup.check_current_task(0) > 0)
+    bpf_trace_printk("file '%s' was opened!\n", path->dentry->d_name.name);
+  return 0;
+}
+)";
+
+int main(int argc, char** argv) {
+  if (argc != 2) {
+    std::cerr << argv[0] << ": requires _one_ cgroup path" << std::endl;
+    return 1;
+  }
+
+  ebpf::BPF bpf;
+  auto init_res = bpf.init(BPF_PROGRAM);
+  if (init_res.code() != 0) {
+    std::cerr << init_res.msg() << std::endl;
+    return 1;
+  }
+
+  auto cgroup_array = bpf.get_cgroup_array("cgroup");
+  auto update_res = cgroup_array.update_value(0, argv[1]);
+  if (update_res.code() != 0) {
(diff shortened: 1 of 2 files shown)


yonghong-song:
[buildbot, test this please]

examples/cpp/CGroupTest.cc
```diff
@@ -0,0 +1,79 @@
...
+  while (true) {
+    if (std::getline(pipe, line)) {
+      std::cout << line << std::endl;
+    } else {
+      std::cout << "Waiting for a sys_clone event" << std::endl;
+      sleep(1);
+    }
+  }
+
+  auto detach_res = bpf.detach_kprobe("vfs_read");
```
yonghong-song: vfs_open?

examples/cpp/CGroupTest.cc
```diff
@@ -0,0 +1,79 @@
...
+  std::string line;
+
+  while (true) {
+    if (std::getline(pipe, line)) {
+      std::cout << line << std::endl;
+    } else {
+      std::cout << "Waiting for a sys_clone event" << std::endl;
+      sleep(1);
+    }
+  }
```
yonghong-song: Looks like we have no way to get out of the while loop. For an example, it is okay. If this is the intention, maybe you want to print out some thing like
```
std::cout << "Started tracing, hit Ctrl-C to terminate." << std::endl;
```
and you do not need anything after this loop.

examples/cpp/CGroupTest.cc
```diff
@@ -0,0 +1,79 @@
...
+
+const std::string BPF_PROGRAM = R"(
+#include <linux/sched.h>
+#include <linux/path.h>
+#include <linux/dcache.h>
+
+BPF_CGROUP_ARRAY(cgroup, 1);
+
+int on_vfs_open(struct pt_regs *ctx, struct path *path) {
+  if (cgroup.check_current_task(0))
```
yonghong-song: This condition is not enough. The kernel implementation is below.
```
BPF_CALL_2(bpf_current_task_under_cgroup, struct bpf_map *, map, u32, idx)
{
        struct bpf_array *array = container_of(map, struct bpf_array, map);
        struct cgroup *cgrp;

        if (unlikely(idx >= array->map.max_entries))
                return -E2BIG;

        cgrp = READ_ONCE(array->ptrs[idx]);
        if (unlikely(!cgrp))
                return -EAGAIN;

        return task_under_cgroup_hierarchy(current, cgrp);
}
```
Initially if there is no cgroup is installed in the cgroup array,
-EAGAIN will be returned. We should not print the opened files
as it will be confusing as cgroup is not really tested.

I suggest the above condition is checked as:
```
  __s64 ret = cgroup.check_current_task(0);
 if (ret == 1) { ...} // if (ret > 0) { ...} should also work
```

Since this is a cgroup test, I think the cgroup path parameter should be required unless we want create a cgroup for user and mimic the test if user did not provide the path.

Please also include instructions how to setup the test environment as the comments in this file, so user does not need to find the commit messages for trying.

examples/cpp/CGroupTest.cc
```diff
@@ -0,0 +1,79 @@
+/*
+ * CPUDistribution Show load distribution across CPU cores during a period of
+ *                 time. For Linux, uses BCC, eBPF. Embedded C.
```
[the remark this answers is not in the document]
jxiong: Done.

examples/cpp/CGroupTest.cc
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * Basic example of BCC and kprobes.
+ *
+ * USAGE: CPUDistribution [duration]
+ *
+ * Copyright (c) Facebook, Inc.
```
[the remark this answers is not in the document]
jxiong: Done

examples/cpp/CGroupTest.cc
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  std::ifstream pipe("/sys/kernel/debug/tracing/trace_pipe");
+  std::string line;
+
+  while (true) {
+    if (std::getline(pipe, line)) {
+      std::cout << line << std::endl;
+    } else {
+      std::cout << "Waiting for a sys_clone event" << std::endl;
```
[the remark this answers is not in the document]
jxiong: Done.

examples/cpp/CGroupTest.cc
jxiong: Yes, I know kernel will clean everything up after the process exits, but let's keep the code after this loop just for completeness.

examples/cpp/CGroupTest.cc
jxiong: Thank you so much for the detailed explanation, I will do that.

yonghong-song:
[buildbot, test this please]

PR #2523, opened 2019-09-23
FAQ: fix error message

jgehrcke:
Just a quick fix for a thing I noticed while reading the FAQ and trying things out...

The package is called bcc, and not bpf:

```
    from bcc import BPF
```

If that is not available then the error message thrown by the Python interpreter reads

```
    ModuleNotFoundError: No module named 'bcc'
```

and not `No module named 'bpf'`, as currently indicated by the FAQ.


palmtenor:
[buildbot, ok to test]

PR #2535, opened 2019-09-30
Link with libbpf package if available

olsajiri:
Adding check for libbpf rpm, which is now available on Fedora
and if it's available using it instead of the local libbpf
sources.

Besides that it's better to share common source of libbpf code,
it also prevents issues when having application that links to
libbpf and libbcc, where you could end up conflicting functions
and segfaults if those 2 libbpf libs are not on the same version.

Signed-off-by: Jiri Olsa <jolsa@kernel.org>


yonghong-song:
@olsajiri thanks for the information. I tried on my fc30 VM and indeed the packages are there.
```
[yhs@localhost ~]$ rpm -ql libbpf-5.2.17-200.fc30.x86_64
/usr/lib/.build-id
/usr/lib/.build-id/a2
/usr/lib/.build-id/a2/7563c0358fb32127cd9f72632d52c86137486e
/usr/lib64/libbpf.so.0
/usr/lib64/libbpf.so.0.0.3
/usr/share/licenses/libbpf
/usr/share/licenses/libbpf/COPYING
[yhs@localhost ~]$ rpm -ql libbpf-devel-5.2.17-200.fc30.x86_64
/usr/include/bpf/bpf.h
/usr/include/bpf/btf.h
/usr/include/bpf/libbpf.h
/usr/include/bpf/libbpf_util.h
/usr/include/bpf/xsk.h
/usr/lib64/libbpf.a
/usr/lib64/libbpf.so
/usr/lib64/pkgconfig/libbpf.pc
/usr/share/licenses/libbpf-devel
/usr/share/licenses/libbpf-devel/COPYING
[yhs@localhost ~]$
```

Currently, bcc also uses a few uapi headers libbpf/include/uapi/linux/{bpf.h, btf.h}. So we have to do some changes to use libbpf-devel package.... Thanks!

olsajiri:
> Currently, bcc also uses a few uapi headers libbpf/include/uapi/linux/{bpf.h, btf.h}. So we have to do some changes to use libbpf-devel package.... Thanks!

hum, I thought that'd be brought in by kernel headers, will check

yonghong-song:
> hum, I thought that'd be brought in by kernel headers, will check
typically we will have a compat version of uapi bpf.h and btf.h so that bcc build itself does not depend on system bpf.h/btf.h which sometimes are old.

olsajiri:
I see, so how about we include those in the the devel package as well, like:

```
/usr/include/bpf/
/usr/include/bpf/uapi
/usr/include/bpf/uapi/linux
/usr/include/bpf/uapi/linux/if_link.h
/usr/include/bpf/uapi/linux/btf.h
/usr/include/bpf/uapi/linux/bpf_common.h
/usr/include/bpf/uapi/linux/bpf.h
/usr/include/bpf/uapi/linux/netlink.h
/usr/include/bpf/uapi/linux/if_xdp.h
/usr/include/bpf/libbpf_util.h
/usr/include/bpf/btf.h
/usr/include/bpf/libbpf.h
/usr/include/bpf/bpf.h
/usr/include/bpf/xsk.h
```

yonghong-song:
@olsajiri Yes. This will help.

olsajiri:
@yonghong-song so there's one thing I'm strugling with now:

currently bcc links `libbpf.a` as a static object into `libbcc.so`,
so even libbpf's core api ends up in the `libbcc.so` dynamic table:

```
$ nm -D  libbcc.so | grep bpf_map_lookup_elem
0000000000b30c22 T bpf_map_lookup_elem
0000000000b30c92 T bpf_map_lookup_elem_flags
```

which conflicts with `libbpf.so` loaded by application,
and could lead to crash if those 2 are on different version

I can fix this by linking `libbpf.so` dynamicly to `libbcc.so`,
but then when building application that links with `libbcc.so`,
you need to add -lbpf as well.. is that acceptable? people
might get angry ;-)

other solution could be to create new library like `libbcc-shared.so`,
which would dynamically linking `libbpf.so`

my goal is to be able to use `libbpf.so` in `bpftrace` without
being worried about the crash ;-)

any ideas? thanks

palmtenor:
This seems indeed would be a problem for your use case.

We can compile BCC with `-fvisibility=hidden` so that symbols (including `libbpf` symbols statically linked in) will not be leaked out, and for the functions we actually want to expose for users to use, tag them with `__attribute__ ((visibility ("default")))` (maybe have a macro like `BCC_API_EXPORT`).

@yonghong-song This may also solve the `libelf` vs. `elfutil` problem when dynamically linking `libbcc` in debug build.

yonghong-song:
@palmtenor using `-fvisibility=hidden` and then selectively expose API functions is a great idea. Do you want to give a quick prototype on this?

palmtenor:
I surely can. But that means we will need to decide what APIs are considered "exported". Should `bcc_syms` ones be exported? Should `bcc_elf` / `bcc_proc` ones be exported?

olsajiri:
> We can compile BCC with `-fvisibility=hidden` so that symbols (including `libbpf` symbols statically linked in) will not be leaked out, and for the functions we actually want to expose for users to use, tag them with `__attribute__ ((visibility ("default")))` (maybe have a macro like `BCC_API_EXPORT`).

I don't like that libbpf will be loaded twice in this case,
however if the other solutions won't work for you, I'm ok with that ;-)

yonghong-song:
The way @palmtenor suggested might be the best solution considering bcc often needs latest bpf/btf uapi headers for features. Another solution is to create a bcc-no-liibbpf.so shared library, but that may have its own maintenance issues and the trunk bcc probably won't test old libbpf versions, ...

olsajiri:
this version seems to works for me when linked with bpftrace,
but it's probably missing some symbols from c++ space,
not sure how to take all of them in some sane way

palmtenor:
I think statically linking would still be a good option for BCC as we can control `libbpf` upgrades and API changes by control the git submodule updates, thus ensuring everything release of BCC would always work on its own regardless of different versions of `libbpf.so` brought in user's environment. Plus the linker script would be harder to maintain.

I also agree that leaking conflicting symbols is bad. I think symbol visibility is pretty standard approach of this issue, either you or I could do that. On the other hand, I know that if I'm calling `dlopen` myself, I can just set `RTLD_DEEPBIND`. I'm not sure if there is a way to make that default to a shared library?

olsajiri:
> I also agree that leaking conflicting symbols is bad. I think symbol visibility is pretty standard approach of this issue, either you or I could do that. On the other hand, I know that if I'm calling `dlopen` myself, I can just set `RTLD_DEEPBIND`. I'm not sure if there is a way to make that default to a shared library?

hum, AFAIU even with `RTLD_DEEPBIND` behaviour we won't prevent duplicated symbols, right? application that links to libcc should still see duplicated libbpf symbols..

I think the best solution would be to have the version script together with new `libcc_shared.so` library, that links libbpf dynamically.. but I'd be happy to go with just one of them ;-)

palmtenor:
My main concern is that if the system pulling in a `libbpf.so` that is different from what BCC is built with, things may break in strange ways.

In fact, this is exactly what the environment I'm currently working with and I believe many other people. It doesn't break today, but may very well break in the future, especially considering `libbpf.so` doesn't have versioned functions like `libc`. So I really enjoyed statically linked BCC...

I will try something for this during the weekend:)

olsajiri:
> In fact, this is exactly what the environment I'm currently working with and I believe many other people. It doesn't break today, but may very well break in the future, especially considering `libbpf.so` doesn't have versioned functions like `libc`. So I really enjoyed statically linked BCC...

hum, libbpf.so does have versionned functions

how about the extra libbcc_shared.so library options, would be minimal change.. we could also link it with specific libbpf version to be on the safe side.. people (like me) would force their applications to
use the shared version if it's detected

> I will try something for this during the weekend:)

thanks

palmtenor:
> how about the extra libbcc_shared.so library options, would be minimal change.. we could also link it with specific libbpf version to be on the safe side.. people (like me) would force their applications to
use the shared version if it's detected

I guess that's OK. But I guess we won't be able to make sure BCC is compatible with all `libbpf.so` versions, other than the one of the BCC git submodule. Would that be OK?

yonghong-song:
@palmtenor what is your opinion? Have you done any investigation for the approach to not expose libbpf symbols? If that approach does not work out, I think the patch provided by @olsajiri is an acceptable approach, which is controlled by a cmake option during build.

palmtenor:
Sorry, I was distracted by some personal errands. Yes, I think the separated library is good and much better than the initial approach

palmtenor:
[buildbot, ok to test]

cmake/FindLibBpf.cmake
```diff
@@ -0,0 +1,59 @@
...
+  NAMES
+    src/bpf.c
+    src/bpf.h
+    src/libbpf.c
+    src/libbpf.h
+
+  PATHS
+    ${libbpf_source_path}
+    ENV CPATH
+)
```
yonghong-song: Which package will have bpf.c and libbpf.c?
From bcc perspective, we do not really need libbpf .c files,
the necessary header files for compilation and linking to libbpf.so should be good enough.

olsajiri: when you set `CMAKE_USE_LIBBPF_PACKAGE`, you are building from libbpf rpm,
bpf.c and libbpf.c come from libbpf source package, check this hunk:

```
-file(GLOB libbpf_sources "libbpf/src/*.c")
+if(LIBBPF_FOUND)
+  # We need to include the libbpf packaged source to get the
+  # headers location, becaseu they are installed under 'bpf'
+  # directory, but the libbpf code references them localy
+
+  include_directories(${LIBBPF_INCLUDE_DIR}/bpf)
+
+  file(GLOB libbpf_sources "${LIBBPF_SOURCE_DIR}/src/*.c")
+  set(libbpf_uapi ${LIBBPF_SOURCE_DIR}/include/uapi/linux)
+else()
+  file(GLOB libbpf_sources "libbpf/src/*.c")
+  set(libbpf_uapi libbpf/include/uapi/linux})
+endif()

```

olsajiri: it's libbpf-debugsource

yonghong-song:
```
[yhs@localhost ~]$ sudo yum install libbpf-debugsource
Last metadata expiration check: 0:02:01 ago on Wed 30 Oct 2019 08:00:30 PM PDT.
No match for argument: libbpf-debugsource
Error: Unable to find a match: libbpf-debugsource
[yhs@localhost ~]$ uname -a
Linux localhost.localdomain 5.2.16-200.fc30.x86_64 #1 SMP Thu Sep 19 16:14:04 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
[yhs@localhost ~]$
```
Maybe I miss some setting?

olsajiri: looks like you need to enable the updates repo, like:
`dnf install --enablerepo=updates-debuginfo libbpf-debugsource`

yonghong-song: Thanks. I can get the source now.

You are missing to add C++ API to bcc-shared.so library.

Also, the name bcc-shared.so does not sound great. It is hard to distinguish libbcc.so and libbcc-shared.so just by name. How about bcc-no-libbpf.so? a little bit long but captures what it is.

Could you have a test case when CMAKE_USE_LIBBPF_PACKAGE is used? The test case should exercise C++ API and should compile successfully on a host with libbpf-debugsource and other libbpf packages installed.

yonghong-song: Please also add some comments in the code to describe what packages are needed to
find all needed headers and sources. For example, people may not be aware of this one `dnf install --enablerepo=updates-debuginfo libbpf-debugsource`.

olsajiri: > Thanks. I can get the source now.
> 
> You are missing to add C++ API to bcc-shared.so library.

ah ok, will check

> Also, the name bcc-shared.so does not sound great. It is hard to distinguish libbcc.so and libbcc-shared.so just by name. How about bcc-no-libbpf.so? a little bit long but captures what it is.

ok

> Could you have a test case when CMAKE_USE_LIBBPF_PACKAGE is used? The test case should exercise C++ API and should compile successfully on a host with libbpf-debugsource and other libbpf packages installed.

ok, thanks for review

olsajiri: @yonghong-song I'm adding the test but I'm getting following error
when running test_libbcc:

```
 : CommandLine Error: Option 'no-x86-call-frame-opt' registered more than once!
LLVM ERROR: inconsistency in registered CommandLine options
```

just making sure it's an issue in my config  as google search suggest..

thanks

yonghong-song: Never seen this error before. This is related compilation options. Most likely it is a user error.

olsajiri: @yonghong-song  I've managed to fix it by linking the libbcc.so dynamically,
which I think should be the way in any case..

I recently upgraded to new fedora that uses llvm-9, what llvm version are you compiling with?

olsajiri: same thing happens for llvm 8 ... I pushed new code with fix for the test,
now both tests give same result

yonghong-song:
tested and the change looks good.

cmake/FindLibBpf.cmake
```diff
@@ -0,0 +1,68 @@
...
+#if (LIBBPF_LIBRARIES AND LIBBPF_INCLUDE_DIR AND LIBBPF_SOURCE_DIR)
+#  set (LibBpf_FIND_QUIETLY TRUE)
+#endif (LIBBPF_LIBRARIES AND LIBBPF_INCLUDE_DIR AND LIBBPF_SOURCE_DIR)
+
+# You'll need following packages to be installed (Fedora names):
+# libbpf
+# libbpf-debugsource
+# libbpf-devel
+#
+# Please note that ou might need to enable updates-debuginfo repo
```
yonghong-song: ou => you

tests/cc/test_c_api.cc
```diff
@@ -234,14 +234,14 @@ TEST_CASE("resolve symbol addresses for a given PID", "[c_api]") {
   }
 
   SECTION("resolve in libbcc.so") {
```
yonghong-song: maybe change here libbcc.so => LIBBCC_NAME?

PR #2549, opened 2019-10-11
Fix Erroneous "gethostlatency" Comments

synical:
Was taking a look around earlier and noticed there was a copied comment from tcpaccept.py in gethostlatency.py. Figured would PR a fixed comment.


palmtenor:
[buildbot, ok to test]

PR #2560, opened 2019-10-19
assign offset properly for btf DataSec

yonghong-song:
The following test case failed:
```
  -bash-4.4$ cat test.py
  #!/usr/bin/python
  from bcc import BPF
  prog="""
  BPF_ARRAY(array1, u64, 1);
  BPF_ARRAY(array2, u64, 1);
  """
  b=BPF(text=prog, debug=0x20)
  -bash-4.4$ sudo ./test.py
  libbpf: Error loading BTF: Invalid argument(22)
  libbpf: magic: 0xeb9f
  version: 1
  flags: 0x0
  hdr_len: 24
  ...
  [57] DATASEC array size=176 vlen=2
           type_id=47 offset=0 size=88
           type_id=49 offset=0 size=88 Invalid offset

  Loading .BTF section failed
  -bash-4.4$
```
The reason is the data offset in DataSec VarInfo
(e.g., type_id=47 offset=0 size=88) always have
offset=0 since JIT won't do a final relocation
for the record. Assign offset properly
in bcc_btf before loading btf to the kernel.
With this patch, test case can load btf properly.
```
  -bash-4.4$ cat test.py
  #!/usr/bin/python
  from bcc import BPF
  prog="""
  BPF_ARRAY(array1, u64, 1);
  BPF_ARRAY(array2, u64, 1);
  """
  b=BPF(text=prog, debug=0x20)
  -bash-4.4$ sudo ./test.py
  -bash-4.4$
```

### DIFF
--- a/src/cc/bcc_btf.cc
+++ b/src/cc/bcc_btf.cc
@@ -114,7 +114,7 @@ void BTF::fixup_datasec(uint8_t *type_sec, uintptr_t type_sec_size,
     case BTF_KIND_VAR:
       next_type += sizeof(struct btf_var);
       break;
-    case BTF_KIND_DATASEC:
+    case BTF_KIND_DATASEC: {
       secname = strings + t->name_off;
       if (sections_.find(secname) != sections_.end()) {
         t->size = std::get<1>(sections_[secname]);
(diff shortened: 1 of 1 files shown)


yonghong-song:
Currently, it is difficult to test btf as it may be ignored if loading failure. In the future, we may reuse the debug mode 0x20 to enforce btf loading, i.e., if it is used, failing btf will fail the program load, so we can proper test it. This will be another patch.

PR #2568, opened 2019-10-23
cpudist: create sufficiently large hash table to avoid missing tasks

### LINKED ISSUE #2567: cpudist stop working when there are too many fork
Running `cpudist -P 1` when a service forks a lot (such as the update-mandb cron) results in incorrect results where only a handfull of process' distributions are taken into account after the forks dies.

This can be reproduced by running this command: `for i in $(seq 10000); do echo fork | dd of=/dev/null & done`. After the command completes, cpudist prints only a few distribution and miss most the load happening after.

It seems like it is related to the default BPF_HASH size and it can be fixed by the proposed PR below, but perhaps there is a better way to avoid that issue?

TristanCacqueray:
This change fixes the cpudist tool to avoid issue when too many tasks
are running.

Fixes #2567 -- cpudist stop working when there are too many fork

### DIFF
--- a/tools/cpudist.py
+++ b/tools/cpudist.py
@@ -64,7 +64,7 @@ typedef struct pid_key {
 } pid_key_t;
 
 
-BPF_HASH(start, u32, u64);
+BPF_HASH(start, u32, u64, MAX_PID);
 STORAGE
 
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

brendangregg:
thanks

PR #2574, opened 2019-10-27
tools/tcpretrans: add tracepoint support

boat0:
If tracepoint tcp:tcp_retransmit_skb exists (kernel version >=4.15)
we use tracepoint instead of kprobe for efficiency.

Co-authored-by: Runlong Lin <linrunlong@bytedance.com>

### DIFF
--- a/tools/tcpretrans.py
+++ b/tools/tcpretrans.py
@@ -91,7 +91,9 @@ struct ipv6_flow_key_t {
     u16 dport;
 };
 BPF_HASH(ipv6_count, struct ipv6_flow_key_t);
+"""
 
+bpf_text_kprobe = """
 static int trace_event(struct pt_regs *ctx, struct sock *skp, int type)
 {
     if (skp == NULL)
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

tools/tcpretrans.py
```diff
@@ -187,11 +252,19 @@
     bpf_text = bpf_text.replace("IPV6_INIT", struct_init['ipv6']['count'])
     bpf_text = bpf_text.replace("IPV4_CORE", count_core_base.replace("COUNT_STRUCT", 'ipv4_count'))
     bpf_text = bpf_text.replace("IPV6_CORE", count_core_base.replace("COUNT_STRUCT", 'ipv6_count'))
+    if BPF.tracepoint_exists("tcp", "tcp_retransmit_skb"):
+        bpf_text_tracepoint = bpf_text_tracepoint.replace("IPV4_CODE", struct_init_tracepoint['ipv4']['count'])
+        bpf_text_tracepoint = bpf_text_tracepoint.replace("IPV6_CODE", struct_init_tracepoint['ipv6']['count'])
+        bpf_text += bpf_text_tracepoint
```
yonghong-song: Looks good in general. Could you do a little bit refactoring not to load `trace_retransmit` function if tracepoint exists? If would be confusing if both `trace_retransmit` and the tracepoint program are loaded.

boat0: Okay. Either kprobe or tracepoint will be loaded, but not both, given '-l' is not there.

yonghong-song:
[buildbot, test this please]

yonghong-song:
The change looks good. Maybe we can just call `BPF.tracepoint_exists("tcp", "tcp_retransmit_skb")` once and remember the answer instead of calling multiple times. But this is minor.

PR #2585, opened 2019-11-04
Support attaching uprobe to offset

mmisono:
This enables to attach uprobe to offset (similar to iovisor/bpftrace#803.)
`attach_kprobe()` already supported it with`event_off`.

This does not check an offset, e.g., whether the offset is the middle of the instruction or not. Currently `attach_kprobe()` also does not perform any check.
It is nice if we can integrate bpftrace's offset check (what above PR does) with bcc, but I think it will be separate PR.

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -1106,9 +1106,14 @@ Examples in situ:
 
 ### 4. attach_uprobe()
 
-Syntax: ```BPF.attach_uprobe(name="location", sym="symbol", fn_name="name")```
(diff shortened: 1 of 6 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
The change looks good. Could you also uprobe offset support to C++ API? This will make uprobe offset support in all user APIs. thanks!

mmisono:
I missed that. I added it and also updated the document.
Could you take a look? @yonghong-song

yonghong-song:
[buildbot, test this please]

yonghong-song:
LGTM. Thanks!

Issue #2588, opened 2019-11-06
[BUG] runqslower reports the same PID to different processes


bodgergely:
Please note that the same PID shows up many times with totally unrelated processes when using **runqslower** which is especially obvious when the latency level is set to low.
I find it very easy to reproduce consistently.
It almost "feels like" the prev PID remains the same, not sure if there is a race condition of some sort.

Kernel: 5.0.0-25-generic

This is something that I observed on other kernels as well

Excerpt (truncated) from running `sudo ./runqslower.py 100`:
Of course many other PIDs suffer from this.

TIME     COMM             PID           LAT(us)
**22:19:12 gnome-software   2435              178**
22:19:12 update-notifier  10581             191
**22:19:12 gsd-media-keys   2435              158**
**22:19:12 gsd-wacom        2435              139**
**22:19:12 ibus-x11         2435              146**
22:19:12 runqslower.py    2613              105
22:19:12 kworker/0:3      2615             1736
22:19:12 kworker/0:3      2615             1357
**22:19:12 swapper/1        2435              105**
22:19:12 chrome           22498             122

Running ps:

greg      **2435**  1.2  0.4 534412 77012 tty2     Sl+  Oct31 114:53 /usr/lib/xorg/Xorg vt2 -displayfd 3 -auth /run/user/1000/gdm/Xauthority -background none -noreset -keeptty -verbose 3

Meaning the real process with pid 2435 is Xorg

bodgergely:
Can anyone else confirm they see this anomaly?
I am happy to help debugging/investigating this issue if others confirm that this is an actual bug.

yonghong-song:
Yes, it does look like some bugs in raw_tracepoint based approach. The kprobe based solution looks more sane. Could you help take a look?

bodgergely:
I created a PR above that fixes this issue. Let me know if it looks good or (bad).

PR #2596, opened 2019-11-14
tools/runqslower.py: Fix raw tracepoint code

bodgergely:
Closes: #2588
It is was observed that the same PID can show up many
times to many unrelated processes. The exact cause of
this is unknown but highly likely (speculation) a race condition.

Removing the default (if supported) raw tracepoint option along
with the raw tracepoint code.

### DIFF
--- a/tools/runqslower.py
+++ b/tools/runqslower.py
@@ -28,6 +28,10 @@
 # Licensed under the Apache License, Version 2.0 (the "License")
 #
 # 02-May-2018   Ivan Babrou   Created this.
+# 18-Nov-2019   Gergely Bod   BUG fix: Use bpf_probe_read_str() to extract the
+#                               process name from 'task_struct* next' in raw tp code.
+#                               bpf_get_current_comm() operates on the current task
+#                               which might already be different than 'next'.
 
 from __future__ import print_function
 from bcc import BPF
@@ -163,13 +167,12 @@ RAW_TRACEPOINT_PROBE(sched_switch)
     // TP_PROTO(bool preempt, struct task_struct *prev, struct task_struct *next)
     struct task_struct *prev = (struct task_struct *)ctx->args[1];
     struct task_struct *next= (struct task_struct *)ctx->args[2];
-    u32 pid, tgid;
+    u32 pid;
     long state;
 
     // ivcsw: treat like an enqueue event and store timestamp
     bpf_probe_read(&state, sizeof(long), (const void *)&prev->state);
     if (state == TASK_RUNNING) {
-        bpf_probe_read(&tgid, sizeof(prev->tgid), &prev->tgid);
         bpf_probe_read(&pid, sizeof(prev->pid), &prev->pid);
         if (!(FILTER_PID || pid == 0)) {
             u64 ts = bpf_ktime_get_ns();
@@ -177,7 +180,6 @@ RAW_TRACEPOINT_PROBE(sched_switch)
         }
     }
 
-    bpf_probe_read(&tgid, sizeof(next->tgid), &next->tgid);
     bpf_probe_read(&pid, sizeof(next->pid), &next->pid);
 
     u64 *tsp, delta_us;
@@ -195,7 +197,7 @@ RAW_TRACEPOINT_PROBE(sched_switch)
     struct data_t data = {};
     data.pid = pid;
     data.delta_us = delta_us;
-    bpf_get_current_comm(&data.task, sizeof(data.task));
+    bpf_probe_read_str(&data.task, sizeof(data.task), next->comm);
 
     // output
     events.perf_submit(ctx, &data, sizeof(data));


bodgergely:
I tested and using the kprobe code instead of the raw tracepoint fixes the issue.

yonghong-song:
I would like to know what is the exact reason for raw tracepoint having issues. If you do not have time, I will take a look some time next week.

bodgergely:
@yonghong-song Actually I got quite interested as well, so I will take a look now.

bodgergely:
It is definitely a race condition. I was able to reduce the amount of times the race happens by moving code around etc (the bpf_probe_reads mostly) but I get a sense we have a deeper dependency/design issue here (either the linux kernel raw tracepoint mechanism or the bcc itself). Still needs investigating whether it is possible to guarantee race elimination at all with the raw tp method.

bodgergely:
I think the  `bpf_get_current_comm(&data.task, sizeof(data.task));` is the culprit here. It is possible that is is operating on a more recently scheduled task than the one we captured by `next`. That explains why we see another task's name associated with the pid. 
I used 
```
    bpf_trace_printk("prev_pid: %u,prev_comm: %s\\n", prev->pid, prev->comm);
    bpf_trace_printk("next_pid: %u,next_comm: %s\\n", next->pid, next->comm);

```
where I can simply log out the current task's name without using `bpf_get_current_comm` and the process name-pid pairs are correct when looking into `trace_pipe`
We should somehow extract the name of the task atomically along with the pid.

bodgergely:
Using the
```
bpf_probe_read_str(&data.task, sizeof(data.task), next->comm);
```
instead of 
```
bpf_get_current_comm(&data.task, sizeof(data.task));
```
eliminates most of the erroneous results.
I updated the PR.

yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks! The change makes sense. The `pid` we got is from `next` pid. The `comm` should be the one from `next`, not from current.

yonghong-song:
[buildbot, test this please]

bodgergely:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

bodgergely:
Should I squash the 2 commits into a single commit?

tools/runqslower.py
```diff
@@ -161,25 +161,23 @@
...
     // TP_PROTO(bool preempt, struct task_struct *prev, struct task_struct *next)
-    struct task_struct *prev = (struct task_struct *)ctx->args[1];
-    struct task_struct *next= (struct task_struct *)ctx->args[2];
-    u32 pid, tgid;
+    u32 pid;
     long state;
+    struct task_struct *next = (struct task_struct *)ctx->args[2];
+    struct task_struct *prev = (struct task_struct *)ctx->args[1];
+
+    bpf_probe_read(&pid, sizeof(next->pid), &next->pid);
```
yonghong-song: This does not sound right. The pid could be overwritten by later at line 174. I suggest keeping code as is, just removing the dead code at line 172 to bpf_probe_read value for tgid.

yonghong-song:
You need to fix a problem I just pointed out. Normally I can squash the commits when I apply. But since you need to change codes any way, please go ahead to squash. Thanks!

tools/runqslower.py
bodgergely: Understood, you are correct about line 174 overwriting the pid. I will fix that. Might as well move the line 169 after the block:
```
    if (state == TASK_RUNNING) {
        bpf_probe_read(&pid, sizeof(prev->pid), &prev->pid);
        if (!(FILTER_PID || pid == 0)) {
            u64 ts = bpf_ktime_get_ns();
            start.update(&pid, &ts);
        }
    }

```
> just removing the dead code at line 172 to bpf_probe_read value for tgid.
Line 172:
```
bpf_probe_read(&state, sizeof(long), (const void *)&prev->state);
```
The above does not seem dead code to me. If I remove line 172, the `state` variable will stay undefined. I might missing something here.

Also the `tgid` is not used at all. So why introduce is back?

yonghong-song: I mean tgid code. The 'state' bpf_probe_read certainly useful. The line number sometimes refers to old, sometimes refer to new code, so maybe confusing.

yonghong-song: I think just remove dead code and use bpf_probe_read_str() change should be good enough.

bodgergely:
Did the fix as we agreed.

bodgergely:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

yonghong-song:
Merged this patch. The change looks good. But current runqslower.py seems filtering in bpf program based on task id instead of user visible pid. 

Suppose pid is the task id and tgid is the actual process id,
the following code
```
bpf_text = bpf_text.replace('FILTER_PID', 'pid != %s' % args.pid)
```
probably should be
```
bpf_text = bpf_text.replace('FILTER_PID', 'tgid != %s' % args.pid)
```

Did you run into any issues with this? We probably should fix this issue as well.
The whole bpf program 'pid' thing can be then removed, `tgid` will be used then.

bodgergely:
Thanks, I created an Issue https://github.com/iovisor/bcc/issues/2607
regarding your comment to look into it later.

PR #2603, opened 2019-11-18
Make StatusTuple code and msg methods const

eklitzke:
I ran into this while trying to write a method that takes a const StatusTuple reference, these methods should be marked const. I also fixed an unnecessary string copy in msg.


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

PR #2608, opened 2019-11-19
add tool: migrations.py: watch tasks migrate across CPUs

bodgergely:
Tool to snoop on processes migrating across CPUs.

Prints a row for each task migration.
Columns:
    * Time
    * Process name
    * PID
    * Prev CPU
    * Curr CPU

### DIFF
--- a/tools/migrations.py
+++ b/tools/migrations.py
@@ -1,0 +1,222 @@
+#!/usr/bin/python
+# @lint-avoid-python-3-compatibility-imports
+#
+# migrations    Trace as processes migrate between CPUs.
+#               For Linux, uses BCC, eBPF.
+#
+# This script traces as tasks migrate between CPUs
+# by taking note of the cpu the current process was
+# scheduled previously.
+#
+# USAGE: migrations [-p PID]
+#
+# REQUIRES: Linux 4.9+ (BPF_PROG_TYPE_PERF_EVENT support).
+# 
+# This observes processes migrating from one CPU to another.
+# It is also possible to observe only a specific PID or
+# by default all threads, except the swapper thread.
+# 
+# Copyright 2019 Cloudflare, Inc.
+# Licensed under the Apache License, Version 2.0 (the "License")
+#
+# 27-Oct-2019   Gergely Bod   Created this.
+
+from __future__ import print_function
+from bcc import BPF
+import argparse
+from time import strftime
+
+# arguments
+examples = """examples:
+    ./migrations          # trace all process migrations system wide
+    ./migrations -p 123   # trace pid 123 only
+"""
+parser = argparse.ArgumentParser(
+    description="Trace process cpu migrations.",
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    epilog=examples)
+parser.add_argument("-p", "--pid", type=int, metavar="PID", dest="pid",
+    help="trace this PID only")
+parser.add_argument("--ebpf", action="store_true",
+    help=argparse.SUPPRESS)
+args = parser.parse_args()
+debug = 0
+
+# define BPF program
+bpf_text = """
+#include <uapi/linux/ptrace.h>
+#include <linux/sched.h>
+#include <linux/nsproxy.h>
+#include <linux/pid_namespace.h>
+
+BPF_HASH(start, u32, u32);
+
+struct rq;
+
+struct data_t {
+    u32 pid;
+    char task[TASK_COMM_LEN];
+    u32 prev_cpu;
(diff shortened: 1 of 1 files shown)


bodgergely:
Let me know if you deem this tool useful to be included in the tools.
I found it useful in scenarios when I want to watch a certain process/thread migration behaviour.

bodgergely:
Also, in addition, watching NUMA node migrations could prove to be useful too. Monitoring if a task gets migrated to a CPU that associated with a different NUMA node than what is was scheduled before.

brendangregg:
```
# migrations.py 
Tracing CPU migrations.
TIME     COMM             PID          PREV_CPU       CURR_CPU
19:46:46 Thread-35        4752           10             29
```

The output is untidy. The last two columns are wider than necessary (why?) and the CPU is not aligned. When I develop tools, I make sure the output is perfect -- that's the easiest part to get right. Getting the logic of what to probe and how is the harder part. (And if you can't get the easy part right, I worry about the harder part.)

Here's a typical prod server for us:

```
# perf stat -e migrations,sched:sched_migrate_task -I 1000
#           time             counts unit events
     1.001148025             38,847      migrations                                                  
     1.001148025             39,528      sched:sched_migrate_task                                    
     2.001837920             49,548      migrations                                                  
     2.001837920             49,880      sched:sched_migrate_task                                    
     3.002563468             45,854      migrations                
```

It's doing about 50,000 migrations per second.

migrations.py prints only 2,000 lines of output per second. Why the large discrepancy?

It's not due to same-CPU migrations being unfiltered above; filtering them:

```
# bpftrace -e 't:sched:sched_migrate_task /args->orig_cpu != args->dest_cpu/ { @ = count(); } 
    interval:s:1 { print(@); clear(@); }'
Attaching 2 probes...
@: 38360

@: 47557

@: 43065
```

We can also already just use `perf record` to get a dump of migrations and CPU details:

```
# perf record -e sched:sched_migrate_task
^C[ perf record: Woken up 1 times to write data ]
[ perf record: Captured and wrote 1.551 MB perf.data (10542 samples) ]
# perf script
      Thread-139  5031 [002]  5461.297541: sched:sched_migrate_task: comm=ForkJoinPool.co pid=5920 prio=120 orig_cpu=19 dest_cpu=4
      Thread-139  5031 [002]  5461.297551: sched:sched_migrate_task: comm=ForkJoinPool.co pid=5906 prio=120 orig_cpu=12 dest_cpu=15
  HystrixTimer-4  6254 [007]  5461.297588: sched:sched_migrate_task: comm=nf.dependency.P pid=6935 prio=120 orig_cpu=45 dest_cpu=15
[...]
```

I can add filters to perf record to avoid the same-CPU migrations if need be.

But the bigger problem: whether the tool is printing 2,000 or 50,000 lines of output per second, what am I supposed to do with it? The key value of BPF is that we can summarize data in kernel. Just this one-liner would be more practical in production:

```
# bpftrace -e 't:sched:sched_migrate_task { @[pid] = count(); }'
Attaching 1 probe...
^C

@[1512]: 1
@[238]: 1
@[40]: 1
@[1684]: 1
@[12392]: 1
@[93]: 1
@[286]: 1
@[1746]: 1
@[76]: 2
@[2270]: 2
@[118]: 3
@[136]: 3
@[2271]: 3
@[64]: 4
@[52]: 4
@[1458]: 7
@[2919]: 17
@[4594]: 105
@[0]: 4838
@[4144]: 72458
```

We can create a migrations analysis tool for bcc, but it needs to go beyond our current toolset.

bodgergely:
@brendangregg First of all thanks for the review! Very much appreciated.
Thanks for showing me those one-liners, and you are right this tool in this form is not useful enough to be included in the tool set. 

> I make sure the output is perfect -- that's the easiest part to get right. Getting the logic of what to probe and how is the harder part. (And if you can't get the easy part right, I worry about the harder part.)

I am still learning the ropes of bcc development, and the conventional presentation format, so excuse me for not getting it right. However I might still be capable to get the 'hard problems' right (this tool is obviously way too simple). 

Unrelated:

I actually have another idea for another scheduling behaviour analysis tool that is related to the **scheduler's load balancer**. Do we have some tool that reports on the cpu load balancer behaviour? I am mostly interested in scenarios when the utilization of the cpus are not perfect due to unbalanced cores.
I will try to think about it in the upcoming days and come up with something.

Thanks for reviewing again.

brendangregg:
The basic CPU tools to check balance include:

```
mpstat -P ALL
runqlen -C
```

there are other more advanced ones, including the `perf sched` subcommand.

PR #2612, opened 2019-11-25
Add map-in-map support

yonghong-song:
Add BPF_MAP_TYPE_HASH_OF_MAPS and BPF_MAP_TYPE_HASH_OF_MAPS
supports in bcc. Two new constructs below are introduced
to bpf program:
```
  BPF_HASH_OF_MAPS(map_name, "inner_map_name", max_entries)
  BPF_ARRAY_OF_MAPS(map_name, "inner_map_name", max_entries)
```
In the above, "inner_map_name" is for metadata purpose and there
must be a map defined in bpf program with map name "inner_map_name".

Both python and C++ APIs are added.

For python, a new Table API get_fd() is introduced to get
the fd of a map so that the fd can be used by a map-in-map
do update. The get_fd() is already exposed as API function
in C++. For C++, without get_fd(), we will need to
templatize basic functions like update_value etc, which
I feed too heavy weight. Because of C++ using get_fd()
mechanism, so I exposed similar API on python side
for parity reason.

For map-in-map, the inner map lookup/update/delete won't have
explicit map names. Considering map-in-map is not
used very frequently, I feel looking primitive
bpf_map_{lookup,update,delete}_elem() probably okay,
so I did not create any new bcc specific
constructs for this purpose.

Added both C++ and python test cases to show how to
use the above two new map type in bcc.

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -759,7 +759,33 @@ Methods (covered later): map.redirect_map().
 Examples in situ:
 [search /examples](https://github.com/iovisor/bcc/search?q=BPF_CPUMAP+path%3Aexamples&type=Code),
 
-### 12. map.lookup()
+### 12. BPF_ARRAY_OF_MAPS
+
+Syntax: ```BPF_ARRAY_OF_MAPS(name, inner_map_name, size)```
+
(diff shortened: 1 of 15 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

PR #2613, opened 2019-11-26
Do not traverse attributes during JSON generation (Fixes #2605)

### LINKED ISSUE #2605: Structs that use __attribute__((aligned(sizeof(...)))) cause invalid JSON to be generated
The following code leads to `json.decoder.JSONDecodeError` in `BPF.get_table()`:
```
from bcc import BPF

b = BPF(text="""
struct S {
        int x;
} __attribute__((aligned(sizeof(long))));

BPF_HISTOGRAM(typetest, struct S);
""")

typetest = b.get_table("typetest")
print(typetest.Key._fields_)
```

The resulting JSON looks like this:
`["S", [["x","int"], ["__pad_end","char",[4]] ], "struct_packed"]"long"`

It seems that any struct that uses `sizeof()` in `__attribute__((aligned(...)))` causes the size to get inserted into the JSON. (For searchability: If the attribute is on the outermost struct, the error message will be `Extra data`, otherwise it'll be `Expecting ',' delimiter`)

alexer:
Previously, `__attribute__((aligned(sizeof(...))))` would cause the type inside `sizeof()` to be appended after the JSON type information for the struct in some clang versions, generating invalid JSON. Inhibit attribute traversal, since we only want to visit the fields.

I'm not that familiar with C++ nor the internals of clang and bcc, so review with extreme prejudice. :)

I've tested that this fixes the issue for me, but I'm not sure if this is the best way to fix this. Since we *only* want to look at the fields, it might be better to just inhibit *all* child traversal for `RecordDecl`, since we manually iterate the fields anyway.

I am not sure what would be the best way to do that, though, since I have no idea what's going on with clang's AST traversal: Given that `TraverseRecordDecl` already seems to try to skipping the children, it's kind of weird that the `AlignedAttr` is visited in the first place - maybe some parent class invoked by `WalkUpFromRecordDecl`, does the attribute visiting? Returning `false` from `VisitRecordDecl` also seems to fix this, but I have no idea what kind of other side effects that might have. (And if returning `false` is the right thing to do, does that mean that most of the other methods in the file should really be returning `false`, too?)

The above roughly translates to "I have no idea what's *really* happening, and the more I look at it, the weirder it seems to get".

### DIFF
--- a/src/cc/json_map_decl_visitor.cc
+++ b/src/cc/json_map_decl_visitor.cc
@@ -43,6 +43,7 @@ class BMapDeclVisitor : public clang::RecursiveASTVisitor<BMapDeclVisitor> {
   bool VisitPointerType(const clang::PointerType *T);
   bool VisitEnumConstantDecl(clang::EnumConstantDecl *D);
   bool VisitEnumDecl(clang::EnumDecl *D);
+  bool VisitAttr(clang::Attr *A);
 
  private:
   bool shouldSkipPadding(const RecordDecl *D);
@@ -183,6 +184,7 @@ bool BMapDeclVisitor::VisitBuiltinType(const BuiltinType *T) {
   result_ += "\"";
   return true;
 }
+bool BMapDeclVisitor::VisitAttr(clang::Attr *A) { return false; }
 
 class JsonMapTypesVisitor : public virtual MapTypesVisitor {
  public:


yonghong-song:
[buildbot, test this please]

yonghong-song:
This works fine. It fixed llvm 8.0 issue. For llvm 10.0, visitAttr() is not called for the test case so it does not have any effect.

PR #2620, opened 2019-11-29
Use libbpf-static instead of libbpf-debugsource for CMAKE_USE_LIBBPF_PACKAGE

olsajiri:
@yonghong-song I thought we could use source packaged within
package build, but it's not the case.. it seems we can use libbpf-static
instead just as fine, with the uapi exception (described below)

I made sure this time and prepared [bcc fedora build](https://koji.fedoraproject.org/koji/taskinfo?taskID=39395362) with this change. Works nicely on all supported archs.

thanks,
jirka

---
Currently when building with CMAKE_USE_LIBBPF_PACKAGE we assume
that 'source' package is available to build libbcc_bpf libraries.

However It turned on that using debugsource (in Fedora) package is
not suitable or even possible, so we are switching to use the
libbpf-static package instead.

Adding libbpf-static package detection and removing source
package detection.

Using ar to extract objects from archive and linking them
with both libbcc_bpf.so and libbcc_bpf.a.

Using always uapi headers from the latest libbpf Github repo,
because there are not packaged, but they are backwards compatible,
so it's no problem to have older libbpf package with newer
uapi headers.

Signed-off-by: Jiri Olsa <jolsa@kernel.org>

### DIFF
--- a/cmake/FindLibBpf.cmake
+++ b/cmake/FindLibBpf.cmake
@@ -3,21 +3,17 @@
 #
 #  LIBBPF_FOUND            - system has libbpf
 #  LIBBPF_INCLUDE_DIR      - the libbpf include directory
-#  LIBBPF_SOURCE_DIR       - the libbpf source directory
+#  LIBBPF_STATIC_LIBRARIES - the libbpf source directory
 #  LIBBPF_LIBRARIES        - link these to use libbpf
 
-#if (LIBBPF_LIBRARIES AND LIBBPF_INCLUDE_DIR AND LIBBPF_SOURCE_DIR)
+#if (LIBBPF_LIBRARIES AND LIBBPF_INCLUDE_DIR AND LIBBPF_STATIC_LIBRARIES)
 #  set (LibBpf_FIND_QUIETLY TRUE)
-#endif (LIBBPF_LIBRARIES AND LIBBPF_INCLUDE_DIR AND LIBBPF_SOURCE_DIR)
+#endif (LIBBPF_LIBRARIES AND LIBBPF_INCLUDE_DIR AND LIBBPF_STATIC_LIBRARIES)
 
 # You'll need following packages to be installed (Fedora names):
 # libbpf
-# libbpf-debugsource
+# libbpf-static
 # libbpf-devel
-#
-# Please note that you might need to enable updates-debuginfo repo
-# for debugsource package like:
-#   dnf install --enablerepo=updates-debuginfo libbpf-debugsource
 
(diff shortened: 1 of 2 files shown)


cmake/FindLibBpf.cmake
```diff
@@ -34,17 +34,16 @@ find_path (LIBBPF_INCLUDE_DIR
 
 file(GLOB libbpf_source_path /usr/src/debug/libbpf-*)
```
yonghong-song: do we still need the above?

cmake/FindLibBpf.cmake
```diff
@@ -3,12 +3,12 @@
...
 #  LIBBPF_INCLUDE_DIR      - the libbpf include directory
-#  LIBBPF_SOURCE_DIR       - the libbpf source directory
+#  LIBBPF_STATIC_LIBRARIES - the libbpf source directory
 #  LIBBPF_LIBRARIES        - link these to use libbpf
 
-#if (LIBBPF_LIBRARIES AND LIBBPF_INCLUDE_DIR AND LIBBPF_SOURCE_DIR)
+#if (LIBBPF_LIBRARIES AND LIBBPF_INCLUDE_DIR AND LIBBPF_STATIC_LIBRARIES)
 #  set (LibBpf_FIND_QUIETLY TRUE)
-#endif (LIBBPF_LIBRARIES AND LIBBPF_INCLUDE_DIR AND LIBBPF_SOURCE_DIR)
+#endif (LIBBPF_LIBRARIES AND LIBBPF_INCLUDE_DIR AND LIBBPF_STATIC_LIBRARIES)
```
yonghong-song: Looks like it should work. Will give a try later.
Could you remove the libbpf-debugsource comments below as well?

cmake/FindLibBpf.cmake
olsajiri: nope, removed

cmake/FindLibBpf.cmake
olsajiri: done, thanks

cmake/FindLibBpf.cmake
```diff
@@ -3,21 +3,17 @@
...
 
 # You'll need following packages to be installed (Fedora names):
 # libbpf
 # libbpf-debugsource
```
yonghong-song: The above `libbpf-debugsource` can be removed.

olsajiri: missed that one, sry, fixed

yonghong-song:
[buildbot, test this please]

PR #2621, opened 2019-11-30
Support a BPF program larger than BPF_MAXINSNS

mmisono:
A privileged user can load bpf program whose size is at most
BPF_COMPLEXITY_LIMIT_INSNS (which is larger than BPF_MAXINSNS) in 
Linux 5.3 (https://github.com/torvalds/linux/commit/c04c0d2b968ac45d6ef020316808ef6c82325a82).
Currently, however, `bcc_prog_load_xattr()` return error if  a bpf
program size is larger than BPF_MAXINSNS before loading the program. To
cope with this, change the function so that first trying to load a bpf
program and print an error message if bpf_program_load_xattr() fails
with -E2BIG.  This strategy is the same as libbpf's [`load_program`](https://github.com/libbpf/libbpf/blob/9ef191ea7dca6815e75dab341f2f7a346d979c66/src/libbpf.c#L3792).
The Error message should use BPF_COMPLEXITY_LIMIT_INSNS when
appropriate, but it is not defined in uapi headers yet.

Signed-off-by: Masanori Misono <m.misono760@gmail.com>

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -512,13 +512,6 @@ int bcc_prog_load_xattr(struct bpf_load_program_attr *attr, int prog_len,
   char prog_name[BPF_OBJ_NAME_LEN] = {};
 
   unsigned insns_cnt = prog_len / sizeof(struct bpf_insn);
-  if (insns_cnt > BPF_MAXINSNS) {
-    errno = EINVAL;
-    fprintf(stderr,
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

pchaigno:
This is going to be particularly useful for `execsnoop --max-args`! Thanks @mmisono!

PR #2622, opened 2019-12-01
Use integer types for enums in JSON (Fixes #2615)

### LINKED ISSUE #2615: Enums seem to have been broken on the Python-side since the support was added
Using structs that have enums in them generates JSON that `BPF._decode_table_type()` can't handle:
```
from bcc import BPF

b = BPF(text="""
struct S {
        enum {FOO} x;
};

BPF_HISTOGRAM(enumtest, struct S);
""")

typetest = b.get_table("enumtest")
```

The resulting JSON is `["S", [["x",["E", ["FOO"], "enum"]] ], "struct_packed"]` and the error is:
```
  File "bcc_enum.py", line 11, in <module>
    typetest = b.get_table("enumtest")
  File "/usr/lib64/python3.6/site-packages/bcc/__init__.py", line 485, in get_table
    keytype = BPF._decode_table_type(json.loads(key_desc))
  File "/usr/lib64/python3.6/site-packages/bcc/__init__.py", line 439, in _decode_table_type
    fields.append((t[0], BPF._decode_table_type(t[1])))
  File "/usr/lib64/python3.6/site-packages/bcc/__init__.py", line 454, in _decode_table_type
    raise Exception("Failed to decode type %s" % str(t))
Exception: Failed to decode type FOO
```

Support for enums in the generated JSON was added in commit f7f873a2 "Fix segfault with enumerations", but the added test doesn't actually reference the table from the Python side, so actually parsing the generated JSON in Python seems to have never been tested. Adding `t = b['drops']` to the test causes it to fail.

The `desc` for the enum is `['E', ['FOO'], 'enum']` and `BPF._decode_table_type()` iterates over `desc[1]`, assuming that each element is [a list] of length 2 or 3. When the element is a longer string instead, it fails.

I'm not sure if `libbcc.bpf_table_key_desc()` is [meant to be] used internally only, or if it might be used by users of bcc too, so I don't know if this should be fixed on the C++ or Python side. (Ie. can we just change the format of the generated JSON "no questions asked")

If we can just change the JSON, I'd propose doing that: Currently, the list of enumeration members is nearly useless anyway, since they don't include the associated values - I'd either remove the members completely, or add the associated values (though I expect that enums will just be represented as `ctypes.c_uint` or similar, so the members would be ignored in the Python code anyway).

There's also the question of size: Enum size is implementation-defined, so it depends on the compiler (some compilers automatically select the smallest type where all the values fit), the compiler flags (`-fshort-enums`), enum attributes (`packed`) and the range of values.

The simplest thing to do seems to be to change the generated JSON from `["fieldname", ["ENUM_MEMBER", ...], "enum"]` to `["fieldname", "unsigned int"]` (or similar), by replacing the code in `VisitEnumDecl()` with `TraverseType(D->getIntegerType()); return false;`. (That fixes the issue for me)

Speaking of `Traverse*()`, it seems that the root cause of the segfault that was referenced in the commit is calling `TraverseDecl()` from `VisitTagType()` (infinite recursion through `BMapDeclVisitor::VisitTagType -> RecursiveASTVisitor::TraverseDecl -> TraverseEnumDecl -> TraverseType -> TraverseEnumType -> WalkUpFromEnumType -> WalkUpFromTagType -> BMapDeclVisitor::VisitTagType`) - the [documentation in AST/RecursiveASTVisitor.h](https://github.com/llvm-mirror/clang/blob/ea521aa6024431138f7890d486a6e3500822b3d5/include/clang/AST/RecursiveASTVisitor.h#L111) seems to say that `Visit*()` may not call `Traverse*()` (nor `WalkUpFrom*()`), so since we're doing that in many places, it might be opening up a can of worms.

I'm not that familiar with clang, so I don't know what would be the best way to avoid the whole problem - maybe the intention is that you'd dispatch to a separate class when you want to call `Traverse*()` from `Visit*()`. You could also get the same effect by tracking a "traversal mode" and only doing things in `Visit*()` if you're in the right mode, though that'd quite likely just end up being messier, since it'd essentially be just a trick to "avoid new classes at any cost".

As for the enum size thing, both gcc and clang seem to use `unsigned int` by default (`int` if there's negative values), but since it's compiler dependent - and we almost certainly aren't using the same compiler that the kernel was compiled with (since we're using clang, whereas the kernel is usually built with gcc) - I think if the compiler used for the kernel were to use a different-sized enum (that wasn't due to `__attribute__((packed))`), we'd end up getting the size wrong. (Or does bcc extract the actual sizes from the kernel, and somehow feed them to clang?)

**In short:**
- There exists an easy fix (using `TraverseType()`)
- Calling `Traverse*()` from `Visit*()` seems to be bad
- In some (probably highly unlikely) configurations, we'll probably get enum sizes wrong

If you want the easy fix, I can make a pull request right away - otherwise, figuring out what is the proper fix will probably need more discussion (or for someone who can see the proper fix without discussion to fix it :P)

alexer:
The JSON generated for enums wasn't being handled, and caused
exceptions. Since the enum values aren't needed (and would've been
useless without the associated numerical values, anyway), just use the
matching integer type instead.

---

`Traverse*()` shouldn't really be called from inside `Visit*()`, but all the other methods are doing it too, so... The original enum code seems to have been a workaround for a problem caused by that, so the whole thing should probably be fixed at some point.

### DIFF
--- a/src/cc/json_map_decl_visitor.cc
+++ b/src/cc/json_map_decl_visitor.cc
@@ -41,7 +41,6 @@ class BMapDeclVisitor : public clang::RecursiveASTVisitor<BMapDeclVisitor> {
   bool VisitTypedefType(const clang::TypedefType *T);
   bool VisitTagType(const clang::TagType *T);
   bool VisitPointerType(const clang::PointerType *T);
-  bool VisitEnumConstantDecl(clang::EnumConstantDecl *D);
   bool VisitEnumDecl(clang::EnumDecl *D);
   bool VisitAttr(clang::Attr *A);
 
@@ -93,22 +92,8 @@ bool BMapDeclVisitor::VisitFieldDecl(FieldDecl *D) {
   return true;
 }
 
-bool BMapDeclVisitor::VisitEnumConstantDecl(EnumConstantDecl *D) {
-  result_ += "\"";
-  result_ += D->getName();
-  result_ += "\",";
-  return false;
-}
-
 bool BMapDeclVisitor::VisitEnumDecl(EnumDecl *D) {
-  result_ += "[\"";
-  result_ += D->getName();
-  result_ += "\", [";
-  for (auto it = D->enumerator_begin(); it != D->enumerator_end(); ++it) {
-    TraverseDecl(*it);
-  }
-  result_.erase(result_.end() - 1);
-  result_ += "], \"enum\"]";
+  TraverseType(D->getIntegerType());
   return false;
 }
 
--- a/tests/python/test_clang.py
+++ b/tests/python/test_clang.py
@@ -944,6 +944,7 @@ struct a {
 BPF_HASH(drops, struct a);
         """
         b = BPF(text=text)
+        t = b['drops']
 
     def test_int128_types(self):
         text = """


yonghong-song:
[buildbot, test this please]

yonghong-song:
> Traverse*() shouldn't really be called from inside Visit*(), but all the other methods are doing it too, so... The original enum code seems to have been a workaround for a problem caused by that, so the whole thing should probably be fixed at some point.

The current mechanism works because we already visit all members for a record declaration, regardless of return value for visiting each field. This might be structured differently though.

PR #2624, opened 2019-12-02
Improved mallocstack

devidasjadhav:
1. to print information about size.
2. using perf_submit now.
3. validating segmentation fault due to missing check before stack walk.


yonghong-song:
So, just check stack id before stack_traces.walk() should fix the problem, right? I would like to avoid using perf_submit(). One advantage of bpf is to do in-kernel accumulation before sending the result to user space. This example is such a use case. Could you just add stack_id checking if it fixed the issue?

devidasjadhav:
@yonghong-song 
this was long pending pull request.
I did exactly as you suggested.

yonghong-song:
[buildbot, test this please]

yonghong-song:
looks good. Thanks!

PR #2632, opened 2019-12-04
man: fix a reference to runqlat in runqslower man page

jeromemarchand:
The runqslower man page obviously derived from the man page of
runqlat. It seems that a reference to runqlat in the NAME section has
been overlooked.


yonghong-song:
[buildbot, test this please]

Issue #2607, opened 2019-11-18
Current runqslower.py seems filtering in bpf program based on task id instead of user visible pid.


bodgergely:
Merged this patch. The change looks good. But current runqslower.py seems filtering in bpf program based on task id instead of user visible pid. 

Suppose pid is the task id and tgid is the actual process id,
the following code
```
bpf_text = bpf_text.replace('FILTER_PID', 'pid != %s' % args.pid)
```
probably should be
```
bpf_text = bpf_text.replace('FILTER_PID', 'tgid != %s' % args.pid)
```

Did you run into any issues with this? We probably should fix this issue as well.
The whole bpf program 'pid' thing can be then removed, `tgid` will be used then.

_Originally posted by @yonghong-song in https://github.com/iovisor/bcc/pull/2596#issuecomment-555230505_

bodgergely:
@yonghong-song But the user can trace any task(process, thread) with the current method. He/she just needs to run pstree or or any other process tool to find the task he/she wants to trace. If we limit it to `tgid` then how will we be able to trace threads for example? Or I might misuderstand something in your proposal.

yonghong-song:
It is totally good idea to trace with any task. What I mention is currently in .py program, we use pid but actually it means tid during filtering. If user want to filter a potential pid and he did so, he may still get events from that pid for other tid's. 

The best is to specify both pid and tid in the tool (making them exclusive to each other) which is similar to many other tools.

Could you draft a pull request for this? We need to 1. add '-t' flag to the tool, 2. properly implementing pid and tid based filtering in the bpf code.

bodgergely:
> Could you draft a pull request for this? We need to 1. add '-t' flag to the tool, 2. properly implementing pid and tid based filtering in the bpf code.

Yep, I will draft a PR for this.

bodgergely:
@yonghong-song
I started working on it but I want to clarify a few things:

1. If user specifies --pid: Means we are filtering based on `tgid`, meaning we are tracing all the tasks that belong to that `tgid`.
2. If user specied --tid: Means we are filtering on `pid` (thread) level and just tracing only that one single task.
3. If user specifies both --tid and --pid: We quit early and show an error message and the examples.

yonghong-song:
@bodgergely yes, the above described change sounds good to me.

bodgergely:
@yonghong-song I just created a PR for this issue. Let me know what you think.

bodgergely:
[...]

PR #2635, opened 2019-12-05
tools/runqslower.py: Option for --pid or --tid

bodgergely:
Closes: #2607

Option for --pid or --tid tracing.
Enable group leader level tracing.
* --pid <pid>:
    will trace every task that has <pid> as group leader.
    Meaning it will trace the parent and its direct children.
* --tid <tid>:
    will trace the given thread

Update example file.

### DIFF
--- a/tools/runqslower.py
+++ b/tools/runqslower.py
@@ -42,7 +42,7 @@ from time import strftime
 examples = """examples:
     ./runqslower         # trace run queue latency higher than 10000 us (default)
     ./runqslower 1000    # trace run queue latency higher than 1000 us
-    ./runqslower -p 123  # trace pid 123 and its children threads
+    ./runqslower -p 123  # trace pid 123
     ./runqslower -t 123  # trace tid 123 (use for threads only)
 """
 parser = argparse.ArgumentParser(
@@ -50,7 +50,7 @@ parser = argparse.ArgumentParser(
     formatter_class=argparse.RawDescriptionHelpFormatter,
     epilog=examples)
 parser.add_argument("-p", "--pid", type=int, metavar="PID", dest="pid",
-    help="trace this PID and its children")
+    help="trace this PID")
 parser.add_argument("-t", "--tid", type=int, metavar="TID", dest="tid",
     help="trace this TID only")
 parser.add_argument("min_us", nargs="?", default='10000',
--- a/tools/runqslower.py
+++ b/tools/runqslower.py
@@ -218,10 +218,6 @@ RAW_TRACEPOINT_PROBE(sched_switch)
 }
 """
 
-
-if args.pid and args.tid:
-    parser.error("specify only one of -p and -t")
-
 is_support_raw_tp = BPF.support_raw_tracepoint()
 if is_support_raw_tp:
     bpf_text += bpf_text_raw_tp
--- a/tools/runqslower.py
+++ b/tools/runqslower.py
@@ -49,15 +49,18 @@ parser = argparse.ArgumentParser(
     description="Trace high run queue latency",
     formatter_class=argparse.RawDescriptionHelpFormatter,
     epilog=examples)
-parser.add_argument("-p", "--pid", type=int, metavar="PID", dest="pid",
-    help="trace this PID")
-parser.add_argument("-t", "--tid", type=int, metavar="TID", dest="tid",
-    help="trace this TID only")
 parser.add_argument("min_us", nargs="?", default='10000',
     help="minimum run queue latecy to trace, in ms (default 10000)")
 parser.add_argument("--ebpf", action="store_true",
     help=argparse.SUPPRESS)
+
+thread_group = parser.add_mutually_exclusive_group()
+thread_group.add_argument("-p", "--pid", metavar="PID", dest="pid",
+    help="trace this PID only", type=int)
+thread_group.add_argument("-t", "--tid", metavar="TID", dest="tid",
+    help="trace this TID only", type=int)
 args = parser.parse_args()
+
 min_us = int(args.min_us)
 debug = 0
 
@@ -215,6 +218,10 @@ RAW_TRACEPOINT_PROBE(sched_switch)
 }
 """
 
+
+if args.pid and args.tid:
+    parser.error("specify only one of -p and -t")
+
 is_support_raw_tp = BPF.support_raw_tracepoint()
 if is_support_raw_tp:
     bpf_text += bpf_text_raw_tp
@@ -226,11 +233,6 @@ if min_us == 0:
     bpf_text = bpf_text.replace('FILTER_US', '0')
 else:
     bpf_text = bpf_text.replace('FILTER_US', 'delta_us <= %s' % str(min_us))
-
-if args.tid and args.pid:
-    print("Can not specify both pid and tid!")
-    print(examples)
-    exit(-1)
 
 if args.tid:
     bpf_text = bpf_text.replace('FILTER_PID', 'pid != %s' % args.tid)
--- a/tools/runqslower.py
+++ b/tools/runqslower.py
@@ -7,7 +7,7 @@
 # This script traces high scheduling delays between tasks being
 # ready to run and them running on CPU after that.
 #
-# USAGE: runqslower [-p PID] [min_us]
+# USAGE: runqslower [-p PID] [-t TID] [min_us]
 #
 # REQUIRES: Linux 4.9+ (BPF_PROG_TYPE_PERF_EVENT support).
 #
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

tools/runqslower.py
```diff
@@ -42,14 +42,17 @@
...
+    ./runqslower -t 123  # trace tid 123 (use for threads only)
 """
 parser = argparse.ArgumentParser(
     description="Trace high run queue latency",
     formatter_class=argparse.RawDescriptionHelpFormatter,
     epilog=examples)
 parser.add_argument("-p", "--pid", type=int, metavar="PID", dest="pid",
-    help="trace this PID only")
+    help="trace this PID and its children")
+parser.add_argument("-t", "--tid", type=int, metavar="TID", dest="tid",
```
yonghong-song: Could you use `parser.add_mutually_exclusive_group()` such that user can only specify `-p` or `-t`, but not both?

tools/runqslower.py
```diff
@@ -80,7 +83,7 @@
 // record enqueue timestamp
 static int trace_enqueue(u32 tgid, u32 pid)
```
yonghong-song: could you change all tgid => pid and pid => tid in both python and C code? This will make code cleaner, less confusion and easy to follow.

tools/runqslower_example.txt
```diff
@@ -41,9 +41,11 @@ positional arguments:
 
 optional arguments:
   -h, --help         show this help message and exit
-  -p PID, --pid PID  trace this PID only
+  -p PID, --pid PID  trace this PID and its children threads
+  -t TID, --tid TID  trace this TID only
```
yonghong-song: since you add/change options, man page also need change.

tools/runqslower.py
```diff
@@ -218,10 +239,22 @@
     bpf_text = bpf_text.replace('FILTER_US', '0')
 else:
     bpf_text = bpf_text.replace('FILTER_US', 'delta_us <= %s' % str(min_us))
-if args.pid:
-    bpf_text = bpf_text.replace('FILTER_PID', 'pid != %s' % args.pid)
+
+if args.tid and args.pid:
+    print("Can not specify both pid and tid!")
+    print(examples)
+    exit(-1)
```
yonghong-song: If you use python mutual exclusive group, you do not need this.

tools/runqslower.py
bodgergely: I was also confused when I was writing the code and switched back and forth the naming of the pids/tids/tgids etc. (Because of the kernel  function interfaces) Finally the `tgid` made is the most obvious as the 'g' signalled that it is a group leader, meaning a pid :) But I can understand the argument for keeping pid and tid.

bodgergely:
For now I just made changes based on your comments except the 
> could you change all tgid => pid and pid => tid in both python and C code? This will make code  cleaner, less confusion and easy to follow.

tools/runqslower.py
```diff
@@ -109,9 +113,12 @@
     if (prev->state == TASK_RUNNING) {
         tgid = prev->tgid;
         pid = prev->pid;
-        if (!(FILTER_PID || pid == 0)) {
-            u64 ts = bpf_ktime_get_ns();
-            start.update(&pid, &ts);
+        u64 ts = bpf_ktime_get_ns();
+        if (pid != 0) {
+            if ( (!(FILTER_PID) && FILTER_TGID == 0) ||
+                 (!(FILTER_TGID) && FILTER_PID == 0)) {
```
yonghong-song: Can the above be simplied to !(FILTER_PID) && !(FILTER_TGID)?
The same for below?

tools/runqslower.py
```diff
@@ -167,19 +174,26 @@
...
+                    start.update(&pid, &ts);
+            }
         }
+
     }
 
+    bpf_probe_read(&tgid, sizeof(next->tgid), &next->tgid);
```
yonghong-song: This probably not needed. Could you double check? We have another place to bpf_probe_read tgid as well, you may want to check as well.

bodgergely:
@yonghong-song  Sorry for not resolving your comments yet... Will do so, today or over the weekend1

bodgergely:
@yonghong-song Thanks for the valuable feedback! I made the changes. Can you take a look
and see if the changes look good?

yonghong-song:
[buildbot, test this please]

yonghong-song:
@bodgergely no problem. will take a look soon.

tools/runqslower.py
```diff
@@ -42,14 +42,17 @@
 examples = """examples:
     ./runqslower         # trace run queue latency higher than 10000 us (default)
     ./runqslower 1000    # trace run queue latency higher than 1000 us
-    ./runqslower -p 123  # trace pid 123 only
+    ./runqslower -p 123  # trace pid 123 and its children threads
```
yonghong-song: just say 'trace pid 123'. people should understand what it means based linux user space convention.

tools/runqslower.py
```diff
@@ -42,14 +42,17 @@
...
+    ./runqslower -t 123  # trace tid 123 (use for threads only)
 """
 parser = argparse.ArgumentParser(
     description="Trace high run queue latency",
     formatter_class=argparse.RawDescriptionHelpFormatter,
     epilog=examples)
 parser.add_argument("-p", "--pid", type=int, metavar="PID", dest="pid",
-    help="trace this PID only")
+    help="trace this PID and its children")
```
yonghong-song: Just say "trace this PID".

tools/runqslower.py
```diff
@@ -42,14 +42,17 @@
...
+parser.add_argument("-t", "--tid", type=int, metavar="TID", dest="tid",
+    help="trace this TID only")
```
yonghong-song: "-p" and "-t" are exclusive. Please use `parser.add_mutually_exclusive_group`. See examples in offcputime.py.

tools/runqslower.py
```diff
@@ -240,7 +260,7 @@ def print_event(cpu, data, size):
     b.attach_kprobe(event="finish_task_switch", fn_name="trace_run")
 
 print("Tracing run queue latency higher than %d us" % min_us)
-print("%-8s %-16s %-6s %14s" % ("TIME", "COMM", "PID", "LAT(us)"))
+print("%-8s %-16s %-6s %14s" % ("TIME", "COMM", "TID", "LAT(us)"))
```
yonghong-song: This does not sound right. You cannot use "TID" all the time. If user specify '-p <pid>', you want to show "PID".

tools/runqslower_example.txt
```diff
@@ -41,9 +48,11 @@ positional arguments:
...
 
 examples:
     ./runqslower         # trace run queue latency higher than 10000 us (default)
     ./runqslower 1000    # trace run queue latency higher than 1000 us
-    ./runqslower -p 123  # trace pid 123 only
+    ./runqslower -p 123  # trace pid 123 and its children
```
yonghong-song: No need to say "and its children".

tools/runqslower_example.txt
```diff
@@ -41,9 +48,11 @@ positional arguments:
 
 optional arguments:
   -h, --help         show this help message and exit
-  -p PID, --pid PID  trace this PID only
+  -p PID, --pid PID  trace this PID and its children threads
```
yonghong-song: No need to say "and its children threads".

tools/runqslower_example.txt
```diff
@@ -5,8 +5,15 @@ runqslower shows high latency scheduling times between tasks being
 ready to run and them running on CPU after that. For example:
 
 # runqslower
-Tracing run queue latency higher than 10000 us
-TIME     COMM             PID           LAT(us)
+
+Note: Showing TID (thread id) in the report column. The smallest
+execution unit becomes a TID when using the --pid flag as
+in that case the tool reports not only the parent pid but
+its children threads as well.
```
yonghong-song: Please do make sure when "-p" is specified, the above contents are presented.

tools/runqslower.py
bodgergely: We might get threads showing up in case of '--pid' flag too. So using PID is also not correct in that case either.
Anyways my 2 cents are on just to simply stay with showing PID as a column value even for threads. Htop, top does the same. Showing PID/TID is bringing quite a lot possible confusion here.
Or what do you vote for?
If you prefer I can make it show PID if '--pid' and TID if  '--tid'.

yonghong-song: Think about this again. For kernel run queue, actually TID indeed makes sense. So, yes, we can just do TID here.  If user filtered through pid, showing TID should be okay too, redundant PID information may not be needed. If user did not do any filterring, TID is the right choice too.

In summary, we can just use "TID" as you changed the above. Please a comment though. Please address other comments and resubmit. Thanks!

bodgergely: Thanks, just to make it clear :)
- In case `--pid` cmd line arg -> show `PID` in the report column
- In case `--tid` cmd line arg specified -> show `TID` in the report column
Is the above what you think would make the most sense?

tools/runqslower_example.txt
bodgergely: Do you mean to emit (the below caveat for example) when running the tool with `--pid` the text:

```
Note: when using the --pid flag the tool reports both the parent pid and any possibile children threads as well.
```

bodgergely:
Updating the `man` file too. If we agree on the reporting column (TID and PID) then I will make more updates to it.

tools/runqslower.py
```diff
@@ -207,6 +218,10 @@
 }
 """
 
+
+if args.pid and args.tid:
+    parser.error("specify only one of -p and -t")
```
yonghong-song: With exclusive group in the above, the above checking should be redundant, python should have warned earlier.

bodgergely: I thought the same! I added this line because `tools/offcputime.py` had that too at line 102.
https://github.com/iovisor/bcc/blob/master/tools/offcputime.py#L102
Maybe that should be removed then too.
I will do so in mine.

bodgergely:
Let me know if good to go. Also probably I should squash the commits into a single unit before you merge.

yonghong-song:
> Also probably I should squash the commits into a single unit before you merge.
No worry. I can squash all commits into one before the merge.

PR #2645, opened 2019-12-11
Fix s390 kprobes support

Stefan-Raspl:
Ever since Linux kernel commit aa0d6e70d, kprobes on s390 could no longer
be found due to a change in symbol names.
Symptom:
  $ ./tools/execsnoop.py
  cannot attach kprobe, probe entry may not exist
  [...]

Signed-off-by: Stefan Raspl <raspl@linux.ibm.com>


yonghong-song:
[buildbot, test this please]

PR #2650, opened 2019-12-16
klockstat: fix caller filter

myllynen:
- args.caller needs to be bytes not string
- also filter calls with missing stack info
- clarify man page a bit

### DIFF
--- a/tools/klockstat.py
+++ b/tools/klockstat.py
@@ -363,7 +363,7 @@ def display(sort, maxs, totals, counts):
             stack  = list(stack_traces.walk(k.value))
             caller = b.ksym(stack[1], show_offset=True)
 
-            if (args.caller and caller.find(args.caller)):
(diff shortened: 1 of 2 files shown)


tools/klockstat.py
```diff
@@ -363,8 +363,8 @@ def display(sort, maxs, totals, counts):
             stack  = list(stack_traces.walk(k.value))
             caller = b.ksym(stack[1], show_offset=True)
 
-            if (args.caller and caller.find(args.caller)):
-                continue
+        if (args.caller and caller.find(args.caller.encode())):
```
olsajiri: I think we should keep the check under the current condition,
because we don't know what data we missed and omitting
we could mislead the user

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

PR #2651, opened 2019-12-16
tools: cgroup filtering in execsnoop/opensnoop

alban:
Add a new option --cgroupmap in execsnoop and opensnoop to only display results from processes that belong to one of the cgroups whose id, returned by bpf_get_current_cgroup_id(), is in a pinned BPF hash map.

Examples of commands:
```
# opensnoop --cgroupmap /sys/fs/bpf/test01
# execsnoop --cgroupmap /sys/fs/bpf/test01
```

Cgroup ids can be discovered in userspace by the system call name_to_handle_at(); an example of C program doing that is available in examples/cgroupid/cgroupid.c.

More complete documentation is added in docs/filtering_by_cgroups.md

The documentation is independent from Kubernetes. However, my goal is to use this feature in Kubernetes: I am preparing to use this in Inspektor Gadget to select specific Kubernetes pods, depending on a Kubernetes label selector. Kubernetes pods matching the label selector can come and go during the execution of the bcc tools; Inspektor Gadget is updating the BPF hash map used by the bcc tools accordingly.

/cc @pothos (this PR needs https://github.com/iovisor/bcc/pull/2378)

### DIFF
--- a/tools/execsnoop.py
+++ b/tools/execsnoop.py
@@ -36,6 +36,7 @@ examples = """examples:
     ./execsnoop -q        # add "quotemarks" around arguments
     ./execsnoop -n main   # only print command lines containing "main"
     ./execsnoop -l tpkg   # only print command where arguments contains "tpkg"
+    ./opensnoop --cgroupmap ./mappath  # only trace cgroups in this BPF map
 """
 parser = argparse.ArgumentParser(
     description="Trace exec() syscalls",
@@ -47,6 +48,8 @@ parser.add_argument("-t", "--timestamp", action="store_true",
     help="include timestamp on output")
 parser.add_argument("-x", "--fails", action="store_true",
     help="include failed exec()s")
+parser.add_argument("--cgroupmap",
+    help="trace cgroups in this BPF map only")
 parser.add_argument("-q", "--quote", action="store_true",
     help="Add quotemarks (\") around arguments."
     )
@@ -84,6 +87,9 @@ struct data_t {
     int retval;
 };
 
+#if CGROUPSET
+BPF_TABLE_PINNED("hash", u64, u64, cgroupset, 1024, "CGROUPPATH");
+#endif
 BPF_PERF_OUTPUT(events);
 
 static int __submit_arg(struct pt_regs *ctx, void *ptr, struct data_t *data)
@@ -108,6 +114,13 @@ int syscall__execve(struct pt_regs *ctx,
     const char __user *const __user *__argv,
     const char __user *const __user *__envp)
 {
+#if CGROUPSET
+    u64 cgroupid = bpf_get_current_cgroup_id();
+    if (cgroupset.lookup(&cgroupid) == NULL) {
+      return 0;
+    }
+#endif
+
     // create data here and pass to submit_arg to save stack space (#555)
     struct data_t data = {};
     struct task_struct *task;
@@ -141,6 +154,13 @@ out:
 
 int do_ret_sys_execve(struct pt_regs *ctx)
 {
+#if CGROUPSET
+    u64 cgroupid = bpf_get_current_cgroup_id();
+    if (cgroupset.lookup(&cgroupid) == NULL) {
+      return 0;
+    }
+#endif
+
     struct data_t data = {};
     struct task_struct *task;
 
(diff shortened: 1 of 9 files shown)


yonghong-song:
[buildbot, test this please]

alban:
Could I get the Jenkins error? It seems I don't have access to it.

yonghong-song:
[buildbot, test this please]

yonghong-song:
@alban previous error is due to buildbot infra problem. Now the issue has been fixed and the test ran fine.

tools/execsnoop.py
```diff
@@ -29,13 +29,14 @@
...
-    ./execsnoop           # trace all exec() syscalls
-    ./execsnoop -x        # include failed exec()s
-    ./execsnoop -T        # include time (HH:MM:SS)
-    ./execsnoop -t        # include timestamps
-    ./execsnoop -q        # add "quotemarks" around arguments
-    ./execsnoop -n main   # only print command lines containing "main"
-    ./execsnoop -l tpkg   # only print command where arguments contains "tpkg"
+    ./execsnoop               # trace all exec() syscalls
+    ./execsnoop -x            # include failed exec()s
+    ./opensnoop -c ./mappath  # only trace cgroups in this BPF map
```
yonghong-song: In trace.py, we already have `-c` to represent a cgroup path such that any processes in the downward hierarchy of that cgroup path will be traced. Here `-c` means different things. I would like `-c` to be consistent across all tools.

Also, in bpf, we have cgroup_array_map which actually helps reference counted cgroup so cgroup won't go away during tracing. The cgroup map here is just a hash or array map, but user could also pin a cgroup array map (BPF_CGROUP_ARRAY) and use it in the bpf program. The advantage of this is cgroup won't go away doing tracing. The drawback is it is not able to trace arbitrary cgroups. So we probably needs both in the future.

So I recommend we just use long name `--cgroupmap` for this option.  In the future, we can implement `-c` for BPF_CGROUP_ARRAY style implementation (similar to trace.py) or `--cgrouparray` for pinned cgroup array.

alban: It makes sense, I'll keep the long name `--cgroupmap` and remove the short name.

If this PR gets accepted, I would like to follow-up with the same cgroup filtering in some other tools (capable, tcpconnect, tcpaccept...). And `tcpconnect` already has a `-c` option too. So that's another argument for using the long name only.

yonghong-song:
[buildbot, ok to test]

tools/execsnoop.py
yonghong-song: sounds good then.

brendangregg:
Yes, we generally use -c for "and run this command". So it can be --cgroupmap for now, and we can add a single letter later if there's one we agree on.

https://www.slideshare.net/brendangregg/designing-tracing-tools-67693476/22

brendangregg:
I'd also put filtering_by_cgroups.txt in /docs

In /tools there's only three things (I hope!):
- toolname.py (or .lua)
- toolname_example.txt
- lib/ for libraries

So generally if you see a *.txt, there's a companion tool. filtering_by_cgroups.txt would break that.

tools/filtering_by_cgroups.txt
```diff
@@ -0,0 +1,53 @@
...
+	--privileged \
+	-v /sys/fs/cgroup:/sys/fs/cgroup \
+	cgroupid \
+	cgroupid $CGROUP | tr -d '\r')
+
+Once you discovered the cgroup id, it can be added in the BPF hash map in the
+following way:
+
+# FILE=/sys/fs/bpf/test01
+# CGROUPID_HEX=$(printf %016x $CGROUPID | sed -e 's/\(..\)/0x\1\n/g' | tac)
```
alban: I think we could add an option in `examples/cgroupid/cgroupid.c` to display the number in the format expected by bpftool (with host endianness), instead of using `sed` + `tac` in the documentation.

yonghong-song: That will be good. People might be confused what to put for `value hex ...` stuff. More clarify is better.

alban: Done

alban:
Patch updated.

yonghong-song:
LGTM. Thanks!

PR #2653, opened 2019-12-17
Extend TableBase with json output and structs with more than 2 keys

wvaske:
The current print methods on python.bcc.table.TableBase are hard coded to print a non-easily machine readable output and are also hard coded to be linear or logarithmic. This extension adds a print method that outputs a machine readable json format and is flexible to logarithmic or linear based on the functions that are passed to the method call. This method is also flexible on the number of keys in the struct that are supported. Additional keys will result in further dict nestings in the eventual output. 

See the example for the included tool: jsonbiohists.py


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

yonghong-song:
[buildbot, test this please]

yonghong-song:
do we need a new tool for this? Can we just add an option to biolatency.py to output with json format?

yonghong-song:
cc @brendangregg

wvaske:
It would be a small effort to implement the json print in biolatency.py and bite_size.py. It would be a larger effort to add the ability to do device AND flag AND pid to both scripts. Also, I was wary of modifying biolatency and bite_size as those are pretty foundational scripts that I think other people use as examples to get started.

I can go either way once Brendan weighs in.

wvaske:
Still waiting for someone to weigh in on this. Any thoughts on changes necessary for this to be merged in?

PR #2655, opened 2019-12-18
Remove kretprobe__ prefix from program names

pchaigno:
Such prefixes are currently removed for kprobes, tracepoints, and raw tracepoints, but not for kretprobes.  This pull request removes it for kretprobe as well, for consistency.

One consequence of this change is that a bcc tool or script with both `kprobe__some_func` and `kretprobe__some_func` will now load both BPF programs under the same name, i.e., `some_func`. I don't know if that's an issue (?)

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -539,6 +539,8 @@ int bcc_prog_load_xattr(struct bpf_load_program_attr *attr, int prog_len,
   if (name_len) {
     if (strncmp(attr->name, "kprobe__", 8) == 0)
       name_offset = 8;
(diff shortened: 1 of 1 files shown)


yonghong-song:
Currently `bpftool p s` will show kretprobe bpf programs and `kprobe` type as well since in kernel it is indeed kprobe type. The difference between kprobe and kretprobe only happens at attach time. But people can use `bpftool perf` to get more detailed information about whether a program is attaching to kprobe or kretprobe.

So I am okay with the change. But let us wait a little bit to see whether we have complaints or other suggestions.

PR #2659, opened 2019-12-21
tools/trace.py: flush stdout for each event

alban:


yonghong-song:
[buildbot, test this please]

PR #2683, opened 2020-01-03
tools/offcputime.py: Redundant arg check

bodgergely:
pid/tid is already set to be mutually exclusive in
python's arg parser so no need for this check manually.


yonghong-song:
[buildbot, test this please]

PR #2685, opened 2020-01-04
rewriter: Fix tracking of pointers with several indirections

pchaigno:
The rewriter tracks, with `nb_derefs`, the number of indirections of external pointers, to rewrite only the appropriate dereference into a call to `bpf_probe_read`.

In `ProbeChecker`, however, `nb_derefs` has a different meaning: it counts the number of dereferences encountered. This difference resulted in an error when `ProbeChecker` is traversing the right-hand side of an assignment. This pull request fixes the error and adds tests for the two cases of assignments: (1) when the right-hand side is an external pointer with several indirection levels and (2) when it is a call to a function returning an external pointer with several indirection levels.

This pull request also changes `ProbeSetter` and `assignsExtPtr` to count dereferences instead of addrof in an effort to use `nb_derefs` more consistently across the code to mean "number of dereferences needed to get to the external pointer".

### DIFF
--- a/tests/python/test_clang.py
+++ b/tests/python/test_clang.py
@@ -499,7 +499,34 @@ int test(struct pt_regs *ctx, struct sock *sk) {
         b = BPF(text=text)
         fn = b.load_func("test", BPF.KPROBE)
 
-    def test_probe_read_nested_deref_func(self):
+    def test_probe_read_nested_deref3(self):
+        text = """
+#include <net/inet_sock.h>
+int test(struct pt_regs *ctx, struct sock *sk) {
+    struct sock **ptr1, **ptr2 = &sk;
+    ptr1 = &sk;
+    return (*ptr1)->sk_daddr + (*ptr2)->sk_daddr;
+}
+"""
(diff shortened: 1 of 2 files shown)


tests/python/test_clang.py
```diff
@@ -499,7 +499,32 @@ def test_probe_read_nested_deref2(self):
...
-    def test_probe_read_nested_deref_func(self):
+    def test_probe_read_nested_deref3(self):
+        text = """
+#include <net/inet_sock.h>
+int test(struct pt_regs *ctx, struct sock *sk) {
+    struct sock **ptr1, **ptr2 = &sk;
+    ptr1 = &sk;
+    return (*ptr1)->sk_daddr + (*ptr2)->sk_daddr;
+}
+"""
```
yonghong-song: Missing
```
b = BPF(text=text)
fn = b.load_func("test", BPF.KPROBE)
```
here?

pchaigno: Ah, yes. I fixed it in local and forgot to commit it. I just updated the branch.

PR #2710, opened 2020-01-17
Don't enter target process mount NS for uprobes

### LINKED ISSUE #2558: USDT, attach_uprobe in  docker containerised java/nodejs  applications (tools/lib/ustat)
Hi, I'm trying to develop some code based on the ustat in tools/lib that will be able to trace the execution of managed runtime java/nodejs applications - thiscode currently works fine on a host without executing in a container.  So my first question is does anyone have any simple get me started examples of docker container approaches to bcc/eBPF analysis of applications? 

I've created an ubuntu 19.04 image ...  and also tried running with stock images.
Running my bcc tool on the host (not within the container), and with containerised execution of the application - I can see the creation of a new java/nodejs process using the same techniques as ustat.  I can attach to the sched_switch tracepoint on the host (my code not in ustat) and can read performance counters for microarchitecture events of the PIDS/TIDS belonging to  the java/nodejs application.

ustat is not attaching to USDTs in the containerised java application. I also have a need to do some function hooking using attach_uprobe for internal java/nodejs events that I want to track and these are also not picking up events - so I need to point to the debug symbols somehow as well.

NOTE: all of this works when not using containerised execution. 

I'm aware of Cilium - but have not attempted to dive into it ...  I guess my understanding of containerised execution is missing something obvious ...

Any help/pointers for information and/or explanations would be very helpful.

Thanks, and apologies in advance for any dumbness on my part.,

Andy

dalehamel:
On the legacy path, for kernels that don't support creating uprobes via
the perf_event_open (kernels prior to e12f03d), the path to the probe is
already absolute, including /proc/PID/root prefix. Unless the target process
is running within the host PID namespace, this path will be invalid.

This should fix #2558 for older kernels

# Description of issue

Older kernels (in my case, 4.14 on Chromium-based OS) use this branch to initialize probes:

https://github.com/iovisor/bcc/blob/ae5a169a0e9e47b57b7e3a8cc2c514e7ad4f3813/src/cc/libbpf.c#L1023-L1028

Rather than the newer (default) approach:

https://github.com/iovisor/bcc/blob/ae5a169a0e9e47b57b7e3a8cc2c514e7ad4f3813/src/cc/libbpf.c#L1020-L1021

The path specified by `config1` was modified in https://github.com/iovisor/bcc/pull/2324 to prepend `/proc/PID/root`, to make sure the path is relative to the target PID's mount namespace. 

However, in this legacy code path, there is already a call to enter the target namespace:

https://github.com/iovisor/bcc/blob/071eef6b1571f23be6f0fc03787523cb06b70659/src/cc/libbpf.c#L990

When the code tries to initialize the probe by writing to the kernel file descriptor:

https://github.com/iovisor/bcc/blob/071eef6b1571f23be6f0fc03787523cb06b70659/src/cc/libbpf.c#L993-L995

The writes on the failing host look like:

```
...
write(13, "p:uprobes/p__proc_3978339_root_usr_local_bin_memcached_7a66_1_bcc_296056 /proc/3978339/root/usr/local/bin/memcached:0x7a66", 122) = -1 ENOENT (No such file or directory)
write(2, "cannot attach uprobe, probe entry may not exist\n", 48cannot attach uprobe, probe entry may not exist
...
```

I reverted BCC to https://github.com/iovisor/bcc/commit/06e30cca8342aed6a1cd83de65d17945499ab6bb and was able to get it to succeed, and the write looks like:

```
...
write(16, "p:uprobes/p__usr_local_bin_memcached_7a66_1_bcc_293830 /usr/local/bin/memcached:0x7a66", 86) = 86
...
```

But the commit from #2324 (https://github.com/iovisor/bcc/commit/01ee0b61a854e06cb62272ddfbb4eeaf16e9ad7c#diff-918ef818c481c73f92c4ce607b1ec386R678) changes this to prepend the namespace.

# Summary of change

The writes can still succeed if the target process is in the host mount namespace, but if the target process is containerized it is very likely not, and the path that is given will be invalid in its mount namespace (it will appear to see pids starting from lower numbers, and not even know its own system PID).

These are the cases where this code fails. The solution is to simply not enter the namespace, as thanks to #2324 we don't need to. This makes the path relative to the host PID namespace.

# Alternatives

An alternative would be to strip the prepended string. This may actually be desirable, as the tracing process (in this case bpftrace) may not be running in the host mount namespace. This would allow for the path resolution to always be accurate, and I'm beginning to convince myself it would be a better approach, with something like this pseudo-ish code:

```c

char* probe_path = config1;
if (strncmp(config1, "/proc", 5)
{
  char* path_component = strtok(probe_path, "/"); // /proc/PID/root/PATH
  path_component = strtok(NULL, "/"); //  /PID/root/PATH
  path_component = strtok(NULL, "/"); // /root/PATH

  if(strncmp(config1, "/root")
  {
    path_component = strtok(NULL, "/");//  /PATH
    probe_path = path_component;
  }
}
  

```

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -880,68 +880,6 @@ static int bpf_attach_tracing_event(int progfd, const char *event_path, int pid,
   return 0;
 }
 
-static int enter_mount_ns(int pid) {
-  struct stat self_stat, target_stat;
-  int self_fd = -1, target_fd = -1;
-  char buf[64];
-
-  if (pid < 0)
-    return -1;
-
-  if ((size_t)snprintf(buf, sizeof(buf), "/proc/%d/ns/mnt", pid) >= sizeof(buf))
-    return -1;
-
-  self_fd = open("/proc/self/ns/mnt", O_RDONLY);
-  if (self_fd < 0) {
-    perror("open(/proc/self/ns/mnt)");
-    return -1;
-  }
-
-  target_fd = open(buf, O_RDONLY);
-  if (target_fd < 0) {
-    perror("open(/proc/<pid>/ns/mnt)");
-    goto error;
-  }
-
-  if (fstat(self_fd, &self_stat)) {
-    perror("fstat(self_fd)");
-    goto error;
-  }
-
-  if (fstat(target_fd, &target_stat)) {
-    perror("fstat(target_fd)");
-    goto error;
-  }
-
-  // both target and current ns are same, avoid setns and close all fds
-  if (self_stat.st_ino == target_stat.st_ino)
-    goto error;
-
-  if (setns(target_fd, CLONE_NEWNS)) {
-    perror("setns(target)");
-    goto error;
-  }
-
-  close(target_fd);
-  return self_fd;
-
-error:
-  if (self_fd >= 0)
-    close(self_fd);
-  if (target_fd >= 0)
-    close(target_fd);
-  return -1;
-}
-
-static void exit_mount_ns(int fd) {
-  if (fd < 0)
-    return;
-
-  if (setns(fd, CLONE_NEWNS))
-    perror("setns");
-  close(fd);
-}
-
 /* Creates an [uk]probe using debugfs.
  * On success, the path to the probe is placed in buf (which is assumed to be of size PATH_MAX).
  */
@@ -950,7 +888,7 @@ static int create_probe_event(char *buf, const char *ev_name,
                               const char *config1, uint64_t offset,
                               const char *event_type, pid_t pid, int maxactive)
 {
-  int kfd = -1, res = -1, ns_fd = -1;
+  int kfd = -1, res = -1;
   char ev_alias[256];
   bool is_kprobe = strncmp("kprobe", event_type, 6) == 0;
 
@@ -988,7 +926,6 @@ static int create_probe_event(char *buf, const char *ev_name,
       close(kfd);
       return -1;
     }
-    ns_fd = enter_mount_ns(pid);
   }
 
   if (write(kfd, buf, strlen(buf)) < 0) {
@@ -1000,14 +937,10 @@ static int create_probe_event(char *buf, const char *ev_name,
     goto error;
   }
   close(kfd);
-  if (!is_kprobe)
-    exit_mount_ns(ns_fd);
   snprintf(buf, PATH_MAX, "/sys/kernel/debug/tracing/events/%ss/%s",
            event_type, ev_alias);
   return 0;
 error:
-  if (!is_kprobe)
-    exit_mount_ns(ns_fd);
   return -1;
 }
 
--- a/tests/cc/test_usdt_probes.cc
+++ b/tests/cc/test_usdt_probes.cc
@@ -184,6 +184,21 @@ static int probe_num_arguments(const char *bin_path, const char *func_name) {
   return num_arguments;
 }
 
+// Unsharing pid namespace requires forking
+// this uses pgrep to find the child process, by searching for a process
+// that has the unshare as its parent
+static int unshared_child_pid(const int ppid) {
+  int child_pid;
+  char cmd[512];
+  const char *cmdfmt = "pgrep -P %d";
+
+  sprintf(cmd, cmdfmt, ppid);
+  if (cmd_scanf(cmd, "%d", &child_pid) != 0) {
+    return -1;
+  }
+  return child_pid;
+}
+
 TEST_CASE("test listing all USDT probes in Ruby/MRI", "[usdt]") {
   size_t mri_probe_count = 0;
 
@@ -291,5 +306,58 @@ TEST_CASE("test listing all USDT probes in Ruby/MRI", "[usdt]") {
       REQUIRE(probe->num_arguments() == exp_arguments);
       REQUIRE(probe->need_enable() == true);
     }
+  }
+}
+
+// These tests are expected to fail if there is no Ruby with dtrace probes
+TEST_CASE("test probing running Ruby process in namespaces",
+          "[usdt][!mayfail]") {
+  SECTION("in separate mount namespace") {
+    static char _unshare[] = "unshare";
+    const char *const argv[4] = {_unshare, "--mount", "ruby", NULL};
+
+    ChildProcess unshare(argv[0], (char **const)argv);
+    if (!unshare.spawned())
+      return;
+    int ruby_pid = unshare.pid();
+
+    ebpf::BPF bpf;
+    ebpf::USDT u(ruby_pid, "ruby", "gc__mark__begin", "on_event");
+    u.set_probe_matching_kludge(1);  // Also required for overlayfs...
+
+    auto res = bpf.init("int on_event() { return 0; }", {}, {u});
+    REQUIRE(res.msg() == "");
+    REQUIRE(res.code() == 0);
+
+    res = bpf.attach_usdt(u, ruby_pid);
+    REQUIRE(res.code() == 0);
+
+    res = bpf.detach_usdt(u, ruby_pid);
+    REQUIRE(res.code() == 0);
+  }
+
+  SECTION("in separate mount namespace and separate PID namespace") {
+    static char _unshare[] = "unshare";
+    const char *const argv[7] = {_unshare,       "--fork", "--mount", "--pid",
+                                 "--mount-proc", "ruby",   NULL};
+
+    ChildProcess unshare(argv[0], (char **const)argv);
+    if (!unshare.spawned())
+      return;
+    int ruby_pid = unshared_child_pid(unshare.pid());
+
+    ebpf::BPF bpf;
+    ebpf::USDT u(ruby_pid, "ruby", "gc__mark__begin", "on_event");
+    u.set_probe_matching_kludge(1);  // Also required for overlayfs...
+
+    auto res = bpf.init("int on_event() { return 0; }", {}, {u});
+    REQUIRE(res.msg() == "");
+    REQUIRE(res.code() == 0);
+
+    res = bpf.attach_usdt(u, ruby_pid);
+    REQUIRE(res.code() == 0);
+
+    res = bpf.detach_usdt(u, ruby_pid);
+    REQUIRE(res.code() == 0);
   }
 }


dalehamel:
@palmtenor @davemarchevsky please review.

I'm not sure what to do for a regression test : /

I tried this out though and on my older kernel it works with / without the target process running in the root process namespace / same process namespace as bpftrace.

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

src/cc/libbpf.c
```diff
@@ -949,7 +949,7 @@ static int create_probe_event(char *buf, const char *ev_name,
                               const char *config1, uint64_t offset,
```
dalehamel: the root issue here is that config1 already contains a path that is within the target process's mount namespace since #2324

dalehamel:
The CI failure seems to be a license issue?

```
34: Test command: /home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1710/build/tests/wrapper.sh "py_test_license" "sudo" "/home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1710/tests/python/test_license.py"
34: Test timeout computed to be: 9.99988e+06
34: /virtual/main.c:2:2: error: BPF_LICENSE cannot be specified through cflags
34: #error BPF_LICENSE cannot be specified through cflags
34:  ^
34: 1 error generated.
34: x/virtual/main.c:2:2: error: BPF_LICENSE cannot be specified through cflags
34: #error BPF_LICENSE cannot be specified through cflags
34:  ^
34: 1 error generated.
34: x.bpf: Failed to load program: Invalid argument
34: 0: (bf) r6 = r1
34: 1: (85) call bpf_ktime_get_ns#5
34: cannot call GPL only function from proprietary program
```

Flaky test maybe?

yonghong-song:
Yes, could be flaky. Will rerun the test.

yonghong-song:
[buildbot, test this please]

dalehamel:
@yonghong-song I thought about this a bit more and I think that stripping the leading string may be a better solution, as calling the probe relative to the process mount namespace is desirable, I think, it's just the PID component that is the issue. This was the behavior prior to the change. I updated the PR description with a more detailed explanation of the issue. What approach do you think is better?

yonghong-song:
> @yonghong-song I thought about this a bit more and I think that stripping the leading string may be a better solution, as calling the probe relative to the process mount namespace is desirable, I think, it's just the PID component that is the issue. This was the behavior prior to the change. I updated the PR description with a more detailed explanation of the issue. What approach do you think is better?

@dalehamel I prefer to use the first approach as the original idea for @davemarchevsky ' patch is to avoid entering namespace and still be able to trace the process running inside the namespace. But I am not  actively using bcc inside containers. Due to pid issue (pid is always the host pid), I suspect not many people are using that either. Once we got the helper to check container/namespace pid in bpf program and more people starting to use bcc inside containers, we can go back to revisit the issue.

@palmtenor @davemarchevsky could you comment on this patch?

dalehamel:
> But I am not actively using bcc inside containers. Due to pid issue (pid is always the host pid), I suspect not many people are using that either. 

Makes sense, I was thinking that resolving the symbols relative to the target process namespace might in the future allow for bcc scripts and bpftrace to have less permissions than they otherwise might need, but for i think that requiring a full view of the host's process namespace is a pretty reasonable privilege capability requirement. This solution is also simpler, and it makes sense to me to keep the path consistent with the one that is passed to the newer codepath.

davemarchevsky:
> The solution is to simply not enter the namespace, as thanks to #2324 we don't need to. This makes the path relative to the host PID namespace.

I think I agree with this logic, or at least 'avoid entering namespace' was the impetus behind #2324.

Some background: none of the `setns`-based stuff in USDT code was working for me because my program using `bcc` was multithreaded. Until a few months ago `setns` manpage explicitly said "A process may not be reassociated with a new mount namespace if it is mutithreaded", until this [was changed](https://github.com/mkerrisk/man-pages/commit/5dd76c44497dd43506f79995a8c7f84a105864e4#diff-bc6cccf396c9e8b47ba4db94b32ec1b5) to be more precise and say "can't reassociate with new mount ns if your process called `clone()` with `CLONE_FS` or is otherwise sharing fs attributes".

This is a pretty onerous restriction, so it seemed like a good idea to stop relying on `setns` to switch mount namespace and instead assume that we can go through procfs to get the right fd to look for USDTs in. This introduces a different set of assumptions / restrictions. Perhaps ultimately the best path forward is to provide an API option which makes minimal assumptions.

Anyways, just wanted to lay out my assumptions explicitly in case they don't work for @dalehamel or otherwise make sense. Code LGTM.

dalehamel:
> manpage explicitly said "A process may not be reassociated with a new mount namespace if it is mutithreaded", until this was changed to be more precise and say "can't reassociate with new mount ns if your process called clone() with CLONE_FS or is otherwise sharing fs attributes".

Thanks for this context @davemarchevsky that makes a lot of sense! In which case I agree that just dropping this call is the best approach. I didn't realize that this multithreaded restriction existed.

In which case, as this change didn't have a failing test associated with it, any suggestions for how to test for this edge case? It looks like some of the Jenkins build run against an old enough kernel that we should be able to catch this code path...

dalehamel:
I see there is an existing framework for launching a subprocess, I think I can use this to make a good regression test, assuming the kernels on the test runners are what would typically ship with fedora / Ubuntu.

For instance, I should be able to just specify the following for the subprocess:

```
unshare -f -m -p --mount-proc ruby
```

This should allow for an appropriate regression test to be written. I'll set to work on this, then I think this should be ok to merge? Ideally I'd like to also verify that the test should fail without my change.

yonghong-song:
@dalehamel yes, it will be good if you can write a regression test which touches the code you modified. I will wait for that before merging the patch. Thanks!

yonghong-song:
The buildbot got the following compilation errors:
```
[ 97%] Building CXX object tests/cc/CMakeFiles/test_libbcc.dir/test_usdt_probes.cc.o
/home/fedora/jenkins/workspace/bcc-pr/label/fc28/tests/cc/test_usdt_probes.cc: In function ‘void ____C_A_T_C_H____T_E_S_T____297()’:
/home/fedora/jenkins/workspace/bcc-pr/label/fc28/tests/cc/test_usdt_probes.cc:300:70: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
     char *const argv[3] = {_unshare, "--mount", "/usr/local/bin/ruby"};
                                                                      ^
/home/fedora/jenkins/workspace/bcc-pr/label/fc28/tests/cc/test_usdt_probes.cc:300:70: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
/home/fedora/jenkins/workspace/bcc-pr/label/fc28/tests/cc/test_usdt_probes.cc:307:12: warning: format ‘%d’ expects argument of type ‘int’, but argument 3 has type ‘size_t’ {aka ‘long unsigned int’} [-Wformat=]
     printf("Ruby pid %d has %d", ruby.pid(), ctx.num_probes());
            ^~~~~~~~~~~~~~~~~~~~              ~~~~~~~~~~~~~~~~
/home/fedora/jenkins/workspace/bcc-pr/label/fc28/tests/cc/test_usdt_probes.cc:335:75: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
                            "--pid",  "--mount-proc", "/usr/local/bin/ruby"};
                                                                           ^
/home/fedora/jenkins/workspace/bcc-pr/label/fc28/tests/cc/test_usdt_probes.cc:335:75: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
/home/fedora/jenkins/workspace/bcc-pr/label/fc28/tests/cc/test_usdt_probes.cc:335:75: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
/home/fedora/jenkins/workspace/bcc-pr/label/fc28/tests/cc/test_usdt_probes.cc:335:75: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
/home/fedora/jenkins/workspace/bcc-pr/label/fc28/tests/cc/test_usdt_probes.cc:335:75: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
[ 97%] Building CXX object tests/cc/CMakeFiles/test_libbcc.dir/utils.cc.o
```

The `test_libbcc` failed. The error messages are below, I guess ruby binary is not available in the buildbot machines.
```
.....
3: unshare: failed to execute /usr/local/bin/ruby: No such file or directory
3: Ruby pid 10839 has 0
3: ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
3: test_libbcc is a Catch v1.4.0 host application.
3: Run with -? for options
3: 
3: -------------------------------------------------------------------------------
3: test probing running Ruby process in namespaces
3:   in separate mount namespace
3: -------------------------------------------------------------------------------
3: /home/fedora/jenkins/workspace/bcc-pr/label/fc28/tests/cc/test_usdt_probes.cc:297
3: 
3: /home/fedora/jenkins/workspace/bcc-pr/label/fc28/tests/cc/test_usdt_probes.cc:309: FAILED:
3:   REQUIRE( ctx.num_probes() > 10 )
3: with expansion:
3:   0 > 10
3: 
3: unshare: failed to execute /usr/local/bin/ruby: No such file or directory
3: -------------------------------------------------------------------------------
3: test probing running Ruby process in namespaces
3:   in separate mount namespace and separate PID namespace
3: -------------------------------------------------------------------------------
3: /home/fedora/jenkins/workspace/bcc-pr/label/fc28/tests/cc/test_usdt_probes.cc:297
3: 
3: /home/fedora/jenkins/workspace/bcc-pr/label/fc28/tests/cc/test_usdt_probes.cc:342: FAILED:
3:   REQUIRE( ctx.num_probes() > 10 )
3: with expansion:
3:   0 > 10
3: 
3: ===============================================================================
3: test cases:  37 |  36 passed | 1 failed
3: assertions: 639 | 637 passed | 2 failed
3: 
3: Failed
 3/40 Test  #3: test_libbcc ......................***Failed   12.28 sec
```

yonghong-song:
Thanks for introducing github workflow! I think bcc should use more of them for testing in the future.

dalehamel:
thanks @yonghong-song the work on this branch isn't my latest, I was aware of that issue and have been trying to debug it on Github Actions VMs. I'm running into some weird issues though, so I've also been trying to debug this locally. Since I can't kick off jenkins runs (I don't think?) I added a basic github actions file to try this out.

I'll keep debugging this, ignore anything I push here for the time being until I say it's ready to review.

> Thanks for introducing github workflow! I think bcc should use more of them for testing in the future.

I was also thinking this, the nice thing about it is that anyone who PRs can easily kick off their own runs, but I think a separate github issue should be cut for this, I've been having some issues replicating the test environment. Is there any documentation available for how to run tests that I've just been missing?

dalehamel:
@davemarchevsky it looks like I have to set `set_probe_matching_kludge(1)` on overlayfs as well, so I don't think it is limited to btrfs : /

I have been trying to write a test case that will actually probe a process and was trying to probe a ruby process, as there's already test cases doing this. In trying to run these tests in docker using the overlayfs graph driver, the major and minor numbers don't match. I inserted some log lines in `_bcc_syms_find_module`:

```
Looking by inode /proc/129261/root/usr/local/bin/ruby /usr/local/bin/ruby 1926452 1926452
MAJ 0 8 MIN 62 3
```

So the inodes match, the paths don't match, and neither do the major or minor numbers. By setting the kludge it matches by inode and works. This is pretty worrisome though, as it requires bcc to have knowledge of the underlying filesystem, which is a leaky abstraction. I have no btrfs anywhere, and I think that using overlayfs is very common in the docker/container ecosystem...

For now I'll set this kludge, but I'm worried about the implications of this leaky API.

dalehamel:
@yonghong-song I now have a working regression test, if I revert my changes the new test for a separate mount and process namespace fails as expected. However, for some reason the unshare command isn't able to find the path to `ruby`, where other USDT tests seem to be able to. Are you able to confirm the path to the Ruby binary in Jenkins? I would have expected that unshare would inherit and use the PATH variable, so I'm a little surprised...

dalehamel:
Oh weird, it looks like there isn't *any* Ruby in jenkins environment? I pushed a temporary commit to do `which ruby`, which locally shows me the path to ruby, and in jenkins shows nothing. It also looks like it is skipping a *lot* of assertions. Locally I see:

```
test cases:   37 |   35 passed | 2 failed
assertions: 2414 | 2412 passed | 2 failed
```

But on jenkins for libbcc I see:

```
3: test cases:  37 |  36 passed | 1 failed
3: assertions: 626 | 624 passed | 2 failed
```

So i guess these other Ruby tests aren't even being run at all on Jenkins?

Another way to work around this problem would be to compile a small test program like bpftrace tests do, and use that as a target instead of Ruby...

tests/cc/test_usdt_probes.cc
```diff
@@ -293,3 +308,58 @@ TEST_CASE("test listing all USDT probes in Ruby/MRI", "[usdt]") {
...
+    res = bpf.detach_usdt(u, ruby_pid);
+    REQUIRE(res.code() == 0);
+  }
+
+  SECTION("in separate mount namespace and separate PID namespace") {
+    static char _unshare[] = "unshare";
+    const char *const argv[7] = {_unshare,       "--fork", "--mount", "--pid",
+                                 "--mount-proc", "ruby",   NULL};
+
+    char rubypath[100];
```
dalehamel: This is a temporary hack that attempts to find Ruby on Jenkins, but it doesn't return anything... I now suspect these other Ruby tests are somehow skipped by Jenkins?

When I run this test myself with docker on Github actions, such as I have in a draft PR #2714 ruby is located without issue by `unshare` above.

yonghong-song:
The buildbot still cannot find ruby
```
which: no ruby in (/sbin:/bin:/usr/sbin:/usr/bin)
3: unshare: failed to execute ruby: No such file or directory
3: Ruby path is �~<��
```
Let me double check.

yonghong-song:
I do not know how to access the jenkins slave machine to check whether ruby is installed or not. Need to ask @drzaeus77 

But I checked the existing test code. Looks like if ruby is not installed, the test will be successfully as it will return okay if ruby is not in the PATH. For example,
```
    USDT::Context ctx("ruby");

    if (!ctx.loaded())
      return;
```
I think your test can have a similar mechanism. If ruby is not available, do not fail the test.
At the same time, let me continue trying to check the VMs...

dalehamel:
> I think your test can have a similar mechanism. If ruby is not available, do not fail the test.
At the same time, let me continue trying to check the VMs...

@yonghong-song there is a special tag in `catch2` called `[!mayfail]` which allows for some tests to fail, still report the failure, but not fail the overall suite. This might be a better option. Just returning from tests would explain why there are so many fewer assertions running in Jenkins than locally or on my Github Actions PR.

In the meantime, I'll clean this PR up as it should be good to merge, I was hoping it would run in Jenkins but it looks like that won't be possible.

tests/cc/test_usdt_probes.cc
```diff
@@ -293,3 +308,56 @@ TEST_CASE("test listing all USDT probes in Ruby/MRI", "[usdt]") {
     }
   }
 }
+
+// These tests are expected to fail if there is no Ruby with dtrace probes
+TEST_CASE("test probing running Ruby process in namespaces",
+          "[usdt][!mayfail]") {
```
dalehamel: @yonghong-song I just tagged this with !mayfail for now, so the failure will be reported but won't cause the run to fail. This feels more correct than just skipping the assertion, which I'd consider to be a testing antipattern.

Overall, I think these checks should actually be run in Jenkins, otherwise what is the point in having them?

dalehamel:
I squashed and rebased

dalehamel:
@yonghong-song this now passes because of the `[!mayfail]` tag, assuming that I can also get #2714 merged, that environment has a ruby that is able to run these tests on each PR to prevent this regression. I think this should be good to merge.

yonghong-song:
Looks good. Thanks for working on this!

PR #2716, opened 2020-01-21
fix line TypeError

DavadDi:
./profile.py -adf -p `pgrep -n main` 5
  File "./profile.py", line 342, in <module>
    print("%s %d" % (b";".join(line).decode('utf-8', 'replace'), v.value))
TypeError: sequence item 5: expected a bytes-like object, str found

### DIFF
--- a/tools/profile.py
+++ b/tools/profile.py
@@ -330,13 +330,13 @@ for k, v in sorted(counts.items(), key=lambda counts: counts[1].value):
         # hash collision (-EEXIST), we still print a placeholder for consistency
         if not args.kernel_stacks_only:
             if stack_id_err(k.user_stack_id):
-                line.append("[Missed User Stack]")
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

yonghong-song:
Your command line in the description does not seem right
```
./profile.py -adf -p pgrep -n main 5
```
`-n` is not supported currently. `-p` should be followed by a number pid.
Could you double check?

DavadDi:
“pgrep -n main” just get pid of main prog，and paas to -p

yonghong-song:
Okay, I see. Probably it means
   ./profile.py -adf -p `pgrep -n main` 5
But the "`" is missing in the description in the github window.

PR #2665, opened 2019-12-23
Changes to fix #2648

jason-chenyixi-chengdu:
Hi, 
This changes is for fix issue #2648 .  Please help to review if it OK.
Main changes are https://github.com/iovisor/bcc/pull/2665/commits/b93e4a3fc128fd58f57fa035b987a1fd2320b516 and https://github.com/iovisor/bcc/pull/2665/commits/e6b95721ba05616c9d69c014ffe84910c0ec27df. Others are some testing in my forked repo.
Thanks!

PR #2721, opened 2020-01-28
Changes to fix #2648(reference to #2665)

### LINKED ISSUE #2648: If the load address is not set to 0 in libc elf, the offset of symbols cannot retrieve correctly
Hi,
When I tried to deploy bcc in an embedded system, i found that invalid argument error orrcured. After troubleshooting, I found that is the offset of 'malloc'(i tried to execute memleak) in libc.so is bigger than the size of libc. That is because the load address is not set to 0 in my libc, so the value of malloc in symbol table is not offset. What i have done to fix it is retrieve the base load address from programe header first, then substracted the value by the base address.

jason-chenyixi-chengdu:
This is another request for #2665

### DIFF
--- a/src/cc/bcc_syms.cc
+++ b/src/cc/bcc_syms.cc
@@ -683,7 +683,17 @@ static int _find_sym(const char *symname, uint64_t addr, uint64_t,
                      void *payload) {
   struct bcc_symbol *sym = (struct bcc_symbol *)payload;
   if (!strcmp(sym->name, symname)) {
-    sym->offset = addr;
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

yonghong-song:
@palmtenor @davemarchevsky Could you help review the patch?

palmtenor:
Just came back from chinese new year. Will take a look shortly. Thanks for the fix.

PR #2723, opened 2020-01-29
Make inode-only matching default, reverse USDT kludge

### LINKED ISSUE #2715: set_probe_matching_kludge is required for overlayfs
To attach to probes on overlayfs, set_probe_matching_kludge is required:

https://github.com/iovisor/bcc/blob/da59f37f14320c0a20300f2c516336ae9ebe8a30/src/cc/bcc_syms.cc#L619-L623

This is further commented in:

https://github.com/iovisor/bcc/blob/da59f37f14320c0a20300f2c516336ae9ebe8a30/src/cc/api/BPF.h#L308-L319

The default behavior when looking for module info for USDT probes is to prefer to match by MAJ:MIN rather than by inode ID

This kludge flag causes the inode ID instead in order to verify that the requested USDT symbol is in a particular file, in order to attach the uprobe to it, and possibly modify semaphores USDT semaphores. This was introduced by https://github.com/iovisor/bcc/commit/01ee0b61a854e06cb62272ddfbb4eeaf16e9ad7c, in #2324 

This means that there are now two filesystems that need this kludge, btrfs (as documented) and overlayfs (not documented except, I think, by this issue). Overlayfs is particularly problematic, because it is the default storage for the graph driver used by most production deployments of docker and containers, such as by kubernetes.

To support USDT tracing on these platforms, code will require this kludge with the majority of tools will probably not know to set.

Philosophically, this also feels like a bit of a leaky abstraction - this is a detail of the filesystem that the binary being probed resides on, which shouldn't really need to be a concern of the caller.

I think we should either:

- Remove this kludge entirely, making resolution by inode the default behavior. In what cases is the MAJ MIN preferred? Is it needed to avoid a collision in some cases? How likely is an inode collision, if it is possible?
- Reverse the kludge, to make the behavior to use MAJ MIN, but make by inode the default. Ok, this is super confusing behavior for a consumer
- Detect what filesystem the target binary resides on, and handle this kludge automatically, and remove the public API as it is no longer necessary.

@davemarchevsky @yonghong-song @palmtenor what are your thoughts?

dalehamel:
This PR is per the discussion in #2715.

Changing the default behavior to use inode only matching fixes these test failures on github actions, as the USDT tests now run in docker (and on overlayfs). Because of this I've been able to remove the `@mayFail` decorator from the python USDT tests, as they now pass.

### DIFF
--- a/src/cc/api/BPF.cc
+++ b/src/cc/api/BPF.cc
@@ -777,7 +777,7 @@ USDT::USDT(const std::string& binary_path, const std::string& provider,
       provider_(provider),
       name_(name),
       probe_func_(probe_func),
-      mod_match_inode_only_(0) {}
+      mod_match_inode_only_(1) {}
 
 USDT::USDT(pid_t pid, const std::string& provider, const std::string& name,
            const std::string& probe_func)
@@ -787,7 +787,7 @@ USDT::USDT(pid_t pid, const std::string& provider, const std::string& name,
       provider_(provider),
       name_(name),
       probe_func_(probe_func),
-      mod_match_inode_only_(0) {}
+      mod_match_inode_only_(1) {}
 
 USDT::USDT(const std::string& binary_path, pid_t pid,
            const std::string& provider, const std::string& name,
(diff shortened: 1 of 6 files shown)


dalehamel:
@davemarchevsky @yonghong-song please take a look

davemarchevsky:
LGTM!

dalehamel:
Passing github actions run: https://github.com/dalehamel/bcc/actions/runs/32702302

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

yonghong-song:
LGTM. Thanks!

dalehamel:
Thanks!

PR #2726, opened 2020-01-30
Introduce {attach|detach}_kfunc API

olsajiri:
Kernel added new probe called trampoline allowing to probe
almost any kernel function when BTF info is available in
the system.
    
Adding the interface under 'kfunc' name, which allows to
probe function entry and 'kretfunc' name, which allows to
trace function return.
    
 The main benefit is really low overhead which is really low
 for trampolines (please see commit for klockstat.py
 with perf comparison).
    
 Another benefit is the ability of kretfunc probe to access
 function arguments, so some tools might need only one program
 instead of entry/exit ones (please see commit for
 opensnoop.py changes).

### DIFF
--- a/tools/opensnoop.py
+++ b/tools/opensnoop.py
@@ -105,8 +105,11 @@ struct data_t {
 #if CGROUPSET
 BPF_TABLE_PINNED("hash", u64, u64, cgroupset, 1024, "CGROUPPATH");
 #endif
-BPF_HASH(infotmp, u64, struct val_t);
 BPF_PERF_OUTPUT(events);
+"""
+
+bpf_text_kprobe = """
+BPF_HASH(infotmp, u64, struct val_t);
 
 int trace_entry(struct pt_regs *ctx, int dfd, const char __user *filename, int flags)
 {
@@ -162,6 +165,41 @@ int trace_return(struct pt_regs *ctx)
     return 0;
 }
 """
+
+bpf_text_kfunc= """
+KRETFUNC_PROBE(do_sys_open, int dfd, const char *filename, int flags, int mode, int ret)
+{
+    u64 id = bpf_get_current_pid_tgid();
+    u32 pid = id >> 32; // PID is higher part
+    u32 tid = id;       // Cast and get the lower part
+    u32 uid = bpf_get_current_uid_gid();
+
+    PID_TID_FILTER
+    UID_FILTER
+    FLAGS_FILTER
+
+    struct data_t data = {};
+    bpf_get_current_comm(&data.comm, sizeof(data.comm));
+
+    u64 tsp = bpf_ktime_get_ns();
+
+    bpf_probe_read(&data.fname, sizeof(data.fname), (void *)filename);
+    data.id    = id;
+    data.ts    = tsp / 1000;
+    data.uid   = bpf_get_current_uid_gid();
+    data.flags = flags; // EXTENDED_STRUCT_MEMBER
+    data.ret   = ret;
+
+    events.perf_submit(ctx, &data, sizeof(data));
+}
+"""
+
+is_support_kfunc = BPF.support_kfunc()
+if is_support_kfunc:
+    bpf_text += bpf_text_kfunc
+else:
+    bpf_text += bpf_text_kprobe
+
 if args.tid:  # TID trumps PID
     bpf_text = bpf_text.replace('PID_TID_FILTER',
         'if (tid != %s) { return 0; }' % args.tid)
@@ -195,8 +233,9 @@ if debug or args.ebpf:
 
 # initialize BPF
 b = BPF(text=bpf_text)
-b.attach_kprobe(event="do_sys_open", fn_name="trace_entry")
-b.attach_kretprobe(event="do_sys_open", fn_name="trace_return")
+if not is_support_kfunc:
+    b.attach_kprobe(event="do_sys_open", fn_name="trace_entry")
+    b.attach_kretprobe(event="do_sys_open", fn_name="trace_return")
 
 initial_ts = 0
 
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -16,6 +16,8 @@ This guide is incomplete. If something feels missing, check the bcc and kernel s
         - [6. USDT probes](#6-usdt-probes)
         - [7. Raw Tracepoints](#7-raw-tracepoints)
         - [8. system call tracepoints](#8-system-call-tracepoints)
+        - [9. kfuncs](#9-kfuncs)
+        - [10. kretfuncs](#9-kretfuncs)
     - [Data](#data)
         - [1. bpf_probe_read()](#1-bpf_probe_read)
         - [2. bpf_probe_read_str()](#2-bpf_probe_read_str)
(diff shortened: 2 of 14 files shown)


yonghong-song:
[buildbot, ok to test]

yonghong-song:
[buildbot, test this please]

yonghong-song:
All tests failed. All the buildbot is pretty old. The latest one in buildbot is fc28 which I do not think supports BTF. For example, the following is the error message:
```
29: ....../bin/sh: line 1: 11657 Killed                  timeout -s KILL -k 5s 5s ../../tools/killsnoop.py > /dev/null
29: .s........WARNING: 15 stack traces lost and could not be displayed.
29: ..libbpf: failed to find valid kernel BTF
29: libbpf: vmlinux BTF is not found
29: /virtual/main.c:27:1: error: redefinition of 'infotmp_table_t'
29: BPF_HASH(infotmp, u64, struct val_t);
29: ^
29: /virtual/include/bcc/helpers.h:176:37: note: expanded from macro 'BPF_HASH'
29:   BPF_HASHX(__VA_ARGS__, BPF_HASH4, BPF_HASH3, BPF_HASH2, BPF_HASH1)(__VA_ARGS__)
29:                                     ^
29: /virtual/main.c:24:1: note: previous definition is here
29: BPF_HASH(infotmp, u64, struct val_t);
29: ^
29: /virtual/include/bcc/helpers.h:176:37: note: expanded from macro 'BPF_HASH'
29:   BPF_HASHX(__VA_ARGS__, BPF_HASH4, BPF_HASH3, BPF_HASH2, BPF_HASH1)(__VA_ARGS__)
29:                                     ^
29: /virtual/main.c:27:10: error: redefinition of 'infotmp'
29: BPF_HASH(infotmp, u64, struct val_t);
29:          ^
29: /virtual/main.c:24:10: note: previous definition is here
29: BPF_HASH(infotmp, u64, struct val_t);
29:          ^
29: /virtual/main.c:27:1: error: redefinition of '____btf_map_infotmp'
29: BPF_HASH(infotmp, u64, struct val_t);
29: ^
29: /virtual/include/bcc/helpers.h:176:37: note: expanded from macro 'BPF_HASH'
29:   BPF_HASHX(__VA_ARGS__, BPF_HASH4, BPF_HASH3, BPF_HASH2, BPF_HASH1)(__VA_ARGS__)
29:                                     ^
29: /virtual/main.c:24:1: note: previous definition is here
29: BPF_HASH(infotmp, u64, struct val_t);
29: ^
29: /virtual/include/bcc/helpers.h:176:37: note: expanded from macro 'BPF_HASH'
29:   BPF_HASHX(__VA_ARGS__, BPF_HASH4, BPF_HASH3, BPF_HASH2, BPF_HASH1)(__VA_ARGS__)
29:                                     ^
29: /virtual/main.c:27:1: error: redefinition of '____btf_map_infotmp'
29: BPF_HASH(infotmp, u64, struct val_t);
29: ^
29: /virtual/include/bcc/helpers.h:176:37: note: expanded from macro 'BPF_HASH'
29:   BPF_HASHX(__VA_ARGS__, BPF_HASH4, BPF_HASH3, BPF_HASH2, BPF_HASH1)(__VA_ARGS__)
29:                                     ^
29: /virtual/main.c:24:1: note: previous definition is here
29: BPF_HASH(infotmp, u64, struct val_t);
29: ^
29: /virtual/include/bcc/helpers.h:176:37: note: expanded from macro 'BPF_HASH'
29:   BPF_HASHX(__VA_ARGS__, BPF_HASH4, BPF_HASH3, BPF_HASH2, BPF_HASH1)(__VA_ARGS__)
29:                                     ^
29: 4 errors generated.
```

We will need to guard the new test with kernel version and availability of kernel BTF. Not all distro's have kernel BTF, I guess.

olsajiri:
ok, I expected some failures, but not that bad.. will check on  some BTF detection for tests

tools/opensnoop.py
```diff
@@ -107,6 +107,10 @@
 #endif
 BPF_HASH(infotmp, u64, struct val_t);
 BPF_PERF_OUTPUT(events);
+"""
+
+bpf_text_kprobe = """
+BPF_HASH(infotmp, u64, struct val_t);
```
yonghong-song: This BPF_HASH is redundant. This will cause compilation error for kprobe case.

yonghong-song:
@olsajiri any update on this one?

brendangregg:
Having the kretfunc retval in args[N] might get pretty confusing. Could it have a separate variable called retval for this?

(I was going to suggest putting retval in args[0] and bumping the others along by one, but that would make it inconsistent with kfunc.)

yonghong-song:
> Having the kretfunc retval in args[N] might get pretty confusing. Could it have a separate variable called retval for this?
I think we could design an interface with return value as a separate variable. The rewriter can kick in to generate proper code. We also need and we already have BTF to get the function signature, so we could parse and get the number of arguments for the particular function.

olsajiri:
> > Having the kretfunc retval in args[N] might get pretty confusing. Could it have a separate variable called retval for this?
> > I think we could design an interface with return value as a separate variable. The rewriter can kick in to generate proper code. We also need and we already have BTF to get the function signature, so we could parse and get the number of arguments for the particular function.

sry, I'm out this week, hence my lag.. I think you're right, we can add some better interface
for arguments.. perhaps even with the argument names, because we have it in the BTF
I'll check on this

jirka

tools/opensnoop.py
olsajiri: ugh, I must have consused my versions and pushed the wrong one,
I recall fixing this, sry.. will fix in next version

yonghong-song:
Oh we could adopt libbpf convention as well.
```
__u64 test4_result = 0;
SEC("fexit/bpf_fentry_test4")
int BPF_PROG(test4, void *a, char b, int c, __u64 d, int ret)
```

Some macros, so we do not need to add BTF analysis code although it is nice to have to double check.

In the long time, I intend to change bcc to be more libbpf friendly so we can get a lot of libbpf funcitonality for free.

olsajiri:
> Oh we could adopt libbpf convention as well.
> 
> ```
> __u64 test4_result = 0;
> SEC("fexit/bpf_fentry_test4")
> int BPF_PROG(test4, void *a, char b, int c, __u64 d, int ret)
> ```
> 
> Some macros, so we do not need to add BTF analysis code although it is nice to have to double check.
> 
> In the long time, I intend to change bcc to be more libbpf friendly so we can get a lot of libbpf funcitonality for free.

ok, that might be actually easier.. will check

thanks,
jirka

olsajiri:
@yonghong-song actually while working on bpftrace support,
I relaized we don't need to pass expected_attach_type through
bcc_func_load chain, but we can get it from the function name
prefix (kfunc__, kretfunc__).

If this kind of interface is ok with you - getting the attr info from
the function name, I'll make the changes. It will ease up bpftrace
inclusion, because we could continue to use  bcc_prog_load.

yonghong-song:
> @yonghong-song actually while working on bpftrace support,
> I relaized we don't need to pass expected_attach_type through
> bcc_func_load chain, but we can get it from the function name
> prefix (kfunc__, kretfunc__).
> 
> If this kind of interface is ok with you - getting the attr info from
> the function name, I'll make the changes. It will ease up bpftrace
> inclusion, because we could continue to use bcc_prog_load.

This should work. In bcc, we already have some other usage with special prefixes.

olsajiri:
> > @yonghong-song actually while working on bpftrace support,
> > I relaized we don't need to pass expected_attach_type through
> > bcc_func_load chain, but we can get it from the function name
> > prefix (kfunc__, kretfunc__).
> > If this kind of interface is ok with you - getting the attr info from
> > the function name, I'll make the changes. It will ease up bpftrace
> > inclusion, because we could continue to use bcc_prog_load.
> 
> This should work. In bcc, we already have some other usage with special prefixes.

cool, changes are already in the branch

yonghong-song:
LGTM. Thanks!

alban:
I get the following error in opensnoop with Linux 5.5.6. Is it a problem in bcc or in the kernel?

```
0: (bf) r6 = r1
1: (79) r1 = *(u64 *)(r6 +32)
2: (7b) *(u64 *)(r10 -304) = r1
3: (79) r7 = *(u64 *)(r6 +8)
func 'do_sys_open' arg1 type INT is not a struct
invalid bpf_context access off=8 size=8
processed 4 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
  File "/usr/share/bcc/tools/opensnoop", line 235, in <module>
    
b = BPF(text=bpf_text)
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 364, in __init__
    
self._trace_autoload()
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 1198, in _trace_autoload
    
self.attach_kretfunc(fn_name=func_name)
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 933, in attach_kretfunc
    
fn = self.load_func(fn_name, BPF.TRACING)
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 404, in load_func
    
(func_name, errstr))
Exception
: 
Failed to load BPF program kretfunc__do_sys_open: Permission denied
```

yonghong-song:
Maybe this is a kernel issue? I mean kernel is not advanced enough to have this patch?
https://www.spinics.net/lists/netdev/msg627180.html

alban:
Thanks for the pointer. Indeed my kernel does not have this patch. I wonder if `BPF.support_kfunc()` should return `false` on kernels without that patch, so that opensnoop can fallback on the previous kprobe code, or if BCC is fine asking users to update their kernels to the latest version.

I just noticed that the new kfunc code does not have the `--cgroupmap` filter.

Issue #2446, opened 2019-07-04
Use proper Python shebang lines


cclauss:
__scripts/py-style-check.sh__ [mandates the use of __#!/usr/bin/python__](https://github.com/iovisor/bcc/blob/master/scripts/py-style-check.sh#L9) but that hardcodes the location of the Python interpreter but on a brew installed Mac or inside a virtualenv, that might not be the desired location.

It would be more standard and flexible to mandate a shebang of __#!/usr/bin/env python__

pchaigno:
Related to #2146.

yonghong-song:
@cclauss The idea is to have /usr/bin/python as the default path and the package installer can have a script to change it to a different appropriate locaiton if needed.

As why use /usr/bin/python instead of `/usr/bin/env python`? This is mostly related what command line is shown (e.g., python vs. opensnoop.py). opensnoop.py is preferred vs. vague python.
https://github.com/iovisor/bcc/issues/2145

PR #2733, opened 2020-01-31
search in PATH for python

brandshaide:
This commit is improving fail safety a bit by instructing to search PATH for python in the - may unlikely - case it’s not located at /bin

### DIFF
--- a/examples/hello_world.py
+++ b/examples/hello_world.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 # Copyright (c) PLUMgrid, Inc.
 # Licensed under the Apache License, Version 2.0 (the "License")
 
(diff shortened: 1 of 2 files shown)


brandshaide:
Sorry, this is failing because my `shebang` is not matching against the expected, origin `shebang` as defined here: 

https://github.com/iovisor/bcc/blob/0e63c5c1ad267281f8d0b652eaf87dd494ddba04/scripts/py-style-check.sh#L9

Do you want me to adapt the shebang in `py-style-check.sh` accordingly to `usr/bin/env python` ?

brandshaide:
added/changed the check on `py-style-check` but apparently some other `*.py` scripts are also affected of this, as they're (no longer) containing the "right" shebang line. I could adjust this for all affected python scripts.

yonghong-song:
HI, we still like to use /usr/bin/python as the default. See issue #2446 for details.

brandshaide:
thanks, yonghong-song for pointing out, I wasn’t aware of this issue. I‘ll close this PR.

PR #2738, opened 2020-02-05
WIP of directly passing ref_ctr_offset to perf event create for USDT uprobes

### LINKED ISSUE #2230: usdt probes requiring semaphore cannot be used on google container OS
This isn't necessarily a bug in bcc per se, more a write-up of what doesn't work and why - hopefully it will at least help others from going down the rabbit hole I did.

There may be a way for bcc to fix this by finding another way to increment the semaphore, but I don't really see how.

In the chromium kernel source code, here is code in `fs/proc/base.c` that prevents processes from writing to their own memory maps for security reasons:

```
static ssize_t mem_write(struct file *file, const char __user *buf,
       size_t count, loff_t *ppos)
{
#ifdef CONFIG_SECURITY_CHROMIUMOS_READONLY_PROC_SELF_MEM
  return -EACCES;
#else
  return mem_rw(file, (char __user*)buf, count, ppos, 1);
#endif
}
```

This variable is enabled by default in the kernel used by container OS, such as in google's GKE offering: https://chromium.googlesource.com/chromiumos/overlays/board-overlays/+/master/overlay-lakitu/sys-kernel/lakitu-kernel-4_14/files/base.config#3016

This means that anyone trying to use bcc on a chromium derived OS, and especially anyone trying to use bcc on GKE, will probably also hit this if they try to use usdt probes.

During the process of enabling a usdt probe, some probes need to be enabled by writing to a semaphore - this isn't true of all usdt probes, but is probably true of many (I ran into this with ruby's usdt probes). As [the dtrace docs](http://man7.org/linux/man-pages/man1/dtrace.1.html) indicate, this is a means to avoid expensive processing around the probe, only adding this extra info/processing if the probe is enabled.

The code that handles this in bcc is here:
https://github.com/iovisor/bcc/blob/c2e2a26b8624492018a14d5eebd4a50b869c911f/src/cc/usdt/usdt.cc#L109-L113

And it is essentially the same as the approach described [here](www.brendangregg.com/blog/2015-07-03/hacking-linux-usdt-ftrace.html).

However, this leads to probes silently failing to be enabled if run against a kernel with the above hardening. Using strace, it is obvious why it fails:

```
openat(AT_FDCWD, "/proc/726288/mem", O_RDWR) = 72
lseek(72, 94200854600568, SEEK_SET)     = 94200854600568
read(72, "\0\0", 2)                     = 2
lseek(72, 94200854600568, SEEK_SET)     = 94200854600568
write(72, "\1\0", 2)                    = -1 EACCES (Permission denied)
close(72)                               = 0
```

Note that this only will happen for probes where `readelf --notes` indicates a value for the sempahore:

```
  stapsdt              0x00000059       NT_STAPSDT (SystemTap probe descriptors)
    Provider: ruby
    Name: cmethod__entry
    Location: 0x000000000019999d, Base: 0x00000000002d8ec0, Semaphore: 0x000000000052bb54
    Arguments: 8@32(%rsp) 8@40(%rsp) 8@48(%rsp) -4@56(%rsp)
```

As there actually is a sempahore indicated here, this USDT probe would be affected. A similar probe in libc would not be affected and can be attached to as the semaphore is not required for the "enable" mechanism:

```
  stapsdt              0x0000003c       NT_STAPSDT (SystemTap probe descriptors)
    Provider: libc
    Name: memory_heap_free
    Location: 0x000000000019bfd0, Base: 0x00000000001bdd48, Semaphore: 0x0000000000000000
    Arguments: 8@%r11 8@%rax
```

dalehamel:
Fixes #2230 

# Overview

This uses https://elixir.bootlin.com/linux/v4.20.17/source/kernel/events/uprobes.c#L426 to manage semaphore rather than the existing code which seeks to the memory address in userspace, and is based on the interface exposed in https://lore.kernel.org/patchwork/patch/998790/

actual code in kernel to increment semaphore added in https://github.com/torvalds/linux/commit/1cc33161a83d20b5462b1e93f95d3ce6388079ee

See:

https://github.com/torvalds/linux/blob/1cc33161a83d20b5462b1e93f95d3ce6388079ee/kernel/events/uprobes.c#L426-L453

Which can now be used rather than the fseek of /proc/PID/mem from userspace

# API changes

## Non-breaking

A new public function is added to libbpf.h, specifically for attaching to uprobes that are a USDT probe, as the kernel now handles this separately from the existing uprobe enable api, and where USDT probes have had to perform their own semaphore management from userspace. 

## Breaking

The signature for each_uprobe callback adds the ref_ctr_offset. Any code that uses this in the wild will need to update to handle this offset being passed.

# To do

- [x] Test from bpftrace / C api
- [ ] Test from C++ api
- [ ] Test from Python
- [ ] Test from Lua
- [ ] Fix lua foreach callback signature

### DIFF
--- a/src/cc/api/BPF.cc
+++ b/src/cc/api/BPF.cc
@@ -226,7 +226,17 @@ StatusTuple BPF::attach_uprobe(const std::string& binary_path,
                                uint64_t symbol_addr,
                                bpf_probe_attach_type attach_type, pid_t pid,
                                uint64_t symbol_offset) {
+  return attach_uprobe(binary_path, symbol, probe_func, symbol_addr,
+                       attach_type, pid, symbol_offset, 0);
+}
 
+StatusTuple BPF::attach_uprobe(const std::string& binary_path,
+                               const std::string& symbol,
+                               const std::string& probe_func,
+                               uint64_t symbol_addr,
+                               bpf_probe_attach_type attach_type, pid_t pid,
+                               uint64_t symbol_offset,
+                               uint32_t ref_ctr_offset) {
   if (symbol_addr != 0 && symbol_offset != 0)
     return StatusTuple(-1,
              "Attachng uprobe with addr %lx and offset %lx is not supported",
@@ -244,17 +254,25 @@ StatusTuple BPF::attach_uprobe(const std::string& binary_path,
   int probe_fd;
   TRY2(load_func(probe_func, BPF_PROG_TYPE_KPROBE, probe_fd));
 
-  int res_fd = bpf_attach_uprobe(probe_fd, attach_type, probe_event.c_str(),
-                                 binary_path.c_str(), offset, pid);
+  int res_fd;
+  printf("Using ref_ctr_offset %08X\n", ref_ctr_offset);
+  if (ref_ctr_offset != 0) {
+    res_fd =
+        bpf_attach_usdt_probe(probe_fd, attach_type, probe_event.c_str(),
+                              binary_path.c_str(), offset, pid, ref_ctr_offset);
+  } else
+    res_fd = bpf_attach_uprobe(probe_fd, attach_type, probe_event.c_str(),
+                               binary_path.c_str(), offset, pid);
 
   if (res_fd < 0) {
     TRY2(unload_func(probe_func));
     return StatusTuple(
         -1,
         "Unable to attach %suprobe for binary %s symbol %s addr %lx "
-        "offset %lx using %s\n",
+        "offset %lx ref_ctr_offset %lx using %s\n",
         attach_type_debug(attach_type).c_str(), binary_path.c_str(),
-        symbol.c_str(), symbol_addr, symbol_offset, probe_func.c_str());
+        symbol.c_str(), symbol_addr, symbol_offset, ref_ctr_offset,
+        probe_func.c_str());
   }
 
   open_probe_t p = {};
--- a/src/cc/api/BPF.h
+++ b/src/cc/api/BPF.h
@@ -79,6 +79,13 @@ class BPF {
                             bpf_probe_attach_type attach_type = BPF_PROBE_ENTRY,
                             pid_t pid = -1,
                             uint64_t symbol_offset = 0);
+
+  StatusTuple attach_uprobe(const std::string& binary_path,
+                            const std::string& symbol,
+                            const std::string& probe_func, uint64_t symbol_addr,
+                            bpf_probe_attach_type attach_type, pid_t pid,
+                            uint64_t symbol_offset, uint32_t ref_ctr_offset);
+
   StatusTuple detach_uprobe(const std::string& binary_path,
                             const std::string& symbol, uint64_t symbol_addr = 0,
                             bpf_probe_attach_type attach_type = BPF_PROBE_ENTRY,
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -862,8 +862,9 @@ static int bpf_try_perf_event_open_with_probe(const char *name, uint64_t offs,
   // the ref_ctr_offset must be packed in the top 32 bits of the 64 bit storage
   // this supports incrementing the reference counter directly from mm_struct
   // rather than seeking /proc/PID/mem in userspace
-  if (ref_ctr_offset > 0)
+  if (ref_ctr_offset > 0) {
     attr.config |= ref_ctr_offset << PERF_UPROBE_REF_CTR_OFFSET_SHIFT;
+  }
 
   /*
    * struct perf_event_attr in latest perf_event.h has the following
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -91,6 +91,8 @@
 
 #define UNUSED(expr) do { (void)(expr); } while (0)
 
+#define NO_REF_CTR 0
+
 struct bpf_helper {
   char *name;
   char *required_version;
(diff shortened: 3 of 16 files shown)


src/cc/api/BPF.cc
```diff
@@ -218,6 +218,18 @@ StatusTuple BPF::attach_uprobe(const std::string& binary_path,
...
+  return attach_uprobe(binary_path, symbol, probe_func, symbol_addr, attach_type, pid,
+                symbol_offset, 0);
+}
+
+StatusTuple BPF::attach_uprobe(const std::string& binary_path,
+                               const std::string& symbol,
+                               const std::string& probe_func,
+                               uint64_t symbol_addr,
+                               bpf_probe_attach_type attach_type, pid_t pid,
+                               uint64_t symbol_offset,
```
dalehamel: can't use default arguments in this one or it is ambiguous, so to use ref_ctr you have to fully specify all values, even ones with defaults.

src/cc/usdt/usdt.cc
```diff
@@ -119,6 +119,35 @@ bool Probe::add_to_semaphore(int16_t val) {
...
+attaching a uprobe, which negates the need to seek to this memory offset in
+userspace to manage semaphores, as the kernel will do it for us.
+
+This helper function checks if this support is available by reading the
+uprobe format for this value, added in a6ca88b241d5e929e6e60b12ad8cd288f0ffa256
+*/
+static bool ref_ctr_offset_supported() {
+  const char *ref_ctr_pmu_path =
+      "/sys/bus/event_source/devices/uprobe/format/ref_ctr_offset";
+  const char *ref_ctr_pmu_expected = "config:32-63\0";
```
dalehamel: not sure null byte is needed

src/cc/usdt/usdt.cc
```diff
@@ -119,6 +119,35 @@ bool Probe::add_to_semaphore(int16_t val) {
...
+  int fd = open(ref_ctr_pmu_path, O_RDONLY);
+
+  if (fd < 0)
+    return false;
+
+  if (read(fd, ref_ctr_pmu_fmt, sizeof(ref_ctr_pmu_fmt)) < 0)
+    return false;
+
+  if (strncmp(ref_ctr_pmu_expected, ref_ctr_pmu_fmt,
+              strlen(ref_ctr_pmu_expected) == 0))
```
dalehamel: should probably do sizeof instead of strlen and ditch null byte

src/cc/usdt/usdt.cc
```diff
@@ -353,7 +383,8 @@ void Context::each_uprobe(each_uprobe_cb callback) {
 
     for (Location &loc : p->locations_) {
       callback(loc.bin_path_.c_str(), p->attached_to_->c_str(), loc.address_,
-               pid_.value_or(-1));
+               pid_.value_or(
+                   -1));  // FIXME ADD semaphore offset to this tuple needed?
```
dalehamel: i don't think that the ref_ctr nedes to be added to this tuple, is there any reason to?

src/cc/api/BPF.h
```diff
@@ -79,6 +79,13 @@ class BPF {
                             bpf_probe_attach_type attach_type = BPF_PROBE_ENTRY,
                             pid_t pid = -1,
                             uint64_t symbol_offset = 0);
+
+  StatusTuple attach_uprobe(const std::string& binary_path,
+                            const std::string& symbol,
+                            const std::string& probe_func, uint64_t symbol_addr,
+                            bpf_probe_attach_type attach_type, pid_t pid,
```
dalehamel: as mentioned above no default arguments, this version must be fully-specified to avoid ambiguity

src/cc/libbpf.c
```diff
@@ -767,9 +767,12 @@ static int bpf_get_retprobe_bit(const char *event_type)
  * creating [k,u]probe with perf_event_open, which makes it easier to clean up
  * the [k,u]probe. This function tries to create pfd with the perf_kprobe PMU.
  */
+
+#define PERF_UPROBE_REF_CTR_OFFSET_SHIFT 32  // from linux/trace_events.h
```
dalehamel: This header wasn't available on the machine i tested, so I'm hardcoding it for compatibility. It seems unlikely to change, but perhaps we could try and detect it instead and fallback to hardcoded only if needed?

src/cc/libbpf.c
```diff
@@ -782,6 +785,14 @@ static int bpf_try_perf_event_open_with_probe(const char *name, uint64_t offs,
   if (is_return)
     attr.config |= 1 << is_return_bit;
 
+  // Per bcc#2230, a6ca88b introduces a ref_ctr_offset to uprobes to set
+  // semaphores directly through uprobe call, rather than seek to memory
+  // the ref_ctr_offset must be packed in the top 32 bits of the 64 bit storage
+  // this supports incrementing the reference counter directly from mm_struct
+  // rather than seeking /proc/PID/mem in userspace
+  if (ref_ctr_offset > 0)
+    attr.config |= (uint64_t)ref_ctr_offset << PERF_UPROBE_REF_CTR_OFFSET_SHIFT;
```
dalehamel: Since the offset is only 32 bits, I decided to preserve this type information. To avoid overflowing it on the bitshift before it is upcast to 64 bits, I explicitly cast it to a 64 bit container.

src/cc/usdt/usdt.cc
dalehamel: This seems like this used by the python API

src/cc/api/BPF.cc
```diff
@@ -235,17 +245,25 @@ StatusTuple BPF::attach_uprobe(const std::string& binary_path,
   int probe_fd;
   TRY2(load_func(probe_func, BPF_PROG_TYPE_KPROBE, probe_fd));
 
-  int res_fd = bpf_attach_uprobe(probe_fd, attach_type, probe_event.c_str(),
-                                 binary_path.c_str(), offset, pid);
+  int res_fd;
+  printf("Using ref_ctr_offset %08X\n", ref_ctr_offset);
```
dalehamel: debug print don't keep tihs

dalehamel:
As i mention on #2230 i encountered a problem with the Kernel API that I think constitutes a legitimate bug. I believe the code here is mostly correct, but it requires kernel patches to work correctly, as the ref counter increment code in the kernel has different behavior than BCC.

dalehamel:
Closing in favor of #3135 from @danobi 

I'll pull anything else that is needed out of this PR once that merges.

Issue #2352, opened 2019-05-14
String loaded using BPF_probe_read being overridden by constant string in some cases


mmarchini:
LLVM: 6, 7 and 8
BCC: 0.9
Kernel: 4.15.0-47-generic

In some cases, strings loaded with BPF_probe_read are being overridden with constant strings. This seems like an LLVM optimization issue.

The code below works as expeced:

```python
#!/usr/bin/env python

from bcc import BPF
from time import sleep


text = """
#include <linux/utsname.h>
#include <linux/sched.h>

#define NODENAME_LEN __NEW_UTS_LEN


typedef struct Foo {
  char loaded[NODENAME_LEN];
  char constant[10];
} Foo_t;

BPF_HASH(foos, Foo_t);

int bug_example(struct pt_regs *ctx) {
  char constant[] = "mystr";
  struct task_struct *task = (struct task_struct *)bpf_get_current_task();

  char loaded[NODENAME_LEN] = "";
  __builtin_memcpy(loaded, task->nsproxy->uts_ns->name.nodename, NODENAME_LEN);

  Foo_t foo;
  __builtin_memcpy(&foo.loaded, loaded, sizeof(foo.loaded));
  __builtin_memcpy(&foo.constant, constant, sizeof(foo.constant));

  u64 ts = 1;
  foos.update(&foo, &ts);

  return 0;
}
"""

b = BPF(text=text, debug=0)
b.attach_kprobe(event="do_nanosleep", fn_name="bug_example")
sleep(2)
counts = b.get_table("foos")
for k, v in counts.items():
    print("%10d \"%s\" \"%s\"" % (v.value, k.loaded, k.constant))
```

Output:

```
# python bug.py
         1 "laul-mmarchini.corp.netflix.com" "mystr"
         1 "laul-mmarchini" "mystr"
```

But if we swap `loaded` and `constant` in the struct `Foo`, the `loaded` string will be overriden with the `constant` string, leading to a erroneous result.

```diff
@@ -12,8 +12,8 @@
 
 
 typedef struct Foo {
-  char loaded[NODENAME_LEN];
   char constant[10];
+  char loaded[NODENAME_LEN];
 } Foo_t;
 
 BPF_HASH(foos, Foo_t);
```

Output:

```
         1 "mystr" "mystr"
         1 "mystr" "mystr"
         1 "mystr" "mystr"
```

This is the correct BPF program:

```
0: (85) call bpf_get_current_task#35
1: (b7) r6 = 0
[...]

PR #2741, opened 2020-02-06
prevent rewriting for array element type

yonghong-song:
Fix issue #2352.
Corresponding llvm bug: https://bugs.llvm.org/show_bug.cgi?id=41918.

If the element type is array type, the rewriter may generate code like
  addr = ({ typeof(__u8 [16]) _val; __builtin_memset(&_val, 0, sizeof(_val));
            bpf_probe_read(&_val, sizeof(_val), (u64)&daddr->s6_addr); _val; })
for something like
  addr = daddr->s6_addr;
where s6_addr is an array.

The above code has an issue then as addr is pointing to some data
which is out of scope, which meaning compiler is free to use the space.

Let us disable such transformation until we find a good solution.


yonghong-song:
Some tools may need adjustment as well. Let me check them later.

PR #2743, opened 2020-02-07
Fix offwaketime regression introduced by #ac00ac

bubble75:
That seems to be unintentional break. This pull request just restores offwaketime functionality.


yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks for the fix!

PR #2749, opened 2020-02-11
bindsnoop BCC tool

pdubovitsky:
bindsnoop tool traces the kernel function performing IP and IPv6 AF socket binding and
prints socket options set before the system call invocation that might
impact bind behavior and bound interface

# ./bindsnoop.py --help
usage: bindsnoop.py [-h] [-t] [-w] [-p PID] [-P PORT] [-E] [-U] [-u UID]
                  [--count] [--cgroupmap CGROUPMAP]

Trace TCP binds

optional arguments:
  -t, --timestamp       include timestamp on output
  -w, --wide            wide column output (fits IPv6 addresses)
  -p PID, --pid PID     trace this PID only
  -P PORT, --port PORT  comma-separated list of ports to trace.
  -E, --errors          include errors in the output.
  -U, --print-uid       include UID on output
  -u UID, --uid UID     trace this UID only
  --count               count binds per src ip and port
  --cgroupmap CGROUPMAP
                        trace cgroups in this BPF map only

examples:
    ./bindsnoop           # trace all TCP bind()s
    ./bindsnoop -t        # include timestamps
    ./bindsnoop -w        # wider columns (fit IPv6)
    ./bindsnoop -p 181    # only trace PID 181
    ./bindsnoop -P 80     # only trace port 80
    ./bindsnoop -P 80,81  # only trace port 80 and 81
    ./bindsnoop -U        # include UID
    ./bindsnoop -u 1000   # only trace UID 1000
    ./bindsnoop -E        # report bind errors
    ./bindsnoop --count   # count bind per src ip
    ./bindsnoop --cgroupmap mappath  # only trace cgroups in this BPF map

it is reporting socket options set before the bins call
impacting system call behavior:
 SOL_IP     IP_FREEBIND              F....
 SOL_IP     IP_TRANSPARENT           .T...
 SOL_IP     IP_BIND_ADDRESS_NO_PORT  ..N..
 SOL_SOCKET SO_REUSEADDR             ...R.
 SOL_SOCKET SO_REUSEPORT             ....r

 SO_BINDTODEVICE interface is reported as "IF" index


yonghong-song:
[buildbot, test this please]

tools/trace_bind.py
```diff
@@ -0,0 +1,430 @@
...
+examples = """examples:
+    ./trace_bind           # trace all TCP bind()s
+    ./trace_bind -t        # include timestamps
+    ./trace_bind -p 181    # only trace PID 181
+    ./trace_bind -P 80     # only trace port 80
+    ./trace_bind -P 80,81  # only trace port 80 and 81
+    ./trace_bind -U        # include UID
+    ./trace_bind -u 1000   # only trace UID 1000
+    ./trace_bind -E        # report bind errors
+    ./trace_bind -c        # count bind per src ip
```
yonghong-song: I try to reserve "-c" for cgroup path, could you use something different? Maybe use a long form like `--per-src-ip`?

tools/trace_bind.py
```diff
@@ -0,0 +1,430 @@
...
+#include <uapi/linux/ptrace.h>
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wtautological-compare"
+#include <net/sock.h>
+#pragma clang diagnostic pop
+#include <net/inet_sock.h>
+#include <net/net_namespace.h>
+#include <bcc/proto.h>
+
+
```
yonghong-song: extra line?

tools/trace_bind.txt
```diff
@@ -0,0 +1,100 @@
+Demonstrations of trace_bind, the Linux eBPF/bcc version.
```
yonghong-song: This file's name should be trace_bind_example.txt.

brendangregg:
Thanks; inconsistent name, though. I'd call it bindsnoop or just bind.

There's only one tool with an underscore in (mysqld_qslower) because it was a jumble of letters without it (mysqldqslower), but even then I'm still annoyed by it (and its length) and may rename it at some point.

tools/trace_bind.py
pdubovitsky: removing "-c", leaving "--count" option only

tools/trace_bind.txt
pdubovitsky: Renaming to bindsnoop_example.txt

brendangregg:
Default output is >80 chars, and doesn't need to be. "BOUND_IF" can be "IF", and other trimming for a start. Did you read https://github.com/iovisor/bcc/blob/master/CONTRIBUTING-SCRIPTS.md ?

If the IPv6 columns bust the 80 characters, my approach is to use narrower columns by default to fit in 80 chars and let IPv6 overflow, plus provide a -w option for the full columns. E.g.:

```
# ./tcplife -w        # wider colums (fit IPv6)
```

I'm working on another book, and again I have to conform to the publisher's template. If it's >80 characters, I can't include your tool. You can go over 80 in extended modes (-x, -v).

In the tool's comments it refers to itself as "tcpbind": was this a solution to BPF Performance Tools, Chapter 10?:

```
2. Write tcpbind(8): a tool for per-event tracing of TCP bind events.
```

It could be called tcpbind(8) (which is a nicer name) if it were TCP only, but I don't see a filter for TCP. What does it do for other address families (AF_'s)?

Also:

```
# ./bindsnoop.py 
/usr/bin/env: ‘bcc-py’: No such file or directory
```

I'd go with /usr/bin/python like the other tools.

pdubovitsky:
I'll make required changes.
It makes sense to follow these conventions, I missed the doc, sorry.
 
> It could be called tcpbind(8) (which is a nicer name) if it were TCP only, but I don't see a filter for TCP. What does it do for other address families (AF_'s)?

The tool came from a real troubleshooting session, I just made it more generic so it might be useful for others. tcpbind was the original name, but I thought it would be unfair to UDP. I'll rename it back to tcpbind.

The address families should be filtered indirectly, because the entry probes are attached to the inet_bind and inet6_bind events only.

Thank you for the book, it is great!

alban:
Thanks!

Would it be possible to add an [option --cgroupmap](https://github.com/iovisor/bcc/pull/2651/files#diff-fb4170c12eaab4e68fa205e0ff6d6306R39)? Not a blocker for the PR, this could be implemented later.

I am thinking that bindsnoop would be useful for debugging connections to Kubernetes services. I would like to use it in https://github.com/kinvolk/inspektor-gadget/issues/34.

pdubovitsky:
I have committed changes that should address output formatting, renaming the tool to tcpbind, updating smoke test and documentation.
--cgroupmap option was added as well.

alban:
I have not reviewed the code but I wanted to report that tcpbind (and `--cgroupmap`) works for me in my setup. Thanks :)

brendangregg:
Looking better, thanks. What does it do for UDP sockets?

Also, I like to see lines similar to this in the header:

```
# Copyright (c) 2015 Brendan Gregg.
#
# 20-Sep-2015   Brendan Gregg   Created this.
```

Some people prefer not to put their name on code, but if it's no problem, I find the last line in particular very helpful. I've often checked it to see when a tool was done and by who; it's much faster than opening a browser window and navigating through the github history (or switching to the repo source and doing a git log filename.)

pdubovitsky:
@brendangregg  tcpbind would behave the same way for both TCP and UDP, reporting socket options set.
It might be a good idea to add socket protocol reporting.

I have added copyright and history line.

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

tools/tcpbind.py
```diff
@@ -0,0 +1,463 @@
+#!/usr/bin/python
+#
+# tcpbind       Trace IPv4 and IPv6 binds()s.
+#               For Linux, uses BCC, eBPF. Embedded C.
+#
+# based on tcpconnect utility from Brendan Gregg's suite.
+#
+# USAGE: tcpbind [-h] [-t] [-E] [-p PID] [-P PORT [PORT ...]] [-w]
```
yonghong-song: The syntax "-P PORT [PORT ...]" does not match the example, should it be comma separated list?

tools/tcpbind.py
```diff
@@ -0,0 +1,463 @@
...
+if args.wide:
+    header_fmt = "%10s %-12.12s %-2s %-39s %-5s %5s %2s"
+    output_fmt = b"%10d %-12.12s %-2d %-39s %5d %-5s %2d"
+
+if args.ebpf:
+    print(bpf_text)
+    exit()
+
+if args.debug_source:
+    b = BPF(text=bpf_text, debug=DEBUG_SOURCE)
```
yonghong-song: Typically we do not have debug_source in the tools as this goes to assembly level and typically for tool writers. But I am okay with it if you feel you may use it when to modify the tools.

tools/tcpbind.py
```diff
@@ -0,0 +1,463 @@
...
+    global start_ts
+    if args.timestamp:
+        if start_ts == 0:
+            start_ts = event.ts_us
+        printb(b"%-9.6f " % ((float(event.ts_us) - start_ts) / 1000000), nl="")
+    if args.print_uid:
+        printb(b"%6d " % event.uid, nl="")
+    if args.errors:
+        printb(b"%3d " % event.return_code, nl="")
+    printb(output_fmt % (event.pid, event.task, event.ip,
```
yonghong-song: Here we print "ip" which essentially 4 or 6 to indicate ip version. Can we imply this information from the address so we do not "IP"?

tools/tcpbind.py
```diff
@@ -0,0 +1,463 @@
...
+    event = b["ipv4_bind_events"].event(data)
+    global start_ts
+    if args.timestamp:
+        if start_ts == 0:
+            start_ts = event.ts_us
+        printb(b"%-9.6f " % ((float(event.ts_us) - start_ts) / 1000000), nl="")
+    if args.print_uid:
+        printb(b"%6d " % event.uid, nl="")
+    if args.errors:
+        printb(b"%3d " % event.return_code, nl="")
```
yonghong-song: Maybe we could print non-zero return code with its actual names for better usability? For example, for return_code "-99" prints EADDRNOTAVAIL?

tools/tcpbind.py
pdubovitsky: Thank you, it was a good catch, fixed

tools/tcpbind.py
pdubovitsky: I have removed debug-source printing.

tools/tcpbind.py
pdubovitsky: It is a good point. I removed IP protocol version reporting.
I added L4 protocol instead. It should be more useful.

tools/tcpbind.py
pdubovitsky: It would be hard to satisfy 80-characters limitation. 
I added return code reporting in human-readable format when "wide" output mode is enabled (-w)

brendangregg:
Since tcpbind does both TCP and UDP, I'd either filter on TCP, or rename it back to bindsnoop (sorry). I'd foresee this as a common source of confusion, with output like:

```

# ./tcpbind.py -t
TIME(s)        PID COMM         PROT ADDR            PORT   OPTS IF
0.000000   3956801 dig          TCP  ::              49611 .....  0
0.011045   3956822 dig          UDP  ::              56343 .....  0
2.310629   3956498 test_bind_op TCP  192.168.1.102   39609 F...r  0
```
... people will wonder why tcpbind is showing UDP.

pdubovitsky:
It makes sense to name it bindsnoop, renaming.

yonghong-song:
python2 has this error,
```
-bash-4.4$ sudo ./bindsnoop.py 
  File "./bindsnoop.py", line 399
    def proto2str(self, proto: int) -> str:
                             ^
SyntaxError: invalid syntax
-bash-4.4$
```
could you fix it?

tools/bindsnoop.py
```diff
@@ -0,0 +1,516 @@
...
+    }
+
+    // pull in details
+    struct socket *skp_ = *skpp;
+    struct sock *skp = skp_->sk;
+
+    struct inet_sock *sockp = (struct inet_sock *)skp;
+
+    u16 sport;
+    bpf_probe_read(&sport, sizeof(sport), &sockp->inet_sport);
```
yonghong-song: Could you initialize sport say to `sport = 0`? ubuntu 16.04, which has kernel 4.4, will fail with verifier due to this.

pdubovitsky: Added initialization. I'll need to create some testing environment to catch it in the future.

yonghong-song:
[buildbot, test this please]

PR #2750, opened 2020-02-11
Draft of new tool "mtcop" for memcached debugging

dalehamel:
WIP, i still need to clean this up and rebase etc.


Dockerfile.ubuntu
```diff
@@ -1,9 +1,9 @@
-FROM ubuntu:bionic
+FROM ubuntu:eoan
```
dalehamel: Probably shouldn't commit this

PR #2755, opened 2020-02-13
Add libbpf-based tools

anakryiko:
With BPF Compile Once - Run Everywhere (CO-RE), global variables, BPF skeleton, and other BPF features developed over last year, it's now feasible to build equivalents of BCC tools without the need to perform BPF code compilation in runtime and without dependency on installed kernel-devel package. This change adds a first such tool (runqslower), converted from its Python origin to a pure C-based tool, using BPF CO-RE. It also sets up a build flow to simplify adding/converting more such tools easily for people willing to help.

It seems like a good idea to keep such tools close to their progenitors and let users choose what works best for them: BCC-based tools are great for experimentation and quick prototyping and customizing to ad-hoc user needs, but sometimes require a high toll w.r.t. runtime resource usage (Clang/LLVM/Python dependency) and assumptions about kernel headers availability. On the other hand, libbpf- and CO-RE-based tools are small, fast and with almost no extra dependencies, but rely on more recent kernel versions. Having both will help better address user needs and will hopefully grow user community further.

See README.md for some more details on a build setup and conventions.


brendangregg:
Thanks, I'll check it out, and have a go at porting some myself (like execsnoop(8), opensnoop(8), biosnoop(8), biolatency(8), ...).

I like to preserve the history of the tools we use. Could runqslower.c have a comment line at the top to say something like:

```C
// Based on runqslower(8) from BCC by Ivan Babrou.
// 11-Feb-2020  Andrii Nakryiko  Created This.
```

And the man/man8/runqslower.8 file could say something like:

```
.SH AUTHOR
Ivan Babrou, original BCC Python version
Andrii Nakryiko, CO-RE version
```

anakryiko:
Sounds good, will add!

anakryiko:
Added history, updated man page. Added `-t TID` option, backported next->comm fix. Should be identical to Python version.

yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks! This is good enough for folks to start experiments!

Issue #2748, opened 2020-02-11
stackcount: address resolution for user-space stack


rohgarg:
I was playing around with `stackcount` when I noticed this issue: address-to-symbol
resolution doesn't work for user-space stacks without the "-P" flag when tracing
single, isolated processes. I'm not sure if this is the expected usage.

For example, the following command is unable to resolve addresses to symbols.

```bash
# # a.out is a simple hello-world program
# python stackcount -p `pgrep a.out` -s -U c:malloc
Tracing 1 functions for "c:malloc"... Hit Ctrl-C to end.

  7f35f5cef2c0     [unknown]
  7f35f5c791e3     [unknown]
  ...
```

Perhaps the documentation for the "-P" option could be improved to indicate
that it's a required parameter for resolving addresses when tracing single,
isolated processes. While the example provided in `stackcount_example.txt`
does use the "-P" option (`stackcount -P -p 4902 c:malloc`), it's not clear
why the option was used in the example. The current documentation just says:

> -P, --perpid          display stacks separately for each process

Another possible fix could be something along the lines of:

```patch
diff --git i/tools/stackcount.py w/tools/stackcount.py
index a58f9e31..d4284a38 100755
--- i/tools/stackcount.py
+++ w/tools/stackcount.py
@@ -358,7 +358,10 @@ class Tool(object):
                     if self.need_delimiter:
                         print("    --")
                     for addr in user_stack:
-                        self._print_uframe(addr, k.tgid)
+                        if self.args.perpid:
+                          self._print_uframe(addr, k.tgid)
+                        else:
+                          self._print_uframe(addr, self.args.pid)
                     if not self.args.pid and k.tgid != 0xffffffff:
                         self._print_comm(k.name, k.tgid)
                     print("    %d\n" % v.value)
```

For reference, this is the BPF program that `stackcount` generates in this case, without the "-P" option.

```C
#include <uapi/linux/ptrace.h>
#include <linux/sched.h>

struct key_t {
    // no pid (thread ID) so that we do not needlessly split this key
    u32 tgid;
    int kernel_stack_id;
    int user_stack_id;
    char name[TASK_COMM_LEN];
};

BPF_HASH(counts, struct key_t);
BPF_STACK_TRACE(stack_traces, 1024);

int trace_count(void *ctx) {
    struct key_t key = {};
    key.tgid = 0xffffffff; // Note that we explicitly set key.tgid to -1


                    key.user_stack_id = stack_traces.get_stackid(
[...]

PR #2757, opened 2020-02-18
tools/stackcount: Fix address resolution for user-space stack

rohgarg:
This patch fixes issue #2748. The bug was that address-to-symbol resolution
didn't work for user-space stacks without the `-P` (per-pid) flag when tracing
single, isolated processes. The current documentation of the `-P` option
indicates that it's used to "display stacks separately for each process",
and this doesn't match with the intended usage.

This patch has two parts:

 - Fix `tools/stackcount.py` to explicitly set perpid to True if `-p <pid>`
   is used
 - Remove the `-P` option from the example of tracing single, isolated process
   in `tools/stackcount_example.txt`, since the usage of the option can
   be confusing (and unnecessary after the current change)

### DIFF
--- a/tools/stackcount.py
+++ b/tools/stackcount.py
@@ -275,6 +275,12 @@ class Tool(object):
             self.kernel_stack = self.args.kernel_stacks_only
             self.user_stack = self.args.user_stacks_only
 
+        # For tracing single processes in isolation, explicitly set perpid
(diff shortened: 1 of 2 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

PR #2764, opened 2020-02-20
tools: execsnoop add -U (--print-uid) and -u (--uid) flags

bacher09:
Add flags to display UID and filter by UID.

It's the same options used in `opensnoop` and `drsnoop`.

### DIFF
--- a/man/man8/execsnoop.8
+++ b/man/man8/execsnoop.8
@@ -2,8 +2,8 @@
 .SH NAME
 execsnoop \- Trace new processes via exec() syscalls. Uses Linux eBPF/bcc.
 .SH SYNOPSIS
-.B execsnoop [\-h] [\-T] [\-t] [\-x] [\-q] [\-n NAME] [\-l LINE]
-.B           [\-\-max-args MAX_ARGS]  [\-\-cgroupmap MAPPATH]
+.B execsnoop [\-h] [\-T] [\-t] [\-x] [\-\-cgroupmap CGROUPMAP] [\-u USER]
+.B           [\-q] [\-n NAME] [\-l LINE] [\-U] [\-\-max-args MAX_ARGS]
 .SH DESCRIPTION
 execsnoop traces new processes, showing the filename executed and argument
 list.
@@ -28,8 +28,14 @@ Print usage message.
 \-T
 Include a time column (HH:MM:SS).
 .TP
+\-U
+Include UID column.
+.TP
 \-t
 Include a timestamp column.
+.TP
+\-u USER
+Filter by UID (or username)
 .TP
 \-x
 Include failed exec()s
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

yonghong-song:
@bacher09 could you add new option to the example file and the man page?

bacher09:
Sure, I'll update man and example.

Also wanted to ask about UID mapping, currently only `cachetop` is doing this, and I think there might be cases when this might be useful. I wasn't sure about this, since there is many cases when this will not work (docker, containers) and most tools doesn't do this.

man/man8/execsnoop.8
```diff
@@ -2,8 +2,8 @@
 .SH NAME
 execsnoop \- Trace new processes via exec() syscalls. Uses Linux eBPF/bcc.
 .SH SYNOPSIS
-.B execsnoop [\-h] [\-T] [\-t] [\-x] [\-q] [\-n NAME] [\-l LINE]
-.B           [\-\-max-args MAX_ARGS]
+.B execsnoop [\-h] [\-T] [\-t] [\-x] [\-\-cgroupmap CGROUPMAP] [\-u USER]
+.B           [\-q] [\-n NAME] [\-l LINE] [\-\-max-args MAX_ARGS]
```
yonghong-song: missing -U here.

man/man8/execsnoop.8
```diff
@@ -56,6 +62,18 @@ Trace all exec() syscalls, and include timestamps:
...
+#
+.B execsnoop \-U
+.TP
+Trace only UID 1000:
+#
+.B execsnoop \-u 1000
+.TP
+Trace only processes launched by user and display UID column:
+#
+.B execsnoop \-Uu user
```
yonghong-song: Since this is an example, you need specific uid instead of `user`. The `execsnoop -Uu user` won't work.

tools/execsnoop.py
```diff
@@ -24,14 +24,35 @@
...
+
+
 # arguments
 examples = """examples:
     ./execsnoop           # trace all exec() syscalls
     ./execsnoop -x        # include failed exec()s
     ./execsnoop -T        # include time (HH:MM:SS)
+    ./execsnoop -U        # include UID
+    ./execsnoop -u 1000   # only trace UID 1000
+    ./execsnoop -u user   # get user UID and trace only them
```
yonghong-song: You do not need this one `... -u user`.

bacher09: I just wanted to emphasise that both: UID and username could be specified as option there.

man/man8/execsnoop.8
bacher09: Well, as I wrote before, now this options also accepts user as argument too, in that case it will lookup UID in `/etc/passwd` (*or somewhere else, depends on configuration*) using `getpwnam`.

tools/execsnoop.py
bacher09: I replaced `user` with `root`, now it should be obvious that this is user login name.

yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks for explanation. Using `-u USER` and use `-u root` in the example sounds good so people copy-and-paste it just works.

Issue #2169, opened 2019-01-29
http_filter complete example breaking


schrepfler:
When capturing http traffic using the http-parse-complete.py script, the program crashes.
```bash
sudo python http-parse-complete.py  -i lo                         
[sudo] password for srdan: 
binding socket to 'lo'
Traceback (most recent call last):
  File "http-parse-complete.py", line 190, in <module>
    ip_src = int(toHex(ip_src_str),16)
  File "http-parse-complete.py", line 37, in toHex
    hv = hex(ord(ch)).replace('0x', '')
TypeError: ord() expected string of length 1, but int found
```

Running on top of Ubuntu 18.04

PR #2767, opened 2020-02-21
Fix the python part of the http-parse-complete example

matdb:
The example wasn't working on my machine, just like in issue #2169

Aside from indentation and other formatting, the only things I changed were the toHex and printUntilCRLF functions, and using bytes instead of strings in some places.


yonghong-song:
[buildbot, test this please]

PR #2768, opened 2020-02-23
tools/oomkill: output PID instead of thread ID (user's view)

boat0:


yonghong-song:
[buildbot, test this please]

PR #2774, opened 2020-02-25
add man pages as html

brendangregg:
This is an initial commit for https://iovisor.github.io/bcc/ to add man pages as html. We can automate this, and make a listing of man pages (scraped from README.md), in future commits.

FWIW, I generated them from bcc master branch using:

```
tmp=/tmp/man
site=https://iovisor.github.io/bcc
mkdir $tmp
cd man/man8
for f in *.8; do
	fname=${f%.*}
	man2html -M $site $f > $tmp/$fname.html
done
```

### DIFF
--- a/man/man8/argdist.html
+++ b/man/man8/argdist.html
@@ -1,0 +1,257 @@
+Content-type: text/html; charset=UTF-8
+
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">
+<HTML><HEAD><TITLE>Man page of argdist</TITLE>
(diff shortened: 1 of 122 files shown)


brendangregg:
[buildbot, ok to test]

brendangregg:
note (for anyone looking) that this is a commit to gh-pages, not master.

brendangregg:
It works!

https://iovisor.github.io/bcc/man/man8/biolatency.html

We can make an index of these somewhere, but we shouldn't change the path since people will start linking to these.

yonghong-song:
Sounds good. Thanks.

PR #2777, opened 2020-02-28
libbpf-tools: covert BCC drsnoop to BPF CO-RE version

ethercflow:


ethercflow:
Hi @anakryiko, I've made a conversion, would you please help me review it? thank you.

libbpf-tools/trace_helpers.h
```diff
@@ -0,0 +1,14 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+#ifndef __TRACE_HELPER_H
+#define __TRACE_HELPER_H
```
anakryiko: this is a great start, but I think this should ultimately be a small, but very useful library for working with kernel symbols and doing stack symbolization (ideally both kernel and user-space). It could be used by both BCC, as well as directly by applications using only libbpf. It's certainly not in scope for this change, just throwing the idea out. cc @brendangregg

ethercflow:
> Great stuff! Probably needs another iteration, but I'm very excited by these new tools using CO-RE and libbpf!
> 
> Sorry for late reply, I don't normally watch BCC notifications very closely, so next time tag me directly, if you need my attention. Thanks!

@anakryiko Thanks a lot for the comments, I've make some updates except `symbolization` and `time` parts. I agree with you, they should be a small, but very useful library. I haven't thought about this yet. I'll think about them after work.

libbpf-tools/drsnoop.bpf.c
```diff
@@ -0,0 +1,95 @@
...
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include "drsnoop.h"
+
+#define BPF_F_INDEX_MASK		0xffffffffULL
+#define BPF_F_CURRENT_CPU		BPF_F_INDEX_MASK
+
+const volatile pid_t targ_pid = 0;
+const volatile pid_t targ_tgid = 0;
+const volatile __u64 targ_extended = 0;
```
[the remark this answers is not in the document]
ethercflow: @anakryiko  declare it as `u64` will get `error: unknown type name 'u64'`. Is this a bug?  Currently I declare it as `const __u64 * const volatile vm_zone_stat_kaddr = 0;` instead.

libbpf-tools/Makefile
```diff
@@ -57,6 +57,10 @@ $(LIBBPF_OBJ): $(wildcard $(LIBBPF_SRC)/*.[ch]) | $(OUTPUT)/libbpf
 		    INCLUDEDIR= LIBDIR= UAPIDIR=			      \
 		    install
 
+$(OUTPUT)/trace_helpers.o: trace_helpers.c | $(OUTPUT)
+	$(call msg,TRACE-HELPERS,$@)
```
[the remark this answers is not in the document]
ethercflow: @anakryiko It seems not covered (tested after remove this rule), I think `$(OUTPUT)/%.o` rule above only cover `app.o`. Am I missing sth?

libbpf-tools/trace_helpers.c
```diff
@@ -0,0 +1,88 @@
...
+	if (!f)
+		return -ENOENT;
+
+	while (fgets(buf, sizeof(buf), f)) {
+		if (sscanf(buf, "%p %c %s", &addr, &symbol, func) != 3)
+			break;
+		if (!addr)
+			continue;
+		syms[i].addr = (long) addr;
+		syms[i].name = strdup(func);
```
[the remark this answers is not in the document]
ethercflow: @anakryiko Thanks a lot, I'm interested with your prototype when you have time. Sorry for late reply, I got some problems last weekend.

libbpf-tools/drsnoop.bpf.c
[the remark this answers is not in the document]
anakryiko: ah, so on user-space side you don't have u64 defined. It is defined as part of vmlinux.h, but that's only for BPF program side. Ok, so never mind, keep using __u64. Pointer declaration got a bit ridiculous, so I won't insist on marking it as a pointer, use your own judgement.

libbpf-tools/trace_helpers.c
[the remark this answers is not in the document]
anakryiko: didn't get to it yet, but shouldn't take too long

libbpf-tools/trace_helpers.h
anakryiko: btw, __TRACE_HELPER(S)_H, just noticed

libbpf-tools/Makefile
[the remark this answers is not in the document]
anakryiko: Alright, I figured out why it didn't work. It's because of extra %.skel.h dependency. We can keep this generic by extracting this dependency and adding it only for $(APPS), like in this diff:
```diff
diff --git a/libbpf-tools/Makefile b/libbpf-tools/Makefile
index cf7f8afe9fe4..393011fe29cf 100644
--- a/libbpf-tools/Makefile
+++ b/libbpf-tools/Makefile
@@ -31,11 +31,13 @@ $(OUTPUT) $(OUTPUT)/libbpf:
        $(call msg,MKDIR,$@)
        $(Q)mkdir -p $@

-$(APPS): %: $(OUTPUT)/%.o $(LIBBPF_OBJ) | $(OUTPUT)
+$(APPS): %: $(OUTPUT)/%.o $(LIBBPF_OBJ) $(OUTPUT)/trace_helpers.o | $(OUTPUT)
        $(call msg,BINARY,$@)
        $(Q)$(CC) $(CFLAGS) $^ -lelf -lz -o $@

-$(OUTPUT)/%.o: %.c $(OUTPUT)/%.skel.h $(wildcard %.h) | $(OUTPUT)
+$(patsubst %,$(OUTPUT)/%.o,$(APPS)): %.o: %.skel.h
+
+$(OUTPUT)/%.o: %.c $(wildcard %.h) | $(OUTPUT)
        $(call msg,CC,$@)
        $(Q)$(CC) $(CFLAGS) $(INCLUDES) -c $(filter %.c,$^) -o $@
```

libbpf-tools/drsnoop.bpf.c
```diff
@@ -0,0 +1,95 @@
...
+	if (targ_tgid && targ_tgid != tgid)
+		return 0;
+	if (targ_pid && targ_pid != pid)
+		return 0;
+
+	piddata.ts = bpf_ktime_get_ns();
+	if (vm_zone_stat_kaddr) {
+		bpf_probe_read(&piddata.nr_free_pages,
+			       sizeof(*vm_zone_stat_kaddr),
+			       (const void*)&vm_zone_stat_kaddr[NR_FREE_PAGES]);
```
anakryiko: nit: you shouldn't need a cast here, does compiler complain without it?

libbpf-tools/drsnoop.bpf.c
```diff
@@ -0,0 +1,95 @@
...
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(u32));
+	__uint(value_size, sizeof(u32));
+} events SEC(".maps");
+
+SEC("tp_btf/mm_vmscan_direct_reclaim_begin")
+int handle__mm_vmscan_direct_reclaim_begin(u64 *ctx)
+{
+	u64 vm_zone_stat[NR_VM_ZONE_STAT_ITEMS];
```
anakryiko: no need for this anymore

libbpf-tools/drsnoop.c
```diff
@@ -0,0 +1,257 @@
...
+		pid = strtol(arg, NULL, 10);
+		if (errno || pid <= 0) {
+			fprintf(stderr, "invalid TID: %s\n", arg);
+			argp_usage(state);
+		}
+		env.tid = pid;
+		break;
+	case 'P':
+		errno = 0;
+		page_cnt = strtol(arg, NULL, 10);
```
anakryiko: check it's power of two:
```
if (page_cnt & (page_cnt - 1))
    fprintf(stderr, "PAGECNT has to be power of 2\n");
```

Alternatively (and probably a bit more user friendly), would be to just find closest power of two number

libbpf-tools/drsnoop.h
```diff
@@ -0,0 +1,15 @@
...
+#ifndef __DRSNOOP_H
+#define __DRSNOOP_H
+
+#define TASK_COMM_LEN 16
+
+struct event {
+	char task[TASK_COMM_LEN];
+	__u64 delta_ns;
+	__u64 nr_reclaimed;
+	pid_t pid;
```
anakryiko: nit: this leaves 4 bytes of padding between pid and nr_free_pages, maybe move pid to the very end, so when new fields are added, they can fill in the padding?

libbpf-tools/runqslower.c
```diff
@@ -14,7 +14,7 @@
 #include "runqslower.h"
 #include "runqslower.skel.h"
 
-struct env {
+static struct env {
```
anakryiko: nice! thanks for fixing, I was thinking about doing it myself :)

libbpf-tools/trace_helpers.h
```diff
@@ -0,0 +1,14 @@
...
+
+struct ksym {
+	long addr;
+	char *name;
+};
+
+int load_kallsyms(void);
+struct ksym *ksym_search(long key);
```
anakryiko: let's agree on this API and then, once your PR is merged, I'll go and optimize internals a bit. It will be easier to coordinate (and test for me) that way. API that I think should be good is to avoid global state, so how about this:

```c
struct ksyms {
    struct ksym syms*;
    int sym_cnt;
};

struct ksyms *ksyms__load(void);
void ksyms__free(struct ksyms *ksyms);
const struct ksym *ksyms__map_addr(const struct ksyms *ksyms, long addr);
const struct ksym *ksyms__get_symbol(const struct ksyms *ksyms, const char *name);
```

Application then will have to:

```c
struct ksyms *syms = ksyms__load();
if (!syms) /* error */

struct ksym *sym = ksyms__map_addr(syms, 123);
if (!sym) /* not found */

/* once done, free ksyms cache */
ksysms__free(syms);
```

Does this API make sense?

brendangregg:
Nice example!

I don't like these:

```
	{ "pagecnt", 'P', "PAGECNT", 0, "Perf buffer size (default 64)" },
	{ "timeout", 'T', "TIMEOUT", 0, "Timeout of perf buffer poll in ms"
					" (default 100)" },
```

I'd remove them and make them #define's in the C code, with reasonable defaults, otherwise people will start adding them to all tools, cluttering up the options and making them harder to learn.

What should be reasonable? With a poll timeout of 100ms, and given this is a low-frequency activity, then 1 page is likely sufficient (although it may not accept a value that small). You could make it 4 pages to be safe. Does that make sense?

brendangregg:
I just did some testing on opensnoop; I'm using:

```
#define PERF_BUFFER_PAGES	16
#define PERF_POLL_TIMEOUT_MS	100
```

Since I can control the rate of open(2)s easily, I set it to 100k/sec and changed the pages size while writing output to a file, to see at what point it started/stopped losing events:

```
# ./opensnoop > out.out
Lost 6 events on CPU #15!
Lost 10 events on CPU #13!
Lost 23 events on CPU #15!
Lost 34 events on CPU #14!
Lost 5 events on CPU #15!
Lost 39 events on CPU #14!
```

(that also means changing the Lost message to be fprintf on stderr.)

It lost events with page sizes 1 - 4, and stopped losing them at 8 and beyond. So that's a yardstick: 8 pages with a 100ms timeout can handle 100k events/sec on my system (and event output struct size). I'll be setting opensnoop to 16 or 32 pages to be safe.

What I can't control are perturbations: if there's some system activity that drives all CPUs to 100% utilization or a short interval, then the user-space opensnoop tool isn't scheduled quickly enough, and gets many lost events.

There's other tunables too, like BPF map sizes, and wherever possible I'll use safe defaults (and not options) so my end users to have to think about them.

anakryiko:
@brendangregg, yeah, I think it's fine to hard-code them and not have them as externally tunable params (but as #defines, as I mentioned on both PRs).

As for 100ms, that one is not relevant at all, it's just a timeout when there are no events at all. Which mostly determines the granularity of timeout. E.g., for opensnoop, if DURATION is 5 seconds, with 100ms perf timeout you'll terminate somewhere between 5 and 5.1 seconds (plus whatever scheduling delays there are in kernel). But if you bump that timeout to 500ms, then it can be anywhere between 5 and 5.5. But if there is data in perf buffer, perf_buffer__poll will return immediately as new data becomes available. For Ctrl-C-terminating cases that perf buffer timeout could be -1 for all the purposes, btw.

brendangregg:
@anakryiko ah thanks, that explains why I can handle 100k/sec events with a 100ms timeout: because it's waking up a lot faster than 100ms. (I was doing some sanity checks of the buffer sizes and it wasn't adding up.)

I'm going to try inserting a usleep() call into my poll loop, to reduce the wakeups. I've had to do this before in some tools https://github.com/iovisor/bcc/blob/master/tools/cpuunclaimed.py#L273 . The goal is to wake up at a gentle rate and read the buffer, not as quick as possible.

anakryiko:
@brendangregg, one more optimization that is possible for perf_buffer (but neither BCC nor libbpf provide ability to tune these as part of their API, yet), is to increase `.wakeup_events` and `.sample_period` values when creating perf buffer. Currently BCC and libbpf set them to 1, which causes kernel to send wake up notification to userspace for every single event, which in a lot of cases is a major overkill. It's a difference between being able to handle 1mln events and handling 1*0*mln events (e.g, when using `.wakeup_event=1000` and `.sample_period=1000`) on single CPU. So if your goal is to reduce overhead at the expense of higher latency, consider those. It should be really simple to add necessary options to libbpf API (and I assume similarly for BCC).

libbpf-tools/drsnoop.bpf.c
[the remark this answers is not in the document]
ethercflow: Got it. I want to mark it as a pointer, as it is more natural to access it as an array, also as example for someone who need define pointers.

libbpf-tools/drsnoop.bpf.c
ethercflow: No compiler complain without it. It's my fault.

libbpf-tools/drsnoop.bpf.c
ethercflow: Oh, sorry for this.

libbpf-tools/drsnoop.h
ethercflow: Yes, it's my fault. Only change `nr_reclaimed` from `__u32` to `__64`,  forgot to check padding things.

brendangregg:
Ah right... So I think I can see it all in vmstat(8).

opensnoop, with a 100k/sec open(2) workload

```
(root) ~ # vmstat -S M 1
procs -----------memory---------- ---swap-- -----io---- -system-- ------cpu-----
 r  b   swpd   free   buff  cache   si   so    bi    bo   in   cs us sy id wa st
 2  0      0  13942     98  16649    0    0     3     2    5    7  0  0 99  0  0
 3  0      0  13939     98  16653    0    0     0    72 125419 80412 11  4 85  0  0
 3  0      0  13934     98  16656    0    0     0     8 124271 79369 11  4 85  0  0
 2  0      0  13929     98  16659    0    0     0     8 124615 79590 12  4 85  0  0
```

now with a usleep(10000) in the polling loop:

```
(root) ~ # vmstat -S M 1
procs -----------memory---------- ---swap-- -----io---- -system-- ------cpu-----
 r  b   swpd   free   buff  cache   si   so    bi    bo   in   cs us sy id wa st
 2  0      0  13951     98  16655    0    0     3     2    7    0  0  0 99  0  0
 3  0      0  13948     98  16659    0    0     0     8 96772  658 12  2 86  0  0
 2  0      0  13943     98  16663    0    0     0     8 96587  601 12  2 86  0  0
 3  0      0  13939     98  16667    0    0     0    76 96750  595 13  2 86  0  0
 2  0      0  13936     98  16670    0    0     0     8 96740  618 13  1 86  0  0
```

So context-switches has gone from 80k/sec to <1k/sec. Interrupts have dropped from 124k/sec to 96k/sec -- I assume it needs setting the sample_period to bring them down further.

Anyway, given the massive drop in context switches, I'll be included the usleep(). I had to do this with cpuunclaimed.py as well. My main look now looks like:

```C
	while (1) {
		usleep(10000);
		if ((err = perf_buffer__poll(pb, PERF_POLL_TIMEOUT_MS)) < 0)
			break;
[...]
```

Such a sleep may not be necessary for very low frequency tools like drsnoop(8).

libbpf-tools/trace_helpers.h
ethercflow: Agree, thanks.

ethercflow:
@brendangregg Thanks a lot for sharing your experience , I'll change them to `#defines` with reasonable defaults.

brendangregg:
I'm still testing, but this is how my code currently reads:

```
/* Tune the buffer size and wakeup rate. These settings cope with roughly
 * 50k opens/sec.
 */
#define PERF_BUFFER_PAGES	64
#define PERF_BUFFER_TIME_MS	10
#define PERF_POLL_TIMEOUT_MS	100
```

drsnoop(8) probably just needs a smaller page size, since its event rate will be low, and doesn't need the usleep().

libbpf-tools/trace_helpers.h
ethercflow: Hi @anakryiko , I've implemented the API and updated the PR, but changed a little, looks like:

- The `trace_helpers.h`

``` c

struct ksym {
 	long addr;
 	char name[0];
};

typedef struct ksyms ksyms_t;

ksyms_t *ksyms__load(void);
void ksyms__free(ksyms_t *ksyms);
const struct ksym *ksyms__map_addr(const ksyms_t *ksyms, long addr);
const struct ksym *ksyms__get_symbol(const ksyms_t *ksyms, const char *name);
```

I ran `cat /proc/kallsyms| wc -l` on` kernel 5.6 rc2` got `144455`. So I think maybe we need to use `rbtree` to get efficient query efficiency, right?

So I defined 

- The `trace_helpers.c` with:

``` c
struct ksym_rb_nodes {
        struct rb_node addr_rb_node;
        struct rb_node name_rb_node;
        struct ksym ksym;
};

struct ksyms {
        struct rb_root addr_entries;
        struct rb_root name_entries;
};
```

The steps are:
1. Read `/proc/kallsyms` line by line and create `addrs_entries` rbtree;
2. After `adds_rb_tree` has completed, use it to gen `names_entries` entree;

So `ksyms__map_addr` visits `addrs_entries` and `ksyms__get_symbol` visits `names_entries` to return `ksym`.

It seems different from your thought:

> given this is going to be potentially used by many tools (kallsyms are also useful for kernel stack symbolization), I think it will pay off to do this a bit more efficiently and cleanly:
> 
> ```
> struct kallsyms_cache {
>     size_t nr_syms;
>     struct ksym *syms;
>     char *strs;
> };
> ```
> 
> `strs` in this case would be a contiguous piece of memory containing all the strings that are needed for all read symbols (this will avoid a lot of small memory allocations performed by strdup). Then each ksym's name pointer will just point somewhere into this piece of memory. I can prototype something like this quickly over weekend, probably, if you are interested.

What do you think about it? Does this implement make sense? Thank you.

libbpf-tools/trace_helpers.h
```diff
@@ -0,0 +1,17 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+#ifndef __TRACE_HELPERS_H
+#define __TRACE_HELPERS_H
+
+struct ksym {
+	long addr;
+	char name[0];
```
anakryiko: please no, it's cumbersome API and will prevent optimization I've been talking about. Keep it as `const char *`.

libbpf-tools/trace_helpers.h
```diff
@@ -0,0 +1,17 @@
...
+};
+
+typedef struct ksyms ksyms_t;
```
anakryiko: Let's follow kernel guidelines here and not create unnecessary typedefs. There is nothing wrong with `struct ksyms`.

libbpf-tools/trace_helpers.h
ethercflow: Oh, sorry for the complexity, I'll drop RB tree and  implement my original version but with the API we discussed. Thank you.

libbpf-tools/trace_helpers.c
```diff
@@ -0,0 +1,88 @@
...
+		result = key - syms[mid].addr;
+		if (result < 0)
+			end = mid;
+		else if (result > 0)
+			start = mid + 1;
+		else
+			return &syms[mid];
+	}
+
+	if (start >= 1 && syms[start - 1].addr < key &&
```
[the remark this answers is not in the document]
ethercflow: @anakryiko I think this should keep `< key` as when `syms[start - 1].addr == key`, it will return at line 66 with binary search.

libbpf-tools/trace_helpers.h
ethercflow: @anakryiko I've update PR with the original version with the API we discussed. PTAL. Thank you.

libbpf-tools/trace_helpers.h
ethercflow: Thanks, I'll read  kernel guidelines to learn more about typedefs.

anakryiko:
Looks great! There are few things I'd tweak in ksyms handling, but given I intend to optimize it and change some internals, I don't think it's worth spending more time on it. Once this lands, I'll do changes I talked about and will put up PR for review.

Thanks for doing the CO-RE conversion!

anakryiko:
@brendangregg, do you have any objections to merging this? I'm going to work on a bit more efficient (mostly in terms of memory layout/allocations) implementation for struct ksyms, but API should stay the same. But I'd rather have this merged first so I can use that as a base and testbed.

Unfortunately, I don't have rights to merge anything, so please take a look. cc @yonghong-song  as well

yonghong-song:
[buildbot, test this please]

PR #2778, opened 2020-02-28
libbpf-tools: add CO-RE opensnoop

brendangregg:
This has one thing to improve:

```C
SEC("kprobe/do_sys_open")
int handle__sys_enter_openat(u64 *ctx)
[...]
	args.fname = (const char *)ctx[13]; /* XXX unstable */
	args.flags = (int)ctx[12];
```

How should I be doing this? Here's what I tried:

- `SEC("tp_bpf...`: the syscall tracepoints aren't available.
- `int BPF_KPROBE(handle_do_sys_open, int dfd, const char __user *filename, int flags)`: I don't think the BPF_KPROBE macros are available properly outside of the selftests yet.
- `PT_REGS_PARM2(ctx);`: needs uapi/linux/bpf.h, which is missing outside of the kernel source.

@anakryiko any other ideas? Thanks.


brendangregg:
oh, and the name "handle__sys_enter_openat()" doesn't match the actual SEC(). I'll update it.

libbpf-tools/opensnoop.h
```diff
@@ -0,0 +1,24 @@
...
+
+struct event {
+	/* user terminology for pid: */
+	pid_t pid;
+	__u64 ts;
+	uid_t uid;
+	int ret;
+	char comm[TASK_COMM_LEN];
+	char fname[NAME_MAX];
+	int flags;
```
anakryiko: some reordering will shave off 8 bytes of paddin in struct event:
```
__u64 ts;
pid_t pid;
uid_t uid;
int ret;
int flags;
char comm[TASK_COMM_LEN];
char fname[NAME_MAX];
```

libbpf-tools/opensnoop.c
```diff
@@ -0,0 +1,298 @@
...
+		printf("%-6s ", "UID");
+	printf("%-6s %-16s %3s %3s ", "PID", "COMM", "FD", "ERR");
+	if (env.extended)
+		printf("%-8s ", "FLAGS");
+	printf("%s\n", "PATH");
+
+	/* setup event callbacks */
+	pb_opts.sample_cb = handle_event;
+	pb_opts.lost_cb = handle_lost_events;
+	pb = perf_buffer__new(bpf_map__fd(obj->maps.events), 64, &pb_opts);
```
anakryiko: might be good idea to have perf_buf size as a constant or event command line arg (I've been lazy with this in runqslower, guilty!)

libbpf-tools/opensnoop.c
```diff
@@ -0,0 +1,298 @@
...
+		pb = NULL;
+		fprintf(stderr, "failed to open perf buffer: %d\n", err);
+		goto cleanup;
+	}
+
+	/* setup duration */
+	if (env.duration) {
+		gettimeofday(&time_now, NULL);
+		time_duration.tv_sec = env.duration;
+		timeradd(&time_now, &time_duration, &time_end);
```
anakryiko: nit: `clock_gettime(CLOCK_MONOTONIC, &value)` might be a better way to measure passage of time?..

libbpf-tools/opensnoop.c
```diff
@@ -0,0 +1,298 @@
...
+#include <string.h>
+#include <sys/resource.h>
+#include <sys/time.h>
+#include <time.h>
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+#include "opensnoop.h"
+#include "opensnoop.skel.h"
+
+struct env {
```
anakryiko: static?

libbpf-tools/opensnoop.c
```diff
@@ -0,0 +1,298 @@
...
+	int duration;
+	bool verbose;
+	bool timestamp;
+	bool print_uid;
+	bool extended;
+	bool failed;
+	char *name;
+} env = {
+	.duration = 0,
+};
```
anakryiko: no need for .duration initialization, just `= {}` should be enough.

libbpf-tools/opensnoop.bpf.c
```diff
@@ -0,0 +1,93 @@
...
+// Copyright (c) 2019 Facebook
+// Copyright (c) 2020 Netflix
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include "opensnoop.h"
+
+#define TASK_RUNNING 0
+
+#define BPF_F_INDEX_MASK		0xffffffffULL
+#define BPF_F_CURRENT_CPU		BPF_F_INDEX_MASK
```
anakryiko: this shouldn't be needed as well, but requires a bunch of work in BTF support for enum, so it has to be copy/pasted for now. It's on my TODO list as well...

brendangregg:
Thanks for the comments, I'll try switching it to `SEC("tracepoint/` (although that means tracing all the open(2) variants, but no big deal.) For other tools we will need the BPF_KPROBE stuff for it to be sane (and/or PT_REGS_PARM).

anakryiko:
btw, this PT_REGS_PARMx macros are available in `#include <bpf/bpf_tracing.h>` header, but it's expecting __KERNEL__ defined to work for BPF case. I'm going to make it work without extra define as well.

anakryiko:
https://patchwork.ozlabs.org/project/netdev/list/?series=161712&state=* adds BPF_KPROBE/BPF_KRETPROBE to libbpf's bpf_tracing.h, in addition to existing PT_REGS_PARM macros. Once this lands, I'll also update BCC's copy of bpftool, which will generate vmlinux.h with a proper header guard, making PT_REGS macro work better and not require explicit __KERNEL__ symbol to be defined.

brendangregg:
I just pushed a commit with the changes (all worked fine, thanks.) We can squash-merge the commit; it's just handy for checking the changes at the moment.

I can fetch net-next with the BPF_KPROBE changes and do another tool to try out kprobes.

As my colleague Jason pointed out, the memory footprint of opensnoop as CO-RE is much lower than opensnoop.py. 9 Mbytes for CO-RE vs 80 Mbytes for Python.

anakryiko:
Looks great!

re: BPF_KPROBE, I already synced that to https://github.com/libbpf/libbpf/commits/master, feel free to update submodule and you should get those from `#include <bpf/bpf_tracing.h>`

I think I need to go and update checked in version of bpftool, it will generate vmlinux.h with proper header guard (__VMLINUX_H__), will post PR in a second.

> As my colleague Jason pointed out, the memory footprint of opensnoop as CO-RE is much lower than opensnoop.py. 9 Mbytes for CO-RE vs 80 Mbytes for Python.

great data point, thanks for sharing!

anakryiko:
https://github.com/iovisor/bcc/pull/2778 will bring in latest bpftool

brendangregg:
I'll merge; a different example will use BPF_KPROBE.

PR #2784, opened 2020-03-02
Fix the missing 'device' argument to bcc_func_load

chaitu-tk:
commit d147588ebe35b7cd2b4d253a7da18bef253ea78d ("Support for hardware
offload (#2502)") added a device_name arg to bcc_fund_load() but ddn't
update one of the caller, this causes a egfault in libc with below
trace

```
(gdb) bt full
#0  strnlen () at ../sysdeps/x86_64/strlen.S:103
No locals.
#1  0x00007f82c1ce053e in __strncpy_sse2 (s1=0x7ffcfe71e690 "", s2=0xa <error: Cannot access memory at address 0xa>, n=16) at ./strncpy.c:29
        size = <optimised out>
#2  0x00007f82c1d62a16 in __GI___if_nametoindex (ifname=0xa <error: Cannot access memory at address 0xa>) at ../sysdeps/unix/sysv/linux/if_index.c:46
        ifr = {ifr_ifrn = {ifrn_name = "\000\351\243\000\000\000\000\000\260.?\302\202\177\000"}, ifr_ifru = {ifru_addr = {sa_family = 13008, sa_data = "?\302\202\177\000\000\335AY\000\000\000\000"},
            ifru_dstaddr = {sa_family = 13008, sa_data = "?\302\202\177\000\000\335AY\000\000\000\000"}, ifru_broadaddr = {sa_family = 13008, sa_data = "?\302\202\177\000\000\335AY\000\000\000\000"},
            ifru_netmask = {sa_family = 13008, sa_data = "?\302\202\177\000\000\335AY\000\000\000\000"}, ifru_hwaddr = {sa_family = 13008, sa_data = "?\302\202\177\000\000\335AY\000\000\000\000"},
            ifru_flags = 13008, ifru_ivalue = -1036045616, ifru_mtu = -1036045616, ifru_map = {mem_start = 140199581397712, mem_end = 5849565, base_addr = 59168, irq = 113 'q', dma = 254 '\376',
              port = 252 '\374'}, ifru_slave = "\320\062?\302\202\177\000\000\335AY\000\000\000\000", ifru_newname = "\320\062?\302\202\177\000\000\335AY\000\000\000\000",
            ifru_data = 0x7f82c23f32d0 "\001"}}
        fd = 5
#3  0x00007f82bd95abcb in ebpf::BPFModule::bcc_func_load(int, char const*, bpf_insn const*, int, char const*, unsigned int, int, char*, unsigned int, char const*) ()
   from /usr/lib/x86_64-linux-gnu/libbcc.so.0
```

### DIFF
--- a/src/cc/api/BPF.cc
+++ b/src/cc/api/BPF.cc
@@ -625,7 +625,7 @@ StatusTuple BPF::load_func(const std::string& func_name, bpf_prog_type type,
   fd = bpf_module_->bcc_func_load(type, func_name.c_str(),
                      reinterpret_cast<struct bpf_insn*>(func_start), func_size,
                      bpf_module_->license(), bpf_module_->kern_version(),
-                     log_level, nullptr, 0);
+                     log_level, nullptr, 0, nullptr);
 
   if (fd < 0)
     return StatusTuple(-1, "Failed to load %s: %d", func_name.c_str(), fd);


chaitu-tk:
Sorry, missed the default value in `bpf_module.h` , but still, need to investigate why it segfaults. Please note that I haven't been able to test this fix as due to a [bug](https://gitmemory.com/issue/llvm/llvm-project/120/584646755) in clang-11 the build fails.

yonghong-song:
We have signature like
```
  int bcc_func_load(int prog_type, const char *name,
                    const struct bpf_insn *insns, int prog_len,
                    int log_level, char *log_buf, unsigned log_buf_size,
                    const char *dev_name = nullptr);
```

so if the caller did not specify `dev_name`. it will be nullptr. Did I miss anything?

chaitu-tk:
> We have signature like
> 
> ```
>   int bcc_func_load(int prog_type, const char *name,
>                     const struct bpf_insn *insns, int prog_len,
>                     const char *license, unsigned kern_version,
>                     int log_level, char *log_buf, unsigned log_buf_size,
>                     const char *dev_name = nullptr);
> ```
> 
> so if the caller did not specify `dev_name`. it will be nullptr. Did I miss anything?

Yes, I realized that immediately after posting see my latest comment

> Sorry, missed the default value in bpf_module.h

chaitu-tk:
But the thing is with 0.12.0 I consistently hit the segfault and unable to move to `HEAD` due to clang linking issue, I don't want compile llvm/clang from source with the linking fix, but probably that's the only thing left to do. (Tried 

Any pointers on debugging this segfault? IMHO the device passed looks like Junk.

yonghong-song:
It would be good to verify whether your change indeed fixed the issue or not. Your suspicion is right based on core dump stack. But I do not understand why it happens. Add some print to dump the pointer value of dev_name?

chaitu-tk:
> Add some print to dump the pointer value of dev_name?

Currently, struggling with build issues to build and install my custom bcc (either with older llvm like 6.0 or latest 11), once I resolve the build then I can give it a shot.

Side note, with latest LLVM I had to write a `FindZLIB.cmake` to resolve `-l ZLIB::ZLIB` to `-lz`, will check again and submit a patch.

yonghong-song:
> Side note, with latest LLVM I had to write a FindZLIB.cmake to resolve -l ZLIB::ZLIB to -lz, will check again and submit a patch.

Thanks. I did not do thorough study and just added -DLLVM_ENABLE_ZLIB=OFF in my llvm cmake command to work around the issue. Your approach to change FindZLIB.cmake should be a right solution.

chaitu-tk:
Instead of building, I have manually modified `/usr/lib/python3/dist-packages/bcc/__init__.py` to pass None for the device in `lib.bcc_func_load` and segfault is resolved (revert and segfault appears).  

I suspect `ctypes` is not handling the default value of arguments for a function well and ends up passing random value to it, causing undefined behavior. I will dig deep in `ctypes` else in the worst case we might have to explicitly pass None from all callers.

chaitu-tk:
Ah, found the root cause, the python packages are `0.10` (latest for ubuntu 16.04) but the libraries (`libbcc_bpf.so` and `libbcc.so` are `0.12`), this mismatch is causing the issue, and that is why explicitly passing None for device works.

Thanks for the help and Sorry for the noise, closing this.

PR #2787, opened 2020-03-03
usdt.py: print errors to stderr in enable_probe_or_bail()

gfx:
This is a proposal to change the behavior of `USDT#enable_probe_or_bail()`: print error messages to `stderr`, instead of `stdout`. I think this PR is the expected behavior, but because it breaks backward compatibility, feel free to reject it.


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

PR #2790, opened 2020-03-04
usdt: fix parsing sp register in arguments on AArch64

nick-arm:
One of the USDT probes for OpenJDK on AArch64 has an argument as an
offset from the stack pointer register like "8@[sp, 112]". This causes
the argument parser to fail:

  Parse error:
      8@x22 8@x20 8@x23 8@x0 8@x26 8@x27 8@[sp, 112] 8@[sp, 120]
  ------------------------------------------^

The error message then repeats forever.

Changed ArgumentParser_aarch64::parse_register so it accepts either
"xNN" or "sp" and outputs the register name rather than the register
number. The stack pointer is in a separate field `sp` in `struct pt_regs`
rather than in the `regs[]` array. See:

https://elixir.bootlin.com/linux/latest/source/arch/arm64/include/asm/ptrace.h#L160

Note that the parser currently accepts "x31" and converts that into a
reference to `regs[31]` but that array only has 31 elements. Made x31 an
alias for `sp` to avoid undefined behaviour from reading past the end of
the array.

Change-Id: I88b6ff741914b5d06ad5798a55bd21ea03f69825
Signed-off-by: Nick Gasson <nick.gasson@arm.com>

### DIFF
--- a/src/cc/usdt.h
+++ b/src/cc/usdt.h
@@ -102,6 +102,8 @@ class ArgumentParser {
   }
   bool error_return(ssize_t error_start, ssize_t skip_start) {
     print_error(error_start);
+    if (isspace(arg_[skip_start]))
+        skip_start++;  // Make sure we skip at least one character
     skip_until_whitespace_from(skip_start);
     return false;
   }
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

PR #2791, opened 2020-03-04
usdt.cc: make bcc_usdt_enable_probe() to accept a fully-specified probe name

gfx:
This is inevitable if multiple providers have the same probe name and we'd like to enable one of it.

There could be another way, for example defining a Python method to use `bcc_usdt_enable_fully_specified_probe()`, but because `enable_probe` is not performance-sensitive, I prefer the way this PR does.

What do you think of it?


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

gfx:
Hmm, this might be confusing for those who uses C API (e.g. iovisor/bpftrace developers).

Probably I change the PR to use `bcc_usdt_enable_fully_specified_probe` from Python, keeping the C API as is.

gfx:
Close this PR because I've re-created the patch to do this. #2799

PR #2795, opened 2020-03-06
table.py: print error messages to stderr and exit(1) instead of exit(0) for fatal errors

gfx:
I believe this is more helpful for those who develop BPF tools with the Python binding.


yonghong-song:
[buildbot, test this please]

PR #2797, opened 2020-03-06
GitHub Actions: optionally publish container image

alban:
This is optional: this GitHub Action will only attempt to publish the
container image if the GitHub repository has been configured with GitHub
secrets (e.g. https://github.com/iovisor/bcc/settings/secrets).

The GitHub secrets can be configured as follows:
- DOCKER_PUBLISH = 1
- DOCKER_IMAGE = docker.io/myorg/bcc
- DOCKER_USERNAME = username
- DOCKER_PASSWORD = password

This is intended to make it easy for anyone to fork iovisor/bcc on
GitHub and publish custom container images with bcc.

-----

Open questions:
- I see that the built image is big: 1.3GB. Are there other attempts to build bcc container images with a smaller size?
- Dockerfile.ubuntu keeps the sources and the .deb files but didn't install them. Is it currently used somewhere? I couldn't find it in the bcc repository.

### DIFF
--- a/.github/workflows/bcc-test.yml
+++ b/.github/workflows/bcc-test.yml
@@ -92,3 +92,38 @@ jobs:
 #     https://github.com/marketplace/actions/debugging-with-tmate
 #    - name: Setup tmate session
 #      uses: mxschmitt/action-tmate@v1
+
+  # Optionally publish container images, guarded by the GitHub secret
+  # DOCKER_PUBLISH.
+  # GitHub secrets can be configured as follows:
+  #   - DOCKER_PUBLISH = 1
+  #   - DOCKER_IMAGE = docker.io/myorg/bcc
+  #   - DOCKER_USERNAME = username
+  #   - DOCKER_PASSWORD = password
+  publish:
+    name: Publish
+    runs-on: ubuntu-latest
+    steps:
+
+    - uses: actions/checkout@v1
+
+    - name: Initialize workflow variables
+      id: vars
+      shell: bash
+      run: |
+          echo ::set-output name=DOCKER_PUBLISH::${DOCKER_PUBLISH}
+      env:
+        DOCKER_PUBLISH: "${{ secrets.DOCKER_PUBLISH }}"
+
+    - name: Build container image and publish to registry
+      id: publish-registry
+      uses: elgohr/Publish-Docker-Github-Action@2.8
+      if: ${{ steps.vars.outputs.DOCKER_PUBLISH }}
+      with:
+        name: ${{ secrets.DOCKER_IMAGE }}
+        username: ${{ secrets.DOCKER_USERNAME }}
+        password: ${{ secrets.DOCKER_PASSWORD }}
+        workdir: .
+        dockerfile: Dockerfile.ubuntu
+        snapshot: true
+        cache: ${{ github.event_name != 'schedule' }}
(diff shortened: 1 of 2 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

yonghong-song:
>    Dockerfile.ubuntu keeps the sources and the .deb files but didn't install them. Is it currently used somewhere? I couldn't find it in the bcc repository.

Probably used by some bcc users. The scripts in the repository are mostly for sharing how to do it. People may customize for their own purpose.

yonghong-song:
Let us merge this. We can tweak it later if needed.

dalehamel:
@yonghong-song I just saw this PR I think we should make how we publish docker images consistent with bpftrace, which is done with https://github.com/iovisor/bpftrace/blob/master/docker/scripts/push.sh

We would need someone with admin access to this repo to add the QUAY_TOKEN secret

This has been very helpful for bpftrace, i've used it for bisecting issues a few times already.

alban:
@dalehamel I am not sure to understand. Do you mean we should enable the GitHub action on this repo (github.com/iovisor/bcc), so it gets built for each commit? As far as I know, it is not enabled. I enabled it in a fork (github.com/kinvolk/bcc) but the build gets done only when I manually sync the fork.

alban:
I did that PR to avoid bcc building logic in Inspektor Gadget:

https://github.com/kinvolk/inspektor-gadget/pull/40/commits/b03c81dba6f94f37901acbbec6ae86cec30fba19#diff-6608fa935740952c262ae1146b372796R13

dalehamel:
@alban yes, like what is done for bpftrace: https://github.com/iovisor/bpftrace/actions/runs/53224984

and then like you've done for inspektor-gadget, we do for kubectl-trace: https://github.com/iovisor/kubectl-trace/blob/master/build/Dockerfile.tracerunner#L2

alban:
I don't have special attachment to `elgohr/Publish-Docker-Github-Action@2.8` versus your script `push.sh`. Is `push.sh` more flexible?

dalehamel:
@alban https://github.com/iovisor/bpftrace/blob/master/INSTALL.md#docker-images details the tags that it currently pushes. bpftrace is a little different because it is producing an executable, where as bcc will probably produce a debfile or bundle of debian packages.

I would like if it pushed the same tags as here, and ideally user just has to set QUAY_TOKEN. It should be easy to push the same tags to dockerhub too, the auth is separated in a different file.

iovisor org is already pushing kubectl-trace and bpftrace images to quay, pushing bcc would make sense to me. There could be two use cases for this:

- A bcc toolchain fully installed, with all tools already on path meant for things like your tool and kubectl-trace to consume. We have a custom toolbox image at my work, inheriting from this image is a use case for that.
- A developer image, which includes the deb file built for that tag. This would allow for projects like bpftrace to very easily target specific bcc images (as a side note, the iovisor apt repo seems out of date, so this is also a work around for distributing bcc images)

I'm not particularly attached to this push.sh script and think it can be extended further, but it would be nice to have it in a script like this that is in-repo. I love the idea of having bcc and other iovisor tools distributed as a docker images, so thanks for getting the ball rolling on this!

As I side note, i think that the github actions could be extended further should build and store the debfile as an artifact for various debian/ubuntu versions, and ideally publish these artifacts on release. This would be consistent with bpftrace, and has been really useful for bisecting to debug issues between releases.

PR #2799, opened 2020-03-07
python binding: allow fully-specified probe names for USDT#enable_probe()

gfx:
`USDT#enable_probe()` cannot enable probes if its basename is not unique, so providing a way to use fully-specified names (e.g. "provider:probe") is inevitable for some cases.

At first, I have created a path as https://github.com/iovisor/bcc/pull/2791, but I prefer this style to keep C API tidy. Sorry for annoying.

### DIFF
--- a/src/cc/bcc_usdt.h
+++ b/src/cc/bcc_usdt.h
@@ -77,6 +77,9 @@ const char *bcc_usdt_genargs(void **ctx_array, int len);
 const char *bcc_usdt_get_probe_argctype(
   void *ctx, const char* probe_name, const int arg_index
 );
+const char *bcc_usdt_get_fully_specified_probe_argctype(
+  void *ctx, const char* provider_name, const char* probe_name, const int arg_index
+);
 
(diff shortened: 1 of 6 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
@gfx The change looks good. What kind of interface we should use in python tools? Currently we use the format `u:<binary>:<probe>` in trace.py.  Maybe we can use `u:<binary>:[<provider>:]<probe>`? Typically, provider is not needed. But the above option will enable it if needed.

It would be good if your patch includes at least one tool which supports the new format.

gfx:
@yonghong-song 

Thanks for the review. It makes sense. Support an optional provider name for "u" type of probe specs in trace.py: d21f8208a81fbefaf5953e7fc57ef806d1eba67a

The commit has also added a C API `bcc_usdt_get_fully_specified_probe_argctype()` to implement this feature.

yonghong-song:
[buildbot, test this please]

yonghong-song:
LGTM. Thanks!

gfx:
Thank you!

PR #2800, opened 2020-03-08
Make -DCMAKE_INSTALL_PREFIX=/usr as the default

gfx:
It is a typical mistake to mix adding "-DCMAKE_INSTALL_PREFIX=/usr " and not when we develop BCC.

Let's set the right default and forget the option.


yonghong-song:
[buildbot, test this please]

yonghong-song:
This makes sense as /usr probably is a sensible and most people just use it. Further people can always provide their own install prefix if needed.

PR #2801, opened 2020-03-08
tools: add option --cgroupmap and --unique to capable.py

alban:
`--cgroupmap` is the same option already introduced in other tools (execsnoop, opensnoop, and others)

Some processes can do a lot of security capability checks, generating a lot of ouput. In this case, the --unique option is useful to only print once the same set of capability, pid (or cgroup if --cgroupmap is used) and kernel/user stacks (if -K or -U are used).
```
  # ./capable.py -K -U --unique
```

Documentation (man page and example text) updated.


yonghong-song:
[buildbot, test this please]

PR #2802, opened 2020-03-08
Reduce container image size with multi-stage builds

alban:
Before this patch, Dockerfile.ubuntu was generating an oversized image.
```
                       Uncompressed size      Compressed size
Before this patch:     1.3GB                  390MB
After this patch:      250MB                  110MB
```
-----
Follow up to https://github.com/iovisor/bcc/pull/2797


yonghong-song:
[buildbot, test this please]

PR #2804, opened 2020-03-09
softirqs: Combined CPU as part of the key is necessary to avoid amiss…

netedwardwu:
… value.

In the environment of massive software interrupts.
<idle>-0     [003] ..s1   106.421020: softirq_entry: vec=6 [action=TASKLET]
<idle>-0     [000] ..s1   106.421063: softirq_entry: vec=3 [action=NET_RX]
<idle>-0     [003] ..s1   106.421083: softirq_exit: vec=6 [action=TASKLET]

Follow the above ftrace logs, we know the correct vec-6 start timestamp is replaced with incorrect vec-3.
Because PID is idle-0. It will produce the wrong result after calculating delta.

### DIFF
--- a/tools/softirqs.py
+++ b/tools/softirqs.py
@@ -54,6 +54,11 @@ debug = 0
 bpf_text = """
 #include <uapi/linux/ptrace.h>
 
+typedef struct entry_key {
+    u32 pid;
+    u32 cpu;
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

PR #2806, opened 2020-03-09
libbpf-tools: add CO-RE xfsslower

ethercflow:
- Parse `fsync`'s offset to `start` and size to `end - start` instead of zero; 

### DIFF
--- a/libbpf-tools/xfsslower.bpf.c
+++ b/libbpf-tools/xfsslower.bpf.c
@@ -1,0 +1,160 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2020 Wenbo Zhang
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_tracing.h>
+#include "xfsslower.h"
+
+#define NULL    0
+
+const volatile pid_t targ_tgid = 0;
+const volatile __u64 min_lat = 0;
+
+struct piddata {
+	u64 ts;
+	loff_t start;
+	loff_t end;
+	struct file *fp;
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 8192);
+	__type(key, u32);
+	__type(value, struct piddata);
+} start SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(u32));
+	__uint(value_size, sizeof(u32));
+} events SEC(".maps");
+
+static __always_inline int
+probe_entry(struct file *fp, loff_t s, loff_t e)
+{
+	u64 id = bpf_get_current_pid_tgid();
+	struct piddata piddata;
+	u32 tgid = id >> 32;
+	u32 pid = id;
+
+	if (!fp)
+		return 0;
+	if (targ_tgid && targ_tgid != tgid)
+		return 0;
+
+	piddata.ts = bpf_ktime_get_ns();
+	piddata.start = s;
+	piddata.end = e;
+	piddata.fp = fp;
+	bpf_map_update_elem(&start, &pid, &piddata, 0);
+	return 0;
+}
+
+SEC("kprobe/xfs_file_read_iter")
+int BPF_KPROBE(kprobe__xfs_file_read_iter, struct kiocb *iocb)
+{
+	struct file *fp = BPF_CORE_READ(iocb, ki_filp);
+	loff_t start = BPF_CORE_READ(iocb, ki_pos);
+
+	return probe_entry(fp, start, 0);
+}
+
+SEC("kprobe/xfs_file_write_iter")
+int BPF_KPROBE(kprobe__xfs_file_write_iter, struct kiocb *iocb)
+{
+	struct file *fp = BPF_CORE_READ(iocb, ki_filp);
+	loff_t start = BPF_CORE_READ(iocb, ki_pos);
+
+	return probe_entry(fp, start, 0);
+}
+
+SEC("kprobe/xfs_file_open")
+int BPF_KPROBE(kprobe__xfs_file_open, struct inode *inode, struct file *file)
+{
+	return probe_entry(file, 0, 0);
+}
+
+SEC("kprobe/xfs_file_fsync")
+int BPF_KPROBE(kprobe__xfs_file_fsync, struct file *file, loff_t start,
+	       loff_t end)
+{
+	return probe_entry(file, start, end);
+}
+
+static __always_inline int
+probe_exit(struct pt_regs *ctx, char type, ssize_t size)
+{
+	u64 id = bpf_get_current_pid_tgid();
+	u64 end_ns = bpf_ktime_get_ns();
+	struct piddata *piddatap;
+	struct event event = {};
+	struct dentry *dentry;
+	const u8 *qs_name_ptr;
+	u32 tgid = id >> 32;
+	struct file *fp;
+	u32 pid = id;
+	u64 delta_us;
+	u32 qs_len;
+
+	if (targ_tgid && targ_tgid != tgid)
+		return 0;
+
+	piddatap = bpf_map_lookup_elem(&start, &pid);
+	if (!piddatap)
+		return 0;    /* missed entry */
+
+	delta_us = (end_ns - piddatap->ts) / 1000;
+	bpf_map_delete_elem(&start, &pid);
+
+	if ((s64)delta_us < 0 || delta_us <= min_lat * 1000)
+		return 0;
+
+	fp = piddatap->fp;
+	dentry = BPF_CORE_READ(fp, f_path.dentry);
+	qs_len = BPF_CORE_READ(dentry, d_name.len);
+	qs_name_ptr = BPF_CORE_READ(dentry, d_name.name);
+	bpf_probe_read_str(&event.file, sizeof(event.file), qs_name_ptr);
+	bpf_get_current_comm(&event.task, sizeof(event.task));
+	event.delta_us = delta_us;
+	event.end_ns = end_ns;
+	event.offset = piddatap->start;
+	if (type != TRACE_FSYNC)
+		event.size = size;
+	else
+		event.size = piddatap->end - piddatap->start;
+	event.type = type;
+	event.tgid = tgid;
+
+	/* output */
+	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU,
+			      &event, sizeof(event));
+	return 0;
+}
+
+SEC("kretprobe/xfs_file_read_iter")
+int BPF_KRETPROBE(kretprobe__xfs_file_read_iters, ssize_t ret)
+{
+	return probe_exit(ctx, TRACE_READ, ret);
+}
+
+SEC("kretprobe/xfs_file_write_iter")
+int BPF_KRETPROBE(kretprobe__xfs_file_write_iter, ssize_t ret)
+{
+	return probe_exit(ctx, TRACE_WRITE, ret);
+}
+
+SEC("kretprobe/xfs_file_open")
+int BPF_KRETPROBE(kretprobe__xfs_file_open)
+{
+	return probe_exit(ctx, TRACE_OPEN, 0);
+}
+
+SEC("kretprobe/xfs_file_fsync")
+int BPF_KRETPROBE(kretprobe__xfs_file_sync)
+{
+	return probe_exit(ctx, TRACE_FSYNC, 0);
+}
+
+char LICENSE[] SEC("license") = "GPL";
--- a/libbpf-tools/.gitignore
+++ b/libbpf-tools/.gitignore
@@ -2,3 +2,4 @@
 /drsnoop
 /opensnoop
 /runqslower
+/xfsslower
--- a/libbpf-tools/Makefile
+++ b/libbpf-tools/Makefile
@@ -9,7 +9,7 @@ INCLUDES := -I$(OUTPUT)
 CFLAGS := -g -O2 -Wall
 ARCH := $(shell uname -m | sed 's/x86_64/x86/')
 
-APPS = drsnoop opensnoop runqslower
+APPS = drsnoop opensnoop runqslower xfsslower
 
 .PHONY: all
 all: $(APPS)
--- a/libbpf-tools/xfsslower.c
+++ b/libbpf-tools/xfsslower.c
@@ -1,0 +1,262 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+// Copyright (c) 2020 Wenbo Zhang
+//
+// Based on xfsslower(8) from BCC by Brendan Gregg & Dina Goldshtein.
+// 9-Mar-2020   Wenbo Zhang   Created this.
+#include <argp.h>
+#include <limits.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/resource.h>
+#include <unistd.h>
+#include <time.h>
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+#include "xfsslower.h"
+#include "xfsslower.skel.h"
+#include "trace_helpers.h"
+
+#define PERF_BUFFER_PAGES	64
+#define PERF_BUFFER_TIME_MS	10
+#define PERF_POLL_TIMEOUT_MS	100
+
+#define NSEC_PER_SEC		1000000000ULL
+
+static struct env {
+	pid_t pid;
+	time_t duration;
+	__u64 min_lat;
+	bool csv;
+	bool verbose;
+} env = {
+	.min_lat = 10000,
+};
+
+const char *argp_program_version = "xfsslower 0.1";
+const char *argp_program_bug_address = "<ethercflow@gmail.com>";
+const char argp_program_doc[] =
+"Trace common XFS file operations slower than a threshold.\n"
+"\n"
+"Usage: xfslower [--help] [-p PID] [-m MIN] [-d DURATION] [-c]\n"
+"\n"
+"EXAMPLES:\n"
+"    xfsslower          # trace operations slower than 10 ms (default)"
+"    xfsslower 0        # trace all operations (warning: verbose)\n"
+"    xfsslower -p 123   # trace pid 123\n"
+"    xfsslower -c -d 1  # ... 1s, parsable output (csv)";
+
+static const struct argp_option opts[] = {
+	{ "csv", 'c', NULL, 0, "Output as csv" },
+	{ "duration", 'd', "DURATION", 0, "Total duration of trace in seconds" },
+	{ NULL, 'h', NULL, OPTION_HIDDEN, "Show the full help"},
+	{ "pid", 'p', "PID", 0, "Process PID to trace" },
+	{ "min", 'm', "MIN", 0, "Min latency of trace in ms (default 10)" },
+	{ "verbose", 'v', NULL, 0, "Verbose debug output" },
+	{},
+};
+
+static error_t parse_arg(int key, char *arg, struct argp_state *state)
+{
+	long long min_lat;
+	time_t duration;
+	int pid;
+
+	switch (key) {
+	case 'v':
+		env.verbose = true;
+		break;
+	case 'h':
+		argp_usage(state);
+		break;
+	case 'c':
+		env.csv = true;
+		break;
+	case 'd':
+		errno = 0;
+		duration = strtol(arg, NULL, 10);
+		if (errno || duration <= 0) {
+			fprintf(stderr, "invalid DURATION: %s\n", arg);
+			argp_usage(state);
+		}
+		env.duration = duration;
+		break;
+	case 'm':
+		errno = 0;
+		min_lat = strtoll(arg, NULL, 10);
+		if (errno || min_lat < 0) {
+			fprintf(stderr, "invalid delay (in ms): %s\n", arg);
+		}
+		env.min_lat = min_lat;
+		break;
+	case 'p':
+		errno = 0;
+		pid = strtol(arg, NULL, 10);
+		if (errno || pid <= 0) {
+			fprintf(stderr, "invalid PID: %s\n", arg);
+			argp_usage(state);
+		}
+		env.pid = pid;
+		break;
+	default:
+		return ARGP_ERR_UNKNOWN;
+	}
+	return 0;
+}
+
+int libbpf_print_fn(enum libbpf_print_level level,
+		const char *format, va_list args)
+{
+	if (level == LIBBPF_DEBUG && !env.verbose)
+		return 0;
+	return vfprintf(stderr, format, args);
+}
+
+static int bump_memlock_rlimit(void)
+{
+	struct rlimit rlim_new = {
+		.rlim_cur	= RLIM_INFINITY,
+		.rlim_max	= RLIM_INFINITY,
+	};
+
+	return setrlimit(RLIMIT_MEMLOCK, &rlim_new);
+}
+
+void handle_event(void *ctx, int cpu, void *data, __u32 data_sz)
+{
+	const struct event *e = data;
+	struct tm *tm;
+	char ts[32];
+	time_t t;
+
+	time(&t);
+	tm = localtime(&t);
+	strftime(ts, sizeof(ts), "%H:%M:%S", tm);
+
+	if (env.csv) {
+		printf("%lld,%s,%d,%c,", e->end_ns, e->task, e->tgid, e->type);
+		if (e->size == LLONG_MAX)
+			printf("LL_MAX,");
+		else
+			printf("%ld,", e->size);
+		printf("%lld,%lld,%s\n", e->offset, e->delta_us, e->file);
+	} else {
+		printf("%-8s %-14.14s %-6d %c ", ts, e->task, e->tgid, e->type);
+		if (e->size == LLONG_MAX)
+			printf("%-7s ", "LL_MAX");
+		else
+			printf("%-7ld ", e->size);
+		printf("%-8lld %7.2f %s\n", e->offset / 1024,
+		       (double)e->delta_us / 1000, e->file);
+	}
+}
+
+void handle_lost_events(void *ctx, int cpu, __u64 lost_cnt)
+{
+	fprintf(stderr, "lost %llu events on CPU #%d\n", lost_cnt, cpu);
+}
+
+uint64_t get_ktime_ns(void)
+{
+	struct timespec ts;
+
+	clock_gettime(CLOCK_MONOTONIC, &ts);
+	return ts.tv_sec * NSEC_PER_SEC + ts.tv_nsec;
+}
+
+int main(int argc, char **argv)
+{
+	static const struct argp argp = {
+		.options = opts,
+		.parser = parse_arg,
+		.doc = argp_program_doc,
+	};
+	struct perf_buffer_opts pb_opts;
+	struct perf_buffer *pb = NULL;
+	struct xfsslower_bpf *obj;
+	__u64 time_end = 0;
+	int err;
+
+	err = argp_parse(&argp, argc, argv, 0, NULL, NULL);
+	if (err)
+		return err;
+
+	libbpf_set_print(libbpf_print_fn);
+
+	err = bump_memlock_rlimit();
+	if (err) {
+		fprintf(stderr, "failed to increase rlimit: %d\n", err);
+		return 1;
+	}
+
+	obj = xfsslower_bpf__open();
+	if (!obj) {
+		fprintf(stderr, "failed to open and/or load BPF ojbect\n");
+		return 1;
+	}
+
+	/* initialize global data (filtering options) */
+	obj->rodata->min_lat = env.min_lat;
+	obj->rodata->targ_tgid = env.pid;
+
+	err = xfsslower_bpf__load(obj);
+	if (err) {
+		fprintf(stderr, "failed to load BPF object: %d\n", err);
+		goto cleanup;
+	}
+
+	err = xfsslower_bpf__attach(obj);
+	if (err) {
+		fprintf(stderr, "failed to attach BPF programs\n");
+		goto cleanup;
+	}
+
+	if (env.csv)
+		printf("ENDTIME_us,TASK,PID,TYPE,BYTES,OFFSET_b,LATENCY_us,FILE");
+	else {
+		if (env.min_lat)
+			printf("Tracing XFS operations slower than %llu ms",
+				env.min_lat);
+		else
+			printf("Tracing XFS operations");
+		if (env.duration)
+			printf(" for %ld secs.\n", env.duration);
+		else
+			printf("... Hit Ctrl-C to end.\n");
+		printf("%-8s %-14s %-6s %1s %-7s %-8s %7s %s",
+			"TIME", "COMM", "PID", "T", "BYTES", "OFF_KB", "LAT(ms)",
+			"FILENAME\n");
+	}
+
+	pb_opts.sample_cb = handle_event;
+	pb_opts.lost_cb = handle_lost_events;
+	pb = perf_buffer__new(bpf_map__fd(obj->maps.events), PERF_BUFFER_PAGES,
+			      &pb_opts);
+	err = libbpf_get_error(pb);
+	if (err) {
+		pb = NULL;
+		fprintf(stderr, "failed to open perf buffer: %d\n", err);
+		goto cleanup;
+	}
+
+	/* setup duration */
+	if (env.duration)
+		time_end = get_ktime_ns() + env.duration * NSEC_PER_SEC;
+
+	/* main: poll */
+	while (1) {
+		usleep(PERF_BUFFER_TIME_MS * 1000);
+		if ((err = perf_buffer__poll(pb, PERF_POLL_TIMEOUT_MS)) < 0)
+			break;
+		if (env.duration && get_ktime_ns() > time_end)
+			goto cleanup;
+	}
+	fprintf(stderr, "failed with polling perf buffer: %d\n", err);
+
+cleanup:
+	perf_buffer__free(pb);
+	xfsslower_bpf__destroy(obj);
+
+	return err != 0;
+}
--- a/libbpf-tools/xfsslower.h
+++ b/libbpf-tools/xfsslower.h
@@ -1,0 +1,24 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+#ifndef __XFSSLOWER_H
+#define __XFSSLOWER_H
+
+#define DNAME_INLINE_LEN 32
+#define TASK_COMM_LEN    16
+
+#define TRACE_READ   'R'
+#define TRACE_WRITE  'W'
+#define TRACE_OPEN   'O'
+#define TRACE_FSYNC  'F'
+
+struct event {
+	char file[DNAME_INLINE_LEN];
+	char task[TASK_COMM_LEN];
+	__u64 delta_us;
+	__u64 end_ns;
+	__s64 offset;
+	ssize_t size;
+	pid_t tgid;
+	char type;
+};
+
+#endif /* __DRSNOOP_H */


ethercflow:
Hi @anakryiko , I want to try `fentry/fexit` replace `kprobe/kretprobe` like this PR. But got runtime error:

> libbpf: fentry/xfs_file_read_iter is not found in vmlinux BTF
> libbpf: failed to load object 'xfsslower_bpf'
> libbpf: failed to load BPF skeleton 'xfsslower_bpf': -2
> failed to load BPF object: -2

Would you please help me to find if there's anything wrong with the code? Thank you.

ethercflow:
I've `xfs` module:

> ➜  ~ cat /proc/kallsyms | grep xfs_file_read_iter
> ffffffffc0cd7860 t xfs_file_read_iter	[xfs]
> ➜  ~ lsmod | grep xfs
> xfs                  1699840  1
> libcrc32c              16384  5 nf_conntrack,nf_nat,bnx2x,xfs,ip_vs

anakryiko:
Right, xfs is built as a module, that's what I suspected. BTF is generated and supported only for vmlinux itself right now, not for kernel modules, unfortunately.  So to make fentry/fexit work with them you'll need to build-in XFS support.

ethercflow:
> Right, xfs is built as a module, that's what I suspected. BTF is generated and supported only for vmlinux itself right now, not for kernel modules, unfortunately. So to make fentry/fexit work with them you'll need to build-in XFS support.

Thank you @anakryiko , I just tried with build-in XFS, but still get the same errror:

> ➜  libbpf-tools git:(libbpf-tools) ✗ ./xfsslower
> libbpf: fentry/xfs_file_read_iter is not found in vmlinux BTF
> libbpf: failed to load object 'xfsslower_bpf'
> libbpf: failed to load BPF skeleton 'xfsslower_bpf': -2
> failed to load BPF object: -2
> ➜  libbpf-tools git:(libbpf-tools) ✗ cat /proc/kallsyms| grep xfs_file_read_iter
> ffffffff92035aa0 t xfs_file_read_iter
> ➜  libbpf-tools git:(libbpf-tools) ✗

To use ` fentry/fexit`, does it need any `CONFIG_XX`? I've checked

```
CONFIG_HAVE_FENTRY=y
CONFIG_FUNCTION_TRACER=y
```

anakryiko:
You have pahole 1.15, isn't that right? fentry/fexit relies on func infos in BTF, which are produced starting from pahole 1.16, can you please try upgrading?

libbpf-tools/xfsslower.bpf.c
```diff
@@ -0,0 +1,163 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2020 Wenbo Zhang
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include "xfsslower.h"
+
+#define BPF_F_INDEX_MASK		0xffffffffULL
+#define BPF_F_CURRENT_CPU		BPF_F_INDEX_MASK
```
anakryiko: you shouldn't need this anymore, now it's part of vmlinux.h

libbpf-tools/xfsslower.bpf.c
```diff
@@ -0,0 +1,163 @@
...
+	piddata.fp = fp;
+	piddata.start = start;
+	piddata.end = end;
+	if (piddata.fp)
+		bpf_map_update_elem(&start, &pid, &piddata, 0);
+	return 0;
+}
+
+SEC("fentry/xfs_file_read_iter")
+int fentry__xfs_file_read_iter(struct args *ctx)
```
anakryiko: I'd recommend using BPF_PROG macro from bpf_tracing.h to access arguments of fentry/fexit, like so:
```
int BPF_PROG(fentry__xfs_file_read_iter, struct kiocb *iocb, struct iov_iter *to) {
...
}
```

libbpf-tools/xfsslower.bpf.c
```diff
@@ -0,0 +1,163 @@
...
+SEC("fentry/xfs_file_write_iter")
+int fentry__xfs_file_write_iter(struct args *ctx)
+{
+	struct kiocb *iocb = (void*)ctx->args[0];
+
+	return probe_entry(iocb->ki_filp, iocb->ki_pos, 0);
+}
+
+SEC("fentry/xfs_file_open")
+int fentry__xfs_file_open(struct args *ctx)
```
anakryiko: same as above, you with BPF_PROG you can just declare `struct inode *inode, struct file *file` as input arguments

libbpf-tools/xfsslower.bpf.c
```diff
@@ -0,0 +1,163 @@
...
+	if (!piddatap)
+		return 0;    /* missed entry */
+
+	delta_us = (end_ns - piddatap->ts) / 1000;
+	bpf_map_delete_elem(&start, &pid);
+
+	if ((s64)delta_us < 0 || delta_us <= min_lat)
+		return 0;
+
+	qs = BPF_CORE_READ(piddatap, fp, f_path.dentry, d_name);
```
anakryiko: given that you are using fentry/fexit programs, you should be able to do this equivalently as :

```
qs = pidatatap->fp->f_path.dentry->d_name;
```

how cool is that, eh? :)

libbpf-tools/xfsslower.bpf.c
```diff
@@ -0,0 +1,163 @@
...
+	event.tgid = tgid;
+
+	/* output */
+	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU,
+			      &event, sizeof(event));
+	return 0;
+}
+
+SEC("fexit/xfs_file_read_iter")
+int fexit__xfs_file_read_iter(struct args *ctx)
```
anakryiko: all fexits can be declared in the same way as fentries, except last argument also captures function return:
```
int BPF_PROG(fexit__xfs_file_read_iter, struct kiocb *iocb, struct iov_iter *to, ssize_t ret) {
...
}
```

BTW, BPF_PROG still implicitly defines `u64 *ctx`, which you can pass along into bpf_perf_event_output.

ethercflow:
> You have pahole 1.15, isn't that right? fentry/fexit relies on func infos in BTF, which are produced starting from pahole 1.16, can you please try upgrading?

Thank you @anakryiko , I updated to 1.16 at host then make clean && rebuild kernel with gcc 8.3.0, and restart qemu to load the recompiled kernel. But still met that error. Did I miss anything else?

1. env
> guest os version: 5.6.0-rc3+
> host pahole version: v1.16
> compile kernel with gcc (Ubuntu 8.3.0-26ubuntu1~18.04) 8.3.0

2. Upload full Kconfig (zcat /proc/config.gz)
[config.txt](https://github.com/iovisor/bcc/files/4310819/config.txt)

3. Upload verbose log in xfsslower (add -v command line arg) and upload it in full. 
[xfsslower_log.txt](https://github.com/iovisor/bcc/files/4310827/xfsslower_log.txt)

4. Upload full dump of bin/bpftool btf dump file /sys/kernel/btf/vmlinux (not a C one)
[vmlinux.txt](https://github.com/iovisor/bcc/files/4310829/vmlinux.txt)

libbpf-tools/xfsslower.bpf.c
ethercflow: So cool! :) I found I missed 

> With tp_btf and fentry/fexit BPF program types, available since Linux 5.5, natural C syntax is possible as well.

from your blog.

anakryiko: Actually, verifier might not allow direct memory access in this particular case, because you are storing `struct file *` in map, at which point verifier has no way to guarantee that this pointer won't be overwritten from userspace. Please try it, but you might need to stick to BPF_CORE_READ for this one, unfortunately.

anakryiko:
So I have bad news for you w.r.t. fentry/fexit usage here. xfs_file_read_iter is static function, and those functions are not recorded (at least right now) in BTF. Which means you can't attach to them with fentry/fexit programs. Your best bet for now is to stick to kprobe, probably.

ethercflow:
> So I have bad news for you w.r.t. fentry/fexit usage here. xfs_file_read_iter is static function, and those functions are not recorded (at least right now) in BTF. Which means you can't attach to them with fentry/fexit programs. Your best bet for now is to stick to kprobe, probably.

Get it, I'll change this to kprobe and try `entry/fexit` with sth else later. Thank you.

libbpf-tools/xfsslower.bpf.c
ethercflow: Thanks for the details, I'll try it later.

ethercflow:
> > So I have bad news for you w.r.t. fentry/fexit usage here. xfs_file_read_iter is static function, and those functions are not recorded (at least right now) in BTF. Which means you can't attach to them with fentry/fexit programs. Your best bet for now is to stick to kprobe, probably.
> 
> Get it, I'll change this to `kprobe` and try `fentry/fexit` with sth else later. Thank you.

Hi @anakryiko ,  are `BPF_KPROBE` && `PT_REGS_PARM` available now?

I tried `BPF_KRPOBE` from your blog's example, 

```c
SEC("kprobe/acct_collect")
int BPF_KPROBE(kprobe__acct_collect, long exit_code, int group_dead)
{

}
```

with https://github.com/libbpf/libbpf/commit/9a424bea428bd7cde6f1bdaa2ee41f48e5686774, but got some warning and error at compile time:

[kprobe.txt](https://github.com/iovisor/bcc/files/4322312/kprobe.txt)



BTW,  from
> So I have bad news for you w.r.t. fentry/fexit usage here. xfs_file_read_iter is static function, and those functions are not recorded (at least right now) in BTF. Which means you can't attach to them with fentry/fexit programs. Your best bet for now is to stick to kprobe, probably.

so if the functions are from kernel modules or defined as static.  I need to use as bcc's way like:

```c
#ifdef __BCC__
int kprobe__acct_collect(struct pt_regs *ctx, long exit_code, int group_dead)
```

instead of CO-RE right?  Thank you.

anakryiko:
Wenbo, right, once libbpf is updated (https://github.com/iovisor/bcc/pull/2814), BPF_KPROBE and PT_REGS_PARM macros should be available. But there is a bit of complication due to using -target bpf, which loses track of host architecture. I forgot about that, given kernel selftests solved that with extra __TARGET_ARCH_$(SRCARCH) defined. I've attempted to fix that in https://github.com/iovisor/bcc/pull/2815, let me know if that works for you. I've only tested on x86_64, though.

anakryiko:
As for the other part of your question. You are conflating few things. When talking about kprobes, we can talk about BCC and libbpf version only. CO-RE is, in the most strict sense, only about field relocations (so all those BPF_CORE_READ macro). Then fentry/fexit are another type of BPF programs, can work without using CO-RE relocations, but in practice those two are used together most frequently. So for your use case, you need libbpf-style kprobes. I'd recomment using BPF_KPROBE, but you can go with:

```
SEC("kprobe/whatever")
int my_kprobe(struct pt_regs *ctx)
{
    int arg0 = PT_REGS_PARM1(ctx);
   ...
}
```

BPF_KPROBE conveniently hides those translations from you, nothing more. We can call this libbpf-style kprobes.

ethercflow:
> As for the other part of your question. You are conflating few things. When talking about kprobes, we can talk about BCC and libbpf version only. CO-RE is, in the most strict sense, only about field relocations (so all those BPF_CORE_READ macro). Then fentry/fexit are another type of BPF programs, can work without using CO-RE relocations, but in practice those two are used together most frequently. So for your use case, you need libbpf-style kprobes. I'd recomment using BPF_KPROBE, but you can go with:
> 
> ```
> SEC("kprobe/whatever")
> int my_kprobe(struct pt_regs *ctx)
> {
>     int arg0 = PT_REGS_PARM1(ctx);
>    ...
> }
> ```
> 
> BPF_KPROBE conveniently hides those translations from you, nothing more. We can call this libbpf-style kprobes.



> As for the other part of your question. You are conflating few things. When talking about kprobes, we can talk about BCC and libbpf version only. CO-RE is, in the most strict sense, only about field relocations (so all those BPF_CORE_READ macro). Then fentry/fexit are another type of BPF programs, can work without using CO-RE relocations, but in practice those two are used together most frequently. So for your use case, you need libbpf-style kprobes. I'd recomment using BPF_KPROBE, but you can go with:
> 
> ```
> SEC("kprobe/whatever")
> int my_kprobe(struct pt_regs *ctx)
> {
>     int arg0 = PT_REGS_PARM1(ctx);
>    ...
> }
> ```
> 
> BPF_KPROBE conveniently hides those translations from you, nothing more. We can call this libbpf-style kprobes.

Thank you @anakryiko so much, I did think that the clang 's precompiler might handle `BPF_xxx` to do special processing to let `xxx's bpf prog` use with BTF and with `BCC` macro, it'll fallback to work without `BTF`, so I got 

> ➜ libbpf-tools git:(libbpf-tools) ✗ ./xfsslower
> libbpf: fentry/xfs_file_read_iter is not found in vmlinux BTF
> libbpf: failed to load object 'xfsslower_bpf'
> libbpf: failed to load BPF skeleton 'xfsslower_bpf': -2
> failed to load BPF object: -2
> ➜ libbpf-tools git:(libbpf-tools) ✗ cat /proc/kallsyms| grep xfs_file_read_iter
> ffffffff92035aa0 t xfs_file_read_iter
> ➜ libbpf-tools git:(libbpf-tools) ✗

It is a bad habit to speculate on the principle, I should throw my doubts out and ask. :(

I think I misunderstood and I have a confusion about:

> So I have bad news for you w.r.t. fentry/fexit usage here. xfs_file_read_iter is static function, and those functions are not recorded (at least right now) in BTF. Which means you can't attach to them with fentry/fexit programs. Your best bet for now is to stick to kprobe, probably.

As `xfs_file_read_iter` is static and not recorded in `BTF` (I think `BTF` == `CO-RE` right?), and  `fentry/fexit` are another type of BPF programs, can work without using `CO-RE` relocations.  So how can I fix 

> ➜ libbpf-tools git:(libbpf-tools) ✗ ./xfsslower
> libbpf: fentry/xfs_file_read_iter is not found in vmlinux BTF
> libbpf: failed to load object 'xfsslower_bpf'
> libbpf: failed to load BPF skeleton 'xfsslower_bpf': -2
> failed to load BPF object: -2
> ➜ libbpf-tools git:(libbpf-tools) ✗ cat /proc/kallsyms| grep xfs_file_read_iter
> ffffffff92035aa0 t xfs_file_read_iter
> ➜ libbpf-tools git:(libbpf-tools) ✗

with `fentry/fexit` + `non-BTF`?

Does it mean I need to use another libbpf's APIs in userspace like libbcc did?  (I think I only changed from `fentry/fexit` to `kprobe` may still meet the similar error as `xfs_file_read_iter`'s BTF doesn't exit.)

ethercflow:
> Wenbo, right, once libbpf is updated (#2814), BPF_KPROBE and PT_REGS_PARM macros should be available. But there is a bit of complication due to using -target bpf, which loses track of host architecture. I forgot about that, given kernel selftests solved that with extra `__TARGET_ARCH_$(SRCARCH)` defined. I've attempted to fix that in #2815, let me know if that works for you. I've only tested on x86_64, though.

Thank you, I tested @anakryiko on x86_64, it works for me. No warnings and errors at compile time. But got runtime error:

> libbpf: prog 'kprobe/xfs_file_read_iter': relo #0: matching error: -22
> libbpf: prog 'kprobe/xfs_file_read_iter': relo #0: failed to relocate: -22
> libbpf: failed to perform CO-RE relocations: -22
> libbpf: failed to load object 'xfsslower_bpf'
> libbpf: failed to load BPF skeleton 'xfsslower_bpf': -22
> failed to load BPF object: -22

I thought it maybe related to https://github.com/iovisor/bcc/pull/2806#issuecomment-598043702 right?

anakryiko:
@ethercflow, BTF is used for many things. At it's essence it's just a type information available form kernel at runtime. BTF is used for CO-RE relocations, of course. But it's also used for fentry/fexit to find kernel's function type definiton. It's also used for .struct_ops, but that's for another day. My point is that BTF != CO-RE, BTF is more generally used.

Now, you can't use fentry/fexit without BTF. Which means you are stuck with kprobe. If you further need to use CO-RE with that kprobe with types defined in a module -- then again, you can't do that. It all comes down to BTF availability for functions and types you need from kernel.

In your case, seems like you only care about struct kiocb and struct file, both should be in kernel. So it seems like kprobe + CO-RE will work for your use-case, no? So please try that approach.

ethercflow:
> @ethercflow, BTF is used for many things. At it's essence it's just a type information available form kernel at runtime. BTF is used for CO-RE relocations, of course. But it's also used for fentry/fexit to find kernel's function type definiton. It's also used for .struct_ops, but that's for another day. My point is that BTF != CO-RE, BTF is more generally used.

@anakryiko Thank you for the details, I think I understand this now.

> Now, you can't use fentry/fexit without BTF. Which means you are stuck with kprobe. If you further need to use CO-RE with that kprobe with types defined in a module -- then again, you can't do that. It all comes down to BTF availability for functions and types you need from kernel.

Please help me check my understanding of this passage：

1. The `fentry/fexit` needs `BTF` to work, the related `BTF` must include function types and parameter types, no matter whether with CO-RE relocations or not;
2. The `kprobe + CO-RE` neeeds `BTF` to work, but the related `BTF` can only include the necessary paramters' types which the kprobe handlers need;


> In your case, seems like you only care about struct kiocb and struct file, both should be in kernel. So it seems like kprobe + CO-RE will work for your use-case, no? So please try that approach.

Thank you, I tried with `kprobe + CO-RE` and got runtime error:

> libbpf: prog 'kprobe/xfs_file_read_iter': relo #0: matching error: -22
> libbpf: prog 'kprobe/xfs_file_read_iter': relo #0: failed to relocate: -22
> libbpf: failed to perform CO-RE relocations: -22
> libbpf: failed to load object 'xfsslower_bpf'
> libbpf: failed to load BPF skeleton 'xfsslower_bpf': -22
> failed to load BPF object: -22

Then I replace xfsslower.bpf.c's `kprobe`s with your example from your blog:

```c
#include "vmlinux.h"
#include <bpf/bpf_helpers.h>
#include <bpf/bpf_core_read.h>
#include <bpf/bpf_tracing.h>

SEC("kprobe/acct_collect")
int BPF_KPROBE(kprobe__acct_collect, long exit_code, int group_dead)
{
        bpf_printk("%ld,%d\n", exit_code, group_dead);
        return 0;
}

char LICENSE[] SEC("license") = "GPL";
```

got similar error (with -v):

> libbpf: loading object 'xfsslower_bpf' from buffer
> libbpf: section(1) .text, size 0, link 0, flags 6, type=1
> libbpf: skip section(1) .text
> libbpf: section(2) kprobe/acct_collect, size 104, link 0, flags 6, type=1
> libbpf: found program kprobe/acct_collect
> libbpf: section(3) license, size 4, link 0, flags 3, type=1
> libbpf: license of xfsslower_bpf is GPL
> libbpf: section(4) .rodata.str1.1, size 9, link 0, flags 32, type=1
> libbpf: skip section(4) .rodata.str1.1
> libbpf: section(5) .BTF, size 837, link 0, flags 0, type=1
> libbpf: section(6) .BTF.ext, size 188, link 0, flags 0, type=1
> libbpf: section(7) .symtab, size 120, link 11, flags 0, type=2
> libbpf: section(8) .rel.BTF, size 16, link 7, flags 0, type=9
> libbpf: skip relo .rel.BTF(8) for section(5)
> libbpf: section(9) .rel.BTF.ext, size 128, link 7, flags 0, type=9
> libbpf: skip relo .rel.BTF.ext(9) for section(6)
> libbpf: section(10) .llvm_addrsig, size 2, link 0, flags 80000000, type=1879002115
> libbpf: skip section(10) .llvm_addrsig
> libbpf: section(11) .strtab, size 147, link 0, flags 0, type=3
> libbpf: skip section(11) .strtab
> libbpf: looking for externs among 5 symbols...
> libbpf: collected 0 externs total
> libbpf: loading kernel BTF '/sys/kernel/btf/vmlinux': 0
> libbpf: prog 'kprobe/acct_collect': performing 2 CO-RE offset relocs
> libbpf: prog 'kprobe/acct_collect': relo #0: kind 0, spec is [2] pt_regs + 0:13 => 104.0 @ &x[0].si
> libbpf: [2] pt_regs: found candidate [198] pt_regs
> libbpf: [2] pt_regs: found candidate [9195] pt_regs
> libbpf: prog 'kprobe/acct_collect': relo #0: matching candidate #0 pt_regs against spec [198] pt_regs + 0:13 => 104.0 @ &x[0].si: 1
> libbpf: prog 'kprobe/acct_collect': relo #0: matching candidate #1 pt_regs against spec [9195] pt_regs + 0 => 0.0 @ &x[0]: -22
> libbpf: prog 'kprobe/acct_collect': relo #0: matching error: -22
> libbpf: prog 'kprobe/acct_collect': relo #0: failed to relocate: -22
> libbpf: failed to perform CO-RE relocations: -22
> libbpf: failed to load object 'xfsslower_bpf'
> libbpf: failed to load BPF skeleton 'xfsslower_bpf': -22
> failed to load BPF object: -22

What's the meaning of this error and how to fix this?

anakryiko:
> @anakryiko Thank you for the details, I think I understand this now.

You're welcome! Glad it was helpful.

> Please help me check my understanding of this passage：
> 
>     1. The `fentry/fexit` needs `BTF` to work, the related `BTF` must include function types and parameter types, no matter whether with CO-RE relocations or not;

Correct.

> 
>     2. The `kprobe + CO-RE` neeeds `BTF` to work, but the related `BTF` can only include the necessary paramters' types which the kprobe handlers need;

Correct again.

> > libbpf: prog 'kprobe/acct_collect': performing 2 CO-RE offset relocs
> > libbpf: prog 'kprobe/acct_collect': relo #0: kind 0, spec is [2] pt_regs + 0:13 => 104.0 @ &x[0].si
> > libbpf: [2] pt_regs: found candidate [198] pt_regs
> > libbpf: [2] pt_regs: found candidate [9195] pt_regs
> > libbpf: prog 'kprobe/acct_collect': relo #0: matching candidate #0 pt_regs against spec [198] pt_regs + 0:13 => 104.0 @ &x[0].si: 1
> > libbpf: prog 'kprobe/acct_collect': relo #0: matching candidate #1 pt_regs against spec [9195] pt_regs + 0 => 0.0 @ &x[0]: -22
> > libbpf: prog 'kprobe/acct_collect': relo #0: matching error: -22
> > libbpf: prog 'kprobe/acct_collect': relo #0: failed to relocate: -22
> > libbpf: failed to perform CO-RE relocations: -22
> > libbpf: failed to load object 'xfsslower_bpf'
> > libbpf: failed to load BPF skeleton 'xfsslower_bpf': -22
> > failed to load BPF object: -22
> 
> What's the meaning of this error and how to fix this?

Ok, I think this is actually a bug in libbpf. It's trying to do relocation against struct pt_regs forward declaration and fails. I'm gonna work on a fix, but meanwhile to unblock yourself you can add the following `#define` before including vmlinux.h:

```
#define BPF_NO_PRESERVE_ACCESS_INDEX
#include "vmlinux.h"
```

This will disable `preserve_access_index` attribute to be applied to all types, which you don't rely on either way (it's mostly useful for fentry/fexit/tp_btf BPF programs).

ethercflow:
> Ok, I think this is actually a bug in libbpf. It's trying to do relocation against struct pt_regs forward declaration and fails. I'm gonna work on a fix, but meanwhile to unblock yourself you can add the following #define before including vmlinux.h:
> 
> #define BPF_NO_PRESERVE_ACCESS_INDEX
> #include "vmlinux.h"
> This will disable preserve_access_index attribute to be applied to all types, which you don't rely on either way (it's mostly useful for fentry/fexit/tp_btf BPF programs).

I'll keep going after work, thanks. :)

anakryiko:
alright, should be fixed as part of this series: https://patchwork.ozlabs.org/project/netdev/list/?series=164156&state=*

ethercflow:
Hi @anakryiko , the current PR with 
```
#define BPF_NO_PRESERVE_ACCESS_INDEX
#include "vmlinux.h"
```
works as expected, PTAL. 

```sh
➜  libbpf-tools git:(master) ✗ ./xfsslower -m 0
Tracing XFS operations... Hit Ctrl-C to end.
TIME     COMM           PID    T BYTES   OFF_KB   LAT(ms) FILENAME
01:44:27 qemu-system-x8 24565  W 4096    553672      0.05 stretch.img
01:44:27 qemu-system-x8 24565  W 4096    167248      0.02 stretch.img
01:44:27 qemu-system-x8 24565  W 4096    12255028    0.02 stretch.img
01:44:33 qemu-system-x8 24565  F LL_MAX  0           0.47 stretch.img
01:44:33 qemu-system-x8 24565  W 4096    19009300    0.05 stretch.img
01:44:33 qemu-system-x8 24565  F LL_MAX  0           0.09 stretch.img
01:44:50 qemu-system-x8 24565  W 16384   19009304    0.08 stretch.img
01:44:50 qemu-system-x8 24565  F LL_MAX  0           0.17 stretch.img
01:44:50 qemu-system-x8 24565  W 4096    19009320    0.02 stretch.img
01:44:50 qemu-system-x8 24565  F LL_MAX  0           0.10 stretch.img
01:45:00 qemu-system-x8 24565  W 16384   19009324    0.07 stretch.img
01:45:00 qemu-system-x8 24565  F LL_MAX  0           0.16 stretch.img
01:45:00 qemu-system-x8 24565  W 4096    19009340    0.02 stretch.img
01:45:00 qemu-system-x8 24565  F LL_MAX  0           0.09 stretch.img
01:45:15 qemu-system-x8 24565  W 4096    553672      0.05 stretch.img
01:45:15 qemu-system-x8 24565  W 4096    167248      0.02 stretch.img
01:45:15 qemu-system-x8 24565  W 4096    12255028    0.02 stretch.img
01:45:20 qemu-system-x8 24565  W 4096    2097292     0.05 stretch.img
01:45:20 qemu-system-x8 24565  W 4096    2097444     0.02 stretch.img
01:45:20 qemu-system-x8 24565  W 4096    2097452     0.02 stretch.img
01:45:21 qemu-system-x8 24565  F LL_MAX  0           0.37 stretch.img
01:45:21 qemu-system-x8 24565  W 4096    19009344    0.04 stretch.img
01:45:21 qemu-system-x8 24565  F LL_MAX  0           0.09 stretch.img
01:46:00 qemu-system-x8 24565  W 16384   19009348    0.10 stretch.img
01:46:00 qemu-system-x8 24565  F LL_MAX  0           0.17 stretch.img
01:46:00 qemu-system-x8 24565  W 4096    19009364    0.02 stretch.img
01:46:00 qemu-system-x8 24565  F LL_MAX  0           0.09 stretch.img
```

With a little regret as https://patchwork.ozlabs.org/patch/1211861/'s state is still `Awaiting Upstream`, we can't get a full path FILENAME. :(

BTW, I found another two issues:
1. I've pulled and merged https://github.com/iovisor/bcc/pull/2820, but found I still need to 
```c
#define NULL    0
```
2. 
> Actually, verifier might not allow direct memory access in this particular case, because you are storing `struct file *` in map, at which point verifier has no way to guarantee that this pointer won't be overwritten from userspace. Please try it, but you might need to stick to BPF_CORE_READ for this one, unfortunately.

I found
```c
qs = BPF_CORE_READ(pidstatp, fp, f_path.dentry, d_name);
```
will be rejected by verifier.

> libbpf: load bpf program failed: Invalid argument
> libbpf: -- BEGIN DUMP LOG ---
> libbpf:
> Unrecognized arg#0 type PTR
> ... ...
> ... ...
> 48: (85) call unknown#195896080
> invalid func unknown#195896080
> processed 43 insns (limit 1000000) max_states_per_insn 0 total_states 2 peak_states 2 mark_read 2

The full log: [xfsslower_log.txt](https://github.com/iovisor/bcc/files/4332334/xfsslower_log.txt) Is this behavior as expected?

libbpf-tools/xfsslower.bpf.c
```diff
@@ -0,0 +1,159 @@
...
+
+	if (targ_tgid && targ_tgid != tgid)
+		return 0;
+
+	piddata.ts = bpf_ktime_get_ns();
+	piddata.start = s;
+	piddata.end = e;
+	piddata.fp = fp;
+	if (piddata.fp)
+		bpf_map_update_elem(&start, &pid, &piddata, 0);
```
anakryiko: you can save a bunch of unnecessary work by exiting early `if (!fp)`, even before targ_tgid check. Also, piddata initialization won't be necessary, you are filling all the fields explicitly either way.

libbpf-tools/xfsslower.bpf.c
```diff
@@ -0,0 +1,159 @@
...
+}
+
+static __always_inline int probe_exit(struct pt_regs *ctx, char type)
+{
+	u64 id = bpf_get_current_pid_tgid();
+	u64 end_ns = bpf_ktime_get_ns();
+	struct piddata *piddatap;
+	struct event event = {};
+	struct file *fp = NULL;
+	struct qstr qs = {};
```
anakryiko: nit: no need for initialization if you are setting all those values unconditionally later

libbpf-tools/xfsslower.bpf.c
```diff
@@ -0,0 +1,159 @@
...
+		return 0;    /* missed entry */
+
+	delta_us = (end_ns - piddatap->ts) / 1000;
+	bpf_map_delete_elem(&start, &pid);
+
+	if ((s64)delta_us < 0 || delta_us <= min_lat * 1000)
+		return 0;
+
+	fp = piddatap->fp;
+	qs = BPF_CORE_READ(fp, f_path.dentry, d_name);
```
anakryiko: what you did here w.r.t. piddatap->fp is right, `BPF_CORE_READ(piddatap, fp, f_path.dentry, d_name)` isn't supposed to work, sorry for misleading you initially. The reason is that `struct piddata *` is local type available only in your BPF application, not in kernel. There is both no need to relocate access to struct piddata, as well as it's going to result in invalid CO-RE relocation (because it can't be matched against kernel type). See error message in https://github.com/libbpf/libbpf/blob/master/src/libbpf.c#L4283

But, I think it's dangerous to read entire qstr like you do here. If the size of this struct changes, you'll read only part of it, potentially. Given you need both qstr.len and qstr.name, it's better to read them individually:

```
dentry = BPF_CORE_READ(fp, f_path.dentry);
qs_len = BPF_CORE_READ(dentry, d_name.len);
qs_name_ptr = BPF_CORE_READ(dentry, d_name.name);
bpf_probe_read_str(&event.file, sizeof(event.file), qs_name_ptr);
```

Note you don't need bpf_core_read_str() because you are using direct pointer to array off bytes, no need for relocation there.

libbpf-tools/xfsslower.bpf.c
```diff
@@ -0,0 +1,159 @@
...
+	event.tgid = tgid;
+
+	/* output */
+	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU,
+			      &event, sizeof(event));
+	return 0;
+}
+
+SEC("kretprobe/xfs_file_read_iter")
+int BPF_KRETPROBE(kretprobe__xfs_file_read_iters)
```
anakryiko: you can add `int ret` arg into BPF_KRETPROBE and pass it into probe_exit directly. That way you won't need explicit PT_REGS_RC() in prove_exit.

libbpf-tools/xfsslower.bpf.c
```diff
@@ -0,0 +1,159 @@
...
+		return 0;
+
+	bpf_core_read_str(&event.file, sizeof(event.file), qs.name);
+
+	bpf_get_current_comm(&event.task, sizeof(event.task));
+	event.delta_us = delta_us;
+	event.end_ns = end_ns;
+	event.offset = piddatap->start;
+	event.size = type != TRACE_FSYNC ? PT_REGS_RC(ctx) :
+		piddatap->end - piddatap->start;
```
anakryiko: nit: just from formatting standpoint, seems like if/else will work a bit better here?

```
if (type != TRACE_FSYNC)
    event.size = PT_REGS_RC(ctx); /* but see comment below about passing ret directly */
else
    event.size = piddatap->end - piddatap->start;
```

libbpf-tools/xfsslower.bpf.c
```diff
@@ -0,0 +1,159 @@
...
+}
+
+SEC("kprobe/xfs_file_read_iter")
+int BPF_KPROBE(kprobe__xfs_file_read_iter, struct kiocb *iocb,
+	       struct iov_iter *to)
+{
+	struct file *fp = BPF_CORE_READ(iocb, ki_filp);
+	loff_t start = BPF_CORE_READ(iocb, ki_pos);
+
+	return 0;
```
anakryiko: was probe_entry supposed to be here?

libbpf-tools/xfsslower.bpf.c
ethercflow: @anakryiko I had a doubt about `int ret` from `BPF_KRETPROBE`'s define 

```c
#define ___bpf_kretprobe_args0() ctx
#define ___bpf_kretprobe_args1(x) \
	___bpf_kretprobe_args0(), (void *)PT_REGS_RET(ctx)
#define ___bpf_kretprobe_args(args...) \
	___bpf_apply(___bpf_kretprobe_args, ___bpf_narg(args))(args)

/*
 * BPF_KRETPROBE is similar to BPF_KPROBE, except, it only provides optional
 * return value (in addition to `struct pt_regs *ctx`), but no input
 * arguments, because they will be clobbered by the time probed function
 * returns.
 */
#define BPF_KRETPROBE(name, args...)					    \
name(struct pt_regs *ctx);						    \
static __attribute__((always_inline)) typeof(name(0))			    \
____##name(struct pt_regs *ctx, ##args);				    \
typeof(name(0)) name(struct pt_regs *ctx)				    \
{									    \
	_Pragma("GCC diagnostic push")					    \
	_Pragma("GCC diagnostic ignored \"-Wint-conversion\"")		    \
	return ____##name(___bpf_kretprobe_args(args));			    \
	_Pragma("GCC diagnostic pop")					    \
}									    \
static __always_inline typeof(name(0)) ____##name(struct pt_regs *ctx, ##args)
```
It looks like:
```c
int ret = ctx->sp;  // ___bpf_kretprobe_args0(), (void *)PT_REGS_RET(ctx)
```
not 
```c
int ret = ctx->ax;  // PT_REGS_RC(ctx)
```
https://github.com/libbpf/libbpf/blob/master/src/bpf_tracing.h#L58
https://github.com/libbpf/libbpf/blob/master/src/bpf_tracing.h#L60

Is this a bug?

libbpf-tools/xfsslower.bpf.c
ethercflow: Oh, sorry for this mistake.

libbpf-tools/xfsslower.bpf.c
ethercflow: Agree, looks much clearer.

libbpf-tools/xfsslower.bpf.c
anakryiko: Yeah, clearly a bug, it should use PT_REGS_RC, not PT_REGS_RET, my bad. Feel free to submit a patch, if not, I'll do it tomorrow or Monday.

libbpf-tools/xfsslower.bpf.c
ethercflow: @anakryiko Thank you for pointing this, I'll remove those unnecessary init, and I've another two questions, the first is about coding style.

 ```c
u64 id = bpf_get_current_pid_tgid();
u64 end_ns = bpf_ktime_get_ns();
struct piddata *piddatap;
struct event event = {}; // if not init, the verifier will complain, I'll describe it as the second question
struct dentry *dentry;
const u8 *qs_name_ptr;
u32 qs_len;
struct file *fp;
u32 tgid = id >> 32;
u32 pid = id;
u64 delta_us;
```

I heard that the order of variable declarations is preferably an inverted Christmas tree, but some variables may be related to each other but may break it, eg

```c
const u8 *qs_name_ptr;
u32 qs_len;
```

and
```c
u32 tgid = id >> 32;
u32 pid = id;
```

Is it ok here and is there a better order?

The second is if I removed `event`'s initialization. The verifier will complain:

> invalid indirect read from stack off -88+85 size 88

Although all members of `event` have been assigned before passing to  `bpf_perf_event_output`


```c
struct event {
        char file[DNAME_INLINE_LEN];
        char task[TASK_COMM_LEN];
        __u64 delta_us;
        __u64 end_ns;
        __s64 offset;
        ssize_t size;
        pid_t tgid;
        char type;
};

fp = piddatap->fp;
dentry = BPF_CORE_READ(fp, f_path.dentry);
qs_len = BPF_CORE_READ(dentry, d_name.len);
qs_name_ptr = BPF_CORE_READ(dentry, d_name.name);
bpf_probe_read_str(&event.file, sizeof(event.file), qs_name_ptr);  // 1

bpf_get_current_comm(&event.task, sizeof(event.task));  // 2
event.delta_us = delta_us;  // 3
event.end_ns = end_ns;  // 4
event.offset = piddatap->start;  // 5
if (type != TRACE_FSYNC)
        event.size = PT_REGS_RC(ctx);  // 6
else
        event.size = piddatap->end - piddatap->start;  // 6
event.type = type;   // 7
event.tgid = tgid;    // 8
```

Is there a way for verifier to be aware of this?

anakryiko: > I heard that the order of variable declarations is preferably an inverted Christmas tree, but some variables may be related to each other but may break it, eg

There is such preference in netdev kernel tree, but that preference is overrated. All other things being equal, reverse Christmas tree is somewhat neater-looking. But if the style gets in a way (e.g., when you need a specific order of initializing variables), then do it the way you need it.

> The second is if I removed `event`'s initialization. The verifier will complain:
> 
> > invalid indirect read from stack off -88+85 size 88
> 

argh... it's a 3 byte padding at the end. I guess that `= {}` initialization is needed here, unfortunately. Either that, or explicit char _padding[3] at the end of your struct, but it will be a bit harder to maintain (because you'd need to re-specify padding every time you add or shuffle around fields). Pick your poison, I guess. No big deal to initialize struct, it's just few extra instructions, but I just thought you wouldn't need it, but seems like you do.

libbpf-tools/xfsslower.bpf.c
ethercflow: Ok, I'll submit it.

libbpf-tools/xfsslower.bpf.c
ethercflow: thanks, I'll fix this.

libbpf-tools/xfsslower.bpf.c
ethercflow: Get it, thanks for the details.

libbpf-tools/xfsslower.bpf.c
ethercflow: Get it, thanks.

anakryiko:
@ethercflow, BPF_KRETPROBE fix is being pulled into BCC in https://github.com/iovisor/bcc/pull/2823

ethercflow:
> @ethercflow, BPF_KRETPROBE fix is being pulled into BCC in #2823

Great! I'll update the PR after work. Thanks!

anakryiko:
Looks great to me, thanks for doing this and helping hammer out  remaining bugs!

ethercflow:
> this is an obvious candidate for moving into helpers, given every single app is re-implementing this. But please do that in separate PR and update all the tools. Thanks!

Get it, I'll revert this PR, and do as you told in separate PR. Thank you for your guidance!

ethercflow:
@yonghong-song @brendangregg Would you please help me to see if it‘s good enough to be merged? After merge I want to submit another PR with `fentry/fexit` example. Thanks!

yonghong-song:
[buildbot, test this please]

yonghong-song:
LGTM. Thanks!

PR #2807, opened 2020-03-10
WIP: libbpf-tools: add CO-RE execsnoop

bacher09:
Hi, I've made version of `execsnoop` using libbpf, but currently it lacks few essential parts (printing `args` and relative timestamps). I just wanted to ask few questions before implementing them.

To implement argument printing I need hash table (or some balanced tree) and wanted to ask if it would be okay if I add some library ([like this one](https://github.com/attractivechaos/klib/blob/master/khash.h)). Or maybe it would be better to use `hsearch` from glibc. Or, as another variant, implement it in CPP and use STL. To be honest, I would prefer to add header only library, since `hsearch` lacks some features  and CPP version will require libstd on host machine, library in the same time could be reused by other tools.

Also, it seems, there is a possibility of **memory leak** in current (python) `execsnoop` implementation. Basically, it receives arguments from [kprobe](https://github.com/iovisor/bcc/blob/master/tools/execsnoop.py#L277) and saves them in memory, but removes them only when receives event from [kretprobe](https://github.com/iovisor/bcc/blob/master/tools/execsnoop.py#L305). So, if event from kretprobe would be lost, it will never remove stored arguments from memory and over time such arguments could pile up. But since events are rarely lost and to be perceivable amount of lost events should be huge I don't think that's an issue. So, I wanted to ask if that's important, in that case I could add periodic GC to my implementation.

### DIFF
--- a/libbpf-tools/execsnoop.bpf.c
+++ b/libbpf-tools/execsnoop.bpf.c
@@ -1,0 +1,128 @@
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include "execsnoop.h"
+
+const volatile bool ignore_failed = true;
+const volatile uid_t targ_uid = INVALID_UID;
+const volatile int max_args = DEFAULT_MAXARGS;
+
+static const struct event empty_event = {};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 10240);
+	__type(key, pid_t);
+	__type(value, struct event);
+} execs SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(u32));
+	__uint(value_size, sizeof(u32));
+} events SEC(".maps");
+
+static __always_inline bool valid_uid(uid_t uid) {
+	return uid != INVALID_UID;
+}
+
+SEC("tracepoint/syscalls/sys_enter_execve")
+int tracepoint__syscalls__sys_enter_execve(struct trace_event_raw_sys_enter* ctx)
+{
+	u64 id;
+	pid_t pid, tgid;
+	unsigned int ret;
+	struct event *event;
+	struct task_struct *task;
+	const char **args = (const char **)(ctx->args[1]);
+	const char *argp;
+	uid_t uid = (u32)bpf_get_current_uid_gid();
+
+	if (valid_uid(targ_uid) && targ_uid != uid)
+		return 0;
+
+	id = bpf_get_current_pid_tgid();
+	pid = (pid_t)id;
+	tgid = id >> 32;
+	if (bpf_map_update_elem(&execs, &pid, &empty_event, BPF_NOEXIST))
+		return 0;
+
+	event = bpf_map_lookup_elem(&execs, &pid);
+	if (!event)
+		return 0;
+
+	event->pid = pid;
+	event->tgid = tgid;
+	event->uid = uid;
+	task = (struct task_struct*)bpf_get_current_task();
+	event->ppid = (pid_t)BPF_CORE_READ(task, real_parent, tgid);
+	event->args_count = 0;
+	event->args_size = 0;
+
+	ret = bpf_probe_read_str(event->args, ARGSIZE, (const char*)ctx->args[0]);
+	if (ret <= ARGSIZE) {
+		event->args_size += ret;
+	} else {
+		/* write an empty string */
+		event->args[0] = '\0';
+		event->args_size++;
+	}
+
+	event->args_count++;
+	#pragma unroll
+	for (int i = 1; i < TOTAL_MAX_ARGS && i < max_args; i++) {
+		bpf_probe_read(&argp, sizeof(argp), &args[i]);
+		if (!argp)
+			return 0;
+
+		if (event->args_size > LAST_ARG)
+			return 0;
+
+		ret = bpf_probe_read_str(&event->args[event->args_size], ARGSIZE, argp);
+		if (ret > ARGSIZE)
+			return 0;
+
+		event->args_count++;
+		event->args_size += ret;
+	}
+	/* try to read one more argument to check if there is one */
+	bpf_probe_read(&argp, sizeof(argp), &args[max_args]);
+	if (!argp)
+		return 0;
+
+	/* pointer to max_args+1 isn't null, asume we have more arguments */
+	event->args_count++;
+	return 0;
+}
+
+SEC("tracepoint/syscalls/sys_exit_execve")
+int tracepoint__syscalls__sys_exit_execve(struct trace_event_raw_sys_exit* ctx)
+{
+	u64 id;
+	pid_t pid;
+	int ret;
+	struct event *event;
+	u32 uid = (u32)bpf_get_current_uid_gid();
+
+	if (valid_uid(targ_uid) && targ_uid != uid)
+		return 0;
+	id = bpf_get_current_pid_tgid();
+	pid = (pid_t)id;
+	event = bpf_map_lookup_elem(&execs, &pid);
+	if (!event)
+		return 0;
+	ret = ctx->ret;
+	if (ignore_failed && ret < 0)
+		goto cleanup;
+
+	event->retval = ret;
+	bpf_get_current_comm(&event->comm, sizeof(event->comm));
+	size_t len = EVENT_SIZE(event);
+	if (len <= sizeof(*event))
+		bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, event, len);
+cleanup:
+	bpf_map_delete_elem(&execs, &pid);
+	return 0;
+}
+
+char LICENSE[] SEC("license") = "GPL";
--- a/libbpf-tools/execsnoop.c
+++ b/libbpf-tools/execsnoop.c
@@ -1,0 +1,340 @@
+// Based on execsnoop(8) from BCC by Brendan Gregg and others.
+//
+#include <argp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/resource.h>
+#include <sys/time.h>
+#include <time.h>
+#include <unistd.h>
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+#include "execsnoop.h"
+#include "execsnoop.skel.h"
+
+#define PERF_BUFFER_PAGES   64
+#define NSEC_PER_SEC 1000000000L
+#define NSEC_PRECISION (NSEC_PER_SEC / 1000)
+#define MAX_ARGS_KEY 259
+
+static struct env {
+	bool time;
+	bool timestamp;
+	bool fails;
+	uid_t uid;
+	bool quote;
+	const char *name;
+	const char *line;
+	bool print_uid;
+	bool verbose;
+	int max_args;
+} env = {
+	.max_args = DEFAULT_MAXARGS,
+	.uid = INVALID_UID
+};
+
+static struct timespec start_time;
+
+const char *argp_program_version = "execsnoop 0.1";
+const char *argp_program_bug_address = "<bpf@vger.kernel.org>";
+const char argp_program_doc[] =
+"Trace open family syscalls\n"
+"\n"
+"USAGE: execsnoop [-h] [-T] [-t] [-x] [-u UID] [-q] [-n NAME] [-l LINE] [-U]\n"
+"                 [--max-args MAX_ARGS]\n"
+"\n"
+"EXAMPLES:\n"
+"   ./execsnoop           # trace all exec() syscalls\n"
+"   ./execsnoop -x        # include failed exec()s\n"
+"   ./execsnoop -T        # include time (HH:MM:SS)\n"
+"   ./execsnoop -U        # include UID\n"
+"   ./execsnoop -u 1000   # only trace UID 1000\n"
+"   ./execsnoop -t        # include timestamps\n"
+"   ./execsnoop -q        # add \"quotemarks\" around arguments\n"
+"   ./execsnoop -n main   # only print command lines containing \"main\"\n"
+"   ./execsnoop -l tpkg   # only print command where arguments contains \"tpkg\"";
+
+static const struct argp_option opts[] = {
+	{ "time", 'T', NULL, 0, "include time column on output (HH:MM:SS)"},
+	{ "timestamp", 't', NULL, 0, "include timestamp on output"},
+	{ "fails", 'x', NULL, 0, "include failed exec()s"},
+	{ "uid", 'u', "UID", 0, "trace this UID only"},
+	{ "quote", 'q', NULL, 0, "Add quotemarks (\") around arguments"},
+	{ "name", 'n', "NAME", 0, "only print commands matching this name, any arg"},
+	{ "line", 'l', "LINE", 0, "only print commands where arg contains this line"},
+	{ "print-uid", 'U', NULL, 0, "print UID column"},
+	{ "max-args", MAX_ARGS_KEY, "MAX_ARGS", 0,
+		"maximum number of arguments parsed and displayed, defaults to 20"},
+	{ "verbose", 'v', NULL, 0, "Verbose debug output" },
+	{},
+};
+
+static error_t parse_arg(int key, char *arg, struct argp_state *state)
+{
+	long int uid, max_args;
+
+	switch (key) {
+	case 'h':
+		argp_usage(state);
+		break;
+	case 'T':
+		env.time = true;
+		break;
+	case 't':
+		env.timestamp = true;
+		break;
+	case 'x':
+		env.fails = true;
+		break;
+	case 'u':
+		errno = 0;
+		uid = strtol(arg, NULL, 10);
+		if (errno || uid < 0 || uid >= INVALID_UID) {
+			fprintf(stderr, "Invalid UID %s\n", arg);
+			argp_usage(state);
+		}
+		env.uid = uid;
+		break;
+	case 'q':
+		env.quote = true;
+		break;
+	case 'n':
+		env.name = arg;
+		break;
+	case 'l':
+		env.line = arg;
+		break;
+	case 'U':
+		env.print_uid = true;
+		break;
+	case 'v':
+		env.verbose = true;
+		break;
+	case MAX_ARGS_KEY:
+		errno = 0;
+		max_args = strtol(arg, NULL, 10);
+		if (errno || max_args < 1 || max_args > TOTAL_MAX_ARGS) {
+			fprintf(stderr, "Invalid MAX_ARGS %s, should be in [1, %d] range\n",
+					arg, TOTAL_MAX_ARGS);
+
+			argp_usage(state);
+		}
+		env.max_args = max_args;
+		break;
+	default:
+		return ARGP_ERR_UNKNOWN;
+	}
+	return 0;
+}
+
+static int bump_memlock_rlimit(void)
+{
+	struct rlimit rlim_new = {
+		.rlim_cur	= RLIM_INFINITY,
+		.rlim_max	= RLIM_INFINITY,
+	};
+
+	return setrlimit(RLIMIT_MEMLOCK, &rlim_new);
+}
+
+int libbpf_print_fn(enum libbpf_print_level level,
+            const char *format, va_list args)
+{
+	if (level == LIBBPF_DEBUG && !env.verbose)
+		return 0;
+	return vfprintf(stderr, format, args);
+}
+
+static void time_since_start()
+{
+	long nsec, sec;
+	static struct timespec cur_time;
+	double time_diff;
+
+	clock_gettime(CLOCK_MONOTONIC, &cur_time);
+	nsec = cur_time.tv_nsec - start_time.tv_nsec;
+	sec = cur_time.tv_sec - start_time.tv_sec;
+	if (nsec < 0) {
+		nsec += NSEC_PER_SEC;
+		sec--;
+	}
+	time_diff = sec + (double)nsec / NSEC_PER_SEC;
+	printf("%-8.3f", time_diff);
+}
+
+static void inline quoted_symbol(char c) {
+	switch(c) {
+		case '"':
+			putchar('\\');
+			putchar('"');
+			break;
+		case '\t':
+			putchar('\\');
+			putchar('t');
+			break;
+		case '\n':
+			putchar('\\');
+			putchar('n');
+			break;
+		default:
+			putchar(c);
+			break;
+	}
+}
+
+static void print_args(const struct event *e, bool quote)
+{
+	int args_counter = 0;
+
+	if (env.quote)
+		putchar('"');
+
+	for (int i = 0; i < e->args_size && args_counter < e->args_count; i++) {
+		char c = e->args[i];
+		if (env.quote) {
+			if (c == '\0') {
+				args_counter++;
+				putchar('"');
+				putchar(' ');
+				if (args_counter < e->args_count) {
+					putchar('"');
+				}
+			} else {
+				quoted_symbol(c);
+			}
+		} else {
+			if (c == '\0') {
+				args_counter++;
+				putchar(' ');
+			} else {
+				putchar(c);
+			}
+		}
+	}
+	if (e->args_count == env.max_args + 1) {
+		fputs(" ...", stdout);
+	}
+}
+
+void handle_event(void *ctx, int cpu, void *data, __u32 data_sz)
+{
+	const struct event *e = data;
+	time_t t;
+	struct tm *tm;
+	char ts[32];
+
+	/* TODO: use pcre lib */
+	if (env.name && strstr(e->comm, env.name) == NULL)
+		return;
+
+	/* TODO: use pcre lib */
+	if (env.line && strstr(e->comm, env.line) == NULL)
+		return;
+
+	time(&t);
+	tm = localtime(&t);
+	strftime(ts, sizeof(ts), "%H:%M:%S", tm);
+
+	if (env.time) {
+		printf("%-8s ", ts);
+	}
+	if (env.timestamp) {
+		time_since_start();
+	}
+
+	if (env.print_uid)
+		printf("%-6d", e->uid);
+
+	printf("%-16s %-6d %-6d %3d ", e->comm, e->pid, e->ppid, e->retval);
+	print_args(e, env.quote);
+	putchar('\n');
+}
+
+void handle_lost_events(void *ctx, int cpu, __u64 lost_cnt)
+{
+	fprintf(stderr, "Lost %llu events on CPU #%d!\n", lost_cnt, cpu);
+}
+
+int main(int argc, char **argv)
+{
+	static const struct argp argp = {
+		.options = opts,
+		.parser = parse_arg,
+		.doc = argp_program_doc,
+	};
+	struct perf_buffer_opts pb_opts;
+	struct perf_buffer *pb = NULL;
+	struct execsnoop_bpf *obj;
+	int err;
+
+	err = argp_parse(&argp, argc, argv, 0, NULL, NULL);
+	if (err)
+		return err;
+
+	libbpf_set_print(libbpf_print_fn);
+
+	err = bump_memlock_rlimit();
+	if (err) {
+		fprintf(stderr, "failed to increase rlimit: %d\n", err);
+		return 1;
+	}
+
+	obj = execsnoop_bpf__open();
+	if (!obj) {
+		fprintf(stderr, "failed to open and/or load BPF object\n");
+		return 1;
+	}
+
+	/* initialize global data (filtering options) */
+	obj->rodata->ignore_failed = !env.fails;
+	obj->rodata->targ_uid = env.uid;
+	obj->rodata->max_args = env.max_args;
+
+	err = execsnoop_bpf__load(obj);
+	if (err) {
+		fprintf(stderr, "failed to load BPF object: %d\n", err);
+		goto cleanup;
+	}
+
+	clock_gettime(CLOCK_MONOTONIC, &start_time);
+	err = execsnoop_bpf__attach(obj);
+	if (err) {
+		fprintf(stderr, "failed to attach BPF programs\n");
+		goto cleanup;
+	}
+	/* print headers */
+	if (env.time) {
+		printf("%-9s", "TIME");
+	}
+	if (env.timestamp) {
+		printf("%-8s ", "TIME(s)");
+	}
+	if (env.print_uid) {
+		printf("%-6s ", "UID");
+	}
+
+	printf("%-16s %-6s %-6s %3s %s\n", "PCOMM", "PID", "PPID", "RET", "ARGS");
+
+	/* setup event callbacks */
+	pb_opts.sample_cb = handle_event;
+	pb_opts.lost_cb = handle_lost_events;
+	pb = perf_buffer__new(bpf_map__fd(obj->maps.events), PERF_BUFFER_PAGES, &pb_opts);
+	err = libbpf_get_error(pb);
+	if (err) {
+		pb = NULL;
+		fprintf(stderr, "failed to open perf buffer: %d\n", err);
+		goto cleanup;
+	}
+
+	/* main: poll */
+	while ((err = perf_buffer__poll(pb, 100)) >= 0)
+		;
+	printf("Error polling perf buffer: %d\n", err);
+
+cleanup:
+	perf_buffer__free(pb);
+	execsnoop_bpf__destroy(obj);
+
+	return err != 0;
+}
--- a/libbpf-tools/.gitignore
+++ b/libbpf-tools/.gitignore
@@ -1,5 +1,6 @@
 /.output
 /drsnoop
+/execsnoop
 /filelife
 /opensnoop
 /runqslower
--- a/libbpf-tools/Makefile
+++ b/libbpf-tools/Makefile
@@ -9,7 +9,7 @@ INCLUDES := -I$(OUTPUT)
 CFLAGS := -g -O2 -Wall
 ARCH := $(shell uname -m | sed 's/x86_64/x86/')
 
-APPS = drsnoop filelife opensnoop runqslower vfsstat xfsslower
+APPS = drsnoop execsnoop filelife opensnoop runqslower vfsstat xfsslower
 
 .PHONY: all
 all: $(APPS)
--- a/libbpf-tools/execsnoop.h
+++ b/libbpf-tools/execsnoop.h
@@ -1,0 +1,27 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+#ifndef __EXECSNOOP_H
+#define __EXECSNOOP_H
+
+#define ARGSIZE  128
+#define TASK_COMM_LEN 16
+#define TOTAL_MAX_ARGS 60
+#define DEFAULT_MAXARGS 20
+#define FULL_MAX_ARGS_ARR (TOTAL_MAX_ARGS * ARGSIZE)
+#define INVALID_UID ((uid_t)-1)
+#define BASE_EVENT_SIZE (size_t)(&((struct event*)0)->args)
+#define EVENT_SIZE(e) (BASE_EVENT_SIZE + e->args_size)
+#define LAST_ARG (FULL_MAX_ARGS_ARR - ARGSIZE)
+
+struct event {
+	pid_t pid;
+	pid_t tgid;
+	pid_t ppid;
+	uid_t uid;
+	int retval;
+	int args_count;
+	unsigned int args_size;
+	char comm[TASK_COMM_LEN];
+	char args[FULL_MAX_ARGS_ARR];
+};
+
+#endif /* __EXECSNOOP_H */


anakryiko:
I think it would be best to actually change the way that BPF probe produces data. Instead of sending each individual parameter as a separate event, collect all of parameters (up to a maximum number, of course) into a single big event. Once done, send it as one whole sample. You can be smart about sending only minimum amount of filled data, not sizeof(sample) -- as a perf_buffer space usage optimization.
This has added benefit that you both don't need a hash table, as well there is no chance of losing some of arguments, while preserving other (due to perf_buffer overflow), which can be super confusing to users. Do you mind trying implementing that? 

So event will look something like:
```
struct event {
    pid_t pid;
    pid_t ppid;
    uid_t uid;
    int retval;
    char comm[TASK_COMM_LEN];
    u32 arg_lens[MAX_ARG_CNT];
    char args[MAX_ARGS_TOTAL_LEN];
};
```

All arguments will be just layed out sequentially in args array, while arg_lens will remember individual length (they are returned from bpf_probe_read_str() on success).

Does it make sense?

bacher09:
@anakryiko I actually started with something similar, but later decided that it's a bad idea, primarily because it forced to deal with arguments at both probes (at `sys_enter` and at `sys_exit`). So basically, arguments handled twice and it's extra overhead (relatively small, but still).
In addition, I had an idea to make something like this:
```c
#define TOTAL_MAXARGS 50
const volatile int maxargs = 20; // could be changed by CLI options
#pragma unroll
for (int i = 1; i < TOTAL_MAXARGS && i < maxargs; i++) {
    const char *argp = NULL;
    bpf_probe_read(&argp, sizeof(argp), &args[i]);

    if (!argp)
        goto out;

    bpf_probe_read_str(event.argv, sizeof(event.argv), (void *)argp);
    bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, &event, sizeof(event));
}
```
So, tool could have `--maxargs` options, as python version. And more arguments you process, more it adds extra overhead.

But, in the same time, your suggestion makes implementation much easier, especially considering that individual argument events could be lost as well (with my current implementation).

Also, wanted to ask you about `CONFIG_DEBUG_INFO_SPLIT` (split debug info)  and BTF. Currently, if you build kernel with `CONFIG_DEBUG_INFO_BTF` and `CONFIG_DEBUG_INFO_SPLIT` generated btf file would be invalid. Is it something that is planned to be fixed or did I missed something ?

bacher09:
Also, if I understood you suggestion correctly it will limit amount of max arguments that could be send somewhere below 512 bytes. Since, we need to assemble event on BPF stack, and it's limited by 512 bytes. So, basically your suggested `struct event` is smaller than 512 bytes, am I correct ?

And there is no such limitation, when using multiple events, there is only limitation on size of each argument. For example, current python implementation could intercept something like [this perfectly][echo_sample] without loosing a single argument.

[echo_sample]: https://paste.ubuntu.com/p/Xx2pWtT2P3/

brendangregg:
Please bear in mind that many BCC tools were first written for Linux kernel versions 4.0 to 4.3, which lacked many BPF capabilities today. There have been rounds of fixes, but as I found out when doing CO-RE opensnoop, the current state of these Python tools can be a mess and need rewriting. I did so for opensnoop, but haven't gone back and updated the Python version yet.

execsnoop is another example: this should be overhauled given BPF functionality of today. The 512-byte stack storage limit today is solved by moving storage to BPF maps: execsnoop can create a key/value for the arguments where the key is in the perf output, and then the user-space tool reads the map value and then deletes the key. Maps are unlimited.

I wouldn't encourage anyone new to BPF to do the CO-RE tool porting, since it's far more work than it may look, and may well be discouraging.

As for execsnoop arguments: yes, tools should match the Python version if at all possible, so that they can share the same man page.

anakryiko:
@bacher09, actually combination of what using single bigger sample for all arguments and using BPF hash map to maintain it across entry/exit would probably make user-space part the simplest.

It will allow very small perf_buffer (just notifying that exec is done), won't require user-space hashmap look-ups (you can just read it from BPF_MAP, when you get perf_buffer sample with pid). In general, people get around 512 byte limitation by using per-cpu array with single element. That way you sort of have your local head :)

As for CONFIG_DEBUG_INFO_SPLIT, I wasn't aware and I never used CONFIG_DEBUG_INFO_SPLIT. Please submit Kconfig bug fix disabling CONFIG_DEBUG_INFO_SPLIT if CONFIG_DEBUG_INFO_BTF is enabled. Thanks!

bacher09:
@anakryiko also wanted to ask you more about 

> sending only minimum amount of filled data, not sizeof(sample)

Do you mean using  calculated size in `bpf_perf_event_output` ?

anakryiko:
@bacher09, right, keep track of actual amount of data and pass that to perf_event_output(). It used to be a bit painful with older kernels, but for those supporting kernel BTF verifier is sophisticated enough to not cause any problems with this.

bacher09:
@anakryiko please check if you have time.

Also, wanted to ask regarding `execveat`, currently python implementation don't handle this syscall. So I wanted to ask if I need to implement it in my PR, or I can add it as another PR.

libbpf-tools/execsnoop.bpf.c
```diff
@@ -0,0 +1,142 @@
...
+	event->uid = uid;
+	event->retval = ret;
+
+	task = (struct task_struct*)bpf_get_current_task();
+	bpf_probe_read(&parent, sizeof(parent), &task->real_parent);
+	bpf_probe_read(&event->ppid, sizeof(pid_t), &parent->tgid);
+	bpf_get_current_comm(&event->comm, sizeof(event->comm));
+
+	/* trick to satisfy verifier */
+	#pragma unroll
```
bacher09: Had to do this trick to satisfy verifier, after loop unrolling size argument (last argument) should be constant.

libbpf-tools/execsnoop.c
```diff
@@ -0,0 +1,338 @@
...
+        int j = 0;
+		if (i > 0) {
+			putchar(' ');
+		}
+		/* arguments should be quoted */
+		if (env.quote) {
+			putchar('"');
+			while(*c && j < ARGSIZE) {
+				switch(*c) {
+					case '"':
```
bacher09: Python version [handle only](https://github.com/iovisor/bcc/blob/master/tools/execsnoop.py#L288) `"`, but I think it's better if quote behavior will be more like python `repr()`.

libbpf-tools/execsnoop.c
```diff
@@ -0,0 +1,338 @@
...
+	if (env.timestamp) {
+		time_since_start(time_since, sizeof(time_since));
+		printf("%-8s ", time_since);
+	}
+
+    if (env.print_uid)
+        printf("%-6d", e->uid);
+
+    printf("%-16s %-6d %-6d %3d ", e->comm, e->pid, e->ppid, e->retval);
+	print_args(e, env.quote);
```
bacher09: Sorry, for inconsistency with tabs and spaces, at first I started with spaces and then when noticed that all other tools in `libbpf-tools` folder using tabs reconfigured editor to tabs.
I'll **fix this** with next update.

libbpf-tools/execsnoop.c
```diff
@@ -0,0 +1,338 @@
...
+};
+
+static struct timespec start_time;
+
+const char *argp_program_version = "execsnoop 0.1";
+const char *argp_program_bug_address = "<bpf@vger.kernel.org>";
+const char argp_program_doc[] =
+"Trace open family syscalls\n"
+"\n"
+"USAGE: execsnoop [-h] [-T] [-t] [-x] [-u UID] [-q] [-n NAME] [-l LINE] [-U]\n"
```
bacher09: All arguments should be the same as python `execsnoop` with one exception `--cgroupmap` (`opensnoop` also doesn't implement this option).

libbpf-tools/execsnoop.c
```diff
@@ -0,0 +1,338 @@
...
+
+    clock_gettime(CLOCK_MONOTONIC, &cur_time);
+	nsec = cur_time.tv_nsec - start_time.tv_nsec;
+	sec = cur_time.tv_sec - start_time.tv_sec;
+	if (nsec < 0) {
+		nsec += NSEC_PER_SEC;
+		sec--;
+	}
+	/* rounded division */
+	nsec = (nsec + (NSEC_PRECISION / 2)) / NSEC_PRECISION;
```
bacher09: Some projects have policy regarding floating-point arithmetic (including userspace ones), so I made this without any floating arithmetic. It's tries to imitate python's `time.time()` [behavior](https://github.com/iovisor/bcc/blob/master/tools/execsnoop.py#L296).

libbpf-tools/execsnoop.bpf.c
```diff
@@ -0,0 +1,142 @@
...
+	if (ret < 0)
+		goto out;
+
+	event->args_count++;
+	const char **args = (const char **)(ctx->args[1]);
+
+	#pragma unroll
+	for (int i = 1; i < TOTAL_MAX_ARGS && i < max_args; i++) {
+		const char *argp = NULL;
+		char *argv_write = &(event->args[ARGSIZE * i]);
```
bacher09: I also have tried to pack arguments in more compact format, but verifier doesn't liked it. It seems `dst` argument should be known at compile time.

libbpf-tools/execsnoop.c
```diff
@@ -0,0 +1,338 @@
...
+	long nsec, sec;
+	static struct timespec cur_time;
+
+	clock_gettime(CLOCK_MONOTONIC, &cur_time);
+	nsec = cur_time.tv_nsec - start_time.tv_nsec;
+	sec = cur_time.tv_sec - start_time.tv_sec;
+	if (nsec < 0) {
+		nsec += NSEC_PER_SEC;
+		sec--;
+	}
```
anakryiko: why so complicated?

```
nsec = cur_time.tv_sec * 1000000000ULL + cur_time.tv_nsec;
```
store start_time in nanoseconds as well, make your life easier

libbpf-tools/execsnoop.c
anakryiko: this is so overengineered and still wrong....
```
snprintf(str, size, "%ld.%09ld", nsec / 1000000000, nsec % 1000000000);
```

libbpf-tools/execsnoop.c
anakryiko: extract this into function, please

libbpf-tools/execsnoop.c
```diff
@@ -0,0 +1,338 @@
...
+	time(&t);
+	tm = localtime(&t);
+	strftime(ts, sizeof(ts), "%H:%M:%S", tm);
+
+	if (env.time) {
+		printf("%-8s ", ts);
+	}
+	if (env.timestamp) {
+		time_since_start(time_since, sizeof(time_since));
+		printf("%-8s ", time_since);
```
anakryiko: just print it here without intermediate snprintf? probably limiting to millisecond precision is good enough (but I haven't checked what Python opensnoop does)

libbpf-tools/execsnoop.bpf.c
```diff
@@ -0,0 +1,142 @@
...
+
+	task = (struct task_struct*)bpf_get_current_task();
+	bpf_probe_read(&parent, sizeof(parent), &task->real_parent);
+	bpf_probe_read(&event->ppid, sizeof(pid_t), &parent->tgid);
+	bpf_get_current_comm(&event->comm, sizeof(event->comm));
+
+	/* trick to satisfy verifier */
+	#pragma unroll
+	for (int i = 0; i <= TOTAL_MAX_ARGS && i <= max_args; i++) {
+		if (i == event->args_count)
```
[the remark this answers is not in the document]
bacher09: In that case verifier rejects code. I mentioned this in previous version of code.
Basically after loop is unrolled generated code should be something like this:

```c
if (args_count == 0)
    bpf_perf_event_output(...); // size of 0 arguments
else if (args_count == 1)
    bpf_perf_event_output(...); // size of 1 arguments
```
In that case verifier accepts code.

libbpf-tools/execsnoop.bpf.c
```diff
@@ -0,0 +1,142 @@
...
+#define NULL 0
+// from linux/bpf.h
+#define BPF_NOEXIST  1
+
+#define BASE_EVENT_SIZE (sizeof(struct event) - FULL_MAX_ARGS_ARR)
+#define EVENT_SIZE(X) (BASE_EVENT_SIZE + (ARGSIZE * (X)->args_count))
+
+const volatile bool ignore_failed = true;
+const volatile uid_t targ_uid = INVALID_UID;
+const volatile int max_args = DEFAULT_MAXARGS;
```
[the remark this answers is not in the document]
bacher09: Python version allows this to be adjustable (`--max-args` argument). Although  it's changes [defined value](https://github.com/iovisor/bcc/blob/master/tools/execsnoop.py#L219).

libbpf-tools/execsnoop.bpf.c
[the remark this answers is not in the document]
anakryiko: you don't need loop:

```
size_t final_size = event->args_count * 128;
if (final_size >= sizeof(struct event))
    /* can't happen */
    return 0;
bpf_perf_event_output(...,...,&event, final_size);
```

libbpf-tools/execsnoop.bpf.c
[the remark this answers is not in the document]
anakryiko: alright, then just stick to max_args, no need for TOTAL_MAX_ARGS (as I mentioned, recent enough verifier will know that max_args is a constant). Verifier also supports proper loop, so you don't need #pragma unroll either (in fact #pragma unroll won't work with const max_args).

libbpf-tools/execsnoop.c
bacher09: In that case `nsec` should be `long long` and not just `long`. I tried to avoid this.

libbpf-tools/execsnoop.c
bacher09: This will produce incorrect result (in terms in imitating python's script behavior). Current script rounds result at [3 digits](https://github.com/iovisor/bcc/blob/master/tools/execsnoop.py#L296).

Also, why it's wrong, I tested it and results was the same as produced by python (for same time value).

bacher09: Basically after this operation there will be only 3 digits (rounded milliseconds). So something like `0.1225` will become `0.123`.

Regarding overengineering agree, but this [isn't mine recipe](https://github.com/torvalds/linux/blob/master/drivers/net/ethernet/sfc/ptp.c#L464). Also I have seen it in few popular game engines as well.

libbpf-tools/execsnoop.bpf.c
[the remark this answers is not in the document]
bacher09: This doesn't work, verifier rejects this.

Also, I tried something like this previously (and result was same).

```c
size_t final_size = event->args_count * 128;                                
if (final_size >= sizeof(struct event))                                     
   return 0;                                                               
                                                                                 
bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, event, final_size); 
```

Results in:

```
from 33 to 102: safe                                                                                                        

from 23 to 102: R0=inv(id=0) R1=inv(id=0,umax_value=4294967295,var_off=(0x0; 0xffffffff)) R2=inv(id=0,umax_value=4294967295,var_off=$0x0; 0xffffffff)) R3=inv4294967295 R6=ctx(id=0,off=0,imm=0) R8=inv(id=0) R9=inv(id=0) R10=fp0
; bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, event,
102: (bf) r2 = r10
; 
103: (07) r2 += -4
; bpf_map_delete_elem(&stored, &pid);
104: (18) r1 = 0xffff8fc34af4f800 
106: (85) call bpf_map_delete_elem#3
invalid indirect read from stack off -4+0 size 4
processed 108 insns (limit 1000000) max_states_per_insn 1 total_states 7 peak_states 7 mark_read 5
```

libbpf-tools/execsnoop.bpf.c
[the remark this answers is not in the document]
bacher09: It's seems in **5.4** it isn't recent enough, cause without this it's rejects code.

libbpf-tools/execsnoop.c
anakryiko: It's wrong because for 1'000'000'009 it will print 1.9 seconds, which clearly is not 1.000000009, isn't it?
Not sure why you need rounding for nanoseconds, you can't measure half a nanosecond... Either way, I'd just stick to %lf if rounding was important.

libbpf-tools/execsnoop.c
anakryiko: sounds good, let's keep formatting nice and consistent

libbpf-tools/execsnoop.c
bacher09: Don't think this function worth arguing and I mostly agree with you.

> 1'000'000'009 it will print 1.9 seconds
nope it will print `0.1000`, but it still isn't correct (cause correct would be 1.000)

But, `tv_nsec` couldn't be 1'000'000'009, mainly because it's larger than one second and `tv_nsec` is truncated to be less than 1 second, at least that's how `struct timespec` supposed to work if I recall it correctly.
And for `.tv_sec =1, .tv_nsec = 9` it will print `1.0` (which still isn't fully correct, correct one should be `1.000`)

I made small [sample program](https://gist.github.com/bacher09/a1005db40fb5c4a5310a86d20122c66a) that demonstrates this.
On my system it has following output:

```bash
1.0            # .tv_sec = 1, .tv_nsec = 9
0.1000         # .tv_sec = 0, .tv_nsec = 1_000_000_009
0.112          # .tv_sec = 0, .tv_nsec = 111500000
0.112          # .tv_sec = 0, .tv_nsec = 111550000
2243 259328533 # sample data from struct timespec
```

So yeah, in some cases result might be incorrect, but it's primarily a formatting issue (`1.0` instead of `1.000`).

bacher09: Also wanted to add that tcpdump has almost [same approach](https://github.com/the-tcpdump-group/tcpdump/blob/master/timeval-operations.h#L67) and here's usage [example](https://github.com/the-tcpdump-group/tcpdump/blob/master/util-print.c#L358). And it's same for [strace](https://github.com/strace/strace/blob/master/util.c#L119) as well, but for formating it using [floating values](https://github.com/strace/strace/blob/master/util.c#L104).

Also, I checked python sources as well and there approach is the same as strace's. Basically this recipe is very common in OS software for dealing with time.

anakryiko: I'm pointing out that you are not zero-padding milliseconds properly and that prints completely incorrect duration. I's not about calculation, it's about formatting. If you have 10'000'000 nanoseconds (10 milliseconds), you'll print out 0.10 (i.e., 100ms), but should print 0.010 (10ms). So in your previous code you had to use "%ld.%03ld" format string. That's all. Rounding algorithm seems ok, I just wouldn't bother with it, but that's beside the point.

bacher09: Yeah, I agree with this. I changed it to be like python and strace implementations.

anakryiko:
@bacher09, do you still plan to work on this?

bacher09:
Sorry I was little bit distracted and also wanted to get answer regarding support of old kernels. Yeah, I plan to work on this on a weekend.

anakryiko:
@bacher09, great, thanks! What was the question about old kernels?

libbpf-tools/execsnoop.bpf.c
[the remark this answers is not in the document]
bacher09: > right, I forgot execsnoop doesn't really need any of CO-RE stuff and can run on much older kernels. This tool can be made to run even on kernels that don't have BTF. You'll need to drop vmlinux.h and just copy/paste trace_event_raw_sys_enter/trace_event_raw_sys_exit struct definitions. It will make this tool quite widely runnable. Not sure how important that is, but something to consider for sure. cc @brendangregg

Basically what kernel version should I target, when I was working on this I was targeting latest LTS (which is `5.4`), but you mentioned that I could target even older kernel version, so question is should I ? Or maybe I should target something newer than `5.4` (because it had some issues with verifier).

bacher09:
This [one](https://github.com/iovisor/bcc/pull/2807#discussion_r421693213) . Basically what kernel version should I target.

bacher09:
@anakryiko please take a look, also this doesn't work anymore on **5.4**, but works fine on **5.6** and newer kernels.

libbpf-tools/execsnoop.bpf.c
```diff
@@ -0,0 +1,129 @@
...
+
+SEC("tracepoint/syscalls/sys_enter_execve")
+int tracepoint__syscalls__sys_enter_execve(struct trace_event_raw_sys_enter* ctx)
+{
+	u64 id;
+	pid_t pid, tgid;
+	int ret;
+	struct event *event;
+	uid_t uid = (u32)bpf_get_current_uid_gid();
+	if (valid_uid(targ_uid) && targ_uid != uid)
```
anakryiko, line 40: nit: kernel code style (which BPF code in libbpf-tools follows) dictates that variable declaration and code should be separated by an empty line, can you please add it before this if? thanks!

libbpf-tools/execsnoop.bpf.c
```diff
@@ -0,0 +1,129 @@
...
+
+	event->pid = pid;
+	event->tgid = tgid;
+	event->uid = uid;
+	event->more_args = false;
+	event->args_count = 0;
+	event->args_size = 0;
+
+	ret = bpf_probe_read_str(event->args, ARGSIZE, (const char*)ctx->args[0]);
+	if (ret > 0 && ret <= ARGSIZE) {
```
anakryiko, line 61: it's simpler to declare ret as unsigned integer, and then this if will turn into:
```
if (ret <= ARGSIZE) {
    event->args_size += ret;
}
```

libbpf-tools/execsnoop.bpf.c
```diff
@@ -0,0 +1,129 @@
...
+	event = bpf_map_lookup_elem(&execs, &pid);
+	if (!event)
+		return 0;
+
+	ret = ctx->ret;
+	if (ignore_failed && ret < 0)
+		goto cleanup;
+
+	event->pid = pid;
+	event->uid = uid;
```
anakryiko, line 116: you already filled this out in kprobe, no need to do it again?

libbpf-tools/execsnoop.bpf.c
```diff
@@ -0,0 +1,129 @@
...
+	event->retval = ret;
+	task = (struct task_struct*)bpf_get_current_task();
+	event->ppid = (pid_t)BPF_CORE_READ(task, real_parent, tgid);
```
anakryiko, line 119: this could be done in kprobe, couldn't it? why splitting uid/pid/tgid in kprobe and ppid in kretprobe?

libbpf-tools/execsnoop.bpf.c
```diff
@@ -0,0 +1,129 @@
...
+	const char **args = (const char **)(ctx->args[1]);
+	for (int i = 1; i < max_args; i++) {
+		const char *argp = NULL;
+
+		bpf_probe_read(&argp, sizeof(argp), &args[i]);
+
+		if (!argp)
+			return 0;
+
+		if (event->args_size < 0 || event->args_size > ARGS_END)
```
anakryiko, line 79: same as suggested above, make args_size unsigned and only check against ARGS_END


Also I think ARGS_END should equal to FULL_MAX_ARGS_ARR, because `i` starts at 1 and `ctx->args[0]` is treated as a first argument (and is counted against same max_args limit). This might be why 5.4 verifier is complaining.

libbpf-tools/execsnoop.bpf.c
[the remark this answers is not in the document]
anakryiko: @bacher09, it seems like you ignored most of the important details I showed in https://github.com/anakryiko/bcc/commit/cda6c3411787504bc7c1dcf8af419a6e86b923c3. Wasn't that demo working for 5.4? See size_t types for len, `strs` variable to keep track of end of argument strings, negative arg_cnt value to mark `had_more_args` fact without wasting 4 bytes in the event struct, there is also a nice trick with reusing return result field between kprobe/kretprobe to store actual full length of the data.

So please do check it again and please let me know if that approach still works on 5.4? If yes, just try to follow the spirit of it and it should work for execsnoop here as well. If not, please paste verifier log showing verification failure on 5.4, we'll need to figure out what's wrong. This whole approach should work on all kernels that have BTF built-in. Don't remember if it's 5.4, 5.2 or whichever, we can check later precisely, but I'm sure 5.4 should be supported.

libbpf-tools/execsnoop.c
```diff
@@ -0,0 +1,340 @@
...
+	pb_opts.lost_cb = handle_lost_events;
+	pb = perf_buffer__new(bpf_map__fd(obj->maps.events), PERF_BUFFER_PAGES, &pb_opts);
+	err = libbpf_get_error(pb);
+	if (err) {
+		pb = NULL;
+		fprintf(stderr, "failed to open perf buffer: %d\n", err);
+		goto cleanup;
+	}
+
+	clock_gettime(CLOCK_MONOTONIC, &start_time);
```
anakryiko, line 329: it's better to record start_time before you do `execsnoop_bpf__attach`, so that your sample can never "go back in time".

libbpf-tools/execsnoop.bpf.c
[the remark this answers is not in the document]
bacher09: `Wasn't that demo working for 5.4?`

Exactly it wasn't, I even wrote a [comment about this](https://github.com/anakryiko/bcc/commit/f001f8cadf69c6beee10476beb34ceee76b284c1#r37837678).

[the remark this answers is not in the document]
anakryiko: Hm.. you must have missed that I actually updated first revision. See https://github.com/anakryiko/bcc/commit/f001f8cadf69c6beee10476beb34ceee76b284c1#r37842380.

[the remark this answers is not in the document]
bacher09: Regarding `more_args` I added to make code more readable, as I explained in another PR and it actually redundant (cause `more_args` true only when `args_count == max_args`).

libbpf-tools/execsnoop.bpf.c
bacher09: IMO, it's very hacky, also it doesn't check that ret isn't equal to `0` (this for some reason had issues with verifier).

libbpf-tools/execsnoop.bpf.c
bacher09: `Also I think ARGS_END should equal to FULL_MAX_ARGS_ARR,`

This would be incorrect and in that case verifier reject code. `ARGS_END` is really a wrong name for such constant, it's actually an offset that has at enough space for at least one argument (there is call `bpf_probe_read_str` after this check passes).

libbpf-tools/execsnoop.bpf.c
```diff
@@ -0,0 +1,142 @@
...
+			goto out;
+
+		ret = bpf_probe_read_str(argv_write, ARGSIZE, (void *)argp);
+		if (ret < 0)
+			goto out;
+
+		event->args_count++;
+	}
+
+	event->more_args = true;
```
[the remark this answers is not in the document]
bacher09: As I said I added it for readability reasons, I don't think being stingy for few bytes is really worth it. But if you insist I'll remove it.

libbpf-tools/execsnoop.bpf.c
anakryiko: It's a very common practice in BPF programs, it can generate simpler and more performance BPF code as well, as well as the one that's easier to verify for verifier. So I'd say still do it. You can leave comment that negative return will turn into a large unsigned ret value.

bpf_probe_read_str() can't return zero if you pass size >0. What kind of problem does it cause for verifier?

libbpf-tools/execsnoop.bpf.c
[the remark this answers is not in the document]
anakryiko: Doing something like below on user-space side:
```
bool more_args = false;

if (event->args_count < 0) {
    event->args_count = -event->args_count;
    more_args = true;
}
```

isn't readable enough?

As for `args_count == max_args` wouldn't it be the case when actual number of args is exactly equal to max_args?

libbpf-tools/execsnoop.bpf.c
anakryiko: Ah, right, yeah, the name is definitely misleading. You could do `event->args_size + ARGSIZE > ARGS_END` and compiler should optimize it to same thing, I think.

libbpf-tools/execsnoop.bpf.c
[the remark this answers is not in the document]
anakryiko: this was comment back from March :) but this bool with 3 bytes of padding just feels sloppy. But if you think it's more readable, I'm fine either way.

libbpf-tools/execsnoop.c
```diff
@@ -0,0 +1,340 @@
...
+
+	libbpf_set_print(libbpf_print_fn);
+
+	err = bump_memlock_rlimit();
+	if (err) {
+		fprintf(stderr, "failed to increase rlimit: %d\n", err);
+		return 1;
+	}
+
+	clock_gettime(CLOCK_MONOTONIC, &start_time);
```
anakryiko, line 283: no-no, not before *open*, not before *load* -- before **attach**, otherwise you might get a noticeable skew

libbpf-tools/execsnoop.bpf.c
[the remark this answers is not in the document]
bacher09: Oops sorry, updated version indeed works on **5.4**, but my version also works on **5.4** if I change `for` loop in enter probe to be unrolled (like this was previously done).

libbpf-tools/execsnoop.bpf.c
```diff
@@ -0,0 +1,123 @@
...
+	if (!event)
+		return 0;
+	ret = ctx->ret;
+	if (ignore_failed && ret < 0)
+		goto cleanup;
+
+	event->retval = ret;
+	bpf_get_current_comm(&event->comm, sizeof(event->comm));
+	size_t len = EVENT_SIZE(event);
+	if (len < sizeof(*event))
```
anakryiko, line 116: len <= sizeof?

bacher09: Yeah, I'll update it in a moment.

libbpf-tools/execsnoop.bpf.c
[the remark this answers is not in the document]
bacher09: So wanted to ask, should I return back `pragma unroll` before for loop to support **5.4** ?

Also wanted to point it out that currently `Makefile` incorrectly tracks dependencies (some headers isn't tracked as dependency). For example if I change `execsnoop.h` and do `make execsnoop` make will output `make: 'execsnoop' is up to date.`, while it isn't correct. And it's same behaviour for all other tools.

And also another question, is any plans to support cmake build, cause I did small sample how  it [could look like](https://gist.github.com/bacher09/beef9131ceece01c6980908a5ca5a560#file-cmakelists-txt-L84) (it just a sample and doesn't handle all the tools). And in that case there would be no need to update `.gitignore` for each tool, in addition it would be possible to make tools that would use multiple bpf sources (if this desirable).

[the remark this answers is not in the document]
anakryiko: Yeah, I'd add `pragma unroll back`.

As for Makefile not tracking depenencies. Yeah, seems like $(wildcard %.h) doesn't record dependency properly. Adding plain %.h will break on tools that don't have corresponding .h header... Not sure, need to find some other way. Thanks for bringing this up.

As for cmake, I have almost 0 experience with cmake, so didn't even attempt to write one. I actually like Makefile for it's simplicity and straightforwardness (once you learn Makefile basics, of course). cmake has too much magic for me. But if someone were to write a good cmake for all libbpf-tools that are easy to extend, that would be great, of course.

libbpf-tools/execsnoop.bpf.c
bacher09: Odd, now it works.

bacher09:
Now it works on **5.4**.

libbpf-tools/execsnoop.bpf.c
```diff
@@ -0,0 +1,124 @@
...
+
+		bpf_probe_read(&argp, sizeof(argp), &args[i]);
+		if (!argp)
+			return 0;
+
+		if (event->args_size > LAST_ARG)
+			return 0;
+
+		ret = bpf_probe_read_str(&event->args[event->args_size], ARGSIZE, argp);
+		if (ret < 0)
```
anakryiko, line 86: ret is unsigned, can't happen, I think you meant `> ARGSIZE`?

libbpf-tools/execsnoop.bpf.c
```diff
@@ -0,0 +1,124 @@
...
+	if (ret <= ARGSIZE) {
+		event->args_size += ret;
+	} else {
+		/* write an empty string */
+		event->args[0] = '\0';
+		event->args_size++;
+	}
+
+	event->args_count++;
+	const char **args = (const char **)(ctx->args[1]);
```
anakryiko, line 73: please declare args variable up top, along other local vars

libbpf-tools/execsnoop.bpf.c
```diff
@@ -0,0 +1,124 @@
...
+
+SEC("tracepoint/syscalls/sys_exit_execve")
+int tracepoint__syscalls__sys_exit_execve(struct trace_event_raw_sys_exit* ctx)
+{
+	u64 id;
+	pid_t pid;
+	int ret;
+	struct event *event;
+	u32 uid = (u32)bpf_get_current_uid_gid();
+	if (valid_uid(targ_uid) && targ_uid != uid)
```
anakryiko, line 103: empty line between variables block and code, please

libbpf-tools/execsnoop.c
```diff
@@ -0,0 +1,340 @@
...
+		} else {
+			if (c == '\0') {
+				args_counter++;
+				putchar(' ');
+			} else {
+				putchar(c);
+			}
+		}
+	}
+	if (e->args_count == env.max_args) {
```
anakryiko, line 215: let me ask again...

> As for args_count == max_args wouldn't it be the case when actual number of args is exactly equal to max_args?

Am I missing why `args_count == max_args` is sufficient to deduce " ..." part?

bacher09: It is the same behavior as `more_args` after loop and same as original [execsnoop](https://github.com/iovisor/bcc/blob/master/tools/execsnoop.py#L181) implementation, and same behavior as [your sample](https://github.com/anakryiko/bcc/commit/cda6c3411787504bc7c1dcf8af419a6e86b923c3#diff-ee7df97255c38d4d9f9fa421b1922f9eR154).

Original execsnoop doesn't distinguish results from `args > max_args` and `args >= max_args`. Here's an example:

   #script1
   /bin/echo 1 2 3 4
   /bin/echo 1 2 3

   $ ./execsnoop.py --max-args=4
    echo             7730   7487     0 /bin/echo 1 2 3 ...
    echo             7731   7487     0 /bin/echo 1 2 3 ...

If another behavior is intended (that distinguish between `args == max_args` and `args > max_args`) in that case implementation should read one argument more (beyond `max_args`). But for such argument implementation could use much smaller read size than `ARGSIZE`.

bacher09: Also, I want to mention that this also applies for argument size too, cause currently all execsnoop implementation doesn't indicated that argument was truncated and the only indication of such possibility is argument that has length of `ARGSIZE`.

libbpf-tools/execsnoop.bpf.c
bacher09: Sorry, I forgot to change this after I changed it to be `unsigned`.

libbpf-tools/execsnoop.c
anakryiko: Right, so existing execsnoop has a bug, and the answer is to preserve it instead of fixing it? Especially given that implementation is very simple (btw, NULL initialization of argp is not necessary):
```
for (int i = 1; i <= TOTAL_MAX_ARGS; i++) {
        /* NULL initialization is not needed */
		const char *argp;

		bpf_probe_read(&argp, sizeof(argp), &args[i]);
		if (!argp)
			return 0;

		if (i == max_args) {
		    event->args_count = -event->args_count;
		    /* or whatever way you insist on */
		    return 0;
		}

		if (event->args_size > LAST_ARG)
		    return 0;

		ret = bpf_probe_read_str(&event->args[event->args_size], ARGSIZE, argp);
		if (ret > ARGSIZE)
			return 0;

		event->args_count++;
		event->args_size += ret;
}
```

My demo already handled such case, btw, if you didn't like that, I'd appreciate you mentioning that, at least, instead of me having to do few more rounds of comments to get you to reply to my concern.

And yes, argument truncation is not indicated in any way, and I think it would be nice if both execsnoop variants would do that. But that can be done separately from this PR.

bacher09: Your demo have exactly same behavior (I tested before writing), and this behavior isn't a bug.

bacher09: Here's an output from your sample:
```
[15826/15826] bash (ret=0), filename /bin/echo, 30 args:/bin/echo,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25,2
6,27,28,29
[15827/15827] bash (ret=0), filename /bin/echo, 30 args:/bin/echo,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25,2
6,27,28,29
```
For: 
```
/bin/echo 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24 25 26 27 28 29
1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24 25 26 27 28 29
$ /bin/echo 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24 25 26 27 28 29 30
1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24 25 26 27 28 29 30
```

As I [mentioned](https://github.com/iovisor/bcc/pull/2807#discussion_r425928238) that same behavior was in [your sample](https://github.com/anakryiko/bcc/commit/cda6c3411787504bc7c1dcf8af419a6e86b923c3#diff-ee7df97255c38d4d9f9fa421b1922f9eR154), moreover you did exactly same comparison (only in bpf code) `if (i == MAX_ARG_CNT)` but for some reason if it succeeded you negated it's value (which could be done in userspace and would achieve same result).

bacher09: Moreover, your sample had extra bug (when `MAX_ARG_CNT` isn't equal to `max_args`).

bacher09:
```
if (i == max_args) {
	event->args_count = -event->args_count;
	/* or whatever way you insist on */
	return 0;
}
```
Again, this code has same **bug**, cause it just checks that `i == max_args` which could be done in userspace and would have same result. As I wrote before:

> in that case implementation should **read one argument more** (**beyond** max_args). But for such argument implementation could use **much smaller read size** than ARGSIZE.

>  (btw, NULL initialization of argp is not necessary)

Can you explain why ?

bacher09:
I updated code to a new behavior (which is different from original `execsnoop` regarding `...`).

libbpf-tools/execsnoop.c
anakryiko: > Your demo have exactly same behavior (I tested before writing), and this behavior isn't a bug.

Yes, good catch, my demo did have those two bugs, but its intent was to handle this correctly (otherwise, as you mentioned, I would do that check in user-space). I think the loop I posted above handles that correctly, unless there is a bug again.

And I do think it's a bug to show extra "...", if there are no more arguments. It's super misleading.

> Again, this code has same bug, cause it just checks that i == max_args which could be done in userspace (which is done in this PR) and would have same result. As I wrote before:

I don't think so. You'll get to this comparison only if `argp != NULL`, which means there was supposed to be another argument, but we are not going to read it, because we are beyond our limit of arguments to read. Say, if you wanted 5 arguments and program had 5 arguments. Algorithm would go on 6th iteration, would see argp == NULL and would correctly conclude that there is no 6th argument. If it had 6+ arguments, we'd read argp, it would not be NULL, we wouldn't attempt to read its contents, but we'd know that there is 6+ arguments (that's where we do negation, extra bit, or flag, whatever). Makes sense? Did I miss anything here?

>> (btw, NULL initialization of argp is not necessary)
> Can you explain why ?

yes, because bpf_probe_read() guarantees that destination (i.e., argp) will be either NULL on error, or will contain correct value (which still can be NULL, but code ignores bpf_probe_read failures, so we'd assume there was no argument, which also sucks, but that's for another time).

bacher09: Regarding misleading behavior, currently if there is an argument that is bigger than `ARGSIZE -1` (which is `127`) it will be truncated and there will no any indication about that, which IMO is also misleading. For instance:

    /bin/echo abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789   # len(arg1) == 186
    # result from original execsnoop
    echo             25603  7487     0 /bin/echo abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789abcdefghijklmnopqrstuvwxyz
ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789abcd

> yes, because bpf_probe_read() guarantees that destination (i.e., argp) will be either NULL on error, or will contain correct value

Thanks, I didn't know that, in that case I'll remove NULL and update PR shortly.

yonghong-song:
[buildbot, test this please]

yonghong-song:
There is conflict with master branch, could you resolve conflict and resubmit?

bacher09:
Done.

brendangregg:
Why was TOTAL_MAX_ARGS 60? What's the limit we're running into?

bacher09:
It's pretty random number, but it has to be limited at something (if we want to target `5.4` kernel, also because struct has to be limited too), original tool has no limit cause it compiles [it at runtime](https://github.com/iovisor/bcc/blob/master/tools/execsnoop.py#L219).

Same actually applies for other constants in original execsnoop, for example `ARGSIZE` which is limited by [128](https://github.com/iovisor/bcc/blob/master/tools/execsnoop.py#L99) (if I had to pick this limit I would probably go with 256 or 512, cause sometimes long path argument could be truncated).

Moreover, the bigger `TOTAL_MAX_ARGS` would be the bigger eBPF code would be (cause loop is unrolled, but this could be dropped if we not going to support **5.4**). Regarding true limits, it works just fine with **2739**, but fails with verifier error when I set it to **2740** (this might be different for different kernel versions). But if we drop loop unrolling, we could set it even much higher.

bacher09:
Regarding limitations, basically it's limited by instructions count (because of loop unrolling) and under different compiler could be different.
Without loop unrolling (this would drop support for **5.4**) it would be probably limited by max structure size, which should be 4 GiB or amount of free ram (depending on which is smaller), but I'm currently not sure about this.

anakryiko:
The bigger the limits, the longer BPF verifier might need to spend veryfing everything. It might use more memory, etc. So limits can't be unreasonable.

brendangregg:
Ok; I think ARGSIZE's limit will be more annoying. Perhaps in a future PR we can explore doing something smarter: I thought with the newer string functions we could find the length of a NULL-terminated string and pack the next arg at the end of the first.

yonghong-song:
[buildbot, test this please]

yonghong-song:
> Ok; I think ARGSIZE's limit will be more annoying. Perhaps in a future PR we can explore doing something smarter: I thought with the newer string functions we could find the length of a NULL-terminated string and pack the next arg at the end of the first.

Do you have a reference for this ("with the newer string functions we could find the length of a NULL-terminated string ")? 

The bpf program could test the return value of say `bpf_probe_read_user_str`. If the return value is equal to the buffer size. That means the kernel copied `size - 1` data and appended '\0' at the end. The bpf program can try again by adjusting the source pointer. A bounded loop could do the work.

Agree this is not a particular issue for this patch.

PR #2809, opened 2020-03-11
libbpf-tools: small clean ups across all tools

anakryiko:
Remove BPF_F_CURRENT_CPU definitions, which are now provided by vmlinux.h
after 1aae4bdd7879 ("bpf: Switch BPF UAPI #define constants used from BPF
program side to enums") commit in kernel.

Fix potential uninitialized read warning in opensnoop.

Also add opensnoop to .gitignore.

Signed-off-by: Andrii Nakryiko <andriin@fb.com>

PR #2811, opened 2020-03-11
libbpf-tools: CO-RE opensnoop minor improvements

bacher09:
* allow passing `0` as valid uid
 * fix type of `targ_uid`

Also wanted to say, that there is another way to implement UID filtering, that will support `0` UID. According to [kernel source][uid_valid] there is only one invalid UID, it is `(uint)-1` which should be `UINT_MAX`. So instead of adding another variable, like I did, it possible to check if `targ_uid` is `UINT_MAX` (or `-1`), but I don't think we should be stingy on variables in this case.

[uid_valid]: https://github.com/torvalds/linux/blob/master/include/linux/uidgid.h#L113

### DIFF
--- a/libbpf-tools/opensnoop.bpf.c
+++ b/libbpf-tools/opensnoop.bpf.c
@@ -10,7 +10,7 @@
 const volatile __u64 min_us = 0;
 const volatile pid_t targ_pid = 0;
 const volatile pid_t targ_tgid = 0;
-const volatile pid_t targ_uid = 0;
+const volatile uid_t targ_uid = 0;
 const volatile bool targ_failed = false;
 
 struct {
@@ -26,6 +26,10 @@ struct {
 	__uint(value_size, sizeof(u32));
 } events SEC(".maps");
 
+static __always_inline bool valid_uid(uid_t uid) {
+	return uid != INVALID_UID;
+}
+
 static __always_inline
 int trace_filtered(u32 tgid, u32 pid)
 {
@@ -36,8 +40,8 @@ int trace_filtered(u32 tgid, u32 pid)
 		return 1;
 	if (targ_pid && targ_pid != pid)
 		return 1;
-	if (targ_uid) {
-		uid = bpf_get_current_uid_gid();
+	if (valid_uid(targ_uid)) {
+		uid = (u32)bpf_get_current_uid_gid();
 		if (targ_uid != uid) {
 			return 1;
 		}
--- a/libbpf-tools/opensnoop.c
+++ b/libbpf-tools/opensnoop.c
@@ -39,7 +39,9 @@ static struct env {
 	bool extended;
 	bool failed;
 	char *name;
-} env = {};
+} env = {
+	.uid = INVALID_UID
+};
 
 const char *argp_program_version = "opensnoop 0.1";
 const char *argp_program_bug_address = "<bpf@vger.kernel.org>";
@@ -79,7 +81,7 @@ static const struct argp_option opts[] = {
 static error_t parse_arg(int key, char *arg, struct argp_state *state)
 {
 	static int pos_args;
-	int pid, uid, duration;
+	long int pid, uid, duration;
 
 	switch (key) {
 	case 'e':
@@ -134,7 +136,7 @@ static error_t parse_arg(int key, char *arg, struct argp_state *state)
 	case 'u':
 		errno = 0;
 		uid = strtol(arg, NULL, 10);
-		if (errno || uid <= 0) {
+		if (errno || uid < 0 || uid >= INVALID_UID) {
 			fprintf(stderr, "Invalid UID %s\n", arg);
 			argp_usage(state);
 		}
--- a/libbpf-tools/opensnoop.h
+++ b/libbpf-tools/opensnoop.h
@@ -4,6 +4,7 @@
 
 #define TASK_COMM_LEN 16
 #define NAME_MAX 255
+#define INVALID_UID ((uid_t)-1)
 
 struct args_t {
 	const char *fname;


libbpf-tools/opensnoop.c
```diff
@@ -229,7 +231,7 @@ int main(int argc, char **argv)
 	struct perf_buffer_opts pb_opts;
 	struct perf_buffer *pb = NULL;
 	struct opensnoop_bpf *obj;
-	__u64 time_end;
+	__u64 time_end = 0;
```
bacher09: This needed to fix compilation warning (`-Wmaybe-uninitialized`).

anakryiko: Fixing this as part of clean up in https://github.com/iovisor/bcc/pull/2809

bacher09: Haven't seen this, I'll update PR.

anakryiko:
I like the idea of using `(uid_t)-1` as a flag for invalid user. Having bool + value doesn't look nice. We can abstract the checking logic into a simple function and keep the logic clean.

```
static __always_inline bool is_valid_uid(uid_t uid) {
    return uid != (uid_t) -1;
}
```

bacher09:
Well, if we add this function into header file and reuse it across other tools, then it's fine. Only one suggestion, I think it should be named `uid_valid`, since it's named this way in [linux kernel](https://github.com/torvalds/linux/blob/master/include/linux/uidgid.h#L111).

libbpf-tools/opensnoop.c
```diff
@@ -76,10 +79,22 @@ static const struct argp_option opts[] = {
 	{},
 };
 
+static int parse_uid(const char *uid_arg, uid_t *uid) {
```
anakryiko: why extracting into a function? it's called from just one place in code

libbpf-tools/utils.h
```diff
@@ -0,0 +1,14 @@
+#ifndef __UTILS_H
+
+#define INVALID_UID ((uid_t)-1)
+#define INVALID_GID ((gid_t)-1)
+
+static __always_inline bool valid_uid(uid_t uid) {
+	return uid != INVALID_UID;
+}
+
+static __always_inline bool valid_gid(gid_t gid) {
```
anakryiko: nothing is using valid_gid() and INVALID_GID, please drop it. No need for utils.h as well, just put valid_uid inside opensnoop.bpf.c

bacher09: I added them because I think it's logical if we have `uid` filter, then `gid` filter is something that has to be. At least most *nix tools has that convention.

But I agree, it would be better to add them when there will be tools that use it. I'll remove them.

Regarding that there is no need in `utils.h`, I recalled why I thought it's bad idea to use only one variable for uid filtering, it's primarily duplication. Basically, in this case we need [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming))  that will repeat both in `opensnoop.c` (we need to initialize variable with invalid uid if we are not doing uid filtering) and `opensnoop.bpf.c`. 

Moreover, porting other tools (e.g. `execsnoop`, `bindsnoop`) would require same functions, since they also implementing uid filtering and it's better to share this code between them. Also, in some sense, python tools has something [similar](https://github.com/iovisor/bcc/blob/master/src/python/bcc/utils.py#L61).

But, maybe it's something that should be added in separate PR.

libbpf-tools/opensnoop.c
bacher09: It's basically same reason, as for `utils.h`. Parsing `uid` is something that done by multiple tools, so I thought that it's something that could be moved into separate file later (maybe when there will be more tools that will require this).

libbpf-tools/opensnoop.c
```diff
@@ -79,7 +81,7 @@ static const struct argp_option opts[] = {
 static error_t parse_arg(int key, char *arg, struct argp_state *state)
 {
 	static int pos_args;
-	int pid, uid, duration;
+	long int pid, uid, duration;
```
bacher09: Also wanted to mention, that currently (and previously before this commit) traced pid could be different from what is specified by user. It could happen because `strtol` returns `long int` while `env` struct is using `int` type for `pid`.

libbpf-tools/opensnoop.c
anakryiko: Yeah, let's think about extracting reusable code (if it makes sense) once we have few use cases. Having extra headers with trivial functions reduces readability without really bringing much benefit (apart from saving 3 lines of trivial code). I'd also be wary about sharing code between BPF and user-space. It might work in simple cases, but could be subtly broken in others.

libbpf-tools/utils.h
anakryiko: Yeah, let's postpone all this until we really need it. Duplicating INVALID_UID definition is not a big deal, IMO.

libbpf-tools/opensnoop.c
anakryiko: right, we could check for valid range in user-space, but likelihood of someone specifying 123123123123123 as uid or pid and expecting it to work is very small :)

libbpf-tools/opensnoop.bpf.c
```diff
@@ -29,6 +29,11 @@ struct {
 	__uint(value_size, sizeof(u32));
 } events SEC(".maps");
 
+static __always_inline bool valid_uid(uid_t uid) {
+	return uid != INVALID_UID;
+}
+
+
```
anakryiko: extra empty line

libbpf-tools/opensnoop.bpf.c
```diff
@@ -39,8 +44,8 @@ int trace_filtered(u32 tgid, u32 pid)
 		return 1;
 	if (targ_pid && targ_pid != pid)
 		return 1;
-	if (targ_uid) {
-		uid = bpf_get_current_uid_gid();
+	if (valid_uid(targ_uid)) {
+		uid = bpf_get_current_uid_gid() & 0xffffffff;
```
anakryiko: nit: `(u32)bpf_get_current_uid_gid()` is shorter and a bit nicer (and harder to get wrong w.r.t. to number of 'f's :) )

libbpf-tools/opensnoop.c
bacher09: It's small, but it still isn't nice. Also, corner cases sometimes accumulate, once I was witness like issue like this in small CLI tool, caused bash script to misbehave in very odd way. But, somebody could claim that it was bug in script (it called tool with wrong arguments).

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

PR #2812, opened 2020-03-11
Strengthen tracepoint format parsing

olsajiri:
There's issue in current RHEL real time kernel with tracepoint format,
which makes bcc-tools to return wrong data.

Two new 'common_' fields were added and it causes 2 issues for tracepoint
format parser.

First issue
  - is the gap between common fields and other fields, which is not
    picked up by the parser, so the resulted struct is not aligned with
    the data.

Second issue
  - is the fact that current parser covers common fields with:
`      u64 __do_not_use__`
    so the new common fields are not accounted for.

I kept both issues and fixes separated to make the change readable.

I posted patch to fix this [1] in RT kernel, but that might take a while,
and we could easily fix our tracepoint parser to workaround this issue.

[1] [https://lore.kernel.org/linux-rt-users/20200221153541.681468-1-jolsa@kernel.org/](https://lore.kernel.org/linux-rt-users/20200221153541.681468-1-jolsa@kernel.org/)


yonghong-song:
[buildbot, test this please]

src/cc/frontends/clang/tp_frontend_action.cc
```diff
@@ -30,6 +30,7 @@
 
 #include "frontend_action_common.h"
 #include "tp_frontend_action.h"
+#include "../../../common.h"
```
yonghong-song: including "common.h" should be enough, I think.

olsajiri: yep, I changed it

yonghong-song:
[buildbot, test this please]

PR #2815, opened 2020-03-12
libbpf-tools: determine target host architecture

anakryiko:
BPF code is compiled with -target bpf, but for PT_REGS_PARM macro (and by
induction for BPF_KPROBE/BPF_KRETPROBE macros as well), it's important to know
what's the target host original architecture was, to use correct definition of
struct pt_regs. Determine that based on output of `uname -m` (taking into
account that both x86_64 and x86 are defined as x86 internally for kernel).

Reported-by: Wenbo Zhang <ethercflow@gmail.com>
Signed-off-by: Andrii Nakryiko <andriin@fb.com>


yonghong-song:
[buildbot, test this please]

PR #2826, opened 2020-03-19
opensnoop: fix --cgroupmap with kfunc

alban:
Commit c347fe6c9f75 ("Support kfunc in opensnoop.py") introduces an
alternative probe on do_sys_open() with kfuncs instead of kprobes. This
new implementation is used if the kernel supports it. But it removed the
--cgroupmap filter added in commit b2aa29fa3269 ("tools: cgroup
filtering in execsnoop/opensnoop").

This patch adds the --cgroupmap filter in the kfunc implementation.

-----

Note: I didn't test it yet; I will need to get a newer kernel first, see https://github.com/iovisor/bcc/pull/2726#issuecomment-601095997.

/cc @olsajiri


yonghong-song:
[buildbot, test this please]

PR #2831, opened 2020-03-22
chore: include Amazon Linux 2 binary option in INSTALL

jfreeland:
I was going through the process of building LLVM and Clang and stumbled upon this gem, https://forums.aws.amazon.com/thread.jspa?threadID=318439.

Lo and behold, it's there and it drastically simplifies my life.


yonghong-song:
[buildbot, test this please]

PR #2838, opened 2020-03-24
cachetop: Change the default sort order to Descending

mkogan1:
currently with ascending sort the useful information
is commonly beyond the bottom of the terminal window
and it is necessary to reverse the sort manually every execution.

Signed-off-by: Mark Kogan <mkogan@redhat.com>


yonghong-song:
[buildbot, test this please]

yonghong-song:
Makes sense to me as the descending order (1). easy to see significant numbers first and (2). similar to perf output. cc @brendangregg

PR #2840, opened 2020-03-25
Translate virtual addresses to binary addresses for shared libs as well.

michalgr:
Quoting https://refspecs.linuxbase.org/elf/elf.pdf:
```
Symbol table entries for different object file types have slightly different
interpretations for the st_value member.
- In relocatable files, st_value holds alignment constraints for a symbol
whose section index is SHN_COMMON.
- In relocatable files, st_value holds a section offset for a defined symbol.
That is, st_value is an offset from the beginning of the section that st_shndx
identifies.
- In executable and shared object files, st_value holds a virtual address. To
make these files' symbols more useful for the dynamic linker, the section offset
(file interpretation) gives way to a virtual address (memory interpretation) for
which the section number is irrelevant.
```

This is a problem in practice as well. I run into this while tracing shared
libraries on Android with bpftrace. Some of them have text sections mmapped
at different offset than file offset which results in probes being placed
at wrong offsets.

### DIFF
--- a/src/cc/bcc_syms.cc
+++ b/src/cc/bcc_syms.cc
@@ -707,6 +707,7 @@ int bcc_resolve_symname(const char *module, const char *symname,
                         const uint64_t addr, int pid,
                         struct bcc_symbol_option *option,
                         struct bcc_symbol *sym) {
+  int module_type;
   static struct bcc_symbol_option default_option = {
     .use_debug_file = 1,
     .check_debug_file_crc = 1,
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
cc @palmtenor could you also take a look?

yonghong-song:
LGTM. Thanks.

PR #2854, opened 2020-04-02
[C++] add BPF::enable_usdt(void) and disble_usdt(void) to handle all the USDT at once

gfx:
It's bothering to write a loop such as:

```c++
  for (auto &probe : probes) {
    ret = bpf->attach_usdt(probe);
    if (!ret.ok()) {
      std::cerr << "attach_usdt: " << ret.msg() << std::endl;
      return EXIT_FAILURE;
    }
  }
```

So I think it makes sense when the BPF.h provides a way to enable/disable all the USDTs at once.

### DIFF
--- a/examples/cpp/FollyRequestContextSwitch.cc
+++ b/examples/cpp/FollyRequestContextSwitch.cc
@@ -98,7 +98,7 @@ int main(int argc, char** argv) {
     return 1;
   }
 
-  auto attach_res = bpf->attach_usdt();
+  auto attach_res = bpf->attach_usdt_all();
   if (attach_res.code() != 0) {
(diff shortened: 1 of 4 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
This may break other applications. Maybe create a new API `attach_usdt_all()`?

gfx:
I don't think these overloadings break the compatibility, but giving new names is absolutely better.

Renamed.

yonghong-song:
[buildbot, test this please]

yonghong-song:
LGTM. Thanks!

PR #2856, opened 2020-04-03
Add `bcc_version.h` to access LIBBCC_VERSION from C/C++

gfx:
I believe this is useful for tools that use C/C++ API.


yonghong-song:
[buildbot, test this please]

yonghong-song:
It might be useful. But do you have a concrete use case? Also, this is just for an example. People outside bcc repo, i.e., using installed headers and librararies still won't be able to access the version string.

gfx:
I'd like to show the BCC version in my BPF tool written in C++: https://github.com/toru/h2olog/tree/v2 

Also, the header will be installed as`$INSTALL_PREFIX/include/bcc/bcc_version.h` (e.g. `/usr/include/bcc/bcc_version.h`). Its content is something like this:

```
#ifndef LIBBCC_VERSION_H
#define LIBBCC_VERSION_H

#define LIBBCC_VERSION "0.13.0-ab54de2e"

#endif
```

yonghong-song:
[buildbot, ok to test]

src/cc/CMakeLists.txt
```diff
@@ -13,7 +13,8 @@ include_directories(${LIBELF_INCLUDE_DIRS})
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/libbpf/include)
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/libbpf/include/uapi)
 add_definitions(${LLVM_DEFINITIONS})
-configure_file(libbcc.pc.in ${CMAKE_CURRENT_BINARY_DIR}/libbcc.pc @ONLY)
+configure_file(libbcc.pc.in libbcc.pc @ONLY)
+configure_file(bcc_version.h.in bcc_version.h @ONLY)
```
yonghong-song, line 17: The above two configure_files do not sound right. The destination file should be ${CMAKE_CURRENT_BINARY_DIR}/libbcc.pc and ${CMAKE_CURRENT_BINARY_DIR}/bcc_version.h, I think.

gfx: Sorry for the lack of explanation, but that is the default in `configure_file`:

> https://cmake.org/cmake/help/latest/command/configure_file.html

> `<output>`
> Path to the output file or directory. A relative path is treated with respect to the value of CMAKE_CURRENT_BINARY_DIR. If the path names an existing directory the output file is placed in that directory with the same file name as the input file.

yonghong-song: Thanks. I see.

The buildbot reported build failures at examples/cpp directory,
```
[ 85%] Building CXX object examples/cpp/CMakeFiles/HelloWorld.dir/HelloWorld.cc.o
/tmp/rpmbuild.wYRV5Q/BUILD/bcc/examples/cpp/HelloWorld.cc:11:10: fatal error: bcc/bcc_version.h: No such file or directory
 #include "bcc/bcc_version.h"
          ^~~~~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [examples/cpp/CMakeFiles/HelloWorld.dir/build.make:63: examples/cpp/CMakeFiles/HelloWorld.dir/HelloWorld.cc.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:924: examples/cpp/CMakeFiles/HelloWorld.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[ 85%] Linking CXX executable CPUDistribution
```

May you need to tweak the CMakefile there since the newer bcc_version.h may not make into install directory yet.

gfx: Right. Fixed in 683f139d by adding bcc_version.h's directory to include dirs.

gfx:
Only `fc27` build has failed. Possibly it fails randomly?

yonghong-song:
Yes, could be a random failure for smoke test biosnoop.py. Will rerun the test.

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

a289672082:
Hi:cmake_minimum_required(VERSION 3.3.0)
project(Test)

set(CMAKE_CXX_STANDARD 17)

#libbcc
find_package(PkgConfig REQUIRED)
pkg_check_modules(LIBBCC REQUIRED libbcc)

message(LIBBCC_INCLUDE_DIRS=${LIBBCC_INCLUDE_DIRS})
message(LIBBCC_LIBRARY_DIRS=${LIBBCC_LIBRARY_DIRS})
message(LIBBCC_LIBRARIES=${LIBBCC_LIBRARIES})
include_directories(${LIBBCC_INCLUDE_DIRS} ${LIBBCC_INCLUDE_DIRS}/..)
link_directories(${LIBBCC_LIBRARY_DIRS})


set(SOURCE_FILES main.cc)


add_executable(HelloWorld ${SOURCE_FILES})


target_link_libraries(HelloWorld ${LIBBCC_LIBRARIES})

PR #2864, opened 2020-04-06
Update filelife pid filter

DavadDi:
Update filelife pid filter;  

u32 pid = bpf_get_current_pid_tgid() >> 32;


yonghong-song:
[buildbot, test this please]

yonghong-song:
The change indeed makes sense. The pid is mostly for filtering and recording purpose. The dentry itself should be enough for the key.

PR #2877, opened 2020-04-17
build: fix clang 10 build

martinetd:
some distros already packaging clang 10 (checked fedora and arch) no longer
ship all the individual libclang*.so component libraries.
Instead, clang from 9.0 onwards provides a new lib, libclang-cpp.so, which
includes everything we need.

Tell cmake to use it if the individual libraries are no longer found.
(Build-wise, if both are present it is more efficient to use the individual
components so keep these first)

---
note I also had to set ENABLE_LLVM_SHARED=ON because various -lLLVM* didn't exist either, but that looks like a problem in my system's `/usr/lib64/cmake/llvm/LLVMConfig.cmake` more than anything else... anyway, this doesn't look related to the clang.

Anyway, thanks for bcc and bpftrace and all the rest! :)


yonghong-song:
[buildbot, test this please]

yonghong-song:
I did not install llvm 10 on my machine, but the change looks correct.

PR #2884, opened 2020-04-22
Fix ELF ABI and endianness checks for powerpc64

sandip4n:
Earlier, it was assumed that ELF ABI v2 is used only on little-endian `powerpc64` environments but it seems this ABI can be used independently of endianness.

It is expected that any C preprocessor that conforms to the ELF ABI v2 specification must predefine the `_CALL_ELF` macro and set its value to 2.

Instead of looking at `__BYTE_ORDER__` to determine whether to use the Local Entry Point (LEP) of symbols, one should look at the `_CALL_ELF` macro instead as this is ABI-related.

Similarly, `_CALL_ELF` should be used only for determining the ABI version and not the endianness.

### DIFF
--- a/src/cc/bcc_elf.c
+++ b/src/cc/bcc_elf.c
@@ -257,16 +257,8 @@ static int list_in_scn(Elf *e, Elf_Scn *section, size_t stridx, size_t symsize,
         continue;
 
 #ifdef __powerpc64__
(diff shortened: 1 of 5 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

Issue #2874, opened 2020-04-15
Reading performace counters at user probes


sahmad98:
Is there a way to write bpf program to read cpu counter values at usdt probes? 

The idea is to profile and gather statistics for a part of code for different metrics which can be computed from performance counters like CPI and Core/Memory Bound etc. Adding two user probes at start and end of the code section and computing the metric in the end and passing the stats to user space application for aggregation and visualization.

sahmad98:
BPF helpers to read performance counters is added in kernel source.

Example: [trace_event_kern.c](https://github.com/torvalds/linux/blob/master/samples/bpf/trace_event_kern.c)

sahmad98:
#2888 Added an example in pull request, since i couldn't find an example for my use case.

PR #2888, opened 2020-04-24
Add perf event data collection example for an userspace application

sahmad98:
Added example for collecting performance counter data at custom user space application probes or USDT probes.

### DIFF
--- a/examples/perf/ipc.py
+++ b/examples/perf/ipc.py
@@ -3,11 +3,7 @@
 # uprobes
 # 24-Apr-2020	Saleem Ahmad	Created this.
 
-from __future__ import print_function
-from bcc import BPF, USDT, Perf
-from bcc.utils import printb
-from time import sleep
-import sys
+from bcc import BPF, utils
 from optparse import OptionParser
 
 # load BPF program
@@ -20,10 +16,12 @@ struct perf_delta {
     u64 time_delta;
 };
 
-// Perf Arrays to read counter values for open
-// perf events.
-BPF_PERF_ARRAY(clk, 16);
-BPF_PERF_ARRAY(inst, 16);
+/*
+Perf Arrays to read counter values for open
+perf events.
+*/
+BPF_PERF_ARRAY(clk, MAX_CPUS);
+BPF_PERF_ARRAY(inst, MAX_CPUS);
 
 // Perf Output
 BPF_PERF_OUTPUT(output);
@@ -37,6 +35,15 @@ void trace_start(struct pt_regs *ctx) {
     u32 time = 2;
 
     int cpu = bpf_get_smp_processor_id();
+    /*
+    perf_read may return negative values for errors.
+    If cpu id is greater than BPF_PERF_ARRAY size,
+    counters values will be very large negative number.
+    NOTE: Use bpf_perf_event_value is recommended over
+    bpf_perf_event_read or map.perf_read() due to
+    issues in ABI. map.perf_read_value() need to be
+    implemented in future.
+    */
     u64 clk_start = clk.perf_read(cpu);
     u64 inst_start = inst.perf_read(cpu);
     u64 time_start = bpf_ktime_get_ns();
@@ -70,35 +77,45 @@ void trace_end(struct pt_regs* ctx) {
     u32 time = 2;
 
     int cpu = bpf_get_smp_processor_id();
+    /*
+    perf_read may return negative values for errors.
+    If cpu id is greater than BPF_PERF_ARRAY size,
+    counters values will be very large negative number.
+    NOTE: Use bpf_perf_event_value is recommended over
+    bpf_perf_event_read or map.perf_read() due to
+    issues in ABI. map.perf_read_value() need to be
+    implemented in future.
+    */
     u64 clk_end = clk.perf_read(cpu);
     u64 inst_end = inst.perf_read(cpu);
     u64 time_end = bpf_ktime_get_ns();
     
     struct perf_delta perf_data = {} ;
-    bool submit = true;
     u64* kptr = NULL;
     kptr = data.lookup(&clk_k);
+
+    // Find elements in map, if not found return
     if (kptr) {
         perf_data.clk_delta = clk_end - *kptr;
+    } else {
+        return;
     }
     
     kptr = data.lookup(&inst_k);
     if (kptr) {
         perf_data.inst_delta = inst_end - *kptr;
     } else {
-        submit = false;
+        return;
     }
 
     kptr = data.lookup(&time);
     if (kptr) {
         perf_data.time_delta = time_end - *kptr;
     } else {
-        submit = false;
+        return;
     }
 
-    if (submit) {
-        output.perf_submit(ctx, &perf_data, sizeof(struct perf_delta));
-    }
+    output.perf_submit(ctx, &perf_data, sizeof(struct perf_delta));
 }
 """
 
@@ -112,7 +129,9 @@ if (not options.lib_name or not options.sym):
     parser.print_help()
     exit()
 
-b = BPF(text=code)
+num_cpus = len(utils.get_online_cpus())
+
+b = BPF(text=code, cflags=['-DMAX_CPUS=%s' % str(num_cpus)])
 
 # Attach Probes at start and end of the trace function
 # NOTE: When attaching to a function for tracing, during runtime relocation
@@ -127,14 +146,11 @@ b.attach_uretprobe(name=options.lib_name, sym=options.sym, fn_name="trace_end")
 
 def print_data(cpu, data, size):
     e = b["output"].event(data)
-    try:
-        print("%-8d %-8d %-8.2f %-8s %d" % (e.clk_delta, e.inst_delta, 
-            1.0* e.inst_delta/e.clk_delta, str(round(e.time_delta * 1e-3, 2)) + ' us', cpu))
-    except Exception as e:
-        print(e)
+    print("%-8d %-12d %-8.2f %-8s %d" % (e.clk_delta, e.inst_delta, 
+        1.0* e.inst_delta/e.clk_delta, str(round(e.time_delta * 1e-3, 2)) + ' us', cpu))
 
 print("Counters Data")
-print("%-8s %-8s %-8s %-8s %s" % ('CLK', 'INST', 'IPC', 'TIME', 'CPU'))
+print("%-8s %-12s %-8s %-8s %s" % ('CLOCK', 'INSTRUCTION', 'IPC', 'TIME', 'CPU'))
 
 b["output"].open_perf_buffer(print_data)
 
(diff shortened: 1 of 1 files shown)


sahmad98:
https://github.com/iovisor/bcc/issues/2874 Issue

yonghong-song:
[buildbot, test this please]

examples/perf/ipc.py
```diff
@@ -0,0 +1,137 @@
...
+        output.perf_submit(ctx, &perf_data, sizeof(struct perf_delta));
+    }
+}
+"""
+
+b = BPF(text=code)
+
+# Attach Probes at start and end of the trace function
+b.attach_uprobe(name="c", sym="strlen", fn_name="trace_start")
+b.attach_uretprobe(name="c", sym="strlen", fn_name="trace_end")
```
yonghong-song, line 108: In my system, I have
```
-bash-4.4$ nm /usr/lib64/libc.so.6 | grep strlen
000000000008cd80 t __GI_strlen
000000000008cd30 i strlen
000000000008cd80 t __strlen_sse2
000000000016c3d0 t __strlen_sse2_no_bsf
000000000016f630 t __strlen_sse2_pminub
0000000000146cc0 t __strlen_sse42
```
Read the `nm` man page, we have
```
           "i" For PE format files this indicates that the symbol is in a section specific to the implementation of
               DLLs.  For ELF format files this indicates that the symbol is an indirect function.  This is a GNU
               extension to the standard set of ELF symbol types.  It indicates a symbol which if referenced by a
               relocation does not evaluate to its address, but instead must be invoked at runtime.  The runtime
               execution will then return the value to be used in the relocation.
```

That is, during runtime relocation stage  by runtime linker, `strlen` is called once to returns a different address
for example `__strlen_sse2` and later calling strlen from application will call `__strlen_sse2` instead. 

What is your local libc `strlen` definition looks like? We should add some comments here  to warn people `strlen` may not work depending on libc.

So this overhead may be big enough compared to actual strlen. Only big input for strlen may have a meaningful output, I think. Please add some comments along this line to warn potential users about how to interpret the results.

yonghong-song:
[buildbot, ok to test]

examples/perf/ipc.py
sahmad98: My strlen definition looks like following

```bash
➜  ~ nm -D /lib/x86_64-linux-gnu/libc-2.27.so | egrep strlen
000000000009dc70 i strlen
```

> Please add some comments along this line to warn potential users about how to interpret the results.
Sure, I will add extra comments about this and also about the context switch from user space to kernel space.

I think since its generic IPC script, better to take library name and function to trace as commnad line argument instead of just calculating for `strlen`.

sahmad98: @yonghong-song I have updated the script with notes on result. I have also added command line argument to run the script on any function inside a library.

examples/perf/ipc.py
```diff
@@ -0,0 +1,156 @@
...
+struct perf_delta {
+    u64 clk_delta;
+    u64 inst_delta;
+    u64 time_delta;
+};
+
+// Perf Arrays to read counter values for open
+// perf events.
+BPF_PERF_ARRAY(clk, 16);
+BPF_PERF_ARRAY(inst, 16);
```
yonghong-song, line 26: As a default, this might be too small, may be we can change to 128 cpus? And add a comment to remind user to increase max_entries if their system has more cpus.

examples/perf/ipc.py
```diff
@@ -0,0 +1,156 @@
...
+BPF_PERCPU_ARRAY(data, u64);
+
+void trace_start(struct pt_regs *ctx) {
+    u32 clk_k = 0;
+    u32 inst_k = 1;
+    u32 time = 2;
+
+    int cpu = bpf_get_smp_processor_id();
+    u64 clk_start = clk.perf_read(cpu);
+    u64 inst_start = inst.perf_read(cpu);
```
yonghong-song, line 41: perf_read may return negative value for errors. In this particular case, if "cpu" id is more than the perf array 'clk'/'inst' size, you will get an enormous large number of clk_start/inst_start and things will go wrong.

Most people do not check this as they know their max entry in the perf array greater than the number of cpus in the system, but a comment will be good. See above.

examples/perf/ipc.py
```diff
@@ -0,0 +1,156 @@
...
+# Perf Event for Unhalted Cycles, The hex value is
+# combination of event, umask and cmask. Read Intel
+# Doc to find the event and cmask. Or use 
+# perf list --details to get event, umask and cmask
+
+PERF_EVENT_RAW = 4
+# Unhalted Clock Cycles
+b["clk"].open_perf_event(PERF_EVENT_RAW, 0x0000003C)
+# Instruction Retired
+b["inst"].open_perf_event(PERF_EVENT_RAW, 0x000000C0)
```
yonghong-song, line 150: If the the pmu for Instruction Retired is multiplexed, the result may not be accurate. The right way is to use bpf_perf_read_value. See
  https://patchwork.ozlabs.org/project/netdev/patch/20171005161923.332790-3-yhs@fb.com/
This needs to create a new builtin function for perf_event.

For this patch, I am okay with it, but please add a comment to warn user potential issues.
Using perf_event_read_value() (or map.perf_read_value() to-be-implemented intrinsic) can be
a followup.

examples/perf/ipc.py
sahmad98: Added comment for the maximum number of cpu.

examples/perf/ipc.py
sahmad98: Added comment for same.

examples/perf/ipc.py
sahmad98: Added comment for same. `map.perf_read_value` can be a follow up.

examples/perf/ipc.py
```diff
@@ -0,0 +1,186 @@
+#!/usr/bin/python
+# IPC - Instructions Per Cycles using Perf Events and
+# uprobes
+# 24-Apr-2020	Saleem Ahmad	Created this.
+
+from __future__ import print_function
+from bcc import BPF, USDT, Perf
```
javierhonduco, line 7: I think we are not using `USDT`

examples/perf/ipc.py
```diff
@@ -0,0 +1,186 @@
...
+from bcc.utils import printb
```
javierhonduco, line 8: Same as above

examples/perf/ipc.py
```diff
@@ -0,0 +1,186 @@
...
+from time import sleep
```
javierhonduco, line 9: Same

examples/perf/ipc.py
```diff
@@ -0,0 +1,186 @@
...
+import sys
```
javierhonduco, line 10: Same

examples/perf/ipc.py
```diff
@@ -0,0 +1,186 @@
...
+
+/*
+Perf Arrays to read counter values for open
+perf events.
+BPF_PERF_ARRAY taken maximum number of cpus.
+Change it to the max number of cpu in case
+it is greater than MAX_CPUS
+*/
+#define MAX_CPUS 128
+BPF_PERF_ARRAY(clk, MAX_CPUS);
```
javierhonduco, line 31: We could potentially [pass the number of CPUs](https://github.com/iovisor/bcc/blob/98c18b640117b10f923c8697be8bfff8ad39834b/src/python/bcc/utils.py#L35) through [Cflags list](https://github.com/iovisor/bcc/blob/master/src/python/bcc/__init__.py#L286), this way it should work in all systems and does not require manually updates :D

examples/perf/ipc.py
```diff
@@ -0,0 +1,186 @@
...
+        perf_data.inst_delta = inst_end - *kptr;
+    } else {
+        submit = false;
+    }
+
+    kptr = data.lookup(&time);
+    if (kptr) {
+        perf_data.time_delta = time_end - *kptr;
+    } else {
+        submit = false;
```
javierhonduco, line 122: Here and above maybe we could short-circuit this with a `return`

examples/perf/ipc.py
```diff
@@ -0,0 +1,186 @@
...
+# This example is to give a reference on how to use perf events with tracing.
+b.attach_uprobe(name=options.lib_name, sym=options.sym, fn_name="trace_start")
+b.attach_uretprobe(name=options.lib_name, sym=options.sym, fn_name="trace_end")
+
+def print_data(cpu, data, size):
+    e = b["output"].event(data)
+    try:
+        print("%-8d %-8d %-8.2f %-8s %d" % (e.clk_delta, e.inst_delta, 
+            1.0* e.inst_delta/e.clk_delta, str(round(e.time_delta * 1e-3, 2)) + ' us', cpu))
+    except Exception as e:
```
javierhonduco, line 159: If the intent is to stop on `<ctrl+c>`, let's use `KeyError`

examples/perf/ipc.py
```diff
@@ -0,0 +1,186 @@
...
+        print(e)
+
+print("Counters Data")
+print("%-8s %-8s %-8s %-8s %s" % ('CLK', 'INST', 'IPC', 'TIME', 'CPU'))
```
javierhonduco, line 163: Can we use the full names here and in the variable names, such as `clock` instead of `clk`? I think it might be more readable :)

examples/perf/ipc.py
javierhonduco: Update: just saw you are handling this below. What kind of exception do you expect this code to raise?

examples/perf/ipc.py
```diff
@@ -0,0 +1,186 @@
...
+        data.update(&time, &time_start);
+    } else {
+        data.insert(&time, &time_start);
+    }
+}
+
+void trace_end(struct pt_regs* ctx) {
+    u32 clk_k = 0;
+    u32 inst_k = 1;
+    u32 time = 2;
```
javierhonduco, line 86: As we use these to access elements in the array, can we move them to some defines?

examples/perf/ipc.py
```diff
@@ -0,0 +1,186 @@
...
+# NOTE: Events can be multiplexed by kernel in case the
+# number of counters is greater than supported by CPU
+# performance monitoring unit, which can result in inaccurate
+# results. Counter values need to be normalized for a more
+# accurate value.
+PERF_EVENT_RAW = 4
+# Unhalted Clock Cycles
+b["clk"].open_perf_event(PERF_EVENT_RAW, 0x0000003C)
+# Instruction Retired
+b["inst"].open_perf_event(PERF_EVENT_RAW, 0x000000C0)
```
javierhonduco, line 180: Maybe we could move these kernel constants in BPF's code in another PR

examples/perf/ipc.py
sahmad98: Removed

examples/perf/ipc.py
sahmad98: Removed

examples/perf/ipc.py
sahmad98: Removed

examples/perf/ipc.py
sahmad98: Removed

examples/perf/ipc.py
sahmad98: Changed to pass CPU number from cflags

examples/perf/ipc.py
sahmad98: Removed submit, and put a return in case of map lookups fail.

examples/perf/ipc.py
sahmad98: I was expecting zero division error, but since i now check for entries in map lookup and then only submit it to output event, the exception is not required now.

examples/perf/ipc.py
sahmad98: Changed

examples/perf/ipc.py
sahmad98: No we can't move them to defines, because `map.lookup()` requires a pointer to key value, which is not possible if we move the keys to defines.

examples/perf/ipc.py
sahmad98: Sure, Not changing in this PR. Could be pushed in another PR.

examples/perf/ipc.py
javierhonduco: Sorry that I was not too clear, I was thinking of something like:
```C
#define CLOCK_ID
...

void trace_end(struct pt_regs* ctx) {
  u32 clk_k = CLOCK_ID;
  ...
}
```

sahmad98: Changed.

marceloamaral:
Any idea how `perf_read_value` can be used?

Issue #1649, opened 2018-03-23
bpf_log2l helper gives an off by one result


ipronin:
When calculating the binary logarithm of a 64-bit integer `bpf_log2l` helper adds 1 to the logarithm of a 32-bit half calculated by `bpf_log2`.
```
value               | bpf_log2 | bpf_log2l
0                   | 0        | 1
1                   | 0        | 1
2                   | 1        | 2
4                   | 2        | 3
4294967296          |          | 33
8589934592          |          | 34
9223372036854775808 |          | 64
```

That doesn't cause errors with code like `dist.increment(bpf_log2l(1ull << 63))`. I suspect that is due to element lookup prior to updating. In real use cases the 0th bucket will always be empty and the 63rd will be lost. Interesting that `bcc.table._print_log2_hist` works around this by printing buckets 1 through 63, skipping the always empty 0th bucket. Was it done intentionally? I can send a patch that fixes this issue if you would like.

brendangregg:
No, I don't remember... I see the commit history is lacking the comment I put here: https://github.com/iovisor/bcc/pull/240  :-)

So it's from tracex2, https://github.com/torvalds/linux/blob/master/samples/bpf/tracex2_kern.c#L56

But I added "+ 1"s. It looks intentional, but I can't remember why...

ipronin:
Should I fix it and see what breaks? :) The code looks intentional, but the results are quite unexpected.

brendangregg:
Be aware that I don't think any end-user tools use bpf_log2() directly. They all use bpf_log2l(). So I don't think their lack of consistency has been a problem yet. And also note that there are tests for this, eg, tests/python/test_histogram.py, but I'm not sure what they are testing.

I still haven't remembered why it's there (I wish I wrote a comment). Here's my guesses:

A) Reserving the 0th slot for negative values (which will get aggregated as one bucket). A common tracing histogram is the retval of the read() syscall, to show a read() I/O size histogram, but read() can also return a negative ERRNO. I may have been thinking of that and reserved 0 on purpose for that case.

B) I wanted to differentiate 0 from 1, but then didn't finish writing the code. And was meaning to do this:

```
value               | bpf_log2l
0                   | 0
1                   | 1
2                   | 2
4                   | 3
```

And then change how histograms are printed in BPF.print_log2_hist() to do something like this (printing both 0 and 1 on separate lines):

```
     usecs               : count     distribution
         0 -> 0          : 0        |                                        |
[...]

PR #2906, opened 2020-05-08
libbpf-tools: add CO-RE cpudist

ethercflow:
- Implement cpudist, which is consistent with https://github.com/iovisor/bcc/blob/master/tools/cpudist_example.txt
- Add a new helper `print_log2_hist` in `trace_helpers`
- Add a general dictionary (based on the glibc's binary search tree) in `trace_helpers`
- Add a bpf helper `bpf_log(l)` in `helpers.h`
- Add  `-Wno-unused-result` to CFLAGS to ignore warnings caused by `asprintf`.

### DIFF
--- a/libbpf-tools/cpudist.bpf.c
+++ b/libbpf-tools/cpudist.bpf.c
@@ -1,0 +1,120 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2020 Wenbo Zhang
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_tracing.h>
+#include "cpudist.h"
+
+#define TASK_RUNNING 0
+
+const volatile bool targ_per_process = false;
+const volatile bool targ_per_thread = false;
+const volatile bool targ_offcpu = false;
+const volatile bool targ_ms = false;
+const volatile pid_t targ_tgid = -1;
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__type(key, u32);
+	__type(value, u64);
+} start SEC(".maps");
+
+static struct hist initial_hist;
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__type(key, u32);
+	__type(value, struct hist);
+} hists SEC(".maps");
+
+static __always_inline u64 log2(u32 v)
+{
+	u32 shift, r;
+
+	r = (v > 0xFFFF) << 4; v >>= r;
+	shift = (v > 0xFF) << 3; v >>= shift; r |= shift;
+	shift = (v > 0xF) << 2; v >>= shift; r |= shift;
+	shift = (v > 0x3) << 1; v >>= shift; r |= shift;
+	r |= (v >> 1);
+
+	return r;
+}
+
+static __always_inline u64 log2l(u64 v)
+{
+	u32 hi = v >> 32;
+
+	if (hi)
+		return log2(hi) + 32;
+	else
+		return log2(v);
+}
+
+static __always_inline void store_start(u32 tgid, u32 pid, u64 ts)
+{
+	if (targ_tgid != -1 && targ_tgid != tgid)
+		return;
+	bpf_map_update_elem(&start, &pid, &ts, 0);
+}
+
+static __always_inline void update_hist(struct task_struct *task,
+					u32 tgid, u32 pid, u64 ts)
+{
+	u64 delta, *tsp, slot;
+	struct hist *histp;
+	u32 id;
+
+	if (targ_tgid != -1 && targ_tgid != tgid)
+		return;
+
+	tsp = bpf_map_lookup_elem(&start, &pid);
+	if (!tsp || ts < *tsp)
+		return;
+
+	if (targ_per_process)
+		id = tgid;
+	else if (targ_per_thread)
+		id = pid;
+	else
+		id = -1;
+	histp = bpf_map_lookup_elem(&hists, &id);
+	if (!histp) {
+		bpf_map_update_elem(&hists, &id, &initial_hist, 0);
+		histp = bpf_map_lookup_elem(&hists, &id);
+		if (!histp)
+			return;
+		BPF_CORE_READ_STR_INTO(&histp->comm, task, comm);
+	}
+	delta = ts - *tsp;
+	if (targ_ms)
+		delta /= 1000000;
+	else
+		delta /= 1000;
+	slot = log2l(delta);
+	if (slot >= MAX_SLOTS)
+		slot = MAX_SLOTS - 1;
+	__sync_fetch_and_add(&histp->slots[slot], 1);
+}
+
+SEC("kprobe/finish_task_switch")
+int BPF_KPROBE(kprobe__finish_task_switch, struct task_struct *prev)
+{
+	u32 prev_tgid = BPF_CORE_READ(prev, tgid);
+	u32 prev_pid = BPF_CORE_READ(prev, pid);
+	u64 id = bpf_get_current_pid_tgid();
+	u32 tgid = id >> 32, pid = id;
+	u64 ts = bpf_ktime_get_ns();
+
+	if (targ_offcpu) {
+		store_start(prev_tgid, prev_pid, ts);
+		update_hist((void*)bpf_get_current_task(), tgid, pid, ts);
+	} else {
+		if (BPF_CORE_READ(prev, state) == TASK_RUNNING)
+			update_hist(prev, prev_tgid, prev_pid, ts);
+		store_start(tgid, pid, ts);
+	}
+	return 0;
+}
+
+char LICENSE[] SEC("license") = "GPL";
--- a/libbpf-tools/cpudist.c
+++ b/libbpf-tools/cpudist.c
@@ -1,0 +1,286 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+// Copyright (c) 2020 Wenbo Zhang
+//
+// Based on cpudist(8) from BCC by Brendan Gregg & Dina Goldshtein.
+// 8-May-2020   Wenbo Zhang   Created this.
+#include <argp.h>
+#include <signal.h>
+#include <stdio.h>
+#include <sys/resource.h>
+#include <unistd.h>
+#include <time.h>
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+#include "cpudist.h"
+#include "cpudist.skel.h"
+#include "trace_helpers.h"
+
+static struct env {
+	time_t interval;
+	pid_t pid;
+	int times;
+	bool offcpu;
+	bool timestamp;
+	bool per_process;
+	bool per_thread;
+	bool milliseconds;
+	bool verbose;
+} env = {
+	.interval = 99999999,
+	.pid = -1,
+	.times = 99999999,
+};
+
+static volatile bool exiting;
+
+const char *argp_program_version = "cpudist 0.1";
+const char *argp_program_bug_address = "<ethercflow@gmail.com>";
+const char argp_program_doc[] =
+"Summarize on-CPU time per task as a histogram.\n"
+"\n"
+"USAGE: cpudist [-h] [-O] [-T] [-m] [-P] [-L] [-p PID] [interval] [count]\n"
+"\n"
+"EXAMPLES:\n"
+"    cpudist              # summarize on-CPU time as a histogram"
+"    cpudist -O           # summarize off-CPU time as a histogram"
+"    cpudist 1 10         # print 1 second summaries, 10 times"
+"    cpudist -mT 1        # 1s summaries, milliseconds, and timestamps"
+"    cpudist -P           # show each PID separately"
+"    cpudist -p 185       # trace PID 185 only";
+
+static const struct argp_option opts[] = {
+	{ NULL, 'h', NULL, OPTION_HIDDEN, "Show the full help" },
+	{ "offcpu", 'O', NULL, 0, "Measure off-CPU time" },
+	{ "timestamp", 'T', NULL, 0, "Include timestamp on output" },
+	{ "milliseconds", 'm', NULL, 0, "Millisecond histogram" },
+	{ "pids", 'P', NULL, 0, "Print a histogram per process ID" },
+	{ "tids", 'L', NULL, 0, "Print a histogram per thread ID" },
+	{ "pid", 'p', "PID", 0, "Trace this PID only" },
+	{ "verbose", 'v', NULL, 0, "Verbose debug output" },
+	{},
+};
+
+static error_t parse_arg(int key, char *arg, struct argp_state *state)
+{
+	static int pos_args;
+
+	switch (key) {
+	case 'v':
+		env.verbose = true;
+		break;
+	case 'h':
+		argp_usage(state);
+		break;
+	case 'm':
+		env.milliseconds = true;
+		break;
+	case 'p':
+		errno = 0;
+		env.pid = strtol(arg, NULL, 10);
+		if (errno) {
+			fprintf(stderr, "invalid PID: %s\n", arg);
+			argp_usage(state);
+		}
+		break;
+	case 'O':
+		env.offcpu = true;
+		break;
+	case 'P':
+		env.per_process = true;
+		break;
+	case 'L':
+		env.per_thread = true;
+		break;
+	case 'T':
+		env.timestamp = true;
+		break;
+	case ARGP_KEY_ARG:
+		errno = 0;
+		if (pos_args == 0) {
+			env.interval = strtol(arg, NULL, 10);
+			if (errno) {
+				fprintf(stderr, "invalid internal\n");
+				argp_usage(state);
+			}
+		} else if (pos_args == 1) {
+			env.times = strtol(arg, NULL, 10);
+			if (errno) {
+				fprintf(stderr, "invalid times\n");
+				argp_usage(state);
+			}
+		} else {
+			fprintf(stderr,
+				"unrecognized positional argument: %s\n", arg);
+			argp_usage(state);
+		}
+		pos_args++;
+		break;
+	default:
+		return ARGP_ERR_UNKNOWN;
+	}
+	return 0;
+}
+
+int libbpf_print_fn(enum libbpf_print_level level,
+		const char *format, va_list args)
+{
+	if (level == LIBBPF_DEBUG && !env.verbose)
+		return 0;
+	return vfprintf(stderr, format, args);
+}
+
+static int bump_memlock_rlimit(void)
+{
+	struct rlimit rlim_new = {
+		.rlim_cur	= RLIM_INFINITY,
+		.rlim_max	= RLIM_INFINITY,
+	};
+
+	return setrlimit(RLIMIT_MEMLOCK, &rlim_new);
+}
+
+static int get_pid_max(void)
+{
+	int pid_max;
+	FILE *f;
+
+	f = fopen("/proc/sys/kernel/pid_max", "r");
+	if (!f)
+		return -1;
+	if (fscanf(f, "%d\n", &pid_max) != 1)
+		pid_max = -1;
+	fclose(f);
+	return pid_max;
+}
+
+static void sig_handler(int sig)
+{
+	exiting = true;
+}
+
+static int print_log2_hists(int fd)
+{
+	char *units = env.milliseconds ? "msecs" : "usecs";
+	__u32 lookup_key = -2, next_key;
+	struct hist hist;
+	int err;
+
+	while (!bpf_map_get_next_key(fd, &lookup_key, &next_key)) {
+		err = bpf_map_lookup_elem(fd, &next_key, &hist);
+		if (err < 0) {
+			fprintf(stderr, "failed to lookup hist: %d\n", err);
+			return -1;
+		}
+		if (env.per_process)
+			printf("\npid = %d %s\n", next_key, hist.comm);
+		if (env.per_thread)
+			printf("\ntid = %d %s\n", next_key, hist.comm);
+		print_log2_hist(hist.slots, MAX_SLOTS, units);
+		lookup_key = next_key;
+	}
+
+	lookup_key = -2;
+	while (!bpf_map_get_next_key(fd, &lookup_key, &next_key)) {
+		err = bpf_map_delete_elem(fd, &next_key);
+		if (err < 0) {
+			fprintf(stderr, "failed to cleanup hist : %d\n", err);
+			return -1;
+		}
+		lookup_key = next_key;
+	}
+	return 0;
+}
+
+int main(int argc, char **argv)
+{
+	static const struct argp argp = {
+		.options = opts,
+		.parser = parse_arg,
+		.doc = argp_program_doc,
+	};
+	struct cpudist_bpf *obj;
+	int pid_max, fd, err;
+	struct tm *tm;
+	char ts[32];
+	time_t t;
+
+	err = argp_parse(&argp, argc, argv, 0, NULL, NULL);
+	if (err)
+		return err;
+
+	libbpf_set_print(libbpf_print_fn);
+
+	err = bump_memlock_rlimit();
+	if (err) {
+		fprintf(stderr, "failed to increase rlimit: %d\n", err);
+		return 1;
+	}
+
+	obj = cpudist_bpf__open();
+	if (!obj) {
+		fprintf(stderr, "failed to open and/or load BPF ojbect\n");
+		return 1;
+	}
+
+	/* initialize global data (filtering options) */
+	obj->rodata->targ_per_process = env.per_process;
+	obj->rodata->targ_per_thread = env.per_thread;
+	obj->rodata->targ_ms = env.milliseconds;
+	obj->rodata->targ_offcpu = env.offcpu;
+	obj->rodata->targ_tgid = env.pid;
+
+	pid_max = get_pid_max();
+	if (pid_max < 0) {
+		fprintf(stderr, "failed to get pid_max\n");
+		return 1;
+	}
+
+	bpf_map__resize(obj->maps.start, pid_max);
+	if (!env.per_process && !env.per_thread)
+		bpf_map__resize(obj->maps.hists, 1);
+	else
+		bpf_map__resize(obj->maps.hists, pid_max);
+
+	err = cpudist_bpf__load(obj);
+	if (err) {
+		fprintf(stderr, "failed to load BPF object: %d\n", err);
+		goto cleanup;
+	}
+
+	err = cpudist_bpf__attach(obj);
+	if (err) {
+		fprintf(stderr, "failed to attach BPF programs\n");
+		goto cleanup;
+	}
+
+	fd = bpf_map__fd(obj->maps.hists);
+
+	signal(SIGINT, sig_handler);
+
+	printf("Tracing %s-CPU time... Hit Ctrl-C to end.\n", env.offcpu ? "off" : "on");
+
+	/* main: poll */
+	while (1) {
+		sleep(env.interval);
+		printf("\n");
+
+		if (env.timestamp) {
+			time(&t);
+			tm = localtime(&t);
+			strftime(ts, sizeof(ts), "%H:%M:%S", tm);
+			printf("%-8s\n", ts);
+		}
+
+		err = print_log2_hists(fd);
+		if (err)
+			break;
+
+		if (exiting || --env.times == 0)
+			break;
+	}
+
+cleanup:
+	cpudist_bpf__destroy(obj);
+
+	return err != 0;
+}
--- a/libbpf-tools/cpudist.h
+++ b/libbpf-tools/cpudist.h
@@ -1,0 +1,13 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+#ifndef __CPUDIST_H
+#define __CPUDIST_H
+
+#define TASK_COMM_LEN	16
+#define MAX_SLOTS	36
+
+struct hist {
+	__u32 slots[MAX_SLOTS];
+	char comm[TASK_COMM_LEN];
+};
+
+#endif // __CPUDIST_H
--- a/libbpf-tools/trace_helpers.c
+++ b/libbpf-tools/trace_helpers.c
@@ -150,3 +150,65 @@ const struct ksym *ksyms__get_symbol(const struct ksyms *ksyms,
 
 	return NULL;
 }
+
+#define min(x, y) ({				 \
+	typeof(x) _min1 = (x);			 \
+	typeof(y) _min2 = (y);			 \
+	(void) (&_min1 == &_min2);		 \
+	_min1 < _min2 ? _min1 : _min2; })
+
+static void print_stars(unsigned int val, unsigned int val_max, int width)
+{
+	int num_stars, num_spaces, i;
+	bool need_plus;
+
+	num_stars = min(val, val_max) * width / val_max;
+	num_spaces = width - num_stars;
+	need_plus = val > val_max;
+
+	for (i = 0; i < num_stars; i++)
+		printf("*");
+	for (i = 0; i < num_spaces; i++)
+		printf(" ");
+	if (need_plus)
+		printf("+");
+}
+
+void print_log2_hist(unsigned int *vals, int vals_size, char *val_type)
+{
+	int stars_max = 40, idx_max = -1;
+	unsigned int val, val_max = 0;
+	unsigned long long low, high;
+	int stars, width, i;
+
+	for (i = 0; i < vals_size; i++) {
+		val = vals[i];
+		if (val > 0)
+			idx_max = i;
+		if (val > val_max)
+			val_max = val;
+	}
+
+	if (idx_max < 0)
+		return;
+
+	printf("%*s%-*s : count    distribution\n", idx_max <= 32 ? 5 : 15, "",
+		idx_max <= 32 ? 19 : 29, val_type);
+
+	if (idx_max <= 32)
+		stars = stars_max;
+	else
+		stars = stars_max / 2;
+
+	for (i = 0; i <= idx_max; i++) {
+		low = (1ULL << (i + 1)) >> 1;
+		high = (1ULL << (i + 1)) - 1;
+		if (low == high)
+			low -= 1;
+		val = vals[i];
+		width = idx_max <= 32 ? 10 : 20;
+		printf("%*lld -> %-*lld : %-8d |", width, low, width, high, val);
+		print_stars(val, val_max, stars);
+		printf("|\n");
+	}
+}
--- a/libbpf-tools/.gitignore
+++ b/libbpf-tools/.gitignore
@@ -1,4 +1,5 @@
 /.output
+/cpudist
 /drsnoop
 /execsnoop
 /filelife
--- a/libbpf-tools/Makefile
+++ b/libbpf-tools/Makefile
@@ -10,6 +10,7 @@ CFLAGS := -g -O2 -Wall
 ARCH := $(shell uname -m | sed 's/x86_64/x86/')
 
 APPS = \
+	cpudist \
 	drsnoop \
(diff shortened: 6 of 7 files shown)


ethercflow:
Hi @anakryiko , long time no see. I've converted the `cpudist`, PTAL when you have time. Thank you. :)

ethercflow:
Hi @anakryiko , long time no see. I've converted the `cpudist`, would you PTAL when you have time? Thank you. :)

anakryiko:
@ethercflow, sure, will take a look when a get a chance in next few days

libbpf-tools/cpudist.c
```diff
@@ -0,0 +1,390 @@
...
+		print_log2_hist(dn->v, max_slots, "usecs");
+}
+
+DICT_ITER_FUNC(__print_per_task_log2_hist);
+
+static void print_per_task_log2_hist(int fd, int pid_max)
+{
+	struct dict d = { .root = NULL, .compar = pid_t_cmp };
+	struct piddata *lookup_key, *next_key, keys[pid_max];
+	DECLARE_LIBBPF_OPTS(bpf_map_batch_opts, opts,
```
anakryiko, line 262: map batch operations are pretty recent addition, you'll be significantly limiting range of supported kernel. Is batching really needed, especially if you do one histogram as one map entry (instead of slot as map entry)?

libbpf-tools/cpudist.bpf.c
```diff
@@ -0,0 +1,107 @@
...
+	} else {
+		piddata.slot = slot;
+		histp = &per_task_hist;
+		slotp = &piddata;
+		if (targ_per_process) {
+			piddata.pid = tgid;
+		} else if (targ_per_thread) {
+			piddata.pid = pid;
+		}
+	}
```
anakryiko, line 78: this feels like an overkill, honestly. Why not unify global and per process/thread data structures by encoding global one as special pid==-1 or something like that? it would all be quite simpler already.

But then, instead of having each histogram slot as a separate entry in hashtable, wouldn't it make more sense to have a fixed-sized array for each possible pid and update corresponding slots? Sure, it's a bit more wasteful, but then how many slots do we need per each pid, right? Plus you can make each slot just u32, instead of u64.

With such a change, you probably also don't need entire dict stuff in user-space, which would further minimize overhead.

ethercflow:
> Thanks a lot for working on this and converting more tools! Keep up the good work, I'll try to keep up with code reviews :)

You're welcome. I'll keep on doing. Thank you for your valuable comments. :)

> I did a first pass, still wrapping my head around entire flow a bit. But thought it would be good to call out stuff I already found. I think all this could be pretty significantly simplified, let's concentrate on that first, and then polish all the other details. See my remarks about histogram representation simplification.

Thank you for calling out these, I will do simplified work this weekend.

> Please also pay extra attention to freeing resources (memory, files). Try running this under ASAN as well and see if it complains about memory leaks and such.

Oh, sorry for this. I'll pay more attention and actions to avoid these. Thank you.

libbpf-tools/cpudist.c
```diff
@@ -0,0 +1,390 @@
...
+
+int libbpf_print_fn(enum libbpf_print_level level,
+		const char *format, va_list args)
+{
+	if (level == LIBBPF_DEBUG && !env.verbose)
+		return 0;
+	return vfprintf(stderr, format, args);
+}
+
+static int bump_memlock_rlimit(void)
```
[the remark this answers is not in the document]
ethercflow: Sure, I'll extract this in the next PR (with get_ktime_ns).

libbpf-tools/helpers.h
```diff
@@ -0,0 +1,35 @@
...
+	shift = (v > 0xFF) << 3; v >>= shift; r |= shift;
+	shift = (v > 0xF) << 2; v >>= shift; r |= shift;
+	shift = (v > 0x3) << 1; v >>= shift; r |= shift;
+	r |= (v >> 1);
+
+	return r;
+}
+
+static inline __attribute__((always_inline))
+unsigned int bpf_log2l(unsigned long v)
```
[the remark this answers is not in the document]
ethercflow: Sure, thank you for sharing your experience.

libbpf-tools/trace_helpers.c
```diff
@@ -150,3 +152,111 @@ const struct ksym *ksyms__get_symbol(const struct ksyms *ksyms,
...
+
+	if (idx_max > 0) {
+		if (idx_max <= 32)
+			asprintf(&header, "     %-19s : count     "
+				"distribution", val_type);
+		else
+			asprintf(&header,  "               %-29s : count     "
+				"distribution", val_type);
+		printf("%s\n", header);
+		free(header);
```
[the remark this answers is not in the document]
ethercflow: Agree, I used `asprintf` by mistake, I've fix it.

libbpf-tools/trace_helpers.c
```diff
@@ -150,3 +152,111 @@ const struct ksym *ksyms__get_symbol(const struct ksyms *ksyms,
...
+		if (idx_max <= 32)
+			asprintf(&body, "%10d -> %-10d : %-8lld |%-*s|",
+				low, high, val, stars, text);
+		else
+			asprintf(&body, "%20d -> %-20d : %-8lld |%-*s|",
+				low, high, val, stars, text);
+		printf("%s\n", body);
+		free(body);
+		free(text);
+	}
```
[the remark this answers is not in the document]
ethercflow: Agree, I used asprintf by mistake, I've fix it. Thank you.

libbpf-tools/cpudist.c
ethercflow: Oh, I didn't consider the issue of version restriction at that time, just to try the batch interface, I've fix it.

libbpf-tools/cpudist.bpf.c
ethercflow: Agree, your design is much better than mine. I complicated things. I've done with such a change, it looks much clearer and less overhead. Thank you.

libbpf-tools/cpudist.h
```diff
@@ -0,0 +1,10 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+#ifndef __CPUDIST_H
+#define __CPUDIST_H
+
+#define MAX_SLOTS    65
```
anakryiko, line 5: I think 65 is way too much and we'll be wasting memory a lot. I think we can be smarter and use only enough slots to represent reasonable periods of times, with all the (unlikely) longer durations just falling into the last bucket. E.g., if we were to track 8 hours with microsecond precision, we'd need only `log2(8 * 60 * 60 * 1000000) = 35` buckets, so allocating 36 instead of 65 would allow to capture everything within 8 hours precisely + 1 overflow bucket for everything longer than 8 hours. With millisecond precision we are talking about 8000 hours, which is beyond reasonable :)

libbpf-tools/cpudist.c
```diff
@@ -0,0 +1,314 @@
...
+#include <string.h>
+#include <stdbool.h>
+#include <sys/resource.h>
+#include <unistd.h>
+#include <time.h>
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+#include "cpudist.h"
+#include "cpudist.skel.h"
+#include "trace_helpers.h"
```
anakryiko, line 22: I bet some of these includes are not necessary?

libbpf-tools/cpudist.bpf.c
```diff
@@ -0,0 +1,121 @@
...
+	delta = ts - *tsp;
+	if (targ_ms)
+		delta /= 1000000;
+	else
+		delta /= 1000;
+	loc = log2l(delta);
+	if (loc >= MAX_SLOTS)
+		return;
+	histp->slots[loc] += 1;
```
anakryiko, line 96: log2l is always >=1, so we are wasting slot 0, plus putting 0 into slot #1, instead of slot #0. So `loc-1` here?

ethercflow:
Thanks for your further review. I'll fix them in next few days.:)

libbpf-tools/cpudist.c
```diff
@@ -0,0 +1,314 @@
...
+	if (pid_max < 0) {
+		fprintf(stderr, "failed to get pid_max\n");
+		return 1;
+	}
+
+	bpf_map__resize(obj->maps.start, pid_max);
+	if (!env.per_process && !env.per_thread)
+		bpf_map__resize(obj->maps.hists, 1);
+	else
+		bpf_map__resize(obj->maps.hists, pid_max);
```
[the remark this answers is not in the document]
ethercflow: Hi @anakryiko For hists, if we changed it to ARRAY, will the traversal overhead in user space be relatively large when we only have hundred threads? 

I simply compared overhead with `syscount`. Based on the map type, the bpf system call is only called a few hundred times , and it will be called 8w times after changing to ARRAY during the same time period.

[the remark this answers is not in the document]
anakryiko: True. You can add BPF_F_MMAPABLE flag to ARRAY and mmap() it. Then accessing it won't require any syscalls and will be extremely fast. But I'm not going to ask you to do this. Let's stick to HASHMAP for now.

libbpf-tools/cpudist.bpf.c
```diff
@@ -0,0 +1,121 @@
...
+	if (!targ_offcpu) {
+		prev_state = BPF_CORE_READ(prev, state);
+		if (prev_state == TASK_RUNNING)
+			update_hist(prev_tgid, prev_pid, ts);
+		store_start(tgid, pid, ts);
+		return 0;
+	}
+	store_start(prev_tgid, prev_pid, ts);
+	update_hist(tgid, pid, ts);
+	return 0;
```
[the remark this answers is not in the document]
ethercflow: Fixed, thanks.

libbpf-tools/cpudist.c
```diff
@@ -0,0 +1,314 @@
...
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+#include "cpudist.h"
+#include "cpudist.skel.h"
+#include "trace_helpers.h"
+
+#define PERF_BUFFER_PAGES	64
+#define PERF_BUFFER_TIME_MS	10
+#define PERF_POLL_TIMEOUT_MS	100
+
```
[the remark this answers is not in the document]
ethercflow: Oh, sorry for this.

libbpf-tools/cpudist.c
```diff
@@ -0,0 +1,314 @@
...
+"    cpudist -O           # summarize off-CPU time as a histogram"
+"    cpudist 1 10         # print 1 second summaries, 10 times"
+"    cpudist -mT 1        # 1s summaries, milliseconds, and timestamps"
+"    cpudist -P           # show each PID separately"
+"    cpudist -p 185       # trace PID 185 only";
+
+static const struct argp_option opts[] = {
+	{ NULL, 'h', NULL, OPTION_HIDDEN, "Show the full help" },
+	{ "offcpu", 'O', NULL, 0, "Measure off-CPU time" },
+	{ "timestamp", 'T',NULL, 0, "Include timestamp on output" },
```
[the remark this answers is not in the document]
ethercflow: Fixed, thanks

libbpf-tools/cpudist.c
```diff
@@ -0,0 +1,314 @@
...
+	case 'L':
+		env.per_thread = true;
+		break;
+	case 'T':
+		env.timestamp = true;
+		break;
+	case ARGP_KEY_ARG:
+		errno = 0;
+		if (pos_args == 0) {
+			interval = strtol(arg, NULL, 10);
```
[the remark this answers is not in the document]
ethercflow: Fix it, thanks

libbpf-tools/cpudist.c
```diff
@@ -0,0 +1,314 @@
...
+	char comm[16];
+	int err;
+
+	while (!bpf_map_get_next_key(fd, &lookup_key, &next_key)) {
+		err = bpf_map_lookup_elem(fd, &next_key, &hist);
+		if (err < 0) {
+			fprintf(stderr, "failed to lookup hist: %d\n", err);
+			return -1;
+		}
+		pid_to_comm(next_key, comm);
```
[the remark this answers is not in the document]
ethercflow: You're right. I've fixed, thanks.

libbpf-tools/cpudist.c
```diff
@@ -0,0 +1,314 @@
...
+	obj->rodata->targ_offcpu = env.offcpu;
+	obj->rodata->targ_tgid = env.pid;
+
+	pid_max = get_pid_max();
+	if (pid_max < 0) {
+		fprintf(stderr, "failed to get pid_max\n");
+		return 1;
+	}
+
+	bpf_map__resize(obj->maps.start, pid_max);
```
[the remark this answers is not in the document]
ethercflow: @anakryiko  Thanks for these details. But I met a problem with using `ARRAY`:

I changed two place:

1. `__uint(type, BPF_MAP_TYPE_HASH)` to `__uint(type, BPF_MAP_TYPE_ARRAY)`
2. 
```c
    if (!tsp || ts < *tsp)
        return;
```
to

 ```c
    if (!tsp || *tsp == 0 || ts < *tsp)
        return;
```

Then got some wrong tsp cause loc >= MAX_SLOTS.  


```shell
           cat-3560  [030] d... 5145650.191445: 0: loc: 35, *ts: 5145651753477337, *tsp: 5145651753477473
           <...>-2790  [024] d... 5145650.321661: 0: loc: 35, *ts: 5145651883695616, *tsp: 5145651883695679
   node_exporter-4345  [014] d... 5145650.324224: 0: loc: 35, *ts: 5145651886258518, *tsp: 5145651886258528
             cat-3560  [027] d... 5145650.348537: 0: loc: 35, *ts: 5145651910571538, *tsp: 5145651910571657
  systemd-logind-2250  [021] d... 5145651.015160: 0: loc: 35, *ts: 5145652577205888, *tsp: 5145652577205942
             cat-3560  [017] d... 5145651.015339: 0: loc: 35, *ts: 5145652577384536, *tsp: 5145652577384614
        libvirtd-3666  [036] d... 5145651.019308: 0: loc: 35, *ts: 5145652581353616, *tsp: 5145652581353662
             cat-3560  [017] d... 5145651.295419: 0: loc: 35, *ts: 5145652857469206, *tsp: 5145652857469271
             cat-3560  [017] d... 5145651.439423: 0: loc: 35, *ts: 5145653001474939, *tsp: 5145653001474999
   rs:main Q:Reg-5835  [025] d... 5145651.762465: 0: loc: 35, *ts: 5145653324521559, *tsp: 5145653324521681
           snapd-29958 [038] d... 5145651.764726: 0: loc: 35, *ts: 5145653326784044, *tsp: 5145653326784101
     dbus-daemon-2254  [021] d... 5145651.765854: 0: loc: 35, *ts: 5145653327912088, *tsp: 5145653327912402
   systemd-udevd-3645  [029] d... 5145651.766828: 0: loc: 35, *ts: 5145653328883935, *tsp: 5145653328884135
   systemd-udevd-3647  [031] d... 5145651.766852: 0: loc: 35, *ts: 5145653328908609, *tsp: 5145653328908646
   systemd-udevd-3648  [033] d... 5145651.933379: 0: loc: 35, *ts: 5145653495438366, *tsp: 5145653495438811
             cat-3560  [019] d... 5145651.963414: 0: loc: 35, *ts: 5145653525475638, *tsp: 5145653525475745
```

I'm confusing about this.

libbpf-tools/cpudist.c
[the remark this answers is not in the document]
ethercflow: Cool, I'll learn more about `BPF_F_MMAPABLE` + mmap later then use it in next tools. Thank you.

libbpf-tools/cpudist.h
ethercflow: Fixed, thanks a lot.

libbpf-tools/cpudist.c
ethercflow: Yes, sorry for this. I'll check more next time.

libbpf-tools/cpudist.bpf.c
ethercflow: Fixed, thanks.

Now I write as this:

```c
loc = log2l(delta);
if (loc > MAX_SLOTS)
        loc = MAX_SLOTS;
loc -= 1;
if (loc >= MAX_SLOTS)
        return;
histp->slots[loc] += 1;
```

I found without check again with

```c 
if (loc >= MAX_SLOTS)
        return;
```

The verifier will reject:
> R1 unbounded memory access, make sure to bounds check any array access into a map

anakryiko: verifier thinks it underflows. You have to `-= 1` first, then check against MAX_SLOTS

anakryiko: there shouldn't be `if (loc >= MAX_SLOTS) return;` at all:

```
loc = log2l(delta) - 1; /* or maybe just remove + 1 from log2l? */
if (loc >= MAX_SLOTS) loc = MAX_SLOTS - 1;
histp->slots[loc]++;
```

btw, why `loc` and not `slot` or something like that?

libbpf-tools/cpudist.bpf.c
```diff
@@ -0,0 +1,124 @@
...
+	else
+		delta /= 1000;
+	loc = log2l(delta);
+	if (loc > MAX_SLOTS)
+		loc = MAX_SLOTS;
+	loc -= 1;
+	if (loc >= MAX_SLOTS)
+		return;
+	histp->slots[loc] += 1;
+	BPF_CORE_READ_STR_INTO(&histp->comm, task, comm);
```
anakryiko, line 101: you should read this just once, when you do initial update into hists above, otherwise it's wasteful

libbpf-tools/cpudist.c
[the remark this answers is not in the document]
anakryiko: sorry, can you point out specifically what's wrong there?

libbpf-tools/cpudist.c
```diff
@@ -0,0 +1,285 @@
...
+		printf("\n");
+
+		if (env.timestamp) {
+			time(&t);
+			tm = localtime(&t);
+			strftime(ts, sizeof(ts), "%H:%M:%S", tm);
+			printf("%-8s\n", ts);
+		}
+
+		if ((err = print_log2_hists(fd)) < 0)
```
anakryiko, line 274: nit: this is too clever and succinct. Sometimes it's necessary in while loops, but here just normal:
```
err = print_log2_hists(fd);
if (err)
    break;
```

libbpf-tools/trace_helpers.c
```diff
@@ -150,3 +150,70 @@ const struct ksym *ksyms__get_symbol(const struct ksyms *ksyms,
...
+
+	if (idx_max <= 32)
+		stars = stars_max;
+	else
+		stars = stars_max / 2;
+
+	for (i = 1; i < idx_max + 1; i++) {
+		low = ((long)1 << i) >> 1;
+		high = ((long)1 << i) - 1;
```
anakryiko, line 208: pro tip: `(long)1` is the same as `1L`. But it can be 32-bit on 32-bit arches. So I'd go with long long: `1LL`. And you probably want unsigned integers everywhere?

libbpf-tools/cpudist.bpf.c
ethercflow: > there shouldn't be `if (loc >= MAX_SLOTS) return;` at all:
> 
> ```
> loc = log2l(delta) - 1; /* or maybe just remove + 1 from log2l? */
> if (loc >= MAX_SLOTS) loc = MAX_SLOTS - 1;
> histp->slots[loc]++;
> ```

Thanks @anakryiko , I tired this before, but still get error:

> ; histp->slots[loc]++;
> 345: (61) r1 = *(u32 *)(r6 +0)
>  R0=inv(id=0,umax_value=16,var_off=(0x0; 0x1f)) R1_w=invP(id=0,umax_value=17179869180,var_off=(0x0; 0x3fffffffc)) R2=inv(id=0,umax_value=34,var_off=(0x0; 0x3f)) R3=inv35 R4=inv0 R6_w=map_value(id=0,off=0,ks=4,vs=160,umax_value=17179869180,var_off=(0x0; 0x3fffffffc)) R7=inv(id=0,umax_value=4294967295,var_off=(0x0; 0xffffffff)) R8=inv(id=0,umax_value=4294967295,var_off=(0x0; 0xffffffff)) R9=inv(id=0) R10=fp0 fp-8=mmmm???? fp-16=mmmmmmmm fp-24=mmmmmmmm fp-32=mmmmmmmm fp-40=mmmmmmmm fp-48=map_value
> R6 unbounded memory access, make sure to bounds check any array access into a map
> processed 154 insns (limit 1000000) max_states_per_insn 0 total_states 12 peak_states 12 mark_read 9

It seems `R2=inv(id=0,umax_value=34,var_off=(0x0; 0x3f))` is limited up to 34, but access array with R6.

> 
> btw, why `loc` and not `slot` or something like that?
Sure I'll change to `slot`, the reason is I used to think `slots[slot]` maybe looks not good.

libbpf-tools/cpudist.c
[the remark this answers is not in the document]
ethercflow: > sorry, can you point out specifically what's wrong there?

@anakryiko Sorry, I didn't describe it clearly. I found when I use HASHMAP,  no slot will >= MAX_SLOTS, but after I changed to ARRAY, it always have slot >= MAX_SLOTS. 

I ran with `cpudist -L 1` (iter hists 1s interval), I found only tid == 0 will produce slot >= MAX_SLOTS when map type is ARRAY.

```
tid = 0 swapper/36
               usecs                         : count     distribution
                   0 -> 1                    : 28       |***                 |
                   2 -> 3                    : 31       |****                |
                   4 -> 7                    : 154      |********************|
                   8 -> 15                   : 68       |********            |
                  16 -> 31                   : 58       |*******             |
                  32 -> 63                   : 100      |************        |
                  64 -> 127                  : 17       |**                  |
                 128 -> 255                  : 11       |*                   |
                 256 -> 511                  : 9        |*                   |
                 512 -> 1023                 : 16       |**                  |
                1024 -> 2047                 : 25       |***                 |
                2048 -> 4095                 : 18       |**                  |
                4096 -> 8191                 : 14       |*                   |
                8192 -> 16383                : 9        |*                   |
               16384 -> 32767                : 3        |                    |
               32768 -> 65535                : 0        |                    |
               65536 -> 131071               : 0        |                    |
              131072 -> 262143               : 0        |                    |
              262144 -> 524287               : 0        |                    |
              524288 -> 1048575              : 0        |                    |
             1048576 -> 2097151              : 0        |                    |
             2097152 -> 4194303              : 0        |                    |
             4194304 -> 8388607              : 0        |                    |
             8388608 -> 16777215             : 0        |                    |
            16777216 -> 33554431             : 0        |                    |
            33554432 -> 67108863             : 0        |                    |
            67108864 -> 134217727            : 0        |                    |
           134217728 -> 268435455            : 0        |                    |
           268435456 -> 536870911            : 0        |                    |
           536870912 -> 1073741823           : 0        |                    |
          1073741824 -> 2147483647           : 0        |                    |
          2147483648 -> 4294967295           : 0        |                    |
          4294967296 -> 8589934591           : 0        |                    |
          8589934592 -> 17179869183          : 0        |                    |
         17179869184 -> 34359738367          : 5        |                    |
```

libbpf-tools/cpudist.bpf.c
ethercflow: Right, sorry for this mistake.

libbpf-tools/trace_helpers.c
ethercflow: Yes, thank you for pointing this.

libbpf-tools/cpudist.c
ethercflow: Fixed, thanks.

libbpf-tools/trace_helpers.c
ethercflow: Fixed, thanks

libbpf-tools/cpudist.c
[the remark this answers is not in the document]
anakryiko: it's probably because bpf_map_lookup_elem() for ARRAY never returns NULL and check doesn't account for that. So your algorithm will assume that tid=0 was already timestamped or something. Either way, let's stick to HASHMAP for now and simplify our lives.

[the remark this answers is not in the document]
anakryiko: Actually, just running your code locally (with HASHMAP), I still got a really large value. Probably also from tid=0. I wonder if it makes sense to just filter that out? What's the behavior of BCC-based cpuhist in such case?

libbpf-tools/cpudist.bpf.c
anakryiko: Ok, so I played with this locally. The problem is that log2 and log2l returns u32, and slot itself is u32, but then it's used as an index into array, which is used as u64. So Clang generates bit extension:

```
r6 <<= 32
r6 >>= 32
```
which makes verifier lose track of the fact that `slot` is actually < 36. If you change `slot` type to u64 and log2/log2l return type to u64, it will work.

libbpf-tools/trace_helpers.c
```diff
@@ -150,3 +150,70 @@ const struct ksym *ksyms__get_symbol(const struct ksyms *ksyms,
...
+
+	for (i = 0; i < vals_size; i++) {
+		val = vals[i];
+		if (val > 0)
+			idx_max = i;
+		if (val > val_max)
+			val_max = val;
+	}
+
+	if (idx_max > 0) {
```
anakryiko, line 192: ok, so if there are no values, the idea is to emit nothing right? It would be way more explicit (and will reduce this nesting):

```
if (idx_max <= 0)
    return;
````

File: libbpf-tools/trace_helpers.c [hunk 21]
```diff
@@ -150,3 +150,70 @@ const struct ksym *ksyms__get_symbol(const struct ksyms *ksyms,
...
+		if (idx_max <= 32)
+			printf("     %-19s : count    distribution\n",
+				val_type);
+		else
+			printf("               %-29s : count     "
+				"distribution\n", val_type);
```
anakryiko, line 198: printf allows to express this better:
```
printf("     %-*s : count    distribution\n", idx_max <= 32 ? 19 : 29, val_type);
```

You can do similar trick to another if/else printf below. Sort of, initialize widths once, then use them in printf as width specifier without extra ifs.

libbpf-tools/trace_helpers.c
```diff
@@ -150,3 +150,70 @@ const struct ksym *ksyms__get_symbol(const struct ksyms *ksyms,
...
+		stars = stars_max;
+	else
+		stars = stars_max / 2;
+
+	for (i = 1; i < idx_max + 1; i++) {
+		low = (1ULL << i) >> 1;
+		high = (1ULL << i) - 1;
+		if (low == high)
+			low -= 1;
+		val = vals[i];
```
anakryiko, line 211: this will ignore slot #0, that doesn't sound right. 

I also think that your histogram logic is wrong and log2l is wrong as well. I printed out what log2 and log2l return  for i=0..32, see for yourself:
```
i=0 log2(0)=0 log2l(0)=1
i=1 log2(1)=0 log2l(1)=1
i=2 log2(2)=1 log2l(2)=2
i=3 log2(3)=1 log2l(3)=2
i=4 log2(4)=2 log2l(4)=3
i=5 log2(5)=2 log2l(5)=3
i=6 log2(6)=2 log2l(6)=3
i=7 log2(7)=2 log2l(7)=3
i=8 log2(8)=3 log2l(8)=4
i=9 log2(9)=3 log2l(9)=4
i=10 log2(10)=3 log2l(10)=4
i=11 log2(11)=3 log2l(11)=4
i=12 log2(12)=3 log2l(12)=4
i=13 log2(13)=3 log2l(13)=4
i=14 log2(14)=3 log2l(14)=4
i=15 log2(15)=3 log2l(15)=4
i=16 log2(16)=4 log2l(16)=5
i=17 log2(17)=4 log2l(17)=5
i=18 log2(18)=4 log2l(18)=5
i=19 log2(19)=4 log2l(19)=5
i=20 log2(20)=4 log2l(20)=5
i=21 log2(21)=4 log2l(21)=5
i=22 log2(22)=4 log2l(22)=5
i=23 log2(23)=4 log2l(23)=5
i=24 log2(24)=4 log2l(24)=5
i=25 log2(25)=4 log2l(25)=5
i=26 log2(26)=4 log2l(26)=5
i=27 log2(27)=4 log2l(27)=5
i=28 log2(28)=4 log2l(28)=5
i=29 log2(29)=4 log2l(29)=5
i=30 log2(30)=4 log2l(30)=5
i=31 log2(31)=4 log2l(31)=5
i=32 log2(32)=5 log2l(32)=6
```
log2l is always +1 from log2, this is completely wrong, I don't understand why log2l has +1 in there. I think log2() results make sense, though, given log2(X) + 1 should give minimum number of bits necessary to represent a given X.

So with that, values of 0 and 1 go into slot #0, while you ignore them and report values in slot #1 as [0, 1], while they should be [2..3], and so on.

So please fix that and double check that everything is accurate. It's hard to trust a tool that doesn't do math correctly :)

ethercflow: Thanks a lot, I'll do more check at the weekends to make sure everything is accurate, and thanks again for your patient guidance.

libbpf-tools/cpudist.bpf.c
ethercflow: Got it, thanks. Fixed.

libbpf-tools/trace_helpers.c
ethercflow: Fixed it, thanks.

libbpf-tools/trace_helpers.c
ethercflow: Fixed, thanks.

libbpf-tools/trace_helpers.c
ethercflow: > log2l is always +1 from log2, this is completely wrong, I don't understand why log2l has +1 in there. I think log2() results make sense, though, given log2(X) + 1 should give minimum number of bits necessary to represent a given X.

Sorry for this! I migrate hist and log2 from bcc version directly. I thought it's ok, so I didn't think about it carefully.  I found https://github.com/iovisor/bcc/issues/1649 today, it looks like a history's legac. 

I've fixed with zero slot as beginning. Thanks.

libbpf-tools/cpudist.c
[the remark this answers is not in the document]
ethercflow: > Actually, just running your code locally (with HASHMAP), I still got a really large value. Probably also from tid=0. I wonder if it makes sense to just filter that out? What's the behavior of BCC-based cpuhist in such case?

@anakryiko with HASHMAP, I ran`crudest -p 0`  for 15 minutes, after ctrl-c, the result is:

```
^C
         0 -> 1          : 25823    |*****                                   |
         2 -> 3          : 30616    |******                                  |
         4 -> 7          : 49292    |**********                              |
         8 -> 15         : 88110    |******************                      |
        16 -> 31         : 123735   |**************************              |
        32 -> 63         : 122768   |*************************               |
        64 -> 127        : 189345   |****************************************|
       128 -> 255        : 34011    |*******                                 |
       256 -> 511        : 26639    |*****                                   |
       512 -> 1023       : 18085    |***                                     |
      1024 -> 2047       : 18400    |***                                     |
      2048 -> 4095       : 54752    |***********                             |
      4096 -> 8191       : 25210    |*****                                   |
      8192 -> 16383      : 9261     |*                                       |
     16384 -> 32767      : 12301    |**                                      |
     32768 -> 65535      : 6091     |*                                       |
     65536 -> 131071     : 446      |                                        |
    131072 -> 262143     : 31       |                                        |
```

No large value found. I tested on a physical machine:

```
CPU(s):              40
On-line CPU(s) list: 0-39
Thread(s) per core:  2
Core(s) per socket:  10
Socket(s):           2
NUMA node(s):        2
```

The bcc version for almost half an hour.

```
➜  tools git:(libbpf-tools) ✗ ./cpudist.py -p 0
^C
         0 -> 1          : 64033    |******                                  |
         2 -> 3          : 59896    |******                                  |
         4 -> 7          : 96904    |**********                              |
         8 -> 15         : 176765   |******************                      |
        16 -> 31         : 247373   |*************************               |
        32 -> 63         : 248718   |*************************               |
        64 -> 127        : 387502   |****************************************|
       128 -> 255        : 68589    |*******                                 |
       256 -> 511        : 53117    |*****                                   |
       512 -> 1023       : 37782    |***                                     |
      1024 -> 2047       : 37948    |***                                     |
      2048 -> 4095       : 109874   |***********                             |
      4096 -> 8191       : 50249    |*****                                   |
      8192 -> 16383      : 18970    |*                                       |
     16384 -> 32767      : 24702    |**                                      |
     32768 -> 65535      : 13212    |*                                       |
     65536 -> 131071     : 1151     |                                        |
    131072 -> 262143     : 67       |                                        |

```

libbpf-tools/cpudist.bpf.c
```diff
@@ -0,0 +1,120 @@
...
+	}
+	delta = ts - *tsp;
+	if (targ_ms)
+		delta /= 1000000;
+	else
+		delta /= 1000;
+	slot = log2l(delta);
+	if (slot >= MAX_SLOTS)
+		slot = MAX_SLOTS - 1;
+	histp->slots[slot] += 1;
```
anakryiko, line 97: I just realized that for per-process mode (due to multi-threaded applications) and especially for the system-wide mode, this increment is quite racy. Consider this a follow up, but I think using atomic increment here would be probably better for accuracy. See `__sync_fetch_and_add()` use in selftests.

libbpf-tools/cpudist.c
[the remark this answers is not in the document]
anakryiko: I mean, this will depend on specific machine's workload, but sure, let's see if that becomes a problem.

anakryiko:
[buildbot, test this please]

anakryiko:
@ethercflow, can you please rebase and update this PR, now that syscount was merged?

ethercflow:
> @ethercflow, can you please rebase and update this PR, now that syscount was merged?

@anakryiko Sure, rebase and updated. PTAL again, Thanks.

libbpf-tools/cpudist.bpf.c
ethercflow: Fixed, thanks.

ethercflow:
> I don't know if you've followed syscount discussion in [another PR](https://github.com/iovisor/bcc/pull/2920), but your solution suffers same kind of race between `bpf_map_lookup_elem()`/`bpf_map_delete_elem()`, which is sort of inherent for this kind of tools. I don't think that should block this to go in or even whether we need to address it, but just keep that in mind for the future.

Thanks, I've followed your dissuasion, I'll keep that in mind for the future. :)

ethercflow:
Hi @yonghong-song I think this may be merged. PTAL, thank you. :)

yonghong-song:
[buildbot, test this please]

libbpf-tools/cpudist.bpf.c
ethercflow: > Ok, so I played with this locally. The problem is that log2 and log2l returns u32, and slot itself is u32, but then it's used as an index into array, which is used as u64. So Clang generates bit extension:
> 
> 
> 
> ```
> 
> r6 <<= 32
> 
> r6 >>= 32
> 
> ```
> 
> which makes verifier lose track of the fact that `slot` is actually < 36. If you change `slot` type to u64 and log2/log2l return type to u64, it will work.

Hi @anakryiko I just found LLVM-7.0 support alu32 to use 32bit subregister, then I tested Makefile with `$(Q)$(CLANG) -g -O2 -target bpf -Xclang -target-feature -Xclang +alu32` and change `log2(l)` return val’s, slot’s type back to `u32`,  found it’s ok with it. So is it better to change to `-Xclang -target-feature -Xclang +alu32`?

anakryiko: There were some pretty recent fixes to verifier related to use of alu32. If we switch libbpf-tools to alu32, they might not run successfully on as wide range of kernels as currently. So I'd defer it for now and stick to no-alu32.

ethercflow: Get it, thanks!

PR #2912, opened 2020-05-12
libbpf-tools: convert BCC vfsstat to libbpf-based version

aspsk:
The vfsstat tool actually doesn't need to use CO-RE, add it as an
example of using batch functions to read/update arrays.

### DIFF
--- a/libbpf-tools/vfsstat.bpf.c
+++ b/libbpf-tools/vfsstat.bpf.c
@@ -1,0 +1,48 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2020 Anton Protopopov
+//
+// Based on vfsstat(8) from BCC by Brendan Gregg
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include "vfsstat.h"
+
+__u64 stats[S_MAXSTAT] = {};
+
+static __always_inline int inc_stats(int key)
+{
+	__atomic_add_fetch(&stats[key], 1, __ATOMIC_RELAXED);
+	return 0;
+}
+
+SEC("kprobe/vfs_read")
+int BPF_KPROBE(kprobe__vfs_read)
+{
+	return inc_stats(S_READ);
+}
+
+SEC("kprobe/vfs_write")
+int BPF_KPROBE(kprobe__vfs_write)
+{
+	return inc_stats(S_WRITE);
+}
+
+SEC("kprobe/vfs_fsync")
+int BPF_KPROBE(kprobe__vfs_fsync)
+{
+	return inc_stats(S_FSYNC);
+}
+
+SEC("kprobe/vfs_open")
+int BPF_KPROBE(kprobe__vfs_open)
+{
+	return inc_stats(S_OPEN);
+}
+
+SEC("kprobe/vfs_create")
+int BPF_KPROBE(kprobe__vfs_create)
+{
+	return inc_stats(S_CREATE);
+}
+
+char LICENSE[] SEC("license") = "GPL";
--- a/libbpf-tools/vfsstat.c
+++ b/libbpf-tools/vfsstat.c
@@ -1,0 +1,188 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2020 Anton Protopopov
+//
+// Based on vfsstat(8) from BCC by Brendan Gregg
+#include <sys/resource.h>
+#include <argp.h>
+#include <unistd.h>
+#include <time.h>
+#include <bpf/bpf.h>
+#include "vfsstat.h"
+#include "vfsstat.skel.h"
+
+const char *argp_program_version = "vfsstat 0.1";
+const char *argp_program_bug_address = "<bpf@vger.kernel.org>";
+static const char argp_program_doc[] =
+	"\nvfsstat: Count some VFS calls\n"
+	"\n"
+	"EXAMPLES:\n"
+	"    vfsstat      # interval one second\n"
+	"    vfsstat 5 3  # interval five seconds, three output lines\n";
+static char args_doc[] = "[interval [count]]";
+
+static const struct argp_option opts[] = {
+	{ "verbose", 'v', NULL, 0, "Verbose debug output" },
+	{},
+};
+
+static struct env {
+	bool verbose;
+	int count;
+	int interval;
+} env = {
+	.interval = 1,	/* once a second */
+};
+
+static error_t parse_arg(int key, char *arg, struct argp_state *state)
+{
+	long interval;
+	long count;
+
+	switch (key) {
+	case 'v':
+		env.verbose = true;
+		break;
+	case ARGP_KEY_ARG:
+		switch (state->arg_num) {
+		case 0:
+			errno = 0;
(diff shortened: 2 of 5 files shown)


brendangregg:
This tool needs a switch to breakdown by file type to be really useful. I've been meaning to add it to the Python version but haven't yet. These VFS operations could be files, sockets, /proc entries, etc, etc. 

What was the reason to port it to CO-RE?

aspsk:
> This tool needs a switch to breakdown by file type to be really useful. I've been meaning to add it to the Python version but haven't yet. These VFS operations could be files, sockets, /proc entries, etc, etc.

Is it possible to do in BPF? E.g., you can tell that `(struct file *)filep` is a socket if `sock_from_file(filep) != NULL`, but how you can access this info from a BPF program?
 
> What was the reason to port it to CO-RE?

I chose it randomly, just to train before porting anything meaningful and to get some feedback on coding style, etc.

In general, what is the purpose of `libbpf-tools`? Are they going to replace the BCC-based tools eventually or just are a set of examples on how to use libbpf CO-RE? If the latter is true, then are there any particular tools which you are interested in porting to CO-RE?

libbpf-tools/vfsstat.bpf.c
```diff
@@ -0,0 +1,57 @@
...
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include "vfsstat.h"
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__uint(max_entries, S_MAXSTAT);
+	__type(key, int);
+	__type(value, u64);
+} stats SEC(".maps");
```
anakryiko, line 15: you can use global array instead, given it's just a fixed-sized array. Try:

```
u64 stats[S_MAX_STAT] = {};
```

Then in code just use it as normal array (no bpf_map_lookup_elem/bpf_map_update_elem necesasry). From user-space, you can access it again as a normal array through `obj->bss->stats[i]`.

libbpf-tools/vfsstat.bpf.c
```diff
@@ -0,0 +1,57 @@
...
+static __always_inline int inc_stats(int key)
+{
+	u64 *val;
+
+	val = bpf_map_lookup_elem(&stats, &key);
+	if (val)
+		++*val;
+	return 0;
+}
+
```
anakryiko, line 26: This is actually racy when multiple CPUs simultaneously do same vfs operation type. I'd suggest using atomic increments. Together with above suggestion to use plain array, it would look like this:

```
__atomic_add_fetch(&stats[key], 1);
```

libbpf-tools/vfsstat.c
```diff
@@ -0,0 +1,210 @@
...
+{
+	struct rlimit rlim_new = {
+		.rlim_cur	= RLIM_INFINITY,
+		.rlim_max	= RLIM_INFINITY,
+	};
+
+	return setrlimit(RLIMIT_MEMLOCK, &rlim_new);
+}
+
+static const char *Strftime(char *s, size_t max, const char *format)
```
anakryiko, line 92: nit: this naming goes against conventions (and is confusing as towards how it is different from strftime). Maybe something like `strftime_now` or something along those lines? Also I'd keep order of format, max, s, as in original strftime, for consistency.

libbpf-tools/vfsstat.c
```diff
@@ -0,0 +1,210 @@
...
+static void print_and_reset_stats(int fd)
+{
+	__u32 count = S_MAXSTAT;
+	__u64 vals[S_MAXSTAT];
+	int keys[S_MAXSTAT];
+	void *batch;
+	char s[16];
+	int err;
+
+	err = bpf_map_lookup_batch(fd, NULL, &batch, keys, vals, &count, NULL);
```
anakryiko, line 135: batch is very recent APIs, this will limit which kernel versions are going to be supported. With using normal array on BPF side, you don't need batching, and your code actually going to be faster due to memory-mapped array in user-space.

libbpf-tools/vfsstat.c
```diff
@@ -0,0 +1,210 @@
...
+		return 1;
+	}
+
+	obj = vfsstat_bpf__open();
+	if (!obj) {
+		fprintf(stderr, "failed to open and/or load BPF object\n");
+		return 1;
+	}
+
+	err = vfsstat_bpf__load(obj);
```
anakryiko, line 186: can be combined in one `vfsstat_bpf__open_and_load()`

libbpf-tools/vfsstat.c
```diff
@@ -0,0 +1,210 @@
...
+		return;
+	}
+
+	printf("%-8s: ", Strftime(s, sizeof(s), "%H:%M:%S"));
+	for (int i = 0; i < S_MAXSTAT; i++)
+		printf(" %8llu", vals[i] / env.interval);
+	printf("\n");
+
+	memset(vals, 0, sizeof(vals));
+	err = bpf_map_update_batch(fd, keys, vals, &count, NULL);
```
anakryiko, line 148: to do it in race-free way, you need to use atomic xchg:

```
for each key:
   stat_val = __atomic_exchange_n(&obj->bss->stats[key], 0, __ATOMIC_RELAXED);
```

libbpf-tools/vfsstat.c
aspsk: Thanks, renamed it to `strftime_now`. As for the order of args, it was actually taken from the original `strftime`:
```
size_t strftime(char *s, size_t max, const char *format, const struct tm *tm);
```

aspsk:
Thanks for your comments @anakryiko! All addressed. Also, I have an almost finished port of `syscount.py`, will push it next weekend if you are ok to review.

I have the same question as above to you about the purpose of libbpf-tools in bcc? Is it going to replace the BCC-based tools eventually or just is a set of examples on how to use libbpf CO-RE?

libbpf-tools/vfsstat.c
```diff
@@ -0,0 +1,186 @@
...
+	long count;
+
+	switch (key) {
+	case 'v':
+		env.verbose = true;
+		break;
+	case ARGP_KEY_ARG:
+		switch (state->arg_num) {
+		case 0:
+			interval = strtol(arg, NULL, 10);
```
anakryiko, line 48: forgot to mention in first round, you have to set errno to zero here, otherwise you can erroneously detect errno != 0 from some previous invocation.

anakryiko:
> Thanks for your comments @anakryiko! All addressed. Also, I have an almost finished port of `syscount.py`, will push it next weekend if you are ok to review.

sounds good.

> I have the same question as above to you about the purpose of libbpf-tools in bcc? Is it going to replace the BCC-based tools eventually or just is a set of examples on how to use libbpf CO-RE?

CO-RE-based (and, more generally, libbpf-based) tools, are meant to be a lightweight alternatives to existing BCC tools, not requiring heavy-weight Python and Clang/LLVM runtime dependencies. So no, it's definitely not just toy examples.

libbpf-tools/vfsstat.c
aspsk: Thanks @anakryiko, fixed here and below.

yonghong-song:
[buildbot, test this please]

PR #2918, opened 2020-05-19
tools/opensnoop: Fix support for kfunc

mauriciovasquezbernal:
The support for kfunc is broken because the filters are returning an integer
when they should be returning void, this commit fixes that by returning 
correct value for both cases.

### DIFF
--- a/tools/opensnoop.py
+++ b/tools/opensnoop.py
@@ -182,7 +182,7 @@ KRETFUNC_PROBE(do_sys_open, int dfd, const char __user *filename,
 #if CGROUPSET
     u64 cgroupid = bpf_get_current_cgroup_id();
     if (cgroupset.lookup(&cgroupid) == NULL) {
-      return 0;
+      return;
     }
 #endif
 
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
@mauriciovasquezbernal You are right that the return value for fentry/fexit bpf programs are not used. 
Previously, the return value can be anything and will be ignored.
But, the commit below
https://kernel.googlesource.com/pub/scm/linux/kernel/git/bpf/bpf/+/e92888c72fbdc6f9d07b3b0604c012e81d7c0da7
tries to enforce return value 0.

The kernel verifier will enforce that a return value 0 at reg r0. So returning nothing may break the program if "r0" is undefined or happens to be a non-zero value.

So I think we should keep "return 0" there.

mauriciovasquezbernal:
@yonghong-song you're totally right, I took the wrong path to solve this. I opened a new PR (https://github.com/iovisor/bcc/pull/2924) that hopefully solves the root of the problem.

PR #2920, opened 2020-05-20
Syscount libbpf

aspsk:
This PR converts BCC syscount to BPF CO-RE version and fixes some simple typos

### DIFF
--- a/libbpf-tools/syscall_helpers.c
+++ b/libbpf-tools/syscall_helpers.c
@@ -1,0 +1,526 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+// Copyright (c) 2020 Anton Protopopov
+#include <stdlib.h>
+#include <string.h>
+#include <limits.h>
+#include <stdio.h>
+#include <errno.h>
+
+static const char **syscall_names;
+static size_t syscall_names_size;
+
+#define warn(...) fprintf(stderr, __VA_ARGS__)
+#define MAX(x, y) (((x) > (y)) ? (x) : (y))
+
+static const char *parse_syscall(const char *buf, int *number)
+{
+	char *end;
+	long x;
+
+	errno = 0;
+	x = strtol(buf, &end, 10);
+	if (errno) {
+		warn("strtol(%s): %s\n", buf, strerror(errno));
+		return NULL;
+	} else if (end == buf) {
+		warn("strtol(%s): no digits found\n", buf);
+		return NULL;
+	} else if (x < 0 || x > INT_MAX) {
+		warn("strtol(%s): bad syscall number: %ld\n", buf, x);
+		return NULL;
+	}
+	if (*end != '\t') {
+		warn("bad input: %s (expected <num>\t<name>)\n", buf);
+		return NULL;
+	}
+
+	*number = x;
+	return ++end;
+}
+
+void init_syscall_names(void)
+{
+	size_t old_size, size = 1024;
+	const char *name;
+	char buf[64];
+	int number;
+	int err;
+	FILE *f;
+
+	f = popen("ausyscall --dump 2>/dev/null", "r");
+	if (!f) {
+		warn("popen: ausyscall --dump: %s\n", strerror(errno));
+		return;
+	}
+
+	syscall_names = calloc(size, sizeof(char *));
+	if (!syscall_names) {
+		warn("calloc: %s\n", strerror(errno));
+		goto close;
+	}
+
+	/* skip the header */
+	fgets(buf, sizeof(buf), f);
+
+	while (fgets(buf, sizeof(buf), f)) {
+		if (buf[strlen(buf) - 1] == '\n')
+			buf[strlen(buf) - 1] = '\0';
+
+		name = parse_syscall(buf, &number);
+		if (!name || !name[0])
+			goto close;
+
+		/* In a rare case when syscall number is > than initial 1024 */
+		if (number >= size) {
+			old_size = size;
+			size = 1024 * (1 + number / 1024);
+			syscall_names = realloc(syscall_names,
+						size * sizeof(char *));
+			if (!syscall_names) {
+				warn("realloc: %s\n", strerror(errno));
+				goto close;
+			}
+			memset(syscall_names+old_size, 0,
+			       (size - old_size) * sizeof(char *));
+		}
+
+		if (syscall_names[number]) {
+			warn("duplicate number: %d (stored: %s)",
+				number, syscall_names[number]);
+			goto close;
+		}
+
+		syscall_names[number] = strdup(name);
+		if (!syscall_names[number]) {
+			warn("strdup: %s\n", strerror(errno));
+			goto close;
+		}
+		syscall_names_size = MAX(number+1, syscall_names_size);
+	}
+
+	if (ferror(f))
+		warn("fgets: %s\n", strerror(errno));
+close:
+	err = pclose(f);
+	if (err < 0)
+		warn("pclose: %s\n", strerror(errno));
+#ifndef __x86_64__
+	/* Ignore the error for x86_64 where we have a table compiled in */
+	else if (err && WEXITSTATUS(err) == 127) {
+		warn("ausyscall required for syscalls number/name mapping\n");
+	} else if (err) {
+		warn("ausyscall exit status (see wait(2)): 0x%x\n", err);
+	}
+#endif
+}
+
+void free_syscall_names(void)
+{
+	for (size_t i = 0; i < syscall_names_size; i++)
+		free((void *) syscall_names[i]);
+	free(syscall_names);
+}
+
+/*
+ * Syscall table for Linux x86_64.
+ *
+ * Semi-automatically generated from strace/linux/x86_64/syscallent.h and
+ * linux/syscallent-common.h using the following commands:
+ *
+ *  awk -F\" '/SEN/{printf("%d %s\n", substr($0,2,3), $(NF-1));}' syscallent.h
+ *  awk '/SEN/ { printf("%d %s\n", $3, $9); }' syscallent-common.h
+ *
+ * (The idea is taken from src/python/bcc/syscall.py.)
+ */
+#ifdef __x86_64__
+static const char *syscall_names_x86_64[] = {
+	[0] = "read",
+	[1] = "write",
+	[2] = "open",
+	[3] = "close",
+	[4] = "stat",
+	[5] = "fstat",
+	[6] = "lstat",
+	[7] = "poll",
+	[8] = "lseek",
+	[9] = "mmap",
+	[10] = "mprotect",
+	[11] = "munmap",
+	[12] = "brk",
+	[13] = "rt_sigaction",
+	[14] = "rt_sigprocmask",
+	[15] = "rt_sigreturn",
+	[16] = "ioctl",
+	[17] = "pread64",
+	[18] = "pwrite64",
+	[19] = "readv",
+	[20] = "writev",
+	[21] = "access",
+	[22] = "pipe",
+	[23] = "select",
+	[24] = "sched_yield",
+	[25] = "mremap",
+	[26] = "msync",
+	[27] = "mincore",
+	[28] = "madvise",
+	[29] = "shmget",
+	[30] = "shmat",
+	[31] = "shmctl",
+	[32] = "dup",
+	[33] = "dup2",
+	[34] = "pause",
+	[35] = "nanosleep",
+	[36] = "getitimer",
+	[37] = "alarm",
+	[38] = "setitimer",
+	[39] = "getpid",
+	[40] = "sendfile",
+	[41] = "socket",
+	[42] = "connect",
+	[43] = "accept",
+	[44] = "sendto",
+	[45] = "recvfrom",
+	[46] = "sendmsg",
+	[47] = "recvmsg",
+	[48] = "shutdown",
+	[49] = "bind",
+	[50] = "listen",
+	[51] = "getsockname",
+	[52] = "getpeername",
+	[53] = "socketpair",
+	[54] = "setsockopt",
+	[55] = "getsockopt",
+	[56] = "clone",
+	[57] = "fork",
+	[58] = "vfork",
+	[59] = "execve",
+	[60] = "exit",
+	[61] = "wait4",
+	[62] = "kill",
+	[63] = "uname",
+	[64] = "semget",
+	[65] = "semop",
+	[66] = "semctl",
+	[67] = "shmdt",
+	[68] = "msgget",
+	[69] = "msgsnd",
+	[70] = "msgrcv",
+	[71] = "msgctl",
+	[72] = "fcntl",
+	[73] = "flock",
+	[74] = "fsync",
+	[75] = "fdatasync",
+	[76] = "truncate",
+	[77] = "ftruncate",
+	[78] = "getdents",
+	[79] = "getcwd",
+	[80] = "chdir",
+	[81] = "fchdir",
+	[82] = "rename",
+	[83] = "mkdir",
+	[84] = "rmdir",
+	[85] = "creat",
+	[86] = "link",
+	[87] = "unlink",
+	[88] = "symlink",
+	[89] = "readlink",
+	[90] = "chmod",
+	[91] = "fchmod",
+	[92] = "chown",
+	[93] = "fchown",
+	[94] = "lchown",
+	[95] = "umask",
+	[96] = "gettimeofday",
+	[97] = "getrlimit",
+	[98] = "getrusage",
+	[99] = "sysinfo",
+	[100] = "times",
+	[101] = "ptrace",
+	[102] = "getuid",
+	[103] = "syslog",
+	[104] = "getgid",
+	[105] = "setuid",
+	[106] = "setgid",
+	[107] = "geteuid",
+	[108] = "getegid",
+	[109] = "setpgid",
+	[110] = "getppid",
+	[111] = "getpgrp",
+	[112] = "setsid",
+	[113] = "setreuid",
+	[114] = "setregid",
+	[115] = "getgroups",
+	[116] = "setgroups",
+	[117] = "setresuid",
+	[118] = "getresuid",
+	[119] = "setresgid",
+	[120] = "getresgid",
+	[121] = "getpgid",
+	[122] = "setfsuid",
+	[123] = "setfsgid",
+	[124] = "getsid",
+	[125] = "capget",
+	[126] = "capset",
+	[127] = "rt_sigpending",
+	[128] = "rt_sigtimedwait",
+	[129] = "rt_sigqueueinfo",
+	[130] = "rt_sigsuspend",
+	[131] = "sigaltstack",
+	[132] = "utime",
+	[133] = "mknod",
+	[134] = "uselib",
+	[135] = "personality",
+	[136] = "ustat",
+	[137] = "statfs",
+	[138] = "fstatfs",
+	[139] = "sysfs",
+	[140] = "getpriority",
+	[141] = "setpriority",
+	[142] = "sched_setparam",
+	[143] = "sched_getparam",
+	[144] = "sched_setscheduler",
+	[145] = "sched_getscheduler",
+	[146] = "sched_get_priority_max",
+	[147] = "sched_get_priority_min",
+	[148] = "sched_rr_get_interval",
+	[149] = "mlock",
+	[150] = "munlock",
+	[151] = "mlockall",
+	[152] = "munlockall",
+	[153] = "vhangup",
+	[154] = "modify_ldt",
+	[155] = "pivot_root",
+	[156] = "_sysctl",
+	[157] = "prctl",
+	[158] = "arch_prctl",
+	[159] = "adjtimex",
+	[160] = "setrlimit",
+	[161] = "chroot",
+	[162] = "sync",
+	[163] = "acct",
+	[164] = "settimeofday",
+	[165] = "mount",
+	[166] = "umount2",
+	[167] = "swapon",
+	[168] = "swapoff",
+	[169] = "reboot",
+	[170] = "sethostname",
+	[171] = "setdomainname",
+	[172] = "iopl",
+	[173] = "ioperm",
+	[174] = "create_module",
+	[175] = "init_module",
+	[176] = "delete_module",
+	[177] = "get_kernel_syms",
+	[178] = "query_module",
+	[179] = "quotactl",
+	[180] = "nfsservctl",
+	[181] = "getpmsg",
+	[182] = "putpmsg",
+	[183] = "afs_syscall",
+	[184] = "tuxcall",
+	[185] = "security",
+	[186] = "gettid",
+	[187] = "readahead",
+	[188] = "setxattr",
+	[189] = "lsetxattr",
+	[190] = "fsetxattr",
+	[191] = "getxattr",
+	[192] = "lgetxattr",
+	[193] = "fgetxattr",
+	[194] = "listxattr",
+	[195] = "llistxattr",
+	[196] = "flistxattr",
+	[197] = "removexattr",
+	[198] = "lremovexattr",
+	[199] = "fremovexattr",
+	[200] = "tkill",
+	[201] = "time",
+	[202] = "futex",
+	[203] = "sched_setaffinity",
+	[204] = "sched_getaffinity",
+	[205] = "set_thread_area",
+	[206] = "io_setup",
+	[207] = "io_destroy",
+	[208] = "io_getevents",
+	[209] = "io_submit",
+	[210] = "io_cancel",
+	[211] = "get_thread_area",
+	[212] = "lookup_dcookie",
+	[213] = "epoll_create",
+	[214] = "epoll_ctl_old",
+	[215] = "epoll_wait_old",
+	[216] = "remap_file_pages",
+	[217] = "getdents64",
+	[218] = "set_tid_address",
+	[219] = "restart_syscall",
+	[220] = "semtimedop",
+	[221] = "fadvise64",
+	[222] = "timer_create",
+	[223] = "timer_settime",
+	[224] = "timer_gettime",
+	[225] = "timer_getoverrun",
+	[226] = "timer_delete",
+	[227] = "clock_settime",
+	[228] = "clock_gettime",
+	[229] = "clock_getres",
+	[230] = "clock_nanosleep",
+	[231] = "exit_group",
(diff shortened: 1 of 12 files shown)


aspsk:
Hi @anakryiko , here's a CO-RE port of `syscount`. One thing which is different from the original BCC-based `syscount` is that I am attaching both tracepoints even if there is no latency measurement requested. I am doing this because there is no official way to not load/attach a particular program using "skel" api. (It seems that it is probably hard to avoid loading the whole object, but it is pretty simple to skip attaching particular programs in the `bpf_object__attach_skeleton()` attach loop.)

anakryiko:
> Hi @anakryiko , here's a CO-RE port of `syscount`.

Thanks, will review shortly.

>One thing which is different from the original BCC-based `syscount` is that I am attaching both tracepoints even if there is no latency measurement requested. I am doing this because there is no official way to not load/attach a particular program using "skel" api. (It seems that it is probably hard to avoid loading the whole object, but it is pretty simple to skip attaching particular programs in the `bpf_object__attach_skeleton()` attach loop.)

Well, skeleton's attach() step is optional, you can attach each program individually with bpf_program__attach(skel->progs.whichever). 

It's true, though, that there's no way of skipping loading and verifying program right now. There was patch on the list to set specific BPF program to not auto-load, but it wasn't landed. I might add it shortly, it's really trivial and has come up before.

libbpf-tools/syscount.bpf.c
```diff
@@ -0,0 +1,103 @@
...
+	ts = bpf_ktime_get_ns();
+	bpf_map_update_elem(&start, &id, &ts, 0);
+	return 0;
+}
+
+SEC("tracepoint/raw_syscalls/sys_exit")
+int tracepoint__raw_syscalls__sys_exit(struct trace_event_raw_sys_exit *args)
+{
+	u64 id = bpf_get_current_pid_tgid();
+	struct data_t zero = {};
```
anakryiko, line 71: this is used for initialization only, right? So it's better to make it a static variable and skip zero-initializing it on every call

libbpf-tools/syscount.bpf.c
```diff
@@ -0,0 +1,103 @@
...
+	struct data_t *val;
+	u64 *start_ts;
+	u32 key;
+
+	/* this happens when there is an interrupt */
+	if (args->id == -1)
+		return 0;
+
+	if (filter_pid && id >> 32 != filter_pid)
```
anakryiko, line 80: please parens around bit manipulation operations to avoid thinking about operator precedence

libbpf-tools/syscount.bpf.c
```diff
@@ -0,0 +1,103 @@
...
+		return 0;
+	if (filter_failed && args->ret >= 0)
+		return 0;
+	if (filter_errno && args->ret != -filter_errno)
+		return 0;
+
+	if (measure_latency) {
+		start_ts = bpf_map_lookup_elem(&start, &id);
```
anakryiko, line 88: there is no need to use entire 64-bits (tgid + tid) as a key, tid should be enough.

libbpf-tools/syscount.bpf.c
```diff
@@ -0,0 +1,103 @@
...
+		if (!start_ts)
+			return 0;
+	}
+
+	key = (count_by_process) ? id >> 32 : args->id;
```
anakryiko, line 93: you use (id >> 32) twice, I think it's better to have a local variable `tgid` instead.

libbpf-tools/syscount.c
```diff
@@ -0,0 +1,1167 @@
...
+{
+	char *end;
+	int val;
+
+	errno = 0;
+	val = strtol(arg, &end, 10);
+	if (errno) {
+		warn("strtol: %s: %s\n", arg, strerror(errno));
+		return -1;
+	} else if (end == arg || val < min || val > max) {
```
anakryiko, line 76: wouldn't `*end == 0` be a better check that entire string argument was a valid integer, parsed completely?

libbpf-tools/syscount.c
```diff
@@ -0,0 +1,1167 @@
...
+	struct rlimit rlim_new = {
+		.rlim_cur	= RLIM_INFINITY,
+		.rlim_max	= RLIM_INFINITY,
+	};
+
+	return setrlimit(RLIMIT_MEMLOCK, &rlim_new);
+}
+
+#ifdef __x86_64__
+static int errno_by_name_x86_64(const char *errno_name)
```
anakryiko, line 104: how about making it a bit cleaner and general by having a list of structs, like this:

```
static struct { int err_code; const char *err_name; } errs = {
    {1, "EPERM"},
    {2, "ENOENT"},
...
};
```
This seems like something that might be used by other tools eventually and will allow bi-directional lookups. code -> name and name -> code. And will avoid any unnecessary macro.

If necessary, it would also be possible to have (optionally) case-insensitive search.

libbpf-tools/syscount.c
```diff
@@ -0,0 +1,1167 @@
...
+	strcase("EMEDIUMTYPE", 124);
+	strcase("ECANCELED", 125);
+	strcase("ENOKEY", 126);
+	strcase("EKEYEXPIRED", 127);
+	strcase("EKEYREVOKED", 128);
+	strcase("EKEYREJECTED", 129);
+	strcase("EOWNERDEAD", 130);
+	strcase("ENOTRECOVERABLE", 131);
+	strcase("ERFKILL", 132);
+	strcase("EHWPOISON", 133);
```
anakryiko, line 242: given it's so long, I'd extract it into a separate header file, honestly. Same for syscall names below. It distracts from the code flow enormously.

libbpf-tools/syscount.c
```diff
@@ -0,0 +1,1167 @@
...
+
+#undef strcase
+
+	return -1;
+
+}
+#endif
+
+/* Try to find the errno number using the errno(1) program */
+static int errno_by_name_dynamic(const char *errno_name)
```
anakryiko, line 252: is there a need for this? For unknown ENAME, why not just allowing specifying errno by integer value instead? This seems bloated and unnecessarily dangerous and complicated.

libbpf-tools/syscount.c
```diff
@@ -0,0 +1,1167 @@
...
+static const char **syscall_names;
+static size_t syscall_names_size;
+
+static const char *parse_syscall(const char *buf, int *number)
+{
+	char *end;
+	long x;
+
+	errno = 0;
+	x = strtol(buf, &end, 10);
```
anakryiko, line 330: use get_int?

libbpf-tools/syscount.c
```diff
@@ -0,0 +1,1167 @@
...
+ * Semi-automatically generated from strace/linux/x86_64/syscallent.h and
+ * linux/syscallent-common.h using the following commands:
+ *
+ *  awk -F\" '/SEN/{printf("%d %s\n", substr($0,2,3), $(NF-1));}' syscallent.h
+ *  awk '/SEN/ { printf("%d %s\n", $3, $9); }' syscallent-common.h
+ *
+ * (The idea is taken from src/python/bcc/syscall.py.)
+ */
+#ifdef __x86_64__
+static const char *syscall_names_x86_64[] = {
```
anakryiko, line 445: same as for errno, can you please split it into a separate file?

libbpf-tools/syscount.c
```diff
@@ -0,0 +1,1167 @@
...
+	return x > y ? -1 : !(x == y);
+}
+
+static const char *comm(pid_t pid, char *buf, size_t size)
+{
+	char path[32];
+	int fd;
+	int n;
+
+	snprintf(path, sizeof(path), "/proc/%d/comm", pid);
```
anakryiko, line 856: this pid to comm mapping from procfs is racy, process could be dead or PID can even be reused in extreme cases. BPF code can easily capture this, so why not do that?

libbpf-tools/syscount.c
```diff
@@ -0,0 +1,1167 @@
...
+{
+	double div = env.milliseconds ? 1000000.0 : 1000.0;
+	char buf[32];
+
+	print_latency_header();
+	for (int i = 0; i < count && i < env.top; i++)
+		printf("%-22s %8llu %16.3lf\n",
+				agg_col(vals[i].key, buf, sizeof(buf)),
+				vals[i].count,
+				vals[i].total_ns / div);
```
anakryiko, line 923: formatting looks off, same in print_count. Is it just how Github renders it or it's really off?

libbpf-tools/syscount.c
```diff
@@ -0,0 +1,1167 @@
...
+	if (env.filter_errno)
+		obj->rodata->filter_errno = env.filter_errno;
+
+	err = syscount_bpf__load(obj);
+	if (err) {
+		warn("failed to load BPF object: %s\n", strerror(-err));
+		goto cleanup_obj;
+	}
+
+	err = syscount_bpf__attach(obj);
```
anakryiko, line 1125: so here you don't have to attach everything, you can control which programs need to be attached and then use 
```
bpf_program__attach(obj->progs.whichone);
```

libbpf-tools/syscount.bpf.c
```diff
@@ -0,0 +1,103 @@
...
+	id = bpf_get_current_pid_tgid();
+	if (filter_pid && id >> 32 != filter_pid)
+		return 0;
+
+	ts = bpf_ktime_get_ns();
+	bpf_map_update_elem(&start, &id, &ts, 0);
+	return 0;
+}
+
+SEC("tracepoint/raw_syscalls/sys_exit")
```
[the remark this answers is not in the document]
aspsk: Am I right that in the case of raw tracepoints we are getting an array of `u64`s pointing to arguments of `TP_PROTO`? In this case prototype for `sys_exit` tracepoint is `TP_PROTO(struct pt_regs *regs, long ret)` and doesn't include the `id` (syscall number). In kernel this is easily obtained as `syscall_get_nr(current, regs)`, but I am not sure how to do this in the BPF case (as it is arch-dependent).

libbpf-tools/syscount.c
aspsk: Well, for example, here https://github.com/iovisor/bcc/pull/2920/files/3c40184699520362fece1d9ae3fe23d6182b9520#diff-68934a0eef57fae11fc290164fd99838R288 this is not the case. When I move the errno/syscall parsing to a separate file, this, probably, will. However, I also wanted to distinguish cases when errno is set vs. no symbols parsed. I will take a better look in the v2.

libbpf-tools/syscount.c
aspsk: It is needed, unluckily, because the list I am compiling in is valid only for `x86_64`, not for other architectures. Funny that there's no reliable way to convert errno/syscalls <-> numbers, this should be provided by some library, probably.

libbpf-tools/syscount.c
aspsk: Does this look better?
```
                col = agg_col(vals[i].key, buf, sizeof(buf));
                printf("%-22s %8llu %16.3lf\n", col, vals[i].count,
                                vals[i].total_ns / div);
```
and
```
                col = agg_col(vals[i].key, buf, sizeof(buf));
                printf("%-22s %8llu\n", col, vals[i].count);
```

libbpf-tools/syscount.c
aspsk: Right!

aspsk:
Thanks for your comments @anakryiko . I will try to address all in the next version. (Except that I do not understand how to use a raw tracepoint in a portable way, as I described above.)

libbpf-tools/syscount.bpf.c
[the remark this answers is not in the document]
anakryiko: Yeah, it's very strange, but sys_exit does not have syscall id provided... You are right that it seems to be arch-dependent. You can get syscall id using pt_regs->orig_ax, but that would break on non-x86. I guess just stick to tracepoint then.

libbpf-tools/syscount.c
anakryiko: link doesn't work, but sure, do whatever makes sense, I guess

libbpf-tools/syscount.c
anakryiko: Sure, I don't object strongly, it just feels wrong to do popen from root-privileged tools for something so basic. Can't we generate similar lists for other architectures using the same approach?

libbpf-tools/syscount.c
anakryiko: it's not about extracting agg_col call for me, it was about aligning each parameter with start of format string. It might be tabs vs spaces or you are using different indentation for 2nd+ arguments, don't know. But too lazy to pull your PR locally and check for myself :)

anakryiko:
> Thanks for your comments @anakryiko . I will try to address all in the next version. (Except that I do not understand how to use a raw tracepoint in a portable way, as I described above.)

Yeah, sys_exit not including syscall number is a downer, I wonder if it would make sense to fix it in kernel...

aspsk:
Hi @anakryiko , I've updated the PR, could you take another look?

One corresponding question: some `libbpf` functions, like `bpf_program__attach` return pointers, and, in case of error, they return kernel-style `ERR_PTR` pointers. However, the `linux/err.h` is not exported by `uapi`, so to use `libbpf` reliably one needs to redefine some of these macros, which is annoying to do for each new app. So, as we can't change API of the functions returning pointers, can't `libbpf` define these macros in `libbpf.h` instead of including `linux/err.h`?

anakryiko:
> Hi @anakryiko , I've updated the PR, could you take another look?
> 

sure

> One corresponding question: some `libbpf` functions, like `bpf_program__attach` return pointers, and, in case of error, they return kernel-style `ERR_PTR` pointers. However, the `linux/err.h` is not exported by `uapi`, so to use `libbpf` reliably one needs to redefine some of these macros, which is annoying to do for each new app. So, as we can't change API of the functions returning pointers, can't `libbpf` define these macros in `libbpf.h` instead of including `linux/err.h`?

libbpf defines libbpf_get_error() function, please use it.

aspsk:
> libbpf defines libbpf_get_error() function, please use it.

Ah, sorry, didn't see it. Updated the PR to use it.

libbpf-tools/syscall_helpers.c
```diff
@@ -0,0 +1,526 @@
...
+		}
+		syscall_names_size = MAX(number+1, syscall_names_size);
+	}
+
+	if (ferror(f))
+		warn("fgets: %s\n", strerror(errno));
+close:
+	err = pclose(f);
+	if (err < 0)
+		warn("pcolse: %s\n", strerror(errno));
```
anakryiko, line 106: typo: pclose

libbpf-tools/syscount.c
```diff
@@ -0,0 +1,466 @@
...
+		err = libbpf_get_error(*link);
+		if (err) {
+			warn("failed to auto-attach program '%s'\n",
+			     bpf_program__name(prog));
+			return err;
+		}
+	}
+
+	return 0;
+}
```
anakryiko, line 364: I don't get why you need this generic function. You have exactly 2 programs, and in some cases you want to attach only one of them. Here's the simplest and most straightforward way to do this:
```
obj->links.sys_exit = bpf_program__attach(obj->progs.sys_exit);
err = libbpf_get_error(obj->links.sys_exit);
if (err) /* ... */
if (env.latency) {
    obj->links.sys_enter = bpf_program__attach(obj->progs.sys_enter);
    err = libbpf_get_error(obj->links.sys_exit);
    if (err) /* ... */
}
```

Instead you are using obj->skeleton and use stuff that's meant for libbpf itself only. For no good reason, actually.

BTW, I'd call sys_enter/sys_exit programs something more short (I used just `sys_enter`/`sys_exit`) in the example above, so that it's easier to use from user-space.

libbpf-tools/syscall_helpers.c
aspsk: <s>I fixed some typos in other tools in this PR, so I can add one!</s>

Ah, fixed. Thanks for reviewing @anakryiko!

libbpf-tools/syscount.c
anakryiko: ping? this approach doesn't work for short-lived processes at all. You can have a separate array for this or just extend start and store process name first time. Is there anything wrong with that?

libbpf-tools/syscount.c
aspsk: > Instead you are using obj->skeleton and use stuff that's meant for libbpf itself only. For no good reason, actually.

You're right, I will fix this.

BTW, I've found a bug which I didn't see before (probably my system wasn't loaded enough, and also I've used batch version of lookup_and_delete initially before you rejected the same thing in vfsstat). There is a race in the `read_vals()`: the loop is not atomic and consists of multiple `bpf` system calls. Thus, the `bpf` entry in the map may be re-populated until it is the last entry in the map, and `bpf` entry could appear multiple times.

In the original tool, it does the printing as follows:
```
items = data.items() // read all values from the map one-by-one
keys = data.keys() // get all map keys
for k in keys: delete(k) // delete one element from the map
```
this also has some problems, because it looses system calls which are listed in keys and happened between getting items and deleting elements by keys.

The right solution which doesn't seem to loose any syscalls would be to do
```
keys = get_keys()
items = get_and_delete_items(keys)
```
but the `bpf_map_lookup_and_delete_elem` is supported only for stacks/queues.

Should I just use the approach used in syscount.py and loose some (relatively few) syscalls? Or maybe the right way is to patch `bpf_map_lookup_and_delete_elem` to support hashes? (Or there is some reason why it wasn't patched when batch lookup_and_delete was added?)

aspsk:
@anakryiko For some reason I can't answer in the thread: will fix the `"/proc/%d/comm"` issue, just forgot about it completely.

libbpf-tools/syscount.h
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2020 Anton Protopopov
+#ifndef __SYSCOUNT_H
+#define __SYSCOUNT_H
+
+struct data_t {
+	__u64 count;
+	__u64 total_ns;
+	__u32 key;
```
anakryiko, line 9: btw, doesn't seem like you are using key field (and you don't need it, right?) 50% memory reduction right there ;)

libbpf-tools/syscount.c
anakryiko: Don't know about lookup_and_delete for hashmap, honestly.

I'd just use two hashmaps and switch between them for each reporting period. From user-space, before you start reporting, increment an index ("current map index"), re-routing BPF logic to a different map. Then work with the one that's not modified, read it, clear it, then on next reporting period bump counter again. On BPF side, before updating data map, you just:
```
int cur_map_idx = 0; /* this gets updated from user-space */

SEC(...)
int prog(...) {
    void *data = (cur_map_idx & 1) ? data1 : data2;
    /* proceed as before */
}
```

There is still tiny race (after you bumped cur_map_idx, there could be still BPF programs that are still updating previous map), but I think they are so tiny that it doesn't matter in practice.

BTW, to mitigate some of the memory overhead, I think it should be fine to use BPF_F_NO_PREALLOC hashmap (just specify it as map_flags). It will make kernel allocate memory for hash map entries on demand. So even with big hashmap, if it doesn't have lots of entries, you'd still have a manageable memory usage.

Makes sense?

libbpf-tools/syscount.h
aspsk: I am using it in user space: I read items into an array of `struct data_t` and sort it by `count` or `total_ns`, but the `key` part should also be sorted to print it. This came from the "batch" version — this way I don't need any extra copy in user space. But now, comparing to syscall overhead this doesn't count, so I can remove this item and use a different data type in user space to store items.

libbpf-tools/syscount.c
aspsk: So, if we still have a race, then I, probably, would prefer to read data the same way the original `syscount.py` — in this way I can re-use functions to read keys and hash items later. But I will try both variants first and see what looks better.

aspsk: > BTW, to mitigate some of the memory overhead, I think it should be fine to use BPF_F_NO_PREALLOC hashmap (just specify it as map_flags). It will make kernel allocate memory for hash map entries on demand. So even with big hashmap, if it doesn't have lots of entries, you'd still have a manageable memory usage.

Thanks, will add this.

aspsk:
Hi @anakryiko I think that I've addressed all your comments. I decided to read values as it is done in the original tool (read keys, read values by keys, delete values by keys), because it will be straightforward to fix this variant when `bpf_map_lookup_and_delete_batch` will be accessible and to correspond to the `syscount.py` beaviour. For command names I've added a LRU hash which I allocate from user space (`bpf_map__resize(obj->maps.comm, MAX_ENTRIES);`) only when needed. Also added some comments where I think it is not obvious why things are happening as they are.

libbpf-tools/syscount.c
```diff
@@ -0,0 +1,455 @@
...
+
+	if (localtime_r(&now, &tm))
+		printf("[%02d:%02d:%02d]\n", tm.tm_hour, tm.tm_min, tm.tm_sec);
+	else
+		warn("localtime_r: %s", strerror(errno));
+}
+
+static bool read_vals(int fd, struct data_ext_t *vals, __u32 *count)
+{
+	__u32 keys[MAX_ENTRIES] = { -1 };
```
anakryiko, line 214: that's quite a big on-the-stack array which is initialized every single time, make it static maybe?

libbpf-tools/syscount.c
```diff
@@ -0,0 +1,455 @@
...
+		}
+		vals[j].count = val.count;
+		vals[j].total_ns = val.total_ns;
+		vals[j].key = keys[j];
+	}
+
+	/* There is a race here: system calls which are represented by keys
+	 * above and happened between lookup and delete will be ignored.  This
+	 * will be fixed in future by using bpf_map_lookup_and_delete_batch,
+	 * but this function is too fresh to use it in bcc. */
```
anakryiko, line 249: this is not bcc, really :) you can use it, you'll be limiting yourself to very recent kernels, though

libbpf-tools/syscount.c
aspsk: Ok, thanks, there is actually no need to init it at all, removed.

libbpf-tools/syscount.bpf.c
```diff
@@ -0,0 +1,133 @@
...
+{
+	char buf[COMM_NAME_SIZE];
+	int err;
+
+	/* We should save the process name every time because a process can
+	 * change its comm. For example, when bash executes a binary program,
+	 * then first syscalls after fork, but before exec, run with the "bash"
+	 * comm, so if we cache this value here, then all processes created by
+	 * bash will be named "bash" */
+	err = bpf_get_current_comm(buf, sizeof(buf));
```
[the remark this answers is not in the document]
aspsk: Yeah, thanks, I will read process name instead (which makes me think about adding a `-t` option to count/display per-thread statistics).

So, for rewriting, I need to do this, because a process/thread can change its name, say, using `prctl` or after `exec`. (Probably it could be done in sync, like, when we obtain a process name once and then watching for syscalls which can change it, but this sounds like a hard thing to implement on practice for a debug tool?)

libbpf-tools/syscount.h
```diff
@@ -0,0 +1,15 @@
...
+#define __SYSCOUNT_H
+
+struct data_t {
+	__u64 count;
+	__u64 total_ns;
+};
+
+#define MAX_ENTRIES 8192
+
+#define COMM_NAME_SIZE 32
```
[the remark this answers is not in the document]
aspsk: Yes, thanks. I was confused by `/proc/*/comm` which can return more bytes (for kernel tasks)

libbpf-tools/syscount.c
aspsk: Ok, what if I do this based on `uname(2)` information (or there is a better way to get linux version)? I am curious if I can actually do race-less code using batch lookup and delete (or I need a `bpf_map_lookup_and_delete_elem` for hashes)

libbpf-tools/syscount.bpf.c
```diff
@@ -0,0 +1,133 @@
...
+	__type(key, u32);
+	__type(value, struct data_t);
+	__uint(map_flags, BPF_F_NO_PREALLOC);
+} data SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_LRU_HASH);
+	__type(key, u32);
+	__type(value, __u8[COMM_NAME_SIZE]);
+} comm SEC(".maps");
```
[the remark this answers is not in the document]
aspsk: Ok, right

aspsk:
@anakryiko I've fixed all "blockers", didn't touch the race with getting values and the optimization with reading process names, planning to to this later (hopefully, together with adding per-thread view)

aspsk:
> @anakryiko I've fixed all "blockers", didn't touch the race with getting values and the optimization with reading process names, planning to to this later (hopefully, together with adding per-thread view)

Upd.: implemented reading values in a raceless way using batch functions, if available

libbpf-tools/syscount.c
```diff
@@ -0,0 +1,482 @@
...
+	for (int j = 0; j < i; j++) {
+		err = bpf_map_lookup_elem(fd, &keys[j], &val);
+		if (err && errno != ENOENT) {
+			warn("failed to lookup element: %s\n", strerror(errno));
+			return false;
+		}
+		vals[j].count = val.count;
+		vals[j].total_ns = val.total_ns;
+		vals[j].key = keys[j];
+		strncpy(vals[j].comm, val.comm, TASK_COMM_LEN);
```
anakryiko, line 272: newer GCCs will complain about potentially non-zero-terminated strings. I'd just do straight memcpy of TASK_COMM_LEN bytes (which we know are zero-terminated).

aspsk: ok, done

aspsk:
> LGTM. Let's land this beast.
> 
> This will probably conflict with cpudist, so one of you guys will need to rebase one more time, sorry about that!

Np, I will rebase mine once `cpudist` is merged (we will have a conflict in `Makefile`).

aspsk:
> > LGTM. Let's land this beast.
> > This will probably conflict with cpudist, so one of you guys will need to rebase one more time, sorry about that!
> 
> Np, I will rebase mine once `cpudist` is merged (we will have a conflict in `Makefile`).

Taking my words back @anakryiko! The `cpudist` already has conflicts, so this would be wise to merge my PR first :-)

yonghong-song:
[buildbot, test this please]

PR #2923, opened 2020-05-21
libbcc-py: Fix libbpf types

mauriciovasquezbernal:
1ad2656a1d9c ("Add support_kfunc function to BPF object") added new functions
to libbcc-py but didn't set the restype and argstypes for those. It's causing
the `bpf_has_kernel_btf` function to return True in systems without support
for BTF, making tools like opensnoop, klockstat and any other using kfuncs
unusable in those systems.

The following Python script reproduces the problem:

```
from bcc import BPF
print(BPF.support_kfunc())
```

Signed-off-by: Mauricio Vásquez <mauricio@kinvolk.io>


yonghong-song:
[buildbot, test this please]

src/python/bcc/libbcc.py
```diff
@@ -108,6 +108,12 @@
 lib.bpf_detach_tracepoint.argtypes = [ct.c_char_p, ct.c_char_p]
 lib.bpf_attach_raw_tracepoint.restype = ct.c_int
 lib.bpf_attach_raw_tracepoint.argtypes = [ct.c_int, ct.c_char_p]
+lib.bpf_detach_kfunc.restype = ct.c_int
+lib.bpf_detach_kfunc.argtypes = [ct.c_int, ct.c_char_p]
```
yonghong-song, line 112: Looks good. Could you remove `bpf_detach_kfunc`? In the python side, it is not needed as python just close the fd, no need to go to C++.

yonghong-song:
[buildbot, ok to test]

src/python/bcc/libbcc.py
mauriciovasquezbernal: Done!

PR #2924, opened 2020-05-21
Fix KFUNC_PROBE return value

mauriciovasquezbernal:
The KFUNC_PROBE macro is using "void" as return type, this is causing problems
in some tools that have a filtering enable that returns 0.

Reproducer: (Notice that it requires BTF support)

```
$ python opensnoop.py --pid 5
/virtual/main.c:33:21: error: void function '____kretfunc__do_sys_open' should not return a value [-Wreturn-type]
    if (pid != 5) { return 0; }
                    ^      ~
1 error generated.
...
```
Signed-off-by: Mauricio Vásquez <mauricio@kinvolk.io>

FIxes changes made in https://github.com/iovisor/bcc/pull/2726.

### DIFF
--- a/src/cc/export/helpers.h
+++ b/src/cc/export/helpers.h
@@ -1012,7 +1012,7 @@ int raw_tracepoint__##event(struct bpf_raw_tracepoint_args *ctx)
 #define BPF_PROG(name, args...)                                 \
 int name(unsigned long long *ctx);                              \
 __attribute__((always_inline))                                  \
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, ok to test]

yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks. LGTM.

PR #2927, opened 2020-05-22
tool: trace process termination by default

syohex:
`sched_process_exit` tracepoint is called when thread terminates.
So exitsnoop shows line per each thread termination if the process
is multi-thread process. This is not useful when people wants to
know why process terminates, not thread.

So this changes exitsnoop default behavior which traces process termination
instead of thread termination. And add `--per-thread` option which behaves
as original exitsnoop implementation.

### Motivation

I suppose `exitsnoop` should check both PID and TID as default behavior because its documentations says **Trace all process termination** but current implementation **traces all thread termination**. However I'm not sure that there is no problem to change default behavior for backward compatibility.

### DIFF
--- a/man/man8/exitsnoop.8
+++ b/man/man8/exitsnoop.8
@@ -2,7 +2,7 @@
 .SH NAME
 exitsnoop \- Trace all process termination (exit, fatal signal). Uses Linux eBPF/bcc.
 .SH SYNOPSIS
-.B exitsnoop [\-h] [\-t] [\-\-utc] [\-x] [\-p PID] [\-\-label LABEL]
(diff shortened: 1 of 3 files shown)


gfx:
+1

Rather I prefer this is the default.

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

yonghong-song:
@brendangregg what do you think? per process as the default?

brendangregg:
Yes, I'd make this the default, and add a --per-thread option instead that provided the old behavior.

syohex:
I got it. I'll update.

syohex:
I have updated PR and PR comments.

yonghong-song:
LGTM. Thanks!

PR #2929, opened 2020-05-23
Add killsnoop arg for filtering on signal

synical:
I was recently using killsnoop to troubleshoot some weird signal behaviour and found that I had to run killsnoop over long periods of time in order to catch the behaviour. A problem I ran into was that filtering for signals I was looking for vs all other signals getting sent around in the system required either logging everything to a file and parsing from there, or running a forked version of killsnoop to filter on the signal.

This PR includes a change to allow optional filtering of the signal that is passed to kill() to help reduce noise in cases like this.

I haven't updated the examples as I'm not sure if this will be considered useful, but happy to do so if this gets the green light.

### DIFF
--- a/man/man8/killsnoop.8
+++ b/man/man8/killsnoop.8
@@ -28,6 +28,9 @@ Only print failed kill() syscalls.
 .TP
 \-p PID
 Trace this process ID only (filtered in-kernel).
+.TP
+\-s SIGNAL
+Trace this signal only (filtered in-kernel).
 .SH EXAMPLES
 .TP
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

yonghong-song:
Yes, this seems a useful option. Could you change man page and example file as well?

synical:
Thanks @yonghong-song, I've updated the example file and man page as requested.

Issue #1754, opened 2018-05-11
ISRA breaks kprobes


bobrik:
GCC can do optimizations and rename functions in the kernel (see [stackoverflow question](https://stackoverflow.com/a/18914402)).

* amd64:

```
$ cat /proc/kallsyms | fgrep do_wakeup
ffffffffa2483420 t ttwu_do_wakeup
```

* arm64:

```
$ cat /proc/kallsyms | fgrep do_wakeup
ffff8692fe2a6ff8 t ttwu_do_wakeup.isra.20
```

We have tools that directly attach to `ttwu_do_wakeup` kprobe, and ISRA breaks those.

What's the suggested way of working around this?

yonghong-song:
I compiled for x86 with `-fipa-sra` and I did not get `.isra.#` symbols.
```
diff --git a/Makefile b/Makefile
index 619a85a..df90968 100644
--- a/Makefile
+++ b/Makefile
@@ -362,7 +362,7 @@ HOST_LFS_LIBS := $(shell getconf LFS_LIBS)
 
 HOSTCC       = gcc
 HOSTCXX      = g++
-HOSTCFLAGS   := -Wall -Wmissing-prototypes -Wstrict-prototypes -O2 \
+HOSTCFLAGS   := -Wall -Wmissing-prototypes -Wstrict-prototypes -O2 -fipa-sra \
                -fomit-frame-pointer -std=gnu89 $(HOST_LFS_CFLAGS)
 HOSTCXXFLAGS := -O2 $(HOST_LFS_CFLAGS)
 HOSTLDFLAGS  := $(HOST_LFS_LDFLAGS)
```

What is you compilation flags form arm64?

bobrik:
I have plenty of `isra` symbols on both arm64 and amd64:

```
$ uname -m; cat /proc/kallsyms | fgrep isra | wc -l
x86_64
785
```

```
$ cat /proc/kallsyms | fgrep isra | wc -l
933
```

We don't set any special compilation flags during build.

This is happening on both Debian Jessie and Stretch, 4.9 and 4.14 kernels. On Docker for mac I see the same picture. I don't have anything else at hand to check.

brendangregg:
I have about 950 isra symbols too (both stock Ubuntu kernel 4.4, and my own built kernels on Ubuntu).

[...]

PR #2930, opened 2020-05-23
Implement ISRA detection in the C++ API

eklitzke:
This change attempts to handles kprobe whose names are mangled by GCC ISRA as described in #1754 

With this change if bcc fails to load a kprobe, it builds a sorted vector of all the .isra. symbols from `/proc/kallsyms`. If a .isra version of a symbol can be found in this vector, that symbol name is used instead. On my Linux desktop there are 810 such symbols (total string size 22 KB). I've written a new test case, and it should be fine if `/proc/kallsyms` isn't available. I didn't implement this in the Python API since I only use the C++ API, but it's probably easy to implement the exact same technique in Python (the Python `bisect` module implements the same functionality as `std::lower_bound`).

As mentioned [here](https://github.com/iovisor/bcc/issues/1754#issuecomment-471868602) GCC might change the function signature of functions optimized this way, so it might  be problematic to do things like inspect function arguments in this case.

One other question is whether this should be opt-in instead. Right now I'm just doing it automatically, but I could see if some people find this to be confusing.

### DIFF
--- a/src/cc/api/BPF.cc
+++ b/src/cc/api/BPF.cc
@@ -20,6 +20,7 @@
 #include <cstdio>
 #include <cstring>
 #include <exception>
+#include <fstream>
 #include <iostream>
 #include <memory>
 #include <sstream>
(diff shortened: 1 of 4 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

yonghong-song:
For python side, we can delay this.
I think we should make this feature as an opt-in feature initially. Otherwise, as you mentioned, user may be be surprised in some times.

PR #2932, opened 2020-05-25
fix: avoid -Wsign-compare warnings

gfx:
There are warnings in lexer.h:

> .../iovisor/bcc/src/cc/frontends/b/lexer.h:54:35: warning: comparison of integer expressions of different signedness: ‘size_t’ {aka ‘long unsigned int’} and ‘const counter_type’ {aka ‘const int’} [-Wsign-compare]
   54 |     for (size_t i = begin.line; i <= end.line; ++i)


yonghong-song:
[buildbot, test this please]

PR #2939, opened 2020-05-28
tools: Add funcinterval bcc tool.

netedwardwu:
Time interval between the same function as a histogram.
Referenced from funclatency.

### DIFF
--- a/tools/funcinterval.py
+++ b/tools/funcinterval.py
@@ -1,0 +1,180 @@
+#!/usr/bin/python
+# @lint-avoid-python-3-compatibility-imports
+#
+# funcinterval   Time interval between the same function, tracepoint
+#                as a histogram.
+#
+# USAGE: funcinterval [-h] [-p PID] [-i INTERVAL] [-T] [-u] [-m] [-v] pattern
+#
+# Run "funcinterval -h" for full usage.
+#
+# Copyright (c) 2020 Realtek, Inc.
+# Licensed under the Apache License, Version 2.0 (the "License")
+#
+# 03-Jun-2020   Edward Wu   Referenced funclatency and created this.
+
+from __future__ import print_function
+from bcc import BPF
+from time import sleep, strftime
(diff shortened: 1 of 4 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

tools/funcinterval.py
```diff
@@ -0,0 +1,161 @@
...
+# signal handler
+def signal_ignore(signal, frame):
+    print()
+
+
+# load BPF program
+b = BPF(text=bpf_text)
+
+# attach probes
+b.attach_kprobe(event_re=func_name, fn_name="trace_func_entry")
```
yonghong-song, line 125: Here, we are matching as regex. Maybe you have an option to control this? Sometimes people really just want to track a particular function.

yonghong-song:
cc @brendangregg

brendangregg:
Well this is easy: I already know it'll be a useful tool as I'm using something similar (bpftrace one-liner) to check arrival distributions. Plus the _example.txt is good. I'm happy with this, thanks.

Like funccount(8), we'll want to do tracepoints as well. That functionality can be added in a later PR.

netedwardwu:
Hi @yonghong-song ,

As you noticed, I remove some funclatency features on funcinterval. Because of those features I never used on funclatency, even I solved many issues by funclatency.
I used funclatency only when I understand the purpose of the specific function name.

I prefer leaving those features to other developers to contribute.
BUT, if you insist on it. For sure that I can do it.
Thanks for your code review.

Hi @brendangregg ,

Thanks for your comment. Yes, the tracepoint feature is what I wanted to do.
I will do another PR for it.

netedwardwu:
Hi all,

I have added tracepoint feature, please help to check it.

Thanks

yonghong-song:
Looks good. Thanks!

PR #2941, opened 2020-05-29
tools/biolatency: add --pid filtering to biolatency

andreimatei:


yonghong-song:
Could you describe in which cases pid filter will be useful? If I understand correctly, the function `blk_mq_start_request` and `blk_account_io_done` are not called in process context. They mostly called in interrupt context, in which cases, pid filtering is not very accurate. But it will be good to hear the details of your use case.

andreimatei:
Hi @yonghong-song! My goal is to monitor the IO latency (and separately size) of requests coming from a particular process - namely the CockroachDB database. My ultimate goal is to embed the tool into the database itself and to display IO information in a dashboard. For that I'm currently reading about libbpf and friends.

For this biolatency tool in isolation it seemed to me that it'd be a common ask to monitor a single process - if technically feasible. Doing the filtering as I've done it "seemed to work" in a quick test - most of the IO on my box seemed to be attributed to the process I was expecting. I haven't checked if the start of the IO in my test was tracked through `blk_mq_start_request` or through `blk_start_request` - you think there could be a difference between them? On the termination of the IO, notice that I have not added any new filtering (so I guess it doesn't matter what the pid is in `blk_account_io_done`).
Anything you happen to know about `blk_start_request` will be greatly appreciated. And generally any thoughts on the topic - like maybe if I should be looking to instrument the filesystem layer rather than the block layer. Any pointers to resources would also be much appreciated!
Thanks!

mcaleavya:
It’s been a while however are we not filtering into the request task
pointer? So each should be unique across functions, so at least there is
commonality, I can’t remember if the tid/pid is also available

yonghong-song:
The following is one of stacks I observed for `blk_mq_start_request`:
```
[  372.276682] Call Trace:
[  372.276988]  dump_stack+0x68/0x9b
[  372.277386]  blk_mq_start_request+0x14/0x190
[  372.277904]  scsi_queue_rq+0x3e8/0xa90
[  372.278357]  blk_mq_dispatch_rq_list+0x90/0x570
[  372.278892]  ? _raw_spin_lock_irqsave_nested+0x81/0x90
[  372.279496]  blk_mq_do_dispatch_sched+0x78/0x100
[  372.280040]  blk_mq_sched_dispatch_requests+0xf7/0x170
[  372.280669]  __blk_mq_run_hw_queue+0x92/0x110
[  372.281188]  process_one_work+0x201/0x580
[  372.281668]  ? process_one_work+0x17a/0x580
[  372.282176]  worker_thread+0x4c/0x3e0
[  372.282613]  kthread+0x114/0x150
[  372.282992]  ? process_one_work+0x580/0x580
[  372.283495]  ? kthread_mod_delayed_work+0xc0/0xc0
[  372.284052]  ret_from_fork+0x3a/0x50
```
The callback function for `blk_account_io_done`:
```
[  372.257694]  <IRQ>
[  372.257948]  dump_stack+0x68/0x9b
[  372.258344]  blk_account_io_done+0x1a/0x380
[  372.258843]  ? recalibrate_cpu_khz+0x10/0x10
[  372.259378]  __blk_mq_end_request+0x6b/0x100
[  372.259913]  scsi_end_request+0x10e/0x320
[  372.260394]  ? scsi_end_request+0x6b/0x320
[  372.260901]  scsi_io_completion+0x6d/0x480
[  372.261389]  blk_done_softirq+0x8d/0xc0
[  372.261870]  __do_softirq+0xd0/0x428
[  372.262286]  irq_exit+0xa5/0xb0
[  372.262677]  call_function_single_interrupt+0xf/0x20
[  372.263276]  </IRQ>
[  372.263542] RIP: 0010:default_idle+0x28/0x170
```
It is not in the process context. The blk operation is in a kthread or softirq context so the pid at that point may be totally something else. The filtering with pid is not guaranteed to work in some cases.

andreimatei:
Thanks @yonghong-song.
Do you think it's worth starting to track the request pointer before the IO is enqueued, while there's (presumably) still a pid available? So I'm thinking we'd maintain a map of queued request pointers, and then the `blk_mq_start_request` kprobe would only start measuring durations for requests in that map. (I think maybe that's what @mcaleavya was suggesting too?). I'm not familiar with the kernel code, but I give it a try. If it works out, would you then welcome the patch?

yonghong-song:
@andreimatei Please go ahead to give a try. If you indeed find a solution, I think it will be a good addition to the tool since it will enable to trace block io operations initialized by a particular process.

PR #2942, opened 2020-05-29
Make reading blacklist from debugfs optional

bobrik:
With lockdown enabled one sees the following:

```
$ sudo /usr/share/bcc/tools/funccount -Ti 1 run_timer_softirq
[Errno 1] Operation not permitted: '/sys/kernel/debug/tracing/../kprobes/blacklist'
```

Which is accompanied by the following in `dmesg`:

```
[Fri May 29 22:12:47 2020] Lockdown: funccount: debugfs access is restricted; see man kernel_lockdown.7
```

Since blacklist is not a required feature, let's make
reading from it optional, so that bcc can work out of the box.

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -546,8 +546,15 @@ class BPF(object):
 
     @staticmethod
     def get_kprobe_functions(event_re):
-        with open("%s/../kprobes/blacklist" % TRACEFS, "rb") as blacklist_f:
-            blacklist = set([line.rstrip().split()[1] for line in blacklist_f])
+        blacklist_file = "%s/../kprobes/blacklist" % TRACEFS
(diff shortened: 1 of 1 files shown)


yonghong-song:
Does this related to LOCKDOWN_TRACEFS?

Anyway, looks good to me although I did not have an environment to reproduce it.

bobrik:
We run in integrity mode, so I think it's more of `LOCKDOWN_DEBUGFS`.

yonghong-song:
> We run in integrity mode, so I think it's more of `LOCKDOWN_DEBUGFS`.

Oh, yes, I searched at kernel/trace directory and did not find `LOCKDOWN_DEBUGFS`, but did find it under `fs/debugfs`.

PR #2951, opened 2020-06-04
Add support for multiple PID/TID for offwaketime

lorddoskias:
Here's multi PID/TID support for offwaketime tool.

### DIFF
--- a/tools/offwaketime.py
+++ b/tools/offwaketime.py
@@ -20,6 +20,16 @@ from sys import stderr
 
 # arg validation
 def positive_int(val):
+    dest = []
+    # Filter up to 5 pids, arbitrary
+    args_list = val.split(",", 5)
+    pids_to_add = min(len(args_list), 5)
+    for i in range(pids_to_add):
+        dest.append(_positive_int(args_list[i]))
+
+    return dest
+
+def _positive_int(val):
     try:
         ival = int(val)
     except ValueError:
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, ok to test]

tools/offwaketime.py
```diff
@@ -20,6 +20,16 @@
 
 # arg validation
 def positive_int(val):
+    dest = []
```
yonghong-song, line 23: The change of this function from returning a int to a list caused the following error
```
  File "../../tools/offwaketime.py", line 125, in <module>
30:     duration = int(args.duration)
30: TypeError: int() argument must be a string or a number, not 'list'
```

lorddoskias: Yes the fix is to use _positive_int for the type of duration. I did catch it and updated my branch. I see that the commit has this updated.

yonghong-song:
[buildbot, test this please]

tools/offwaketime.py
yonghong-song: > Yes the fix is to use _positive_int for the type of duration. I did catch it and updated my branch. I see that the commit has this updated.

The error still there. Could you double check?

lorddoskias: Ops, I had missed the usage of positive_int in positive_nonzero_int. Pushed fixed branch now. Restart the tests.

yonghong-song:
LGTM. Thanks!

PR #2952, opened 2020-06-04
Fix runqslower to indicate that the latency param is in microseconds.

adityamandaleeka:
This is a minor fix... I was using `runqslower` when I noticed that the help text indicated that the latency parameter was in milliseconds, when it's actually in microseconds. This change fixes that, and I also fixed a typo in that line while I was there.


brendangregg:
[buildbot, ok to test]

PR #2953, opened 2020-06-05
usdt: Add helpers to set semaphore values

danobi:
While debugging a high memory consumption issue in bpftrace, I noticed
that a USDT::Context object can take ~10M per instance [0]. Along with
the new --usdt-file-activation feature in bpftrace
( https://github.com/iovisor/bpftrace/pull/1317 ), bpftrace can
potentially hold onto many dozens of USDT:Context instances, causing
memory issues.

While reducing the amount of memory USDT::Context uses is one option,
we can potentially side step it by allowing the usdt semaphore count to
be set independently. Before, the only way to increment the count (by 1)
is to call bcc_usdt_enable*(). bcc_usdt_enable*() has checks that limit
it to a single increment per context. The only way to decrement the
count is by calling bcc_usdt_close() which naturally only allows for
one decrement.

With independent semaphore helpers, we can avoid holding onto a
USDT::Context instance for the lifetime of the tracing session. We can
simply:

1. create a USDT::Context
2. increment the semaphore count for the probe we care about
3. destroy the USDT::Context
4. repeat 1-3 for all probes we want to attach to
5. do our tracing
6. create a USDT::Context for the probe we care about
7. decrement the semaphore count
8. destroy the USDT::Context
9. repeat 6-8 for all the probes we're attached to

This approach also has the benefit of 1 USDT::Context instance being
alive at a time which can help keep memory high watermark low.

[0]: Through gdb single stepping and /proc/pid/status. Exact process is
not described here b/c memory usage probably varies based on tracee
binary.

### DIFF
--- a/src/cc/bcc_usdt.h
+++ b/src/cc/bcc_usdt.h
@@ -70,9 +70,12 @@ int bcc_usdt_get_argument(void *usdt, const char *provider_name,
                           struct bcc_usdt_argument *argument);
 
 int bcc_usdt_enable_probe(void *, const char *, const char *);
+int bcc_usdt_addsem_probe(void *, const char *, const char *, int16_t);
 #define BCC_USDT_HAS_FULLY_SPECIFIED_PROBE
 int bcc_usdt_enable_fully_specified_probe(void *, const char *, const char *,
                                           const char *);
+int bcc_usdt_addsem_fully_specified_probe(void *, const char *, const char *,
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

yonghong-song:
LGTM. Thanks!

PR #2954, opened 2020-06-05
usdt: Have Context::addsem_probe() nop if pid not specified

danobi:
This makes bcc_usdt_addsem*() more consistent with the bcc_usdt_enable*()
interface where if a USDT::Context was not constructed with a pid the
semaphore enablement nops.


yonghong-song:
[buildbot, test this please]

PR #2957, opened 2020-06-08
libbpf-tools: refactor, move public methods to trace_helpers.c

ethercflow:
- extract `bump_memlock_rlimit` and put into `trace_helpers.c`
- extract `get_ktime_ns` and put into `trace_helpers.c`
- Resort struct's fields by alignment size instead of field size

### DIFF
--- a/libbpf-tools/trace_helpers.c
+++ b/libbpf-tools/trace_helpers.c
@@ -3,7 +3,15 @@
 #include <stdlib.h>
 #include <string.h>
 #include <stdbool.h>
+#include <sys/resource.h>
+#include <time.h>
 #include "trace_helpers.h"
+
+#define min(x, y) ({				 \
+	typeof(x) _min1 = (x);			 \
+	typeof(y) _min2 = (y);			 \
+	(void) (&_min1 == &_min2);		 \
+	_min1 < _min2 ? _min1 : _min2; })
 
 struct ksyms {
 	struct ksym *syms;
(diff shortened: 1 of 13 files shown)


ethercflow:
Hi @anakryiko I've refactor as you mentioned, PTAL when you're free. Thanks.

libbpf-tools/trace_helpers.c
```diff
@@ -212,3 +214,21 @@ void print_log2_hist(unsigned int *vals, int vals_size, char *val_type)
 		printf("|\n");
 	}
 }
+
+unsigned long long get_ktime_ns(void)
+{
+	struct timespec ts;
+
+	clock_gettime(CLOCK_MONOTONIC, &ts);
+	return ts.tv_sec * NSEC_PER_SEC + ts.tv_nsec;
```
anakryiko, line 223: can you add (unsigned long long) cast to ts.tv_sec to avoid overflow on 32-bit architectures?

anakryiko: ah, never mind, NSEC_PER_SEC is already unsigned long long

libbpf-tools/xfsslower.h
```diff
@@ -11,14 +11,14 @@
...
-	char file[DNAME_INLINE_LEN];
-	char task[TASK_COMM_LEN];
 	__u64 delta_us;
 	__u64 end_ns;
 	__s64 offset;
 	ssize_t size;
 	pid_t tgid;
 	char type;
+	char file[DNAME_INLINE_LEN];
+	char task[TASK_COMM_LEN];
```
anakryiko, line 21: this actually might be harmful, as it puts file and task at non-aligned offset. Given both file (32 bytes) and task (16 bytes) have 8-aligned sizes, it's ok to keep them at the bottom. Please revert this part. It's similar for other places with multiple-of-8-sized arrays, they are not going to harm alignment.

ethercflow: Thanks @anakryiko , I've revert them back to ensure that all fields of the structure at 8-aligned offset if they have 8-aligned sizes.

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

PR #2963, opened 2020-06-11
dbstat: fix overflowing timestamp

jeromemarchand:
The current default value of interval (99999999999) in dbstat is too
high to be used in the sleep() function in python 3. I couldn't find a
authoritative source on the issue, but it seems the max value is
2^63/10^9 (9223372036). Anyway, 99999999 is the de facto standard for
a very big number here, so just use that. It's over 3 years, that
should be enough.

For consistency, I also change a couple of value in klockstat even
though they didn't overflow.

It fixes the following error:
$ dbstat mysql
Tracing database queries for pids  slower than 0 ms...
  File "./dbstat", line 112, in <module>
    sleep(args.interval)
OverflowError: timestamp too large to convert to C _PyTime_t


yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks. The new number is still good enough.

PR #2965, opened 2020-06-12
Explicitly use NULL macro in dref check

djosephsen:
As a new-person going through the tutorial myself, I thought it might be a little clearer to use the NULL macro in checks like these.  This could be accompanied by a _thing to learn_ in lesson four to the effect of:  _The verifier requires that pointer values derived from a map lookup must be checked for a null value before they can be dereferenced and used_


yonghong-song:
[buildbot, test this please]

PR #2966, opened 2020-06-14
Add install steps for Alpine Linux

acj:
Recent changes in Alpine have made it possible to build bcc without ugly patches, so it seems like a good time to add the instructions here. I'm hopeful that the Python 3 caveats can be removed soon and that the tools will work out of the box on more systems. It should be fairly easy to get up and running, though.


yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks!

PR #2967, opened 2020-06-15
isotack tool proposal

alekseymmm:
Hi everyone, I am new to bcc. Recently I used it to create a tool like other bcc/tools, that helps me analyze the performance, load balancing, and datapath within our custom SW raid system. And I think of maybe merging my work into main bcc tree. Since it may be useful for anyone who analyzes io subsystems: raids, VM, individual devices. This tool is different from other like disksnoop, stackcount and bitesize (most similar ones) because it focuses on the block devices, rather than processes, store stack traces to analyze the hottest datapathes, count total io sizes to measure device endurance and io balancing. For me initially, it started as a tool to investigate RAID read-modify-write parity updates costs.

Besides this tool works on the most generic kernel block layer function `generic_make_request` not the blk_account_io_start or _done or similar so that it makes it possible to track io to any kind of block devices: bio- based, single queue req-based, mq.

Additionally, I implemented io size counting in thread-safe manner so we do not lose any io information with multithreaded workloads. I performed testing by fio and Linux mdraid, lvm xfs, and different types of block devices and queue types: null_blk with different queue_mode parameter, SSD, NVMe drives.
Also, flamegraphs are beautiful and informative for analysis with this tool.

Please let me know if it is not "on the level" to be in iovisor tools and what possible improvements should I do.

### DIFF
--- a/tools/iostack.py
+++ b/tools/iostack.py
@@ -151,8 +151,8 @@ for k, v in sorted(counts.items(), key=lambda counts: counts[1].value):
             print("    %s" % b.sym(addr, k.tgid_pid).decode('utf-8', 'replace'))
 
         if k.tgid_pid != 0xffffffff:
-            print("    %s [%d]" % (k.comm_name.decode('utf-8', 'replace'),
+            print("        %s [%d]" % (k.comm_name.decode('utf-8', 'replace'),
                                    k.tgid_pid))
         else:
-            print("    %s" % (k.comm_name.decode('utf-8', 'replace')))
+            print("        %s" % (k.comm_name.decode('utf-8', 'replace')))
         print("        %d\n" % v.value)
--- a/tools/iostack.py
+++ b/tools/iostack.py
@@ -32,7 +32,6 @@ BPF_STACK_TRACE(stack_traces, 16384);
 int _generic_make_request(struct pt_regs *ctx, struct bio *bio) {
     struct disk_data_t data = {};
     int dir = op_is_write((bio)->bi_opf & REQ_OP_MASK) ? WRITE : READ;
-    
     u32 bi_size = bio->bi_iter.bi_size;
     
     struct gendisk *bio_disk = bio->bi_disk;
@@ -125,7 +124,7 @@ duration = int(args.duration)
 
 # header
 if not args.folded:
-    print("Tracing total io sizes (bytes) to block devices", end="")
+    print("Tracing io (bytes) to block devices", end="")
     if duration < 99999999:
         print(" for %d secs." % duration)
     else:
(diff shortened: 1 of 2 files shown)


brendangregg:
It doesn't seem to be solving something far beyond the existing tools (e.g., using a combination of biotop and biostacks, etc).

It's default output is worse than biotop.

It's by-stack output will be more verbose than biostacks, as it breaks down by so many keys. This produces hundreds of pages of output. Now, if you're a storage engineer working an issue, you have time to dig through this stuff. But that's too specialized. I can image other people running it and wondering: what am I supposed to do with all this output?

For specialized cases like that, I'd just do on-the-fly bpftrace:

```
# bpftrace -e 't:block:block_rq_insert { @[args->dev, args->rwbs, kstack] = sum(args->bytes); }'
Attaching 1 probe...
^C

@[0, N, 
    __elv_add_request+358
    blk_execute_rq_nowait+203
    blk_execute_rq+80
    scsi_execute+227
    scsi_test_unit_ready+96
    sd_check_events+248
    disk_check_events+101
    disk_events_workfn+22
    process_one_work+478
    worker_thread+50
    kthread+289
    ret_from_fork+53
]: 0
@[271581184, WS, 
    blk_mq_insert_requests+203
    blk_mq_sched_insert_requests+111
    blk_mq_flush_plug_list+467
    blk_flush_plug_list+234
    blk_finish_plug+44
    dmcrypt_write+593
    kthread+289
    ret_from_fork+53
]: 32768
```

We're just missing a device number to device name helper; with that, we can produce the same output as iostack -K with a bpftrace one-liner.

Adding tools to bcc should be cases of "wow, yeah, we really need that!", and this does not pass that test. I thought this was explained in the CONTRIBUTING-SCRIPTS.md file -- perhaps we need to make that more clear?

People are welcome to create their own repos of bpf tools (as I do for specailized things I don't want in bcc). Maybe some tool over time will prove useful enough for addition (especially as it evolves to solve more problems).

alekseymmm:
@brendangregg  Fair points, totally agree.
You are right and with a real workloads output is overwhelming, so I usually combine it with flamegraph
```
/iostack.py -K -f > stack
flamegraph.pl --reverse stack  --countname "bytes" > out.svg
```

Another point is that the bpftrace one-liner you showed and soem other tools from bcc/tools do not work with bio-based block devices. Like bio based null block dev
```
modprobe null_blk queue_mode=0 gb=1000 bs=4096 irqmode=0 nr_devices=1
````
and do any io to /dev/nullb0 and you do not see them in the output of this one-liner.

What do you think about atomic increments I used instead of just map.increment(...)? I noticed some bcc tools lose the information about io if the workload is performed in multiple io threads since I believe map.increment(...) is not thread safe.


Nevertheless, thanks for you feedback, If this tool is not really that good then I think you can just close this PR.

PR #2970, opened 2020-06-16
libbpf-tools: add CO-RE bitesize

ethercflow:
- Based on https://github.com/iovisor/bcc/blob/master/tools/bitesize.py, add a comm filter.
- Extract `log/log2` into a BPF-only helper header `bits.bpf.h`.

### DIFF
--- a/libbpf-tools/bitesize.bpf.c
+++ b/libbpf-tools/bitesize.bpf.c
@@ -1,0 +1,59 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2020 Wenbo Zhang
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include "bitesize.h"
+#include "bits.bpf.h"
+
+const volatile char targ_comm[TASK_COMM_LEN] = {};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 10240);
+	__type(key, struct hist_key);
+	__type(value, struct hist);
+	__uint(map_flags, BPF_F_NO_PREALLOC);
+} hists SEC(".maps");
+
+static struct hist initial_hist;
+
+static __always_inline bool comm_filtered(const char *comm)
+{
+	int i;
+
+	for (i = 0; targ_comm[i] != '\0' && i < TASK_COMM_LEN; i++) {
+		if (comm[i] != targ_comm[i])
+			return false;
+	}
+	return true;
+}
+
+SEC("tp_btf/block_rq_issue")
+int BPF_PROG(tp_btf__block_rq_issue, struct request_queue *q,
+	     struct request *rq)
+{
+	struct hist_key hkey;
+	struct hist *histp;
+	u64 slot;
+
+	bpf_get_current_comm(&hkey.comm, sizeof(hkey.comm));
+	if (!comm_filtered(hkey.comm))
+		return 0;
+
+	histp = bpf_map_lookup_elem(&hists, &hkey);
+	if (!histp) {
+		bpf_map_update_elem(&hists, &hkey, &initial_hist, 0);
+		histp = bpf_map_lookup_elem(&hists, &hkey);
+		if (!histp)
+			return 0;
+	}
+	slot = log2l(rq->__data_len / 1024);
+	if (slot >= MAX_SLOTS)
+		slot = MAX_SLOTS - 1;
+	__sync_fetch_and_add(&histp->slots[slot], 1);
+
+	return 0;
+}
+
+char LICENSE[] SEC("license") = "GPL";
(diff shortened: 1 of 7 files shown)


ethercflow:
Hi @anakryiko I'm converting and testing `bio*` tools, this is the first one. PTAL when you're free. Thanks. :)

libbpf-tools/bitesize.h
```diff
@@ -0,0 +1,15 @@
+#ifndef __BITESIZE_H
+#define __BITESIZE_H
+
+#define TASK_COMM_LEN	16
+#define MAX_SLOTS	36
```
anakryiko, line 5: 2^36 is a huge block size, do we really need so many? 30 (up to 1GB) should be plenty, no?

UPD: oh, just noticed that you are counting in KBs, so subtract 10 from that: 20 should be plenty.

libbpf-tools/bitesize.bpf.c
```diff
@@ -0,0 +1,81 @@
...
+}
+
+static __always_inline bool comm_filtered(const char *comm)
+{
+	const volatile char *cond = targ_comm;
+
+	while (*cond) {
+		if (*cond++ != *comm++)
+			return false;
+	}
```
anakryiko, line 50: I'm surprised this is OK with verifier. It should be considered an infinite loop. Also, on a bit older kernels, where loops are not supported, this will fail. Which kernels have you tested?

I think you should rewrite this as a for loop with up to `TASK_COMM_LEN` iterations, and add `#pragma unroll`

libbpf-tools/bitesize.bpf.c
```diff
@@ -0,0 +1,81 @@
...
+
+static __always_inline u64 log2l(u64 v)
+{
+	u32 hi = v >> 32;
+
+	if (hi)
+		return log2(hi) + 32;
+	else
+		return log2(v);
+}
```
anakryiko, line 41: ok, now that we have two tools using this (and more probably are coming, right?), I think it's time to extract this into a BPF-only helper header. Something like `bits.bpf.h` to designate it's BPF-only header (not supposed to be included from user-space code, and some more bit-manipulation functions could be added there eventually?

libbpf-tools/bitesize.bpf.c
ethercflow: I only tested it on `5.7.0-rc7+`, I'll test it on some older kernels, I agree bounded loop with for is better.

libbpf-tools/bitesize.h
ethercflow: Oh sorry for this, I forgot to recalculate this value, although you mentioned it in the last pr, I'll fix it and keep this in mind. Thank you.

libbpf-tools/bitesize.bpf.c
ethercflow: Sure, thanks.

libbpf-tools/bitesize.bpf.c
ethercflow: > I'm surprised this is OK with verifier. It should be considered an infinite loop. Also, on a bit older kernels, where loops are not supported, this will fail. Which kernels have you tested?

@anakryiko Update: I tested on v5.2 ~ v5.5, this is OK from v5.5, I found most of libbpf-tools cann't work on v5.2 ~ v5.4,  for example, this tool will get error on v5.4:

```
libbpf: prog 'tp_btf/block_rq_issue': relo #0: matching candidate #0 request against spec [2473] request + 0:7 => 40.0 @ &x[0].__data_len: 1
libbpf: prog 'tp_btf/block_rq_issue': relo #0: patched insn #36 (LDX/ST/STX) off 40 -> 40
libbpf: block_rq_issue is not found in vmlinux BTF
libbpf: failed to load object 'bitesize_bpf'
libbpf: failed to load BPF skeleton 'bitesize_bpf': -2
failed to load BPF object: -2
```

So we minimum support v5.5 for all libbpf-tools? If so, can we use this kind of loop or we should always use `for` loop with `#pragma unroll`

anakryiko: Right, tp_btf programs were added after BPF loop support was added, it seems. So yeah, we don't have to use `#pragma unroll`. But we need to make it a bounded loop anyways (with explicit cap on number of iterations).

ethercflow: Get it, thanks. I'll fix these soon.

anakryiko:
Looks good and simple, thanks!

It would be nice to follow-up also with adjustments to Makefile to make all *.bpf.h headers be a dependency for all *.bpf.o files, so that re-build is detected more reliably. But it doesn't have to be a part of this PR.

ethercflow:
Hi, @yonghong-song PTAL, thanks.

yonghong-song:
[buildbot, test this please]

PR #2973, opened 2020-06-17
Introduce task detector

qu10wenhao:
The task detector is first introduced on Mail-Archive by Michael Wang, which could trace the related schedule events of a specified task and be helpful on debugging the competition on CPU resource.

As Andrii suggested, now we have rewrite this tool based on libbpf and CO-RE.

We open this pull request to merge this task detector tool into BCC tools.

### DIFF
--- a/libbpf-tools/task_detector.bpf.c
+++ b/libbpf-tools/task_detector.bpf.c
@@ -1,7 +1,5 @@
 // SPDX-Licence-Identifier: GPL-2.0
-// Copyright (c) 2020 Wenhao Qu
-//
-// Based on task_detector by Yun Wang
+// Copyright (c) 2020 Alibaba Cloud
 #include "vmlinux.h"
 #include <bpf/bpf_helpers.h>
 #include "task_detector.h"
--- a/libbpf-tools/task_detector.bpf.c
+++ b/libbpf-tools/task_detector.bpf.c
@@ -1,0 +1,266 @@
+// SPDX-Licence-Identifier: GPL-2.0
+// Copyright (c) 2020 Wenhao Qu
+//
+// Based on task_detector by Yun Wang
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include "task_detector.h"
+
+#define DEBUG_ON 0
+#define TASK_RUNNING 0x0000
+#define TASK_REPORT_MAX 0x0100
+
+struct{
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__uint(max_entries, 1);
+	__type(key, int);
+	__type(value, struct user_info);
+} user_info_maps SEC(".maps");
+
+struct{
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__uint(max_entries, NR_ENTRY_MAX);
+	__type(key, int);
+	__type(value, struct trace_info);
+} trace_info_maps SEC(".maps");
+
+struct{
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, NR_ENTRY_MAX);
+	__type(key, struct si_key);
+	__type(value, u64);
+} syscall_info_maps SEC(".maps");
+
+struct{
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__uint(max_entries, 1);
+	__type(key, int);
+	__type(value, int);
+} start_maps SEC(".maps");
+
+struct{
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__uint(max_entries, 1);
+	__type(key, int);
+	__type(value, int);
+} end_maps SEC(".maps");
+
+struct{
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__uint(max_entries, 1);
+	__type(key, int);
+	__type(value, int);
+} trace_on_maps SEC(".maps");
+
+static inline void *get_user_info(void)
+{
+	int key = 0;
+
+	return bpf_map_lookup_elem(&user_info_maps, &key);
+}
+
+static inline void set_trace_on(int cpu)
+{
+	int key = 0;
+	int *trace_on = bpf_map_lookup_elem(&trace_on_maps, &key);
+
+	if (trace_on)
+		*trace_on = cpu + 1;
+}
+
+static inline void set_trace_off(void)
+{
+	int key = 0;
+	int *trace_on = bpf_map_lookup_elem(&trace_on_maps, &key);
+
+	if (trace_on)
+		*trace_on = 0;
+}
+
+static inline int should_trace(int cpu)
+{
+	int key = 0;
+	int *trace_on = bpf_map_lookup_elem(&trace_on_maps, &key);
+
+	return (trace_on && *trace_on == cpu + 1);
+}
+
+static inline void *get_trace(int key)
+{
+	return bpf_map_lookup_elem(&trace_info_maps, &key);
+}
+
+static inline void add_trace(struct trace_info ti)
+{
+	int orig_end, key = 0;
+	struct trace_info *tip;
+	int *start, *end;
+
+	start = bpf_map_lookup_elem(&start_maps, &key);
+	end = bpf_map_lookup_elem(&end_maps, &key);
+
+	if (!start || !end)
+		return;
+
+	orig_end = *end;
+	*end = (*end + 1);
+
+	if (*end == NR_ENTRY_MAX)
+		*end = 0;
+
+	if (*end == *start)
+		return;
+
+	//tip = get_trace(orig_end);
+	//if (!tip) {
+	//	*end = orig_end;
+	//	return;
+	//}
+
+	//memcpy(tip, &ti, sizeof(ti));
+	ti.ts = bpf_ktime_get_ns();
+
+	if (ti.type == TYPE_SYSCALL_ENTER ||
+	    ti.type == TYPE_SYSCALL_EXIT ||
+	    ti.type == TYPE_WAIT ||
+	    ti.type == TYPE_DEQUEUE)
+		bpf_get_current_comm(&ti.comm, sizeof(ti.comm));
+	bpf_map_update_elem(&trace_info_maps, &orig_end, &ti, 0);
+}
+
+SEC("tp_btf/sched_process_exit")
+int handle__sched_process_exit(u64 *ctx)
+{
+	struct task_struct *p = (void *)ctx[0];
+	struct user_info *ui = get_user_info();
+
+	if (ui && ui->pid && ui->pid == p->pid)
+		ui->exit = 1;
+
+	return 0;
+}
+
+SEC("tp_btf/sched_migrate_task")
+int handle__sched_migrate_task(u64 *ctx)
+{
+	struct task_struct *p = (void *) ctx[0];
+	int dest_cpu = (int) ctx[1];
+	struct trace_info ti = {
+		.cpu = dest_cpu,
+		.pid = p->pid,
+		.type = TYPE_MIGRATE,
+	};
+	struct user_info *ui = get_user_info();
+
+	if (!ui || !ui->pid || ui->pid != p->pid)
+		return 0;
+
+	set_trace_on(dest_cpu);
+
+	add_trace(ti);
+
+	return 0;
+}
+
+SEC("tp_btf/sched_wakeup")
+int handle__sched_wakeup(u64 *ctx)
+{
+	struct task_struct *p = (void *) ctx[0];
+	//int target_cpu = task_cpu(p);
+	struct trace_info ti = {
+		.cpu = p->wake_cpu,
+		.pid = p->pid,
+		.type = TYPE_ENQUEUE,
+	};
+	struct user_info *ui = get_user_info();
+
+	if (!ui || !ui->pid || ui->pid != p->pid)
+		return 0;
+
+	set_trace_on(p->wake_cpu);
+
+	add_trace(ti);
+
+	return 0;
+}
+
+SEC("tp_btf/sched_switch")
+int handle__sched_switch(u64 *ctx)
+{
+	struct task_struct *prev = (void *) ctx[1];
+	struct task_struct *next = (void *) ctx[2];
+	struct trace_info ti = {
+		.cpu = bpf_get_smp_processor_id(),
+	}, *tip;
+	struct user_info *ui = get_user_info();
+
+	if (!ui || !ui->pid)
+		return 0;
+
+	if (!should_trace(ti.cpu)) {
+		if (ui->pid != prev->pid &&
+		    ui->pid != next->pid)
+			return 0;
+
+		set_trace_on(ti.cpu);
+
+		ti.pid = ui->pid;
+		ti.type = TYPE_MIGRATE;
+		add_trace(ti);
+	}
+
+	if (prev->state != TASK_RUNNING &&
+	    prev->state != TASK_REPORT_MAX) {
+		if (ui->pid == prev->pid)
+			set_trace_off();
+		ti.type = TYPE_DEQUEUE;
+	} else
+		ti.type = TYPE_WAIT;
+	ti.pid = prev->pid;
+	add_trace(ti);
+
+	if (!should_trace(ti.cpu))
+		return 0;
+
+	ti.type = TYPE_EXECUTE;
+	ti.pid = next->pid,
+	add_trace(ti);
+
+	return 0;
+}
+
+SEC("tracepoint/raw_syscalls/sys_enter")
+int bpf_trace_sys_enter(struct trace_event_raw_sys_enter *args)
+{
+	struct trace_info ti = {
+		.cpu = bpf_get_smp_processor_id(),
+		.pid = bpf_get_current_pid_tgid(),
+		.type = TYPE_SYSCALL_ENTER,
+		.syscall = args->id,
+	}, *tip;
+	struct user_info *ui = get_user_info();
+
+	if (args->id && ui && ui->trace_syscall && should_trace(ti.cpu))
+		add_trace(ti);
+
+	return 0;
+}
+
+SEC("tracepoint/raw_syscalls/sys_exit")
+int bpf_trace_sys_exit(struct trace_event_raw_sys_exit *args)
+{
+	struct trace_info ti = {
+		.cpu = bpf_get_smp_processor_id(),
+		.pid = bpf_get_current_pid_tgid(),
+		.type = TYPE_SYSCALL_EXIT,
+		.syscall = args->id,
+	}, *tip;
+	struct user_info *ui = get_user_info();
+
+	if (args->id && ui && ui->trace_syscall && should_trace(ti.cpu))
+		add_trace(ti);
+
+	return 0;
+}
+
+char _license[] SEC("license") = "GPL";
--- a/libbpf-tools/task_detector.c
+++ b/libbpf-tools/task_detector.c
@@ -1,7 +1,12 @@
 // SPDX-Licence-Identifier: GPL-2.0
-// Copyright (c) 2020 Wenhao Qu
+// Copyright (c) 2020 Alibaba Cloud
 //
-// Based on task_detector by Yun Wang
+// Based on task_detector by Michael Wang
+// Created by Wenhao Qu
+//
+// Maintainers:
+// Michael Wang <yun.wang@linux.alibaba.com>
+// Wenhao Qu <quxi.qwh@alibaba-inc.com>
 #include <argp.h>
 #include <stdio.h>
 #include <signal.h>
@@ -24,15 +29,15 @@
 #include "task_detector.skel.h"
 
 const char *argp_program_version = "task_detector 0.1";
-const char *argp_program_bug_address = "<yun.wang@xxxxxxxxxxxxxxxxx>";
+const char *argp_program_bug_address = "<bpf@vger.kernel.org>";
 static const char argp_program_doc[] =
 "Trace the related schedule events of a specified task.\n"
 "\n"
 "USAGE: task_detector -p PID [-s]\n"
 "\n"
 "EXAMPLES:\n"
-"    runqslower -p 49870       	# trace pid 49870\n"
-"    runqslower -p 49870 -s    	# trace pid 49870 and system call\n";
+"    task_detector -p 49870       	# trace pid 49870\n"
+"    task_detector -p 49870 -s    	# trace pid 49870 and system call\n";
 
 static const struct argp_option opts[] = {
 	{ "pid", 'p', "PID", 0, "Process PID to trace"},
--- a/libbpf-tools/task_detector.c
+++ b/libbpf-tools/task_detector.c
@@ -1,0 +1,371 @@
+// SPDX-Licence-Identifier: GPL-2.0
+// Copyright (c) 2020 Wenhao Qu
+//
+// Based on task_detector by Yun Wang
+#include <argp.h>
+#include <stdio.h>
+#include <signal.h>
+#include <sched.h>
+#include <string.h>
+#include <unistd.h>
+#include <fcntl.h>
+#include <locale.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <sys/time.h>
+#include <sys/resource.h>
+#include <sys/wait.h>
+#include <dirent.h>
+#include <bpf/bpf.h>
+#include "errno_helpers.h"
+#include "trace_helpers.h"
+#include "syscall_helpers.h"
+#include "task_detector.h"
+#include "task_detector.skel.h"
+
+const char *argp_program_version = "task_detector 0.1";
+const char *argp_program_bug_address = "<yun.wang@xxxxxxxxxxxxxxxxx>";
+static const char argp_program_doc[] =
+"Trace the related schedule events of a specified task.\n"
+"\n"
+"USAGE: task_detector -p PID [-s]\n"
+"\n"
+"EXAMPLES:\n"
+"    runqslower -p 49870       	# trace pid 49870\n"
+"    runqslower -p 49870 -s    	# trace pid 49870 and system call\n";
+
+static const struct argp_option opts[] = {
+	{ "pid", 'p', "PID", 0, "Process PID to trace"},
+	{ "syscall", 's', NULL, 0, "Trace SYSCALL Info"},
+	{},
+};
+
+static struct env {
+	int ui_map_fd;
+	int ti_map_fd;
+	int si_map_fd;
+	int start_map_fd;
+	int end_map_fd;
+	bool exiting;
+	int nr_cpus;
+	int trace_syscall;
+	int target;
+} env = {
+	.exiting = false,
+};
+
+static error_t parse_arg(int key, char *arg, struct argp_state *state)
+{
+	int pid;
+	switch (key) {
+	case 'p':
+		if(arg == NULL) {
+			fprintf(stderr, "PID is required\n");
+			argp_usage(state);
+		} else {
+			pid = strtol(arg, NULL, 10);
+			if (pid <= 0) {
+				fprintf(stderr, "Invalid PID: %s\n", arg);
+				argp_usage(state);
+			}
+			env.target = pid;
+		}
+		break;
+	case 's':
+		env.trace_syscall = 1;
+		break;
+	case ARGP_KEY_END:
+		if (!env.target) {
+			fprintf(stderr, "No target PID\n");
+			argp_usage(state);
+		}
+		break;
+	default:
+		return ARGP_ERR_UNKNOWN;
+	}
+	return 0;
+}
+
+
+static inline int get_comm(int pid, char comm[])
+{
+	int fd, ret = 1;
+	char path_buf[512];
+
+	comm[0] = '\0';
+
+	if (!pid) {
+		strcpy(comm, "IDLE");
+		return 0;
+	}
+
+	snprintf(path_buf, sizeof(path_buf), "/proc/%d/comm", pid);
+	fd = open(path_buf, O_RDONLY);
+	if (fd) {
+		int cnt;
+
+		cnt = read(fd, comm, 16);
+		if (cnt > 0) {
+			comm[cnt - 1] = '\0';
+			ret = 0;
+		}
+	}
+	close(fd);
+	return ret;
+}
+
+static inline int time_to_str(u64 ns, char *buf, size_t len)
+{
+	u64 us = ns / 1000;
+	u64 ms = us / 1000;
+	u64 s = ms / 1000;
+
+	if (s > 10)
+		snprintf(buf, len, "%llus", s);
+	else if (ms > 10)
+		snprintf(buf, len, "%llums", ms);
+	else if (us > 10)
+		snprintf(buf, len, "%lluus", us);
+	else
+		snprintf(buf, len, "%lluns", ns);
+
+	return (s || ms > 10);
+}
+
+static inline void pr_ti(struct trace_info *ti, char *opt, char *delay)
+{
+	char comm[16];
+
+	if (get_comm(ti->pid, comm))
+		memcpy(comm, ti->comm, sizeof(comm));
+
+	printf("%-27lluCPU=%-7dPID=%-7dCOMM=%-20s%-37s%-17s\n",
+				ti->ts, ti->cpu, ti->pid, comm, opt,
+				delay ? delay : "");
+}
+
+static int print_trace_info(int start, int end)
+{
+	int key;
+	char d_str[16];
+	char comm_buf[2*TASK_COMM_LEN];
+	static u64 w_start, p_start;
+	static struct trace_info ti_last;
+
+	for (key = start; key != end; key = (key + 1) % NR_ENTRY_MAX) {
+		char func[80];
+		struct trace_info ti;
+		struct si_key sik;
+		u64 siv;
+
+		if (bpf_map_lookup_elem(env.ti_map_fd, &key, &ti))
+			continue;
+
+		time_to_str(ti.ts - ti_last.ts, d_str, sizeof(d_str));
+
+		switch (ti.type) {
+		case TYPE_MIGRATE:
+			w_start = p_start = ti.ts;
+			pr_ti(&ti, "MIGRATE", NULL);
+			break;
+		case TYPE_ENQUEUE:
+			w_start = p_start = ti.ts;
+			printf("----------------------------\n");
+			pr_ti(&ti, "ENQUEUE", NULL);
+			break;
+		case TYPE_WAIT:
+			if (ti.pid == env.target) {
+				w_start = ti.ts;
+				pr_ti(&ti, "WAIT AFTER EXECUTED", d_str);
+			} else {
+				time_to_str(ti.ts - p_start,
+						d_str, sizeof(d_str));
+				pr_ti(&ti, "PREEMPTED", d_str);
+			}
+			break;
+		case TYPE_EXECUTE:
+			if (ti.pid == env.target) {
+				time_to_str(ti.ts - w_start,
+						d_str, sizeof(d_str));
+				pr_ti(&ti, "EXECUTE AFTER WAITED", d_str);
+			} else {
+				p_start = ti.ts;
+				pr_ti(&ti, "PREEMPT", NULL);
+			}
+			break;
+		case TYPE_DEQUEUE:
+			if (ti.pid == env.target)
+				pr_ti(&ti, "DEQUEUE AFTER EXECUTED", d_str);
+			else {
+				time_to_str(ti.ts - p_start,
+						d_str, sizeof(d_str));
+				pr_ti(&ti, "DEQUEUE AFTER PREEMPTED", d_str);
+			}
+			break;
+		case TYPE_SYSCALL_ENTER:
+			siv = ti.ts;
+			sik.cpu = ti.cpu;
+			sik.pid = ti.pid;
+			sik.syscall = ti.syscall;
+			bpf_map_update_elem(env.si_map_fd, &sik, &siv, BPF_ANY);
+			syscall_name(ti.syscall, comm_buf, sizeof(comm_buf));
+			snprintf(func, sizeof(func), "SC [%d:%s] ENTER",
+					ti.syscall, comm_buf);
+			pr_ti(&ti, func, NULL);
+			break;
+		case TYPE_SYSCALL_EXIT:
+			sik.cpu = ti.cpu;
+			sik.pid = ti.pid;
+			sik.syscall = ti.syscall;
+			if (bpf_map_lookup_elem(env.si_map_fd, &sik, &siv))
+				break;
+			time_to_str(ti.ts - siv, d_str, sizeof(d_str));
+			bpf_map_delete_elem(env.si_map_fd, &sik);
+			syscall_name(ti.syscall, comm_buf, sizeof(comm_buf));
+			snprintf(func, sizeof(func), "SC [%d:%s] TAKE %s TO EXIT",
+					ti.syscall, comm_buf, d_str);
+			pr_ti(&ti, func, NULL);
+			break;
+		default:
+			break;
+		}
+
+		memcpy(&ti_last, &ti, sizeof(ti));
+	}
+
+	return end;
+}
+
+static int setup_user_info(int pid)
+{
+	int key = 0;
+	DIR *dir;
+	char buf[256];
+	struct user_info ui;
+	snprintf(buf, sizeof(buf), "/proc/%d", pid);
+	dir = opendir(buf);
+	if (!dir) {
+		printf("Open %s failed: %s\n", buf, strerror(errno));
+		return -1;
+	}
+	memset(&ui, 0, sizeof(ui));
+	ui.pid = pid;
+	ui.trace_syscall = env.trace_syscall;
+	bpf_map_update_elem(env.ui_map_fd, &key, &ui, BPF_ANY);
+
+	closedir(dir);
+
+	return 0;
+}
+
+static void int_exit(int sig)
+{
+	env.exiting = true;
+}
+
+int main(int argc, char **argv)
+{
+	struct task_detector_bpf *obj;
+	static const struct argp argp = {
+		.options = opts,
+		.parser = parse_arg,
+		.doc = argp_program_doc,
+	};
+	
+	int err;
+
+	err = argp_parse(&argp, argc, argv, 0, NULL, NULL);
+	if (err)
+		return err;
+
+	init_syscall_names();
+	
+	/* Check cpu number */
+	env.nr_cpus = sysconf(_SC_NPROCESSORS_CONF);
+	if (env.nr_cpus > NR_CPU_MAX) {
+		printf("Support Maximum %d cpus\n", NR_CPU_MAX);
+		goto freename;
+	}
+	
+	/* Increase rlimit */
+	err = bump_memlock_rlimit();
+	if (err) {
+		fprintf(stderr, "failed to increase rlimit: %d\n", err);
+		goto freename;
+	}
+	
+	/* Open bpf object */
+	obj = task_detector_bpf__open();
+	if(!obj){
+		fprintf(stderr, "failed to open and/or load BPF object\n");
+		goto freename;
+	}
+
+	/* Load bpf program */
+	err = task_detector_bpf__load(obj);
+	if(err){
+		fprintf(stderr, "failed to load BPF object: %d\n", err);
+		goto cleanup;
+	}
+	
+	/* Attach bpf program */
+	err = task_detector_bpf__attach(obj);
+	if (err) {
+		fprintf(stderr, "failed to attach BPF programs\n");
+		goto cleanup;
+	}
+
+	/* Setup global map fd */
+	env.ui_map_fd = bpf_map__fd(obj->maps.user_info_maps);
+	env.ti_map_fd = bpf_map__fd(obj->maps.trace_info_maps);
+	env.si_map_fd = bpf_map__fd(obj->maps.syscall_info_maps);
+	env.start_map_fd = bpf_map__fd(obj->maps.start_maps);
+	env.end_map_fd = bpf_map__fd(obj->maps.end_maps);
+
+	/* Setip user info */
+	if (setup_user_info(env.target)) {
+		printf("Illegal Target %d\n", env.target);
+		goto cleanup;
+	}
+	
+	signal(SIGINT, int_exit);
+	signal(SIGTERM, int_exit);
+
+	printf("Start tracing schedule events ");
+	if (env.trace_syscall)
+		printf("(include SYSCALL)");
+	printf("\nTarget task pid %d\n", env.target);
+	
+	/* main: print trace info */
+	while (1) {
+		if (env.exiting)
+			goto cleanup;
+		int key = 0, start = 0, end = 0;
+		struct user_info ui = {
+			.exit = 0,
+		};
+
+		bpf_map_lookup_elem(env.ui_map_fd, &key, &ui);
+		if (ui.exit) {
+			printf("Target \"%d\" Destroyed\n", env.target);
+			goto cleanup;
+		}
+
+		bpf_map_lookup_elem(env.start_map_fd, &key, &start);
+		bpf_map_lookup_elem(env.end_map_fd, &key, &end);
+		if (start == end) {
+			sleep(1);
+			continue;
+		}
+
+		start = print_trace_info(start, end);
+
+		bpf_map_update_elem(env.start_map_fd, &key, &start, BPF_ANY);
+	}
+
+cleanup:
+	task_detector_bpf__destroy(obj);
+freename:
+	free_syscall_names();
+	return err != 0;
+}
--- a/libbpf-tools/.gitignore
+++ b/libbpf-tools/.gitignore
@@ -8,3 +8,4 @@
 /syscount
 /vfsstat
 /xfsslower
+/task_detector
--- a/libbpf-tools/Makefile
+++ b/libbpf-tools/Makefile
@@ -19,6 +19,7 @@ APPS = \
 	syscount \
 	vfsstat \
 	xfsslower \
+	task_detector \
 	#
 
 COMMON_OBJ = \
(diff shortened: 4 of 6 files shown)


anakryiko:
As I said on mailing list, I think this is useful and generic tool, I like it. I'd call it something like `schedsnoop`, task detector doesn't ring a bell for me, honestly.

I'll go over the code itself, but I hope @brendangregg could give a high-level feedback w.r.t. proper documentation and similar. Thanks!

libbpf-tools/task_detector.bpf.c
```diff
@@ -0,0 +1,264 @@
...
+	return 0;
+}
+
+SEC("tp_btf/sched_switch")
+int handle__sched_switch(u64 *ctx)
+{
+	struct task_struct *prev = (void *) ctx[1];
+	struct task_struct *next = (void *) ctx[2];
+	struct trace_info ti = {
+		.cpu = bpf_get_smp_processor_id(),
```
anakryiko, line 191: you call this before you check for pid, so it will be a wasted work for a lot of cases. It's better to check pid, then do all the initializations. That way you spend resources only on events that are going to be useful.

libbpf-tools/task_detector.c
```diff
@@ -0,0 +1,376 @@
...
+		}
+		break;
+	default:
+		return ARGP_ERR_UNKNOWN;
+	}
+	return 0;
+}
+
+
+static inline int get_comm(int pid, char comm[])
```
anakryiko, line 95: getting process name from procfs is not great, because process could be dead by that time. It's very simple to get this on BPF side and just have it recorded. Consider doing that instead.

libbpf-tools/task_detector.c
```diff
@@ -0,0 +1,376 @@
...
+	if (s > 10)
+		snprintf(buf, len, "%llus", s);
+	else if (ms > 10)
+		snprintf(buf, len, "%llums", ms);
+	else if (us > 10)
+		snprintf(buf, len, "%lluus", us);
+	else
+		snprintf(buf, len, "%lluns", ns);
+
+	return (s || ms > 10);
```
anakryiko, line 137: why this return result? it's confusing and not used anywhere

libbpf-tools/task_detector.c
```diff
@@ -0,0 +1,376 @@
...
+	u64 s = ms / 1000;
+
+	if (s > 10)
+		snprintf(buf, len, "%llus", s);
+	else if (ms > 10)
+		snprintf(buf, len, "%llums", ms);
+	else if (us > 10)
+		snprintf(buf, len, "%lluus", us);
+	else
+		snprintf(buf, len, "%lluns", ns);
```
anakryiko, line 135: nit: I know CPUs are fast, but there is no reason to do all three divisions. If you write this like this, it will be as simple to follow, but a bit less sloppy, engineering-wise:

```
#define NS_IN_SEC 1000000000LLU
#define NS_IN_MS 1000000LLU
#define NS_IN_US 1000LLU

if (ns > 10 * NS_IN_SEC)
  snprintf(buf, len, "%llus", ns / NS_IN_SEC);
else if (ns > 10 * NS_IN_MS)
  snprintf(buf, len, "%llums", ns / NS_IN_MS);
else if (ns > 10 * NS_IN_US)
  snprintf(buf, len, "%lluus", ns / NS_IN_US);
else
  snprintf(buf, len, "%lluns", ns);
```

libbpf-tools/task_detector.c
```diff
@@ -0,0 +1,376 @@
...
+}
+
+static inline void pr_ti(struct trace_info *ti, char *opt, char *delay)
+{
+	char comm[16];
+
+	if (get_comm(ti->pid, comm))
+		memcpy(comm, ti->comm, sizeof(comm));
+
+	printf("%-27lluCPU=%-7dPID=%-7dCOMM=%-20s%-37s%-17s\n",
```
anakryiko, line 147: timestamp as a raw number from some unknown original seems a bit confusing. It would be nice to convert it to something human-readable. you can either use a new bpf_ktime_boot_ns() (but then you'll need very fresh kernel), or you can remember kernel time at the beginning of profiling and then subtract that from each sample timestamp to show relative time from the start of profiling. But either was having us/ms/s is better than raw ns-based timestamp with arbitrary origin, IMO.

libbpf-tools/task_detector.bpf.c
```diff
@@ -0,0 +1,264 @@
...
+
+struct{
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__uint(max_entries, 1);
+	__type(key, int);
+	__type(value, struct user_info);
+} user_info_maps SEC(".maps");
+
+struct{
+	__uint(type, BPF_MAP_TYPE_ARRAY);
```
anakryiko, line 19: this can also be just a global variable array:

```
struct trace_info trace_infos[NR_ENTRY_MAX];
```

From user-space, iterating it will be also nice and simple:

```
for (i = 0; i < NR_ENTRY_MAX; i++) {
    struct trace_info *ti = &skel->bss->trace_infos[i];
   ...
}
```

libbpf-tools/task_detector.bpf.c
```diff
@@ -0,0 +1,264 @@
...
+	add_trace(ti);
+
+	return 0;
+}
+
+SEC("tp_btf/sched_wakeup")
+int handle__sched_wakeup(u64 *ctx)
+{
+	struct task_struct *p = (void *) ctx[0];
+	//int target_cpu = task_cpu(p);
```
anakryiko, line 167: debug comments sprinkled around?

libbpf-tools/task_detector.c
```diff
@@ -0,0 +1,376 @@
...
+
+static int print_trace_info(int start, int end)
+{
+	int key;
+	char d_str[16];
+	char comm_buf[2*TASK_COMM_LEN];
+	static u64 w_start, p_start;
+	static struct trace_info ti_last;
+
+	for (key = start; key != end; key = (key + 1) % NR_ENTRY_MAX) {
```
anakryiko, line 160: see comment above about perf_buffer, this seems like a better fit here -- you'd just "listen" to events and emit them as they come. See pretty much all other libbpf-tools, they always do perf_buffer.

libbpf-tools/task_detector.c
```diff
@@ -0,0 +1,376 @@
...
+}
+
+static int setup_user_info(int pid)
+{
+	int key = 0;
+	DIR *dir;
+	char buf[256];
+	struct user_info ui;
+	snprintf(buf, sizeof(buf), "/proc/%d", pid);
+	dir = opendir(buf);
```
anakryiko, line 251: so you are checking if process with such PID exists, is that right? Do other BCC tools ever do this?...

libbpf-tools/task_detector.c
```diff
@@ -0,0 +1,376 @@
...
+	bpf_map_update_elem(env.ui_map_fd, &key, &ui, BPF_ANY);
+
+	closedir(dir);
+
+	return 0;
+}
+
+static void int_exit(int sig)
+{
+	env.exiting = true;
```
anakryiko, line 268: please move `exiting` from struct env, and make it `static volatile` variable. Env is for input parameters, not for internal state. Plus you need volatile, as it's updated from interrupt handler.

libbpf-tools/task_detector.c
```diff
@@ -0,0 +1,376 @@
...
+		fprintf(stderr, "failed to attach BPF programs\n");
+		goto cleanup;
+	}
+
+	/* Setup global map fd */
+	env.ui_map_fd = bpf_map__fd(obj->maps.user_info_maps);
+	env.ti_map_fd = bpf_map__fd(obj->maps.trace_info_maps);
+	env.si_map_fd = bpf_map__fd(obj->maps.syscall_info_maps);
+	env.start_map_fd = bpf_map__fd(obj->maps.start_maps);
+	env.end_map_fd = bpf_map__fd(obj->maps.end_maps);
```
anakryiko, line 328: as I mentioned above, env is not intended to keep all programs state, for that just static global variables are a nicer alternative for such small tools. env is for what's parsed from command line arguments.

libbpf-tools/task_detector.bpf.c
```diff
@@ -0,0 +1,264 @@
...
+	end = bpf_map_lookup_elem(&end_maps, &key);
+
+	if (!start || !end)
+		return;
+
+	orig_end = *end;
+	*end = (*end + 1);
+
+	if (*end == NR_ENTRY_MAX)
+		*end = 0;
```
anakryiko, line 107: that's an interesting approach. Did you consider using perf_buffer instead and send samples to user-space as they arrive? Any downsides to doing that?

qu10wenhao:
Hello Andrii,

I have followed all of your suggestions, which inspire me a lot. By using the global variable and the perf_buff, the amount of code has been greatly reduced. Also I have renamed this tool as schedsnoop. :)

However, I failed to use the BPF_PROG marco and I also noticed that other tools in libbpf-tools do not use this marco either. The definition of this marco is found under bcc/src/cc/export/helpers.h but I can't find any samples or any document about it.  It will be very pleasure if you could provide some links for me to study.

Thanks

anakryiko:
BPF_PROG comes from libbpf header `bpf_tracing.h`. You can see a lot of examples in kernel selftests/bpf. E.g. : https://github.com/torvalds/linux/blob/master/tools/testing/selftests/bpf/progs/test_vmlinux.c#L75

libbpf-tools/schedsnoop.c
```diff
@@ -0,0 +1,309 @@
...
+		pr_ti(ti, "MIGRATE", NULL);
+		break;
+	case TYPE_ENQUEUE:
+		w_start = p_start = ti->ts;
+		printf("----------------------------\n");
+		pr_ti(ti, "ENQUEUE", NULL);
+		break;
+	case TYPE_WAIT:
+		if (ti->pid == env.target) {
+			w_start = ti->ts;
```
anakryiko, line 146: you will actually have a problem with re-ordering of events. If they happen on two different CPUs (e.g., when task is migrated) you can get events out of order, because perfbuf has per-CPU buffers.

I ignored that problem initially, because I assumed all the events are independent, but now I see that you do care about those timestamps.

Luckily, I actually have an even better way to do this for you, except it is quite bleeding edge (you'll need a 5.8 kernel). Do you mind trying that out?

See: https://github.com/torvalds/linux/blob/master/tools/testing/selftests/bpf/progs/test_ringbuf.c and https://github.com/torvalds/linux/blob/master/tools/testing/selftests/bpf/prog_tests/ringbuf.c

BPF_MAP_TYPE_RINGBUF is a shared ringbuffer across all CPUs, so ordering like scheduling events are preserved. It also has a nice reserve/commit API, which seems rather convenient for you here.

libbpf-tools/schedsnoop.c
```diff
@@ -0,0 +1,309 @@
...
+			pr_ti(ti, "WAIT AFTER EXECUTED", d_str);
+		} else {
+			time_to_str(ti->ts - p_start,
+					d_str, sizeof(d_str));
```
anakryiko, line 150: nit: formatting is off?

libbpf-tools/schedsnoop.c
```diff
@@ -0,0 +1,309 @@
...
+			pr_ti(ti, "DEQUEUE AFTER PREEMPTED", d_str);
+		}
+		break;
+	case TYPE_SYSCALL_ENTER:
+		siv = ti->ts;
+		sik.cpu = ti->cpu;
+		sik.pid = ti->pid;
+		sik.syscall = ti->syscall;
+		bpf_map_update_elem(si_map_fd, &sik, &siv, BPF_ANY);
```
anakryiko, line 178: why is this done from user-space and not from BPF? there might be significant delays and it's just less performant overall (and causes extra syscall for every other syscall?)

libbpf-tools/schedsnoop.c
```diff
@@ -0,0 +1,309 @@
...
+	else
+		snprintf(buf, len, "%lluns", ns);
+
+	return 0;
+}
+
+static inline void pr_ti(struct trace_info *ti, char *opt, char *delay)
+{
+	char buf[27];
+	snprintf(buf, sizeof(buf), "%lluus", ti->ts / NS_IN_US);
```
anakryiko, line 115: so you are still printing absolute timestamp and then optionally prepend relative timestamp? why both? why not just using relative all the time?

libbpf-tools/schedsnoop.c
```diff
@@ -0,0 +1,309 @@
...
+
+void handle_event(void *ctx, int cpu, void *data, __u32 data_sz)
+{
+	struct trace_info *ti = data;
+	char d_str[16];
+	char comm_buf[2 * TASK_COMM_LEN];
+	char func[80];
+	struct si_key sik;
+	u64 siv;
+	static u64 w_start, p_start, last_time;
```
anakryiko, line 129: first value of last_time will be zero, which will give you huge delta for the very first sample, that's probably not what you want. You probably want to fetch kernel timestamp right before you attach BPF program.

It's not immediately obvious to me how to get same timestamp from user-space. I suspect it might be one of clock_gettime() with  CLOCK_MONOTONIC or CLOCK_MONOTONIC_RAW, but I haven't checked. It would be good to figure it out, though, I think this question comes up from time to time.

Alternatively, there is bpf_get_ktime_boot_ns(), with more well-defined origin.

libbpf-tools/schedsnoop.c
```diff
@@ -0,0 +1,309 @@
...
+	pb = perf_buffer__new(bpf_map__fd(obj->maps.events), PERF_BUFFER_PAGES, &pb_opts);
+	err = libbpf_get_error(pb);
+	if (err) {
+		pb = NULL;
+		fprintf(stderr, "failed to open perf buffer: %d\n", err);
+		goto cleanup;
+	}
+	
+	/* main: poll */
+	while (!exiting && !obj->bss->targ_exit && \
```
anakryiko, line 293: hmm, you don't need '\' in C statements, I'm even surprised that compiles (but I guess it makes sense). Still, quite unconventional and unnecessary.

libbpf-tools/schedsnoop.c
```diff
@@ -0,0 +1,309 @@
...
+			(err = perf_buffer__poll(pb, 100)) >= 0);
+	if (exiting)
+		goto cleanup;
+	if (obj->bss->targ_exit) {
+		printf("Target %d Destroyed!\n", env.target);
```
anakryiko, line 298: typo: Destroyed -> exited?

libbpf-tools/schedsnoop.h
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2020 Alibaba Cloud
+#ifndef SCHEDSNOOP_H
+#define SCHEDSNOOP_H
+
+#define NR_ENTRY_MAX		10000
+#define NR_CPU_MAX		512
```
anakryiko, line 7: you don't seem to use it, just for that unnecessary check

libbpf-tools/schedsnoop.h
```diff
@@ -0,0 +1,38 @@
...
+#ifndef SCHEDSNOOP_H
+#define SCHEDSNOOP_H
+
+#define NR_ENTRY_MAX		10000
+#define NR_CPU_MAX		512
+#define TASK_COMM_LEN		16
+
+
+typedef unsigned long long u64;
+typedef int pid_t;
```
anakryiko, line 12: these come from vmlinux.h already, drop it

libbpf-tools/schedsnoop.c
```diff
@@ -0,0 +1,309 @@
...
+{
+	struct trace_info *ti = data;
+	char d_str[16];
+	char comm_buf[2 * TASK_COMM_LEN];
+	char func[80];
+	struct si_key sik;
+	u64 siv;
+	static u64 w_start, p_start, last_time;
+
+
```
anakryiko, line 131: nit: extra line

libbpf-tools/schedsnoop.c
anakryiko: So it seems like you should be able to get the same-origin clock in user-space with clock_gettime(CLOCK_MONOTONIC), according to https://github.com/torvalds/linux/blob/master/include/uapi/linux/bpf.h#L694

libbpf-tools/schedsnoop.bpf.c
```diff
@@ -0,0 +1,190 @@
...
+		add_trace(ctx, &ti);
+	}
+
+	if (prev->state != TASK_RUNNING &&
+	    prev->state != TASK_REPORT_MAX) {
+		if (targ_pid == prev->pid)
+			set_trace_off();
+		ti.type = TYPE_DEQUEUE;
+	} else
+		ti.type = TYPE_WAIT;
```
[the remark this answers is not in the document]
qu10wenhao: When the task switch happened, we want to know whether the previous task wants to dequeue or this task still wants to run but a preemption happens.  If it is a preemption, the status of previous task will be set as TASK_REPORT_MAX (check https://github.com/torvalds/linux/blob/master/include/trace/events/sched.h#L121). Also if the previous status is TASK_RUNNING, it definitely don't want to dequeue.

So in these two cases (prev->state is TASK_REPORT_MAX or TASK_RUNNING), the previous task is still waiting and we should set the trace info type as TYPE_WAIT. Otherwise, the trace info type should be TYPE_DEQUEUE. I will add some comments to explain it later.

BTW, since the preemption force to set the previous status as TASK_REPORT_MAX, I'm not so sure whether TASK_RUNNING is still required, just in case some special situation exist.

libbpf-tools/schedsnoop.c
```diff
@@ -0,0 +1,309 @@
...
+const char *argp_program_version = "schedsnoop 0.1";
+const char *argp_program_bug_address = "<bpf@vger.kernel.org>";
+static const char argp_program_doc[] =
+"Trace the related schedule events of a specified task.\n"
+"\n"
+"USAGE: schedsnoop -p PID [-s]\n"
+"\n"
+"EXAMPLES:\n"
+"    schedsnoop -p 49870       	# trace pid 49870\n"
```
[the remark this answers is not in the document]
qu10wenhao: Tracing all threads of a given process is a great idea and we are working on it. Currently, if our target process has multiple threads, we will open multiple consoles and use this tool to trace each threads respectively.

A small problem is, how to show the traceinfos? Simply output them to the console is not so good. Even now when we trace one single thread, it output plenty of information to the console. When we  trace multiple threads at the same time, it will be a big chaos.

Our idea is that we only output very little information to the console, such as how many threads are we tracing, which of them are working and which of them are preempted. Detail information will be written to certain files related to each threads. Do you think it is ok? Or could you please give us some suggestions.

libbpf-tools/schedsnoop.c
```diff
@@ -0,0 +1,309 @@
...
+		return err;
+
+	init_syscall_names();
+	
+	/* Check cpu number */
+	env.nr_cpus = sysconf(_SC_NPROCESSORS_CONF);
+	if (env.nr_cpus > NR_CPU_MAX) {
+		printf("Support Maximum %d cpus\n", NR_CPU_MAX);
+		goto freename;
+	}
```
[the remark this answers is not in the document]
qu10wenhao: We found this limitation in some old eBPF samples. If it is unnecessary now, we will remove it.

libbpf-tools/schedsnoop.c
qu10wenhao: Well, we do care about the order. we will run some experiments so see how serious the out of order problem is. 

If the perfbuf has per-CPU buffers, does it means that there are per-CPU handlers? E.G. if both cpu#0 and cpu#1 emit an event to the their buffers, will these two events be processed 
simultaneously by related cpu?

And about the shared ringbuffer in kernel 5.8, we are very interesting in this new feature. However, we are worried whether this feature is stable or not. Will the mainline kernel 5.8 include this feature in the future? If it does, we are very willing to have a try

libbpf-tools/schedsnoop.h
qu10wenhao: NR_ENTRY_MAX is used for setting the max entry of syscall_info_maps.

libbpf-tools/schedsnoop.c
qu10wenhao: last_time is used to calculate the time duration between two specific events, which shows how long a certain tasks has been running or how long it has been preempted. So it doesn't matter that the first value is zero or the timestamp that I attach the BPF program.

The only problem is, if the target task has been running for a while before we start tracing it, the first sample may output something strange, e.g. the target task may be preempted for a very long time because of the zero first value. This seems unsolvable since we cannot get any information before we attach the BPF program. (Maybe we can read procfs to get some information? But it's hard to get any information about preemption i think.)

the further output information won't be affected.

libbpf-tools/schedsnoop.c
qu10wenhao: the relative timestamp is not the time different between two adjacent events. It is used to show how long our target task has been executed, how long another task has preempted the target task. It is optional because some events don't have a duration.

the absolute timestamp is not so important here, but we believe it is necessary to keep it. the absolute timestamp may be helpful for developer, especially kernel developer, to trace the system and locate the problem.

qu10wenhao: Maybe I can add a debug options in argp and only output raw timestamp when dubug is on.

qu10wenhao:
I add new commits to fix most of the problem you mentioned. We have also left some ideas and questions. Please check those unresolved conversation :)

What's more, during our previous experiments with old version eBPF, we have noticed a strange scenario. When the tasks are really heavy, we may lose the track on some events. i.e. certain event should happened, but it didn't trigger the attached BPF program so that we won't record it. The most obvious example is, we may lose the sched_migrate event and trace some unexpected sched_switch events on untargeted cpu. 

E.g. 
Target task execute on cpu#1 and thus we keep tracing the schedule events on cpu#1. And suddenly, we find a sched_switch event happened on cpu#2 which switch our target task to another one. This switch event indicates that the target task is now working on cpu#2 but we didn't record any relative sched_migrate event.

This is a very rare situation and thus we add a check in the sched_switch handler. 
[https://github.com/qu10wenhao/bcc/blob/master/libbpf-tools/schedsnoop.bpf.c#L132](url)
We don't know what cause this scenario, whether there is a limitation of perf event (since bpf program links to the tracepoint by perf event) or something else. 

If you have any idea about it, please share it with us. Also if you think it is unnecessary, we may remove this check later.

Thanks!

libbpf-tools/schedsnoop.bpf.c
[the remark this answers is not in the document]
anakryiko: Ah, awesome. I'd suggest to put all these into comments in relevant parts of the code. People can learn a lot just by reading the code that way.

libbpf-tools/schedsnoop.c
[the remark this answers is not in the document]
anakryiko: dumping to file to visualize with some other tool is definitely an approach. Not sure how suitable it is for BCC tools, though if you have an option of outputting, e.g., JSON events, then you can build a pipeline that would parse it and dump in whatever closed/proprietary format is suitable for your visualization tool.

I don't think you need to implement any of that in the first version. More important to get right from the get go is to establishing good terminology: use tid for "thread ID", use "thread" or "task" terminology throughout, because that's what you are tracing.

libbpf-tools/schedsnoop.c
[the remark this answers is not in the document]
anakryiko: I just don't see where this is used or required. So I'd drop, unless we know why we need it.

libbpf-tools/schedsnoop.c
anakryiko: I used to trace fork/exec/exit events, and re-ordering happened regularly enough to be a big problem. For you it will mostly be happening when thread is migrated to a different CPU and keeps running there.

re: per-CPU buffers. Each buffer is independent, but your user-space callback is going to be executed on single CPU. So reordering happens only because you might get events from CPU X before events from CPU Y, even though some of the events on CPU Y happened sooner.

re: BPF RINGBUF, yes, it's stable. If there are any bugs, we'll be fixing them, but so far we haven't seen any report. It's extremely unlikely this will be pulled out. We are going to use it at FB in multiple applications, so I think you can definitely use it.

libbpf-tools/schedsnoop.c
anakryiko: Yeah, command line option to switch between relative vs absolute timestamps make sense. Relative is more user-friendly by default, IMO.

libbpf-tools/schedsnoop.h
anakryiko: that comment is specifically for NR_CPU_MAX

anakryiko:
> This is a very rare situation and thus we add a check in the sched_switch handler.
> https://github.com/qu10wenhao/bcc/blob/master/libbpf-tools/schedsnoop.bpf.c#L132
> We don't know what cause this scenario, whether there is a limitation of perf event (since bpf program links to the tracepoint by perf event) or something else.

Did you see dropped perf events during that time? The simplest reason might be that a spike of events on a CPU would case perf buffer to run out of available space. At which point bpf_perf_event_output() would drop the same.

It might be something else. E.g., tracepoint sub-system might have limitations around re-entrancy and stuff, so I'm not sure. But I'd check perfbuffer overflow first.

libbpf-tools/schedsnoop.c
```diff
@@ -0,0 +1,278 @@
...
+"Trace the related schedule events of a specified task.\n"
+"\n"
+"USAGE: schedsnoop -p PID [-s]\n"
+"\n"
+"EXAMPLES:\n"
+"    schedsnoop -p 49870       	# trace pid 49870\n"
+"    schedsnoop -p 49870 -s    	# trace pid 49870 and system call\n";
+
+static const struct argp_option opts[] = {
+	{ "pid", 'p', "PID", 0, "Process PID to trace"},
```
anakryiko, line 48: see here, it's not process PID, it's thread ID (TID) (from user-space perspective).

libbpf-tools/schedsnoop.c
```diff
@@ -0,0 +1,278 @@
...
+	{ "syscall", 's', NULL, 0, "Trace SYSCALL Info"},
+	{},
+};
+
+static struct env {
+	int trace_syscall;
+	int targ_pid;
+} env;
+
+int volatile si_map_fd;
```
anakryiko, line 58: not used anymore, and volatile is unnecessary

libbpf-tools/schedsnoop.c
```diff
@@ -0,0 +1,278 @@
...
+	
+	/* Attach bpf program */
+	err = schedsnoop_bpf__attach(obj);
+	if (err) {
+		fprintf(stderr, "failed to attach BPF programs\n");
+		goto cleanup;
+	}
+	
+	/* Setup global map fd */
+	si_map_fd = bpf_map__fd(obj->maps.syscall_info_maps);
```
anakryiko, line 241: no need

libbpf-tools/schedsnoop.bpf.c
```diff
@@ -0,0 +1,241 @@
...
+	__u64 *last_ts;
+	__u64 ts = bpf_ktime_get_ns();
+	switch (ti->type) {
+	case TYPE_SYSCALL_ENTER:
+		bpf_map_update_elem(&syscall_info_maps, sik, &ts, BPF_ANY);
+		break;
+	case TYPE_SYSCALL_EXIT:
+		last_ts = bpf_map_lookup_elem(&syscall_info_maps, sik);
+		if(!last_ts)
+			return;
```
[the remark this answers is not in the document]
qu10wenhao: when a syscall exits we need to know when does it start and calculate the duration. If we couldn't find any record in syscall_info_maps, we just drop this event instead of emit it to perfbuffer. That's why we return here.

qu10wenhao:
Well, we did not use perfbuffer to record and output at that time.

So may be the question is, how does perf event trigger the BPF program? E.g. if we create a perf event based on sched_switch tracepoint and attach a BPF program on it, how does the whole system works? 

For my own understanding, each time when schedule switch happened and it comes to the tracepoint, it will call certain function (maybe perf_trace_##call) and emit some information (like perf event name, timestamp, parameters) to the ring buffer. the attached BPF program will also be executed at that time. So the buffer size doesn't matter.

If I'm wrong, maybe there is another thread that keep reading the ring buffer, check each perf events and execute the related BPF programs, then it is most likely the perfbuffer overflow problem

brendangregg:
Does this really trace sched events and print output for each event?

Didn't I reject something similar not that long ago (last year)?

It's a bad idea for two reasons:

A) Overhead. I can imagine my Netflix colleagues running this on our big systems and causing an outage (if it "only" eats one CPU, that'll be enough to trigger fail over), then asking me to help. 
B) If you've followed observability, the _whole_ point of using eBPF was in-kernel summaries so we didn't have to do per-event output. http://www.brendangregg.com/blog/2015-05-15/ebpf-one-small-step.html

And there's another reason: how does it compare to "perf sched record", which already has an optimized binary output file and different tools to post-process? And, there's also "perf record -e sched:*; perf script" which prints a format string that already has similar arguments to what you're printing.

If you think this belongs in BCC, I'd:

1. Publish it on your own repo first.
2. Publish documentation (blog posts) showing real world case studies where it, and only it, can solve the problem.
3. Show overhead measurements in production environments.

You'd really need to make a case that despite the overhead, and despite other tools, there is a good real world production reason for this.

libbpf-tools/schedsnoop.bpf.c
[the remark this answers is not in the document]
anakryiko: ah, right, makes sense

anakryiko:
> Well, we did not use perfbuffer to record and output at that time.
> 
> So may be the question is, how does perf event trigger the BPF program? E.g. if we create a perf event based on sched_switch tracepoint and attach a BPF program on it, how does the whole system works?
> 

This is more complicated than what I can try to explain here. If you are curious to learn all the details, please see kernel sources for how tracepoints are handled by perf subsystem.

> For my own understanding, each time when schedule switch happened and it comes to the tracepoint, it will call certain function (maybe perf_trace_##call) and emit some information (like perf event name, timestamp, parameters) to the ring buffer. the attached BPF program will also be executed at that time. So the buffer size doesn't matter.

I'm talking about PERF_EVENT_ARRAY map and ring buffers that are created by libbpf for each CPU. When you do bpf_perf_event_output(), you are using some amount of data in such ring buffer. E.g., in your case you have 64 4KB pages allocated per each CPU. So if there is a high-frequency scheduling events, you might overflow the buffer before user-space part can consume all the enqueued events.

> 
> If I'm wrong, maybe there is another thread that keep reading the ring buffer, check each perf events and execute the related BPF programs, then it is most likely the perfbuffer overflow problem

Not really. BPF program is executed by kernel on each scheduling event tracepoint. Your BPF program emits data into perf buffer (independent from tracepoints). Then your user-space program does poll those perf buffers for new events, and when available, consumes them. So there is a delay between when BPF program emits data into perf buffer, and when user-space consumes them and frees up that space for more events.

anakryiko:
> Does this really trace sched events and print output for each event?

Yes, but only for events relevant to requested thread.

> 
> Didn't I reject something similar not that long ago (last year)?
> 
> It's a bad idea for two reasons:
> 
> A) Overhead. I can imagine my Netflix colleagues running this on our big systems and causing an outage (if it "only" eats one CPU, that'll be enough to trigger fail over), then asking me to help.

There is certainly an overhead, but I'm not sure that's good enough reason to disqualify the tool? Any tool can be misused. E.g., trace.py on malloc/free or even these scheduling events, and you are pretty surely will overwhelm the system.

We at Facebook actually have a somewhat similar tool, used by many users in production, that records similar events for all threads within the process for some short amount of time, and then visualizes it in a browser in a nice UI, for more detailed inspection. It's certainly quite heavyweight, but if used for short periods of time, it is immensely useful to understand the behavior (especially blocking pattern) of an application.

> B) If you've followed observability, the _whole_ point of using eBPF was in-kernel summaries so we didn't have to do per-event output. http://www.brendangregg.com/blog/2015-05-15/ebpf-one-small-step.html

There are a bunch of applications at Facebook that do 24/7 monitoring and emit events to user-space for post-processing without in-kernel pre-aggregation. Things like fork/exec/exit events, for instance. There is certainly benefit to being able to do in-kernel aggregations, but there are cases where user-space really needs every single event. We've even switched from kernel aggregated stack traces for profiling into just emitting every stack trace straight to user-space (e.g., check bpf_get_stack() helper, which can also return build ID of executable, if requested).

> 
> And there's another reason: how does it compare to "perf sched record", which already has an optimized binary output file and different tools to post-process? And, there's also "perf record -e sched:*; perf script" which prints a format string that already has similar arguments to what you're printing.

Haven't used perf to record scheduling events, but can it do it as intelligently as this schedsnoop tool for one particular task, categorizing what type of event occurred (e.g., pre-empted vs waiting on lock, etc)? I'm sure you can post-process perf recording to extract this, but it will probably also record extra data that is simply filtered by this tool.
> 
> If you think this belongs in BCC, I'd:
> 
>     1. Publish it on your own repo first.
> 
>     2. Publish documentation (blog posts) showing real world case studies where it, and only it, can solve the problem.
> 

This one would be really nice, I agree.

>     3. Show overhead measurements in production environments.

This will very much depend on the workload itself, wouldn't it? So sure they can show some numbers, but I'm not sure how easy would it be to qualify it generically.

qu10wenhao:
Hi @brendangregg,

Our initial target is to find out **WHO** and **WHEN** has preempted our target task for **HOW LONG**. What schedsnoop (aka task detector) does is to filter schedule events, extract schedule behaviors and finally help us understand the whole life time of the certain task we interest. By analyzing the life time, we may figure out the biggest "resource thief" and optimize our system.

>Does this really trace sched events and print output for each event?
>
>Didn't I reject something similar not that long ago (last year)?

Schedsnoop doesn't simply trace the schedule events and print them. We focus on the schedule behaviors relevant to the target task rather than events. E.g. in schedule switch event, there will be two different schedule behaviors. Also  we need to distinguish whether the preemption cause the switch or the task just dequeue by itself.

> And there's another reason: how does it compare to "perf sched record", which already has an optimized binary output file and different tools to post-process? And, there's also "perf record -e sched:*; perf script" which prints a format string that already has similar arguments to what you're printing.

We used to use "perf sched record" before but it does not perform good enough for us. To the best of our knowledge, although perf sched has many useful features (like timehist, map) ,  it's hard to record all events related to a specific task. 

-  "perf record -e sched:*" will record all schedule events, which includes too many useless information and make it hard to target our specific task. Also there will be a huge overhead.

- "perf record -e sched:* -p pid" could only trace schedule events that called by target pid. However, there is a problem about schedule switch. I.e. when task TARGET firstly switches to task B and then task B switches back to task TARGET, only the first switch event will be recorded because the second one is called by task B. Thus in this way, we can only know **WHO** and **When** has preempted our target but we can not know **HOW LONG**.

>Publish it on your own repo first.
>Publish documentation (blog posts) showing real world case studies where it, and only it, can solve the problem.

Yes, I have published it on my own repo [task_detector](https://github.com/qu10wenhao/task_detector)
There is a small README to show some example output. Maybe later I can find some real world scenario and write a document (or just update the README).

qu10wenhao:
>I'm talking about PERF_EVENT_ARRAY map and ring buffers that are created by libbpf for each CPU. When you do bpf_perf_event_output(), you are using some amount of data in such ring buffer. E.g., in your case you have 64 4KB pages allocated per each CPU. So if there is a high-frequency scheduling events, you might overflow the buffer before user-space part can consume all the enqueued events.

Well, you have misunderstand my question. I was wondered how the BPF program is triggered in kernel-space and will there be such a case that certain tracepoint is hit but the attached BPF program does not execute correctly.

Maybe I should bypass this question right now. Based on current design, perfbuffer overflow problem could be serious under heavy scheduling.

anakryiko:
> > I'm talking about PERF_EVENT_ARRAY map and ring buffers that are created by libbpf for each CPU. When you do bpf_perf_event_output(), you are using some amount of data in such ring buffer. E.g., in your case you have 64 4KB pages allocated per each CPU. So if there is a high-frequency scheduling events, you might overflow the buffer before user-space part can consume all the enqueued events.
> 
> Well, you have misunderstand my question. I was wondered how the BPF program is triggered in kernel-space and will there be such a case that certain tracepoint is hit but the attached BPF program does not execute correctly.

That I addressed with "This is more complicated than what I can try to explain here. " :) There is definitely the case where this can happen for certain perf events, if BPF program is "preempted" by another BPF program. You can't do that. But whether that can happen at all depends on tracepoint/perf event, I can't tell you if that's the case here.

> 
> Maybe I should bypass this question right now. Based on current design, perfbuffer overflow problem could be serious under heavy scheduling.

Well you had the same problem with your initial array approach. You had only 10000 elements, right? After that you were either dropping or overwriting data (not sure which one). You single event is 64 bytes long. For BPF ringbuf, overhead per each event is extra 8 bytes. So 72 byte * 10000 ~ 70KB. So if you set BPF ringbuf size to anything >= 70KB, you'll have the same amount of data capacity. For perfbuf, I'm not sure about overhead per event, plus you'd need to decide how big each per-CPU array has to be, but you can do similar calculations.

Hope this makes it more clear.

qu10wenhao:
That's very kind of you. :)

BTW, about the undebug timestamp. I'm going to output the system time like 2020-6-24 15:34:38.xxxxxxxxx, which is calculated based on the event timestamp and the initial timestamp that the BPF program is attached. Is that OK? or maybe less accuracy (ms/us).

brendangregg:
> > Does this really trace sched events and print output for each event?
> 
> Yes, but only for events relevant to requested thread.

I saw that, but that still means a ton of output.

> 
> > Didn't I reject something similar not that long ago (last year)?
> > It's a bad idea for two reasons:
> > A) Overhead. I can imagine my Netflix colleagues running this on our big systems and causing an outage (if it "only" eats one CPU, that'll be enough to trigger fail over), then asking me to help.
> 
> There is certainly an overhead, but I'm not sure that's good enough reason to disqualify the tool? Any tool can be misused. E.g., trace.py on malloc/free or even these scheduling events, and you are pretty surely will overwhelm the system.

In both cases you are choosing to trace something with high frequency ("trace.py something_silly"). But in this case you are just picking a PID -- it would be the first and only tool that can shoot yourself in the foot like this.

> 
> We at Facebook actually have a somewhat similar tool, used by many users in production, that records similar events for all threads within the process for some short amount of time, and then visualizes it in a browser in a nice UI, for more detailed inspection. It's certainly quite heavyweight, but if used for short periods of time, it is immensely useful to understand the behavior (especially blocking pattern) of an application.

Sounds like LTTng Trace Compass or Kernel Shark.

> 
> > B) If you've followed observability, the _whole_ point of using eBPF was in-kernel summaries so we didn't have to do per-event output. http://www.brendangregg.com/blog/2015-05-15/ebpf-one-small-step.html
> 
> There are a bunch of applications at Facebook that do 24/7 monitoring and emit events to user-space for post-processing without in-kernel pre-aggregation. Things like fork/exec/exit events, for instance. 

You never needed eBPF for that -- you could have coded that in perf or Ftrace all along (and I did create such tools). My perf tools _needed_ eBPF for in-kernel summaries, so that high frequency events could be traced in a sane way.

I'm not complaining about execsnoop/etc being BPF since they are low frequency and don't need to be summarized. But to write a high frequency event tool that uses BPF but doesn't summarize is missing the point.

> There is certainly benefit to being able to do in-kernel aggregations, but there are cases where user-space really needs every single event. We've even switched from kernel aggregated stack traces for profiling into just emitting every stack trace straight to user-space (e.g., check bpf_get_stack() helper, which can also return build ID of executable, if requested).
>

Sure, but that's why I brought up perf...

> > And there's another reason: how does it compare to "perf sched record", which already has an optimized binary output file and different tools to post-process? And, there's also "perf record -e sched:*; perf script" which prints a format string that already has similar arguments to what you're printing.
> 
> Haven't used perf to record scheduling events, but can it do it as intelligently as this schedsnoop tool for one particular task, categorizing what type of event occurred (e.g., pre-empted vs waiting on lock, etc)? I'm sure you can post-process perf recording to extract this, but it will probably also record extra data that is simply filtered by this tool.

perf has -p options (e.g., on sched timehist) and you can see see prev_state in the sched_switch events; e.g., pre-empted is "R". Oh, and I forgot to mention Ftrace's custom preemption tracer.

But it's pretty rare that I use these tools (another reason I'd hesitate about adding yet another).

> 
> > If you think this belongs in BCC, I'd:
> > ```
> > 1. Publish it on your own repo first.
> > 
> > 2. Publish documentation (blog posts) showing real world case studies where it, and only it, can solve the problem.
> > ```
> 
> This one would be really nice, I agree.
> 

Given the actual use cases, we can see the target problems it's solving. Then they should really think about whether those same problems can be solved with a summary tool or not. And given it's high frequency, I'd really want a summary if at all possible.

> > ```
> > 3. Show overhead measurements in production environments.
> > ```
> 
> This will very much depend on the workload itself, wouldn't it? So sure they can show some numbers, but I'm not sure how easy would it be to qualify it generically.

I often use stress tests to figure out the worst case scenario. And if that worst case isn't high, then the overhead should be even less in production.

E.g., drive a high context switch & syscall workload to CPU saturation (0% idle), then find out the application context switch rate with and without the tool running. Since there's no CPU headroom, the overhead of the tool slows down the app, which can be measured, and expressed as a ratio. "Worst case, this tool will slow the target application by 15% (or whatever)."

This tool traces both context switches and syscalls: I would not want to run it on our busy 64-CPU prod systems for more than 1 second. (Although that may not be possible: if request latency increased high enough it will fail over in <1 second, so I wouldn't even get a full second of output!)

anakryiko:
> > > It's a bad idea for two reasons:
> > > A) Overhead. I can imagine my Netflix colleagues running this on our big systems and causing an outage (if it "only" eats one CPU, that'll be enough to trigger fail over), then asking me to help.
> > 
> > 
> > There is certainly an overhead, but I'm not sure that's good enough reason to disqualify the tool? Any tool can be misused. E.g., trace.py on malloc/free or even these scheduling events, and you are pretty surely will overwhelm the system.
> 
> In both cases you are choosing to trace something with high frequency ("trace.py something_silly"). But in this case you are just picking a PID -- it would be the first and only tool that can shoot yourself in the foot like this.
> 

Just like `runqslower.py 1`? To me it's more about whether you know which tool are you picking and what do you want to achieve with it? Tools are just tools, user needs to know how to use them regardless.

> > We at Facebook actually have a somewhat similar tool, used by many users in production, that records similar events for all threads within the process for some short amount of time, and then visualizes it in a browser in a nice UI, for more detailed inspection. It's certainly quite heavyweight, but if used for short periods of time, it is immensely useful to understand the behavior (especially blocking pattern) of an application.
> 
> Sounds like LTTng Trace Compass or Kernel Shark.

Oh cool, didn't know about Kernel Shark. It does look similar :)

> 
> > > B) If you've followed observability, the _whole_ point of using eBPF was in-kernel summaries so we didn't have to do per-event output. http://www.brendangregg.com/blog/2015-05-15/ebpf-one-small-step.html
> > 
> > 
> > There are a bunch of applications at Facebook that do 24/7 monitoring and emit events to user-space for post-processing without in-kernel pre-aggregation. Things like fork/exec/exit events, for instance.
> 
> You never needed eBPF for that -- you could have coded that in perf or Ftrace all along (and I did create such tools). My perf tools _needed_ eBPF for in-kernel summaries, so that high frequency events could be traced in a sane way.

It's obviously over-simplification. We also grab a bunch of extra data with custom BPF logic, that perf/ftrace just can't provide. So we still need eBPF in practice.

> 
> I'm not complaining about execsnoop/etc being BPF since they are low frequency and don't need to be summarized. But to write a high frequency event tool that uses BPF but doesn't summarize is missing the point.

Maybe for bcc tools, but not in general. BPF is used for far more than just summarizing data in kernel.

> > 
> > This will very much depend on the workload itself, wouldn't it? So sure they can show some numbers, but I'm not sure how easy would it be to qualify it generically.
> 
> I often use stress tests to figure out the worst case scenario. And if that worst case isn't high, then the overhead should be even less in production.
> 
> E.g., drive a high context switch & syscall workload to CPU saturation (0% idle), then find out the application context switch rate with and without the tool running. Since there's no CPU headroom, the overhead of the tool slows down the app, which can be measured, and expressed as a ratio. "Worst case, this tool will slow the target application by 15% (or whatever)."
> 

Ok, fair point, I think it would be nice to see those numbers.

> This tool traces both context switches and syscalls: I would not want to run it on our busy 64-CPU prod systems for more than 1 second. (Although that may not be possible: if request latency increased high enough it will fail over in <1 second, so I wouldn't even get a full second of output!)

This tool traces single thread, so it's not all that relevant how many CPUs are there. Or are you worried about just tracing scheduling events with (almost) no-op BPF program overall? I got impressions that BPF program call overhead is quite low to actually do this without too much impact.

qu10wenhao:
Hi @anakryiko & @brendangregg 

we have pushed a new commit that change schedsnoop a lot. You can also check my own repo [task_detector](https://github.com/qu10wenhao/task_detector).

First of all, schedsnoop could now record the preemption and do some statistics at the kernel side. It will output a preemption report when schedsnoop exits (use CTRL+C to stop it or the target thread exits). This report will show the top 10 threads that have preempted our target during tracing, sorted by the average preemption time. Output information includes the preemption count, the average time and the longest time. (More information could be available if needed)

> > > Does this really trace sched events and print output for each event?
> > 
> > Yes, but only for events relevant to requested thread.
> 
>  I saw that, but that still means a ton of output.

This was a serious problem. Our target used to be debugging instead of 24/7 tracing, so we haven't run this tool for vert long time and get a ton of output. Now, we won't print each events in default (enable it with an extra -l option). 

> > > This will very much depend on the workload itself, wouldn't it? So sure they can show some numbers, but I'm not sure how easy would it be to qualify it generically.
> >
> > I often use stress tests to figure out the worst case scenario. And if that worst case isn't high, then the overhead should be even less in production.
> >
> > E.g., drive a high context switch & syscall workload to CPU saturation (0% idle), then find out the application context switch rate with and without the tool running. Since there's no CPU headroom, the overhead of the tool slows down the app, which can be measured, and expressed as a ratio. "Worst case, this tool will slow the target application by 15% (or whatever)."
> 
> Ok, fair point, I think it would be nice to see those numbers.

For this part, we used schbench to test the performance of schedsnoop, which we focus on the 99th wakeup latency and show how schedsnoop will affect it.  We run schbench with multiple threads (more than #cpu) and 0 sleep time, so that CPU has 0% idle and switch/migration events happen frequently. The results are listed below:
- default, no SYSCALL : 0.2%
- default, include SYSCALL: 4.6%
- print events to file, no SYSCALL : 9.9%
- print events to file, include SYSCALL : 12.0%
- print events to console, no SYSCALL : 23.5%
- print events to console, include SYSCALL : 24%

Here is the conclusion. In default case, which we don't print the events, schedsnoop has very little affect to the wakeup latency and perform better than perf (use perf to record all schedule events will increase the latency by 3%). With events output, it is acceptable if we print the events to the file (add "> events.out" at the end of command), which it increases 10% of the latency. Print events to the console is the worst case, which increase the wakeup latency by more than 20%. 

I'm not so familiar with performance part but, to the best of my understanding, I think 5% impact on wakeup latency is quite OK. So we can first use default mode to get a preemption report and then print the events in a short time interval to see detail information. 

If you focus on other performance metrics, please let me know and I will do other tests. It will be nice if you could suggest more tool for performance testing (just like schbench). :)

qu10wenhao:
What's more, we've found a logical endless loop in schedsnoop while tracing the SYSCALL and outputting the events. Currently, we handle the events in user-side with "perf_buffer__poll". This function will call SYSCALL "epoll_wait". 

When schedsnoop itself preempts the target thread, this "epoll_wait" will be recorded and an SYSCALL event will be emitted to user-side schedsnoop for handle. After that, schedsnoop will call "perf_buffer__poll" again to wait next event and this is will call another SYSCALL "epoll_wait", which will be recorded again. So this is an endless loop and schedsnoop will continuously print plenty of "epoll_wait" events. This endless loop won't stop until schedsnoop doesn't preempt the target thread.

It is same for SYSCALL "write" when we try to print the events to the console.

Currently, I avoid this endless loop by adding an extra judgement before the BPF program emit such event to user-side. These SYSCALL events ("epoll_wait" and "write" called by schedsnoop itself) won't be emitted to user-side for output, but they will still be record and showed in the final syscall report.

@anakryiko Do you have any suggestion about this scenario? 
Does the new shared ringbuffer in kernel 5.8 still using such SYSCALL at the backend?

anakryiko:
> @anakryiko Do you have any suggestion about this scenario?
> Does the new shared ringbuffer in kernel 5.8 still using such SYSCALL at the backend?

epoll_wait is used for both, so no, BPF ringbuf doesn't change anything. I don't know what's the ideal solution. You can try to filter out context switches to/from your tool's process by filtering based on PID (but then you might lose some transitions because of X -> T -> Y -> X, where X is process you are tracing, T is your tool's process, Y is some other process). 

Or you might want to minimize context switched by not using epoll_wait, and instead sleep for short period of time, wake up and then use perf_buffer__consume() or ring_buffer__consumer(), which don't do any syscalls, to consume available data. Consider that.


@brendangregg, do these changes make sense to you? If you still think it's a bad fit for libbpf-tools, I'd rather avoid doing another round of code reviewing.

qu10wenhao:
Hi @brendangregg & @anakryiko 

It's been one month since we last chat. Hope you have time to take a short view of our newest progress.

As we mentioned before, we are willing to trace not only one single thread but also the whole process. Now, we have enable schedsnoop to trace multiple threads with a specific process id (aka. TGID). Schedsnoop is able to trace on maximal 64 threads at the same time and for each target threads, it will output a preemption report respectively. 

the limitation of the tracing thread number is due to the CPU mask. Currently, we use a u64 param as a mask for each CPU to indicate which threads is working on this CPU. We can enlarge this limitation by using a bigger mask (such as using two u64 params could allow us to trace 128 threads simultaneously), but it requires recompilation. Till now, we have not found a good solution to break this limitation.

Hope you are still interest in this tool and please provide more advises for us to improve it. THX!

anakryiko:
@qu10wenhao, thanks for working and updating this. I'll hold off on code review until we figure out if this is going to get accepted into BCC tools or not. Hopefully your changes made it appealing enough to @brendangregg to consider this as an addition to a set of tools.

PR #2984, opened 2020-06-21
libbpf-tools: remove unnecessary header include from syscount.bpf.c

anakryiko:
It causes build failure on my system due to trying to include GCC-specific
header. It doesn't seem to be necessary, though, so remove it.

Signed-off-by: Andrii Nakryiko <andriin@fb.com>


yonghong-song:
[buildbot, test this please]

PR #2986, opened 2020-06-22
bpf: use bpf_probe_read in implicitly generated kernel mem read

yonghong-song:
Currently, bcc rewriter uses bpf_probe_read_kernel() for implicitly specified
kernel memory read. This is not totally correct. Some user memory in kernel
data structure may be accessed. bpf_probe_read_kernel() may fail with
later kernels. Let us revert back to old bpf_probe_read().


yonghong-song:
cc @davemarchevsky

PR #2987, opened 2020-06-23
[Documentation] Fix kernel version for ringbuf and add ringbuf to maps table

willfindlay:
kernel-versions.md lists ringbufs as being supported in 5.7+. They are in fact supported in **5.8+**. This pull request changes it to 5.8 and adds an entry for the ringbuf map itself.


yonghong-song:
[buildbot, test this please]

PR #2988, opened 2020-06-23
tools/profile: fix suggestion about when to increase stack-storage-size

boat0:
When we do stack traces via stackmaps, hash collisions (-EEXIST) may indicate
that the map size is too small. Not -ENOMEM.

### DIFF
--- a/tools/profile.py
+++ b/tools/profile.py
@@ -305,17 +305,18 @@ def aksym(addr):
 
 # output stacks
 missing_stacks = 0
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
This looks correct since we removed BPF_F_REUSE_STACKID from bpf_get_stack_id() helper sometimes back.

PR #2990, opened 2020-06-24
Fix KFUNC_PROBE calls in vfs_stat.py by adding a return value

willfindlay:
After #2978, `vfsstat.py` causes verifier to complain due to missing return values from `KFUNC_PROBE`s. This pull request adds explicit return values to these BPF programs.


yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks. for the new kernel kentry/kexit return value has to be 0.

PR #2992, opened 2020-06-24
libbpf-tools: add CO-RE biolatency

ethercflow:
- implement [biolatency](https://github.com/iovisor/bcc/blob/master/tools/biolatency.py) with `BTF_PROG_TYPE_TRACING` instead of `kprobes`
- extract string filter to `bits.bpf.h`
- modify makefile to add *bpf.h as *.bpf.o's dep for re-build detection

### DIFF
--- a/libbpf-tools/Makefile
+++ b/libbpf-tools/Makefile
@@ -10,6 +10,7 @@ CFLAGS := -g -O2 -Wall
 ARCH := $(shell uname -m | sed 's/x86_64/x86/')
 
 APPS = \
+	biolatency \
 	bitesize \
 	cpudist \
 	drsnoop \
--- a/libbpf-tools/biolatency.bpf.c
+++ b/libbpf-tools/biolatency.bpf.c
@@ -1,0 +1,121 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2020 Wenbo Zhang
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_tracing.h>
+#include "biolatency.h"
+#include "bits.bpf.h"
+
+#define MAX_ENTRIES 10240
+
+const volatile char targ_disk[DISK_NAME_LEN] = {};
+const volatile bool targ_per_disk = false;
+const volatile bool targ_per_flag = false;
+const volatile bool targ_queued = false;
+const volatile bool targ_ms = false;
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_ENTRIES);
+	__type(key, struct request *);
+	__type(value, u64);
+	__uint(map_flags, BPF_F_NO_PREALLOC);
+} start SEC(".maps");
+
+static struct hist initial_hist;
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_ENTRIES);
+	__type(key, struct hist_key);
+	__type(value, struct hist);
+	__uint(map_flags, BPF_F_NO_PREALLOC);
+} hists SEC(".maps");
+
+static __always_inline bool disk_filtered(const char *disk)
+{
+	int i;
+
+	for (i = 0; targ_disk[i] != '\0' && i < DISK_NAME_LEN; i++) {
+		if (disk[i] != targ_disk[i])
+			return false;
+	}
+	return true;
+}
+
+static __always_inline
+int trace_rq_start(struct request *rq)
+{
+	u64 ts = bpf_ktime_get_ns();
+	char disk[DISK_NAME_LEN];
+
+	bpf_probe_read_kernel_str(&disk, sizeof(disk), rq->rq_disk->disk_name);
+	if (!disk_filtered(disk))
+		return 0;
+
+	bpf_map_update_elem(&start, &rq, &ts, 0);
+	return 0;
+}
+
+SEC("tp_btf/block_rq_insert")
+int BPF_PROG(tp_btf__block_rq_insert, struct request_queue *q,
+	     struct request *rq)
+{
+	return trace_rq_start(rq);
+}
+
+SEC("tp_btf/block_rq_issue")
+int BPF_PROG(tp_btf__block_rq_issue, struct request_queue *q,
+	     struct request *rq)
+{
+	if (targ_queued && BPF_CORE_READ(q, elevator))
+		return 0;
+	return trace_rq_start(rq);
+}
+
+SEC("tp_btf/block_rq_complete")
+int BPF_PROG(tp_btf__block_rq_complete, struct request *rq, int error,
+	     unsigned int nr_bytes)
+{
+	u64 slot, *tsp, ts = bpf_ktime_get_ns();
+	struct hist_key hkey = {};
+	struct hist *histp;
+	s64 delta;
+
+	tsp = bpf_map_lookup_elem(&start, &rq);
+	if (!tsp)
+		return 0;
+	delta = (s64)(ts - *tsp);
+	if (delta < 0)
+		goto cleanup;
+
+	if (targ_per_disk)
+		bpf_probe_read_kernel_str(&hkey.disk, sizeof(hkey.disk),
+					rq->rq_disk->disk_name);
+	if (targ_per_flag)
+		hkey.cmd_flags = rq->cmd_flags;
+
+	histp = bpf_map_lookup_elem(&hists, &hkey);
+	if (!histp) {
+		bpf_map_update_elem(&hists, &hkey, &initial_hist, 0);
+		histp = bpf_map_lookup_elem(&hists, &hkey);
+		if (!histp)
+			goto cleanup;
+	}
+
+	if (targ_ms)
+		delta /= 1000000;
+	else
+		delta /= 1000;
+	slot = log2l(delta);
+	if (slot >= MAX_SLOTS)
+		slot = MAX_SLOTS - 1;
+	__sync_fetch_and_add(&histp->slots[slot], 1);
+
+cleanup:
+	bpf_map_delete_elem(&start, &rq);
+	return 0;
+}
+
+char LICENSE[] SEC("license") = "GPL";
--- a/libbpf-tools/biolatency.c
+++ b/libbpf-tools/biolatency.c
@@ -1,0 +1,319 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+// Copyright (c) 2020 Wenbo Zhang
+//
+// Based on biolatency(8) from BCC by Brendan Gregg.
+// 15-Jun-2020   Wenbo Zhang   Created this.
+#include <argp.h>
+#include <signal.h>
+#include <stdio.h>
+#include <unistd.h>
+#include <time.h>
+#include <bpf/libbpf.h>
+#include <sys/resource.h>
+#include <bpf/bpf.h>
+#include "blk_types.h"
+#include "biolatency.h"
+#include "biolatency.skel.h"
+#include "trace_helpers.h"
+
+#define ARRAY_SIZE(x) (sizeof(x) / sizeof(*(x)))
+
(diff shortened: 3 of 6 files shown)


ethercflow:
Hi @anakryiko  I've converted `biolatency` PTAL when you're free. Thanks.

libbpf-tools/blk_types.h
```diff
@@ -0,0 +1,126 @@
...
+ *   - if the least significant bit is not set transfers are FROM the device
+ *
+ * If a operation does not transfer data the least significant bit has no
+ * meaning.
+ */
+#define REQ_OP_BITS	8
+#define REQ_OP_MASK	((1 << REQ_OP_BITS) - 1)
+#define REQ_FLAG_BITS	24
+
+enum req_opf {
```
anakryiko, line 24: this is already part of vmlinux.h, why adding it here?

libbpf-tools/blk_types.h
```diff
@@ -0,0 +1,126 @@
...
+	REQ_OP_SCSI_IN		= 32,
+	REQ_OP_SCSI_OUT		= 33,
+	/* Driver private requests */
+	REQ_OP_DRV_IN		= 34,
+	REQ_OP_DRV_OUT		= 35,
+
+	REQ_OP_LAST,
+};
+
+enum req_flag_bits {
```
anakryiko, line 60: same, it's in vmlinux.h

libbpf-tools/bits.bpf.h
```diff
@@ -25,4 +25,16 @@ static __always_inline u64 log2l(u64 v)
 		return log2(v);
 }
 
+static __always_inline
+bool str_filtered(const char *target, const volatile char *cond, int cond_len)
```
anakryiko, line 29: this is not bit manipulation thing, it's also not complicated and tricky at all... so I don't know, it definitely doesn't belong to bits.bpf.h, but creating another header just for this also seems like a waste

libbpf-tools/biolatency.c
```diff
@@ -0,0 +1,351 @@
...
+		return 0;
+	return vfprintf(stderr, format, args);
+}
+
+static void sig_handler(int sig)
+{
+	exiting = true;
+}
+
+static void print_cmd_flags(int cmd_flags)
```
anakryiko, line 136: have you consider doing something like this instead:
```
static struct { int bit; const char * str; } flags[] = {
  { __REQ_NOWAIT, "NoWait-" },
  { __REQ_BACKGROUND, "Background-"},
  ...
};
for (i = 0; i < ARRAY_SIZE(flags); i++) {
    if (cmd_flags & flags[i].bit)
        printf(flags[i].str);
}
```

For REQ_OP_MASK similar approach:
```
static const char *op[] = {
  [REQ_OP_READ] = "Read",
  [REQ_OP_WRITE] = "Write",
  ...
};

if ((cmd_flags & REQ_OP_MASK) < ARRAY_SIZE(op)) {
    printf("Unknown");
else
    printf(op[cmd_flags & REQ_OP_MASK]);
}
```

It will be a bit nicer, IMO.

libbpf-tools/biolatency.c
```diff
@@ -0,0 +1,351 @@
...
+	}
+	obj->links.tp_btf__block_rq_complete =
+		bpf_program__attach(obj->progs.tp_btf__block_rq_complete);
+	err = libbpf_get_error(obj->links.tp_btf__block_rq_complete);
+	if (err) {
+		fprintf(stderr, "failed to attach: %s\n", strerror(-err));
+		goto cleanup;
+	}
+
+	fd = bpf_map__fd(obj->maps.hists);
```
anakryiko, line 321: you use it in one place below, I'd just do it there without extra `fd` variable. It's extremely cheap function.

libbpf-tools/Makefile
```diff
@@ -63,7 +64,7 @@ $(OUTPUT)/%.skel.h: $(OUTPUT)/%.bpf.o | $(OUTPUT)
 	$(call msg,GEN-SKEL,$@)
 	$(Q)$(BPFTOOL) gen skeleton $< > $@
 
-$(OUTPUT)/%.bpf.o: %.bpf.c $(LIBBPF_OBJ) $(wildcard %.h) vmlinux.h | $(OUTPUT)
+$(OUTPUT)/%.bpf.o: %.bpf.c $(LIBBPF_OBJ) $(wildcard %.h) $(wildcard *.bpf.h) vmlinux.h | $(OUTPUT)
```
anakryiko, line 67: nit: it's getting a bit long, maybe split into two lines?

libbpf-tools/biolatency.bpf.c
```diff
@@ -0,0 +1,109 @@
...
+	     struct request *rq)
+{
+	return trace_rq_start(rq);
+}
+
+SEC("tp_btf/block_rq_issue")
+int BPF_PROG(tp_btf__block_rq_issue, struct request_queue *q,
+	     struct request *rq)
+{
+	if (targ_queued && BPF_CORE_READ(q, elevator))
```
anakryiko, line 61: it's a tp_btf program, so you don't have to use BPF_CORE_READ macro, just `q->elevator` will work and will be faster.

libbpf-tools/biolatency.bpf.c
```diff
@@ -0,0 +1,109 @@
...
+	__uint(map_flags, BPF_F_NO_PREALLOC);
+} hists SEC(".maps");
+
+static __always_inline
+int trace_rq_start(struct request *rq)
+{
+	u64 ts = bpf_ktime_get_ns();
+	char disk[DISK_NAME_LEN];
+
+	BPF_CORE_READ_STR_INTO(&disk, rq, rq_disk, disk_name);
```
anakryiko, line 42: same as below, no need for BPF_CORE_READ_STR_INTO, you can just do:
```
bpf_probe_read_kernel_str(&disk, sizeof(disk), rq->rq_disk->disk_name);
```

libbpf-tools/biolatency.bpf.c
```diff
@@ -0,0 +1,109 @@
...
+	if (!tsp)
+		return 0;
+	delta = (s64)(ts - *tsp);
+	if (delta < 0)
+		goto cleanup;
+
+	if (targ_per_disk)
+		BPF_CORE_READ_STR_INTO(&hkey.disk, rq, rq_disk, disk_name);
+	if (targ_per_flag)
+		hkey.cmd_flags = rq->cmd_flags;
```
anakryiko, line 85: see, here you are accessing fields directly

libbpf-tools/blk_types.h
ethercflow: Oh, I need them in user space (biolatency.c), so I defined these.

libbpf-tools/bits.bpf.h
ethercflow: Oh, I think I misunderstood the `bit-manipulation`, your meaning is `bit operation`?  I misunderstood it as light(bit) helpers. I’ll change it back. Thanks.

libbpf-tools/biolatency.c
ethercflow: Thanks, it looks much nicer.

libbpf-tools/biolatency.bpf.c
ethercflow: Yes, @anakryiko I used `if (targ_queued && q->elevator)` first, but I found I always get nothing output (with none I/O scheduler: `echo none > /sys/block/sdc/queue/scheduler`, which set `q->elevator = NULL;`), but with `BPF_CORE_READ` is ok.

The kernel I tested is on `5.7.0-rc7+` and `5.8.0-rc1+`, both have this problem. I also found if the struct's member's type is pointer,  `if(struct->pointer_member)` will not work. So is this a bug or am I missed sth?

The workload source is 

```go
package main

import (
        "flag"
        "log"
        "os"
        "time"
)

func main() {
        name := flag.String("file", "", "filename to use")
        size := flag.Int("size", 1024*10, "chunk size")
        sync := flag.Int("sync", 1, "sync file after that many writes, <= 0 to disable")
        sleep := flag.Duration("sleep", time.Second, "time to sleep between writes")
        count := flag.Int("count", 0, "number of iterations, <=0 for infinity")

        flag.Parse()

        if *name == "" {
                flag.PrintDefaults()
        }

        file, err := os.OpenFile(*name, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0666)
        if err != nil {
                log.Fatalf("Error opening file %q: %s", *name, err)
        }

        buf := make([]byte, *size)

        i := 0
        for {
                i++

               started := time.Now()

                verb := "Written"

                if _, err := file.Write(buf); err != nil {
                        log.Fatalf("Error writing: %s", err)
                }     
                if *sync > 0 && i%*sync == 0 {
                        if err := file.Sync(); err != nil {
                                log.Fatalf("Error syncing: %s", err)
                        }

                        verb = "Synced"
                }

                elapsed := time.Since(started)

                log.Printf("%s in %d μs | %v", verb, int(elapsed.Seconds() * 1e6), elapsed)

                if *count > 0 && i >= *count {
                        break
                }

                time.Sleep(*sleep)
        }
}
```

Then run:

`./biolatency   -d sdc  -QDF`
`go run ./paranoid.go -file=/dev/sdc -size=64 -count=10.   (write disk without fs)`


WIth 

`if (targ_queued && BPF_CORE_READ(q, elevator))`

I can always get

```
disk = sdc	flags = Sync-Write
        16 -> 31         : 8        |                                        |
        32 -> 63         : 2        |****************************************|
```

but with

`if(struct->pointer_member)`

Output nothing.

anakryiko: It does sound like a bug. Can you please create a smallest repro and submit it to bpf@vger.kernel.org? Please also attach a BPF assembly dump ( with `llvm-objdump -d`). Thanks a lot for your help!

ethercflow: Sure, you're welcome.

ethercflow: @anakryiko Has reported, the subject is `tp_btf: if (!struct->pointer_member) always actually false although pointer_member == NULL`

libbpf-tools/bits.bpf.h
ethercflow: Fixed.

libbpf-tools/biolatency.c
ethercflow: Fixed

libbpf-tools/biolatency.c
ethercflow: Fixed.

libbpf-tools/Makefile
ethercflow: Fixed

libbpf-tools/biolatency.bpf.c
ethercflow: Fixed.

ethercflow:
Hi @anakryiko! Thanks for the review, pushed fixes for all your comments, but left two places:
1. I keep `BPF_CORE_READ(q, elevator)` instead of `if (targ_queued && q->elevator)`, I'll keep this in mind, when related kernel versions are all fixed, I'll change to it, is this OK?
2. I keep redefine `REQ_OP_MASK` and so on to use it in user space prog, I tried include `vmlinux.h` in user space, but I got a lot of redefined errors, so `vmlinux.h` is designed to be used only for bpf programs, right?  Any good suggestions for parsing kernel definitions in user mode?

anakryiko:
Yeah, vmlinux.h is not supposed to be used from user-space, so that's not a viable option. And given REQ_OP_MASK is defined in kernel internal header, you have no other choice as to copy/paste its definitions.

BPF_CORE_READ(q, elevator) is better for kernel compatibility, so I'd leave it as is for a long time as well.

So I approved your changes, thanks!

ethercflow:
@yonghong-song  PTAL, thanks. :)

ethercflow:
Hi @yonghong-song , I think this may get merged, would you PTAL when you have time? Thank you. :)

yonghong-song:
sorry, forgot this one.

yonghong-song:
[buildbot, test this please]

PR #2998, opened 2020-07-02
valloc and pvalloc is deprecated in bionic on Android

b-ripper:
memleak -p PID will failed on Android
https://android.googlesource.com/platform/bionic/+/master/libc/bionic/malloc_common.cpp#196


yonghong-song:
[buildbot, test this please]

PR #2999, opened 2020-07-03
Adjust the order of linux kernel header include paths

lcp:
The current order of include paths could cause the following error:

In file included from /virtual/main.c:3:
In file included from include/linux/sched.h:13:
In file included from include/linux/pid.h:4:
In file included from include/linux/rculist.h:10:
In file included from include/linux/rcupdate.h:38:
In file included from include/linux/spinlock.h:50:
In file included from include/linux/preempt.h:80:
In file included from /lib/modules/4.12.14-197.45-default/build/arch/s390/include/generated/asm/preempt.h:1:
include/asm-generic/preempt.h:10:42: error: no member named 'preempt_count' in 'struct thread_info'
        return READ_ONCE(current_thread_info()->preempt_count);
               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~
include/linux/compiler.h:349:34: note: expanded from macro 'READ_ONCE'
                     ~~~~~~~~~~~~^~~~~
include/linux/compiler.h:342:17: note: expanded from macro '__READ_ONCE'
        union { typeof(x) __val; char __c[1]; } __u;
\
...
7 errors generated.

It's supposed to load "asm/preempt.h" from "source/"(*) but accidentally
loaded the one from "build/", so the error showed.
(x86_64 didn't suffer this error because it didn't install the extra
 "asm/preempt.h" to "build/".)

For the distros, e.g. SUSE/openSUSE and debian,  with separate "build/"
and "source/", all those "generated/" paths only exist in "build/". To
avoid the potential compilation issue, this commit moves down all those
"generated/" paths in "build/" and adds them conditionally.

(*) /lib/modules/4.12.14-197.45-default/source/arch/s390/include/

### DIFF
--- a/src/cc/frontends/clang/kbuild_helper.cc
+++ b/src/cc/frontends/clang/kbuild_helper.cc
@@ -74,27 +74,47 @@ int KBuildHelper::get_flags(const char *uname_machine, vector<string> *cflags) {
   cflags->push_back("-isystem");
   cflags->push_back("/virtual/lib/clang/include");
 
-  // some module build directories split headers between source/ and build/
+  // The include order from kernel top Makefile:
+  //
+  // # Use USERINCLUDE when you must reference the UAPI directories only.
+  // USERINCLUDE    := \
+  //                 -I$(srctree)/arch/$(SRCARCH)/include/uapi \
+  //                 -I$(objtree)/arch/$(SRCARCH)/include/generated/uapi \
+  //                 -I$(srctree)/include/uapi \
+  //                 -I$(objtree)/include/generated/uapi \
+  //                 -include $(srctree)/include/linux/kconfig.h
+  //
+  // # Use LINUXINCLUDE when you must reference the include/ directory.
+  // # Needed to be compatible with the O= option
+  // LINUXINCLUDE    := \
+  //                 -I$(srctree)/arch/$(SRCARCH)/include \
+  //                 -I$(objtree)/arch/$(SRCARCH)/include/generated \
+  //                 $(if $(building_out_of_srctree),-I$(srctree)/include) \
+  //                 -I$(objtree)/include \
+  //                 $(USERINCLUDE)
+  //
+  // Some distros such as openSUSE/SUSE and Debian splits the headers between
+  // source/ and build/. In this case, just $(srctree) is source/ and
+  // $(objtree) is build/.
   if (has_source_dir_) {
-    cflags->push_back("-I" + kdir_ + "/build/arch/"+arch+"/include");
-    cflags->push_back("-I" + kdir_ + "/build/arch/"+arch+"/include/generated/uapi");
+    cflags->push_back("-Iarch/"+arch+"/include/");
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
The following is the include order from kernel top Makefile.
```
# Use USERINCLUDE when you must reference the UAPI directories only.
USERINCLUDE    := \
                -I$(srctree)/arch/$(SRCARCH)/include/uapi \
                -I$(objtree)/arch/$(SRCARCH)/include/generated/uapi \
                -I$(srctree)/include/uapi \
                -I$(objtree)/include/generated/uapi \
                -include $(srctree)/include/linux/kconfig.h

# Use LINUXINCLUDE when you must reference the include/ directory.
# Needed to be compatible with the O= option
LINUXINCLUDE    := \
                -I$(srctree)/arch/$(SRCARCH)/include \
                -I$(objtree)/arch/$(SRCARCH)/include/generated \
                $(if $(building_out_of_srctree),-I$(srctree)/include) \
                -I$(objtree)/include \
                $(USERINCLUDE)
```

Do you think we could organize the include orders following the above convention to fix the problem?
I mention this just want to be careful and try to solve the issue in a way similar to kernel module.

lcp:
It should work. For SUSE/openSUSE, we can do a simple conversion: $(srctree) => "source/", $(objtree) => "build/". Not sure whether this rule holds for debian though.

yonghong-song:
> It should work. For SUSE/openSUSE, we can do a simple conversion: $(srctree) => "source/", $(objtree) => "build/". Not sure whether this rule holds for debian though.

This is what I am afraid of. The makefile change may break on other systems. But I guess if you adhere to what kernel did, we have less chance it will break.

lcp:
> > It should work. For SUSE/openSUSE, we can do a simple conversion: $(srctree) => "source/", $(objtree) => "build/". Not sure whether this rule holds for debian though.
> 
> This is what I am afraid of. The makefile change may break on other systems. But I guess if you adhere to what kernel did, we have less chance it will break.

I've followed the include order in kernel top Makefile to update the build flags.

lcp:
BTW, I checked the debian kernel headers, and the installation is basically the same as openSUSE/SUSE.

The file lists:
linux-headers-4.19.0-9-amd64
https://packages.debian.org/buster/amd64/linux-headers-4.19.0-9-amd64/filelist

linux-headers-4.19.0-9-common
https://packages.debian.org/buster/all/linux-headers-4.19.0-9-common/filelist

linux-headers-4.19.0-9-amd64 links build/ to the linux-headers-4.19.0-9-amd64 directory and source/ to the linux-headers-4.19.0-9-common directory, and those generated directories are only in linux-headers-4.19.0-9-amd64, i.e. build/.

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

yonghong-song:
LGTM. Thanks!

PR #3005, opened 2020-07-06
memleak: Add workaround to alleviate misjudgments when free is missing

netedwardwu:
Profiling in memory part is hard to be accurate because of BPF infrastructure.
memleak keeps misjudging memory leak on the complicated environment which has
the action of free in hard/soft irq.

For example, in my misjudged case:

640 bytes in 10 allocations from stack
--
__kmalloc+0x178 [kernel]
__kmalloc+0x178 [kernel]
xhci_urb_enqueue+0x140 [kernel]
usb_hcd_submit_urb+0x5e0 [kernel]

This result looks like kernel doesn't free urb_priv. However, it's not true.
The reason for this leak is because xhci hw irq interrupts during the BPF program.
BPF program is not finished on that CPU, and xhci_irq() will call xhci_urb_free_priv()
before the end. But the kernel doesn't permit this isr to go into BPF program again.
Because BPF infrastructure(trace_call_bpf) denied this action.
So we miss this free action and cause memory leak misjudgment.

Side-effect:
- Increase overhead for each memory allocation.
- A higher chance to be interrupted at the allocation part causes ignore more allocations.

This workaround doesn't solve all misjudgments, the improvement in BPF infrastructure
is the only solution.

### DIFF
--- a/man/man8/memleak.8
+++ b/man/man8/memleak.8
@@ -3,8 +3,8 @@
 memleak \- Print a summary of outstanding allocations and their call stacks to detect memory leaks. Uses Linux eBPF/bcc.
 .SH SYNOPSIS
 .B memleak [-h] [-p PID] [-t] [-a] [-o OLDER] [-c COMMAND] [--combined-only]
-[-s SAMPLE_RATE] [-T TOP] [-z MIN_SIZE] [-Z MAX_SIZE] [-O OBJ] [INTERVAL]
-[COUNT]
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks for the patch! Looks good to me. Could you also add an example in memleak_example.txt file and also the map page?

netedwardwu:
Hi @yonghong-song ,

Thanks for your reminder.
I make this commitment more complete. Please check again.

yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks! LGTM.

PR #3008, opened 2020-07-07
Replace kprobe function "blk_account_io_completion"

owen-chenhn:
Starting from kernel version 5.8.0, the block layer function "blk_account_io_completion" is not available anymore as kprobe attach function. This makes errors occur when running some of the BCC tools and examples in kernel version >= 5.8.0. Therefore, we propose to use function "blk_account_io_done" instead in every kprobe attachment, because this function is compatible to previous kernel versions.  

This pull request changes "blk_account_io_completion" to "blk_account_io_done" at everywhere the former function occurs, including the doc files.

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -1602,7 +1602,7 @@ Example:
 b = BPF(text="""
 BPF_HISTOGRAM(dist);
 
-int kprobe__blk_account_io_completion(struct pt_regs *ctx, struct request *req)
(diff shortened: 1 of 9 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
Agree that `blk_account_io_done` is a better choice as `biolatency.py` and `biolatpcts.py` already used it.

PR #3011, opened 2020-07-09
Add .lazy_symbolize field to bcc_symbol_option

rhdxmr:
`bcc.libbcc.bcc_symbol_option` class defined at `libbcc.py` is different from the structure defined at `bcc_syms.h`

The former lacks of `.lazy_symbolize` field. So, the field is added to the class.

And also `perf_event_attr` defined at `perf.py` is different from the structure defined at `vmlinux.h`

So, missing fields are added to the class.


yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks for the fix!

PR #3012, opened 2020-07-09
Fix opensnoop and improve error reporting

realhezhe:


yonghong-song:
[buildbot, test this please]

src/cc/libbpf.c
```diff
@@ -588,8 +588,18 @@ int bcc_prog_load_xattr(struct bpf_load_program_attr *attr, int prog_len,
...
+      ret = libbpf_find_vmlinux_btf_id(attr->name + name_offset,
+                                       expected_attach_type);
+      if (ret == -EINVAL) {
+        fprintf(stderr, "bpf: vmlinux BTF is not found\n");
+        return ret;
+      } else if (ret < 0) {
+        fprintf(stderr, "bpf: %s is not found in vmlinux BTF\n",
+                attr->name + name_offset);
+        return ret;
+      }
```
yonghong-song, line 600: The above implementation is correct based on current libbpf_find_vmlinux_btf_id implementation.

tools/opensnoop.py
```diff
@@ -235,8 +237,12 @@
 # initialize BPF
 b = BPF(text=bpf_text)
 if not is_support_kfunc:
-    b.attach_kprobe(event="do_sys_open", fn_name="trace_entry")
-    b.attach_kretprobe(event="do_sys_open", fn_name="trace_return")
+    if parse_version(platform.uname().release.split('-')[0]) > parse_version('5.6.0'):
+        entry = "do_sys_openat2"
```
yonghong-song, line 241: This change is tricky. 

First, we cannot do kernel version comparison in tools since there are kernels backported or forward-ported with features and the version checking does not make sense to these kernels. In bcc, we check /proc/kallsyms for a particular global variable to indicate whether a particular feature exists or not.

Second, `do_sys_openat2` is not a safe alternative either. Depending on the compiler, `do_sys_openat2` may be inlined into `do_sys_open`. The safest way is to trace the syscall (__x64_sys_open, __x64_sys_openat, etc.) itself. You can check `execsnoop.py` for an example.

BTW, do_sys_open2 and do_sys_open has different signatures. So this change won't work for kfunc's.

realhezhe: Thanks for explanation.

For kprobe part, how about the following one?
https://github.com/iovisor/bcc/pull/3012/commits/5d9963e552a6888e2d39e770597d3cc6c580c7e1

For kfunc part,
1) I was trying to catch __x64_sys_openat but met the following error.

bpf_text_kfunc= """
KRETFUNC_PROBE(__x64_sys_openat, int dfd, const char __user *filename, int flags, int mode, int ret)
...

arg#0 type is not a struct
Unrecognized arg#0 type PTR
; KRETFUNC_PROBE(__x64_sys_openat, int dfd, const char __user *filename, int flags, int mode, int ret)
0: (bf) r6 = r1
; KRETFUNC_PROBE(__x64_sys_openat, int dfd, const char __user *filename, int flags, int mode, int ret)
1: (79) r1 = *(u64 *)(r6 +32)
func '__x64_sys_openat' doesn't have 5-th argument
invalid bpf_context access off=32 size=8
processed 2 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0

  File "/usr/share/bcc/tools/opensnoop", line 282, in <module>
    b = BPF(text=bpf_text, debug=4)
  File "/usr/lib/python3.8/site-packages/bcc/__init__.py", line 366, in __init__
    self._trace_autoload()
  File "/usr/lib/python3.8/site-packages/bcc/__init__.py", line 1239, in _trace_autoload
    self.attach_kretfunc(fn_name=func_name)
  File "/usr/lib/python3.8/site-packages/bcc/__init__.py", line 951, in attach_kretfunc
    fn = self.load_func(fn_name, BPF.TRACING)
  File "/usr/lib/python3.8/site-packages/bcc/__init__.py", line 405, in load_func
    raise Exception("Failed to load BPF program %s: %s" %
Exception: Failed to load BPF program b'kretfunc____x64_sys_openat': Permission denied



2) And then I tried to do_sys_openat2, but it seems it's in /proc/kallsyms but not in /sys/kernel/btf/vmlinux

$ sudo /usr/share/bcc/tools/opensnoop
traceback (most recent call last):
  File "/usr/share/bcc/tools/opensnoop", line 283, in <module>
    b = BPF(text=bpf_text, debug=4)
  File "/usr/lib/python3.8/site-packages/bcc/__init__.py", line 366, in __init__
    self._trace_autoload()
  File "/usr/lib/python3.8/site-packages/bcc/__init__.py", line 1239, in _trace_autoload
    self.attach_kretfunc(fn_name=func_name)
  File "/usr/lib/python3.8/site-packages/bcc/__init__.py", line 951, in attach_kretfunc
    fn = self.load_func(fn_name, BPF.TRACING)
  File "/usr/lib/python3.8/site-packages/bcc/__init__.py", line 405, in load_func
    raise Exception("Failed to load BPF program %s: %s" %
Exception: Failed to load BPF program b'kretfunc__do_sys_openat2': Invalid argument

$ strings /sys/kernel/btf/vmlinux |grep sys_open
__ia32_compat_sys_open
__ia32_compat_sys_open_by_handle_at
__ia32_compat_sys_openat
__ia32_sys_open
__ia32_sys_open_by_handle_at
__ia32_sys_open_tree
__ia32_sys_openat
__ia32_sys_openat2
__x64_sys_open
__x64_sys_open_by_handle_at
__x64_sys_open_tree
__x64_sys_openat
__x64_sys_openat2
do_sys_open

$ cat /proc/kallsyms |grep sys_open
0000000000000000 t do_sys_openat2
0000000000000000 T __ia32_sys_openat2
0000000000000000 T __x64_sys_openat2
0000000000000000 T __x64_sys_open
0000000000000000 T __ia32_compat_sys_open
0000000000000000 T __x64_sys_openat
0000000000000000 T __ia32_sys_open
0000000000000000 T __ia32_compat_sys_openat
0000000000000000 T __ia32_sys_openat
0000000000000000 T do_sys_open
0000000000000000 T __ia32_sys_open_tree
0000000000000000 T __x64_sys_open_tree
0000000000000000 T __x64_sys_open_by_handle_at
0000000000000000 T __ia32_sys_open_by_handle_at
0000000000000000 T __ia32_compat_sys_open_by_handle_at
0000000000000000 t proc_sys_open
0000000000000000 d sys_open_test
0000000000000000 d _eil_addr___ia32_compat_sys_openat
0000000000000000 d _eil_addr___ia32_compat_sys_open
0000000000000000 d _eil_addr___ia32_sys_openat2
0000000000000000 d _eil_addr___x64_sys_openat2
0000000000000000 d _eil_addr___ia32_sys_openat
0000000000000000 d _eil_addr___x64_sys_openat
0000000000000000 d _eil_addr___ia32_sys_open
0000000000000000 d _eil_addr___x64_sys_open
0000000000000000 d _eil_addr___ia32_sys_open_tree
0000000000000000 d _eil_addr___x64_sys_open_tree
0000000000000000 d _eil_addr___ia32_compat_sys_open_by_handle_at
0000000000000000 d _eil_addr___ia32_sys_open_by_handle_at
0000000000000000 d _eil_addr___x64_sys_open_by_handle_at

yonghong-song:
There is an issue with your understanding w.r.t. __x64_sys_open function signature in 5.6. 
For __x64_sys_open and other __x64_sys_<> functions, the argument is a single pt_regs and you need futher
dereference of the first parameter to get additional parameters.

Look at assembly code,
```
ffffffff81264610 <__x64_sys_openat>:
ffffffff81264610:       e8 2b a6 de ff          callq  ffffffff8104ec40 <__fentry__>
ffffffff81264615:       8b 57 60                mov    0x60(%rdi),%edx
ffffffff81264618:       0f b7 4f 38             movzwl 0x38(%rdi),%ecx
ffffffff8126461c:       48 8b 77 68             mov    0x68(%rdi),%rsi
ffffffff81264620:       48 8b 7f 70             mov    0x70(%rdi),%rdi
ffffffff81264624:       80 ce 80                or     $0x80,%dh
ffffffff81264627:       e9 24 ff ff ff          jmpq   ffffffff81264550 <do_sys_open>
```

Trying to do the following change to execsnoop.py and it will have rewriter output to
show what bpf program is expecting.
```
bash-4.4$ git diff
diff --git a/tools/execsnoop.py b/tools/execsnoop.py
index 53052d39..b16e21a0 100755
--- a/tools/execsnoop.py
+++ b/tools/execsnoop.py
@@ -224,7 +224,7 @@ if args.ebpf:
     exit()
 
 # initialize BPF
-b = BPF(text=bpf_text)
+b = BPF(text=bpf_text, debug=4)
 execve_fnname = b.get_syscall_fnname("execve")
 b.attach_kprobe(event=execve_fnname, fn_name="syscall__execve")
 b.attach_kretprobe(event=execve_fnname, fn_name="do_ret_sys_execve")
-bash-4.4$
```

yonghong-song:
```
bpf_text_kfunc= """
KRETFUNC_PROBE(__x64_sys_openat, int dfd, const char __user *filename, int flags, int mode, int ret)
...

arg#0 type is not a struct
Unrecognized arg#0 type PTR
; KRETFUNC_PROBE(__x64_sys_openat, int dfd, const char __user *filename, int flags, int mode, int ret)
0: (bf) r6 = r1
; KRETFUNC_PROBE(__x64_sys_openat, int dfd, const char __user *filename, int flags, int mode, int ret)
1: (79) r1 = *(u64 *)(r6 +32)
func '__x64_sys_openat' doesn't have 5-th argument
invalid bpf_context access off=32 size=8
processed 2 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
```

The above error is expected, __x64_sys_openat only has one parameter "struct pt_regs *regs", so the correct probe definition should be
```
KRETFUNC_PROBE(__x64_sys_openat, struct pt_regs *regs, int ret)
```

realhezhe:
Thanks. By adding code extracting from regs, I can now make kfunc work with sys_openat.

But I found another problem. For kprobe, if I attached the trace_entry on __x64_sys_openat, the "PATH" column coming from the 2nd parameter "filename" would be empty. Comparing debug ouput with execsnoop, I noticed there is some difference in the re-writing part as below.


In execsnoop,
int trace_entry(struct pt_regs *ctx)
{
 int dfd = ctx->di; const char __user *filename = ctx->si; int flags = ctx->dx;
#ifdef CONFIG_ARCH_HAS_SYSCALL_WRAPPER
 struct pt_regs * __ctx = ctx->di;
 const char __user *filename; bpf_probe_read(&filename, sizeof(filename), &__ctx->di);
 const char __user *const __user *__argv; bpf_probe_read(&__argv, sizeof(__argv), &__ctx->si);
 const char __user *const __user *__envp; bpf_probe_read(&__envp, sizeof(__envp), &__ctx->dx);
#else
 const char __user *filename = ctx->di; const char __user *const __user *__argv = ctx->si; const char __user *const __user *__envp = ctx->dx;
#endif
...


In opensnoop,
int trace_entry(struct pt_regs *ctx)
{
 int dfd = ctx->di; const char __user *filename = ctx->si; int flags = ctx->dx;
...


CONFIG_ARCH_HAS_SYSCALL_WRAPPER is enabled in my case.
I can't find how they're treated differently in bcc code.

BTW, I also check the parameters passing in the syscall path, the consequence of the re-writing seems just right.

yonghong-song:
For execsnoop.py, we have
```
int syscall__execve(struct pt_regs *ctx,
    const char __user *filename,
    const char __user *const __user *__argv,
    const char __user *const __user *__envp)
{
...
```
The prefix `syscall__` has a special meaning and the rewriter will generate code properly like what you showed in the above.

Alternatively, if you want to avoid the unstable nature of kprobe, we could try syscall tracepoint which is relatively stable. 
```
root@arch-fb-vm1:/sys/kernel/debug/tracing/events/syscalls ls sys_enter_open*
sys_enter_open:
enable  filter  format  id  trigger

sys_enter_openat:
enable  filter  format  id  trigger

sys_enter_openat2:
enable  filter  format  id  trigger

sys_enter_open_by_handle_at:
enable  filter  format  id  trigger

sys_enter_open_tree:
enable  filter  format  id  trigger
```

`open`, `openat` and `openat2` are what we are interested in.

realhezhe:
Thanks. I just updated the pull request with 3 three commits.

yonghong-song:
[buildbot, test this please]

src/python/bcc/__init__.py
```diff
@@ -642,7 +642,10 @@ def get_syscall_prefix(self):
     # return "sys_clone" or "__x64_sys_clone".
     def get_syscall_fnname(self, name):
         name = _assert_is_bytes(name)
-        return self.get_syscall_prefix() + name
+        fullname = self.get_syscall_prefix() + name
+        if self.ksymname(fullname) != -1:
+            return fullname
+        return None
```
yonghong-song, line 648: can we use ksymname() to test return value instead of modifying the function here?

src/python/bcc/__init__.py
```diff
@@ -651,7 +654,11 @@ def fix_syscall_fnname(self, name):
         name = _assert_is_bytes(name)
         for prefix in self._syscall_prefixes:
             if name.startswith(prefix):
-                return self.get_syscall_fnname(name[len(prefix):])
+                fnname = self.get_syscall_fnname(name[len(prefix):])
+                if fnname:
+                    return fnname
+                else:
+                    return name
```
yonghong-song, line 661: why changing this function? The tool did not use this function, right?

tools/opensnoop.py
```diff
@@ -137,38 +178,54 @@
...
-    data.uid = bpf_get_current_uid_gid();
-    data.flags = valp->flags; // EXTENDED_STRUCT_MEMBER
-    data.ret = PT_REGS_RC(ctx);
 
-    events.perf_submit(ctx, &data, sizeof(data));
-    infotmp.delete(&id);
+    if (bpf_get_current_comm(&val.comm, sizeof(val.comm)) == 0) {
+        val.id = id;
+        val.fname = filename;
+        val.flags = how->flags; // EXTENDED_STRUCT_MEMBER
```
yonghong-song, line 202: For this function, I think the main difference between it and generic trace_entry is the above line.
Could you add  something like
```
  int flags = how->flags;
```
in bpf_text_kprobe_header_openat2, it will eliminate this function, right?
something like
```
bpf_text_kprobe_header_openat2 = """
#include <uapi/linux/openat2.h>
int syscall__trace_entry_openat2(struct pt_regs *ctx, int dfd, const char __user *filename, struct open_how *how)
{
   int flags = how->flags;
"""
You need to adjust `syscall__trace_entry_open` and `syscall__trace_entry_openat` as well.

File: tools/opensnoop.py [hunk 6]
```diff
@@ -199,12 +256,70 @@
...
+    events.perf_submit(ctx, &data, sizeof(data));
+
+    return 0;
+}
+"""
+
+b = BPF(text='')
+fnname_open = b.get_syscall_fnname("open").decode()
+fnname_openat = b.get_syscall_fnname("openat").decode()
+fnname_openat2 = b.get_syscall_fnname("openat2").decode()
```
yonghong-song, line 293: Like I mentioned above, may be test fnname_openat2? There is no need to test open and openat as openat exists since linux 2.6.16.

src/python/bcc/__init__.py
realhezhe: OK. I'll move this bit of logic to opensnoop.

BTW, get_syscall_prefix, get_syscall_fnname and fix_syscall_fnname seems a bit weird to me, as they would return a symbol that does not exist at all if all trial had failed. As stated in the comment of get_syscall_prefix, "If no valid value found, will return the first possible value which would probably lead to error in later API calls." It feels returning a None or error is better than returning a non-existing one that might get users confused later.

src/python/bcc/__init__.py
realhezhe: I thought I should keep the logic as same as possible after changing get_syscall_fnname. But yes, this will be kept untouched in the later version.

tools/opensnoop.py
realhezhe: Thanks. I'll update with it.

tools/opensnoop.py
realhezhe: Yes, indeed, I'll remove the checks against open and openat in next version.

tools/opensnoop.py
```diff
@@ -137,38 +181,28 @@
...
-    bpf_probe_read_user(&data.fname, sizeof(data.fname), (void *)valp->fname);
-    data.id = valp->id;
-    data.ts = tsp / 1000;
-    data.uid = bpf_get_current_uid_gid();
-    data.flags = valp->flags; // EXTENDED_STRUCT_MEMBER
-    data.ret = PT_REGS_RC(ctx);
-
-    events.perf_submit(ctx, &data, sizeof(data));
-    infotmp.delete(&id);
+    const char __user *filename = (char *)regs->di; int flags = regs->si;
```
yonghong-song, line 189: The same as below, we cannot use x86 specific registers. This will make tool not portable. Also, `regs` cannot be used to access all registers. You need to do something like below for x64 and arm64.
```
#if defined(CONFIG_ARCH_HAS_SYSCALL_WRAPPER) && !defined(__s390x__)
   struct pt_regs *real_regs = PT_REGS_PARM1(regs);
#else
   struct pt_regs *real_regs = regs;
#endif
   const char __user *filename = PT_REGS_PARAM1(real_regs);
   int flags = PT_REGS_PARAM2(real_regs);
   ...
```

It would be good if we can hide the above macros from user. But this can be delayed until future work.

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

tools/opensnoop.py
realhezhe: > ```
> #if defined(CONFIG_ARCH_HAS_SYSCALL_WRAPPER) && !defined(__s390x__)
>    struct pt_regs *real_regs = PT_REGS_PARM1(regs);
> #else
>    struct pt_regs *real_regs = regs;
> #endif
>    const char __user *filename = PT_REGS_PARAM1(real_regs);
>    int flags = PT_REGS_PARAM2(real_regs);
>    ...
> ```
This only works for kprobes which collects parameter(s) as pt_regs and passes it to the registered hook, so we need to extract twice for SYSCALL_WRAPPER.

For kfunc, we have direct access to the parameter(s). Do you mean the following definition?
```
#if defined(CONFIG_ARCH_HAS_SYSCALL_WRAPPER) && !defined(__s390x__)
KRETFUNC_PROBE(FNNAME, struct pt_regs *regs, int ret)
{
    const char __user *filename = (char *)PT_REGS_PARM1(regs);
    int flags = PT_REGS_PARM2(regs);
#else
KRETFUNC_PROBE(FNNAME, const char __user *filename, int flags, int ret)
{
#endif
```

yonghong-song:
Thanks. The new change w.r.t. KRETFUNC_PROBE looks correct to me. It is a little bit complicated for user programs. I guess we could simplify this through additional macro magic. This can be done in a followup.

PR #3014, opened 2020-07-12
use bpf_probe_read_kernel for implicit kernel mem read on s390

yonghong-song:
Commit f579bf8d60c8 ("bpf: use bpf_probe_read in implicitly
generated kernel mem read") unconditionally use bpf_probe_read()
for implicit kernel memory read in bpf programs.

This won't work for s390 with recent kernels since s390 has
overlap user/kernel addresses and bpf_probe_read() is not
available any more.

This patch partially reverted Commit f579bf8d60c8 such
that for s390, bpf_probe_read_kernel() will be used
while other architectures bpf_probe_read() is used.

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -316,7 +316,10 @@ bool MapVisitor::VisitCallExpr(CallExpr *Call) {
 ProbeVisitor::ProbeVisitor(ASTContext &C, Rewriter &rewriter,
                            set<Decl *> &m, bool track_helpers) :
   C(C), rewriter_(rewriter), m_(m), track_helpers_(track_helpers),
-  addrof_stmt_(nullptr), is_addrof_(false) {}
+  addrof_stmt_(nullptr), is_addrof_(false) {
+  const char **calling_conv_regs = get_call_conv();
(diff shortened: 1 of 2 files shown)


sumanthkorikkar:
Hi @yonghong-song 

Patch works. Tested with biosnoop.py and test_clang.py. Thank you.
     
I think it is worth mentioning this in the reference manual for future users.
Bcc rewriter shall try to follow the function parameter pointer deferences to bpf_probe_read_kernel in arch which has overlappping address space. All the user pointer deferences should be performed using bpf_probe_read_user.  However, For architectures which does not have overlapping address space (like x86), bcc rewriter will rewrite these pointer deferences to bpf_probe_read.

yonghong-song:
> Hi @yonghong-song
> 
> Patch works. Tested with biosnoop.py and test_clang.py. Thank you.
> 
> I think it is worth mentioning this in the reference manual for future users.
> Bcc rewriter shall try to follow the function parameter pointer deferences to bpf_probe_read_kernel in arch which has overlappping address space. All the user pointer deferences should be performed using bpf_probe_read_user. However, For architectures which does not have overlapping address space (like x86), bcc rewriter will rewrite these pointer deferences to bpf_probe_read.

Good point. Will have a followup patch for this later.

PR #3015, opened 2020-07-12
fix cc queue/stack test kernel version

s41m0n:
This PR fixes the Kernel version check in test_queuestack_table.cc .
The correct one, as discussed in #3013, is 4.20 (not 5.0).

Signed-off-by: Simone Magnani <simonemagnani.96@gmail.com>


yonghong-song:
[buildbot, test this please]

PR #3022, opened 2020-07-17
libbpf-tools: add CO-RE biopattern

ethercflow:

### DIFF
--- a/libbpf-tools/biopattern.bpf.c
+++ b/libbpf-tools/biopattern.bpf.c
@@ -1,0 +1,44 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2020 Wenbo Zhang
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include "biopattern.h"
+#include "maps.bpf.h"
+
+const volatile dev_t targ_dev = -1;
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 64);
+	__type(key, dev_t);
+	__type(value, struct counter);
+	__uint(map_flags, BPF_F_NO_PREALLOC);
+} counters SEC(".maps");
+
+SEC("tracepoint/block/block_rq_complete")
+int tp__block__block_rq_complete(struct trace_event_raw_block_rq_complete *ctx)
+{
+	sector_t *last_sectorp,  sector = ctx->sector;
+	struct counter *counterp, zero = {};
+	u32 nr_sector = ctx->nr_sector;
+	dev_t dev = ctx->dev;
+
+	if (targ_dev != -1 && targ_dev != dev)
+		return 0;
+
+	counterp = bpf_map_lookup_or_try_init(&counters, &dev, &zero);
+	if (!counterp)
+		return 0;
+	if (counterp->last_sector) {
+		if (counterp->last_sector == sector)
+			__sync_fetch_and_add(&counterp->sequential, 1);
+		else
+			__sync_fetch_and_add(&counterp->random, 1);
+		__sync_fetch_and_add(&counterp->bytes, nr_sector * 512);
+	}
+	counterp->last_sector = sector + nr_sector;
+	return 0;
+}
+
+char LICENSE[] SEC("license") = "GPL";
--- a/libbpf-tools/biopattern.c
+++ b/libbpf-tools/biopattern.c
@@ -1,0 +1,230 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+// Copyright (c) 2020 Wenbo Zhang
+//
+// Based on biopattern(8) from BPF-Perf-Tools-Book by Brendan Gregg.
+// 17-Jun-2020   Wenbo Zhang   Created this.
+#include <argp.h>
+#include <signal.h>
+#include <stdio.h>
+#include <unistd.h>
+#include <time.h>
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+#include "biopattern.h"
+#include "biopattern.skel.h"
+#include "trace_helpers.h"
+
+#define MINORBITS	20
+#define MINORMASK	((1U << MINORBITS) - 1)
+
+#define MAJOR(dev)	((unsigned int) ((dev) >> MINORBITS))
+#define MINOR(dev)	((unsigned int) ((dev) & MINORMASK))
+#define MKDEV(ma,mi)	(((ma) << MINORBITS) | (mi))
+
+static struct env {
+	time_t interval;
+	bool timestamp;
+	bool verbose;
+	int times;
+	__u32 dev;
+} env = {
+	.interval = 99999999,
+	.times = 99999999,
+	.dev = -1,
+};
+
+static volatile bool exiting;
+
+const char *argp_program_version = "biopattern 0.1";
+const char *argp_program_bug_address = "<ethercflow@gmail.com>";
+const char argp_program_doc[] =
+"Show block device I/O pattern.\n"
+"\n"
+"USAGE: biopattern [-h] [-T] [-m] [interval] [count]\n"
+"\n"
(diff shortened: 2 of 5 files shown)


libbpf-tools/biopattern.bpf.c
```diff
@@ -0,0 +1,60 @@
...
+
+	last_sectorp = bpf_map_lookup_elem(&last_sectors, &dev);
+	if (last_sectorp) {
+		counterp = bpf_map_lookup_elem(&counters, &dev);
+		if (!counterp) {
+			bpf_map_update_elem(&counters, &dev, &zero, 0);
+			counterp = bpf_map_lookup_elem(&counters, &dev);
+			if (!counterp)
+				goto update_last_sector;
+		}
```
anakryiko, line 46: can you please use `bpf_map_lookup_or_try_init` from maps.bpf.h?

libbpf-tools/biopattern.bpf.c
```diff
@@ -0,0 +1,60 @@
...
+		if (*last_sectorp == sector)
+			__sync_fetch_and_add(&counterp->sequential, 1);
+		else
+			__sync_fetch_and_add(&counterp->random, 1);
+		__sync_fetch_and_add(&counterp->bytes, nr_sector * 512);
```
anakryiko, line 51: curious, is the sector size always 512?

libbpf-tools/biopattern.bpf.c
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+update_last_sector:
+	last_sector = sector + nr_sector;
+	bpf_map_update_elem(&last_sectors, &dev, &last_sector, 0);
```
anakryiko, line 56: why not combining last_sectors and counters in a single hashmap? Seems like it would be simpler and faster?

libbpf-tools/biopattern.c
```diff
@@ -0,0 +1,228 @@
...
+		if (env.timestamp) {
+			time(&t);
+			tm = localtime(&t);
+			strftime(ts, sizeof(ts), "%H:%M:%S", tm);
+			printf("%-9s ", ts);
+		}
+		total = counter.sequential + counter.random;
+		printf("%3d:%-2d %5d %5d %8d %10lld\n", MAJOR(next_key),
+			MINOR(next_key), counter.random * 100 / total,
+			counter.sequential * 100 / total, total,
```
anakryiko, line 142: technically, you can have an overflow, so use 100L and %ld in format string?

anakryiko:
just curious, are you using all the tools you've converted in practice?

ethercflow:
> just curious, are you using all the tools you've converted in practice?

Currently, all the bio* tools were used in a few weeks ago (I converted these in Jun) to help my colleagues analyze fio's IO characteristics in our IDC and want to recommend bcc/bpf tools to them. Other tools are not.  I am continuing to read Gregg's BPF-Perf-Tools-Book & kernel books & some parts of kernel source, so I hope to do sth useful, which can help the community and get a better understanding about kernel. Is this the right way? Would you please give me some suggestions? Thank you. :)

anakryiko:
@ethercflow, I think the biggest help would be if you can post a walk-through of some of the conversions you did as a blog post or something like that. BPF is sorely lacking an approachable tutorials and guides, so anything helps in that regard.

The reason I asked if you use the tools was just to get an idea of how much real use libbpf-tools get. Having someone who does use tools in practice to do the conversion is the best, because testing is better and its easier for that person to spot when data is bogus.

ethercflow:
> @ethercflow, I think the biggest help would be if you can post a walk-through of some of the conversions you did as a blog post or something like that. BPF is sorely lacking an approachable tutorials and guides, so anything helps in that regard.

@anakryiko Thanks for your advise!  I will think about it seriously and then implement it. 

> The reason I asked if you use the tools was just to get an idea of how much real use libbpf-tools get. Having someone who does use tools in practice to do the conversion is the best, because testing is better and its easier for that person to spot when data is bogus.

Agree, currently some of my personal usage scenarios are relatively simple，I look forward to more people using them too.  I am currently trying to recommend to my colleagues to use them.
 
The two chances are:
- I am currently working with our [TIKV](https://github.com/tikv/tikv/) performance optimization team and explaining some performance indicators from the principle level of the Linux kernel to help them determine the optimization direction. And have a plan to share BPF with colleagues at the end of August.
- Our company is very open to new technologies, using and sharing are encouraged internally.

The two difficulties are：
- As most of our customers are still using Centos 7（3.10 kernel, only from Centos 7.6's kernel support bpf) , so I think most of time we can only use BCC at there；
- For us internally, because `bpf` lacks support for DWARF stack unwinding, and we kv team cann't  recompile  [TIKV](https://github.com/tikv/tikv/) to enable `-fno-omit-frame-pointer` for some reason, so my colleagues prefer to use perf.  I have a plan to draw a library from perf's libdw/LBR,  but I am not sure for `bpf` copy 8k or more user stack from kernel back to user space is a good decision.  What do you think about?

anakryiko:
> > @ethercflow, I think the biggest help would be if you can post a walk-through of some of the conversions you did as a blog post or something like that. BPF is sorely lacking an approachable tutorials and guides, so anything helps in that regard.
> 
> @anakryiko Thanks for your advise! I will think about it seriously and then implement it.

Great, that would be very helpful.

>     * As most of our customers are still using Centos 7（3.10 kernel, only from Centos 7.6's kernel support bpf) , so I think most of time we can only use BCC at there；

yeah, can't really do anything about that, we'll just need to wait this out

> 
>     * For us internally, because `bpf` lacks support for DWARF stack unwinding, and we kv team cann't  recompile  [TIKV](https://github.com/tikv/tikv/) to enable `-fno-omit-frame-pointer` for some reason, so my colleagues prefer to use perf.  I have a plan to draw a library from perf's libdw/LBR,  but I am not sure for `bpf` copy 8k or more user stack from kernel back to user space is a good decision.  What do you think about?

See  bpf_get_stack() helper, it will tell you how long the stack really is, so you can send over just the right amount. It works just fine in practice, we at Facebook use that pretty extensively. Having build_id added to stack trace bloats stack trace size even more, yet it's still ok. It's the matter of a sampling frequency to balance out the overhead.

As for DWARF-based stack symbolization, that's extremely slow and expensive process, we try to avoid it as much as possible. I believe we use `-fno-omit-frame-pointer` essentially for all the software we control. So might be worth it to fight harder with kv team. Try showing that benefits of omitting frame pointer are not big enough to a better profiling tool support?

ethercflow:
> See bpf_get_stack() helper, it will tell you how long the stack really is, so you can send over just the right amount. It works just fine in practice, we at Facebook use that pretty extensively. Having build_id added to stack trace bloats stack trace size even more, yet it's still ok. It's the matter of a sampling frequency to balance out the overhead.

Thanks, I'll see it and try it later.
 
> As for DWARF-based stack symbolization, that's extremely slow and expensive process, we try to avoid it as much as possible. I believe we use `-fno-omit-frame-pointer` essentially for all the software we control. So might be worth it to fight harder with kv team. Try showing that benefits of omitting frame pointer are not big enough to a better profiling tool support?

Thanks, I'll try harder, I am not too clear about the construction of the rust program, I will try to figure out the reason for KV's rejection.

libbpf-tools/biopattern.bpf.c
ethercflow: From latest code: https://github.com/torvalds/linux/blob/master/block/blk-core.c#L1595, it seems the sector size always 512.

libbpf-tools/biopattern.bpf.c
ethercflow: Fixed

libbpf-tools/biopattern.bpf.c
ethercflow: Fixed

libbpf-tools/biopattern.c
ethercflow: Fixed

ethercflow:

yonghong-song:
[buildbot, test this please]

PR #3030, opened 2020-07-24
Add example for new helper bpf_get_ns_current_pid_tgid() to documentation.

cneira:
This PR completes the work done for [PID filtering issues when running bpf script inside container ](https://github.com/iovisor/bcc/issues/1329).

- Added  bpf_get_ns_current_pid_tgid helper to developer reference documentation. 
- Added example cloning hello_perfoutput.py as hello_perfoutput_using_ns.py replacing  bpf_get_current_pid_tgid() with the new helper  bpf_get_ns_current_pid_tgid().

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -31,6 +31,7 @@ This guide is incomplete. If something feels missing, check the bcc and kernel s
         - [9. bpf_get_prandom_u32()](#9-bpf_get_prandom_u32)
         - [10. bpf_probe_read_user()](#10-bpf_probe_read_user)
         - [11. bpf_probe_read_user_str()](#11-bpf_probe_read_user_str)
+        - [12. bpf_get_ns_current_pid_tgid()](#4-bpf_get_ns_current_pid_tgid)
     - [Debugging](#debugging)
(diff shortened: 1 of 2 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks!

cneira:
> Thanks!

Would it be worth to change calls from  bpf_get_current_pid_tgid() with the new helper bpf_get_ns_current_pid_tgid() on the tools that use that helper?

yonghong-song:
> > Thanks!
> > @yonghong-song
> > Would it be worth to change calls from  bpf_get_current_pid_tgid() with the new helper bpf_get_ns_current_pid_tgid() on the tools that use that helper?

That will be a good idea. We need to test whether the helper is available or not before using it. Also, we may want to have an option to control. We can first try one or two and see what is the best user interface. Thanks for working on this!

cneira:
@yonghong-song 

I like that approach!, I'll work on the first approach:
- if  bpf_get_ns_current_pid_tgid()  exists then use that, if not defer to bpf_get_current_pid_tgid() helper.

Thanks for your feedback!.

Issue #2555, opened 2019-10-16
trace -U cannot show second function from the top of the stack


SPYFF:
Hello! The title of my issue migh leave you with confusion, so let me explain it:
I have a very simple test program, test.c: 
```
#include <unistd.h>
#include <stdio.h>

void func_d() {
	printf("Hello\n");
}
void func_c() {
	func_d();
}
void func_b() {
        func_c();
}
void func_a() {
	func_b();
}
int main() {
         func_a();
}
```

For tracing the `func_d` I use the following command:
```
sudo /usr/share/bcc/tools/trace p:/home/user/test:func_d -U
```

Which give me the following output:
```
PID     TID     COMM            FUNC             
19438   19438   test         func_d           
        func_d+0x0 [test]
        func_b+0x1e [test]
        func_a+0x12 [test]
        main+0x12 [test]
        __libc_start_main+0xf3 [libc-2.30.so]
        [unknown]

19438   19438   test         func_d           
        func_d+0x0 [test]
        func_b+0x1e [test]
        func_a+0x12 [test]
        main+0x12 [test]
        __libc_start_main+0xf3 [libc-2.30.so]
        [unknown]
```

The `func_c` call which should be just under the top of the callstack is missing for some reason. I'm sure this is not because some compiler optimization, because if I check the backtrace ( `bt` ) int `gdb` its looks normal (func_a -> func_b -> func_c -> func_d). Also, if I BCC trace the `func_c`  I get this output:
```
ID     TID     COMM            FUNC             
19438   19438   sleeper         func_c           
        func_c+0x0 [sleeper]
        func_a+0x12 [sleeper]
        main+0x12 [sleeper]
        __libc_start_main+0xf3 [libc-2.30.so]
        [unknown]

19438   19438   sleeper         func_c           
        func_c+0x0 [sleeper]
        func_a+0x12 [sleeper]
        main+0x12 [sleeper]
        __libc_start_main+0xf3 [libc-2.30.so]
        [unknown]

```

As far as I understand, there is certainly something wrong with the second backtrace element, because I lost `func_b` from the output

yonghong-song:
I suspect this is a bcc specific thing. kernel stack addresses seem correct. If you create another func_e, you will get,
```
-bash-4.4$ sudo trace.py p:/home/yhs/tmp/a.out:func_e -U                                                         
PID     TID     COMM            FUNC             
1975599 1975599 a.out           func_e           
        func_e+0x0 [a.out]
[...]

Issue #2150, opened 2019-01-17
BCC trace for kernel function local variables


SPYFF:
Hi! I want to trace local variables of kernel functions, without success so far. Someting similar is possible with BCC trace according to this: https://github.com/iovisor/bcc/pull/439

What I tried so far is to insert a custom probe with perf, which gives me the `size_goal` variable around the 133 line of the `tcp_sendmsg_locked`
```
perf probe tcp_sendmsg_locked:133 size_goal
```

Then I list the available probes:
```
# /usr/share/bcc/tools/tplist | grep tcp

probe:tcp_sendmsg_locked
tcp:tcp_retransmit_skb
tcp:tcp_send_reset
tcp:tcp_receive_reset
tcp:tcp_destroy_sock
tcp:tcp_rcv_space_adjust
tcp:tcp_retransmit_synack
tcp:tcp_probe
```

After that I check the available variables from the probe:
```
# /usr/share/bcc/tools/tplist -v probe:tcp_sendmsg_locked

probe:tcp_sendmsg_locked
    unsigned long __probe_ip;
    s32 size_goal;
```
This shows me that `size_goal` is available as an argument. 

But when I try to trace that I got an error message:
```
# /usr/share/bcc/tools/trace 't:probe:tcp_sendmsg_locked "size_goal=%d", args->size_goal'
ioctl(PERF_EVENT_IOC_SET_BPF): Invalid argument
Failed to attach BPF program tracepoint__probe__tcp_sendmsg_locked to tracepoint probe:tcp_sendmsg_locked
```

Is that normal (this feature not supported) or I miss something? Most of the predefined probes working, for example.
```
# /usr/share/bcc/tools/trace 't:tcp:tcp_rcv_space_adjust "dport=%d", args->dport'
```

Thanks for the help!

yonghong-song:
Which system you are using? In latest net-next, I see `tcp_sendmsg_locked` is a kernel function and there is no kernel tracepoint named with `tcp_sendmsg_locked`.
```
-bash-4.4$ egrep -r tcp_sendmsg_locked
net/tcp.h:int tcp_sendmsg_locked(struct sock *sk, struct msghdr *msg, size_t size);
-bash-4.4$ pwd
/home/yhs/work/net-next/include
-bash-4.4$ egrep -r tcp_rcv_space_adjust
net/tcp.h:void tcp_rcv_space_adjust(struct sock *sk);
[...]

PR #3036, opened 2020-08-01
Offset support for uprobe and kprobe in tools/trace.py

SPYFF:
Offset support for kprobes and uprobes are provided by the API of the BCC. However the tools/trace.py currently has no support for invoking that.

This modification let the user define the offset where the kprobe or uprobe should be injected. The format is simple: `func_name+offset` where the offset could be decimal (objdump like) or hexadecimal.

This modification let the trace user to read register values in the middle of the function (#2150) or access the correct callstack where the base pointer already set (#2555) (both after manual assembly analysis, this patch dont support the variable name resolution nor function prologue analysis of the binary)

### DIFF
--- a/man/man8/trace.8
+++ b/man/man8/trace.8
@@ -86,7 +86,7 @@ information. See PROBE SYNTAX below.
 .SH PROBE SYNTAX
 The general probe syntax is as follows:
 
-.B [{p,r}]:[library]:function[(signature)] [(predicate)] ["format string"[, arguments]]
+.B [{p,r}]:[library]:function[+offset][(signature)] [(predicate)] ["format string"[, arguments]]
 
 .B {t:category:event,u:library:probe} [(predicate)] ["format string"[, arguments]]
 .TP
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

yonghong-song:
Thanks for adding offset support for uprobe and kprobe in trace.py. This should be useful for some advanced users. Could you also update trace_example.txt with a few examples and man/man8/trace.8 to explain `function` for uprobe and kprobe can have optional `+offset`? Thanks!

SPYFF:
I wrote a lengthy example which clarify how this new `+offset` option works with the uprobe case for print register value. I assume the same results with kprobes but I not discuss that, because that looks the same after the advanced user extract the archived kernel image for the vmlinux and analyze the disassembly of that.

yonghong-song:
LGTM. Thanks!

PR #3050, opened 2020-08-12
libbpf-tools: add CO-RE biostacks

ethercflow:
```sh
etcd           12849  sda
bpf_prog_df594d080d725b41_fentry__blk_account_io_start
bpf_prog_df594d080d725b41_fentry__blk_account_io_start
bpf_trampoline_38122
blk_account_io_start
generic_make_request
submit_bio
ext4_io_submit
ext4_writepages
do_writepages
__filemap_fdatawrite_range
file_write_and_wait_range
ext4_sync_file
vfs_fsync_range
do_fsync
__x64_sys_fdatasync
do_syscall_64
        32 -> 63         : 7        |****************************************|

kworker        6539   sda
blk_account_io_merge_bio
blk_attempt_plug_merge
blk_mq_make_request
generic_make_request
submit_bio
ext4_io_submit
ext4_writepages
do_writepages
__writeback_single_inode
writeback_sb_inodes
__writeback_inodes_wb
wb_writeback
wb_workfn
process_one_work
worker_thread
kthread
ret_from_fork
```

### DIFF
--- a/libbpf-tools/biostacks.bpf.c
+++ b/libbpf-tools/biostacks.bpf.c
@@ -1,0 +1,115 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2020 Wenbo Zhang
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_tracing.h>
+#include "biostacks.h"
+#include "bits.bpf.h"
+#include "maps.bpf.h"
+
+#define MAX_ENTRIES	10240
+#define NULL		0
+
+const volatile bool targ_ms = false;
+const volatile dev_t targ_dev = -1;
+
+struct internal_rqinfo {
+	u64 start_ts;
+	struct rqinfo rqinfo;
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_ENTRIES);
+	__type(key, struct request *);
+	__type(value, struct internal_rqinfo);
+	__uint(map_flags, BPF_F_NO_PREALLOC);
+} rqinfos SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_ENTRIES);
+	__type(key, struct rqinfo);
+	__type(value, struct hist);
+	__uint(map_flags, BPF_F_NO_PREALLOC);
+} hists SEC(".maps");
+
+static struct hist zero;
+
+static __always_inline
+int trace_start(void *ctx, struct request *rq, bool merge_bio)
+{
+	struct internal_rqinfo *i_rqinfop = NULL, i_rqinfo = {};
+	struct gendisk *disk = BPF_CORE_READ(rq, rq_disk);
+	dev_t dev;
+
+	dev = disk ? MKDEV(BPF_CORE_READ(disk, major),
+			BPF_CORE_READ(disk, first_minor)) : 0;
+	if (targ_dev != -1 && targ_dev != dev)
+		return 0;
+
+	if (merge_bio)
+		i_rqinfop = bpf_map_lookup_elem(&rqinfos, &rq);
+	if (!i_rqinfop)
+		i_rqinfop = &i_rqinfo;
+
+	i_rqinfop->start_ts = bpf_ktime_get_ns();
+	i_rqinfop->rqinfo.pid = bpf_get_current_pid_tgid();
+	i_rqinfop->rqinfo.kern_stack_size =
+		bpf_get_stack(ctx, i_rqinfop->rqinfo.kern_stack,
+			sizeof(i_rqinfop->rqinfo.kern_stack), 0);
+	bpf_get_current_comm(&i_rqinfop->rqinfo.comm,
+			sizeof(&i_rqinfop->rqinfo.comm));
+	i_rqinfop->rqinfo.dev = dev;
+
+	if (i_rqinfop == &i_rqinfo)
+		bpf_map_update_elem(&rqinfos, &rq, i_rqinfop, 0);
+	return 0;
+}
+
+SEC("fentry/blk_account_io_start")
+int BPF_PROG(blk_account_io_start, struct request *rq)
+{
+	return trace_start(ctx, rq, false);
+}
+
+SEC("kprobe/blk_account_io_merge_bio")
+int BPF_KPROBE(blk_account_io_merge_bio, struct request *rq)
+{
+	return trace_start(ctx, rq, true);
+}
+
+SEC("fentry/blk_account_io_done")
+int BPF_PROG(blk_account_io_done, struct request *rq)
+{
+	u64 slot, ts = bpf_ktime_get_ns();
+	struct internal_rqinfo *i_rqinfop;
+	struct rqinfo *rqinfop;
+	struct hist *histp;
+	s64 delta;
+
+	i_rqinfop = bpf_map_lookup_elem(&rqinfos, &rq);
+	if (!i_rqinfop)
+		return 0;
+	delta = (s64)(ts - i_rqinfop->start_ts);
+	if (delta < 0)
+		goto cleanup;
+	histp = bpf_map_lookup_or_try_init(&hists, &i_rqinfop->rqinfo, &zero);
+	if (!histp)
+		goto cleanup;
+	if (targ_ms)
+		delta /= 1000000;
+	else
+		delta /= 1000;
+	slot = log2l(delta);
+	if (slot >= MAX_SLOTS)
+		slot = MAX_SLOTS - 1;
+	__sync_fetch_and_add(&histp->slots[slot], 1);
+
+cleanup:
+	bpf_map_delete_elem(&rqinfos, &rq);
+	return 0;
+}
+
+char LICENSE[] SEC("license") = "GPL";
--- a/libbpf-tools/biostacks.c
+++ b/libbpf-tools/biostacks.c
@@ -1,0 +1,232 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+// Copyright (c) 2020 Wenbo Zhang
+//
+// Based on biostacks(8) from BPF-Perf-Tools-Book by Brendan Gregg.
+// 10-Aug-2020   Wenbo Zhang   Created this.
+#include <argp.h>
+#include <signal.h>
+#include <stdio.h>
+#include <unistd.h>
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+#include "biostacks.h"
+#include "biostacks.skel.h"
+#include "trace_helpers.h"
+
+static struct env {
+	char *disk;
+	int duration;
+	bool milliseconds;
+	bool verbose;
+} env = {
+	.duration = -1,
+};
+
+const char *argp_program_version = "biostacks 0.1";
+const char *argp_program_bug_address = "<ethercflow@gmail.com>";
+const char argp_program_doc[] =
+"Tracing block I/O with init stacks.\n"
+"\n"
+"USAGE: biostacks [--help] [-d disk] [duration]\n"
+"\n"
+"EXAMPLES:\n"
+"    biostacks              # trace block I/O with init stacks.\n"
+"    biostacks 1            # trace for 1 seconds only\n"
+"    biostacks -d sdc       # trace sdc only\n";
+
+static const struct argp_option opts[] = {
+	{ "disk",  'd', "DISK",  0, "Trace this disk only" },
+	{ "milliseconds", 'm', NULL, 0, "Millisecond histogram" },
+	{ "verbose", 'v', NULL, 0, "Verbose debug output" },
+	{},
+};
+
+static error_t parse_arg(int key, char *arg, struct argp_state *state)
+{
+	static int pos_args;
+
+	switch (key) {
+	case 'v':
+		env.verbose = true;
+		break;
+	case 'h':
+		argp_usage(state);
+		break;
+	case 'd':
+		env.disk = arg;
+		if (strlen(arg) + 1 > DISK_NAME_LEN) {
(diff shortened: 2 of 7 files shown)


ethercflow:

libbpf-tools/biostacks.bpf.c
```diff
@@ -0,0 +1,128 @@
...
+	__uint(map_flags, BPF_F_NO_PREALLOC);
+} hists SEC(".maps");
+
+static __always_inline bool disk_filtered(const char *disk)
+{
+	int i;
+
+	for (i = 0; targ_disk[i] != '\0' && i < DISK_NAME_LEN; i++) {
+		if (disk[i] != targ_disk[i])
+			return false;
```
anakryiko, line 43: I see you do it in few tools already. It would be simpler on BPF side and more efficient to filter disk based on its major/minor numbers, no? Any reason you didn't do it that way? I.e., translate disk name into numbers on user-space side at the start, and then filter by numbers, not by string match, on BPF side.

libbpf-tools/biostacks.bpf.c
```diff
@@ -0,0 +1,128 @@
...
+	bpf_map_update_elem(&rqinfos, &rq, i_rqinfo, 0);
+	return 0;
+}
+
+SEC("fentry/blk_account_io_start")
+int BPF_PROG(fentry__blk_account_io_start, struct request *rq)
+{
+	u32 pid = bpf_get_current_pid_tgid();
+	struct internal_rqinfo i_rqinfo = {};
+	u64 ts = bpf_ktime_get_ns();
```
anakryiko, line 68: nit: doing extra work before you know it's I/O for the desired disk

actually, simpler and nicer way to do this would be to make trace_start capture pid and timestamp

libbpf-tools/biostacks.bpf.c
```diff
@@ -0,0 +1,128 @@
...
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_ENTRIES);
+	__type(key, struct rqinfo);
+	__type(value, struct hist);
+	__uint(map_flags, BPF_F_NO_PREALLOC);
+} hists SEC(".maps");
+
+static __always_inline bool disk_filtered(const char *disk)
```
anakryiko, line 37: nit: filtered is a bit ambiguous (could mean both matched, as well as rejected). disk_allowed() would be more obvious.

libbpf-tools/biostacks.bpf.c
```diff
@@ -0,0 +1,128 @@
...
+	}
+	return trace_start(i_rqinfop, ctx, rq, pid, ts);
+}
+
+SEC("fentry/blk_account_io_done")
+int BPF_PROG(fentry__blk_account_io_done, struct request *rq)
+{
+	u64 slot, ts = bpf_ktime_get_ns();
+	struct internal_rqinfo *i_rqinfop;
+	struct hist *histp, zero = {};
```
anakryiko, line 101: this zero stuff is going to be zero-initialized at every call, it's a bit wasteful. Just declare it as static volatile variable once.

libbpf-tools/biostacks.c
```diff
@@ -0,0 +1,213 @@
...
+	if (err) {
+		fprintf(stderr, "failed to attach blk_account_io_done: %s\n",
+			strerror(err));
+		goto cleanup;
+	}
+
+	signal(SIGINT, sig_handler);
+
+	printf("Tracing block I/O with init stacks. Hit Ctrl-C to end.\n");
+	sleep(env.duration);
```
anakryiko, line 205: this will return prematurely if Ctrl-C hits, is that the intent?

libbpf-tools/biostacks.c
```diff
@@ -0,0 +1,213 @@
...
+		if (err < 0) {
+			fprintf(stderr, "failed to lookup hist: %d\n", err);
+			return -1;
+		}
+		printf("%-14.14s %-6d %-7s\n",
+			next_key.comm, next_key.pid, next_key.disk);
+		num_stack = next_key.kern_stack_size /
+			sizeof(next_key.kern_stack[0]);
+		for (i = 0; i < num_stack; i++) {
+			ksym = ksyms__map_addr(ksyms, next_key.kern_stack[i]);
```
anakryiko, line 116: ksym could be NULL, you need to handle that

libbpf-tools/biostacks.c
```diff
@@ -0,0 +1,213 @@
...
+	if (!ksyms) {
+		fprintf(stderr, "failed to load kallsyms\n");
+		goto cleanup;
+	}
+	if (ksyms__get_symbol(ksyms, "blk_account_io_merge_bio")) {
+		obj->links.kprobe__blk_account_io_merge_bio =
+			bpf_program__attach(obj->
+					progs.kprobe__blk_account_io_merge_bio);
+		err = libbpf_get_error(obj->
+				links.kprobe__blk_account_io_merge_bio);
```
anakryiko, line 185: I really think you should drop "kprobe__" and "fentry__" prefixes, will make for shorter identifiers and they are just redundant and not-so-important prefixes, IMO

libbpf-tools/biostacks.bpf.c
ethercflow: No special reason, actually I used to filter dev_t directly https://github.com/iovisor/bcc/blob/master/libbpf-tools/biopattern.bpf.c#L9.  I can't remember why I use this way in other bio tools,  I'll all other bio tools to filter dev_t in next PR.

libbpf-tools/biostacks.bpf.c
ethercflow: Sure, thanks.

libbpf-tools/biostacks.bpf.c
ethercflow: Agree, thanks.

libbpf-tools/biostacks.bpf.c
ethercflow: Agree, thanks.

libbpf-tools/biostacks.c
ethercflow: Yes, if the user does not specify a duration,  it will sleep forever, so use ctrl-c to end the sleep and output the result.

libbpf-tools/biostacks.c
ethercflow: Oh, sorry for forgot this, I'll fix this then.

libbpf-tools/biostacks.c
ethercflow: Sure, I'll drop these prefixes, and drop other tools in next PR. Thanks!

ethercflow:
Hi @anakryiko! Thanks for the review, pushed fixes for all your comments. :)

libbpf-tools/trace_helpers.c
```diff
@@ -159,6 +166,134 @@ const struct ksym *ksyms__get_symbol(const struct ksyms *ksyms,
...
+{
+	if (!parts)
+		return;
+
+	free(parts->parts);
+	free(parts->strs);
+	free(parts);
+}
+
+const struct part *parts__map_dev(const struct parts *parts, unsigned int dev)
```
anakryiko, line 272: map_dev is not coherent with get_name below. How about `partitions__get_by_dev` and `partitions__get_by_name`?

libbpf-tools/trace_helpers.c
```diff
@@ -159,6 +166,134 @@ const struct ksym *ksyms__get_symbol(const struct ksyms *ksyms,
 	return NULL;
 }
 
+struct parts {
```
anakryiko, line 169: I think parts is not a great name. Let's spell it out completely: `partitions`

libbpf-tools/trace_helpers.c
```diff
@@ -159,6 +166,134 @@ const struct ksym *ksyms__get_symbol(const struct ksyms *ksyms,
...
+};
+
+static int parts__add_part(struct parts *parts, const char *name,
+			   unsigned int dev)
+{
+	size_t new_cap, name_len = strlen(name) + 1;
+	struct part *part;
+	void *tmp;
+
+	if (parts->strs_sz + name_len > parts->strs_cap) {
```
anakryiko, line 185: nit: I don't think you had to do all this complexity for a list of partitions, we don't expect to have tens of thousands of them, right? It's ok to do it the same way as kallsyms, but just doing straightforward strdup() and malloc() would work just fine and be simpler to maintain.

libbpf-tools/biostacks.bpf.c
```diff
@@ -0,0 +1,115 @@
...
+
+static __always_inline
+int trace_start(void *ctx, struct request *rq, bool merge_bio)
+{
+	struct internal_rqinfo *i_rqinfop = NULL, i_rqinfo = {};
+	struct gendisk *disk = BPF_CORE_READ(rq, rq_disk);
+	dev_t dev;
+
+	dev = disk ? MKDEV(BPF_CORE_READ(disk, major),
+			BPF_CORE_READ(disk, first_minor)) : 0;
```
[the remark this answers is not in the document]
ethercflow: I am intentional, I think as we have `partitions__get_by_dev`, so I changed 

```c
struct rqinfo {
 	__u32 pid;
 	int kern_stack_size;
 	__u64 kern_stack[MAX_STACK];
 	char comm[TASK_COMM_LEN];
 	__u32 dev;
 };
```

last member from `char disk[32] to _u32 dev`;

I think it's cheaper than calling `bpf_probe_read_kernel_str` and `BPF_CORE_READ_STR_INTO`, am I right?

libbpf-tools/trace_helpers.c
ethercflow: Agree, yours are much better, thanks!

libbpf-tools/trace_helpers.c
ethercflow: Agree, I'll change to do `strdup()` and `malloc()`, thanks!

libbpf-tools/trace_helpers.c
ethercflow: Agree, thanks.

libbpf-tools/biostacks.bpf.c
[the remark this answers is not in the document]
anakryiko: I meant that you are doing those BPF_CORE_READ()s even if you don't need to filter by device, so I suggested to read that only if targ_dev != -1:

```
if (targ_dev != -1) {
    struct gendisk *disk = BPF_CORE_READ(rq, rq_disk);
    dev_t dev;
    
    dev = disk ? MKDEV(BPF_CORE_READ(disk, major), BPF_CORE_READ(disk, first_minor)) : 0;
    if (targ_dev != dev)
        return 0;
}
```
?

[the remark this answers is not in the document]
ethercflow: Yes I knew what you mean,  but I want to print disk name before printing call stack:

comm pid **disk_name**     # eg:     etcd           12849  **sda**
kernel stack
hist

So I always need `dev`, and in userspace call 

```c
part = partitions__get_by_dev(parts, next_key.dev);
printf("%-14.14s %-6d %-7s\n", next_key.comm, next_key.pid, part ? part->name : "Unknown");
```

Before walking kernel stacks.

> I think it's cheaper than calling bpf_probe_read_kernel_str and BPF_CORE_READ_STR_INTO, am I right?

I meant at last, I read & save the disk name in the kernel by calling `bpf_probe_read_kernel_str` and `BPF_CORE_READ_STR_INTO`  every time,  I think it may be more expensive than just read & save dev. Am I right?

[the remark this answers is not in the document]
anakryiko: ah, I totally missed dev use at line 64, sorry, never mind then

re: overhead, tbh, I don't know what's more or less expensive. I'm just always trying to avoid string comparisons on BPF side, so I prefer dev_t. But then a bunch of user-space complexity could have been saved by just having dev_name. So it's not clear what's the best solution. Both would be fine, I think.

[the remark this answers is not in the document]
ethercflow: Get it, thanks for sharing your experience. :)

ethercflow:
Hi @anakryiko! Thanks for the review,  I have changed to linked lists to manage partitions to simplify. PTAL, thanks! :)

libbpf-tools/trace_helpers.c
```diff
@@ -159,6 +166,108 @@ const struct ksym *ksyms__get_symbol(const struct ksyms *ksyms,
...
+struct partitions {
+	struct list_head list;
+};
+
+static int partitions__add_partition(struct partitions *partitions,
+				     const char *name, unsigned int dev)
+{
+	struct partition *partition;
+
+	partition = malloc(sizeof(*partition));
```
anakryiko, line 178: oh come one, why are you jumping from one extreme to another? Is linked list really necessary? Please stop adding unnecessary "infrastructure" code. Use realloc for dynamically grown array, it's much easier and nicer to work with, compared to linked list.  Do use strdup() for name, my previous objection was against insisting on having a single array of all concatenated strings, that's simply not worth it. But why linked list?... Just an array of struct partitions, you can just `realloc(p, (N+1) * sizeof(struct partition))`. Nice and simple.

ethercflow: @anakryiko Oh so sorry, I find that I often complicate things,  I will think more to avoid this. For now I have changed to dynamic array.  PTAL again. Thank you for your patient guidance.

libbpf-tools/trace_helpers.c
```diff
@@ -159,6 +166,103 @@ const struct ksym *ksyms__get_symbol(const struct ksyms *ksyms,
...
+	fclose(f);
+	return NULL;
+}
+
+void partitions__free(struct partitions *partitions)
+{
+	if (!partitions)
+		return;
+
+	free(partitions->items);
```
anakryiko, line 236: you need to iterate over partitions->items and free ->name as well

ethercflow: Fixed, so terrible, I'm going to face the wall and think ...

yonghong-song:
[buildbot, test this please]

PR #3063, opened 2020-08-19
Forbid trampolines for archs other than x86_64

olsajiri:
The trampoline support check in bcc does not work properly,
so the feature is detected even on architectures that do not
support it - all archs other than x86_64.

We are checking for bpf_trampoline_link_prog to exist in
kernel, which works fine on x86_64 to check if the feature
is supported, but it's global function, so it exists also
in other archs even when the feature is not supported
so it returns True also on other archs.

Adding explicit x86_64 check to support_kfunc function.

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -22,6 +22,7 @@ import re
 import struct
 import errno
 import sys
+import platform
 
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
LGTM. Thanks!

PR #3068, opened 2020-08-23
tools/tcprtt: add tcprtt to trace the RTT of TCP

pacepi:
This program traces TCP RTT(round-trip time) to analyze the quality of
network, then help us to distinguish the network latency trouble is
from user process or physical network.

### DIFF
--- a/man/man8/tcprtt.8
+++ b/man/man8/tcprtt.8
@@ -1,0 +1,76 @@
+.TH tcprtt 8  "2020-08-23" "USER COMMANDS"
+.SH NAME
+tcprtt \- Trace TCP RTT of established connections. Uses Linux eBPF/bcc.
+.SH SYNOPSIS
+.B tcprtt [\-h] [\-T] [\-D] [\-m] [\-i INTERVAL] [\-d DURATION]
+.SH DESCRIPTION
+This tool traces established connections RTT(round-trip time) to analyze the
+quality of network. This can be useful for general troubleshooting to
+distinguish the network latency is from user process or physical network.
+
+Since this uses BPF, only the root user can use this tool.
+.SH REQUIREMENTS
+CONFIG_BPF and bcc.
+.SH OPTIONS
+.TP
+\-h
+Print usage message.
+.TP
+\-T
+Include a time column on output (HH:MM:SS).
+.TP
+\-D
+Show debug infomation of bpf text.
+.TP
+\-m
+Output histogram in milliseconds.
+.TP
+\-i INTERVAL
+Print output every interval seconds.
+.TP
+\-d DURATION
+Total duration of trace in seconds.
+.TP
+\-p SPORT
+Filter for source port.
+.TP
+\-P DPORT
+Filter for destination port.
+.TP
+\-a SADDR
+Filter for source address.
+.TP
+\-A DADDR
+Filter for destination address.
+.SH EXAMPLES
+.TP
+Trace TCP RTT and print 1 second summaries, 10 times:
+#
+.B tcprtt \-i 1 \-d 10
+.TP
+Summarize in millisecond, and timestamps:
+#
+.B tcprtt \-m \-T
+.TP
+Only trace TCP RTT for destination address 192.168.1.100 and destination port 80:
+#
+.B tcprtt \-i 1 \-d 10 -A 192.168.1.100 -P 80
+.SH OVERHEAD
+This traces the kernel tcp_rcv_established function and collects TCP RTT. The
+rate of this depends on your server application. If it is a web or proxy server
+accepting many tens of thousands of connections per second.
+.SH SOURCE
+This is from bcc.
+.IP
+https://github.com/iovisor/bcc
+.PP
+Also look in the bcc distribution for a companion _examples.txt file containing
+example usage, output, and commentary for this tool.
+.SH OS
+Linux
+.SH STABILITY
+Unstable - in development.
+.SH AUTHOR
+zhenwei pi
+.SH SEE ALSO
+tcptracer(8), tcpconnect(8), funccount(8), tcpdump(8)
--- a/tools/tcprtt.py
+++ b/tools/tcprtt.py
@@ -1,0 +1,169 @@
+#!/usr/bin/python
+# @lint-avoid-python-3-compatibility-imports
+#
+# tcprtt    Summarize TCP RTT as a histogram. For Linux, uses BCC, eBPF.
+#
+# USAGE: tcprtt [-h] [-T] [-D] [-m] [-i INTERVAL] [-d DURATION]
+#           [-p SPORT] [-P DPORT] [-a SADDR] [-A DADDR]
+#
+# Copyright (c) 2020 zhenwei pi
+# Licensed under the Apache License, Version 2.0 (the "License")
+#
+# 23-AUG-2020  zhenwei pi  Created this.
+
+from __future__ import print_function
+from bcc import BPF
+from time import sleep, strftime
+import socket, struct
+import argparse
+
+# arguments
+examples = """examples:
+    ./tcprtt            # summarize TCP RTT
+    ./tcprtt -i 1 -d 10 # print 1 second summaries, 10 times
+    ./tcprtt -m -T      # summarize in millisecond, and timestamps
+    ./tcprtt -p         # filter for source port
+    ./tcprtt -P         # filter for destination port
+    ./tcprtt -a         # filter for source address
+    ./tcprtt -A         # filter for destination address
+    ./tcprtt -D         # show debug bpf text
+"""
+parser = argparse.ArgumentParser(
+    description="Summarize TCP RTT as a histogram",
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    epilog=examples)
+parser.add_argument("-i", "--interval",
+    help="summary interval, seconds")
+parser.add_argument("-d", "--duration", type=int, default=99999,
+    help="total duration of trace, seconds")
+parser.add_argument("-T", "--timestamp", action="store_true",
+    help="include timestamp on output")
+parser.add_argument("-m", "--milliseconds", action="store_true",
+    help="millisecond histogram")
+parser.add_argument("-p", "--sport",
+    help="source port")
+parser.add_argument("-P", "--dport",
+    help="destination port")
+parser.add_argument("-a", "--saddr",
+    help="source address")
+parser.add_argument("-A", "--daddr",
+    help="destination address")
+parser.add_argument("-D", "--debug", action="store_true",
+    help="print BPF program before starting (for debugging purposes)")
+parser.add_argument("--ebpf", action="store_true",
+    help=argparse.SUPPRESS)
+args = parser.parse_args()
+if not args.interval:
+    args.interval = args.duration
+
+# define BPF program
+bpf_text = """
+#ifndef KBUILD_MODNAME
+#define KBUILD_MODNAME "bcc"
+#endif
+#include <uapi/linux/ptrace.h>
+#include <linux/tcp.h>
+#include <net/sock.h>
+#include <net/inet_sock.h>
+#include <bcc/proto.h>
+
+BPF_HISTOGRAM(hist_srtt);
+
+int trace_tcp_rcv(struct pt_regs *ctx, struct sock *sk, struct sk_buff *skb)
+{
+    struct tcp_sock *ts = tcp_sk(sk);
+    u32 srtt = ts->srtt_us >> 3;
+    const struct inet_sock *inet = inet_sk(sk);
+
+    SPORTFILTER
+    DPORTFILTER
+    SADDRFILTER
+    DADDRFILTER
+    FACTOR
+
+    hist_srtt.increment(bpf_log2l(srtt));
+
+    return 0;
+}
+"""
+
+# filter for source port
+if args.sport:
+    bpf_text = bpf_text.replace(b'SPORTFILTER',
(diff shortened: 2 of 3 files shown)


netedwardwu:
Hi @pacepi ,

I think this tool is useful but complicated. I don't think your data in tcprtt_example.txt is accurate.
https://github.com/iovisor/bcc/pull/3068/files#diff-44a966a7f2896eca28a300ab24f2d4a8R8
May I ask, did you really use this tool to debug the real-world problem?
Do you really think 256~1023 us RTT is possible for a real server-client environment?

Thanks

tools/tcprtt_example.txt
```diff
@@ -0,0 +1,25 @@
...
+         8 -> 15         : 0        |                                        |
+        16 -> 31         : 0        |                                        |
+        32 -> 63         : 0        |                                        |
+        64 -> 127        : 3        |                                        |
+       128 -> 255        : 17       |                                        |
+       256 -> 511        : 23268    |*                                       |
+       512 -> 1023       : 707578   |****************************************|
+
+The above output shows that the TCP RTT is stable and the network seems good,
+if we hit a latency spike of a process, it maybe a problem of user code.
```
netedwardwu, line 25: 1. Should add the USAGE message
Example: https://github.com/iovisor/bcc/blob/master/tools/funclatency_example.txt#L331

2. Should add more examples for the real-world problem.
For example, what kind of output means network problem instead of user code

pacepi:
> Hi @pacepi ,
> 
> I think this tool is useful but complicated. I don't think your data in tcprtt_example.txt is accurate.
> https://github.com/iovisor/bcc/pull/3068/files#diff-44a966a7f2896eca28a300ab24f2d4a8R8
> May I ask, did you really use this tool to debug the real-world problem?
> Do you really think 256~1023 us RTT is possible for a real server-client environment?
> 
> Thanks

I implemented a kmod version which collects tcp rtt and generates a proc file to read. And I ported it to a bcc version recently.
Typical real-world problem cases:
1, from overlay network to underlay network case. the packets need to walk through software gateway (Ex, ovs-dpdk).
2, database case. several replications of database are deployed in different IDCs.
3, network request -> proxy -> database case.

If hitting a network service latency spike, I usually need to know the quality of network is good or not.
The example is JUST a output sample(two servers connect with the same TOR), it actually can't show the real world problem.

netedwardwu:
Can srtt_us this value measure the RTT of RX???
I think it can only measure the RTT of TX. Am I wrong?

pacepi:
> Can srtt_us this value measure the RTT of RX???
> I think it can only measure the RTT of TX. Am I wrong?

emmm, the achievement looks like this [PAPER](https://medium.com/@alitech_2017/tcprt-transforming-real-time-service-quality-of-large-scale-cloud-databases-34e82f9ab6a3). But it's hard to build the full environment. A bcc version only supports an easy method to debug.

netedwardwu:
> > Can srtt_us this value measure the RTT of RX???
> > I think it can only measure the RTT of TX. Am I wrong?
> 
> emmm, the achievement looks like this [PAPER](https://medium.com/@alitech_2017/tcprt-transforming-real-time-service-quality-of-large-scale-cloud-databases-34e82f9ab6a3). But it's hard to build the full environment. A bcc version only supports an easy method to debug.

Thanks! Actually, my question is just like 
```
TcpRT: Instrument and Diagnostic Analysis System for Service Quality of Cloud Databases at Massive Scale in Real-time
```
image
```
RTT time = T2 − T2’
```
It can only check TX part but not RX. In the above image, you don't know RTT of RspN_1, RspN_2, because you don't have T2 and T2'

But tcprtt this tool will also count RX part in log2histogram.
After my testing, TX part is accurate and really useful but not RX

pacepi:
@netedwardwu Add a specified case and usage in _example.txt.

pacepi:
Hey, @yonghong-song @netedwardwu 
I want to know something about the next step. would you give me some advice?

netedwardwu:
I have no right to merge, need @yonghong-song 's help.

I'm just interested in this tool and do some tests on it.
I test tcprtt with TX throughput and verify the latency with tcpdump, the result is correct and match.
This tool is useful on TX side, and I will use this tool in the future on my NIC TX performance tuning.

However, srtt_us with RX side that I'm not sure what its latency means, and not sure how this calculates.
Like I said before, is it possible to calculate RTT in RX side?

iamkafai:
A few thoughts.
It is kprobing into tcp_rcv_established.   In the bpf prog, the tp->srtt_us sample is taken on every received packet.  I don't think the kernel is updating the RTT unless new data is ack-ed.  Even it is acking new data, the rtt is likely to be very similar (if not the same) as the previous one unless there is something happened drastically.

Also, a faster flow (shorter RTT) will have more packet received which will then have more counts in the bpfmap.

In short, I am not sure what the output is telling me.

yonghong-song:
@pacepi could you respond to @iamkafai 's concern?

pacepi:
> A few thoughts.
> It is kprobing into tcp_rcv_established. In the bpf prog, the tp->srtt_us sample is taken on every received packet. I don't think the kernel is updating the RTT unless new data is ack-ed. Even it is acking new data, the rtt is likely to be very similar (if not the same) as the previous one unless there is something happened drastically.

Of cause, kprobe hurts the performance. Luckly, we usually don't run bcc tool as a resident process. And I have run a kmod (also kprobe into tcp_rcv_established and collocts rtt/mdev into a proc file) on a server(25G NIC), it works fine.
And if the rtt changes a few, it means that the network is stable, the latency spike of network service may be caused by other reasons.

> 
> Also, a faster flow (shorter RTT) will have more packet received which will then have more counts in the bpfmap.
> 

Yes, you are right. In my plans,  if the tool could be accepted, the more arguments/filter may be needed. Ex, saddr/sport/daddr/dport.

> In short, I am not sure what the output is telling me.

The example in _expample.txt:
wrk shows the avg latency(86.75ms) and distribution:
     50%   14.55ms
     75%  119.21ms
     90%  230.22ms
     99%  726.90ms

The latency jitter is very obvious, but from this report, I don't know if the network is unstable or my networking service has a bug.

During wrk testing, run tcprtt:
# ./tcprtt.py -i 1 -d 10 -m
Tracing TCP RTT... Hit Ctrl-C to end.
     msecs               : count     distribution
         0 -> 1          : 4        |                                        |
         4 -> 7          : 1055     |****************************************|
         8 -> 15         : 26       |                                        |
        64 -> 127        : 18       |                                        |
       128 -> 255        : 14       |                                        |
       256 -> 511        : 14       |                                        |
       512 -> 1023       : 12       |                                        |

From this output, I can see the TCP RTT also has jitter. So I think I should check the network firstly rather than checking my code blindly.

yonghong-song:
@pacepi for
>  if the tool could be accepted, the more arguments/filter may be needed. Ex, saddr/sport/daddr/dport.
could you describe (or better implement) what you plan to do here to address @iamkafai 's concern?

pacepi:
> @pacepi for
> 
> > if the tool could be accepted, the more arguments/filter may be needed. Ex, saddr/sport/daddr/dport.
> > could you describe (or better implement) what you plan to do here to address @iamkafai 's concern?

Pushed better implement which supports source address/port and destination address/port as filter.

pacepi:
@iamkafai @yonghong-song 
Add source address/port and destination address/port as filter to drop some streams which we don't care.

Let me show a specified case:
A web service written by golang hits a latency spike, We need to find the reason: golang gc mechanism, or schedule latency, or network jitter ...
During running web server, tcprtt finds that the rtt is unstable. GOOD,  so we should make sure the network is stable firstly.
and if the rtt is stable (as @iamkafai said 'the rtt is likely to be very similar'), GOOD, the latency spike is probably not caused by network.

iamkafai:
Having ip filtering may make it good enough for some ad-hoc debugging.

pacepi:
Hi, @yonghong-song 
It seems that we have reached an agreement with ip filtering for tcprtt.

yonghong-song:
[buildbot, test this please]

yonghong-song:
> Hi, @yonghong-song
> It seems that we have reached an agreement with ip filtering for tcprtt.
Sounds good! Looking at the code now.

man/man8/tcprtt.8
```diff
@@ -0,0 +1,76 @@
...
+\-i INTERVAL
+Print output every interval seconds.
+.TP
+\-d DURATION
+Total duration of trace in seconds.
+.TP
+\-p SPORT
+Filter for source port.
+.TP
+\-p DPORT
```
yonghong-song, line 37: both SPORT and DPORT have -p, typo?

man/man8/tcprtt.8
```diff
@@ -0,0 +1,76 @@
...
+Filter for destination port.
+.TP
+\-a SADDR
+Filter for source address.
+.TP
+\-a DADDR
```
yonghong-song, line 43: both SADDR and DADDR are -a, typo?

man/man8/tcprtt.8
```diff
@@ -0,0 +1,76 @@
...
+https://github.com/iovisor/bcc
+.PP
+Also look in the bcc distribution for a companion _examples.txt file containing
+example usage, output, and commentary for this tool.
+.SH OS
+Linux
+.SH STABILITY
+Unstable - in development.
+.SH AUTHOR
+zhenwei pi
```
yonghong-song, line 74: Zhenwei Pi?
The same for all other names in this patch.

tools/tcprtt.py
```diff
@@ -0,0 +1,170 @@
...
+#include <net/inet_sock.h>
+#include <bcc/proto.h>
+
+BPF_HISTOGRAM(hist_srtt);
+
+int trace_tcp_rcv(struct pt_regs *ctx, struct sock *sk, struct sk_buff *skb)
+{
+    struct tcp_sock *ts = tcp_sk(sk);
+    u32 srtt_usec = ts->srtt_us >> 3;
+    u32 srtt = srtt_usec;
```
yonghong-song, line 76: srtt_usec is not used. You can just do `u32 srtt = ts->srtt_us >> 3;`.

tools/tcprtt.py
```diff
@@ -0,0 +1,170 @@
...
+    bpf_probe_read_kernel(&dport, sizeof(dport), (void *)&inet->inet_dport);
+    if (ntohs(dport) != %d)
+        return 0;""" % int(args.dport))
+else:
+    bpf_text = bpf_text.replace(b'DPORTFILTER', b'')
+
+# filter for source address
+if args.saddr:
+    bpf_text = bpf_text.replace(b'SADDRFILTER',
+        b"""u32 saddr = 0;
```
yonghong-song, line 114: Here, only ipv4 address is available for filtering. It would be good we can recognize and filter for ipv6 address. But this can be in a followup patch.

tools/tcprtt.py
```diff
@@ -0,0 +1,170 @@
...
+    bpf_probe_read_kernel(&saddr, sizeof(saddr), (void *)&inet->inet_saddr);
+    if (saddr != %d)
+        return 0;""" % struct.unpack("=I", socket.inet_aton(args.saddr))[0])
+else:
+    bpf_text = bpf_text.replace(b'SADDRFILTER', b'')
+
+# filter for source address
+if args.daddr:
+    bpf_text = bpf_text.replace(b'DADDRFILTER',
+        b"""u32 daddr = 0;
```
yonghong-song, line 124: Same as the above, it would be good if we support ipv6 address filtering as well here. Note that tool captures srtt for both ipv4 and ipv6 connections.

yonghong-song (commented):
Mostly looks good. A few minor comments and a followup request which you can also do in this patch if you want. Thanks!

yonghong-song:
[buildbot, ok to test]

man/man8/tcprtt.8
pacepi: A typo, need fix. Thanks.

man/man8/tcprtt.8
pacepi: A typo, need fix. Thanks.

man/man8/tcprtt.8
pacepi: I prefer name in lower-case letters  ...
And keep all names in lower-case in this patch.

brendangregg:
Thanks; a couple of suggested follow-up PRs:

- Add an entry for this to README.md
- I think it's missing what should be the primary use case: show a breakdown of remotehost RTT, to look for RTT differences across different hosts/paths. See biolatency -D: how this option changes biolatency from printing a single histogram to doing a per-disk breakdown.

Picture a basic example (and this is the first thing I wanted to do with this tool at Netflix): you have a web server and clients complaining about latency. You want to do a RTT breakdown by client to look for patterns. I'd want to run a command like this:

```
tcprtt -l 80 -R
```

meaning local port 80 (-l 80) and remote address breakdown (-R) to see a histogram for each remote address.

We'll run out of letters, so these options may need to be full-form (following the convention of the other options):

```
tcprtt --lport 80 --by-raddr
```

The current tcprtt allows me to filter by source or destination port: how would I use those options in this scenario? Is it --sport 80 to match on the outbound packet? Or --dport 80 to match on the inbound packet?

I think a problem is that srtt_us is a by-socket metric, and at the socket level you have local and remote ports (or endpoints A and B), but this tool is using by-packet terminology (source/destination) to analyze the socket. I find it very confusing, and I'm familiar with tracing srtt_us.

If these options are added: --lport, --rport, --by-laddr, --by-raddr: I imagine they'll be the primarily used options, and not sport/dport.

PR #3073, opened 2020-08-26
fix the exits abnormal due to trace buffer

chenquanquan:
Some data may be left in the trace buffer, and the format cannot be
parsed by trace_fields(), which whill cause an abnormality in the
system.

Re-running the program cannot solved, only restart kernel.

Call trace:
      File "/usr/lib/python2.7/site-packages/kwaibcc/__init__.py", line 1273, in trace_fields
        pid, cpu, flags, ts = line[:ts_end].split()
    ValueError: need more than 2 values to unpack

Prevent problems with exception handling.

Signed-off-by: Chen Guanqiao <chen.chenchacha@foxmail.com>

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -1278,7 +1278,10 @@ class BPF(object):
             task = line[:16].lstrip()
             line = line[17:]
             ts_end = line.find(b":")
-            pid, cpu, flags, ts = line[:ts_end].split()
+            try:
+                pid, cpu, flags, ts = line[:ts_end].split()
+            except Exception as e:
+                continue
             cpu = cpu[1:-1]
             # line[ts_end:] will have ": [sym_or_addr]: msgs"
             # For trace_pipe debug output, the addr typically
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

src/python/bcc/__init__.py
```diff
@@ -1290,7 +1293,10 @@ def trace_fields(self, nonblocking=False):
             line = line[ts_end + 1:]
             sym_end = line.find(b":")
             msg = line[sym_end + 2:]
-            return (task, int(pid), int(cpu), flags, float(ts), msg)
+            try:
+                return (task, int(pid), int(cpu), flags, float(ts), msg)
+            except Exception as e:
+                return (task, pid, cpu, flags, ts, msg)
```
yonghong-song, line 1299: Is the change here necessary? Even if we return in Exception to caller, it may still have exceptions due to ill-formed fields, right?

yonghong-song:
[buildbot, ok to test]

src/python/bcc/__init__.py
chenquanquan: You're right，I need to fill the return:
return ("Unknown", 0, 0, "Unknown", 0.0, "Unknown")

yonghong-song:
LGTM. Thanks.

PR #3076, opened 2020-08-28
libbpf-tools: add softirqs

ethercflow:
```

16:34:08
SOFTIRQ          TOTAL_usecs
timer                    168
net_rx                  1251
tasklet                   22
sched                   1005
rcu                      212
```

Or

```

softirq = timer
         0 -> 1          : 152      |****************************************|
         2 -> 3          : 20       |*****                                   |
         4 -> 7          : 7        |*                                       |

softirq = net_rx
         8 -> 15         : 1        |********************                    |
        16 -> 31         : 2        |****************************************|

softirq = tasklet
         4 -> 7          : 1        |**********                              |
         8 -> 15         : 4        |****************************************|
        16 -> 31         : 3        |******************************          |

softirq = sched
         0 -> 1          : 17       |******                                  |
         2 -> 3          : 112      |****************************************|
         4 -> 7          : 49       |*****************                       |
         8 -> 15         : 1        |                                        |
        32 -> 63         : 1        |                                        |
        64 -> 127        : 1        |                                        |

softirq = rcu
         0 -> 1          : 146      |****************************************|
         2 -> 3          : 12       |***                                     |
         4 -> 7          : 3        |                                        |
         8 -> 15         : 9        |**                                      |
        16 -> 31         : 3        |                                        |
```

### DIFF
--- a/libbpf-tools/Makefile
+++ b/libbpf-tools/Makefile
@@ -22,6 +22,7 @@ APPS = \
 	opensnoop \
 	readahead \
 	runqslower \
+	softirqs \
 	syscount \
 	tcpconnect \
 	tcpconnlat \
--- a/libbpf-tools/softirqs.bpf.c
+++ b/libbpf-tools/softirqs.bpf.c
@@ -1,0 +1,67 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2020 Wenbo Zhang
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include "softirqs.h"
+#include "bits.bpf.h"
+#include "maps.bpf.h"
+
+const volatile bool targ_dist = false;
+const volatile bool targ_ns = false;
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__uint(max_entries, 1);
+	__type(key, u32);
+	__type(value, u64);
+} start SEC(".maps");
+
+__u64 counts[NR_SOFTIRQS];
+struct hist hists[NR_SOFTIRQS];
+
+SEC("tp_btf/softirq_entry")
+int BPF_PROG(softirq_entry, unsigned int vec_nr)
+{
+	u64 ts = bpf_ktime_get_ns();
+	u32 key = 0;
+
+	bpf_map_update_elem(&start, &key, &ts, 0);
+	return 0;
+}
+
+SEC("tp_btf/softirq_exit")
+int BPF_PROG(softirq_exit, unsigned int vec_nr)
+{
+	u32 key = 0;
+	s64 delta;
+	u64 *tsp;
+
+	if (vec_nr >= NR_SOFTIRQS)
+		return 0;
+	tsp = bpf_map_lookup_elem(&start, &key);
+	if (!tsp || !*tsp)
+		return 0;
+	delta = bpf_ktime_get_ns() - *tsp;
+	if (delta < 0)
+		return 0;
+	if (!targ_ns)
+		delta /= 1000U;
+
+	if (!targ_dist) {
+		__sync_fetch_and_add(&counts[vec_nr], delta);
+	} else {
+		struct hist *hist;
+		u64 slot;
+
+		hist = &hists[vec_nr];
+		slot = log2(delta);
+		if (slot >= MAX_SLOTS)
+			slot = MAX_SLOTS - 1;
+		__sync_fetch_and_add(&hist->slots[slot], 1);
+	}
+
+	return 0;
+}
+
+char LICENSE[] SEC("license") = "GPL";
--- a/libbpf-tools/softirqs.c
+++ b/libbpf-tools/softirqs.c
@@ -1,0 +1,255 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+// Copyright (c) 2020 Wenbo Zhang
+//
+// Based on softirq(8) from BCC by Brendan Gregg & Sasha Goldshtein.
+// 15-Aug-2020   Wenbo Zhang   Created this.
+#include <argp.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+#include <unistd.h>
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+#include "softirqs.h"
+#include "softirqs.skel.h"
+#include "trace_helpers.h"
+
+struct env {
+	bool distributed;
+	bool nanoseconds;
+	time_t interval;
+	int times;
+	bool timestamp;
+	bool verbose;
+} env = {
+	.interval = 99999999,
+	.times = 99999999,
+};
+
+static volatile bool exiting;
+
+const char *argp_program_version = "softirqs 0.1";
+const char *argp_program_bug_address = "<bpf@vger.kernel.org>";
+const char argp_program_doc[] =
+"Summarize soft irq event time as histograms.\n"
+"\n"
+"USAGE: softirqs [--help] [-T] [-N] [-d] [interval] [count]\n"
+"\n"
+"EXAMPLES:\n"
+"    softirqss            # sum soft irq event time\n"
+"    softirqss -d         # show soft irq event time as histograms\n"
+"    softirqss 1 10       # print 1 second summaries, 10 times\n"
+"    softirqss -NT 1      # 1s summaries, nanoseconds, and timestamps\n";
+
+static const struct argp_option opts[] = {
+	{ "distributed", 'd', NULL, 0, "Show distributions as histograms" },
+	{ "timestamp", 'T', NULL, 0, "Include timestamp on output" },
+	{ "nanoseconds", 'N', NULL, 0, "Output in nanoseconds" },
+	{ "verbose", 'v', NULL, 0, "Verbose debug output" },
+	{},
+};
+
+static error_t parse_arg(int key, char *arg, struct argp_state *state)
+{
+	static int pos_args;
+
+	switch (key) {
+	case 'v':
+		env.verbose = true;
+		break;
+	case 'd':
+		env.distributed = true;
+		break;
+	case 'N':
+		env.nanoseconds = true;
+		break;
+	case 'T':
+		env.timestamp = true;
+		break;
+	case ARGP_KEY_ARG:
+		errno = 0;
+		if (pos_args == 0) {
+			env.interval = strtol(arg, NULL, 10);
+			if (errno) {
+				fprintf(stderr, "invalid internal\n");
+				argp_usage(state);
+			}
+		} else if (pos_args == 1) {
+			env.times = strtol(arg, NULL, 10);
+			if (errno) {
+				fprintf(stderr, "invalid times\n");
+				argp_usage(state);
+			}
+		} else {
+			fprintf(stderr,
+				"unrecognized positional argument: %s\n", arg);
+			argp_usage(state);
+		}
+		pos_args++;
+		break;
+	default:
+		return ARGP_ERR_UNKNOWN;
+	}
+	return 0;
+}
+
+int libbpf_print_fn(enum libbpf_print_level level,
+		const char *format, va_list args)
+{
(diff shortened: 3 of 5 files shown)


ethercflow:

libbpf-tools/softirqs.bpf.c
```diff
@@ -0,0 +1,84 @@
...
+
+	tsp = bpf_map_lookup_elem(&start, &key);
+	if (!tsp || !*tsp)
+		return 0;
+
+	delta = bpf_ktime_get_ns() - *tsp;
+	if (delta < 0)
+		return 0;
+	if (!targ_ns)
+		delta /= 1000;
```
anakryiko, line 59: BPF doesn't have signed division instruction, if this works, it's only by accident due to some Clang optimization. Please use unsigned integers when dividing

libbpf-tools/softirqs.c
```diff
@@ -0,0 +1,277 @@
...
+
+static int print_hist(int fd)
+{
+	char *units = env.nanoseconds ? "nsecs" : "usecs";
+	struct hist hist, zero = {};
+	__u32 vec;
+	int err;
+
+	for (vec = 0; vec < NR_SOFTIRQS; vec++) {
+		err = bpf_map_lookup_elem(fd, &vec, &hist);
```
anakryiko, line 176: given `NR_SOFTIRQS` is small and fixed, there is no reason to use ARRAY map, just use global array variable for this. It will be faster and simpler on both BPF and user-space side.

libbpf-tools/softirqs.c
```diff
@@ -0,0 +1,277 @@
...
+			time(&t);
+			tm = localtime(&t);
+			strftime(ts, sizeof(ts), "%H:%M:%S", tm);
+			printf("%-8s\n", ts);
+		}
+
+		if (!env.distributed)
+			err = print_count(bpf_map__fd(obj->maps.counts));
+		else
+			err = print_hist(bpf_map__fd(obj->maps.hists));
```
anakryiko, line 265: similar to another PR comment, maybe pass `struct bpf_map *` as arguments?

libbpf-tools/softirqs.bpf.c
ethercflow: Fixed, thanks!

libbpf-tools/softirqs.c
ethercflow: Fixed, thanks!

libbpf-tools/softirqs.c
ethercflow: Sure,  as we've change bpf arrays to global values, so I change it to 

```c
static int print_hist(struct softirqs_bpf__bss *bss);
static int print_count(struct softirqs_bpf__bss *bss);
```

instead.

libbpf-tools/Makefile
```diff
@@ -59,7 +60,7 @@ $(OUTPUT) $(OUTPUT)/libbpf:
 
 $(APPS): %: $(OUTPUT)/%.o $(LIBBPF_OBJ) $(COMMON_OBJ) | $(OUTPUT)
 	$(call msg,BINARY,$@)
-	$(Q)$(CC) $(CFLAGS) $^ -lelf -lz -o $@
+	$(Q)$(CC) $(CFLAGS) $^ -latomic -lelf -lz -o $@
```
anakryiko, line 63: why is this needed? those built-ins shouldn't require extra libraries

libbpf-tools/softirqs.c
```diff
@@ -0,0 +1,258 @@
...
+
+static int print_count(struct softirqs_bpf__bss *bss)
+{
+	const char *units = env.nanoseconds ? "nsecs" : "usecs";
+	__u32 vec;
+
+	printf("%-16s %6s%5s\n", "SOFTIRQ", "TOTAL_", units);
+
+	for (vec = 0; vec < NR_SOFTIRQS; vec++) {
+		if (bss->counts[vec] > 0)
```
anakryiko, line 146: use __atomic_load_n() to read it

libbpf-tools/softirqs.c
```diff
@@ -0,0 +1,258 @@
...
+		hist = &bss->hists[vec];
+		if (!memcmp(&zero, hist, sizeof(*hist)))
+			continue;
+		printf("softirq = %s\n", vec_names[vec]);
+		print_log2_hist(hist->slots, MAX_SLOTS, units);
+		printf("\n");
+	}
+
+	for (vec = 0; vec < NR_SOFTIRQS; vec++)
+		__atomic_store(&bss->hists[vec], &zero, __ATOMIC_RELAXED);
```
anakryiko, line 175: how is this supposed to work? atomics are only available for a single word integers, not arbitrary structs. I guess for histograms you'll have to use map after all :(

anakryiko: or we can just leave with races and zero out non-atomically with memset, which is probably ok as well

libbpf-tools/Makefile
ethercflow: I found when calling   generic version of an atomic operation, such as `__atomic_store` without `_n`, we need to link `libatomic`.

libbpf-tools/softirqs.c
ethercflow: > how is this supposed to work? atomics are only available for a single word integers, not arbitrary structs. I guess for histograms you'll have to use map after all :(

Oh I think I misunderstood the meaning of "generic version"

> Built-in Function: void __atomic_store (type *ptr, type *val, int memorder)
> This is the generic version of an atomic store. It stores the value of *val into *ptr.

ethercflow: > how is this supposed to work? 

Um, I am a bit confused about this at the moment,  I do agree 

> atomics are only available for a single word integers, not arbitrary structs. 

But from  https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html I saw

> Built-in Function: void __atomic_store (type *ptr, type *val, int memorder)
> This is the generic version of an atomic store. It stores the value of *val into *ptr.

And from the test results, it does modify the entire structure （ `softirq -d 1` # clear  once a second）

```
softirq = timer
         0 -> 1          : 600      |****************************************|
         2 -> 3          : 39       |**                                      |
         4 -> 7          : 53       |***                                     |
         8 -> 15         : 4        |                                        |
        16 -> 31         : 13       |                                        |
        32 -> 63         : 3        |                                        |
        64 -> 127        : 2        |                                        |

softirq = timer
         0 -> 1          : 679      |****************************************|
         2 -> 3          : 30       |*                                       |
         4 -> 7          : 50       |**                                      |
         8 -> 15         : 5        |                                        |
        16 -> 31         : 5        |                                        |
        32 -> 63         : 1        |                                        |
        64 -> 127        : 3        |                                        |

softirq = timer
         0 -> 1          : 483      |****************************************|
         2 -> 3          : 30       |**                                      |
         4 -> 7          : 46       |***                                     |
         8 -> 15         : 4        |                                        |
        16 -> 31         : 4        |                                        |
        32 -> 63         : 3        |                                        |
        64 -> 127        : 3        |                                        |

softirq = timer
         0 -> 1          : 426      |****************************************|
         2 -> 3          : 22       |**                                      |
         4 -> 7          : 40       |***                                     |
         8 -> 15         : 4        |                                        |
        16 -> 31         : 2        |                                        |
        32 -> 63         : 4        |                                        |
        64 -> 127        : 1        |                                        |

softirq = timer
         0 -> 1          : 565      |****************************************|
         2 -> 3          : 41       |**                                      |
         4 -> 7          : 44       |***                                     |
         8 -> 15         : 7        |                                        |
        16 -> 31         : 9        |                                        |
        32 -> 63         : 4        |                                        |
        64 -> 127        : 1        |                                        |

softirq = timer
         0 -> 1          : 391      |****************************************|
         2 -> 3          : 22       |**                                      |
         4 -> 7          : 41       |****                                    |
         8 -> 15         : 4        |                                        |
        16 -> 31         : 1        |                                        |
        32 -> 63         : 4        |                                        |
        64 -> 127        : 1        |                                        |
```

I will determine the reason.

ethercflow: > how is this supposed to work? atomics are only available for a single word integers, not arbitrary structs. I guess for histograms you'll have to use map after all :(

From
https://sourcegraph.com/github.com/gcc-mirror/gcc@e1336703f8220dcffdeddb5e19dd032c766fbb8f/-/blob/libatomic/config/posix/lock.c#L91
https://sourcegraph.com/github.com/gcc-mirror/gcc@e1336703f8220dcffdeddb5e19dd032c766fbb8f/-/blob/libatomic/gstore.c#L100
I see  `__atomic_store` is implement by `pthread_mutex` + `memcpy`.

I perfer 
> or we can just leave with races and zero out non-atomically with memset, which is probably ok as well

than map, thanks!

libbpf-tools/softirqs.c
ethercflow: Fixed, thanks!

libbpf-tools/softirqs.c
ethercflow: Fixed, thanks!

ethercflow:
Hi @anakryiko! Thanks for the review, pushed fixes for all your comments. :)

libbpf-tools/softirqs.c
```diff
@@ -0,0 +1,258 @@
...
+	printf("%-16s %6s%5s\n", "SOFTIRQ", "TOTAL_", units);
+
+	for (vec = 0; vec < NR_SOFTIRQS; vec++) {
+		count = __atomic_load_n(&bss->counts[vec], __ATOMIC_RELAXED);
+		if (count > 0)
+			printf("%-16s %11llu\n", vec_names[vec], count);
+	}
+
+	for (vec = 0; vec < NR_SOFTIRQS; vec++)
+		__atomic_store_n(&bss->counts[vec], 0, __ATOMIC_RELAXED);
```
anakryiko, line 153: `__atomic_exchange_n` is what you want here to read value and atomically set it to zero at the same time

libbpf-tools/softirqs.c
```diff
@@ -0,0 +1,258 @@
...
+	for (vec = 0; vec < NR_SOFTIRQS; vec++) {
+		hist = &bss->hists[vec];
+		if (!memcmp(&zero, hist, sizeof(*hist)))
+			continue;
+		printf("softirq = %s\n", vec_names[vec]);
+		print_log2_hist(hist->slots, MAX_SLOTS, units);
+		printf("\n");
+	}
+
+	memset(&bss->hists, 0, sizeof(bss->hists));
```
anakryiko, line 175: do memset for each individual histogram maybe? it looks cleaner and behaves a bit saner, minimizing the race between reading histogram and resetting it

anakryiko: also, given histogram is modified at the same time as you are printing it out (with all those fancy columns of starts, etc), copy one histogram at a time locally, reset it to zero immediately, then emit it from local copy. That way it's safe on output and minimizes the race between reading and resetting to all zeros.

ethercflow: @anakryiko Fixed, thanks!

libbpf-tools/softirqs.c
ethercflow: Fixed, thanks!

libbpf-tools/softirqs.c
```diff
@@ -0,0 +1,255 @@
...
+
+static int print_hist(struct softirqs_bpf__bss *bss)
+{
+	const char *units = env.nanoseconds ? "nsecs" : "usecs";
+	__u32 vec;
+
+	for (vec = 0; vec < NR_SOFTIRQS; vec++) {
+		struct hist hist = bss->hists[vec];
+
+		memset(&bss->hists[vec], 0, sizeof(bss->hists[vec]));
```
anakryiko, line 166: you copied struct implicitly (through struct assignment), but set it to zero through memset :)) this line could have been:
```
bss->hists[vec] = zero;
```

anakryiko (approved):
Looks good. Approved.

libbpf-tools/softirqs.c
ethercflow: Oh,  sorry for this, haha. Fixed, thanks!

ethercflow:

yonghong-song:
[buildbot, test this please]

PR #3078, opened 2020-08-31
bcc/tools: add lustredist and lustreslower for lustre client

KnightKu:
Add tools lustredist and lustreslower for lustre client.

Signed-off-by: Gu Zheng <cengku@gmail.com>


yonghong-song:
I did not find kernel funciton `ll_file_open` in the mainline linux source code. So `lustre` file system is still out-of-tree and in development, right? If this is the case, I suggest to put these tools under `examples/tracing/lustre`. You can add a simple readme file in that directory to say why these two tools are there. I do not want to put tools in tools directory where most people do not use. Thanks!

KnightKu:
@yonghong-song Ah, I see. Let me rework it accordingly, thanks.

PR #3080, opened 2020-08-31
libbpf-tools: add hardirqs

ethercflow:
The output looks like: 

```
Tracing hard irq event time... Hit Ctrl-C to end.
^C
HARDIRQ                    TOTAL_usecs
nvme0q1                             33
ens33                               18
ahci[0000:02:05.0]                  76
nvme0q5                             17
nvme0q13                            14
nvme0q2                             31
nvme0q4                             61
nvme0q7                             16
```
```
Tracing hard irq event time... Hit Ctrl-C to end.
^C
HARDIRQ                    TOTAL_count
nvme0q1                              3
nvme0q5                              1
nvme0q4                              3
nvme0q7                              4
ens33                                4
nvme0q3                              3
nvme0q13                             2
nvme0q10                             2
nvme0q8                              2
nvme0q9                              2
```
```
Tracing hard irq event time... Hit Ctrl-C to end.
^C
hardirq = nvme0q6
         8 -> 15         : 2        |****************************************|
hardirq = nvme0q4
hardirq = nvme0q11
         8 -> 15         : 1        |****************************************|
        16 -> 31         : 1        |****************************************|
hardirq = ens33
         0 -> 1          : 4        |****************************************|
         4 -> 7          : 4        |****************************************|
         8 -> 15         : 1        |**********                              |
hardirq = ahci[0000:02:05.0]
         2 -> 3          : 1        |****************************************|
hardirq = nvme0q8
         8 -> 15         : 1        |****************************************|
```

### DIFF
--- a/libbpf-tools/hardirqs.bpf.c
+++ b/libbpf-tools/hardirqs.bpf.c
@@ -1,0 +1,93 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2020 Wenbo Zhang
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include "hardirqs.h"
+#include "bits.bpf.h"
+#include "maps.bpf.h"
+
+#define MAX_ENTRIES	256
+
+const volatile bool targ_dist = false;
+const volatile bool targ_ns = false;
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__uint(max_entries, 1);
+	__type(key, u32);
+	__type(value, u64);
+} start SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_ENTRIES);
+	__type(key, struct irq_key);
+	__type(value, struct info);
+} infos SEC(".maps");
+
+static struct info zero;
+
+SEC("tracepoint/irq/irq_handler_entry")
+int handle__irq_handler(struct trace_event_raw_irq_handler_entry *ctx)
+{
+	struct irq_key key = {};
+	struct info *info;
+
+	bpf_probe_read_kernel_str(&key.name, sizeof(key.name), ctx->__data);
+	info = bpf_map_lookup_or_try_init(&infos, &key, &zero);
+	if (!info)
+		return 0;
+	info->count += 1;
+	return 0;
+}
+
+SEC("tp_btf/irq_handler_entry")
+int BPF_PROG(irq_handler_entry)
+{
+	u64 ts = bpf_ktime_get_ns();
+	u32 key = 0;
+
+	bpf_map_update_elem(&start, &key, &ts, 0);
+	return 0;
+}
+
+SEC("tp_btf/irq_handler_exit")
+int BPF_PROG(irq_handler_exit_exit, int irq, struct irqaction *action)
+{
+	struct irq_key ikey = {};
+	struct info *info;
+	u32 key = 0;
+	s64 delta;
+	u64 *tsp;
+
+	tsp = bpf_map_lookup_elem(&start, &key);
+	if (!tsp || !*tsp)
+		return 0;
+
+	delta = bpf_ktime_get_ns() - *tsp;
+	if (delta < 0)
+		return 0;
+	if (!targ_ns)
+		delta /= 1000U;
+
+	bpf_probe_read_kernel_str(&ikey.name, sizeof(ikey.name), action->name);
+	info = bpf_map_lookup_or_try_init(&infos, &ikey, &zero);
+	if (!info)
+		return 0;
+
+	if (!targ_dist) {
+		info->count += delta;
+	} else {
+		u64 slot;
+
+		slot = log2(delta);
+		if (slot >= MAX_SLOTS)
+			slot = MAX_SLOTS - 1;
+		info->slots[slot]++;
+	}
+
+	return 0;
+}
+
+char LICENSE[] SEC("license") = "GPL";
--- a/libbpf-tools/hardirqs.c
+++ b/libbpf-tools/hardirqs.c
@@ -1,0 +1,269 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+// Copyright (c) 2020 Wenbo Zhang
+//
+// Based on hardirq(8) from BCC by Brendan Gregg.
+// 31-Aug-2020   Wenbo Zhang   Created this.
+#include <argp.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+#include <unistd.h>
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+#include "hardirqs.h"
+#include "hardirqs.skel.h"
+#include "trace_helpers.h"
+
+struct env {
+	bool count;
+	bool distributed;
+	bool nanoseconds;
+	time_t interval;
+	int times;
+	bool timestamp;
+	bool verbose;
+} env = {
+	.interval = 99999999,
+	.times = 99999999,
+};
+
+static volatile bool exiting;
+
+const char *argp_program_version = "hardirqs 0.1";
+const char *argp_program_bug_address = "<bpf@vger.kernel.org>";
+const char argp_program_doc[] =
+"Summarize hard irq event time as histograms.\n"
+"\n"
+"USAGE: hardirqs [--help] [-T] [-N] [-d] [interval] [count]\n"
+"\n"
+"EXAMPLES:\n"
+"    hardirqs            # sum hard irq event time\n"
+"    hardirqs -d         # show hard irq event time as histograms\n"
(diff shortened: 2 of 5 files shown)


ethercflow:
BTW, I found it seems that we can only define one handler for an event right? 
Eg:  I want to define  
```c
SEC("tp_btf/irq_handler_entry")
int BPF_PROG(handler1) {}

SEC("tp_btf/irq_handler_entry")
int BPF_PROG(handler2) {}
```

at the same time, then use `env.xxx` to determine
```
bpf_program__attach(obj->progs.handler1);
```
or
```
bpf_program__attach(obj->progs.handler2);
```

I found  `skeleton` only generate `handler1` , Is this intentional?

anakryiko:
> Hi @anakryiko, another tool PTAL when you have time, thank you. :)
> BTW, I found it seems that we can only define one handler for an event right?

Right now libbpf supports only one, yes. [This patch set](https://patchwork.ozlabs.org/project/netdev/list/?series=198728&state=*) adds support for it. Once it lands and we update libbpf you should be able to do this easily.

> 
> I found `skeleton` only generate `handler1` , Is this intentional?

It's the consequence of lack of libbpf support.

libbpf-tools/hardirqs.bpf.c
```diff
@@ -0,0 +1,107 @@
...
+
+static struct hist zero_hist;
+
+SEC("tracepoint/irq/irq_handler_entry")
+int handle__irq_handler(struct trace_event_raw_irq_handler_entry *ctx)
+{
+	struct irq_key key = {};
+	u64 *count, zero = 0;
+
+	bpf_probe_read_kernel_str(&key.name, sizeof(key.name), ctx->__data);
```
anakryiko, line 46: why using symbolic name as a key? It's inconvenient and slow. There seems to be `irq` int field, use that for key look up?

libbpf-tools/hardirqs.bpf.c
```diff
@@ -0,0 +1,107 @@
...
+	count = bpf_map_lookup_or_try_init(&irqs, &key, &zero);
+	if (!count)
+		return 0;
+	*count += 1;
```
anakryiko, line 50: should be atomic?

libbpf-tools/hardirqs.bpf.c
```diff
@@ -0,0 +1,107 @@
...
+
+	bpf_probe_read_kernel_str(&ikey.name, sizeof(ikey.name), action->name);
+
+	if (!targ_dist) {
+		u64 *elapsed, zero = 0;
+
+		elapsed = bpf_map_lookup_or_try_init(&irqs, &ikey, &zero);
+		if (!elapsed)
+			return 0;
+		*elapsed += delta;
```
anakryiko, line 90: atomic?

libbpf-tools/hardirqs.c
```diff
@@ -0,0 +1,293 @@
...
+static int print_count_or_latency(int fd)
+{
+	struct irq_key lookup_key = {}, next_key;
+	__u64 count;
+	int err;
+
+	if (env.count) {
+		printf("%-26s %11s\n", "HARDIRQ", "TOTAL_count");
+	} else {
+		char *units = env.nanoseconds ? "nsecs" : "usecs";
```
anakryiko, line 125: const char *?

libbpf-tools/hardirqs.c
```diff
@@ -0,0 +1,293 @@
...
+			tm = localtime(&t);
+			strftime(ts, sizeof(ts), "%H:%M:%S", tm);
+			printf("%-8s\n", ts);
+		}
+
+		if (env.distributed)
+			err = print_hist(bpf_map__fd(obj->maps.hists));
+		else
+			err = print_count_or_latency(
+				bpf_map__fd(obj->maps.irqs));
```
anakryiko, line 280: nit: pass `struct bpf_map *` as arguments to those functions. It will be shorter here and less error-prone, as you can't pass any random integer anymore.

libbpf-tools/hardirqs.bpf.c
```diff
@@ -0,0 +1,107 @@
...
+	__type(value, u64);
+} start SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_ENTRIES);
+	__type(key, struct irq_key);
+	__type(value, struct hist);
+	__uint(map_flags, BPF_F_NO_PREALLOC);
+} hists SEC(".maps");
```
anakryiko, line 36: have you considered combining `hists` and `irqs` maps into one? In counting mode you wouldn't use histogram part. It might be a bit slower on user-space side, because you'll be reading entire map value for each entry, but I wonder if it makes any difference. But keeping just one map definitely would save some memory.

ethercflow:
Thanks, I'll fix these soon.

libbpf-tools/hardirqs.bpf.c
ethercflow: > why using symbolic name as a key

Because I think irq cannot distinguish shared interrupts，such as (cp fromFrom http://static.lwn.net/images/pdf/LDD3/ch10.pdf ):

> All the handlers installed for the same interrupt number appear on the same line of
> /proc/interrupts. The following output (from an x86_64 system) shows how shared
> interrupt handlers are displayed:
>  CPU0
>  0: 892335412 XT-PIC timer
>  1: 453971 XT-PIC i8042
>  2: 0 XT-PIC cascade
>  5: 0 XT-PIC libata, ehci_hcd
>  8: 0 XT-PIC rtc
>  9: 0 XT-PIC acpi
>  10: 11365067 XT-PIC ide2, uhci_hcd, uhci_hcd, SysKonnect SK-98xx, EMU10K1
>  11: 4391962 XT-PIC uhci_hcd, uhci_hcd
>  12: 224 XT-PIC i8042
>  14: 2787721 XT-PIC ide0
>  15: 203048 XT-PIC ide1
> NMI: 41234
> LOC: 892193503
> ERR: 102
> MIS: 0
> This system has several shared interrupt lines. IRQ 5 is used for the serial ATA and
> IEEE 1394 controllers; IRQ 10 has several devices, including an IDE controller, two
> USB controllers, an Ethernet interface, and a sound card; and IRQ 11 also is used by
> two USB controllers.

libbpf-tools/hardirqs.bpf.c
ethercflow: In my mind, only one instance of the irq handler would be running on one of the processes.  So I think we don't need atomic?

libbpf-tools/hardirqs.bpf.c
ethercflow: same above

libbpf-tools/hardirqs.c
ethercflow: Thanks, will fix it after make sure the aboves

libbpf-tools/hardirqs.c
ethercflow: Thanks, will fix it after make sure the aboves

libbpf-tools/hardirqs.bpf.c
ethercflow: Thanks, I'll merge it and give the bench result of two. :)

ethercflow: combined，thanks.

libbpf-tools/hardirqs.c
ethercflow: Fixed, thanks!

libbpf-tools/hardirqs.c
ethercflow: Fixed, thanks!

ethercflow:
Hi @anakryiko! Thanks for the review, pushed fixes for your comments, except key type and atomic :)

libbpf-tools/hardirqs.bpf.c
anakryiko: oh, ok, that makes sense, thanks!

libbpf-tools/hardirqs.bpf.c
anakryiko: I'm not an expert on how IRQs work and how they are handled in Linux, so don't know. But my assumption was that each CPU can handle them independently. But again, no idea.

libbpf-tools/hardirqs.bpf.c
```diff
@@ -0,0 +1,94 @@
...
+	__type(key, u32);
+	__type(value, u64);
+} start SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_ENTRIES);
+	__type(key, struct irq_key);
+	__type(value, struct info);
+	__uint(map_flags, BPF_F_NO_PREALLOC);
```
anakryiko, line 27: with MAX_ENTRIES = 256, there is probably no need for BPF_F_NO_PREALLOC. BPF_F_NO_PREALLOC is quite slower, so I'd suggest not using it here.

anakryiko (approved):
LGTM.

libbpf-tools/hardirqs.bpf.c
ethercflow: Got it, thanks!

ethercflow:

yonghong-song:
[buildbot, test this please]

PR #3085, opened 2020-09-03
memleak: fix false positive on failed allocations

amir73il:
memleak tool reports a failed allocation as a leak,
because the NULL address is never freed.

Signed-off-by: Amir Goldstein <amir73il@gmail.com>


yonghong-song:
[buildbot, test this please]

PR #3092, opened 2020-09-10
examples/tracing: add usb_buffer - Monitor USB transfer

gpioblink:
This script lets you know the contents of the buffer the USB is sending and receiving.


brendangregg:
examples/tracing are for code examples for others to read and study, mostly for demonstrating BCC capabilities to new programmers, but can also demonstrate common targets of tracing. This has no comments, and the lesson isn't clear. It is short, which is good for examples/tracing.

I'm still not sure about it, but if you added some comments I think it would be a better proposal. E.g., have a header line that reads something like:

```
# This is an example of USB event tracing, showing important members from USB structs.
```

And then add more comments in the code, something like:

```
// Instrument usb_hcd_giveback_urb() because ... _insert reason here_
 int kprobe____usb_hcd_giveback_urb(struct pt_regs *ctx, struct urb *urb) {
[...]
// fetch details from struct urb and its members:
struct usb_device *dev = urb->dev;
[...]
```

To help newcomers who are looking in examples/tracing.

If this program is useful for some real-world issues, then maybe it should be tools/usbsnoop. But there'd have to be some good real-world use cases for it; otherwise it can stay as an example for now.

BTW, did you try bpftrace? E.g.:

```
# bpftrace --include linux/usb.h -e 'k:usb_hcd_giveback_urb {
    $urb = (struct urb *)arg0; printf("vendor: 0x%x, product: 0x%x\n", $urb->dev->descriptor.idVendor, $urb->dev->descriptor.idProduct); }'
Attaching 1 probe...
vendor: 0x0, product: 0x0
vendor: 0x0, product: 0x0
```

gpioblink:
Thank you for your response!

I wanted to create this example to show how easy it is to replicate a USBPcap-like feature in BCC.
This is a example that displays the input and output buffers of a USB.

I've added a few explanations. I hope you will check it out 🙇‍♂️

yonghong-song:
cc @brendangregg

PR #3098, opened 2020-09-15
tool: add readahead bcc tool

tuxology:
This tool provides two read-ahead cache statistics to understand the read-ahead cache performance for the system:

-  Number of pages in the cache which are left unaccessed
-  Age of pages in the cache along with their count presented as a histogram 

This tool is based on the `readahead` example in @brendangregg 's BPF Performance Tools Book.

### DIFF
--- a/tools/readahead_example.txt
+++ b/tools/readahead_example.txt
@@ -1,0 +1,60 @@
+Demonstration of readahead, the Linux eBPF/bcc version
+
+Read-ahead mechanism is used by operation sytems to optimize sequential operations
+by reading ahead some pages to avoid more expensive filesystem operations. This tool
+shows the performance of the read-ahead caching on the system under a given load to 
+investigate any caching issues. It shows a count for unused pages in the cache and
+also prints a histogram showing how long they have remianed there.
+
+Usage Scenario
+==============
+
+Consider that you are developing a React Native application which performs aggressive 
+reads while re-encoding a video in local-storage. Usually such an app would be multi-
+layered and have transitional library dependencies. The actual read may be performed 
+by some unknown native library which may or may not be using hints to the OS, such as  
+madvise(p, LEN, MADV_SEQUENTIAL). If high IOPS is observed in such an app, running 
+readahead may pin the issue much faster in this case as the developer digs deeper 
+into what may be causing this. 
+
+An example where such an issue can surface is: https://github.com/boltdb/bolt/issues/691
+
+# readahead -d 30
+Tracing... Hit Ctrl-C to end.
+^C
+Read-ahead unused pages: 6765
+Histogram of read-ahead used page age (ms):
+
+     age (ms)            : count     distribution
+         0 -> 1          : 4236     |****************************************|
+         2 -> 3          : 394      |***                                     |
+         4 -> 7          : 1670     |***************                         |
+         8 -> 15         : 2132     |********************                    |
+        16 -> 31         : 401      |***                                     |
+        32 -> 63         : 1256     |***********                             |
+        64 -> 127        : 2352     |**********************                  |
+       128 -> 255        : 357      |***                                     |
+       256 -> 511        : 369      |***                                     |
+       512 -> 1023       : 366      |***                                     |
+      1024 -> 2047       : 181      |*                                       |
+      2048 -> 4095       : 439      |****                                    |
+      4096 -> 8191       : 188      |*                                       |
+
+In the example above, we recorded system-wide stats for 30 seconds. We can observe that 
+while most of the pages stayed in the readahead cache for quite less time, after 30
+seconds 6765 pages still remained in the cache, yet unaccessed.
+
+Note on Kprobes Usage
+=====================
+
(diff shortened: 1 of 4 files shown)


tools/readaheadstat.py
```diff
@@ -0,0 +1,122 @@
...
+        birth.delete(&arg0); // remove the entry from hashmap
+    }
+    return 0;
+}
+"""
+
+b = BPF(text=program)
+b.attach_kprobe(event="__do_page_cache_readahead", fn_name="entry__do_page_cache_readahead")
+b.attach_kretprobe(event="__do_page_cache_readahead", fn_name="exit__do_page_cache_readahead")
+b.attach_kretprobe(event="__page_cache_alloc", fn_name="exit__page_cache_alloc")
```
yonghong-song, line 100: The same for all other kprobe based tracing. These functions may be inlined so the result from the tool may not represent what really happens in the system. I think this should be spelled out in the example and man page as warning, so if people double or cannot interpret their results, they should check vmlinux source and binary to confirm whether inlining happens or not. Analyzing dwarf probably can get similar information.

tools/readaheadstat_example.txt
```diff
@@ -0,0 +1,32 @@
+Demonstration of readaheadstat, the Linux eBPF/bcc version
+
+Read-ahead mechanism is used by operation sytems to optimize sequential operations
+by reading ahead some pages to avoid more epensive filesystem operations. This tool
+shows the performance of the read-ahead caching on the system under a given load to 
+investigate any caching issues. It collects stats for unused pages in the cache and
+also prints a histogram showing how long they have remianed there.
```
yonghong-song, line 7: I guess we need more explanation about *when* people should use this tool. What kind of performance issues may trigger people to use this tool to check readahead performance.

yonghong-song (commented):
cc @brendangregg

tools/readaheadstat.py
tuxology: Added notes in man page and example

tools/readaheadstat_example.txt
tuxology: Explained this with a usage scenario as well

brendangregg:
I wanted to refer to the style of other ports of book tools, but it seems like I wrote the code but never did the PR. Let me do the PR so we have it to refer to!

tuxology:
@brendangregg So if this is still required, let me know and I can conform to the style of other ports but if you feel this tool is not needed in bcc, we can close this PR

brendangregg:
It is actually an important tool, and I demoed it in my keynote at LSFMM. I think it's ok to include. I'd call it readahead, not readaheadstat, since it is a port of readahead (right?) and also "stat" tools by convention are for per-interval summary statistics (vmstat/iostat).

I'd like the man page to mention the BPF book as well, like in the other tools I did the PR for recently.

tuxology:
@brendangregg 

> It is actually an important tool, and I demoed it in my keynote at LSFMM. I think it's ok to include. I'd call it readahead, not readaheadstat, since it is a port of readahead (right?) and also "stat" tools by convention are for per-interval summary statistics (vmstat/iostat).

Understood. Yes, it is now a proper port of `readahead`. So, we actually started building the bcc version from scratch as a part of a conference demo that included live graphs with snapshots at regular intervals (and hence `stat`). We quickly discovered that you had already built the bpftrace version as part of the book. So I then used that as a reference and created this port. I will rename this to `readahead` and will now reference your original work in the code comments and man pages.

> I'd like the man page to mention the BPF book as well, like in the other tools I did the PR for recently.

Yes, will do that as well. 

Btw, I also created a small graphic for this tool to explain it. Maybe someone can use it sometime as a learning material if they reach this comment:

image

tuxology:
@brendangregg its ready. I squashed previous commits to make the git message more relevant

yonghong-song:
[buildbot, test this please]

yonghong-song (commented):
The file readahead.py is not an executable.
```
-bash-4.4$ ls -l readahead.py
-rw-r--r--. 1 yhs users 3674 Sep 25 10:47 readahead.py
-bash-4.4$
```
Could you make it an executable?

tools/readahead_example.txt
```diff
@@ -0,0 +1,60 @@
+Demonstration of readahead, the Linux eBPF/bcc version
+
+Read-ahead mechanism is used by operation sytems to optimize sequential operations
+by reading ahead some pages to avoid more epensive filesystem operations. This tool
```
yonghong-song, line 4: epensive => expensive

yonghong-song:
[buildbot, ok to test]

tools/readahead_example.txt
tuxology: Fixed

tuxology:
@yonghong-song I've made the changes you suggested in latest update

yonghong-song:
LGTM. Thanks!

brendangregg:
Thanks Suchakra!

PR #3105, opened 2020-09-24
tools/tcpstates: remove unnecessary TCP checking

boat0:


yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks for the fix. Make sense. tcp_set_state kernel function only accepts tcp sockets :-)

Issue #2955, opened 2020-06-06
error: field has incomplete type 'struct memcg_cache_params'


Daniel15:
I installed this on Debian Bullseye using:
```
apt install bpfcc-tools
```

but when I run `slabratetop-bpfcc`, I get this error:
```
In file included from /virtual/main.c:6:
include/linux/slub_def.h:112:28: error: field has incomplete type 'struct memcg_cache_params'
        struct memcg_cache_params memcg_params;
                                  ^
include/linux/slub_def.h:112:9: note: forward declaration of 'struct memcg_cache_params'
        struct memcg_cache_params memcg_params;
               ^
1 error generated.
Traceback (most recent call last):
  File "/usr/sbin/slabratetop-bpfcc", line 109, in <module>
    b = BPF(text=bpf_text)
  File "/usr/lib/python3/dist-packages/bcc/__init__.py", line 347, in __init__
    raise Exception("Failed to compile BPF module %s" % (src_file or "<text>"))
Exception: Failed to compile BPF module <text>
```

Any ideas?

```
root@lux01:~# cat /etc/debian_version
bullseye/sid

root@lux01:~# uname -a
Linux lux01 5.6.0-2-cloud-amd64 #1 SMP Debian 5.6.14-1 (2020-05-23) x86_64 GNU/Linux
```

yonghong-song:
Yes, this is a problem with 5.6 kernel where `struct memcg_cache_params` is defined in `linux/mm/slab.h`.  This can be fixed by copying related structures to the bpf program, e.g., `runqlen.py`. If you are interested, please go ahead to try and submit a pull request. Otherwise, I will find some time later to fix it.

PR #3109, opened 2020-09-28
slabratetop: Add memcg_cache_params struct def

dwrank:
The following kernel commit moves struct memcg_cache_params from
 include/linux/slab.h to mm/slab.h.

    commit 9adeaa226988b97bc15928e12f40a9863134467c
    Author: Waiman Long <longman@redhat.com>
    Date:   Mon Sep 23 15:33:49 2019 -0700

        mm, slab: move memcg_cache_params structure to mm/slab.h

        The memcg_cache_params structure is only embedded into the kmem_cache of
        slab and slub allocators as defined in slab_def.h and slub_def.h and used
        internally by mm code.  There is no needed to expose it in a public
        header.  So move it from include/linux/slab.h to mm/slab.h.  It is just a
        refactoring patch with no code change.

        ...

The commit is in kernel v5.4, causing a compiler error when including
slub_def.h or slab_def.h in slabratetop's bpf program.

Add an empty definition of the struct in slabratetop's bpf
program so it will compile with kernel version >= v5.4.
The actual struct definition is not required for slabratetop's bpf program.

Tested with kernel v5.8.8 and v5.3.7.

Fixes: https://github.com/iovisor/bcc/issues/2955

Signed-off-by: Daniel Rank dwrank@gmail.com

### DIFF
--- a/tools/slabratetop.py
+++ b/tools/slabratetop.py
@@ -62,7 +62,13 @@ def signal_ignore(signal, frame):
 bpf_text = """
 #include <uapi/linux/ptrace.h>
 #include <linux/mm.h>
-#include <linux/slab.h>
+
+// memcg_cache_params is a part of kmem_cache, but is not publicly exposed in
+// kernel versions 5.4 to 5.8.  Define an empty struct for it here to allow the
+// bpf program to compile.  It has been completely removed in kernel version
+// 5.9, but it does not hurt to have it here for versions 5.4 to 5.8.
+struct memcg_cache_params {};
+
 #ifdef CONFIG_SLUB
 #include <linux/slub_def.h>
 #else


yonghong-song:
[buildbot, test this please]

yonghong-song:
I cannot reproduce the issue in my 5.6 based machine. Could you show the error message? The structure `memcg_cache_params` seems not used in any structures in linux/include/linux/.... Also, if this is indeed the case only pointer of `memcg_cache_params` is used, maybe a forward declaration is enough?

dwrank:
I tried again with kernel 5.6.6, with the same results.
I tried with a forward declaration, but I get the same error.  

Here is the error, the same reported in https://github.com/iovisor/bcc/issues/2955:

    $ sudo slabratetop
    In file included from /virtual/main.c:7:
    include/linux/slub_def.h:102:28: error: field has incomplete type 'struct memcg_cache_params'
            struct memcg_cache_params memcg_params;
                                      ^
    include/linux/slub_def.h:102:9: note: forward declaration of 'struct memcg_cache_params'
            struct memcg_cache_params memcg_params;
                   ^
    1 error generated.
      File "/usr/share/bcc/tools/slabratetop", line 110, in <module>
        b = BPF(text=bpf_text)
      File "/usr/lib/python3.8/site-packages/bcc/__init__.py", line 364, in __init__

dwrank:
@yonghong-song It looks like memcg_cache_params got removed from include/linux/{slub_def.h, slab_def.h} and mm/slab.h with the following commit in kernel v5.9-rc1, which should avoid the issue, but it may be good to have the empty struct def for versions >= 5.4 and < 5.9.

    commit 10befea91b61c4e2c2d1df06a2e978d182fcf792
    Author: Roman Gushchin <guro@fb.com>
    Date:   Thu Aug 6 23:21:27 2020 -0700

        mm: memcg/slab: use a single set of kmem_caches for all allocations

        Instead of having two sets of kmem_caches: one for system-wide and
        non-accounted allocations and the second one shared by all accounted
        allocations, we can use just one.

        ...

yonghong-song:
> @yonghong-song It looks like memcg_cache_params got removed from include/linux/{slub_def.h, slab_def.h} and mm/slab.h with the following commit in kernel v5.9-rc1, which should avoid the issue, but it may be good to have the empty struct def for versions >= 5.4 and < 5.9.
> 
> ```
> commit 10befea91b61c4e2c2d1df06a2e978d182fcf792
> Author: Roman Gushchin <guro@fb.com>
> Date:   Thu Aug 6 23:21:27 2020 -0700
> 
>     mm: memcg/slab: use a single set of kmem_caches for all allocations
> 
>     Instead of having two sets of kmem_caches: one for system-wide and
>     non-accounted allocations and the second one shared by all accounted
>     allocations, we can use just one.
> 
>     ...
> ```

Thanks for clarification. My 5.6 kernel with backport including this commit, that is why it does not have issues. I checked out 5.7 stable and did see the issue.

Your change looks good. I cannot find a much better way. The best could be like
```
#ifndef __struct_type_exist(memcg_cache_params) 
struct memcg_cache_params {};
#endif 
```
But unfortunately, we do not have the above builtin macro yet.

Could you update the comments to say that this work is needed for
5.4 - 5.8 and 5.9 does not need this workaround but didn't hurt? Thanks!

yonghong-song:
[buildbot, ok to test]

dwrank:
@yonghong-song Amended the commit to update the code and commit comments.

PR #3111, opened 2020-09-29
Fix tcpdrop v6 source (remote) and dest (local) address

nspring:
For v6 connections, tcpdrop.py would report the source and destination addresses swapped - tcp_drop() covers the input path, where the source of the received packet is the daddr stored in the socket.  The destination of the received packet is the local address. This commit swaps them to be correct and leaves a comment since it is not obvious (they can be removed).

### DIFF
--- a/tools/tcpdrop.py
+++ b/tools/tcpdrop.py
@@ -128,10 +128,12 @@ int trace_tcp_drop(struct pt_regs *ctx, struct sock *sk, struct sk_buff *skb)
         struct ipv6_data_t data6 = {};
         data6.pid = pid;
         data6.ip = 6;
+        // The remote address (skc_v6_daddr) was the source
         bpf_probe_read_kernel(&data6.saddr, sizeof(data6.saddr),
-            sk->__sk_common.skc_v6_rcv_saddr.in6_u.u6_addr32);
-        bpf_probe_read_kernel(&data6.daddr, sizeof(data6.daddr),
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
cc @iamkafai could you help take a look as well? Thanks!

iamkafai:
The change makes sense to me.

It will be easier to read going forward if both v4 and v6 use the same way to avoid similar confusion.  (i.e. either both get from the ip[46] hdr or both get from the sk)

nspring:
@iamkafai Getting it from the header for v6 is not straightforward - on my test machine I think I'm seeing v4 packets (via tcp_v4_do_rcv) being delivered to v6 sockets that use ::ffff:127.0.0.1 as the address.   I'll see if I can get v4 addresses from the sock, though I would usually avoid touching the well-tested not-buggy code path.

iamkafai:
No problem for doing it later.  Thanks for explaining the situation o n v6 hdr. 

I think the v4 address should be in skc_daddr and skc_rcv_saddr of sock_common.  However, the TCP_LISTEN sk probably does not have the daddr but I guess tcp_drop won't be called with TCP_LISTEN so far, not sure.

yonghong-song:
Thanks for you both. I think we can do another patch if getting ip v4 address from packet is not always correct.

PR #3112, opened 2020-09-29
tools/biolatency: Handle signals from user

kcalvinalvin:
Rationale: `KeyboardInterrupt` only handles interrupts from the keyboard. This results in biolatency not printing anything during exit when given a signal that's not from the keyboard. This doesn't play well with commands such as `timeout` https://man7.org/linux/man-pages/man1/timeout.1.html


yonghong-song:
[buildbot, test this please]

brendangregg:
Most tools, including biolatency, already support a duration. You want biolatency to run for 5 seconds? Then it already supports it:

```
biolatency 5 1
```

This PR has changed the code to support an external tool imposing durations. Did you not read the man page?

This PR seems bad to me. Yes, I know there are situations where external tools might send other signals (SIGTERM, SIGQUIT), but are we going to clutter EVERY tool in this way for something that might never happen?

kcalvinalvin:
> This PR has changed the code to support an external tool imposing durations.

This PR has changed the code to handle signals. KeyboardInterrupt is strictly a python built in exception.

> Did you not read the man page?

I have read the man page.

> This PR seems bad to me. Yes, I know there are situations where external tools might send other signals (SIGTERM, SIGQUIT), but are we going to clutter EVERY tool in this way for something that might never happen?

I changed biolatency since it fitted my use-case and made a PR since I thought it may be useful for others. It's a very minimal code change that handles extra use cases. In my view, I found it to be a more correct code.

I don't think the criticism here is valid since it's a basis on future actions that is not a scope of this PR.

brendangregg:
> I don't think the criticism here is valid since it's a basis on future actions that is not a scope of this PR.

Everything we allow sets a precedent. This PR sets a precedent that we will add code to support timeout, in addition to the tool's own support for durations. That's duplication and code clutter. I think we should revert this change to avoid setting a bad precedent.

I'll consider other use cases for this change (this PR only mentions timeout) if you would like to explain.

yonghong-song:
Okay, let me revert the change. I agree that we should get consensus on how to handle external signals/events for tools, so we can maintain consistency across tools.

PR #3115, opened 2020-10-01
Revert "tools/biolatency: Handle signals from user"

yonghong-song:
This reverts commit 87792ce6b783941ea1d8bcd29812a55520930c95.

Let us have a consensus about how bcc tools should interact with
external signals etc. before applying such changes to this tool
and other tools.


brendangregg:
thanks!

PR #3123, opened 2020-10-06
add latency tracing for bio-based block devices

alekseymmm:
The current behavior of biolatency.py is that it traces the submission and completion of requests in corresponding functions. Thus it only works for request-based block devices. This patch adds latency tracing for bio-based block devices. It is useful for Linux device mapper monitoring (like mdraid and lvm are common examples).
I tested it on 
* mdraid raids
* lvm volumes
* null blk block devices created with queue_mode=0 (bio based)

statistics (collected with -D for example) looks like 
```
disk = b'dm-0'
       256 -> 511        : 2        |*                                       |
      2048 -> 4095       : 9        |********                                |
      4096 -> 8191       : 45       |****************************************|
      8192 -> 16383      : 18       |****************                        |
```

### DIFF
--- a/tools/biolatency.py
+++ b/tools/biolatency.py
@@ -43,6 +43,8 @@ parser.add_argument("interval", nargs="?", default=99999999,
     help="output interval, in seconds")
 parser.add_argument("count", nargs="?", default=99999999,
     help="number of outputs")
+parser.add_argument("-b", "--bio", action="store_true",
+    help="trace bio submition and complition")
 parser.add_argument("--ebpf", action="store_true",
     help=argparse.SUPPRESS)
 args = parser.parse_args()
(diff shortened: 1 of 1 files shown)


brendangregg:
Are you sure this isn't going to double count some I/O?

alekseymmm:
@brendangregg  Oh yes, You are right. I fixed this by moving bio tracing to a separate flag (-b or --bio) of biolatency, and request tracing as a default. So only one kind of probes (req or bio) is active at the same time.
What do you think about this approach?

yonghong-song:
[buildbot, test this please]

yonghong-song:
I found the following comments for `bio_endio`:
```
/**
 * bio_endio - end I/O on a bio
 * @bio:        bio
 *
 * Description:
 *   bio_endio() will end I/O on the whole bio. bio_endio() is the preferred
 *   way to end I/O on a bio. No one should call bi_end_io() directly on a
 *   bio unless they own it and thus know that it has an end_io function.
 *
 *   bio_endio() can be called several times on a bio that has been chained
 *   using bio_chain().  The ->bi_end_io() function will only be called the
 *   last time.  At this point the BLK_TA_COMPLETE tracing event will be
 *   generated if BIO_TRACE_COMPLETION is set.
 **/
void bio_endio(struct bio *bio)
```
How you do handle chained bio where one bio submit followed by
multiple bio_endio? Do we have a way to tell whether a bio_endio
is the last one in the chain?

yonghong-song:
since a new option is added, please change example file and man page as well.

PR #3127, opened 2020-10-08
tools: Use correct key key when deleting item from map

mauriciovasquezbernal:
The correct key to use is "tig" instead of "pid".


yonghong-song:
[buildbot, test this please]

PR #3133, opened 2020-10-13
set larger range to get probes evnet for test_usdt2.py

xuchunmei000:
I test test_usdt2.py on aarch64 with kernel-5.9-rc5,
gcc-9.2.1 and glibc2.30.
with the current range of 3, always get failure:
FAIL: test_attach1 (__main__.TestUDST)
  File "test_usdt2.py", line 170, in test_attach1
    self.assertTrue(self.evt_st_6 == 1)
AssertionError: False is not true

Ran 1 test in 1.068s

FAILED (failures=1)

Signed-off-by: Chunmei Xu <xuchunmei@linux.alibaba.com>


yonghong-song:
[buildbot, test this please]

PR #3134, opened 2020-10-14
Update installation instructions for Arch Linux

sam-lunt:
`bcc`, `bcc-tools`, and `python-bcc` were recently moved from the AUR to the official Arch repositories, so the installation instructions need to be updated.

I also removed the note about upgrading to kernel 4.3.1-1. The `linux-lts` package was upgraded from 4.1.21 to 4.4.7 by [this commit](https://github.com/archlinux/svntogit-packages/commit/20d5e134c5154161886d4f049811c9b30cd6544b#diff-3e341d2d9c67be01819b25b25d5e53ea3cdf3a38d28846cda85a195eb9b7203a) in April 2016. Considering that `linux-lts` trails behind the other Arch kernel packages (`linux`, `linux-zen`, `linux-hardened`), I think that any Arch installation that is still in use can be assumed to be on a kernel later than 4.3.1

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -4,7 +4,7 @@
 * [Packages](#packages)
   - [Ubuntu](#ubuntu---binary)
   - [Fedora](#fedora---binary)
-  - [Arch](#arch---aur)
+  - [Arch](#arch---binary)
   - [Gentoo](#gentoo---portage)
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks!

PR #3135, opened 2020-10-15
uprobe: Add ref_cnt_offset arg to bpf_attach_uprobe

danobi:
This argument allows callers to tell the kernel to manage USDT semaphore
counts. It's better than fiddling with the counts in userspace b/c if
the userspace program crashes then the semaphore count does not get
decremented.

It's also better b/c the kernel can activate the semaphore for all
processes that have the target mapped into memory. Currently with bcc,
you have to provide the pid of each process you want to activate a
semaphore for and then bcc will go poke /proc/pid/mem at the right
offset.

### DIFF
--- a/src/cc/api/BPF.cc
+++ b/src/cc/api/BPF.cc
@@ -225,7 +225,8 @@ StatusTuple BPF::attach_uprobe(const std::string& binary_path,
                                const std::string& probe_func,
                                uint64_t symbol_addr,
                                bpf_probe_attach_type attach_type, pid_t pid,
-                               uint64_t symbol_offset) {
+                               uint64_t symbol_offset,
+                               uint32_t ref_ctr_offset) {
 
   if (symbol_addr != 0 && symbol_offset != 0)
     return StatusTuple(-1,
@@ -245,7 +246,8 @@ StatusTuple BPF::attach_uprobe(const std::string& binary_path,
   TRY2(load_func(probe_func, BPF_PROG_TYPE_KPROBE, probe_fd));
 
   int res_fd = bpf_attach_uprobe(probe_fd, attach_type, probe_event.c_str(),
-                                 binary_path.c_str(), offset, pid);
+                                 binary_path.c_str(), offset, pid,
+                                 ref_ctr_offset);
 
   if (res_fd < 0) {
     TRY2(unload_func(probe_func));
--- a/src/cc/api/BPF.cc
+++ b/src/cc/api/BPF.cc
@@ -20,9 +20,12 @@
 #include <cstdio>
 #include <cstring>
 #include <exception>
+#include <fcntl.h>
 #include <iostream>
 #include <memory>
 #include <sstream>
+#include <sys/stat.h>
+#include <sys/types.h>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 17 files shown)


danobi:
cc @liu-song-6 

I've been testing the kernel feature out in https://github.com/danobi/bpftrace/tree/uprobe_refcnt and it seems something's broken.
The perf event attaches correctly and the uprobe triggers the bpf prog to run. But the semaphore doesn't seem to be incremented.

Here's the strace output I collected:
```
$ sudo strace -e trace=perf_event_open -e abbrev=none ../src/bpftrace -e 'usdt::tracetest:testprobe { printf("%s\n", str(arg1) ); exit(); }' -p $(pidof usdt_semaphore_test)
[...]
perf_event_open({type=0x7 /* PERF_TYPE_??? */, size=PERF_ATTR_SIZE_VER6, config=0x40404000000000, sample_period=1, sample_type=0, read_format=0, disabled=0, inherit=0, pinned=0, exclusive=0, exclusive_user=0, exclude_kernel=0, exclude_hv=0, exclude_idle=0, mmap=0, comm=0, freq=0, inherit_stat=0, enable_on_exec=0, task=0, watermark=0, precise_ip=0 /* arbitrary skid */, mmap_data=0, sample_id_all=0, exclude_host=0, exclude_guest=0, exclude_callchain_kernel=0, exclude_callchain_user=0, mmap2=0, comm_exec=0, use_clockid=0, context_switch=0, write_backward=0, namespaces=0, wakeup_events=1, config1=0x5638a47f1430, config2=0x1198, sample_regs_user=0, sample_regs_intr=0, aux_watermark=0, sample_max_stack=0, aux_sample_size=0}, 843309, -1, -1, PERF_FLAG_FD_CLOEXEC) = 10
tracetest_testprobe_semaphore: 0
```

`config=0x40404000000000` looks right to me. Top 32 bits are 0x404040. My test program has a semaphore at that address:
```
$ readelf -n ./testprogs/usdt_semaphore_test
[...]
    Provider: tracetest
    Name: testprobe
    Location: 0x0000000000401198, Base: 0x000000000040202b, Semaphore: 0x0000000000404040
```

I'm printf debugging the kernel right now but was hoping you might have an idea what could be going wrong.

liu-song-6:
I guess the problem is we passed in "address of semaphore" instead of "offset of semaphore". 

If after fixing that, it still doesn't work. Could you please try whether the text based interface works? Something like 

`echo "p:some/uprobe ./testprogs/usdt_semaphore_test:<uprobe_offset>(<semaphore_offset>)" > /sys/kernel/debug/tracing/uprobe_events `

danobi:
I'm trying to follow the tutorial at https://www.kernel.org/doc/html/latest/trace/uprobetracer.html but I can't figure out how to enable the event:
```
[root@maharaja tracing]# cat uprobe_events
p:uprobes/p_usdt_0x401198 /home/daniel/dev/bpftrace/build/tests/testprogs/usdt_semaphore_test:0x0000000000401198(0x404040) arg1=%ip

[root@maharaja tracing]# echo 1 > events/uprobes/enable
bash: echo: write error: Invalid argument

[root@maharaja tracing]# echo 1 > events/uprobes/p_usdt_0x401198/enable
bash: echo: write error: Invalid argument

[root@maharaja tracing]# echo 1 > events/enable
bash: echo: write error: Invalid argument
```

danobi:
> I guess the problem is we passed in "address of semaphore" instead of "offset of semaphore".

What's the difference between "address" and "offset". A true offset from the uprobe addr could be negative, right? And the kernel API only takes unsigned offsets.

danobi:
Ah, figured it out. It's the offset from the base of the section (I think).

liu-song-6:
> Ah, figured it out. It's the offset from the base of the section (I think).

It is the offset from the beginning the file, not from the base of that section.

yonghong-song:
[buildbot, test this please]

yonghong-song:
it is possible we can write a test for this? The test can be guarded with proper kernel version.

danobi:
> It is the offset from the beginning the file, not from the base of that section.

Yeah you're right. I had to stare at the perf code a bit. I managed to get this to work but it requires another commit. I'll write a test too.

danobi:
I'll take the PR out of draft when it's ready for review again

dalehamel:
@danobi i'm working on the same thing in https://github.com/iovisor/bcc/pull/2738 since february, looks like i took too long :upside_down_face: we should swap notes. I have it mostly working i believe just need tests.

danobi:
@dalehamel ah, I forgot you were doing it too. I knew at some point...

I've actually got the whole thing working end to end in bpftrace too: https://github.com/danobi/bpftrace/tree/uprobe_refcnt . As soon as I write a test for this bcc PR I'll send both PRs.

danobi:
I added a commit to have the C++ API use the refcnt feature if available so I could write a test. I ran the usdt tests locally and my new test passes (although some old tests are still failing). Probably an environment thing.

dalehamel:
@danobi i think the test failures are legit https://github.com/danobi/bcc/runs/1277920687?check_suite_focus=true the 16.04 github actions instances use the older kernel, so maybe the python code path that is failing is still hitting the older api? My branch passes on both the 4.15 kernel and the 5.4 kernel, it uses the check in a few more places, on my most recent change https://github.com/dalehamel/bcc/runs/1261615940?check_suite_focus=true

I think that means the current implementation will fail on older kernels though, if it is forcing to use the kernel API when it's not supported.

dalehamel:
@danobi reviewed this and it looks great, thanks so much for getting this over the line. Really excited to see this in bpftrace too

yonghong-song:
[buildbot, test this please]

yonghong-song:
@davemarchevsky Could you also take a look at this patch since it impacts usdt?

davemarchevsky:
LGTM. Excited that we'll be able to avoid doing hacky `/proc/PID/mem` writes!

yonghong-song:
@danobi Thanks!

PR #3048, opened 2020-08-12
Netqtop 3037

Yolandajn:
# Issue #3037 
The tool netqtop uses tracepoints NET_DEV_START_XMIT and NETIF_RECEIVE_SKB to intercept every transmitted and received packet, as a result, considerable performance descent is expected.

### Environment 
Kernel version: 4.19.91-19.1.al7.x86_64
CPU: Intel(R) Xeon(R) Platinum 8269CY CPU @ 2.50GHz
Number of cpu cores: 104
### Throughput test
**Commands**
taskset -c ${cpuid} sockperf -tp -i 127.0.0.1 -t 20
taskset -c 0 netqtop -n lo
\# Use lo NIC to test the full impact of netqtop when both tx queues and rx queues are busy. 
**Method**
a) Run sockperf on multiple cpus, record PPS  
b) Run sockperf on multiple cpus and run netqtop on cpu0, record PPS  
**Result**
| Threads | a) PPS [msg/sec] | b) PPS [msg/sec] | PPS drop [%] |
| --- | --- | --- | --- |
| 1 | 301880 | 214680 | 28.89 |
| 2 | 398041 | 292487 | 26.52 |
| 4 | 765540 | 556819 | 27.26 |
| 8 | 1529395 | 1108588 | 27.51 |
| 16 | 2945627 | 2092091 | 28.98 |
| 32 | 5608652 | 3777976 | 32.64 |
| 64 | 6179109 | 4584678 | 25.80 |
| 104 | 6744903 | 4852358 | 28.06 |

From the above chart we see netqtop causes at most 32.64% decrease of PPS.

### Latency test
**Commands**
\# start sockperf server
taskset -c 0 sockperf sr -p 13344 
\# start sockperf client
taskset -c 103 sockperf pp -t 20 -i 127.0.0.1 -p 13344
\# start netqtop and start sockperf client again
taskset -c 102 netqtop -n lo -i 1
taskset -c 103 sockperf pp -t 20 -i 127.0.0.1 -p 13344
**Method & Result**
a) Run sockperf in pingpong mode and record latency: 7.424 usec   
b) Run netqtop and then run sockperf in pp mode and record latency: 8.590 usec   
Comparing a) and b), the latency increased **1.17 usec**.  

### Reason for not using sampling feature
[...]

PR #3140, opened 2020-10-19
fix netqtop python3 compatible

Yolandajn:
fixed python3 compatible issue referred in #3139 and [pr#3048](https://github.com/iovisor/bcc/pull/3048#issuecomment-710483485)

### DIFF
--- a/tools/netqtop.py
+++ b/tools/netqtop.py
@@ -6,7 +6,6 @@ from ctypes import *
 import argparse
 import os
 from time import sleep,time,localtime,asctime
-import types
 
 # pre defines -------------------------------
 ROOT_PATH = "/sys/class/net"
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

tools/netqtop.py
```diff
@@ -28,7 +29,7 @@ def to_str(num):
     elif num > 1000:
         return str(round(num/1024.0, 2)) + 'K'
     else:
-        if type(num) == types.FloatType:
+        if isinstance(num, float):
```
yonghong-song, line 32: Since we removed using `types` module, we do not need to import it (line 9), right?

yonghong-song:
[buildbot, ok to test]

tools/netqtop.py
Yolandajn: yes, thank you. I deleted this line.

PR #3141, opened 2020-10-19
Push additional build artifacts in CI

dalehamel:
This enables pushing docker images to quay.io, but I will need to have github admin of iovisor/bcc become an admin of the quay repository, in order to add the necessary secret for this to push the images.

I believe this will be very useful, for projects that need an up-to-date image pulled from CI on github actions, and make it easy to bisect revisions, using the docker repo as an artifact repository. In addition to docker images (for distribution purposes only, really), this also publishes debian packages as part of a tarball in the github actions output https://github.com/dalehamel/bcc/actions/runs/311964261

This builds on  @alban 's approach of guarding the image build with a variable, but replaces the explicit variable by detecting if `QUAY_TOKEN` is present or absent. The generic dockerhub approach is still available, but I've augmented it slightly, and moved it to a separate github workflow. Since I have set this up on my personal repo by adding my own personal token, I can demonstrate the images that this publishes and tags: https://quay.io/repository/dalehamel/bcc?tab=tags

As an example, to test out the latest ubuntu focal (20.04) image:

```
docker run -ti quay.io/dalehamel/bcc:quay-push-focal-release bash
ls /root/bcc
ls /usr/share/bcc
```

I have set this up to build packages for all the most recent ubuntu LTS releases, but the process is generic enough that explicit builds for various LTS debian releases could be done fairly easily as well. I happened to add ubuntu focal support to the debian control file, as I think it is safe to do so for a new LTS release, and this is implicitly adding a CI check for this build.

This builds `release` artifacts using the existing debian build script, and results in a container being pushed and tagged in various ways (by branch, by commit, etc). This allows the user to fetch from quay.io

We used to have https://repo.iovisor.org/apt/ as a way to get reasonably up-to-date apt packages, this serves as a substitute for that. The latest apt package published to this repo is for bcc 0.10.0, for ubuntu bionic.

### DIFF
--- a/.github/workflows/bcc-test.yml
+++ b/.github/workflows/bcc-test.yml
@@ -92,38 +92,3 @@ jobs:
 #     https://github.com/marketplace/actions/debugging-with-tmate
 #    - name: Setup tmate session
 #      uses: mxschmitt/action-tmate@v1
-
-  # Optionally publish container images, guarded by the GitHub secret
-  # DOCKER_PUBLISH.
-  # GitHub secrets can be configured as follows:
-  #   - DOCKER_PUBLISH = 1
-  #   - DOCKER_IMAGE = docker.io/myorg/bcc
-  #   - DOCKER_USERNAME = username
-  #   - DOCKER_PASSWORD = password
-  publish:
-    name: Publish
-    runs-on: ubuntu-latest
-    steps:
-
-    - uses: actions/checkout@v1
-
-    - name: Initialize workflow variables
-      id: vars
-      shell: bash
-      run: |
-          echo ::set-output name=DOCKER_PUBLISH::${DOCKER_PUBLISH}
-      env:
-        DOCKER_PUBLISH: "${{ secrets.DOCKER_PUBLISH }}"
-
-    - name: Build container image and publish to registry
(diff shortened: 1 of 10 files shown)


dalehamel:
An example, built for this pull request:

https://github.com/dalehamel/bcc/actions/runs/315668215

dalehamel:
As an example of how to use this in docker builds to get the latest BCC packages:


```
FROM quay.io/dalehamel/bcc:quay-push-focal-release as bcc

...

FROM ubuntu:20.04
....

# Install bcc by copying apt packages from docker image
COPY --from=bcc /root/bcc /tmp/bcc
RUN \
  apt-get update -y && \
  DEBIAN_FRONTEND=noninteractive apt-get install -y python python3 binutils libelf1 kmod && \
  dpkg -i /tmp/bcc/*.deb && rm -rf /tmp/bcc

```

And with an official bcc image, we can instead use the URL: "quay.io/iovisor/bcc" rather than my personal repo.

Right now bpftrace builds bcc from source, but this above snippet can allow bpftrace and other projects to avoid needing to do so.

yonghong-song:
[buildbot, test this please]

PR #3145, opened 2020-10-25
Update INSTALL.md

Jakhotiya:
For ubuntu focal fossa only instructions given for Eoan worked for me. If we don't add Focal fossa to list, people will go to "For other versions" section. That refers to libllvm3.7 which could not be installed on ubuntu focal. Hence this proposed change


yonghong-song:
[buildbot, test this please]

PR #3153, opened 2020-11-04
use correct arch register for the 4th param of x86_64 syscalls

### LINKED ISSUE #3150: Probing syscall__renameat in Ubuntu 20.04 - can't read 4th param
I created kprobes for the `renameat` and `renameat2` syscalls:

```
int syscall__renameat(struct pt_regs *ctx, int oldfd, const char __user *oldname,                                                                        
                      int newfd, const char __user *newname)                                                                                             
{                                                                                                                                                        
.
.
}

int syscall__renameat2(struct pt_regs *ctx, int oldfd, const char __user *oldname,
                       int newfd, const char __user *newname, unsigned int flags)
{  
.
.
}
```

I can read the data at `oldname` fine, but `newname` seems to be garbage. `oldfd` and `newfd` are correct.

I had a similar problem with RHEL 8 (#2597) but that fix doesn't work here. In fact, none of the PT_REGS_PARMx point to usable strings.

I'm using kernel 5.4.0-52-generic on Ubuntu 20.04.1 LTS. BCC version is v0.16.0. LLVM tools are 8.0.1 (built from source).

Thank you

yonghong-song:
Fix #3150

On x86_64, for syscalls, the calling convension is to use
r10 instead of rcx for the 4th parameter. I have verified this
with disassembling vmlinux codes.
  https://www.systutorials.com/x86-64-calling-convention-by-gcc/

bcc previously used rcx for the 4th parameter for all cases.
This patch fixed the issue by using r10 for syscalls.
A macro PT_REGS_PARM4_SYSCALL() is also introduced in helpers.h
to access the 4th parameter for r10.

### DIFF
--- a/src/cc/export/helpers.h
+++ b/src/cc/export/helpers.h
@@ -1099,6 +1099,7 @@ int bpf_usdt_readarg_p(int argc, struct pt_regs *ctx, void *buf, u64 len) asm("l
 #define PT_REGS_PARM2(ctx)	((ctx)->si)
 #define PT_REGS_PARM3(ctx)	((ctx)->dx)
 #define PT_REGS_PARM4(ctx)	((ctx)->cx)
+#define PT_REGS_PARM4_SYSCALL(ctx)	((ctx)->r10) /* for syscall only */
 #define PT_REGS_PARM5(ctx)	((ctx)->r8)
 #define PT_REGS_PARM6(ctx)	((ctx)->r9)
 #define PT_REGS_RET(ctx)	((ctx)->sp)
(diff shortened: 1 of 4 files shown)


craig65535:
Hello,
I was wondering what the reasoning was behind the decision to have PT_REGS_PARM4_SYSCALL defined for x86 only, and only for param 4. It would seem that a full set of generic _SYSCALL macros for all architectures would be more useful as helpers

yonghong-song:
> Hello,
> I was wondering what the reasoning was behind the decision to have PT_REGS_PARM4_SYSCALL defined for x86 only, and only for param 4. It would seem that a full set of generic _SYSCALL macros for all architectures would be more useful as helpers

No particular reason. But your suggestion is good to have a full set of PT_REGS_PARM*_SYSCALL. Could you help submit a pull request? Thanks!

craig65535:
@yonghong-song See #3155

Issue #2792, opened 2020-03-05
tcptop TX_KB always gives 0 values


abhay2101:
OS environment: centos7
Kenrel version: 3.10.0-1062.12.1.el7.x86_64
bcc tools: version 11 and 12 both.

tcpop  TX_KB field always gives 0 TX_KB. RX_RB is right value.
tcplife gives right values though.

we can repro with traffic or scp of last file back and forth.

abhay2101:
@brendangregg

netedwardwu:
I'm not sure the below is the reason or not.

https://elixir.bootlin.com/linux/v3.10.0/ident/tcp_sendmsg
Kernel 3.10 has no tcp_sendmsg to do kprobe__tcp_sendmsg.

Could you share your tcptop.py code like the below?
https://github.com/iovisor/bcc/blob/master/tools/tcptop.py#L97

abhay2101:
tcp_sendmsg function is present in kernel code .

https://git.centos.org/rpms/kernel/tree/c7

netedwardwu:
Hi @abhay2101 

I don't really understand why your kernel version is 3.10, and bcc is working???

Follow the below link
https://github.com/iovisor/bcc/blob/master/docs/kernel-versions.md
3.10 is too old to **every main features**...

Few questions:
1. Does your kernel have backward patch?
2. Could you add some bpf_trace_printk() to debug?
https://github.com/iovisor/bcc/blob/master/tools/tcptop.py#L103
```
    if (family == AF_INET) {
        struct ipv4_key_t ipv4_key = {.pid = pid};
        ipv4_key.saddr = sk->__sk_common.skc_rcv_saddr;
        ipv4_key.daddr = sk->__sk_common.skc_daddr;
        ipv4_key.lport = sk->__sk_common.skc_num;
        dport = sk->__sk_common.skc_dport;
        ipv4_key.dport = ntohs(dport);
        ipv4_send_bytes.increment(ipv4_key, size);
```
Because RX is working you said. Check above variables have value?

abhay2101:
ebof tracing is suppoed to work on centos7 

https://www.redhat.com/en/blog/introduction-ebpf-red-hat-enterprise-linux-7
https://github.com/iovisor/bcc/blob/master/INSTALL.md#centos---source

           sshd-32462 [000] .... 50101.618810: 0: abhay bpf trace size=84 
            sshd-32462 [000] .... 50101.618819: 0: abhay bpf trace size=84 
[...]

PR #3156, opened 2020-11-05
Add kernel versioning for the tcp_sendmsg function; Add date printing…

DavadDi:
Add kernel version macros so that `tcptop` can work in centos 7.x, because this version of the kernel is generally 3.10.x; in order to facilitate the analysis of `tcptop` output after locating to the file, add the printed time information.

See also: https://github.com/iovisor/bcc/issues/2792

### DIFF
--- a/examples/cpp/CMakeLists.txt
+++ b/examples/cpp/CMakeLists.txt
@@ -19,7 +19,7 @@ foreach(EXAMPLE ${EXAMPLES})
   if(NOT CMAKE_USE_LIBBPF_PACKAGE)
     target_link_libraries(${NAME} bcc-static)
   else()
-    target_link_libraries(${NAME} bcc-shared-no-libbpf)
+    target_link_libraries(${NAME} bcc-shared)
   endif()
 
   if(INSTALL_CPP_EXAMPLES)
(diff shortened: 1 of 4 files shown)


yonghong-song:
[buildbot, test this please]

brendangregg:
Thanks, however, this seems like a bad idea: SystemTap's scripts became a mess trying to satisfy ancient kernels. It's a maintenance and testing burden, and makes these worse examples for people to learn from.

I'd suggest writing a convincing explanation for why we should support 3.10.x kernels. If you cannot, we'll close this PR.

DavadDi:
@brendangregg  Thank you very much for your quick reply, and also for providing `tcptop`, an excellent tool that is a handy tool for locating difficult problems in our production environment.

As mentioned in the reply about the maintenance burden, it's true that too much kernel versioning can be quite costly. I checked the CentOS website and found that the maintenance cycle for **CentOS7 runs until July 2024**, while the default kernel for both **CentOS7 and Redhat 7 is 3.10.xx**, and the CentOS and Redhat series has a lot of features as mainstream Linux systems.  This is a huge market for our BCC tools to better deliver performance.

The 7.x family also provides porting and support for many technologies in `ebpf`, and continues to support a richer set of `ebpf` features, and the `tcp_sendmsg` base function changes very infrequently, even with the addition of kernel macro support, not as much as Systemtap, in my opinion. This is for information only.

It is also advisable to provide a hint of the `tcp_sendmsg` function change or a minimum version requirement for this tool if you consider not adding a Linux macro, as this is more user friendly than running it correctly but the results are not as expected, since most users are not aware of such changes in the signatures of the underlying kernel functions.

For myself, I usually spend quite a bit of time trying to locate and troubleshoot when the tool is working properly, but the results are never as expected.

PR #3158, opened 2020-11-05
tools/offcputime: Fix python3 string output

lava:
When using non-folded output, some of the strings would
be printed as raw byte arrays on python3, like this:

    b'finish_task_switch'
    b'__schedule'
    b'schedule'
    b'worker_thread'
    b'kthread'
    b'ret_from_fork'
    -                kworker/u16:1 (22022)
        2267942

This commit updates the code to treat these as utf8 strings,
consistent with all the other strings printed from this tool.


yonghong-song:
[buildbot, test this please]

PR #3128, opened 2020-10-08
add DNS correlation to connect tracking

nabilschear:
often i'd like to be able to track DNS domains that were associated with an IP at the time of connect.  So, I wired up a simple and hopefully low-overhead kprobe to udp_recvmsg that will capture DNS response packets and put them into a cache dictionary in user space.  When connect syscalls come in, then it will look for a corresponding DNS query in the cache.  It will also note the time delta between when the DNS response was received and when the connect happened.

This requires dnslib and expiringdict python packages.  dnslib is already required by an example in the bcc tree: https://github.com/iovisor/bcc/blob/b0bf04ac4042a6c004b15dcc5e40e12ae78020f9/examples/networking/dns_matching/dns_matching.py but I couldn't find anywhere to add these dependencies for a core tool.  Looking for help here.

PR #3162, opened 2020-11-10
Drop ubuntu xenial support from release, it is EOL

dalehamel:
https://github.com/iovisor/bcc/pull/3128 broke the release process, as it tries to install packages that are not available on ubuntu xenial (16.04)

This should repair this problem by checking the version and not installing this libraries on ubuntu xenial

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -15,8 +15,6 @@ jobs:
     strategy:
       matrix:
         env:
-        - NAME: xenial-release
-          OS_RELEASE: 16.04
         - NAME: bionic-release
           OS_RELEASE: 18.04
         - NAME: focal-release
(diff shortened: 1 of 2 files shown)


dalehamel:
FYI @nabilschear @yonghong-song

dalehamel:
Actually it looks like the xenial build is failing due to missing packages, I think it is no longer supported by ubuntu anyways, so I am going to drop xenial instead.

dalehamel:
Here is an example run, releasing against my own copy of bcc: https://github.com/dalehamel/bcc/runs/1381460245?check_suite_focus=true

dalehamel:
Ugh still issues... these packages only exist in ubuntu bionic it seems : /

dalehamel:
Ok here is the passing run https://github.com/dalehamel/bcc/runs/1381597505?check_suite_focus=true

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

PR #3166, opened 2020-11-13
tools/biolatpcts: Add accumulative reporting on SIGUSR2

htejun:
SIGUSR1's behavior is the same - print and start a new interval. SIGUSR2
triggers output without starting a new interval and can be used to monitor
progress without affecting the accumulation of data points.


yonghong-song:
[buildbot, test this please]

PR #3173, opened 2020-11-18
dirtop.py : Fix a race condition which is a consequence of counts.clear(). 

egobillot:
`table.clear()` suffers from race condition if there is no call to bpf_spin_lock/bpf_spin_unlock. The problem is bpf_spin_lock appears in kernel v5.1. So I have removed the call to `counts.clear()` since it is not mandatory. Instead I compute the delta during the interval.

### DIFF
--- a/tools/dirtop.py
+++ b/tools/dirtop.py
@@ -202,16 +202,25 @@ DNAME_INLINE_LEN = 32  # linux/dcache.h
 print('Tracing... Output every %d secs. Hit Ctrl-C to end' % interval)
 
 
-def sort_fn(counts):
+def sort_metrics(metrics):
     """Define how to sort the columns"""
+    inode = metrics[1]
     if args.sort == "all":
-        return (counts[1].rbytes + counts[1].wbytes + counts[1].reads + counts[1].writes)
+        return (inode["rbytes"] +
+                inode["wbytes"] +
+                inode["reads"] +
+                inode["writes"])
     else:
-        return getattr(counts[1], args.sort)
+        return inode[args.sort]
 
 
 # output
 exiting = 0
+
+# metric is a dict where key is an inode_id and value a
+# dictionary of datas collected for this inode _id
+metrics = {}
+
 while 1:
     try:
         sleep(interval)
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
@egobillot could you explain what is the exact problem? Missing or incorrect output or what? This will help evaluate whether the fix is sufficient or not.

egobillot:
@yonghong-song The problem is simply the tool is not accurate enough.
My first explanation was not complete and the commit message not clear. My apologies. Let me clarify.

The accuracy issue is the consequence of 2 things:
1. The algo itself
    - Step1 : we get the data from the hash map
    - Step2 : we print it to the terminal
    - Step3 : we clear the hash map 
eBPF continues to collect data between step2 and step3. It will be lost. This is the first accuracy issue.

2. counts.clear() suffers from race condition:
    - there is concurrency between kernel and userspace on the eBPF map.
    - The probability to enter in race condition is inversely proportional to the read/write rate
 
<ins>How to test/reproduce this issue :</ins> 
Create a file in /tmp/mydrectory
`$ mkdir -p /tmp/mydirectory`
`$ fallocate -l 1000000 /tmp/mydirectory/file `

run dirtop.py
`sudo python3 ./dirtop.py -s rbytes -d /tmp/mydirectory  -C 1 > ~/t.txt` ==> Wait the bpf is loaded (~5 s).

Then generate 1 million read with dd
`dd if=/tmp/mydirectory/file bs=1 count=1000000`

Count how many reads have been captured with :
`grep -E "^[0-9]+ " ~/t.txt| awk {'print res+=$1'}` => the last line is the nb of reads dirtop has captured

Emilien

yonghong-song:
@egobillot Thanks for explanation. This is indeed an issue and it has been there for a long time. Now the kernel actually has a better interface to do batched lookup_and_delete. https://lore.kernel.org/bpf/20200115184308.162644-1-brianvv@google.com/. If you allocate sufficient large buffer, you will be able to get all elements back to user space at the same time those elements will be removed from the kernel. This way, we won't have race you mentioned in the above. This will require some bcc infrastructure work but I think this is a better generic method which may benefit other tools which have similar issues. Do you want to take a look at this direction? Thanks!

egobillot:
@yonghong-song Yes your solution looks better. I will have a look and go back to you if I can do it. I need to deep dive in the bcc infrastructure to understand what needs to be done.

yonghong-song:
> @yonghong-song Yes your solution looks better. I will have a look and go back to you if I can do it. I need to deep dive in the bcc infrastructure to understand what needs to be done.

Thanks. Just post to the mailing list if you need help to understand any bcc/kernel internals.

egobillot:
@yonghong-song I have a draft of the support of `BPF_MAP_LOOKUP_AND_DELETE_BATCH` in bcc. The problem is that is fixes only half of the problem.
Let me remind you the issue.
> The accuracy issue is the consequence of 2 things:
>
>    1. The algo itself
>        Step1 : we get the data from the hash map
>        Step2 : we print it to the terminal
>        Step3 : we clear the hash map
>        eBPF continues to collect data between step2 and step3. It will be lost. This is the first accuracy issue.
>
>    2. counts.clear() suffers from race condition:
>        there is concurrency between kernel and userspace on the eBPF map.
>        The probability to enter in race condition is inversely proportional to the read/write rate

lookup and delete in one batch solve the first point. 
For the second point, even if there is no more call to counts.clear() with the support of `BPF_MAP_LOOKUP_AND_DELETE_BATCH` I am still facing race condition: concurrency between kernel and userspace on the eBPF map. For this point I have no solution.  bpf_spin_unlock is not supported for tracing programs on the latest kernel (https://elixir.bootlin.com/linux/v5.11-rc2/source/include/uapi/linux/bpf.h#L2670)
```
 *		* Tracing programs and socket filter programs cannot use
 *		  **bpf_spin_lock**\ () due to insufficient preemption checks
 *		  (but this may change in the future).
```
Any ideas ?

yonghong-song:
@egobillot Thanks for working on this! For your 2nd point, yes, there are concurrent access to the map by both user space and kernel. But kernel batch_and_delete will hold a bucket lock until all elements are copied and removed. And then pending kernel update() (spin lock on the same bucket lock) can then do an update(insert an entry). So kernel already handled this
race. https://github.com/torvalds/linux/blob/master/kernel/bpf/hashtab.c#L1405-L1618

For your question about bpf_spin_lock(), there is no need to use in bpf program as the above reason, for an update, if that bucket is busy, update will spin-lock wait on that bucket lock already.

bpf_spin_lock() is disabled for tracing problem mostly to avoid dead lock due to nested tracing program. For example,
bpf_prog 1:    bpf_spin_lock(); < an interrupt > ; bpf_spin_unlock();
during the interrupt another bpf program is invoked doing bpf_spin_lock() to the same map. This will cause dead lock.
Some kernel work is required to prevent such cases from happening.

yonghong-song:
Let us close this one. We have proper map_lookup_and_delete_batch support now (https://github.com/iovisor/bcc/pull/3234).

PR #3182, opened 2020-12-02
libbpf-tools: optimize cpudist, use mmap to eliminate bpf system calls

ethercflow:
When using `per_process` or `per_thread` mode, if there are many tasks, the bpf system call will be called many times to traverse the `hists` results. So use `BPF_MAP_TYPE_ARRAY` with `BPF_F_MMAPABLE` instead.

### DIFF
--- a/libbpf-tools/cpudist.bpf.c
+++ b/libbpf-tools/cpudist.bpf.c
@@ -17,14 +17,14 @@ const volatile pid_t targ_tgid = -1;
 
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(map_flags, BPF_F_NO_PREALLOC);
 	__type(key, u32);
 	__type(value, u64);
 } start SEC(".maps");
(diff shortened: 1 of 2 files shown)


ethercflow:
Hi @anakryiko , I've change d this tool with `BPF_MAP_TYPE_ARRAY` with `BPF_F_MMAPABLE` to eliminate bpf system calls. PTAL when you have time, thank you! :)

anakryiko:
if the system has max PID number set up high, but is not itself highly-loaded, you'll end up with tons of empty (never set) histograms. With this approach, you'd be iterating over all of them unnecessarily. So it's not at all clear that this will be an improvement. Adding no-prealloc flag might be useful to save memory, but I'm not sure if it's safe to do from the context in which finish_task_switch is called, TBH. So basically, I'd leave it as is.

ethercflow:
Ok, agree, thanks!

Issue #3186, opened 2020-12-08
Kernel 4.14 - unable to insert value returned from map into another map?


craig65535:
I'm using Amazon Linux 2, kernel 4.14.203-156.332.amzn2.x86_64. BCC version 0.17.0, llvm 8.0.1.

I wrote a BPF program that reads from one map and inserts into another.

```C
BPF_ARRAY(zeroed_event_arr, event_t, 1);
// ...
    u64 tgid_kpid = bpf_get_current_pid_tgid();
    u32 kpid = tgid_kpid & 0xffffffff;
    event_t *zeroed_event = zeroed_event_arr.lookup(&idx);
    if (zeroed_event) {
        calls.insert(&kpid, zeroed_event);
        // ...
```

This works on other kernels (4.18.0-80.11.2.el8_0.x86_64 on RHEL 8.0, 5.4.0-56-generic on Ubuntu 20.04.1). However, it fails on Amazon Linux 2:

> bpf: Failed to load program: Permission denied
0: (79) r6 = *(u64 *)(r1 +104)
1: (85) call bpf_get_current_pid_tgid#14
2: (bf) r8 = r0
3: (63) *(u32 *)(r10 -4) = r8
4: (b7) r1 = 0
5: (63) *(u32 *)(r10 -8) = r1
6: (18) r1 = 0xffff888172217c00
8: (bf) r2 = r10
9: (07) r2 += -8
10: (85) call bpf_map_lookup_elem#1
11: (15) if r0 == 0x0 goto pc+71
 R0=map_value(id=0,off=0,ks=4,vs=448,imm=0) R6=inv(id=0) R8=inv(id=0) R10=fp0
12: (18) r1 = 0xffff8881a8546600
14: (bf) r7 = r10
15: (07) r7 += -4
16: (bf) r2 = r7
17: (bf) r3 = r0
18: (b7) r4 = 1
19: (85) call bpf_map_update_elem#2
R3 type=map_value expected=fp

> terminate called after throwing an instance of 'std::runtime_error'
  what():  Failed to load syscall__rename: -1
Aborted

This appears to have been reported before in #347 and there was a patch proposed in https://lists.linuxfoundation.org/pipermail/iovisor-dev/2017-September/001105.html . It looks like that patch is in mainline 4.18: https://github.com/amazonlinux/linux/commit/d71962f3e627b5941804036755c844fabfb65ff5 and Amazon's 4.19 kernel.

[...]

Issue #347, opened 2016-02-07
reusing map values as map keys


brendangregg:
Low priority, but I've found that map values can't be reused as keys. Eg:

``` C
struct info_t {
    u32 pid;
    char name[TASK_COMM_LEN];
};
BPF_HASH(infobyreq, struct request *, struct info_t);
BPF_HASH(counts, struct info_t, u64);
[...]
infop = infobyreq.lookup(&req);
valp = counts.lookup_or_init(infop, &zero);
```

Gives:

``` Shell
from 17 to 43: R0=map_value_or_null(ks=8,vs=20) R6=map_value(ks=8,vs=20) R10=fp
43: (18) r1 = 0x2f70f300
45: (bf) r2 = r6
46: (85) call 1
R2 type=map_value expected=fp
```

I can __builtin_memcpy() as a workaround to a new key.

pchaigno:
@4ast Would allowing map helpers access to map element values in the verifier (in the vein of https://patchwork.ozlabs.org/patch/712889/) be a sensible fix for this? If so, I have a patch ready, which I can submit to the iovisor mailing list.

4ast:
there was quite a bit work on the verifier side recently and helpers should be able to access map values already. What's missing? and yes, the patch would be great :)

pchaigno:
In the 4.13-rc7 version, other helpers appear to be able to access map values, but not map helpers. I'll try to send the patch today. Best to discuss it over code.

pchaigno:
I just posted the patch on the iovisor mailing list.

matthew-olson-intel:
@pchaigno Since that patch seems to be merged, do we need to re-test and confirm that this issue is resolved?

pchaigno:
I believe this can be closed. My patch was released in Linux 4.18.

brendangregg:
Thanks @pchaigno and @matthew-olson-intel .

PR #3187, opened 2020-12-10
4.18 - Pass map values to map helpers

craig65535:
Document this limitation reported in #347 and #3186


yonghong-song:
[buildbot, test this please]

Issue #3190, opened 2020-12-13
opensnoop: error when trying to show/filter flags


zb3:
I'm on Fedora 33, after running:
```
sudo /usr/share/bcc/tools/opensnoop -f O_WRONLY -f O_RDWR -e
```
I got:
```
bpf: Failed to load program: Permission denied
arg#0 type is not a struct
Unrecognized arg#0 type PTR
; KRETFUNC_PROBE(__x64_sys_openat2, struct pt_regs *regs, int ret)
0: (bf) r6 = r1
; KRETFUNC_PROBE(__x64_sys_openat2, struct pt_regs *regs, int ret)
1: (79) r9 = *(u64 *)(r6 +8)
2: (79) r1 = *(u64 *)(r6 +0)
func '__x64_sys_openat2' arg0 has btf_id 178 type STRUCT 'pt_regs'
; const char __user *filename = (char *)PT_REGS_PARM2(regs);
3: (79) r8 = *(u64 *)(r1 +104)
; struct open_how __user *how = (struct open_how *)PT_REGS_PARM3(regs);
4: (79) r1 = *(u64 *)(r1 +96)
; int flags = how->flags;
5: (79) r7 = *(u64 *)(r1 +0)
R1 invalid mem access 'inv'
processed 6 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
```
I got similar error when using only the `-e` flag.

## Fix
Now, I have no idea how all this works (yet...), so my fix below might not make sense, but intuitively, it seems to me that the pointer is dereferenced without being read, so I fixed it like this:
```patch
--- /tmp/opensnoop1	2020-12-13 13:07:21.415405454 +0100
+++ /tmp/opensnoop2	2020-12-13 13:08:29.936981179 +0100
@@ -215,8 +215,11 @@
 {
     int dfd = PT_REGS_PARM1(regs);
     const char __user *filename = (char *)PT_REGS_PARM2(regs);
-    struct open_how __user *how = (struct open_how *)PT_REGS_PARM3(regs);
-    int flags = how->flags;
+    struct open_how __user how;
+    int flags;
+    
+    bpf_probe_read_user(&how, sizeof(struct open_how), (struct open_how*)PT_REGS_PARM3(regs));
+    flags = how.flags;
 #else
 KRETFUNC_PROBE(FNNAME, int dfd, const char __user *filename, struct open_how __user *how, int ret)
 {
```

[...]

PR #3194, opened 2020-12-17
tools/opensnoop: Use bpf_probe_read_user explicitly

zb3:
This is a workaround for #3190


yonghong-song:
[buildbot, test this please]

PR #3196, opened 2020-12-17
biolatency.py Add -j flag

swiftomkar:
### Why
In it's current form, storing the histogram data and analyzing the same for a large set of machines is difficult.
With this flag, it is possible to store the histogram as a dictionary/json file which can be imported easily to be analyzed with libraries like Pandas. 

### Example output
```
{
0: 0, 
2: 0, 
4: 0, 
8: 0, 
16: 0, 
32: 0, 
64: 13, 
128: 44, 
256: 41, 
512: 14, 
1024: 27, 
2048: 1, 
4096: 10,  
8192: 0
}
```
### Updated example output - v1
```
[
    {'interval-start': 0, 'interval-end': 1, 'count': 0}, 
    {'interval-start': 2, 'interval-end': 3, 'count': 5}, 
    {'interval-start': 4, 'interval-end': 7, 'count': 143}
]
```
### Updated example output - v2
```
{
    'ts': '2020-12-30 14:49:37', 
    'val_type': 'msecs', 
    'data': 
        [
            {'interval-start': 0, 'interval-end': 1, 'count': 2}, 
            {'interval-start': 2, 'interval-end': 3, 'count': 1}
        ], 
    'flags': 'Sync-Write'
}

```

### DIFF
--- a/man/man8/biolatency.8
+++ b/man/man8/biolatency.8
@@ -36,6 +36,12 @@ Print a histogram per disk device.
 \-F
 Print a histogram per set of I/O flags.
 .TP
+\-j
+Print a histogram dictionary along with the histogram
+.TP
+\-f
+Write the histogram dictionary to the specified file path. Used with -j
+.TP
 interval
 Output interval, in seconds.
 .TP
@@ -63,6 +69,14 @@ Include OS queued time in I/O time:
 Show a latency histogram for each disk device separately:
 #
 .B biolatency \-D
+.TP
+Show a latency histogram in a dictionary format:
+#
+.B biolatency \-j
+.TP
+Store the histogram dictionary in the specified file:
+#
+.B biolatency \-jf ./histogram.json
 .SH FIELDS
 .TP
 usecs
--- a/man/man8/biolatency.8
+++ b/man/man8/biolatency.8
@@ -37,10 +37,7 @@ Print a histogram per disk device.
 Print a histogram per set of I/O flags.
 .TP
 \-j
-Print a histogram dictionary along with the histogram
-.TP
-\-f
-Write the histogram dictionary to the specified file path. Used with -j
+Print a histogram dictionary
 .TP
 interval
 Output interval, in seconds.
@@ -73,10 +70,6 @@ Show a latency histogram for each disk device separately:
 Show a latency histogram in a dictionary format:
 #
 .B biolatency \-j
-.TP
-Store the histogram dictionary in the specified file:
-#
-.B biolatency \-jf ./histogram.json
 .SH FIELDS
 .TP
 usecs
--- a/man/man8/biolatency.8
+++ b/man/man8/biolatency.8
@@ -1,4 +1,4 @@
-.TH biolatency 8  "2020-12-29" "USER COMMANDS"
+.TH biolatency 8  "2020-12-30" "USER COMMANDS"
 .SH NAME
 biolatency \- Summarize block device I/O latency as a histogram.
 .SH SYNOPSIS
--- a/man/man8/biolatency.8
+++ b/man/man8/biolatency.8
@@ -1,4 +1,4 @@
-.TH biolatency 8  "2019-12-12" "USER COMMANDS"
+.TH biolatency 8  "2020-12-29" "USER COMMANDS"
 .SH NAME
 biolatency \- Summarize block device I/O latency as a histogram.
 .SH SYNOPSIS
--- a/tools/biolatency.py
+++ b/tools/biolatency.py
@@ -34,21 +34,21 @@ parser = argparse.ArgumentParser(
     formatter_class=argparse.RawDescriptionHelpFormatter,
     epilog=examples)
 parser.add_argument("-T", "--timestamp", action="store_true",
-                    help="include timestamp on output")
+    help="include timestamp on output")
 parser.add_argument("-Q", "--queued", action="store_true",
-                    help="include OS queued time in I/O time")
+    help="include OS queued time in I/O time")
 parser.add_argument("-m", "--milliseconds", action="store_true",
-                    help="millisecond histogram")
+    help="millisecond histogram")
 parser.add_argument("-D", "--disks", action="store_true",
-                    help="print a histogram per disk device")
+    help="print a histogram per disk device")
 parser.add_argument("-F", "--flags", action="store_true",
-                    help="print a histogram per set of I/O flags")
+    help="print a histogram per set of I/O flags")
 parser.add_argument("interval", nargs="?", default=99999999,
-                    help="output interval, in seconds")
+    help="output interval, in seconds")
 parser.add_argument("count", nargs="?", default=99999999,
-                    help="number of outputs")
+    help="number of outputs")
 parser.add_argument("--ebpf", action="store_true",
-                    help=argparse.SUPPRESS)
+    help=argparse.SUPPRESS)
 parser.add_argument("-j", "--json", action="store_true",
     help="json output")
 parser.add_argument("-f", "--file", type=str)
@@ -68,16 +68,20 @@ if args.flags and args.disks:
 bpf_text = """
 #include <uapi/linux/ptrace.h>
 #include <linux/blkdev.h>
+
 typedef struct disk_key {
     char disk[DISK_NAME_LEN];
     u64 slot;
 } disk_key_t;
+
 typedef struct flag_key {
     u64 flags;
     u64 slot;
 } flag_key_t;
+
 BPF_HASH(start, struct request *);
 STORAGE
+
 // time block I/O
 int trace_req_start(struct pt_regs *ctx, struct request *req)
 {
@@ -85,10 +89,12 @@ int trace_req_start(struct pt_regs *ctx, struct request *req)
     start.update(&req, &ts);
     return 0;
 }
+
 // output
 int trace_req_done(struct pt_regs *ctx, struct request *req)
 {
     u64 *tsp, delta;
+    
     // fetch timestamp and calculate delta
     tsp = start.lookup(&req);
     if (tsp == 0) {
@@ -96,8 +102,10 @@ int trace_req_done(struct pt_regs *ctx, struct request *req)
     }
     delta = bpf_ktime_get_ns() - *tsp;
     FACTOR
+    
     // store as histogram
     STORE
+    
     start.delete(&req);
     return 0;
 }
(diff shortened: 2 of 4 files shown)


yonghong-song:
[buildbot, test this please]

tools/biolatency.py
```diff
@@ -15,39 +15,51 @@
...
+    ./biolatency -j                 # print a dictionary
+    ./biolatency -j -f name.json    # Dump the histogram dictionary to name.json file
 """
 parser = argparse.ArgumentParser(
     description="Summarize block device I/O latency as a histogram",
     formatter_class=argparse.RawDescriptionHelpFormatter,
     epilog=examples)
 parser.add_argument("-T", "--timestamp", action="store_true",
-    help="include timestamp on output")
+                    help="include timestamp on output")
```
yonghong-song, line 37: Could you avoid format change here so we can avoid unnecessary code churn? The same for other places in this commit.

tools/biolatency.py
```diff
@@ -139,9 +143,9 @@
         b.attach_kprobe(event="blk_start_request", fn_name="trace_req_start")
     b.attach_kprobe(event="blk_mq_start_request", fn_name="trace_req_start")
 b.attach_kprobe(event="blk_account_io_done",
-    fn_name="trace_req_done")
+                fn_name="trace_req_done")
 
-print("Tracing block device I/O... Hit Ctrl-C to end.")
+# print("Tracing block device I/O... Hit Ctrl-C to end.")
```
yonghong-song, line 148: You do not want to unconditionally comment out the above banner. You can guard it for non-jason output.

tools/biolatency.py
```diff
@@ -15,39 +15,51 @@
...
-    ./biolatency -D         # show each disk device separately
-    ./biolatency -F         # show I/O flags separately
+    ./biolatency                    # summarize block I/O latency as a histogram
+    ./biolatency 1 10               # print 1 second summaries, 10 times
+    ./biolatency -mT 1              # 1s summaries, milliseconds, and timestamps
+    ./biolatency -Q                 # include OS queued time in I/O time
+    ./biolatency -D                 # show each disk device separately
+    ./biolatency -F                 # show I/O flags separately
+    ./biolatency -j                 # print a dictionary
+    ./biolatency -j -f name.json    # Dump the histogram dictionary to name.json file
```
yonghong-song, line 30: We have new flags, could you add to examples files and man pages?

yonghong-song:
[buildbot, ok to test]

tools/biolatency.py
swiftomkar: Yes, I'll fix all of that!

tools/biolatency.py
swiftomkar: Sure!

tools/biolatency.py
swiftomkar: I have put it within an `if` block now. Will not print banner only when the json output is being written to a file.

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

tools/biolatency_example.txt
```diff
@@ -292,11 +292,62 @@ flags = Priority-Metadata-Write
...
 
+The -j option prints a dictionary of the histogram.
+In most cases, this flag will be used with the -f flag. For example:
+
+# ./biolatency.py -j
+Tracing block device I/O... Hit Ctrl-C to end.
+^C
+{32: 4, 2: 0, 4096: 32, 4: 0, 1024: 10, 8192: 0, 8: 0, 64: 38, 128: 42, 256: 5, 16: 0, 512: 1, 2048: 0}
+     usecs               : count     distribution
+         0 -> 1          : 0        |                                        |
```
yonghong-song, line 303: Look at bpftool json output example
   https://manpages.ubuntu.com/manpages/focal/man8/bpftool-prog.8.html
can we design something similar?

If we want to output to json, we should just output in json format, there is no need to output old way format. So the above format maybe looks like
  [{
     "usecs-start": 0,
     "usecs-end" : 1,
     "count" : 4096
   },{
     "usecs-start": 2,
     "usecs-end": 3,
     "count": 0
    } ...
  }

The above bpftool example is with pretty print. I guess your json output does not need to be pretty print.

I think we do not need "-f" option. json format is mostly for automatic/advanced processing, we do not need to display banner "Tracing block device I/O... Hit Ctrl-C to end." so the output can be redirected to a file if needed. In man page, we can mention "-j" option will omit banner to formatted output. WDYT?

yonghong-song (commented):
For completeness, you should also handle output with different options e.g. -D (histogram per disk device) and -F (histogram per set of I/O flags) -T (timestamp) etc.

tools/biolatency_example.txt
swiftomkar: Yes, I think that is a better design. Pushing that change soon. Thanks!

swiftomkar:
@yonghong-song I could add the json printing function in `table.py` file (https://github.com/iovisor/bcc/blob/master/src/python/bcc/table.py) so that other tools that need can use the function as a method. WDYT? 
The json would print something like:
[{'interval-start': 0, 'interval-end': 1, 'count': 0}, {'interval-start': 2, 'interval-end': 3, 'count': 5}, {'interval-start': 4, 'interval-end': 7, 'count': 143}]

man/man8/biolatency.8
```diff
@@ -36,6 +36,12 @@ Print a histogram per disk device.
 \-F
 Print a histogram per set of I/O flags.
 .TP
+\-j
+Print a histogram dictionary along with the histogram
```
yonghong-song, line 40: -j only prints dictionary

man/man8/biolatency.8
```diff
@@ -36,6 +36,12 @@ Print a histogram per disk device.
...
+.TP
+\-f
+Write the histogram dictionary to the specified file path. Used with -j
```
yonghong-song, line 43: -f is not available any more.

man/man8/biolatency.8
```diff
@@ -63,6 +69,14 @@ Include OS queued time in I/O time:
...
 #
 .B biolatency \-D
+.TP
+Show a latency histogram in a dictionary format:
+#
+.B biolatency \-j
+.TP
+Store the histogram dictionary in the specified file:
+#
+.B biolatency \-jf ./histogram.json
```
yonghong-song, line 79: -jf is not available any more

tools/biolatency.py
```diff
@@ -18,12 +18,13 @@
...
-    ./biolatency -Q         # include OS queued time in I/O time
-    ./biolatency -D         # show each disk device separately
-    ./biolatency -F         # show I/O flags separately
+    ./biolatency                    # summarize block I/O latency as a histogram
+    ./biolatency 1 10               # print 1 second summaries, 10 times
+    ./biolatency -mT 1              # 1s summaries, milliseconds, and timestamps
+    ./biolatency -Q                 # include OS queued time in I/O time
+    ./biolatency -D                 # show each disk device separately
+    ./biolatency -F                 # show I/O flags separately
+    ./biolatency -j                 # print a dictionary
```
yonghong-song, line 27: Suppose we display results with -D
```
./biolatency.py -D
Tracing block device I/O... Hit Ctrl-C to end.
^C

disk = b'vda'
        64 -> 127        : 3        |******                                  |
       128 -> 255        : 4        |********                                |
       256 -> 511        : 18       |****************************************|
       512 -> 1023       : 10       |**********************                  |
      1024 -> 2047       : 3        |******                                  |
```

-D -j is not supported:
```
./biolatency.py -j -D
^C
  File "./biolatency.py", line 240, in <module>
    print_json_hist(dist)
  File "./biolatency.py", line 208, in print_json_hist
    max_nonzero_idx = k.value
AttributeError: 'disk_key' object has no attribute 'value'
```

I think we should support -j with all existing other flags. Could you check to make sure this is implemented properly?

tools/biolatency_example.txt
```diff
@@ -292,11 +292,62 @@ flags = Priority-Metadata-Write
 These can be handled differently by the storage device, and this mode lets us
 examine their performance in isolation.
 
+The -j option prints a dictionary of the histogram.
+In most cases, this flag will be used with the -f flag. For example:
```
yonghong-song, line 296: Please update examples properly here. -f is not available any more.

yonghong-song:
> @yonghong-song I could add the json printing function in `table.py` file (https://github.com/iovisor/bcc/blob/master/src/python/bcc/table.py) so that other tools that need can use the function as a method. WDYT?
> The json would print something like:
> [{'interval-start': 0, 'interval-end': 1, 'count': 0}, {'interval-start': 2, 'interval-end': 3, 'count': 5}, {'interval-start': 4, 'interval-end': 7, 'count': 143}]

Let us delay this until we have a second user. At this point, let us just ensure it is implemented properly in biolatency.py.

Please also change the commit subject from "biolatency.py Add -j and -f flags" to just "biolatency.py, add -j flag".

tools/biolatency_example.txt
swiftomkar: Updated.

tools/biolatency.py
swiftomkar: Fix pushed.

man/man8/biolatency.8
swiftomkar: Updated

man/man8/biolatency.8
swiftomkar: Updated

man/man8/biolatency.8
swiftomkar: Updated

swiftomkar:
> > @yonghong-song I could add the json printing function in `table.py` file (https://github.com/iovisor/bcc/blob/master/src/python/bcc/table.py) so that other tools that need can use the function as a method. WDYT?
> > The json would print something like:
> > [{'interval-start': 0, 'interval-end': 1, 'count': 0}, {'interval-start': 2, 'interval-end': 3, 'count': 5}, {'interval-start': 4, 'interval-end': 7, 'count': 143}]
> 
> Let us delay this until we have a second user. At this point, let us just ensure it is implemented properly in biolatency.py.
> 
> Please also change the commit subject from "biolatency.py Add -j and -f flags" to just "biolatency.py, add -j flag for json output".

@yonghong-song We can delay this to another PR. But, I didn't quite understand what you mean by second user.

yonghong-song:
Sorry, "second user" should be "second use case" which also implemented json output.

swiftomkar:
Got it. Thanks!

tools/biolatency_example.txt
```diff
@@ -292,11 +292,47 @@ flags = Priority-Metadata-Write
...
+^C
+{'ts': '2020-12-30 14:33:03', 'val_type': 'usecs', 'data': [{'interval-start': 0, 'interval-end': 1, 'count': 0}, {'interval-start': 2, 'interval-end': 3, 'count': 0}, {'interval-start': 4, 'interval-end': 7, 'count': 0}, {'interval-start': 8, 'interval-end': 15, 'count': 0}, {'interval-start': 16, 'interval-end': 31, 'count': 0}, {'interval-start': 32, 'interval-end': 63, 'count': 2}, {'interval-start': 64, 'interval-end': 127, 'count': 75}, {'interval-start': 128, 'interval-end': 255, 'count': 7}, {'interval-start': 256, 'interval-end': 511, 'count': 0}, {'interval-start': 512, 'interval-end': 1023, 'count': 6}, {'interval-start': 1024, 'interval-end': 2047, 'count': 3}, {'interval-start': 2048, 'interval-end': 4095, 'count': 31}]}
+
+the key `data` is the list of the log2 histogram intervals. The `interval-start` and `interval-end` define the
+latency bucket and `count` is the number of I/O's that lie in that latency range.
+
+
+The -j and -f option together lets you write the histogram dictionary into a json file.
+With these flags, it is possible to store the histogram as a dictionary/json file which
+can be imported easily to be analyzed with libraries like Pandas.
```
yonghong-song, line 308: The above paragraph can be removed.

tools/biolatency.py
```diff
@@ -193,6 +200,68 @@ def flags_print(flags):
...
+
+            prev = index
+    histogram = {"ts": strftime("%Y-%m-%d %H:%M:%S"), "val_type": val_type, "data": hist_list}
+    if section_bucket:
+        histogram[section_bucket[0]] = section_bucket[1]
+    print(histogram)
+
+
+def print_json_hist(self, val_type="value", section_header="Bucket ptr",
+                       section_print_fn=None, bucket_fn=None, bucket_sort_fn=None):
```
yonghong-song, line 231: Looking at this function again. It has a lot of duplicates with print_log2_dist. It looks like what you suggested earlier make more sense. Let me implement print_json_hist as a library function. In the bcc python library (table.py), we can factor out common part of print_json_hist/print_log2_dist to avoid code duplication. WDYT?

tools/biolatency_example.txt
swiftomkar: Ah, I missed that one

tools/biolatency.py
swiftomkar: @yonghong-song Yes, that is why I was suggesting implementing the function in table.py. I had already implemented the function in table.py and can easily push it upstream so you won't have to write it again. WDYT?

yonghong-song: sounds good. thanks!

swiftomkar: @yonghong-song Looking carefully at `table.py`, I see that `print_linear_hist` also has a lot of overlap with `print_json_hist/print_log2_dist`. 
How about we factor out all of that in a separate PR? I can take it up.

yonghong-song: Sounds good. I will merge the current PR as you can have a followup to do proper refactoring. Thanks!

PR #3200, opened 2020-12-21
tools/netqtop: fix rx queue id

pizhenwei:
Net device driver uses 'skb_record_rx_queue' to mark rx queue
mapping during receiving a packet. Also need to call
'skb_get_rx_queue' to get rx queue id.

### DIFF
--- a/tools/netqtop.c
+++ b/tools/netqtop.c
@@ -95,19 +95,33 @@ TRACEPOINT_PROBE(net, net_dev_start_xmit){
 }
 
 TRACEPOINT_PROBE(net, netif_receive_skb){
-    struct sk_buff* skb = (struct sk_buff*)args->skbaddr;
-    if(!name_filter(skb)){
+    struct sk_buff skb;
+
+    bpf_probe_read(&skb, sizeof(skb), args->skbaddr);
+    if(!name_filter(&skb)){
         return 0;
     }
 
-    u16 qid = skb->queue_mapping;
+    /* case 1: if the NIC does not support multi-queue feature, there is only
+     *         one queue(qid is always 0).
+     * case 2: if the NIC supports multi-queue feature, there are several queues
+     *         with different qid(from 0 to n-1).
+     * The net device driver should mark queue id by API 'skb_record_rx_queue'
+     * for a recieved skb, otherwise it should be a BUG(all of the packets are
+     * reported as queue 0). For example, virtio net driver is fixed for linux:
+     * commit: 133bbb18ab1a2("virtio-net: per-queue RPS config")
+     */
+    u16 qid = 0;
+    if (skb_rx_queue_recorded(&skb))
+        qid = skb_get_rx_queue(&skb);
+
     struct queue_data newdata;
     __builtin_memset(&newdata, 0, sizeof(newdata));
     struct queue_data *data = rx_q.lookup_or_try_init(&qid, &newdata);
     if(!data){
         return 0;
     }
-    updata_data(data, skb->len);
+    updata_data(data, skb.len);
     
     return 0;
 }


pizhenwei:
@Yolandajn Hi, I always got zero PPS/BFS of **rx queue 0** on my enviroment (Mellanox Technologies MT27800 Family [ConnectX-5]).
And I checked mlnx driver code and TAP driver code, I guess we need to get queue id by kernel API `skb_get_rx_queue`. Then netqtop could show the right result. What do you think about this?

Yolandajn:
> @Yolandajn Hi, I always got zero PPS/BFS of **rx queue 0** on my enviroment (Mellanox Technologies MT27800 Family [ConnectX-5]).
> And I checked mlnx driver code and TAP driver code, I guess we need to get queue id by kernel API `skb_get_rx_queue`. Then netqtop could show the right result. What do you think about this?

hi, thank you for the advice. The new version also works fine on my machine, so i think it is a good idea to make this change. BTW what did you get by skb->queue_mapping? was it `-1` or something else?
Besides, i was thinking that we could also change the probe for net_dev_start_xmit to avoid potential problems. But using `skb_get_tx_queue` caused compiling errors. Im not sure if it is necessary, or we should use other functions? If it's not necessay then we can just keep this version.

pizhenwei:
> > @Yolandajn Hi, I always got zero PPS/BFS of **rx queue 0** on my enviroment (Mellanox Technologies MT27800 Family [ConnectX-5]).
> > And I checked mlnx driver code and TAP driver code, I guess we need to get queue id by kernel API `skb_get_rx_queue`. Then netqtop could show the right result. What do you think about this?
> 
> hi, thank you for the advice. The new version also works fine on my machine, so i think it is a good idea to make this change. BTW what did you get by skb->queue_mapping? was it `-1` or something else?
> Besides, i was thinking that we could also change the probe for net_dev_start_xmit to avoid potential problems. But using `skb_get_tx_queue` caused compiling errors. Im not sure if it is necessary, or we should use other functions? If it's not necessay then we can just keep this version.

It seems that you missed the change of this PR... Actually I'm trying to use `skb_get_tx_queue` to get the queue id.

I also tested for virtio-net case:
```
Linux-5.4
Virtio-net 4 queues
```
The original version netqop also got zero PPS/BFS of **rx queue 0**. The new version(with patch of this PR) worked fine.

And we can check [source code](https://github.com/torvalds/linux/blob/master/drivers/net/virtio_net.c) of function **static void receive_buf(...)** which calls **skb_record_rx_queue(...)** to mark rx queue id.

pizhenwei:
@yonghong-song hi, what do you think about this change?

tools/netqtop.c
```diff
@@ -95,19 +95,24 @@ TRACEPOINT_PROBE(net, net_dev_start_xmit){
 }
 
 TRACEPOINT_PROBE(net, netif_receive_skb){
-    struct sk_buff* skb = (struct sk_buff*)args->skbaddr;
-    if(!name_filter(skb)){
+    struct sk_buff skb;
+
+    bpf_probe_read(&skb, sizeof(skb), args->skbaddr);
```
yonghong-song, line 100: Is this change necessary? We can directly directly get fields from skb pointer. What is the need to copy skb structure to the stack? Did you hit any issues?

tools/netqtop.c
```diff
@@ -95,19 +95,24 @@ TRACEPOINT_PROBE(net, net_dev_start_xmit){
...
+    if(!name_filter(&skb)){
         return 0;
     }
 
-    u16 qid = skb->queue_mapping;
+    u16 qid = 0;
+    if (skb_rx_queue_recorded(&skb))
+        qid = skb_get_rx_queue(&skb);
```
yonghong-song, line 107: From my understanding, 
```
static inline u16 skb_get_rx_queue(const struct sk_buff *skb)
{
        return skb->queue_mapping - 1;
} 
```

Here qid = 0 could mean two things: there is no mapping or actual qid of 0. Do we need to differentiate this? For example, if `!skb_rx_queue_recorded(&skb)`, simply return from the future as it is not in any queue? Or if not in any queue, it means queue 0?

yonghong-song (commented):
sorry for late reply. In vacation and also with Christmas holiday last week...

yonghong-song:
[buildbot, ok to test]

tools/netqtop.c
pizhenwei: Actually we directly can get `queue_mapping` field from from skb pointer, but we can't directly call function `skb_rx_queue_recorded` & `skb_get_rx_queue` because bcc reports error. So two options:

- 1, re-implement `skb_rx_queue_recorded` & `skb_get_rx_queue` in this program.
- 2, copy skb and use kernel implemented function.

Solution 1 is not a hard work, but from my understanding, keeping the same API with kernel is better.

tools/netqtop.c
pizhenwei: - If the NIC does **not** support multi-queue feature, there is only one queue(qid is always 0).
- And if the NIC supports multi-queue feature, there are several queues with different qid(from 0 to n-1).

So I guess current logic could represent the two cases. And if we want to know weather a NIC supports multi-queue feature, we can use `ethtool -l NICNAME` to test it.

tools/netqtop.c
yonghong-song: Okay, I see. Let us use option 2 then. Re-implementing is not desirable.

tools/netqtop.c
yonghong-song: What is the return value for skb_rx_queue_recorded(&skb) for single-queue feature? If this is false, then current code should work fine. Just adding some comments to explain is good enough.

pizhenwei: OK, added some comments and force-pushed!

yonghong-song:
LGTM. Thanks!

PR #3203, opened 2020-12-24
allowing examples and tests to not be built

juchem:
Building examples and tests take up a lot of time and disk space.

In many use-cases, though, `bcc` is built solely for the library portion, as a dependency of another project.

This patch allows one to opt out of building the examples and tests, saving up time and disk space in the process. Default behavior (build them) is kept.

In my machine this lowers build time from 1.5 hours to 18 minutes.


yonghong-song:
[buildbot, test this please]

yonghong-song:
LGTM. Thanks!

PR #3210, opened 2021-01-01
cmake: always link to packaged libbpf if CMAKE_USE_LIBBPF_PACKAGE is set

bluca:
For some binaries, even if CMAKE_USE_LIBBPF_PACKAGE is set static linking is still used, breaking expectations. If the option to use the packaged version is set, it should be respected.


yonghong-song:
[buildbot, test this please]

bluca:
CMake's NAME_WLE was introduced in v3.14, changed to NAME_WE which is available in older releases. Let's see if this works or it needs permissions:

bluca:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

bluca:
The build failure on the Ubuntu CIs was due to the order of linker flags being changed - on some compilers it matters. I've fix it and restored the original order - bpf-static first, then its dependencies (elf/zlib).

bluca:
@yonghong-song gentle ping for review. We will be using these patches in the upcoming Debian 11 release (and as a consequence in the next Ubuntu release and other derivatives). TIA!

yonghong-song:
The change looks good to me. Thanks!

PR #3214, opened 2021-01-06
bcc/libbpf-tools: Use fentry for vfsstat

KentaTada:
Use fentry like vfsstat.py.

Signed-off-by: Kenta Tada <Kenta.Tada@sony.com>

### DIFF
--- a/libbpf-tools/trace_helpers.c
+++ b/libbpf-tools/trace_helpers.c
@@ -500,3 +500,22 @@ slow_path:
 	fclose(f);
 	return false;
 }
+
+bool vmlinux_btf_exists(void)
+{
+	if (!access("/sys/kernel/btf/vmlinux", R_OK))
+		return true;
+	return false;
+}
+
+bool module_btf_exists(const char *mod)
+{
+	char sysfs_mod[80];
+
+	if (mod) {
+		snprintf(sysfs_mod, sizeof(sysfs_mod), "/sys/kernel/btf/%s", mod);
+		if (!access(sysfs_mod, R_OK))
+			return true;
+	}
+	return false;
+}
--- a/libbpf-tools/vfsstat.bpf.c
+++ b/libbpf-tools/vfsstat.bpf.c
@@ -16,31 +16,61 @@ static __always_inline int inc_stats(int key)
 }
 
 SEC("kprobe/vfs_read")
-int BPF_KPROBE(vfs_read)
+int BPF_KPROBE(kprobe_vfs_read)
 {
 	return inc_stats(S_READ);
 }
 
 SEC("kprobe/vfs_write")
-int BPF_KPROBE(vfs_write)
+int BPF_KPROBE(kprobe_vfs_write)
 {
 	return inc_stats(S_WRITE);
 }
 
 SEC("kprobe/vfs_fsync")
-int BPF_KPROBE(vfs_fsync)
+int BPF_KPROBE(kprobe_vfs_fsync)
 {
 	return inc_stats(S_FSYNC);
 }
 
 SEC("kprobe/vfs_open")
-int BPF_KPROBE(vfs_open)
+int BPF_KPROBE(kprobe_vfs_open)
 {
 	return inc_stats(S_OPEN);
 }
 
 SEC("kprobe/vfs_create")
-int BPF_KPROBE(vfs_create)
+int BPF_KPROBE(kprobe_vfs_create)
+{
+	return inc_stats(S_CREATE);
+}
+
+SEC("fentry/vfs_read")
+int BPF_PROG(fentry_vfs_read)
+{
+	return inc_stats(S_READ);
+}
+
+SEC("fentry/vfs_write")
+int BPF_PROG(fentry_vfs_write)
+{
+	return inc_stats(S_WRITE);
+}
+
+SEC("fentry/vfs_fsync")
+int BPF_PROG(fentry_vfs_fsync)
+{
+	return inc_stats(S_FSYNC);
+}
+
+SEC("fentry/vfs_open")
+int BPF_PROG(fentry_vfs_open)
+{
+	return inc_stats(S_OPEN);
+}
+
+SEC("fentry/vfs_create")
+int BPF_PROG(fentry_vfs_create)
 {
 	return inc_stats(S_CREATE);
 }
--- a/libbpf-tools/vfsstat.c
+++ b/libbpf-tools/vfsstat.c
@@ -140,7 +140,7 @@ int main(int argc, char **argv)
 		.doc = argp_program_doc,
 		.args_doc = args_doc,
 	};
-	struct vfsstat_bpf *obj;
+	struct vfsstat_bpf *skel;
 	int err;
 
 	err = argp_parse(&argp, argc, argv, 0, NULL, NULL);
@@ -156,13 +156,34 @@ int main(int argc, char **argv)
 		return 1;
 	}
 
-	obj = vfsstat_bpf__open_and_load();
-	if (!obj) {
-		fprintf(stderr, "failed to open and/or load BPF object\n");
+	skel = vfsstat_bpf__open();
+	if (!skel) {
+		fprintf(stderr, "failed to open BPF skelect\n");
 		return 1;
 	}
 
-	err = vfsstat_bpf__attach(obj);
+	/* It fallbacks to kprobes when kernel does not support fentry. */
+	if (vmlinux_btf_exists() && fentry_exists("vfs_read", NULL)) {
+		bpf_program__set_autoload(skel->progs.kprobe_vfs_read, false);
+		bpf_program__set_autoload(skel->progs.kprobe_vfs_write, false);
+		bpf_program__set_autoload(skel->progs.kprobe_vfs_fsync, false);
+		bpf_program__set_autoload(skel->progs.kprobe_vfs_open, false);
+		bpf_program__set_autoload(skel->progs.kprobe_vfs_create, false);
+	} else {
+		bpf_program__set_autoload(skel->progs.fentry_vfs_read, false);
+		bpf_program__set_autoload(skel->progs.fentry_vfs_write, false);
+		bpf_program__set_autoload(skel->progs.fentry_vfs_fsync, false);
+		bpf_program__set_autoload(skel->progs.fentry_vfs_open, false);
+		bpf_program__set_autoload(skel->progs.fentry_vfs_create, false);
+	}
+
+	err = vfsstat_bpf__load(skel);
+	if (err) {
+		fprintf(stderr, "failed to load BPF skelect: %d\n", err);
+		goto cleanup;
+	}
+
+	err = vfsstat_bpf__attach(skel);
 	if (err) {
 		fprintf(stderr, "failed to attach BPF programs: %s\n",
 				strerror(-err));
@@ -172,11 +193,11 @@ int main(int argc, char **argv)
 	print_header();
 	do {
 		sleep(env.interval);
-		print_and_reset_stats(obj->bss->stats);
+		print_and_reset_stats(skel->bss->stats);
 	} while (!env.count || --env.count);
 
 cleanup:
-	vfsstat_bpf__destroy(obj);
+	vfsstat_bpf__destroy(skel);
 
 	return err != 0;
 }
--- a/libbpf-tools/trace_helpers.h
+++ b/libbpf-tools/trace_helpers.h
@@ -73,4 +73,7 @@ bool fentry_exists(const char *name, const char *mod);
  */
 bool kprobe_exists(const char *name);
 
+bool vmlinux_btf_exists(void);
+bool module_btf_exists(const char *mod);
+
 #endif /* __TRACE_HELPERS_H */


libbpf-tools/vfsstat.bpf.c
```diff
@@ -15,32 +15,32 @@ static __always_inline int inc_stats(int key)
 	return 0;
 }
 
-SEC("kprobe/vfs_read")
-int BPF_KPROBE(vfs_read)
+SEC("fentry/vfs_read")
```
anakryiko, line 18: using fentry/fexit bumps the required minimal Linux version, so it's a regression in the behavior in that sense. Best way to get better performance by using fentry/fexit entries would be to have both kprobe/kretprobe and fentry/fexit and then detect fentry/fexit support at runtime and load (or rather disable auto-load of) only appropriate set of programs

KentaTada: Although the minimum required kernel version was not explicitly stated in README,  I thought we could use BPF trampoline  without considering older kernels.
Actually, fentry/fexit is used in tcpconnlat without considering older kernels:
https://github.com/iovisor/bcc/blob/v0.18.0/libbpf-tools/tcpconnlat.bpf.c#L49

What is the minimum kernel version required to use libbpf-tools?
If the minimum kernel version is 5.5 or later, I think that we do not need to support kprobe/kretprobe for vfsstat.

anakryiko: There is no single minimal kernel version for all libbpf-tools. vfsstat in particular is extremely simple and doesn't use CO-RE, so it can run on very old kernels. It's a good property, so we should try to preserve it. This tool could be a good example on how to opportunistically use fentry/fexit (for better performance), but fallback to kprobe/kretprobe (for wider kernel support).

KentaTada: Thank you for your review.
In addition to that, I remembered that BPF trampoline was only supported on x86_64 currently.
So the fallback to kprobe/kretprobe is needed for not only older kernels support but also multi-architecture support.
I'll try on it.

KentaTada: @anakryiko 
I'm sorry for the late reply.

I tried to implement fallback to kprobe/kretprobe but it was needed to prepare for two skel files which are for kprobes and fentry.
In addition, vfsstat.c needed to implement the abstraction layer to use different tracing mechanism depending on the kernel version and architecture.
I think that it is redundant for all tools to implement same facilities.
So libbpf may provide new APIs to realize fallback. Or, Could you give me your idea if you have?
Anyway, I'll continue this activity when I have some time but chnage this PR to draft.

anakryiko: You shouldn't need two skeletons. Please see recent ext4dist PR on how it is dealing with having both kprobe and fentry functions (https://github.com/iovisor/bcc/pull/3309).

I think before we add some extra functionality to libbpf, let's just first build up libbpf-tools infra to be used by libbpf-tools for starters (and to understand what's essential and generalizable vs what's just specific to these tools). How about we implement feature-probing helpers for common scenarios, just like BPF.support_kfunc in Python side of BCC:

```
 908     @staticmethod¬
 909     def support_kfunc():¬
 910         # there's no trampoline support for other than x86_64 arch¬
 911         if platform.machine() != 'x86_64':¬
 912             return False;¬
 913         if not lib.bpf_has_kernel_btf():¬
 914             return False;¬
 915         # kernel symbol "bpf_trampoline_link_prog" indicates kfunc support¬
 916         if BPF.ksymname("bpf_trampoline_link_prog") != -1:¬
 917             return True¬
 918         return False¬
```

Checking kernel BTF is simple (`access("/sys/kernel/btf/vmlinux", F_OK);`). We probably need a fast implementation for checking the existence of a single (or maybe small set of) kernel symbols. See trace_helpers.c, it has efficient ksyms cache implementation, but that's overkill if you only need to check existence of a symbol. platform.machine() is probably implementable with utsname() call? WDYT?

KentaTada: Thank you for your valuable information.

>You shouldn't need two skeletons. Please see recent ext4dist PR on how it is dealing with having both kprobe and fentry functions (#3309).

I could not execute `ext4dist` in my environment without BTF.
I think that libbpf checks the existence of BTF when it actually loads the skel object using bpf_object__load_skeleton().

I just glanced at https://github.com/libbpf/libbpf/blob/99bc176337071326d7d6b9552142eb6a0450fc88/src/libbpf.c and I found that libbpf checks the existence of BTF at startup if the bpf object needs BTF.
Below code checks whether the bpf object needs BTF	
https://github.com/libbpf/libbpf/blob/99bc176337071326d7d6b9552142eb6a0450fc88/src/libbpf.c#L2588-L2589
For example, above code will not return 0 and execute libbpf_find_kernel_btf() and fail to find BTF if I change `vfsstat.bpf.c` to use fentry. 
So I could not include both SEC() of fentry and kprobes in `vfsstat.bpf.c`.

I apologize if there are any mistakes in my understanding.
I'll investigate it more.

>I think before we add some extra functionality to libbpf, let's just first build up libbpf-tools infra to be used by libbpf-tools for starters (and to understand what's essential and generalizable vs what's just specific to these tools). How about we implement feature-probing helpers for common scenarios, just like BPF.support_kfunc in Python side of BCC:

I agree with your idea.

>Checking kernel BTF is simple (access("/sys/kernel/btf/vmlinux", F_OK);). We probably need a fast implementation for checking the existence of a single (or maybe small set of) kernel symbols. See trace_helpers.c, it has efficient ksyms cache implementation, but that's overkill if you only need to check existence of a symbol. platform.machine() is probably implementable with utsname() call? WDYT?

I'll read trace_helpers.c at first.
Thanks!

KentaTada: >For example, above code will not return 0 and execute libbpf_find_kernel_btf() and fail to find BTF if I change vfsstat.bpf.c to use fentry.

I understand the mechanism of bpf_program__set_autoload() and it bypasses the check at runtime when false.
But below code may not be correct. I think that `if (!should_fallback()) {` is correct instead of `if (should_fallback()) {` 
https://github.com/iovisor/bcc/blob/7fd8cad30f910af0470e00130817442091376ab3/libbpf-tools/ext4dist.c#L195

anakryiko: > For example, above code will not return 0 and execute libbpf_find_kernel_btf() and fail to find BTF if I change `vfsstat.bpf.c` to use fentry.
> So I could not include both SEC() of fentry and kprobes in `vfsstat.bpf.c`.

You can have fentry in your BPF code and if you will do `bpf_program__set_autoload(false)` on it, vmlinux BTF won't be required by libbpf.

anakryiko: > > For example, above code will not return 0 and execute libbpf_find_kernel_btf() and fail to find BTF if I change vfsstat.bpf.c to use fentry.
> 
> I understand the mechanism of bpf_program__set_autoload() and it bypasses the check at runtime when false.
> But below code may not be correct. I think that `if (!should_fallback()) {` is correct instead of `if (should_fallback()) {` . Or should_fallback() should return false when BTF is supported.
> 
> https://github.com/iovisor/bcc/blob/7fd8cad30f910af0470e00130817442091376ab3/libbpf-tools/ext4dist.c#L195
> 
> I'll create PR to fix above issue and some minor issues.

I think that code is correct. The difference between what you do and what ext4dist is doing is that ext4dist will always try to use fentry, unless it detects that ext4 is a kernel module and kernel doesn't support kernel module BTF.

We might change it to be still more conservative and fallback to kprobe, if no fentry and vmlinux BTF are supported, but that's a separate issue. As is, the logic is valid: it disables fentry if it is detected that ext4 is a module and module BTF isn't yet supported.

KentaTada:
@anakryiko 
As of now, I updated the fentry support for vfsstat in a simple way although I read https://github.com/iovisor/bcc/pull/3347#discussion_r612096999
Off course, I'll change the implementation when new helper methods are ready.
Could you review the current code?

libbpf-tools/vfsstat.c
```diff
@@ -132,6 +132,13 @@ static void print_and_reset_stats(__u64 stats[S_MAXSTAT])
 	printf("\n");
 }
 
+static bool is_btf_enabled(void)
+{
+	if (!access("/sys/kernel/btf/vmlinux", R_OK))
```
anakryiko, line 137: the problem is that /sys/kernel/btf/vmlinux is present in 5.4 kernel, while fentry/fexit is supported starting from 5.5, so on 5.4 vfsstat will be failing

We should probably implement helper to detect if fentry programs are supported and use that everywhere. I'll reply with some more details on the other PR.

KentaTada: I'm sorry I forgot to check the kernel version.
I'll wait for helpers.
Thank you for your kind help!

KentaTada:
@anakryiko 
I updated vfsstat.c using fentry_exists().

I have one concern about the error message when it fallbacks to kprobes.
When it fallbacks to kprobes, the below messages are always printed to stderr.
```
# ./vfsstat
failed to parse vmlinux BTF at '/sys/kernel/btf/vmlinux': No such file or directory
failed to parse vmlinux BTF at '/sys/kernel/btf/vmlinux': No such file or directory
failed to parse vmlinux BTF at '/sys/kernel/btf/vmlinux': No such file or directory
failed to parse vmlinux BTF at '/sys/kernel/btf/vmlinux': No such file or directory
failed to parse vmlinux BTF at '/sys/kernel/btf/vmlinux': No such file or directory
TIME         READ/s  WRITE/s  FSYNC/s   OPEN/s CREATE/s
21:53:20:      1785     1145        0       94        0
```
It's a trivial but we may include the message that fallback is succeeded and kprobes is used instead of fentry.

anakryiko:
> ```
> # ./vfsstat
> failed to parse vmlinux BTF at '/sys/kernel/btf/vmlinux': No such file or directory
> failed to parse vmlinux BTF at '/sys/kernel/btf/vmlinux': No such file or directory
> failed to parse vmlinux BTF at '/sys/kernel/btf/vmlinux': No such file or directory
> failed to parse vmlinux BTF at '/sys/kernel/btf/vmlinux': No such file or directory
> failed to parse vmlinux BTF at '/sys/kernel/btf/vmlinux': No such file or directory
> TIME         READ/s  WRITE/s  FSYNC/s   OPEN/s CREATE/s
> 21:53:20:      1785     1145        0       94        0
> ```
> 
> It's a trivial but we may include the message that fallback is succeeded and kprobes is used instead of fentry.

No, it's very bad for usability. Can you please add access() checks for /sys/kernel/btf/vmlinux and /sys/kernel/btf/<module>. It's not an error if those are missing, it's an error only if it's present and we can't parse them.

libbpf-tools/vfsstat.c
```diff
@@ -156,13 +156,46 @@ int main(int argc, char **argv)
...
-		fprintf(stderr, "failed to open and/or load BPF object\n");
+	skel = vfsstat_bpf__open();
+	if (!skel) {
+		fprintf(stderr, "failed to open BPF skelect\n");
 		return 1;
 	}
 
-	err = vfsstat_bpf__attach(obj);
+	/* It fallbacks to kprobes when kernel does not support fentry. */
+	if (fentry_exists("vfs_read", NULL)) {
```
anakryiko, line 166: If fentry for vfs_read is supported ("exists"), there is no need to repeat the check for other vfs_xxx operations. Please roll everything under single if/else check.

anakryiko (changes requested):

libbpf-tools/vfsstat.c
KentaTada: Fixed it

KentaTada:
>No, it's very bad for usability. Can you please add access() checks for /sys/kernel/btf/vmlinux and /sys/kernel/btf/. It's not an error if those are missing, it's an error only if it's present and we can't parse them.

I added two helpers in `trace_helpers.c` because check functions are needed for every tools to use fentry.

libbpf-tools/trace_helpers.c
```diff
@@ -500,3 +500,22 @@ bool kprobe_exists(const char *name)
 	fclose(f);
 	return false;
 }
+
+bool is_btf_enabled_for_kernel(void)
```
anakryiko, line 504: maybe let's call them consistent with _exists pattern for fentry/fexit: vmlinux_btf_exists() and module_btf_exists()?

anakryiko (approved):
Other than the naming suggestion looks good. Thanks.

libbpf-tools/trace_helpers.c
KentaTada: I agree with you.
I changed the name of functions.
Thanks!

KentaTada:
@anakryiko @yonghong-song
Could you review my modification?
Especially, I want to use vmlinux_btf_exists() and module_btf_exists() for other tools.

anakryiko:
I already approved it. It still looks good :) @yonghong-song, feel free to merge it. Thanks!

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

PR #3217, opened 2021-01-07
add tool: cpuload

liuchao173:
When the CPU is not idle for a period of time, it displays the tasks which run out of CPU resources.


yonghong-song:
[buildbot, test this please]

yonghong-song:
The `atop` tool should be able to capture tasks with hot cpu usage. Have you tried `atop`? I try to understand what is the advantage of this tool vs. `atop`.

liuchao173:
> The `atop` tool should be able to capture tasks with hot cpu usage. Have you tried `atop`? I try to understand what is the advantage of this tool vs. `atop`.

The minimum precision of `atop` is one second, and cpuload only print when cpu load is high. It focus on hot cpu usage.

brendangregg:
Thanks, I'm trying to see the use case for this. We have `pidstat` for rolling summaries of CPU usage, `perf sched` for event logging including runtimes, and then BCC/bpftrace tools including runqlen/runqlat for debugging scheduler latency, and cpuunclaimed for identifying scheduler failures (failure to run when idle CPUs present).

The example file has:

```cpuload can be useful for discovering which tasks cause CPU high load.```

I don't think that's a good use case given how many other tools already do this. I'd identify other strong use cases or not include this tool.

PR #3218, opened 2021-01-08
Remove libbcc-no-libbpf shared library, change libbcc linkage instead

bluca:
The current upstream split does not work very well, as the SONAME is mangled so
nothing recognises it. Applications link against libbcc, not libbcc_no_bpf.
Remove it entirely, and switch the libbcc.so to dynamically
link with libbpf if CMAKE_USE_LIBBPF_PACKAGE is set.

### DIFF
--- a/src/cc/CMakeLists.txt
+++ b/src/cc/CMakeLists.txt
@@ -83,6 +83,9 @@ set(bcc_sym_sources bcc_syms.cc bcc_elf.c bcc_perf_map.c bcc_proc.c)
 set(bcc_common_headers libbpf.h perf_reader.h "${CMAKE_CURRENT_BINARY_DIR}/bcc_version.h")
 set(bcc_table_headers file_desc.h table_desc.h table_storage.h)
 set(bcc_api_headers bcc_common.h bpf_module.h bcc_exception.h bcc_syms.h bcc_proc.h bcc_elf.h)
+if(LIBBPF_FOUND)
+  set(bcc_common_sources ${bcc_common_sources} libbpf.c perf_reader.c)
+endif()
 
 if(ENABLE_CLANG_JIT)
 add_library(bcc-shared SHARED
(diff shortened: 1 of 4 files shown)


bluca:
@yonghong-song if removing the no-bcc flavour entirely is too much, although I really don't think it can be used that much due to the wrong SONAME in the elf headers, I am OK with doing additional work to add a new config option or make the existing one tri-state (off, on, enforce).

yonghong-song:
[buildbot, test this please]

src/cc/CMakeLists.txt
```diff
@@ -123,41 +116,42 @@ set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} ${llvm_lib_exclude_f
...
   list(APPEND bcc_common_libs_for_s usdt-static)
-  list(APPEND bcc_common_libs_for_n usdt-static)
   list(APPEND bcc_common_libs_for_lua usdt-static)
 endif()
 
 add_subdirectory(frontends)
 
 # Link against LLVM libraries
-target_link_libraries(bcc-shared ${bcc_common_libs_for_s})
+target_link_libraries(bcc-shared ${bcc_common_libs_for_s} ${LIBBPF_LIBRARIES})
```
yonghong-song, line 151: ${LIBBPF_LIBRARIES} is not needed here since ${bcc_common_libs_for_s} has already linked with it, right?

yonghong-song:
> @yonghong-song if removing the no-bcc flavour entirely is too much, although I really don't think it can be used that much due to the wrong SONAME in the elf headers, I am OK with doing additional work to add a new config option or make the existing one tri-state (off, on, enforce).

I do not think a lot of people use it yet. Also it just a library name change to make it work at the same time it simplifies thing a lot. So I am okay with this.

yonghong-song:
[buildbot, ok to test]

src/cc/CMakeLists.txt
bluca: Yes indeed, removed.

PR #3225, opened 2021-01-13
Revert "Remove libbcc-no-libbpf shared library, change libbcc linkage…

olsajiri:
I'm surprised you'd remove library that's being used.

bpftrace in fedora is using it:
  -DLIBBCC_LIBRARIES:PATH=/usr/lib64/libbcc-no-libbpf.so

It's the only way to have libcc together with dynamicaly linked libbpf.so.

Please let's keep it in.

### DIFF
--- a/examples/cpp/CMakeLists.txt
+++ b/examples/cpp/CMakeLists.txt
@@ -19,7 +19,7 @@ foreach(EXAMPLE ${EXAMPLES})
   if(NOT CMAKE_USE_LIBBPF_PACKAGE)
     target_link_libraries(${NAME} bcc-static)
   else()
-    target_link_libraries(${NAME} bcc-shared)
+    target_link_libraries(${NAME} bcc-shared-no-libbpf)
   endif()
 
   if(INSTALL_CPP_EXAMPLES)
--- a/examples/cpp/pyperf/CMakeLists.txt
+++ b/examples/cpp/pyperf/CMakeLists.txt
@@ -10,7 +10,7 @@ target_link_libraries(PyPerf bcc-static)
 if(NOT CMAKE_USE_LIBBPF_PACKAGE)
   target_link_libraries(PyPerf bcc-static)
 else()
-  target_link_libraries(PyPerf bcc-shared)
+  target_link_libraries(PyPerf bcc-shared-no-libbpf)
 endif()
 
 if(INSTALL_CPP_EXAMPLES)
--- a/src/cc/CMakeLists.txt
+++ b/src/cc/CMakeLists.txt
@@ -83,9 +83,6 @@ set(bcc_sym_sources bcc_syms.cc bcc_elf.c bcc_perf_map.c bcc_proc.c)
 set(bcc_common_headers libbpf.h perf_reader.h "${CMAKE_CURRENT_BINARY_DIR}/bcc_version.h")
 set(bcc_table_headers file_desc.h table_desc.h table_storage.h)
 set(bcc_api_headers bcc_common.h bpf_module.h bcc_exception.h bcc_syms.h bcc_proc.h bcc_elf.h)
-if(LIBBPF_FOUND)
-  set(bcc_common_sources ${bcc_common_sources} libbpf.c perf_reader.c)
-endif()
 
 if(ENABLE_CLANG_JIT)
 add_library(bcc-shared SHARED
@@ -93,6 +90,16 @@ add_library(bcc-shared SHARED
   ${bcc_util_sources})
 set_target_properties(bcc-shared PROPERTIES VERSION ${REVISION_LAST} SOVERSION 0)
 set_target_properties(bcc-shared PROPERTIES OUTPUT_NAME bcc)
+
+# If there's libbpf detected we build the libbcc-no-libbpf.so library, that
+# dynamicaly links libbpf.so, in comparison to static link in libbcc.so.
+if(LIBBPF_FOUND)
+  add_library(bcc-shared-no-libbpf SHARED
+    link_all.cc ${bcc_common_sources} ${bcc_table_sources} ${bcc_sym_sources}
+    ${bcc_util_sources} libbpf.c perf_reader.c)
+  set_target_properties(bcc-shared-no-libbpf PROPERTIES VERSION ${REVISION_LAST} SOVERSION 0)
+  set_target_properties(bcc-shared-no-libbpf PROPERTIES OUTPUT_NAME bcc-no-libbpf)
+endif()
 
 if(ENABLE_USDT)
   set(bcc_usdt_sources usdt/usdt.cc usdt/usdt_args.cc)
@@ -116,25 +123,18 @@ set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} ${llvm_lib_exclude_f
 set(bcc_common_libs b_frontend clang_frontend
   -Wl,--whole-archive ${clang_libs} ${llvm_libs} -Wl,--no-whole-archive
   ${LIBELF_LIBRARIES})
-set(bcc_common_libs_for_a ${bcc_common_libs})
-set(bcc_common_libs_for_s ${bcc_common_libs})
-set(bcc_common_libs_for_lua b_frontend clang_frontend
+set(bcc_common_libs_for_a ${bcc_common_libs} bpf-static)
+set(bcc_common_libs_for_s ${bcc_common_libs} bpf-static)
+set(bcc_common_libs_for_n ${bcc_common_libs})
+set(bcc_common_libs_for_lua b_frontend clang_frontend bpf-static
   ${clang_libs} ${llvm_libs} ${LIBELF_LIBRARIES})
-if(LIBBPF_FOUND)
-  list(APPEND bcc_common_libs_for_a ${LIBBPF_LIBRARIES})
-  list(APPEND bcc_common_libs_for_s ${LIBBPF_LIBRARIES})
-  list(APPEND bcc_common_libs_for_lua ${LIBBPF_LIBRARIES})
-else()
-  list(APPEND bcc_common_libs_for_a bpf-static)
-  list(APPEND bcc_common_libs_for_s bpf-static)
-  list(APPEND bcc_common_libs_for_lua bpf-static)
-endif()
 
 if(ENABLE_CPP_API)
   add_subdirectory(api)
   list(APPEND bcc_common_libs_for_a api-static)
   # Keep all API functions
   list(APPEND bcc_common_libs_for_s -Wl,--whole-archive api-static -Wl,--no-whole-archive)
+  list(APPEND bcc_common_libs_for_n -Wl,--whole-archive api-static -Wl,--no-whole-archive)
 endif()
 
 if(ENABLE_USDT)
@@ -142,6 +142,7 @@ if(ENABLE_USDT)
   add_subdirectory(usdt)
   list(APPEND bcc_common_libs_for_a usdt-static)
   list(APPEND bcc_common_libs_for_s usdt-static)
+  list(APPEND bcc_common_libs_for_n usdt-static)
   list(APPEND bcc_common_libs_for_lua usdt-static)
 endif()
 
@@ -151,6 +152,11 @@ add_subdirectory(frontends)
 target_link_libraries(bcc-shared ${bcc_common_libs_for_s})
 target_link_libraries(bcc-static ${bcc_common_libs_for_a} bcc-loader-static)
 set(bcc-lua-static ${bcc-lua-static} ${bcc_common_libs_for_lua})
+
+if(LIBBPF_FOUND)
+  target_link_libraries(bcc-shared-no-libbpf ${bcc_common_libs_for_n} ${LIBBPF_LIBRARIES})
+  install(TARGETS bcc-shared-no-libbpf LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR})
+endif()
 
 install(TARGETS bcc-shared bcc-static bcc-loader-static bpf-static LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
 install(FILES ${bcc_table_headers} DESTINATION include/bcc)
--- a/tests/cc/CMakeLists.txt
+++ b/tests/cc/CMakeLists.txt
@@ -10,7 +10,7 @@ add_executable(test_static test_static.c)
 if(NOT CMAKE_USE_LIBBPF_PACKAGE)
   target_link_libraries(test_static bcc-static)
 else()
-  target_link_libraries(test_static bcc-shared)
+  target_link_libraries(test_static bcc-shared-no-libbpf)
 endif()
 
 add_test(NAME c_test_static COMMAND ${TEST_WRAPPER} c_test_static sudo ${CMAKE_CURRENT_BINARY_DIR}/test_static)
@@ -55,11 +55,11 @@ endif()
 
 if(LIBBPF_FOUND)
   add_executable(test_libbcc_no_libbpf ${TEST_LIBBCC_SOURCES})
-  add_dependencies(test_libbcc_no_libbpf bcc-shared)
+  add_dependencies(test_libbcc_no_libbpf bcc-shared-no-libbpf)
 
-  target_link_libraries(test_libbcc_no_libbpf ${PROJECT_BINARY_DIR}/src/cc/libbcc.so dl usdt_test_lib ${LIBBPF_LIBRARIES})
+  target_link_libraries(test_libbcc_no_libbpf ${PROJECT_BINARY_DIR}/src/cc/libbcc-no-libbpf.so dl usdt_test_lib ${LIBBPF_LIBRARIES})
   set_target_properties(test_libbcc_no_libbpf PROPERTIES INSTALL_RPATH ${PROJECT_BINARY_DIR}/src/cc)
-  target_compile_definitions(test_libbcc_no_libbpf PRIVATE -DLIBBCC_NAME=\"libbcc.so\")
+  target_compile_definitions(test_libbcc_no_libbpf PRIVATE -DLIBBCC_NAME=\"libbcc-no-libbpf.so\")
 
   add_test(NAME test_libbcc_no_libbpf COMMAND ${TEST_WRAPPER} c_test_all_no_libbpf sudo ${CMAKE_CURRENT_BINARY_DIR}/test_libbcc_no_libbpf)
 endif()


yonghong-song:
[buildbot, test this please]

yonghong-song:
I do see @bluca approach is to static link libbpf package libbpf.a into shared libbcc.so on the system. libbcc-no-libbpf.so already became an ABI, so agree that we need to revert it. Sorry about this.

I guess either debian can also use libbcc-no-libbpf.so or they want libbcc.so to statically link with host libbpf.a, a cmake
option can be added to control this behavior.

yonghong-song:
@olsajiri looks like you already reverted commit version for libbpf submodule, could you checkout latest master, do a revert and resubmit?

yonghong-song:
[buildbot, ok to test]

bluca:
> It's the only way to have libcc together with dynamicaly linked libbpf.so.

This is not true though? There's the actual option to have the real library, with the real soname, dynamically linked - CMAKE_USE_LIBBPF_PACKAGE

bluca:
> > It's the only way to have libcc together with dynamicaly linked libbpf.so.
> 
> This is not true though? There's the actual option to have the real library, with the real soname, dynamically linked - CMAKE_USE_LIBBPF_PACKAGE

I mean if you have a dependency on the elf library with the wrong soname, that's fine I guess, but please add it as an option. It doesn't match the pkg-config interface, the package names, nothing.

olsajiri:
> > It's the only way to have libcc together with dynamicaly linked libbpf.so.
> 
> This is not true though? There's the actual option to have the real library, with the real soname, dynamically linked - CMAKE_USE_LIBBPF_PACKAGE

yes, but people expect libbcc.so without libbpf.so dependency
having libbpf statically linked inside

so we had to add libbcc-no-libbpf.so to be able to use it with libbpf.so

bluca:
> > > It's the only way to have libcc together with dynamicaly linked libbpf.so.
> > 
> > 
> > This is not true though? There's the actual option to have the real library, with the real soname, dynamically linked - CMAKE_USE_LIBBPF_PACKAGE
> 
> yes, but people expect libbcc.so without libbpf.so dependency
> having libbpf statically linked inside

I might be missing something, but if someone wants libbcc statically linked, the option can be turned off? It's off by default in fact, one has to opt-in

> so we had to add libbcc-no-libbpf.so to be able to use it with libbpf.so

This is not really usable, I'm afraid. The soname is wrong. pkg-config doesn't match. I mean it's fine as an option I guess, but there needs to be a way to do the right thing as well.

olsajiri:
> @olsajiri looks like you already reverted commit version for libbpf submodule, could you checkout latest master, do a revert and resubmit?

sorry, re-pushed

olsajiri:
> > > > It's the only way to have libcc together with dynamicaly linked libbpf.so.
> > > 
> > > 
> > > This is not true though? There's the actual option to have the real library, with the real soname, dynamically linked - CMAKE_USE_LIBBPF_PACKAGE
> > 
> > 
> > yes, but people expect libbcc.so without libbpf.so dependency
> > having libbpf statically linked inside
> 
> I might be missing something, but if someone wants libbcc statically linked, the option can be turned off? It's off by default in fact, one has to opt-in
> 
> > so we had to add libbcc-no-libbpf.so to be able to use it with libbpf.so
> 
> This is not really usable, I'm afraid. The soname is wrong. pkg-config doesn't match. I mean it's fine as an option I guess, but there needs to be a way to do the right thing as well.

I need to check new bcc code, but at the time we were adding this, it was the only way
to have libbcc without libbpf being statically linked in it, which causes problems for programs
that dynamically linked libbpf, like bpftrace

d4b3bf03d9c6 Add libbcc-no-libbpf.so library

bluca:
> > > > > It's the only way to have libcc together with dynamicaly linked libbpf.so.
> > > > 
> > > > 
> > > > This is not true though? There's the actual option to have the real library, with the real soname, dynamically linked - CMAKE_USE_LIBBPF_PACKAGE
> > > 
> > > 
> > > yes, but people expect libbcc.so without libbpf.so dependency
> > > having libbpf statically linked inside
> > 
> > 
> > I might be missing something, but if someone wants libbcc statically linked, the option can be turned off? It's off by default in fact, one has to opt-in
> > > so we had to add libbcc-no-libbpf.so to be able to use it with libbpf.so
> > 
> > 
> > This is not really usable, I'm afraid. The soname is wrong. pkg-config doesn't match. I mean it's fine as an option I guess, but there needs to be a way to do the right thing as well.
> 
> I need to check new bcc code, but at the time we were adding this, it was the only way
> to have libbcc without libbpf being statically linked in it, which causes problems for programs
> that dynamically linked libbpf, like bpftrace
> 
> [d4b3bf0](https://github.com/iovisor/bcc/commit/d4b3bf03d9c67324ff85f9d3c74b201b7e427615) Add libbcc-no-libbpf.so library

Right, the commit you are reverting makes libbcc dynamically link with libbpf when CMAKE_USE_LIBBPF_PACKAGE is used. Isn't that what you want? It's opt-in, and users who don't want that simply don't set it and get the static linking.

yonghong-song:
@olsajiri could you check whether bpftrace works properly with `-DLIBBCC_LIBRARIES:PATH=/usr/lib64/libbcc.so` or not if CMAKE_USE_LIBBPF_PACKAGE is used to build bcc?

olsajiri:
> @olsajiri could you check whether bpftrace works properly with `-DLIBBCC_LIBRARIES:PATH=/usr/lib64/libbcc.so` or not if CMAKE_USE_LIBBPF_PACKAGE is used to build bcc?

it probably will work, I'll try

but the point is that I was told libbcc needs to stay as is with libbpf compiled in,
so that's why we created with libbcc-no-libbpf

I think I won't be able to persuade libbcc fedora maintainer to change that,
if there's a chance we will break existing users

bluca:
> > @olsajiri could you check whether bpftrace works properly with `-DLIBBCC_LIBRARIES:PATH=/usr/lib64/libbcc.so` or not if CMAKE_USE_LIBBPF_PACKAGE is used to build bcc?
> 
> it probably will work, I'll try
> 
> but the point is that I was told libbcc needs to stay as is with libbpf compiled in,
> so that's why we created with libbcc-no-libbpf

And that will keep happening with CMAKE_USE_LIBBPF_PACKAGE=off which is the default, as mentioned, so it shouldn't be a problem? Unless I'm missing something.

Again, if you also want at the same time to keep the separate elf shared object with the broken soname that doesn't match pkg-config, I don't have any problem with that - but please add a separate option that can be disabled independently so that we can also do the right thing as well. Thanks!

olsajiri:
> > > @olsajiri could you check whether bpftrace works properly with `-DLIBBCC_LIBRARIES:PATH=/usr/lib64/libbcc.so` or not if CMAKE_USE_LIBBPF_PACKAGE is used to build bcc?
> > 
> > 
> > it probably will work, I'll try
> > but the point is that I was told libbcc needs to stay as is with libbpf compiled in,
> > so that's why we created with libbcc-no-libbpf
> 
> And that will keep happening with CMAKE_USE_LIBBPF_PACKAGE=off which is the default, as mentioned, so it shouldn't be a problem? Unless I'm missing something.

well, we need libbcc (with libbpf compiled in) and libbcc-no-libbpf at the same time
compiled for bcc package

and that was the case until this commit:
 300296a59861 cmake: always link to packaged libbpf if CMAKE_USE_LIBBPF_PACKAGE is set (#3210)

> 
> Again, if you also want at the same time to keep the separate elf shared object with the broken soname that doesn't match pkg-config, I don't have any problem with that - but please add a separate option that can be disabled independently so that we can also do the right thing as well. Thanks!

what's wrong with the soname?

$ readelf -a /usr/lib64/libbcc.so.0 | grep SONAME
 0x000000000000000e (SONAME)             Library soname: [libbcc.so.0]

$ readelf -a /usr/lib64/libbcc-no-libbpf.so.0 | grep SONAME
 0x000000000000000e (SONAME)             Library soname: [libbcc-no-libbpf.so.0]

thanks

bluca:
> > > > @olsajiri could you check whether bpftrace works properly with `-DLIBBCC_LIBRARIES:PATH=/usr/lib64/libbcc.so` or not if CMAKE_USE_LIBBPF_PACKAGE is used to build bcc?
> > > 
> > > 
> > > it probably will work, I'll try
> > > but the point is that I was told libbcc needs to stay as is with libbpf compiled in,
> > > so that's why we created with libbcc-no-libbpf
> > 
> > 
> > And that will keep happening with CMAKE_USE_LIBBPF_PACKAGE=off which is the default, as mentioned, so it shouldn't be a problem? Unless I'm missing something.
> 
> well, we need libbcc (with libbpf compiled in) and libbcc-no-libbpf at the same time
> compiled for bcc package
> 
> and that was the case until this commit:
> [300296a](https://github.com/iovisor/bcc/commit/300296a598613912df5dc61f4b327b7102e52011) cmake: always link to packaged libbpf if CMAKE_USE_LIBBPF_PACKAGE is set (#3210)

Sorry, maybe I misunderstood - I thought the different cases were needed by different people in different situations? Are you saying that you need that _at the exact same time for the exact same build_ ? What's the use case, if you don't mind me asking?

> > Again, if you also want at the same time to keep the separate elf shared object with the broken soname that doesn't match pkg-config, I don't have any problem with that - but please add a separate option that can be disabled independently so that we can also do the right thing as well. Thanks!
> 
> what's wrong with the soname?
> 
> $ readelf -a /usr/lib64/libbcc.so.0 | grep SONAME
> 0x000000000000000e (SONAME) Library soname: [libbcc.so.0]
> 
> $ readelf -a /usr/lib64/libbcc-no-libbpf.so.0 | grep SONAME
> 0x000000000000000e (SONAME) Library soname: [libbcc-no-libbpf.so.0]
> 
> thanks

It's mangled with the no-libbpf suffix, so it doesn't match the pkg-config metadata. It also makes it seem like it's a separate interface with a separate ABI, when in reality it's exactly the same thing, which is a bit weird.

olsajiri:
> > > > > @olsajiri could you check whether bpftrace works properly with `-DLIBBCC_LIBRARIES:PATH=/usr/lib64/libbcc.so` or not if CMAKE_USE_LIBBPF_PACKAGE is used to build bcc?
> > > > 
> > > > 
> > > > it probably will work, I'll try
> > > > but the point is that I was told libbcc needs to stay as is with libbpf compiled in,
> > > > so that's why we created with libbcc-no-libbpf
> > > 
> > > 
> > > And that will keep happening with CMAKE_USE_LIBBPF_PACKAGE=off which is the default, as mentioned, so it shouldn't be a problem? Unless I'm missing something.
> > 
> > 
> > well, we need libbcc (with libbpf compiled in) and libbcc-no-libbpf at the same time
> > compiled for bcc package
> > and that was the case until this commit:
> > [300296a](https://github.com/iovisor/bcc/commit/300296a598613912df5dc61f4b327b7102e52011) cmake: always link to packaged libbpf if CMAKE_USE_LIBBPF_PACKAGE is set (#3210)
> 
> Sorry, maybe I misunderstood - I thought the different cases were needed by different people in different situations? Are you saying that you need that _at the exact same time for the exact same build_ ? What's the use case, if you don't mind me asking?

bcc package providing libraries is the use case

it provides libbcc (with libbpf compiled in)
and libbcc-no-libbpf that we use to compile bpftrace with

when we compile bpftrace with libbcc, the dynamic linker gets confused,
because of the extra compiled libbpf inside libbcc

my first fix was to have libcc dynamically linked with libbpf (like you did),
but I was told libbcc needs to stay as is because:
  1) libbpf package is not widespread
  2) libbcc is using latest libbpf from git submodule and package will not keep up
  3) peoples' apps linking libbcc will be broken 

while 1) and 2) might not be a big issue anymore, I'm worried about 3)

> 
> > > Again, if you also want at the same time to keep the separate elf shared object with the broken soname that doesn't match pkg-config, I don't have any problem with that - but please add a separate option that can be disabled independently so that we can also do the right thing as well. Thanks!
> > 
> > 
> > what's wrong with the soname?
> > $ readelf -a /usr/lib64/libbcc.so.0 | grep SONAME
> > 0x000000000000000e (SONAME) Library soname: [libbcc.so.0]
> > $ readelf -a /usr/lib64/libbcc-no-libbpf.so.0 | grep SONAME
> > 0x000000000000000e (SONAME) Library soname: [libbcc-no-libbpf.so.0]
> > thanks
> 
> It's mangled with the no-libbpf suffix, so it doesn't match the pkg-config metadata. It also makes it seem like it's a separate interface with a separate ABI, when in reality it's exactly the same thing, which is a bit weird.

I guess we can just fix pkg-config metadata then?

bluca:
> > > > > > @olsajiri could you check whether bpftrace works properly with `-DLIBBCC_LIBRARIES:PATH=/usr/lib64/libbcc.so` or not if CMAKE_USE_LIBBPF_PACKAGE is used to build bcc?
> > > > > 
> > > > > 
> > > > > it probably will work, I'll try
> > > > > but the point is that I was told libbcc needs to stay as is with libbpf compiled in,
> > > > > so that's why we created with libbcc-no-libbpf
> > > > 
> > > > 
> > > > And that will keep happening with CMAKE_USE_LIBBPF_PACKAGE=off which is the default, as mentioned, so it shouldn't be a problem? Unless I'm missing something.
> > > 
> > > 
> > > well, we need libbcc (with libbpf compiled in) and libbcc-no-libbpf at the same time
> > > compiled for bcc package
> > > and that was the case until this commit:
> > > [300296a](https://github.com/iovisor/bcc/commit/300296a598613912df5dc61f4b327b7102e52011) cmake: always link to packaged libbpf if CMAKE_USE_LIBBPF_PACKAGE is set (#3210)
> > 
> > 
> > Sorry, maybe I misunderstood - I thought the different cases were needed by different people in different situations? Are you saying that you need that _at the exact same time for the exact same build_ ? What's the use case, if you don't mind me asking?
> 
> bcc package providing libraries is the use case
> 
> it provides libbcc (with libbpf compiled in)
> and libbcc-no-libbpf that we use to compile bpftrace with
> 
> when we compile bpftrace with libbcc, the dynamic linker gets confused,
> because of the extra compiled libbpf inside libbcc
> 
> my first fix was to have libcc dynamically linked with libbpf (like you did),
> but I was told libbcc needs to stay as is because:
> 
> 1. libbpf package is not widespread
> 2. libbcc is using latest libbpf from git submodule and package will not keep up

Those only applies to local builds though, it doesn't apply to a distribution (where in fact it's the opposite - vendoring means out of date libbpf that doesn't track latest changes and always falls behind). The existing option allows users to do this, and it's the default so there should be no surprise nor regressions for direct users and developers, right?

> 3. peoples' apps linking libbcc will be broken
> 
> while 1) and 2) might not be a big issue anymore, I'm worried about 3)

How would that happen? The breakages, as you said, come from the linker getting confused because of multiple built-in and dynamic objects. What cases are there where the opposite is true?

> > > > Again, if you also want at the same time to keep the separate elf shared object with the broken soname that doesn't match pkg-config, I don't have any problem with that - but please add a separate option that can be disabled independently so that we can also do the right thing as well. Thanks!
> > > 
> > > 
> > > what's wrong with the soname?
> > > $ readelf -a /usr/lib64/libbcc.so.0 | grep SONAME
> > > 0x000000000000000e (SONAME) Library soname: [libbcc.so.0]
> > > $ readelf -a /usr/lib64/libbcc-no-libbpf.so.0 | grep SONAME
> > > 0x000000000000000e (SONAME) Library soname: [libbcc-no-libbpf.so.0]
> > > thanks
> > 
> > 
> > It's mangled with the no-libbpf suffix, so it doesn't match the pkg-config metadata. It also makes it seem like it's a separate interface with a separate ABI, when in reality it's exactly the same thing, which is a bit weird.
> 
> I guess we can just fix pkg-config metadata then?

By all means feel free to add an entirely separate metadata file, but that doesn't work for me at all. The real thing being the real thing is required. This is just one library out of millions, it's not special in any way, and having special casing and special handling and special everything beyond a build-time config flag is just unsustainable, sorry. The build flag can be opt-in and disabled by default, that's not an issue.

olsajiri:
> > > > > > > @olsajiri could you check whether bpftrace works properly with `-DLIBBCC_LIBRARIES:PATH=/usr/lib64/libbcc.so` or not if CMAKE_USE_LIBBPF_PACKAGE is used to build bcc?
> > > > > > 
> > > > > > 
> > > > > > it probably will work, I'll try
> > > > > > but the point is that I was told libbcc needs to stay as is with libbpf compiled in,
> > > > > > so that's why we created with libbcc-no-libbpf
> > > > > 
> > > > > 
> > > > > And that will keep happening with CMAKE_USE_LIBBPF_PACKAGE=off which is the default, as mentioned, so it shouldn't be a problem? Unless I'm missing something.
> > > > 
> > > > 
> > > > well, we need libbcc (with libbpf compiled in) and libbcc-no-libbpf at the same time
> > > > compiled for bcc package
> > > > and that was the case until this commit:
> > > > [300296a](https://github.com/iovisor/bcc/commit/300296a598613912df5dc61f4b327b7102e52011) cmake: always link to packaged libbpf if CMAKE_USE_LIBBPF_PACKAGE is set (#3210)
> > > 
> > > 
> > > Sorry, maybe I misunderstood - I thought the different cases were needed by different people in different situations? Are you saying that you need that _at the exact same time for the exact same build_ ? What's the use case, if you don't mind me asking?
> > 
> > 
> > bcc package providing libraries is the use case
> > it provides libbcc (with libbpf compiled in)
> > and libbcc-no-libbpf that we use to compile bpftrace with
> > when we compile bpftrace with libbcc, the dynamic linker gets confused,
> > because of the extra compiled libbpf inside libbcc
> > my first fix was to have libcc dynamically linked with libbpf (like you did),
> > but I was told libbcc needs to stay as is because:
> > 
> > 1. libbpf package is not widespread
> > 2. libbcc is using latest libbpf from git submodule and package will not keep up
> 
> Those only applies to local builds though, it doesn't apply to a distribution (where in fact it's the opposite - vendoring means out of date libbpf that doesn't track latest changes and always falls behind). The existing option allows users to do this, and it's the default so there should be no surprise nor regressions for direct users and developers, right?
> 
> > 1. peoples' apps linking libbcc will be broken
> > 
> > while 1) and 2) might not be a big issue anymore, I'm worried about 3)
> 
> How would that happen? The breakages, as you said, come from the linker getting confused because of multiple built-in and dynamic objects. What cases are there where the opposite is true?

missing libbpf dependency

but as you mentioned above that's probably not big deal for distributions,
because they provide the needed libbpf, just local builds, maybe I could
sell this to bcc maintainer, I'll ask

I tested bpftrace with libbcc and it seems to work

bluca:
> > > > > > > > @olsajiri could you check whether bpftrace works properly with `-DLIBBCC_LIBRARIES:PATH=/usr/lib64/libbcc.so` or not if CMAKE_USE_LIBBPF_PACKAGE is used to build bcc?
> > > > > > > 
> > > > > > > 
> > > > > > > it probably will work, I'll try
> > > > > > > but the point is that I was told libbcc needs to stay as is with libbpf compiled in,
> > > > > > > so that's why we created with libbcc-no-libbpf
> > > > > > 
> > > > > > 
> > > > > > And that will keep happening with CMAKE_USE_LIBBPF_PACKAGE=off which is the default, as mentioned, so it shouldn't be a problem? Unless I'm missing something.
> > > > > 
> > > > > 
> > > > > well, we need libbcc (with libbpf compiled in) and libbcc-no-libbpf at the same time
> > > > > compiled for bcc package
> > > > > and that was the case until this commit:
> > > > > [300296a](https://github.com/iovisor/bcc/commit/300296a598613912df5dc61f4b327b7102e52011) cmake: always link to packaged libbpf if CMAKE_USE_LIBBPF_PACKAGE is set (#3210)
> > > > 
> > > > 
> > > > Sorry, maybe I misunderstood - I thought the different cases were needed by different people in different situations? Are you saying that you need that _at the exact same time for the exact same build_ ? What's the use case, if you don't mind me asking?
> > > 
> > > 
> > > bcc package providing libraries is the use case
> > > it provides libbcc (with libbpf compiled in)
> > > and libbcc-no-libbpf that we use to compile bpftrace with
> > > when we compile bpftrace with libbcc, the dynamic linker gets confused,
> > > because of the extra compiled libbpf inside libbcc
> > > my first fix was to have libcc dynamically linked with libbpf (like you did),
> > > but I was told libbcc needs to stay as is because:
> > > 
> > > 1. libbpf package is not widespread
> > > 2. libbcc is using latest libbpf from git submodule and package will not keep up
> > 
> > 
> > Those only applies to local builds though, it doesn't apply to a distribution (where in fact it's the opposite - vendoring means out of date libbpf that doesn't track latest changes and always falls behind). The existing option allows users to do this, and it's the default so there should be no surprise nor regressions for direct users and developers, right?
> > > 1. peoples' apps linking libbcc will be broken
> > > 
> > > while 1) and 2) might not be a big issue anymore, I'm worried about 3)
> > 
> > 
> > How would that happen? The breakages, as you said, come from the linker getting confused because of multiple built-in and dynamic objects. What cases are there where the opposite is true?
> 
> missing libbpf dependency
> 
> but as you mentioned above that's probably not big deal for distributions,
> because they provide the needed libbpf, just local builds, maybe I could
> sell this to bcc maintainer, I'll ask

It's packaged, so the libbc package will depend on libbpf - it cannot be missing. It can only be missing if it wasn't used at build time because the option was disabled, but then it would be statically linked so no issue either.

As far as I can see, missing libbpf is only a problem if someone installs the libbpf-dev package _and_ does a manual, local build of bcc _and_ enables the option (disabled by default) _and_ separately, manually ships the result somewhere else. I mean, it's possible, but there's a lot of opt-in steps that need to be voluntarily taken to get there, and one can just not opt-in...
I do not see how this is a problem for _distro_ packages of bcc in any way though, since RPM takes care of tracking everything that needs to be tracked. I might be missing something, of course.

> I tested bpftrace with libbcc and it seems to work

Nice!

yonghong-song:
@olsajiri @bluca Looks the current mechanism works. I will hold this patch then.

olsajiri:
> @olsajiri @bluca Looks the current mechanism works. I will hold this patch then.

yep, I'm waiting for the fedora bcc maintainer reply, I'll report back ;-)

olsajiri:
> > @olsajiri @bluca Looks the current mechanism works. I will hold this patch then.
> 
> yep, I'm waiting for the fedora bcc maintainer reply, I'll report back ;-)

I talked to fedora's bcc maintainer and he's fine with that,
so we can scratch this

thanks and sorry for the noise ;-)

yonghong-song:
> > > @olsajiri @bluca Looks the current mechanism works. I will hold this patch then.
> > 
> > 
> > yep, I'm waiting for the fedora bcc maintainer reply, I'll report back ;-)
> 
> I talked to fedora's bcc maintainer and he's fine with that,
> so we can scratch this
> 
> thanks and sorry for the noise ;-)

Great. Thanks for good news :-)

bluca:
> > > @olsajiri @bluca Looks the current mechanism works. I will hold this patch then.
> > 
> > 
> > yep, I'm waiting for the fedora bcc maintainer reply, I'll report back ;-)
> 
> I talked to fedora's bcc maintainer and he's fine with that,
> so we can scratch this
> 
> thanks and sorry for the noise ;-)

Great news, thanks for double checking!

PR #3234, opened 2021-01-18
add bpf_map_lookup_and_delete_batch in bcc

egobillot:
After a discussion with @yonghong-song in PR https://github.com/iovisor/bcc/pull/3173 , I have added the support of batch lookup and delete.

In bcc, I have added items_lookup_and_delete_batch(). Feel free to propose a better name.

Batch works fine, I get all the map with a single BPF call.
But I still see race condition while it is not supposed to be. I have tested with kernel 5.8 and 5.11.0-051100rc2. This is annoying since this is the issue I wanted to address. Any help on this would be appreciated. 

Can you review the code please ?

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -359,12 +359,8 @@ int bpf_lookup_and_delete(int fd, void *key, void *value)
 int bpf_lookup_and_delete_batch(int fd, __u32 *in_batch, __u32 *out_batch, void *keys,
                                 void *values, __u32 *count)
 {
-  DECLARE_LIBBPF_OPTS(bpf_map_batch_opts,
-                      opts,
-                      .elem_flags = 0,
-                      .flags = 0,);
   return bpf_map_lookup_and_delete_batch(fd, in_batch, out_batch, keys, values,
-                                         count, &opts);
+                                         count, NULL);
 }
 
 int bpf_get_first_key(int fd, void *key, size_t key_size)
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -356,6 +356,17 @@ int bpf_lookup_and_delete(int fd, void *key, void *value)
   return bpf_map_lookup_and_delete_elem(fd, key, value);
 }
 
+int bpf_lookup_and_delete_batch(int fd, __u32 *in_batch, __u32 *out_batch, void *keys,
+                                void *values, __u32 *count)
+{
+  DECLARE_LIBBPF_OPTS(bpf_map_batch_opts,
+                      opts,
+                      .elem_flags = 0,
+                      .flags = 0,);
+  return bpf_map_lookup_and_delete_batch(fd, in_batch, out_batch, keys, values,
+                                         count, &opts);
+}
+
 int bpf_get_first_key(int fd, void *key, size_t key_size)
 {
   int i, res;
--- a/src/python/bcc/table.py
+++ b/src/python/bcc/table.py
@@ -397,6 +397,34 @@ class TableBase(MutableMapping):
         for k in self.keys():
             self.__delitem__(k)
 
+    def items_lookup_and_delete_batch(self, batch_size=0):
+        # By default batch size is the biggest possible (max_entries in
+        # the map) unless user wants to override it.
+        if batch_size <= 0:  # no value set by user
+            batch_size = self.max_entries
+
+        in_batch = ct.c_void_p(0)
+        out_batch = ct.c_void_p(0)
+        keys = (type(self.Key()) * batch_size)()
+        values = (type(self.Leaf()) * batch_size)()
+        count = ct.c_uint32(batch_size)
+        res = lib.bpf_lookup_and_delete_batch(self.map_fd,
+                                              ct.byref(in_batch),
+                                              ct.byref(out_batch),
+                                              ct.byref(keys),
+                                              ct.byref(values),
+                                              ct.byref(count)
+                                              )
+        errcode = ct.get_errno()
+        if (errcode == errno.EINVAL):
+            raise Exception("BPF_MAP_LOOKUP_AND_DELETE_BATCH is invalid.")
+
+        if (res != 0 and errcode != errno.ENOENT):
+            raise Exception("BPF_MAP_LOOKUP_AND_DELETE_BATCH has failed")
+
+        for i in range(0, count.value):
+            yield (keys[i], values[i])
+
     def zero(self):
         # Even though this is not very efficient, we grab the entire list of
         # keys before enumerating it. This helps avoid a potential race where
@@ -584,6 +612,8 @@ class TableBase(MutableMapping):
 class HashTable(TableBase):
     def __init__(self, *args, **kwargs):
         super(HashTable, self).__init__(*args, **kwargs)
+        self.max_entries = int(lib.bpf_table_max_entries_id(self.bpf.module,
+                                                            self.map_id))
 
     def __len__(self):
         i = 0
--- a/src/python/bcc/table.py
+++ b/src/python/bcc/table.py
@@ -397,22 +397,21 @@ class TableBase(MutableMapping):
         for k in self.keys():
             self.__delitem__(k)
 
-
     def items_lookup_and_delete_batch(self):
         # batch size is set to the maximum
         batch_size = self.max_entries
-        in_batch = ct.c_void_p(0)
-        out_batch = ct.c_void_p(0)
+        out_batch = ct.c_uint32(0)
         keys = (type(self.Key()) * batch_size)()
         values = (type(self.Leaf()) * batch_size)()
         count = ct.c_uint32(batch_size)
         res = lib.bpf_lookup_and_delete_batch(self.map_fd,
-                                              ct.byref(in_batch),
+                                              None,
                                               ct.byref(out_batch),
                                               ct.byref(keys),
                                               ct.byref(values),
                                               ct.byref(count)
                                               )
+
         errcode = ct.get_errno()
         if (errcode == errno.EINVAL):
             raise Exception("BPF_MAP_LOOKUP_AND_DELETE_BATCH is invalid.")
--- a/src/python/bcc/table.py
+++ b/src/python/bcc/table.py
@@ -397,12 +397,10 @@ class TableBase(MutableMapping):
         for k in self.keys():
             self.__delitem__(k)
 
-    def items_lookup_and_delete_batch(self, batch_size=0):
-        # By default batch size is the biggest possible (max_entries in
-        # the map) unless user wants to override it.
-        if batch_size <= 0:  # no value set by user
-            batch_size = self.max_entries
 
+    def items_lookup_and_delete_batch(self):
+        # batch size is set to the maximum
+        batch_size = self.max_entries
         in_batch = ct.c_void_p(0)
         out_batch = ct.c_void_p(0)
         keys = (type(self.Key()) * batch_size)()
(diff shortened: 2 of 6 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
@egobillot Thanks for the patch! For
> But I still see race condition while it is not supposed to be. I have tested with kernel 5.8 and 5.11.0-051100rc2. This is 
> annoying since this is the issue I wanted to address. Any help on this would be appreciated.
Could you provide more details about race conditions? Even better if you could provide a test case. During batch_lookup_and_delete, bpf programs and user spaces still can lookup over the hash table, but they cannot
do update or delete if that hash buck happens in the middle of lookup and delete.

egobillot:
Here is my stress test : https://gist.github.com/egobillot/9aa94f4993d0395d6d53a7c55411b4d9
build readzero which is the stress tool
`cc readzero.c -o readzero`

run `sudo python3 ./readtop.py` then run `./readzero` and check the bpf catch 10 000 000 +1 reads system call.

Do you also see the race condition ? My laptop generate ~ 1.6M reads per second. At this rate race condition is always visible.

yonghong-song:
@egobillot I ran your test case and actually it looks good to me. I made some change to your test program,
```
diff --git a/readtop.c b/readtop.c
index 6e0a307..13b2477 100644
--- a/readtop.c
+++ b/readtop.c
@@ -45,6 +45,7 @@ static int syscall_enter(void * ctx, char * fname){
 }
 #ifdef LATENCY
 static int syscall_return(void * ctx, char * fname){
+#if 0
     struct key_t key = {};
     struct value_t *pValue;
     // set asap the endtime to improve precision
@@ -63,6 +64,7 @@ static int syscall_return(void * ctx, char * fname){
     pValue->cumLat += endTime - pValue->startTime;
     // now update the map
     map.update(&key, pValue);
+#endif
 
     return 0;
 }
diff --git a/readtop.py b/readtop.py
index d4c5d19..c73cb7d 100755
--- a/readtop.py
+++ b/readtop.py
@@ -101,7 +101,7 @@ if len(argv) > 1:
 
 
 syscall_list = ['read']
-latency = False
+latency = True
 
 # load BPF program
 b = create_and_load_bpf(syscalls=syscall_list, lat=latency)
```
Basically, I enabled latency so I won't have any load error for kretprobe.
But I also commented out code for kretprobe so it won't update the map.
I got the following result:
```
bash-4.4$ sudo ./readtop.py
      PID-     COMM-   fname-  counter CUMULATED
  1711421- readzero-    read-  1141055 1141055
  1711421- readzero-    read-  2486536 3627591
  1711421- readzero-    read-  2462848 6090439
  1711421- readzero-    read-  2467749 8558188
  1711421- readzero-    read-  1441813 10000001
^CTraceback (most recent call last):
  File "./readtop.py", line 114, in <module>
    for k, v in sorted(b['map'].items_lookup_and_delete_batch(), key=lambda kv: (kv[0]).pid):
  File "/usr/lib/python2.7/site-packages/bcc/table.py", line 408, in items_lookup_and_delete_batch
    keys = (type(self.Key()) * batch_size)()
KeyboardInterrupt
```
You can see the accumulated result is exactly 10M + 1.

I then tried the following change,
```
diff --git a/readtop.c b/readtop.c
index 6e0a307..2f86af1 100644
--- a/readtop.c
+++ b/readtop.c
@@ -22,6 +22,7 @@ ACTIVATEALLSYSCALL
 ACTIVATELATENCY
 
 static int syscall_enter(void * ctx, char * fname){
+#if 0
     struct key_t key = {};
     struct value_t valZero = {0,0,0};
     struct value_t *pValue;
@@ -40,28 +41,29 @@ static int syscall_enter(void * ctx, char * fname){
 
     // update the map
     map.update(&key, pValue);
+#endif
 
     return 0;
 }
 #ifdef LATENCY
 static int syscall_return(void * ctx, char * fname){
     struct key_t key = {};
+    struct value_t valZero = {0,0,0};
     struct value_t *pValue;
-    // set asap the endtime to improve precision
-    u64 endTime = bpf_ktime_get_ns();
+    u64 stime = bpf_ktime_get_ns();
 
-    //Build the key ...
     strcpy(key.fname, fname);
     bpf_get_current_comm(key.comm, sizeof(key.comm));
     key.pid = bpf_get_current_pid_tgid();
 
-    // ... to lookup its value
-    pValue=map.lookup(&key);
-    if(!pValue || pValue->startTime == 0)
+    pValue = map.lookup_or_init(&key, &valZero);
+    if(!pValue)
         return 0;
 
-    pValue->cumLat += endTime - pValue->startTime;
-    // now update the map
+    pValue->counter++;
+    pValue->startTime = stime; // usefull to let userspace clear old map entries
+
+    // update the map
     map.update(&key, pValue);
 
     return 0;
diff --git a/readtop.py b/readtop.py
index d4c5d19..c73cb7d 100755
--- a/readtop.py
+++ b/readtop.py
@@ -101,7 +101,7 @@ if len(argv) > 1:
 
 
 syscall_list = ['read']
-latency = False
+latency = True
 
 # load BPF program
 b = create_and_load_bpf(syscalls=syscall_list, lat=latency)
```
The above only tried to accumulate counters for kretprobe and I got
```
^Cbash-4.4$ sudo ./readtop.py
      PID-     COMM-   fname-  counter CUMULATED
  1787330- readzero-    read-  1484019 1484019
  1787330- readzero-    read-  1500009 2984028
  1787330- readzero-    read-  2986507 5970535
  1787330- readzero-    read-  1539252 7509787
  1787330- readzero-    read-  3035736 10545523
  1787330- readzero-    read-  4509701 15055224
  1787330- readzero-    read-  1016048 16071272
```
I got 16M+ counters. If you enable both kretprobe and kprobe,
you will get similar 16M+ counters. I have to dig out why this happens.
It is possible if we enable kretprobe, then kprobe itself will be
called more times and that is why if enabling both, you will get
more counters.

I think bpf_map_lookup_and_delete_patch() works. It will be good
if you can add a test to use the new APIs. thanks!

egobillot:
Hello @yonghong-song. Thanks for your help. I have applied your first patch on my test program. It got 10M +1  as result but only the first time I ran it. The problem is that I got this good result only once. My next trials have all failed.
Can you re-run the first test several times? and if you want to make sure you stress the "lock" you can run ./readtop.py 0.01.

I don't think the issue is in bcc but in the kernel...

yonghong-song:
I did some further debugging and I think this is an application issue. The following change can make output consistent with 10M + 1 output.
```
--- a/readtop.c
+++ b/readtop.c
@@ -39,7 +39,7 @@ static int syscall_enter(void * ctx, char * fname){
     pValue->startTime = stime; // usefull to let userspace clear old map entries
 
     // update the map
-    map.update(&key, pValue);
+    // map.update(&key, pValue);
 
     return 0;
 }
@@ -62,7 +62,7 @@ static int syscall_return(void * ctx, char * fname){
 
     pValue->cumLat += endTime - pValue->startTime;
     // now update the map
-    map.update(&key, pValue);
+    // map.update(&key, pValue);
 
     return 0;
 }
```

The problem is you did `map.update` at the end of the program which actually has a big counter, which you will increment next time on top of it. Here `map.update` is not needed. Once you did `pValue=map.lookup(&key)` and got a valid
pointer pValue, directly change pValue pointee value is the same as update. This way, the updated value will be copied
to user space and then the kernel will delete it from the map. Next time it will start from counter 0.

src/python/bcc/table.py
```diff
@@ -397,6 +397,34 @@ def clear(self):
...
             self.__delitem__(k)
 
+    def items_lookup_and_delete_batch(self, batch_size=0):
+        # By default batch size is the biggest possible (max_entries in
+        # the map) unless user wants to override it.
+        if batch_size <= 0:  # no value set by user
+            batch_size = self.max_entries
+
+        in_batch = ct.c_void_p(0)
+        out_batch = ct.c_void_p(0)
```
yonghong-song, line 407: So, in_batch and out_batch are both pointers to `0` right? The `batch` is an opaque number used for continuous batch processing in the case of smaller buffer. For example:
  1. first run, in_batch = NULL, out_patch content not matter
  2. second run, in_batch = previous out_patch, out patch content not matter,
  3. ...
Since here, we allocate enough buffer and only want to do one batch processing,
the in_batch needs to be a NULL pointer.

egobillot: Yes this is inspired by what I have seen in /kernel-src/tools/testing/selftests/bpf/map_tests/htab_map_batch_ops.c

Your comment make me change my mind about the optional argument : batch_size. It does not do what we could intuitively imagine. It does a lookup_and_delete on only batch_size elements in the table then returns. I does not loop until the map is empty and returns batch_size elements on each steps.
I think this is confusing and not useful. What do you think about it ?

yonghong-song: > Yes this is inspired by what I have seen in /kernel-src/tools/testing/selftests/bpf/map_tests/htab_map_batch_ops.c
> 
> Your comment make me change my mind about the optional argument : batch_size. It does not do what we could intuitively imagine. It does a lookup_and_delete on only batch_size elements in the table then returns. I does not loop until the map is empty and returns batch_size elements on each steps.
> I think this is confusing and not useful. What do you think about it ?

Yes, let us just remove the optional argument and users in general do
not really care about batch size. If there is a need for map with
*huge* size, we can improve to using smaller batch size inside
the API.

yonghong-song:
[buildbot, test this please]

src/cc/libbpf.c
```diff
@@ -356,6 +356,17 @@ int bpf_lookup_and_delete(int fd, void *key, void *value)
...
 
+int bpf_lookup_and_delete_batch(int fd, __u32 *in_batch, __u32 *out_batch, void *keys,
+                                void *values, __u32 *count)
+{
+  DECLARE_LIBBPF_OPTS(bpf_map_batch_opts,
+                      opts,
+                      .elem_flags = 0,
+                      .flags = 0,);
+  return bpf_map_lookup_and_delete_batch(fd, in_batch, out_batch, keys, values,
+                                         count, &opts);
```
yonghong-song, line 367: You can skip DECLARE_LIBBPF_OPTS, just replace `&opts` with NULL.

src/python/bcc/table.py
```diff
@@ -397,6 +397,32 @@ def clear(self):
...
+    def items_lookup_and_delete_batch(self):
+        # batch size is set to the maximum
+        batch_size = self.max_entries
+        in_batch = ct.c_void_p(0)
+        out_batch = ct.c_void_p(0)
+        keys = (type(self.Key()) * batch_size)()
+        values = (type(self.Leaf()) * batch_size)()
+        count = ct.c_uint32(batch_size)
+        res = lib.bpf_lookup_and_delete_batch(self.map_fd,
+                                              ct.byref(in_batch),
```
yonghong-song, line 410: We should pass a NULL pointer for in_batch. I think `ct.byref(in_batch)` here is a pointer to value 0? I see some document saying `None` can be represented as NULL pointer. But I did not do experiments, so not sure.

egobillot:
@yonghong-song It should be ok now. I have also update the reference guide.

yonghong-song:
[buildbot, test this please]

yonghong-song:
LGTM. Thanks!

yonghong-song:
The test failed.
```
41: Test command: /home/fedora/jenkins/workspace/bcc-pr/label/fc25/build/tests/wrapper.sh "py_test_map_batch_ops" "sudo" "/home/fedora/jenkins/workspace/bcc-pr/label/fc25/tests/python/test_test_map_batch_ops.py"
41: Test timeout computed to be: 9.99988e+06
41: bash: /home/fedora/jenkins/workspace/bcc-pr/label/fc25/tests/python/test_test_map_batch_ops.py: No such file or directory
41: Failed
41/45 Test #41: py_test_map_batch_ops ............***Failed    0.02 sec
```
For test name `test_test_map_batch_ops.py`, maybe `test_map_batch_ops.py`?

yonghong-song:
[buildbot, ok to test]

PR #3236, opened 2021-01-19
Rename tools/swapin_example.py to tools/swapin_example.txt

rantala:
Use correct *.txt suffix for swapin example file.


yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks for the fix!

PR #3243, opened 2021-01-26
aarch64: turn off jump table optimization during jit compilation

xuchunmei000:
test_clang.py: test_jump_table will get failed with error messgae:
bpf: Failed to load program: Invalid argument
unknown opcode a0
processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0
peak_states 0 mark_read 0

HINT: The 'unknown opcode' can happen if you reference a global or
static variable, or data in read-only section. For example, 'char *p =
"hello"' will result in p referencing a read-only section, and 'char p[]
= "hello"' will have "hello" stored on the stack.

Signed-off-by: Chunmei Xu <xuchunmei@linux.alibaba.com>


yonghong-song:
[buildbot, test this please]

Issue #3224, opened 2021-01-13
Required dependencies for libbpf-tools?


mika:
Following up on Brendan Gregg's excellent http://www.brendangregg.com/blog/2020-11-04/bpf-co-re-btf-libbpf.html I looked into packaging the `libbpf-tools` binaries (`biolatency`, `biopattern`, `biosnoop`, `biostacks`, `bitesize`, `cpudist`, `drsnoop`, `execsnoop`, `filelife`, `hardirqs`, `llcstat`, `numamove`, `opensnoop`, `readahead`, `runqlat`, `runqlen`, `runqslower`, `softirqs`, `syscount`, `tcpconnect`, `tcpconnlat`, `vfsstat` + `xfsslower`), to be able to use them on systems without having to install lots of additional packages.

I did my packaging work on current Debian testing/bullseye, see https://gist.github.com/mika/e55ab72659bda90d8fecbbb42830d250

Most of the binaries work as intended, though `readahead`, `xfsslower` + `llcstat` are failing for me (both on the system where I built the binaries as well as on another Debian/bullseye system with just the compiled binaries and an according kernel with `CONFIG_DEBUG_INFO_BTF` enabled):

```
% sudo readahead
libbpf: failed to find kernel BTF type ID of '__do_page_cache_readahead': -3
libbpf: failed to load object 'readahead_bpf'
libbpf: failed to load BPF skeleton 'readahead_bpf': -3
failed to open and/or load BPF ojbect

% sudo xfsslower
libbpf: kprobe perf_event_open() failed: No such file or directory
libbpf: prog 'xfs_file_read_iter': failed to create kprobe 'xfs_file_read_iter' perf event: No such file or directory
libbpf: failed to auto-attach program 'xfs_file_read_iter': -2
failed to attach BPF programs

% sudo llcstat
failed to init perf sampling: No such file or directory
% sudo strace -f -s500 llcstat 2>&1 | tail
munmap(0x7f17e10f2000, 360448)          = 0
perf_event_open({type=PERF_TYPE_HARDWARE, size=0 /* PERF_ATTR_SIZE_??? */, config=PERF_COUNT_HW_CACHE_MISSES, ...}, -1, 0, -1, 0) = -1 ENOENT (No such file or directory)
write(2, "failed to init perf sampling: No such file or directory\n", 56failed to init perf sampling: No such file or directory) = 56
[...]
```

[...]

PR #3245, opened 2021-01-27
libbpf-tools: fix some tools error info and usage

ethercflow:
- Fix the error message in the open phase
- Fix spelling error: ojbect -> object
- Complete parameter information in usage
- Capitalize parameters that do not require parameter values ​​for easy memory

### DIFF
--- a/libbpf-tools/biolatency.c
+++ b/libbpf-tools/biolatency.c
@@ -40,12 +40,12 @@ const char *argp_program_bug_address = "<bpf@vger.kernel.org>";
 const char argp_program_doc[] =
 "Summarize block device I/O latency as a histogram.\n"
 "\n"
-"USAGE: biolatency [--help] [-T] [-m] [-Q] [-D] [-F] [-d] [interval] [count]\n"
+"USAGE: biolatency [--help] [-T] [-M] [-Q] [-D] [-F] [-d DISK] [interval] [count]\n"
 "\n"
 "EXAMPLES:\n"
 "    biolatency              # summarize block I/O latency as a histogram\n"
 "    biolatency 1 10         # print 1 second summaries, 10 times\n"
-"    biolatency -mT 1        # 1s summaries, milliseconds, and timestamps\n"
+"    biolatency -MT 1        # 1s summaries, milliseconds, and timestamps\n"
 "    biolatency -Q           # include OS queued time in I/O time\n"
 "    biolatency -D           # show each disk device separately\n"
 "    biolatency -F           # show I/O flags separately\n"
@@ -53,7 +53,7 @@ const char argp_program_doc[] =
 
 static const struct argp_option opts[] = {
 	{ "timestamp", 'T', NULL, 0, "Include timestamp on output" },
-	{ "milliseconds", 'm', NULL, 0, "Millisecond histogram" },
+	{ "milliseconds", 'M', NULL, 0, "Millisecond histogram" },
 	{ "queued", 'Q', NULL, 0, "Include OS queued time in I/O time" },
 	{ "disk", 'D', NULL, 0, "Print a histogram per disk device" },
 	{ "flag", 'F', NULL, 0, "Print a histogram per set of I/O flags" },
@@ -70,7 +70,7 @@ static error_t parse_arg(int key, char *arg, struct argp_state *state)
 	case 'v':
 		env.verbose = true;
 		break;
-	case 'm':
+	case 'M':
 		env.milliseconds = true;
 		break;
 	case 'Q':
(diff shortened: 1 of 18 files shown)


libbpf-tools/biolatency.c
```diff
@@ -40,20 +40,20 @@ const char *argp_program_bug_address = "<bpf@vger.kernel.org>";
...
 "Summarize block device I/O latency as a histogram.\n"
 "\n"
-"USAGE: biolatency [--help] [-T] [-m] [-Q] [-D] [-F] [-d] [interval] [count]\n"
+"USAGE: biolatency [--help] [-T] [-M] [-Q] [-D] [-F] [-d DISK] [interval] [count]\n"
 "\n"
 "EXAMPLES:\n"
 "    biolatency              # summarize block I/O latency as a histogram\n"
 "    biolatency 1 10         # print 1 second summaries, 10 times\n"
-"    biolatency -mT 1        # 1s summaries, milliseconds, and timestamps\n"
+"    biolatency -MT 1        # 1s summaries, milliseconds, and timestamps\n"
```
anakryiko, line 48: biolatency.py is using `-m`, why sudden and diverging change?

libbpf-tools/cpudist.c
```diff
@@ -37,20 +37,20 @@ const char *argp_program_bug_address = "<bpf@vger.kernel.org>";
...
 "\n"
-"USAGE: cpudist [--help] [-O] [-T] [-m] [-P] [-L] [-p PID] [interval] [count]\n"
+"USAGE: cpudist [--help] [-O] [-T] [-M] [-P] [-L] [-p PID] [interval] [count]\n"
 "\n"
 "EXAMPLES:\n"
 "    cpudist              # summarize on-CPU time as a histogram"
 "    cpudist -O           # summarize off-CPU time as a histogram"
 "    cpudist 1 10         # print 1 second summaries, 10 times"
-"    cpudist -mT 1        # 1s summaries, milliseconds, and timestamps"
+"    cpudist -MT 1        # 1s summaries, milliseconds, and timestamps"
```
anakryiko, line 46: similarly. py version is sticking with -m. What's your rationale for making these incompatible changes?

libbpf-tools/drsnoop.c
```diff
@@ -31,18 +31,18 @@ const char *argp_program_bug_address = "<bpf@vger.kernel.org>";
...
 "    drsnoop -p 123  # trace pid 123\n"
 "    drsnoop -t 123  # trace tid 123 (use for threads only)\n"
 "    drsnoop -d 10   # trace for 10 seconds only\n"
-"    drsnoop -e      # trace all direct reclaim events with extended faileds\n";
+"    drsnoop -E      # trace all direct reclaim events with extended faileds\n";
 
 static const struct argp_option opts[] = {
 	{ "duration", 'd', "DURATION", 0, "Total duration of trace in seconds" },
-	{ "extended", 'e', NULL, 0, "Extended fields output" },
+	{ "extended", 'E', NULL, 0, "Extended fields output" },
```
anakryiko, line 45: this seems to be libbpf-tools exclusive argument, but again, why changing it in the first place?

anakryiko (changes requested):
There are a bunch more cases of arbitrary changing the case of existing command line arguments with no justification. Please provide rationale for the changes in your commit messages, not just the "what".

Also, while I have your attention. Can you please check https://github.com/iovisor/bcc/issues/3224#issuecomment-759909082? Would you be able to add that check? Thanks!

libbpf-tools/biolatency.c
ethercflow: > There are a bunch more cases of arbitrary changing the case of existing command line arguments with no justification. Please provide rationale for the changes in your commit messages, not just the "what".

Oh, my motivation is I think maybe it's easier to remember. Lowercase parameters need parameter values ​​(except -v), and uppercase ones belong to the switch class, so you don't need to add parameter values.

> Also, while I have your attention. Can you please check [#3224 (comment)](https://github.com/iovisor/bcc/issues/3224#issuecomment-759909082)? Would you be able to add that check? Thanks!

Oh sorry, I haven't checked the email from iovisor/bcc recently, I missed it. I will look at it later （Maybe weekend）.

anakryiko: I haven't heard about such naming convention, tbh. I think the harm of changing existing arguments and, especially, diverging from .py tool, is much bigger. So let's not do such renamings unnecessarily.

And thanks for checking the other PR.

ethercflow: Got it,  I will be consistent with .py tool. Thanks!

PR #3246, opened 2021-01-27
cmake: look for either static or dynamic libraries

teknoraver:
On some distro, static libraries are shipped in a separate package.
Look for either a static or dynamic libbpf, and only fail if neither is found.


yonghong-song:
[buildbot, test this please]

yonghong-song:
@bluca could you help take a look whether this change works fine in your environment?

bluca (approved):
Yep it works just fine, thanks for the ping!

PR #3258, opened 2021-02-03
Increase storage size of usdt argument constants to 64 bits

rsanger:
[Systemtap](https://sourceware.org/systemtap/wiki/UserSpaceProbeImplementation) defines the size of an argument as either 1, 2, 4, or 8 bytes (signed or unsigned). Thus the constant variable should be stored in a `long long` instead of an `int` to ensure 8 bytes size.

I've opted for `long long` here rather than `int64_t` to match with `strtoull`'s type.

I've kept this as a signed integer for backwards compatibility, and I've used `strtoull` for the conversion as it correctly handles both the maximum unsigned value and negative values. The sign of the _size_ argument indicates whether the _constant_ value stored is signed or unsigned.

Note: 16-byte constants are possible but AFIK only from non-standard types like `__int128`.
This patch doesn't support that case.

### DIFF
--- a/src/cc/bcc_usdt.h
+++ b/src/cc/bcc_usdt.h
@@ -54,7 +54,7 @@ struct bcc_usdt_location {
 struct bcc_usdt_argument {
     int size;
     int valid;
-    int constant;
+    long long constant;
     int deref_offset;
     const char *deref_ident;
     const char *base_register_name;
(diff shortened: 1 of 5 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
@davemarchevsky could you help take a look as well?

PR #3259, opened 2021-02-03
Wakeuptime: Use tracepoints if available

ananthmg:
wakeuptime.py uses kprobes by default and traces schedule(). Switch to
using tracepoints (sched_switch and wakeup) instead, if available.


ananthmg:
Working on a potentially better patch than this one... will open a new request... please ignore this one.

PR #3260, opened 2021-02-03
tools/wakeuptime: Switch to using tracepoints if available

ananthmg:
wakeuptime currently uses kprobes to trace schedule(). Switch to using
tracepoints if available.

Also, in some builds, try_to_wake_up() may get optimized away causing
the script to fail altogether.

With this switch, we however see stack trace entries related to BPF too
in the end for each trace. While correct, it takes up screen real estate
for each backtrace.

    target:          InputThread
    ffffffffaf000107 secondary_startup_64_no_verify
    ffffffffb12495f1 start_kernel
    ffffffffaf10e609 cpu_startup_entry
    ffffffffaf10e40b do_idle
    ffffffffaf946849 cpuidle_enter
    ffffffffaf946577 cpuidle_enter_state
    ffffffffafc00d02 asm_sysvec_apic_timer_interrupt
    ffffffffafbbbc86 sysvec_apic_timer_interrupt
    ffffffffaf0dcc32 irq_exit_rcu
    ffffffffaf02bc47 do_softirq_own_stack
    ffffffffafc01112 asm_call_irq_on_stack
    ffffffffafe000ca __softirqentry_text_start
    ffffffffaf163a66 run_timer_softirq
    ffffffffaf1639df __run_timers.part.0
    ffffffffaf162f09 call_timer_fn
    ffffffffaf8c46c3 input_repeat_key
    ffffffffaf8c318e input_pass_values.part.0
    ffffffffaf8c2085 input_to_handler
    ffffffffaf8c9e81 evdev_events
    ffffffffaf124a1c __wake_up_common_lock
    ffffffffaf1248b0 __wake_up_common
    ffffffffaf375e49 ep_poll_callback
    ffffffffaf124a1c __wake_up_common_lock
    ffffffffaf1248b0 __wake_up_common
    ffffffffaf375e0d ep_poll_callback
    ffffffffaf124a1c __wake_up_common_lock
    ffffffffaf1248b0 __wake_up_common
    ffffffffaf124681 autoremove_wake_function
    ffffffffaf108e57 try_to_wake_up
    ffffffffaf107dcc ttwu_do_wakeup
    ffffffffaf1f48a2 bpf_trace_run1
    ffffffffc16879d3 ftrace_trampoline
    ffffffffaf1f51c9 bpf_get_stackid_raw_tp
    waker:           swapper/0
        169325

### DIFF
--- a/tools/wakeuptime.py
+++ b/tools/wakeuptime.py
@@ -102,7 +102,7 @@ BPF_HASH(counts, struct key_t);
 BPF_HASH(start, u32);
 BPF_STACK_TRACE(stack_traces, STACK_STORAGE_SIZE);
 
-int offcpu(struct pt_regs *ctx) {
+static int offcpu_sched_switch() {
     u32 pid = bpf_get_current_pid_tgid();
     struct task_struct *p = (struct task_struct *) bpf_get_current_task();
     u64 ts;
@@ -115,7 +115,7 @@ int offcpu(struct pt_regs *ctx) {
     return 0;
 }
 
-int waker(struct pt_regs *ctx, struct task_struct *p) {
+static int wakeup(ARG0, struct task_struct *p) {
     u32 pid = p->pid;
     u64 delta, *tsp, ts;
 
@@ -143,6 +143,38 @@ int waker(struct pt_regs *ctx, struct task_struct *p) {
     return 0;
 }
 """
+
+bpf_text_kprobe = """
+int offcpu(struct pt_regs *ctx) {
+    return offcpu_sched_switch();
+}
+
+int waker(struct pt_regs *ctx, struct task_struct *p) {
+    return wakeup(ctx, p);
+}
+"""
+
+bpf_text_raw_tp = """
+RAW_TRACEPOINT_PROBE(sched_switch)
+{
+    // TP_PROTO(bool preempt, struct task_struct *prev, struct task_struct *next)
+    return offcpu_sched_switch();
+}
+
+RAW_TRACEPOINT_PROBE(sched_wakeup)
+{
+    // TP_PROTO(struct task_struct *p)
+    struct task_struct *p = (struct task_struct *) bpf_get_current_task();
+    return wakeup(ctx, p);
+}
+"""
+
+is_supported_raw_tp = BPF.support_raw_tracepoint()
+if is_supported_raw_tp:
+    bpf_text += bpf_text_raw_tp
+else:
+    bpf_text += bpf_text_kprobe
+
 if args.pid:
     filter = 'pid != %s' % args.pid
 elif args.useronly:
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

netedwardwu:
Hi @ananthmg ,

May I ask
1. Why should we change to use tracepoint?
2. Did you compare the result of kprobe from tracepoint? Is the same result in the same scenario?

Thanks!

netedwardwu:
> Hi @ananthmg ,
> 
> May I ask
> 
> 1. Why should we change to use tracepoint?
> 2. Did you compare the result of kprobe from tracepoint? Is the same result in the same scenario?
> 
> Thanks!

Okay... the symbol be optimized or stripped in some builds...

Should we concern the below case that is the difference between kprobe and tracepoint?
https://elixir.bootlin.com/linux/latest/source/kernel/sched/core.c#L2854
```
	/*
	 * If we are going to wake up a thread waiting for CONDITION we
	 * need to ensure that CONDITION=1 done by the caller can not be
	 * reordered with p->state check below. This pairs with smp_store_mb()
	 * in set_current_state() that the waiting thread does.
	 */
	raw_spin_lock_irqsave(&p->pi_lock, flags);
	smp_mb__after_spinlock();
	if (!(p->state & state))
		goto unlock;

	trace_sched_waking(p);
```
Thanks

ananthmg:
> 1. Why should we change to use tracepoint?

1. Tracepoints are generally orders of magnitude faster than kprobes since they execute in the task context. As a general rule, if data we need can be obtained using tracepoints, that should be preferred over using kprobes.
2. schedule() is a very sensitive function to probe on. See the section on detaching in the wakeuptime_example.txt. With tracepoints, that overhead will reduce considerably.
3. Tracepoints are a kind of ABI (debatable). Function names and compiler optimizations are fluid.

> 2. Did you compare the result of kprobe from tracepoint? Is the same result in the same scenario?

The output is identical, except for the few bpf entries that are new in the tracepoint case.

ananthmg:
> https://elixir.bootlin.com/linux/latest/source/kernel/sched/core.c#L2854
> 
> ```
> 	/*
> 	 * If we are going to wake up a thread waiting for CONDITION we
> 	 * need to ensure that CONDITION=1 done by the caller can not be
> 	 * reordered with p->state check below. This pairs with smp_store_mb()
> 	 * in set_current_state() that the waiting thread does.
> 	 */
> 	raw_spin_lock_irqsave(&p->pi_lock, flags);
> 	smp_mb__after_spinlock();
> 	if (!(p->state & state))
> 		goto unlock;
> 
> 	trace_sched_waking(p);
> ```

This is the case where the task is in the process of 'waking up', but not 'woken' yet. I would wager the kprobe mechanism may contain some false positive entries of tasks that are in this state. Migrating to tracepoints solves this problem, since trace_sched_wakeup() gets called from try_to_wake_up() and ttwu_do_wakeup() after the task->state is set to RUNNING.

Incidentally, this is another reason why the tracepoint method is more accurate for this case.

netedwardwu:
Very clear, thank you

yonghong-song:
Agree that tracepoint is better in this case. Let us use it if available. Thanks!

PR #3266, opened 2021-02-10
Mention kernel version which fixed the bug

fungusakafungus:
I'm not familiar with the kernel, I just looked these commits up and was confused by the mention that the bug is still not fixed.

Apparently that kernel bug is fixed in 4.8-rc3 and what broke there is fixed in 4.8.10
according to https://cdn.kernel.org/pub/linux/kernel/v4.x/ChangeLog-4.8.10

mentioned as:

bpf: fix htab map destruction when extra reserve is in use
Fixes: a6ed3ea65d98 ("bpf: restore behavior of bpf_map_update_elem")


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

PR #3265, opened 2021-02-09
Add multi-architecture vmlinux.h support

anakryiko:
Add support for per-architecture vmlinux.h headers. Add arm64 and powerpc vmlinux.h, generated by @vt-alt in https://github.com/iovisor/bcc/issues/3252.

PR #3271, opened 2021-02-15
bcc/libbpf-tools: tweak compilation to the architecture

rsudarikov:
Note: would like to add bpftool binary for arm64


libbpf-tools/Makefile
```diff
@@ -2,14 +2,14 @@
 OUTPUT := .output
 CLANG ?= clang
 LLVM_STRIP ?= llvm-strip
-BPFTOOL ?= bin/bpftool
+ARCH := $(shell uname -m | sed 's/x86_64/x86/' | sed 's/aarch64/arm64/' | sed 's/ppc64le/powerpc/')
+BPFTOOL ?= bin/$(ARCH)/bpftool
```
anakryiko, line 6: using ARCH here might not be a good idea (consider x86 and x86_64, 64-bit binary won't run on 32-bit OS). So let's just use the result of `uname -m` without the "unification" substitutions?

anakryiko (changes requested):
Thanks for doing this change. We were just discussing it recently (https://github.com/iovisor/bcc/pull/3265#issuecomment-776893572). I'd just rename the commit to talk about re-organizing bpftool location and making it architecture-specific, not really adding arm64 binary (you are just moving it around).

PR #3277, opened 2021-02-21
Fix bytes - strings confusion in trace tool

jongiddy:
Using `trace` with the `-n` or `-f` option gives me a `TypeError: string argument without an encoding`. This is due to bare use of the `bytes` constructor to convert a string. The [docs](https://docs.python.org/3.6/library/functions.html#func-bytearray) for `bytes` refer to `bytearray` which states "If it is a string, you must also give the encoding".

For `-n`, the comparison is the only place that `self.name` is used, so I moved the the conversion to the `__init__` method. For `-f`, the compared `msg` is already a string (because `self.python_format` is a string), so the conversion is unnecessary.

I've also made some minor cleanups such as removing redundant backslashes and semicolons.

### DIFF
--- a/tools/trace.py
+++ b/tools/trace.py
@@ -13,7 +13,7 @@
 from __future__ import print_function
 from bcc import BPF, USDT, StrcmpRewrite
 from functools import partial
-from time import sleep, strftime
+from time import strftime
 import time
 import argparse
 import re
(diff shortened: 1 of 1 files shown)


jongiddy:
Note that this occurs on Ubuntu 20.04, where `trace` from the `bpfcc-tools` package is modified to use `/usr/bin/python3`. It could be considered a distro problem, but Python 3 is the default on several distros now, and is likely to become default on more.

yonghong-song:
[buildbot, test this please]

yonghong-song:
There is a test failure:
```
  File "/home/fedora/jenkins/workspace/bcc-pr/label/fc28/tests/python/test_tools_smoke.py", line 357, in test_trace
30:     self.run_with_int("trace.py do_sys_open")
```
Could you try `trace.py do_sys_open` in both python3 and python2 environment?

yonghong-song:
[buildbot, ok to test]

jongiddy:
Updated to use `str.encode` rather than `bytes` constructor. This works on Python 2 and 3.

yonghong-song:
LGTM. Thanks!

PR #3284, opened 2021-02-25
Converted tcptop from BCC to libbpf + CORE

Am4teur:
Converted the tool tcptop.py (that can be found on the bcc/tools folder) to libbpf + CORE.

### DIFF
--- a/libbpf-tools/tcptop.bpf.c
+++ b/libbpf-tools/tcptop.bpf.c
@@ -6,19 +6,16 @@
 #include <bpf/bpf_core_read.h>
 #include "tcptop.h"
 
-
 char LICENSE[] SEC("license") = "Dual BSD/GPL";
 
 /* Define here, because there are conflicts with include files */
-#define AF_INET		2 || 5
+#define AF_INET		2
 #define AF_INET6	10
-
 
 struct {
 	__uint(type, BPF_MAP_TYPE_RINGBUF);
 	__uint(max_entries, 256 * 1024);
 } rb SEC(".maps");
-
 
 struct {
     __uint(type, BPF_MAP_TYPE_HASH);
@@ -48,9 +45,7 @@ struct {
     __type(value, int);
 } ipv6_recv_bytes SEC(".maps");
 
-
 const volatile pid_t filter_pid = 0;
-
 
 static __always_inline void
 trace_v4(struct pt_regs *ctx, struct ipv4_key_t k4, bool is_send, u16 family)
@@ -124,18 +119,15 @@ trace_v6(struct pt_regs *ctx, struct ipv6_key_t k6, bool is_send, u16 family)
     bpf_ringbuf_submit(e, 0);
 }
 
-static __always_inline
-int tcp_sendmsg_handler(struct pt_regs *ctx, struct sock *sk,
-    struct msghdr *msg, size_t size)
+SEC("kprobe/tcp_sendmsg")
+int BPF_KPROBE(tcp_sendmsg, struct sock *sk, struct msghdr *msg, size_t size)
 {
     u32 pid = bpf_get_current_pid_tgid() >> 32;
-
 
 	if (filter_pid && pid != filter_pid) {
 		return 0;
 	}
 
-    //u16 dport = 0, family;
     u16 family;
 	BPF_CORE_READ_INTO(&family, sk, __sk_common.skc_family);
 
@@ -157,14 +149,12 @@ int tcp_sendmsg_handler(struct pt_regs *ctx, struct sock *sk,
         bpf_map_update_elem(&ipv6_send_bytes, &ipv6_key, &size, 0 /* flags: BPF_ANY */);
         trace_v6(ctx, ipv6_key, true, family);
     }
-    /*else { //drop | this is commented for testing purposes
-		return 0;
-	}*/
+
 	return 0;
 }
 
-static __always_inline
-int tcp_cleanup_rbuf_handler(struct pt_regs *ctx, struct sock *sk, int copied)
+SEC("kprobe/tcp_cleanup_rbuf")
+int BPF_KPROBE(tcp_cleanup_rbuf, struct sock *sk, int copied)
 {
     u32 pid = bpf_get_current_pid_tgid() >> 32;
 
@@ -196,20 +186,6 @@ int tcp_cleanup_rbuf_handler(struct pt_regs *ctx, struct sock *sk, int copied)
         bpf_map_update_elem(&ipv6_recv_bytes, &ipv6_key, &copied, 0 /* flags: BPF_ANY */);
         trace_v6(ctx, ipv6_key, false, family);
     }
-    /*else { //drop
-		return 0;
-	}*/	
+
 	return 0;
-}
-
-SEC("kprobe/tcp_sendmsg")
-int BPF_KPROBE(tcp_sendmsg, struct sock *sk, struct msghdr *msg, size_t size)
-{
-	return tcp_sendmsg_handler(ctx, sk, msg, size);
-}
-
-SEC("kprobe/tcp_cleanup_rbuf")
-int BPF_KPROBE(tcp_cleanup_rbuf, struct sock *sk, int copied)
-{
-	return tcp_cleanup_rbuf_handler(ctx, sk, copied);
 }
--- a/libbpf-tools/tcptop.bpf.c
+++ b/libbpf-tools/tcptop.bpf.c
@@ -1,0 +1,215 @@
+// SPDX-License-Identifier: GPL-2.0 OR BSD-3-Clause
+/* Copyright (c) 2020 Facebook */
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include <bpf/bpf_core_read.h>
+#include "tcptop.h"
+
+
+char LICENSE[] SEC("license") = "Dual BSD/GPL";
+
+/* Define here, because there are conflicts with include files */
+#define AF_INET		2 || 5
+#define AF_INET6	10
+
+
+struct {
+	__uint(type, BPF_MAP_TYPE_RINGBUF);
+	__uint(max_entries, 256 * 1024);
+} rb SEC(".maps");
+
+
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __uint(max_entries, MAX_ENTRIES);
+    __type(key, struct ipv4_key_t);
+    __type(value, size_t);
+} ipv4_send_bytes SEC(".maps");
+
+struct {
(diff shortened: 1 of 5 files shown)


libbpf-tools/tcptop.bpf.c
```diff
@@ -0,0 +1,215 @@
...
+#include <bpf/bpf_tracing.h>
+#include <bpf/bpf_core_read.h>
+#include "tcptop.h"
+
+
+char LICENSE[] SEC("license") = "Dual BSD/GPL";
+
+/* Define here, because there are conflicts with include files */
+#define AF_INET		2 || 5
+#define AF_INET6	10
```
anakryiko, line 14: weird formatting?.. and what's (2 || 5) ?

libbpf-tools/tcptop.bpf.c
```diff
@@ -0,0 +1,215 @@
...
+
+
```
anakryiko, line 16: some more double empty lines, please leave just one

libbpf-tools/tcptop.bpf.c
```diff
@@ -0,0 +1,215 @@
...
+        BPF_CORE_READ_INTO(&ipv6_key.saddr, sk, __sk_common.skc_v6_rcv_saddr.in6_u.u6_addr32);
+        BPF_CORE_READ_INTO(&ipv6_key.daddr, sk, __sk_common.skc_v6_daddr.in6_u.u6_addr32);
+        BPF_CORE_READ_INTO(&ipv6_key.lport, sk, __sk_common.skc_num);
+        BPF_CORE_READ_INTO(&ipv6_key.dport, sk, __sk_common.skc_dport);
+        bpf_map_update_elem(&ipv6_recv_bytes, &ipv6_key, &copied, 0 /* flags: BPF_ANY */);
+        trace_v6(ctx, ipv6_key, false, family);
+    }
+    /*else { //drop
+		return 0;
+	}*/
```
anakryiko, line 201: debug leftovers

anakryiko (changes requested):
This feels a bit too raw. Formatting is off everywhere, please fix that.

Also, seems like tcptop.py doesn't use ringbuf, while this version does. What's the rationale? Please describe all this.

libbpf-tools/tcptop.bpf.c
```diff
@@ -0,0 +1,215 @@
+// SPDX-License-Identifier: GPL-2.0 OR BSD-3-Clause
+/* Copyright (c) 2020 Facebook */
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include <bpf/bpf_core_read.h>
+#include "tcptop.h"
+
+
```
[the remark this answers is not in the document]
Am4teur: Deleted the unnecessary line as requested. There was no specific meaning with the extra empty line.

libbpf-tools/tcptop.bpf.c
Am4teur: Corrected the formatting and the debug value. I used tcpconnect.bpf.c (and tcpconnect.c) to make tcptop.bpf.c as standard as possible. These defines are like tcpconnect (apart from the 2 || 5, which is now corrected).

libbpf-tools/tcptop.bpf.c
Am4teur: Deleted the unnecessary line as requested. There was no specific meaning with the extra empty line.

libbpf-tools/tcptop.bpf.c
```diff
@@ -0,0 +1,215 @@
...
+    /*else { //drop
+		return 0;
+	}*/	
+	return 0;
+}
+
+SEC("kprobe/tcp_sendmsg")
+int BPF_KPROBE(tcp_sendmsg, struct sock *sk, struct msghdr *msg, size_t size)
+{
+	return tcp_sendmsg_handler(ctx, sk, msg, size);
```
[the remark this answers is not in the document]
Am4teur: Again, following tcpconnect.bpf.c, this function indirection was applied. Nevertheless, followed your suggestion and corrected the unnecessary function indirection.

libbpf-tools/tcptop.bpf.c
Am4teur: This was left behind on purpose, just like tcptop.py left it behind. Nevertheless, followed your suggestion and deleted the comment.

Am4teur:
"Also, seems like tcptop.py doesn't use ringbuf, while this version does. What's the rationale? Please describe all this."

Yes, in fact, tcptop.py uses perfbuf, but after reading your article ( https://nakryiko.com/posts/bpf-ringbuf/ ) and understanding that ringbuf supports a familiar functionality from perfbuf and also has great advantages, such as, not copying the same data twice, I decided that this conversion (bcc -> libbpf), was the ideal place to also change from perfbuf to ringbuf. This change can create incompatibility problems but since this is a "new" tool, this won't affect much.

libbpf-tools/tcptop.bpf.c
anakryiko: tcptop.py has "// else drop" comment, it doesn't have commented out code. so feel free to leave a single comment, but not a commented out block of code. Please also prefer C89 style comments (/* */), that's what BPF code in libbpf-tools follow, as it's the kernel style.

libbpf-tools/tcptop.bpf.c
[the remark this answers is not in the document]
anakryiko: tcpconnect.bpf.c reuses single function across 4 BPF programs. Here, each BPF program has its own unique logic. So two cases are completely different. Please, stick to tcptop.py's logical structure, not some other unrelated tool.

anakryiko:
> "Also, seems like tcptop.py doesn't use ringbuf, while this version does. What's the rationale? Please describe all this."
> 
> Yes, in fact, tcptop.py uses perfbuf, but after reading your article ( https://nakryiko.com/posts/bpf-ringbuf/ ) and understanding that ringbuf supports a familiar functionality from perfbuf and also has great advantages, such as, not copying the same data twice, I decided that this conversion (bcc -> libbpf), was the ideal place to also change from perfbuf to ringbuf. This change can create incompatibility problems but since this is a "new" tool, this won't affect much.

Are we talking about the same tools/tcptop.py? Because I don't see any perfbuf usage there. It seems like it just sleep, then dumps BPF map values. I've double-checked again, I don't spot any perfbuf usage. I do agree in general about ringbuf over perfbuf, of course. Just in this case it seems like a completely different business logic between BCC and libbpf versions.

Am4teur:
> Are we talking about the same tools/tcptop.py? Yes, there is no use of perfbuf or ringbuf in the original tcptop.py. The addition of the ringbuf was made by me to adapt the tool, since it is easier to print the event from the user-space this way, in my opinion.

I do not think that it is a completely different business logic between the BCC tcptop.py tool and this one. However, I do understand that it is not 100% mapped, i.e., the important logic is there but a few minor aspects are not mapped from the bcc to the libbpf (such as, arguments like -cgroup that you can find on the tcptop.py).
Since I am using this conversion, maybe it would be useful to someone, but tcptop.bpf.c and tcptop.c are not as similar as possible to tcptop.py. They print the main logic (destination addr, source addr, etc) but minor aspects are missing.

Am4teur:
> Are we talking about the same tools/tcptop.py?

Yes, there is no use of perfbuf or ringbuf in the original tcptop.py. The addition of the ringbuf was made by me to adapt the tool, since it is easier to print the event from the user-space this way, in my opinion.

I do not think that it is a completely different business logic between the BCC tcptop.py tool and this one. However, I do understand that it is not 100% mapped, i.e., the important logic is there but a few minor aspects are not mapped from the bcc to the libbpf (such as, arguments like -cgroup that you can find on the tcptop.py).
Since I am using this conversion, maybe it would be useful to someone, but tcptop.bpf.c and tcptop.c are not as similar as possible to tcptop.py. They print the main logic (destination addr, source addr, etc) but minor aspects are missing.

libbpf-tools/tcptop.bpf.c
Am4teur: Changed to "// else drop" as suggested. I understand your point, having some code commented is different than what tcptop.py had. I appreciate your suggestions because I improve as a contributor and as a software engineer.

anakryiko: so few nits about that comment:
  1) // is not C89 style comment (which libbpf-tools are trying to follow, in line with what kernel style dictates)
  2) the comment itself is misleading. There is no packet drop, we just ignore other protocols, so something like "/* ignore other protocols */" is way more clear, IMHO.

But I'd just drop the comment altogether, tbh. It's not very useful regardless.

anakryiko:
> > Are we talking about the same tools/tcptop.py?
> 
> Yes, there is no use of perfbuf or ringbuf in the original tcptop.py. The addition of the ringbuf was made by me to adapt the tool, since it is easier to print the event from the user-space this way, in my opinion.

Adapt it to what? You've built a tool that has a different logic than original tcptop.py. Calling it tcptop and referencing tcptop.py is just misleading. tcptop.py aggregate stats for a period of time and prints out summary for that period. While your variant prints out non-aggregated stats. The performance will suffer if there is a lot of network activity. You'll most probably also will drop data and will print wrong stats. You can have millions of packets per second, emitting one BPF ringbuf sample for each such event is very-very expensive approach.

I get that doing the output from Python is simple due to dict()'s, but you can achieve a similar logic in C with just sorting receive and send stats by ip route and joining them on the fly. It will be O(NlogN), but still plenty fast in practice.

> 
> I do not think that it is a completely different business logic between the BCC tcptop.py tool and this one. However, I do understand that it is not 100% mapped, i.e., the important logic is there but a few minor aspects are not mapped from the bcc to the libbpf (such as, arguments like --cgroupmap that you can find on the tcptop.py).
> Since I am using this conversion, maybe it would be useful to someone, but tcptop.bpf.c and tcptop.c are not as similar as possible to tcptop.py. They print the main logic (destination addr, source addr, etc) but minor aspects are missing.

If your goal is not to have BPF CO-RE variant of tcptop.py, then please don't call your tool tcptop, as mentioned above. cc @brendangregg for an independent opinion

One thing that would be interesting and an overall improvement would be to use fentry BPF programs instead of kprobes. That would give a better performance at the expense of raising the minimum kernel version requirement. So something to consider, but is not a requirement.

anakryiko:
Oh, and you completely disregarded the formatting and styling issues:

- use tabs consistently throughout;
- follow C89 requirements to have variable declaration block at the top of lexical block, separated by single empty line from statments.

That would be a good start.

Am4teur:
Given your feedback, the best thing to do is close this PR with no merging. In fact, it was wrong calling it tcptop since the logic is different.

Also, to my surprise, in my editor the tabbing is correct, but in the committed files, the tabbing is inconsistent. Now I realize your point.

PR #3286, opened 2021-02-25
biotop and biosnoop: save __data_len at blk_start_request

### LINKED ISSUE #3099: biosnoop/biotop always return BYTES 0
Hi, I'm trying on running biosnoop.py on my desktop machine, but I'm always getting 0 for BYTES as shown below. (same for biotop.py)

image

This is the result while my desktop is on idle state, and also getting 0 for BYTES when I'm writing actual data to a file.

Any idea why I'm not getting the right value for the BYTES?

andreasgerstmayr:
Commit 95c9229ea9f029a1b9e8dcbe86fc67f037c0dfa2 replaced the
blk_account_io_completion kprobe with blk_account_io_done. Unfortunately
the req->__data_len field is 0 in blk_account_io_done, therefore we need
to save the __data_len field in blk_start_request

Resolves #3099

### DIFF
--- a/tools/biosnoop.py
+++ b/tools/biosnoop.py
@@ -39,6 +39,12 @@ bpf_text="""
 #include <uapi/linux/ptrace.h>
 #include <linux/blkdev.h>
 
+// for saving the timestamp and __data_len of each request
+struct start_req_t {
+    u64 ts;
+    u64 data_len;
+};
+
 struct val_t {
     u64 ts;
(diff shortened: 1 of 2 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
@andreasgerstmayr Thanks a lot for fixing this long-standing issue!

PR #3287, opened 2021-02-26
libbpf-tools: add funclatency

brho:
This is a port of BCC's funclatency.  Usage:

---------
Time functions and print latency as a histogram

Usage: funclatency [-h] [-m|-u] [-p PID] [-d DURATION] [ -i INTERVAL ]
                   [-T] FUNCTION
       Choices for FUNCTION: FUNCTION         (kprobe)
                             LIBRARY:FUNCTION (uprobe a library in -p PID)
                             :FUNCTION        (uprobe the binary of -p PID)

  -m, --milliseconds         Output in milliseconds
  -u, --microseconds         Output in microseconds
  -p, --pid=PID              Process ID to trace
  -d, --duration=DURATION    Duration to trace
  -i, --interval=INTERVAL    Summary interval in seconds
  -T, --timestamp            Print timestamp

  -?, --help                 Give this help list
      --usage                Give a short usage message
  -V, --version              Print program version

Mandatory or optional arguments to long options are also mandatory or optional
for any corresponding short options.

Examples:
  ./funclatency do_sys_open         # time the do_sys_open() kernel function
  ./funclatency -m do_nanosleep     # time do_nanosleep(), in milliseconds
  ./funclatency -u vfs_read         # time vfs_read(), in microseconds
  ./funclatency -p 181 vfs_read     # time process 181 only
  ./funclatency -p 181 c:read       # time the read() C library function
  ./funclatency -p 181 :foo         # time foo() from pid 181's userspace
  ./funclatency -i 2 -d 10 vfs_read # output every 2 seconds, for 10s
  ./funclatency -mTi 5 vfs_read     # output every 5 seconds, with timestamps

---------

It supports kprobes and has limited support for uprobes.  Currently, you
cannot uprobe a library unless you provide a PID.  You also cannot
uprobe a binary without a PID.  This is due to a limitation of libbpf.
Specifically, perf_event_open_probe() will only trace cpu0 for all
processes (pid == -1).  funclatency is set up to support tracing
programs by name for whenever libbpf adds that feature.

It does not support wildcard patterns.

Some of the functions for uprobes are useful for other programs, so I
put those in uprobe_helpers.{c,h}.

Signed-off-by: Barret Rhoden <brho@google.com>

### DIFF
--- a/libbpf-tools/funclatency.bpf.c
+++ b/libbpf-tools/funclatency.bpf.c
@@ -1,0 +1,71 @@
+// SPDX-License-Identifier: GPL-2.0
+/* Copyright (c) 2021 Google LLC. */
+#include "vmlinux.h"
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include "funclatency.h"
+#include "bits.bpf.h"
+
+const volatile pid_t targ_tgid;
+const volatile int units;
+
+/* key: pid.  value: start time */
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_PIDS);
+	__type(key, u32);
+	__type(value, u64);
+} starts SEC(".maps");
+
+__u32 hist[MAX_SLOTS];
+
+SEC("kprobe/dummy_kprobe")
+int BPF_KPROBE(dummy_kprobe)
+{
+	u64 id = bpf_get_current_pid_tgid();
+	u32 tgid = id >> 32;
+	u32 pid = id;
+	u64 nsec;
+
+	if (targ_tgid && targ_tgid != tgid)
+		return 0;
+	nsec = bpf_ktime_get_ns();
+	bpf_map_update_elem(&starts, &pid, &nsec, BPF_ANY);
+
+	return 0;
+}
+
+SEC("kretprobe/dummy_kretprobe")
+int BPF_KRETPROBE(dummy_kretprobe)
+{
+	u64 *start;
+	u64 nsec = bpf_ktime_get_ns();
+	u64 id = bpf_get_current_pid_tgid();
+	u32 pid = id;
+	u64 slot, delta;
+
+	start = bpf_map_lookup_elem(&starts, &pid);
+	if (!start)
+		return 0;
+
+	delta = nsec - *start;
+
+	switch (units) {
+	case USEC:
+		delta /= 1000;
+		break;
+	case MSEC:
+		delta /= 1000000;
+		break;
+	}
+
+	slot = log2l(delta);
+	if (slot >= MAX_SLOTS)
+		slot = MAX_SLOTS - 1;
+	__sync_fetch_and_add(&hist[slot], 1);
+
+	return 0;
+}
+
+char LICENSE[] SEC("license") = "GPL";
--- a/libbpf-tools/funclatency.c
+++ b/libbpf-tools/funclatency.c
@@ -1,0 +1,335 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+/* Copyright (c) 2021 Google LLC.
+ *
+ * Based on funclatency from BCC by Brendan Gregg and others
+ * 2021-02-26   Barret Rhoden   Created this.
+ *
+ * TODO:
+ * - support uprobes on libraries without -p PID. (parse ld.so.cache)
+ * - support regexp pattern matching and per-function histograms
+ */
+#include <argp.h>
+#include <errno.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+#include <unistd.h>
+
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+#include "funclatency.h"
+#include "funclatency.skel.h"
+#include "trace_helpers.h"
+#include "map_helpers.h"
+#include "uprobe_helpers.h"
+
+#define warn(...) fprintf(stderr, __VA_ARGS__)
+
+static struct prog_env {
+	int units;
+	pid_t pid;
+	unsigned int duration;
+	unsigned int interval;
+	unsigned int iterations;
+	bool timestamp;
+	char *funcname;
+} env = {
+	.interval = 99999999,
+	.iterations = 99999999,
+};
+
+const char *argp_program_version = "funclatency 0.1";
+const char *argp_program_bug_address = "<bpf@vger.kernel.org>";
+static const char args_doc[] = "FUNCTION";
+static const char program_doc[] =
+"Time functions and print latency as a histogram\n"
+"\n"
+"Usage: funclatency [-h] [-m|-u] [-p PID] [-d DURATION] [ -i INTERVAL ]\n"
+"                   [-T] FUNCTION\n"
+"       Choices for FUNCTION: FUNCTION         (kprobe)\n"
+"                             LIBRARY:FUNCTION (uprobe a library in -p PID)\n"
+"                             :FUNCTION        (uprobe the binary of -p PID)\n"
+"                             PROGRAM:FUNCTION (uprobe the binary PROGRAM)\n"
+"\v"
+"Examples:\n"
+"  ./funclatency do_sys_open         # time the do_sys_open() kernel function\n"
+"  ./funclatency -m do_nanosleep     # time do_nanosleep(), in milliseconds\n"
+"  ./funclatency -u vfs_read         # time vfs_read(), in microseconds\n"
+"  ./funclatency -p 181 vfs_read     # time process 181 only\n"
+"  ./funclatency -p 181 c:read       # time the read() C library function\n"
+"  ./funclatency -p 181 :foo         # time foo() from pid 181's userspace\n"
+"  ./funclatency -i 2 -d 10 vfs_read # output every 2 seconds, for 10s\n"
+"  ./funclatency -mTi 5 vfs_read     # output every 5 seconds, with timestamps\n"
+;
+
+static const struct argp_option opts[] = {
+	{ "milliseconds", 'm', NULL, 0, "Output in milliseconds"},
+	{ "microseconds", 'u', NULL, 0, "Output in microseconds"},
+	{0, 0, 0, 0, ""},
+	{ "pid", 'p', "PID", 0, "Process ID to trace"},
+	{0, 0, 0, 0, ""},
+	{ "interval", 'i', "INTERVAL", 0, "Summary interval in seconds"},
+	{ "duration", 'd', "DURATION", 0, "Duration to trace"},
+	{ "timestamp", 'T', NULL, 0, "Print timestamp"},
+	{ NULL, 'h', NULL, OPTION_HIDDEN, "Show the full help"},
+	{},
+};
+
+static error_t parse_arg(int key, char *arg, struct argp_state *state)
+{
+	struct prog_env *env = state->input;
+	long duration, interval, pid;
+
+	switch (key) {
+	case 'p':
+		errno = 0;
+		pid = strtol(arg, NULL, 10);
+		if (errno || pid <= 0) {
+			warn("Invalid PID: %s\n", arg);
+			argp_usage(state);
+		}
+		env->pid = pid;
+		break;
+	case 'm':
+		if (env->units != NSEC) {
+			warn("only set one of -m or -u\n");
+			argp_usage(state);
+		}
+		env->units = MSEC;
+		break;
+	case 'u':
+		if (env->units != NSEC) {
+			warn("only set one of -m or -u\n");
+			argp_usage(state);
+		}
+		env->units = USEC;
+		break;
+	case 'd':
+		errno = 0;
+		duration = strtol(arg, NULL, 10);
+		if (errno || duration <= 0) {
+			warn("Invalid duration: %s\n", arg);
+			argp_usage(state);
+		}
+		env->duration = duration;
+		break;
+	case 'i':
+		errno = 0;
+		interval = strtol(arg, NULL, 10);
+		if (errno || interval <= 0) {
+			warn("Invalid interval: %s\n", arg);
+			argp_usage(state);
+		}
+		env->interval = interval;
+		break;
+	case 'T':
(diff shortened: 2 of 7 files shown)


libbpf-tools/funclatency.bpf.c
```diff
@@ -0,0 +1,84 @@
...
+{
+	u64 id = bpf_get_current_pid_tgid();
+	u32 tgid = id >> 32;
+
+	u64 nsec;
+
+	if (targ_tgid && targ_tgid != tgid)
+		return 0;
+	nsec = bpf_ktime_get_ns();
+	bpf_map_update_elem(&starts, &id, &nsec, BPF_ANY);
```
anakryiko, line 40: here you are relying on little-endian byte order for key. Your map expects u32, but you are passing a pointer to u64. Better to split id into pid and tgid explicitly.

libbpf-tools/funclatency.bpf.c
```diff
@@ -0,0 +1,84 @@
...
+		break;
+	case MSEC:
+		val /= 1000000;
+		break;
+	}
+
+	slot = log2l(val);
+	if (slot >= MAX_SLOTS)
+		slot = MAX_SLOTS - 1;
+	count = bpf_map_lookup_elem(&hist, &slot);
```
anakryiko, line 62: again, map expects u32, but you are passing u64

libbpf-tools/funclatency.bpf.c
```diff
@@ -0,0 +1,84 @@
...
+	if (!count)
+		return;
+	*count += 1;
```
anakryiko, line 65: did you consider instead of using per-cpu array and explicit bpf_map_lookup_elem() each time to use global array and use atomic increment? It would use less memory, might be even faster, and will be simpler to consume on user-space part.

libbpf-tools/funclatency.bpf.c
```diff
@@ -0,0 +1,84 @@
...
+}
+
+SEC("kretprobe/dummy_kretprobe")
+int BPF_KRETPROBE(dummy_kretprobe)
+{
+	u64 *start;
+	u64 nsec = bpf_ktime_get_ns();
+	u64 id = bpf_get_current_pid_tgid();
+
+	start = bpf_map_lookup_elem(&starts, &id);
```
anakryiko, line 75: same problem with key size mismatch

libbpf-tools/funclatency.c
```diff
@@ -0,0 +1,370 @@
...
+#include "trace_helpers.h"
+#include "map_helpers.h"
+#include "uprobe_helpers.h"
+
+/* Adapted from linux/err.h */
+#define MAX_ERRNO 4095
+#define ERR_PTR(err) ((void *)((intptr_t)(err)))
+#define PTR_ERR(ptr) ((intptr_t)(ptr))
+#define IS_ERR(x) ((unsigned long)(void *)(x) >= (unsigned long)-MAX_ERRNO)
+#define IS_ERR_OR_NULL(ptr) (!(ptr) || IS_ERR(ptr))
```
anakryiko, line 33: you don't need this, please use `libbpf_get_error()` instead

libbpf-tools/funclatency.c
```diff
@@ -0,0 +1,370 @@
...
+"  ./funclatency -p 181 c:read       # time the read() C library function\n"
+"  ./funclatency -p 181 :foo         # time foo() from pid 181's userspace\n"
+"  ./funclatency -i 2 -d 10 vfs_read # output every 2 seconds, for 10s\n"
+"  ./funclatency -mTi 5 vfs_read     # output every 5 seconds, with timestamps\n"
+;
+
+static const struct argp_option opts[] = {
+	{ "milliseconds", 'm', NULL, 0, "Output in milliseconds"},
+	{ "microseconds", 'u', NULL, 0, "Output in microseconds"},
+	{0, 0, 0, 0, ""},
```
anakryiko, line 80: what are these empty slots for?

libbpf-tools/funclatency.c
```diff
@@ -0,0 +1,370 @@
...
+	print_log2_hist(hist, MAX_SLOTS, unit_str());
+}
+
+static int attach_kprobes(struct funclatency_bpf *obj)
+{
+	struct bpf_link *l;
+
+	l = bpf_program__attach_kprobe(obj->progs.dummy_kprobe, false,
+				       env.funcname);
+	if (IS_ERR(l)) {
```
anakryiko, line 209: err = libbpf_get_error(l);
if (err) {
    warn(..., err);
    return -1;
}

libbpf-tools/funclatency.c
```diff
@@ -0,0 +1,370 @@
...
+
+static int attach_uprobes(struct funclatency_bpf *obj)
+{
+	char *binary, *function, *bin_path;
+	off_t func_off;
+	int ret = -1;
+	struct bpf_link *l;
+
+	/*
+	 * TODO: libbpf cannot attach uprobes to all processes on all cpus.  It
```
anakryiko, line 233: Are you sure about this? How did you test? We have kernel selftest that does uretprobe for any PID. I just modified that test to use pid==-1 for both uprobe and uretprobe and triggered uprobe from a forked process. It worked. So I wonder what didn't work in your case?

As far as I can see, BCC doesn't do anything special beyond what libbpf does when setting up uprobes, so both should work the same.

libbpf-tools/funclatency.c
```diff
@@ -0,0 +1,370 @@
...
+	 */
+	if (!env.pid) {
+		warn("Need a PID for uprobes (for now)\n");
+		return -1;
+	}
+
+	binary = strdup(env.funcname);
+	if (!binary)
+		handle_error("strdup");
+	function = strchr(binary, ':');
```
anakryiko, line 245: probably worthwhile to leave a comment that binary is guaranteed to contain ':' (handled by attach_probes below)

libbpf-tools/funclatency.c
```diff
@@ -0,0 +1,370 @@
...
+			time(&t);
+			tm = localtime(&t);
+			strftime(ts, sizeof(ts), "%H:%M:%S", tm);
+			printf("%-8s\n", ts);
+		}
+
+		print_hist(bpf_map__fd(obj->maps.hist));
+
+		if (exiting)
+			break;
```
anakryiko, line 361: nit: make it part of for exit condition above?

anakryiko (commented):
I only skimmed the uprobe_helpers part for now. I wonder if putting them into trace_helpers.c would be better (uprobes are tracing after all).

libbpf-tools/funclatency.c
brho: They help with sorting the arg options by breaking the options into 'groups'.  Argp sorts them alphabetically within these groups.  The empty slot will make it so that -m and -u are next to each other.

libbpf-tools/funclatency.c
brho: I woke up this morning and realized that, yeah, this might be wrong.  I was basing it off of perf_event_open not supporting pid = -1 and cpu = -1, and that BCC was doing a "for each cpu" in perf_event_open().

after testing, it does work as intended.  i taskset the traced process to a different cpu than funclatency just to make sure.

libbpf-tools/funclatency.bpf.c
brho: I was trying to avoid the overhead of the atomic operation and cache line bouncing.  Though that was 'preventative'. 

I tried out both versions and didn't see a clear winner either way, so I'll switch it to the global array.

libbpf-tools/funclatency.bpf.c
brho: agreed that should be a u32.  FWIW, the reason i had `u64 slot` was from earlier prototyping on a global array:

`__sync_fetch_and_add(&hist[slot], 1);`

for that, slot needs to be a u64, otherwise the program was failing verification.

libbpf-tools/funclatency.c
brho: added a check for !function and warn() + return.

brho:
> I only skimmed the uprobe_helpers part for now. I wonder if putting them into trace_helpers.c would be better (uprobes are tracing after all).

Either way works for me.  The stuff in uprobe_helpers is mostly about finding programs, libraries, and offsets within elf files.

brho:
hi anakryiko, i made the changes you pointed out, other than moving stuff to a new file.  i clicked the 'request review' button, but it might not work due to permissions issues.  not trying to rush things - just wanted to make sure you saw the latest on this patchset.  =)

libbpf-tools/funclatency.c
```diff
@@ -0,0 +1,335 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+/* Copyright (c) 2021 Google LLC.
+ *
+ * Based on funclatency from BCC by Brendan Gregg and others
+ * 2021-02-26   Barret Rhoden   Created this.
+ *
+ * TODO:
+ * - support uprobes on libraries without -p PID. (parse ld.so.cache)
```
anakryiko, line 8: looking briefly at command line args of funclatency.py, it seems like there are more features that are not supported (e.g., regexp pattern matching and per-function historgrams). It's fine, but let's document all the lacking features somewhere here?

anakryiko (changes requested):
it looks good overall, but I had a bunch of naming and formatting nits below. And also the overuse of malloc() seems avoidable without sacrificing anything, yet getting simpler APIs, please check the comments.

libbpf-tools/uprobe_helpers.c
```diff
@@ -0,0 +1,263 @@
...
+#include <string.h>
+#include <unistd.h>
+#include <errno.h>
+#include <limits.h>
+#include <gelf.h>
+
+#define warn(...) fprintf(stderr, __VA_ARGS__)
+
+/* Returns the full path to the program for pid.  free() the result. */
+char *get_path_to_pid(pid_t pid)
```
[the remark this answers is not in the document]
brho: agreed.

libbpf-tools/uprobe_helpers.c
```diff
@@ -0,0 +1,263 @@
...
+/* Returns the offset of a function in the elf file `path`, or -1 on failure. */
+off_t func_offset_in_elf(const char *path, const char *func)
+{
+	off_t ret = -1;
+	int fd = -1;
+	Elf *e;
+	Elf_Scn *scn;
+	Elf_Data *data;
+	GElf_Shdr shdr[1];
+	GElf_Sym sym[1];
```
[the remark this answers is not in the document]
brho: Yeah, that way you can use `sym` and `shdr` as pointers instead of & and `.`.  

It's more useful when you have a lot of code that already treats an object as a pointer, e.g. other helper functions operating on the same object.  That way, all of the functions operate similarly, and you don't have to write one function specially (with the & and .).

In this case here, it doesn't matter much.  It'd be a minor annoyance if we refactored the code and pulled the elf_nextscn loop into a helper function.  Once I picked up that trick, I've been using it proactively.  =)

Can change it if you'd like.

libbpf-tools/uprobe_helpers.c
```diff
@@ -0,0 +1,263 @@
...
+
+	if (elf_getshdrstrndx(e, &shstrndx) != 0)
+		goto out;
+
+	scn = NULL;
+	while ((scn = elf_nextscn(e, scn))) {
+		if (!gelf_getshdr(scn, shdr))
+			continue;
+		if (!(shdr->sh_type == SHT_SYMTAB ||
+		      shdr->sh_type == SHT_DYNSYM))
```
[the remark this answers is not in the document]
brho: it hit the 80 column limit.  (82)  though i guess if it was '.' instead of '->', then we'd be ok.  =)

do you all have a limit max?  i usually just run my patches through checkpatch and have vim set up to not go over 80 by default.

libbpf-tools/uprobe_helpers.c
```diff
@@ -0,0 +1,263 @@
...
+		warn("path /proc/PID/maps failed");
+		return NULL;
+	}
+	maps = fopen(proc_pid_maps, "r");
+	free(proc_pid_maps);
+	if (!maps) {
+		warn("No such pid %d\n", pid);
+		return NULL;
+	}
+	path = malloc(path_sz);
```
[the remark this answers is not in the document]
brho: good idea.

brho:
pushed a bunch of fixes.  PTAL.  thanks for the review!

libbpf-tools/uprobe_helpers.c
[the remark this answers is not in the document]
anakryiko: kernel line length limit was relaxed to 100 characters, so if I can keep the line from wrapping with 100 characters, I usually do

libbpf-tools/funclatency.c
```diff
@@ -0,0 +1,333 @@
...
+	if (resolve_binary_path(binary, env.pid, bin_path, sizeof(bin_path)))
+		goto out_binary;
+
+	func_off = get_elf_func_offset(bin_path, function);
+	if (func_off < 0) {
+		warn("Could not find %s in %s\n", function, bin_path);
+		goto out_binary;
+	}
+
+	l = bpf_program__attach_uprobe(obj->progs.dummy_kprobe, false,
```
anakryiko, line 226: ok, so somehow I didn't pay attention to this initially.

You are leaking link here (and you are not explicitly detaching it, neither on error, nor in the normal flow).

You can store returned link into the skeleton like this:

```
obj->links.dummy_kprobe = bpf_program__attach_uprobe(...);
```

Later, when you destory the skeleton, those links will be destroyed as well. It's fine if link themselves failed (so no need to NULL them out)

brho: thanks.  i was just using 'l' for the error check; didn't know they needed to be set like that.

fixed and pushed.

anakryiko (commented):
So looks good overall except the bpf_link leak issue. Please fix that.

And I'd unwrap a bunch of lines that are just barely over 80 characters :)

anakryiko (approved):
got distracted for a bit and you already updated the code :) lgtm

brho:
> And I'd unwrap a bunch of lines that are just barely over 80 characters :)

Done!  Found and unwrapped a few.

Thanks.  =)

yonghong-song:
[buildbot, test this please]

PR #3288, opened 2021-02-28
Use string type for funccount comparison to PATH elements

jongiddy:
`BPF.find_exe` requires a string to match against `PATH`, and the `USDT` constructor requires a string for the `path` parameter, so that it can be encoded. However, the `ArgString` class returns a `bytes` from the `__str__` method. With Python 3, these type mismatches cause errors.

The fix modifies `ArgString` to return a `str` from `__str__` and uses that to pass a `str` in `funccount.py`.

### DIFF
--- a/src/python/bcc/utils.py
+++ b/src/python/bcc/utils.py
@@ -77,7 +77,7 @@ class ArgString(object):
         return self.s.encode(FILESYSTEMENCODING)
 
     def __str__(self):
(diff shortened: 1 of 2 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

PR #3289, opened 2021-03-01
fix wakeuptime's raw_tp: get awakened task from ctx not current

ethercflow:


yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks for the fix! There are cases where p in `trace_sched_wakeup(p)` is the *current* but in some other cases they are not equal. Make sense to use tracepoint argument instead of *current*.

PR #3296, opened 2021-03-06
Add mypy type checking

jongiddy:
I've previously found issues with strings and bytes being mixed when running bcc tools using Python 3 (e.g. https://github.com/iovisor/bcc/pull/3277, https://github.com/iovisor/bcc/pull/3288). Rather than hit these randomly while using the tools, this PR adds mypy testing for the Python code.

This PR adds a GitHub Actions test that checks both Python 2 and Python 3 types. It adds code fixes and ignores to get the repo to pass both these checks. The type annotations are in comment form that is acceptable to both Python 2 and 3.

I also added function type descriptions to the `ArgString` type, to ensure that the change from https://github.com/iovisor/bcc/pull/3288 didn't break anything unexpected. The intent is that more function type descriptions can be added over time to ensure that the code works well with both Python 2 and Python 3. But I want to check that this change is acceptable before adding more annotations.


yonghong-song:
[buildbot, test this please]

yonghong-song:
In general, additional checking/enforcement is a good thing. But I am not 100% sure about the long-term maintenance and what kind of requirement we will have for new python codes. @brendangregg what do you think?

jongiddy:
One alternative would be to remove the GitHub Action that does type checking, but keep the type comments and minor code fixes. Even without being checked, the comments are useful.

Like all unchecked comments, they may drift into being inaccurate over time. But this ensures that no-one will ever be blocked by working code that breaks the type signatures (e.g. adding a parameter to a function but forgetting to change the type signature).

PR #3298, opened 2021-03-08
tools: add LPORT option for tcpconnlat output

suresh2514:


yonghong-song:
[buildbot, test this please]

yonghong-song:
@suresh2514 could you add a description in the commit message why you want add LPORT? the example file and the man page should also change with LPORT addition. @brendangregg could you take a look at well? It must be a reason why you did not add LPORT when you first wrote the tool.

suresh2514:
Hi,

I am closing this request as @brendangregg  suggested try something like using '-L' option to include LPORT.

PR #3299, opened 2021-03-08
tools: add LPORT output for tcpconnect tool

suresh2514:
LPORT information will help in uniquely  identifying a socket. 
For example, this info will be of much help when analyzing a tcpdump.

### DIFF
--- a/tools/tcpconnect.py
+++ b/tools/tcpconnect.py
@@ -19,6 +19,7 @@
 # 09-Jan-2019   Takuma Kume     Support filtering by UID
 # 30-Jul-2019   Xiaozhou Liu    Count connects.
 # 07-Oct-2020   Nabil Schear    Correlate connects with DNS responses
+# 08-Mar-2021   Suresh Kumar    Added LPORT option
 
 from __future__ import print_function
 from bcc import BPF
@@ -87,6 +88,7 @@ struct ipv4_data_t {
     u32 saddr;
     u32 daddr;
     u64 ip;
+    u16 lport;
     u16 dport;
     char task[TASK_COMM_LEN];
 };
@@ -99,6 +101,7 @@ struct ipv6_data_t {
     unsigned __int128 saddr;
     unsigned __int128 daddr;
     u64 ip;
+    u16 lport;
     u16 dport;
     char task[TASK_COMM_LEN];
 };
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
@suresh2514 the example file and the man page also need change. @brendangregg could you take a look at this lport addition as well?

brendangregg:
Well, I'm glad it doesn't bust >80 char width, but for info that's rarely needed I'd rather see it under switches (like -x for extended columns) to avoid clutter. I've never needed LPORT on connect; I understand if you are cross-referencing with tcpdump you'd want it, but that's not the common case (tcpconnect also dates back to 2010, and this is the first time someone has asked for LPORT). Every extra bit we add to the default output forces everyone to consider it, adding to the burden of learning these tools, so I try hard to keep the default output to the bare minimum to solve the common problems.

This is also the case with other Unix tools (e.g., `iostat` vs `iostat -x`, `ls` vs `ls -l`, etc.).

brendangregg:
... or a -L for adding the LPORT column, similar to how it already has -U for UID.

suresh2514:
Hi @brendangregg,
thanks. 
We have been using other bcc tools like tcptop, tcpaccept, tcpstates etc.  They all carry the LPORT information as default output. Eg:
+++
./tcptop
PID    COMM         LADDR                 RADDR                  RX_KB  TX_KB
5953   chrome       192.168.1.3:56852     192.168.122.205:443         2      2

# ./tcpaccept 
PID     COMM         IP RADDR            RPORT LADDR            LPORT

# ./tcpretrans 
Tracing retransmits ... Hit Ctrl-C to end
TIME     PID    IP LADDR:LPORT          T> RADDR:RPORT          STATE
+++

Only for 'tcpconnlat' and 'tcpconnect', we miss these outputs by default. May be we can include it along with ip-address  to keep it short  like 'LADDR:LPORT?. Before I add another commit with '-L' option, thought of checking with you again.

brendangregg:
Right, because LPORT is the common use case in those tools. With tcpaccept it's showing the server port. With tcpretrans it could be the server port, or it's identifying the socket.

With tcpconnect the common use case is: what hosts and destination ports are which processes connecting to? The default output solves that common use case.

suresh2514:
@brendangregg 

cool. thanks.  I will close this request and will try with '-L' option.

suresh2514:
@brendangregg , 
I have made the changes to use '-L' option to print LPORT
New PR: https://github.com/iovisor/bcc/pull/3301

PR #3302, opened 2021-03-10
tools/virtiostat: add filter

pizhenwei:
Add device driver/name filter for virtiostat.

### DIFF
--- a/tools/virtiostat.py
+++ b/tools/virtiostat.py
@@ -4,7 +4,7 @@
 # virtiostat    Show virtio devices input/output statistics.
 #               For Linux, uses BCC, eBPF.
 #
-# USAGE: virtiostat [-h] [-T] [-D] [INTERVAL] [COUNT]
+# USAGE: virtiostat [-h] [-T] [-D] [-d DRIVER] [-n DEVNAME] [INTERVAL] [COUNT]
 #
 # Copyright (c) 2021 zhenwei pi
 # Licensed under the Apache License, Version 2.0 (the "License")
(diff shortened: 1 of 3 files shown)


pizhenwei:
@brendangregg Sorry, I tried to implement '-x' mode (add latency stat), but failed. It's hard to distinguish the each requests' done time from virtio layer(Ex, 128 IO-depth case).

yonghong-song:
[buildbot, test this please]

tools/virtiostat.py
```diff
@@ -52,6 +58,25 @@
 /* typically virtio blk has max SEG of 128 */
 #define SG_MAX          128
 
+/* local strcmp function, max length 16 to protect instruction loops */
+#define CMPMAX	16
+
+static int local_strcmp(const char *cs, const char *ct)
+{
+	int len = 0;
+        unsigned char c1, c2;
```
yonghong-song, line 67: misaligned here. For below while loop as well, maybe to use 4 byte indentation consistently, also similar to other codes.

yonghong-song:
[buildbot, ok to test]

tools/virtiostat.py
pizhenwei: Sure.

tools/virtiostat.py
```diff
@@ -194,6 +222,29 @@
 }
 """
 
+# filter for driver name
+if args.driver:
+    bpf_text = bpf_text.replace('DRIVERFILTER',
+        """char filter_driver[] = \"%s\";
+        char driver[16];
+        bpf_probe_read_kernel(driver, sizeof(driver), vq->vdev->dev.driver->name);
```
yonghong-song, line 230: Since vq->vdev->dev.driver->name is really a string, let us use bpf_probe_read_kernel_str here.

tools/virtiostat.py
```diff
@@ -194,6 +222,29 @@
...
+        return;""" % (args.driver))
+else:
+    bpf_text = bpf_text.replace('DRIVERFILTER', '')
+
+# filter for dev name
+if args.devname:
+    bpf_text = bpf_text.replace('DEVNAMEFILTER',
+        """char filter_devname[] = \"%s\";
+        char devname[16];
+        bpf_probe_read_kernel(devname, sizeof(devname), vq->vdev->dev.kobj.name);
```
yonghong-song, line 241: bpf_probe_read_kernel_str()?

pizhenwei: bpf_probe_read_kernel_str is better to copy a string from kernel, so also replace other codes together.

yonghong-song:
[buildbot, test this please]

PR #3306, opened 2021-03-11
tools/skbstat: add skbstat tool

Dongdong-Wang:
This tool can help us find out the cause of packet(skb) loss,
by dumping the backtrace of the kernel function `kfree_skb`.

We can pass IP address/protocol number/port number to `skbstat`,
then it will filter none related packets dropped by kernel.

In addition, this tool can record the skb's main processing routines
by the kernel network stack.

Signed-off-by: Dongdong Wang <wangdongdong.6@bytedance.com>

### DIFF
--- a/tools/skbstat.py
+++ b/tools/skbstat.py
@@ -10,6 +10,10 @@
 #
 # This uses dynamic tracing of kernel functions, and will need to be updated
 # to match kernel changes.
+# 
+# WARNING: When run with `--input/output/consume` options,
+# this tool will trace high traffic kernel functions and may lead to
+# some noticeable overhead. Use it carefully on production system.
 #
 # Copyright (c) 2021 Dongdong Wang <wangdongdong.6@bytedance.com>
 # Licensed under the Apache License, Version 2.0 (the "License")
@@ -40,13 +44,7 @@ examples = """examples:
     skbstat -p tcp -P 80                  # only trace input or output HTTP packet
 """
 parser = argparse.ArgumentParser(
-    description="""
-    Trace skb event (input/output/consume/drop) of the kernel.
-
-    WARNING: When run with `--input/output/consume` options, 
-    this tool will trace high traffic kernel functions and may lead to 
-    some noticeable overhead. Use it carefully on production system.
-""",
+    description="Trace skb event (input/output/consume/drop) of the kernel",
     formatter_class=argparse.RawDescriptionHelpFormatter,
     epilog=examples)
 
@@ -269,9 +267,8 @@ static inline int port_match1(u16 sport, u16 dport, struct filter *filter)
 
 static inline int port_match(u16 sport, u16 dport, struct filter *filter)
 {
-	return (filter->flags & SKB_PORT_MATCH_1) ? 
-            port_match1(sport, dport, filter) :
-	    port_match2(sport, dport, filter);
+	return (filter->flags & SKB_PORT_MATCH_1) ?
+	    port_match1(sport, dport, filter) : port_match2(sport, dport, filter);
 }
 
 static inline int proto_match(u8 proto, struct filter *filter)
@@ -282,10 +279,10 @@ static inline int proto_match(u8 proto, struct filter *filter)
 static inline int ip6_addr_match2(unsigned __int128 saddr,
 				  unsigned __int128 daddr, struct filter *filter)
 {
-	unsigned __int128 saddr_expect = *((unsigned __int128 *)(filter->addr6[0])), 
-            daddr_expect = *((unsigned __int128 *)(filter->addr6[1]));
-	unsigned __int128 saddr_mask = *((unsigned __int128 *)(filter->mask6[0])), 
-            daddr_mask = *((unsigned __int128 *)(filter->mask6[1]));
+	unsigned __int128 saddr_expect = *((unsigned __int128 *)(filter->addr6[0])),
+	    daddr_expect = *((unsigned __int128 *)(filter->addr6[1]));
+	unsigned __int128 saddr_mask = *((unsigned __int128 *)(filter->mask6[0])),
+	    daddr_mask = *((unsigned __int128 *)(filter->mask6[1]));
 
 	return (saddr_expect ? (saddr & saddr_mask) == saddr_expect : 1)
 	    && (daddr_expect ? (daddr & daddr_mask) == daddr_expect : 1);
@@ -294,10 +291,8 @@ static inline int ip6_addr_match2(unsigned __int128 saddr,
 static inline int ip6_addr_match1(unsigned __int128 saddr,
 				  unsigned __int128 daddr, struct filter *filter)
 {
-	unsigned __int128 addr_expect =
-	    *((unsigned __int128 *)(filter->addr6[0]));
-	unsigned __int128 addr_mask =
-	    *((unsigned __int128 *)(filter->mask6[0]));
+	unsigned __int128 addr_expect = *((unsigned __int128 *)(filter->addr6[0]));
+	unsigned __int128 addr_mask = *((unsigned __int128 *)(filter->mask6[0]));
 
 	return ((saddr & addr_mask) == addr_expect)
 	    || ((daddr & addr_mask) == addr_expect);
@@ -306,17 +301,14 @@ static inline int ip6_addr_match1(unsigned __int128 saddr,
 static inline int ip6_addr_match(unsigned __int128 saddr,
 				 unsigned __int128 daddr, struct filter *filter)
 {
-	return (filter->flags & SKB_ADDR_MATCH_1) ? 
-            ip6_addr_match1(saddr, daddr, filter) :
-	    ip6_addr_match2(saddr, daddr, filter);
+	return (filter->flags & SKB_ADDR_MATCH_1) ?
+	    ip6_addr_match1(saddr, daddr, filter) : ip6_addr_match2(saddr, daddr, filter);
 }
 
 static inline int ip_addr_match2(__be32 saddr, __be32 daddr, struct filter *filter)
 {
-	return (filter->addr[0] ? (saddr & filter->mask[0]) ==
-		filter->addr[0] : 1)
-	    && (filter->addr[1] ? (daddr & filter->mask[1]) ==
-		filter->addr[1] : 1);
+	return (filter->addr[0] ? (saddr & filter->mask[0]) == filter->addr[0] : 1)
+	    && (filter->addr[1] ? (daddr & filter->mask[1]) == filter->addr[1] : 1);
 }
 
 static inline int ip_addr_match1(__be32 saddr, __be32 daddr, struct filter *filter)
@@ -327,9 +319,8 @@ static inline int ip_addr_match1(__be32 saddr, __be32 daddr, struct filter *filt
 
 static inline int ip_addr_match(__be32 saddr, __be32 daddr, struct filter *filter)
 {
-	return (filter->flags & SKB_ADDR_MATCH_1) ? 
-            ip_addr_match1(saddr, daddr, filter) :
-	    ip_addr_match2(saddr, daddr, filter);
+	return (filter->flags & SKB_ADDR_MATCH_1) ?
+	    ip_addr_match1(saddr, daddr, filter) : ip_addr_match2(saddr, daddr, filter);
 }
 
 static inline int record_skb(struct pt_regs *ctx, union filter_args *args)
@@ -343,11 +334,9 @@ static inline int record_skb(struct pt_regs *ctx, union filter_args *args)
 	} else if (args->forward) {
 		count = forward.lookup_or_try_init(&ip, &zero);
 	}
-
 	if (count) {
 		*count += 1;
 	}
-
 	return 0;
 }
 
@@ -356,8 +345,7 @@ static inline int record_sk(struct pt_regs *ctx, union filter_args *args)
 	return record_skb(ctx, args);
 }
 
-static inline int record_skb_backtrace(struct pt_regs *ctx,
-				       union filter_args *args)
+static inline int record_skb_backtrace(struct pt_regs *ctx, union filter_args *args)
 {
 	u32 stack_id = stack_traces.get_stackid(ctx, 0);
 	u64 *count, zero = 0;
@@ -367,16 +355,13 @@ static inline int record_skb_backtrace(struct pt_regs *ctx,
 	} else {
 		count = consume_stacks.lookup_or_try_init(&stack_id, &zero);
 	}
-
 	if (count) {
 		*count += 1;
 	}
-
 	return 0;
 }
 
-static inline int skb_l4_match(u8 proto, union l4hdr *hdr,
-			       struct filter *filter)
+static inline int skb_l4_match(u8 proto, union l4hdr *hdr, struct filter *filter)
 {
 	switch (proto) {
 	case IPPROTO_TCP:
@@ -387,32 +372,27 @@ static inline int skb_l4_match(u8 proto, union l4hdr *hdr,
 			return port_match(src, dst, filter);
 		}
 	default:
-                // TODO: parse other protocols and do some match.
+		// TODO: parse other protocols and do some match.
 		break;
 	}
 	return 1;
 }
 
-static inline int skb_l4hdr(struct sk_buff *skb, union l4hdr *l4h,
-			    int l3_offset)
+static inline int skb_l4hdr(struct sk_buff *skb, union l4hdr *l4h, int l3_offset)
 {
 	// transport header was set
 	if (skb->transport_header != (typeof(skb->transport_header)) ~ 0U &&
 	    skb->transport_header != skb->network_header) {
 		if (bpf_probe_read_kernel
-		    (l4h, sizeof(union l4hdr),
-		     skb->head + skb->transport_header)) {
+		    (l4h, sizeof(union l4hdr), skb->head + skb->transport_header)) {
 			goto err;
 		}
 		goto success;
 	}
-
 	if (bpf_probe_read_kernel
-	    (l4h, sizeof(union l4hdr),
-	     skb->head + skb->network_header + l3_offset)) {
+	    (l4h, sizeof(union l4hdr), skb->head + skb->network_header + l3_offset)) {
 		goto err;
 	}
-
 success:
 	return 0;
 err:
@@ -425,19 +405,15 @@ static inline int skb_l3hdr(struct sk_buff *skb, void *l3h, int l3h_len)
 	if (!skb->network_header) {
 		goto err;
 	}
-
-	if (bpf_probe_read_kernel
-	    (l3h, l3h_len, skb->head + skb->network_header)) {
+	if (bpf_probe_read_kernel(l3h, l3h_len, skb->head + skb->network_header)) {
 		goto err;
 	}
-
 	return 0;
 err:
 	return -1;
 }
 
-static inline int sk_ip6_match(struct sock *sk, struct filter *filter,
-			       union filter_args *args)
+static inline int sk_ip6_match(struct sock *sk, struct filter *filter, union filter_args *args)
 {
 	unsigned __int128 saddr, daddr;
 	u16 sport, dport;
@@ -456,7 +432,6 @@ static inline int sk_ip6_match(struct sock *sk, struct filter *filter,
 	} else {
 		goto miss;
 	}
-
 	if (ip6_addr_match(saddr, daddr, filter)) {
 		return port_match(htons(sport), htons(dport), filter);
 	}
@@ -472,25 +447,20 @@ static inline int skb_ip6_match(struct sk_buff *skb, struct filter *filter)
 	if (skb_l3hdr(skb, &ip6h, sizeof(struct ipv6hdr))) {
 		goto miss;
 	}
-
 	if (!ip6_addr_match
-	    (*((unsigned __int128 *)(&ip6h.saddr)),
-	     *((unsigned __int128 *)(&ip6h.daddr)), filter)
+	    (*((unsigned __int128 *)(&ip6h.saddr)), *((unsigned __int128 *)(&ip6h.daddr)), filter)
 	    || !proto_match(ip6h.nexthdr, filter)) {
 		goto miss;
 	}
-
 	if (skb_l4hdr(skb, &l4h, sizeof(struct ipv6hdr))) {
 		goto miss;
 	}
-
 	return skb_l4_match(ip6h.nexthdr, &l4h, filter);
 miss:
 	return 0;
 }
 
-static inline int sk_ip_match(struct sock *sk, struct filter *filter,
-			      union filter_args *args)
+static inline int sk_ip_match(struct sock *sk, struct filter *filter, union filter_args *args)
 {
 	u32 saddr, daddr;
 	u16 sport, dport;
@@ -509,7 +479,6 @@ static inline int sk_ip_match(struct sock *sk, struct filter *filter,
 	} else {
 		goto miss;
 	}
-
 	if (ip_addr_match(saddr, daddr, filter)) {
 		return port_match(htons(sport), htons(dport), filter);
 	}
@@ -525,16 +494,13 @@ static inline int skb_ip_match(struct sk_buff *skb, struct filter *filter)
 	if (skb_l3hdr(skb, &iph, sizeof(struct iphdr))) {
 		goto miss;
 	}
-
 	if (!ip_addr_match(iph.saddr, iph.daddr, filter)
 	    || !proto_match(iph.protocol, filter)) {
 		goto miss;
 	}
-
 	if (skb_l4hdr(skb, &l4h, iph.ihl * 4)) {
 		goto miss;
 	}
-
 	return skb_l4_match(iph.protocol, &l4h, filter);
 miss:
 	return 0;
@@ -565,16 +531,15 @@ static inline void filter_init(struct filter *filter)
 	SADDR_FILTER
 	DADDR_FILTER
 	SPORT_FILTER
-	DPORT_FILTER 
-	PROTO_FILTER 
-	FLAGS_FILTER 
+	DPORT_FILTER
+	PROTO_FILTER
+	FLAGS_FILTER
 	IP_VERSION_FILTER
 }
 
 static inline int sk_match_and_trace(struct pt_regs *ctx, struct sock *sk,
 				     union filter_args *args,
-				     int (*cb) (struct pt_regs *,
-						union filter_args *))
+				     int (*cb) (struct pt_regs *, union filter_args *))
 {
 	struct filter filter = { };
 
@@ -591,14 +556,12 @@ static inline int sk_match_and_trace(struct pt_regs *ctx, struct sock *sk,
 			return cb(ctx, args);
 		}
 	}
-
 	return 0;
 }
 
 static inline int skb_match_and_trace(struct pt_regs *ctx, struct sk_buff *skb,
 				      union filter_args *args,
-				      int (*cb) (struct pt_regs *,
-						 union filter_args *))
+				      int (*cb) (struct pt_regs *, union filter_args *))
 {
 	struct filter filter = { };
 
@@ -615,7 +578,6 @@ static inline int skb_match_and_trace(struct pt_regs *ctx, struct sk_buff *skb,
 			return cb(ctx, args);
 		}
 	}
-
 	return 0;
 }
 
@@ -624,7 +586,6 @@ static inline int __trace_sk_input(struct pt_regs *ctx, struct sock *sk)
 	union filter_args args = { };
 
 	args.input = true;
-
 	return sk_match_and_trace(ctx, sk, &args, &record_sk);
 }
 
@@ -633,7 +594,6 @@ static inline int __trace_skb_input(struct pt_regs *ctx, struct sk_buff *skb)
 	union filter_args args = { };
 
 	args.input = true;
-
 	return skb_match_and_trace(ctx, skb, &args, &record_skb);
 }
 
@@ -642,7 +602,6 @@ static inline int __trace_sk_output(struct pt_regs *ctx, struct sock *sk)
 	union filter_args args = { };
 
 	args.output = true;
-
 	return sk_match_and_trace(ctx, sk, &args, &record_sk);
 }
 
@@ -651,7 +610,6 @@ static inline int __trace_skb_output(struct pt_regs *ctx, struct sk_buff *skb)
 	union filter_args args = { };
 
 	args.output = true;
-
 	return skb_match_and_trace(ctx, skb, &args, &record_skb);
 }
 
@@ -660,8 +618,7 @@ int trace_sk_input(struct pt_regs *ctx, struct sock *sk)
 	return __trace_sk_input(ctx, sk);
 }
 
-int trace_skb_input_arg3(struct pt_regs *ctx, void *arg1, void *arg2,
-			 struct sk_buff *skb)
+int trace_skb_input_arg3(struct pt_regs *ctx, void *arg1, void *arg2, struct sk_buff *skb)
 {
 	return __trace_skb_input(ctx, skb);
 }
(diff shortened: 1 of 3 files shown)


tools/skbstat.py
```diff
@@ -0,0 +1,996 @@
...
+import ipaddress
+from time import strftime
+from struct import unpack
+from time import sleep
+
+# arguments
+examples = """examples:
+    ./skbstat                               # dump all skb dropping backtraces
+    ./skbstat --ingress                     # dump skb ingress routines
+    ./skbstat --egress                      # dump skb egress routines
```
yonghong-song, line 28: egress -> output?

You may add an example command line like
    skbstat --consume
?

tools/skbstat.py
```diff
@@ -0,0 +1,996 @@
...
+    ./skbstat -s 10.0.0.1                   # only trace packet comming from 10.0.0.1
```
yonghong-song, line 29: I hit an error here on python3
```
$ sudo ./skbstat.py -s 10.0.0.1
  File "./skbstat.py", line 110, in <module>
    saddr_filter, ip_version = init_addr_filter(args.saddr, 0)
  File "./skbstat.py", line 81, in init_addr_filter
    net = ipaddress.ip_network(unicode(addr_str))
NameError: name 'unicode' is not defined
$
```

tools/skbstat.py
```diff
@@ -0,0 +1,996 @@
...
+if args.ebpf:
+    print(bpf_text)
+    exit()
+
+# initialize BPF
+b = BPF(text=bpf_text)
+
+# probe list
+probe_list_input_dev = {
+    "netif_receive_skb_internal": ("trace_skb_input", 30),
```
yonghong-song, line 721: I hit the following error:
```
$ sudo ./skbstat.py --input
ERROR: netif_receive_skb_internal() kernel function not found or traceable. Older kernel versions not supported.
```

The function is inlined in netif_receive_skb
```
int netif_receive_skb(struct sk_buff *skb)
{       
        int ret;
        
        trace_netif_receive_skb_entry(skb);

        ret = netif_receive_skb_internal(skb);
        trace_netif_receive_skb_exit(ret);
        
        return ret;
}       
EXPORT_SYMBOL(netif_receive_skb);
```
Any particular reason you did not trace func `netif_receive_skb` or tracepoint/raw_tracepoint for netif_receive_skb_entry?

tools/skbstat.py
```diff
@@ -0,0 +1,996 @@
...
+}
+
+probe_list_input_ip = {
+    "ip_error": ("trace_skb_input", 19),
```
yonghong-song, line 725: I see below codes that the number "19" is used for sorting. Some comments will be good to understand algorithm without digging into the implementation.

tools/skbstat.py
```diff
@@ -0,0 +1,996 @@
...
+    "ip_rcv": ("trace_skb_input", 19),
+    "ip_rcv_finish": ("trace_skb_input_arg3", 18),
+    "ip_route_input_noref": ("trace_skb_input", 17),
+    "ip_local_deliver": ("trace_skb_input", 16),
+    "ip_local_deliver_finish": ("trace_skb_input_arg3", 15),
+}
+
+probe_list_input_ip6 = {
+    "ip6_pkt_drop": ("trace_skb_input", 19),
```
yonghong-song, line 734: Some other kernel functions in this tool may also be subject to potential inlining issues, e.g., this one `ip6_pkt_drop`. I checked vmlinux source code, it seems fine at least for my machine. It would be
good to put a comment here to indicate a potential issue as in the future the compiler might inline
one or more call sites. Other functions may have this issues are:
```
    ip6_input
    tcp_drop
    ip_local_out
    __ip_local_out
    __dev_queue_xmit
   dev_hard_start_xmit
```

tools/skbstat.py
```diff
@@ -0,0 +1,996 @@
...
+    except KeyboardInterrupt:
+        exit()
+
+    if args.input:
+        print_input_routines()
+
+    if args.output:
+        print_output_routines()
+
+    print_all_backtraces(args.consume, True)
```
yonghong-song, line 996: drop is always true here.

tools/skbstat_example.txt
```diff
@@ -0,0 +1,82 @@
...
+  -P PORT, --port PORT  source or destination tcp/udp port
+  -p PROTO, --proto PROTO
+                        protocol ( tcp/udp/icmp ... )
+  --input               trace skb input
+  --output              trace skb output
+  --consume             dump consume_skb backtrace
+
+examples:
+    ./skbstat                               # dump all skb dropping backtraces
+    ./skbstat --ingress                     # dump skb ingress routines
```
yonghong-song, line 76: --input

tools/skbstat_example.txt
```diff
@@ -0,0 +1,82 @@
...
+    ./skbstat --egress                      # dump skb egress routines
```
yonghong-song, line 77: --output

tools/skbstat_example.txt
```diff
@@ -0,0 +1,82 @@
...
+                ip_rcv+0x52
+                __netif_receive_skb_one_core+0x50
+                netif_receive_skb_internal+0x2f
+                napi_gro_receive+0x107
+
+[...]
+
+The last column shows the number of function calls or skb dropped stacks.
+
+This tool is useful for debugging why a particular packet (skb) was dropped.
```
yonghong-song, line 43: Is it possible tcp_drop itself (tcpdrop.py) would achieve the same purpose? Any other use cases?

tools/skbstat.py
```diff
@@ -0,0 +1,996 @@
...
+from bcc import BPF
+import argparse
+import ipaddress
+from time import strftime
+from struct import unpack
+from time import sleep
+
+# arguments
+examples = """examples:
+    ./skbstat                               # dump all skb dropping backtraces
```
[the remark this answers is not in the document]
Dongdong-Wang: Ok, I'll fix that error, thx.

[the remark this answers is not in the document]
Dongdong-Wang: `skbstat` is a little bit different with `tcpdrop`. It probe `kfree_skb` function call which leads to packet drops. The packet can be any protocol (TCP/UDP/ICMP...), not just TCP.
There are so many `kfree_skb` routines in kernel, so it may not be easy to tell the user why the packet got lost.  :-(
`skbstat` will provide filter arguments (source/dest IP ...) to let the users see the drop backtrace of the packet which  they really care about.

tools/skbstat.py
```diff
@@ -0,0 +1,996 @@
...
+    ./skbstat --ingress                     # dump skb ingress routines
```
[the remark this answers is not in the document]
Dongdong-Wang: yeah, input is fine. I'll fix it.

tools/skbstat.py
Dongdong-Wang: ok, output is file.
I'll add some example of `skbstat --consume`.

tools/skbstat.py
Dongdong-Wang: I will test carefully with Python3.

tools/skbstat.py
Dongdong-Wang: OK, I'll add some comments here.

tools/skbstat.py
Dongdong-Wang: Yeah. DROP is the default action here by design. In most cases, people may really care about: "why kernel drops my packet?". So `skbstat` dump drops backtrace by default.

tools/skbstat_example.txt
Dongdong-Wang: yeah. input

tools/skbstat_example.txt
Dongdong-Wang: ok, output.

tools/skbstat_example.txt
Dongdong-Wang: the drop of none TCP packet is also hard to debug, and `skbstat` can filter some unrelated/uncared packet for us.

tools/skbstat.py
Dongdong-Wang: OK, I'll add some comments here.

tools/skbstat.py
Dongdong-Wang: I think I can add `--drop` argument, for better output control.

tools/skbstat.py
Dongdong-Wang: Most of the driver code in kernel source tree, they call `napi_gro_receive` instead of `netif_receive_skb`. The backtrace of input routine on my machine showed like this:
```
                ip_protocol_deliver_rcu+0x36
                ip_local_deliver_finish+0x44
                ip_local_deliver+0x6b
                ip_rcv+0x52
                __netif_receive_skb_one_core+0x50
                netif_receive_skb_internal+0x2f
                napi_gro_receive+0x107
```

Dongdong-Wang: Also, code with tracepoint and kprobe are so different and hard to write neatly. So this version of skbstat only contains kprobes, no tracepoints. Maybe next version will solve this issue.

yonghong-song: Another alternative is tracepoint net:napi_gro_receive_entry if you cares about the caller is napi_gro_receive. But not sure whether it covers all cases. Or you could use net:netif_receive_skb directly as the function __netif_receive_skb_core is prone to potential inlining problem.

tools/skbstat.py
```diff
@@ -0,0 +1,1015 @@
...
+	case IPPROTO_UDP:
+		{
+			u16 dst = hdr->tcpudphdr.dst, src = hdr->tcpudphdr.src;
+
+			return port_match(src, dst, filter);
+		}
+	default:
+		break;
+	}
+	return 1;
```
yonghong-song, line 390: So any other protocols will consider as match?

tools/skbstat.py
```diff
@@ -0,0 +1,1015 @@
...
+
+# PROBE LIST
+# the dict key is function to probe, value is tuple.
+# the first entry of tuple is probe function, the second
+# one is the output priority, which can be used by sorter()
+# function. the name of probe function with biggest output 
+# priority number sits at last of output lines.
+
+# NOTE: those kernel functions may be inlined by compiler
+# or just not exists in older kernel. so it may lead to some
```
yonghong-song, line 738: Let us not say "older kernel". It is totally possible that inlining or function renaming may happen in future kernels. Let us just mention "in some kernel versions". The same for below a couple of places.

tools/skbstat_example.txt
```diff
@@ -0,0 +1,88 @@
...
+# skbstat.py -p tcp -P 80 --input --output
+TIME     NUM    FUNC/STACK
+20:04:29        [INGRESS]
+         1      tcp_v4_rcv
+         1      ip_local_deliver_finish
+         1      ip_local_deliver
+         1      ip_route_input_noref
+         1      ip_rcv_finish
+         1      ip_rcv
+         1      netif_receive_skb_internal
```
yonghong-song, line 16: If you can have an example which does not have all 1 numbers (as in the above) and have some explanation, that will be much better.

tools/skbstat_example.txt
```diff
@@ -0,0 +1,88 @@
...
+                ip_local_deliver_finish+0x44
+                ip_local_deliver+0x6b
+                ip_rcv+0x52
+                __netif_receive_skb_one_core+0x50
+                netif_receive_skb_internal+0x2f
+                napi_gro_receive+0x107
+
+[...]
+
+The last column shows the number of function calls or skb dropped stacks.
```
yonghong-song, line 41: You mean the "NUM" column, right?

tools/skbstat.py
```diff
@@ -0,0 +1,1015 @@
+#!/usr/bin/python
+# @lint-avoid-python-3-compatibility-imports
+#
+# skbstat Trace skb events(ingress/egress/consume/drop) of kernel.
+#         For Linux, uses BCC, eBPF. Embedded C.
+#
+# USAGE: skbstat [-h] [-i INTERVAL] [-s SADDR] [-d DADDR] [-N ADDR]
```
yonghong-song, line 7: Could you add a warning on potential performance impact on production system? Similar to tcptop.

tools/skbstat_example.txt
```diff
@@ -0,0 +1,88 @@
...
+
+For example, when you see HTTP requsts arrive at your machine by tcpdump,
+and get lost in kernel. Then you can run `skbstat -p tcp -P 80` to see which
+kernel routine dropped those packets.
+
+USAGE:
+
+usage: skbstat.py [-h] [-i INTERVAL] [-s SADDR] [-d DADDR] [-N ADDR]
+                  [-S SPORT] [-D DPORT] [-P PORT] [-p PROTO] [--input]
+                  [--output] [--consume] [--no_drop]
```
yonghong-song, line 53: Could you provide a man page as well? All tools need to have a man page. Thanks!

tools/skbstat.py
```diff
@@ -0,0 +1,1015 @@
...
+#                  [-S SPORT] [-D DPORT] [-P PORT] [-p PROTO] [--input]
+#                                    [--output] [--consume]
```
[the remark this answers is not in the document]
Dongdong-Wang: ok, I'll fix it.

tools/skbstat.py
```diff
@@ -0,0 +1,1015 @@
...
+
+# arguments
+examples = """examples:
+    skbstat                               # dump kernel skb drop backtraces (default)
+    skbstat --no_drop                     # do not dump kernel skb drop backtraces
+    skbstat --input                       # dump kernel skb input backtraces
+    skbstat --output                      # dump kernel skb output backtraces
+    skbstat --consume                     # dump kernel skb consume routines
+    skbstat -s 10.0.0.1                   # only trace packet comming from 10.0.0.1
+    skbstat -N 10.0.0.1                   # only trace packet comming from or going to 10.0.0.1
```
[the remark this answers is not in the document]
Dongdong-Wang: ok, example of ip6 address will be added here.

tools/skbstat.py
```diff
@@ -0,0 +1,1015 @@
...
+# define BPF program
+bpf_text = """
+#include <uapi/linux/ptrace.h>
+#include <uapi/linux/ip.h>
+#include <uapi/linux/ipv6.h>
+#include <uapi/linux/if_ether.h>
+#include <net/sock.h>
+#include <bcc/proto.h>
+
+BPF_TABLE("percpu_hash", u32, u64, drop_stacks, 256);
```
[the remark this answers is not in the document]
Dongdong-Wang: nice, I did not find `BPF_PERCPU_HASH` before. I'll replace with it.

tools/skbstat.py
```diff
@@ -0,0 +1,1015 @@
...
+from struct import unpack
+from time import sleep
+
+# arguments
+examples = """examples:
+    skbstat                               # dump kernel skb drop backtraces (default)
+    skbstat --no_drop                     # do not dump kernel skb drop backtraces
+    skbstat --input                       # dump kernel skb input backtraces
+    skbstat --output                      # dump kernel skb output backtraces
+    skbstat --consume                     # dump kernel skb consume routines
```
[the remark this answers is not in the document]
Dongdong-Wang: It seems the the `stack_id` did not exist in the `stack_traces`. I add a {try: except: } block here to catch the `KeyError` exception and print some error messages.

tools/skbstat.py
```diff
@@ -0,0 +1,1015 @@
...
+static inline int port_match1(u16 sport, u16 dport, struct filter *filter)
+{
+	return port_range_match(sport, filter->sport_range) ||
+	    port_range_match(dport, filter->sport_range);
+}
+
+static inline int port_match(u16 sport, u16 dport, struct filter *filter)
+{
+	return (filter->flags & SKB_PORT_MATCH_1) ? port_match1(sport, dport,
+								filter) :
```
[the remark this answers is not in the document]
Dongdong-Wang: I formated the C code by kernel tools `Lindent.sh`. Looks like not so tidy. I'll change this by my own hands.

tools/skbstat.py
Dongdong-Wang: Yeah, make sense. not just older kernel.

tools/skbstat_example.txt
Dongdong-Wang: I'll fix this.

tools/skbstat_example.txt
Dongdong-Wang: Oh, sorry, I mean the "NUM" column.

tools/skbstat.py
Dongdong-Wang: Sure. In fact, this tool will actually add some overhead. I'll add a warning.

tools/skbstat_example.txt
Dongdong-Wang: Ok, no problem.

tools/skbstat.py
```diff
@@ -0,0 +1,1015 @@
...
+	    port_match2(sport, dport, filter);
+}
+
+static inline int proto_match(u8 proto, struct filter *filter)
+{
+	return (filter->proto && proto) ? filter->proto == proto : 1;
+}
+
+static inline int ip6_addr_match2(unsigned __int128 saddr,
+				  unsigned __int128 daddr,
```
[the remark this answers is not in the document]
Dongdong-Wang: OK.

tools/skbstat.py
Dongdong-Wang: `skb_l4_match` will check the detail of L4 protocol, such as TCP port, ICMP id. Protocols as ICMP will just return 1 and skip the L4 check currently. Protocol number will be checked in `proto_match` function.

tools/skbstat.py
Dongdong-Wang: yes, tracepoint is always stable. I'll try to implement with tracepoint.

tools/skbstat.py
```diff
@@ -0,0 +1,1079 @@
...
+    skbstat -p tcp -D 80                  # only trace input HTTP packet
+    skbstat -p tcp -P 80                  # only trace input or output HTTP packet
+"""
+parser = argparse.ArgumentParser(
+    description="""
+    Trace skb event (input/output/consume/drop) of the kernel.
+
+    WARNING: When run with `--input/output/consume` options, 
+    this tool will trace high traffic kernel functions and may lead to 
+    some noticeable overhead. Use it carefully on production system.
```
yonghong-song, line 48: There is no need to put a warning description here. Just put in the beginning of the file around line 13.

tools/skbstat.py
```diff
@@ -0,0 +1,1079 @@
...
+#include <uapi/linux/ip.h>
+#include <uapi/linux/ipv6.h>
+#include <uapi/linux/if_ether.h>
+#include <net/sock.h>
+#include <bcc/proto.h>
+
+BPF_TABLE("percpu_hash", u32, u64, drop_stacks, 256);
+BPF_TABLE("percpu_hash", u32, u64, consume_stacks, 256);
+BPF_TABLE("percpu_hash", u64, u64, input, 256);
+BPF_TABLE("percpu_hash", u64, u64, forward, 256);
```
yonghong-song, line 220: You forgot to use BPF_PERCPU_HASH, since another revision is needed, suggest you use BPF_PERCPU_HASH since it is simpler.

tools/skbstat.py
```diff
@@ -0,0 +1,1079 @@
...
+
+static inline int skb_l4hdr(struct sk_buff *skb, union l4hdr *l4h,
+			    int l3_offset)
+{
+	// transport header was set
+	if (skb->transport_header != (typeof(skb->transport_header)) ~ 0U &&
+	    skb->transport_header != skb->network_header) {
+		if (bpf_probe_read_kernel
+		    (l4h, sizeof(union l4hdr),
+		     skb->head + skb->transport_header)) {
```
yonghong-song, line 404: The code like the above 3 lines are hard to read. For C code here, I think we do not need adhere to 80 char's per line. 100 char's per line is perfect fine.

tools/skbstat.py
```diff
@@ -0,0 +1,1079 @@
...
+
+	return 0;
+}
+
+static inline int __trace_sk_input(struct pt_regs *ctx, struct sock *sk)
+{
+	union filter_args args = { };
+
+	args.input = true;
+
```
yonghong-song, line 627: We can save a line here. Otherwise, the code looks too sparse. The same for other functions below.

tools/skbstat_example.txt
```diff
@@ -0,0 +1,103 @@
...
+The "FUNC/STACK" column shows either the function name or function backtrace.
+
+This tool is useful for debugging why a particular packet (skb) was dropped.
+
+For example, when you see HTTP requsts arrive at your machine by tcpdump,
+and get lost in kernel. Then you can run `skbstat -p tcp -P 80` to see which
+kernel routine dropped those packets.
+
+WARNING: When run with `--input/output/consume` options,
+this tool will trace high traffic kernel functions and may lead to
```
yonghong-song, line 56: The example file is used to show examples. Let us remove the WARNING here.

tools/skbstat_example.txt
```diff
@@ -0,0 +1,103 @@
...
+
+USAGE:
+
+usage: skbstat.py [-h] [-i INTERVAL] [-s SADDR] [-d DADDR] [-N ADDR]
+                  [-S SPORT] [-D DPORT] [-P PORT] [-p PROTO] [--input]
+                  [--output] [--consume] [--no_drop]
+
+    Trace skb event (input/output/consume/drop) of the kernel.
+
+    WARNING: When run with `--input/output/consume` options,
```
yonghong-song, line 67: The same here, let us remove WARNING here.

yonghong-song (commented):
Thanks, much better now. Just added some minor comments.

brendangregg:
I've run this and I'm a bit confused: the tool's documentation talks about identifying dropped packets, and I see this:

```
# skbstat.py
TIME     NUM    FUNC/STACK
21:17:04        [DROP]
         1     
                kfree_skb+0x1
                inet_csk_destroy_sock+0x64
                tcp_done+0xbc
                tcp_time_wait+0x1a2
                tcp_rcv_state_process+0x379
                tcp_v4_do_rcv+0xc4
                tcp_v4_rcv+0xcef
                ip_protocol_deliver_rcu+0x30
                ip_local_deliver_finish+0x48
                ip_local_deliver+0x73
                ip_rcv_finish+0x87
                ip_rcv+0xbc
                __netif_receive_skb_core+0x817
                __netif_receive_skb_list_core+0x102
                netif_receive_skb_list_internal+0x1a1
                netif_receive_skb_list+0x2a
                iwl_pcie_rx_handle+0x4d0
                iwl_pcie_irq_rx_msix_handler+0x58
                irq_thread_fn+0x28
                irq_thread+0xda
                kthread+0x114
                ret_from_fork+0x1f
         1     
                kfree_skb+0x1
                tcp_v4_rcv+0xcef
                ip_protocol_deliver_rcu+0x30
                ip_local_deliver_finish+0x48
                ip_local_deliver+0x73
                ip_rcv_finish+0x87
                ip_rcv+0xbc
                __netif_receive_skb_core+0x817
                __netif_receive_skb_list_core+0x102
                netif_receive_skb_list_internal+0x1a1
                gro_normal_list.part.0+0x1e
                gro_normal_one+0x49
                napi_gro_complete+0x6e
                dev_gro_receive+0x346
                napi_gro_receive+0x39
                ieee80211_invoke_fast_rx+0x387
                ieee80211_prepare_and_rx_handle+0x3f
                ieee80211_rx_napi+0x28a
                iwl_mvm_rx_mpdu_mq+0xc2e
                iwl_mvm_rx_mq_rss+0x54
                iwl_pcie_rx_handle_rb.constprop.0+0xee
                iwl_pcie_rx_handle+0x1db
                iwl_pcie_irq_rx_msix_handler+0x58
                irq_thread_fn+0x28
                irq_thread+0xda
                kthread+0x114
                ret_from_fork+0x1f
         15    
                kfree_skb+0x1
                ip_protocol_deliver_rcu+0x30
                ip_local_deliver_finish+0x48
                ip_local_deliver+0x73
                ip_rcv_finish+0x87
                ip_rcv+0xbc
                __netif_receive_skb_core+0x817
                __netif_receive_skb_list_core+0x102
                netif_receive_skb_list_internal+0x1a1
                gro_normal_list.part.0+0x1e
                gro_normal_one+0x49
                napi_gro_complete+0x6e
                dev_gro_receive+0x346
                napi_gro_receive+0x39
                ieee80211_invoke_fast_rx+0x387
                ieee80211_prepare_and_rx_handle+0x3f
                ieee80211_rx_napi+0x28a
                iwl_mvm_rx_mpdu_mq+0xc2e
                iwl_mvm_rx_mq_rss+0x54
                iwl_pcie_rx_handle_rb.constprop.0+0xee
                iwl_pcie_rx_handle+0x1db
                iwl_pcie_irq_rx_msix_handler+0x58
                irq_thread_fn+0x28
                irq_thread+0xda
                kthread+0x114
                ret_from_fork+0x1f
         25    
                kfree_skb+0x1
                ip_protocol_deliver_rcu+0x30
                ip_local_deliver_finish+0x48
                ip_local_deliver+0x73
                ip_rcv_finish+0x87
                ip_rcv+0xbc
                __netif_receive_skb_core+0x817
                __netif_receive_skb_list_core+0x102
                netif_receive_skb_list_internal+0x1a1
                netif_receive_skb_list+0x2a
                iwl_pcie_rx_handle+0x4d0
                iwl_pcie_irq_rx_msix_handler+0x58
                irq_thread_fn+0x28
                irq_thread+0xda
                kthread+0x114
                ret_from_fork+0x1f
         33    
                kfree_skb+0x1
                skb_release_all+0x26
                __kfree_skb+0x13
                tcp_recvmsg+0x8ba
                inet_recvmsg+0x5e
                sock_recvmsg+0x69
                __sys_recvfrom+0xe5
                __x64_sys_recvfrom+0x29
                do_syscall_64+0x49
                entry_SYSCALL_64_after_hwframe+0x44
```

But my system is not currently dropping packets that I know of (or that I can see in `netstat`/`nstat`). What's it showing?

I can guess that it's showing some normal packet free paths. Not all: my system is doing a higher packet rate than shown here.

What exactly is it showing?

I think a tool that showed actual drop paths (such as IpOutDiscards, etc.) only, along with the port and protocol filtering it does, would be a useful debug aid. Then all the output is easy to interpret: anything shown is a problem. But this tool isn't it.

There's also a lot of code to parse ports and addresses, but this isn't shown in the output, which is a pity. Right now it's a >1000 line tool -- which would be the largest tool in bcc -- but it doesn't have that much outward functionality. This will be a burden to maintain (again, it will be the biggest tool in the entire repository), and for something I'm not even sure what the output means, I'm not sure it belongs.

tools/skbstat.py
Dongdong-Wang: agree.

tools/skbstat.py
Dongdong-Wang: OK, will change it.

tools/skbstat.py
Dongdong-Wang: Emm, you are right.

tools/skbstat_example.txt
Dongdong-Wang: ok

tools/skbstat_example.txt
Dongdong-Wang: ok

tools/skbstat.py
Dongdong-Wang: I have concern with older bcc version which has no BPF_PERCPU_HASH. This tool will not work when copied to some machine which only has old bcc version.

yonghong-song: okay, I won't insist on this one.

Dongdong-Wang:
> I've run this and I'm a bit confused: the tool's documentation talks about identifying dropped packets, and I see this:
> 
> ```
> # skbstat.py
> TIME     NUM    FUNC/STACK
> 21:17:04        [DROP]
>          1     
>                 kfree_skb+0x1
>                 inet_csk_destroy_sock+0x64
>                 ...
> ```
> 
> But my system is not currently dropping packets that I know of (or that I can see in `netstat`/`nstat`). What's it showing?
In most cases, kernel drops a packet (skb) by calling `kfree_skb`. But there are also normal packet free paths using the `kfree_skb` function.
> 
> I can guess that it's showing some normal packet free paths. Not all: my system is doing a higher packet rate than shown here.
> 
> What exactly is it showing?
> 
> I think a tool that showed actual drop paths (such as IpOutDiscards, etc.) only, along with the port and protocol filtering it does, would be a useful debug aid. Then all the output is easy to interpret: anything shown is a problem. But this tool isn't it.
> 
> There's also a lot of code to parse ports and addresses, but this isn't shown in the output, which is a pity. Right now it's a >1000 line tool -- which would be the largest tool in bcc -- but it doesn't have that much outward functionality. This will be a burden to maintain (again, it will be the biggest tool in the entire repository), and for something I'm not even sure what the output means, I'm not sure it belongs.

brendangregg:
This just doesn't seem useful enough for inclusion.

As a tool to investigate packet drops: we shouldn't be making the end user pick through network stack traces to identify what's abnormal if it can be avoided. And it can. I'd have made the tool only print packet drops (filter normal paths), and print the SNMP counter name for the drop followed by the packet details: IP addresses and ports and stack trace, at a minimum. The tool doesn't do that.

As a tool for just looking at the kfree_skb() function: well, I'd rename it to "skbfree" for a start, and in the documentation make it clear it's for advanced users only who are already familiar with the network stack. The output should really include more useful details. You'd end up with something similar to tcpdrop(8), although 5x the number of lines of code. And in hindsight, I've only used tcpdrop(8) a few times since adding it; if we were to reduce the number of tools (too many tools is a burden for new users), ditching tcpdrop(8) would be high on my list. As would an skbfree(8) tool.

Other tracers (I won't name here) accepted every tool people wrote, and ended up with an unmaintained broken mess. I'm trying not to let that happen here. Tools must be obviously useful, and short enough for us to maintain.

I'd put this on your own github repository and develop it over time and see where it goes. Post some real world use cases. Show some amazing wins that only this tool can do. This would make a convincing case for BCC inclusion. But that doesn't have to be the objective either. Really, every tool people write does not have to live here.

Dongdong-Wang:
> This just doesn't seem useful enough for inclusion.
> 
> As a tool to investigate packet drops: we shouldn't be making the end user pick through network stack traces to identify what's abnormal if it can be avoided. And it can. I'd have made the tool only print packet drops (filter normal paths), and print the SNMP counter name for the drop followed by the packet details: IP addresses and ports and stack trace, at a minimum. The tool doesn't do that.
> 
I agree with you. The drop backtraces are not so accurate. I'll try to filter normal paths and append SNMP counter and IP address and ports to them.
> As a tool for just looking at the kfree_skb() function: well, I'd rename it to "skbfree" for a start, and in the documentation make it clear it's for advanced users only who are already familiar with the network stack. The output should really include more useful details. You'd end up with something similar to tcpdrop(8), although 5x the number of lines of code. And in hindsight, I've only used tcpdrop(8) a few times since adding it; if we were to reduce the number of tools (too many tools is a burden for new users), ditching tcpdrop(8) would be high on my list. As would an skbfree(8) tool.
> 
`tcpdrop` can not show the UDP/ICMP loss backtraces.
> Other tracers (I won't name here) accepted every tool people wrote, and ended up with an unmaintained broken mess. I'm trying not to let that happen here. Tools must be obviously useful, and short enough for us to maintain.
> 
I understand and agree with you. 
> I'd put this on your own github repository and develop it over time and see where it goes. Post some real world use cases. Show some amazing wins that only this tool can do. This would make a convincing case for BCC inclusion. But that doesn't have to be the objective either. Really, every tool people write does not have to live here.
In fact, this tool is really helpful to those kernel network developers. They developed a new feature and may cause the packet loss and did not know why. So with this tool, they could pass the exact IP addresses and ports and wait for the backtraces of `kfree_skb`. I will append some real use cases in the example file.
This tool is not so good for now. I'll improve it. Thanks for your reply and suggestion. When it seems good and useful, I'll create a PR again.

PR #3307, opened 2021-03-11
tcpstates: forget sockets when connection is closed

jeromemarchand:
The adress of struct sock, which are used as the map key, are often
reused. When it happens random duration appears in the MS field for
new connections (CLOSE->SYN_SENT and LISTEN->SYN_RECV
transitions). Let's forget about the socket when the connection is
closed.


yonghong-song:
[buildbot, test this please]

yonghong-song:
Make sense. Thanks for the fix!

PR #3309, opened 2021-03-11
libbpf-tools: add ext4dist

ethercflow:
```sh
➜  libbpf-tools git:(ext4dist) ✗ sudo ./ext4dist  1 5
Tracing ext4 operation latency... Hit Ctrl-C to end.


operation = 'read_iter'
         0 -> 1          : 257      |****************************************|
         2 -> 3          : 12       |*                                       |
         4 -> 7          : 2        |                                        |

operation = 'write_iter'
         0 -> 1          : 79       |***                                     |
         2 -> 3          : 831      |****************************************|
         4 -> 7          : 543      |**************************              |
         8 -> 15         : 105      |*****                                   |
        16 -> 31         : 5        |                                        |

operation = 'open'
         0 -> 1          : 997      |****************************************|
         2 -> 3          : 2        |                                        |
         4 -> 7          : 1        |                                        |


operation = 'read_iter'
         0 -> 1          : 1813     |****************************************|
         2 -> 3          : 102      |**                                      |
         4 -> 7          : 14       |                                        |
```

### DIFF
--- a/libbpf-tools/ext4dist.bpf.c
+++ b/libbpf-tools/ext4dist.bpf.c
@@ -1,0 +1,165 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2021 Wenbo Zhang
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_tracing.h>
+#include "bits.bpf.h"
+#include "ext4dist.h"
+
+const volatile bool targ_ms = false;
+const volatile pid_t targ_tgid = 0;
+
+#define MAX_ENTRIES	10240
+
+struct hist hists[__MAX_FOP_TYPE];
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_ENTRIES);
+	__type(key, u32);
+	__type(value, u64);
+} starts SEC(".maps");
+
+static int trace_entry(void)
+{
+	u64 id = bpf_get_current_pid_tgid();
+	u32 tgid = id >> 32;
+	u32 pid = id;
+	u64 ts;
+
+	if (targ_tgid && targ_tgid != tgid)
+		return 0;
+	ts = bpf_ktime_get_ns();
+	bpf_map_update_elem(&starts, &pid, &ts, BPF_ANY);
+
+	return 0;
+}
+
+SEC("kprobe/ext4_file_read_iter")
+int BPF_KPROBE(kprobe1)
+{
+	return trace_entry();
+}
+
+SEC("kprobe/ext4_file_write_iter")
+int BPF_KPROBE(kprobe2)
+{
+	return trace_entry();
+}
+
+SEC("kprobe/ext4_file_open")
+int BPF_KPROBE(kprobe3)
+{
+	return trace_entry();
+}
+
+SEC("kprobe/ext4_sync_file")
+int BPF_KPROBE(kprobe4)
+{
+	return trace_entry();
+}
+
+SEC("fentry/ext4_file_read_iter")
+int BPF_PROG(fentry1)
+{
+	return trace_entry();
+}
+
+SEC("fentry/ext4_file_write_iter")
+int BPF_PROG(fentry2)
+{
+	return trace_entry();
+}
+
+SEC("fentry/ext4_file_open")
+int BPF_PROG(fentry3)
+{
+	return trace_entry();
+}
+
+SEC("fentry/ext4_sync_file")
+int BPF_PROG(fentry4)
+{
+	return trace_entry();
+}
+
+static int trace_return(enum ext4_fop_type type)
+{
+	u64 *tsp, slot, ts = bpf_ktime_get_ns();
+	u64 id = bpf_get_current_pid_tgid();
+	u32 pid = id;
+	s64 delta;
+
+	tsp = bpf_map_lookup_elem(&starts, &pid);
+	if (!tsp)
+		return 0;
+	delta = (s64)(ts - *tsp);
+	if (delta < 0)
+		goto cleanup;
+
+	if (targ_ms)
+		delta /= 1000000U;
+	else
+		delta /= 1000U;
+	slot = log2l(delta);
+	if (slot >= MAX_SLOTS)
+		slot = MAX_SLOTS - 1;
+	if (type >= __MAX_FOP_TYPE)
+		goto cleanup;
+	__sync_fetch_and_add(&hists[type].slots[slot], 1);
+
+cleanup:
+	bpf_map_delete_elem(&starts, &pid);
+	return 0;
+}
+
+SEC("kretprobe/ext4_file_read_iter")
+int BPF_KRETPROBE(kretprobe1)
+{
+	return trace_return(READ_ITER);
+}
+
+SEC("kretprobe/ext4_file_write_iter")
+int BPF_KRETPROBE(kretprobe2)
+{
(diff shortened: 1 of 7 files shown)


libbpf-tools/ext4dist.h
```diff
@@ -0,0 +1,19 @@
...
+	__MAX_FOP_TYPE,
+};
+
+#define MAX_SLOTS	27
+
+struct hist {
+	__u32 slots[MAX_SLOTS];
+};
+
+#endif // __EXT4DIST_H
```
anakryiko, line 19: please don't use C++ comments, should be `/* __EXT4DIST_H */`

libbpf-tools/ext4dist.bpf.c
```diff
@@ -0,0 +1,117 @@
...
+		goto cleanup;
+	__sync_fetch_and_add(&hists[type].slots[slot], 1);
+
+cleanup:
+	bpf_map_delete_elem(&starts, &pid);
+	return 0;
+}
+
+SEC("kretprobe/ext4_file_read_iter")
+int BPF_KRETPROBE(ext4_file_read_iter_ret)
```
anakryiko, line 94: how about using fentry/fexit? even if they are in modules, they should work on latest kernel. Would be actually nice to try it out on real tool, btw ;)

anakryiko (approved):
So looks good, please fix the comment.

But consider trying fentry/fexit. At least would be great if you can test it with ext4 in a module on latest released kernels.

It might be a good time to add some helpers to detect kernel support for module BTFs, and choose between fentry/fexit and kprobe/kretprobe accordingly. WDYT?

libbpf-tools/ext4dist.h
ethercflow: Oh, I forgot to change it, I am going to modify the auto-generated template

ethercflow:
> So looks good, please fix the comment.
> 
> But consider trying fentry/fexit. At least would be great if you can test it with ext4 in a module on latest released kernels.
> 
> It might be a good time to add some helpers to detect kernel support for module BTFs, and choose between fentry/fexit and kprobe/kretprobe accordingly. WDYT?

Agree, I will change to this

yonghong-song:
[buildbot, test this please]

ethercflow:

Hi, @yonghong-song  Please do not merge this for now, I am adding fentry/fexit and fallback into kprobe logic, thanks!

yonghong-song:
okay.

ethercflow:
> So looks good, please fix the comment.
> 
> But consider trying fentry/fexit. At least would be great if you can test it with ext4 in a module on latest released kernels.
> 
> It might be a good time to add some helpers to detect kernel support for module BTFs, and choose between fentry/fexit and kprobe/kretprobe accordingly. WDYT?

Hi @anakryiko,  I tested with the module BTF on 5.12 rc2+, it works very well. I add two helpers to make the choose.  Does it make sense?

libbpf-tools/ext4dist.c
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	return 0;
+}
+
+static bool should_fallback(void)
+{
+	if (!is_kernel_module("ext4"))
+		return false;
+	return !is_kernel_support_btf_modules();
```
anakryiko, line 149: this is way too complicated and unreliable, I think. E.g., relying on /boot/config or /proc/kconfig.gz is making unnecessary assumptions.

Just check that /sys/kernel/btf/ext4 exists, that's it. If it does, kernel definitely supports module BTF for ext4 module. If not, ext4 module might not be loaded or module BTFs are not supported. So just attempt to attach kprobe/kretprobe, and if that fails -- report failure. If not, good, you've got everything working on old kernel.

anakryiko (changes requested):

libbpf-tools/ext4dist.c
ethercflow: Fixed, thanks!

libbpf-tools/ext4dist.c
```diff
@@ -0,0 +1,252 @@
...
+		print_log2_hist(hist.slots, MAX_SLOTS, units);
+		printf("\n");
+	}
+
+	return 0;
+}
+
+static bool should_fallback(void)
+{
+	if (is_kernel_module("ext4") && !access("/sys/kernel/btf/ext4", R_OK))
```
anakryiko, line 147: you don't need to check is_kernel_module(), if /sys/kernel/btf/ext4 is missing, then it's either not loaded or module BTF is not supported. So it's sufficient to check /sys/kernel/btf/ext4.

BTW, please leave a comment here on what you are doing and why.

libbpf-tools/ext4dist.c
```diff
@@ -0,0 +1,252 @@
...
+		return true;
+	return false;
+}
+
+static void ext4dist__choose_progs(struct ext4dist_bpf *skel, bool fallback)
```
anakryiko, line 152: maybe just inline this, this function doesn't serve much purpose beyond distracting from linearly following the code in main()

anakryiko (changes requested):

libbpf-tools/ext4dist.c
ethercflow: > you don't need to check is_kernel_module(), if /sys/kernel/btf/ext4 is missing, then it's either not loaded or module BTF is not supported. So it's sufficient to check /sys/kernel/btf/ext4.

Oh, the reason I checked is that I found that EXT4 is compiled into the kernel in some environments. eg:  VMware + Ubuntu 20.10.

> BTW, please leave a comment here on what you are doing and why.
Sure

libbpf-tools/ext4dist.c
ethercflow: Thanks for your suggestion, I will fix it

libbpf-tools/ext4dist.c
ethercflow: Hi @anakryiko   I found 5.12 rc2+  with `make defconfig` , 

```
➜  linux-5.12-rc2 cat .config | grep CONFIG_EXT4_FS
CONFIG_EXT4_FS=y
CONFIG_EXT4_FS_POSIX_ACL=y
CONFIG_EXT4_FS_SECURITY=y
```

EXT4 is compiled into kernel by default. So I think we need `is_kernel_module` to handle this?

anakryiko: Oh, you are right. If it's compiled in you'll want to use fentry, of course, right. So never mind this.

libbpf-tools/ext4dist.c
```diff
@@ -0,0 +1,241 @@
...
+
+static void ext4dist__choose_progs(struct bpf_object *obj, bool fallback)
+{
+	const char func_prefix = fallback ? 'k' : 'f';
+	struct bpf_program *prog;
+
+	bpf_object__for_each_program(prog, obj) {
+		if (bpf_program__name(prog)[0] != func_prefix) {
+			bpf_program__set_autoload(prog, false);
+		}
```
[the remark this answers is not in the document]
ethercflow: It has been fixed, I found that I always used loops instead of repeating more than 3, but ignored the readability. Thanks for pointing this out.

libbpf-tools/ext4dist.c
ethercflow: Fixed, thanks!

libbpf-tools/ext4dist.c
ethercflow: Added commit, thanks!

anakryiko (approved):
Awesome, thanks!

yonghong-song:
[buildbot, test this please]

PR #3315, opened 2021-03-14
In GCC10.2 suffix '.isra.0' was appended to 'finish_task_switch'

docularxu:
When buildiing kernel with GCC10 [2] in Debian on an Arm64 machine, it was
found the new "inter-procedural optimization improvements" [1] makes symbol
name 'finish_task_switch' changed to 'finish_task_switch.isra.0'.

Details:
The results, when built with gcc 9.3:
nm ../linux.buildout/kernel/sched/core.o | grep finish_task_switch
0000000000001288 t finish_task_switch

However, when built with gcc 10.2:
nm ../linux.buildout/kernel/sched/core.o | grep finish_task_switch
00000000000012d0 t finish_task_switch.isra.0

The same symbols (with xxx.isra.0 or without, respectively of course) also
appear in final file 'System.map' and in '/proc/kallsyms' when booting. This
negatively impact the tracing tools commonly used in kernel debugging, such
as bcc tools offcputime and runqlat. They hardcode 'finish_task_switch'
(without the .isra.0 suffix) into their scripts.

This patch fix the issue by changing the hardcoded 'finish_task_switch' string
to a python regular expression pattern who can match both the traditional form
'finish_task_switch' and the new gcc10 form 'finish_task_switch.isra.0'
(with single digit at the end). attach_kprobe()'s input parameter 'event_re'
is used for this type of pattern matching.

[1] https://gcc.gnu.org/gcc-10/changes.html
[2] ARCH=arm64 make Image

Signed-off-by: Guodong Xu <guodong.xu@linaro.org>

### DIFF
--- a/examples/tracing/task_switch.py
+++ b/examples/tracing/task_switch.py
@@ -6,7 +6,8 @@ from bcc import BPF
 from time import sleep
 
 b = BPF(src_file="task_switch.c")
-b.attach_kprobe(event="finish_task_switch", fn_name="count_sched")
+b.attach_kprobe(event_re="^finish_task_switch$|^finish_task_switch\.isra\.\d$",
+                fn_name="count_sched")
 
 # generate many schedule events
 for i in range(0, 100): sleep(0.01)
--- a/tests/python/test_trace2.py
+++ b/tests/python/test_trace2.py
@@ -31,7 +31,8 @@ class TestTracingEvent(TestCase):
     def setUp(self):
         b = BPF(text=text, debug=0)
         self.stats = b.get_table("stats")
-        b.attach_kprobe(event="finish_task_switch", fn_name="count_sched")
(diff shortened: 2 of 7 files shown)


docularxu:
I tested building with docker on arm64. 
```
$ ./scripts/docker/build.sh  iovisor/bcc.git  v0.18.0 18.04 ubuntu
```
, and had a test run with docker image,
```$ docker run -it --rm \
  --privileged \
  -v /lib/modules:/lib/modules:ro \
  -v /usr/src:/usr/src:ro \
  -v /etc/localtime:/etc/localtime:ro \
  -v /sys/kernel/debug:/sys/kernel/debug:rw \
  --workdir /usr/share/bcc/tools \
  iovisor/bcc.git:v0.18.0```

I checked the tools with both gcc9-built kernel and gcc10-built kernel. All can work as expected.

yonghong-song:
[buildbot, test this please]

PR #3317, opened 2021-03-15
[libbpf-tools]block tracepoints no longer have struct request_queue arg

rsudarikov:
stabilizing llibbpf-tools/biolatency since TP_PROTO & TP_ARGS for block_rq_insert and block_rq_issue tracepoints has been changed; 
see https://github.com/torvalds/linux/commit/a54895fa changes

### DIFF
--- a/libbpf-tools/biolatency.bpf.c
+++ b/libbpf-tools/biolatency.bpf.c
@@ -10,6 +10,10 @@
 
 #define MAX_ENTRIES	10240
 
+#define KERNEL_VERSION(a, b, c) (((a) << 16) + ((b) << 8) + (c))
+
+extern int LINUX_KERNEL_VERSION __kconfig;
+
 const volatile bool targ_per_disk = false;
 const volatile bool targ_per_flag = false;
 const volatile bool targ_queued = false;
@@ -53,34 +57,30 @@ int trace_rq_start(struct request *rq)
 }
 
 SEC("tp_btf/block_rq_insert")
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 11, 1)
-int BPF_PROG(block_rq_insert, struct request *rq)
+int block_rq_insert (u64 *ctx)
 {
+	/**
+	 * commit a54895fa (v5.11-rc1) changed tracepoint argument list
+	 * from TP_PROTO(struct request_queue *q, struct request *rq)
+	 * to TP_PROTO(struct request *rq)
+	 */
+	struct request *rq = LINUX_KERNEL_VERSION >= KERNEL_VERSION(5, 11, 1) ? (void *)ctx[0] : (void *)ctx[1];
 	return trace_rq_start(rq);
 }
-#else
-int BPF_PROG(block_rq_insert, struct request_queue *q, struct request *rq)
-{
-	return trace_rq_start(rq);
-}
-#endif
 
 SEC("tp_btf/block_rq_issue")
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 11, 1)
-int BPF_PROG(block_rq_issue, struct request *rq)
+int block_rq_issue (u64 *ctx)
 {
+	/**
+	 * commit a54895fa (v5.11-rc1) changed tracepoint argument list
+	 * from TP_PROTO(struct request_queue *q, struct request *rq)
+	 * to TP_PROTO(struct request *rq)
+	 */
+	struct request *rq = LINUX_KERNEL_VERSION >= KERNEL_VERSION(5, 11, 1) ? (void *)ctx[0] : (void *)ctx[1];
 	if (targ_queued && BPF_CORE_READ(rq->q, elevator))
 		return 0;
 	return trace_rq_start(rq);
 }
-#else
-int BPF_PROG(block_rq_issue, struct request_queue *q, struct request *rq)
-{
-	if (targ_queued && BPF_CORE_READ(q, elevator))
-		return 0;
-	return trace_rq_start(rq);
-}
-#endif
 
 SEC("tp_btf/block_rq_complete")
 int BPF_PROG(block_rq_complete, struct request *rq, int error,
--- a/libbpf-tools/biolatency.bpf.c
+++ b/libbpf-tools/biolatency.bpf.c
@@ -57,29 +57,42 @@ int trace_rq_start(struct request *rq)
 }
 
 SEC("tp_btf/block_rq_insert")
-int block_rq_insert (u64 *ctx)
+int block_rq_insert(u64 *ctx)
 {
 	/**
 	 * commit a54895fa (v5.11-rc1) changed tracepoint argument list
 	 * from TP_PROTO(struct request_queue *q, struct request *rq)
 	 * to TP_PROTO(struct request *rq)
 	 */
-	struct request *rq = LINUX_KERNEL_VERSION >= KERNEL_VERSION(5, 11, 1) ? (void *)ctx[0] : (void *)ctx[1];
-	return trace_rq_start(rq);
+	if (LINUX_KERNEL_VERSION <= KERNEL_VERSION(5, 10, 0))
+		return trace_rq_start((void *)ctx[1]);
+	else
+		return trace_rq_start((void *)ctx[0]);
+}
+
+static __always_inline
+int f(u64 *ctx, int idx)
+{
+	/**
+	 * Verifier happens to be more friendly to pointer arithmetic starting v5.11
+	 */
+	if (targ_queued && BPF_CORE_READ(((struct request *)ctx[idx])->q, elevator))
+		return 0;
+	return trace_rq_start((void *)ctx[idx]);
 }
 
 SEC("tp_btf/block_rq_issue")
-int block_rq_issue (u64 *ctx)
+int block_rq_issue(u64 *ctx)
 {
 	/**
 	 * commit a54895fa (v5.11-rc1) changed tracepoint argument list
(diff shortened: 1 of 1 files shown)


libbpf-tools/biolatency.bpf.c
```diff
@@ -52,18 +53,34 @@ int trace_rq_start(struct request *rq)
 }
 
 SEC("tp_btf/block_rq_insert")
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 11, 1)
```
anakryiko, line 56: this can't be done at the compilation time, because Linux version on which you compile BPF code has nothing to do with the kernel version that it will be run on. There are two possible approaches here.

First, is to do what you did here (two sets of BPF programs), but without any compile-time `#if`/`#else`. Then from user-space side detect which set of BPF programs needs to be used based on actual running kernel support. And then bpf_program__set_autoload(false) for the BPF programs that won't work on the kernel. This is most flexible approach, but requires quite a bit more code.

Alternatively, libbpf supports .kconfig externs, so one can do kernel version check in runtime in BPF code. So the solution would be like this:

```
#define KERNEL_VERSION(a, b, c) (((a) << 16) + ((b) << 8) + (c))

extern int LINUX_KERNEL_VERSION __kconfig;

SEC("tp_btf/block_rq_insert")
int block_rq_insert(u64 *ctx)
{
    /* comment here pointing to the commit that changed tracepoint argument list */
    struct request *rq = LINUX_KERNEL_VERSION >= KERNEL_VERSION(5, 11, 1) ? ctx[1] : ctx[0];

    /* the rest is untouched */
}
```

libbpf will substitute actual runtime kernel version into LINUX_KERNEL_VERSION extern.

Making such decisions based on kernel version is a bit bad, because there are kernels that get a lot of patches backported, but kernels themselves stay on old kernel version. But given the minimal amount of code, it might be a good start (and a first demo of .kconfig externs in libbpf-tools ;) ).

rsudarikov:
@anakryiko please review

libbpf-tools/biolatency.bpf.c
```diff
@@ -6,9 +6,14 @@
 #include <bpf/bpf_tracing.h>
 #include "biolatency.h"
 #include "bits.bpf.h"
+#include <linux/version.h>
```
anakryiko, line 9: please remove this include, we don't need it, do we?

libbpf-tools/biolatency.bpf.c
```diff
@@ -52,15 +57,27 @@ int trace_rq_start(struct request *rq)
 }
 
 SEC("tp_btf/block_rq_insert")
-int BPF_PROG(block_rq_insert, struct request_queue *q, struct request *rq)
+int block_rq_insert (u64 *ctx)
```
anakryiko, line 60: extra space after function name

libbpf-tools/biolatency.bpf.c
```diff
@@ -52,15 +57,27 @@ int trace_rq_start(struct request *rq)
...
 {
+	/**
+	 * commit a54895fa (v5.11-rc1) changed tracepoint argument list
+	 * from TP_PROTO(struct request_queue *q, struct request *rq)
+	 * to TP_PROTO(struct request *rq)
+	 */
+	struct request *rq = LINUX_KERNEL_VERSION >= KERNEL_VERSION(5, 11, 1) ? (void *)ctx[0] : (void *)ctx[1];
```
anakryiko, line 67: it's a very long line, can you wrap it? Or maybe cleaner is to do it with if/else after variable declaration.

plus also empty line between variable declarations and the rest of function's code

libbpf-tools/biolatency.bpf.c
```diff
@@ -52,15 +57,27 @@ int trace_rq_start(struct request *rq)
...
 	return trace_rq_start(rq);
 }
 
 SEC("tp_btf/block_rq_issue")
-int BPF_PROG(block_rq_issue, struct request_queue *q, struct request *rq)
+int block_rq_issue (u64 *ctx)
```
anakryiko, line 72: same here with extra space

anakryiko (commented):
few styling nits. Did you test this approach on both old and new kernels?

libbpf-tools/biolatency.bpf.c
chenhengqi: Remove this line got the following warining:
```
warning: implicit declaration of function 'KERNEL_VERSION' is invalid in C99 [-Wimplicit-function-declaration]
```

rsudarikov:
> few styling nits. Did you test this approach on both old and new kernels?

@anakryiko Good suggestions! Seems that verifier is getting smarter for v5.11 and accepts the following code:
```
SEC("tp_btf/block_rq_insert")
int block_rq_insert(u64 *ctx)
{
    int idx = (LINUX_KERNEL_VERSION <= KERNEL_VERSION(5, 10, 0));
    struct request *rq = (void *)ctx[idx];

    return trace_rq_start(rq);
}
```
For kernels < v5.11, verifier doesn't accept the code above and complains with the following error:
```
libbpf: -- BEGIN DUMP LOG ---
libbpf:
arg#0 type is not a struct
Unrecognized arg#0 type PTR
; int idx = (LINUX_KERNEL_VERSION <= KERNEL_VERSION(5, 10, 0));
0: (18) r2 = 0xfffffe00353f0000
2: (61) r3 = *(u32 *)(r2 +0)
R1=ctx(id=0,off=0,imm=0) R2_w=map_value(id=0,off=0,ks=4,vs=4,imm=0) R10=fp0
3: (67) r3 <<= 32
4: (c7) r3 s>>= 32
5: (b7) r2 = 1
6: (b7) r4 = 330241
; int idx = (LINUX_KERNEL_VERSION <= KERNEL_VERSION(5, 10, 0));
7: (6d) if r4 s> r3 goto pc+1
last_idx 7 first_idx 0
regs=10 stack=0 before 6: (b7) r4 = 330241
last_idx 7 first_idx 0
regs=8 stack=0 before 6: (b7) r4 = 330241
regs=8 stack=0 before 5: (b7) r2 = 1
regs=8 stack=0 before 4: (c7) r3 s>>= 32
regs=8 stack=0 before 3: (67) r3 <<= 32
regs=8 stack=0 before 2: (61) r3 = *(u32 *)(r2 +0)
; struct request *rq = (void *)ctx[idx];
9: (67) r2 <<= 3
10: (0f) r1 += r2
last_idx 10 first_idx 0
regs=4 stack=0 before 9: (67) r2 <<= 3
regs=4 stack=0 before 7: (6d) if r4 s> r3 goto pc+1
regs=4 stack=0 before 6: (b7) r4 = 330241
regs=4 stack=0 before 5: (b7) r2 = 1
11: (79) r6 = *(u64 *)(r1 +0)
dereference of modified ctx ptr R1 off=8 disallowed
processed 10 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0

libbpf: -- END LOG --
libbpf: failed to load program 'block_rq_insert'
libbpf: failed to load object 'biolatency_bpf'
libbpf: failed to load BPF skeleton 'biolatency_bpf': -4007

```
New changes has been tested against v5.8 to v5.11, new helper function is to minimize boilerplate code and make verifier happy.
Please review

libbpf-tools/biolatency.bpf.c
anakryiko: ping this

libbpf-tools/biolatency.bpf.c
```diff
@@ -6,9 +6,14 @@
 #include <bpf/bpf_tracing.h>
 #include "biolatency.h"
 #include "bits.bpf.h"
+#include <linux/version.h>
 
 #define MAX_ENTRIES	10240
 
+#define KERNEL_VERSION(a, b, c) (((a) << 16) + ((b) << 8) + (c))
```
anakryiko, line 13: we should be able to remove this, once latest libbpf sync gets through and libbpf is updated in BCC repo. bpf_helpers.h is not providing this macro

libbpf-tools/biolatency.bpf.c
```diff
@@ -52,17 +57,42 @@ int trace_rq_start(struct request *rq)
...
+	if (LINUX_KERNEL_VERSION <= KERNEL_VERSION(5, 10, 0))
+		return trace_rq_start((void *)ctx[1]);
+	else
+		return trace_rq_start((void *)ctx[0]);
 }
 
-SEC("tp_btf/block_rq_issue")
-int BPF_PROG(block_rq_issue, struct request_queue *q, struct request *rq)
+static __always_inline
+int f(u64 *ctx, int idx)
```
anakryiko, line 74: f() is not a good name, please come up with something more sane. And why passing ctx and idx separately, not just desired pointer type (`struct request *`)?  Makes no sense to do ctx[idx] everywhere, if you can do it down in block_rq_issue very explicitly.

anakryiko (changes requested):

libbpf-tools/biolatency.bpf.c
rsudarikov: done

libbpf-tools/biolatency.bpf.c
rsudarikov: done

libbpf-tools/biolatency.bpf.c
rsudarikov: agree, reworked that part.

anakryiko (approved):
LGTM. Thanks.

yonghong-song:
[buildbot, test this please]

PR #3318, opened 2021-03-15
libbpf-tools: fix non-C89-compliant for loop variable declarations

anakryiko:
Fix all the found instances of declaring loop variable inside for() construct,
which is not supported in C89 standard, which is what libbpf-tools are trying
to adhere to, both in user-space and BPF source code. It actually causes
compilation errors on some versions of GCC, as reported by customers in
private conversations.

Signed-off-by: Andrii Nakryiko <andrii@kernel.org>


yonghong-song:
[buildbot, test this please]

PR #3322, opened 2021-03-16
FAQ: Update LD_LIBRARY_PATH and PYTHONPATH

seehearfeel:
libbcc.so is actually installed to /usr/lib64 instead of
/usr/local/lib in my computer system:

$ find /usr -name libbcc.so
/usr/lib64/libbcc.so

And also we can find out bcc-python:

$ find /usr/lib -name bcc
/usr/lib/python2.7/site-packages/bcc

It is better to use one line command to export environment
variable LD_LIBRARY_PATH and PYTHONPATH.

Signed-off-by: Tiezhu Yang <yangtiezhu@loongson.cn>


yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks for putting such info in FAQ!

Issue #3293, opened 2021-03-05
'finish_task_switch' is optimized away by GCC 10.2, attach_kprobe failed


docularxu:
Since 'finish_task_switch' is such a common tracing point where many bcc tools use, such as in: 
`tools/offcputime.py:b.attach_kprobe(event="finish_task_switch", fn_name="oncpu")`
`libbpf-tools/cpudist.bpf.c:int BPF_KPROBE(finish_task_switch, struct task_struct *prev)
`
 I think it's useful to report this issue here.

I am using GCC 10 and GCC 9 from debian sid stable.
`echo 'deb http://ftp.debian.org/debian sid main' >> /etc/apt/sources.list
`
 Their version details are respectively:
`# gcc --version`
`gcc (Debian 10.2.1-6) 10.2.1 20210110`

`# gcc --version`
`gcc (Debian 9.3.0-22) 9.3.0
`
My environment is an Arm64 machine. And I'm doing a native arm64 build with GCC. The kernel build commands I am using is this, in both gcc 9 and gcc 10 cases:
`  cd linux.git`
`   git checkout -b working-5.11 v5.11`
`   ARCH=arm64 make mrproper O=../linux.buildout`
`   ARCH=arm64 make defconfig O=../linux.buildout`
`   ARCH=arm64 make Image -j 96 O=../linux.buildout`

The results, when built with gcc 9.3: 
`nm ../linux.buildout/kernel/sched/core.o | grep finish_task_switch`
`0000000000001288 t finish_task_switch`

However, when built with gcc 10.2:
`nm ../linux.buildout/kernel/sched/core.o | grep finish_task_switch`
`00000000000012d0 t finish_task_switch.isra.0`

The same symbols (with xxx.isra.0 or without, respectively of course) also appear in 'System.map' and in '/proc/kallsyms' when booting. So, here you can see the problem if I use the kernel built by gcc 10:

bcc tools such as 'offcputime' will report: Failed to attach BPF program ... kprobe finish_task_switch. 
But what really exists in gcc 10 built kernel is "finish_task_switch.isra.0"

docularxu:
In the source level, ( linux.git/kernel/sched/core.c), neither adding `noinline`, nor adding ` asm{""}` can change the behavior of GCC 10 built result.
`static struct rq *finish_task_switch(struct task_struct *prev)
`

The only way to make GCC 10 generate a __clean__ 'finish_task_switch' is deleting the `static`. Sadly.

[...]

PR #3334, opened 2021-03-19
libbpf-tools: add offcputime

ethercflow:
Compared with the previous version, raw_tp/sched_switch is used instead of kprobe/finish_task_switch, because the latter has symbolic name problems under gcc 10 https://github.com/iovisor/bcc/issues/3293.

Functionally compared with the old version, the following features are missing, will be completed in subsequent PR.

Todo list

- [x] Support user stack symbol parsing from ELF & DYN
- [x] Support user stack symbol parsing from vdso
- [ ] Support user stack symbol parsing from perf-map
- [ ] Support user stack symbol parsing with mount ns
- [ ] Sort the results in ascending order according to the length of offcpu time
- [ ] Support folding output results
- [ ] missing stack event report

The output looks like below:
```
    bpf_prog_f8f33eb3e61e9447_sched_switch
    bpf_prog_f8f33eb3e61e9447_sched_switch
    bpf_trace_run3
    __bpf_trace_sched_switch
    __schedule
    schedule
    futex_wait_queue_me
    futex_wait
    do_futex
    __x64_sys_futex
    do_syscall_64
    runtime.futex
    runtime.notesleep
    runtime.stopm
    runtime.findrunnable
    runtime.schedule
    runtime.park_m
    runtime.mcall
    [unknown]
    -                containerd (1172)
        69917
```

### DIFF
--- a/libbpf-tools/offcputime.bpf.c
+++ b/libbpf-tools/offcputime.bpf.c
@@ -1,0 +1,110 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2021 Wenbo Zhang
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_tracing.h>
+#include "offcputime.h"
+
+#define PF_KTHREAD		0x00200000	/* I am a kernel thread */
+#define MAX_ENTRIES		10240
+
+const volatile bool kernel_threads_only = false;
+const volatile bool user_threads_only = false;
+const volatile __u64 max_block_ns = -1;
+const volatile __u64 min_block_ns = 1;
+const volatile pid_t targ_tgid = -1;
+const volatile pid_t targ_pid = -1;
+const volatile long state = -1;
+
+struct internal_key {
+	u64 start_ts;
+	struct key_t key;
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__type(key, u32);
+	__type(value, struct internal_key);
+	__uint(max_entries, MAX_ENTRIES);
+} start SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_STACK_TRACE);
+	__uint(key_size, sizeof(u32));
+} stackmap SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__type(key, struct key_t);
+	__type(value, struct val_t);
+	__uint(max_entries, MAX_ENTRIES);
+} info SEC(".maps");
+
+static bool allow_record(struct task_struct *t)
+{
+	if (targ_tgid != -1 && targ_tgid != t->tgid)
+		return false;
+	if (targ_pid != -1 && targ_pid != t->pid)
+		return false;
+	if (user_threads_only && t->flags & PF_KTHREAD)
+		return false;
+	else if (kernel_threads_only && !(t->flags & PF_KTHREAD))
+		return false;
+	if (state != -1 && t->state != state)
+		return false;
+	return true;
+}
+
+SEC("tp_btf/sched_switch")
+int BPF_PROG(sched_switch, bool preempt, struct task_struct *prev,
+	     struct task_struct *next)
+{
+	struct internal_key *i_keyp, i_key;
+	struct val_t *valp, val;
+	s64 delta;
+	u32 pid;
+
+	if (allow_record(prev)) {
+		pid = prev->pid;
+		/* To distinguish idle threads of different cores */
+		if (!pid)
+			pid = bpf_get_smp_processor_id();
+		i_key.key.pid = pid;
+		i_key.key.tgid = prev->tgid;
+		i_key.start_ts = bpf_ktime_get_ns();
+
+		if (prev->flags & PF_KTHREAD)
+			i_key.key.user_stack_id = -1;
+		else
+			i_key.key.user_stack_id =
+				bpf_get_stackid(ctx, &stackmap,
+						BPF_F_USER_STACK);
+		i_key.key.kern_stack_id = bpf_get_stackid(ctx, &stackmap, 0);
+		bpf_map_update_elem(&start, &pid, &i_key, 0);
+		bpf_probe_read_str(&val.comm, sizeof(prev->comm), prev->comm);
+		val.delta = 0;
+		bpf_map_update_elem(&info, &i_key.key, &val, BPF_NOEXIST);
+	}
+
+	pid = next->pid;
+	i_keyp = bpf_map_lookup_elem(&start, &pid);
+	if (!i_keyp)
+		return 0;
+	delta = (s64)(bpf_ktime_get_ns() - i_keyp->start_ts);
+	if (delta < 0)
+		goto cleanup;
+	delta /= 1000U;
+	if (delta < min_block_ns || delta > max_block_ns)
+		goto cleanup;
+	valp = bpf_map_lookup_elem(&info, &i_keyp->key);
+	if (!valp)
+		goto cleanup;
+	__sync_fetch_and_add(&valp->delta, delta);
+
+cleanup:
+	bpf_map_delete_elem(&start, &pid);
+	return 0;
+}
+
+char LICENSE[] SEC("license") = "GPL";
--- a/libbpf-tools/offcputime.c
+++ b/libbpf-tools/offcputime.c
@@ -1,0 +1,342 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+// Copyright (c) 2021 Wenbo Zhang
+//
+// Based on offcputime(8) from BCC by Brendan Gregg.
+// 19-Mar-2021   Wenbo Zhang   Created this.
+#include <argp.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <time.h>
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+#include "offcputime.h"
+#include "offcputime.skel.h"
+#include "trace_helpers.h"
+
+static struct env {
+	pid_t pid;
+	pid_t tid;
+	bool user_threads_only;
+	bool kernel_threads_only;
+	int stack_storage_size;
+	int perf_max_stack_depth;
+	__u64 min_block_time;
+	__u64 max_block_time;
+	long state;
+	int duration;
+	bool verbose;
+} env = {
+	.pid = -1,
+	.tid = -1,
+	.stack_storage_size = 1024,
+	.perf_max_stack_depth = 127,
+	.min_block_time = 1,
+	.max_block_time = -1,
+	.state = -1,
+	.duration = 99999999,
+};
+
+static volatile bool exiting;
+
+const char *argp_program_version = "offcputime 0.1";
+const char *argp_program_bug_address =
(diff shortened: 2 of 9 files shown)


libbpf-tools/offcputime.h
```diff
@@ -0,0 +1,15 @@
...
+#define __OFFCPUTIME_H
+
+#define TASK_COMM_LEN		16
+
+struct key_t {
+	__u32 pid;
+	__u32 tgid;
+	int user_stack_id;
+	int kern_stack_id;
+	char comm[TASK_COMM_LEN]; /* in order to distinguish idle threads */
```
ethercflow, line 12: The reason why it has not been moved to val is to distinguish between different core idle threads. I remember you mentioned that comparing char [] in key is slower. 

Since the id of idle is 0, I think we can set the pid field to the value of `bpf_get_smp_processor_id()`  instead for better performance?

libbpf-tools/offcputime.bpf.c
```diff
@@ -0,0 +1,101 @@
...
+	if (allow_record(prev)) {
+		pid = prev->pid;
+		i_key.start_ts = bpf_ktime_get_ns();
+		i_key.key.pid = pid;
+		i_key.key.tgid = prev->tgid;
+		i_key.key.user_stack_id = -1; /* will support it later */
+		i_key.key.kern_stack_id = bpf_get_stackid(ctx, &stackmap, 0);
+		bpf_probe_read_str(&i_key.key.comm, sizeof(prev->comm),
+				   prev->comm);
+		bpf_map_update_elem(&start, &pid, &i_key, 0);
```
ethercflow, line 76: From

> If you are using recently added BTF_PROG_TYPE_TRACING BPF programs, then you have a smartness of BPF verifier on your side, which now understands and tracks BTF types natively and allows you to follow pointers and read kernel memory directly (and safely), avoiding bpf_probe_read() calls, so you don’t need compiler rewriting magic to get the same nice and familiar syntax.
> 
> Libbpf + BPF_PROG_TYPE_TRACING way:
> 
> pid_t pid = task->pid;

I tried to use `&prev->pid ` as the map key, but got BPF verifier's complain:


```
106: (85) call bpf_map_update_elem#2
R2 type=ptr_ expected=fp, pkt, pkt_meta, map_value
```

Is this as expected?

davemarchevsky (approved):
@anakryiko asked me to take a look at this as I've worked on similar symbolication stuff in the past. This LGTM and I think it's mergeable as-is. 

The symbolication helpers were easy to follow and can be a good foundation for a general symbolication library. I have some suggestions that I think would move it in that direction, but don't think any of them are necessary to address in this diff. 
  * Allowing the user to provide a path to a file - or provide a string - with same format as `/proc/PID/maps`, to the fns which open this file now, would make this code easier to test.
    * Also, would enable some use-cases like using [`jemalloc` heap dump files](http://jemalloc.net/jemalloc.3.html#heap_profile_format), which include a `/proc/PID/maps` section, to symbolicate
 * More flexible behavior when dealing with a malformed ELF. Maybe this is a niche use-case, but I encounter ELFs with malformed symbol tables or symbols semi-regularly. Being able to ask the symbolication code to complain loudly about parse/format issues when it encounters them, but try to continue, helps track down the source of the issue.
 * Exited process support / caching of map info. For tools which are meant to run for a long time, but symbolicate and print data when exiting, it's possible that some pids the tool has collected data for die before the tool exits. 
 * Dealing with mount namespaces. The paths in `/proc/PID/maps` are relative to that process' mount namespace. If the tool is running as root and the process is in some container with isolated mounts, it's likely that opening the path provided will not work.

libbpf-tools/uprobe_helpers.c
```diff
@@ -169,10 +169,8 @@ static Elf *open_elf(const char *path, int *fd_close)
 		return NULL;
 	}
 	fd = open(path, O_RDONLY);
-	if (fd < 0) {
-		warn("Could not open %s\n", path);
```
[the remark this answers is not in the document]
ethercflow: Oh, I deleted it during debugging, I will put it back. Thanks!

ethercflow:
> @anakryiko asked me to take a look at this as I've worked on similar symbolication stuff in the past. This LGTM and I think it's mergeable as-is.
> 
> The symbolication helpers were easy to follow and can be a good foundation for a general symbolication library. I have some suggestions that I think would move it in that direction, but don't think any of them are necessary to address in this diff.
> 
> * Allowing the user to provide a path to a file - or provide a string - with same format as `/proc/PID/maps`, to the fns which open this file now, would make this code easier to test.
>   
>   * Also, would enable some use-cases like using [`jemalloc` heap dump files](http://jemalloc.net/jemalloc.3.html#heap_profile_format), which include a `/proc/PID/maps` section, to symbolicate
> * More flexible behavior when dealing with a malformed ELF. Maybe this is a niche use-case, but I encounter ELFs with malformed symbol tables or symbols semi-regularly. Being able to ask the symbolication code to complain loudly about parse/format issues when it encounters them, but try to continue, helps track down the source of the issue.
> * Exited process support / caching of map info. For tools which are meant to run for a long time, but symbolicate and print data when exiting, it's possible that some pids the tool has collected data for die before the tool exits.
> * Dealing with mount namespaces. The paths in `/proc/PID/maps` are relative to that process' mount namespace. If the tool is running as root and the process is in some container with isolated mounts, it's likely that opening the path provided will not work.

Thanks @davemarchevsky for your suggestion, I think it makes a lot of sense. I will continue to follow up in this direction.

yonghong-song:
@ethercflow There is a conflict, could you rebase?

libbpf-tools/offcputime.c
```diff
@@ -0,0 +1,350 @@
...
+static const struct argp_option opts[] = {
+	{ "pid", 'p', "PID", 0, "Trace this PID only" },
+	{ "tid", 't', "TID", 0, "Trace this TID only" },
+	{ "user-threads-only", 'u', NULL, 0,
+	  "User threads only (no kernel threads)" },
+	{ "kernel-threads-only", 'k', NULL, 0,
+	  "Kernel threads only (no user threads)" },
+	{ "perf-max-stack-depth", OPT_PERF_MAX_STACK_DEPTH,
+	  "PERF-MAX-STACK-DEPTH", 0, "the limit for both kernel and user "
+	  "frames (deault 127)" },
```
anakryiko, line 75: please don't split strings

also typo: default

and "stack traces"?

libbpf-tools/offcputime.c
```diff
@@ -0,0 +1,350 @@
...
+	if (level == LIBBPF_DEBUG && !env.verbose)
+		return 0;
+	return vfprintf(stderr, format, args);
+}
+
+static void sig_handler(int sig)
+{
+}
+
+
```
anakryiko, line 192: nit: extra empty line

libbpf-tools/offcputime.c
```diff
@@ -0,0 +1,350 @@
...
+	struct ksyms *ksyms = NULL;
+	struct offcputime_bpf *obj;
+	int err;
+
+	err = argp_parse(&argp, argc, argv, 0, NULL, NULL);
+	if (err)
+		return err;
+	if (env.user_threads_only && env.kernel_threads_only) {
+		fprintf(stderr, "user_threads_only, kernel_threads_only "
+			"cann't be used together.\n");
```
anakryiko, line 278: typo: can't (and don't split strings, please check all the other strings as well)

libbpf-tools/trace_helpers.c
```diff
@@ -169,6 +175,604 @@ const struct ksym *ksyms__get_symbol(const struct ksyms *ksyms,
...
+}
+
+static int dso__add_sym(struct dso *dso, const char *name, uint64_t start,
+			uint64_t size)
+{
+	size_t new_cap, name_len = strlen(name) + 1;
+	struct sym *sym;
+	void *tmp;
+
+	if (dso->strs_sz + name_len > dso->strs_cap) {
```
anakryiko, line 393: so, libbpf's `struct btf` is actually a pretty efficient "set of strings" data structure, how about using it instead? see btf__add_str() and btf__find_str().

libbpf-tools/trace_helpers.h
```diff
@@ -20,6 +20,25 @@ const struct ksym *ksyms__map_addr(const struct ksyms *ksyms,
...
 
+struct sym {
+	const char *name;
+	unsigned long start;
+	unsigned long size;
+};
+
+struct syms;
+
+struct syms *syms__load(int tgid);
```
anakryiko, line 31: as Dave mentioned, you probably would want to have "syms__load_pid" and "syms_load_file" variants

libbpf-tools/trace_helpers.c
```diff
@@ -169,6 +175,604 @@ const struct ksym *ksyms__get_symbol(const struct ksyms *ksyms,
...
+	return dso__find_sym(dso, offset);
+}
+
+const struct sym *syms__get_symbol(const struct syms *syms,
+				   const char *name)
+{
+	return NULL;
+}
+
+struct syms_vec {
```
anakryiko, line 723: I think syms_vec is misleading name. This is a cache of per-file/per-process symbols, right? It doesn't have to be internally implemented as a vector (dynamic array). So maybe syms_cache or something along those lines?

libbpf-tools/trace_helpers.c
```diff
@@ -169,6 +175,604 @@ const struct ksym *ksyms__get_symbol(const struct ksyms *ksyms,
...
+	dso = syms__find_dso(syms, addr, &offset);
+	if (!dso)
+		return NULL;
+	return dso__find_sym(dso, offset);
+}
+
+const struct sym *syms__get_symbol(const struct syms *syms,
+				   const char *name)
+{
+	return NULL;
```
anakryiko, line 720: if this is not implemented, let's drop the API for now?

anakryiko (commented):
I didn't read the actual logic of parsing ELF and extracting symbols, just few remarks. Lots of work was put into this, thanks!

libbpf-tools/offcputime.c
ethercflow: Fixed, thanks. I will submit a pr later to fix all string split problems in other tools. I understand the reason for this from your last pr comment. This time is too busy to update this pr.

libbpf-tools/offcputime.c
ethercflow: Fixed, thanks!

libbpf-tools/offcputime.c
ethercflow: Fixed, thanks!

libbpf-tools/trace_helpers.c
ethercflow: Cool, changed to this, thanks!

libbpf-tools/trace_helpers.h
ethercflow: Yes, changed to these two variants, thanks!

libbpf-tools/trace_helpers.c
ethercflow: Sure, fixed, thanks!

libbpf-tools/trace_helpers.c
ethercflow: Sure, removed.

ethercflow:
> @ethercflow There is a conflict, could you rebase?

Rebased, thanks!

ethercflow:
> I didn't read the actual logic of parsing ELF and extracting symbols, just few remarks. Lots of work was put into this, thanks!

Thanks for your suggestions, the current problems have been fixed.

yonghong-song:
[buildbot, test this please]

PR #3339, opened 2021-03-20
tools: handle renamed lookup_fast function in dcache tools

jeromemarchand:
The lookup_fast function can be rename lookup_fast.constprop.x by gcc
constant propagation optimization and that breaks dcstat and
dcsnoop. Let's look for both name using a regular expression.

### DIFF
--- a/tools/dcsnoop.py
+++ b/tools/dcsnoop.py
@@ -129,7 +129,7 @@ if args.ebpf:
 # initialize BPF
 b = BPF(text=bpf_text)
 if args.all:
-    b.attach_kprobe(event="lookup_fast", fn_name="trace_fast")
+    b.attach_kprobe(event_re="^lookup_fast$|^lookup_fast.constprop.*.\d$", fn_name="trace_fast")
 
 mode_s = {
     0: 'M',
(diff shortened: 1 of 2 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
which gcc version you are using? is it possible we can dig out what are const propagated? In `look_fast` function, dcstat is using `nd->last.name`, where `name` is the filename. I don't think nd->last.name can be constant propagated. it is possible nd->flags or other nd fields.

jeromemarchand:
> which gcc version you are using? is it possible we can dig out what are const propagated? In `look_fast` function, dcstat is using `nd->last.name`, where `name` is the filename. I don't think nd->last.name can be constant propagated. it is possible nd->flags or other nd fields.

It's gcc 11.0.0. I don't much about constant propagation, but hardly imagine `nd->last.name` is a constant. I just know that the function name was suffixed with .constprop.0

yonghong-song:
I don't have gcc11 at hand. But I agree `nd->last.name` should not be the one becoming constant in the func as it is never explicitly reset. Some flags are possible though. Thanks for the fix!

PR #3346, opened 2021-03-31
execsnoop: reliably fetch arguments and environ variable on arm64

lebauce:
Accessing argv and envp parameters of execve syscall can cause minor page faults
on arm64, which cause bpf_probe_user_read functions to return EFAULT. Therefore,
we hook into either 'prepare_binprm' or 'bprm_execve' functions that are called
just after the kernel accessed it first.

### DIFF
--- a/tools/execsnoop.py
+++ b/tools/execsnoop.py
@@ -117,6 +117,13 @@ struct data_t {
     int retval;
 };
 
+struct val_t {
+    const char *filename;
+    const char *const *argv;
+    const char *const *envp;
+};
+
+BPF_HASH(entryinfo, u64, struct val_t);
 BPF_PERF_OUTPUT(events);
 
 static int __submit_arg(struct pt_regs *ctx, void *ptr, struct data_t *data)
@@ -136,19 +143,20 @@ static int submit_arg(struct pt_regs *ctx, void *ptr, struct data_t *data)
     return 0;
 }
 
-int syscall__execve(struct pt_regs *ctx,
-    const char __user *filename,
-    const char __user *const __user *__argv,
-    const char __user *const __user *__envp)
+int parse_argv_envp(struct pt_regs *ctx)
 {
+    struct val_t *valp;
+    u64 id = bpf_get_current_pid_tgid();
 
-    u32 uid = bpf_get_current_uid_gid() & 0xffffffff;
-
-    UID_FILTER
-
-    if (container_should_be_filtered()) {
+    valp = entryinfo.lookup(&id);
+    if (valp == 0) {
+        // missed tracing issue or filtered
         return 0;
     }
+
+    const char *filename = valp->filename;
+    const char *const *__argv = valp->argv;
+    const char *const *__envp = valp->envp;
 
     // create data here and pass to submit_arg to save stack space (#555)
     struct data_t data = {};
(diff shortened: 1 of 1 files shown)


tools/execsnoop.py
```diff
@@ -227,6 +259,10 @@ def parse_uid(user):
 b = BPF(text=bpf_text)
 execve_fnname = b.get_syscall_fnname("execve")
 b.attach_kprobe(event=execve_fnname, fn_name="syscall__execve")
+if BPF.get_kprobe_functions(b'prepare_binprm'):
+    b.attach_kprobe(event="prepare_binprm", fn_name="parse_argv_envp")
+else:
+    b.attach_kprobe(event="bprm_execve", fn_name="parse_argv_envp")
```
yonghong-song, line 265: In latest kernel (I checked bpf-next only), both prepare_binprm and bprm_execve become static variable and prepare_binprm became a much smaller function and easily subject to inlining.

Since we are saving pointer and wait until a later pointer to do the actually probe_read. Can we do this during return probe?

lebauce: When the `execve` succeeds, the return probe is never executed (the kretprobe seems to be there only to get the args of `execve` that failed).

`prepare_binrpm` is supposed to be used on old kernels where it was an exported symbol using EXPORT_SYMBOL (I missed the fact that `prepare_binrpm` is still present in early kernels but almost always inlined as you said because it's very small).

`bprm_execve` is indeed declared static but almost never inlined as it's pretty big. We could switch to using `prepare_exec_creds` which is declared as `extern` and called shortly after the args were accessed by the kernel.

To sum up, I suggest trying to first attach to `prepare_exec_creds` and, if it fails, attach to `prepare_binprm`. WDYT ?

yonghong-song: > When the execve succeeds, the return probe is never executed (the kretprobe seems to be there only to get the args of execve that failed).

This seems not true for me on x64. I see
```
$ sudo trace.py 'r::__x64_sys_execve "retval=%d", retval'
2987496 2987496 env             __x64_sys_execve retval=0
2987497 2987497 env             __x64_sys_execve retval=0
2987498 2987498 env             __x64_sys_execve retval=0
2987496 2987496 timeout         __x64_sys_execve retval=0
2987497 2987497 timeout         __x64_sys_execve retval=0
2987498 2987498 timeout         __x64_sys_execve retval=0
2987500 2987500 sh              __x64_sys_execve retval=0
2987499 2987499 sh              __x64_sys_execve retval=0
2987501 2987501 sh              __x64_sys_execve retval=0
2987499 2987499 wdb-up.sh       __x64_sys_execve retval=0
2987501 2987501 check_binary_ru __x64_sys_execve retval=0
2987502 2987502 sar             __x64_sys_execve retval=0
2987503 2987503 awk             __x64_sys_execve retval=0
```

The function latency indicates these short live processes takes
average 0.2 millisecond,
```
$ sudo funclatency.py __x64_sys_execve
Tracing 1 functions for "__x64_sys_execve"... Hit Ctrl-C to end.
^C
     nsecs               : count     distribution
      1024 -> 2047       : 2        |                                        |
      2048 -> 4095       : 6        |*                                       |
      4096 -> 8191       : 10       |*                                       |
      8192 -> 16383      : 0        |                                        |
     16384 -> 32767      : 0        |                                        |
     32768 -> 65535      : 0        |                                        |
     65536 -> 131071     : 56       |**********                              |
    131072 -> 262143     : 215      |****************************************|
    262144 -> 524287     : 17       |***                                     |
    524288 -> 1048575    : 1        |                                        |
   1048576 -> 2097151    : 0        |                                        |
   2097152 -> 4194303    : 7        |*                                       |

avg = 205112 nsecs, total: 64610445 nsecs, count: 315
```

so arm64 behave differently?

> To sum up, I suggest trying to first attach to prepare_exec_creds and, if it fails, attach to prepare_binprm. WDYT ?

`prepare_exec_creds` indeed looks better. But I would like to get a confirmation execve retprobe indeed not working in arm64 compared to x86_64.

PR #3347, opened 2021-03-31
[libbpf-tools/readahead] switching to kprobes/kretprobes instead of fentry/fexit for arm64

rsudarikov:
libbpf-tools/readahead fails on arm64 with ENOTSUPP since eBPF fentry/fexit is not supported so fallback to kprobes/kretprobes.

Note: dynamic attach to program also fails with EINVAL for kprobes (not supported yet?) so currently just guarded off for arm64.
Any hints for more elegant solution for dynamic attach are welcome)

### DIFF
--- a/libbpf-tools/readahead.bpf.c
+++ b/libbpf-tools/readahead.bpf.c
@@ -26,8 +26,8 @@ struct {
 
 static struct hist hist;
 
-SEC("fentry/do_page_cache_ra")
-int BPF_PROG(do_page_cache_ra)
+static __always_inline
+int __do_page_cache_ra()
 {
 	u32 pid = bpf_get_current_pid_tgid();
 	u64 one = 1;
@@ -36,8 +36,17 @@ int BPF_PROG(do_page_cache_ra)
 	return 0;
 }
 
-SEC("fexit/__page_cache_alloc")
-int BPF_PROG(page_cache_alloc_ret, gfp_t gfp, struct page *ret)
+static __always_inline
+int __do_page_cache_ra_ret()
+{
+	u32 pid = bpf_get_current_pid_tgid();
+
+	bpf_map_delete_elem(&in_readahead, &pid);
+	return 0;
+}
+
+static __always_inline
+int __page_cache_alloc_ret(struct page *ret)
 {
 	u32 pid = bpf_get_current_pid_tgid();
 	u64 ts;
@@ -53,17 +62,8 @@ int BPF_PROG(page_cache_alloc_ret, gfp_t gfp, struct page *ret)
 	return 0;
 }
 
-SEC("fexit/do_page_cache_ra")
-int BPF_PROG(do_page_cache_ra_ret)
-{
-	u32 pid = bpf_get_current_pid_tgid();
-
-	bpf_map_delete_elem(&in_readahead, &pid);
-	return 0;
-}
-
-SEC("fentry/mark_page_accessed")
-int BPF_PROG(mark_page_accessed, struct page *page)
+static __always_inline
+int __mark_page_accessed(struct page *page)
 {
 	u64 *tsp, slot, ts = bpf_ktime_get_ns();
 	s64 delta;
@@ -84,6 +84,53 @@ update_and_cleanup:
 	bpf_map_delete_elem(&birth, &page);
 
 	return 0;
+}
+
+SEC("kprobe/do_page_cache_ra")
+int BPF_KPROBE(kprobe_do_page_cache_ra)
+{
+	return __do_page_cache_ra();
+}
+
+SEC("fentry/do_page_cache_ra")
+int BPF_PROG(fentry_do_page_cache_ra)
+{
+	return __do_page_cache_ra();
+}
+
+SEC("kretprobe/do_page_cache_ra")
+int BPF_KPROBE(kretprobe_do_page_cache_ra)
+{
+	return __do_page_cache_ra_ret();
+}
+
+SEC("fexit/do_page_cache_ra")
+int BPF_PROG(fexit_do_page_cache_ra)
+{
+	return __do_page_cache_ra_ret();
+}
+
+SEC("kretprobe/__page_cache_alloc")
+int BPF_KRETPROBE(kretprobe_page_cache_alloc_ret, struct page *ret)
+{
+	return __page_cache_alloc_ret(ret);
+}
+
+SEC("fexit/__page_cache_alloc")
+int BPF_PROG(fexit_page_cache_alloc_ret, struct page *ret)
+{
+	return __page_cache_alloc_ret(ret);
+}
+
+SEC("kprobe/mark_page_accessed")
+int BPF_KPROBE(kprobe_mark_page_accessed, struct page *page)
+{
+	return __mark_page_accessed(page);
+}
+SEC("fentry/mark_page_accessed")
+int BPF_PROG(fentry_mark_page_accessed, struct page *page)
+{
+	return __mark_page_accessed(page);
 }
 
 char LICENSE[] SEC("license") = "GPL";
--- a/libbpf-tools/readahead.c
+++ b/libbpf-tools/readahead.c
@@ -73,6 +73,10 @@ static void sig_handler(int sig)
 	exiting = true;
 }
 
+/*
+ * Starting from v5.10-rc1 (8238287), __do_page_cache_readahead has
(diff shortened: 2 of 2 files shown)


libbpf-tools/biolatency.bpf.c
```diff
@@ -9,6 +9,10 @@
 
 #define MAX_ENTRIES	10240
 
+#define KERNEL_VERSION(a, b, c) (((a) << 16) + ((b) << 8) + (c))
```
anakryiko, line 12: libbpf was already updated, so you should get this macro from bpf_helpers.h

libbpf-tools/biolatency.bpf.c
```diff
@@ -52,17 +59,31 @@ int trace_rq_start(struct request *rq)
 }
 
 SEC("tp_btf/block_rq_insert")
-int BPF_PROG(block_rq_insert, struct request_queue *q, struct request *rq)
+int block_rq_insert(u64 *ctx)
 {
-	return trace_rq_start(rq);
+	/**
+	 * commit a54895fa (v5.11-rc1) changed tracepoint argument list
```
anakryiko, line 65: we already landed PR with these changes. Please rebase.

libbpf-tools/readahead.c
```diff
@@ -119,7 +119,7 @@ int main(int argc, char **argv)
 		fprintf(stderr, "failed to open BPF object\n");
 		return 1;
 	}
-
+#ifndef __aarch64__
```
anakryiko, line 122: the problem here is that you still have to handle function rename between kernel versions, just not for fentry but for kprobe. As is, tool will only work for v5.10+.

I think a bit cleaner solution would be to have both kprobe and fentry functions defined in BPF (with no arm64-specific `#ifndef` in BPF code). Then in user-space, do detection of all the cases we need to handle: arm64 vs non-arm64, do_page_cache_ra vs __do_page_cache_readahead.

To make this simpler, instead of using skeleton's readahead_bpf__attach(), we can attach each program individually. So in some cases you'll do bpf_program__attach_trace() (for fentry, preceded by correct bpf_program__set_attach_target), in others you'll do bpf_program__attach_kprobe() with the right kernel function name.

While at it, and with that approach, we can also add fallback to kprobe/kretprobe for x86, if we detect that fentry isn't supported. It's a bit complicated logic on user-space side, but we are trying to handle vastly differing environments, so I think it's justified.

WDYT?

rsudarikov: reworked; basic idea was to rely on the arch_prepare_bpf_trampoline return value which is (-ENOTSUPP) for everything except x86

libbpf-tools/biolatency.bpf.c
rsudarikov: fixed

libbpf-tools/biolatency.bpf.c
rsudarikov: fixed

libbpf-tools/readahead.c
```diff
@@ -119,28 +139,35 @@ int main(int argc, char **argv)
...
-		goto cleanup;
+	link = bpf_program__attach_trace(obj->progs.fexit_page_cache_alloc_ret);
+	if ((long)link == -524 /*ENOTSUPP*/) {
+		err = readahead__probe_target(obj->progs.kprobe_do_page_cache_ra, false);
+		if (err)
+			goto cleanup;
+		err = readahead__probe_target(obj->progs.kprobe_do_page_cache_ra, true);
+		if (err)
+			goto cleanup;
+		bpf_program__attach_kprobe(obj->progs.kretprobe_page_cache_alloc_ret,
```
chenhengqi, line 156: The return value of `bpf_program__attach_kprobe` should assign to `obj->links.xxx`, and we should do error check here.

chenhengqi (changes requested):

libbpf-tools/readahead.c
rsudarikov: It will not work since bpf_program__attach_kprobe returns perf_event_open_probe's return value if the latter fails.

rsudarikov: > The return value of `bpf_program__attach_kprobe` should assign to `obj->links.xxx`, and we should do error check here.

It will not work since bpf_program__attach_kprobe returns perf_event_open_probe's return value if the latter fails.

libbpf-tools/readahead.bpf.c
```diff
@@ -86,4 +86,51 @@ int BPF_PROG(mark_page_accessed, struct page *page)
...
+{
+	return __page_cache_alloc_ret(ret);
+}
+
+SEC("kprobe/mark_page_accessed")
+int BPF_KPROBE(kprobe_mark_page_accessed, struct page *page)
+{
+	return __mark_page_accessed(page);
+}
+SEC("fentry/mark_page_accessed")
```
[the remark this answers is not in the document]
rsudarikov: will fix next commit

libbpf-tools/readahead.c
```diff
@@ -119,28 +139,35 @@ int main(int argc, char **argv)
...
-	 * renamed to do_page_cache_ra. So we specify the function dynamically.
-	 */
-	err = readahead__set_attach_target(obj->progs.do_page_cache_ra);
-	if (err)
-		goto cleanup;
-	err = readahead__set_attach_target(obj->progs.do_page_cache_ra_ret);
-	if (err)
-		goto cleanup;
-
 	err = readahead_bpf__load(obj);
```
anakryiko, line 142: how did you test these changes? I suspect it won't work on some kernels, because for fentry/fexit programs you have to `bpf_program__set_autoload(false)` them on old kernels before `readahead_bpf__load()` is called, otherwise BPF verifier will reject them. Then, for cases where you do want to use fentry/fexit you need to do `readahead__set_attach_target()` before load as well to set correct target functions.

Lastly, for bpf_program__attach_kprobe, given you specify attach function name at the attach type, you don't need two copies of kretprobe and kprobe in BPF code, you can reuse the same function and just specify the right function name.

cc @ethercflow  I think given that now the conditions of attachment is more complicated, it is probably a good idea to invest in two helper method to check whether some function exists in kernel or not. How about we add few helpers:
  - `bool fentry_exists(const char *name, const char *mod);`. If `mod != NULL`, load `/sys/kernel/btf/<mod>` (if it exists) and check if FUNC with a given name exists; if not, load `/sys/kernel/btf/vmlinux` and check if FUNC with given name exists;
  - `bool kprobe_exists(const char *name);` which should scan `/sys/kernel/tracing/available_filter_functions` (probably with fallback to /proc/kallsyms, which is more expensive to parse). This can be used to detect `__do_page_cache_readahead` vs `do_page_cache_ra` for kprobe attachment case;
  
  So in the end, readahead logic would be something like:
```
if (fentry_exists("do_page_cache_ra", NULL)) {
    bpf_program__set_attach_target(obj->progs.do_page_cache_ra, "do_page_cache_ra");
  /* same for fexit */
} else if (fentry_exists("__do_page_cache_readahead")) {
    bpf_program__set_attach_target(obj->progs.do_page_cache_ra, "__do_page_cache_readahead");
} else if (kprobe_exists("do_page_cache_ra")) {
   /* use bpf_program__attach_kprobe(..., "do_page_cache_ra");
} else {
   bpf_program__attach_kprobe(..., "__do_page_cache_readahead")
}
```

anakryiko (changes requested):

libbpf-tools/readahead.c
ethercflow: Yes, I agree. Compared with the previous way, the introduction of these two helpers will make the logic clearer.

anakryiko: So for fentry_exists case, it's not enough to check that FUNC exists, we need to check that kernel actually supports fentry. As I pointed out in https://github.com/iovisor/bcc/pull/3214#discussion_r617058439, kernel BTF is exposed starting from 5.4 kernel, but fentry/fexit is actually supported starting from 5.5, so there is a gap between kernel versions. But good thing is that with BTF it's actually easy to detect if fentry/fexit is supported:

1. btf__find_by_name_kind(btf, "bpf_attach_type", BTF_KIND_ENUM).
2. If that enum has BPF_TRACE_FENTRY enum value, then kernel supports FENTRY programs.

@ethercflow, would you be able to contribute discussed helpers?

ethercflow: @anakryiko Sure, I will do this in the next few days.

PR #3349, opened 2021-04-02
libbpf-tools: fix some minor issues of ext4dist

KentaTada:
Fix below minor issues of ext4dist

* Fix the check of the existence of module BTFs
* Initialize the global variable

Signed-off-by: Kenta Tada <Kenta.Tada@sony.com>

### DIFF
--- a/libbpf-tools/ext4dist.bpf.c
+++ b/libbpf-tools/ext4dist.bpf.c
@@ -12,7 +12,7 @@ const volatile pid_t targ_tgid = 0;
 
 #define MAX_ENTRIES	10240
 
-struct hist hists[__MAX_FOP_TYPE];
+struct hist hists[__MAX_FOP_TYPE] = {};
 
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
--- a/libbpf-tools/ext4dist.c
+++ b/libbpf-tools/ext4dist.c
@@ -152,7 +152,7 @@ static bool should_fallback(void)
 	 * we can use fentry to get better performance, otherwise we need
 	 * to fall back to use kprobe to be compatible with the old kernel.
 	 */
-	if (is_kernel_module("ext4") && !access("/sys/kernel/btf/ext4", R_OK))
+	if (is_kernel_module("ext4") && access("/sys/kernel/btf/ext4", R_OK))
 		return true;
 	return false;
 }


libbpf-tools/ext4dist.c
```diff
@@ -153,8 +153,8 @@ static bool should_fallback(void)
 	 * to fall back to use kprobe to be compatible with the old kernel.
 	 */
 	if (is_kernel_module("ext4") && !access("/sys/kernel/btf/ext4", R_OK))
-		return true;
-	return false;
+		return false;
+	return true;
```
anakryiko, line 157: as I mentioned on another PR, I think this is wrong. We fallback to kprobe if ext4 is a module and kernel doesn't support module BTFs. That's what current logic is doing, and I think correctly.

libbpf-tools/ext4dist.bpf.c
```diff
@@ -12,7 +12,7 @@ const volatile pid_t targ_tgid = 0;
 
 #define MAX_ENTRIES	10240
 
-struct hist hists[__MAX_FOP_TYPE];
+struct hist hists[__MAX_FOP_TYPE] = {};
```
anakryiko, line 15: this is good, thank you!

anakryiko (changes requested):

libbpf-tools/ext4dist.c
KentaTada: >as I mentioned on another PR, I think this is wrong. We fallback to kprobe if ext4 is a module and kernel doesn't support module BTFs. That's what current logic is doing, and I think correctly.

I misunderstood the objective of this function. Thanks.

BTW,
If kernel doesn't support module BTFs, I think `access("/sys/kernel/btf/ext4", R_OK)` will not return 0 but return -1.
So, I think this check is ` if (is_kernel_module("ext4") && (access("/sys/kernel/btf/ext4", R_OK) == -1 ))`

anakryiko: you are right, `!access()` is wrong here. It should be just `is_kernel_module() && access()` (no need for == -1, enough to check for non-zero return).

KentaTada: Thank for very much for your support.
I fixed it.

anakryiko (approved):
Looks good, thanks!

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

PR #3350, opened 2021-04-03
BCC tools: compatibility modification of scheduler related tools

ethercflow:
Since some companies are still using Centos 7 distribution, I made this compatibility modification, the tools include:

- cpudist
- offcputime
- offwaketime
- runqlat
- runqslower

### DIFF
--- a/tools/offwaketime.py
+++ b/tools/offwaketime.py
@@ -188,7 +188,13 @@ int waker(struct pt_regs *ctx, struct task_struct *p) {
     return 0;
 }
 
+#if LINUX_VERSION_CODE < KERNEL_VERSION(3, 19, 0)
+struct rq;
+
+int oncpu(struct pt_regs *ctx, struct rq *rq, struct task_struct *p) {
+#else
 int oncpu(struct pt_regs *ctx, struct task_struct *p) {
+#endif
     // PID and TGID of the previous Process (Process going into waiting)
     u32 pid = p->pid;
     u32 tgid = p->tgid;
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -560,7 +560,7 @@ class BPF(object):
             with open(blacklist_file, "rb") as blacklist_f:
                 blacklist = set([line.rstrip().split()[1] for line in blacklist_f])
         except IOError as e:
-            if e.errno != errno.EPERM:
+            if e.errno != errno.EPERM and e.errno != errno.ENOENT:
                 raise e
             blacklist = set([])
 
--- a/tools/cpudist.py
+++ b/tools/cpudist.py
@@ -94,7 +94,13 @@ static inline void update_hist(u32 tgid, u32 pid, u64 ts)
     STORE
 }
 
-int sched_switch(struct pt_regs *ctx, struct task_struct *prev)
+#if LINUX_VERSION_CODE < KERNEL_VERSION(3, 19, 0)
+struct rq;
+
+int sched_switch(struct pt_regs *ctx, struct rq *rq, struct task_struct *prev) 
+#else
+int sched_switch(struct pt_regs *ctx, struct task_struct *prev) 
+#endif
 {
     u64 ts = bpf_ktime_get_ns();
     u64 pid_tgid = bpf_get_current_pid_tgid();
--- a/tools/offcputime.py
+++ b/tools/offcputime.py
@@ -134,7 +134,13 @@ struct warn_event_t {
 };
 BPF_PERF_OUTPUT(warn_events);
 
+#if LINUX_VERSION_CODE < KERNEL_VERSION(3, 19, 0)
+struct rq;
+
+int oncpu(struct pt_regs *ctx, struct rq *rq, struct task_struct *prev) {
+#else
 int oncpu(struct pt_regs *ctx, struct task_struct *prev) {
+#endif
     u32 pid = prev->pid;
     u32 tgid = prev->tgid;
     u64 ts, *tsp;
--- a/tools/runqlat.py
+++ b/tools/runqlat.py
@@ -111,7 +111,14 @@ int trace_ttwu_do_wakeup(struct pt_regs *ctx, struct rq *rq, struct task_struct 
 }
 
 // calculate latency
-int trace_run(struct pt_regs *ctx, struct task_struct *prev)
+
+#if LINUX_VERSION_CODE < KERNEL_VERSION(3, 19, 0)
+struct rq;
+
+int trace_run(struct pt_regs *ctx, struct rq *rq, struct task_struct *prev) 
+#else
+int trace_run(struct pt_regs *ctx, struct task_struct *prev) 
+#endif
 {
     u32 pid, tgid;
 
--- a/tools/runqslower.py
+++ b/tools/runqslower.py
@@ -107,7 +107,13 @@ int trace_ttwu_do_wakeup(struct pt_regs *ctx, struct rq *rq, struct task_struct 
 }
 
 // calculate latency
-int trace_run(struct pt_regs *ctx, struct task_struct *prev)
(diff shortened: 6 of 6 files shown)


tools/offwaketime.py
```diff
@@ -188,7 +188,13 @@ def signal_ignore(signal, frame):
     return 0;
 }
 
+#if LINUX_VERSION_CODE < KERNEL_VERSION(3, 19, 0)
+struct rq;
+
+int oncpu(struct pt_regs *ctx, struct rq *rq, struct task_struct *prev) {
```
chenhengqi, line 194: IIUC, `prev` should be renamed as `p`.
Or change the `p` below to `prev` to align with other files.

ethercflow: @chenhengqi Fixed, thanks! :)

brendangregg:
This looks like it sets a bad precedent. The SystemTap tools did this, and became an awful mess of ifdefs to support different kernels.

What's the real problem here? Is this to temporarily support some backported kernels? If so I'd be inclined to not support them.

ethercflow:
Thanks! @brendangregg 

> This looks like it sets a bad precedent. The SystemTap tools did this, and became an awful mess of ifdefs to support different kernels.

Agree, I also don't like mess of ifdefs.

> What's the real problem here? Is this to temporarily support some backported kernels? If so I'd be inclined to not support them.

The problem I face is that many customers still insist on using CentOS 7 to avoid introducing unknown bugs by changing the release version or upgrading the kernel. They are very careful about the installation of diagnostic tools, so they usually do not choose to use SystemTap.

On the other hand, the current BCC installation document describes the support for CentOS 7, and the tools mentioned in the patch can run without any error, but actually parsed the wrong parameters, resulting in no output from these tools. This may cause misjudgment for users who do not know the relevant details.

brendangregg:
Apart from the messyness, it also makes testing worse for everyone. If you're on Ubuntu and you change these tools, do you now need to spin up a CentOS server to check that they still work? And vice versa.

Here's a couple of ideas:
- Someone in the CentOS community could fork bcc and call it "bcc CentOS" or something, and change the tools to be CentOS only (no ifdefs). Then when CentOS catches up to a recent kernel, that fork becomes legacy. This pushes the problem that CentOS created back to the CentOS community.
- We create a subdirectory of "variants", e.g., tools/variants, and then make copies of cpudist.py, offcputime.py, etc. and call them something like cpudist_centos7.py (maybe include more details in the filename to describe the kernel version) and change them directly (no ifdefs). I did this in another tracing project years ago to support older kernel versions, and it was fine. And we already have tools/old (it's been so long I forget why we created it: we should have changed the filenames).

In both cases, we don't have ifdefs and we aren't forcing Ubuntu people to test their changes on both Ubuntu and CentOS and vice versa. It may mean the _centos7.py versions get out of date over time (as new features are added to the main versions) but if this is a temporary measure to support weird CentOS kernel backports then that may be fine.

Personally, I think this is a fundamental problem with the Linux community: that everyone is free to do build whatever they want, without considering the burden for interfacing projects. My attitude is: you're free to build whatever you want, and I'm free not to support it. If that forces you to think about interoperability across Linux, then we are all better off for it.

netedwardwu:
What if we change to use tracepoint instead? For example cpudist.
We can get the same data from tracepoint sched:sched_switch, although we have no "prev->state" in the tracepoint.
```
TRACEPOINT_PROBE(sched, sched_switch) {
    u64 ts = bpf_ktime_get_ns();
    u64 pid_tgid = bpf_get_current_pid_tgid();
    u32 tgid = pid_tgid >> 32, pid = pid_tgid;

    // We can't check task state due to current task state is
    // what we need at sched:sched_switch tracepoint.
    // However have no any BPF helper is supportive.
    // if (prev->state == TASK_RUNNING) {
    if (1)
#ifdef ONCPU
        update_hist(pid, ts);
#else
        store_start(pid, ts);
#endif

BAIL:
#ifdef ONCPU
    store_start(args->next_pid, ts);
#else
    update_hist(args->next_pid, ts);
#endif
    return 0;
}
```
Above BPF program is my tracepoint version, the result is the same as original cpudist

The different parameter in the function is truly a problem.
This case is simple because the unsupported version is really old.
But if someone changes the function we need in the newest kernel version (5.11).
Then it's hard to say we don't support 5.10 but only 5.11

But don't like an awful mess of ifdefs either.

mcaleavya:
Hey

I agree with Brendan on this and I think keeping older tools in
old/variants at least keeps its all in the same place. However If the
overhead of testing these builds at older versions becomes a burden they
should go into their own vintage repository. As far as I recall old were
versions pre BPF_PERF_OUTPUT

Al

ethercflow:
> Apart from the messyness, it also makes testing worse for everyone. If you're on Ubuntu and you change these tools, do you now need to spin up a CentOS server to check that they still work? And vice versa.

> Personally, I think this is a fundamental problem with the Linux community: that everyone is free to do build whatever they want, without considering the burden for interfacing projects. My attitude is: you're free to build whatever you want, and I'm free not to support it. If that forces you to think about interoperability across Linux, then we are all better off for it.

Thanks a lot for sharing your views and attitudes, which made me realize that my thinking is not comprehensive enough. 

> Someone in the CentOS community could fork bcc and call it "bcc CentOS" or something, and change the tools to be CentOS only (no ifdefs). Then when CentOS catches up to a recent kernel, that fork becomes legacy. This pushes the problem that CentOS created back to the CentOS community.

I prefer this one, thanks!

ethercflow:
> What if we change to use tracepoint instead? For example cpudist.
> We can get the same data from tracepoint sched:sched_switch, although we have no "prev->state" in the tracepoint.
> 
> ```
> TRACEPOINT_PROBE(sched, sched_switch) {
>     u64 ts = bpf_ktime_get_ns();
>     u64 pid_tgid = bpf_get_current_pid_tgid();
>     u32 tgid = pid_tgid >> 32, pid = pid_tgid;
> 
>     // We can't check task state due to current task state is
>     // what we need at sched:sched_switch tracepoint.
>     // However have no any BPF helper is supportive.
>     // if (prev->state == TASK_RUNNING) {
>     if (1)
> #ifdef ONCPU
>         update_hist(pid, ts);
> #else
>         store_start(pid, ts);
> #endif
> 
> BAIL:
> #ifdef ONCPU
>     store_start(args->next_pid, ts);
> #else
>     update_hist(args->next_pid, ts);
> #endif
>     return 0;
> }
> ```
> 
> Above BPF program is my tracepoint version, the result is the same as original cpudist
> 
> The different parameter in the function is truly a problem.
> This case is simple because the unsupported version is really old.
> But if someone changes the function we need in the newest kernel version (5.11).
> Then it's hard to say we don't support 5.10 but only 5.11
> 
> But don't like an awful mess of ifdefs either.

Thanks, I have used tp in libbpf-tools to solve the gcc optimization problem, as you said they can work very well.  After reading your discussion, I think it is more appropriate for the CentOS community to adapt each tool. So I decided to close this PR.

ethercflow:
> Hey
> 
> I agree with Brendan on this and I think keeping older tools in old/variants at least keeps its all in the same place. However If the overhead of testing these builds at older versions becomes a burden they should go into their own vintage repository. As far as I recall old were versions pre BPF_PERF_OUTPUT
>
> Al

Thanks for your advise, I think it is more appropriate for the CentOS community to adapt each tool. So I decided to close this PR.

brendangregg:
@netedwardwu thanks, yes, I wrote many of these tools before we had tracepoint support, so some can and should be switched to tracepoints -- which will help with compatibility.

netedwardwu:
Hi @brendangregg ,

I can help cpudist part, but tracepoint sched:sched_switch has no tgid but only pid.
```
__entry->prev_pid	= prev->pid;
```
It was assigned, we can't use shift operation to get the tgid.

So it seems like a problem if we don't sacrifice tgid feature?
Thanks

PR #3351, opened 2021-04-03
libbpf-tools: add gethostlatency

chenhengqi:
```shell
$ sudo ./gethostlatency
TIME        PID    COMM                 LATms      HOST            
22:35:13    694482 node                 4.20       dc.services.visualstudio.com
22:35:17    694482 node                 1.47       default.exp-tas.com
...  
```

### DIFF
--- a/libbpf-tools/gethostlatency.bpf.c
+++ b/libbpf-tools/gethostlatency.bpf.c
@@ -1,0 +1,78 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2021 Hengqi Chen
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_tracing.h>
+#include "gethostlatency.h"
+
+#define MAX_ENTRIES 10240
+
+const volatile pid_t targ_tgid = 0;
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_ENTRIES);
+	__type(key, __u32);
+	__type(value, struct val_t);
+} start SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, sizeof(__u32));
+} events SEC(".maps");
+
+static __always_inline
+int probe_entry(struct pt_regs *ctx) {
+	if (!PT_REGS_PARM1(ctx))
+		return 0;
+
+	struct val_t val = {};
+	__u32 pid = bpf_get_current_pid_tgid() >> 32;
+
+	if (targ_tgid && targ_tgid != pid)
+		return 0;
+
+	if (bpf_get_current_comm(&val.comm, sizeof(val.comm)) == 0) {
+		bpf_probe_read_user(&val.host, sizeof(val.host),
+					   (void *)PT_REGS_PARM1(ctx));
+		val.pid = pid;
+		val.time = bpf_ktime_get_ns();
+		bpf_map_update_elem(&start, &pid, &val, BPF_ANY);
+	}
+
+	return 0;
+}
+
+static __always_inline
+int probe_return(struct pt_regs *ctx) {
+	struct val_t *valp;
+	__u32 pid = bpf_get_current_pid_tgid() >> 32;
+	__u64 now = bpf_ktime_get_ns();
+
+	valp = bpf_map_lookup_elem(&start, &pid);
+	if (!valp)
+		return 0;
+
+	// update time from timestamp to delta
+	valp->time = now - valp->time;
+	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, valp,
+			sizeof(*valp));
+	bpf_map_delete_elem(&start, &pid);
+	return 0;
+}
+
+SEC("kprobe/handle_entry")
+int handle_entry(struct pt_regs *ctx)
+{
+	return probe_entry(ctx);
+}
+
+SEC("kretprobe/handle_return")
+int handle_return(struct pt_regs *ctx)
+{
+	return probe_return(ctx);
+}
+
+char LICENSE[] SEC("license") = "GPL";
--- a/libbpf-tools/gethostlatency.c
+++ b/libbpf-tools/gethostlatency.c
@@ -1,0 +1,285 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+// Copyright (c) 2021 Hengqi Chen
+//
+// Based on gethostlatency(8) from BCC by Brendan Gregg.
+// 24-Mar-2021   Hengqi Chen   Created this.
+#include <argp.h>
+#include <errno.h>
+#include <signal.h>
+#include <time.h>
+
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+#include "gethostlatency.h"
+#include "gethostlatency.skel.h"
+#include "trace_helpers.h"
+#include "uprobe_helpers.h"
+
+#define PERF_BUFFER_PAGES 16
+#define PERF_POLL_TIMEOUT_MS 100
+#define warn(...) fprintf(stderr, __VA_ARGS__)
+
+volatile sig_atomic_t canceled = 0;
+pid_t traced_pid = 0;
+
+const char *argp_program_version = "gethostlatency 0.1";
+const char *argp_program_bug_address =
+	"https://github.com/iovisor/bcc/tree/master/libbpf-tools";
+const char argp_program_doc[] =
+"Show latency for getaddrinfo/gethostbyname[2] calls.\n"
+"\n"
+"USAGE: gethostlatency [-h] [-p PID]\n"
+"\n"
+"EXAMPLES:\n"
+"    gethostlatency             # time getaddrinfo/gethostbyname[2] calls\n"
+"    gethostlatency -p 1216     # only trace PID 1216\n";
+
+static const struct argp_option opts[] = {
+	{"pid", 'p', "PID", 0, "Process ID to trace"},
+	{NULL, 'h', NULL, OPTION_HIDDEN, "Show the full help"},
+	{},
+};
+
+static error_t parse_arg(int key, char *arg, struct argp_state *state)
+{
+	long pid;
+
+	switch (key) {
+	case 'p':
+		errno = 0;
+		pid = strtol(arg, NULL, 10);
+		if (errno || pid <= 0) {
+			warn("Invalid PID: %s\n", arg);
+			argp_usage(state);
+		}
+		traced_pid = pid;
+		break;
+	case 'h':
+		argp_state_help(state, stderr, ARGP_HELP_STD_HELP);
+		break;
+	default:
+		return ARGP_ERR_UNKNOWN;
+	}
+	return 0;
+}
+
+static void sig_int(int signo)
+{
+	canceled = 1;
+}
+
+static const char *strftime_now(char *s, size_t max, const char *format)
+{
+	struct tm *tm;
+	time_t t;
+
+	t = time(NULL);
+	tm = localtime(&t);
+	if (tm == NULL) {
+		warn("localtime: %s\n", strerror(errno));
+		return "<failed>";
+	}
+	if (strftime(s, max, format, tm) == 0) {
+		warn("strftime error\n");
+		return "<failed>";
+	}
+	return s;
+}
+
+void handle_event(void *ctx, int cpu, void *data, __u32 data_sz)
+{
+	const struct val_t *e = data;
+	char s[16] = {};
+	const char *now;
+
+	now = strftime_now(s, sizeof(s), "%H:%M:%S");
+	printf("%-11s %-10d %-20s %-10.2f %-16s\n",
+		now, e->pid, e->comm, (double)e->time/1000000, e->host);
+}
+
+void handle_lost_events(void *ctx, int cpu, __u64 lost_cnt)
+{
+	warn("lost %llu events on CPU #%d\n", lost_cnt, cpu);
+}
+
+static int get_libc_path(char *path)
+{
+	FILE *f;
+	char buf[256] = {};
+	char *filename;
+	float version;
+
+	f = fopen("/proc/self/maps", "r");
+	if (!f) {
+		return -errno;
+	}
+
+	while (fscanf(f, "%*x-%*x %*s %*s %*s %*s %[^\n]\n", buf) != EOF) {
+		if (strchr(buf, '/') != buf) {
+			continue;
+		}
+		filename = strrchr(buf, '/') + 1;
+		if (sscanf(filename, "libc-%f.so", &version) == 1) {
+			memcpy(path, buf, strlen(buf));
+			fclose(f);
+			return 0;
+		}
+	}
+
+	fclose(f);
+	return -1;
+}
+
+static int attach_uprobes(struct gethostlatency_bpf *obj)
+{
+	int err;
+	char libc_path[PATH_MAX] = {};
+	off_t func_off;
+
+	err = get_libc_path(libc_path);
+	if (err) {
+		warn("could not find libc.so\n");
+		return -1;
+	}
+
+	func_off = get_elf_func_offset(libc_path, "getaddrinfo");
+	if (func_off < 0) {
+		warn("could not find getaddrinfo in %s\n", libc_path);
+		return -1;
+	}
+	obj->links.handle_entry =
+		bpf_program__attach_uprobe(obj->progs.handle_entry, false,
+					   traced_pid ?: -1, libc_path, func_off);
+	err = libbpf_get_error(obj->links.handle_entry);
+	if (err) {
+		warn("failed to attach getaddrinfo: %d\n", err);
+		return -1;
+	}
+	obj->links.handle_return =
+		bpf_program__attach_uprobe(obj->progs.handle_return, true,
+					   traced_pid ?: -1, libc_path, func_off);
+	err = libbpf_get_error(obj->links.handle_return);
+	if (err) {
+		warn("failed to attach getaddrinfo: %d\n", err);
+		return -1;
+	}
+
+	func_off = get_elf_func_offset(libc_path, "gethostbyname");
+	if (func_off < 0) {
+		warn("Could not find gethostbyname in %s\n", libc_path);
+		return -1;
+	}
+	obj->links.handle_entry =
+		bpf_program__attach_uprobe(obj->progs.handle_entry, false,
+					   traced_pid ?: -1, libc_path, func_off);
+	err = libbpf_get_error(obj->links.handle_entry);
+	if (err) {
+		warn("failed to attach gethostbyname: %d\n", err);
+		return -1;
+	}
+	obj->links.handle_return =
+		bpf_program__attach_uprobe(obj->progs.handle_return, true,
+					   traced_pid ?: -1, libc_path, func_off);
+	err = libbpf_get_error(obj->links.handle_return);
+	if (err) {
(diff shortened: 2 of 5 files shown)


danobi (commented):
Didn't look past the ld.so.cache parsing -- maybe we can discuss that first:

I'm not sure you're parsing the old/new versions correctly. I'm looking at glibc's code ( https://github.com/bminor/glibc/blob/595c22ecd8e87a27fd19270ed30fdbae9ad25426/elf/cache.c#L316 ) and I think the new/old formats require 2 completely different code paths.

Regardless, I wonder if parsing ld.so.cache is the best solution. On the surface it seems more reliable but I wonder about that. I deleted `/etc/ld.so.cache` from my system and things still work OK. I think the dynamic loader has a slow path where it manually finds all the .so's in the system. And your code doesn't have this slow path.

Maybe we can hard code a few common libc locations on the system and take the first one that exists. And also provide a command line arg to override libc path for uncommon systems. That ought to simplify things a great deal.

anakryiko:
So I was going to propose to just run `ldconfig -p` instead of parsing binary /etc/ld.so.cache, but if one removes /etc/ld.so.cache then `ldconfig -p` fails, so that's not great. But hard-coding a list of paths to check also seems suboptimal.

What do you guys think about one of these two:

a) this should be done without shelling out, I'm just demonstrating with shell command

```
$ sudo cat /proc/self/maps | grep libc | head -n1
7f2c0c7a2000-7f2c0c966000 r-xp 00000000 08:04 21238613                   /usr/lib64/libc-2.17.so
```

b) this will have to shell out, unfortunately (but we'll use `/proc/self/exe`, of course):

```
$ sudo ldd /proc/1077168/exe
        linux-vdso.so.1 =>  (0x00007fff832b8000)
        libelf.so.1 => /lib64/libelf.so.1 (0x00007fc72223f000)
        libz.so.1 => /lib64/libz.so.1 (0x00007fc722029000)
        libc.so.6 => /lib64/libc.so.6 (0x00007fc721c5b000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fc722457000)
```
I prefer a), if that indeed covers this use case completely.

libbpf-tools/gethostlatency.bpf.c
```diff
@@ -0,0 +1,112 @@
...
+
+SEC("uretprobe/c/gethostbyname")
+int BPF_PROG(gethostbyname_return)
+{
+	probe_return((void *)ctx);
+	return 0;
+}
+
+SEC("uretprobe/c/gethostbyname2")
+int BPF_PROG(gethostbyname2_return)
```
anakryiko, line 106: you don't really need all these program copies, it's enough to have one uprobe and one uretprobe program and attach them to multiple functions at runtime.

also BPF_PROG is not intended for uprobe/uretprobe, it's for fentry/fexit and alike, so please just do

```
int hanlde_entry(struct pt_regs *ctx) { ... }
```

chenhengqi: Thanks for the insight.

chenhengqi:
> So I was going to propose to just run `ldconfig -p` instead of parsing binary /etc/ld.so.cache, but if one removes /etc/ld.so.cache then `ldconfig -p` fails, so that's not great. But hard-coding a list of paths to check also seems suboptimal.
> 
> What do you guys think about one of these two:
> 
> a) this should be done without shelling out, I'm just demonstrating with shell command
> 
> ```
> $ sudo cat /proc/self/maps | grep libc | head -n1
> 7f2c0c7a2000-7f2c0c966000 r-xp 00000000 08:04 21238613                   /usr/lib64/libc-2.17.so
> ```
> 
> b) this will have to shell out, unfortunately (but we'll use `/proc/self/exe`, of course):
> 
> ```
> $ sudo ldd /proc/1077168/exe
>         linux-vdso.so.1 =>  (0x00007fff832b8000)
>         libelf.so.1 => /lib64/libelf.so.1 (0x00007fc72223f000)
>         libz.so.1 => /lib64/libz.so.1 (0x00007fc722029000)
>         libc.so.6 => /lib64/libc.so.6 (0x00007fc721c5b000)
>         /lib64/ld-linux-x86-64.so.2 (0x00007fc722457000)
> ```
> 
> I prefer a), if that indeed covers this use case completely.

Have seen something like this in libbpf-bootstrap's uprobe example. I will have a try.

chenhengqi:
https://github.com/iovisor/bcc/blob/0cf8166505ba89f14545d6fc54d088aad127d870/libbpf-tools/funclatency.c#L8

BTW, I got the idea of parsing `ld.so.cache` from `funclatency` tool.

libbpf-tools/ldcache_helpers.c
```diff
@@ -0,0 +1,207 @@
...
+			return NULL;
+		}
+	}
+
+	pos = (char *)align(pos, __alignof__(struct header_new));
+	if (pos >= data + size) {
+		return NULL;
+	}
+
+	hdr_new = (struct header_new *)pos;
```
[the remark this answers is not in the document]
chenhengqi: There are three formats for `ld.so.cache`:
```c
enum opt_format
    {
      opt_format_old = 0, /* Use struct cache_file.  */
      opt_format_compat = 1, /* Use both, old format followed by new.  */
      opt_format_new = 2, /* Use struct cache_file_new.  */
    };
```
Current implementation supports last two format.

libbpf-tools/ldcache_helpers.c
```diff
@@ -0,0 +1,207 @@
...
+
+static struct ldcache *parse(const char *data, long size, int old) {
+	char *pos, *strs;
+	struct header *hdr;
+	struct header_new *hdr_new;
+	struct entry_new *entry_new;
+	struct ldcache *cache;
+	int i, n;
+
+	pos = (char *)data;
```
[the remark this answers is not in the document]
chenhengqi: Ah, casting here to avoid compilation warning.

libbpf-tools/ldcache_helpers.c
```diff
@@ -0,0 +1,207 @@
...
+	if (!data) {
+		fclose(f);
+		return NULL;
+	}
+	n = fread(data, 1, size, f);
+	if (!n) {
+		fclose(f);
+		return NULL;
+	}
+	fclose(f);
```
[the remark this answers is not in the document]
chenhengqi: I will drop these files.

chenhengqi:
@anakryiko @danobi 

Hello guys, I've updated the program as you suggested. PTAL. Thanks.

libbpf-tools/gethostlatency.c
```diff
@@ -0,0 +1,286 @@
...
+	struct gethostlatency_bpf *obj;
+	int err;
+
+	err = argp_parse(&argp, argc, argv, 0, NULL, NULL);
+	if (err)
+		return err;
+
+	err = bump_memlock_rlimit();
+	if (err) {
+		fprintf(stderr, "failed to increase rlimit: %d\n", err);
```
danobi, line 232: Should this just use `warn()`?

libbpf-tools/gethostlatency.c
```diff
@@ -0,0 +1,286 @@
...
+		return 1;
+	}
+
+	obj = gethostlatency_bpf__open();
+	if (!obj) {
+		fprintf(stderr, "failed to open BPF object\n");
```
danobi, line 238: Same as above

libbpf-tools/gethostlatency.c
```diff
@@ -0,0 +1,286 @@
...
+		return 1;
+	}
+
+	obj->rodata->targ_tgid = traced_pid;
+
+	err = gethostlatency_bpf__load(obj);
+	if (err) {
+		fprintf(stderr, "failed to load BPF object: %d\n", err);
```
danobi, line 246: Same as above

libbpf-tools/gethostlatency.c
```diff
@@ -0,0 +1,286 @@
...
+#include "gethostlatency.h"
+#include "gethostlatency.skel.h"
+#include "trace_helpers.h"
+#include "uprobe_helpers.h"
+
+#define PERF_BUFFER_PAGES 16
+#define PERF_POLL_TIMEOUT_MS 100
+#define warn(...) fprintf(stderr, __VA_ARGS__)
+
+int canceled = 0;
```
danobi, line 22: Not sure how much it actually matters in practice, but this should be a `volatile sig_atomic_t cancelled;`. See https://wiki.sei.cmu.edu/confluence/display/c/SIG31-C.+Do+not+access+shared+objects+in+signal+handlers

libbpf-tools/gethostlatency.c
```diff
@@ -0,0 +1,286 @@
...
+	printf("%-11s %-6s %-20s %-10s %-16s\n",
+		"TIME", "PID", "COMM", "LATms", "HOST");
+
+	while (1) {
+		if ((err = perf_buffer__poll(pb, PERF_POLL_TIMEOUT_MS)) < 0)
+			break;
+		if (canceled)
+			goto cleanup;
+	}
+	printf("error polling perf buffer: %d\n", err);
```
danobi, line 280: Use `warn()` instead?

libbpf-tools/gethostlatency.bpf.c
```diff
@@ -0,0 +1,82 @@
...
+	data.pid = valp->pid;
+	data.delta = tsp - valp->ts;
+	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, &data,
+			sizeof(data));
+	bpf_map_delete_elem(&start, &pid);
+	return 0;
+}
+
+SEC("kprobe/hanlde_entry")
+int hanlde_entry(struct pt_regs *ctx)
```
danobi, line 71: Nit: spelling
```suggestion
SEC("kprobe/handle_entry")
int handle_entry(struct pt_regs *ctx)
```

libbpf-tools/gethostlatency.bpf.c
```diff
@@ -0,0 +1,82 @@
...
+{
+	return probe_entry(ctx);
+}
+
+SEC("kretprobe/hanlde_return")
+int hanlde_return(struct pt_regs *ctx)
```
danobi, line 77: Ditto
```suggestion
SEC("kretprobe/handle_return")
int handle_return(struct pt_regs *ctx)
```

libbpf-tools/gethostlatency.bpf.c
```diff
@@ -0,0 +1,82 @@
...
+	__uint(value_size, sizeof(u32));
+} events SEC(".maps");
+
+static __always_inline
+int probe_entry(struct pt_regs *ctx) {
+	if (!PT_REGS_PARM1(ctx))
+		return 0;
+
+	struct val_t val = {};
+	u32 pid = bpf_get_current_pid_tgid();
```
danobi, line 32: Maybe nicer to explicitly take the bottom 32 bits
```suggestion
	u32 pid = bpf_get_current_pid_tgid() & 0xFFFFFFFF;
```

libbpf-tools/gethostlatency.bpf.c
```diff
@@ -0,0 +1,82 @@
...
+
+	return 0;
+}
+
+static __always_inline
+int probe_return(struct pt_regs *ctx) {
+	struct val_t *valp;
+	struct data_t data = {};
+	u64 delta;
+	u32 pid = bpf_get_current_pid_tgid();
```
danobi, line 53: Same as above

libbpf-tools/gethostlatency.bpf.c
```diff
@@ -0,0 +1,82 @@
...
+		return 0;
+
+	struct val_t val = {};
+	u32 pid = bpf_get_current_pid_tgid();
+
+	if (targ_tgid && targ_tgid != pid)
+		return 0;
+
+	if (bpf_get_current_comm(&val.comm, sizeof(val.comm)) == 0) {
+		bpf_probe_read_user(&val.host, sizeof(val.host),
```
danobi, line 38: Check return value here too?

libbpf-tools/gethostlatency.h
```diff
@@ -0,0 +1,21 @@
...
+	char comm[TASK_COMM_LEN];
+	char host[80];
+	__u64 ts;
+};
+
+struct data_t {
+	__u32 pid;
+	__u64 delta;
+	char comm[TASK_COMM_LEN];
+	char host[80];
```
danobi, line 18: Not a big deal, but I think you can probably reuse `val_t` and save an extra 2 probe_read(). You can just fixup `ts` to actually hold `delta` before transmitting to userspace.

libbpf-tools/gethostlatency.c
```diff
@@ -0,0 +1,286 @@
...
+
+	err = gethostlatency_bpf__load(obj);
+	if (err) {
+		fprintf(stderr, "failed to load BPF object: %d\n", err);
+		goto cleanup;
+	}
+
+	err = attach_uprobes(obj);
+	if (err) {
+		warn("failed to attach BPF programs: %s\n", strerror(-err));
```
danobi, line 252: `attach_uprobes()` seems to return -1 for errors, so I don't think calling strerror() is right here. Could probably just drop the error message from this warning -- attach_uprobes() already prints out a helpful message

libbpf-tools/gethostlatency.bpf.c
chenhengqi: I am afraid that this is a misuse of bpf_get_current_pid_tgid, I will file a issue and let's discuss there

libbpf-tools/gethostlatency.bpf.c
chenhengqi: Most tools in this repo do NOT check return value of `bpf_probe_read_user` and `bpf_probe_read_kernel`, how about just keep consistent with the code base ?

libbpf-tools/gethostlatency.h
chenhengqi: Nice idea. Will do.

libbpf-tools/gethostlatency.bpf.c
danobi: What do you mean? I think either version is technically correct

libbpf-tools/gethostlatency.bpf.c
danobi: Sure, I guess it doesn't hurt. An empty string doesn't hurt.

libbpf-tools/gethostlatency.bpf.c
```diff
@@ -0,0 +1,78 @@
...
+	__uint(value_size, sizeof(__u32));
+} events SEC(".maps");
+
+static __always_inline
+int probe_entry(struct pt_regs *ctx) {
+	if (!PT_REGS_PARM1(ctx))
+		return 0;
+
+	struct val_t val = {};
+	__u32 pid = bpf_get_current_pid_tgid() >> 32;
```
danobi, line 32: You changed this from the last version, right? I think tid is also OK. Each thread can have its own comm so it might be confusing to print out tgid instead of tid.

chenhengqi: Yes, I changed it.

There are more than 70 tools in this repo use `bpf_get_current_pid_tgid` and they expose option named `PID`.

43 of them make explicit conversion using `bpf_get_current_pid_tgid >> 32`, the rest simply using something like `__u32 pid = bpf_get_current_pid_tgid()`, I suspect some of them misuse the `bpf_get_current_pid_tgid` helper.

chenhengqi:
@danobi  Can we get this landed ?

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

drzaeus77:
[buildbot, test this please]

libbpf-tools/gethostlatency.bpf.c
evverx: > it's enough to have one uprobe and one uretprobe program and attach them to multiple functions at runtime

@anakryiko @chenhengqi I ran `gethostlatency` under ASan and it reported several memory leaks:
```
==14089==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 56 byte(s) in 1 object(s) allocated from:
    #0 0x7f03aa453527 in __interceptor_calloc (/lib64/libasan.so.6+0xab527)
    #1 0x46c65d in bpf_program__attach_perf_event_opts /home/vagrant/bcc/src/cc/libbpf/src/libbpf.c:9163
    #2 0x470cca in bpf_program__attach_uprobe_opts /home/vagrant/bcc/src/cc/libbpf/src/libbpf.c:9668
    #3 0x471271 in bpf_program__attach_uprobe /home/vagrant/bcc/src/cc/libbpf/src/libbpf.c:9699
    #4 0x405f97 in attach_uprobes /home/vagrant/bcc/libbpf-tools/gethostlatency.c:162
    #5 0x406db3 in main /home/vagrant/bcc/libbpf-tools/gethostlatency.c:253
    #6 0x7f03a98651e1 in __libc_start_main (/lib64/libc.so.6+0x281e1)

Direct leak of 56 byte(s) in 1 object(s) allocated from:
    #0 0x7f03aa453527 in __interceptor_calloc (/lib64/libasan.so.6+0xab527)
    #1 0x46c65d in bpf_program__attach_perf_event_opts /home/vagrant/bcc/src/cc/libbpf/src/libbpf.c:9163
    #2 0x470cca in bpf_program__attach_uprobe_opts /home/vagrant/bcc/src/cc/libbpf/src/libbpf.c:9668
    #3 0x471271 in bpf_program__attach_uprobe /home/vagrant/bcc/src/cc/libbpf/src/libbpf.c:9699
    #4 0x40622c in attach_uprobes /home/vagrant/bcc/libbpf-tools/gethostlatency.c:176
    #5 0x406db3 in main /home/vagrant/bcc/libbpf-tools/gethostlatency.c:253
    #6 0x7f03a98651e1 in __libc_start_main (/lib64/libc.so.6+0x281e1)

Direct leak of 56 byte(s) in 1 object(s) allocated from:
    #0 0x7f03aa453527 in __interceptor_calloc (/lib64/libasan.so.6+0xab527)
    #1 0x46c65d in bpf_program__attach_perf_event_opts /home/vagrant/bcc/src/cc/libbpf/src/libbpf.c:9163
    #2 0x470cca in bpf_program__attach_uprobe_opts /home/vagrant/bcc/src/cc/libbpf/src/libbpf.c:9668
    #3 0x471271 in bpf_program__attach_uprobe /home/vagrant/bcc/src/cc/libbpf/src/libbpf.c:9699
    #4 0x40643e in attach_uprobes /home/vagrant/bcc/libbpf-tools/gethostlatency.c:184
    #5 0x406db3 in main /home/vagrant/bcc/libbpf-tools/gethostlatency.c:253
    #6 0x7f03a98651e1 in __libc_start_main (/lib64/libc.so.6+0x281e1)

Direct leak of 56 byte(s) in 1 object(s) allocated from:
    #0 0x7f03aa453527 in __interceptor_calloc (/lib64/libasan.so.6+0xab527)
    #1 0x46c65d in bpf_program__attach_perf_event_opts /home/vagrant/bcc/src/cc/libbpf/src/libbpf.c:9163
    #2 0x470cca in bpf_program__attach_uprobe_opts /home/vagrant/bcc/src/cc/libbpf/src/libbpf.c:9668
    #3 0x471271 in bpf_program__attach_uprobe /home/vagrant/bcc/src/cc/libbpf/src/libbpf.c:9699
    #4 0x405d85 in attach_uprobes /home/vagrant/bcc/libbpf-tools/gethostlatency.c:154
    #5 0x406db3 in main /home/vagrant/bcc/libbpf-tools/gethostlatency.c:253
    #6 0x7f03a98651e1 in __libc_start_main (/lib64/libc.so.6+0x281e1)

SUMMARY: AddressSanitizer: 224 byte(s) leaked in 4 allocation(s).
```
Looks like either `libbpf` can't free the memory allocated by `bpf_program__attach_uprobe` or `bpf_program__attach_uprobe` isn't used correctly.

chenhengqi: @evverx Thanks for the report. Will take a look at this.

PR #3360, opened 2021-04-09
Fix ttysnoop for kernel > v5.11 and other config changes

olsajiri:
- Kernel commit [1] changed arguments of tty_write function, changing the probe function to new prototypes.

- Use array map to store data

- Adding the possibility to define transmitting data size (--datasize option) and number of times we ask for this amount (--datacount option).

### DIFF
--- a/tools/ttysnoop.py
+++ b/tools/ttysnoop.py
@@ -62,6 +62,7 @@ except:
 bpf_text = """
 #include <uapi/linux/ptrace.h>
 #include <linux/fs.h>
+#include <linux/uio.h>
 
 #define BUFSIZE 256
 struct data_t {
@@ -71,12 +72,8 @@ struct data_t {
 
 BPF_PERF_OUTPUT(events);
 
-int kprobe__tty_write(struct pt_regs *ctx, struct file *file,
-    const char __user *buf, size_t count)
+static int do_tty_write(void *ctx, const char __user *buf, size_t count)
 {
-    if (file->f_inode->i_ino != PTS)
-        return 0;
-
     // bpf_probe_read_user() can only use a fixed size, so truncate to count
     // in user space:
     struct data_t data = {};
@@ -89,6 +86,40 @@ int kprobe__tty_write(struct pt_regs *ctx, struct file *file,
 
     return 0;
 };
+
+/**
+ * commit 9bb48c82aced (v5.11-rc4) tty: implement write_iter
+ * changed arguments of tty_write function
+ */
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5, 11, 0)
+int kprobe__tty_write(struct pt_regs *ctx, struct file *file,
+    const char __user *buf, size_t count)
+{
+    if (file->f_inode->i_ino != PTS)
+        return 0;
+
+    return do_tty_write(ctx, buf, count);
+}
+#else
+KFUNC_PROBE(tty_write, struct kiocb *iocb, struct iov_iter *from)
+{
+    const char __user *buf;
+    const struct kvec *kvec;
+    size_t count;
+
+    if (iocb->ki_filp->f_inode->i_ino != PTS)
+        return 0;
+
+    if (from->type != (ITER_IOVEC + WRITE))
+        return 0;
+
+    kvec  = from->kvec;
+    buf   = kvec->iov_base;
+    count = kvec->iov_len;
+
+    return do_tty_write(ctx, kvec->iov_base, kvec->iov_len);
+}
+#endif
 """
 
 bpf_text = bpf_text.replace('PTS', str(pi.st_ino))
--- a/tools/ttysnoop.py
+++ b/tools/ttysnoop.py
@@ -70,19 +70,33 @@ struct data_t {
     char buf[BUFSIZE];
 };
 
+BPF_ARRAY(data_map, struct data_t, 1);
 BPF_PERF_OUTPUT(events);
 
 static int do_tty_write(void *ctx, const char __user *buf, size_t count)
 {
+    int zero = 0;
+    struct data_t *data;
+
+/* We can't read data to map data before v4.11 */
+#if LINUX_VERSION_CODE < KERNEL_VERSION(4, 11, 0)
+    struct data_t _data = {};
+
+    data = &_data;
+#else
+    data = data_map.lookup(&zero);
+    if (!data)
+        return 0;
+#endif
+
     // bpf_probe_read_user() can only use a fixed size, so truncate to count
     // in user space:
-    struct data_t data = {};
-    bpf_probe_read_user(&data.buf, BUFSIZE, (void *)buf);
+    bpf_probe_read_user(&data->buf, BUFSIZE, (void *)buf);
     if (count > BUFSIZE)
-        data.count = BUFSIZE;
+        data->count = BUFSIZE;
     else
-        data.count = count;
-    events.perf_submit(ctx, &data, sizeof(data));
+        data->count = count;
+    events.perf_submit(ctx, data, sizeof(*data));
 
     return 0;
 };
--- a/tools/ttysnoop.py
+++ b/tools/ttysnoop.py
@@ -28,10 +28,13 @@ def usage():
 
 # arguments
 examples = """examples:
-    ./ttysnoop /dev/pts/2    # snoop output from /dev/pts/2
-    ./ttysnoop 2             # snoop output from /dev/pts/2 (shortcut)
-    ./ttysnoop /dev/console  # snoop output from the system console
-    ./ttysnoop /dev/tty0     # snoop output from /dev/tty0
+    ./ttysnoop /dev/pts/2          # snoop output from /dev/pts/2
+    ./ttysnoop 2                   # snoop output from /dev/pts/2 (shortcut)
+    ./ttysnoop /dev/console        # snoop output from the system console
+    ./ttysnoop /dev/tty0           # snoop output from /dev/tty0
+    ./ttysnoop /dev/pts/2 -s 1024  # snoop output from /dev/pts/2 with data size 1024
+    ./ttysnoop /dev/pts/2 -c 2     # snoop output from /dev/pts/2 with 2 checks for 256 bytes of data in buffer
+                                     (potentially retrieving 512 bytes)
 """
 parser = argparse.ArgumentParser(
     description="Snoop output from a pts or tty device, eg, a shell",
@@ -41,6 +44,10 @@ parser.add_argument("-C", "--noclear", action="store_true",
     help="don't clear the screen")
 parser.add_argument("device", default="-1",
     help="path to a tty device (eg, /dev/tty0) or pts number")
+parser.add_argument("-s", "--datasize", default="256",
+    help="size of the transmitting buffer (default 256)")
+parser.add_argument("-c", "--datacount", default="16",
+    help="number of times we check for 'data-size' data (default 16)")
 parser.add_argument("--ebpf", action="store_true",
     help=argparse.SUPPRESS)
 args = parser.parse_args()
@@ -64,7 +71,7 @@ bpf_text = """
 #include <linux/fs.h>
 #include <linux/uio.h>
 
-#define BUFSIZE 256
+#define BUFSIZE USER_DATASIZE
 struct data_t {
     int count;
     char buf[BUFSIZE];
@@ -75,7 +82,7 @@ BPF_PERF_OUTPUT(events);
 
 static int do_tty_write(void *ctx, const char __user *buf, size_t count)
 {
-    int zero = 0;
+    int zero = 0, i;
     struct data_t *data;
 
 /* We can't read data to map data before v4.11 */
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, test this please]

olsajiri:
I removed the trampoline and kept kprobe for both pre/post 5.11 case, let's see ;-)

olsajiri:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

tools/ttysnoop.py
```diff
@@ -28,10 +28,13 @@ def usage():
...
-    ./ttysnoop 2             # snoop output from /dev/pts/2 (shortcut)
-    ./ttysnoop /dev/console  # snoop output from the system console
-    ./ttysnoop /dev/tty0     # snoop output from /dev/tty0
+    ./ttysnoop /dev/pts/2           # snoop output from /dev/pts/2
+    ./ttysnoop 2                    # snoop output from /dev/pts/2 (shortcut)
+    ./ttysnoop /dev/console         # snoop output from the system console
+    ./ttysnoop /dev/tty0            # snoop output from /dev/tty0
+    ./ttysnoop /dev/pts/2  -s 1024  # snoop output from /dev/pts/2 with data size 1024
+    ./ttysnoop /dev/pts/2  -c 2     # snoop output from /dev/pts/2 with 2 checks for 4096 bytes of data in buffer
+                                      (potentially retrieving 8192 bytes)
```
yonghong-song, line 37: could you explain why '-c' option is needed? I see the default below is 20? Why choose 20?

tools/ttysnoop.py
```diff
@@ -41,6 +44,10 @@ def usage():
     help="don't clear the screen")
 parser.add_argument("device", default="-1",
     help="path to a tty device (eg, /dev/tty0) or pts number")
+parser.add_argument("-s", "--datasize", default="4096",
+    help="size of the transmitting buffer (default 4096)")
```
yonghong-song, line 48: Previous default is 256, now we increased to 4096, maybe in commit messages on why such an increase?

tools/ttysnoop.py
```diff
@@ -62,33 +69,75 @@ def usage():
...
+{
+    int zero = 0, i;
+    struct data_t *data;
+
+    data = data_map.lookup(&zero);
+    if (!data)
+        return 0;
+
+    for (i = 0; i < USER_DATACOUNT; i++) {
+        bpf_probe_read_user(&data->buf, BUFSIZE, (void *)buf);
```
yonghong-song, line 93: I did not see `buf` pointer change. That means, every iteration, you try to read the same source. Is this what you want? Also, we can use bpf_probe_read_user() return value which should tells you precisely the bytes read.

tools/ttysnoop.py
```diff
@@ -62,33 +69,75 @@ def usage():
...
-};
+    if (iocb->ki_filp->f_inode->i_ino != PTS)
+        return 0;
+
+    if (from->type != (ITER_IOVEC + WRITE))
+        return 0;
+
+    kvec  = from->kvec;
+    buf   = kvec->iov_base;
+    count = kvec->iov_len;
```
yonghong-song, line 136: buf and count are not used.

tools/ttysnoop_example.txt
```diff
@@ -73,11 +73,16 @@ positional arguments:
   device         path to a tty device (eg, /dev/tty0) or pts number
 
 optional arguments:
-  -h, --help     show this help message and exit
-  -C, --noclear  don't clear the screen
+  -h, --help      show this help message and exit
+  -C, --noclear   don't clear the screen
+  -s, --datasize  size of the transmitting buffer (default 4096)
+  -c, --datacount number of times we check for 'datasize' data (default 20)
```
yonghong-song, line 79: Please add these new options to man page as well.

yonghong-song (commented):
I think kprobe is fine here. tty_write probably not performance critical part of the kernel. If we want, we can add kfunc support later.

yonghong-song:
for ubuntu 16.04, I see the failure:
```
30: back-edge from insn 61 to 34
30: 
30: Traceback (most recent call last):
30:   File "../../tools/ttysnoop.py", line 153, in <module>
30:     b = BPF(text=bpf_text)
30:   File "/home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1604/build/src/python/bcc-python/bcc/__init__.py", line 373, in __init__
30:     self._trace_autoload()
30:   File "/home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1604/build/src/python/bcc-python/bcc/__init__.py", line 1236, in _trace_autoload
30:     fn = self.load_func(func_name, BPF.KPROBE)
30:   File "/home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1604/build/src/python/bcc-python/bcc/__init__.py", line 413, in load_func
30:     (func_name, errstr))
30: Exception: Failed to load BPF program kprobe__tty_write: Invalid argument
```

You probably need to fully unroll the loop (with clang pragma) for old kernels.

olsajiri:
I'll make the fixes and resubmit, thanks

tools/ttysnoop.py
olsajiri: ok, I kept the 256, with the buf fix you pointed out, I'm getting nice snoop output on ncurses apps as well

tools/ttysnoop.py
olsajiri: ugh, yep.. fixed ;-)

tools/ttysnoop.py
olsajiri: leftover from trampoline removal, fixed, thanks

tools/ttysnoop_example.txt
olsajiri: ok

tools/ttysnoop.py
olsajiri: with most 'big bcurses' apps I'm getting 4096 as maximum for count,
so I guess it's the limit somewhere in the kernel code, I can change the default to be closer to that

yonghong-song:
We have one error for ubuntu16.04 (kernel 4.4.0),
```
30: 0: (bf) r6 = r1
30: 1: (79) r8 = *(u64 *)(r6 +96)
30: 2: (79) r9 = *(u64 *)(r6 +104)
30: 3: (79) r3 = *(u64 *)(r6 +112)
30: 4: (b7) r7 = 0
30: 5: (7b) *(u64 *)(r10 -8) = r7
30: 6: (7b) *(u64 *)(r10 -16) = r7
30: 7: (07) r3 += 32
30: 8: (bf) r1 = r10
30: 9: (07) r1 += -16
30: 10: (b7) r2 = 8
30: 11: (85) call 4
30: 12: (79) r3 = *(u64 *)(r10 -16)
30: 13: (07) r3 += 64
30: 14: (bf) r1 = r10
30: 15: (07) r1 += -8
30: 16: (b7) r2 = 8
30: 17: (85) call 4
30: 18: (79) r1 = *(u64 *)(r10 -8)
30: 19: (55) if r1 != 0xe goto pc+582
30:  R0=inv R1=imm14 R6=ctx R7=imm0 R8=inv R9=inv R10=fp
30: 20: (7b) *(u64 *)(r10 -24) = r9
30: 21: (63) *(u32 *)(r10 -8) = r7
30: 22: (18) r1 = 0xffff8800bb905000
30: 24: (bf) r2 = r10
30: 25: (07) r2 += -8
30: 26: (85) call 1
30: 27: (15) if r0 == 0x0 goto pc+574
30:  R0=map_value(ks=4,vs=260) R6=ctx R7=imm0 R8=inv R9=inv R10=fp
30: 28: (bf) r1 = r0
30: 29: (07) r1 += 4
30: 30: (7b) *(u64 *)(r10 -32) = r1
30: 31: (b7) r9 = 256
30: 32: (bf) r7 = r0
30: 33: (b7) r2 = 256
30: 34: (79) r3 = *(u64 *)(r10 -24)
30: 35: (85) call 4
30: R1 type=inv expected=fp
30: 
30: Traceback (most recent call last):
30:   File "../../tools/ttysnoop.py", line 153, in <module>
30:     b = BPF(text=bpf_text)
30:   File "/home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1604/build/src/python/bcc-python/bcc/__init__.py", line 373, in __init__
30:     self._trace_autoload()
30:   File "/home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1604/build/src/python/bcc-python/bcc/__init__.py", line 1236, in _trace_autoload
30:     fn = self.load_func(func_name, BPF.KPROBE)
30:   File "/home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1604/build/src/python/bcc-python/bcc/__init__.py", line 413, in load_func
30:     (func_name, errstr))
30: Exception: Failed to load BPF program kprobe__tty_write: Permission denied
```
perf_event.submit in 4.4 expects a fp for data, but later kernel supports map value for data. Hence we have the above error.

In this case, we add a fix for newer kernel (5.11) and unfortunately it MAY make 4.4 kernel not working.

tools/ttysnoop.py
```diff
@@ -62,33 +69,75 @@ def usage():
...
+    }
+
+    return 0;
+};
+
+/**
+ * commit 9bb48c82aced (v5.11-rc4) tty: implement write_iter
+ * changed arguments of tty_write function
+ */
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5, 11, 0)
```
yonghong-song, line 116: I forgot to mention last time. bcc tools typically do not use this LINUX_VERSION_CODE comparison approach. The reason is due to backported kernels whose kernel version say is less then 5.11, but may actually have this feature supported. You can search various "supported" clause and may find some examples. Is it possible we could find
some variable/macro/structure we could use to identify this tty_write change?

Also, for ubuntu 16.04 failure issue, I think it is OKAY to only support the new feature with new tty_write signature so the tool can continue to work with old kernels.

olsajiri: > I forgot to mention last time. bcc tools typically do not use this LINUX_VERSION_CODE comparison approach. The reason is due to backported kernels whose kernel version say is less then 5.11, but may actually have this feature supported. You can search various "supported" clause and may find some examples. Is it possible we could find
> some variable/macro/structure we could use to identify this tty_write change?

I'll check but I'm not sure

> 
> Also, for ubuntu 16.04 failure issue, I think it is OKAY to only support the new feature with new tty_write signature so the tool can continue to work with old kernels.

hum,  you mean the tool WILL NOT continue to work with older kernels, right? if we support only new tty_write signature

yonghong-song: No, for the program with *new* function signature, only newer kernel is supported, you are free to use the newer features (like map value for bpf_perf_event_output). The *old* function signature should still be supported, e.g., with the original bpf code. This way, both old kernels and new kernels are supported by the tool.

yonghong-song: > I'll check but I'm not sure.
Thanks!

yonghong-song:
[buildbot, test this please]

yonghong-song:
fedora25/26/27 and ubuntu16.04, 17.10, 18.04 all failed with something like
```
30: ...cannot attach kprobe, Device or resource busy
30: Traceback (most recent call last):
30:   File "../../tools/ttysnoop.py", line 154, in <module>
30:     b.attach_kprobe(event="tty_write", fn_name="trace_tty_write")
30:   File "/home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1804/build/src/python/bcc-python/bcc/__init__.py", line 691, in attach_kprobe
30:     (fn_name, event))
30: Exception: Failed to attach BPF program trace_tty_write to kprobe tty_write
```

fedora28 is fine. Could you take a look?

olsajiri:
> fedora25/26/27 and ubuntu16.04, 17.10, 18.04 all failed with something like
> 30: ...cannot attach kprobe, Device or resource busy

support to attach multiple kprobes to single function is for very long time,
so not sure what's the problem

> fedora28 is fine. Could you take a look?

I managed to get fedora 27, let's see if it'll fail for me

olsajiri:
> I managed to get fedora 27, let's see if it'll fail for me

hum, it's working for me

yonghong-song:
> > I managed to get fedora 27, let's see if it'll fail for me
> 
> hum, it's working for me

Okay, let me take a look.

tools/ttysnoop.py
```diff
@@ -97,9 +144,15 @@ def usage():
...
         exit()
 
+bpf_text = bpf_text.replace('USER_DATASIZE', '%s' % args.datasize)
+bpf_text = bpf_text.replace('USER_DATACOUNT', '%s' % args.datacount)
+
 # initialize BPF
 b = BPF(text=bpf_text)
 
+if BPF.get_kprobe_functions(b'tty_write'):
+    b.attach_kprobe(event="tty_write", fn_name="trace_tty_write")
```
yonghong-song, line 154: This won't work. I tried on a 5.2 kernel and see:
```
207: kprobe  name tty_write  tag 41976baf6bf5cfd7  gpl
        loaded_at 2021-04-19T16:56:28-0700  uid 0
        xlated 3488B  jited 2092B  memlock 4096B  map_ids 100,99
        btf_id 45
208: kprobe  name trace_tty_write  tag ab95056fe92735a8  gpl
        loaded_at 2021-04-19T16:56:29-0700  uid 0
        xlated 3672B  jited 2204B  memlock 4096B  map_ids 100,99
        btf_id 45

// run bpftool perf
...
pid 120451  fd 7: prog_id 207  kprobe  func tty_write  offset 0
pid 120451  fd 9: prog_id 208  kprobe  func tty_write  offset 0
```

Basically we have 2 kprobe programs attached to tty_write. There are two problems here for old kernel: first really old kernel like 4.4 may be 4.9 not supporting kprobe attaching two bpf programs, and second I see your code still use bpf_perf_event_output with map value which is not supported for 4.4 either.

I checked the commit changing the signature and some surrounding commits,
```
9bb48c82aced07698a2d08ee0f1475a6c4f6b266 tty: implement write_iter
dd78b0c483e33225e0e0782b0ed887129b00f956 tty: implement read_iter
3b830a9c34d5897be07176ce4e6f2d75e2c8cfd7 tty: convert tty_ldisc_ops 'read()' function to take a kernel pointer
```
and cannot find useful info (visitable header macro/struct_type change) etc. which the script can be used to detect whether the signature is changed. So in this case, I think use version is OK, I don't think people backports tty stuff, which should be extremely rare.

Your first version of code looks largely fine. But please keep the original code for kernel versions < 5.11 and the new code >= 5.11. This way, the old kernel including 4.4 will be still supported. Please also add some comments in the code around the place where you check kernel versions to mention we do kernel version check since we could not find a reliable way to detect the precise point where the kernel signature is changed. Thanks!

olsajiri: ok, will post new version, thanks

yonghong-song:
This version looks good to me. Thanks!

PR #3363, opened 2021-04-18
add bpf_map_lookup_batch and bpf_map_delete_batch in bcc

egobillot:
Add **bpf_map_lookup_batch** and **bpf_map_delete_batch** in bcc
* add bpf_map_lookup_batch in bcc
* add bpf_map_delete_batch in bcc
* add test_map_batch_ops.py to test batch lookup and delete on map
* add items_lookup_batch() and items_delete_batch()  in the reference guide

The bpf_map_delete_batch in table.py is not using bpf_map_delete_batch from libbcc. It rather uses bpf_map_lookup_and_delete_batch without using the return values. I did not succeed in writing bpf_map_delete_batch with only one BPF syscall with  BPF_MAP_DELETE_BATCH. I doubt this is possible.

bpf_map_delete_batch could be used to improve the table.clear() (less system calls) but it would only works for kernel v5.6 and later. How do you you manage this ? Do you let the kernel dependencies management in libbpf ?

### DIFF
--- a/src/python/bcc/table.py
+++ b/src/python/bcc/table.py
@@ -422,13 +422,41 @@ class TableBase(MutableMapping):
         for i in range(0, count.value):
             yield (keys[i], values[i])
 
-    def items_delete_batch(self):
-        """Delete all the key-value pairs in the map.
-        It is faster to call bpf_lookup_and_delete_batch than create keys list
-        and then call bpf_delete_batch on these keys.
+    def items_delete_batch(self, keys=None):
+        """Delete all the key-value pairs in the map if no key are given.
+        In that case, it is faster to call lib.bpf_lookup_and_delete_batch than
+        create keys list and then call lib.bpf_delete_batch on these keys.
+        If a list of keys is given then it deletes the related key-value.
         """
-        for _ in self.items_lookup_and_delete_batch():
-            return
+        if keys is not None:
+            # a list is expected
+            if type(keys) != list:
+                raise TypeError
+
+            batch_size = len(keys)
+            if batch_size < 1 and batch_size > self.max_entries:
+                raise KeyError
+
+            count = ct.c_uint32(batch_size)
+
+            # build the array aka list of key_t that will be deleted
+            keylist = (type(self.Key()) * batch_size)()
+            for i, k in enumerate(keys):
+                keylist[i] = k
+
+            res = lib.bpf_delete_batch(self.map_fd,
+                                       ct.byref(keylist),
+                                       ct.byref(count)
+                                       )
+            errcode = ct.get_errno()
+            if (errcode == errno.EINVAL):
+                raise Exception("BPF_MAP_DELETE_BATCH is invalid.")
+
+            if (res != 0 and errcode != errno.ENOENT):
+                raise Exception("BPF_MAP_DELETE_BATCH has failed")
+        else:
+            for _ in self.items_lookup_and_delete_batch():
+                return
 
     def items_lookup_and_delete_batch(self):
         # batch size is set to the maximum
--- a/src/python/bcc/table.py
+++ b/src/python/bcc/table.py
@@ -397,6 +397,64 @@ class TableBase(MutableMapping):
         for k in self.keys():
             self.__delitem__(k)
 
+    def items_update_batch(self, dictionary):
+        # batch size is set to the maximum
+        nb_key = len(dictionary)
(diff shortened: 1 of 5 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks @egobillot for the contribution. To answer some of your questions below:
> The bpf_map_delete_batch in table.py is not using bpf_map_delete_batch from libbcc. It rather uses bpf_map_lookup_and_delete_batch without using the return values. I did not succeed in writing bpf_map_delete_batch with only one BPF syscall with BPF_MAP_DELETE_BATCH. I doubt this is possible.

Currently, to BPF_MAP_DELETE_BATCH needs an array of keys, most likely these keys will come from BPF_MAP_LOOKUP_BATCH or BPF_MAP_LOOKUP_AND_DELETE_BATCH. With these keys, you can delete them in one syscall.

> bpf_map_delete_batch could be used to improve the table.clear() (less system calls) but it would only works for kernel v5.6 and later. How do you you manage this ? Do you let the kernel dependencies management in libbpf ?

It is tricky to use table.clear() without using proceeding BPF_MAP_LOOKUP_BATCH, e.g., example, user space may lookup some keys like k1, k2, k3, and then when you call table.clear(), maybe some new keys are already added like k4. In this case, k4 will be removed. and this may make analysis result in-precise. I know today we already have this issue, but it would be good we can use the interface to improve the situation.

You can test whether /proc/kallsyms having this symbol `generic_map_lookup_batch` or not. If it exists, the feature is supported.
```
ffffffff811b8d30 t bpf_map_do_batch
ffffffff811bb2c0 T generic_map_delete_batch
ffffffff811bb460 T generic_map_update_batch
ffffffff811bb660 T generic_map_lookup_batch
ffffffff811d5b40 t __htab_map_lookup_and_delete_batch
ffffffff811d6310 t htab_map_lookup_and_delete_batch
ffffffff811d6330 t htab_map_lookup_batch
ffffffff811d6350 t htab_lru_map_lookup_and_delete_batch
ffffffff811d6370 t htab_lru_map_lookup_batch
ffffffff811d6390 t htab_percpu_map_lookup_and_delete_batch
ffffffff811d63b0 t htab_percpu_map_lookup_batch
ffffffff811d63d0 t htab_lru_percpu_map_lookup_and_delete_batch
ffffffff811d63f0 t htab_lru_percpu_map_lookup_batch
```

src/python/bcc/table.py
```diff
@@ -397,6 +397,39 @@ def clear(self):
...
+        if (errcode == errno.EINVAL):
+            raise Exception("BPF_MAP_LOOKUP_BATCH is invalid.")
+
+        if (res != 0 and errcode != errno.ENOENT):
+            raise Exception("BPF_MAP_LOOKUP_BATCH has failed")
+
+        for i in range(0, count.value):
+            yield (keys[i], values[i])
+
+    def items_delete_batch(self):
```
yonghong-song, line 425: I would like items_delete_patch takes an optional argument which is an array of keys. If users provide this array, we should be call self.items_delete_batch to delete these keys only. If user provides no keys, we can call items_lookup_and_delete_batch() to delete all of keys in the map. WDYT?

egobillot: Yes, good idea.

egobillot: And to be consistent, I should also provide the same optional argument, an array of keys, for items_lookup_and_delete_batch() and items_lookup_batch(). The src code should be almost the same than items_delete_batch(). If you are Ok with this commit https://github.com/iovisor/bcc/commit/4dbaeef03d284dc0ed931a918a36d69e646521cd, I will continue with the 2 other functions.

yonghong-song:
[buildbot, test this please]

src/python/bcc/table.py
yonghong-song: > And to be consistent, I should also provide the same optional argument, an array of keys, for items_lookup_and_delete_batch() and items_lookup_batch(). The src code should be almost the same than items_delete_batch(). If you are Ok with this commit [4dbaeef](https://github.com/iovisor/bcc/commit/4dbaeef03d284dc0ed931a918a36d69e646521cd), I will continue with the 2 other functions.

@egobillot The new change looks good to me. Agree that let us make API consistent to have keys for batched lookup, lookup_and_delete as well. Thanks for working on this!

PR #3367, opened 2021-04-21
Add NFS tools

yzgyyang:


yonghong-song:
[buildbot, test this please]

yzgyyang:
@yonghong-song From the build log I see:

```
Setting up python-bcc (0.19.0-21.git.8a0db755) ...
  File "/usr/lib/python2.7/dist-packages/bcc/topclass.py", line 52
    def __init__(self, column_fmt: dict, app_name: str, args, stdscr=None):
                                 ^
SyntaxError: invalid syntax
```

These scripts are tested using Python 3.6 - I may put some effort to write Python 2-3 compatible code, but just wondering if it is possible not to do so? Also, is there a plan for Python 3?

yonghong-song:
@yzgyyang I will take a look at tools soon. For your question related to python3.6, please take a look at tcpconnect.py, which put some 3.6 packages under an option so the tool will be still be useful for python2 users. Is it possible for you?

tools/filetop_getattr.py
```diff
@@ -0,0 +1,412 @@
...
+        bpf_text = bpf_text.replace("_FILTER_COMM_BLOCK_", filter_comm)
+
+        # dry run option
+        if self.args.dry_run:
+            print(bpf_text)
+            sys.exit()
+
+        # initialize BPF
+        self.bpf = bcc.BPF(text=bpf_text)
+        self.bpf.attach_kprobe(event="vfs_getattr", fn_name="trace_getattr_entry")
```
yonghong-song, line 315: Here, we attached to vfs_getattr(), which is tracing not only nfs but also stat() syscall etc. Is this really what you want? I thought you try to check and debug nfs layer. Is it possible to kprobe some nfs related functions?

tools/filetop_getattr.py
```diff
@@ -0,0 +1,412 @@
...
+ * container_of - cast a member of a structure out to the containing structure
+ * @ptr:        the pointer to the member.
+ * @type:       the type of the container struct this is embedded in.
+ * @member:     the name of the member within the struct.
+ */
+#ifndef containerof
+#define containerof(ptr, type, member) ({                      \
+    const typeof( ((type *)0)->member ) *__mptr = (ptr);    \
+    (type *)( (char *)__mptr - offsetof(type,member) );})
+#endif
```
yonghong-song, line 102: I just added the above two *generic* macros in the helpers.h
https://github.com/iovisor/bcc/commit/cda7acdbddb19a5a258f881711845e41463dabbf
so you don't need to define here.

yonghong-song (commented):
First, thanks for contributing to bcc! Please take a look at
   https://github.com/iovisor/bcc/blob/master/CONTRIBUTING-SCRIPTS.md
for general guideline for a new tool.

The current version lacks details about why these tools will be useful to you and how they helped you monitor/debug your production issues. These are important metrics to gauge how useful these tools are for bcc/bpf community. Please take some time to describe in detail.

cc @brendangregg

yzgyyang:
> please take a look at tcpconnect.py, which put some 3.6 packages under an option so the tool will still be useful for python2 users

@yonghong-song By "put some 3.6 packages under an option", do you mean this? https://github.com/iovisor/bcc/blob/e83019bdf6c400b589e69c7d18092e38088f89a8/tools/tcpconnect.py#L432-L444

I don't think the scripts in this PR use any external packages though - I can surely change the code to be both python2/3 compatible, by doing things like this: https://github.com/iovisor/bcc/blob/e83019bdf6c400b589e69c7d18092e38088f89a8/tools/tcpconnect.py#L24

And remove python3 features like type hinting, etc.

What do you think?

tools/filetop_getattr.py
yzgyyang: Thank you! Will test and remove them soon.

tools/filetop_getattr.py
yzgyyang: Yes - took a similar idea from the original `filetop` script here, but we added the `-c` option which filters the process name, so when we do `-c nfsd` it captures NFSd-related calls while the tool itself is still generic.

yzgyyang:
> Please take a look at
> https://github.com/iovisor/bcc/blob/master/CONTRIBUTING-SCRIPTS.md
> for general guideline for a new tool.
> 
> The current version lacks details about why these tools will be useful to you and how they helped you monitor/debug your production issues. These are important metrics to gauge how useful these tools are for bcc/bpf community. Please take some time to describe in detail.

@yonghong-song Sounds good, thanks! I will follow the guide to create the documentation (example.txt, man page and README entries) to provide some contexts, and let you know.

brendangregg:
Ok, thanks for working on NFS tools, but in their current state they seem far from a suitable bcc addition. Have you read CONTRIBUTING-SCRIPTS.md? Just a few quick comments:

- Why do we have 3 NFS top's? If you want to write nfstop, do so and call it "nfstop". Make it like top(1): it prints everything by default (user, IP, files), and you can then control those columns using command line options. Adding 3 tools when 1 would be suitable just adds clutter and makes end-users needlessly learn more things. We already have a lot of tools, and that's overwhelming for newcomers, so we have to be extra careful to not add anything that's unnecessary.
- 400 lines for a Python library. A) I doubt the multiple tops should exist anyway, so I doubt this library should exist. B) That's a lot of lines for us to maintain, given that there are already top's in bcc that are entirely smaller. C) We're moving to C libbpf-tools and are not really investing in Python anymore.
- The names are all inconsistent with the other bcc tools. E.g., filetop_getattr.py should be nfsgetattrtop.py, iptop.py and usertop.py are VERY misleading and should have an nfs prefix. But I don't think there should be three in the first place: there should just be one 'nfstop'.

If an nfstop ends up more than 300-400 lines, I'd question whether it's gone too far. I can take any of the current tools and make them 2000 lines by adding a ton of options, but it doesn't serve the community in the long run as we end up with things that are not well tested and not well maintained. If it can solve a bulk of the problems with 300 lines, then stop right there. Let people write ad hoc bpftrace one-liners for the edge cases.

yzgyyang:
@brendangregg Thank you for the constructive comments! I read `CONTRIBUTING-SCRIPTS.md` just now and will now work on that checklist if these tools are still deemed as valuable for this repo - this PR was put up to see if I am on the right track of submission, and should be a "draft PR" really, as we basically just pulled the code from the internal repo and wanted to identify the steps for them to be accepted. Apologize for any confusion!

> * Why do we have 3 NFS top's? If you want to write nfstop, do so and call it "nfstop". Make it like top(1): it prints everything by default (user, IP, files), and you can then control those columns using command line options.
We try to summarize incoming NFS requests by different aspects: hot files, hot users and hot IPs (clients). The output, BPF map and focus of these are very different, so I feel like "do one thing and do it well" applies here.

Also, `filetop_getattr` is a rewrite/extension to the original `filetop` script and still focuses on the VFS layer, while `iptop` and `usertop` captures `nfsd3_proc_getattr`/`nfsd4_getattr` and are more similar.

> * 400 lines for a Python library. A) I doubt the multiple tops should exist anyway, so I doubt this library should exist. B) That's a lot of lines for us to maintain, given that there are already top's in bcc that are entirely smaller. C) We're moving to C libbpf-tools and are not really investing in Python anymore.

The 400 lines of the library are mainly to abstract the top tools, it handles multiple types of outputs (json, plain text, interactive with curses) and run modes (run once/continuously) and contains the same moving parts for all top tools. We put up the library to accommodate more top tools in the future - also, there are already 6 top tools in the repo, and they may adopt the library to instantly get features for free, instead of their own (slightly different between each) implementation to outputs etc.

Of course, if these are not enough to justify a library, we can surely put these functions back into the scripts and strip unwanted features.

> * The names are all inconsistent with the other bcc tools. E.g., filetop_getattr.py should be nfsgetattrtop.py, iptop.py and usertop.py are VERY misleading and should have an nfs prefix. But I don't think there should be three in the first place: there should just be one 'nfstop'.

The naming should definitely be changed =) We used these internally but could surely make them consistent with the repo (maybe `nfs{file, ip, user}top.py`?). The name filetop_getattr is because the tool is a rewrite and expansion of the existing filetop, and is not final for sure.

> If an nfstop ends up more than 300-400 lines, I'd question whether it's gone too far. I can take any of the current tools and make them 2000 lines by adding a ton of options, but it doesn't serve the community in the long run as we end up with things that are not well tested and not well maintained. If it can solve a bulk of the problems with 300 lines, then stop right there. Let people write ad hoc bpftrace one-liners for the edge cases.

Please take a look at the current `iptop` and `usertop` as they are around 200 lines of code. `filetop` is a bit lengthy since we are trying to solve the "full path" + mountpoint and assemble in Python. A lot of the code is to transform the data and make output "complete" and "nicer", so I personally wouldn't say they are just a ton of extra options to make things lengthy.

These features are what we thought we needed for our debugging/monitoring purposes, and they surely can change to better fit in the repo.

brendangregg:
I still don't see a reason to have three tools when one will suffice. There should only be one nfstop.

Also consider: I use "top" tools when I'm debugging my laptop, but for production server analysis I rarely do. It's because I need a log of what's happening to attach to Jira tickets, drop in SRE chatrooms, attach to vendor emails, etc. The first time you do serious performance work and miss capturing a screenshot because top cleared the screen too quickly, is when you realize that these top tools are really just toys. Ok for personal use. Ok for debugging your laptop. Not recommended for serious production use by performance engineers. That's an additional reason why I don't think we want a top library in bcc. It's ok to have a few top tools, but I don't want to encourage too many.

I still haven't looked at the code, but tracing VFS and then filtering on NFS should only be done as a **last resort** when you can't get that info from NFS tracepoints or NFS kprobes. It adds too much overhead, as VFS gets called by everything. Yes, I used it in filetop, because (until there are VFS tracepoints) there's no other way to do it. And I made it clear in the man page that it costs dearly:

```
OVERHEAD
       Depending on the frequency of application reads  and  writes,  overhead
       can  become significant, in the worst case slowing applications by over
       50%. Hopefully for real world workloads the overhead is  much  less  --
       test before use. The reason for the high overhead is that VFS reads and
       writes can be a frequent event, and despite  the  eBPF  overhead  being
       very  small per event, if you multiply this small overhead by a million
       events per second, it becomes a million times worse. Literally. You can
       gauge  the  number  of reads and writes using the vfsstat(8) tool, also
       from bcc.
```

Now, I'm wondering how we got here. Looks like I'm missing advice in CONTRIBUTING_SCRIPTS.md about tracepoints vs kprobes vs uprobes. I might submit an update to add more detail. But how do we get newcomers to read it? I'd hate to have a PR checklist saying "[ ] Yes, I've read CONTRIBUTING_SCRIPTS.md", but I don't have a better idea.

PR #3368, opened 2021-04-21
Tools: add the PPID field in mountsnoop

w-simon:
It is found that in the production environment, the system() function
or shell command is often used to start the mount process temporarily,
so the PPID field needs to be added to find the corresponding program.

Signed-off-by: Wen Yang <wenyang@linux.alibaba.com>

### DIFF
--- a/tools/mountsnoop.py
+++ b/tools/mountsnoop.py
@@ -72,6 +72,8 @@ struct data_t {
             /* current->nsproxy->mnt_ns->ns.inum */
             unsigned int mnt_ns;
             char comm[TASK_COMM_LEN];
+            char pcomm[TASK_COMM_LEN];
+            pid_t ppid;
             unsigned long flags;
         } enter;
         /*
@@ -102,6 +104,8 @@ int syscall__mount(struct pt_regs *ctx, char __user *source,
     bpf_get_current_comm(event.enter.comm, sizeof(event.enter.comm));
     event.enter.flags = flags;
     task = (struct task_struct *)bpf_get_current_task();
+    event.enter.ppid = task->real_parent->tgid;
+    bpf_probe_read_kernel_str(&event.enter.pcomm, TASK_COMM_LEN, task->real_parent->comm);
     nsproxy = task->nsproxy;
     mnt_ns = nsproxy->mnt_ns;
     event.enter.mnt_ns = mnt_ns->ns.inum;
@@ -157,6 +161,8 @@ int syscall__umount(struct pt_regs *ctx, char __user *target, int flags)
     bpf_get_current_comm(event.enter.comm, sizeof(event.enter.comm));
     event.enter.flags = flags;
     task = (struct task_struct *)bpf_get_current_task();
+    event.enter.ppid = task->real_parent->tgid;
+    bpf_probe_read_kernel_str(&event.enter.pcomm, TASK_COMM_LEN, task->real_parent->comm);
     nsproxy = task->nsproxy;
     mnt_ns = nsproxy->mnt_ns;
     event.enter.mnt_ns = mnt_ns->ns.inum;
@@ -243,6 +249,8 @@ class EnterData(ctypes.Structure):
     _fields_ = [
         ('mnt_ns', ctypes.c_uint),
         ('comm', ctypes.c_char * TASK_COMM_LEN),
+        ('pcomm', ctypes.c_char * TASK_COMM_LEN),
+        ('ppid', ctypes.c_uint),
         ('flags', ctypes.c_ulong),
     ]
 
@@ -330,7 +338,7 @@ else:
         return '"{}"'.format(''.join(escape_character(c) for c in s))
 
 
-def print_event(mounts, umounts, cpu, data, size):
+def print_event(mounts, umounts, parent, cpu, data, size):
     event = ctypes.cast(data, ctypes.POINTER(Event)).contents
 
     try:
@@ -341,6 +349,8 @@ def print_event(mounts, umounts, cpu, data, size):
                 'mnt_ns': event.union.enter.mnt_ns,
                 'comm': event.union.enter.comm,
                 'flags': event.union.enter.flags,
+                'ppid': event.union.enter.ppid,
+                'pcomm': event.union.enter.pcomm,
             }
         elif event.type == EventType.EVENT_MOUNT_SOURCE:
             mounts[event.pid]['source'] = event.union.str
@@ -358,6 +368,8 @@ def print_event(mounts, umounts, cpu, data, size):
                 'mnt_ns': event.union.enter.mnt_ns,
                 'comm': event.union.enter.comm,
                 'flags': event.union.enter.flags,
+                'ppid': event.union.enter.ppid,
+                'pcomm': event.union.enter.pcomm,
             }
         elif event.type == EventType.EVENT_UMOUNT_TARGET:
             umounts[event.pid]['target'] = event.union.str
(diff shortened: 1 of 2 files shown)


yonghong-song:
[buildbot, test this please]

tools/mountsnoop.py
```diff
@@ -379,9 +408,9 @@ def print_event(mounts, umounts, cpu, data, size):
                     target=decode_mount_string(syscall['target']),
                     flags=decode_umount_flags(syscall['flags']),
                     retval=decode_errno(event.union.retval))
-            print('{:16} {:<7} {:<7} {:<11} {}'.format(
-                syscall['comm'].decode('utf-8', 'replace'), syscall['tgid'],
-                syscall['pid'], syscall['mnt_ns'], call))
+            print('{:16} {:<7} {:<7} {:<7} {:<11} {}'.format(
```
davemarchevsky, line 411: Looks like this is the first time field ordering is being changed since `mountsnoop`'s [initial commit](https://github.com/iovisor/bcc/commit/e822a8189d9ba376112960417223191296417431). Is there any chance someone might be relying on the current order? I'm imagining a situation where someone is parsing the output of `mountsnoop` and doing `pid = fields[1]` instead of using first line to determine order.

I don't know if this is a reasonable concern for this tool even if it could happen. My hunch would be not to worry about it, but wanted to get your thoughts.

davemarchevsky (changes requested):
LGTM aside from nits, and one hypothetical concern

I tested the changes out on a host with recent kernel (5.6) and bcc, worked as expected.

tools/mountsnoop.py
davemarchevsky: The manpage says

> .SH STABILITY
> Unstable - in development.

as do all the other tool manpages

tools/mountsnoop.py
```diff
@@ -330,6 +342,21 @@ def decode_mount_string(s):
         return '"{}"'.format(''.join(escape_character(c) for c in s))
 
 
+# This is best-effort PPID matching. Short-lived processes may exit
+# before we get a chance to read the PPID.
+# This is a fallback for when fetching the PPID from task->real_parent->tgip
```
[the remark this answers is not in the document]
w-simon: OK, thanks

tools/mountsnoop.py
```diff
@@ -330,6 +342,21 @@ def decode_mount_string(s):
...
+# returns 0, which happens in some kernel versions.
```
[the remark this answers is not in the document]
w-simon: OK, we will improve it. Thanks.

tools/mountsnoop.py
w-simon: Thank you very much.
Compatibility is very important. We really need to consider it carefully here. If we add the two fields later, will it be better? It's like this:
COMM PID TID MNT_ NS CALL PCOMM PPID

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

davemarchevsky (changes requested):
Commented on and un-resolved the thread talking about adding new columns.

tools/mountsnoop.py
w-simon: This is a good idea. We'll optimize it later. 
Thanks.

w-simon:
> Commented on and un-resolved the thread talking about adding new columns.

Sorry for the delay. 
Please take a look again.Thanks.

tools/mountsnoop.py
```diff
@@ -341,6 +374,8 @@ def print_event(mounts, umounts, cpu, data, size):
                 'mnt_ns': event.union.enter.mnt_ns,
                 'comm': event.union.enter.comm,
                 'flags': event.union.enter.flags,
+                'ppid': event.ppid if event.ppid > 0 else get_ppid(event.pid),
```
yonghong-song, line 377: Did you really see a case where you need to call get_ppid(event.pid) to get ppid? I though kernel found find event.ppid properly in all cases. If we missed something, we should understand and add proper bpf logic.

tools/mountsnoop.py
```diff
@@ -341,6 +374,8 @@ def print_event(mounts, umounts, cpu, data, size):
...
+                'pcomm': pid_to_comm(event.ppid),
```
yonghong-song, line 378: You can get pcomm in bpf program. This way, you don't need to worry about short lived processes.

w-simon: Thanks.We will revise it later according to your suggestions.

tools/mountsnoop.py
```diff
@@ -393,6 +413,8 @@ def main():
     )
     parser.add_argument("--ebpf", action="store_true",
         help=argparse.SUPPRESS)
+    parser.add_argument("-P", "--parent_process", action="store_true",
+        help="also snoop the parent process")
```
danobi, line 417: Why add a flag if the bpf prog is going to unconditionally collect the parent info? Would keep the code simpler to remove the branches for `args.parent_process` too.

I'd say either gate the collection prog side or remove the gate altogether.

w-simon: Thanks.
This is a compatibility issue, which has been discussed with @davemarchevsky and @yonghong-song  before. 
The brief contents are as follows:
Looks like this is the first time field ordering is being changed since mountsnoop's initial commit. Is there any chance someone might be relying on the current order? I'm imagining a situation where someone is parsing the output of mountsnoop and doing pid = fields[1] instead of using first line to determine order.

So we only print the columns when a flag is passed. 
This way default output is not affected.

Nevertheless, if the community strongly recommends deleting the flag, we will follow it.

davemarchevsky: I think the conclusions reached in previous discussion still hold: we should keep output format stable and gate these new columns.

Gating BPF-side collection behind the flag feels like a nice-to-have that doesn't need to be addressed in this PR. I'm in favor of keeping the BPF simple to ease eventual conversion to `libbpf-tools` style and don't think the extraneous 20 bytes of copying will affect perf significantly since this tool is already copying ~400 bytes.

tools/mountsnoop.py
```diff
@@ -157,6 +162,9 @@
     bpf_get_current_comm(event.enter.comm, sizeof(event.enter.comm));
     event.enter.flags = flags;
     task = (struct task_struct *)bpf_get_current_task();
+    event.enter.ppid = task->real_parent->tgid;
+    __builtin_memcpy(event.enter.pcomm, task->real_parent->comm, TASK_COMM_LEN);
```
davemarchevsky, line 166: can you use `bpf_probe_read_kernel` here, and in the other `__builtin_memcpy` instance?

davemarchevsky (changes requested):
Aside from `__builtin_memcpy` nit I think this is good to go

tools/mountsnoop.py
w-simon: OK. Thanks for your advice.

brendangregg:
Thanks, looks good.

Just a minor suggestion: Since the output is getting a bit wide (I know it's a -P option, so that makes it ok; I hate wide default outputs), I'd still consider making the COMM and PCOMM fields 12 characters wide, which will usually be sufficient, and letting them overflow up to 16 for the few bigger process names. See for yourself: Do you have many process names bigger than 12 chars?

davemarchevsky (approved):

yonghong-song:
@w-simon could you address @brendangregg 's above comments?

w-simon:
> Thanks, looks good.
> 
> Just a minor suggestion: Since the output is getting a bit wide (I know it's a -P option, so that makes it ok; I hate wide default outputs), I'd still consider making the COMM and PCOMM fields 12 characters wide, which will usually be sufficient, and letting them overflow up to 16 for the few bigger process names. See for yourself: Do you have many process names bigger than 12 chars?

Thank you.
We did an experiment to reduce the width of COMM and PCOMM 12 characters. However, since the contents of the table increase with the mount/umout events, the columns may not be aligned, as shown below：

```
COMM         PID      TID       PCOMM        PPID     MNT_NS      CALL
umount       3348    3348    short.sh       3346    1075831104  umount("/root/2", 0x0) = 0
...
umount       3360    3360    long_long_long_ 3358    1075831104  umount("/root/2", 0x0) = 0
```

In addition,  we can only save a few characters in the COMM and PCOMM fields, but the call field can be super long due to path, far more than 80 characters, which could occupy several lines, as follows:

```
COMM             PID     TID     MNT_NS      CALL
(_exit.sh)       2055623 2055623 2162281600  mount("", "", "", MS_REC|MS_SLAVE, "") = 0
(_exit.sh)       2055623 2055623 2162281600  mount("/", "/proc/self/fd/3", "", MS_BIND|MS_REC, "") = 0
(_exit.sh)       2055623 2055623 2162281600  umount("/run/systemd/unit-root/proc", UMOUNT_NOFOLLOW) = -EBUSY
(_exit.sh)       2055623 2055623 2162281600  umount("/run/systemd/unit-root/proc/sys/fs/binfmt_misc", UMOUNT_NOFOLLOW) = 0

```
Thank you very much. 
Please give us more guidance.

PR #3370, opened 2021-04-22
libbpf-tools: add two helpers

ethercflow:
add discussed helpers https://github.com/iovisor/bcc/pull/3347#discussion_r612096999

### DIFF
--- a/libbpf-tools/trace_helpers.c
+++ b/libbpf-tools/trace_helpers.c
@@ -6,8 +6,11 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <unistd.h>
 #include <sys/resource.h>
 #include <time.h>
+#include <bpf/btf.h>
+#include <bpf/libbpf.h>
 #include "trace_helpers.h"
 
 #define min(x, y) ({				 \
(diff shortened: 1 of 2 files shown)


ethercflow:
Currently only tested on 5.12 rc2+ and CentOS Stream 8.3/ 4.18.0-240.22.1.el8_3. Not tested on 5.4 with BTF yet.

libbpf-tools/trace_helpers.c
```diff
@@ -395,3 +398,103 @@ bool is_kernel_module(const char *name)
...
+		if ((ret == EOF && feof(f)) || ret != 1)
+			break;
+		if (!strcmp(name, sym_name)) {
+			fclose(f);
+			return true;
+		}
+	}
+
+	fclose(f);
+	return false;
```
chenhengqi, line 481: Should we fallthrough to slow path?

anakryiko: No we shouldn't. If /sys/kernel/debug/tracing/available_filter_functions exists, it will contains all the kprobe'able functions. Even if kallsyms contains sym_name, that would mean that it's not attachable anyway.

libbpf-tools/trace_helpers.c
```diff
@@ -395,3 +398,103 @@ bool is_kernel_module(const char *name)
...
+	FILE *f;
+	int ret;
+
+	f = fopen("/sys/kernel/debug/tracing/available_filter_functions", "r");
+	if (!f)
+		goto slow_path;
+
+	while (true) {
+		ret = fscanf(f, "%s%*[^\n]\n", sym_name);
+		if ((ret == EOF && feof(f)) || ret != 1)
```
anakryiko, line 472: if ret != 1, we should at least log some error message, not just silently assume EOF. Similarly below.

anakryiko (commented):
It looks great, that's exactly what I had in mind. But can you please add comments to each fentry_exists and kprobe_exists explaining what they are for and on a high-level what they are doing? See also the point about silently ignoring reading errors.

ethercflow:
> It looks great, that's exactly what I had in mind. But can you please add comments to each fentry_exists and kprobe_exists explaining what they are for and on a high-level what they are doing? See also the point about silently ignoring reading errors.

Fixed, PTAL again. :)

ethercflow:
> It looks great, that's exactly what I had in mind. But can you please add comments to each fentry_exists and kprobe_exists explaining what they are for and on a high-level what they are doing? See also the point about silently ignoring reading errors.

Fixed, PTAL again. :)

libbpf-tools/trace_helpers.c
```diff
@@ -395,3 +398,112 @@ bool is_kernel_module(const char *name)
...
+		base = NULL;
+		goto err_out;
+	}
+	if (mod) {
+		snprintf(sysfs_mod, sizeof(sysfs_mod), "/sys/kernel/btf/%s", mod);
+		btf = btf__parse_split(sysfs_mod, base);
+		if (libbpf_get_error(btf)) {
+			fprintf(stderr, "failed to load BTF from %s: %s\n",
+				sysfs_mod, strerror(-libbpf_get_error(btf)));
+			btf = NULL;
```
anakryiko, line 424: no need to NULL it out, same above

libbpf-tools/trace_helpers.c
```diff
@@ -395,3 +398,112 @@ bool is_kernel_module(const char *name)
...
 	return found;
 }
+
+bool fentry_exists(const char *name, const char *mod)
+{
+	const char sysfs_vmlinux[] = "/sys/kernel/btf/vmlinux";
+	struct btf *base, *btf = NULL;
+	const struct btf_type *type;
+	const struct btf_enum *e;
+	char sysfs_mod[64];
```
anakryiko, line 408: max module name length is 56 , so this should be 16 + 1 + 56 = 73 to accomodate any possible module name, let's make it 80 for the round value?

libbpf-tools/trace_helpers.c
```diff
@@ -395,3 +398,112 @@ bool is_kernel_module(const char *name)
...
+	}
+
+	id = btf__find_by_name_kind(btf, "bpf_attach_type", BTF_KIND_ENUM);
+	if (id < 0)
+		goto err_out;
+	type = btf__type_by_id(btf, id);
+	if (!type) {
+		id = -1;
+		goto err_out;
+	}
```
anakryiko, line 438: this can't happen, if btf__find_by_name_kind() returned ID, it will be valid, so please drop this check altogether

libbpf-tools/trace_helpers.c
```diff
@@ -395,3 +398,112 @@ bool is_kernel_module(const char *name)
...
+	if (!f)
+		goto slow_path;
+
+	while (true) {
+		ret = fscanf(f, "%s%*[^\n]\n", sym_name);
+		if (ret == EOF && feof(f))
+			break;
+		if (ret != 1) {
+			fprintf(stderr, "failed to read symbol from "
+				"available_filter_functions\n");
```
anakryiko, line 476: please keep error strings on the single line, even if it's a very long one. It makes it easier to grep for it, if necessary. That's the rule for kernel, for example

libbpf-tools/trace_helpers.h
```diff
@@ -43,4 +43,34 @@ int bump_memlock_rlimit(void);
 
 bool is_kernel_module(const char *name);
 
+/*
+ * When we need for kprobe/kretprobe, please check out new fentry/fexit
```
anakryiko, line 47: Maybe "When attempting to use kprobe/kretprobe" ?

libbpf-tools/trace_helpers.h
```diff
@@ -43,4 +43,34 @@ int bump_memlock_rlimit(void);
...
+ * probes, as they provide better performance and usability. But in some
+ * situations we have to fallback to kprobe/kretprobe probes. This helper
+ * is used to detect those ones before attaching.
```
anakryiko, line 50: "those ones" is ambiguous. Maybe "used to detect fentry/fexit support for the specified kernel function" ?

libbpf-tools/trace_helpers.h
```diff
@@ -43,4 +43,34 @@ int bump_memlock_rlimit(void);
...
+ *
+ *	1. A gap between kernel versions, kernel BTF is exposed
+ * 	   starting from 5.4 kernel. but fentry/fexit is actually
+ * 	   supported starting from 5.5.
+ *	2. Wether kernel support module BTF or not
+ *
+ * *name* is the name of a function to be attached, which can be from
```
anakryiko, line 57: again, ambiguous. "kernel function to be attached to"?

libbpf-tools/trace_helpers.h
```diff
@@ -43,4 +43,34 @@ int bump_memlock_rlimit(void);
...
+ * vmlinux or a kernel module.
+ * *mod* is the name of a kernel module, if NULL, it means *name*
+ * belongs to vmlinux.
+ */
+bool fentry_exists(const char *name, const char *mod);
+
+/*
+ * The name of a function to be attached may be changed between kernel
```
anakryiko, line 65: same, this is not a name of a function we are attaching, it's the name of a function we are attaching *to*. So, as above, be more specific: "the name of a kernel function to be attached to"

libbpf-tools/trace_helpers.h
```diff
@@ -43,4 +43,34 @@ int bump_memlock_rlimit(void);
...
+ * releases. This helper is used to confirm whether the target kernel
+ * uses a certain function name before attaching.
+ *
+ * It is achieved by scaning
+ * 	/sys/kernel/debug/tracing/available_filter_functions
+ * If this file does not exist, it fallbacks to parse /proc/kallsyms,
+ * which is more slower.
```
anakryiko, line 72: typo: "more slow" or "slower"

anakryiko (changes requested):
few more nits and comments

ethercflow:
> few more nits and comments

Fixed, thanks!

anakryiko (approved):
LGTM, let's use it in vfsstat (cc @KentaTada) and other tools (ext4dist as well?)

ethercflow:
> LGTM, let's use it in vfsstat (cc @KentaTada) and other tools (ext4dist as well?)

Sure

yonghong-song:
[buildbot, test this please]

PR #3379, opened 2021-04-26
introduced map batch update in Python and Queue/Stack itervalues utility

s41m0n:
This pr aims at introducing:
* items_update_batch: batch operation to update multiple key-value pairs at the same time
* Queue/Stack itervalues: to lookup all values consequently

Signed-off-by: Simone Magnani <simonemagnani.96@gmail.com>

### DIFF
--- a/src/python/bcc/table.py
+++ b/src/python/bcc/table.py
@@ -460,6 +460,34 @@ class TableBase(MutableMapping):
             for _ in self.items_lookup_and_delete_batch():
                 return
 
+    def items_update_batch(self, keys, values):
+        """Update all the key-value pairs in the map provided.
+        The lists must be the same length, between 1 and the maximum number of entries.
+        """
+        # two ct.Array are expected
+        if not isinstance(keys, ct.Array) or not isinstance(values, ct.Array):
+            raise TypeError
+
+        batch_size = len(keys)
+
+        # check that batch between limits and coherent with the provided values
+        if batch_size < 1 or batch_size > self.max_entries or batch_size != len(values):
+            raise KeyError
+
+        count = ct.c_uint32(batch_size)
+        res = lib.bpf_update_batch(self.map_fd,
+                                   ct.byref(keys),
+                                   ct.byref(values),
+                                   ct.byref(count)
+                                   )
+
+        errcode = ct.get_errno()
+        if (errcode == errno.EINVAL):
+            raise Exception("BPF_MAP_UPDATE_BATCH is invalid.")
+
+        if (res != 0 and errcode != errno.ENOENT):
+            raise Exception("BPF_MAP_UPDATE_BATCH has failed")
+
     def items_lookup_and_delete_batch(self):
         # batch size is set to the maximum
         batch_size = self.max_entries
--- a/src/python/bcc/table.py
+++ b/src/python/bcc/table.py
@@ -314,6 +314,8 @@ class TableBase(MutableMapping):
         self.flags = lib.bpf_table_flags_id(self.bpf.module, self.map_id)
         self._cbs = {}
         self._name = name
+        self.max_entries = int(lib.bpf_table_max_entries_id(self.bpf.module,
+                self.map_id))
 
     def get_fd(self):
         return self.map_fd
@@ -670,9 +672,7 @@ class TableBase(MutableMapping):
 class HashTable(TableBase):
     def __init__(self, *args, **kwargs):
         super(HashTable, self).__init__(*args, **kwargs)
-        self.max_entries = int(lib.bpf_table_max_entries_id(self.bpf.module,
-                                                            self.map_id))
-
+        
     def __len__(self):
         i = 0
         for k in self: i += 1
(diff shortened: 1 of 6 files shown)


yonghong-song:
[buildbot, test this please]

src/python/bcc/table.py
```diff
@@ -458,6 +460,41 @@ def items_delete_batch(self, keys=None):
...
+        batch_size = len(keys)
+
+        # check that batch between limits and coherent with the provided values
+        if (batch_size < 1 and batch_size > self.max_entries) or (batch_size != len(values)):
+            raise KeyError
+        
+        # build the array aka list of key_t that will be deleted
+        keylist = (type(self.Key()) * batch_size)()
+        valuelist = (type(self.Leaf()) * batch_size)()
+        for i, k, v in enumerate(zip(keys, values)):
```
s41m0n, line 480: I am not sure if this for loop is needed, as for the *items_delete_batch* function.
* Isn't better from the performance point of view that the user already provides the arrays correctly formatted?
* What if the data type instead of being a primitive one (e.g., ct.c_uint32) is a user-defined struct?

yonghong-song: This is a good point. it is possible that batch_size could be big, so the library here may do double work if user already prepared an array. This is a library function, we can enforce the input parameters to be an array. The same applies to items_delete_batch(). Thanks!

yonghong-song (commented):
BTW, you can push several commits in the pull request, e.g., one commit for map_batch_update, another for queue/stack itervalue utility, and another one for items_delete_batch() and these commits will be merged into master without squashing. This will be better for later people to check commits for a particular feature/change.

yonghong-song:
BTW, the test failed due to
```
41: Test command: /home/fedora/jenkins/workspace/bcc-pr/label/fc28/build/tests/wrapper.sh "py_test_map_batch_ops" "sudo" "/home/fedora/jenkins/workspace/bcc-pr/label/fc28/tests/python/test_map_batch_ops.py"
41: Test timeout computed to be: 10000000
41:   File "/home/fedora/jenkins/workspace/bcc-pr/label/fc28/tests/python/test_map_batch_ops.py", line 12
41:     from __future__ import print_function
41:     ^
41: SyntaxError: from __future__ imports must occur at the beginning of the file
41: Failed
41/45 Test #41: py_test_map_batch_ops ............***Failed    0.04 sec
```

src/python/bcc/table.py
s41m0n: No problem :)

s41m0n:
> BTW, you can push several commits in the pull request, e.g., one commit for map_batch_update, another for queue/stack itervalue utility, and another one for items_delete_batch() and these commits will be merged into master without squashing. This will be better for later people to check commits for a particular feature/change.

I agree, sorry for not doing that before. It should be all right now.

src/python/bcc/table.py
```diff
@@ -458,6 +457,34 @@ def items_delete_batch(self, keys=None):
...
+        The lists must be the same length, between 1 and the maximum number of entries.
+        """
+        # two ct.Array are expected
+        if not isinstance(keys, ct.Array) or not isinstance(values, ct.Array):
+            raise TypeError
+
+        batch_size = len(keys)
+
+        # check that batch between limits and coherent with the provided values
+        if (batch_size < 1 and batch_size > self.max_entries) or (batch_size != len(values)):
```
yonghong-song, line 471: LGTM. Only one comment: here 'and' should be 'or', right? Otherwise, for any `max_entries >= 1`, `batch_size < 1 and batch_size > self.max_entries` is always false.

yonghong-song:
[buildbot, ok to test]

yonghong-song:
Also, there is a 2.7 python error,
```
Bytecompiling .py files below /tmp/rpmbuild.RuNXTK/BUILDROOT/bcc-0.19.0-31.git.7bbc553e.x86_64/usr/lib/python2.7 using /usr/bin/python2.7
Compiling /tmp/rpmbuild.RuNXTK/BUILDROOT/bcc-0.19.0-31.git.7bbc553e.x86_64/usr/lib/python2.7/site-packages/bcc/table.py ...
  File "/usr/lib/python2.7/site-packages/bcc/table.py", line 427
    def items_delete_batch(self, keys: ct.Array=None):
                                     ^
SyntaxError: invalid syntax
```

src/python/bcc/table.py
s41m0n: You're absolutely right, thanks for noticing! I fixed it also in items_delete_batch()

s41m0n:
> Also, there is a 2.7 python error,
> 
> ```
> Bytecompiling .py files below /tmp/rpmbuild.RuNXTK/BUILDROOT/bcc-0.19.0-31.git.7bbc553e.x86_64/usr/lib/python2.7 using /usr/bin/python2.7
> Compiling /tmp/rpmbuild.RuNXTK/BUILDROOT/bcc-0.19.0-31.git.7bbc553e.x86_64/usr/lib/python2.7/site-packages/bcc/table.py ...
>   File "/usr/lib/python2.7/site-packages/bcc/table.py", line 427
>     def items_delete_batch(self, keys: ct.Array=None):
>                                      ^
> SyntaxError: invalid syntax
> ```

That's true, type checking it's from Python3 on, and I forgot to run the tests also with Python2.
Should be ok now, even though that would be really useful feature to have :)

yonghong-song:
LGTM. Thanks!

PR #3381, opened 2021-04-27
Fix BPF(src_file="foo")

jeromemarchand:
Since commit 75f20a15 ("Use string type for comparison to PATH
elements"), src_file isn't working anymore. Somehow, two wrongs
(ArgString __str__() returning a bytes object and joining a bytes and
what was supposed to be a string) did make a right.

It fixes the following error in netqtop and deadlock:
  File "/usr/share/bcc/tools/netqtop", line 207, in <module>
    b = BPF(src_file = EBPF_FILE)
  File "/usr/lib/python3.6/site-packages/bcc/__init__.py", line 335, in __init__
    src_file = BPF._find_file(src_file)
  File "/usr/lib/python3.6/site-packages/bcc/__init__.py", line 255, in _find_file
    t = b"/".join([os.path.abspath(os.path.dirname(argv0.__str__())), filename])
TypeError: sequence item 0: expected a bytes-like object, str found

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -252,7 +252,7 @@ class BPF(object):
         if filename:
             if not os.path.isfile(filename):
                 argv0 = ArgString(sys.argv[0])
-                t = b"/".join([os.path.abspath(os.path.dirname(argv0.__str__())), filename])
+                t = b"/".join([os.path.abspath(os.path.dirname(argv0.__bytes__())), filename])
                 if os.path.isfile(t):
                     filename = t
                 else:


yonghong-song:
[buildbot, test this please]

yonghong-song:
I tried netqtop.py with both python2.7 and python3.6, the existing code works fine. But this patch also works with both python2.7 and python3.7. Not sure what is the exact problem.
@jongiddy could you also take a look at this change?

jeromemarchand:
> I tried netqtop.py with both python2.7 and python3.6, the existing code works fine. But this patch also works with both python2.7 and python3.7. Not sure what is the exact problem.
> @jongiddy could you also take a look at this change?

I hit the issue with python 3.6.8. There has been some changes in os.path.* interfaces in 3.6 that might explain the change in behavior ("Changed in version 3.6: Accepts a path-like object."):
https://docs.python.org/3/library/os.path.html

jongiddy:
I agree that it is broken. In #3296 I proposed the alternative fix of
```
 t = "/".join([os.path.abspath(os.path.dirname(str(argv0))), filename])
```

jeromemarchand:
> I agree that it is broken. In #3296 I proposed the alternative fix of
> 
> ```
>  t = "/".join([os.path.abspath(os.path.dirname(str(argv0))), filename])
> ```

The type of filename is bytes here (there the following code a bit earlier `src_file = _assert_is_bytes(src_file)`). With python 3.6.8, I get the following error:
```
# /usr/share/bcc/tools/netqtop -n lo
  File "/usr/share/bcc/tools/netqtop", line 207, in <module>
    b = BPF(src_file = EBPF_FILE)
  File "/usr/lib/python3.6/site-packages/bcc/__init__.py", line 335, in __init__
    src_file = BPF._find_file(src_file)
  File "/usr/lib/python3.6/site-packages/bcc/__init__.py", line 255, in _find_file
    t = "/".join([os.path.abspath(os.path.dirname(str(argv0))), filename])
```

jongiddy:
Yeah, #3296 would need to be adopted as a whole for my proposed change to work.

jeromemarchand:
Oh, I see. If #3296 would help prevent such issues in the future, I definitely welcome it. These kind of error pop up all the time.

yonghong-song:
Thanks, guys. I will take a look at #3296 again later.

Issue #3366, opened 2021-04-20
clang macro-redefined warnings in newer kernel versions


Mic92:
When running on 5.10.29, I get the following warning:

```
╭─[~/git/vmsh]─[main]
╰─ % sudo profile -F 99 -f > folded
In file included from <built-in>:2:
In file included from /virtual/include/bcc/bpf.h:12:
In file included from include/linux/types.h:6:
In file included from include/uapi/linux/types.h:14:
In file included from include/uapi/linux/posix_types.h:5:
In file included from include/linux/stddef.h:5:
In file included from include/uapi/linux/stddef.h:2:
In file included from include/linux/compiler_types.h:69:
include/linux/compiler-clang.h:45:9: warning: '__HAVE_BUILTIN_BSWAP32__' macro redefined [-Wmacro-redefined]
#define __HAVE_BUILTIN_BSWAP32__
        ^
<command line>:4:9: note: previous definition is here
#define __HAVE_BUILTIN_BSWAP32__ 1
        ^
In file included from <built-in>:2:
In file included from /virtual/include/bcc/bpf.h:12:
In file included from include/linux/types.h:6:
In file included from include/uapi/linux/types.h:14:
In file included from include/uapi/linux/posix_types.h:5:
In file included from include/linux/stddef.h:5:
In file included from include/uapi/linux/stddef.h:2:
In file included from include/linux/compiler_types.h:69:
include/linux/compiler-clang.h:46:9: warning: '__HAVE_BUILTIN_BSWAP64__' macro redefined [-Wmacro-redefined]
#define __HAVE_BUILTIN_BSWAP64__
        ^
<command line>:5:9: note: previous definition is here
#define __HAVE_BUILTIN_BSWAP64__ 1
        ^
In file included from <built-in>:2:
In file included from /virtual/include/bcc/bpf.h:12:
In file included from include/linux/types.h:6:
In file included from include/uapi/linux/types.h:14:
In file included from include/uapi/linux/posix_types.h:5:
In file included from include/linux/stddef.h:5:
In file included from include/uapi/linux/stddef.h:2:
In file included from include/linux/compiler_types.h:69:
include/linux/compiler-clang.h:47:9: warning: '__HAVE_BUILTIN_BSWAP16__' macro redefined [-Wmacro-redefined]
#define __HAVE_BUILTIN_BSWAP16__
        ^
<command line>:3:9: note: previous definition is here
#define __HAVE_BUILTIN_BSWAP16__ 1
        ^
[...]

PR #3391, opened 2021-04-30
Move HAVE_BUILTIN_BSWAP includes to separate header

davemarchevsky:
As reported in #3366, on newer kernels bcc complains about macro
redefinition when compiling bpf programs:

```
include/linux/compiler-clang.h:46:9: warning: '__HAVE_BUILTIN_BSWAP64__' macro redefined [-Wmacro-redefined]
\#define __HAVE_BUILTIN_BSWAP64__
        ^
<command line>:5:9: note: previous definition is here
\#define __HAVE_BUILTIN_BSWAP64__ 1
```

Since these macros are passed in as `-D` cflags, they appear first
before any \#define statements in code. Since an [upstream kernel
patch](https://lore.kernel.org/linux-csky/20210226161151.2629097-1-arnd@kernel.org/)
added these defines in a kernel header, we see the warning.

This patch moves these definitions to a separate 'virtual' header that's included
after virtual_bpf.h and adds an ifndef guard. As a result, newer kernels
with the patch will not trigger the warning, while older kernels will
not lose the definition.

This should be safe based on my digging - some existing bcc programs use
`__builtin_bswap` methods, but without checking HAVE_BUILTIN_BSWAP.
Macros that may be conditionally defined based on HAVE_BUILTIN_BSWAP,
like those in `bpf_endian.h`, aren't. If a similar macro or struct def
in virtual_bpf.h - or any header it pulls in - changes depending on
HAVE_BUILTIN_BSWAP this could cause problems on older kernels, but I
don't believe that this is the case, or will be based on how
infrequently the defines are checked.

### DIFF
--- a/src/cc/export/bpf_workaround.h
+++ b/src/cc/export/bpf_workaround.h
@@ -1,0 +1,11 @@
+R"********(
+#ifndef __HAVE_BUILTIN_BSWAP16__
+#define __HAVE_BUILTIN_BSWAP16__
+#endif
+#ifndef __HAVE_BUILTIN_BSWAP32__
+#define __HAVE_BUILTIN_BSWAP32__
+#endif
+#ifndef __HAVE_BUILTIN_BSWAP64__
+#define __HAVE_BUILTIN_BSWAP64__
+#endif
+)********"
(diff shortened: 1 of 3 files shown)


hhoffstaette:
Seems fine to me; obviously better than disabling any warnings and hoping for the best. ;)

davemarchevsky:
rebased, renamed to `bpf_workaround.h`

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

PR #3400, opened 2021-05-05
loader: include bpf_workaround.h header

davemarchevsky:
The intent of #3391 was to have `bpf_workaround.h` included after
`bpf.h` when compiling bcc headers. However, that PR only added the file
to the `headers_` map of files that can be included. To actually
include, need to adjust compiler input flags as well.

Fixes: d089013e ("Move HAVE_BUILTIN_BSWAP includes to separate header")


yonghong-song:
[buildbot, test this please]

PR #3404, opened 2021-05-06
Use arch-specific libdir with pkgconfig

lgbaldoni:


yonghong-song:
[buildbot, test this please]

lgbaldoni:
> @lgbaldoni could you give a brief description what the problem you hit and how this patch solved your problem? We need such information in the commit message. Thanks!

On systems with arch-dependent library dirs (e.g. `/usr/lib64`) this will pass the correct search path to the linker.

PR #3417, opened 2021-05-10
libbpf-tools: add statsnoop

chenhengqi:
```shell
sudo ./statsnoop -x -t -p 3534689
TIME       PID        COMM                 RET  ERR  PATH
0.000000   3534689    YDService            -1   2    /proc/2111125
0.000021   3534689    YDService            -1   2    /proc/2111125
0.000029   3534689    YDService            -1   2    /proc/2111126
```

### DIFF
--- a/libbpf-tools/statsnoop.bpf.c
+++ b/libbpf-tools/statsnoop.bpf.c
@@ -1,0 +1,94 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2021 Hengqi Chen
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include "statsnoop.h"
+
+#define MAX_ENTRIES 10240
+
+const volatile pid_t target_pid = 0;
+const volatile bool  trace_failed_only = false;
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_ENTRIES);
+	__type(key, __u32);
+	__type(value, const char *);
+} values SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, sizeof(__u32));
+} events SEC(".maps");
+
+static int probe_entry(void *ctx, const char *pathname)
+{
+	__u64 id = bpf_get_current_pid_tgid();
+	__u32 pid = id >> 32;
+	__u32 tid = (__u32)id;
+
+	if (!pathname)
+		return 0;
+
+	if (target_pid && target_pid != pid)
+		return 0;
+
+	bpf_map_update_elem(&values, &tid, &pathname, BPF_ANY);
+	return 0;
+};
+
+static int probe_return(void *ctx, int ret)
+{
+	__u64 id = bpf_get_current_pid_tgid();
+	__u32 pid = id >> 32;
+	__u32 tid = (__u32)id;
+	const char **pathname;
+	struct event event = {};
+
+	pathname = bpf_map_lookup_elem(&values, &tid);
+	if (!pathname)
+		return 0;
+
+	if (trace_failed_only && ret >= 0) {
+		bpf_map_delete_elem(&values, &tid);
+		return 0;
+	}
+
+	event.pid = pid;
+	event.ts_ns = bpf_ktime_get_ns();
+	event.ret = ret;
+	bpf_get_current_comm(&event.comm, sizeof(event.comm));
+	bpf_probe_read_user_str(event.pathname, sizeof(event.pathname), *pathname);
+
+	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, &event, sizeof(event));
+	bpf_map_delete_elem(&values, &tid);
+	return 0;
+}
+
+SEC("tracepoint/syscalls/sys_enter_statfs")
+int handle_statfs_entry(struct trace_event_raw_sys_enter *ctx)
+{
+	return probe_entry(ctx, (const char *)ctx->args[0]);
+}
+
+SEC("tracepoint/syscalls/sys_exit_statfs")
+int handle_statfs_return(struct trace_event_raw_sys_exit *ctx)
+{
+	return probe_return(ctx, (int)ctx->ret);
+}
+
+SEC("tracepoint/syscalls/sys_enter_newstat")
+int handle_newstat_entry(struct trace_event_raw_sys_enter *ctx)
+{
+	return probe_entry(ctx, (const char *)ctx->args[0]);
+}
+
+SEC("tracepoint/syscalls/sys_exit_newstat")
+int handle_newstat_return(struct trace_event_raw_sys_exit *ctx)
+{
+	return probe_return(ctx, (int)ctx->ret);
+}
+
+char LICENSE[] SEC("license") = "GPL";
--- a/libbpf-tools/statsnoop.c
+++ b/libbpf-tools/statsnoop.c
@@ -1,0 +1,186 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+// Copyright (c) 2021 Hengqi Chen
+//
+// Based on statsnoop(8) from BCC by Brendan Gregg.
+// 09-May-2021   Hengqi Chen   Created this.
+#include <argp.h>
+#include <errno.h>
+#include <signal.h>
+#include <time.h>
+
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+#include "statsnoop.h"
+#include "statsnoop.skel.h"
+#include "trace_helpers.h"
+
+#define PERF_BUFFER_PAGES       16
+#define PERF_POLL_TIMEOUT_MS    100
+#define warn(...) fprintf(stderr, __VA_ARGS__)
+
+static volatile sig_atomic_t exiting = 0;
+
+static pid_t target_pid = 0;
+static bool trace_failed_only = false;
+static bool emit_timestamp = false;
+
+const char *argp_program_version = "statsnoop 0.1";
+const char *argp_program_bug_address =
+	"https://github.com/iovisor/bcc/tree/master/libbpf-tools";
+const char argp_program_doc[] =
+"Trace stat syscalls.\n"
+"\n"
+"USAGE: statsnoop [-h] [-t] [-x] [-p PID]\n"
+"\n"
+"EXAMPLES:\n"
+"    statsnoop             # trace all stat syscalls\n"
+"    statsnoop -t          # include timestamps\n"
+"    statsnoop -x          # only show failed stats\n"
+"    statsnoop -p 1216     # only trace PID 1216\n";
+
+static const struct argp_option opts[] = {
+	{"pid", 'p', "PID", 0, "Process ID to trace"},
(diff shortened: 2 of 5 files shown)


libbpf-tools/statsnoop.c
```diff
@@ -0,0 +1,188 @@
...
+	err = statsnoop_bpf__attach(obj);
+	if (err) {
+		warn("failed to attach BPF programs: %d\n", err);
+		goto cleanup;
+	}
+
+	pb_opts.sample_cb = handle_event;
+	pb_opts.lost_cb = handle_lost_events;
+	pb = perf_buffer__new(bpf_map__fd(obj->maps.events), PERF_BUFFER_PAGES,
+			&pb_opts);
```
danobi, line 158: Style nit: line up `&pb_opts` with first character of `bpf_map__fd`

libbpf-tools/statsnoop.c
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	if (signal(SIGINT, sig_int) == SIG_ERR) {
+		warn("can't set signal handler: %s\n", strerror(-errno));
+		goto cleanup;
+	}
+
+	if (output_with_timestamp) {
+		printf("%-11s", "TIME");
+	}
```
danobi, line 172: Style nit: no curly braces for single statement

libbpf-tools/statsnoop.c
```diff
@@ -0,0 +1,188 @@
...
+	printf("%-10s %-20s %-4s %-4s %-s\n",
+			"PID", "COMM", "RET", "ERR", "PATH");
```
danobi, line 174: Style nit: too much indent

libbpf-tools/statsnoop.bpf.c
```diff
@@ -0,0 +1,99 @@
...
+static int probe_entry(void *ctx, const char *pathname)
+{
+	struct value_t val = {};
+	__u64 id = bpf_get_current_pid_tgid();
+	__u32 pid = id >> 32;
+	__u32 tid = id & 0xFFFFFFFF;
+
+	if (!pathname) {
+		return 0;
+	}
```
danobi, line 35: Nit: no braces

libbpf-tools/statsnoop.bpf.c
```diff
@@ -0,0 +1,99 @@
...
+
+	if (traced_pid && traced_pid != pid) {
+		return 0;
+	}
```
danobi, line 39: Same as above

libbpf-tools/statsnoop.bpf.c
```diff
@@ -0,0 +1,99 @@
...
+{
+	__u64 id = bpf_get_current_pid_tgid();
+	__u32 pid = id >> 32;
+	__u32 tid = id & 0xFFFFFFFF;
+	struct value_t *val;
+
+	val = bpf_map_lookup_elem(&values, &tid);
+	if (!val) {
+		return 0;
+	}
```
danobi, line 56: Same as above

libbpf-tools/statsnoop.h
```diff
@@ -0,0 +1,20 @@
...
+#define __STATSNOOP_H
+
+#define TASK_COMM_LEN   16
+#define NAME_MAX        255
+
+struct value_t {
+	const char *pathname;
+};
+
+struct event_t {
```
danobi, line 12: I think the `_t` suffix is a reserved suffix for POSIX APIs. I think these can just be called `struct value` and `struct event`

libbpf-tools/statsnoop.bpf.c
```diff
@@ -0,0 +1,99 @@
...
+
+	if (trace_failed_only && ret >= 0) {
+		bpf_map_delete_elem(&values, &tid);
+		return 0;
+	}
+
+	struct event_t event = {};
```
danobi, line 63: Nit: declare this at top of function

libbpf-tools/statsnoop.c
```diff
@@ -0,0 +1,188 @@
...
+	if (e->ret >= 0) {
+		fd = e->ret;
+		err = 0;
+	} else {
+		fd = -1;
+		err = -e->ret;
+	}
+	if (!start_timestamp) {
+		start_timestamp = e->ts_ns;
+	}
```
danobi, line 99: Why is `start_timestamp` initialized here? Seems like its value depends on when the first sample comes in. Why not do this in main()?

chenhengqi: Just follow its counterpart in tools:
https://github.com/iovisor/bcc/blob/f129fda996a73fd58e61e7446d89d9333bb6e9c2/tools/statsnoop.py#L159-L160

libbpf-tools/statsnoop.h
chenhengqi: OK, will do.

I find some of the tools already use `_t` suffix.

libbpf-tools/statsnoop.c
chenhengqi: kernel code style, right ?

danobi: Yeah, I figure we ought to stick to kernel style b/c the other tools seem to follow it

libbpf-tools/statsnoop.c
danobi: I guess it's fine if the intention is to compare time between events

danobi (approved):

libbpf-tools/statsnoop.bpf.c
```diff
@@ -0,0 +1,96 @@
...
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, sizeof(__u32));
+} events SEC(".maps");
+
+static int probe_entry(void *ctx, const char *pathname)
+{
+	struct value val = {};
+	__u64 id = bpf_get_current_pid_tgid();
+	__u32 pid = id >> 32;
+	__u32 tid = id & 0xFFFFFFFF;
```
anakryiko, line 31: nit: `(u32) id` is shorter and no one needs to count the number of 'F's :)

libbpf-tools/statsnoop.bpf.c
```diff
@@ -0,0 +1,96 @@
...
+
+	if (!pathname)
+		return 0;
+
+	if (target_pid && target_pid != pid)
+		return 0;
+
+	val.pathname = pathname;
```
anakryiko, line 39: why do you need `struct value`, if it has only one field? Why not store pathname directly?

libbpf-tools/statsnoop.c
```diff
@@ -0,0 +1,186 @@
...
+
+#define PERF_BUFFER_PAGES       16
+#define PERF_POLL_TIMEOUT_MS    100
+#define warn(...) fprintf(stderr, __VA_ARGS__)
+
+static volatile sig_atomic_t canceled = 0;
+
+pid_t target_pid = 0;
+bool trace_failed_only = false;
+bool output_with_timestamp = false;
```
anakryiko, line 25: statics?

libbpf-tools/statsnoop.c
```diff
@@ -0,0 +1,186 @@
...
+		break;
+	default:
+		return ARGP_ERR_UNKNOWN;
+	}
+	return 0;
+}
+
+static void sig_int(int signo)
+{
+	canceled = 1;
```
anakryiko, line 81: most tools call such variable 'exiting'

libbpf-tools/statsnoop.c
```diff
@@ -0,0 +1,186 @@
...
+}
+
+void handle_event(void *ctx, int cpu, void *data, __u32 data_sz)
```
anakryiko, line 84: static

libbpf-tools/statsnoop.c
```diff
@@ -0,0 +1,186 @@
...
+	double ts = 0.0;
+
+	if (e->ret >= 0) {
+		fd = e->ret;
+		err = 0;
+	} else {
+		fd = -1;
+		err = -e->ret;
+	}
+	if (!start_timestamp)
```
anakryiko, line 97: start_timestamp is used only inside handle_event, let's make it static within that function only, instead of for the entire file

libbpf-tools/statsnoop.c
```diff
@@ -0,0 +1,186 @@
...
+		start_timestamp = e->ts_ns;
+	if (output_with_timestamp) {
+		ts = (double)(e->ts_ns - start_timestamp) / 1000000000;
+		printf("%-11.6f", ts);
```
anakryiko, line 101: no spaces after %f?

also why bother with doubles, if you can just output seconds and microseconds separately with `/ 1000000000` and `/ 1000 % 1000000`?

libbpf-tools/statsnoop.c
```diff
@@ -0,0 +1,186 @@
...
+	}
+	printf("%-7d %-20s %-4d %-4d %-s\n", e->pid, e->comm, fd, err, e->pathname);
+}
+
+void handle_lost_events(void *ctx, int cpu, __u64 lost_cnt)
```
anakryiko, line 106: static

libbpf-tools/statsnoop.h
```diff
@@ -0,0 +1,20 @@
...
+#define NAME_MAX        255
+
+struct value {
+	const char *pathname;
+};
+
+struct event {
+	__u32 pid;
+	__u64 ts_ns;
+	int ret;
```
anakryiko, line 15: reorder `__u64, __u32, int` to avoid padding

libbpf-tools/statsnoop.h
```diff
@@ -0,0 +1,20 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+#ifndef __STATSNOOP_H
+#define __STATSNOOP_H
+
+#define TASK_COMM_LEN   16
+#define NAME_MAX        255
```
anakryiko, line 6: 256 for nice rounded (in binary) number?

anakryiko (changes requested):
Few high-level points.

1. Use raw_tracepoint instead of tracepoint, it's faster and has the same (or more) data available.
2. Better still, using fexit here would eliminate the need to keep per-tid state completely. You can check return result and use input argument all in one BPF program.

With fexit, I'd also argue to go for BPF ringbuf instead of BPF perfbuf to save memory and have a more succinct *and* performant BPF code.

libbpf-tools/statsnoop.c
chenhengqi: The output format is aligned with its counterpart in tools.

libbpf-tools/statsnoop.h
chenhengqi: `linux/limits.h` has this macro which also set to 255. Maybe because it can fit in a `u8` ?

anakryiko: oh, ok, it doesn't matter much, compiler will just pad one byte at the end. Please disregard, I didn't realize it's matching some Linux UAPI constant.

libbpf-tools/statsnoop.c
anakryiko: Right, I'm not talking about the width of the argument (11), but more about converting __u64 to double just to format it as seconds and microseconds. You can do the latter with just integer division without any rounding behavior of double. But it's minor, so if you insist them it's fine, it's not broken, just unnecessary.

chenhengqi: So be it. :)

I don't like to handle the alignment.

anakryiko (approved):
lgtm

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

PR #3423, opened 2021-05-12
libbpf-tools/opensnoop: disable open on aarch64

### LINKED ISSUE #3344: opensnoop on arm64 fails because no open syscall
(that is true for both libbpf-tools/opensnoop.bpf.c, the tools/opensnoop.py and bpftrace's opensnoop.bt)

I'm surprised I don't find any issue about this, for example the libbpf-tool variant fails at runtime like this:
```
libbpf: failed to determine tracepoint 'syscalls/sys_enter_open' perf event ID: No such file or directory
libbpf: prog 'tracepoint__syscalls__sys_enter_open': failed to create tracepoint 'syscalls/sys_enter_open' perf event: No such file or directory
libbpf: failed to auto-attach program 'tracepoint__syscalls__sys_enter_open': -2
failed to attach BPF programs
```

Is there a way to automatically detect if a tracepoint is available? I'm mostly only interested in the libbpf-tools variant at this point.
If possible it would be great to catch this kind of problems at compile time, even if it might not work on another kernel if we're able to ensure the probes exist through some generated header file at compile time (optionally?) that would help catch typos or such.

martinetd:
aarch64 has no open syscall, do not attempt to trace it.

Fixes #3344.

Sorry for the delay, here it is.
I've successfully tested opensnoop with the patch on aarch64 debian.

### DIFF
--- a/libbpf-tools/opensnoop.c
+++ b/libbpf-tools/opensnoop.c
@@ -242,6 +242,16 @@ int main(int argc, char **argv)
 	obj->rodata->targ_uid = env.uid;
 	obj->rodata->targ_failed = env.failed;
 
+#ifdef __aarch64__
+	/* aarch64 has no open syscall, only openat variants.
+	 * Disable associated tracepoints that do not exist. See #3344.
+	 */
+	bpf_program__set_autoload(
+		obj->progs.tracepoint__syscalls__sys_enter_open, false);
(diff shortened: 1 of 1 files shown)


anakryiko (approved):
Great, LGTM.

yonghong-song:
[buildbot, test this please]

PR #3424, opened 2021-05-12
GCC 7: Coverity 'Uninitialized scalar field'

rkubik:
Probably not a real issue but it doesn't hurt to initialize this field to suppress warnings.

```
1. member_decl: Class member declaration for code_.
Code code_;

2. Uninitialized scalar field (UNINIT_CTOR). uninit_member: Non-static class member code_ is not initialized in this constructor nor in any functions that it calls.
StatusTuple(int ret) : ret_(ret) {}
```

### DIFF
--- a/src/cc/bcc_exception.h
+++ b/src/cc/bcc_exception.h
@@ -78,7 +78,7 @@ private:
   int ret_;
 
   bool use_enum_code_ = false;
-  Code code_;
+  Code code_ = Code::UNKNOWN;
 
   std::string msg_;
 };


yonghong-song:
[buildbot, test this please]

yonghong-song:
Yes, it is not a real issue. `code_` is used only if `use_enum_code_` is true. And in the place `use_enum_code_` is set to true, `code_` is initialized. This may be a little bit complicated for compiler to figure out. So agree that let us just initialized `code_` to silence the compiler warning.

Issue #3407, opened 2021-05-07
Misuse of bpf_get_current_pid_tgid or PID should be TID ?


chenhengqi:
I found some tools use `u32 pid = bpf_get_current_pid_tgid` in their BPF program and expose a `PID` option, is this a misuse of bpf_get_current_pid_tgid or we should change `PID` to `TID` ?

- [x] btrfsdist.py
- [ ] bashreadline.py
- [ ] cachetop.py
- [ ] dcsnoop.py
- [x] ext4dist.py
- [x] gethostlatency.py
- [ ] killsnoop.py
- [ ] llcstat.py
- [ ] mdflush.py
- [ ] mysqld_qslower.py
- [x] nfsdist.py
- [x] solisten.py
- [x] sslsniff.py
- [x] statsnoop.py
- [x] tcpdrop.py
- [ ] wakeuptime.py
- [x] xfsdist.py
- [x] zfsdist.py

I don't know it's intended or I misunderstand this due to my ignorance. :)
Can anyone clarify this ?

brendangregg:
Please see prior tickets where we've discussed this and sorted it out. There may be some cases where the pid still needs to be fixed, but I can't recall offhand all the discussions we've had on this: I'd need to look them up just like you can. Also check the git history for cases where ones have been fixed.

chenhengqi:
Thanks, Brendan. (I am a big fan of you :))

chenhengqi:
Some related commits:

- https://github.com/iovisor/bcc/commit/15d7955e07a942340610f71e44f44046b960e3a8
- https://github.com/iovisor/bcc/pull/674/commits/66bf2e8e0bd7b7367f9a01502dbcac994a063be2
- https://github.com/iovisor/bcc/pull/739/commits/ca7ac22c2be51d6cbe7d7a48ea96f3cde3e50b05

chenhengqi:
The tools listed above either has a `PID` option or has a `PID` column in their output, but use `u32 pid = bpf_get_current_pid_tgid` directly.

I think some of these tools still need to be fixed.

yonghong-song:
@chenhengqi If you feel it needs a change, please help submit a patch. In the past, we indeed made some adjustment from tid (incorrectly named pid) -> real pid. Thanks!

brendangregg:
[...]

PR #3433, opened 2021-05-16
tools: filter/display using PID intead of TID

chenhengqi:
This PR fixes PID filtering in

* gethostlatency.py
* solisten.py
* sslsniff.py
* tcpdrop.py

For background, see #3407 .

### DIFF
--- a/tools/gethostlatency.py
+++ b/tools/gethostlatency.py
@@ -61,14 +61,16 @@ int do_entry(struct pt_regs *ctx) {
         return 0;
 
     struct val_t val = {};
-    u32 pid = bpf_get_current_pid_tgid();
+    u64 pid_tgid = bpf_get_current_pid_tgid();
+    u32 pid = pid_tgid >> 32;
+    u32 tid = (u32)pid_tgid;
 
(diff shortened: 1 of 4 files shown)


chenhengqi:
The `tcpdrop.py` not works in recent kernel. Maybe due to the `tcp_drop` function get inlined.

I wonder if there are workarounds for that.

yonghong-song:
[buildbot, test this please]

yonghong-song:
> The `tcpdrop.py` not works in recent kernel. Maybe due to the `tcp_drop` function get inlined.
> 
> I wonder if there are workarounds for that.

Unfortunately, we don't have workaround for this if tcp_drop is completely gone due to inlining. People may be more interesting in what exactly leads to tcp drop. There is a recent commit
   https://git.kernel.org/pub/scm/linux/kernel/git/netdev/net-next.git/commit/?id=709c0314239992162cba26a860f04319a15860c4
will adds a new tracepoint for csum_errors. I think tcp_drop can be enhanced to utilize this tracepoint to list drops (kfree_sub's) due to csum errors.

chenhengqi:
Thanks for the insight. Will try.

PR #3436, opened 2021-05-17
libbpf-tools: add fsdist

chenhengqi:

### DIFF
--- a/libbpf-tools/.gitignore
+++ b/libbpf-tools/.gitignore
@@ -15,6 +15,7 @@
 /gethostlatency
 /hardirqs
 /llcstat
+/nfsdist
 /numamove
 /offcputime
 /opensnoop
--- a/libbpf-tools/Makefile
+++ b/libbpf-tools/Makefile
@@ -32,6 +32,7 @@ APPS = \
 	gethostlatency \
 	hardirqs \
 	llcstat \
+	nfsdist \
 	numamove \
 	offcputime \
 	opensnoop \
--- a/libbpf-tools/nfsdist.bpf.c
+++ b/libbpf-tools/nfsdist.bpf.c
@@ -1,0 +1,193 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2021 Hengqi Chen
+#include <vmlinux.h>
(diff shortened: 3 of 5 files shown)


anakryiko (commented):
See feedback for xfsdist, it applies here 100% with s/xfs/nfs/ :)

I also started wondering if it makes sense to try to generalise this as sort sort of fsdist tool that would work for ext4, xfs, nfs, btrfs, etcfs :) Thoughts? They all are almost identical.

chenhengqi:
> See feedback for xfsdist, it applies here 100% with s/xfs/nfs/ :)
> 
> I also started wondering if it makes sense to try to generalise this as sort sort of fsdist tool that would work for ext4, xfs, nfs, btrfs, etcfs :) Thoughts? They all are almost identical.

Yeah, but I am afraid that would break something like design principles / Unix philosophy.

The tools directory contains [ext4/nfs/xfs/btrfs]slower and [ext4/nfs/xfs/btrfs]dist, should we follow this ?

anakryiko:
> > See feedback for xfsdist, it applies here 100% with s/xfs/nfs/ :)
> > I also started wondering if it makes sense to try to generalise this as sort sort of fsdist tool that would work for ext4, xfs, nfs, btrfs, etcfs :) Thoughts? They all are almost identical.
> 
> Yeah, but I am afraid that would break something like design principles / Unix philosophy.

If anything, it will apply DRY where it's due. Any improvements or extra features will apply to all FS-specific tools at the same time without 5x duplication. The more I think about this, the more I like it. It's pretty easy to generalize this to generic fsdist tool, which will accept FS type as an argument. But also, if such tool detects that it is actually called as, say, nfsdist, it will just assume nfs as a FS. So in the end we'll physically have one tool with common functionality, a single embedded libbpf, yet which would be easily handling multiple FSs. And then we can just create few symlinks from ext4dist -> fsdist, btrfsdist -> fsdist, and so on.

> 
> The tools directory contains [ext4/nfs/xfs/btrfs]slower and [ext4/nfs/xfs/btrfs]dist, should we follow this ?

Yeah, slower variants should probably be generalized as well.

PR #3440, opened 2021-05-18
finish to add support of subset in items_*_batch()

egobillot:
Hello,
Here is a partial PR in order to :
 - add items_lookup_batch() on subset keys
 - add items_lookup_and_delete_batch() on subset keys
 - add docstring on items_*_batch()
 - update the reference_guide.md
This is a partial PT because I am stuck with items_lookup_batch and items_lookup_and_delete_batch on a subset. Sometimes tests are Ok and sometimes they fails. See the output of tests.
``` 
python git:(add_subset_for_batch_ops) ✗ sudo python3 ./test_map_batch_ops.py
...F....
FAIL: test_lookup_and_delete_batch_subset (__main__.TestMapBatch)
  File "./test_map_batch_ops.py", line 94, in test_lookup_and_delete_batch_subset
    self.assertEqual(count, self.SUBSET_SIZE)
AssertionError: 31 != 32

Ran 8 tests in 2.635s

FAILED (failures=1)
➜  python git:(add_subset_for_batch_ops) ✗ sudo python3 ./test_map_batch_ops.py
........
Ran 8 tests in 2.613s

OK
```
Do I need to call the lib.bpf_lookup_and_delete_batch() several times until the sum of the nb of elements returned matches the subset size ? For example, If my subset size is 32 and get only 31 elem on the first call, do I need to call it again for the last element ? :thinking:

### DIFF
--- a/src/python/bcc/table.py
+++ b/src/python/bcc/table.py
@@ -399,19 +399,99 @@ class TableBase(MutableMapping):
         for k in self.keys():
             self.__delitem__(k)
 
-    def items_lookup_batch(self):
-        # batch size is set to the maximum
-        batch_size = self.max_entries
-        out_batch = ct.c_uint32(0)
-        keys = (type(self.Key()) * batch_size)()
-        values = (type(self.Leaf()) * batch_size)()
-        count = ct.c_uint32(batch_size)
+    def _alloc_keys_values(self, alloc_k=False, alloc_v=False, count=None):
+        """Allocate keys and/or values arrays. Useful for in items_*_batch.
+
+        Args:
+            alloc_k (bool): True to allocate keys array, False otherwise.
+            Default is False.
+            alloc_v (bool): True to allocate values array, False otherwise.
+            Default is False.
+            count (int): number of elements in the array(s) to allocate. If
+            count is None then it allocates the maximum nb of elements i.e
+            self.max_entries.
+
+        Returns:
+            tuple: (count, keys, values). Where count is ct.c_uint32,
+            and keys and values an instance of ct.Array
+        Raises:
+            ValueError: If count is not < 1 or > self.max_entries
+        """
+        keys = values = None
+        if not alloc_k and not alloc_v:
+            return (ct.c_uint32(0), None, None)
+
+        if not count:  # means alloc maximum size
+            count = self.max_entries
+        elif count < 1 or count > self.max_entries:
+            raise ValueError("Wrong count")
+
+        if alloc_k:
+            #keys = (self.Key * count)()
+            keys = (self.Key * self.max_entries)()
+        if alloc_v:
+            #values = (self.Leaf * count)()
+            values = (self.Leaf * self.max_entries)()
+
+        return (ct.c_uint32(count), keys, values)
+
+    def sanity_check_keys_values(self, keys=None, values=None):
+        """Check if the given keys or values have the right type and size.
+
+        Args:
+            keys (ct.Array): keys array to check
+            values (ct.Array): values array to check
+        Returns:
+            ct.c_uint32 : the size of the array(s)
+        Raises:
+            ValueError: If length of arrays is < 1 or > self.max_entries, or
+            when both arrays length are different.
+            TypeError: If the keys and values are not an instance of ct.Array
+        """
+        arr_len = 0
+        for elem in [keys, values]:
+            if elem:
+                if not isinstance(elem, ct.Array):
+                    raise TypeError
+
+                arr_len = len(elem)
+                if arr_len < 1 or arr_len > self.max_entries:
+                    raise ValueError("Array's length is wrong")
+
+        if keys and values:
+            # check both length are equal
+            if len(keys) != len(values):
+                raise ValueError("keys array length != values array length")
+
+        return ct.c_uint32(arr_len)
+
+    def items_lookup_batch(self, ct_keys=None):
+        """Lookup the key-value pairs related to the keys given as parameters.
+
+        Args:
+            ct_keys (ct.Array): keys array to lookup. If an array
+            of keys is given then it lookups all the related keys-values.
+            If keys is None (default) then it looks up every entries.
+        Yields:
+            tuple: The tuple of (key,value) for every entries that have
+            been looked up.
+        Raises:
+            Exception: If bpf syscall has failed or is invalid
+        """
+        if ct_keys is None:
+            ct_cnt, ct_keys, ct_values = self._alloc_keys_values(alloc_k=True,
+                                                                 alloc_v=True)
+        else:
+            ct_cnt = self.sanity_check_keys_values(keys=ct_keys)
+            ct_cnt, _, ct_values = self._alloc_keys_values(alloc_v=True,
+                                                           count=ct_cnt.value)
+
         res = lib.bpf_lookup_batch(self.map_fd,
                                    None,
-                                   ct.byref(out_batch),
-                                   ct.byref(keys),
-                                   ct.byref(values),
-                                   ct.byref(count)
+                                   ct.byref(ct_cnt),
+                                   ct.byref(ct_keys),
+                                   ct.byref(ct_values),
+                                   ct.byref(ct_cnt)
                                    )
 
         errcode = ct.get_errno()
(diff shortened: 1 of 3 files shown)


egobillot:
Thank you @yonghong-song . It was not clear to me that subset batch was not supported for lookup and lookup_and_delete by the kernel. So I remove the code regarding subset for lookup_*.

yonghong-song:
[buildbot, test this please]

src/python/bcc/table.py
```diff
@@ -399,19 +399,91 @@ def clear(self):
...
+    def _sanity_check_keys_values(self, keys=None, values=None):
+        """Check if the given keys or values have the right type and size.
+
+        Args:
+            keys (ct.Array): keys array to check
+            values (ct.Array): values array to check
+        Returns:
+            ct.c_uint32 : the size of the array(s)
+        Raises:
+            ValueError: If length of arrays is < 1 or > self.max_entries, or
```
yonghong-song, line 445: is less than 1 or greater than self.max_entries?

src/python/bcc/table.py
```diff
@@ -399,19 +399,91 @@ def clear(self):
...
-        count = ct.c_uint32(batch_size)
+        """Lookup the key-value pairs related to the keys given as parameters.
+
+        Args:
+            None
+        Yields:
+            tuple: The tuple of (key,value) for every entries that have
+            been looked up.
+        Raises:
+            Exception: If bpf syscall has failed or is invalid
```
yonghong-song, line 475: How about `Exception: if bpf syscall return value indicates an error`

src/python/bcc/table.py
```diff
@@ -399,19 +399,91 @@ def clear(self):
...
+        ct_cnt, ct_keys, ct_values = self._alloc_keys_values(alloc_k=True,
+                                                             alloc_v=True)
+
         res = lib.bpf_lookup_batch(self.map_fd,
                                    None,
-                                   ct.byref(out_batch),
-                                   ct.byref(keys),
-                                   ct.byref(values),
-                                   ct.byref(count)
+                                   ct.byref(ct_cnt),
```
yonghong-song, line 483: To avoid future confusion, let us just use a separate variable `out_batch` here by not overloading `ct_cnt`.

src/python/bcc/table.py
```diff
@@ -399,19 +399,91 @@ def clear(self):
...
+                                   ct.byref(ct_keys),
+                                   ct.byref(ct_values),
+                                   ct.byref(ct_cnt)
                                    )
```
yonghong-song, line 487: Think it more. I think we may need a loop here. The following is the scenario. Say there are five buckets, max_entries = 5,
   1. lookup_batch find there are 3 elements in bucket 0, and copied three key/values.
   2. lookup_batch finished bucket 0 and not lock bucket 1 yet, in such an interval, some other threads/processes may delete all 3 elements in bucket 0 and add 3 new elements in bucket 1.
   3. lookup_batch lock bucket 1 and find 3 elements, not able to copy all of them and hence returns to user space with 3 elements.

So at the end, we got 3 elements, out_batch = 1 which indicates next time to start with bucket 1 internally.
The above case should be really rare but it may still happen if there are intensive mixed lookup/update/delete operations. But with max_entries as a buffer size, we are guaranteed to make further progress at least one bucket at a time and the loop will eventually terminate.

For checking return code, a successful syscall return codes should be:
  0 indicating successful get some values and not finished yet
  errno.ENOENT: done
Any other error code indicates an error, we don't need to single out errno.EINVAL here.

You can take a look at linux:tools/testing/selftests/bpf/map_tests/htab_map_batch_ops.c for more details.

src/python/bcc/table.py
```diff
@@ -421,31 +493,30 @@ def items_lookup_batch(self):
...
+    def items_delete_batch(self, ct_keys=None):
+        """Delete the key-value pairs related to the keys given as parameters.
+        Note that if no key are given, it is faster to call
+        lib.bpf_lookup_and_delete_batch than create keys array and then call
+        lib.bpf_delete_batch on these keys.
+
+        Args:
+            ct_keys (ct.Array): keys array to delete. If an array of keys is
+            given then it deletes all the related keys-values.
+            If keys is None (default) then it deletes every entries.
```
yonghong-song, line 508: deletes all entries.

src/python/bcc/table.py
```diff
@@ -457,25 +528,22 @@ def items_delete_batch(self, keys=None):
...
         """Update all the key-value pairs in the map provided.
-        The lists must be the same length, between 1 and the maximum number of entries.
+        The arrays must be the same length, between 1 and the maximum number
+        of entries.
+
+        Args:
+            ct_keys (ct.Array): keys array to update
+            ct_values (ct.Array): values array to update
+        Raises:
+            Exception: If bpf syscall has failed or is invalid
```
yonghong-song, line 540: Exception: if bpf syscall returns an error

src/python/bcc/table.py
```diff
@@ -421,31 +493,30 @@ def items_lookup_batch(self):
...
+
+        Args:
+            ct_keys (ct.Array): keys array to delete. If an array of keys is
+            given then it deletes all the related keys-values.
+            If keys is None (default) then it deletes every entries.
+        Yields:
+            tuple: The tuple of (key,value) for every entries that have
+            been deleted.
+        Raises:
+            Exception: If bpf syscall has failed or is invalid
```
yonghong-song, line 513: if bpf syscall returns an error.

src/python/bcc/table.py
```diff
@@ -421,31 +493,30 @@ def items_lookup_batch(self):
...
+        if ct_keys is not None:
+            ct_cnt = self._sanity_check_keys_values(keys=ct_keys)
             res = lib.bpf_delete_batch(self.map_fd,
-                                       ct.byref(keys),
-                                       ct.byref(count)
+                                       ct.byref(ct_keys),
+                                       ct.byref(ct_cnt)
                                        )
             errcode = ct.get_errno()
             if (errcode == errno.EINVAL):
```
yonghong-song, line 522: The error chacking here and below is not quite right. The delete_batch and update_batch use kernel internal function `generic_map_{delete,update}_batch` function, they return 0 to indicate a success and non-0 as an error code. Just check errcode is 0 or not is enough.

src/python/bcc/table.py
```diff
@@ -457,25 +528,22 @@ def items_delete_batch(self, keys=None):
...
         res = lib.bpf_update_batch(self.map_fd,
-                                   ct.byref(keys),
-                                   ct.byref(values),
-                                   ct.byref(count)
+                                   ct.byref(ct_keys),
+                                   ct.byref(ct_values),
+                                   ct.byref(ct_cnt)
                                    )
 
         errcode = ct.get_errno()
```
yonghong-song, line 549: similar to delete_batch, just check whether errcode is 0 or not is enough below.

src/python/bcc/table.py
```diff
@@ -486,18 +554,29 @@ def items_update_batch(self, keys, values):
...
+        """Look up and delete the key-value pairs related to the keys given
+        as parameters.
+
+        Args:
+            None
+        Yields:
+            tuple: The tuple of (key,value) for every entries that have
+            been looked up and deleted.
+        Raises:
+            Exception: If bpf syscall has failed or is invalid
```
yonghong-song, line 566: if bpf syscall return code indicates an error.

src/python/bcc/table.py
```diff
@@ -486,18 +554,29 @@ def items_update_batch(self, keys, values):
...
         res = lib.bpf_lookup_and_delete_batch(self.map_fd,
                                               None,
-                                              ct.byref(out_batch),
-                                              ct.byref(keys),
-                                              ct.byref(values),
-                                              ct.byref(count)
+                                              ct.byref(ct_cnt),
+                                              ct.byref(ct_keys),
+                                              ct.byref(ct_values),
+                                              ct.byref(ct_cnt)
```
yonghong-song, line 579: similar to lookup_batch, we may need a loop here.

yonghong-song (commented):
Added some comments. Most importantly for the robust implementation, I suggest to have a loop for lookup_batch and lookup_and_delete_batch, could you take a look?

egobillot:
Yes, I will have a look at your comments.

egobillot:
@yonghong-song It should be better now with https://github.com/iovisor/bcc/pull/3440/commits/7cbf766e2d0b6c60787f73999f607d0375adac40 . I have factorized the code for items_lookup_and_delete_batch() and items_lookup_batch() since it was almost the same. 
I have also took into account all your comments. I hope I did not missed one.

yonghong-song:
@egobillot Thanks for the update! Much better. I just added one more comment and I think we should be fine after that.

yonghong-song:
[buildbot, ok to test]

src/python/bcc/table.py
```diff
@@ -399,116 +399,196 @@ def clear(self):
...
-        for i in range(0, count.value):
-            yield (keys[i], values[i])
+        total = 0
+        while True:
+            res = bpf_batch(self.map_fd,
+                            ct.byref(out_batch) if total else None,
+                            ct.byref(out_batch),
+                            ct.byref(ct_keys, ct.sizeof(self.Key) * total),
+                            ct.byref(ct_values, ct.sizeof(self.Leaf) * total),
+                            ct.byref(ct_cnt)
```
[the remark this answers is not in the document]
egobillot: Tell me what you think of https://github.com/iovisor/bcc/pull/3440/commits/59394fbf054b942223c38b5e4d552c28f4988254

yonghong-song:
LGTM,  Thanks!

PR #3441, opened 2021-05-19
libbpf-tools: add fsdist

chenhengqi:

### DIFF
--- a/libbpf-tools/fsdist.c
+++ b/libbpf-tools/fsdist.c
@@ -1,0 +1,453 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+
+/*
+ * fsdist  Summarize file system operations latency.
+ *
+ * Copyright (c) 2021 Hengqi Chen
+ * 20-May-2021   Hengqi Chen   Created this.
+ */
+#include <argp.h>
+#include <libgen.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+#include <unistd.h>
+
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+
+#include "fsdist.h"
+#include "fsdist.skel.h"
+#include "trace_helpers.h"
+
+#define warn(...) fprintf(stderr, __VA_ARGS__)
+
+enum fs_type {
+	NONE,
+	BTRFS,
+	EXT4,
+	NFS,
+	XFS,
+};
+
+static struct fs_config {
+	const char *fs;
+	const char *op_funcs[MAX_OP];
+} fs_configs[] = {
+	[BTRFS] = { "btrfs", {
+		[READ] = "btrfs_file_read_iter",
+		[WRITE] = "btrfs_file_write_iter",
+		[OPEN] = "btrfs_file_open",
+		[FSYNC] = "btrfs_sync_file",
+		[GETATTR] = NULL, /* not supported */
+	}},
+	[EXT4] = { "ext4", {
+		[READ] = "ext4_file_read_iter",
+		[WRITE] = "ext4_file_write_iter",
+		[OPEN] = "ext4_file_open",
+		[FSYNC] = "ext4_sync_file",
+		[GETATTR] = "ext4_file_getattr",
+	}},
+	[NFS] = { "nfs", {
+		[READ] = "nfs_file_read",
+		[WRITE] = "nfs_file_write",
+		[OPEN] = "nfs_file_open",
+		[FSYNC] = "nfs_file_fsync",
+		[GETATTR] = "nfs_getattr",
+	}},
+	[XFS] = { "xfs", {
+		[READ] = "xfs_file_read_iter",
+		[WRITE] = "xfs_file_write_iter",
+		[OPEN] = "xfs_file_open",
+		[FSYNC] = "xfs_file_fsync",
+		[GETATTR] = NULL, /* not supported */
+	}},
+};
+
+static char *file_op_names[] = {
+	[READ] = "read",
+	[WRITE] = "write",
+	[OPEN] = "open",
+	[FSYNC] = "fsync",
+	[GETATTR] = "getattr",
+};
+
+static struct hist zero;
+static volatile sig_atomic_t exiting;
+
+/* options */
+static enum fs_type fs_type = NONE;
+static bool emit_timestamp = false;
+static bool timestamp_in_ms = false;
+static pid_t target_pid = 0;
+static int interval = 99999999;
+static int count = 99999999;
+static bool verbose = false;
+
+const char *argp_program_version = "fsdist 0.1";
+const char *argp_program_bug_address =
+	"https://github.com/iovisor/bcc/tree/master/libbpf-tools";
+const char argp_program_doc[] =
+"Summarize file system operations latency.\n"
+"\n"
+"Usage: fsdist [-h] [-t] [-T] [-m] [-p PID] [interval] [count]\n"
+"\n"
+"EXAMPLES:\n"
+"    fsdist -t ext4             # show ext4 operations latency as a histogram\n"
+"    fsdist -t nfs -p 1216      # trace nfs operations with PID 1216 only\n"
+"    fsdist -t xfs 1 10         # trace xfs operations, 1s summaries, 10 times\n"
+"    fsdist -t btrfs -m 5       # trace btrfs operation, 5s summaries, in ms\n";
+
+static const struct argp_option opts[] = {
+	{ "timestamp", 'T', NULL, 0, "Print timestamp" },
+	{ "milliseconds", 'm', NULL, 0, "Millisecond histogram" },
+	{ "pid", 'p', "PID", 0, "Process ID to trace" },
+	{ "type", 't', "Filesystem", 0, "Which filesystem to trace, [btrfs/ext4/nfs/xfs]" },
+	{ "verbose", 'v', NULL, 0, "Verbose debug output" },
+	{ NULL, 'h', NULL, OPTION_HIDDEN, "Show the full help" },
+	{},
+};
+
+static error_t parse_arg(int key, char *arg, struct argp_state *state)
+{
+	static int pos_args;
+
+	switch (key) {
+	case 'v':
+		verbose = true;
+		break;
+	case 'T':
+		emit_timestamp = true;
+		break;
+	case 'm':
+		timestamp_in_ms = true;
+		break;
+	case 't':
+		if (!strcmp(arg, "btrfs")) {
+			fs_type = BTRFS;
+		} else if (!strcmp(arg, "ext4")) {
+			fs_type = EXT4;
+		} else if (!strcmp(arg, "nfs")) {
+			fs_type = NFS;
+		} else if (!strcmp(arg, "xfs")) {
+			fs_type = XFS;
+		} else {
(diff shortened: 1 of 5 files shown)


libbpf-tools/fsdist.c
```diff
@@ -0,0 +1,373 @@
...
+const char *argp_program_bug_address =
+	"https://github.com/iovisor/bcc/tree/master/libbpf-tools";
+const char argp_program_doc[] =
+"Summarize btrfs/ext4/nfs/xfs operation latency.\n"
+"\n"
+"Usage: fsdist [-h] [-x] [-T] [-m] [-p PID] [interval] [count]\n"
+"\n"
+"EXAMPLES:\n"
+"    fsdist                     # show ext4 operation latency as a histogram\n"
+"    fsdist -x nfs -p 1216      # trace nfs operation with PID 1216 only\n"
```
anakryiko, line 50: -x seems like a weird flag for filesystem type, why not `-t`, similar to how `mount -t nfs` works?

libbpf-tools/fsdist.h
```diff
@@ -0,0 +1,20 @@
...
+enum fs_file_op {
+	READ,
+	WRITE,
+	OPEN,
+	FSYNC,
+	GETATTR,
+	MAX_OP,
+};
+
+#define MAX_SLOTS	32
```
anakryiko, line 14: why the bump in number of slots? ext4dist uses 27, is that not enough?

libbpf-tools/fsdist.c
```diff
@@ -0,0 +1,373 @@
...
+		enable_ext4fs_probes(skel);
+		break;
+	case NFS:
+		enable_nfs_probes(skel);
+		break;
+	case XFS:
+		enable_xfs_probes(skel);
+		break;
+	}
+
```
anakryiko, line 333: So I was thinking to avoid so many BPF programs and hard-coding all those fs-specific function names.

Think about this approach.

1. You define "generic" fentry/fexit programs for each possible type of fs_op. You don't even have to specify real function names in .bpf.c, because we are going to override them at runtime. Similarly, you define BPF_KRETPROBE for each such operation, again with fake attach function names. You can probably re-use the same BPF_KPROBE, btw, given they do the same thing that's not operation-specific.

2. On user-space side, we have a table of supported file systems and operations (and some operations might not be supported at all, like NFS has more stuff than other FSs):
```
struct {
    const char *fs;
    const char *op_funcs[MAX_OP];
} fs_configs[] = {
  { "btrfs", {
      [OPEN] = "ext4_file_open",
      [FSYNC] = "ext4_sync_file",
      [GETATTR] = NULL, /* not supported */
  },
  { "nfs", { ... } },
  /* and so on */
}
```

3. At runtime you search for your desired FS, iterate over supported operations and see which kernel function has to be traced. You do fentry_exists() check and either pick, say, generic fentry_open/fexit_open pair (for OPEN operation) or disable them and pick kprobe/kretrprobe. For fentry/fexit you have to use bpf_program__set_attach_target() call to specify exact function you want to eventually attach to. That has to happen before skeleton load. For kprobe/kretprobe, you'll do bpf_program__attach_kprobe() after skeleton loads (so you'll have two loops to handle both fentry and kprobe, that's a mild complication).

And that's it. Your BPF side will have just minimum necessary set of BPF programs. Your user-space will be easily extensible to new FSs by just adding "config". Done.

4. Let's also add the logic of checking with which name fsdist was exec()'ed. If it is, say, ext4dist, then just automatically pretend if `-t ext4` was provided. Then we can just have symlinks from ext4dist to fsdist to preserve the behavior of dedicated ext4dist, btrfsdist, xfsdist, and so on.

chenhengqi: Ah, thanks. Nice idea, will do in this way. :)

(When I implemented this, I found myself didn't follow the DRY rule, but couldn't figure out a good way to organize the code.)

libbpf-tools/fsdist.c
chenhengqi: My first thought was to avoid `-t` since it was used for `TID` and `timestamp` in other tools.

Will change it to `-t` just like the `mount` utility.

chenhengqi: My first thought was to avoid `-t` since it was used for `TID` and `timestamp` in other tools.

Will change it to `-t` just like the `mount` utility.

libbpf-tools/fsdist.h
chenhengqi: For NFS we may need a large value.

libbpf-tools/fsdist.c
chenhengqi: > checking with which name fsdist was exec()'ed

Do you mean `basename(argv[0])` ? I don't know how to do that. :facepalm:

libbpf-tools/fsdist.c
```diff
@@ -0,0 +1,489 @@
...
+	[WRITE] = "write",
+	[OPEN] = "open",
+	[FSYNC] = "fsync",
+	[GETATTR] = "getattr",
+};
+
+static struct hist zero;
+static volatile sig_atomic_t exiting;
+
+// options
```
anakryiko, line 74: C++ comment, please use /* */

libbpf-tools/fsdist.c
```diff
@@ -0,0 +1,489 @@
...
+#include <bpf/bpf.h>
+
+#include "fsdist.h"
+#include "fsdist.skel.h"
+#include "trace_helpers.h"
+
+#define warn(...) fprintf(stderr, __VA_ARGS__)
+
+enum fs_type {
+	BTRFS,
```
anakryiko, line 23: it's better to have also NONE for default (invalid) value

libbpf-tools/fsdist.c
```diff
@@ -0,0 +1,489 @@
...
+	[OPEN] = "open",
+	[FSYNC] = "fsync",
+	[GETATTR] = "getattr",
+};
+
+static struct hist zero;
+static volatile sig_atomic_t exiting;
+
+// options
+static enum fs_type fs_type = EXT4;
```
anakryiko, line 75: see above, we shouldn't default to EXT4, start with NONE and require it to be either specified (if it's used as `fstype -t ext4` or derived, if called as `ext4dist`)

libbpf-tools/fsdist.c
```diff
@@ -0,0 +1,489 @@
...
+static pid_t target_pid = 0;
+static int interval = 99999999;
+static int count = 99999999;
+static bool verbose = false;
+
+const char *argp_program_version = "fsdist 0.1";
+const char *argp_program_bug_address =
+	"https://github.com/iovisor/bcc/tree/master/libbpf-tools";
+const char argp_program_doc[] =
+"Summarize btrfs/ext4/nfs/xfs operation latency.\n"
```
anakryiko, line 87: "Summarize file system operations latency." to keep it short and generic?

Then we can have a list of supported file systems below.

libbpf-tools/fsdist.c
```diff
@@ -0,0 +1,489 @@
...
+			support_fentry = false;
+			break;
+		}
+	}
+	return support_fentry;
+}
+
+static void fentry_set_attach_target(struct fsdist_bpf *obj)
+{
+	bpf_program__set_attach_target(obj->progs.file_read_fentry, 0,
```
anakryiko, line 223:
```
struct fs_config cfg = &fs_configs[fs_type];

bpf_program__set_attach_target(obj->progs.file_read_fexit, 0, cfg->op_funcs[READ]);
```

And keep all of the below single line?

libbpf-tools/fsdist.c
```diff
@@ -0,0 +1,489 @@
...
+	bpf_program__set_autoload(obj->progs.getattr_entry, false);
+	bpf_program__set_autoload(obj->progs.getattr_exit, false);
+}
+
+static int attach_kprobes(struct fsdist_bpf *obj)
+{
+	long err;
+
+	// READ
+	obj->links.file_read_entry = bpf_program__attach_kprobe(
```
anakryiko, line 283: try to keep them on a single line as well, it's way more readable. It's fine to have a bit longer lines, if necessary

libbpf-tools/fsdist.c
```diff
@@ -0,0 +1,489 @@
...
+		obj->progs.file_read_exit,
+		true,
+		fs_configs[fs_type].op_funcs[READ]
+	);
+	err = libbpf_get_error(obj->links.file_read_exit);
+	if (err) {
+		warn("failed to attach kprobe: %ld\n", err);
+		return -1;
+	}
+	// WRITE
```
anakryiko, line 303: no C++ style comments, please

libbpf-tools/fsdist.c
```diff
@@ -0,0 +1,489 @@
...
+		fs_configs[fs_type].op_funcs[READ]);
+	bpf_program__set_attach_target(obj->progs.file_read_fexit, 0,
+		fs_configs[fs_type].op_funcs[READ]);
+	bpf_program__set_attach_target(obj->progs.file_write_fentry, 0,
+		fs_configs[fs_type].op_funcs[WRITE]);
+	bpf_program__set_attach_target(obj->progs.file_write_fexit, 0,
+		fs_configs[fs_type].op_funcs[WRITE]);
+	bpf_program__set_attach_target(obj->progs.file_open_fentry, 0,
+		fs_configs[fs_type].op_funcs[OPEN]);
+	bpf_program__set_attach_target(obj->progs.file_open_fexit, 0,
```
anakryiko, line 233: you need to check errors for each set_attach_target

you can do it concisely:

```
int err = 0;

err = err ?: bpf_program__set_attach_target(...);
err = err ?: bpf_program__set_attach_target(...);
err = err ?: bpf_program__set_attach_target(...);
err = err ?: bpf_program__set_attach_target(...);

if (err) { too bad }
```

libbpf-tools/fsdist.c
```diff
@@ -0,0 +1,489 @@
...
+
+	skel->rodata->target_pid = target_pid;
+	skel->rodata->in_ms = timestamp_in_ms;
+
+	// before load
+	// if fentry is supported, we set attach target and disable kprobes
+	// otherwise, we disable fentry and attach kprobes after loading
+	support_fentry = check_fentry();
+	if (support_fentry) {
+		fentry_set_attach_target(skel);
```
anakryiko, line 432: this should be a failing function, see above

libbpf-tools/fsdist.c
```diff
@@ -0,0 +1,489 @@
...
+			warn("failed to attach BPF programs: %d\n", err);
+			goto cleanup;
+		}
+	} else {
+		err = attach_kprobes(skel);
+		if (err) {
+			warn("failed to attach BPF programs: %d\n", err);
+			goto cleanup;
+		}
+	}
```
anakryiko, line 459: no need to duplicate error checking:

```
err = support_fentry ? fsdist_bpf__attach(skel) : attach_kprobes(skel);
if (err) {
}
```

libbpf-tools/fsdist.c
anakryiko: yeah, something like that. Doesn't it work? I've seen perf tool doing something like that.

anakryiko (approved):
Very nice, thank you!

Please follow up with removal of existing xxxdist tools and creating symlinks for them.

chenhengqi:
ping @yonghong-song

yonghong-song:
[buildbot, test this please]

PR #3443, opened 2021-05-21
Update tcpconnect to use "__u32 af" instead of "int af"

marselester:
Hi! Please consider updating `tcpconnect` to use `__u32 af` instead of `int af` so it's less error prone to decode an address family on Go side, for example https://github.com/marselester/libbpf-tools/blob/master/cmd/tcpconnect/main.go.


yonghong-song:
[buildbot, test this please]

yonghong-song:
Agreed. It indeed makes sense to use __u32 in this case.

Issue #3453, opened 2021-05-25
tools/deadlock.py could not open bpf map: thread_to_held_mutexes, error: Cannot allocate memory


chenyuezhou:
```
$./deadlock.py 101565 --binary /lib64/libpthread.so.0
could not open bpf map: thread_to_held_mutexes, error: Cannot allocate memory
Traceback (most recent call last):
  File "./deadlock.py", line 556, in <module>
    main()
  File "./deadlock.py", line 468, in main
    bpf = BPF(src_file=b'deadlock.c')
  File "/usr/lib/python2.7/site-packages/bcc/__init__.py", line 372, in __init__
    raise Exception("Failed to compile BPF module %s" % (src_file or "<text>"))
Exception: Failed to compile BPF module deadlock.c
```

`libbcc` seems to have set `RLIMIT_MEMLOCK` to `RLIM_INFINITY`. 

I wonder what caused `bpf_create_map_xattr` to fail.

yonghong-song:
The memory requirement for this map `thread_to_held_mutexes` actually is very big:
```
  2097152 * (8 + 16 * 16) = 553648128
```
Could you reduce its size in `deadlock.c` to see whether it passed or not.

Which kernel are you using? In the recent/latest kernel (I didn't check versions), bpf program map/prog memory
won't compared against locked memory any more, instead it compared against cgroup memory. Maybe this is the reason?

chenyuezhou:
@yonghong-song Thanks.

```
$uname -r
5.10.35
```

I've tried to reduce the size of `deadlock.c`,  and it works fine.

reducing the size of `deadlock.c` may not be the best solution. I want to know exactly what limits it has exceeded, and maybe we can output better error messages in `deadlock.c`.

chenyuezhou:
```
$uname -r 
5.10.35
```

I checked my kernel.

The kernel still seems to check for locked memory.

`kernel/bpf/hashtab.c@htab_map_alloc+471`

```
/* if map size is larger than memlock limit, reject it */
err = bpf_map_charge_init(&htab->map.memory, cost);
```

I use `tools/trace.py` find.

```
$./trace.py 'r::bpf_map_area_alloc (!retval)'
PID     TID     COMM            FUNC
5004    5004    deadlock.py     bpf_map_area_alloc
```

`kernel/bpf/hashtab.c@htab_map_alloc+477`

```
err = -ENOMEM;
htab->buckets = bpf_map_area_alloc(htab->n_buckets * 
[...]

PR #3455, opened 2021-05-27
tools/deadlock: support specifies maxnum of threads and edge cases

chenyuezhou:
https://github.com/iovisor/bcc/issues/3453

support specifies maxnum of threads and edge cases.

The default size of memory that deadlock needs to allocate is too large. Sometimes this will cause OOM for particular system, or the memory limit has been exceeded.

```
[16536.857153] oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=/,mems_allowed=0,global_oom,task_memcg=/user.slice/user-0.slice/session-1.scope,task=trace.py,pid=4971,uid=0
[16536.857622] Out of memory: Killed process 4971 (trace.py) total-vm:470116kB, anon-rss:0kB, file-rss:0kB, shmem-rss:0kB, UID:0 pgtables:528kB oom_score_adj:0
```

### DIFF
--- a/tools/deadlock.py
+++ b/tools/deadlock.py
@@ -475,7 +475,8 @@ def main():
             print('%s. Is the process (pid=%d) running?' % (str(e), args.pid))
             sys.exit(1)
 
-    text = open('deadlock.c').read()
+    with open('deadlock.c') as f:
+        text = f.read()
     text = text.replace(b'MAX_THREADS', str(args.threads));
     text = text.replace(b'MAX_EDGES', str(args.edges));
     bpf = BPF(text=text)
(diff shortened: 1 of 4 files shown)


tools/deadlock.py
```diff
@@ -465,7 +475,10 @@ def main():
             print('%s. Is the process (pid=%d) running?' % (str(e), args.pid))
             sys.exit(1)
 
-    bpf = BPF(src_file=b'deadlock.c')
+    text = open('deadlock.c').read()
```
chenhengqi, line 478: It would be better to use something like `with open('deadlock', 'r') as f:` :)

chenyuezhou: adjusted.

chenhengqi (approved):

yonghong-song:
[buildbot, test this please]

PR #3458, opened 2021-05-28
libbpf-tools: add fsslower

chenhengqi:
fsslower is a multitool which show file system operations slower
than a threshold. Now it supports btrfs/ext4/nfs/xfs filesystems.
It behaves the same as its counterpart in BCC tools named
btrfsslower.py/ext4slower.py/nfsslower.py/xfsslower.py

```shell
$ sudo ./fsslower -t ext4 -m 1 -d 20
Tracing ext4 operations slower than 1 ms for 20 secs.
TIME     COMM             PID     T BYTES   OFF_KB   LAT(ms) FILENAME
13:29:01 journal-offline  2992396 F LL_MAX  0          23.99 system.journal
```

### DIFF
--- a/libbpf-tools/fsslower.bpf.c
+++ b/libbpf-tools/fsslower.bpf.c
@@ -1,0 +1,202 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+/* Copyright (c) 2021 Hengqi Chen */
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_tracing.h>
+#include "bits.bpf.h"
+#include "fsslower.h"
+
+#define MAX_ENTRIES	10240
+
+const volatile pid_t target_pid = 0;
+const volatile __u64 min_lat_ms = 0;
+
+struct data {
+	__u64 ts;
+	loff_t start;
+	loff_t end;
+	struct file *fp;
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_ENTRIES);
+	__type(key, __u32);
+	__type(value, struct data);
+} starts SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, sizeof(__u32));
+} events SEC(".maps");
+
+static int probe_entry(struct file *fp, loff_t start, loff_t end)
+{
+	__u64 pid_tgid = bpf_get_current_pid_tgid();
+	__u32 pid = pid_tgid >> 32;
+	__u32 tid = (__u32)pid_tgid;
+	struct data data = {};
+
+	if (!fp)
+		return 0;
+
+	if (target_pid && target_pid != pid)
+		return 0;
+
+	data.ts = bpf_ktime_get_ns();
+	data.fp = fp;
+	data.start = start;
+	data.end = end;
+	bpf_map_update_elem(&starts, &tid, &data, BPF_ANY);
+	return 0;
+}
+
+static int probe_exit(void *ctx, enum fs_file_op op, ssize_t size)
+{
+	__u64 pid_tgid = bpf_get_current_pid_tgid();
+	__u32 pid = pid_tgid >> 32;
+	__u32 tid = (__u32)pid_tgid;
+	__u64 end_ns, delta_us;
+	const __u8 *file_name;
+	struct data *datap;
+	struct event event = {};
+	struct dentry *dentry;
+	struct file *fp;
+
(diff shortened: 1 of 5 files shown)


libbpf-tools/fsslower.bpf.c
```diff
@@ -0,0 +1,202 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+/* Copyright (c) 2021 Hengqi Chen */
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_tracing.h>
+#include "bits.bpf.h"
+#include "fsslower.h"
+
+#define MAX_ENTRIES	10240
```
anakryiko, line 10: xfsslower is using 8096, any specific reason to bump this?

libbpf-tools/fsslower.bpf.c
```diff
@@ -0,0 +1,202 @@
...
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, sizeof(__u32));
+} events SEC(".maps");
+
+static int probe_entry(struct file *fp, loff_t start, loff_t end)
+{
+	__u64 pid_tgid = bpf_get_current_pid_tgid();
+	__u32 pid = pid_tgid >> 32;
+	__u32 tid = (__u32)pid_tgid;
+	struct data data = {};
```
anakryiko, line 40: no need to zero-initialize, you are initializing it completely below

libbpf-tools/fsslower.bpf.c
```diff
@@ -0,0 +1,202 @@
...
+	__u64 pid_tgid = bpf_get_current_pid_tgid();
+	__u32 pid = pid_tgid >> 32;
+	__u32 tid = (__u32)pid_tgid;
+	__u64 end_ns, delta_us;
+	const __u8 *file_name;
+	struct data *datap;
+	struct event event = {};
+	struct dentry *dentry;
+	struct file *fp;
+
```
anakryiko, line 67: check targ_pid to avoid unnecessary map lookup? just like xfsslower does it

libbpf-tools/fsslower.bpf.c
```diff
@@ -0,0 +1,202 @@
...
+	datap = bpf_map_lookup_elem(&starts, &tid);
+	if (!datap)
+		return 0;
+
+	bpf_map_delete_elem(&starts, &tid);
+
+	end_ns = bpf_ktime_get_ns();
+	delta_us = (end_ns - datap->ts) / 1000;
+	if (delta_us <= min_lat_ms * 1000)
```
anakryiko, line 76: instead of having min_lat_ms in milliseconds and do nanosecond to millisecond convertion, why not use nanosecond units for min_lat and do a simple and fast comparison?

libbpf-tools/fsslower.c
```diff
@@ -0,0 +1,463 @@
...
+		goto cleanup;
+	}
+
+	print_headers();
+
+	if (duration)
+		time_end = get_ktime_ns() + duration * NSEC_PER_SEC;
+
+	while (1) {
+		usleep(PERF_BUFFER_TIME_MS * 1000);
```
anakryiko, line 450: Why is this usleep() needed here? I see that xfsslower is doing the same, but I can't recall why it was added in the first place...

libbpf-tools/fsslower.c
```diff
@@ -0,0 +1,463 @@
...
+ *
+ * Copyright (c) 2021 Hengqi Chen
+ *
+ * Based on fsdist and previous work by
+ *     - Brendan Gregg
+ *     - Dina Goldshtein
+ *     - Joe Yin
+ *     - Samuel Nair
+ *     - Wenbo Zhang
+ * 27-May-2021   Hengqi Chen   Created this.
```
anakryiko, line 14: This is very much based on xfsslower, so I think it's fair to keep "Copyright (c) 2020 Wenbo Zhang" and "9-Mar-2020   Wenbo Zhang   Created this."?

anakryiko (changes requested):
This overall looks good, see few comments.

But I'd prefer if this PR was a refactoring of xfsslower into fsslower, instead of a completely seperate copy of the code. That way we can keep track of history and re-use big chunks of code.

Also, similarly what I asked to do for fsdist, let's add symlinks for individual ext4slower, btrfsslower, nfsslower and xfsslower from the very beginning.

Thanks!

chenhengqi:
> This overall looks good, see few comments.
> 
> But I'd prefer if this PR was a refactoring of xfsslower into fsslower, instead of a completely seperate copy of the code. That way we can keep track of history and re-use big chunks of code.
> 
> Also, similarly what I asked to do for fsdist, let's add symlinks for individual ext4slower, btrfsslower, nfsslower and xfsslower from the very beginning.
> 
> Thanks!

Will do. But how to add symlinks for those tools, since we only have source code here.

chenhengqi:
close in favor of #3472

PR #3460, opened 2021-05-30
libbpf-tools: add mountsnoop

chenhengqi:
This commit adds a new libbpf tool mountsnoop.
It has the same functionalities just as its
counterpart in BCC tools. The default output
is the same.

```
$ mountsnoop
COMM             PID     TID     MNT_NS      CALL
dockerd          1827    1903    4026531840  mount("overlay", "/data/docker/overlay2/153e6b58322c64cf4b2aac1b9caba42d390481a7d33a2bffe0eb858943d49fb6-init/merged", "overlay", 0x0, "index=off,lowerdir=/data/docker/overlay2/l/GWTHHZ2C3PYGAJ5GLTWLHMHHKR,upperdir=/data/docker/overlay2/153e6b58322c64cf4b2aac1b9caba42d390481a7d33a2bffe0eb858943d49fb6-init/diff,workdir=/data/docker/overlay2/153e6b58322c64cf4b2aac1b9caba42d390481a7d33a2bffe0eb858943d49fb6-init/work") = 0
dockerd          1827    1903    4026531840  umount("/data/docker/overlay2/153e6b58322c64cf4b2aac1b9caba42d390481a7d33a2bffe0eb858943d49fb6-init/merged", MS_NOSUID) = 0
```

Also, we provide a detailed mode enabled by -d
option which displays each mount/umount syscall
vertically with more field. In this way, the
output looks more friendly.

```
$ mountsnoop -d -t
PID:    1827
TID:    1864
COMM:   dockerd
OP:     MOUNT
RET:    0
LAT:    246us
MNT_NS: 4026531840
FS:     overlay
SOURCE: overlay
TARGET: /data/docker/overlay2/5fc51d4e4820082177751a8aadf3f42a751c86aff1e0efbc1a5e6af345ee205a-init/merged
DATA:   index=off,lowerdir=/data/docker/overlay2/l/GWTHHZ2C3PYGAJ5GLTWLHMHHKR,upperdir=/data/docker/overlay2/5fc51d4e4820082177751a8aadf3f42a751c86aff1e0efbc1a5e6af345ee205a-init/diff,workdir=/data/docker/overlay2/5fc51d4e4820082177751a8aadf3f42a751c86aff1e0efbc1a5e6af345ee205a-init/work
FLAGS:  0x0

PID:    1827
TID:    1864
COMM:   dockerd
OP:     UMOUNT
RET:    0
LAT:    95us
MNT_NS: 4026531840
FS:
SOURCE:
TARGET: /data/docker/overlay2/5fc51d4e4820082177751a8aadf3f42a751c86aff1e0efbc1a5e6af345ee205a-init/merged
DATA:
FLAGS:  MS_NOSUID
```

### DIFF
--- a/libbpf-tools/mountsnoop.bpf.c
+++ b/libbpf-tools/mountsnoop.bpf.c
@@ -1,0 +1,137 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+/* Copyright (c) 2021 Hengqi Chen */
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_tracing.h>
+#include "mountsnoop.h"
+
+#define MAX_ENTRIES 10240
+
+const volatile pid_t target_pid = 0;
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_ENTRIES);
+	__type(key, __u32);
+	__type(value, struct arg);
+} args SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__uint(max_entries, 1);
+	__type(key, int);
+	__type(value, struct event);
+} heap SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, sizeof(__u32));
+} events SEC(".maps");
+
+static int probe_entry(const char *src, const char *dest, const char *fs,
+		       __u64 flags, const char *data, enum op op)
+{
+	__u64 pid_tgid = bpf_get_current_pid_tgid();
+	__u32 pid = pid_tgid >> 32;
+	__u32 tid = (__u32)pid_tgid;
+	struct arg arg = {};
+
+	if (target_pid && target_pid != pid)
+		return 0;
+
+	arg.ts = bpf_ktime_get_ns();
+	arg.flags = flags;
+	arg.src = src;
+	arg.dest = dest;
+	arg.fs = fs;
+	arg.data= data;
+	arg.op = op;
+	bpf_map_update_elem(&args, &tid, &arg, BPF_ANY);
+	return 0;
+};
+
+static int probe_exit(void *ctx, int ret)
+{
+	__u64 pid_tgid = bpf_get_current_pid_tgid();
+	__u32 pid = pid_tgid >> 32;
+	__u32 tid = (__u32)pid_tgid;
+	struct arg *argp;
+	struct event *eventp;
+	struct task_struct *task;
+	int zero = 0;
+
+	argp = bpf_map_lookup_elem(&args, &tid);
+	if (!argp)
+		return 0;
+
+	eventp = bpf_map_lookup_elem(&heap, &zero);
+	if (!eventp)
+		return 0;
+
+	task = (struct task_struct *)bpf_get_current_task();
+	eventp->delta = bpf_ktime_get_ns() - argp->ts;
+	eventp->flags = argp->flags;
+	eventp->pid = pid;
+	eventp->tid = tid;
+	eventp->mnt_ns = BPF_CORE_READ(task, nsproxy, mnt_ns, ns.inum);
+	eventp->ret = ret;
+	eventp->op = argp->op;
+	bpf_get_current_comm(&eventp->comm, sizeof(eventp->comm));
+	if (argp->src)
+		bpf_probe_read_user_str(eventp->src, sizeof(eventp->src), argp->src);
+	else
+		eventp->src[0] = '\0';
+	if (argp->dest)
+		bpf_probe_read_user_str(eventp->dest, sizeof(eventp->dest), argp->dest);
+	else
+		eventp->dest[0] = '\0';
+	if (argp->fs)
+		bpf_probe_read_user_str(eventp->fs, sizeof(eventp->fs), argp->fs);
+	else
+		eventp->fs[0] = '\0';
+	if (argp->data)
+		bpf_probe_read_user_str(eventp->data, sizeof(eventp->data), argp->data);
+	else
+		eventp->data[0] = '\0';
+	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, eventp, sizeof(*eventp));
+
+	bpf_map_delete_elem(&args, &tid);
+	return 0;
+}
+
+SEC("tracepoint/syscalls/sys_enter_mount")
+int mount_entry(struct trace_event_raw_sys_enter *ctx)
+{
+	const char *src = (const char *)ctx->args[0];
+	const char *dest = (const char *)ctx->args[1];
+	const char *fs = (const char *)ctx->args[2];
+	__u64 flags = (__u64)ctx->args[3];
+	const char *data = (const char *)ctx->args[4];
+
+	return probe_entry(src, dest, fs, flags, data, MOUNT);
+}
+
+SEC("tracepoint/syscalls/sys_exit_mount")
+int mount_exit(struct trace_event_raw_sys_exit *ctx)
+{
+	return probe_exit(ctx, (int)ctx->ret);
+}
+
+SEC("tracepoint/syscalls/sys_enter_umount")
+int umount_entry(struct trace_event_raw_sys_enter *ctx)
+{
+	const char *dest = (const char *)ctx->args[0];
+	__u64 flags = (__u64)ctx->args[1];
+
+	return probe_entry(NULL, dest, NULL, flags, NULL, UMOUNT);
+}
+
+SEC("tracepoint/syscalls/sys_exit_umount")
+int umount_exit(struct trace_event_raw_sys_exit *ctx)
+{
+	return probe_exit(ctx, (int)ctx->ret);
+}
+
+char LICENSE[] SEC("license") = "Dual BSD/GPL";
--- a/libbpf-tools/mountsnoop.c
+++ b/libbpf-tools/mountsnoop.c
@@ -1,0 +1,308 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+
+/*
+ * mountsnoop  Trace mount and umount[2] syscalls
+ *
+ * Copyright (c) 2021 Hengqi Chen
+ * 30-May-2021   Hengqi Chen   Created this.
+ */
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+#include <argp.h>
+#include <errno.h>
+#include <signal.h>
+#include <string.h>
+#include <time.h>
+
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+#include "mountsnoop.h"
+#include "mountsnoop.skel.h"
+#include "trace_helpers.h"
+
+#define PERF_BUFFER_PAGES	64
+#define PERF_POLL_TIMEOUT_MS	100
+#define warn(...) fprintf(stderr, __VA_ARGS__)
+
+/* https://www.gnu.org/software/gnulib/manual/html_node/strerrorname_005fnp.html */
+#if !defined(__GLIBC__) || __GLIBC__ < 2 || (__GLIBC__ == 2 && __GLIBC_MINOR__ < 32)
+	const char *strerrorname_np(int errnum)
+	{
+		return NULL;
+	}
+#endif
+
+static volatile sig_atomic_t exiting = 0;
+
+static pid_t target_pid = 0;
+static bool emit_timestamp = false;
+static bool output_vertically = false;
+static const char *flag_names[] = {
+	[0] = "MS_RDONLY",
+	[1] = "MS_NOSUID",
+	[2] = "MS_NODEV",
+	[3] = "MS_NOEXEC",
+	[4] = "MS_SYNCHRONOUS",
+	[5] = "MS_REMOUNT",
+	[6] = "MS_MANDLOCK",
+	[7] = "MS_DIRSYNC",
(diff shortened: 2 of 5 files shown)


chenhengqi:
@anakryiko 

I need some help here. I tried implement `mountsnoop` using raw tracepoint but got the following error:
```shell
libbpf: prog 'mount_entry': failed to attach to raw tracepoint 'sys_enter_mount': No such file or directory
libbpf: failed to auto-attach program 'mount_entry': -2
failed to attach BPF programs: -2
```

In available events:
```shell
$ cat /sys/kernel/debug/tracing/available_events | grep sys_enter_mount
syscalls:sys_enter_mount
```

anakryiko:
raw_tracepoints don't support per-syscall tracepoints, they only have generic sys_enter and sys_exit tracepoints. So if you want to use raw tracepoint, you'd need to filter on syscall ID. Otherwise you have to use tracepoints.

chenhengqi:
> raw_tracepoints don't support per-syscall tracepoints, they only have generic sys_enter and sys_exit tracepoints. So if you want to use raw tracepoint, you'd need to filter on syscall ID. Otherwise you have to use tracepoints.

Thanks for the detail. Will switch to tracepoint instead.

chenhengqi:
Currently, the tool has the following output format, not sure it's acceptable or not
```shell
16:51:46
    PID:    574587
    COMM:   runc:[2:INIT]
    OP:     MOUNT
    RET:    0
    LAT:    4us
    FS:     
    SOURCE: 
    TARGET: /proc/self/fd/6
    DATA:   
    FLAGS:  MS_VERBOSE | MS_PRIVATE

16:51:46
    PID:    574587
    COMM:   runc:[2:INIT]
    OP:     MOUNT
    RET:    0
    LAT:    14us
    FS:     bind
    SOURCE: /data/docker/containers/a5053ad18fa4b074087166415c8d74b0f43c6c3bf9649a87b56b0950b910ae6e/hosts
    TARGET: /proc/self/fd/6
    DATA:   
    FLAGS:  MS_MOVE | MS_VERBOSE

16:52:22
    PID:    1039
    COMM:   systemd
    OP:     UMOUNT
    RET:    -2
    LAT:    25us
    FS:     
    SOURCE: 
    TARGET: /run/user/1000/credentials/data-docker-overlay2-ccb8d238b4e6753e1fa72096d41b683d94c602a3c8d13608d58256d0217f2a26-merged.mount
    DATA:   
    FLAGS:  MS_NOSUID | MS_NOEXEC

16:52:22
    PID:    1
    COMM:   systemd
    OP:     UMOUNT
    RET:    -2
    LAT:    23us
    FS:     
    SOURCE: 
    TARGET: /run/credentials/data-docker-overlay2-ccb8d238b4e6753e1fa72096d41b683d94c602a3c8d13608d58256d0217f2a26-merged.mount
    DATA:   
    FLAGS:  MS_NOSUID | MS_NOEXEC
```

anakryiko:
@chenhengqi seems like there is existing Python-based mountsnoop.py, right? We should try to preserve the output format of the original tool as much as possible. If there is extra data that you think is useful to output, but it doesn't fit into single-line format, then I think more appropriate approach is to have a detailed view mode, which user can specify to get more data. WDYT?

chenhengqi:
> @chenhengqi seems like there is existing Python-based mountsnoop.py, right? We should try to preserve the output format of the original tool as much as possible. If there is extra data that you think is useful to output, but it doesn't fit into single-line format, then I think more appropriate approach is to have a detailed view mode, which user can specify to get more data. WDYT?

Nice idea. Will make the default output the same as mountsnoop.py and some add some new options to alter the output format.

chenhengqi:
@anakryiko  Updated. Please take a look again. Thanks.

libbpf-tools/mountsnoop.bpf.c
```diff
@@ -0,0 +1,131 @@
...
+		bpf_probe_read_user_str(eventp->data, sizeof(eventp->data), argp->data);
+	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, eventp, sizeof(*eventp));
+
+	bpf_map_delete_elem(&args, &tid);
+	bpf_map_delete_elem(&values, &tid);
+	return 0;
+}
+
+SEC("tracepoint/syscalls/sys_enter_mount")
+int mount_entry(struct trace_event_raw_sys_enter *ctx)
```
anakryiko, line 99: not sure if you realized, but you could have written this as 

```
int BPF_PROG(mount_entry, const char *src, const char *dst, const char *fs, __u64 flags, const char data) {
    return probe_entry(src, dest, fs, flags, data, MOUNT);
}
```

libbpf-tools/mountsnoop.c
```diff
@@ -0,0 +1,318 @@
...
+"\n"
+"EXAMPLES:\n"
+"    mountsnoop         # trace mount and umount syscalls\n"
+"    mountsnoop -v      # output vertically(one line per column value)\n"
+"    mountsnoop -p 1216 # only trace PID 1216\n";
+
+static const struct argp_option opts[] = {
+	{ "pid", 'p', "PID", 0, "Process ID to trace" },
+	{ "timestamp", 't', NULL, 0, "Include timestamp on output" },
+	{ "vertical", 'v', NULL, 0, "Output result vertically" },
```
anakryiko, line 85: -v usually means "verbose". "Vertical" is quite unusual way to phrase something like this. Why not "--detailed"?

libbpf-tools/mountsnoop.c
```diff
@@ -0,0 +1,318 @@
...
+static const char *strflags(__u64 flags)
+{
+	static char str[512];
+	int i;
+	bool first = true;
+
+	if (!flags)
+		return "0x0";
+
+	memset(str, 0, sizeof(str));
```
anakryiko, line 133: `str[0] = '\0';` should be enough

libbpf-tools/mountsnoop.c
```diff
@@ -0,0 +1,318 @@
...
+	for (i = 0; i < flag_count; i++) {
+		if (!((1 << i) & flags))
+			continue;
+		if (first)
+			first = false;
+		else
+			strcat(str, " | ");
```
anakryiko, line 140:
```
if (str[0] != '\0')
    strcat(str, " | ");
```

libbpf-tools/mountsnoop.c
```diff
@@ -0,0 +1,318 @@
...
+
+static const char *strerrno(int errnum)
+{
+	const char *errstr;
+	static char ret[32] = {};
+
+	if (!errnum)
+		return "0";
+
+	memset(ret, 0, sizeof(ret));
```
anakryiko, line 154: same as above, making sure first char is 0 is enough to make it an empty string

libbpf-tools/mountsnoop.c
```diff
@@ -0,0 +1,318 @@
...
+		strcat(call, "\", \"");
+		strcat(call, e->dest);
+		strcat(call, "\", \"");
+		strcat(call, e->fs);
+		strcat(call, "\", ");
+		strcat(call, strflags(e->flags));
+		strcat(call, ", \"");
+		strcat(call, e->data);
+		strcat(call, "\") = ");
+		strcat(call, strerrno(e->ret));
```
anakryiko, line 190: uhm... sprintf()?..

anakryiko:
Can you please also expand the commit with example of output in both detailed and default modes? It's worth calling out that detailed view is new one and BCC-based version of mountsnoop doesn't support it.

anakryiko (changes requested):

libbpf-tools/mountsnoop.bpf.c
```diff
@@ -0,0 +1,131 @@
...
+
+	arg.ts = bpf_ktime_get_ns();
+	arg.flags = flags;
+	arg.src = src;
+	arg.dest = dest;
+	arg.fs = fs;
+	arg.data= data;
+	arg.op = op;
+	bpf_map_update_elem(&args, &tid, &arg, BPF_ANY);
+	bpf_map_update_elem(&values, &tid, &empty, BPF_ANY);
```
anakryiko, line 53: hm... I don't think you need two separate maps for this. Just store both args and event in one struct, but at the end submit only the event part. Makes everything simpler and more efficient.

anakryiko: hm... thinking some more, you don't even need hashmap for event/values.
1. Either use per-cpu array for temporary storage (you need event only during probe_exit execution, so using per-cpu is fine, because that BPF program is not going to be migrated to another CPU nor will it be interrupted by the same tracepoint).
2. Or use BPF ringbuf and just avoid extra map for struct event. Do bpf_ringbuf_reserve(), fill out data, submit. Done.

anakryiko (changes requested):

libbpf-tools/mountsnoop.bpf.c
chenhengqi: Thanks. Will look into this later.

`bpf_map_update_elem(&values, &tid, &empty, BPF_ANY);` this one is used to avoid stack limit of BPF programs.

libbpf-tools/mountsnoop.bpf.c
chenhengqi: Just like the way we use `fentry`, Right? 
Actually, I didn't realize this. Thanks for the reminder.
Maybe we can add an example to libbpf-bootstrap.

chenhengqi: `mount` has the following format, so the first argument has to be an int.
```
field:int __syscall_nr;	offset:8;	size:4;	signed:1;
field:char * dev_name;	offset:16;	size:8;	signed:0;
field:char * dir_name;	offset:24;	size:8;	signed:0;
field:char * type;	offset:32;	size:8;	signed:0;
field:unsigned long flags;	offset:40;	size:8;	signed:0;
field:void * data;	offset:48;	size:8;	signed:0;
```

chenhengqi: I will do this for the `kvmexit` tool.

libbpf-tools/mountsnoop.c
chenhengqi: :)

libbpf-tools/mountsnoop.bpf.c
anakryiko: that format is for non-raw tracepoint. BPF_PROG() is just rewriting those ctx[0], ctx[1] accesses into a nice input parameters. So don't overthink this.

libbpf-tools/mountsnoop.bpf.c
anakryiko: > `bpf_map_update_elem(&values, &tid, &empty, BPF_ANY);` this one is used to avoid stack limit of BPF programs.

Right, that's what I realized and that's why I propose two better alternative. BPF ringbuf is the best variant, but it sets higher minimum kernel requirements.

libbpf-tools/mountsnoop.bpf.c
chenhengqi: @anakryiko  I got this error when I tried this method.

```
libbpf: -- BEGIN DUMP LOG ---
libbpf: 
arg#0 type is not a struct
Unrecognized arg#0 type PTR
; int BPF_PROG(mount_entry, const char *src, const char *dest, const char *fs, __u64 flags, const char *data)
0: (79) r2 = *(u64 *)(r1 +32)
1: (7b) *(u64 *)(r10 -80) = r2
2: (79) r2 = *(u64 *)(r1 +24)
3: (7b) *(u64 *)(r10 -72) = r2
4: (79) r6 = *(u64 *)(r1 +16)
5: (79) r9 = *(u64 *)(r1 +8)
6: (79) r8 = *(u64 *)(r1 +0)
invalid bpf_context access off=0 size=8
processed 7 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0

libbpf: -- END LOG --
libbpf: failed to load program 'mount_entry'
libbpf: failed to load object 'mountsnoop_bpf'
libbpf: failed to load BPF skeleton 'mountsnoop_bpf': -4007
```

anakryiko: Oh, wait. I got confused, sorry. I somehow decided that you are using *raw tracepoint* (`SEC("raw_tp/sys_enter")`) but clearly you are not. BPF_PROG() approach I described above is applicable only to raw_tracepoints, so never mind.

libbpf-tools/mountsnoop.bpf.c
chenhengqi: I've read your blog about BPF ringbuf, that's awesome. Thanks.

Since the BPF ringbuf requires bump kernel version and we have a bunch of tools using perfbuf already, let's
use per-cpu array here. WDYT ?

libbpf-tools/mountsnoop.bpf.c
```diff
@@ -0,0 +1,129 @@
...
+	eventp->op = argp->op;
+	bpf_get_current_comm(&eventp->comm, sizeof(eventp->comm));
+	if (argp->src)
+		bpf_probe_read_user_str(eventp->src, sizeof(eventp->src), argp->src);
+	if (argp->dest)
+		bpf_probe_read_user_str(eventp->dest, sizeof(eventp->dest), argp->dest);
+	if (argp->fs)
+		bpf_probe_read_user_str(eventp->fs, sizeof(eventp->fs), argp->fs);
+	if (argp->data)
+		bpf_probe_read_user_str(eventp->data, sizeof(eventp->data), argp->data);
```
anakryiko, line 89: you are re-using event between multiple invocation. If previous event had src/dst/fs/data filled in and current event doesn't, you'll still have the data from the previous event, which is going to be very confusing. You should add `else { argp->src[0] = '\0'; }` to each of these to clear out the string.

chenhengqi: Ah, another gotcha, didn't realize that.

Updated, thanks.

anakryiko (approved):
LGTM

yonghong-song:
[buildbot, test this please]

PR #3461, opened 2021-05-31
libbpf-tools: add solisten

chenhengqi:
```shell
$ sudo solisten
PID     COMM             RET LAT(us)  BACKLOG PROTO PORT  ADDR                            
241011  netcat           0   12       1       TCPv4 1216  0.0.0.0                         
241158  nginx            0   14       511     TCPv4 80    0.0.0.0                         
241158  nginx            0   4        511     TCPv6 80    ::                              
241158  nginx            0   1        511     TCPv4 80    0.0.0.0                         
241158  nginx            0   1        511     TCPv6 80    ::
```

### DIFF
--- a/libbpf-tools/solisten.bpf.c
+++ b/libbpf-tools/solisten.bpf.c
@@ -1,0 +1,101 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+/* Copyright (c) 2021 Hengqi Chen */
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_endian.h>
+#include <bpf/bpf_tracing.h>
+#include "solisten.h"
+
+#define MAX_ENTRIES	10240
+#define AF_INET	2
+#define AF_INET6	10
+
+const volatile pid_t target_pid = 0;
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_ENTRIES);
+	__type(key, __u32);
+	__type(value, struct event);
+} values SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, sizeof(__u32));
+} events SEC(".maps");
+
+static void fill_event(struct event *event, struct socket *sock)
+{
+	__u16 family, type;
+	struct sock *sk;
+	struct inet_sock *inet;
+
+	sk = BPF_CORE_READ(sock, sk);
+	inet = (struct inet_sock *)sk;
+	family = BPF_CORE_READ(sk, __sk_common.skc_family);
+	type = BPF_CORE_READ(sock, type);
+
+	event->proto = ((__u32)family << 16) | type;
+	event->port = bpf_ntohs(BPF_CORE_READ(inet, inet_sport));
+	if (family == AF_INET)
+		event->addr[0] = BPF_CORE_READ(sk, __sk_common.skc_rcv_saddr);
+	else if (family == AF_INET6)
+		BPF_CORE_READ_INTO(event->addr, sk, __sk_common.skc_v6_rcv_saddr.in6_u.u6_addr32);
+	bpf_get_current_comm(event->task, sizeof(event->task));
+}
+
+SEC("kprobe/inet_listen")
+int BPF_KPROBE(inet_listen_entry, struct socket *sock, int backlog)
+{
+	__u64 pid_tgid = bpf_get_current_pid_tgid();
+	__u32 pid = pid_tgid >> 32;
+	__u32 tid = (__u32)pid_tgid;
+	struct event event = {};
+
+	if (target_pid && target_pid != pid)
+		return 0;
+
+	fill_event(&event, sock);
+	event.pid = pid;
+	event.backlog = backlog;
+	bpf_map_update_elem(&values, &tid, &event, BPF_ANY);
+	return 0;
+}
+
+SEC("kretprobe/inet_listen")
+int BPF_KRETPROBE(inet_listen_exit, int ret)
+{
+	__u32 tid = bpf_get_current_pid_tgid();
+	struct event *eventp;
+
+	eventp = bpf_map_lookup_elem(&values, &tid);
+	if (!eventp)
+		return 0;
+
+	eventp->ret = ret;
+	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, eventp, sizeof(*eventp));
+	bpf_map_delete_elem(&values, &tid);
+	return 0;
+}
+
+SEC("fexit/inet_listen")
+int BPF_PROG(inet_listen_fexit, struct socket *sock, int backlog, int ret)
+{
+	__u64 pid_tgid = bpf_get_current_pid_tgid();
+	__u32 pid = pid_tgid >> 32;
+	struct event event = {};
+
+	if (target_pid && target_pid != pid)
+		return 0;
+
+	fill_event(&event, sock);
+	event.pid = pid;
+	event.backlog = backlog;
+	event.ret = ret;
+	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, &event, sizeof(event));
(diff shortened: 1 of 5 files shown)


libbpf-tools/solisten.bpf.c
```diff
@@ -0,0 +1,88 @@
...
+/* Copyright (c) 2021 Hengqi Chen */
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_endian.h>
+#include <bpf/bpf_tracing.h>
+#include "solisten.h"
+
+#define MAX_ENTRIES	10240
+#define AF_INET		2
```
danobi, line 11: Weird indentation
```suggestion
#define AF_INET	2
```

libbpf-tools/solisten.c
```diff
@@ -0,0 +1,191 @@
...
+	if (type == SOCK_STREAM)
+		prot = "TCP";
+	else if (type == SOCK_DGRAM)
+		prot = "UDP";
+	else
+		prot = "UNK";
+	snprintf(proto, sizeof(proto), "%s%s", prot, family == AF_INET ? "v4" : "v6");
+	inet_ntop(family, e->addr, addr, sizeof(addr));
+	printf("%-7d %-16s %-3d %-8d %-7d %-5s %-5d %-32s\n",
+	       e->pid, e->task, e->ret, (int)e->time / 1000, e->backlog, proto, e->port, addr);
```
danobi, line 110: Why cast `e->time` to int? Why not just use `%lu` format specifier?

libbpf-tools/solisten.bpf.c
```diff
@@ -0,0 +1,88 @@
...
+	return 0;
+}
+
+SEC("kprobe/inet_listen")
+int BPF_KPROBE(inet_listen_entry, struct socket *sock, int backlog)
+{
+	return probe_entry(sock, backlog);
+}
+
+SEC("kretprobe/inet_listen")
```
danobi, line 82: Can these 2 probes be replaced with a single `fexit/inet_listen` probe? Would let us remove the `values` map.

libbpf-tools/solisten.bpf.c
```diff
@@ -0,0 +1,88 @@
...
+	__u32 pid = pid_tgid >> 32;
+	__u32 tid = (__u32)pid_tgid;
+	__u16 family, type;
+	struct sock *sk = BPF_CORE_READ(sock, sk);
+	struct inet_sock *inet = (struct inet_sock *)sk;
+	struct event event = {
+		.time = bpf_ktime_get_ns(),
+		.pid = pid,
+		.backlog = backlog,
+	};
```
anakryiko, line 41: all this work before we even decide if we do the tracing or not (target_pid check below). Let's not be wasteful here

libbpf-tools/solisten.bpf.c
```diff
@@ -0,0 +1,88 @@
...
+		return 0;
+
+	family = BPF_CORE_READ(sk, __sk_common.skc_family);
+	type = BPF_CORE_READ(sock, type);
+	event.proto = ((__u32)family << 16) | type;
+	event.port = bpf_ntohs(BPF_CORE_READ(inet, inet_sport));
+	if (family == AF_INET)
+		event.addr[0] = BPF_CORE_READ(sk, __sk_common.skc_rcv_saddr);
+	else if (family == AF_INET6)
+		bpf_probe_read_kernel(event.addr, sizeof(event.addr), sk->__sk_common.skc_v6_rcv_saddr.in6_u.u6_addr32);
```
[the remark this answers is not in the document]
anakryiko: you could also do BPF_CORE_READ_INTO() here for consistency with BPF_CORE_READ() usage above

libbpf-tools/solisten.bpf.c
```diff
@@ -0,0 +1,88 @@
+/* SPDX-License-Identifier: GPL-2.0 */
```
anakryiko, line 1: LGPL-2.1 OR BSD-2-Clause to match LICENSE variable below?

libbpf-tools/solisten.c
```diff
@@ -0,0 +1,191 @@
...
+		return 1;
+	}
+
+	obj = solisten_bpf__open();
+	if (!obj) {
+		warn("failed to open BPF object\n");
+		return 1;
+	}
+
+	err = solisten_bpf__load(obj);
```
anakryiko, line 146: you can combine open and load into a single solisten_bpf__open_and_load(), given you are not doing any adjustments between those two steps

chenhengqi: I missed the `pid` filtering part. Added it.

libbpf-tools/solisten.bpf.c
chenhengqi: Nice idea. Add `fexit` and fallback to `kprobe/kretprobe` if not available.

libbpf-tools/solisten.bpf.c
```diff
@@ -0,0 +1,111 @@
...
+SEC("kprobe/inet_listen")
+int BPF_KPROBE(inet_listen_entry, struct socket *sock, int backlog)
+{
+	return probe_entry(sock, backlog);
+}
+
+SEC("kretprobe/inet_listen")
+int BPF_KRETPROBE(inet_listen_exit, int ret)
+{
+	return probe_exit(ctx, ret);
```
anakryiko, line 90: given probe_entry/probe_exit are not re-used in fexit program, I'd just inline them to have less need to jump around the source code to follow the logic

libbpf-tools/solisten.c
```diff
@@ -0,0 +1,203 @@
...
+
+	obj = solisten_bpf__open();
+	if (!obj) {
+		warn("failed to open BPF object\n");
+		return 1;
+	}
+
+	obj->rodata->target_pid = target_pid;
+
+	if (vmlinux_btf_exists() && fentry_exists("inet_listen", NULL)) {
```
anakryiko, line 151: doesn't fentry_exist() return false if vmlinux BTF can't be found?

libbpf-tools/solisten.c
```diff
@@ -0,0 +1,203 @@
...
+	err = solisten_bpf__attach(obj);
+	if (err) {
+		warn("failed to attach BPF programs: %d\n", err);
+		goto cleanup;
+	}
+
+	pb_opts.sample_cb = handle_event;
+	pb_opts.lost_cb = handle_lost_events;
+	pb = perf_buffer__new(bpf_map__fd(obj->maps.events), PERF_BUFFER_PAGES,
+			      &pb_opts);
```
anakryiko, line 173: nit: single line should be fine here

anakryiko (approved):
few minor nits, but otherwise looks good to me

danobi:
Is there a concrete use case for the kprobe fallback? solisten already exists as a bcc-based tool; why not target the latest and greatest linux for a new tool?

The way I've seen fallback code like this go is it never, ever gets removed b/c of concerns over breaking existing users. But for new tools there are no existing users.

chenhengqi:
> Is there a concrete use case for the kprobe fallback? solisten already exists as a bcc-based tool; why not target the latest and greatest linux for a new tool?
> 
> The way I've seen fallback code like this go is it never, ever gets removed b/c of concerns over breaking existing users. And then you're forever maintaining 2 of everything. But for new tools there are no existing users.

I saw first discussion here #3214  , and followed this way in #3441 and #3472 .

anakryiko:
> Is there a concrete use case for the kprobe fallback? solisten already exists as a bcc-based tool; why not target the latest and greatest linux for a new tool?
> 

If we are trying to build a practical light-weight replacements for Python-based tools, we shouldn't just target the latest and greatest kernels, but rather try to make it work on as many version as possible without too much maintenance burden. You are right that we probably won't be able to remove kprobes any time soon, but overall I don't think it adds much of a problem.

anakryiko:
Keep in mind that some embedded system might package and deploy libbpf-based variants, but never Python/BCC-based ones. Artificially crippling libbpf-based tools by chasing only the very latest features isn't great for end users.

This is one of the main reasons why I haven't tried forcing more of BPF ringbuf usage, for example.

chenhengqi:
@danobi  Please take a look at this again.

danobi (approved):

yonghong-song:
[buildbot, test this please]

PR #3472, opened 2021-06-05
libbpf-tools: migrate xfsslower to fsslower

chenhengqi:
This commit migrates xfsslower to a generic fsslower which supports
tracing multiple file systems. It works the same way as the original
tool except that the users are supposed to specify which file systems
to trace using -t option.

```shell
sudo ./fsslower -t ext4 -m 1
Tracing ext4 operations slower than 1 ms... Hit Ctrl-C to end.
TIME     COMM             PID     T BYTES   OFF_KB   LAT(ms) FILENAME
10:36:07 code             6896    F LL_MAX  0           2.40 state.vscdb-journal
10:36:07 code             6896    F LL_MAX  0           1.74 state.vscdb-journal
10:36:07 code             6896    F LL_MAX  0           1.78 state.vscdb
```


libbpf-tools/fsslower.c
```diff
@@ -0,0 +1,458 @@
...
+{
+	int i;
+	const char *fn_name, *module;
+	bool support_fentry = true;
+
+	for (i = 0; i < MAX_OP; i++) {
+		fn_name = fs_configs[fs_type].op_funcs[i];
+		module = fs_configs[fs_type].fs;
+		if (fn_name && !fentry_exists(fn_name, NULL)
+		    && !fentry_exists(fn_name, module)) {
```
anakryiko, line 206: hm... this made me realize that fentry_exists behavior is suboptimal. I think it should treat module as a hint, not as a mandatory thing. I.e., if we do fentry_exists("my_func", "my_module") and there is no /sys/kernel/btf/my_module, then just ignore "my_module" and try to find the type in the vmlinux. So that you can check existence of a function with just one fentry_exists() call. If functionality is in a module and that module exists -- good, we know which module to look at, but if it was compiled in or module BTFs are not supported, still look into vmlinux BTF.

In either case, such behavior will return true only if we did find the function in BTF somewhere. Can you please fix fentry_exists() to behave like that?

libbpf-tools/Makefile
```diff
@@ -46,7 +47,6 @@ APPS = \
 	tcpconnect \
 	tcpconnlat \
 	vfsstat \
-	xfsslower \
```
anakryiko, line 49: We still need to generate xfsslower. I think best way is to add something like

```
xfsslower ext4slower nfsslower btrfsslower: fsslower
    $(call msg,SYMLINK,$@)
    $(Q)ln -s $^ $@
```

Just make sure to also clean those up (probably would make sense to have `$(APP_ALIASES) = btrfsslower ext4slower ...` for that purpose.

anakryiko (commented):
Next time, let's not unnecessarily close/open PRs. It's better to preserve the history within the same PR, IMO.

chenhengqi:
> Next time, let's not unnecessarily close/open PRs. It's better to preserve the history within the same PR, IMO.

Sorry. Will follow this.

libbpf-tools/Makefile
chenhengqi: Updated.

libbpf-tools/fsslower.c
chenhengqi: Done.

libbpf-tools/trace_helpers.c
```diff
@@ -1043,7 +1023,38 @@ bool fentry_exists(const char *name, const char *mod)
...
+		snprintf(sysfs_mod, sizeof(sysfs_mod), "/sys/kernel/btf/%s", mod);
+		btf = btf__parse_split(sysfs_mod, base);
+		if (libbpf_get_error(btf)) {
+			fprintf(stderr, "failed to load BTF from %s: %s\n",
+				sysfs_mod, strerror(-libbpf_get_error(btf)));
+			goto out;
+		}
+		id = btf_id_by_name(btf, name);
+		if (id > 0)
+			goto out;
```
anakryiko, line 1052: it's unnecessary to do btf_id_by_name() twice. If module BTF was successfully loaded, then iterating it from type #1 to the last one will also iterate vmlinux BTFs (module BTF is "built on top" of vmlinux BTF and includes all the vmlinux BTF types as well). So you can simplify this and you won't need btf_id_by_name() function, you can leave the logic as is. Basically, the idea can be experessed in pseudo code as:

```
base = btf__parse();
if (error) /* exit, no BTF at all */
btf = btf__parse_split(sysfs_mod, base);
if (error)
    btf = base; /* no module, but fallback to plain vmlinux BTF */
/* search for type in btf just once */
```

libbpf-tools/fsslower.c
anakryiko: if it's done, you should remove `fentry_exists(fn_name, NULL)`

libbpf-tools/fsslower.c
```diff
@@ -0,0 +1,458 @@
...
+		printf("%lld,%lld,%s\n", e->offset, e->delta_us, e->file);
+		return;
+	}
+
+	time(&t);
+	tm = localtime(&t);
+	strftime(ts, sizeof(ts), "%H:%M:%S", tm);
+
+	printf("%-8s %-16s %-7d %c ", ts, e->task, e->pid, file_op[e->op]);
+	e->size == LLONG_MAX ? printf("%-7s ", "LL_MAX") : printf("%-7ld ", e->size);
```
anakryiko, line 344: this is an abuse of ternary operator, please use proper if/else here

libbpf-tools/trace_helpers.c
chenhengqi: Done.

libbpf-tools/fsslower.c
anakryiko: ping?

chenhengqi: Sorry, I missed this. Use if/else instead.

chenhengqi: I wonder any rationale about this ?

Reduce readability ?

anakryiko: ternary  operator is supposed to be used for returning expressions, it's not really a replacement for if/else statement; it's just confusing to read and easy to miss

libbpf-tools/fsslower.c
```diff
@@ -0,0 +1,460 @@
...
+static void handle_event(void *ctx, int cpu, void *data, __u32 data_sz)
+{
+	const struct event *e = data;
+	struct tm *tm;
+	char ts[32];
+	time_t t;
+
+	if (csv) {
+		printf("%lld,%s,%d,%c,", e->end_ns, e->task, e->pid, file_op[e->op]);
+		e->size == LLONG_MAX ? printf("LL_MAX,") : printf("%ld,", e->size);
```
anakryiko, line 333: same? can you please fix all the code?

libbpf-tools/fsslower.c
chenhengqi: OK. Thanks.

libbpf-tools/fsslower.c
chenhengqi: Fix two of them. Those in `fentry_set_attach_target` are left as it is.

anakryiko (approved):
LGTM, thanks!

yonghong-song:
[buildbot, test this please]

anakryiko:
@chenhengqi symboling linking breaks the (re-)build:

```
$ rm fsslower
06/10 09:35:13.823 andriin@devbig012:~/local/bcc/libbpf-tools (master)
$ make -j60
  BINARY   fsslower
  SYMLINK  btrfsslower
  SYMLINK  ext4slower
  SYMLINK  nfsslower
  SYMLINK  xfsslower
ln: failed to create symbolic link ‘btrfsslower’: File exists
ln: failed to create symbolic link ‘ext4slower’make: *** [btrfsslower] Error 1
make: : File exists*** Deleting file `btrfsslower'
```

We need to do `ln -f -s` to force over-writing existing binary or symlink. Cab you please follow up with the fix?

chenhengqi:
> @chenhengqi symboling linking breaks the (re-)build:
> 
> ```
> $ rm fsslower
> 06/10 09:35:13.823 andriin@devbig012:~/local/bcc/libbpf-tools (master)
> $ make -j60
>   BINARY   fsslower
>   SYMLINK  btrfsslower
>   SYMLINK  ext4slower
>   SYMLINK  nfsslower
>   SYMLINK  xfsslower
> ln: failed to create symbolic link ‘btrfsslower’: File exists
> ln: failed to create symbolic link ‘ext4slower’make: *** [btrfsslower] Error 1
> make: : File exists*** Deleting file `btrfsslower'
> ```
> 
> We need to do `ln -f -s` to force over-writing existing binary or symlink. Cab you please follow up with the fix?

OK. Will do this together with `fsdist`.

PR #3473, opened 2021-06-07
libbcc: support BPF_SOCKHASH specify the key type

chenyuezhou:
I think one of the main reasons for providing `sockhash` is because people want to use more flexible keys. So I think we need to provide a way for users to define their own key types.


yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks! Agree that for sockhash, we do want to flexible key type compared to u32 only (which is for sockarray).

PR #3477, opened 2021-06-08
Add kvmexit.py and kvmexit_example.txt

ShirleyFei:
Display the exit_reason and its statistics of each kvm_exit for all
vcpus of all virtual machines.

Signed-off-by: Fei Li <lifei.shirley@bytedance.com>

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -1123,6 +1123,18 @@ class BPF(object):
             return True
         return False
 
+    @staticmethod
+    def support_raw_tracepoint_in_module():
+        # kernel symbol "bpf_trace_modules" indicates raw tp support in modules, ref: kernel commit a38d1107
+        kallsyms = "/proc/kallsyms"
+        with open(kallsyms) as syms:
+            for line in syms:
+                (_, _, name) = line.rstrip().split(" ", 2)
+                name = name.split("\t")[0]
+                if name == "bpf_trace_modules":
+                    return True
+            return False
+
     def detach_tracepoint(self, tp=b""):
         """detach_tracepoint(tp="")
 
--- a/tools/kvmexit.py
+++ b/tools/kvmexit.py
@@ -1,0 +1,389 @@
+#!/usr/bin/env python
+#
+# kvmexit.py
+#
+# Display the exit_reason and its statistics of each vm exit
+# for all vcpus of all virtual machines. For example:
+# $./kvmexit.py
+#  PID      TID      KVM_EXIT_REASON                     COUNT
+#  1273551  1273568  EXIT_REASON_MSR_WRITE               6
+#  1274253  1274261  EXIT_REASON_EXTERNAL_INTERRUPT      1
+#  1274253  1274261  EXIT_REASON_HLT                     12
+#  ...
+#
+# Besides, we also allow users to specify one pid, tid(s), or one
+# pid and its vcpu. See kvmexit_example.txt for more examples.
+#
+# @PID: each vitual machine's pid in the user space.
+# @TID: the user space's thread of each vcpu of that virtual machine.
+# @KVM_EXIT_REASON: the reason why the vm exits.
+# @COUNT: the counts of the @KVM_EXIT_REASONS.
+#
+# REQUIRES: Linux 4.7+ (BPF_PROG_TYPE_TRACEPOINT support)
+#
+# Copyright (c) 2021 ByteDance Inc. All rights reserved.
+#
+# Author(s):
+#   Fei Li <lifei.shirley@bytedance.com>
+
+
+from __future__ import print_function
+from time import sleep, strftime
+from bcc import BPF
+import argparse
+import multiprocessing
+import os
+import signal
+import subprocess
+
+#
+# Process Arguments
+#
+def valid_args_list(args):
+    args_list = args.split(",")
+    for arg in args_list:
+        try:
+            int(arg)
+        except:
+            raise argparse.ArgumentTypeError("must be valid integer")
+    return args_list
+
+# arguments
+examples = """examples:
+    ./kvmexit                              # Display kvm_exit_reason and its statistics in real-time until Ctrl-C
+    ./kvmexit 5                            # Display in real-time after sleeping 5s
+    ./kvmexit -p 3195281                   # Collpase all tids for pid 3195281 with exit reasons sorted in descending order
+    ./kvmexit -p 3195281 20                # Collpase all tids for pid 3195281 with exit reasons sorted in descending order, and display after sleeping 20s
+    ./kvmexit -p 3195281 -v 0              # Display only vcpu0 for pid 3195281, descending sort by default
+    ./kvmexit -p 3195281 -a                # Display all tids for pid 3195281
+    ./kvmexit -t 395490                    # Display only for tid 395490 with exit reasons sorted in descending order
+    ./kvmexit -t 395490 20                 # Display only for tid 395490 with exit reasons sorted in descending order after sleeping 20s
+    ./kvmexit -T '395490,395491'           # Display for a union like {395490, 395491}
+"""
+parser = argparse.ArgumentParser(
+    description="Display kvm_exit_reason and its statistics at a timed interval",
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    epilog=examples)
+parser.add_argument("duration", nargs="?", default=99999999, type=int, help="show delta for next several seconds")
+parser.add_argument("-p", "--pid", type=int, help="trace this PID only")
+exgroup = parser.add_mutually_exclusive_group()
+exgroup.add_argument("-t", "--tid", type=int, help="trace this TID only")
+exgroup.add_argument("-T", "--tids", type=valid_args_list, help="trace a comma separated series of tids with no space in between")
+exgroup.add_argument("-v", "--vcpu", type=int, help="trace this vcpu only")
+exgroup.add_argument("-a", "--alltids", action="store_true", help="trace all tids for this pid")
+args = parser.parse_args()
+duration = int(args.duration)
+
+#
+# Setup BPF
+#
+
+# load BPF program
+bpf_text = """
+#include <linux/delay.h>
+
+#define REASON_NUM 69
+#define TGID_NUM 1024
+
+struct exit_count {
+    u64 exit_ct[REASON_NUM];
+};
+BPF_PERCPU_ARRAY(init_value, struct exit_count, 1);
+BPF_TABLE("percpu_hash", u64, struct exit_count, pcpu_kvm_stat, TGID_NUM);
+
+struct cache_info {
+    u64 cache_pid_tgid;
+    struct exit_count cache_exit_ct;
+};
+BPF_PERCPU_ARRAY(pcpu_cache, struct cache_info, 1);
+
+FUNC_ENTRY {
+    int cache_miss = 0;
+    int zero = 0;
+    u32 er = GET_ER;
+    if (er >= REASON_NUM) {
+        return 0;
+    }
+
+    u64 cur_pid_tgid = bpf_get_current_pid_tgid();
+    u32 tgid = cur_pid_tgid >> 32;
+    u32 pid = cur_pid_tgid;
+
+    if (THREAD_FILTER)
+        return 0;
+
+    struct exit_count *tmp_info = NULL, *initial = NULL;
+    struct cache_info *cache_p;
+    cache_p = pcpu_cache.lookup(&zero);
+    if (cache_p == NULL) {
+        return 0;
+    }
+
+    if (cache_p->cache_pid_tgid == cur_pid_tgid) {
+        //a. If the cur_pid_tgid hit this physical cpu consecutively, save it to pcpu_cache
+        tmp_info = &cache_p->cache_exit_ct;
+    } else {
+        //b. If another pid_tgid matches this pcpu for the last hit, OR it is the first time to hit this physical cpu.
+        cache_miss = 1;
+
+        // b.a Try to load the last cache struct if exists.
+        tmp_info = pcpu_kvm_stat.lookup(&cur_pid_tgid);
+
+        // b.b If it is the first time for the cur_pid_tgid to hit this pcpu, employ a
+        // per_cpu array to initialize pcpu_kvm_stat's exit_count with each exit reason's count is zero
+        if (tmp_info == NULL) {
+            initial = init_value.lookup(&zero);
+            if (initial == NULL) {
+                return 0;
+            }
+
+            pcpu_kvm_stat.update(&cur_pid_tgid, initial);
+            tmp_info = pcpu_kvm_stat.lookup(&cur_pid_tgid);
+            // To pass the verifier
+            if (tmp_info == NULL) {
+                return 0;
+            }
+        }
+    }
+
+    if (er < REASON_NUM) {
+        tmp_info->exit_ct[er]++;
+        if (cache_miss == 1) {
+            if (cache_p->cache_pid_tgid != 0) {
+                // b.*.a Let's save the last hit cache_info into kvm_stat.
+                pcpu_kvm_stat.update(&cache_p->cache_pid_tgid, &cache_p->cache_exit_ct);
+            }
+            // b.* As the cur_pid_tgid meets current pcpu_cache_array for the first time, save it.
+            cache_p->cache_pid_tgid = cur_pid_tgid;
+            bpf_probe_read(&cache_p->cache_exit_ct, sizeof(*tmp_info), tmp_info);
+        }
+        return 0;
+    }
+
+    return 0;
+}
+"""
+
+# format output
+exit_reasons = (
+    "EXCEPTION_NMI",
+    "EXTERNAL_INTERRUPT",
+    "TRIPLE_FAULT",
+    "INIT_SIGNAL",
+    "N/A",
+    "N/A",
+    "N/A",
+    "INTERRUPT_WINDOW",
+    "NMI_WINDOW",
+    "TASK_SWITCH",
+    "CPUID",
+    "N/A",
+    "HLT",
+    "INVD",
+    "INVLPG",
+    "RDPMC",
+    "RDTSC",
+    "N/A",
+    "VMCALL",
+    "VMCLEAR",
+    "VMLAUNCH",
+    "VMPTRLD",
+    "VMPTRST",
+    "VMREAD",
+    "VMRESUME",
+    "VMWRITE",
+    "VMOFF",
+    "VMON",
+    "CR_ACCESS",
+    "DR_ACCESS",
+    "IO_INSTRUCTION",
+    "MSR_READ",
+    "MSR_WRITE",
+    "INVALID_STATE",
+    "MSR_LOAD_FAIL",
+    "N/A",
+    "MWAIT_INSTRUCTION",
+    "MONITOR_TRAP_FLAG",
+    "N/A",
+    "MONITOR_INSTRUCTION",
+    "PAUSE_INSTRUCTION",
+    "MCE_DURING_VMENTRY",
+    "N/A",
+    "TPR_BELOW_THRESHOLD",
+    "APIC_ACCESS",
+    "EOI_INDUCED",
+    "GDTR_IDTR",
+    "LDTR_TR",
+    "EPT_VIOLATION",
+    "EPT_MISCONFIG",
+    "INVEPT",
+    "RDTSCP",
+    "PREEMPTION_TIMER",
+    "INVVPID",
+    "WBINVD",
+    "XSETBV",
+    "APIC_WRITE",
+    "RDRAND",
+    "INVPCID",
+    "VMFUNC",
+    "ENCLS",
+    "RDSEED",
+    "PML_FULL",
+    "XSAVES",
+    "XRSTORS",
+    "N/A",
+    "N/A",
+    "UMWAIT",
+    "TPAUSE"
+)
+
+#
+# Do some checks
+#
+try:
+    # Currently, only adapte on intel architecture
+    cmd = "cat /proc/cpuinfo | grep vendor_id | head -n 1"
+    arch_info = subprocess.check_output(cmd, shell=True).strip()
+    if b"Intel" in arch_info:
+        pass
+    else:
+        raise Exception("Currently we only support Intel architecture, please do expansion if needs more.")
+
+    # Check if kvm module is loaded
+    if os.access("/dev/kvm", os.R_OK | os.W_OK):
+        pass
+    else:
+        raise Exception("Please insmod kvm module to use kvmexit tool.")
+except Exception as e:
+    raise Exception("Failed to do precondition check, due to: %s." % e)
+
+try:
+    if BPF.support_raw_tracepoint_in_module():
+        # Let's firstly try raw_tracepoint_in_module
+        func_entry = "RAW_TRACEPOINT_PROBE(kvm_exit)"
+        get_er = "ctx->args[0]"
+    else:
+        # If raw_tp_in_module is not supported, fall back to regular tp
+        func_entry = "TRACEPOINT_PROBE(kvm, kvm_exit)"
+        get_er = "args->exit_reason"
+except Exception as e:
+    raise Exception("Failed to catch kvm exit reasons due to: %s" % e)
+
+
+def find_tid(tgt_dir, tgt_vcpu):
+    for tid in os.listdir(tgt_dir):
+        path = tgt_dir + "/" + tid + "/comm"
+        fp = open(path, "r")
+        comm = fp.read()
+        if (comm.find(tgt_vcpu) != -1):
+            return tid
+    return -1
+
+# set process/thread filter
+thread_context = ""
+header_format = ""
+need_collapse = not args.alltids
+if args.tid is not None:
+    thread_context = "TID %s" % args.tid
+    thread_filter = 'pid != %s' % args.tid
+elif args.tids is not None:
+    thread_context = "TIDS %s" % args.tids
+    thread_filter = "pid != " + " && pid != ".join(args.tids)
+    header_format = "TIDS     "
+elif args.pid is not None:
+    thread_context = "PID %s" % args.pid
+    thread_filter = 'tgid != %s' % args.pid
+    if args.vcpu is not None:
+        thread_context = "PID %s VCPU %s" % (args.pid, args.vcpu)
+        # transfer vcpu to tid
+        tgt_dir = '/proc/' + str(args.pid) + '/task'
+        tgt_vcpu = "CPU " + str(args.vcpu)
+        args.tid = find_tid(tgt_dir, tgt_vcpu)
+        if args.tid == -1:
+            raise Exception("There's no v%s for PID %d." % (tgt_vcpu, args.pid))
+        thread_filter = 'pid != %s' % args.tid
+    elif args.alltids:
+        thread_context = "PID %s and its all threads" % args.pid
+        header_format = "TID      "
+else:
+    thread_context = "all threads"
+    thread_filter = '0'
+    header_format = "PID      TID      "
+bpf_text = bpf_text.replace('THREAD_FILTER', thread_filter)
+
+# For kernel >= 5.0, use RAW_TRACEPOINT_MODULE for performance consideration
+bpf_text = bpf_text.replace('FUNC_ENTRY', func_entry)
+bpf_text = bpf_text.replace('GET_ER', get_er)
+b = BPF(text=bpf_text)
+
+
+# header
+print("Display kvm exit reasons and statistics for %s" % thread_context, end="")
+if duration < 99999999:
+    print(" after sleeping %d secs." % duration)
+else:
+    print("... Hit Ctrl-C to end.")
+print("%s%-35s %s" % (header_format, "KVM_EXIT_REASON", "COUNT"))
+
+# signal handler
+def signal_ignore(signal, frame):
+        print()
+try:
+    sleep(duration)
+except KeyboardInterrupt:
+    signal.signal(signal.SIGINT, signal_ignore)
+
+
+# Currently, sort multiple tids in descending order is not supported.
+if (args.pid or args.tid):
+    ct_reason = []
+    if args.pid:
+        tgid_exit = [0 for i in range(len(exit_reasons))]
+
+# output
+pcpu_kvm_stat = b["pcpu_kvm_stat"]
+pcpu_cache = b["pcpu_cache"]
+for k, v in pcpu_kvm_stat.items():
+    tgid = k.value >> 32
+    pid = k.value & 0xffffffff
+    for i in range(0, len(exit_reasons)):
+        sum1 = 0
+        for inner_cpu in range(0, multiprocessing.cpu_count()):
+            cachePIDTGID = pcpu_cache[0][inner_cpu].cache_pid_tgid
+            # Take priority to check if it is in cache
+            if cachePIDTGID == k.value:
+                sum1 += pcpu_cache[0][inner_cpu].cache_exit_ct.exit_ct[i]
+            # If not in cache, find from kvm_stat
+            else:
+                sum1 += v[inner_cpu].exit_ct[i]
+        if sum1 == 0:
+            continue
+
+        if (args.pid and args.pid == tgid and need_collapse):
+            tgid_exit[i] += sum1
+        elif (args.tid and args.tid == pid):
+            ct_reason.append((sum1, i))
+        elif not need_collapse or args.tids:
+            print("%-8u %-35s %-8u" % (pid, exit_reasons[i], sum1))
+        else:
+            print("%-8u %-8u %-35s %-8u" % (tgid, pid, exit_reasons[i], sum1))
+
+    # Display only for the target tid in descending sort
+    if (args.tid and args.tid == pid):
+        ct_reason.sort(reverse=True)
+        for i in range(0, len(ct_reason)):
+            if ct_reason[i][0] == 0:
+                continue
+            print("%-35s %-8u" % (exit_reasons[ct_reason[i][1]], ct_reason[i][0]))
+        break
+
+
+# Aggregate all tids' counts for this args.pid in descending sort
+if args.pid and need_collapse:
+    for i in range(0, len(exit_reasons)):
+        ct_reason.append((tgid_exit[i], i))
+    ct_reason.sort(reverse=True)
+    for i in range(0, len(ct_reason)):
+        if ct_reason[i][0] == 0:
+            continue
+        print("%-35s %-8u" % (exit_reasons[ct_reason[i][1]], ct_reason[i][0]))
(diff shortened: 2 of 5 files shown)


tools/kvmexit.py
```diff
@@ -0,0 +1,293 @@
...
+
+    if ival < 0:
+        raise argparse.ArgumentTypeError("must be positive")
+    return ival
+
+def positive_nonzero_int(val):
+    ival = positive_int(val)
+    if ival == 0:
+        raise argparse.ArgumentTypeError("must be nonzero")
+    return ival
```
chenhengqi, line 55: Existing tools in BCC simply use `int`, you can perform this check after `parser.parse_args`.

tools/kvmexit.py
```diff
@@ -0,0 +1,293 @@
...
+        print()
+try:
+    sleep(duration)
+except KeyboardInterrupt:
+    signal.signal(signal.SIGINT, signal_ignore)
+
+# output
+pcpu_kvm_stat = b["pcpu_kvm_stat"]
+pcpu_cache = b["pcpu_cache"]
+EXIT_REASON_NUM = 69
```
chenhengqi, line 277: Why hard code this, use `len(exit_reasons)` instead.

tools/kvmexit.py
```diff
@@ -0,0 +1,293 @@
...
+    } else {
+        return -1;
+    }
+
+    if (tmp_info == NULL) {
+        // b.a Try to load the last cache struct if existed.
+        tmp_info = pcpu_kvm_stat.lookup(&cur_pid_tgid);
+    }
+    if (tmp_info == NULL) {
+        // b.b If it is the first time for the cur_pid_tgid to hit this pcpu, employ a per_cpu array to initialize pcpu_kvm_stat's exit_count with each exit reason's count is zero
```
chenhengqi, line 125: Why not merge these two `if` statements ?

chenhengqi (changes requested):

tools/kvmexit.py
```diff
@@ -0,0 +1,293 @@
+#!/usr/bin/env python
+#
+# kvmexit.py
+#
+# Display the exit_reason and its statistics of each kvm_exit
+# for all vcpus of all virtual machines. For example,
+# $./kvmexit.py
+#  TGID     PID      KVM_EXIT_REASON                     STAT
```
yonghong-song, line 8: Instead of STAT, let us use COUNT, which states what the data is more accurately.

tools/kvmexit.py
```diff
@@ -0,0 +1,293 @@
...
+#  1273551  1273568  EXIT_REASON_MSR_WRITE               6
+#  1274253  1274261  EXIT_REASON_EXTERNAL_INTERRUPT      1
+#  1274253  1274261  EXIT_REASON_HLT                     12
+#  ...
+#
+# @TGID: each vitual machine's pid in the user space.
+# @PID: the user space's thread of each vcpu of that virtual machine.
+# @KVM_EXIT_REASON: the reason why the vm exits.
+# @STAT: the counts of the @KVM_EXIT_REASON since the vm starts, and
+#        it is recorded every second by setting `sleep(1)`.
```
yonghong-song, line 18: @COUNT: the counts of the @KVM_EXIT_REASONS.

"since vm starts" is not accurate. the counting starts when bpf starts to run. The "sleep(1)" is not true either.

tools/kvmexit.py
```diff
@@ -0,0 +1,293 @@
...
+# Display the exit_reason and its statistics of each kvm_exit
+# for all vcpus of all virtual machines. For example,
+# $./kvmexit.py
+#  TGID     PID      KVM_EXIT_REASON                     STAT
+#  1273551  1273568  EXIT_REASON_MSR_WRITE               6
+#  1274253  1274261  EXIT_REASON_EXTERNAL_INTERRUPT      1
+#  1274253  1274261  EXIT_REASON_HLT                     12
+#  ...
+#
+# @TGID: each vitual machine's pid in the user space.
```
yonghong-song, line 14: This is kernel terminology. Use "PID" instead.

tools/kvmexit.py
```diff
@@ -0,0 +1,293 @@
...
+# @PID: the user space's thread of each vcpu of that virtual machine.
```
yonghong-song, line 15: Using "TID". Also, do you think TID is useful? I am not quite sure about it.

tools/kvmexit.py
```diff
@@ -0,0 +1,293 @@
...
+# @KVM_EXIT_REASON: the reason why the vm exits.
+# @STAT: the counts of the @KVM_EXIT_REASON since the vm starts, and
+#        it is recorded every second by setting `sleep(1)`.
+#
+# REQUIRES: Linux 4.7+ (BPF_PROG_TYPE_TRACEPOINT support)
+#
+# Copyright (c) 2019 BYTEDANCE Inc.
```
yonghong-song, line 22: 2021

tools/kvmexit.py
```diff
@@ -0,0 +1,293 @@
...
+    try:
+        ival = int(val)
+    except ValueError:
+        raise argparse.ArgumentTypeError("must be an integer")
+
+    if ival < 0:
+        raise argparse.ArgumentTypeError("must be positive")
+    return ival
+
+def positive_nonzero_int(val):
```
yonghong-song, line 51: I think positive_int and positive_nonzero_int are not needed.

tools/kvmexit.py
```diff
@@ -0,0 +1,293 @@
...
+examples = """examples:
+    ./kvmexit             # Display kvm_exit_reason and its statistics until Ctrl-C
+    ./kvmexit 5           # Display kvm_exit_reason and its statistics for 5 seconds only
+"""
+parser = argparse.ArgumentParser(
+    description="Display kvm_exit_reason and its statistics at a timed interval",
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    epilog=examples)
+parser.add_argument("duration", nargs="?", default=99999999,
+    type=positive_nonzero_int,
```
yonghong-song, line 67: Just use "type=int" and we are done with it. Similar to most of other tools, no need to validate actual value. Tools users are supposed to look at example/man page and provides a sane input.

tools/kvmexit.py
```diff
@@ -0,0 +1,293 @@
...
+};
+BPF_PERCPU_ARRAY(pcpu_cache, struct cache_info, 1);
+
+FUNC_ENTRY {
+    u64 cur_pid_tgid = bpf_get_current_pid_tgid();
+    int zero = 0;
+    int cache_miss = 0;
+    u32 er = GET_ER;
+    if (er >= REASON_NUM) {
+        return -1;
```
yonghong-song, line 102: Just return 0 here. For raw_tracepoint, return value is not used. for tracepoint, return 1 is different from return 0. return 1 permits kernel perf event output continues, this is important when you enable debugfs tracepoint. but I prefers to return 0 in this function for such complexity. It is unlikely people will use debugfs tracepoint if this tool is used.

tools/kvmexit.py
```diff
@@ -0,0 +1,293 @@
...
+        if (cache_p->cache_pid_tgid == cur_pid_tgid) {
+            //a. If the cur_pid_tgid hit this physical cpu consecutively, save it to pcpu_cache
+            tmp_info = &cache_p->cache_exit_ct;
+        } else {
+            //b. If another pid_tgid matches this pcpu for the last hit, OR it is the first time to hit this physical cpu.
+            cache_miss = 1;
+        }
+    } else {
+        return -1;
+    }
```
yonghong-song, line 118: typically we should do

  if (cache_p == NULL)
    return 0;
  ...

to reduce nesting.

Also this caching stuff seems complicated. Will go back to this once I understand use case better.

tools/kvmexit.py
```diff
@@ -0,0 +1,293 @@
...
+try:
+    has_kvm = subprocess.check_output("lsmod | grep kvm", shell=True)
+except Exception as e:
+    raise Exception("Please insmod kvm module to use kvmexit")
+
+try:
+    uname = subprocess.check_output("uname -r", shell=True).split('.')[:2]
+    if int(uname[0]) >= 5:
+        func_entry = "RAW_TRACEPOINT_PROBE(kvm_exit)"
+        get_er = "ctx->args[0]"
```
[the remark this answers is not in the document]
yonghong-song: Agree, please use `BPF.support_raw_tracepoint()`.

tools/kvmexit.py
```diff
@@ -0,0 +1,293 @@
...
+b = BPF(text=bpf_text)
+
+
+# header
+print("Display kvm exit reasons and statistics", end="")
+if duration < 99999999:
+    print(" for %d secs." % duration)
+else:
+    print("... Hit Ctrl-C to end.")
+print("%-8s %-8s %-35s %s" % ("TGID", "PID", "KVM_EXIT_REASON", "STAT"))
```
yonghong-song, line 264: COUNT

tools/kvmexit.py
```diff
@@ -0,0 +1,293 @@
...
+    "EXIT_REASON_TPAUSE"
+)
+
+try:
+    has_kvm = subprocess.check_output("lsmod | grep kvm", shell=True)
+except Exception as e:
+    raise Exception("Please insmod kvm module to use kvmexit")
+
+try:
+    uname = subprocess.check_output("uname -r", shell=True).split('.')[:2]
```
yonghong-song, line 236: In python3, I need the following to make the tool work:
```
 try:
-    uname = subprocess.check_output("uname -r", shell=True).split('.')[:2]
+    uname = subprocess.check_output("uname -r", shell=True).split(b'.')[:2]
```

tools/kvmexit_example.txt
```diff
@@ -0,0 +1,106 @@
+Demonstrations of kvm exit reasons, the Linux eBPF/bcc version.
+
+
+This tool displays the detail exit_reason and its statistics of each
+kvm_exit for all virtual machines running on this physical machine.
```
yonghong-song, line 5: The above explanation only shows what you intended to do. It would be good to explain how
such information will help you solve any particular issues. You mentioned you have many VMs and performance is important, how many VMs do you have in each physical box?

For example, I tried on simple qemu/kvm run (start and at login prompt, quit), I got:
```
$ sudo python3.6 kvmexit.py
Display kvm exit reasons and statistics... Hit Ctrl-C to end.
TGID     PID      KVM_EXIT_REASON                     STAT
^C4036993  4037007  EXIT_REASON_EXTERNAL_INTERRUPT      728     
4036993  4037007  EXIT_REASON_CPUID                   196     
4036993  4037007  EXIT_REASON_HLT                     1447    
4036993  4037007  EXIT_REASON_VMCALL                  424     
4036993  4037007  EXIT_REASON_CR_ACCESS               11      
4036993  4037007  EXIT_REASON_DR_ACCESS               1       
4036993  4037007  EXIT_REASON_IO_INSTRUCTION          19642   
4036993  4037007  EXIT_REASON_MSR_READ                92      
4036993  4037007  EXIT_REASON_MSR_WRITE               1989    
4036993  4037007  EXIT_REASON_PAUSE_INSTRUCTION       666     
4036993  4037007  EXIT_REASON_GDTR_IDTR               6       
4036993  4037007  EXIT_REASON_LDTR_TR                 3       
4036993  4037007  EXIT_REASON_EPT_VIOLATION           2658    
4036993  4037007  EXIT_REASON_PREEMPTION_TIMER        690     
4036993  4037007  EXIT_REASON_WBINVD                  1       
4036993  4037007  EXIT_REASON_XSETBV                  1       
4036993  4037007  EXIT_REASON_APIC_WRITE              1182
```
How does the above information helps you?

You could you use debugfs kvm/kvm_exit tracepoint to get the same information, I guess the reason might be due to performance reason so you want to do in-kernel aggregation. So clarifying your use case is important here.

tools/kvmexit.py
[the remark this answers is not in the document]
ShirleyFei: OK, thanks for the suggestion. :)

tools/kvmexit.py
ShirleyFei: > Use `len(exit_reasons)` instead please.

Thanks for the reminder.

tools/kvmexit.py
ShirleyFei: > Just use "type=int" and we are done with it. Similar to most of other tools, no need to validate actual value. Tools users are supposed to look at example/man page and provides a sane input.

Got it, thanks for the reminder :)

tools/kvmexit.py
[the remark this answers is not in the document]
ShirleyFei: > BPF.support_raw_tracepoint()

How about the latter judgement on whether kernel version is newer than 4.7? Would it be more friendly to use `uname` to do the judgement and leave a message, or just omit the judgement like https://github.com/iovisor/bcc/blob/master/tools/runqlat.py#L207 does?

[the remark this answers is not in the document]
ShirleyFei: > Use `BPF.support_raw_tracepoint()` instead.

How about the latter judgement on whether kernel version is newer than 4.7? Would it be more friendly to use uname to do the judgement and leave a message, or just omit the judgement like https://github.com/iovisor/bcc/blob/master/tools/runqlat.py#L207 does?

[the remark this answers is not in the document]
ShirleyFei: > Use `BPF.support_raw_tracepoint()` instead.

Actually there are two stages to fully support raw_tracepoint: the first stage is only for in-kernel functions, and the second stage is for kernel modules and it is merged into kernel later than v5.0 (Ref: https://lore.kernel.org/patchwork/patch/1009374/). Considering kvm is a kernel module, maybe add a new function: **support_raw_tracepoint_module()**  to check this will be more appropriate? Just as follows: 
```
@@ -1010,6 +1010,12 @@ class BPF(object):
             return True
         return False
    
    def support_raw_tracepoint_module():
        # kernel symbol "bpf_trace_modules" indicates raw_tracepoint module support
        if BPF.ksymname("bpf_trace_modules") != -1:
            return True
        return False
```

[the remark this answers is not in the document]
ShirleyFei: > > Use `BPF.support_raw_tracepoint()` instead.
> 
> Actually there are two stages to fully support raw_tracepoint: the first stage is only for in-kernel functions, and the second stage is for kernel modules and it is merged into kernel later than v5.0 (Ref: https://lore.kernel.org/patchwork/patch/1009374/). Considering kvm is a kernel module, maybe add a new function: **support_raw_tracepoint_module()** to check this will be more appropriate? Just as follows:
> 
> ```
> --- a/src/python/bcc/__init__.py
> +++ b/src/python/bcc/__init__.py
> @@ -1010,6 +1010,12 @@ class BPF(object):
>              return True
>          return False
>     
>     def support_raw_tracepoint_module():
>         # kernel symbol "bpf_trace_modules" indicates raw_tracepoint module support
>         if BPF.ksymname("bpf_trace_modules") != -1:
>             return True
>         return False
> ```

Besides, how about the latter judgement on whether kernel version is older than 4.7? Would it be more friendly to use uname to do the judgement and leave a message, or just omit the judgement like https://github.com/iovisor/bcc/blob/master/tools/runqlat.py#L207 does?

tools/kvmexit.py
ShirleyFei: > Existing tools in BCC simply use `int`, you can perform this check after `parser.parse_args`.

Sorry that I did not get the point. I guess I referred to the `duration` in profile.py and offcputime.py. Would it be better to check the positive_int after parser.parse_args?

tools/kvmexit.py
ShirleyFei: > Instead of STAT, let us use COUNT, which states what the data is more accurately.

No problem, will reword this and the following.

tools/kvmexit.py
ShirleyFei: > Using "TID". Also, do you think TID is useful? I am not quite sure about it.

Thanks. Yes, in some cases only one specific TID exits frequently or has an abnormal KVM_EXIT_REASON.

tools/kvmexit.py
ShirleyFei: > Please merge these two `if` statements.

Actually the first statements should be `if (tmp_info != NULL) {`, so sorry for the mistake. :( For more detail, please refer to the `Help to understand` and that workflow in tools/kvmexit_example.txt

tools/kvmexit.py
ShirleyFei: > This is kernel terminology. Use "PID" instead.

Ok, will change them to "PID" and "TID".

tools/kvmexit.py
ShirleyFei: > Just return 0 here. For raw_tracepoint, return value is not used. for tracepoint, return 1 is different from return 0. return 1 permits kernel perf event output continues, this is important when you enable debugfs tracepoint. but I prefers to return 0 in this function for such complexity. It is unlikely people will use debugfs tracepoint if this tool is used.

Would you mind pointing the code saying "return 1 permits kernel perf event output"?  BTW, will "return -1" bother this logic?

tools/kvmexit.py
ShirleyFei: > typically we should do
> 
> if (cache_p == NULL)
> return 0;
> ...
> 
> to reduce nesting.
NP
> 
> Also this caching stuff seems complicated. Will go back to this once I understand use case better.
Would the `Help` and `flow` in "tools/kvmexit_example.txt" be helpful to understand the caching stuff? Hope they  work, as that's what they're supposed to..

tools/kvmexit_example.txt
ShirleyFei: > The above explanation only shows what you intended to do. It would be good to explain how
> such information will help you solve any particular issues. You mentioned you have many VMs and performance is important, how many VMs do you have in each physical box?
> 
The mainly purpose of doing this is for performance optimization in the virtual machine scenario. As we know, vms'  frequent exits can cause performance problems, thus we want to locate the frequent exited reasons and then find solutions to reduce or avoid the exit. 

> For example, I tried on simple qemu/kvm run (start and at login prompt, quit), I got:
> 
> ```
> $ sudo python3.6 kvmexit.py
> Display kvm exit reasons and statistics... Hit Ctrl-C to end.
> TGID     PID      KVM_EXIT_REASON                     STAT
> ^C4036993  4037007  EXIT_REASON_EXTERNAL_INTERRUPT      728     
> 4036993  4037007  EXIT_REASON_CPUID                   196     
> 4036993  4037007  EXIT_REASON_HLT                     1447    
> 4036993  4037007  EXIT_REASON_VMCALL                  424     
> 4036993  4037007  EXIT_REASON_CR_ACCESS               11      
> 4036993  4037007  EXIT_REASON_DR_ACCESS               1       
> 4036993  4037007  EXIT_REASON_IO_INSTRUCTION          19642   
> 4036993  4037007  EXIT_REASON_MSR_READ                92      
> 4036993  4037007  EXIT_REASON_MSR_WRITE               1989    
> 4036993  4037007  EXIT_REASON_PAUSE_INSTRUCTION       666     
> 4036993  4037007  EXIT_REASON_GDTR_IDTR               6       
> 4036993  4037007  EXIT_REASON_LDTR_TR                 3       
> 4036993  4037007  EXIT_REASON_EPT_VIOLATION           2658    
> 4036993  4037007  EXIT_REASON_PREEMPTION_TIMER        690     
> 4036993  4037007  EXIT_REASON_WBINVD                  1       
> 4036993  4037007  EXIT_REASON_XSETBV                  1       
> 4036993  4037007  EXIT_REASON_APIC_WRITE              1182
> ```
> 
> How does the above information helps you?
> 
> You could you use debugfs kvm/kvm_exit tracepoint to get the same information, I guess the reason might be due to performance reason so you want to do in-kernel aggregation. So clarifying your use case is important here.

Yes, using bcc/ebpf is for performance considerations.  Thanks for the kindly reminder, I will do the polish. And then push a new version for all modifications later. :)

pizhenwei: Linux only supports several exit reasons 'exits  halt_exits  io_exits  irq_exits  mmio_exits  signal_exits' for a virtual machine, actually we need to know the full reasons especially wrmsr/rdmsr/ept-violation/ple on a x86 server. This could not be implemented in  upstream kernel(debugfs kvm) because it's x86-specific([discuss](https://patchwork.kernel.org/project/kvm/patch/1555939499-30854-1-git-send-email-pizhenwei@bytedance.com/) about this).

tools/kvmexit.py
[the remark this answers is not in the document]
yonghong-song: support_raw_tracepoint_module() sounds okay to me. We don't need to print out the message related to kernel version. Some kernels have backports, so in general, bcc tries very hard not to tie the feature with a particular kernel  version.

tools/kvmexit.py
yonghong-song: Although it is nice to check, but I think it is okay, some of tools have the check but most tools does not. But it is okay for you to do a check of the positive_int. I won't insist.

tools/kvmexit_example.txt
yonghong-song: All these are important reasons to have a separate tool where tracepoints cannot provide, e.g., 
   - tracepoint won't do in-kernel aggregation, and bpf program provides nice kernel aggregation, bpf in-kernel cache may further improve performance.
   - tracepoint missed some arch (e.g. x86) specific kvm reasons.
   - the detailed exit reason distribution may help tuning qemu and guest kernel to avoid hot vm exit.
It would be good if you can put some versions of the above reasoning in the beginning of the tool as the comments os it would be clear of why and what this tool can help.

ShirleyFei: > All these are important reasons to have a separate tool where tracepoints cannot provide, e.g.,
> 
> * tracepoint won't do in-kernel aggregation, and bpf program provides nice kernel aggregation, bpf in-kernel cache may further improve performance.
> * tracepoint missed some arch (e.g. x86) specific kvm reasons.
> * the detailed exit reason distribution may help tuning qemu and guest kernel to avoid hot vm exit.
>   It would be good if you can put some versions of the above reasoning in the beginning of the tool as the comments os it would be clear of why and what this tool can help.

Hi @yonghong-song , thanks for your kindly suggestions, and I have prepared a version 2 based on v1, please help to review. Thanks a lot. :)

src/python/bcc/__init__.py
```diff
@@ -1123,6 +1123,18 @@ def support_raw_tracepoint():
             return True
         return False
 
+    @staticmethod
+    def support_raw_tracepoint_module():
+        # kernel symbol "bpf_trace_modules" indicates raw tp module support, ref: kernel commit a38d1107
```
yonghong-song, line 1128: Let us name it as `support_raw_tracepoint_in_module` to be more clear what we are testing here.
So "indicates raw tp module support" => "indicates raw tp support in modules".

tools/kvmexit.py
```diff
@@ -0,0 +1,393 @@
...
+#
+# Besides, we also allow users to specify one pid, tid(s), or one
+# pid and its vcpu. See kvmexit_example.txt for more examples.
+#
+# @PID: each vitual machine's pid in the user space.
+# @TID: the user space's thread of each vcpu of that virtual machine.
+# @KVM_EXIT_REASON: the reason why the vm exits.
+# @COUNT: the counts of the @KVM_EXIT_REASONS.
+#
+# REQUIRES: Linux 4.7+ (BPF_PROG_TYPE_TRACEPOINT support)
```
yonghong-song, line 22: I have
```
$ git describe a38d1107
v4.20-rc6-1353-ga38d1107f937
```
So looks like we need 4.20+.

tools/kvmexit.py
yonghong-song: In kernel/events/core.c, we have
```
  struct trace_event_call *call, u64 count,
                               struct pt_regs *regs, struct hlist_head *head,
                               struct task_struct *task)
{
        if (bpf_prog_array_valid(call)) {
                *(struct pt_regs **)raw_data = regs;
                if (!trace_call_bpf(call, raw_data) || hlist_empty(head)) {
                        perf_swevent_put_recursion_context(rctx);
                        return;
                }
        }
        perf_tp_event(call->event.type, count, raw_data, size, regs, head,
                      rctx, task);
}
EXPORT_SYMBOL_GPL(perf_trace_run_bpf_submit);
```

If bpf program returns none zero, it is possible control may go to perf_tp_event(). But this is for regular tracepoint, previously I probably didn't pay attention to the fact that it used raw tracepoint here.

For raw tracepoint, the return value is ignored. See c4f6699dfcb85 ("bpf: introduce BPF_RAW_TRACEPOINT"). So let us just return 0 in all places.

tools/kvmexit.py
```diff
@@ -0,0 +1,393 @@
...
+    if os.access("/dev/kvm", os.R_OK | os.W_OK):
+        pass
+    else:
+        raise Exception("Please insmod kvm module to use kvmexit tool.")
+except Exception as e:
+    raise Exception("Failed to do precondition check, due to: %s." % e)
+
+try:
+    # Let's firstly try raw_tracepoint_module
+    if support_raw_tracepoint_module():
```
yonghong-song, line 264: This won't work. The correct way is `BPF.support_raw_tracepoint_in_module()`.

tools/kvmexit.py
```diff
@@ -0,0 +1,393 @@
...
+    header_format = "PID      TID      "
+bpf_text = bpf_text.replace('THREAD_FILTER', thread_filter)
+
+# For kernel >= 5.0, use RAW_TRACEPOINT_MODULE for performance consideration
+bpf_text = bpf_text.replace('FUNC_ENTRY', func_entry)
+bpf_text = bpf_text.replace('GET_ER', get_er)
+b = BPF(text=bpf_text)
+
+
+
```
yonghong-song, line 322: one less empty line?

tools/kvmexit.py
```diff
@@ -0,0 +1,393 @@
...
+# signal handler
+def signal_ignore(signal, frame):
+        print()
+try:
+    sleep(duration)
+except KeyboardInterrupt:
+    signal.signal(signal.SIGINT, signal_ignore)
+
+
+
```
yonghong-song, line 340: one less empty line?

tools/kvmexit_example.txt
```diff
@@ -0,0 +1,242 @@
...
+# ./kvmexit.py -h
+usage: kvmexit.py [-h] [-p PID [-v VCPU | -a] ] [-t TID | -T 'TID1,TID2'] [duration]
+
+Display kvm_exit_reason and its statistics at a timed interval
+
+optional arguments:
+  -h, --help            show this help message and exit
+  -p PID, --pid PID     display process with this PID only, collpase all tids with exit reasons sorted in descending order
+   -v VCPU, --v VCPU    display this VCPU only for this PID
+   -a, --alltids        display all TIDS for this PID
```
yonghong-song, line 227: some alignment issue here.

tools/kvmexit.py
ShirleyFei: > I have
> 
> ```
> $ git describe a38d1107
> v4.20-rc6-1353-ga38d1107f937
> ```
> 
> So looks like we need 4.20+.

When this tool is used by users, it will try to firstly employ raw tracepoints in modules, if fails then fall back to regular tracepoint. Thus when referring to 'REQUIRES', I think we'd better mention the time when regular tracepoint is introduced, instead of the time when raw tp is introduced. (The commit: a38d1107 is introduced to support raw tracepoints in modules) :)

tools/kvmexit.py
ShirleyFei: > This won't work. The correct way is `BPF.support_raw_tracepoint_in_module()`.

Thanks for the reminder

tools/kvmexit.py
ShirleyFei: > one less empty line?

OK, no problem :)

tools/kvmexit_example.txt
ShirleyFei: > some alignment issue here.

Emmm, actually '-v' and '-a' is not aligned on purpose, as they are sub-parameters for '-p'.  But now I am intending to make them aligned,, after all, the "examples" and "usage" have explained all. :)

tools/kvmexit.py
yonghong-song: Okay. 4.7+ is fine then.

tools/kvmexit_example.txt
```diff
@@ -0,0 +1,242 @@
...
+Considering virtual machines' frequent exits can cause performance problems,
+this tool aims to locate the frequent exited reasons and then find solutions
+to reduce or even avoid the exit, by displaying the detail exit reasons and
+the counts of each vm exit for all vms running on one physical machine.
+
+
+#Features of this tool
+- Although there is a patch: [KVM: x86: add full vm-exit reason debug entries]
+  (https://patchwork.kernel.org/project/kvm/patch/1555939499-30854-1-git-send-email-pizhenwei@bytedance.com/)
+  tring to fill more vm-exit reason debug entries, just as the comments said,
```
yonghong-song, line 13: tring => trying

yonghong-song (commented):
A few minor comments. And could you add a man page in the next revision?

tools/kvmexit_example.txt
ShirleyFei: Thanks, will update them. :)

tools/kvmexit.py
```diff
@@ -0,0 +1,393 @@
...
+    if (cache_p->cache_pid_tgid == cur_pid_tgid) {
+        //a. If the cur_pid_tgid hit this physical cpu consecutively, save it to pcpu_cache
+        tmp_info = &cache_p->cache_exit_ct;
+    } else {
+        //b. If another pid_tgid matches this pcpu for the last hit, OR it is the first time to hit this physical cpu.
+        cache_miss = 1;
+    }
+
+    if (tmp_info == NULL) {
```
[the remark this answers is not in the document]
ShirleyFei: > We can remove line 128 - 130. They are not needed.

Indeed,  removed them already.

ShirleyFei:
Hi @yonghong-song, I add a new "man/man8/kvmexit.8" file in the 3rd version, and also add a ref in "README.md". Please help to review, thanks a lot. :)

yonghong-song:
[buildbot, test this please]

yonghong-song:
LGTM. Thanks!

anakryiko:
I'm a bit late to the party here. But were there any strong reasons to write a new tool in Python vs doing it as part of libbpf-tools? According to http://www.brendangregg.com/blog/2020-11-04/bpf-co-re-btf-libbpf.html, specifically this part:

> Coding performance tools in BCC Python is now considered deprecated as we move to libbpf C with BTF and CO-RE

... it probably would be better to start off as libbpf-tool. May I interest anyone to now try to re-create it with libbpf?

chenhengqi:
> I'm a bit late to the party here. But were there any strong reasons to write a new tool in Python vs doing it as part of libbpf-tools? According to http://www.brendangregg.com/blog/2020-11-04/bpf-co-re-btf-libbpf.html, specifically this part:
> 
> > Coding performance tools in BCC Python is now considered deprecated as we move to libbpf C with BTF and CO-RE
> 
> ... it probably would be better to start off as libbpf-tool. May I interest anyone to now try to re-create it with libbpf?

Cheers. Will do.

ShirleyFei:
> I'm a bit late to the party here. But were there any strong reasons to write a new tool in Python vs doing it as part of libbpf-tools? According to http://www.brendangregg.com/blog/2020-11-04/bpf-co-re-btf-libbpf.html, specifically this part:
> 
> > Coding performance tools in BCC Python is now considered deprecated as we move to libbpf C with BTF and CO-RE
> 
> ... it probably would be better to start off as libbpf-tool. May I interest anyone to now try to re-create it with libbpf?
Hi @anakryiko, thanks for the suggestion, and sorry for the late reply. Yes, we know libbpf-tool is better than BCC python when referring to friendly debugging, lightweight dependencies and CO-RE. And the main reason to write in Python is trying to make it a more universal tool, especially considering many companies’ servers in production environment do not have the appropriate libc6 package which must be higher than v2.26 due to the limitation mentioned in https://www.brendangregg.com/blog/2020-11-04/bpf-co-re-btf-libbpf.html: "Note to distro maintainers: it requires pahole >= 1.16" (The dwarves package needs higher version of libc6).  Except the above backward compatibility reason, we indeed should embrace libbpf-tools instead of BCC python when introducing a new tool. So I am in favor of rewriting the kvm_exit_reason tool as libbpf-tool, and will commence on it. But I still think leave it with BCC python is a good idea at least in short term to make it an universal tool, and later in one day we can remove all existed tools writing BCC python together. :) Thanks for the suggestion again, have a nice day.

ShirleyFei:
> > I'm a bit late to the party here. But were there any strong reasons to write a new tool in Python vs doing it as part of libbpf-tools? According to http://www.brendangregg.com/blog/2020-11-04/bpf-co-re-btf-libbpf.html, specifically this part:
> > > Coding performance tools in BCC Python is now considered deprecated as we move to libbpf C with BTF and CO-RE
> > 
> > 
> > ... it probably would be better to start off as libbpf-tool. May I interest anyone to now try to re-create it with libbpf?
> 
> Cheers. Will do.

Hi @chenhengqi, have you commenced on the rewriting?  The logic for this tool is a little obscure, maybe we can work together if proper. Or if you have not started, I'd like to work on it then. :)

chenhengqi:
@ShirleyFei Feel free to do that. Cheers.

Yeah, IMO, the current implementation is a little complicated. :)

I've read the *BPF Performance Tools* book which has a relative simple kvmexit tool, I would like to stick to that.

Also, add an option for measuring latency between kvm_exit and kvm_entry as Brendan commented.

ShirleyFei:
> @ShirleyFei Feel free to do that. Cheers.
> 
> Yeah, IMO, the current implementation is a little complicated. :)

Yes, especially the per cpu array and hash. Actually they are introduced in order to improve performance, but at the same time this increases the difficulty of understanding. For background, realization and examples, you can refer to the chart in kvmexit_example.txt.
> 
> I've read the _BPF Performance Tools_ book which has a relative simple kvmexit tool, I would like to stick to that.

Sure, and any questions please feel free to contact me. :)
> 
> Also, add an option for measuring latency between kvm_exit and kvm_entry as Brendan commented.
 
Emmm, do you mean "examples/tracing/kvm_hypercall.py"? I am afraid these are two different tools, and please do not confuse them. You can make them two parts, and submit separately. :)

chenhengqi:
> > @ShirleyFei Feel free to do that. Cheers.
> > Yeah, IMO, the current implementation is a little complicated. :)
> 
> Yes, especially the per cpu array and hash. Actually they are introduced in order to improve performance, but at the same time this increases the difficulty of understanding. For background, realization and examples, you can refer to the chart in kvmexit_example.txt.
> 
> > I've read the _BPF Performance Tools_ book which has a relative simple kvmexit tool, I would like to stick to that.
> 
> Sure, and any questions please feel free to contact me. :)
> 
> > Also, add an option for measuring latency between kvm_exit and kvm_entry as Brendan commented.
> 
> Emmm, do you mean "examples/tracing/kvm_hypercall.py"? I am afraid these are two different tools, and please do not confuse them. You can make them two parts, and submit separately. :)

See the comment in this commit: https://github.com/iovisor/bcc/commit/423656396e3c8454924daefc9a8dc0496414d748

PR #3480, opened 2021-06-09
bashreadline: show UID and GID

cristiklein:
After this PR, UID and GID are also shown:

```
$ sudo ./bashreadline.py 
TIME      PID    UID    GID    COMMAND
12:15:40  289760 1000   1000   how-are-you
12:16:25  291983 1000   1000   git checkout -b bashreadline-add-uid-gid
12:16:27  291983 1000   1000   git add -p
12:16:29  291983 1000   1000   git status
12:16:38  291983 1000   1000   git commit -m "bashreadline: show UID and GID"
12:16:40  291983 1000   1000   git push
12:16:44  291983 1000   1000       git push --set-upstream origin bashreadline-add-uid-gid
12:17:01  291983 1000   1000   git remote add ck git@github.com:cristiklein/bcc
12:17:08  291983 1000   1000   git push -u ck bashreadline-add-uid-gid
12:17:20  291983 1000   1000   git push -u ck bashreadline-add-uid-gid
```

### DIFF
--- a/tools/bashreadline.py
+++ b/tools/bashreadline.py
@@ -39,6 +39,8 @@ bpf_text = """
 
 struct str_t {
     u64 pid;
+    u32 uid;
+    u32 gid;
     char str[80];
 };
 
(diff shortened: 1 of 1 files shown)


brendangregg:
Thanks, but this is clutter that makes the default output more difficult to read through (and increasing the possibility of >80 chars) while also adding very little value (this is a basic example tool for uprobe capabilities). If there was an actual production use case (if!) then extra fields would be an option (like -v). I hope it's especially clear in this case that the tool is a hello-world style example of uprobes, and adding stuff to it is really not the point. I need to add a note to CONTRIBUTING-SCRIPTS.md about this type of PR. Please think long and hard before changing the _default_ output of anything.

When I wrote the tools, if I thought the default output needed something, I already put it there. The only tool where the default output is lacking (that springs to mind) is biosnoop -- which should include the filename or pathname (if available) for disk I/O.

cristiklein:
Thanks for the quick feedback. I guess I got too enthusiastic when I read "Advanced: develop bcc tools, contribute to bcc & bpftrace" in the eBPF tutorial. :smile:

PR #3483, opened 2021-06-11
Avoid attempting to build the cpp examples unless needed

sevan:
Though the cpp components take a lot of resources as they're statically
linked, it is much worse when llvm is unstripped and was built with debug
symbols. The build consumes 10s of gigabytes of RAM before being killed if
the resources are not available, for components which are not going to be
installed by default, hence only attempt to build them if actually required.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,10 +1,14 @@
 set(EXAMPLE_PROGRAMS hello_world.py)
 install(PROGRAMS ${EXAMPLE_PROGRAMS} DESTINATION share/bcc/examples)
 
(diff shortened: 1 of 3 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

yonghong-song:
Thinking twice. I think we should have a separate CMake variable to control build, e.g., we gave user three options:
  - no build and install for examples
  - build for examples, but not install
  - build and install for examples
The reason is some users may want to build and experiment locally without install. Forcing build + install together may not be good for these users. WDYT?

yonghong-song:
[buildbot, ok to test]

sevan:
Makes sense, ok

PR #3485, opened 2021-06-11
libbpf-tools: remove ext4dist

chenhengqi:
In #3441, we introduce a new libbpf tools named fsdist, which is
built on the idea by @anakryiko and previous work by @ethercflow.
fsdist extends ext4dist to support multiple file systems in a flexable
way. Now we can replace ext4dist and treat it as an alias to fsdist.
This commit removes ext4dist and replaces it with a symlink to fsdist.

References:
    #3430, #3436


anakryiko (approved):
Thanks for quick turnaround!

yonghong-song:
[buildbot, test this please]

PR #3486, opened 2021-06-11
Decode bytes when formatting them as strings

### LINKED ISSUE #3478: Inconsistent string encoding breaks tools/tplist UI and behavior
When I use `tplist`, it prints out identifiers which include a Python string kind identifier in front:

```
-> % sudo tplist -p $(pgrep -f 'python main.py')                     
b'/proc/17593/root/usr/bin/python3.8' b'python':b'gc__start'
b'/proc/17593/root/usr/bin/python3.8' b'python':b'gc__done'
b'/proc/17593/root/usr/bin/python3.8' b'python':b'audit'
b'/proc/17593/root/usr/bin/python3.8' b'python':b'import__find__load__start'
b'/proc/17593/root/usr/bin/python3.8' b'python':b'import__find__load__done'
[...]
```

I would have expected output like this:
```
-> % sudo tplist -p $(pgrep -f 'python main.py')                     
/proc/17593/root/usr/bin/python3.8 python:gc__start
/proc/17593/root/usr/bin/python3.8 python:gc__done
/proc/17593/root/usr/bin/python3.8 python:audit
/proc/17593/root/usr/bin/python3.8 python:import__find__load__start
/proc/17593/root/usr/bin/python3.8 python:import__find__load__done
[...]
```

Furthermore, these mangled strings break filtering. I get no results if I do something like this:
```
-> % sudo tplist -p $(pgrep -f 'python main.py') 'python:*'
-> %
```

Instead, I need to explicitly include string type markers:
```
-> % sudo tplist -p $(pgrep -f 'python main.py') "b'python':*"
b'/proc/17593/root/usr/bin/python3.8' b'python':b'gc__start'
b'/proc/17593/root/usr/bin/python3.8' b'python':b'gc__done'
b'/proc/17593/root/usr/bin/python3.8' b'python':b'audit'
b'/proc/17593/root/usr/bin/python3.8' b'python':b'import__find__load__start'
b'/proc/17593/root/usr/bin/python3.8' b'python':b'import__find__load__done'
```

I think this is because `bcc/usdt.py` [creates `USDTProbe` objects with the raw `probe.contents`](https://github.com/iovisor/bcc/blob/1aeec1ffc5517ed47b249ba74a3e2870af36f865/src/python/bcc/usdt.py#L197) from its C calls. The same is true for [`USDTProbeLocation`s created by `USDTProbe`](https://github.com/iovisor/bcc/blob/1aeec1ffc5517ed47b249ba74a3e2870af36f865/src/python/bcc/usdt.py#L120-L125). These are then [formatted with the `%s` directive](https://github.com/iovisor/bcc/blob/1aeec1ffc5517ed47b249ba74a3e2870af36f865/src/python/bcc/usdt.py#L116-L117) in `USDTProbe.short_name()`, and `tplist` uses `short_name()` for its filtering.

It's tempting to change the `USDTProbe` (and `USDTProbeLocation`) constructors to decode their input values when assigning them to attributes, but I think that would break lots of stuff unless we go through and re-encode them, since those attributes are used directly to make lots of calls to the BCC C libraries. Changing only the display side is maybe safer, but will involve chasing down lots of cases where the strings are displayed. For example, [tplist uses `%s` on probe attributes too directly](https://github.com/iovisor/bcc/blob/master/tools/tplist.py#L83-L84).

I built HEAD from source on Ubuntu 20.04. I'm using Python 3.8.5.

spenczar:
Fixes #3478.

USDTProbe objects (and USDTProbeArguments and USDTProbeLocations) are instantiated with data that's sourced from libccc calls. That means that their attributes are bytes-typed, not string-typed.

When a bytes-typed value is rendered into a string with Python's '%s' formatting directive, it gets a wrapped in single quotes and prefixed with b. For example, b'probe-location'. This is visually noisy, but also breaks some tool behavior which uses string-formatted values for stuff like filters.

This is only an issue in Python 3. In Python 2, the bytes type is just an alias for the string type, and so byte sequences from libcc were implicitly decoded as ASCII text.

There are probably more spots like this to be fixed, but maybe we can improve things as we see them.

### DIFF
--- a/src/python/bcc/usdt.py
+++ b/src/python/bcc/usdt.py
@@ -50,24 +50,24 @@ class USDTProbeArgument(object):
         if self.valid & BCC_USDT_ARGUMENT_FLAGS.CONSTANT != 0:
             return "%d" % self.constant
         if self.valid & BCC_USDT_ARGUMENT_FLAGS.DEREF_OFFSET == 0:
-            return "%s" % self.base_register_name
+            return "%s" % self.base_register_name.decode()
         if self.valid & BCC_USDT_ARGUMENT_FLAGS.DEREF_OFFSET != 0 and \
            self.valid & BCC_USDT_ARGUMENT_FLAGS.DEREF_IDENT == 0:
             if self.valid & BCC_USDT_ARGUMENT_FLAGS.INDEX_REGISTER_NAME != 0:
-                index_offset = " + %s" % self.index_register_name
+                index_offset = " + %s" % self.index_register_name.decode()
                 if self.valid & BCC_USDT_ARGUMENT_FLAGS.SCALE != 0:
                     index_offset += " * %d" % self.scale
             else:
                 index_offset = ""
             sign = '+' if self.deref_offset >= 0 else '-'
-            return "*(%s %s %d%s)" % (self.base_register_name,
-                                    sign, abs(self.deref_offset), index_offset)
+            return "*(%s %s %d%s)" % (self.base_register_name.decode(),
+                                      sign, abs(self.deref_offset), index_offset)
         if self.valid & BCC_USDT_ARGUMENT_FLAGS.DEREF_OFFSET != 0 and \
            self.valid & BCC_USDT_ARGUMENT_FLAGS.DEREF_IDENT != 0 and \
            self.valid & BCC_USDT_ARGUMENT_FLAGS.BASE_REGISTER_NAME != 0 and \
            self.base_register_name == "ip":
             sign = '+' if self.deref_offset >= 0 else '-'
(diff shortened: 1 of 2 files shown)


spenczar:
cc @yonghong-song

yonghong-song:
[buildbot, test this please]

yonghong-song:

davemarchevsky:
LGTM. After talking with @yonghong-song, we want to continue supporting python2 for the foreseeable future, so I wonder if it makes sense to add some more general decoding wrapper around all byte-strings coming to python from libbcc.

Regardless, I don't think that thought should prevent this from shipping.

PR #3497, opened 2021-06-19
libbpf-tools: Don't redefine _GNU_SOURCE to avoid redefinition warning

davemarchevsky:
Similar to past commits like 667988ce9e2a051ff608b727f6c89a5baa01fa67,
my toolchain complains that `_GNU_SOURCE` is redefined. Let's only
define it when it passes `ifndef`


yonghong-song:
[buildbot, test this please]

anakryiko:
Where does `_GNU_SOURCE` come from in your setup, Dave? Is it defined through compiler command line (`-D_GNU_SOURCE`?)

Issue #3487, opened 2021-06-13
Stack table symbol resolution uses wrong path for containerized processes


d3dave:
When calling `BPFStackTable::get_stack_symbol(stack_id, pid)` with a PID of a containerized process, symbols are resolved incorrectly, if at all. I tracked the issue down to `ProcSyms::load_exe` where the path to the executable of the process within the container is treated as a path on the host (https://github.com/iovisor/bcc/blob/master/src/cc/bcc_syms.cc#L128). Modifying `get_pid_exe` to simply return the `/proc/<pid>/exe` path instead of the read link appears to work:

```diff
 std::string get_pid_exe(pid_t pid) {
-  char exe_path[4096];
-  int res;
-
-  std::string exe_link = tfm::format("/proc/%d/exe", pid);
-  res = readlink(exe_link.c_str(), exe_path, sizeof(exe_path));
-  if (res == -1)
-    return "";
-  if (res >= static_cast<int>(sizeof(exe_path)))
-    res = sizeof(exe_path) - 1;
-  exe_path[res] = '\0';
-  return std::string(exe_path);
+  return tfm::format("/proc/%d/exe", pid);
 }
```

Note this is only relevant for the executable itself and not any loaded modules because `_add_module` properly handles the `enter_ns` argument. This is not a 100% proper solution as the result of `get_pid_exe` is also used by some USDT-related code which appears to rely on the error check of `readlink`.

Jongy:
Alternatively, since current code already assumes `pid` is running, it can return `/proc/%d/root/path/to/exe`.

Using `/proc/%d/exe` has the benefit of being accessible also when the executable was already deleted, so it's a superior solution to `/proc/%d/root/` IMO. See a [similar fix](https://github.com/benfred/py-spy/pull/364) I made in py-spy to allow profiling "deleted" Python binaries.

yonghong-song:
@d3dave Just want to clarify my understanding. In this case, the bcc tool is running inside the container or outside the container? I assume it should be outside the container, right? Why `readlink` result won't work here? Maybe you can give a step-by-step explanation which will be clear what is the issue?

davemarchevsky:
I think this is a TL;DR of the case @d3dave is describing. Please let me know if I'm not understanding correctly.

[...]

PR #3498, opened 2021-06-19
ProcSyms should treat the executable like any other mapped file when symbolizing

davemarchevsky:
As reported in #3487, when `/proc/PID/exe`'s symlink points to a
mountns-relative path from a different mountns than the tracing process,
we can fail to open it as we don't prepend `/proc/PID/root` .

A few potential solutions were discussed in that issue, we settled on
treating the main exe like any other map in `/proc/PID/maps`. Since it's
always the first map we can reuse existing code and get rid of
exe-specific helpers.

While writing this I noticed that the remaining place `get_pid_exe` is used, the USDT code, potentially also has a bug due to `get_pid_exe` and shouldn't be using it either if so. Will follow up

I still need to validate this. Plan on doing so with `libbpf-tools`' offcputime and hopefully some example that can be committed

### DIFF
--- a/src/cc/bcc_syms.cc
+++ b/src/cc/bcc_syms.cc
@@ -116,29 +116,7 @@ ProcSyms::ProcSyms(int pid, struct bcc_symbol_option *option)
   load_modules();
 }
 
-int ProcSyms::_add_load_sections(uint64_t v_addr, uint64_t mem_sz,
-                                 uint64_t file_offset, void *payload) {
-  auto module = static_cast<Module *>(payload);
-  module->ranges_.emplace_back(v_addr, v_addr + mem_sz, file_offset);
-  return 0;
-}
-
-void ProcSyms::load_exe() {
-  std::string exe = ebpf::get_pid_exe(pid_);
-  Module module(exe.c_str(), exe.c_str(), &symbol_option_);
-
-  if (module.type_ != ModuleType::EXEC)
-    return;
-
-
-  bcc_elf_foreach_load_section(exe.c_str(), &_add_load_sections, &module);
-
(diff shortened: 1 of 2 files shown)


src/cc/bcc_syms.cc
```diff
@@ -116,29 +116,7 @@ ProcSyms::ProcSyms(int pid, struct bcc_symbol_option *option)
...
-
-  bcc_elf_foreach_load_section(exe.c_str(), &_add_load_sections, &module);
-
-  if (!module.ranges_.empty())
-    modules_.emplace_back(std::move(module));
-}
-
 void ProcSyms::load_modules() {
-  load_exe();
   bcc_procutils_each_module(pid_, _add_module, this);
```
davemarchevsky, line 120: No code needs to be added in this PR because the `_add_module` callback passed to the `/proc/PID/maps` iterator compares module 'name's of `/proc/PID/maps` lines - the file path - with already-added `Module` objects. Since `load_exe` was adding to `modules_` a `Module` with name = result of `get_pid_exe`, if the `/proc/PID/maps` entry is the same path, it'll just get ignored.

yonghong-song:
[buildbot, test this please]

davemarchevsky:
This is good to go, changes in functionality are covered by [this existing test](https://github.com/iovisor/bcc/blob/master/tests/cc/test_c_api.cc#L220-L234), which I had trouble getting good signal from on the github actions testrun, so recreated it locally to be sure.

yonghong-song:
>  I had trouble getting good signal from on the github actions testrun, so recreated it locally to be sure.
Does github actions work for you? What do you mean having trouble getting good signals? Anything we can improve here?

PR #3499, opened 2021-06-19
libbpf-tools: display pid instead of tid

chenhengqi:
execsnoop displays tid in its output with header PID,
which is wrong and differs from the original BCC tool.
This commit fixes that with some code cleanup.


danobi (approved):
LGTM

yonghong-song:
[buildbot, test this please]

PR #3504, opened 2021-06-22
bcc-test: fix test error

chenyuezhou:
https://github.com/iovisor/bcc/runs/2882081015#step:7:133 (https://github.com/iovisor/bcc/runs/2882081015#step:4:1786)

I tested it and found that `python2` reports this error when using the version of `pyroute2` after 0.6.1.

In view of the fact that `python2` is no longer updated, the installation version of all the packages that `python2` depends on is specified here to avoid the test failure caused by the update of the related dependencies. (https://github.com/iovisor/bcc/runs/2515123487#step:4:1741)

### DIFF
--- a/Dockerfile.tests
+++ b/Dockerfile.tests
@@ -51,7 +51,7 @@ RUN apt-get update && apt-get install -y \
       libtinfo-dev
 
 RUN pip3 install pyroute2 netaddr dnslib cachetools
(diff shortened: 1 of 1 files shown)


yonghong-song:
Thanks for the fix!

yonghong-song:
[buildbot, test this please]

PR #3506, opened 2021-06-23
tcprtt: fix compatibility for python3

pizhenwei:
Suggested by Yonghong, tcprtt report error on python3:
    TypeError: can't concat str to bytes

Both python2 and python3, inet_ntop returns a string type, there is
no need to encode any more.

Test for python2 and python3, both work fine.


yonghong-song:
[buildbot, test this please]

PR #3508, opened 2021-06-25
tools/tcpaccept compatible with kernel version >= 5.6

chenyuezhou:
After kernel version 5.6, `sk_protocol` is no longer bitfield.

https://patchwork.ozlabs.org/project/netdev/patch/20191219220557.17823-3-mathew.j.martineau@linux.intel.com/

My patch tries to access `->sk_protocol` directly. If the kernel version is before 5.6, it will output the error message `error: address of bit-field requested`, and then works as before. If the kernel version is 5.6 or later, tcpaccept will work properly.

If my patch is acceptable, I will adjust the `bindsnoop` tool in the same way in the next PR.

### DIFF
--- a/tools/tcpaccept.py
+++ b/tools/tcpaccept.py
@@ -108,40 +108,7 @@ int kretprobe__inet_csk_accept(struct pt_regs *ctx)
     if (newsk == NULL)
         return 0;
 
-    // check this is TCP
-    u8 protocol = 0;
-    // workaround for reading the sk_protocol bitfield:
-
-    // Following comments add by Joe Yin:
-    // Unfortunately,it can not work since Linux 4.10,
-    // because the sk_wmem_queued is not following the bitfield of sk_protocol.
-    // And the following member is sk_gso_max_segs.
-    // So, we can use this:
-    // bpf_probe_read_kernel(&protocol, 1, (void *)((u64)&newsk->sk_gso_max_segs) - 3);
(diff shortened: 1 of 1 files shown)


chenyuezhou:
It seems that using `BPF_CORE_READ_BITFIELD_PROBED` would be a better choice, but I got this error when I tried to use it.

```
/virtual/main.c:57:14: warning: implicit declaration of function 'BPF_CORE_READ_BITFIELD_PROBED' is invalid in C99 [-Wimplicit-function-declaration]
```

chenyuezhou:
> It seems that using `BPF_CORE_READ_BITFIELD_PROBED` would be a better choice, but I got this error when I tried to use it.

I didn't add the correct header file, let me solve this problem first.

yonghong-song:
This is not supported yet. There is an effort for bcc to use more libbpf functionalities. cc @davemarchevsky

chenyuezhou:
> This is not supported yet. There is an effort for bcc to use more libbpf functionalities. cc @davemarchevsky

@yonghong-song Excuse me. Is anyone working on this issue?

yonghong-song:
@chenyuezhou @davemarchevsky has some interest on this subject. I think the first step is compile into an object file instead of llvm jit. Once we got an object file (or a buffer with ELF object), we can then take advantage of libbpf parsing functionalities. One example is from rust redbpf,
   https://github.com/foniod/redbpf/blob/main/cargo-bpf/src/llvm.rs
specifically pay attention to function LLVMTargetMachineEmitToFile().

If you have any ideas, feel free to share here.

PR #3516, opened 2021-06-30
cmake: Make libbcc_bpf.so the BCC runtime

danobi:
This commit adds more functionality into libbcc_bpf.so such that
libbcc_bpf.so contains all of BCC's runtime components. "Runtime" in
this context means everything except the stuff that depends on
clang/LLVM.

libbcc_bpf.so was originally created in fa07345 ("make libbpf
standalone-ready") with (I'm guessing) the intent of creating
bcc-libbpf. That has been superceded by libbpf (separate repo) so I
don't think it should be used much anymore.

This updated libbcc_bpf.so will be used by [ahead-of-time compiled
bpftrace scripts][0] to drop the dependency on LLVM/clang for the
runtime component.

[0]: https://dxuuu.xyz/aot-bpftrace.html


yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

yonghong-song:
LGTM. I think libbcc_bpf.{a,so} mostly will be used in bpftrace. Originally it is not even installed. AFAIK, majority users should and will use libbcc.{a,so}.

PR #3519, opened 2021-07-01
libbpf-tools: fix uprobe helper get_elf_func_offset

chenhengqi:
get_elf_func_offset didn't work properly when use with
statically linked binary. It seems like not subtract the
base load address cause the problem. This commits fixes
that like BCC does. see [0] and [1].

[0]: https://github.com/iovisor/bcc/blob/v0.20.0/src/cc/bcc_syms.cc#L751-L764
[1]: https://github.com/iovisor/bcc/blob/v0.20.0/src/cc/bcc_elf.c#L723-L756

### DIFF
--- a/libbpf-tools/uprobe_helpers.c
+++ b/libbpf-tools/uprobe_helpers.c
@@ -228,12 +228,17 @@ off_t get_elf_func_offset(const char *path, const char *func)
 	Elf *e;
 	Elf_Scn *scn;
 	Elf_Data *data;
+	GElf_Ehdr ehdr;
 	GElf_Shdr shdr[1];
+	GElf_Phdr phdr;
 	GElf_Sym sym[1];
-	size_t shstrndx;
+	size_t shstrndx, nhdrs;
 	char *n;
 
 	e = open_elf(path, &fd);
+
+	if (!gelf_getehdr(e, &ehdr))
+		goto out;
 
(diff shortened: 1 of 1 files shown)


libbpf-tools/uprobe_helpers.c
```diff
@@ -261,6 +266,24 @@ off_t get_elf_func_offset(const char *path, const char *func)
...
+			if (phdr.p_type != PT_LOAD || !(phdr.p_flags & PF_X))
+				continue;
+			if (phdr.p_vaddr <= ret && ret < (phdr.p_vaddr + phdr.p_memsz)) {
+				ret = ret - phdr.p_vaddr + phdr.p_offset;
+				goto errout;
+			}
+		}
+		ret = -1;
+	}
+errout:
```
davemarchevsky, line 286: Since this is used for a non-error condition on line 281, could you rename it to something without `err` in the name, or add a second label so that 281's `goto` is easily understood to not be an error state?

davemarchevsky (changes requested):
Aside from my label nitpick, LGTM

libbpf-tools/uprobe_helpers.c
chenhengqi: Done.

davemarchevsky (approved):

yonghong-song:
[buildbot, test this please]

PR #3534, opened 2021-07-14
[py3:tools/deadlock.py] fix usage of str.replace() method to make it py3 compatible

veshij:
deadlock.py fails on py3 with `TypeError: replace() argument 1 must be str, not bytes`

```
  File "/home/oleg/tinfoil.runfiles/libbcc/tools/deadlock.py", line 570, in <module>
    main()
  File "/home/oleg/tinfoil.runfiles/libbcc/tools/deadlock.py", line 480, in main
    text = text.replace(b'MAX_THREADS', str(args.threads));
TypeError: replace() argument 1 must be str, not bytes
```

### DIFF
--- a/tools/deadlock.py
+++ b/tools/deadlock.py
@@ -477,8 +477,8 @@ def main():
 
     with open('deadlock.c') as f:
         text = f.read()
-    text = text.replace(b'MAX_THREADS', str(args.threads));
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

PR #3536, opened 2021-07-15
Add some basic support for LoongArch and fix error when run script

seehearfeel:
bcc: Add some basic support for LoongArch
bcc: Modify {RAW_}TRACEPOINT_PROBE to fix error about incomplete definition

### DIFF
--- a/src/cc/export/helpers.h
+++ b/src/cc/export/helpers.h
@@ -1140,6 +1140,9 @@ int bpf_usdt_readarg_p(int argc, struct pt_regs *ctx, void *buf, u64 len) asm("l
 #elif defined(__TARGET_ARCH_mips)
 #define bpf_target_mips
 #define bpf_target_defined
+#elif defined(__TARGET_ARCH_loongarch)
+#define bpf_target_loongarch
+#define bpf_target_defined
 #else
 #undef bpf_target_defined
 #endif
@@ -1156,6 +1159,8 @@ int bpf_usdt_readarg_p(int argc, struct pt_regs *ctx, void *buf, u64 len) asm("l
 #define bpf_target_powerpc
 #elif defined(__mips__)
 #define bpf_target_mips
+#elif defined(__loongarch__)
+#define bpf_target_loongarch
 #endif
 #endif
 
@@ -1216,6 +1221,18 @@ int bpf_usdt_readarg_p(int argc, struct pt_regs *ctx, void *buf, u64 len) asm("l
 #define PT_REGS_RC(x) ((x)->regs[2])
 #define PT_REGS_SP(x) ((x)->regs[29])
 #define PT_REGS_IP(x) ((x)->cp0_epc)
+#elif defined(bpf_target_loongarch)
+#define PT_REGS_PARM1(x) ((x)->regs[4])
+#define PT_REGS_PARM2(x) ((x)->regs[5])
+#define PT_REGS_PARM3(x) ((x)->regs[6])
+#define PT_REGS_PARM4(x) ((x)->regs[7])
+#define PT_REGS_PARM5(x) ((x)->regs[8])
+#define PT_REGS_PARM6(x) ((x)->regs[9])
+#define PT_REGS_RET(x) ((x)->regs[1])
+#define PT_REGS_FP(x) ((x)->regs[22]) /* Works only with CONFIG_FRAME_POINTER */
+#define PT_REGS_RC(x) ((x)->regs[4])
+#define PT_REGS_SP(x) ((x)->regs[3])
+#define PT_REGS_IP(x) ((x)->csr_epc)
 #else
 #error "bcc does not support this platform yet"
 #endif
--- a/src/cc/export/helpers.h
+++ b/src/cc/export/helpers.h
@@ -1260,10 +1260,10 @@ int bpf_usdt_readarg_p(int argc, struct pt_regs *ctx, void *buf, u64 len) asm("l
 #define lock_xadd(ptr, val) ((void)__sync_fetch_and_add(ptr, val))
 
 #define TRACEPOINT_PROBE(category, event) \
-int tracepoint__##category##__##event(struct tracepoint__##category##__##event *args)
+tracepoint__##category##__##event(struct tracepoint__##category##__##event *args)
 
 #define RAW_TRACEPOINT_PROBE(event) \
-int raw_tracepoint__##event(struct bpf_raw_tracepoint_args *ctx)
+raw_tracepoint__##event(struct bpf_raw_tracepoint_args *ctx)
 
 /* BPF_PROG macro allows to define trampoline function,
  * borrowed from kernel bpf selftest code.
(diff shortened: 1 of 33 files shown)


src/cc/usdt.h
```diff
@@ -156,6 +156,9 @@ class ArgumentParser_x64 : public ArgumentParser {
     REG_SI,
     REG_DI,
     REG_BP,
+#ifdef REG_SP
+#undef REG_SP
+#endif
```
xen0n, line 161: This seems a bit hacky.

src/cc/export/helpers.h
```diff
@@ -1216,6 +1221,18 @@ int bpf_usdt_readarg_p(int argc, struct pt_regs *ctx, void *buf, u64 len) asm("l
...
+#define PT_REGS_PARM2(x) ((x)->regs[5])
+#define PT_REGS_PARM3(x) ((x)->regs[6])
+#define PT_REGS_PARM4(x) ((x)->regs[7])
+#define PT_REGS_PARM5(x) ((x)->regs[8])
+#define PT_REGS_PARM6(x) ((x)->regs[9])
+#define PT_REGS_RET(x) ((x)->regs[1])
+#define PT_REGS_FP(x) ((x)->regs[22]) /* Works only with CONFIG_FRAME_POINTER */
+#define PT_REGS_RC(x) ((x)->regs[4])
+#define PT_REGS_SP(x) ((x)->regs[3])
+#define PT_REGS_IP(x) ((x)->csr_epc)
```
xen0n, line 1235: These bits depend on the kernel ABI being stable. We might have to wait until the LoongArch port of Linux gets upstreamed or at least settles down, only then this could be merged without code churn.

src/cc/usdt.h
seehearfeel: Yes, I agree with you, it is more a matter of necessity.
Maybe we can suggest to modify the macro name REG_SP in libc-dev, 
but it is not so easy, so just undef it now to make it works.

src/cc/export/helpers.h
seehearfeel: Thank you for your review. We actually have an ABI register conversion as below, it is settled for a long time.

```
================= =============== =================== ============
Name              Alias           Usage               Preserved
                                                      across calls
================= =============== =================== ============
``$r0``           ``$zero``       Constant zero       Unused
``$r1``           ``$ra``         Return address      No
``$r2``           ``$tp``         TLS                 Unused
``$r3``           ``$sp``         Stack pointer       Yes
``$r4``-``$r11``  ``$a0``-``$a7`` Argument registers  No
``$r4``-``$r5``   ``$v0``-``$v1`` Return value        No
``$r12``-``$r20`` ``$t0``-``$t8`` Temp registers      No
``$r21``          ``$x``          Reserved            Unused
``$r22``          ``$fp``         Frame pointer       Yes
``$r23``-``$r31`` ``$s0``-``$s8`` Static registers    Yes
================= =============== =================== ============
```
https://git.kernel.org/pub/scm/linux/kernel/git/chenhuacai/linux-loongson.git/tree/Documentation/loongarch/introduction.rst?h=loongarch-next

xen0n: > Thank you for your review. We actually have an ABI register conversion as below, it is settled for a long time.
> 
> ```
> ================= =============== =================== ============
> Name              Alias           Usage               Preserved
>                                                       across calls
> ================= =============== =================== ============
> ``$r0``           ``$zero``       Constant zero       Unused
> ``$r1``           ``$ra``         Return address      No
> ``$r2``           ``$tp``         TLS                 Unused
> ``$r3``           ``$sp``         Stack pointer       Yes
> ``$r4``-``$r11``  ``$a0``-``$a7`` Argument registers  No
> ``$r4``-``$r5``   ``$v0``-``$v1`` Return value        No
> ``$r12``-``$r20`` ``$t0``-``$t8`` Temp registers      No
> ``$r21``          ``$x``          Reserved            Unused
> ``$r22``          ``$fp``         Frame pointer       Yes
> ``$r23``-``$r31`` ``$s0``-``$s8`` Static registers    Yes
> ================= =============== =================== ============
> ```
> https://git.kernel.org/pub/scm/linux/kernel/git/chenhuacai/linux-loongson.git/tree/Documentation/loongarch/introduction.rst?h=loongarch-next

Hi, what I mean by "ABI" is not limited to register convention; things like the field name `regs` also count.

src/cc/export/helpers.h
```diff
@@ -1243,10 +1260,10 @@ int bpf_usdt_readarg_p(int argc, struct pt_regs *ctx, void *buf, u64 len) asm("l
 #define lock_xadd(ptr, val) ((void)__sync_fetch_and_add(ptr, val))
 
 #define TRACEPOINT_PROBE(category, event) \
-int tracepoint__##category##__##event(struct tracepoint__##category##__##event *args)
+tracepoint__##category##__##event(struct tracepoint__##category##__##event *args)
 
 #define RAW_TRACEPOINT_PROBE(event) \
-int raw_tracepoint__##event(struct bpf_raw_tracepoint_args *ctx)
+raw_tracepoint__##event(struct bpf_raw_tracepoint_args *ctx)
```
chenhengqi, line 1266: Is this change necessary ? If yes, I think docs should be updated accordingly.

I am afraid that tools using BCC might be broken due to this change.

davemarchevsky:
Is https://loongson.github.io/LoongArch-Documentation/ the best place to find LoongArch docs?

xen0n:
> Is https://loongson.github.io/LoongArch-Documentation/ the best place to find LoongArch docs? 

Currently yes; that repo is maintained by Loongson employees. However note that only the Volume 1 is available for now; other downloads are placeholders.

davemarchevsky (changes requested):
In addition to comments, can you take a look at what the various arch-specific `ArgumentParser`s are doing in the `src/cc/usdt` dir and see if it's necessary to make any changes there? It [currently defaults to](https://github.com/iovisor/bcc/blob/master/src/cc/usdt/usdt.cc#L45) the x86 ArgumentParser, which doesn't sound like what you want. 

If the tracepoint `#define` changes are due to the commit preceding them, I have a feeling there are some missing changes in `b_frontend_action` or related files. Will dig around to see if anything stands out.

src/cc/usdt.h
seehearfeel: Sorry for the late reply.  Thanks for your suggestion.

The sys/ucontext.h of LoongArch glibc has only an internal version now, I submitted a bug to the related engineer to modify it as LOONGARCH_REG_SP before upstream, so it is not necessary to do this change in src/cc/usdt.h, I will resubmit this patch until LoongArch glibc is settled.

src/cc/export/helpers.h
seehearfeel: I agree with you, I do not want to do this change too, this commit is just a RFC patch,
because this change only makes it work well when expand {RAW_}TRACEPOINT_PROBE on LoongArch,
it needs to do some more analysis about the macro expansion.

PR #3539, opened 2021-07-15
libbpf-tools: add CO-RE tcpretrans

michaelgugino:
This commit adds tcpretrans port to CO-RE. This port currently contains
only code for attaching to trace point tcp_retransmit_skb and does not
fall back to kprobe/kretprobe as in the python implementation.

The count and lossprobe options are not yet implemented.

Signed-off-by: Michael Gugino <mgugino@redhat.com>


libbpf-tools/tcpretrans.h
```diff
@@ -0,0 +1,27 @@
...
+
+struct event {
+	union {
+		__u32 saddr_v4;
+		__u8 saddr_v6[16];
+	};
+	union {
+		__u32 daddr_v4;
+		__u8 daddr_v6[16];
+	};
```
chenhengqi, line 18: Since we use same struct for v4/v6 event, this can be a single field like `__u8 saddr[16]`.
In user space, `inet_ntop` can handle this correctly.

libbpf-tools/tcpretrans.h
```diff
@@ -0,0 +1,27 @@
...
+	__u32 af; // AF_INET or AF_INET6
+	__u32 pid;
+	__u16 dport;
+	__u16 sport;
+	__u64 type;
+	int state;
```
chenhengqi, line 24: Consider reorder these field to avoid padding.

libbpf-tools/Makefile
```diff
@@ -47,6 +47,7 @@ APPS = \
 	syscount \
 	tcpconnect \
 	tcpconnlat \
+	tcpretrans \
```
chenhengqi, line 50: There should be a corresponding entry in `.gitignore`.

libbpf-tools/tcpretrans.bpf.c
```diff
@@ -0,0 +1,68 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2020 Anton Protopopov
+// Copyright (c) 2021 Red Hat
+//
+// Based on tcpconnect.c by Anton Protopopov and
+// tcpretrans(8) from BCC by Brendan Gregg
```
chenhengqi, line 6: C style comment.

libbpf-tools/tcpretrans.bpf.c
```diff
@@ -0,0 +1,68 @@
...
+#include <vmlinux.h>
+
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+
+#include "tcpretrans.h"
+
+/* Define here, because there are conflicts with include files */
+#define AF_INET		2
+#define AF_INET6	10
```
chenhengqi, line 16: weird indention.

libbpf-tools/tcpretrans.c
```diff
@@ -0,0 +1,243 @@
...
+	return vfprintf(stderr, format, args);
+}
+
+static void print_events_header()
+{
+	printf("%-8s %-6s %-2s %-20s %1s> %-20s %-4s\n", "TIME", "PID", "IP",
+		"LADDR:LPORT", "T", "RADDR:RPORT", "STATE");
+}
+
+void handle_event(void *ctx, int cpu, void *data, __u32 data_sz)
```
chenhengqi, line 105: static

libbpf-tools/tcpretrans.c
```diff
@@ -0,0 +1,243 @@
...
+	"TIME_WAIT",
+	"CLOSE",
+	"CLOSE_WAIT",
+	"LAST_ACK",
+	"LISTEN",
+	"CLOSING",
+	"NEW_SYN_RECV"};
+
+
+static volatile sig_atomic_t hang_on = 1;
```
chenhengqi, line 49: Existing tools use `exiting` for this purpose. I think it's good to follow that.

libbpf-tools/tcpretrans.c
```diff
@@ -0,0 +1,243 @@
...
+int main(int argc, char **argv)
+{
+	static const struct argp argp = {
+		.options = opts,
+		.parser = parse_arg,
+		.doc = argp_program_doc,
+		.args_doc = NULL,
+	};
+
+	struct ring_buffer *rb = NULL;
```
chenhengqi, line 192: This `rb` is unused.

libbpf-tools/tcpretrans.bpf.c
michaelgugino: Many files in this directory follow this pattern.

libbpf-tools/tcpretrans.bpf.c
michaelgugino: I think this is a rendering problem with this diff format.  It looks good here and locally: https://github.com/iovisor/bcc/blob/b709dc49c65d2d651a93e66a6451e520955ba09b/libbpf-tools/tcpretrans.bpf.c

libbpf-tools/tcpretrans.c
michaelgugino: syscount.c and tcpconnect. also use the 'hang_on' pattern.

libbpf-tools/tcpretrans.h
michaelgugino: I'm not quite sure what you mean, other event structs like tcpconnect do not pad elements.

libbpf-tools/tcpretrans.h
michaelgugino: I went ahead and attempted this, seems to work locally, let me know what you think.

libbpf-tools/tcpretrans.bpf.c
chenhengqi: You are right.

libbpf-tools/tcpretrans.bpf.c
chenhengqi: Many review comments about code style I posted here was proposed before by maintainers/reviewers of this repo. I think is good to follow those rules which mainly comes from kernel code style guide.

Recent commits added by me all use C style comments which is suggeted by Andrii.

libbpf-tools/tcpretrans.h
chenhengqi: This should make the struct more compact.

libbpf-tools/tcpretrans.h
```diff
@@ -0,0 +1,37 @@
...
+	__u16 dport;
+	__u16 sport;
+};
+
+struct ipv6_flow_key {
+	__u8 saddr[16];
+	__u8 daddr[16];
+	__u16 dport;
+	__u16 sport;
+};
```
chenhengqi, line 35: Consider merge this two struct and adding a type field to distinguish v4/v6.
This way we can save a map in BPF side.

libbpf-tools/tcpretrans.h
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2020 Anton Protopopov
+// Copyright (c) 2021 Red Hat, Inc.
+#ifndef __TCPRETRANS_H
+#define __TCPRETRANS_H
+
+#define MAX_ENTRIES 8192
```
chenhengqi, line 7: This could be defined in .bpf.c.

libbpf-tools/tcpretrans.bpf.c
```diff
@@ -0,0 +1,147 @@
...
+#include <bpf/bpf_tracing.h>
+
+#include "maps.bpf.h"
+#include "tcpretrans.h"
+
+/* Define here, because there are conflicts with include files */
+#define AF_INET		2
+#define AF_INET6	10
+
+SEC(".rodata") const volatile bool do_count = 0;
```
chenhengqi, line 20: `SEC(".rodata")` is not necessary, right ?

libbpf-tools/tcpretrans.bpf.c
```diff
@@ -0,0 +1,147 @@
...
+	__uint(map_flags, BPF_F_NO_PREALLOC);
+} ipv4_count SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_ENTRIES);
+	__type(key, struct ipv6_flow_key);
+	__type(value, u64);
+	__uint(map_flags, BPF_F_NO_PREALLOC);
+} ipv6_count SEC(".maps");
```
chenhengqi, line 42: We can save a map here.

libbpf-tools/tcpretrans.bpf.c
```diff
@@ -0,0 +1,147 @@
...
+
+static  __always_inline void count_v4(const struct sock *skp)
```
chenhengqi, line 44: `__always_inline` is not required.

libbpf-tools/tcpretrans.bpf.c
```diff
@@ -0,0 +1,147 @@
...
+				   __sk_common.skc_v6_daddr.in6_u.u6_addr32);
+	}
+	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU,
+				  &e, sizeof(e));
+	return 0;
+}
+
+SEC("tp/tcp/tcp_retransmit_skb")
+int tracepoint__tcp__tcp_retransmit_skb(struct trace_event_raw_tcp_event_sk_skb* ctx) {
+	const struct sock *skp;
```
chenhengqi, line 129: Should be a blank line below.

libbpf-tools/tcpretrans.bpf.c
```diff
@@ -0,0 +1,147 @@
...
+int BPF_KPROBE(tcp_send_loss_probe, struct sock *sk)
+{
+	return trace_event(ctx, sk, TLP);
+}
+
+SEC("kprobe/tcp_retransmit_skb")
+int BPF_KPROBE(tcp_retransmit_skb, struct sock *sk)
+{
+	trace_event(ctx, sk, RETRANSMIT);
+	return 0;
```
chenhengqi, line 144: `return trace_event(ctx, sk, RETRANSMIT);` ?

libbpf-tools/tcpretrans.bpf.c
```diff
@@ -0,0 +1,147 @@
...
+	pid_tgid = bpf_get_current_pid_tgid();
+	pid = pid_tgid >> 32;
+	e.pid = pid;
+
+	BPF_CORE_READ_INTO(&dport, skp, __sk_common.skc_dport);
+	e.dport = dport;
+	BPF_CORE_READ_INTO(&sport, skp, __sk_common.skc_num);
+	e.sport = sport;
+	state = BPF_CORE_READ(skp, __sk_common.skc_state);
+	e.state = state;
```
chenhengqi, line 111: Though no big deal, I think you can save many of these local variables.

libbpf-tools/tcpretrans.c
```diff
@@ -0,0 +1,354 @@
...
+
+const char *argp_program_version = "tcpretrans 0.1";
+const char *argp_program_bug_address =
+	"https://github.com/iovisor/bcc/tree/master/libbpf-tools";
+static const char argp_program_doc[] =
+	"\ntcpretrans: Trace TCP retransmits\n"
+	"\n"
+	"EXAMPLES:\n"
+	"    tcpretrans             # display all TCP retransmissions\n"
+	// TODO "    tcpconnect -c          # count occurred retransmits per flow\n"
```
chenhengqi, line 31: This `TODO` can be removed.

libbpf-tools/tcpretrans.h
michaelgugino: I considered this but I would prefer not to.  The existing tcpretrans from BCC prints all ipv4 then ipv6, having two maps makes that a little easier.

libbpf-tools/tcpretrans.h
michaelgugino: MAX_ENTRIES is also used in tcpretrans.c

libbpf-tools/tcpretrans.h
michaelgugino: I don't understand, checkout bindsnoop.h, no padding for structs there either.

libbpf-tools/tcpretrans.bpf.c
michaelgugino: e.state complains about type casting, I'll remove the port variables, they are unneeded.

libbpf-tools/tcpretrans.c
```diff
@@ -0,0 +1,359 @@
...
+const char *argp_program_version = "tcpretrans 0.1";
+const char *argp_program_bug_address =
+	"https://github.com/iovisor/bcc/tree/master/libbpf-tools";
+static const char argp_program_doc[] =
+	"\ntcpretrans: Trace TCP retransmits\n"
+	"\n"
+	"EXAMPLES:\n"
+	"    tcpretrans             # display all TCP retransmissions\n"
+	"    tcpconnect -c          # count occurred retransmits per flow\n"
+	"    tcpconnect -l      	# include tail loss probe attempts\n"
```
chenhengqi, line 37: tcpretrans ?

libbpf-tools/tcpretrans.h
chenhengqi: I would still suggest that using one map and sort the result in userspace.
This can save a map in BPF side for efficiency and merge two print functions into one in userspace.
WDYT ?

michaelgugino: Are maps really expensive?  We would need to double the ipv6 map's entries, and that would reserve more memory (is memory reserved for map entries?), my math says it will increase total reservation by 458752 bytes.

In any case, I think it's helpful to have the code structure similar to tcpconnect.c and tcpretrans.py.  Unless there is a compelling reason to change the structs, I propose we keep them the same.

libbpf-tools/tcpretrans.c
```diff
@@ -0,0 +1,359 @@
...
+	return vfprintf(stderr, format, args);
+}
+
+static void print_count_ipv4(int map_fd)
+{
+	static struct ipv4_flow_key keys[MAX_ENTRIES];
+	__u32 value_size = sizeof(__u64);
+	__u32 key_size = sizeof(keys[0]);
+	static struct ipv4_flow_key zero;
+	static __u64 counts[MAX_ENTRIES];
```
anakryiko, line 118: you are mixing static and non-static variables here, that's quite confusing to read

libbpf-tools/tcpretrans.c
```diff
@@ -0,0 +1,359 @@
...
+	char s[INET_ADDRSTRLEN];
+	char d[INET_ADDRSTRLEN];
+	__u32 i, n = MAX_ENTRIES;
+	struct in_addr src;
+	struct in_addr dst;
+	char *ep_fmt = "[%s]#%d";
```
anakryiko, line 124: const char?

libbpf-tools/tcpretrans.c
```diff
@@ -0,0 +1,359 @@
...
+	if (dump_hash(map_fd, keys, key_size, counts, value_size, &n, &zero)) {
+		warn("dump_hash: %s", strerror(errno));
+		return;
+	}
+
+	for (i = 0; i < n; i++) {
+		src.s_addr = keys[i].saddr;
+		dst.s_addr = keys[i].daddr;
+		sprintf(local, ep_fmt, inet_ntop(AF_INET, &src, s, sizeof(s)), keys[i].sport);
+		sprintf(remote, ep_fmt, inet_ntop(AF_INET, &dst, d, sizeof(d)), ntohs(keys[i].dport));
```
anakryiko, line 137: nit: trying to do too much in one line, inet_ntop() call can  be moved out into local variable to keep everything simpler

libbpf-tools/tcpretrans.c
```diff
@@ -0,0 +1,359 @@
...
+		pause();
+
+	printf("\n%-25s %-25s %-10s\n", "LADDR:LPORT", "RADDR:RPORT", "RETRANSMITS");
+	print_count_ipv4(map_fd_ipv4);
+	print_count_ipv6(map_fd_ipv6);
+}
+
+static void print_events_header()
+{
+	printf("%-8s %-6s %-2s %-20s %1s> %-20s %-4s\n", "TIME", "PID", "IP",
```
anakryiko, line 184: The output is slightly misaligned on my machine. Original tcpretrans.py has the same problem, so it might be my local values. But feels like it should be improved a bit, WDYT?

```
$ sudo ./tcpretrans
Tracing retransmits ... Hit Ctrl-C to end
TIME     PID    IP LADDR:LPORT          T> RADDR:RPORT          STATE
17:06:22 2142809 6  ::1:64056            R> ::1:8297             ESTABLISHED
17:06:22 2142809 6  ::1:64058            R> ::1:8297             ESTABLISHED
17:06:22 2142809 6  ::1:64060            R> ::1:8297             ESTABLISHED
17:06:22 2894834 6  ::1:37958            R> ::1:4331             ESTABLISHED
17:06:23 2142809 6  ::1:38622            R> ::1:16262            ESTABLISHED
17:06:23 2142809 6  ::1:38624            R> ::1:16262            ESTABLISHED
```

```
$ sudo ./tcpretrans -c
Tracing retransmits ... Hit Ctrl-C to end
^C
LADDR:LPORT               RADDR:RPORT               RETRANSMITS
[::1]#36560          <-> [::1]#9176                    1
[::1]#41082          <-> [::1]#9099                    1
[::1]#50392          <-> [::1]#3201                    1
[::1]#36744          <-> [::1]#9176                    1
```

libbpf-tools/tcpretrans.c
```diff
@@ -0,0 +1,359 @@
...
+	sprintf(remote, "%s:%d", inet_ntop(e->af, &d, dst, sizeof(dst)), ntohs(e->dport));
+
+	printf("%-8s %-6d %-2d %-20s %1s> %-20s %s\n",
+		   ts,
+		   e->pid,
+		   e->af == AF_INET ? 4 : 6,
+		   local,
+		   e->type == RETRANSMIT ? "R" : "L",
+		   remote,
+		   TCPSTATE[e->state - 1]);
```
anakryiko, line 227: if e->state somehow becomes 0 (e.g., failed read or something), you'll be reading way outside of array. Maybe let's add entry 0 in TCPSTATE like "<INVALID>"? And then also check that e->state < ARRAY_SIZE(TCPSTATE)?

libbpf-tools/tcpretrans.c
```diff
@@ -0,0 +1,359 @@
...
+	}
+
+	obj = tcpretrans_bpf__open();
+	if (!obj) {
+		warn("failed to open BPF object\n");
+		return 1;
+	}
+
+	// bpf will load non-existant trace points but fail at the attach stage, so
+	// check to ensure our tp exists before we load it.
```
anakryiko, line 298: please don't use C++-style comments

libbpf-tools/tcpretrans.c
```diff
@@ -0,0 +1,359 @@
...
+		}
+	}
+
+	if (!env.lossprobe) {
+		prog = bpf_object__find_program_by_name(obj->obj, "tcp_send_loss_probe");
+		err = bpf_program__set_autoload(prog, false);
+		if (err) {
+			warn("Unable to set autoload for tcp_send_loss_probe\n");
+			return err;
+		}
```
anakryiko, line 325: you have BPF skeleton, why are you using generic APIs to look up by string names?...

libbpf-tools/tcpretrans.c
```diff
@@ -0,0 +1,359 @@
...
+
+	err = tcpretrans_bpf__attach(obj);
+	if (err) {
+		warn("failed to attach BPF programs: %s\n", strerror(-err));
+		goto cleanup;
+	}
+
+	if (signal(SIGINT, sig_int) == SIG_ERR || signal(SIGTERM, sig_int) == SIG_ERR) {
+		warn("can't set signal handler: %s\n", strerror(-errno));
+		goto cleanup;
```
anakryiko, line 345: set err here

libbpf-tools/tcpretrans.bpf.c
```diff
@@ -0,0 +1,150 @@
...
+		BPF_CORE_READ_INTO(&e.daddr, skp,
+				   __sk_common.skc_v6_daddr.in6_u.u6_addr32);
+	}
+	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU,
+				  &e, sizeof(e));
+	return 0;
+}
+
+SEC("tp/tcp/tcp_retransmit_skb")
+int tracepoint__tcp__tcp_retransmit_skb(struct trace_event_raw_tcp_event_sk_skb* ctx)
```
anakryiko, line 130: this is super long and inconvenient name, it will also get truncated to generic "tracepoint_" or something like that, if you dump maps with bpftool. We still have few original tools using such BCC-style long names (where it's actually part of defining which tracepoint to connect too), but generally we've moved to shorter names, like  just "tcp_retransmit_skb" would be totally appropriate in this case.

anakryiko (changes requested):

anakryiko:
btw, seems like your PR description is out of date, you do have kprobe fallback, no?

libbpf-tools/tcpretrans.bpf.c
```diff
@@ -0,0 +1,150 @@
...
+	__uint(map_flags, BPF_F_NO_PREALLOC);
+} ipv6_count SEC(".maps");
+
+static void count_v4(const struct sock *skp)
+{
+	struct ipv4_flow_key key = {};
+	static __u64 zero;
+	__u64 *val;
+
+	BPF_CORE_READ_INTO(&key.saddr, skp, __sk_common.skc_rcv_saddr);
```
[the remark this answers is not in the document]
anakryiko: ah, makes sense

anakryiko (approved):
Github doesn't allow me to reply to your comment...

> but I can't imagine the kernel gives us a garbage value that is larger than the array.

It's highly unlikely, but if kernel ever adds another enum and someone runs old version of this tool, they'll have value bigger than they expected. It's probably unlikely to happen, so I'm ok if we don't touch it right now, but it's something to always keep in mind with tracing: you can never be 100% sure that value you are getting is in the expected range.


The rest looks good to me.

michaelgugino:
@yonghong-song @brendangregg PTAL.

libbpf-tools/tcpretrans.c
```diff
@@ -0,0 +1,364 @@
...
+	}
+
+	err = tcpretrans_bpf__attach(obj);
+	if (err) {
+		warn("failed to attach BPF programs: %s\n", strerror(-err));
+		goto cleanup;
+	}
+
+	if (signal(SIGINT, sig_handler) == SIG_ERR || signal(SIGTERM, sig_handler) == SIG_ERR) {
+		warn("can't set signal handler: %s\n", strerror(-errno));
```
mauriciovasquezbernal, line 348:
```suggestion
		warn("can't set signal handler: %s\n", strerror(errno));
```

seizethedave:
Something we can do to get this tool merged? It seems to be old enough now that it no longer builds.

chenhengqi:
ping @michaelgugino

michaelgugino:
@chenhengqi I will try to make some updates to the patch later this week.

PR #3540, opened 2021-07-16
tools: Fix filtering by mount namespace

mauriciovasquezbernal:
The filtering by mount namespace implementation relies on the
redefinition of the "struct mnt_namespace" internal kernel structure.
The layout of this structure changed in Linux 5.11 (https://github.com/torvalds/linux/commit/1a7b8969e664d6af328f00fe6eb7aabd61a71d13),
this commit adds a conditional on the kernel version to adapt to this
change.

/cc @alban

### DIFF
--- a/src/python/bcc/containers.py
+++ b/src/python/bcc/containers.py
@@ -51,7 +51,10 @@ def _mntns_filter_func_writer(mntnsmap):
     * more likely to change.
     */
     struct mnt_namespace {
+    // This field was removed in https://github.com/torvalds/linux/commit/1a7b8969e664d6af328f00fe6eb7aabd61a71d13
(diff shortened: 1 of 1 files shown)


danobi (approved):
Version number comparison sucks b/c distro could backport. But I don't see any other way other than using BTF (which I'm not sure bcc tools support).

yonghong-song:
Agree. In general, we don't like kernel versions as this may complicate backport. In this particular case, there is no easy checking existence of kernel functions/symbols to differentiate whether the patch is applied to kernel or not. So using kernel version here is OK for this case.

yonghong-song:
[buildbot, test this please]

mauriciovasquezbernal:
I agree as well. The only real solution for this is to use BTF, however this change is specific for BCC tools without BTF support and it's the "best-effort" we can do in that case.

PR #3541, opened 2021-07-17
libbpf-tools: add filetop

chenhengqi:
```shell
14:24:14 loadavg: 1.31 1.06 1.17 1/1369 837906

TID     COMM             READS  WRITES R_Kb    W_Kb    T FILE
1197    gnome-shell      1      0      0       0       - [timerfd]
811093  Timer            0      1      0       0       -
```

### DIFF
--- a/libbpf-tools/filetop.bpf.c
+++ b/libbpf-tools/filetop.bpf.c
@@ -1,0 +1,91 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+/* Copyright (c) 2021 Hengqi Chen */
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_tracing.h>
+#include "filetop.h"
+#include "stat.h"
+
+#define MAX_ENTRIES	10240
+
+const volatile pid_t target_pid = 0;
+const volatile bool regular_file_only = true;
+static struct file_stat zero_value = {};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_ENTRIES);
+	__type(key, struct file_id);
+	__type(value, struct file_stat);
+} entries SEC(".maps");
+
+static void get_file_path(struct file *file, char *buf, size_t size)
+{
+	struct qstr dname;
+
+	dname = BPF_CORE_READ(file, f_path.dentry, d_name);
+	bpf_probe_read_kernel(buf, size, dname.name);
+}
+
+static int probe_entry(struct pt_regs *ctx, struct file *file, size_t count, enum op op)
+{
+	__u64 pid_tgid = bpf_get_current_pid_tgid();
+	__u32 pid = pid_tgid >> 32;
+	__u32 tid = (__u32)pid_tgid;
+	int mode;
+	struct file_id key = {};
+	struct file_stat *valuep;
+
+	if (target_pid && target_pid != pid)
+		return 0;
+
+	mode = BPF_CORE_READ(file, f_inode, i_mode);
+	if (regular_file_only && !S_ISREG(mode))
+		return 0;
+
+	key.dev = BPF_CORE_READ(file, f_inode, i_rdev);
+	key.inode = BPF_CORE_READ(file, f_inode, i_ino);
+	key.pid = pid;
+	key.tid = tid;
+	valuep = bpf_map_lookup_elem(&entries, &key);
+	if (!valuep) {
+		bpf_map_update_elem(&entries, &key, &zero_value, BPF_ANY);
+		valuep = bpf_map_lookup_elem(&entries, &key);
+		if (!valuep)
+			return 0;
+		valuep->pid = pid;
+		valuep->tid = tid;
+		bpf_get_current_comm(&valuep->comm, sizeof(valuep->comm));
+		get_file_path(file, valuep->filename, sizeof(valuep->filename));
+		if (S_ISREG(mode)) {
+			valuep->type = 'R';
+		} else if (S_ISSOCK(mode)) {
+			valuep->type = 'S';
+		} else {
+			valuep->type = 'O';
+		}
+	}
+	if (op == READ) {
+		valuep->reads++;
+		valuep->read_bytes += count;
+	} else {	/* op == WRITE */
+		valuep->writes++;
+		valuep->write_bytes += count;
+	}
+	return 0;
+};
+
+SEC("kprobe/vfs_read")
+int BPF_KPROBE(vfs_read_entry, struct file *file, char *buf, size_t count, loff_t *pos)
+{
+	return probe_entry(ctx, file, count, READ);
+}
+
+SEC("kprobe/vfs_write")
+int BPF_KPROBE(vfs_write_entry, struct file *file, const char *buf, size_t count, loff_t *pos)
+{
+	return probe_entry(ctx, file, count, WRITE);
+}
+
+char LICENSE[] SEC("license") = "Dual BSD/GPL";
--- a/libbpf-tools/filetop.c
+++ b/libbpf-tools/filetop.c
@@ -1,0 +1,313 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+
+/*
+ * filetop Trace file reads/writes by process.
+ * Copyright (c) 2021 Hengqi Chen
+ *
+ * Based on filetop(8) from BCC by Brendan Gregg.
+ * 17-Jul-2021   Hengqi Chen   Created this.
+ */
+#include <argp.h>
+#include <errno.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+#include <unistd.h>
+
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+#include "filetop.h"
+#include "filetop.skel.h"
+#include "trace_helpers.h"
+
+#define warn(...) fprintf(stderr, __VA_ARGS__)
+#define OUTPUT_ROWS_LIMIT 10240
+
+enum SORT {
+	ALL,
+	READS,
+	WRITES,
+	RBYTES,
+	WBYTES,
+};
+
+static volatile sig_atomic_t exiting = 0;
+
+static pid_t target_pid = 0;
+static bool clear_screen = true;
+static bool regular_file_only = true;
+static int output_rows = 20;
+static int sort_by = ALL;
+static int interval = 1;
+static int count = 99999999;
+
+const char *argp_program_version = "filetop 0.1";
+const char *argp_program_bug_address =
+	"https://github.com/iovisor/bcc/tree/master/libbpf-tools";
+const char argp_program_doc[] =
+"Trace file reads/writes by process.\n"
+"\n"
+"USAGE: filetop [-h] [-p PID] [interval] [count]\n"
+"\n"
+"EXAMPLES:\n"
+"    filetop            # file I/O top, refresh every 1s\n"
+"    filetop -p 1216    # only trace PID 1216\n"
+"    filetop 5 10       # 5s summaries, 10 times\n";
+
+static const struct argp_option opts[] = {
+	{ "pid", 'p', "PID", 0, "Process ID to trace" },
+	{ "noclear", 'C', NULL, 0, "Don't clear the screen" },
+	{ "all", 'a', NULL, 0, "Include special files" },
+	{ "sort", 's', "SORT", 0, "Sort columns, default all [all, reads, writes, rbytes, wbytes]" },
+	{ "rows", 'r', "ROWS", 0, "Maximum rows to print, default 20" },
+	{ NULL, 'h', NULL, OPTION_HIDDEN, "Show the full help" },
+	{},
+};
+
+static error_t parse_arg(int key, char *arg, struct argp_state *state)
+{
+	long pid, rows;
+	static int pos_args;
+
+	switch (key) {
+	case 'p':
+		errno = 0;
+		pid = strtol(arg, NULL, 10);
+		if (errno || pid <= 0) {
+			warn("invalid PID: %s\n", arg);
+			argp_usage(state);
+		}
+		target_pid = pid;
+		break;
+	case 'C':
+		clear_screen = false;
+		break;
+	case 'a':
+		regular_file_only = false;
+		break;
+	case 's':
+		if (!strcmp(arg, "all")) {
+			sort_by = ALL;
+		} else if (!strcmp(arg, "reads")) {
+			sort_by = READS;
+		} else if (!strcmp(arg, "writes")) {
+			sort_by = WRITES;
+		} else if (!strcmp(arg, "rbytes")) {
+			sort_by = RBYTES;
+		} else if (!strcmp(arg, "wbytes")) {
+			sort_by = WBYTES;
+		} else {
+			warn("invalid sort method: %s\n", arg);
+			argp_usage(state);
+		}
+		break;
+	case 'r':
+		errno = 0;
+		rows = strtol(arg, NULL, 10);
+		if (errno || rows <= 0) {
+			warn("invalid rows: %s\n", arg);
+			argp_usage(state);
+		}
+		output_rows = rows;
+		if (output_rows > OUTPUT_ROWS_LIMIT)
+			output_rows = OUTPUT_ROWS_LIMIT;
+		break;
+	case 'h':
+		argp_state_help(state, stderr, ARGP_HELP_STD_HELP);
+		break;
+	case ARGP_KEY_ARG:
+		errno = 0;
+		if (pos_args == 0) {
+			interval = strtol(arg, NULL, 10);
+			if (errno || interval <= 0) {
+				warn("invalid interval\n");
+				argp_usage(state);
+			}
+		} else if (pos_args == 1) {
+			count = strtol(arg, NULL, 10);
+			if (errno || count <= 0) {
+				warn("invalid count\n");
+				argp_usage(state);
+			}
+		} else {
+			warn("unrecognized positional argument: %s\n", arg);
+			argp_usage(state);
+		}
+		pos_args++;
+		break;
+	default:
+		return ARGP_ERR_UNKNOWN;
+	}
+	return 0;
+}
+
+static void sig_int(int signo)
+{
+	exiting = 1;
+}
+
+static int sort_column(const void *obj1, const void *obj2)
+{
+	struct file_stat *s1 = (struct file_stat *)obj1;
+	struct file_stat *s2 = (struct file_stat *)obj2;
+
+	if (sort_by == READS) {
+		return s2->reads - s1->reads;
+	} else if (sort_by == WRITES) {
+		return s2->writes - s1->writes;
+	} else if (sort_by == RBYTES) {
+		return s2->read_bytes - s1->read_bytes;
+	} else if (sort_by == WBYTES) {
+		return s2->write_bytes - s1->write_bytes;
+	} else {
+		return (s2->reads + s2->writes + s2->read_bytes + s2->write_bytes)
+		     - (s1->reads + s1->writes + s1->read_bytes + s1->write_bytes);
+	}
+}
+
+static int print_stat(struct filetop_bpf *obj)
+{
+	FILE *f;
+	time_t t;
+	struct tm *tm;
+	char ts[16], buf[256];
+	struct file_id key, *prev_key = NULL;
+	static struct file_stat values[OUTPUT_ROWS_LIMIT];
+	int n, i, err = 0, rows = 0;
+	int fd = bpf_map__fd(obj->maps.entries);
+
+	f = fopen("/proc/loadavg", "r");
+	if (f) {
+		time(&t);
+		tm = localtime(&t);
+		strftime(ts, sizeof(ts), "%H:%M:%S", tm);
+		memset(buf, 0 , sizeof(buf));
+		n = fread(buf, 1, sizeof(buf), f);
+		if (n)
+			printf("%8s loadavg: %s\n", ts, buf);
+		fclose(f);
+	}
+
+	printf("%-7s %-16s %-6s %-6s %-7s %-7s %1s %s\n",
+	       "TID", "COMM", "READS", "WRITES", "R_Kb", "W_Kb", "T", "FILE");
+
+	while (1) {
+		err = bpf_map_get_next_key(fd, prev_key, &key);
+		if (err) {
+			if (errno == ENOENT) {
+				err = 0;
+				break;
+			}
+			warn("bpf_map_get_next_key failed: %s\n", strerror(errno));
+			return err;
+		}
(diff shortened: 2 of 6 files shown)


chenhengqi:
When I implement this tool.  Two problems occurred to me.

* Why there is no `bpf_map_clear` helper provided like in C++ `std::map::clear()` does ?  Several libbpf tools use `bpf_map_get_next_key` and `bpf_map_delete_elem` in a loop to achieve this purpose. I am wondering about the decide decision behind this.
* I tried adding `vfs_readv`/`vfs_writev` to this tool, but can't figure out a way to convince the verifier about the arguments access.(i.e. sum all iovec.iov_len)
```c
SEC("kprobe/vfs_readv")
+int BPF_KPROBE(vfs_readv_entry, struct file *file, const struct iovec *vec, unsigned long vlen, loff_t *pos, rwf_t flags)
```

davemarchevsky:
Did some functional comparison with the python tool, looks like this is matching the python tool's functionality pretty closely!

Only thing I noticed was that the sort order for the python tool appears to be descending by default, while this tool's is ascending:

Python tool output (`-s writes -C`):
```
TID    COMM             READS  WRITES R_Kb    W_Kb    T FILE
1388555 bsdacctd         0      34     0       2       R bsdacctdfile
1089657 rs:main Q:Reg    0      31     0       8       R proxy.log
1089657 rs:main Q:Reg    0      19     0       3       R cppagentd.log
2156639 exec_info_logge  0      17     0       12      R 643633.log
2785604 CanaryMgr        0      10     0       0       R 339209.log
1      systemd          0      5      0       0       R cgroup.subtree_control
1089657 rs:main Q:Reg    0      3      0       0       R secure
1089657 rs:main Q:Reg    0      3      0       0       R messages
1089657 rs:main Q:Reg    0      2      0       0       R serverd.log
844    systemd-journal  0      1      0       0       R .#9:2001486305Qg7fa3
1      systemd          0      1      0       0       R cpuset.mems
1089657 rs:main Q:Reg    0      1      0       0       R proxy2.log
1      systemd          0      1      0       0       R cgroup.procs
1      systemd          0      1      0       0       R io.weight
3359031 dd               0      1      0       4       R writable_root.check
1      systemd          0      1      0       0       R memory.high
1      systemd          0      1      0       0       R cpuset.cpus
1      systemd          0      1      0       0       R cpu.weight
1      systemd          0      1      0       0       R memory.min
1      systemd          0      1      0       0       R memory.max
```

this tool's output:
```
TID     COMM             READS  WRITES R_Kb    W_Kb    T FILE
176615  log          3      0      64      0       R stat
176615  log          1      0      7       0       R status
176615  log          3      0      64      0       R stat
176615  log          3      0      64      0       R stat
3359980 expr             3      0      0       0       R libgmp.so.10.3.2
3360027 sh               3      0      0       0       R timeout
176615  log          2      0      4       0       R environ
176615  log          1      0      7       0       R status
2879739 oomd      2      0      8       0       R memory.current
3359986 sh               2      0      0       0       R libc-2.28.so
3360029 awk              1      0      0       0       R libdl-2.28.so
3359921 netedit_collect  3      0      0       0       R netedit_collector.sh
3360020 timeout          1      0      0       0       R librt-2.28.so
176615  log          2      0      15      0       R status
3359918 timeout          1      0      0       0       R libpthread-2.28.so
2879739 oomd      2      0      8       0       R memory.current
176615  log          3      0      64      0       R stat
3359963 python3.8        2      0      8       0       R secrets_tool
176615  log          3      0      64      0       R stat
3359963 python3.8        3      0      8       0       R secrets_tool

```

Can we do desc-by-default here as well?

Doing a pass on the code itself now

libbpf-tools/filetop.h
```diff
@@ -0,0 +1,31 @@
...
+
+#define PATH_MAX	4096
+#define TASK_COMM_LEN	16
+
+enum op {
+	READ,
+	WRITE,
+};
+
+struct file_id {
```
davemarchevsky, line 13: so the key -> val mapping of the python program was:
`(pid, name_len, comm, name, type) -> (reads, writes, rbytes, wbytes)`

while this tool's is:
`(file, pid, tid) -> (reads, rbytes, writes, wbytes, pid, tid, filename, comm, type)`
where file is a `struct file*`

I think the new way is better but wanted to run through some edge cases and ask questions to make sure I'm understanding correctly:
* If a file is closed and re-opened, do we expect to see a different `struct file *`? Similarly, is there any chance we get the same `struct file *` if we close file A and open file B shortly after?
  * Latter hypothetical is more concerning to me as we'd be attributing io to wrong filename
* Why do `pid, tid` exist in both key and value?
* Was the purpose of this change to avoid reading `comm` and `filename` unnecessarily?

libbpf-tools/filetop.bpf.c
```diff
@@ -0,0 +1,110 @@
...
+/* Copyright (c) 2021 Hengqi Chen */
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_tracing.h>
+#include "filetop.h"
+
+#define MAX_ENTRIES	10240
+
+#define S_IFMT		00170000
```
davemarchevsky, line 11: Looks like these defines are coming from `include/uapi/linux/stat.h`. Since they're not making it into `vmlinux.h`, and other tools might want to use these same defs, can they be pulled out into a separate file? Similarly to `libbpf-tools/blk_types.h`

davemarchevsky (changes requested):

chenhengqi:
> Did some functional comparison with the python tool, looks like this is matching the python tool's functionality pretty closely!
> 
> Only thing I noticed was that the sort order for the python tool appears to be descending by default, while this tool's is ascending:
> 
> Python tool output (`-s writes -C`):
> 
> ```
> TID    COMM             READS  WRITES R_Kb    W_Kb    T FILE
> 1388555 bsdacctd         0      34     0       2       R bsdacctdfile
> 1089657 rs:main Q:Reg    0      31     0       8       R proxy.log
> 1089657 rs:main Q:Reg    0      19     0       3       R cppagentd.log
> 2156639 exec_info_logge  0      17     0       12      R 643633.log
> 2785604 CanaryMgr        0      10     0       0       R 339209.log
> 1      systemd          0      5      0       0       R cgroup.subtree_control
> 1089657 rs:main Q:Reg    0      3      0       0       R secure
> 1089657 rs:main Q:Reg    0      3      0       0       R messages
> 1089657 rs:main Q:Reg    0      2      0       0       R serverd.log
> 844    systemd-journal  0      1      0       0       R .#9:2001486305Qg7fa3
> 1      systemd          0      1      0       0       R cpuset.mems
> 1089657 rs:main Q:Reg    0      1      0       0       R proxy2.log
> 1      systemd          0      1      0       0       R cgroup.procs
> 1      systemd          0      1      0       0       R io.weight
> 3359031 dd               0      1      0       4       R writable_root.check
> 1      systemd          0      1      0       0       R memory.high
> 1      systemd          0      1      0       0       R cpuset.cpus
> 1      systemd          0      1      0       0       R cpu.weight
> 1      systemd          0      1      0       0       R memory.min
> 1      systemd          0      1      0       0       R memory.max
> ```
> 
> this tool's output:
> 
> ```
> TID     COMM             READS  WRITES R_Kb    W_Kb    T FILE
> 176615  log          3      0      64      0       R stat
> 176615  log          1      0      7       0       R status
> 176615  log          3      0      64      0       R stat
> 176615  log          3      0      64      0       R stat
> 3359980 expr             3      0      0       0       R libgmp.so.10.3.2
> 3360027 sh               3      0      0       0       R timeout
> 176615  log          2      0      4       0       R environ
> 176615  log          1      0      7       0       R status
> 2879739 oomd      2      0      8       0       R memory.current
> 3359986 sh               2      0      0       0       R libc-2.28.so
> 3360029 awk              1      0      0       0       R libdl-2.28.so
> 3359921 netedit_collect  3      0      0       0       R netedit_collector.sh
> 3360020 timeout          1      0      0       0       R librt-2.28.so
> 176615  log          2      0      15      0       R status
> 3359918 timeout          1      0      0       0       R libpthread-2.28.so
> 2879739 oomd      2      0      8       0       R memory.current
> 176615  log          3      0      64      0       R stat
> 3359963 python3.8        2      0      8       0       R secrets_tool
> 176615  log          3      0      64      0       R stat
> 3359963 python3.8        3      0      8       0       R secrets_tool
> ```
> 
> Can we do desc-by-default here as well?
> 
> Doing a pass on the code itself now

I can't reproduce this. The output order should be the same as the python tool.

libbpf-tools/filetop.c
```diff
@@ -0,0 +1,319 @@
...
+	}
+	return 0;
+}
+
+static void sig_int(int signo)
+{
+	exiting = 1;
+}
+
+static int sort_column(const void *obj1, const void *obj2)
```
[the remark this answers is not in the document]
chenhengqi: This descending order thing is handled at output stage.

libbpf-tools/filetop.bpf.c
chenhengqi: Look like the fileslower tool also requires this, let us put them in a separate file.

libbpf-tools/filetop.h
chenhengqi: > * If a file is closed and re-opened, do we expect to see a different struct file *? Similarly, is there any chance we get the same struct file * if we close file A and open file B shortly after?
>     *  Latter hypothetical is more concerning to me as we'd be attributing io to wrong filename

I can't answer the first question. :(

But since the original tool only use the filename (without full file path), I think that is more unreliable.
Say, the same process writes to `/foo/bar/xxx` and `/fuz/baz/xxx` would be added up to the same key.

> * Why do pid, tid exist in both key and value?

pid and tid exist in value for userspace output, since we sort and output the values only.

> * Was the purpose of this change to avoid reading comm and filename unnecessarily?

Yes, I think this is better from the perspective of performance.

libbpf-tools/filetop.c
```diff
@@ -0,0 +1,319 @@
...
+		err = bpf_map_lookup_elem(fd, &key, &values[rows++]);
+		if (err) {
+			warn("bpf_map_lookup_elem failed: %s\n", strerror(errno));
+			return err;
+		}
+		prev_key = &key;
+	}
+
+	qsort(values, rows, sizeof(struct file_stat), sort_column);
+	rows = rows < output_rows ? rows : output_rows;
```
davemarchevsky, line 220: I see what you mean by "the descending order is handled at the output stage". I think this is where the bug is.

Let's say we have 10k items and the items, after qsort, are arranged in ascending order by some column `[0, 1, 2, ..., 9999]`. Let's also say we only want 20, so this statement becomes `rows = 10000 < 20 ? 10000 : 20` resulting in `rows = 20`. Then the loop on next line will print items `[19, 18, ... 0]` when in reality we wanted it to print items `[9999, 9998, ..., 9980]`.

Guess we were seeing different output due to our test hosts having different # of processes doing fileio, if `rows < output_rows` this will work correctly.

Can do something like:

```
rows_to_print = rows < output_rows ? rows : output_rows;
for (i = rows - 1; i >= (rows - rows_to_print); i--) {
   ...
}
```

chenhengqi: :facepalm:  You are right. Thanks.

Update the sort_column function instead.

davemarchevsky: confirmed that this fixes the issue

libbpf-tools/filetop.h
davemarchevsky: > But since the original tool only use the filename (without full file path), I think that is more unreliable.
Say, the same process writes to /foo/bar/xxx and /fuz/baz/xxx would be added up to the same key.

Ah, thanks for noting that, that's definitely worth improving on.

Unfortunately after digging into where `struct file *`s are coming from I think the scenarios I was worried about can happen. When the kernel [allocates a new `struct file`](https://elixir.bootlin.com/linux/latest/source/fs/file_table.c#L96) it uses memory provided by a [slab allocator](https://elixir.bootlin.com/linux/latest/source/fs/file_table.c#L41) that just serves `struct file`-sized memory. 

So for my two hypotheticals: if a file is closed and re-opened, there's no guarantee of seeing the same `struct file *` or a different one. Similarly for the more concerning one, two different files could have the same `struct file *` if first one is closed before second opened.

Not sure how often this happens in practice, but given that this tool is more likely to be used on systems that are doing a lot of "file stuff" and therefore more likely to trigger these cases, we should find some other identifier.  

The inode's i_ino and device number of the superblock (`inode->i_sb->s_dev`) should do it.

libbpf-tools/filetop.bpf.c
```diff
@@ -0,0 +1,91 @@
...
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_tracing.h>
+#include "filetop.h"
+#include "stat.h"
+
+#define MAX_ENTRIES	10240
+
+const volatile pid_t target_pid = 0;
+const volatile bool full_path = false;
```
davemarchevsky, line 13: This isn't being used anywhere. Same comment applies to `show_full_path` flag setting in the userspace runner.

libbpf-tools/filetop.h
chenhengqi: Thanks.

> The inode's i_ino and device number of the superblock (inode->i_sb->s_dev) should do it.

Will do.

libbpf-tools/filetop.bpf.c
chenhengqi: Will remove this for now and add this option once bpf_d_path helper expose to vfs_*

libbpf-tools/filetop.h
chenhengqi: I read this [page](https://www.oreilly.com/library/view/linux-device-drivers/0596000081/ch03s02.html) which says:

> Every time the kernel calls a device driver, it tells the driver which device is being acted upon. The major and minor numbers are paired in a single data type that the driver uses to identify a particular device. The combined device number (the major and minor numbers concatenated together) resides in the field i_rdev of the inode structure, which we introduce later. Some driver functions receive a pointer to struct inode as the first argument. So if you call the pointer inode (as most driver writers do), the function can extract the device number by looking at inode->i_rdev.

Also, the kernel has two functions `imajor` and `iminor` which extract device number from `inode->i_rdev`.

https://elixir.bootlin.com/linux/latest/source/include/linux/fs.h#L875

I think it's reasonable to just use `inode->i_rdev`, WDYT?

chenhengqi: Updated. :)

davemarchevsky: I think just using the `rdev` directly is fine, no need to pull out major/minor

davemarchevsky (approved):
Revalidated, LGTM!

/cc @yonghong-song

davemarchevsky:
[buildbot, test this please]

yonghong-song:
[buildbot, test this please]

chenhengqi:
> When I implement this tool. Two problems occurred to me.
> 
> * Why there is no `bpf_map_clear` helper provided like in C++ `std::map::clear()` does ?  Several libbpf tools use `bpf_map_get_next_key` and `bpf_map_delete_elem` in a loop to achieve this purpose. I am wondering about the decide decision behind this.
> * I tried adding `vfs_readv`/`vfs_writev` to this tool, but can't figure out a way to convince the verifier about the arguments access.(i.e. sum all iovec.iov_len)
> 
> ```c
> SEC("kprobe/vfs_readv")
> +int BPF_KPROBE(vfs_readv_entry, struct file *file, const struct iovec *vec, unsigned long vlen, loff_t *pos, rwf_t flags)
> ```

@yonghong-song  Can you help answer these two questions ?  Thanks.

yonghong-song:
> > When I implement this tool. Two problems occurred to me.
> > 
> > * Why there is no `bpf_map_clear` helper provided like in C++ `std::map::clear()` does ?  Several libbpf tools use `bpf_map_get_next_key` and `bpf_map_delete_elem` in a loop to achieve this purpose. I am wondering about the decide decision behind this.

bcc map (table) has an API called clear_table_non_atomic() which can be used to clear the table.
Currently it used the following loop like
```
    while (this->first(key.get()))
      if (!this->remove(key.get())) {
        return StatusTuple(-1,
                           "Failed to delete element when clearing table %s",
                           desc.name.c_str());
      }
```
this should work for a seldomly changed map but not working for a frequently-updated map.

The real typical use case should not just clear the map but clear the map with a set of elements,
e.g.,
   1. use batch interface to get all the map elements.
   2. do something with these map elements
   3. remove these elements from the map
or just
   1. use batch interface to get, and the same time delete all the map elements
 this way, you do not need to remove these elements from the map as they have been removed.

> > * I tried adding `vfs_readv`/`vfs_writev` to this tool, but can't figure out a way to convince the verifier about the arguments access.(i.e. sum all iovec.iov_len)

In such cases, you will need a bounded loop as otherwise verifier will assume U64_MAX vlen. You may also need to use llvm12 or latest compiler. Please see virtiostat.py. If you still cannot make verifier happy, please post a reproducible test case here and I will take a look.

> > 
> > ```c
> > SEC("kprobe/vfs_readv")
> > +int BPF_KPROBE(vfs_readv_entry, struct file *file, const struct iovec *vec, unsigned long vlen, loff_t *pos, rwf_t flags)
> > ```
> 
> @yonghong-song Can you help answer these two questions ? Thanks.

chenhengqi:
> > > When I implement this tool. Two problems occurred to me.
> > > 
> > > * Why there is no `bpf_map_clear` helper provided like in C++ `std::map::clear()` does ?  Several libbpf tools use `bpf_map_get_next_key` and `bpf_map_delete_elem` in a loop to achieve this purpose. I am wondering about the decide decision behind this.
> 
> bcc map (table) has an API called clear_table_non_atomic() which can be used to clear the table.
> Currently it used the following loop like
> 
> ```
>     while (this->first(key.get()))
>       if (!this->remove(key.get())) {
>         return StatusTuple(-1,
>                            "Failed to delete element when clearing table %s",
>                            desc.name.c_str());
>       }
> ```
> 
> this should work for a seldomly changed map but not working for a frequently-updated map.
> 
> The real typical use case should not just clear the map but clear the map with a set of elements,
> e.g.,
> 
> 1. use batch interface to get all the map elements.
> 2. do something with these map elements
> 3. remove these elements from the map
>    or just
> 4. use batch interface to get, and the same time delete all the map elements
>    this way, you do not need to remove these elements from the map as they have been removed.
> 
> > > * I tried adding `vfs_readv`/`vfs_writev` to this tool, but can't figure out a way to convince the verifier about the arguments access.(i.e. sum all iovec.iov_len)
> 
> In such cases, you will need a bounded loop as otherwise verifier will assume U64_MAX vlen. You may also need to use llvm12 or latest compiler. Please see virtiostat.py. If you still cannot make verifier happy, please post a reproducible test case here and I will take a look.

```diff
diff --git a/libbpf-tools/filetop.bpf.c b/libbpf-tools/filetop.bpf.c
index c02a2054..377e5218 100644
--- a/libbpf-tools/filetop.bpf.c
+++ b/libbpf-tools/filetop.bpf.c
@@ -88,4 +88,30 @@ int BPF_KPROBE(vfs_write_entry, struct file *file, const char *buf, size_t count
        return probe_entry(ctx, file, count, WRITE);
 }
 
+SEC("kprobe/vfs_readv")
+int BPF_KPROBE(vfs_readv_entry, struct file *file, const struct iovec *vec, unsigned long vlen, loff_t *pos, rwf_t flags)
+{
+       struct iovec *arr = (struct iovec *)vec, *elem;
+       int count = 0, i;
+
+       for (i = 0; i < vlen && i < 0xff; i++) {
+               elem = arr + i;
+               count += BPF_CORE_READ(elem, iov_len);
+       }
+       return probe_entry(ctx, file, count, READ);
+}
+
+SEC("kprobe/vfs_writev")
+int BPF_KPROBE(vfs_writev_entry, struct file *file, const struct iovec *vec, unsigned long vlen, loff_t *pos, rwf_t flags)
+{
+       struct iovec *arr = (struct iovec *)vec, *elem;
+       int count = 0, i;
+
+       for (i = 0; i < vlen && i < 0xff; i++) {
+               elem = arr + i;
+               count += BPF_CORE_READ(elem, iov_len);
+       }
+       return probe_entry(ctx, file, count, WRITE);
+}
+
 char LICENSE[] SEC("license") = "Dual BSD/GPL";
```
This magic works:
```c
for (i = 0; i < vlen && i < 0xff; i++) {
```

While this don't:
```c
int len = vlen & 0xff;
for (i = 0; i < len; i++) {
```

I remembered you also referred this in some issues to workaround verifier bound check, so I tried this and failed.

> 
> > > ```c
> > > SEC("kprobe/vfs_readv")
> > > +int BPF_KPROBE(vfs_readv_entry, struct file *file, const struct iovec *vec, unsigned long vlen, loff_t *pos, rwf_t flags)
> > > ```
> > 
> > 
> > @yonghong-song Can you help answer these two questions ? Thanks.

yonghong-song:
For  below code:
  > int len = vlen & 0xff;
  > for (i = 0; i < len; i++) {
I tried with latest llvm (build from source) and 5.6 kernel, it works fine.
Could you try whether latest llvm works for you or not?

chenhengqi:
@yonghong-song 

```c
int len = vlen & 0xff;

for (i = 0; i < len; i++) {
	elem = arr + i;
	count += BPF_CORE_READ(elem, iov_len);
}
```

This code causes the following error:
```
libbpf: load bpf program failed: Argument list too long
libbpf: -- BEGIN DUMP LOG ---
libbpf: 
Unrecognized arg#0 type PTR
; int BPF_KPROBE(vfs_writev_entry, struct file *file, const struct iovec *vec, unsigned long vlen, loff_t *pos, rwf_t flags)
0: (b7) r2 = 0
; int BPF_KPROBE(vfs_writev_entry, struct file *file, const struct iovec *vec, unsigned long vlen, loff_t *pos, rwf_t flags)
1: (79) r7 = *(u64 *)(r1 +104)
2: (79) r3 = *(u64 *)(r1 +112)
3: (7b) *(u64 *)(r10 -16) = r3
4: (79) r8 = *(u64 *)(r1 +96)
; int len = vlen & 0xff;
5: (57) r8 &= 255
; for (i = 0; i < len; i++) {
6: (15) if r8 == 0x0 goto pc+16
 R1=ctx(id=0,off=0,imm=0) R2_w=invP0 R3_w=invP(id=0) R7_w=invP(id=0) R8_w=invP(id=0,umax_value=255,var_off=(0x0; 0xff)) R10=fp0 fp-16_w=mmmmmmmm
7: (b7) r6 = 0
8: (b7) r9 = 8
; count += BPF_CORE_READ(elem, iov_len);
9: (bf) r3 = r7
10: (0f) r3 += r9
11: (bf) r1 = r10
......
```

yonghong-song:
Okay, the error is "Argument list too long". That typically means the verification exceeds some internal limits. Could you try latest llvm-project trunk? I just fixed a loop exit condition optimization in llvm (https://reviews.llvm.org/D107483)

chenhengqi:
OK. Will try this later.

chenhengqi:
> Okay, the error is "Argument list too long". That typically means the verification exceeds some internal limits. Could you try latest llvm-project trunk? I just fixed a loop exit condition optimization in llvm (https://reviews.llvm.org/D107483)

Yes, the llvm-project trunk works for me. Thanks.

PR #3554, opened 2021-07-28
bcc/tools: use device number and inode number to identify a file

chenhengqi:
Currently, the filetop tool use (tid, filename, type) tuple to
key a file, which is not enough to uniquely identify a file.
A thread write to multi files with the same name would add up to
same value in the map which can be repro by the following command:

$ cat somefile | tee /foo/bar/xxx /fuz/baz/xxx

Let us add device number and inode number to uniquely identify
a file.

### DIFF
--- a/tools/filetop.py
+++ b/tools/filetop.py
@@ -65,6 +65,8 @@ bpf_text = """
 
 // the key for the output summary
 struct info_t {
+    unsigned long inode;
+    dev_t dev;
     u32 pid;
     u32 name_len;
     char comm[TASK_COMM_LEN];
@@ -100,7 +102,11 @@ static int do_entry(struct pt_regs *ctx, struct file *file,
         return 0;
 
     // store counts and sizes by pid & file
-    struct info_t info = {.pid = pid};
+    struct info_t info = {
+        .pid = pid,
+        .inode = file->f_inode->i_ino,
+        .dev = file->f_inode->i_rdev,
+    };
     bpf_get_current_comm(&info.comm, sizeof(info.comm));
     info.name_len = d_name.len;
     bpf_probe_read_kernel(&info.name, sizeof(info.name), d_name.name);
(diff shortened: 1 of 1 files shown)


chenhengqi:
Thanks @davemarchevsky  for pointing this out.

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

anakryiko (commented):
Does this work with btrfs? I remember we had some issues with btrfs due to some inode or dev reporting peculiarities.

davemarchevsky:
@anakryiko I think you're referring to [this kludge](https://github.com/iovisor/bcc/blob/722cf83941879c52ebea5e5a1692b2976de6ad62/src/cc/api/BPF.h#L368-L382) and the various PRs and conversations around it ([1](https://github.com/iovisor/bcc/pull/2324) [2](https://github.com/iovisor/bcc/issues/2715) [3](https://github.com/iovisor/bcc/pull/2723) - just listing for completeness).

TL;DR of the issue is 

> btrfs and overlayfs report different device numbers for files in /proc/PID/maps vs stat syscall

so if we're getting device numbers from different sources and expecting to be able to compare across them it could be a problem e.g. common case of "pull out inode,dev from using BPF and find matching `/proc/PID/maps` entry". Since this tool is just using the device nums from one source it won't cause problem here.

davemarchevsky:
LGTM, test failure only occurs on `fc27` and is unrelated to this PR

anakryiko:
> @anakryiko I think you're referring to [this kludge](https://github.com/iovisor/bcc/blob/722cf83941879c52ebea5e5a1692b2976de6ad62/src/cc/api/BPF.h#L368-L382) and the various PRs and conversations around it ([1](https://github.com/iovisor/bcc/pull/2324) [2](https://github.com/iovisor/bcc/issues/2715) [3](https://github.com/iovisor/bcc/pull/2723) - just listing for completeness).
> 
> TL;DR of the issue is
> 
> > btrfs and overlayfs report different device numbers for files in /proc/PID/maps vs stat syscall
> 
> so if we're getting device numbers from different sources and expecting to be able to compare across them it could be a problem e.g. common case of "pull out inode,dev from using BPF and find matching `/proc/PID/maps` entry". Since this tool is just using the device nums from one source it won't cause problem here.

Ah, I see, thanks for explaining!

yonghong-song:
[buildbot, test this please]

PR #3564, opened 2021-08-05
libbpf-tools: add exitsnoop

chenhengqi:
```shell
$ sudo ./exitsnoop
PCOMM            PID     PPID    TID     AGE(s)  EXIT_CODE
awk              2128720 2128718 2128720 0.01    0
sh               2128718 289432  2128718 0.01    0
zsh              2128721 2126939 2128721 0.00    0
a.out            2128722 2126939 2128722 0.28    signal 11 (Segmentation fault), core dumped
zsh              2128726 2126939 2128726 0.00    0
zsh              2128729 2128728 2128729 0.00    0
wc               2128730 2128728 2128730 0.00    0
zsh              2128728 2128727 2128728 0.01    0
git              2128731 2128727 2128731 0.00    code 1
git              2128732 2128727 2128732 0.00    code 128
```

### DIFF
--- a/libbpf-tools/.gitignore
+++ b/libbpf-tools/.gitignore
@@ -12,6 +12,7 @@
 /cpufreq
 /drsnoop
 /execsnoop
+/exitsnoop
 /ext4dist
 /ext4slower
 /filelife
--- a/libbpf-tools/Makefile
+++ b/libbpf-tools/Makefile
@@ -27,6 +27,7 @@ APPS = \
 	cpufreq \
 	drsnoop \
 	execsnoop \
+	exitsnoop \
 	filelife \
 	filetop \
 	fsdist \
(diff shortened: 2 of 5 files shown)


anakryiko (changes requested):
Some of the options from exitsnoop.py are missing. They don't seem to be too hard to implement. Any reason you omitted them? Please try to keep this tool as close to the original Python version so that it can be a complete drop-in replacement.

chenhengqi:
Yes, it's trivial to implement them, but they are left out intentionally.

* The `label` option adds nothing valuable to tracing output and seems redundant. None of the existing tools has this option. Omit `utc` option for the same reason.
* The `timespec` option is intended for `RHEL 7.6` (with kernel 3.10), which has no support for BTF, I think it's not possible to get it run on a `RHEL 7.6` instance. While on Centos 8 (RHEL 8), this tool works properly.

anakryiko (approved):
Makes sense (though I can see how UTC might be useful, I'd still probably add it). LGTM.

yonghong-song:
[buildbot, test this please]

PR #3565, opened 2021-08-05
Added IPv4/IPv6 filter support for tcp trace tools

hariharan-a:
Added command line optional argument [-4 | -6] similar to ping/traceroute to support filtering IPv4/IPv6 tcp stats. 

+ @brendangregg

### DIFF
--- a/tools/tcpconnlat.py
+++ b/tools/tcpconnlat.py
@@ -41,7 +41,7 @@ examples = """examples:
     ./tcpconnlat -p 181    # only trace PID 181
     ./tcpconnlat -L        # include LPORT while printing outputs
     ./tcpconnlat -4        # trace IPv4 family only
-    ./tcpconnlat -6        # trace IPv6 family ony
+    ./tcpconnlat -6        # trace IPv6 family only
 """
 parser = argparse.ArgumentParser(
     description="Trace TCP connects and show connection latency",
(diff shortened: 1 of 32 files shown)


yonghong-song:
[buildbot, test this please]

tools/tcpconnlat.py
```diff
@@ -40,6 +40,8 @@ def positive_float(val):
     ./tcpconnlat -t        # include timestamps
     ./tcpconnlat -p 181    # only trace PID 181
     ./tcpconnlat -L        # include LPORT while printing outputs
+    ./tcpconnlat -4        # trace IPv4 family only
+    ./tcpconnlat -6        # trace IPv6 family ony
```
yonghong-song, line 44: family only

yonghong-song (commented):
I think adding -4 -6 make sense as a lot of tools already have "-4" "-6" options like ping, ss etc. Could you also add new options to man pages? Thanks!

yonghong-song:
[buildbot, ok to test]

tools/tcpconnlat.py
hariharan-a: Done. Fixed typo and added options to man pages.

brendangregg:
Looks good, thanks, and follows the precedent of -4/-6 from other network tools. Thanks for updating all the man pages and example files.

anakryiko:
@hariharan-a Would you be able to check libbpf-tools and see about adding the same feature there?

hariharan-a:
> @hariharan-a Would you be able to check libbpf-tools and see about adding the same feature there?

Sure. I will check.

PR #3573, opened 2021-08-09
Add libbpf-based tools to container image

mauriciovasquezbernal:
This PR extends the dockerfile to also include the tools based on libbpf.

### DIFF
--- a/Dockerfile.ubuntu
+++ b/Dockerfile.ubuntu
@@ -17,6 +17,14 @@ WORKDIR /root/bcc
 RUN /usr/lib/pbuilder/pbuilder-satisfydepends && \
     ./scripts/build-deb.sh ${BUILD_TYPE}
 
+# libbpf-tools
+RUN apt-get -qq update && \
+    apt-get -y install build-essential libelf-dev libfl-dev clang-10 libcap-dev
+
+RUN cd /root/bcc/src/cc/libbpf/src/ && make
+RUN cd /root/bcc/libbpf-tools/ && \
+  CLANG=clang-10 LLVM_STRIP=llvm-strip-10 make -j $(nproc) DESTDIR=/libbpf-tools/ prefix="" install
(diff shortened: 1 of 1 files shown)


Dockerfile.ubuntu
```diff
@@ -17,6 +17,14 @@ WORKDIR /root/bcc
 RUN /usr/lib/pbuilder/pbuilder-satisfydepends && \
     ./scripts/build-deb.sh ${BUILD_TYPE}
 
+# libbpf-tools
+RUN apt-get -qq update && \
+    apt-get -y install build-essential libelf-dev libfl-dev clang-10 libcap-dev
+
+RUN cd /root/bcc/src/cc/libbpf/src/ && make
+RUN cd /root/bcc/libbpf-tools/ && \
+  CLANG=clang-10 LLVM_STRIP=llvm-strip-10 make -j $(nproc) DESTDIR=/libbpf-tools/ prefix="" install
```
anakryiko, line 26: would it be possible to use a more recent Clang version?

mauriciovasquezbernal: That's the most recent version shipped with bionic :/. We could use clang-12 if we build the libbpf tools using focal. (I think It'll work just fine to have the build image on focal even if the final container is built with bionic). If you think it's better to use a more recent version I'll switch to that approach.

anakryiko: Newer Clang has better code generation for BPF, generally. But whether that is important for current versions of libbpf-tools, I don't really know. I was just curious what's the limitations for using newer Clang. I satisfied my curiosity, thanks!

davemarchevsky:
Sorry that I didn't review this properly :( .

I'm addressing this in a different way in #3779 and future PRs, so we can reach a state where we're building and doing some basic "does this run?" testing of all libbpf-tools.

PR #3576, opened 2021-08-11
Proof of concept: Add offcpu-sampling capability to offcputime.py 

mstange:
This adds a -F <frequency> flag to offcputime.py.

Example usage:

```
$ sudo ./tools/offcputime.py -F 1000 2 # Sample at 1000 Hz for 2 seconds
```

This achieves something like offcpu sampling without actually using
a wall-clock time based timer interrupt:
First, we calculate a sample period from the supplied frequency, in nanoseconds.
Then, whenever a thread wakes up, we accumulate its sleep time into a per-thread field.
Whenever this accumulated sleep time counter overflows a multiple of the sample
period, we sample.
This is an alternative to using --min-block-time: It achieves low overhead
without missing short sleeps.

## Backstory

This proof of concept was born out of confusion after reading https://www.brendangregg.com/offcpuanalysis.html
The page has this warning:

> WARNING: With off-CPU tracing, be aware that scheduler events can be very frequent – in extreme cases, millions of events per second – and although tracers may only add a tiny amount of overhead to each event, due to the event rate that overhead can add up and become significant. Off-CPU sampling has overhead issues too, as systems may have tens of thousands of threads which must be constantly sampled, orders of magnitude higher overhead than CPU sampling across the CPU count only.

There is also a section about "Off-CPU Sampling", which says:

> An off-CPU sampler must work in a different way: either setting up timers in each application thread to wake them up and capture stacks, or having the kernel walk all threads at an interval and capture their stacks.

The approach in this PR doesn't have any of the mentioned drawbacks:
 1. It is tracing-based, it only does work when the thread starts running. There are no extra wakeups.
 2. It has low overhead, because it only samples at the specified rate.

## Comparison

I ran the following three commands, one after the other, with the same workload running in the background. Specifically, I had Firefox Nightly running which was displaying an animation and also running its own built-in profiler which calls `sleep` in ~0.7ms increments in each process's SamplerThread.

```
sudo ./tools/offcputime.py -uUf 2 | ../FlameGraph/flamegraph.pl --color=io --title="Off-CPU Time Flame Graph" --countname=us > out-traced.svg
sudo ./tools/offcputime.py -F 1000 -uUf 2 | ../FlameGraph/flamegraph.pl --color=io --title="Off-CPU Time Flame Graph" --countname=us > out-sampled.svg
sudo ./tools/offcputime.py -m 1000 -uUf 2 | ../FlameGraph/flamegraph.pl --color=io --title="Off-CPU Time Flame Graph" --countname=us > out-with-minimum.svg
```

Results:

 - [out-traced.svg](http://tests.themasta.com/tmp/out-traced.svg)
 - [out-sampled.svg](http://tests.themasta.com/tmp/out-sampled.svg)
 - [out-with-minimum.svg](http://tests.themasta.com/tmp/out-with-minimum.svg)

You can see that `out-traced.svg` and `out-sampled.svg` look very similar. Importantly, they both show off-cpu time in Firefox's `SamplerThread`. In contrast, `out-with-minimum.svg` does not show the off-cpu time in the `SamplerThread` because it filtered out those short sleep times.

### DIFF
--- a/tools/offcputime.py
+++ b/tools/offcputime.py
@@ -3,7 +3,7 @@
 # offcputime    Summarize off-CPU time by stack trace
 #               For Linux, uses BCC, eBPF.
 #
-# USAGE: offcputime [-h] [-p PID | -u | -k] [-U | -K] [-f] [duration]
+# USAGE: offcputime [-h] [-p PID | -u | -k] [-U | -K] [-F frequency] [-f] [duration]
 #
 # Copyright 2016 Netflix, Inc.
 # Licensed under the Apache License, Version 2.0 (the "License")
@@ -45,6 +45,7 @@ examples = """examples:
     ./offcputime             # trace off-CPU stack time until Ctrl-C
     ./offcputime 5           # trace for 5 seconds only
     ./offcputime -f 5        # 5 seconds, and output in folded format
+    ./offcputime -F 997      # sample off-CPU stack time at frequency 997 Hz
     ./offcputime -m 1000     # trace only events that last more than 1000 usec
     ./offcputime -M 10000    # trace only events that last less than 10000 usec
     ./offcputime -p 185      # only trace threads for PID 185
@@ -82,6 +83,11 @@ parser.add_argument("--stack-storage-size", default=1024,
     type=positive_nonzero_int,
     help="the number of unique stack traces that can be stored and "
          "displayed (default 1024)")
+parser.add_argument("-F", "--frequency", metavar="FREQUENCY", dest="frequency",
+    type=positive_nonzero_int,
+    help="the frequency at which sampling happens, per thread - we sample " +
+         "the wakeup stack whenever a thread's accumulated sleep time " +
+         "crosses a multiple of the sample period")
 parser.add_argument("duration", nargs="?", default=99999999,
     type=positive_nonzero_int,
     help="duration of trace, in seconds")
@@ -123,7 +129,18 @@ struct key_t {
     char name[TASK_COMM_LEN];
 };
 BPF_HASH(counts, struct key_t);
-BPF_HASH(start, u32);
+
+struct thread_timing_t {
+    // The timestamp at which the most recent sleep period 
+    // of this thread started.
+    u64 sleep_start_ts;
+
+    // The sleep delta time, in nanoseconds, which has accumulated
+    // since the last multiple of SAMPLE_PERIOD was crossed.
+    // Only used if SAMPLE_PERIOD is non-zero.
+    u64 accum_delta_ns;
+};
+BPF_HASH(thread_timing, u32, struct thread_timing_t);
 BPF_STACK_TRACE(stack_traces, STACK_STORAGE_SIZE);
 
 struct warn_event_t {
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
It is not clear to me what is the advantage of out-sampled.svg compared to out-traced.svg. It looks both approach will accumulate all needed events. Could you help explain in more details? cc @brendangregg

mstange:
The gathering of out-sampled.svg should have much lower overhead. I haven't measured the overhead myself, but there's a whole chapter of overhead warnings here: https://www.brendangregg.com/offcpuanalysis.html#Overhead

brendangregg:
No, this is not solving the overhead problem because it still traces scheduler events. This tool's use of -F implies to the end-user that it is performing collection by sampling, when it is not.

```
    ./offcputime -F 997      # sample off-CPU stack time at frequency 997 Hz
```

No, It's really doing a `trace all scheduler events and filter based on a 997 Hz sample rate`.

Yes, I know it is _simulating_ sampling. But it's still tracing scheduler events, which can be millions per second, and can hurt many production environments (64-CPU server under heavy load).

I've been working on real off-CPU sampling using timer-based interrupts, which I'll add to this tool (or if it ends up too different, a new tool). It won't trace scheduler events, which can be millions per second.

mstange:
Ooh, intriguing!

I concur that it's a bit misleading to call it "sampling". On the other hand, the `-m` flag is similarly misleading - it still traces events under that minimum, it just doesn't record stacks for those events.

I'm extremely curious to see a timer-based sampling tool. I also wonder how the overhead compares if you do a system-wide trace of all threads, most of which will be sleeping for most of the time. Tracing all scheduler events definitely has overhead, but sampling all threads on each timer interrupt will be quite costly, too.

brendangregg:
Yes, I see the -m and -M descriptions should say "include events that last..." instead of "trace events that last..." in the USAGE. The man page should also be a bit more clear and state that tracing happens anyway.

As for sampling: Imagine -p PID -F 49, on a 64-threaded application. That's not dissimilar to a -F 49 CPU profile of a 64-CPU system, which is acceptable. Not using a PID filter, and sampling all threads on the system (including kernel threads) is going to get costly. Some apps do have >1000 threads, and I'd want to reduce the sample rate there.

It's also possible that we could make the kernel cache stack traces by thread and clear the cache on context switch (as in, natively via a kernel patch, not a BPF program). So -F 49 sampling on a cold thread just fetches the cached stack instead of doing a stack walk every time.

mstange:
Makes sense. For the record, I help maintain two profilers which currently use the timer-based interrupt-all-threads approach: The built-in Firefox profiler, and a macOS profiler called [perfrecord](https://github.com/mstange/perfrecord/). In both projects, the overhead from sampling sleeping threads is the biggest performance problem. However, since both profilers run in user space, some of this overhead might not apply to a kernel-based timer profiler.
In the browser profiling use case there about 500 threads across multiple processes, we target a sampling frequency of 1000Hz, and we're fine with overhead in the single-digit percentages. In this scenario my intuition would be that tracing all context switching events (with simulated sampling) would have lower overhead than sampling all threads. But I'm eager to see actual data for this - I don't have any concrete numbers myself. And I welcome anything that makes it easier to get wall-clock time based sampling profiles on Linux.

mstange:
Thanks a lot for your feedback, Brendan!

I'm going to close this PR as it was mostly a proof-of-concept to gather your opinions on this approach, with a potential side-benefit that the simulated-sampling option might be useful to others.

brendangregg:
Thanks Markus. 1000 Hz is a lot. I used to recommend 999 Hz >15 years ago when the stacks I was looking at were usually 10-30 frames deep, for C and C++ apps, but nowadays at Netflix I'm often seeing >1000-frame stacks, which I don't want to sample that fast due to the costs with stack walking. 49 Hz is what we use by default. Actually, a Firefox profile I'm looking at at the moment is mostly 10-30 frames, so I guess higher rates is still ok there.

mstange:
1000Hz has served us quite well. We can't always "loop" the workload to the point where aggregates become useful; we many profiling scenarios such as application startup, cpu-bound page load, opening a tab, or scrolling, where we want to see sufficient profiling detail from a recording of just a single interaction. The high sampling resolution lets us see those details. It also lets us see the order in which things happen, i.e. the execution over time. [Startup example](https://share.firefox.dev/3EfJrs0)
The overhead is mitigated by the fact that the Firefox profiler only samples a small set of threads by default (< 20), and it's not an always-on profiler. Users have to explicitly turn the profiler on and off. There is no disruption of a production service or anything like that.

PR #3578, opened 2021-08-12
Funcslower improvements

dmonakhov:
Add tracing feature for slow functions only.

### DIFF
--- a/tools/funcslower.py
+++ b/tools/funcslower.py
@@ -27,14 +27,15 @@ import argparse
 import time
 
 examples = """examples:
-  ./funcslower vfs_write        # trace vfs_write calls slower than 1ms
+  ./funcslower syscall:write    # trace syscall write calls slower than 1ms
+  ./funcslower vfs_write        # same  but trace vfs_write
   ./funcslower -m 10 vfs_write  # same, but slower than 10ms
   ./funcslower -u 10 c:open     # trace open calls slower than 10us
   ./funcslower -p 135 c:open    # trace pid 135 only
   ./funcslower c:malloc c:free  # trace both malloc and free slower than 1ms
   ./funcslower -a 2 c:open      # show first two arguments to open
   ./funcslower -UK -m 10 c:open # Show user and kernel stack frame of open calls slower than 10ms
-  ./funcslower -f -UK c:open    # Output in folded format for flame graphs
+  ./funcslower -f -UK :open    # Output in folded format for flame graphs
 """
 parser = argparse.ArgumentParser(
     description="Trace slow kernel or user function calls.",
@@ -241,9 +242,16 @@ if args.verbose or args.ebpf:
 
 b = BPF(text=bpf_text)
 
+# Attach  probes
 for i, function in enumerate(args.functions):
+    library=""
+
     if ":" in function:
         library, func = function.split(":")
+        if library == "syscall":
+            library = ""
+            function = b.get_syscall_fnname(func)
+    if library:
         b.attach_uprobe(name=library, sym=func, fn_name="trace_%d" % i)
         b.attach_uretprobe(name=library, sym=func, fn_name="trace_return")
     else:
--- a/tools/funcslower.py
+++ b/tools/funcslower.py
@@ -23,6 +23,7 @@
 
 from __future__ import print_function
 from bcc import BPF
+from bcc.containers import filter_by_containers
 import argparse
 import time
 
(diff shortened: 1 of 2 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
@dmonakhov There is a bpf program compilation error:
```
30: .../virtual/main.c:201:22: error: variable has incomplete type 'struct profile_t'
30:     struct profile_t pe = {.count = 1, .first_ns = delta_ns, .last_ns = delta_ns};
30:                      ^
30: /virtual/main.c:201:12: note: forward declaration of 'struct profile_t'
30:     struct profile_t pe = {.count = 1, .first_ns = delta_ns, .last_ns = delta_ns};
30:            ^
30: /virtual/main.c:206:27: error: use of undeclared identifier 'blocked'
30:         u64 *blocked_ns = blocked.lookup(&tgid_pid);
30:                           ^
30: /virtual/main.c:212:29: error: use of undeclared identifier 'profileinfo'
30:     struct profile_t *old = profileinfo.lookup(&event);
30:                             ^
30: /virtual/main.c:214:23: error: incomplete definition of type 'struct profile_t'
30:         pe.count = old->count +1;
30:                    ~~~^
30: /virtual/main.c:201:12: note: forward declaration of 'struct profile_t'
30:     struct profile_t pe = {.count = 1, .first_ns = delta_ns, .last_ns = delta_ns};
30:            ^
30: /virtual/main.c:215:26: error: incomplete definition of type 'struct profile_t'
30:         pe.first_ns = old->first_ns;
30:                       ~~~^
30: /virtual/main.c:201:12: note: forward declaration of 'struct profile_t'
30:     struct profile_t pe = {.count = 1, .first_ns = delta_ns, .last_ns = delta_ns};
30:            ^
30: /virtual/main.c:216:27: error: incomplete definition of type 'struct profile_t'
30:         pe.total_ns += old->total_ns;
30:                        ~~~^
30: /virtual/main.c:201:12: note: forward declaration of 'struct profile_t'
30:     struct profile_t pe = {.count = 1, .first_ns = delta_ns, .last_ns = delta_ns};
30:            ^
30: /virtual/main.c:218:5: error: use of undeclared identifier 'profileinfo'
30:     profileinfo.update(&event, &pe);
30:     ^
30: /virtual/main.c:233:5: error: use of undeclared identifier 'blocked'
30:     blocked.update(&tgid_pid, &ts);
30:     ^
30: 8 errors generated.
30: Traceback (most recent call last):
30:   File "../../tools/funcslower.py", line 364, in <module>
30:     b = BPF(text=bpf_text)
30:   File "/home/fedora/jenkins/workspace/bcc-pr/label/fc28/build/src/python/bcc-python/bcc/__init__.py", line 452, in __init__
30:     raise Exception("Failed to compile BPF module %s" % (src_file or "<text>"))
```

yonghong-song:
@dmonakhov I will review the patch, but it would be good if you can explain in summary what are changes in the patch and why they are useful. Thanks!

yonghong-song:
If I understand correctly, you tried to collect some stack information through sampling or tracing specific scheduler functions or other user specified functions. I can see scheduling functions MIGHT be useful so you can identify why that function is slower. I am not 100% sure on sampling and user specified functions as they may just capture some stack traces and not sure whether you indeed capture the hot stack or not. cc @brendangregg

brendangregg:
I'd like to understand this better. So two problems right away:

1. Scheduler tracing is expensive due to the frequency of events. For scheduler tools that's usually unavoidable and is to be expected. But to add scheduler tracing to other tools seems like this is creating an unexpected gotcha.
2. Adding profiling...This seems like it may be violating "do one thing and do it well" philosophy.

Need better examples. Here's a path that _might_ be acceptable; I'm just thinking out loud here:

1. `funcslower`: show me slow functions
2. `funcslower -x`: include a column for CPU time (fetch from task_struct at no additional cost; no extra probes!)
3. If CPU time == latency: `funcslower --cpu-profile`: Include CPU profiling information for the slow functions only
4. If CPU time << latency: `funcslower --offcpu-profile`: include Off-CPU (blocking) trace information for the slow functions only

That gives people a path where by steps 3 and 4 they are choosing to use more expensive instrumentation for cases where it will help, and expect the extra expense (I would not include short option names, like -B and -S, since I'd want to make it abundantly clear that these options are making funcslower do a lot of extra work).

Is that what you had in mind?

brendangregg:
BTW, I did dig into this earlier, but I can't remember right now if it was possible to get the elapsed CPU times from the task_struct at arbitrary points. The times are in there (usr/sys breakdown) for getrusage(2).

If there's no current way, a workaround would be 99-Hz profiling to create a coarse CPU metric, with a granularity of ~10ms, for the -x option. No fetching stack traces at that point. The source code should have a block comment saying "This is a workaround until better support exists; don't copy this code" and the man page should make it clear that the CPU column is "currently" based on 10ms sampling and has a large degree of error.

I really wouldn't want users running `funcslower --cpu-profile --offcpu-profile` all the time, because they don't know which option to use, and causing needless performance issues. (In fact, it should reject both options and tell the user to "pick one based on the CPU/LAT ratio from the -x output".) Hence having a -x to guide them.

yonghong-song:
> BTW, I did dig into this earlier, but I can't remember right now if it was possible to get the elapsed CPU times from the task_struct at arbitrary points. The times are in there (usr/sys breakdown) for getrusage(2).

if I remember correctly, elapsed cpu is not available at arbitrary points. They are calculated during context switch/irq time as well as specific syscall (e.g., getrusage) time.

PR #3579, opened 2021-08-14
bcc/python: Improvements to python perf_event_attr ctype

davemarchevsky:
Discussion in #3571 identified some low-hanging fruit improvements to the perf_event_attr ctype. This PR adds these improvements as well as adding the new test from #3571 to CMakeFiles.

First commit focuses on `perf_event_attr` ctype improvements: 

```
    bcc/python: extend perf_event_attr ctype

    This commit brings the Perf.perf_event_attr ctype in line with version 6
    of struct perf_event_attr (see uapi/linux/perf_event.h kernel header).
    Specifically:
      * All named fields are added, including field names within anonymous
      unions and bitfields
      * Perf.perf_event_attr now complains when a field which isn't part of
      the ctype struct is set.
        * Goal here is to prevent users from setting a
        recently-added field - which we haven't updated the ctype _fields_ to
        include - and getting confused when it doesn't propagate to the
        perf_event_open syscall. This bit me in #3571 and I am pretty
        familiar with bcc internals so I'd like to prevent this from
        confusing others down the line.
      * Perf.perf_event_attr's 'flags' field is removed as it was a standin
      for the bitfields. The _old_ profile.py was the only script in bcc
      tools that I could find using this.

    The last bullet is a breaking change. Although `tools/old/profile.py`
    has been migrated to use the bitfield it was flipping using `flags`,
    there could be some scripts out in the wild which break. I don't think
    this is likely: this stuff hasn't been significantly touched since 2016
    and I suspect if users of the python interface were writing lots of
    perf_event programs we would've seen more python tools or activity here.

    Regardless, there is probably a way to keep `flags` field working while
    also exposing named bitfields, but I suspect it'll be ugly and wanted to
    see if anyone thought it was necessary.
```

second pulls out kernel_version_ge into utils for use in further commits:

```
    bcc/python tests: pull kernel_version_ge into utils

    This helper is replicated in a few different places, let's pull it out.
```

third adds `test_attach_perf_event.py` to test infra. The single test in that file fails due to old kernel headers on ubuntu 16.04, and regardless fails on everything due to lack of hw perf event support

```
    bcc/python: Add test_attach_perf_event.py to CMake tests

    Add to CMakeLists.txt of tests so that the test is run as part of github
    actions test suite. Shorten the sleep duration so test finishes faster -
    since it's just testing attach currently the extra time isn't producing
    more signal.

    Also add python equivalent of `perf_event_sample_format` enum so
    `sample_type` can be more clearly set.

    v2: The test doesn't work on ubuntu 16.04 due to old kernel headers. It
    doesn't work on the rest of the github actions VMs due to hardware perf
    events not being supported, so add necessary check / skip.
```

Fourth commit:
```
bcc/python: Add x86 and sw test to test_attach_perf_event.py
Since the current test can't run on github actions since there's no HW
perf counter access, add a test using software page faults perf
event, which might work.

Also, rename the current HW test in there to highlight that it'll work
for PowerPC, and add a similar test for x86.
```

/cc @athira-rajeev

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -141,6 +141,34 @@ class PerfSWConfig:
     DUMMY = 9
     BPF_OUTPUT = 10
 
+class PerfEventSampleFormat:
+    # from perf_event_sample_format in uapi/linux/bpf.h
+    IP = (1 << 0)
+    TID = (1 << 1)
+    TIME = (1 << 2)
+    ADDR = (1 << 3)
+    READ = (1 << 4)
+    CALLCHAIN = (1 << 5)
+    ID = (1 << 6)
+    CPU = (1 << 7)
+    PERIOD = (1 << 8)
+    STREAM_ID = (1 << 9)
+    RAW = (1 << 10)
+    BRANCH_STACK = (1 << 11)
+    REGS_USER = (1 << 12)
+    STACK_USER = (1 << 13)
+    WEIGHT = (1 << 14)
+    DATA_SRC = (1 << 15)
+    IDENTIFIER = (1 << 16)
+    TRANSACTION = (1 << 17)
+    REGS_INTR = (1 << 18)
+    PHYS_ADDR = (1 << 19)
+    AUX = (1 << 20)
+    CGROUP = (1 << 21)
+    DATA_PAGE_SIZE = (1 << 22)
+    CODE_PAGE_SIZE = (1 << 23)
+    WEIGHT_STRUCT = (1 << 24)
+
 class BPFProgType:
     # From bpf_prog_type in uapi/linux/bpf.h
     SOCKET_FILTER = 1
--- a/src/python/bcc/perf.py
+++ b/src/python/bcc/perf.py
@@ -16,31 +16,113 @@ import ctypes as ct
 import os
 from .utils import get_online_cpus
 
+class _sample_period_union(ct.Union):
+        _fields_ = [
+                ('sample_period', ct.c_ulong),
+                ('sample_freq', ct.c_ulong),
+        ]
+
+class _wakeup_events_union(ct.Union):
+        _fields_ = [
+                ('wakeup_events', ct.c_uint),
+                ('wakeup_watermark', ct.c_uint),
+        ]
+
+class _bp_addr_union(ct.Union):
+        _fields_ = [
+                ('bp_addr', ct.c_ulong),
+                ('kprobe_func', ct.c_ulong),
+                ('uprobe_path', ct.c_ulong),
+                ('config1', ct.c_ulong),
+        ]
+
+class _bp_len_union(ct.Union):
+        _fields_ = [
+                ('bp_len', ct.c_ulong),
+                ('kprobe_addr', ct.c_ulong),
+                ('probe_offset', ct.c_ulong),
+                ('config2', ct.c_ulong),
+        ]
+
 class Perf(object):
+
         class perf_event_attr(ct.Structure):
+                _anonymous_ = [
+                        "_sample_period_union",
+                        "_wakeup_events_union",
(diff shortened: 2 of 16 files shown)


athira-rajeev:
@davemarchevsky Hi Dave,

few doubts here from my testing:
- If perf_event_attr has an undefined attr field used , I am getting a warning now with the improvements in this patch. which is expected and good result. example:
```Warning: Setting field testing on perf_event_attr that isn't part of the ctype - testing won't make it to perf_event_open```

But I see that attach proceeds with other valid fields in perf_event_attr. Is my understanding right ?

- I tried with setting using both `sample_period` and `sample_freq` in perf_event_attr and test proceeded. example:
```           event_attr = Perf.perf_event_attr()
            event_attr.type = Perf.PERF_TYPE_HARDWARE
            event_attr.config = PerfHWConfig.CACHE_MISSES
            event_attr.sample_period = 100
            event_attr.sample_type = PerfEventSampleFormat.ADDR
            event_attr.sample_freq = 12
```

but we are allowing only one of them with the union defined in `sample_period_union`. Is it upto the user to pass relevant field or any restriction for this union is needed from bcc checks ?

davemarchevsky:
@athira-rajeev The "print warning, but still proceed" is intended functionality. I'd like to avoid throwing an exception or failing in a similarly aggressive way so as not to break existing tools. Of course, if an existing tool would stop working due to an aggressive failure, it would be due to it setting a field that doesn't make it to the struct, so perhaps it's worth it. On the other hand, the perf_event_attr API here isn't used by many tools in this repo, it's quite old so some tools out in the wild could be relying on it, and I'd rather give them a smoother transition.

Re: the union issue, I think this results in similar behavior to what would happen if you did those assignments in C:

```python
>>> import ctypes as ct
>>> class TestUnion(ct.Union):
...  _fields_ = [
...   ('sample_period', ct.c_uint),
...   ('sample_freq', ct.c_uint)
...  ]
...
>>> a = TestUnion()
>>> a.sample_period = 10
>>> print(a.sample_period)
10
>>> a.sample_freq = 20
>>> print(a.sample_period)
20
>>> print(a.sample_freq)
20
```

Not that such behavior doesn't have the potential to be confusing, but at least it's consistent, so I think it's fine like this. Maybe I'm biased because the situation I added some logic to prevent - setting a field on perf_event_attr ctype that wouldn't make it to perf event - bit me last week.

athira-rajeev:
> @athira-rajeev The "print warning, but still proceed" is intended functionality. I'd like to avoid throwing an exception or failing in a similarly aggressive way so as not to break existing tools. Of course, if an existing tool would stop working due to an aggressive failure, it would be due to it setting a field that doesn't make it to the struct, so perhaps it's worth it. On the other hand, the perf_event_attr API here isn't used by many tools in this repo, it's quite old so some tools out in the wild could be relying on it, and I'd rather give them a smoother transition.
> 
> Re: the union issue, I think this results in similar behavior to what would happen if you did those assignments in C:
> 
> ```python
> >>> import ctypes as ct
> >>> class TestUnion(ct.Union):
> ...  _fields_ = [
> ...   ('sample_period', ct.c_uint),
> ...   ('sample_freq', ct.c_uint)
> ...  ]
> ...
> >>> a = TestUnion()
> >>> a.sample_period = 10
> >>> print(a.sample_period)
> 10
> >>> a.sample_freq = 20
> >>> print(a.sample_period)
> 20
> >>> print(a.sample_freq)
> 20
> ```
> 
> Not that such behavior doesn't have the potential to be confusing, but at least it's consistent, so I think it's fine like this. Maybe I'm biased because the situation I added some logic to prevent - setting a field on perf_event_attr ctype that wouldn't make it to perf event - bit me last week.

Sure @davemarchevsky Thanks for the response and the improvements in perf_event_attr. 

Thanks
Athira

PR #3585, opened 2021-08-18
Ksnoop

alan-maguire:
ksnoop is libbpf-tools tool which can be used to examine kernel function entry arguments and return values. It's intended to function as a working example of the benefits of using deep data inspection via libbpf + BTF. ksnoop can carry out some basic tracing actions; choosing specific arguments/return value to trace, filtering by pid, tracing a specific set of functions etc. For more details see ksnoop.8. ksnoop requires a kernel built with BTF, and it can also be used to trace module functions+data if module BTF is available.

ksnoop was originally posted to the bpf kernel list as an RFC; Daniel Borkmann suggested that bcc's libbpf-tools might be an appropriate home.

### DIFF
--- a/libbpf-tools/Makefile
+++ b/libbpf-tools/Makefile
@@ -34,6 +34,7 @@ APPS = \
 	funclatency \
 	gethostlatency \
 	hardirqs \
+	ksnoop \
 	llcstat \
 	mountsnoop \
 	numamove \
--- a/libbpf-tools/ksnoop.bpf.c
+++ b/libbpf-tools/ksnoop.bpf.c
@@ -7,7 +7,6 @@
 #include <bpf/bpf_tracing.h>
 #include <bpf/bpf_core_read.h>
 
-#include <asm-generic/errno.h>
 #include "ksnoop.h"
 
 /* For kretprobes, the instruction pointer in the struct pt_regs context
@@ -21,8 +20,8 @@
  */
 #define FUNC_MAX_STACK_DEPTH	16
 
-#ifndef NULL
-#define NULL			0
+#ifndef ENOSPC
+#define ENOSPC			28
 #endif
 
 struct func_stack {
@@ -53,7 +52,7 @@ struct {
 	__uint(key_size, sizeof(int));
 } ksnoop_perf_map SEC(".maps");
 
-static inline void clear_trace(struct trace *trace)
+static void clear_trace(struct trace *trace)
 {
 	__builtin_memset(&trace->trace_data, 0, sizeof(trace->trace_data));
 	trace->data_flags = 0;
--- a/libbpf-tools/ksnoop.bpf.c
+++ b/libbpf-tools/ksnoop.bpf.c
@@ -33,7 +33,8 @@ struct func_stack {
 /* function call stack hashed on a per-task key */
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
-	__uint(max_entries, 512);
+	/* function call stack for functions we are tracing */
+	__uint(max_entries, (8 * MAX_FUNC_TRACES));
 	__type(key, __u64);
 	__type(value, struct func_stack);
 } ksnoop_func_stack SEC(".maps");
@@ -41,7 +42,7 @@ struct {
 /* per-cpu trace info hashed on function address */
 struct {
 	__uint(type, BPF_MAP_TYPE_PERCPU_HASH);
-	__uint(max_entries, 8);
+	__uint(max_entries, MAX_FUNC_TRACES);
 	__type(key, __u64);
 	__type(value, struct trace);
 } ksnoop_func_map SEC(".maps");
@@ -59,7 +60,7 @@ static void clear_trace(struct trace *trace)
 	trace->buf_len = 0;
 }
 
-static inline struct trace *get_trace(struct pt_regs *ctx, bool entry)
+static struct trace *get_trace(struct pt_regs *ctx, bool entry)
 {
 	__u8 stack_depth, last_stack_depth;
 	struct func_stack *func_stack;
@@ -184,7 +185,7 @@ static inline struct trace *get_trace(struct pt_regs *ctx, bool entry)
 	return trace;
 }
 
-static inline void output_trace(struct pt_regs *ctx, struct trace *trace)
+static void output_trace(struct pt_regs *ctx, struct trace *trace)
 {
 	__u16 trace_len;
 
@@ -212,7 +213,7 @@ skip:
 	clear_trace(trace);
 }
 
-static inline void output_stashed_traces(struct pt_regs *ctx,
+static void output_stashed_traces(struct pt_regs *ctx,
 					 struct trace *currtrace,
 					 bool entry)
 {
@@ -267,7 +268,7 @@ static inline void output_stashed_traces(struct pt_regs *ctx,
 	output_trace(ctx, currtrace);
 }
 
-static inline __u64 get_arg(struct pt_regs *ctx, enum arg argnum)
+static __u64 get_arg(struct pt_regs *ctx, enum arg argnum)
 {
 	switch (argnum) {
 	case KSNOOP_ARG1:
@@ -287,7 +288,7 @@ static inline __u64 get_arg(struct pt_regs *ctx, enum arg argnum)
 	}
 }
 
-static inline int ksnoop(struct pt_regs *ctx, bool entry)
+static int ksnoop(struct pt_regs *ctx, bool entry)
 {
 	void *data_ptr = NULL;
 	struct trace *trace;
--- a/libbpf-tools/ksnoop.bpf.c
+++ b/libbpf-tools/ksnoop.bpf.c
@@ -30,11 +30,13 @@ struct func_stack {
 	__u8 stack_depth;
 };
 
+#define MAX_TASKS		2048
+
 /* function call stack hashed on a per-task key */
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
 	/* function call stack for functions we are tracing */
-	__uint(max_entries, (8 * MAX_FUNC_TRACES));
+	__uint(max_entries, MAX_TASKS);
 	__type(key, __u64);
 	__type(value, struct func_stack);
 } ksnoop_func_stack SEC(".maps");
--- a/libbpf-tools/ksnoop.bpf.c
+++ b/libbpf-tools/ksnoop.bpf.c
@@ -1,0 +1,455 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+/* Copyright (c) 2021, Oracle and/or its affiliates. */
+
+#include "vmlinux.h"
+
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include <bpf/bpf_core_read.h>
+
+#include <asm-generic/errno.h>
+#include "ksnoop.h"
+
+/* For kretprobes, the instruction pointer in the struct pt_regs context
+ * is the kretprobe_trampoline.  We derive the instruction pointer
+ * by pushing it onto a function stack on entry and popping it on return.
+ *
+ * We could use bpf_get_func_ip(), but "stack mode" - where we
+ * specify functions "a", "b and "c" and only want to see a trace if "a"
+ * calls "b" and "b" calls "c" - utilizes this stack to determine if trace
+ * data should be collected.
+ */
+#define FUNC_MAX_STACK_DEPTH	16
+
+#ifndef NULL
+#define NULL			0
+#endif
+
+struct func_stack {
+	__u64 task;
+	__u64 ips[FUNC_MAX_STACK_DEPTH];
+	__u8 stack_depth;
+};
+
+/* function call stack hashed on a per-task key */
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 512);
+	__type(key, __u64);
+	__type(value, struct func_stack);
+} ksnoop_func_stack SEC(".maps");
+
+/* per-cpu trace info hashed on function address */
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_HASH);
+	__uint(max_entries, 8);
+	__type(key, __u64);
+	__type(value, struct trace);
+} ksnoop_func_map SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(value_size, sizeof(int));
+	__uint(key_size, sizeof(int));
+} ksnoop_perf_map SEC(".maps");
+
+static inline void clear_trace(struct trace *trace)
+{
+	__builtin_memset(&trace->trace_data, 0, sizeof(trace->trace_data));
+	trace->data_flags = 0;
+	trace->buf_len = 0;
+}
+
+static inline struct trace *get_trace(struct pt_regs *ctx, bool entry)
+{
+	__u8 stack_depth, last_stack_depth;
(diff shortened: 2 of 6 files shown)


chenhengqi:
This tool  looks pretty cool !

man/man8/ksnoop.8
```diff
@@ -0,0 +1,298 @@
...
+                                      .skc_rcv_saddr = (__be32)0x1701a8c0,
+                                     },
+                                    },
+                                  ...
+.ft P
+.fi
+.UNINDENT
+.UNINDENT
+.sp
+Trace meber information associated with argument.  Only one level of
```
chenhengqi, line 217: typo: meber

libbpf-tools/Makefile
```diff
@@ -34,6 +34,7 @@ APPS = \
 	funclatency \
 	gethostlatency \
 	hardirqs \
+	ksnoop \
```
chenhengqi, line 37: Please also add an entry to .gitignore

brendangregg:
G'Day Alan; It's a cool tool (good use of BTF) but why wouldn't we add that BTF pretty printing support to the existing bcc trace.py? (We do need a libbpf C version of trace.py.) I once gave an internal talk at a major tech company, and I asked what their favorite tool was. They said trace.py. It does have some adoption.

Also, trace.py was created before we had bpftrace. It's like a lightweight bpftrace. But that's all the more reason why I don't think we want to crowd this space further: End users shouldn't have to spend time choosing between trace.py, bpftrace, and ksnoop, if it's possible we can do the same with fewer tools. Imagine we add BTF-style output to bpftrace (either as a printf() %, enhancing print(), or a new print builtin like printt() for print type).

man/man8/ksnoop.8
alan-maguire: thanks for taking a look; will fix!

libbpf-tools/Makefile
alan-maguire: good catch, thanks!

alan-maguire:
> G'Day Alan; It's a cool tool (good use of BTF) but why wouldn't we add that BTF pretty printing support to the existing bcc trace.py? (We do need a libbpf C version of trace.py.) I once gave an internal talk at a major tech company, and I asked what their favorite tool was. They said trace.py. It does have some adoption.
> 
> Also, trace.py was created before we had bpftrace. It's like a lightweight bpftrace. But that's all the more reason why I don't think we want to crowd this space further: End users shouldn't have to spend time choosing between trace.py, bpftrace, and ksnoop, if it's possible we can do the same with fewer tools. Imagine we add BTF-style output to bpftrace (either as a printf() %, enhancing print(), or a new print builtin like printt() for print type).

Hey Brendan

I'd really like to see bpftrace with BTF-style output; my long-term aim is to drive adoption of this functionality across tracers, hence this example and my upcoming eBPF summit lightning talk today. I can't do that directly for bpftrace, but I'd hope others might pick that up and I'd be happy to help with any core libbpf/BTF issues uncovered as part of that effort if needed. 

In the meantime, all I can say is this tracing tool fills a real need for me; a standalone (no toolchain required) way to facilitate kernel debugging function data in depth. That's not for everybody clearly, but it's got some unique features that solve some specific problems I run into a lot
- show me the function entry/return data for a specific stack signature; 
- show me the entry arguments when the function returns a particular value (such as a failure)

These would be doable in more general tracers of course with BTF dump data support, but in the interim it's useful to have something that solves those problems. I have to confess I like the simplicity of how those questions are expressed;
for example, "show me the skbs for ip_send_skb() calls that fail" becomes

ksnoop "ip_send_skb(skb, return != 0)"

My main aim with this work though is to get the code that suports this functionality "out there" as a reference others can use.

libbpf-tools/ksnoop.c
```diff
@@ -0,0 +1,979 @@
...
+static int trace_to_value(struct btf *btf, struct func *func, char *argname,
+			  char *membername, char *predicate, struct value *val)
+{
+	__u8 i;
+
+	strncpy(val->name, argname, sizeof(val->name));
+	if (strlen(membername) > 0) {
+		strncat(val->name, "->", sizeof(val->name));
+		strncat(val->name, membername, sizeof(val->name));
+	}
```
chenhengqi, line 314: I saw a compilation warning here:
```shell
In file included from /usr/include/string.h:495,
                 from ksnoop.c:10:
In function ‘strncat’,
    inlined from ‘trace_to_value’ at ksnoop.c:313:3,
    inlined from ‘parse_trace’ at ksnoop.c:602:7:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:136:10: warning: ‘__builtin___strncat_chk’ specified bound 256 equals destination size [-Wstringop-overflow=]
  136 |   return __builtin___strncat_chk (__dest, __src, __len, __bos (__dest));
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

How about using snprintf here ?

libbpf-tools/ksnoop.bpf.c
```diff
@@ -0,0 +1,455 @@
...
+	__type(value, struct trace);
+} ksnoop_func_map SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(value_size, sizeof(int));
+	__uint(key_size, sizeof(int));
+} ksnoop_perf_map SEC(".maps");
+
+static inline void clear_trace(struct trace *trace)
```
chenhengqi, line 56: `inline` is not required.

libbpf-tools/ksnoop.c
alan-maguire: good idea: i've simplified to use

if (strlen(membername) > 0)
                snprintf(val->name, sizeof(val->name), "%s->%s",
                         argname, membername);
        else
                strncpy(val->name, argname, sizeof(val->name));

libbpf-tools/ksnoop.bpf.c
alan-maguire: will fix, thanks!

libbpf-tools/ksnoop.bpf.c
```diff
@@ -0,0 +1,455 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+/* Copyright (c) 2021, Oracle and/or its affiliates. */
+
+#include "vmlinux.h"
+
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include <bpf/bpf_core_read.h>
+
+#include <asm-generic/errno.h>
```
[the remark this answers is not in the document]
alan-maguire: sure, sounds good.

libbpf-tools/ksnoop.bpf.c
```diff
@@ -0,0 +1,455 @@
...
+ * We could use bpf_get_func_ip(), but "stack mode" - where we
+ * specify functions "a", "b and "c" and only want to see a trace if "a"
+ * calls "b" and "b" calls "c" - utilizes this stack to determine if trace
+ * data should be collected.
+ */
+#define FUNC_MAX_STACK_DEPTH	16
+
+#ifndef NULL
+#define NULL			0
+#endif
```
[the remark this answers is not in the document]
alan-maguire: no, leftover from an earlier version,thanks for catching!

alan-maguire:
Any additional thoughts/comments here? I did some investigation on trace.py; given that it is quite a general-purpose tracer, I don't think converting it to a libbpf tool would be simple. Having used ksnoop for a while now in kernel debugging I strongly believe it provides a value in tasks centred on digging into kernel failure modes (specific stack signatures, entry arguments when return fails etc) and the benefits of having it available - at least until bpftrace grows equivalent functionality - are worthwhile. It can also operate without LLVM being libbpf+CO-RE based, and is functional on a range of kernels provided BTF is available (I use it to debug our 5.4-based kernel which ships with kernel BTF).

libbpf-tools/ksnoop.c
```diff
@@ -0,0 +1,979 @@
...
+			      libbpf_errstr(vmlinux_btf));
+			return NULL;
+		}
+	}
+	if (!name || strlen(name) == 0)
+		return vmlinux_btf;
+
+	snprintf(path, sizeof(path), "/sys/kernel/btf/%s", name);
+
+	mod_btf = btf__parse_raw_split(path, vmlinux_btf);
```
chenhengqi, line 361: Let us use `btf__load_vmlinux_btf` and `btf__load_module_btf` instead.

libbpf-tools/ksnoop.c
```diff
@@ -0,0 +1,979 @@
...
+			err = -EINVAL;
+			goto out;
+		}
+		if (strcmp(func->name, sym_name) != 0)
+			continue;
+		func->ip = sym_addr;
+		func->mod[0] = '\0';
+		/* get module name from [modname] */
+		if (ret == 4) {
+			if (sscanf(mod_info, "%*[\t ]\[%[^]]", func->mod) < 1) {
```
chenhengqi, line 487: Should be `"%*[\t ][%[^]]"`

brendangregg:
I need to catch up and watch your eBPF summit talk video. I'm not convinced we need a 3rd tool; if bpftrace came first, I would have argued against trace.py (for obvious reasons), and this just seems similar. But I'll watch the video to really understand it!

alan-maguire:
> I need to catch up and watch your eBPF summit talk video. I'm not convinced we need a 3rd tool; if bpftrace came first, I would have argued against trace.py (for obvious reasons), and this just seems similar. But I'll watch the video to really understand it!

Thanks! The key differences probably don't get teased out within the scope of that talk unfortunately (5 minutes meant I couldn't describe too much detail) but the essence of the value (aside from the basic arg/return value tracing) is it implements a form of speculative tracing;

- if I specify I want to see an argument for a function which later returns a -1 value, I will say "ksnoop ip_send_skb(skb, return==--1)". What happens under the covers is the argument values are stashed until return time, and if the predicate is satisfied then the perf event is issued, otherwise the data is discarded. So we speculatively record the args for later evaluation, and this is useful for understanding specific function failures. Similarly;
- if I specify a set of functions and use the "-s" option, again the argument data is stashed until the set of functions is called one after the other (though not necessarily directly), and only then a set of perf events - one for each function - is sent with the stashed arguments. Useful for cases where I have for example a syzcall reproducer which gives a failure with a specific stack signature, and I want to dig in deeper examining the arguments.

So, in essence ksnoop - as well as providing basic arg/return value display - also provides a way to speculatively trace on the basis of future outcomes. The outcomes I chose are closely tied to kernel debugging tasks (why did this function fail? what's going on when this particular sequence of functions gets called?). This is all expressible in different tracers in different ways of course, but there's an advantage in having a specific tool for those jobs I think.

libbpf-tools/ksnoop.bpf.c
```diff
@@ -0,0 +1,454 @@
...
+	__u8 stack_depth;
+};
+
+/* function call stack hashed on a per-task key */
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 512);
+	__type(key, __u64);
+	__type(value, struct func_stack);
+} ksnoop_func_stack SEC(".maps");
```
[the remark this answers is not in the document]
alan-maguire: Yep, the map is populated with one entry per function specified on the commandline, and there is no need to delete those entries - they live for the lifetime of the program to record tracing information for the function traced. max_entries should probably be lower now I look at it;  we only allow MAX_FUNC_TRACES (currently 8) different functions to be traced. I think we can probably bump MAX_FUNC_TRACES up to 64 and use it explicitly in the map max_entries to reflect the limitation. We probably don't want BPF_F_NO_PREALLOC since the map size is small I think?

libbpf-tools/ksnoop.c
alan-maguire: hmm, don't we need to escape the "[" to match the start of the "[modulename]"? You might notice that despite it being escaped, we have a matching "]" at the end of the string; that's because we get a warning:
ksnoop.c:487:39: warning: no closing ‘]’ for ‘%[’ format [-Wformat=]

...if we don't include it.

libbpf-tools/ksnoop.c
alan-maguire: good idea, much neater!

libbpf-tools/ksnoop.c
```diff
@@ -0,0 +1,979 @@
...
+	p_err("\t/* lost %llu events */", cnt);
+}
+
+static int add_traces(struct bpf_map *func_map, struct trace *traces,
+		      int nr_traces)
+{
+	int i, j, ret, nr_cpus = libbpf_num_possible_cpus();
+	struct trace *map_traces;
+
+	map_traces = calloc(nr_cpus, sizeof(struct trace));
```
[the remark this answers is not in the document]
alan-maguire: it should, will fix!

libbpf-tools/ksnoop.c
```diff
@@ -0,0 +1,979 @@
...
+	printf("%16s %4s %8s %s\n", "TIME", "CPU", "PID", "FUNCTION/ARGS");
+
+	while (1) {
+		ret = perf_buffer__poll(pb, 1);
+		if (ret < 0 && ret != -EINTR) {
+			p_err("Polling failed: %s", strerror(-ret));
+			return 1;
+		}
+	}
+
```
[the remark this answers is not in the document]
alan-maguire: good idea, will add this.

libbpf-tools/ksnoop.c
```diff
@@ -0,0 +1,979 @@
...
+#include "ksnoop.h"
+#include "ksnoop.skel.h"
+
+#ifndef KSNOOP_VERSION
+#define KSNOOP_VERSION	"0.1"
+#endif
+
+struct btf *vmlinux_btf;
+const char *bin_name;
+int pages = PAGES_DEFAULT;
```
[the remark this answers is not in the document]
alan-maguire: yep, will fix these (and filter_pid, log_level etc).

libbpf-tools/ksnoop.c
chenhengqi: Actually, I've written a small program to test the logic here, remove the `\` works for me.

libbpf-tools/ksnoop.c
```diff
@@ -0,0 +1,979 @@
...
+				      kretprobe ? "kretprobe" : "kprobe",
+				      traces[i].func.name,
+				      strerror(-ret));
+				return ret;
+			}
+			p_debug("Attached %s for '%s'",
+				kretprobe ? "kretprobe" : "kprobe",
+				traces[i].func.name);
+		}
+	}
```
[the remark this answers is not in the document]
alan-maguire: yep, this is simpler alright. will fix.

libbpf-tools/ksnoop.c
alan-maguire: you're right, sorry; i thought i remembered this not working. i've removed it now and module tracing still works..

libbpf-tools/ksnoop.bpf.c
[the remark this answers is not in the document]
chenhengqi: My concern here is, if we have 1000(> 512) process trigger the same kernel function, we may suffer data loss.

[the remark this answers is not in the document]
alan-maguire: hmm, for now I've made a change to make size 2048; since each map entry is ~150 bytes we can afford to do this I think.

chenhengqi (approved):
LGTM

alan-maguire:
> LGTM

Thanks very much for the review!

yonghong-song:
[buildbot, test this please]

PR #3591, opened 2021-08-27
Add the support for detaching a single kprobe/kretprobe handler

hao-lee:
The uprobe also has a similar problem. I will fix it in a subsequent patch.

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -99,6 +99,8 @@ This guide is incomplete. If something feels missing, check the bcc and kernel s
         - [9. attach_xdp()](#9-attach_xdp)
         - [10. attach_func()](#10-attach_func)
         - [11. detach_func()](#11-detach_func)
+        - [12. detach_kprobe()](#12-detach_kprobe)
+        - [13. detach_kretprobe()](#13-detach_kretprobe)
     - [Debug Output](#debug-output)
         - [1. trace_print()](#1-trace_print)
         - [2. trace_fields()](#2-trace_fields)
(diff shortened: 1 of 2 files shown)


chenhengqi:
I wonder if there are any realworld use cases for this change.

yonghong-song:
[buildbot, test this please]

yonghong-song:
[buildbot, ok to test]

yonghong-song:
The change looks okay. But could you explain how this change will help you in practice? This seems only happen when you try to attach two bpf programs to the *same* event (kernel or user function). I wonder what is the use case for that in bcc context.

hao-lee:
@chenhengqi @yonghong-song We want to measure both the execution count, execution time, and some other metrics of a kernel function. For flexibility, we want to use separate handlers for each metric to disable them individually if any of them incur some performance penalties. Without this interface, we have to disable all handlers on the kernel function.

yonghong-song:
> @chenhengqi @yonghong-song We want to measure both the execution count, execution time, and some other metrics of a kernel function. For flexibility, we want to use separate handlers for each metric to disable them individually if any of them incur some performance penalties. Without this interface, we have to disable all handlers on the kernel function.

Could you add this into commit message?

yonghong-song:
There are some test failures, e.g.,
```
28: Traceback (most recent call last):
28:   File "/home/fedora/jenkins/workspace/bcc-pr/label/fc26/tests/python/test_percpu.py", line 122, in test_struct_custom_func
28:     bpf_code.detach_kprobe(event_name)
28: TypeError: detach_kprobe() takes exactly 3 arguments (2 given)
28: 
28: ======================================================================
28: ERROR: test_u32 (__main__.TestPercpu)
28: ----------------------------------------------------------------------
28: Traceback (most recent call last):
28:   File "/home/fedora/jenkins/workspace/bcc-pr/label/fc26/tests/python/test_percpu.py", line 74, in test_u32
28:     bpf_code.attach_kprobe(event=event_name, fn_name="hello_world")
28:   File "/home/fedora/jenkins/workspace/bcc-pr/label/fc26/build/src/python/bcc-python/bcc/__init__.py", line 834, in attach_kprobe
28:     (fn_name, event))
28: Exception: Failed to attach BPF program hello_world to kprobe sys_clone
```
could you take a look?

hao-lee:
> Could you add this into commit message?
Sure.

> There are some test failures

I will fix it in the next pushing. Thanks!

hao-lee:
Hello, I have added more information to the commit message and have fixed the above problems. Thanks for the review.

Issue #3592, opened 2021-08-27
update killsnoop to use tracepoints


brendangregg:
This is a request for help.

I wrote killsnoop back in 2015 before tracepoint support, and so I kprobe'd sys_kill(). It still does some derivation of that. But now there's a report it no longer works on Linux 5.11: https://github.com/iovisor/bcc/pull/3572#issuecomment-900357032 CC @chenhengqi 

Can someone please update killsnoop (both Python and libbpf-tools) to use tracepoints instead of kprobes (if it works as expected). All of these:

```
  syscalls:sys_enter_kill                            [Tracepoint event]
  syscalls:sys_enter_tgkill                          [Tracepoint event]
  syscalls:sys_enter_tkill                           [Tracepoint event]
  syscalls:sys_exit_kill                             [Tracepoint event]
  syscalls:sys_exit_tgkill                           [Tracepoint event]
  syscalls:sys_exit_tkill                            [Tracepoint event]
```

chenhengqi:
I made a mistake. I think the behavior of killsnoop.py is correct.

The manpage of kill(2) has the following description:
```
If sig is 0, then no signal is sent, but existence and permission checks are still  performed;  this  can  be
used to check for the existence of a process ID or process group ID that the caller is permitted to signal.
```

And using strace also confirmed this:
```
kill(1240, 0)                           = -1 EPERM (Operation not permitted)
kill(2733, 0)                           = -1 EPERM (Operation not permitted)
kill(1396, 0)                           = 0
kill(11199, 0)                          = -1 EPERM (Operation not permitted)
kill(685, 0)                            = -1 EPERM (Operation not permitted)
kill(757, 0)                            = 0
kill(1449, 0)                           = -1 EPERM (Operation not permitted)
kill(839, 0)                            = 0
kill(686, 0)                            = -1 EPERM (Operation not permitted)
kill(715, 0)                            = -1 EPERM (Operation not permitted)
kill(721, 0)                            = 0
kill(1148, 0)                           = -1 EPERM (Operation not permitted)
kill(1168, 0)                           = -1 EPERM (Operation not permitted)
kill(754, 0)                            = 0
kill(758, 0)                            = 0
kill(1173, 0)                           = -1 EPERM (Operation not permitted)
kill(8680, 0)                           = -1 EPERM (Operation not permitted)
kill(11327, 0)                          = -1 EPERM (Operation not permitted)
kill(1162, 0)                           = -1 EPERM (Operation not permitted)
[...]

PR #3593, opened 2021-08-29
bcc/tools: Update killsnoop to use tracepoints

chenhengqi:
As suggested by @brendangregg in #3592, update killsnoop
to use tracepoints. This commit adds kill/tkill/tgkill as
tracing targets using tracepoints. The output looks the
same as using kprobe:

sudo ./killsnoop.py
00:17:54  151828 code             0    11326  0
00:17:54  151828 code             0    11326  0
00:17:54  834    containerd       23   40277  0


brendangregg:
Thanks; the man page needs to be updated to say "killsnoop traces the kill() family of syscalls".

Also, if you don't mind, please change PID to be %7d and TPID to be %8d, to fit the bigger PID numbers we see on today's systems (which I find pretty annoying) and to include the "-" in the TPID column for signals to process groups (it should always have been one char bigger).

chenhengqi:
> Thanks; the man page needs to be updated to say "killsnoop traces the kill() family of syscalls".
> 
> Also, if you don't mind, please change PID to be %7d and TPID to be %8d, to fit the bigger PID numbers we see on today's systems (which I find pretty annoying) and to include the "-" in the TPID column for signals to process groups (it should always have been one char bigger).

Updated.

brendangregg:
[buildbot, ok to test]

devidasjadhav:
the build error was due to "failed to find valid kernel BTF"

31: Traceback (most recent call last):
31:   File "../../tools/killsnoop.py", line 171, in <module>
31:     b = BPF(text=bpf_text)
31:   File "/home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1710/build/src/python/bcc-python/bcc/__init__.py", line 488, in __init__
31:     self._trace_autoload()
31:   File "/home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1710/build/src/python/bcc-python/bcc/__init__.py", line 1428, in _trace_autoload
31:     self.attach_tracepoint(tp=tp, fn_name=fn.name)
31:   File "/home/iovisor/jenkins/workspace/bcc-pr/label/ubuntu1710/build/src/python/bcc-python/bcc/__init__.py", line 1007, in attach_tracepoint
31:     (fn_name, tp))
31: Exception: Failed to attach BPF program tracepoint__syscalls__sys_enter_kill to tracepoint syscalls:sys_enter_kill
31: Fs..........libbpf: failed to find valid kernel BTF
31: libbpf: vmlinux BTF is not found
31: ......'unknown': I need something more specific.
31: .In file included from /virtual/main.c:6:

what would be right approach ?
1.  having conditional kernel check and changing behavior for backward compatibility. 
2.  exiting gracefully with error this behavior is not supported. and retaining old python in [old directory](https://github.com/iovisor/bcc/tree/master/tools/old) .

chenhengqi:
I prefer option 1.

chenhengqi:
@brendangregg  Please take a look at this again.

I've updated this tool to use tracepoints if available and fallback to kprobe so that it won't break old CI instances.

brendangregg:
[buildbot, ok to test]

chenhengqi:
The CI still failed. `tracepoint_exists` is not suitable to detect whether tracepoint exists.

This [commit](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=98b5c2c65c2951772a8fc661f50d675e450e8bce) allows BPF attached to tracepoints and exports a symbol `perf_swevent_put_recursion_context `. I think we can use this to detect tracepoint support. Is it feasible ?

cc @yonghong-song  for help.

chenhengqi:
BTW, the CI complains about vmlinux BTF not exist, I can't figure out why.

devidasjadhav:
Analysis of logs lead to 
`if BPF.tracepoint_exists("syscalls", "sys_enter_kill")` doesn't work as expected.

for reference definition of function used is [here](https://github.com/iovisor/bcc/blob/b00e6b417641a053bdf29db5235af213af5a9617/src/python/bcc/__init__.py#L989 )

which means it will check directory `events/syscalls/sys_enter_kill/`

The reason is there is syscall in 4.4 which is kernel in ubuntu 16.04.
```
root@ubuntu:/home/a# lsb_release -d
Description:	Ubuntu 16.04.7 LTS
root@ubuntu:/home/a# ls /sys/kernel/debug/tracing/events/syscalls/sys_enter_kill/
enable  filter  format  id  trigger
root@ubuntu:/home/a# uname -a
Linux ubuntu 4.4.0-186-generic #216-Ubuntu SMP Wed Jul 1 05:34:05 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

So we need to come-up with better way to detect this.
easiest would be checking kernel version and matching with supported version.
this would be quick fix which can be changed back when we resolve `tracepoint_exists` with proper way. 
or coming up with `is_tracepoint` function. 

issue is caused because syscalls and tracepoints have same name and tracepoints doesn't have separate directory.

devidasjadhav:
Here is code I have added to compare versions.

replace if condition in you code with this.

```python
import platform

def version2tuple(v):
    return tuple(map(int, (v.split("."))))

def kernel_version_greater_than(ver_str):
    cur_ver = platform.release()
    abs_ver = cur_ver.split("-")[0]
    return version2tuple(abs_ver) > version2tuple(ver_str)

# initialize BPF
if kernel_version_greater_than("5.4"):
```

tested if this works by changing multiple versions higher and smaller than current version.
Hope this helps.

Note:
I thought of adding PR but most of work was already done by @chenhengqi . so just sharing code snippet.

chenhengqi:
> Here is code I have added to compare versions.
> 
> replace if condition in you code with this.
> 
> ```python
> import platform
> 
> def version2tuple(v):
>     return tuple(map(int, (v.split("."))))
> 
> def kernel_version_greater_than(ver_str):
>     cur_ver = platform.release()
>     abs_ver = cur_ver.split("-")[0]
>     return version2tuple(abs_ver) > version2tuple(ver_str)
> 
> # initialize BPF
> if kernel_version_greater_than("5.4"):
> ```
> 
> tested if this works by changing multiple versions higher and smaller than current version.
> Hope this helps.
> 
> Note:
> I thought of adding PR but most of work was already done by @chenhengqi . so just sharing code snippet.

This is not accurate due to kernel backport. This kind of work had been proposed before.

As I mentioned above, maybe we could use `perf_swevent_put_recursion_context` symbol to detect BPF tracepoint support. I am not 100% sure, so still wait for Yonghong's confirmation.

yonghong-song:
Not sure whether the symbol `perf_swevent_put_recursion_context` will work or not. How about `perf_event_attach_bpf_prog`?

chenhengqi:
> Not sure whether the symbol `perf_swevent_put_recursion_context` will work or not. How about `perf_event_attach_bpf_prog`?

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=98b5c2c65c2951772a8fc661f50d675e450e8bce

`perf_swevent_put_recursion_context` is exported with `allow bpf programs attach to tracepoints` feature.

While `perf_event_attach_bpf_prog` seems too generic. :(

yonghong-song:
> > Not sure whether the symbol `perf_swevent_put_recursion_context` will work or not. How about `perf_event_attach_bpf_prog`?
> 
> https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=98b5c2c65c2951772a8fc661f50d675e450e8bce
> 
> `perf_swevent_put_recursion_context` is exported with `allow bpf programs attach to tracepoints` feature.
> 
> While `perf_event_attach_bpf_prog` seems too generic. :(

I am fine with perf_swevent_put_recursion_context if it works :-) I just didn't closely check it. `perf_event_attach_bpf_prog` is introduced later so it is definitely safe but may be too conservative.

chenhengqi:
> > > Not sure whether the symbol `perf_swevent_put_recursion_context` will work or not. How about `perf_event_attach_bpf_prog`?
> > 
> > 
> > https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=98b5c2c65c2951772a8fc661f50d675e450e8bce
> > `perf_swevent_put_recursion_context` is exported with `allow bpf programs attach to tracepoints` feature.
> > While `perf_event_attach_bpf_prog` seems too generic. :(
> 
> I am fine with perf_swevent_put_recursion_context if it works :-) I just didn't closely check it. `perf_event_attach_bpf_prog` is introduced later so it is definitely safe but may be too conservative.

After some investigation, I think checking existence of `perf_swevent_put_recursion_context` ksym is not feasible.

matthew-olson-intel:
Is this conversation resolved? It looks to me like it was only failing in kernel 4.4, which the CI no longer runs, so is the CI succeeding now?

brendangregg:
[buildbot, ok to test]

PR #3596, opened 2021-08-31
Add filtering capabilities to mountsnoop.py

eiffel-fl:
Hi.

First, I hope you are fine and the same for your relatives.

In  kinvolk/inspektor-gadget#169, I added support for `mountsnoop` gadget.
To be able to use `--selector`, I needed to modify BPF code generated in `mountsnoop.py`.
First, I needed to call `filter_by_containers()` and also add check to `container_should_be_filtered()` in BPF code itself.
As `struct mnt_namespace` is defined in `mountsnoop.py` BPF code (as we want to snoop `mount` operations), I added guard around this definition, so we do not define this structure two times (one time in `mountnsoop.py` and the other in `containers.py`).

I tested these modifications inside `minikube` and using `inspektor-gadget` to interact with `mountsnoop.py`.

If you see any ways to improve this code, feel free to share.

Best regards.

### DIFF
--- a/tools/mountsnoop.py
+++ b/tools/mountsnoop.py
@@ -13,6 +13,7 @@
 from __future__ import print_function
 import argparse
 import bcc
+from bcc.containers import filter_by_containers
 import ctypes
 import errno
 import functools
@@ -31,7 +32,14 @@ bpf_text = r"""
  * VFS and not installed in any kernel-devel packages. So, let's duplicate the
  * important part of the definition. There are actually more members in the
  * real struct, but we don't need them, and they're more likely to change.
+ *
+ * To add support for --selector option, we need to call filter_by_containers().
+ * But this function adds code which defines struct mnt_namespace.
+ * To avoid having this structure twice, we define MNT_NAMESPACE_DEFINED in
+ * filter_by_containers(), then here we check if macro is already defined before
+ * adding struct definition.
  */
+#ifndef MNT_NAMESPACE_DEFINED
 struct mnt_namespace {
     // This field was removed in https://github.com/torvalds/linux/commit/1a7b8969e664d6af328f00fe6eb7aabd61a71d13
     #if LINUX_VERSION_CODE < KERNEL_VERSION(5, 11, 0)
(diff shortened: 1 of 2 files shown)


tools/mountsnoop.py
```diff
@@ -416,14 +433,19 @@ def main():
...
 
     mounts = {}
     umounts = {}
+    local_bpf_text = filter_by_containers(args) + bpf_text
     if args.ebpf:
-        print(bpf_text)
+        print(local_bpf_text)
         exit()
-    b = bcc.BPF(text=bpf_text)
+    b = bcc.BPF(text=local_bpf_text)
```
mauriciovasquezbernal, line 448: I don't think you need an additional variable. 

```    
bpf_text = filter_by_containers(args) + bpf_text
```
and use `bpf_text` everywhere else should work just fine.

mauriciovasquezbernal (approved):
Besides my comment on the variable name. LGTM! 
(I'm not an approver here so just approving to mark my review)

tools/mountsnoop.py
eiffel-fl: Thank you for the review.

I wanted to address it this evening, but I deleted minikube and my connection is quite unstable, so it is a bit hard to finish the `go mod download`...
I will try to correct it before Sunday.

yonghong-song:
[buildbot, test this please]

tools/mountsnoop.py
eiffel-fl: I tested the code you proposed and I remembered why I use the `local` thing because I got this error:
```
  File "/usr/share/bcc/tools/mountsnoop", line 474, in <module>
    main()
  File "/usr/share/bcc/tools/mountsnoop", line 444, in main
    bpf_text = filter_by_containers(args) + bpf_text
UnboundLocalError: local variable 'bpf_text' referenced before assignment
```

Indeed, since this code `bpf_text = filter_by_containers(args) + bpf_text` is in `main()`, python allocates a local variable called `bpf_text` which is used uninitialized, so I got an error.
So, using a local variable called `local_bpf_text` avoids this problem.
Finally, I solved it using:
```python
global bpf_text
bpf_text = filter_by_containers(args) + bpf_text
```
which seems to be a more pythonic way than having `local_bpf_text`.

yonghong-song:
[buildbot, test this please]

eiffel-fl:
Thank you for the merge!

PR #3603, opened 2021-09-04
libbpf-tools/syscount: use atomic_add for counter

fuweid:
Signed-off-by: Wei Fu <fuweid89@gmail.com>

### DIFF
--- a/libbpf-tools/syscount.bpf.c
+++ b/libbpf-tools/syscount.bpf.c
@@ -91,11 +91,11 @@ int sys_exit(struct trace_event_raw_sys_exit *args)
 	key = (count_by_process) ? pid : args->id;
 	val = bpf_map_lookup_or_try_init(&data, &key, &zero);
 	if (val) {
-		val->count++;
+		__sync_fetch_and_add(&val->count, 1);
 		if (count_by_process)
 			save_proc_name(val);
 		if (measure_latency)
-			val->total_ns += bpf_ktime_get_ns() - *start_ts;
+			__sync_fetch_and_add(&val->total_ns, bpf_ktime_get_ns() - *start_ts);
 	}
 	return 0;
 }


fuweid:
With atomic operation, the count result seems reliable.

But I still have question about this `syscount` tool: Can `syscount` guarantee
that the `sys_exit` is called for every single syscall call?

For non raw_tracepoint, the libbpf-tools will open perf_event and attach the
bpf prog to the event. The [include/trace/perf.h](https://github.com/torvalds/linux/blob/v5.11/include/trace/perf.h#L33-L71)
defines the `perf_trace_sys_exit` which registered into the
`__tracepoint_sys_exit->funcs`. When syscall exit, the `trace_sys_exit` will be
called, like:

```
trace_sys_exit
  preempt_disable_notrace
    __traceiter_sys_exit
      perf_trace_sys_exit
        perf_trace_run_bpf_submit
          trace_call_bpf
```

Based on v5.11 kernel, the `trace_call_bpf` is

```c
// https://github.com/torvalds/linux/blob/v5.11/kernel/trace/bpf_trace.c#L101
unsigned int trace_call_bpf(struct trace_event_call *call, void *ctx)
{
	unsigned int ret;

	cant_sleep();

	if (unlikely(__this_cpu_inc_return(bpf_prog_active) != 1)) {
		/*
		 * since some bpf program is already running on this cpu,
		 * don't call into another bpf program (same or different)
		 * and don't send kprobe event into ring-buffer,
		 * so return zero here
		 */
		ret = 0;
		goto out;
	}

	/*
	 * Instead of moving rcu_read_lock/rcu_dereference/rcu_read_unlock
	 * to all call sites, we did a bpf_prog_array_valid() there to check
	 * whether call->prog_array is empty or not, which is
	 * a heuristic to speed up execution.
	 *
	 * If bpf_prog_array_valid() fetched prog_array was
	 * non-NULL, we go into trace_call_bpf() and do the actual
	 * proper rcu_dereference() under RCU lock.
	 * If it turns out that prog_array is NULL then, we bail out.
	 * For the opposite, if the bpf_prog_array_valid() fetched pointer
	 * was NULL, you'll skip the prog_array with the risk of missing
	 * out of events when it was updated in between this and the
	 * rcu_dereference() which is accepted risk.
	 */
	ret = BPF_PROG_RUN_ARRAY_CHECK(call->prog_array, ctx, BPF_PROG_RUN);

 out:
	__this_cpu_dec(bpf_prog_active);

	return ret;
}
```

Does `bpf_prog_active` impact the `syscount`? In v5.11 kernel, the
raw_tracepoint doesn't increase the `bpf_prog_active`. Not sure that
`bpf_prog_active` will make that single counter bpf program not reliable.

yonghong-song:
[buildbot, test this please]

yonghong-song:
trace_call_bpf should be okay for sys_exit tracepoint and it is guaranteed to be called. The logic of bpf_prog_active is mostly to handled nested bpf calling in the context of interrupt, sys_exit tracepoint context cannot be in interrupt so it won't be skipped.

fuweid:
Thanks @yonghong-song !

PR #3605, opened 2021-09-06
refresh snapcraft rules

ColinIanKing:
Minor changes to add bcc commands to snapcraft instrall rules and add snapcraft architectures to build when using snapcraft remote-build


yonghong-song:
Thanks!

PR #3612, opened 2021-09-10
tools/ttysnoop: Fix KFUNC_PROBE to support new iov_iter signature

Sin4wd:
Kernel commit [1] used ->iter_type and ->data_source instead of ->type

[1] 8cd54c1c8480 iov_iter: separate direction from flavour

Signed-off-by: Sina Radmehr <sina_rad@hamravesh.com>

### DIFF
--- a/tools/ttysnoop.py
+++ b/tools/ttysnoop.py
@@ -138,9 +138,20 @@ KFUNC_PROBE(tty_write, struct kiocb *iocb, struct iov_iter *from)
 
     if (iocb->ki_filp->f_inode->i_ino != PTS)
         return 0;
-
+/**
+ * commit 8cd54c1c8480 iov_iter: separate direction from flavour 
+ * `type` is represented by iter_type and data_source seperately
+ */
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5, 14, 0)
     if (from->type != (ITER_IOVEC + WRITE))
         return 0;
+#else
+    if (from->iter_type != ITER_IOVEC)
+        return 0;
+    if (from->data_source != WRITE)
+        return 0;
+#endif
+
 
     kvec  = from->kvec;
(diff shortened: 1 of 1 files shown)


tools/ttysnoop.py
```diff
@@ -138,9 +138,20 @@ def usage():
...
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5, 14, 0)
     if (from->type != (ITER_IOVEC + WRITE))
         return 0;
+#else
+    if (from->iter_type != ITER_IOVEC)
+        return 0;
+    if (from->data_source != WRITE)
+        return 0;
+#endif
+
```
chenhengqi, line 154: Looks good.

How about also remove `buf` and `count`, they are defined but not used actually.

chenhengqi:
@Sin4wd  Could you please help testing #3613 on your 5.14 machine. Thanks.

yonghong-song:
[buildbot, test this please]

yonghong-song:
We used kernel version check to accommodate structure field change (deleting and adding). I checked the kernel source tree and didn't find a good symbol in nearby commits related to iov_iter, maybe because I am not really familiar with fs/iov_iter subsystem. Let us use the kernel version checking for now since the file actually has another one already. If somebody complains, we then will have a concrete case and may find some symbol making it work.

tools/ttysnoop.py
yonghong-song: Agree with @chenhengqi Since you are touching the code, maybe you can remove those unused variables.

yonghong-song:
[buildbot, ok to test]

yonghong-song:
@Sin4wd Hi, I want to cut a release for 5.14, so I will go ahead to merge this patch. If possible, could you send a followup patch later to remove unused variables, thanks!

PR #3618, opened 2021-09-14
Add FENTRY_PROBE/FEXIT_PROBE

chenhengqi:
I was prototyping some new tools with fentry using BCC, but the docs has nothing related to `fentry`.

After some research I found that fentry/fexit are implemented as kfunc/kretfunc in BCC. Actually, only this [issue](https://github.com/iovisor/bpftrace/issues/1833) mentioned this.

I think it's helpful to add this to BCC.

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -18,8 +18,10 @@ This guide is incomplete. If something feels missing, check the bcc and kernel s
         - [8. system call tracepoints](#8-system-call-tracepoints)
         - [9. kfuncs](#9-kfuncs)
         - [10. kretfuncs](#10-kretfuncs)
-        - [11. lsm probes](#11-lsm-probes)
-        - [12. bpf iterators](#12-bpf-iterators)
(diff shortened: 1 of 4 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
@chenhengqi bcc and bpftrace have used kfunc/kretfunc for quite sometimes and as you mentioned FENTRY_PROBE/FEXIT_PROBE is just an alias to KFUNC_PROBE/KRETFUNC_PROBE. I feel let us keep
KFUNC_PROBE/KRETFUNC_PROBE for now. As long as we have documentation to explain what it is, it should be okay. Most bcc users may not really be aware of what underlying technology and that is purpose for bcc to provide a little bit high-level abstraction. But your adding kfunc support for readahead.py is good. Maybe you can create a pull request for that?

chenhengqi:
OK, I am fine with that.

Will drop the first two commits.

PR #3624, opened 2021-09-16
threadsnoop: look for pthread_create in libc too

### LINKED ISSUE #3623: tools: threadsnoop doesn't work with glibc 2.34
In glibc 2.34, the features of several libraries, including libpthread, have been integrated in libc directly.
https://sourceware.org/pipermail/libc-alpha/2021-August/129718.html

libpthread.so.0 is now an empty stub:
$ nm /usr/lib64/libpthread.so.0 
nm: /usr/lib64/libpthread.so.0: no symbols

Changing the uprobe name to "c" does fixes the issue, but of course would break the tools for older version of glibc:
`b.attach_uprobe(name="c", sym="pthread_create", fn_name="do_entry")`

I wonder if there is some kind of indirection in libpthread that we can use.

jeromemarchand:
Since glibc 2.34, pthread features are integrated in libc directly.
Look for pthread_create there too when it is not found in libpthread.

Fixes #3623

### DIFF
--- a/tools/threadsnoop.py
+++ b/tools/threadsnoop.py
@@ -38,7 +38,12 @@ void do_entry(struct pt_regs *ctx) {
     events.perf_submit(ctx, &data, sizeof(data));
 };
 """)
-b.attach_uprobe(name="pthread", sym="pthread_create", fn_name="do_entry")
+
+# Since version 2.34, pthread features are integrated in libc
+try:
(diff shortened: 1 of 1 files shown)


yonghong-song:
[buildbot, test this please]

yonghong-song:
I think it is okay to do retry with first libpthread and then libc. This is simple and effective.

PR #3628, opened 2021-09-17
fix debian build settings for LLVM package names

gfx:
There's no `llvm-8.0`, but instead, it's `llvm-8` in Ubuntu.

Ubuntu 16.04 is affected by this change, choosing llvm-8 instead of llvm-3.8.


yonghong-song:
[buildbot, test this please]

yonghong-song:
Thanks, at some point, we should add more recent llvm compilers (up to llvm12).

PR #3633, opened 2021-09-23
ksnoop: remove duplicate include

davemarchevsky:
`ksnoop` is the only libbpf tool which is including both `<linux/bpf.h>` and `<bpf/bpf.h>` - the rest of the tools just include the latter

build fails for me because of redefinition errors as a result. Let's use `<bpf/bpf.h>` like the rest of the tools

### DIFF
--- a/libbpf-tools/ksnoop.c
+++ b/libbpf-tools/ksnoop.c
@@ -4,7 +4,6 @@
 #include <ctype.h>
 #include <errno.h>
 #include <getopt.h>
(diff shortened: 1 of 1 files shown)


davemarchevsky:
build-and-test will fail due to infra issues which i'm also investigating

chenhengqi (approved):
LGTM.

chenhengqi:
I test this with a clean VM, with or without `linux/bpf.h` both work.

PR #3634, opened 2021-09-23
tools/kvmexit: Display header after KeyboardInterrupt

chenhengqi:
When Ctrl+C is hit, `^C` messes up the output header. Fix that
by adding a blank line before printing. Also remove unused import
and signal handler.


yonghong-song:
[buildbot, test this please]

PR #3637, opened 2021-09-24
add ubuntu-20.04 to bcc-test.yml

davemarchevsky:
`publish.yml` is using 20.04, so let's add it here as well

Need to migrate this to match libbpf's CI eventually


yonghong-song:
The buildbot test failure is due to a flaky test. rerun will mostly fix it. I guess we should also fix that flaky test...

PR #3639, opened 2021-09-28
Update tcpv4connect.py

bozakov:
fix byte string comparison so we can run with python3


yonghong-song:
[buildbot, test this please]

PR #3645, opened 2021-10-05
Add function to print additional information when using `--containersmap`

eiffel-fl:
Hi.

In this PR, I added `print_additional_info_header` which basically prints:
```
NODE             NAMESPACE        POD          CONTAINER
```
when `--containersmap` is used.
It makes the code easier to change, as now if we want to modify header name, we just need to modify corresponding values in `containers.py`.

Best regards.


eiffel-fl:
Oh!
Sorry! I mistaken this repo (upstream) with our fork!

PR #3647, opened 2021-10-07
get the pid namespace by following task_active_pid_ns()

ChengyuanLiCY:
When unsharing of pid namespace, though nsproxy->pid_ns_for_children is
new, the process is still in the orignal pid namespace, only the forked
children processes will be in the new pid namespace.
So it's not correct to get process's pid namespace by nsproxy->pid_ns_for_children,
should get pid namespace by task_active_pid_ns() way, i.e.
pid->numbers[pid->level].ns

### DIFF
--- a/libbpf-tools/runqlat.bpf.c
+++ b/libbpf-tools/runqlat.bpf.c
@@ -48,6 +48,24 @@ int trace_enqueue(u32 tgid, u32 pid)
 	return 0;
 }
 
+static __always_inline unsigned int pid_namespace(struct task_struct *task)
+{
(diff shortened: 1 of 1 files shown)


yonghong-song:
@ChengyuanLiCY Thanks for the patch, I think you are correct. But let me dig into the kernel source a little bit and tried a couple of experiments to ensure it is correct. The original tool's result should mostly be correct too and most unsharing probably happened at bash level, which won't incur a lot of scheduling events. BTW, bcc python tools have the same issue.

Issue #2623, opened 2019-12-01
invalid indirect read from stack off -16+4 size 16 (when tracing spinlocks)


pavlix:
Hello! I jumped right into the hard job, tracing spinlocks. I have issues copying a pointer value from the argument of `_raw_spin_lock` to a `BPF_PERF_OUTPUT`.

```
#include <uapi/linux/ptrace.h>
#include <linux/irq.h>
#include <linux/spinlock.h>

struct event {
    int direction;
    u64 lock;
};

BPF_PERF_OUTPUT(events);

int kprobe___raw_spin_lock(struct pt_regs *ctx, u64 lock)
{
    struct event event = {
        .direction = 0,
        .lock = lock,
    };
    events.perf_submit(ctx, &event, sizeof event);
    bpf_trace_printk("lock %ld\n", lock);
    return 0;
}
```

This cannot be loaded to the kernel.

```
bpf: Failed to load program: Permission denied
; u64 lock = ctx->di;
0: (79) r6 = *(u64 *)(r1 +112)
; struct event event = {
1: (7b) *(u64 *)(r10 -8) = r6
2: (b7) r2 = 0
3: (63) *(u32 *)(r10 -16) = r2
last_idx 3 first_idx 0
regs=4 stack=0 before 2: (b7) r2 = 0
; bpf_perf_event_output(ctx, bpf_pseudo_fd(1, -1), CUR_CPU_IDENTIFIER, &event, sizeof event);
4: (18) r2 = 0xffff904e72e57600
6: (bf) r4 = r10
; 
7: (07) r4 += -16
; bpf_perf_event_output(ctx, bpf_pseudo_fd(1, -1), CUR_CPU_IDENTIFIER, &event, sizeof event);
8: (18) r3 = 0xffffffff
10: (b7) r5 = 16
11: (85) call bpf_perf_event_output#25
invalid indirect read from stack off -16+4 size 16
processed 10 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
```

Surprisingly, a trivial change from static initialization from the arguments to an individual assignment worked for me. Is there a reason for that or is this a bug?

The following modification works.

```
int kprobe___raw_spin_lock(struct pt_regs *ctx, u64 lock)
{
    struct event event = {
        .direction = 0,
    };
    event.lock = lock,
    events.perf_submit(ctx, &event, sizeof event);
    bpf_trace_printk("lock %ld\n", lock);
    return 0;
}
```

[...]

PR #3651, opened 2021-10-11
tools/capable: Set data to zero before setting fields.

eiffel-fl:
Hi.

In our fork, we added a field to `struct data_t`.
Thus, with the actual code and if all fields are not used (*e.g.* by not using `-U` and `-K` options), we would get BPF valider error like the following:
```
...
; bpf_perf_event_output(ctx, bpf_pseudo_fd(1, -1), CUR_CPU_IDENTIFIER, &data, sizeof(data));
74: (bf) r1 = r6
75: (18) r3 = 0xffffffff
77: (b7) r5 = 56
78: (85) call bpf_perf_event_output#25
invalid indirect read from stack off -80+44 size 56
processed 76 insns (limit 1000000) max_states_per_insn 0 total_states 6 peak_states 6 mark_read 6
```

In this commit, I added call to `__builtin_memset` to ensure `data` is full of 0, so unaligned access do not cause trouble anymore.
To be honest, this code is not really needed, as code inside `iovisor/bcc` works.
But I though it could be a good idea to submit you this modification as it makes the code safer (for example if one day a new field is added to the `struct data_t`).
Note that this comment put me on the way:
https://github.com/iovisor/bcc/issues/2623#issuecomment-560214481

If you see any way to improve this PR, feel free to share.

Best regards.

### DIFF
--- a/tools/capable.py
+++ b/tools/capable.py
@@ -182,7 +182,15 @@ int kprobe__cap_capable(struct pt_regs *ctx, const struct cred *cred,
     }
 
     u32 uid = bpf_get_current_uid_gid();
-    struct data_t data = {.tgid = tgid, .pid = pid, .uid = uid, .cap = cap, .audit = audit, .insetid = insetid};
+
+    struct data_t data = {};
+
+    data.tgid = tgid;
+    data.pid = pid;
+    data.uid = uid;
+    data.cap = cap;
+    data.audit = audit;
+    data.insetid = insetid;
 #ifdef KERNEL_STACKS
     data.kernel_stack_id = stacks.get_stackid(ctx, 0);
 #endif


tools/capable.py
```diff
@@ -182,7 +182,16 @@ def __getattr__(self, name):
...
+
+    struct data_t data;
+    __builtin_memset(&data, 0, sizeof(data));
+
+    data.tgid = tgid;
+    data.pid = pid;
+    data.uid = uid;
+    data.cap = cap;
+    data.audit = audit;
+    data.insetid = insetid;
```
chenhengqi, line 194: Yeah, adding a new field may introduce some hole in the struct, that is why the verifier complains.

I think `struct data_t data = {}` should be enough, right ?

eiffel-fl: Hum...
This is a bit strange because the previous code should have lead to unused fields being 0:
https://gcc.gnu.org/onlinedocs/gcc/Designated-Inits.html

I will investigate, I do not really know what is proper to C, to GCC, to BPF, to BCC and to Clang to give a correct answer about your suggestion.

tools/capable.py
```diff
@@ -182,7 +182,15 @@ def __getattr__(self, name):
     }
 
     u32 uid = bpf_get_current_uid_gid();
-    struct data_t data = {.tgid = tgid, .pid = pid, .uid = uid, .cap = cap, .audit = audit, .insetid = insetid};
+
+    struct data_t data = {};
```
yonghong-song, line 186: The C standard doesn't enforce compiler to initialize all memory occupied with the structure with 0. It only dictates all named fields to be initialized with 0 and all padding value are undefined and the compiler has the right leaving them with any value. But typically clang compiler will try to initialize the whole thing. So let us keep this change until/if it breaks.

PR #3652, opened 2021-10-11
libbpf-tools/mountsnoop: Fix example strings.

eiffel-fl:
Hi.

I modified `mountsnoop.c` in our fork and figured out the example strings were not correct.
I though it can be a good idea to upstream this at it will profit everyone.

Best regards.

### DIFF
--- a/libbpf-tools/mountsnoop.c
+++ b/libbpf-tools/mountsnoop.c
@@ -84,7 +84,7 @@ const char argp_program_doc[] =
 "\n"
 "EXAMPLES:\n"
 "    mountsnoop         # trace mount and umount syscalls\n"
-"    mountsnoop -v      # output vertically(one line per column value)\n"
+"    mountsnoop -d      # detailed output (one line per column value)\n"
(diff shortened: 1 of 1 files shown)


chenhengqi (approved):
LGTM. Thanks.

libbpf-tools/mountsnoop.c
```diff
@@ -84,7 +84,7 @@ const char argp_program_doc[] =
 "\n"
 "EXAMPLES:\n"
 "    mountsnoop         # trace mount and umount syscalls\n"
-"    mountsnoop -v      # output vertically(one line per column value)\n"
+"    mountsnoop -d      # detailled output (one line per column value)\n"
```
davemarchevsky, line 87: nit: detailled -> detailed

davemarchevsky (changes requested):
Aside from spelling nit, I think this makes sense

libbpf-tools/mountsnoop.c
eiffel-fl: Addressed! Thank you for the review!

PR #3655, opened 2021-10-12
tools/fdleak: Add FD leak monitor

vachelyang:
Trace and display over-threshold fds to detect FD leaks process.

`In UNIX-like, everything is a file`, file leakage problems may occur when
most resources are used improperly, eg:file, socket ...
It's a light and easy to use ebpf probe fd alloc
and monitor over-threshold fds process, collects stack and list open files.

### DIFF
--- a/tools/fdleak.py
+++ b/tools/fdleak.py
@@ -1,18 +1,20 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 #
-# fdleak    Trace and display fd alloc/close to detect FD leaks in user-mode
-#           processes.
+# fdleak    Trace and display fd alloc/close to detect FD leaks processes.
 #           For Linux, uses BCC,BPF. Embedded C.
 #
-# USAGE: fdleak [-h] [-p PID] [-w WARN] [--dumpstack] [--lsof] [-a]
+# USAGE: fdleak.py [-h] [-p PID] [--csv] [--lsof] [-i INTERVAL]
 #
-# If the survival time is longer than the minimum allowable survival time,
-# and the fd number opened by the process is keeps growing,that will trigger
-# fdleak to collect allocator information: user stack, max fd number,
-# thread name, tid.
+# Fdleak divided into 3 stages: loose L1, loose L2, strict L3
+# The first two stages check whether the allocation number of fd has been
+# increasing (for processes). If it keeps increasing, strict mode will be
+# triggered,strictly check the time of each allocation and close. If the
+# survivor is greater than 30s, strict inspection will print out the
+# monitored Process name, PID, Thread name, TID, Backtrace, Trace Hits
 #
 # This tool only works on Linux 4.6+.
+# The dump and plot argument need installed pandas and matplotlib module.
 #
 # # Copyright (c) 2021 Vachel Yang.
 # Licensed under the Apache License, Version 2.0 (the "License")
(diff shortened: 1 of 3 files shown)


tools/fdleak_example.txt
```diff
@@ -0,0 +1,113 @@
...
+  -w WARN, --warn WARN  set warning threshold,default is 819(1024*0.8)
+  --dumpstack           show alloc fd stack
+  --lsof                list of open files
+  -a, --showall         show all debug info
+
+examples:
+    ./fdleak                # trace all process alloc_fd()
+    ./fdleak -p 181         # only trace PID 181
+    ./fdleak -w 900         # set warning threshold 900
+    ./fdleak --dumpstack    # show alloc fd stack
```
netedwardwu, line 111: Sometimes the number of fd descriptor operations is so often, so dumpstack will output too much info that causes unorganized, one process has many different fd use cases. hundreds of processes in real-world case.

tools/fdleak.py
```diff
@@ -0,0 +1,175 @@
...
+};
+
+inline static void submit_to_user(int fds,struct pt_regs *ctx){
+    struct data_t data = {0};
+    data.pid = bpf_get_current_pid_tgid();
+    data.fds = fds;
+    bpf_get_current_comm(&data.taskname, sizeof(data.taskname));
+    data.stack_user = user_stack.get_stackid(ctx, BPF_F_USER_STACK | BPF_F_REUSE_STACKID);
+    data.stack_kernel = kernel_stack.get_stackid(ctx, 0);
+    events.perf_submit(ctx, &data, sizeof(data));
```
netedwardwu, line 85: Shouldn't use perf submit, because a lot of normal FD operations for production case.

tools/fdleak.py
```diff
@@ -0,0 +1,175 @@
...
+}
+
+int kretprobe____alloc_fd(struct pt_regs *ctx)
+{
+    int ret = PT_REGS_RC(ctx);
+
+    const int FILTER_THRESHOLD = ##FILTER_THRESHOLD## ;
+    ##FILTER_PID##
```
netedwardwu, line 93: Why not just FILTER_PID?

tools/fdleak_example.txt
```diff
@@ -0,0 +1,113 @@
...
+----
+USAGE message:
+usage: my_test.py [-h] [-p PID] [-w WARN] [--dumpstack] [--lsof] [-a]
+
+Trace FD leak
+
+optional arguments:
+  -h, --help            show this help message and exit
+  -p PID, --pid PID     trace this PID only
+  -w WARN, --warn WARN  set warning threshold,default is 819(1024*0.8)
```
netedwardwu, line 102: User won't know how to use this argument if you don't describe in example

netedwardwu (commented):
@brendangregg is right, have to think about "a lot of normal FDs for our production services"
And also yes, memleak is a good tool to follow.
Maybe you could use __close_fd as free operation.
memleak doesn't record the operation that BPF program didn't insert to kernel yet.
Should we care in fdleak?

Thanks

tools/fdleak_example.txt
```diff
@@ -0,0 +1,113 @@
...
+	kretprobe_trampoline+0x0 [kernel]
+	SYSC_accept4+0x9f [kernel]
+	SyS_accept+0x10 [kernel]
+	entry_SYSCALL_64_after_hwframe+0x41 [kernel]
+
+----
+^ `test_tcp_leak` open too much socket, trace at 'main+0x13 [test_tcp_leak]'
+----
+USAGE message:
+usage: my_test.py [-h] [-p PID] [-w WARN] [--dumpstack] [--lsof] [-a]
```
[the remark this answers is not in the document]
vachelyang: Sorry, I made a mistake, I will strengthen my own inspection.

tools/fdleak_example.txt
vachelyang: I studied the memleak monitor method and optimized a [new  fdleak version](https://github.com/vachelyang/bcc/commit/f9fbcea73b8b013c09b3b8c13688110ff415ea7e). 
Pls help to review it. And the automatic recycling method has not been resolved now, if you had better way, pls let me know,thx

This version removes those inefficient options. like this:
    ./fdleak                # trace all process alloc/close file descriptor
    ./fdleak -p 181         # only trace PID 181
    ./fdleak --lsof         # list the files opened by the monitor process
    ./fdleak -i 10          # interval to scan for outstanding allocations (in seconds), default is 5s
    ./fdleak -m 60          # minimum allowable survival time (in seconds), default is 30s

vachelyang:
> Am I right that this just prints FD allocs beyond 819? That's going to match a lot of normal FDs for our production services.
> 
> If you have a copy of my BPF book, please see pages 107-109, where I identify leaks as long duration FDs. I've used this successfully. As I said there, memleak(8) could be adapted for file descriptors.
> 
> An ideal FD leak tool could offer both approaches. I'd go with the long duration survivors approach by default, but an option could also be provided to do your threshold approach. The man page should discuss the pros and cons of each approach.

hello brendangregg 
first, i learned how to use your `memleak` tool, and update(or adpted) to [fdleak](https://github.com/vachelyang/bcc/commit/f9fbcea73b8b013c09b3b8c13688110ff415ea7e). it may had some problems also, i will continue to  optimize and learning
others, i found that memleak can't monitor the memory reclaimed by the system ,such as process is killed and exit without calling free ..
did you have any suggestions or methods~ thx~

tools/fdleak.py
vachelyang: This problem has also been fixed, replace  BPF_PERF_ARRAY with BPF_HASH. Thanks for pointing out the mistake.

tools/fdleak.py
```diff
@@ -0,0 +1,329 @@
...
+  char comm[TASK_COMM_LEN];
+};
+
+struct combined_alloc_info_t {
+  u64 last_longest_survivor;
+  int number_of_allocs;
+};
+
+BPF_STACK_TRACE(user_stack, 10240);
+BPF_HASH(fd_allocs_hash, u64, struct alloc_info_t, 1024000);
```
netedwardwu, line 74: Too much memory cost....is 1024000 a must?

tools/fdleak.py
```diff
@@ -0,0 +1,329 @@
...
+    bpf_text = bpf_text.replace("##PRINT_DEBUG##", "if(0)")
+
+if args.ebpf:
+    print(bpf_text)
+
+# load c text
+bpf = BPF(text=bpf_text)
+
+def get_pname(id):
+    comm_path=("/proc/%d/comm"%id)
```
netedwardwu, line 217: Did you test your code with pep8 for the coding style?
https://github.com/iovisor/bcc/blob/master/CONTRIBUTING-SCRIPTS.md => Step 12.

aaa.py:193:19: E226 missing whitespace around arithmetic operator
        t = t_name+"-"+str(tid)

tools/fdleak.py
```diff
@@ -0,0 +1,329 @@
+#!/usr/bin/python
```
netedwardwu, line 1: You missed another two files(README.md, man8), please check Brendan's lead
https://github.com/brendangregg/bcc/commit/9fa156273b395cfc5505f0fff5d6b7b1396f7daa

netedwardwu (commented):
I will help to test on ARM64 platform WHEN the code is ready. Thanks

tools/fdleak.py
```diff
@@ -0,0 +1,382 @@
...
+def find_allocinfo_by_stack_id(in_stack_id):
+    pidfd_allocs_hash = bpf["pidfd_allocs_hash"].items()
+    stack_traces = bpf["user_stack"]
+    process_alloc_dic = {}
+
+    for k, alloc_info in pidfd_allocs_hash:
+        if alloc_info.stack_id == in_stack_id:
+            if(BPF.monotonic_time() - alloc_info.timestamp_ns < min_allow_survival_ns):
+                continue
+            survival_time = 1+((int)(BPF.monotonic_time()
```
netedwardwu, line 320: E226 missing whitespace around arithmetic operator
            survival_time = 1+((int)(BPF.monotonic_time()
                             ^

tools/fdleak.py
```diff
@@ -0,0 +1,382 @@
...
+if args.pid != -1:
+    bpf_text = bpf_text.replace('##FILTER_PID##',
+        'if (bpf_get_current_pid_tgid() >> 32 != %s) { return 0; }' % args.pid)
+    print("start trace pid= %s" % args.pid)
+else:
+    bpf_text = bpf_text.replace('##FILTER_PID##', '')
+
+interval = args.interval
+min_allow_survival = args.min_allow
+min_allow_survival_ns = min_allow_survival*1000000000
```
netedwardwu, line 211: E226 missing whitespace around arithmetic operator
min_allow_survival_ns = min_allow_survival*1000000000

tools/fdleak.py
```diff
@@ -0,0 +1,405 @@
...
+  struct alloc_info_t info = {0};
+  u64 pidfd = bpf_get_current_pid_tgid() & 0xffffffff00000000L;
+  u32 pid = bpf_get_current_pid_tgid()>>32;
+  u32 tid = (u32)(bpf_get_current_pid_tgid() & 0xffffffffL);
+  struct combined_alloc_info_t *cinfo;
+  //other errors are not the monitor focus
+  if(ret_fd < 0 && ret_fd != -EMFILE)
+    return 0;
+
+  pidfd += (u32)ret_fd;
```
netedwardwu, line 132: You use TGID + fd as the key, what if the below case happens?

TGID = 5000, fd ret = 10, pidfd = 5010
TGID = 5008, fd ret = 2, pidfd = 5010

tools/fdleak.py
```diff
@@ -0,0 +1,405 @@
...
+
+static inline int fd_alloc_return(struct pt_regs *ctx) {
+  int ret_fd = PT_REGS_RC(ctx);
+  struct alloc_info_t info = {0};
+  u64 pidfd = bpf_get_current_pid_tgid() & 0xffffffff00000000L;
+  u32 pid = bpf_get_current_pid_tgid()>>32;
+  u32 tid = (u32)(bpf_get_current_pid_tgid() & 0xffffffffL);
+  struct combined_alloc_info_t *cinfo;
+  //other errors are not the monitor focus
+  if(ret_fd < 0 && ret_fd != -EMFILE)
```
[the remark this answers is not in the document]
vachelyang: Hanlde `-EMFILE` i think can make the user-level scanner display errors more obvious.

 ```
    def get_max_fd(self):
        # -24 is EMFILE, most of the time it means too many open files
        if(self.fd.count(-24) > 0):
            return "\033[1;31m-24\033[0m %d" % max(self.fd)
        else:
            return max(self.fd)
```

File: tools/fdleak.py [hunk 13]
```diff
@@ -0,0 +1,405 @@
...
+  u8 is_exit = 1;
+  pid_exit_hash.update(&pid, &is_exit);
+}
+
+static inline int fd_alloc_return(struct pt_regs *ctx) {
+  int ret_fd = PT_REGS_RC(ctx);
+  struct alloc_info_t info = {0};
+  u64 pidfd = bpf_get_current_pid_tgid() & 0xffffffff00000000L;
+  u32 pid = bpf_get_current_pid_tgid()>>32;
+  u32 tid = (u32)(bpf_get_current_pid_tgid() & 0xffffffffL);
```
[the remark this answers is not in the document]
vachelyang: The `pidfd` is u64, the upper 32 bits are `tgid`, and the lower 32 bits are `fd`. 
This is the unique key that describe that a process applies for a file descriptor.

tools/fdleak.py
```diff
@@ -0,0 +1,405 @@
...
+                     unsigned int fd)
+#endif
+{
+  ##FILTER_PID##
+  FD_MONITOR_TRIGGER(fd)
+  return fd_close_enter(fd);
+}
+
+// alloc_fd
+int kretprobe__get_unused_fd_flags(struct pt_regs *ctx)
```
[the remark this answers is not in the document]
vachelyang: You are right, `f_dupfd` does not call `get_unused_fd_flags`, so the fdleak misses `f_dupfd` in open file trace.. unfortunately, `close_fd` can be monitored close.
The reason why i don't monitor __alloc_fd is [`__alloc_fd` will merged to *static* alloc_fd](https://github.com/torvalds/linux/commit/aa384d1) . it will not be extened by fstable.  
It seems need to add a  trace of `f_dupfd`, is that right?

tools/fdleak.py
```diff
@@ -0,0 +1,405 @@
...
+  u32 tid = (u32)(bpf_get_current_pid_tgid() & 0xffffffffL);
+  struct combined_alloc_info_t *cinfo;
+  //other errors are not the monitor focus
+  if(ret_fd < 0 && ret_fd != -EMFILE)
+    return 0;
+
+  pidfd += (u32)ret_fd;
+  info.timestamp_ns = bpf_ktime_get_ns();
+  info.stack_id =
+      user_stack.get_stackid(ctx, BPF_F_USER_STACK | BPF_F_REUSE_STACKID);
```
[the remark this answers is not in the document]
vachelyang: I simply thought that the kernel thread would not have complex fd operations (compare memory). 
I had not encountered it before developing on android..Let me think about it and do some experiment.

tools/fdleak.py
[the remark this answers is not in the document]
vachelyang: Okay, I will add this. Thanks for your refine.

netedwardwu:
About you use tgid fd as key...I'm not sure it's okay...

When process fork(), it will call copy_process() => dup_fd() to duplicate fd from TGID
So fd number is double in that TGID, but current fdleak doesn't handle that...so "MAX_FD" in fdleak will be incorrect?

Am I right?
Or you already handle such scenario?
Thanks

tools/fdleak_example.txt
```diff
@@ -0,0 +1,140 @@
...
+lr-x------ 1 vachel.yang vachel.yang 64 oct 16 17:49 123 -> /dev/null
+lr-x------ 1 vachel.yang vachel.yang 64 oct 16 17:49 124 -> /dev/null
+lr-x------ 1 vachel.yang vachel.yang 64 oct 16 17:49 125 -> /dev/null
+lr-x------ 1 vachel.yang vachel.yang 64 oct 16 17:49 126 -> /dev/null
+lr-x------ 1 vachel.yang vachel.yang 64 oct 16 17:49 127 -> /dev/null
+
+---->
+^ fdleak thread open too much `/dev/null`
+----
+Currently i don't know how to deal with the case that recycling program resources by system `do_exit` -> `exit_files` .
```
netedwardwu, line 97: Yes! This is a real problem...can we kprobe filp_close() to handle sources from __close_fd and also exit_files?

PR #3665, opened 2021-10-17
bcc: Allow KFUNC_PROBE to instrument function without arguments

chenhengqi:
Update KFUNC_PROBE and its family to allow instrument kernel
function without specifying arguments. Sometimes, we don't need
to bookkeep arguments at function entry, just store a timestamp.
This fix would allow this use case.


davemarchevsky:
[buildbot, test this please]

davemarchevsky:
[buildbot, ok to test]

PR #3671, opened 2021-10-21
Fix LINUX_KERNEL_VERSION check for TP of block_rq_insert/block_rq_issue

ChengyuanLiCY:
The TP_PROTO for block_rq_insert and block_rq_issue are still
TP_PROTO(struct request_queue *q, struct request *rq) in kernel version
5.10.x which is bigger than 5.10.0, so the version check should be:
LINUX_KERNEL_VERSION < KERNEL_VERSION(5, 11, 0)

Signed-off-by: Li Chengyuan chengyuanli@hotmail.com


chenhengqi:
I am inclined to fix all these kernel version checks in a CO-RE way.

PR #3673, opened 2021-10-23
libbpf-tools: Fix memory leaks in ksnoop/gethostlatency

### LINKED ISSUE #3664: Memory leaks in libbpf tools
Some of the libbpf tools have memory leak problems. This is because we have one BPF program but attach it to multiple functions, each attaching creates a new `struct bpf_link *` and these `bpf_link`s are not freed by `xxx_bpf__destroy`.

Take `gethostlatency` as an example:

https://github.com/iovisor/bcc/blob/44fc17fc8ca0a53f37e82aa82a6a000ec28384c4/libbpf-tools/gethostlatency.c#L152-L154

https://github.com/iovisor/bcc/blob/44fc17fc8ca0a53f37e82aa82a6a000ec28384c4/libbpf-tools/gethostlatency.c#L174-L176

The first assignment to `handle_entry` is leaked.

The fix can be done by updating the tool itself. But I wonder whether we should also update bpftool to generate code to cover these use cases.

cc @anakryiko  @davemarchevsky

chenhengqi:
There are memory leaks when attach a BPF program to multiple
targets in these tools. This is because we misuse the
bpf_program__attach_kprobe function, the returned struct bpf_link
object is not freed after use. Closes #3664.

### DIFF
--- a/libbpf-tools/gethostlatency.c
+++ b/libbpf-tools/gethostlatency.c
@@ -133,7 +133,7 @@ static int get_libc_path(char *path)
 	return -1;
 }
 
-static int attach_uprobes(struct gethostlatency_bpf *obj)
+static int attach_uprobes(struct gethostlatency_bpf *obj, struct bpf_link *links[])
 {
 	int err;
 	char libc_path[PATH_MAX] = {};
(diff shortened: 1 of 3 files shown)


davemarchevsky:
[buildbot, test this please]

davemarchevsky:
[buildbot, ok to test]

PR #3688, opened 2021-11-03
libbpf-tools: add klockstat

brho:
This is a port of BCC's klockstat.  Differences from BCC:
- can specify a lock by ksym name, using -L
- tracks whichever task had the max time for acquire and hold, outputted
when -s > 1 (otherwise it's cluttered).
- does not reset stats each interval by default.  Can request with -R.

-------------
Usage: klockstat [-hRT] [-p PID] [-t TID] [-c FUNC] [-L LOCK] [-n NR_LOCKS]
                 [-s NR_STACKS] [-S SORT] [-d DURATION] [-i INTERVAL]

  -p, --pid=PID              Filter by process ID
  -t, --tid=TID              Filter by thread ID
  -c, --caller=FUNC          Filter by caller string prefix
  -L, --lock=LOCK            Filter by specific ksym lock name
  -n, --locks=NR_LOCKS       Number of locks to print
  -s, --stacks=NR_STACKS     Number of stack entries to print per lock
  -S, --sort=SORT            Sort by field:
                               acq_[max|total|count]
                               hld_[max|total|count]
  -d, --duration=SECONDS     Duration to trace
  -i, --interval=SECONDS     Print interval
  -R, --reset                Reset stats each interval
  -T, --timestamp            Print timestamp

  -?, --help                 Give this help list
      --usage                Give a short usage message
  -V, --version              Print program version

Mandatory or optional arguments to long options are also mandatory or optional
for any corresponding short options.

Examples:
  klockstat                     # trace system wide until ctrl-c
  klockstat -d 5                # trace for 5 seconds
  klockstat -i 5                # print stats every 5 seconds
  klockstat -p 181              # trace process 181 only
  klockstat -t 181              # trace thread 181 only
  klockstat -c pipe_            # print only for lock callers with 'pipe_'
                                # prefix
  klockstat -L cgroup_mutex     # trace the cgroup_mutex lock only
  klockstat -S acq_count        # sort lock acquired results by acquire count
  klockstat -S hld_total        # sort lock held results by total held time
  klockstat -S acq_count,hld_total  # combination of above
  klockstat -n 3                # display top 3 locks
  klockstat -s 6                # display 6 stack entries per lock

-------------

Signed-off-by: Barret Rhoden <brho@google.com>

### DIFF
--- a/libbpf-tools/klockstat.bpf.c
+++ b/libbpf-tools/klockstat.bpf.c
@@ -1,0 +1,230 @@
+// SPDX-License-Identifier: GPL-2.0
+/* Copyright (c) 2021 Google LLC.
+ *
+ * Based on klockstat from BCC by Jiri Olsa and others
+ * 2021-10-26   Barret Rhoden   Created this.
+ */
+#include "vmlinux.h"
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include "klockstat.h"
+#include "bits.bpf.h"
+
+const volatile pid_t targ_tgid = 0;
+const volatile pid_t targ_pid = 0;
+struct mutex *const volatile targ_lock = NULL;
+
+struct {
+	__uint(type, BPF_MAP_TYPE_STACK_TRACE);
+	__uint(max_entries, MAX_ENTRIES);
+	__uint(key_size, sizeof(u32));
+	__uint(value_size, PERF_MAX_STACK_DEPTH * sizeof(u64));
+} stack_map SEC(".maps");
+
+/*
+ * Uniquely identifies a task grabbing a particular lock; a task can only hold
+ * the same lock once (non-recursive mutexes).
+ */
+struct task_lock {
+	u64 task_id;
+	u64 lock_ptr;
+};
+
+struct lockholder_info {
+	s32 stack_id;
+	u64 task_id;
+	u64 try_at;
+	u64 acq_at;
+	u64 rel_at;
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_ENTRIES);
+	__type(key, struct task_lock);
+	__type(value, struct lockholder_info);
+} lockholder_map SEC(".maps");
+
+/*
+ * Keyed by stack_id.
+ *
+ * Multiple call sites may have the same underlying lock, but we only know the
+ * stats for a particular stack frame.  Multiple tasks may have the same
+ * stackframe.
+ */
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_ENTRIES);
+	__type(key, s32);
+	__type(value, struct lock_stat);
+} stat_map SEC(".maps");
+
+static bool tracing_task(u64 task_id)
+{
+	u32 tgid = task_id >> 32;
+	u32 pid = task_id;
+
+	if (targ_tgid && targ_tgid != tgid)
+		return false;
+	if (targ_pid && targ_pid != pid)
+		return false;
+	return true;
+}
+
+static void lock_contended(void *ctx, struct mutex *lock)
+{
+	u64 task_id;
+	struct lockholder_info li[1] = {0};
+	struct task_lock tl = {};
+
+	if (targ_lock && targ_lock != lock)
+		return;
+	task_id = bpf_get_current_pid_tgid();
+	if (!tracing_task(task_id))
+		return;
+
+	li->task_id = task_id;
+	/*
+	 * Skip 4 frames, e.g.:
+	 *       __this_module+0x34ef
+	 *       __this_module+0x34ef
+	 *       __this_module+0x8c44
+	 *             mutex_lock+0x5
+	 *
+	 * Note: if you make major changes to this bpf program, double check
+	 * that you aren't skipping too many frames.
+	 */
+	li->stack_id = bpf_get_stackid(ctx, &stack_map,
+				       4 | BPF_F_FAST_STACK_CMP);
+	/* Legit failures include EEXIST */
+	if (li->stack_id < 0)
+		return;
+	li->try_at = bpf_ktime_get_ns();
+
+	tl.task_id = task_id;
+	tl.lock_ptr = (u64)lock;
+	bpf_map_update_elem(&lockholder_map, &tl, li, BPF_ANY);
+}
+
+static void lock_acquired(struct mutex *lock)
+{
+	u64 task_id;
+	struct lockholder_info *li;
+	struct task_lock tl = {};
+
+	if (targ_lock && targ_lock != lock)
+		return;
+	task_id = bpf_get_current_pid_tgid();
+	if (!tracing_task(task_id))
+		return;
+
+	tl.task_id = task_id;
+	tl.lock_ptr = (u64)lock;
+	li = bpf_map_lookup_elem(&lockholder_map, &tl);
+	if (!li)
+		return;
+
+	li->acq_at = bpf_ktime_get_ns();
+}
+
+static void account(struct lockholder_info *li)
+{
+	struct lock_stat *ls;
+	u64 delta;
+
+	/*
+	 * Multiple threads may have the same stack_id.  Even though we are
+	 * holding the lock, dynamically allocated mutexes can have the same
+	 * callgraph but represent different locks.  They will be accounted as
+	 * the same lock, which is what we want, but we need to use atomics to
+	 * avoid corruption, especially for the total_time variables.
+	 */
+	ls = bpf_map_lookup_elem(&stat_map, &li->stack_id);
+	if (!ls) {
+		struct lock_stat fresh = {0};
+
+		bpf_map_update_elem(&stat_map, &li->stack_id, &fresh, BPF_ANY);
+		ls = bpf_map_lookup_elem(&stat_map, &li->stack_id);
+		if (!ls)
+			return;
+	}
+
+	delta = li->acq_at - li->try_at;
+	__sync_fetch_and_add(&ls->acq_count, 1);
+	__sync_fetch_and_add(&ls->acq_total_time, delta);
+	if (delta > READ_ONCE(ls->acq_max_time)) {
+		WRITE_ONCE(ls->acq_max_time, delta);
+		WRITE_ONCE(ls->acq_max_id, li->task_id);
+		/*
+		 * Potentially racy, if multiple threads think they are the max,
+		 * so you may get a clobbered write.
+		 */
+		bpf_get_current_comm(ls->acq_max_comm, TASK_COMM_LEN);
+	}
+
+	delta = li->rel_at - li->acq_at;
+	__sync_fetch_and_add(&ls->hld_count, 1);
+	__sync_fetch_and_add(&ls->hld_total_time, delta);
+	if (delta > READ_ONCE(ls->hld_max_time)) {
+		WRITE_ONCE(ls->hld_max_time, delta);
+		WRITE_ONCE(ls->hld_max_id, li->task_id);
+		bpf_get_current_comm(ls->hld_max_comm, TASK_COMM_LEN);
+	}
+}
+
+static void lock_released(struct mutex *lock)
+{
+	u64 task_id;
+	struct lockholder_info *li;
+	struct task_lock tl = {};
+
+	if (targ_lock && targ_lock != lock)
+		return;
+	task_id = bpf_get_current_pid_tgid();
+	if (!tracing_task(task_id))
+		return;
+	tl.task_id = task_id;
+	tl.lock_ptr = (u64)lock;
+	li = bpf_map_lookup_elem(&lockholder_map, &tl);
+	if (!li)
+		return;
+
+	li->rel_at = bpf_ktime_get_ns();
+	account(li);
+
+	bpf_map_delete_elem(&lockholder_map, &tl);
+}
+
+SEC("fentry/mutex_lock")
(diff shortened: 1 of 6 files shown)


libbpf-tools/klockstat.bpf.c
```diff
@@ -0,0 +1,265 @@
...
+#include <bpf/bpf_tracing.h>
+#include "klockstat.h"
+#include "bits.bpf.h"
+
+const volatile pid_t targ_tgid = 0;
+const volatile pid_t targ_pid = 0;
+struct mutex *const volatile targ_lock = NULL;
+
+/* Make sure all programs are loaded before we start tracing */
+bool enabled = false;
```
chenhengqi, line 19: Why we need this variable ?

libbpf-tools/klockstat.bpf.c
```diff
@@ -0,0 +1,265 @@
...
+		return false;
+	return true;
+}
+
+static void lock_contended(void *ctx, struct mutex *lock)
+{
+	u64 task_id;
+	struct task_info *ti;
+	struct lockholder_info li[1] = {0};
+	struct depth_id di[1] = {0};
```
chenhengqi, line 92: Don't think you need this single element array, `struct foo bar = {}` should be enough.

libbpf-tools/klockstat.bpf.c
```diff
@@ -0,0 +1,265 @@
...
+SEC("tp_btf/sched_switch")
+int BPF_PROG(sched_switch, bool preempt, struct task_struct *prev,
+	     struct task_struct *next)
+{
+	#define TASK_DEAD 0x0080
+	/*
+	 * Need CORE_READ in a conditional for now:
+	 * https://lore.kernel.org/bpf/20200630171240.2523722-1-yhs@fb.com/
+	 */
+	if (BPF_CORE_READ(prev, state) != TASK_DEAD)
```
chenhengqi, line 259: Please use https://github.com/iovisor/bcc/blob/master/libbpf-tools/core_fixes.bpf.h#L20 here.

davemarchevsky:
[buildbot, test this please]

davemarchevsky:
[buildbot, ok to test]

libbpf-tools/klockstat.bpf.c
brho: there are a couple of mutexes that are grabbed while the three programs are inserted, and they lead to weird results.  i think the lock bookkeeping gets thrown off when we aren't tracing all three lock-entry/lock-exit/unlock in sync.  we'll think a task is at a certain depth, but not see the unlock.  later on, we see an unlock that we erroneously pair with an older lock.

specifically, this backtrace appears as if a mutex was held for the duration of klockstat.  this was taken from a 4 second duration.  the mutex wasn't actually held 4 seconds, but we thought it was.
```
           modify_ftrace_direct+0x18b 4002238347        1 4002238347   4002238347
          bpf_trampoline_update+0x427
       bpf_trampoline_link_prog+0x133
        bpf_tracing_prog_attach+0x26f
                      SYSC_bpf+0x2b3a
                              Max PID 2028, COMM klockstat
```

and if you periodically sample (with -i), you get some other bonkers results:

```
                               Caller  Avg Spin    Count   Max Spin   Total Spin
           modify_ftrace_direct+0x18b 6148913868173880962        3 18446741604521630976 18446741604521642888
          bpf_trampoline_update+0x427
       bpf_trampoline_link_prog+0x133
        bpf_tracing_prog_attach+0x26f
                      SYSC_bpf+0x2b3a
                              Max PID 2067, COMM klockstat
```

i'll add a comment to the source.

libbpf-tools/klockstat.bpf.c
brho: i do that so you can use the variable as a pointer.  otherwise you'd have 'li' being accessed with '.' in some functions and '->' in others.

though di is always local to the function, so i changed that.

i can change it for 'li' too if you want.

chenhengqi: It's ok. I don't have strong opinion about this.
Someone does this before, see: https://github.com/iovisor/bcc/blob/4d08181d194bc2961f5d3e599f6db5d870648d15/libbpf-tools/uprobe_helpers.c#L232-L234

libbpf-tools/klockstat.bpf.c
chenhengqi: I create a new PR #3691 to resolve what you concern here. So, no worry.

For the two questions:
* We include `vmlinux.h` which has `preserve_access_index` applied and we use that in `fentry` context, which allows kernel pointer chasing, so it's OK.
* This is compiler(clang) and loader(libbpf) magic. From libbpf's perspective, `struct task_struct` and `struct task_struct___x ` are the same, and `bpf_core_field_exists` is not a function call.

Please refer to Andrii's blog: https://nakryiko.com/posts/bpf-core-reference-guide/

libbpf-tools/klockstat.bpf.c
```diff
@@ -0,0 +1,276 @@
...
+#include "bits.bpf.h"
+#include "core_fixes.bpf.h"
+
+const volatile pid_t targ_tgid = 0;
+const volatile pid_t targ_pid = 0;
+struct mutex *const volatile targ_lock = NULL;
+
+/*
+ * Make sure all programs are loaded before we start tracing so that we see
+ * lock/lock_acquried/unlock for all locks.  Otherwise, the bookkeeping can be
```
chenhengqi, line 21: Don't think this is right. Even if we have all BPF programs loaded, we can also see partial lock/unlock event sequence.
What we often do is filter out those event, like: https://github.com/iovisor/bcc/blob/4d08181d194bc2961f5d3e599f6db5d870648d15/tools/tcpconnect.py#L164-L166

brho: i think the issue is that the lock-acquired or lock-unlocked events are getting paired up with the wrong lock-contended, possibly due to a mismatch with the depth.  the code already aborts if it can't find a lockholder_info for a given depth, but the issues could arise if you use the wrong depth.  

TBH, i haven't traced out the exact line sequence of events.  i see a bunch of locks grabbed in the kernel when the bpf program is inserted, but i don't know exactly when the fentry/fexit programs start to run in all of that code.

FWIW, the original klockstat.py also used an enabled flag, so they might have had the same issue, especially considering this takes the same approach (tracking via task->depth).   though they also used it when extracting stats.

brho: here's what i think happens: if you trace mutex_lock, but not unlock, and then grab and release a bunch of locks, then we'll increment task->depth a bunch of times.  once we start tracing unlock, then all future locks will be paired with an unlock.  those initial unpaired struct lockholder_info will just sit around, and the task->depth will always appear to be holding more locks than it does.

eventually, we'll have some mutex_unlock that wasn't paired with a traced call.  in my original code, this was from mutex_trylock.  there are some other places in the kernel where you can grab a mutex without calling mutex_lock or trylock.  when those callsites unlock, we'll trace it and pair it up with one of those initial locks from way back when klockstat set up.  that's how i was able to get a trace where it thought a lock was held for several seconds.

once a task->depth hits 0, we'll ignore any extra 'unbalanced' unlocks.  it's possible that we'll get some incorrect results, such as 
  mutex_lock X
    weird_untraced_mutex_lock Y
    mutex_unlock Y
  mutex_unlock X

the unlock of Y will be paired with the lock of X, and the unlock of X will be ignored.  this is better than having that unlock of X pair with some mutex_lock from a long time ago.

brho:
just checking if there's anything else needed on this PR.

chenhengqi:
> just checking if there's anything else needed on this PR.

Sorry, I missed the GitHub notification. Will take a look at this again this week.

brho:
hi - 

two updates on this in my recent change (9 Dec):

1) instead of tracking 'lock depth' as a lookup (which was interpreted as "task X's Nth lock acquired"), do lookups by task_id + lock_ptr, which is "Task X grabbing lock Y".  

the biggest benefit is that we don't have "mismatched" lock/unlock pairs.  so if a task grabs mutex A, then mutex B, it can unlock A before unlocking B.  if you tracked by "depth", you'd just assume the first unlocked was the last to be locked.

an ancillary benefit is we don't have to worry about having unlock get paired with "ancient" locks, which means we don't need that "enabled" flag, or the bulk of comments justifying it.

finally, it cleans up the code a bit: one less MAP to maintain, since we no longer care about a task's "depth".

2) use atomics or other shared memory sync when accessing a lockstat structure.  these are keyed by a callstack.  although callstacks identify a line of code, e.g. vfs_read+0x12, they do not uniquely identify a lock.  so you can have multiple threads working on the same struct in account().  i put the WRITE_ONCE/READ_ONCE in bits.bpf.h, but that could be moved elsewhere.  

thanks!

libbpf-tools/klockstat.h
```diff
@@ -0,0 +1,19 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+#pragma once
```
chenhengqi, line 2: Consider replace this `#pragma once` with `#ifndef` guard.

libbpf-tools/klockstat.h
```diff
@@ -0,0 +1,19 @@
...
+
+#define MAX_ENTRIES 102400
+#define TASK_COMM_LEN 16
+#define PERF_MAX_STACK_DEPTH 127
+
+struct lock_stat {
+	uint64_t acq_count;
```
chenhengqi, line 9: u64 or __u64 ?

libbpf-tools/klockstat.bpf.c
```diff
@@ -0,0 +1,230 @@
...
+		if (!ls)
+			return;
+	}
+
+	delta = li->acq_at - li->try_at;
+	__sync_fetch_and_add(&ls->acq_count, 1);
+	__sync_fetch_and_add(&ls->acq_total_time, delta);
+	if (delta > READ_ONCE(ls->acq_max_time)) {
+		WRITE_ONCE(ls->acq_max_time, delta);
+		WRITE_ONCE(ls->acq_max_id, li->task_id);
```
chenhengqi, line 158: Why do you need these `READ_ONCE` and `WRITE_ONCE`s ?

libbpf-tools/klockstat.c
```diff
@@ -0,0 +1,557 @@
...
+"  klockstat -L cgroup_mutex     # trace the cgroup_mutex lock only\n"
+"  klockstat -S acq_count        # sort lock acquired results by acquire count\n"
+"  klockstat -S hld_total        # sort lock held results by total held time\n"
+"  klockstat -S acq_count,hld_total  # combination of above\n"
+"  klockstat -n 3                # display top 3 locks\n"
+"  klockstat -s 6                # display 6 stack entries per lock\n"
+;
+
+static const struct argp_option opts[] = {
+	{ "pid", 'p', "PID", 0, "Filter by process ID"},
```
chenhengqi, line 84: Add a space before `}`.

libbpf-tools/klockstat.c
```diff
@@ -0,0 +1,557 @@
...
+}
+
+static char *symname(struct ksyms *ksyms, uint64_t pc, char *buf, size_t n)
+{
+	const struct ksym *ksym = ksyms__map_addr(ksyms, pc);
+
+	if (!ksym)
+		return "Unknown";
+	snprintf(buf, n, "%s+0x%lx", ksym->name, pc - ksym->addr);
+	buf[n - 1] = 0;
```
chenhengqi, line 313: snprintf already handle this, no ?

libbpf-tools/klockstat.c
```diff
@@ -0,0 +1,557 @@
...
+{
+	char buf[40];
+
+	printf("%37s %9lu %8lu %10lu %12lu\n",
+	       symname(ksyms, ss->bt[0], buf, sizeof(buf)),
+	       ss->ls.acq_total_time / ss->ls.acq_count,
+	       ss->ls.acq_count,
+	       ss->ls.acq_max_time,
+	       ss->ls.acq_total_time);
+	for (int i = 1; i < nr_stack_entries; i++) {
```
chenhengqi, line 333: Declare this `i` first to follow the kernel code style. Same for the five for loop below.

libbpf-tools/klockstat.h
brho: Done

libbpf-tools/klockstat.h
brho: Done

libbpf-tools/klockstat.bpf.c
brho: in part, for documentation that those are read from and written to concurrently.  that's what i'd do for in-kernel code.  

the bpf compiler might make those u64 accesses atomic, but i don't know if that is a guarantee, and i'd rather not have issues with load/store tearing.

libbpf-tools/klockstat.c
brho: Done

libbpf-tools/klockstat.c
brho: done

libbpf-tools/klockstat.c
brho: probably.  the man page doesn't say anything about that.  https://pubs.opengroup.org/onlinepubs/9699919799/functions/snprintf.html says it does.  my workstation's glibc does.  i've been burnt in the past with weird snprintfs, but i guess times have changed.  =)

brho:
pushed fixes.  thanks!

brho:
just checking if this needs anything else.  thanks.

davemarchevsky:
Sorry for the merge delay here.

I would normally push back against the change to default behavior here, but I can see why it makes sense and am not an active user of `klockstat` so don't think it's a good idea to hold this up further.

Did a quick followup PR (#3796) adding "Differences from BCC" comment.

Issue #3687, opened 2021-11-03
cachestat broken by recent kernel change


chenhengqi:
The BCC tool and libbpf-tool cachestat are broken on latest kernel.
Kernel commit [203a31516616 ("mm/writeback: Add __folio_mark_dirty()")](https://github.com/torvalds/linux/commit/203a31516616111b8eaaf00c16fa3fcaa25e6f81)
converted `account_page_dirtied` to `folio_account_dirtied`, so the
attach target of these tools disppeared, making them broken.

```shell
$ sudo ./cachestat                                 
libbpf: failed to find kernel BTF type ID of 'account_page_dirtied': -3
libbpf: prog 'account_page_dirtied': failed to prepare load attributes: -3
libbpf: failed to load program 'account_page_dirtied'
libbpf: failed to load object 'cachestat_bpf'
libbpf: failed to load BPF skeleton 'cachestat_bpf': -3
failed to open and/or load BPF object
```

```shell
sudo ./cachestat.py 
cannot attach kprobe, probe entry may not exist
Traceback (most recent call last):
  File "./cachestat.py", line 99, in <module>
    b.attach_kprobe(event="account_page_dirtied", fn_name="do_count")
  File "/usr/lib/python2.7/dist-packages/bcc/__init__.py", line 833, in attach_kprobe
    (fn_name, event))
Exception: Failed to attach BPF program do_count to kprobe account_page_dirtied
```

chenhengqi:
BCC tool cachetop has the same problem.

chenhengqi:
Maybe this problem date back to kernel v5.14 with https://github.com/torvalds/linux/commit/6e1cae881a0646f31fe2bda90297d820da1137eb .

yonghong-song:
Thanks @chenhengqi I will take a look at this soon to see whether we can make a compatible change without changing tool's functionality.

ShawnZhong:
I guess this issue is fixed by https://github.com/iovisor/bcc/pull/3747 & https://github.com/iovisor/bcc/pull/3692 and can be closed?

PR #3692, opened 2021-11-08
tools: fix cachestat.py with 5.15 kernel

yonghong-song:
Fix issue #3687.
The tool cachestat.py doesn't work with 5.15 kernel due to
kprobe function renaming. Adapt to the new function.
Also added a comment that static functions might
get inlined and the result may not be accurate if this happens.
More work can be done in the future to make the tool
more robust.

### DIFF
--- a/tools/cachestat.py
+++ b/tools/cachestat.py
@@ -96,7 +96,15 @@ if debug or args.ebpf:
 b = BPF(text=bpf_text)
 b.attach_kprobe(event="add_to_page_cache_lru", fn_name="do_count")
 b.attach_kprobe(event="mark_page_accessed", fn_name="do_count")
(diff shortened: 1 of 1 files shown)


chenhengqi:
I forgot to update the issue title, would you fix `cachetop` in the same commit ?

https://github.com/iovisor/bcc/issues/3687#issuecomment-958672755

yonghong-song:
> I forgot to update the issue title, would you fix `cachetop` in the same commit ?
> 
> [#3687 (comment)](https://github.com/iovisor/bcc/issues/3687#issuecomment-958672755)

Sorry. Just found this when I tried to clean up emails. Will send a separate commit to fix cachetop issue.

PR #3694, opened 2021-11-09
Fix garbled java class name problem of uobjnew.py

D-D-H:
Hi,

Please help review this patch that fixes a garbled java class name problem of uobjnew.py.

Without this patch, the output may look like this:
```
NAME/TYPE                      # ALLOCS      # BYTES
Test                                  1           16
[B��;*                               1         1016
[Ljava/lang/Thread;                   1         4016
[Ljava/lang/String;                   1         4016
[Ljava/lang/Float;%                   1         4016
[D���                               1         8016
```

Apply this patch:
```
NAME/TYPE                      # ALLOCS      # BYTES
Test                                  1           16
[B                                    1         1016
[Ljava/lang/String;                   1         4016
[Ljava/lang/Thread;                   1         4016
[Ljava/lang/Float;                    1         4016
[D                                    1         8016

```

In this patch, I read the third parameter twice with different types. The reason for this is for compatibility reasons: the declaration and definition of this probe in OpenJDK use different types:

- declaration: https://github.com/openjdk/jdk/blob/master/src/hotspot/os/posix/dtrace/hotspot.d#L75
- definition: https://github.com/openjdk/jdk/blob/master/src/hotspot/share/runtime/sharedRuntime.cpp#L1018
  - name->utf8_length() return a int value

### DIFF
--- a/tools/lib/uobjnew.py
+++ b/tools/lib/uobjnew.py
@@ -96,9 +96,11 @@ int alloc_entry(struct pt_regs *ctx) {
     struct key_t key = {};
     struct val_t *valp, zero = {};
     u64 classptr = 0, size = 0;
+    u32 length = 0;
     bpf_usdt_readarg(2, ctx, &classptr);
+    bpf_usdt_readarg(3, ctx, &length);
     bpf_usdt_readarg(4, ctx, &size);
-    bpf_probe_read_user(&key.name, sizeof(key.name), (void *)classptr);
+    bpf_probe_read_user(&key.name, min(sizeof(key.name), (size_t)length), (void *)classptr);
     valp = allocs.lookup_or_try_init(&key, &zero);
     if (valp) {
         valp->total_size += size;


chenhengqi:
What is the output after this fix ?

D-D-H:
> What is the output after this fix ?

```
NAME/TYPE                      # ALLOCS      # BYTES
Test                                  1           16
[B                                    1         1016
[Ljava/lang/String;                   1         4016
[Ljava/lang/Thread;                   1         4016
[Ljava/lang/Float;                    1         4016
[D                                    1         8016
```

tools/lib/uobjnew.py
```diff
@@ -98,7 +98,16 @@
...
-    bpf_probe_read_user(&key.name, sizeof(key.name), (void *)classptr);
+    u32 len32 = 0;
+    u64 len64 = 0;
+    size_t len = sizeof(key.name);
+    if (!bpf_usdt_readarg(3, ctx, &len32)) {
+      len = min(len, (size_t)len32);
+    } else {
+      bpf_usdt_readarg(3, ctx, &len64);
+      len = min(len, (size_t)len64);
+    }
```
yonghong-song, line 109: could you add some comments here esp. w.r.t. why attempt to read both u32 and u64 len?

yonghong-song (commented):
As @chenhengqi asked, could you add the result after the fix in the commit message? This will make it easy to understand the motivation of the change. Thanks!

tools/lib/uobjnew.py
D-D-H: Updated, thanks

D-D-H:
> As @chenhengqi asked, could you add the result after the fix in the commit message? This will make it easy to understand the motivation of the change. Thanks!

Added, please help review if the message is appropriate or not.

tools/lib/uobjnew.py
```diff
@@ -98,7 +98,20 @@
     u64 classptr = 0, size = 0;
     bpf_usdt_readarg(2, ctx, &classptr);
     bpf_usdt_readarg(4, ctx, &size);
-    bpf_probe_read_user(&key.name, sizeof(key.name), (void *)classptr);
+    // The reason why read the third arg twice with different types here:
+    // According to the source code of OpenJDK:
+    // the declaration type of the third arg is uintptr_t(see hotspot.d)
+    // but the actual type is int(see SharedRuntime::dtrace_object_alloc)
```
yonghong-song, line 104: If the actual type is int, should we always read len32?

tools/lib/uobjnew.py
```diff
@@ -98,7 +98,20 @@
...
+    u32 len32 = 0;
+    u64 len64 = 0;
+    size_t len = sizeof(key.name);
+    if (!bpf_usdt_readarg(3, ctx, &len32)) {
+      len = min(len, (size_t)len32);
+    } else {
+      bpf_usdt_readarg(3, ctx, &len64);
```
yonghong-song, line 111: Do we have any chance to hit this branch?

tools/lib/uobjnew.py
D-D-H: At present, it should only be int, but I am not sure whether the OpenJDK community will fix this problem in the future.
If it will be fixed, we are currently not sure whether it is int or uintptr_t at the end.
So my current implementation is for better compatibility considerations.

tools/lib/uobjnew.py
D-D-H: This branch is never hit for the JDK version(the mainline).
Not sure about other versions (8, 11 ...)

yonghong-song: Let us just use 'int' type, i.e., len32 version. It is unlikely it will change to len64. Also if the argument type is indeed like uintptr_t. The current implementation is not correct anyway since you will need to do an additional dereference to get the actual length when reading argument which is really unnecessary for usdt.

D-D-H: okay, fixed.

yonghong-song:
[buildbot, test this please]

PR #3695, opened 2021-11-11
tools: fix cachetop.py with 5.15 kernel

yonghong-song:
The tool cachetop.py doesn't work with 5.15 kernel due to
kprobe function renaming. Adapt to the new function.
Commit 61087b961716 ("tools: fix cachestat.py with 5.15 kernel")
fixed a similar issue for cachestat.py.


brendangregg:
thanks!

PR #3711, opened 2021-11-18
fix: add batch methods into libbpf.h

hsqStephenZhang:
some batch methods are missing in libbpf.h, but it's ok for python because the libbpf.so has these symbols, but i think it's better to add these requisite methods in header file


hsqStephenZhang:
thanks

PR #3713, opened 2021-11-18
Enable CMP0074 to allow `${pkg}_ROOT`, especially for LLVM_ROOT

gfx:
Since CMake 3.12, `${pkg}_ROOT` is supported if `CMP0074` is set to `NEW`. This allows users to select a particular version of LLVM very easily. For example, `cmake ... -DLLVM_ROOT=/usr/lib/llvm-13`.

See https://cmake.org/cmake/help/latest/policy/CMP0074.html#policy:CMP0074 for details.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,7 @@
 # Copyright (c) PLUMgrid, Inc.
 # Licensed under the Apache License, Version 2.0 (the "License")
 cmake_minimum_required(VERSION 2.8.7)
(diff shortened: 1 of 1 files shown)


yonghong-song:
The patch got a failure on 18.04:
```
CMake Error at CMakeLists.txt:4 (cmake_policy):
  Policy "CMP0074" is not known to this version of CMake.
```

gfx:
Fixed for older dists!

PR #3718, opened 2021-11-21
tools: improve sslsniff (send buffer & filtering)

bacher09:
This makes few improvements to `sslsniff` tool:

* This can send much larger data payload and also adds `--max-buffer-size` CLI option which allow changing this param.
* Fixes dealing with non ASCII protocols, previously struct was defined as array of chars which made python ctypes treat it as NULL terminated string and it prevents from displaying any data past the null byte (which is very common for http2 protocol).
* Adds more filtering and displaying options (`--print-uid`, `--print-tid`, `--uid <uid>`)
* Avoids printing any data if `SSL_read` have failed with error ( `<= 0` return value), buffer in such cases has meaningless data, since it wasn't really decrypted.

This also deals correctly with rare cases when `bpf_probe_read_user` fails (so buffer should be empty and should not be displayed).

Basically, in order to escape BPF stack size limitation it uses per cpu array, which allows handling much larger payload.

### DIFF
--- a/tools/sslsniff.py
+++ b/tools/sslsniff.py
@@ -4,7 +4,8 @@
 #           GnuTLS and NSS
 #           For Linux, uses BCC, eBPF.
 #
-# USAGE: sslsniff.py [-h] [-p PID] [-c COMM] [-o] [-g] [-d]
+# USAGE: sslsniff.py [-h] [-p PID] [-u UID] [-x] [-c COMM] [-o] [-g] [-n] [-d]
+#                    [--hexdump] [--max-buffer-size SIZE]
 #
 # Licensed under the Apache License, Version 2.0 (the "License")
 #
@@ -23,17 +24,23 @@ import textwrap
 examples = """examples:
     ./sslsniff              # sniff OpenSSL and GnuTLS functions
     ./sslsniff -p 181       # sniff PID 181 only
+    ./sslsniff -u 1000      # sniff only UID 1000
     ./sslsniff -c curl      # sniff curl command only
     ./sslsniff --no-openssl # don't show OpenSSL calls
     ./sslsniff --no-gnutls  # don't show GnuTLS calls
     ./sslsniff --no-nss     # don't show NSS calls
     ./sslsniff --hexdump    # show data as hex instead of trying to decode it as UTF-8
+    ./sslsniff -x           # show process UID and TID
 """
 parser = argparse.ArgumentParser(
     description="Sniff SSL data",
     formatter_class=argparse.RawDescriptionHelpFormatter,
     epilog=examples)
 parser.add_argument("-p", "--pid", type=int, help="sniff this PID only.")
+parser.add_argument("-u", "--uid", type=int, default=None,
+                    help="sniff this UID only.")
+parser.add_argument("-x", "--extra", action="store_true",
+                    help="show extra fields (UID, TID)")
 parser.add_argument("-c", "--comm",
                     help="sniff only commands matching string.")
 parser.add_argument("-o", "--no-openssl", action="store_false", dest="openssl",
@@ -48,6 +55,8 @@ parser.add_argument("--ebpf", action="store_true",
                     help=argparse.SUPPRESS)
 parser.add_argument("--hexdump", action="store_true", dest="hexdump",
                     help="show data as hexdump instead of trying to decode it as UTF-8")
+parser.add_argument('--max-buffer-size', type=int, default=8192,
+                    help='Size of captured buffer')
 args = parser.parse_args()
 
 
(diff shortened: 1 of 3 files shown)


bacher09:
I also wanted to suggest to add few different output formats (aside from which is currently provided), because sometimes binary data is mixed out with plain ASCII and viewing this via plain hex view or plain ascii is not very convenient, so I was going to suggest something similar to `-A` option of `tcpdump`.

tools/sslsniff.py
```diff
@@ -94,47 +130,79 @@
...
+        } else {
+            buf_copy_size = 0;
+        }
+
+        perf_SSL_read.perf_submit(ctx, data, EVENT_SIZE(buf_copy_size));
         return 0;
 }
 """
 
+TASK_COMM_LEN = 16  # linux/sched.h
```
chenhengqi, line 193: This `TASK_COMM_LEN` is unused.

bacher09: You absolutely right, but it was the case even before my change :smile: . Anyway, I'll remove it.

tools/sslsniff.py
```diff
@@ -23,17 +23,25 @@
 examples = """examples:
     ./sslsniff              # sniff OpenSSL and GnuTLS functions
     ./sslsniff -p 181       # sniff PID 181 only
+    ./sslsniff -u 1000      # sniff only UID 1000
```
yonghong-song, line 26: Since you are touching this file, could you also update line 7:
 USAGE: sslsniff.py [-h] [-p PID] [-c COMM] [-o] [-g] [-d]
clearly, we has missed some flags there. You don't need to add --ebpf which is mostly hidden from tool users.

tools/sslsniff.py
```diff
@@ -23,17 +23,25 @@
...
     description="Sniff SSL data",
     formatter_class=argparse.RawDescriptionHelpFormatter,
     epilog=examples)
 parser.add_argument("-p", "--pid", type=int, help="sniff this PID only.")
+parser.add_argument("-u", "--uid", type=int, default=None,
+                    help="sniff this UID only.")
+parser.add_argument("-U", "--print-uid", action="store_true",
+                    help="print UID column")
+parser.add_argument("-T", "--print-tid", action="store_true",
+                    help="print TID column")
```
yonghong-song, line 44: I wonder whether you can replace '-U'/'-T' with '-x' to indicate extra columns to print. Please see capable.py for an example.

tools/sslsniff.py
```diff
@@ -48,41 +56,69 @@
...
+        int buf_filled;
+        char comm[TASK_COMM_LEN];
+        u8 buf[MAX_BUF_SIZE];
 };
 
+#define BASE_EVENT_SIZE ((size_t)(&((struct probe_SSL_data_t*)0)->buf))
+#define EVENT_SIZE(X) (BASE_EVENT_SIZE + ((size_t)(X)))
+
+
+BPF_PERCPU_ARRAY(ssl_data, struct probe_SSL_data_t, 1);
```
yonghong-song, line 85: percpu array is supported since 4.6. I think we are fine here as ubuntu 16.04 (4.4 kernel) is EOL now.

tools/sslsniff.py
```diff
@@ -48,41 +56,69 @@
...
+        data->tid = tid;
+        data->uid = uid;
+        data->len = num;
+        data->buf_filled = 0;
+        bpf_get_current_comm(&data->comm, sizeof(data->comm));
+        u32 buf_copy_size = min((size_t)MAX_BUF_SIZE, (size_t)num);
+
+        if (buf != 0) {
+                ret = bpf_probe_read_user(data->buf, buf_copy_size, buf);
+        }
```
yonghong-song, line 113: Let us just to make format consistent similar to line 99-100 by removing the '{}' as it only contains a single statement.

tools/sslsniff.py
```diff
@@ -48,41 +56,69 @@
...
-
-        bpf_get_current_comm(&__data.comm, sizeof(__data.comm));
-
-        if ( buf != 0) {
-                bpf_probe_read_user(&__data.v0, sizeof(__data.v0), buf);
+        if (!ret) {
+            data->buf_filled = 1;
+        } else {
+            buf_copy_size = 0;
         }
```
yonghong-song, line 119: The same as above, let remove '{}' since both then and else branch only have a single statement.

tools/sslsniff.py
```diff
@@ -94,47 +130,77 @@
...
+        int ret;
 
-        FILTER
+        PID_FILTER
+        UID_FILTER
 
         u64 *bufp = bufs.lookup(&tid);
         if (bufp == 0) {
                 return 0;
         }
```
yonghong-song, line 156: remove '{}'.

tools/sslsniff.py
```diff
@@ -94,47 +130,77 @@
...
+        data->len = (u32)len;
+        data->buf_filled = 0;
+        u32 buf_copy_size = min((size_t)MAX_BUF_SIZE, (size_t)len);
+
+        bpf_get_current_comm(&data->comm, sizeof(data->comm));
 
         if (bufp != 0) {
-                bpf_probe_read_user(&__data.v0, sizeof(__data.v0), (char *)*bufp);
+                ret = bpf_probe_read_user(&data->buf, buf_copy_size, (char *)*bufp);
         }
```
yonghong-song, line 178: ditto

tools/sslsniff.py
```diff
@@ -94,47 +130,77 @@
...
 
         bufs.delete(&tid);
 
-        perf_SSL_read.perf_submit(ctx, &__data, sizeof(__data));
+        if (!ret) {
+            data->buf_filled = 1;
+        } else {
+            buf_copy_size = 0;
+        }
```
yonghong-song, line 186: ditto

yonghong-song (commented):
Could you also add some examples in sslsniff_example.txt file? Could you also add the new options in the man page? Thanks!

tools/sslsniff.py
bacher09: I've tested it on `4.19` and `4.15` kernels (`Ubuntu 18.04`), so it's fine in this regard.

tools/sslsniff.py
```diff
@@ -94,47 +126,74 @@
...
+        if (bufp != 0)
+                ret = bpf_probe_read_user(&data->buf, buf_copy_size, (char *)*bufp);
 
         bufs.delete(&tid);
 
-        perf_SSL_read.perf_submit(ctx, &__data, sizeof(__data));
+        if (!ret)
+            data->buf_filled = 1;
+        else
+            buf_copy_size = 0;
```
chenhengqi, line 179: This indention is weird. The original code use 8 spaces.

yonghong-song (commented):
Thanks for the fix. The sslsniff.py looks good to me. Could you also add new examples to sslsniff_example.py and also change man page?

yonghong-song (commented):
Just one minor request for man page consistency. thanks!

bacher09:
Sorry, but could you clarify which inconsistency I need to fix in the man page ? I updated few things (in certain places order of arguments was different), but it wasn't in the man page. And thanks for the review.

yonghong-song:
> Sorry, but could you clarify which inconsistency I need to fix in the man page ? I updated few things (in certain places order of arguments was different), but it wasn't in the man page. And thanks for the review.

Let us check another man page, biolatency.8, beyond field, we also have options:
```
.SH OPTIONS
\-h
Print usage message.
.TP
\-T
Include timestamps on output.
.TP
\-m
Output histogram in milliseconds.
.TP
\-D
Print a histogram per disk device.
.TP
\-F
Print a histogram per set of I/O flags.
.TP
\-j
Print a histogram dictionary
.TP
\-e
Show extension summary(total, average)
```

In sslsniff.8, we only have fields, but not options,
```
.SH FIELDS
.TP
FUNC
Which function is being called (write/send or read/recv)
.TP
TIME
Time of the command, in seconds.
.TP
COMM
Entered command.
.TP
PID
Process ID calling SSL.
.TP
LEN
Bytes written or read by SSL functions.
```

That is what I mean inconsistency. We should also describe options in the man page. As a matter of fact, the field is just part of options. See trace.8,
```
.SH OPTIONS
.TP
\-h
Print usage message.
.TP
\-p PID
Trace only functions in the process PID.
.TP
\-L TID
Trace only functions in the thread TID.
.TP
\-v
Display the generated BPF program, for debugging purposes.
.TP
\-z STRING_SIZE
When collecting string arguments (of type char*), collect up to STRING_SIZE
characters. Longer strings will be truncated.
...
```

Since you are touching this file, could you change from `.SH FIELDS` to `.SH OPTIONS` and document all options (except --ebpf) in the man page?

bacher09:
Just updated, is it fine now ? And thanks for the clarification.

yonghong-song:
> Just updated, is it fine now ? And thanks for the clarification.

Yes, LGTM. Thanks for the update of the man page as well!

PR #3722, opened 2021-11-22
Add slabgroth script

poelzi:
This script monitors the difference each slab cache uses over time by
counting allocation and deallocation. If the system leaks memory in a slab
subsystem, this leakage will accumulate in the counter


brendangregg:
Thanks; seems too specialized as a stand-alone tool when I think this could be a flag to slabratetop(8) or even just slabtop(1).

Do you have a real world example of solving an issue with it where the other tools didn't work?

PR #3723, opened 2021-11-24
bcc: remove trailing semicolon of macro

JackyYin:
The trailing semicolon of a do-while style macro will cause an if-else condition without braces to fail compilation.

For example, Let's say we have a simple if-else statement:
```
if (condition)
    foo(x);
else
    bar(x);
```

Now we want to replace the `foo()` function with other macro `FOO`:

```
#define FOO(a)         \
do {                   \
// do something with a \
}   while(0);

if (condition)
    FOO(x);
else
    bar(x);
```

The above code will fail the compilation because of the redundant semicolon will end the statement of `if` clause, causing an orphan `else` clause.

Also notice that other do-while style macros are free of the trailing semicolon, so I think it's a good idea to make them align.


chenhengqi (approved):
LGTM. Thanks.

PR #3726, opened 2021-11-29
libbpf-tools: fix local/remote address byte ordering in tcprtt

chendotjs:
Since inet_aton() converts IPv4 numbers-and-dots notation to binary in network byte order, there is no need to do htonl() again.

Signed-off-by: chendotjs <chendotjs@gmail.com>

### DIFF
--- a/libbpf-tools/tcprtt.c
+++ b/libbpf-tools/tcprtt.c
@@ -10,7 +10,6 @@
 #include <signal.h>
 #include <unistd.h>
 #include <time.h>
-#include <arpa/inet.h>
 #include <bpf/libbpf.h>
 #include <bpf/bpf.h>
 #include "tcprtt.h"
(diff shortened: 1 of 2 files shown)


libbpf-tools/tcprtt.c
```diff
@@ -131,14 +130,14 @@ static error_t parse_arg(int key, char *arg, struct argp_state *state)
...
-		env.laddr = htonl(addr.s_addr);
+		env.laddr = addr.s_addr;
 		break;
 	case 'A':
 		if (inet_aton(arg, &addr) < 0) {
 			fprintf(stderr, "invalid remote address: %s\n", arg);
 			argp_usage(state);
 		}
-		env.raddr = htonl(addr.s_addr);
+		env.raddr = addr.s_addr;
```
chenhengqi, line 140: IIRC, this single `htonl` here is to avoid multiple `bpf_ntohl` in BPF program. This change here seems to be wrong.

chenhengqi (approved):
LGTM

libbpf-tools/tcprtt.c
chenhengqi: You are right.

> inet_aton() converts the Internet host address cp from the IPv4 numbers-and-dots notation into binary  form  (in
       network  byte  order) and stores it in the structure that inp points to.

ethercflow (approved):
LGTM

PR #3729, opened 2021-12-02
libbpf-tools/sysinjector: modify the return value of the syscall

KentaTada:
sysinjector is a new libbpf-based tool to inject an error into
any system call.
This tool helps to understand how to use error injection on libbpf
especially when users want to develop chaos engineering tools.

To change the return value of mknodat(2) to ENOMEM on Linux x86_64
we can use the following:
```
$ sudo ./sysinjector -r -12 -p 1111 __x64_sys_mknodat
```

The sleep command will fail after we execute the following:
```
$ sudo ./sysinjector -r -22 -c sleep __x64_sys_clock_nanosleep
```


libbpf-tools/sysinjector.bpf.c
```diff
@@ -0,0 +1,40 @@
...
+
+	bpf_get_current_comm(current_comm, sizeof(current_comm));
+	if (!is_same_comm(current_comm))
+		return 0;
+
+	bpf_override_return(ctx, retval);
+	return 0;
+}
+
+char LICENSE[] SEC("license") = "GPL";
```
chenhengqi, line 40: Can you dual license this tool like we do for other tools ?

libbpf-tools/sysinjector.bpf.c
```diff
@@ -0,0 +1,40 @@
...
+const volatile int retval = 0;
+
+static __always_inline int is_same_comm(const char *comm)
+{
+	int i;
+	for (i = 0; targ_comm[i] != '\0' && i < TASK_COMM_LEN; i++) {
+		if (comm[i] != targ_comm[i])
+			return false;
+	}
+	return true;
```
chenhengqi, line 19: targ_comm = `abc` and comm = `abcd` will be evaluated as true.

libbpf-tools/sysinjector.c
```diff
@@ -0,0 +1,175 @@
...
+	if (env.comm)
+		strncpy((char*)obj->rodata->targ_comm, env.comm, env.comm_len);
+
+	err = sysinjector_bpf__load(obj);
+	if (err) {
+		warn("failed to load BPF object\n");
+		return 1;
+	}
+
+	err = attach_kprobes(obj);
```
chenhengqi, line 162: You can use `bpf_program__set_attach_target` and let libbpf do auto-attach.

libbpf-tools/sysinjector.c
```diff
@@ -0,0 +1,175 @@
...
+	struct prog_env *env = state->input;
+	long pid;
+	int len, retval;
+
+	switch (key) {
+	case 'h':
+		argp_state_help(state, stderr, ARGP_HELP_STD_HELP);
+		break;
+	case 'p':
+		errno = 0;
```
chenhengqi, line 58: You don't need to set errno explicitly, do you ?

libbpf-tools/sysinjector.c
```diff
@@ -0,0 +1,175 @@
...
+		pid = strtol(arg, NULL, 10);
+		if (errno || pid <= 0) {
+			warn("Invalid PID: %s\n", arg);
+			argp_usage(state);
+		}
+		env->pid = pid;
+		break;
+	case 'c':
+		env->comm = arg;
```
chenhengqi, line 67: strdup ?

chenhengqi (changes requested):

libbpf-tools/sysinjector.bpf.c
```diff
@@ -0,0 +1,40 @@
...
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include "sysinjector.h"
+
+const volatile pid_t targ_tgid = 0;
+const volatile char targ_comm[TASK_COMM_LEN] = {};
+const volatile int retval = 0;
+
+static __always_inline int is_same_comm(const char *comm)
```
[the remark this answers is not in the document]
ethercflow: How about rename to `bool comm_allowed(const char *comm)` like we did for other tools?

libbpf-tools/sysinjector.c
KentaTada: _strtol()_  can legitimately return 0, LONG_MAX, or LONG_MIN on both success and failure. So, this code should set errno to 0 before calling _strtol()_ .
https://man7.org/linux/man-pages/man3/strtol.3.html

libbpf-tools/sysinjector.bpf.c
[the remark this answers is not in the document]
KentaTada: Thanks!
Basically, I'll reuse comm_allowed() but I think we should not use bool because of C89-compliant and will modify it.

libbpf-tools/sysinjector.bpf.c
KentaTada: I'll reuse comm_allowed().

libbpf-tools/sysinjector.bpf.c
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: GPL-2.0
```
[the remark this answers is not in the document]
KentaTada: Fixed

libbpf-tools/sysinjector.bpf.c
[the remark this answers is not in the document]
KentaTada: Fixed

libbpf-tools/sysinjector.c
KentaTada: Fixed

libbpf-tools/sysinjector.bpf.c
KentaTada: Yes. Fixed it.
Thanks!

libbpf-tools/sysinjector.bpf.c
```diff
@@ -0,0 +1,40 @@
...
+#include <bpf/bpf_tracing.h>
+#include "sysinjector.h"
+
+const volatile pid_t targ_tgid = 0;
+const volatile char targ_comm[TASK_COMM_LEN] = {};
+const volatile int retval = 0;
+
+static __always_inline int is_same_comm(const char *comm)
+{
+	int i;
```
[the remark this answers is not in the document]
KentaTada: Fixed

libbpf-tools/Makefile
```diff
@@ -49,6 +49,7 @@ APPS = \
 	solisten \
 	statsnoop \
 	syscount \
+	sysinjector \
```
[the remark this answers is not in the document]
KentaTada: >BCC has a tool called inject, should this be renamed to inject ?

No. _sysinjector_ is not based on the BCC's inject. Actually, this tool modifies the return code of syscall instead of kmalloc, bio, alloc_page.
The objective of this tool shows how to implement chaos engineering tools for the application layer using libbpf .

libbpf-tools/sysinjector.c
KentaTada: >You can use bpf_program__set_attach_target and let libbpf do auto-attach.

This tool uses kprobes to override the return value.
I think it is the only way to do it.

KentaTada:
@chenhengqi, @ethercflow
I appreciate your prompt help.
I modified my code and answered your question.

libbpf-tools/Makefile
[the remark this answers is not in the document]
chenhengqi: Same functionality, no ?

libbpf-tools/sysinjector.c
chenhengqi:
```c
bpf_program__set_attach_target(obj->progs.handle_retval_at_enter, 0, env.syscall);
sysinjector_bpf__attach();
```

libbpf-tools/sysinjector.bpf.c
chenhengqi: The case above not yet resolved.

libbpf-tools/sysinjector.c
KentaTada: bpf_program__set_attach_target() is used for BTF-based kernel attach targets.
As of now, when the user-space program specifies attach_btf_id using bpf_program__set_attach_target(), BPF_PROG_TYPE_TRACING, BPF_PROG_TYPE_LSM , BPF_PROG_TYPE_STRUCT_OPS, BPF_PROG_TYPE_EXT are only supported as the program type.
https://github.com/torvalds/linux/blob/v5.16-rc4/kernel/bpf/syscall.c#L2053-L2075
Actually, the above code could not load the BPF skeleton.

Or do you mean we should support both kprove and fmod_ret?

libbpf-tools/Makefile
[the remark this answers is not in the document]
KentaTada: >Same functionality, no ?
No.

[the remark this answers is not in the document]
chenhengqi: I won't insist on this, but I think we can extend this tool to support the inject functionalities and replace it.

libbpf-tools/sysinjector.c
chenhengqi: You are right, currently libbpf don't support set attach target for kprobe.

libbpf-tools/Makefile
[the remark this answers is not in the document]
KentaTada: I'm sorry for the late reply.
There is a difference in concept between _inject_ and _sysinjector_.
So I think _sysinjector_ should be independent from _inject_.
However, I also want to implement _inject_  using libbpf in the near future.

KentaTada:
I think I resolved all problems.

libbpf-tools/sysinjector.bpf.c
```diff
@@ -0,0 +1,41 @@
...
+
+const volatile pid_t targ_tgid = 0;
+const volatile char targ_comm[TASK_COMM_LEN] = {};
+const volatile int retval = 0;
+
+static __always_inline int comm_allowed(const char *comm)
+{
+	int i;
+
+	for (i = 0; i < TASK_COMM_LEN && targ_comm[i] != '\0'; i++) {
```
chenhengqi, line 16: This will treat "abc" == "abcd", right ?

KentaTada: Yes.
I think it is better to match the original implementation of comm_allowed() because we unify the behavior of all tools.
But if you and @ethercflow and maintainers don't mind, I'll try to modify comm_allowed() and move it to _trace_helpers.c_ in the another PR.

chenhengqi: SGTM

chenhengqi (approved):
LGTM

PR #3732, opened 2021-12-04
use probe_limt from env

woodgear:


woodgear:
@brendangregg @drzaeus77 @goldshtn seems i need a approve to run the workflows.

yonghong-song:
@woodgear could you add commit message (details on why you proposed this patch and what the motivation and how the change fixed your problem, etc.) to the patch and also add the commit message to the github description?

woodgear:
currently, I'm trying to use funccount.py to observe a running Nginx process, I have no sense for Nginx before, just want to observe how many userspace function-call a normal HTTP request will cause, which I thought is what funccount.py try deal with. but Nginx has 4777 functions, and the current hard cord probe limit in bcc is 1000, so I create this pr which could read PROBE_LIMIT from env. if not env set I will use default probe limit 1000 just as before

chenhengqi:
@yonghong-song  Is there a hard limit in the kernel ?

yonghong-song:
> @yonghong-song Is there a hard limit in the kernel ?

No, there is no hard limit. But as far as I know, the kprobe attachment will become slower as the number of probes gets bigger and bigger. In kernel, currently there are some efforts to do batch attachment (the same bpf program to many kernel functions).

But I agree we should not limit this at least to make it configurable. So @woodgear I think you patch looks okay. But your patch commit message is empty. Could you resubmit with proper commit message to explain "why" the change as you did in the above comment?

woodgear:
update commit message.

src/python/bcc/__init__.py
```diff
@@ -751,9 +751,17 @@ def get_kprobe_functions(event_re):
 
     def _check_probe_quota(self, num_new_probes):
         global _num_open_probes
-        if _num_open_probes + num_new_probes > _probe_limit:
+        if _num_open_probes + num_new_probes > BPF.get_probe_limit():
             raise Exception("Number of open probes would exceed global quota")
 
+    @staticmethod
+    def get_probe_limit():
+        env_probe_limit = os.environ.get('PROBE_LIMIT')
```
yonghong-song, line 759: One more request, could you change env name from PROBE_LIMIT to BCC_PROBE_LIMIT? This way, we have less possible conflict with other environ variables.

woodgear: update

PR #3740, opened 2021-12-08
hardirqs: fix issue if irq is triggered while idle task (tid=0)

ismhong:
Currently, hardirqs use tid as key to store information while tracepoint
irq_handler_entry. It works fine if irq is triggered while normal task
running, but there is a chance causing overwrite issue while irq is
triggered while idle task (a.k.a swapper/x, tid=0) running on multi-core
system.

Let's say there are two irq event trigger simultaneously on both CPU
core, irq A @ core #0, irq B @ core #1, and system load is pretty light,
so BPF program will get tid=0 since current task is swapper/x for both cpu
core. In this case, the information of first irq event stored in map could
be overwritten by incoming second irq event.

Use tid and cpu_id together to make sure the key is unique for each
event in this corner case.

Please check for more details at merge request #2804, #3733.

### DIFF
--- a/tools/hardirqs.py
+++ b/tools/hardirqs.py
@@ -67,6 +67,14 @@ bpf_text = """
 #include <linux/irqdesc.h>
 #include <linux/interrupt.h>
 
+// Add cpu_id as part of key for irq entry event to handle the case which irq
+// is triggered while idle thread(swapper/x, tid=0) for each cpu core.
+// Please see more detail at pull request #2804, #3733.
+typedef struct entry_key {
+    u32 tid;
+    u32 cpu_id;
+} entry_key_t;
+
 typedef struct irq_key {
     char name[32];
     u64 slot;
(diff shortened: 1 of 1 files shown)


chenhengqi (commented):
Looks good overall.

Have to mention here, it seems like the libbpf-tool version has a better implementation.

ismhong:
Hi @chenhengqi 

Thanks for review.
> Have to mention here, it seems like the libbpf-tool version has a better implementation.

I just checked the hardirqs implementation of libbpf-tool version.
For count mode, maybe we can just only attach BPF program to raw tracepoint of tracepoint:irq:irq_handler_exit for both `res` field and `struct irqaction * action` for name of irq. This way should also keep things simple for count mode.

The reason using raw tracepoint is the `struct irqaction* action` is not exported to fields of tracepoint:irq:irq_handler_exit.

How do you think for this? Thanks.
```
TRACE_EVENT(irq_handler_exit,

    TP_PROTO(int irq, struct irqaction *action, int ret),

    TP_ARGS(irq, action, ret),

    TP_STRUCT__entry(
        __field(    int,    irq )
        __field(    int,    ret )
    ),

    TP_fast_assign(
        __entry->irq    = irq;
        __entry->ret    = ret;
    ),

    TP_printk("irq=%d ret=%s",
          __entry->irq, __entry->ret ? "handled" : "unhandled")
);
```

For time/hist mode, it seems the current libbpf version hardirqs didn't deal with the unique key for start timestamp or I misunderstand something?
Do you have any better suggestion for time/hist mode? Thank you. :)
https://github.com/iovisor/bcc/blob/814c264d8088fc566dd9a5650691158281792cb4/libbpf-tools/hardirqs.bpf.c#L49

chenhengqi:
For count mode, `tracepoint/irq/irq_handler_entry` should work, right ?

ismhong:
Hi @chenhengqi

No, for count mode, we need the `ret` field of tracepoint:irq:irq_handler_exit for fixing the shared irq issue (https://github.com/iovisor/bcc/pull/3733).

chenhengqi:
Sorry, I forgot that. Let me fix the libbpf tool.

Issue #2697, opened 2020-01-10
[API user friendliness] StatusTuple can have canonical error code instead of an integer


yzhao1012:
Right now, StatusTuple accepts an 'int' as "error code". This because hard to enforce meaningful error code across bcc code base. And most of BCC code simply returns -1.

Blindly using -1 or inconsistent error code makes users difficult to react to error correctly.

For example, attach_uprobe() can return error when a probe was already attached before. Without a distinguished error code than -1, user code has no idea what to do with it. But normally user code want to just ignore this failure and move on.

I suggest to predefine a few canonical error code, where simple errors like duplicated operation can be reported as such.

This would means a tedious laborious changes to update all StatusTuple returning code, so far there seems at least 256 places. But that should be straightforward with a few lines of shell code.
$ grep 'return StatusTuple' third_party/bcc -R | wc -l
256

WDYT?

yonghong-song:
This is a good idea. Maybe you can help?

yzhao1012:
Glad to hear your consent, Yonghong! I can pull out draft. Stay tuned.

yzhao1012:
Sorry for long inactivity. I started to allocate time on this from next week. Changes should be coming next week.

yzhao1012:
Current plan:
* Replace error checking using StatusTuple::code() (the forms include code() != 0, !code()) with StatusTuple::ok(). PRs are created for src/ tests/ examples/ respectively.
* Replace more exotic form of using StatusTuple::code() for error checking with StatusTuple::ok(). For example, StatusTuple::code() < 0. But that requires additional research.
* Replace the use of the ctors of StatusTuple which uses integer code with the one using enum class Code. Replace
  * StatusTuple(int ret)
  * StatusTuple(int ret, const char *msg)
  * StatusTuple(int ret, const std::string &msg)
  With appropriate forms of StatusTuple(Code code, const std::string &msg). This part is more complicated than the previous steps, as of 2021-12-13, this comment was written, there hasn't been detailed study yet, so the actual steps are not yet decided.
* Any left-over cleanup work.
* Removing the StatusTuple ctors that uses integer error code.

PR #3742, opened 2021-12-08
!StatusTuple::ok() replaced StatusTuple::code() != 0 for files under src/ dir

yzhao1012:
This is done with the shell command:
```
sed -i 's|\([A-Z_a-z]\+\).code() != 0|!\1.ok()|' $(grep '\.code() != 0' src -rl)
```
This is part of #2697

### DIFF
--- a/src/cc/bcc_exception.h
+++ b/src/cc/bcc_exception.h
@@ -86,7 +86,7 @@ private:
 #define TRY2(CMD)                    \
   do {                               \
     ebpf::StatusTuple __stp = (CMD); \
-    if (!__stp.ok()) {         \
+    if (!__stp.ok()) {               \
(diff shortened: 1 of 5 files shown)


src/cc/bcc_exception.h
```diff
@@ -86,7 +86,7 @@ class StatusTuple {
 #define TRY2(CMD)                    \
   do {                               \
     ebpf::StatusTuple __stp = (CMD); \
-    if (__stp.code() != 0) {         \
+    if (!__stp.ok()) {         \
```
chenhengqi, line 89: nit: Please keep the `\` aligned.

yzhao1012: Done.

Issue #3737, opened 2021-12-07
tools/trace.py: add option to filter tracing by UID


evilpan:
Tools like opensnoop and execsnoop supports `-u UID` to filter output by user ID, which is very useful in practice. Maybe we should add this to strace.py to make it better?

yonghong-song:
Yes, we can do it. Could you help with a pull request?

PR #3743, opened 2021-12-10
Add --uid option to filter by user ID

evilpan:
This is for issue #3737 .

We should clear up the messy options in `strace.py`, and use `-u` for UID filtering (like opensnoop/execsnoop). But for now I'll just leave it as is to avoid breaking the documents.


yonghong-song (commented):
LGTM. Could you also change example file and man page to include this option?

PR #3751, opened 2021-12-13
Replace !StatusTuple::code() with StatusTuple::ok() in src/cc/api/BPFTable.h

yzhao1012:
!StatusTuple::code() is equivalent to !StatusTuple::ok().

### DIFF
--- a/src/cc/api/BPFTable.h
+++ b/src/cc/api/BPFTable.h
@@ -48,7 +48,7 @@ class BPFQueueStackTableBase {
   StatusTuple leaf_to_string(const ValueType* value, std::string& value_str) {
     char buf[8 * desc.leaf_size];
     StatusTuple rc = desc.leaf_snprintf(buf, sizeof(buf), value);
-    if (!rc.code())
+    if (rc.ok())
       value_str.assign(buf);
     return rc;
   }
@@ -90,7 +90,7 @@ class BPFTableBase {
   StatusTuple key_to_string(const KeyType* key, std::string& key_str) {
     char buf[8 * desc.key_size];
     StatusTuple rc = desc.key_snprintf(buf, sizeof(buf), key);
-    if (!rc.code())
+    if (rc.ok())
       key_str.assign(buf);
     return rc;
   }
@@ -98,7 +98,7 @@ class BPFTableBase {
   StatusTuple leaf_to_string(const ValueType* value, std::string& value_str) {
     char buf[8 * desc.leaf_size];
     StatusTuple rc = desc.leaf_snprintf(buf, sizeof(buf), value);
-    if (!rc.code())
+    if (rc.ok())
       value_str.assign(buf);
     return rc;
   }


src/cc/api/BPFTable.h
```diff
@@ -48,7 +48,7 @@ class BPFQueueStackTableBase {
   StatusTuple leaf_to_string(const ValueType* value, std::string& value_str) {
     char buf[8 * desc.leaf_size];
     StatusTuple rc = desc.leaf_snprintf(buf, sizeof(buf), value);
-    if (!rc.code())
+    if (!rc.ok())
```
chenhengqi, line 51: `!rc.code()` is equivalent to `rc.code() == 0`. So this should be `rc.ok()`.

yzhao1012:
I think the test failure is not relevant to this change.

How to restart the build?

yonghong-song:
Just kicked the rerun. See https://github.community/t/manually-restart-actions-and-entire-workflows/16262/3. Inside the workflow, you can trigger rerun's.

davemarchevsky:
I'm not sure if the test failure here is spurious. Maybe it's related to @chenhengqi's comment? Haven't been seeing similar spurious test failures on other PRs

src/cc/api/BPFTable.h
yzhao1012: Oh, you are right. I must have went through a weird mental wiring that exactly reversed the meaning of 0...

yzhao1012:
> I'm not sure if the test failure here is spurious. Maybe it's related to @chenhengqi's comment? Haven't been seeing similar spurious test failures on other PRs

You are right, I got the meaning of !code() wrong.

yzhao1012:
I changed the if condition from !StatusTuple.ok() to StatusTuple::ok() as @chenhengqi noted. But the test still failed, as shown in the build.

I tried testing command locally. It still failed, but seems the warning at the beginning shows it's expected (as logged below)?
```
$ tests/wrapper.sh "py_test_tools_smoke" "sudo" ../tests/python/test_tools_smoke.py   
[sudo] password for yzhao: 
timeout: failed to run command ‘../../tools/argdist.py’: No such file or directory
CRITICAL:root:WARNING! Test test_argdist (__main__.SmokeTests) failed, but marked as passed because it is decorated with @mayFail.
CRITICAL:root:  The reason why this mayFail was: This fails on github actions environment, and needs to be fixed
CRITICAL:root:  The failure was: "0 != 127"
```

yzhao1012:
> Just kicked the rerun. See https://github.community/t/manually-restart-actions-and-entire-workflows/16262/3. Inside the workflow, you can trigger rerun's.

It seems not the same for you and myself, as I followed the instruction at https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs (which was linked in the comment of https://github.community/t/manually-restart-actions-and-entire-workflows/16262/3 mentioned above), there is no drop-down menu as claimed for me.

image

davemarchevsky:
@yzhao1012 could you rebase and force push? This is also a convenient way to retrigger tests

yonghong-song:
> > Just kicked the rerun. See https://github.community/t/manually-restart-actions-and-entire-workflows/16262/3. Inside the workflow, you can trigger rerun's.
> 
> It seems not the same for you and myself, as I followed the instruction at https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs (which was linked in the comment of https://github.community/t/manually-restart-actions-and-entire-workflows/16262/3 mentioned above), there is no drop-down menu as claimed for me.
> 
> image

Maybe this is a permission issue. Maybe need to check github workflow documentation at least pull request owner can rerun the test for their own pull request.

PR #3756, opened 2021-12-16
Implement bashreadline with libbpf.

ThinkerYzu1:
Bashreadline will print user inputs, returning from readline, of every
instance of bash shell.  Readline is in bash itself, linked
statically, for some devices, while others may link to libreadline.so.
This implementation finds the symbol in bash if possible. Or, it tries
to find libreadline.so using ldd if the symbol is not in bash.

Signed-off-by: Kui-Feng Lee <kuifeng@fb.com>

### DIFF
--- a/libbpf-tools/Makefile
+++ b/libbpf-tools/Makefile
@@ -16,6 +16,7 @@ $(error Architecture $(ARCH) is not supported yet. Please open an issue)
 endif
 
 APPS = \
+	bashreadline \
 	bindsnoop \
 	biolatency \
 	biopattern \
--- a/libbpf-tools/bashreadline.bpf.c
+++ b/libbpf-tools/bashreadline.bpf.c
@@ -1,0 +1,38 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+/* Copyright (c) 2021 Facebook */
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include "bashreadline.h"
+
+#define TASK_COMM_LEN 16
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, sizeof(__u32));
+} events SEC(".maps");
+
+SEC("uretprobe/readline")
+int BPF_KRETPROBE(printret, const void *ret) {
+	struct str_t data;
+	char comm[TASK_COMM_LEN];
+	u32 pid;
+
+	if (!ret)
+		return 0;
+
+	bpf_get_current_comm(&comm, sizeof(comm));
+	if (comm[0] != 'b' || comm[1] != 'a' || comm[2] != 's' || comm[3] != 'h' || comm[4] != 0 )
+		return 0;
+
+	pid = bpf_get_current_pid_tgid() >> 32;
+	data.pid = pid;
+	bpf_probe_read_user_str(&data.str, sizeof(data.str), ret);
+
+	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, &data, sizeof(data));
+
+	return 0;
+};
+
+char LICENSE[] SEC("license") = "GPL";
--- a/libbpf-tools/bashreadline.c
+++ b/libbpf-tools/bashreadline.c
@@ -1,0 +1,215 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+/* Copyright (c) 2021 Facebook */
+#include <argp.h>
+#include <stdio.h>
+#include <errno.h>
+#include <signal.h>
+#include <time.h>
+#include <unistd.h>
+#include <ctype.h>
+
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+#include "bashreadline.h"
+#include "bashreadline.skel.h"
+#include "trace_helpers.h"
+#include "uprobe_helpers.h"
+
+#define PERF_BUFFER_PAGES	16
+#define PERF_POLL_TIMEOUT_MS	100
+#define warn(...) fprintf(stderr, __VA_ARGS__)
+
+static volatile sig_atomic_t exiting = 0;
+
+const char *argp_program_version = "bashreadline 1.0";
+const char *argp_program_bug_address =
+	"https://github.com/iovisor/bcc/tree/master/libbpf-tools";
+const char argp_program_doc[] =
+"Print entered bash commands from all running shells.\n"
+"\n"
+"USAGE: bashreadline [-s <path/to/libreadline.so>]\n"
+"\n"
+"EXAMPLES:\n"
+"    bashreadline\n"
+"    bashreadline -s /usr/lib/libreadline.so\n";
+
+static const struct argp_option opts[] = {
+	{ "shared", 's', "PATH", 0, "the location of libreadline.so library" },
+	{ NULL, 'h', NULL, OPTION_HIDDEN, "Show the full help" },
+	{},
+};
+
+static char *libreadline_path = NULL;
+
+static error_t parse_arg(int key, char *arg, struct argp_state *state)
+{
+	switch (key) {
+	case 's':
+		libreadline_path = strdup(arg);
+		if (libreadline_path == NULL)
+			return ARGP_ERR_UNKNOWN;
+		break;
+	case 'h':
+		argp_state_help(state, stderr, ARGP_HELP_STD_HELP);
+		break;
+	default:
+		return ARGP_ERR_UNKNOWN;
+	}
+	return 0;
+}
+
+static void handle_event(void *ctx, int cpu, void *data, __u32 data_size)
+{
+	struct str_t *e = data;
+	struct tm *tm;
+	char ts[16];
+	time_t t;
+
+	time(&t);
+	tm = localtime(&t);
+	strftime(ts, sizeof(ts), "%H:%m:%S", tm);
+
+	printf("%-9s %-7d %s\n", ts, e->pid, e->str);
+}
+
+static void handle_lost_events(void *ctx, int cpu, __u64 lost_cnt)
+{
+	warn("lost %llu events on CPU #%d\n", lost_cnt, cpu);
+}
+
+static char *find_readline_so()
+{
+	const char *bash_path = "/bin/bash";
+	FILE *fp;
+	off_t func_off;
+	char *line = NULL;
+	size_t line_sz = 0;
+	char path[128];
+	char *result = NULL;
+
+	func_off = get_elf_func_offset(bash_path, "readline");
+	if (func_off >= 0)
+		return strdup(bash_path);
+
+	/*
+	 * Try to find libreadline.so if readline is not defined in
+	 * bash itself.
+	 *
+	 * ldd will print a list of names of shared objects,
+	 * dependencies, and their paths.  The line for libreadline
+	 * would looks like
+	 *
+	 *      libreadline.so.8 => /usr/lib/libreadline.so.8 (0x00007b....)
+	 *
+	 * Here, it finds a line with libreadline.so and extracts the
+	 * path after the arrow, '=>', symbol.
+	 */
+	fp = popen("ldd /bin/bash", "r");
+	if (fp == NULL)
+		goto cleanup;
+	while (getline(&line, &line_sz, fp) >= 0) {
+		if (sscanf(line, "%*s => %127s", path) < 1)
+			continue;
+		if (strstr(line, "/libreadline.so")) {
+			result = strdup(path);
+			break;
+		}
+	}
+
+cleanup:
+	if (line)
+		free(line);
+	if (fp)
+		fclose(fp);
+	return result;
+}
+
+static void sig_int(int signo)
+{
+	exiting = 1;
+}
+
+int main(int argc, char **argv)
+{
+	static const struct argp argp = {
+		.options = opts,
+		.parser = parse_arg,
+		.doc = argp_program_doc,
+	};
+	struct bashreadline_bpf *obj = NULL;
+	struct perf_buffer_opts pb_opts;
+	struct perf_buffer *pb = NULL;
+	char *readline_so_path;
+	off_t func_off;
+	int err;
+
+	err = argp_parse(&argp, argc, argv, 0, NULL, NULL);
+	if (err)
+		return err;
+
+	if (libreadline_path) {
+		readline_so_path = libreadline_path;
+	} else if ((readline_so_path = find_readline_so()) == NULL) {
+		warn("failed to find readline\n");
+		return 1;
+	}
+
+	err = bump_memlock_rlimit();
+	if (err) {
+		warn("failed to increase rlimit: %d\n", err);
+		goto cleanup;
+	}
+
+	obj = bashreadline_bpf__open_and_load();
+	if (!obj) {
+		warn("failed to open and load BPF object\n");
+		goto cleanup;
+	}
+
+	func_off = get_elf_func_offset(readline_so_path, "readline");
+	if (func_off < 0) {
+		warn("cound not find readline in %s\n", readline_so_path);
+		goto cleanup;
+	}
+
+	obj->links.printret = bpf_program__attach_uprobe(obj->progs.printret, true, -1,
+							 readline_so_path, func_off);
+	err = libbpf_get_error(obj->links.printret);
+	if (err) {
+		warn("failed to attach readline: %d\n", err);
+		goto cleanup;
+	}
+
+	pb_opts.sample_cb = handle_event;
+	pb_opts.lost_cb = handle_lost_events;
+	pb = perf_buffer__new(bpf_map__fd(obj->maps.events), PERF_BUFFER_PAGES, &pb_opts);
+	err = libbpf_get_error(pb);
+	if (err) {
+		warn("failed to open perf buffer: %d\n", err);
+		goto cleanup;
+	}
+
+	if (signal(SIGINT, sig_int) == SIG_ERR) {
+		warn("can't set signal handler: %s\n", strerror(errno));
+		err = 1;
+		goto cleanup;
+	}
+
+	printf("%-9s %-7s %s\n", "TIME", "PID", "COMMAND");
+	while (!exiting) {
+		err = perf_buffer__poll(pb, PERF_POLL_TIMEOUT_MS);
+		if (err < 0 && errno != EINTR) {
+			warn("error polling perf buffer: %s\n", strerror(errno));
+			goto cleanup;
+		}
+		err = 0;
+	}
+
+cleanup:
+	if (readline_so_path)
+		free(readline_so_path);
+	perf_buffer__free(pb);
+	bashreadline_bpf__destroy(obj);
+
+	return err != 0;
+}
--- a/libbpf-tools/bashreadline.h
+++ b/libbpf-tools/bashreadline.h
@@ -1,0 +1,13 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+/* Copyright (c) 2021 Facebook */
+#ifndef __BASHREADLINE_H
+#define __BASHREADLINE_H
+
+#define MAX_LINE_SIZE 80
+
+struct str_t {
+	__u32 pid;
+	char str[MAX_LINE_SIZE];
+};
+
+#endif /* __BASHREADLINE_H */
--- a/libbpf-tools/.gitignore
+++ b/libbpf-tools/.gitignore
@@ -1,4 +1,5 @@
 /.output
+/bashreadline
 /bindsnoop
 /biolatency
 /biopattern


libbpf-tools/bashreadline.bpf.c
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2020 Wenbo Zhang
```
anakryiko, line 2: wrong copyright (and year as well)?

libbpf-tools/bashreadline.bpf.c
```diff
@@ -0,0 +1,35 @@
...
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
```
anakryiko, line 5: probably don't need this one

libbpf-tools/bashreadline.h
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+#ifndef __BASHREADLINE_H
+#define __BASHREADLINE_H
+
+#define TASK_COMM_LEN 16
```
anakryiko, line 5: you are using this only in single .c file, please move #define there

libbpf-tools/bashreadline.bpf.c
```diff
@@ -0,0 +1,35 @@
...
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, sizeof(__u32));
+} events SEC(".maps");
+
+SEC("uretprobe/readline")
+int printret(struct pt_regs *ctx) {
+	struct str_t data  = {};
```
anakryiko, line 17: no need to initialize data, you are filling it out below

libbpf-tools/bashreadline.bpf.c
```diff
@@ -0,0 +1,35 @@
...
+	char comm[TASK_COMM_LEN] = {};
```
anakryiko, line 18: same, you'll fill it out with bpf_get_current_comm()

libbpf-tools/bashreadline.bpf.c
```diff
@@ -0,0 +1,35 @@
...
+	u32 pid;
+	if (!PT_REGS_RC(ctx))
```
anakryiko, line 20: let's follow consistent kernel code style, add empty line after variable declarations

libbpf-tools/bashreadline.bpf.c
```diff
@@ -0,0 +1,35 @@
...
+#include "bashreadline.h"
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, sizeof(__u32));
+} events SEC(".maps");
+
+SEC("uretprobe/readline")
+int printret(struct pt_regs *ctx) {
```
anakryiko, line 16: try using BPF_KRETPROBE() macro, see other tools

libbpf-tools/bashreadline.bpf.c
```diff
@@ -0,0 +1,35 @@
...
+	struct str_t data  = {};
+	char comm[TASK_COMM_LEN] = {};
+	u32 pid;
+	if (!PT_REGS_RC(ctx))
+		return 0;
+	pid = bpf_get_current_pid_tgid() >> 32;
+	data.pid = pid;
+	bpf_probe_read_user(&data.str, sizeof(data.str) - 1, (void *)PT_REGS_RC(ctx));
```
anakryiko, line 24: should be sizeof(data.str), bpf_probe_read_user() will properly zero-terminate

libbpf-tools/bashreadline.bpf.c
```diff
@@ -0,0 +1,35 @@
...
+	data.str[sizeof(data.str) - 1] = 0;
```
anakryiko, line 25: no need, see above

libbpf-tools/bashreadline.bpf.c
```diff
@@ -0,0 +1,35 @@
...
+
+	bpf_get_current_comm(&comm, sizeof(comm));
+	if (comm[0] == 'b' && comm[1] == 'a' && comm[2] == 's' && comm[3] == 'h' && comm[4] == 0 ) {
+		bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, &data, sizeof(data));
+	}
```
anakryiko, line 30: nit: unnecessary {} for single-line if body

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,232 @@
...
+#include "bashreadline.h"
+#include "bashreadline.skel.h"
+#include "trace_helpers.h"
+#include "uprobe_helpers.h"
+
+#define PERF_BUFFER_PAGES	16
+#define PERF_POLL_TIMEOUT_MS	100
+#define warn(...) fprintf(stderr, __VA_ARGS__)
+
+
```
anakryiko, line 21: extra empty line?

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,232 @@
...
+static volatile sig_atomic_t exiting = 0;
+
+const char *argp_program_version = "bashredline 1.0";
```
anakryiko, line 24: typo: read

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,232 @@
...
+const char *argp_program_bug_address =
+	"https://github.com/iovisor/bcc/tree/master/libbpf-tools";
+const char argp_program_doc[] =
+"Print entered bash commands from all running shells.\n"
+"\n"
+"USAGE: bashreadline [-s <readline>]\n"
```
anakryiko, line 30: probably better to write "<path/to/libreadline.so>" to make it clearer?

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,232 @@
...
+	{},
+};
+
+static char *libreadline_path = NULL;
+
+static error_t parse_arg(int key, char *arg, struct argp_state *state)
+{
+	switch (key) {
+	case 's':
+		libreadline_path = strdup(arg);
```
anakryiko, line 48: check for NULL?

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,232 @@
...
+
+	printf("%-9s %-6lld %s\n", ts, e->pid, e->str);
+}
+
+static void handle_lost_events(void *ctx, int cpu, __u64 lost_cnt)
+{
+	warn("lost %llu events on CPU #%d\n", lost_cnt, cpu);
+}
+
+static char *find_readline_so() {
```
anakryiko, line 78: nit: { on new line

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,232 @@
...
+	char *line = NULL;
+	char *sep, *path_start, *sp;
+	size_t line_sz;
+	off_t func_off;
+	char *result = NULL;
+
+	func_off = get_elf_func_offset(bash_path, "readline");
+	if (func_off >= 0) {
+		return strdup(bash_path);
+	}
```
anakryiko, line 90: unnecessary {}

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,232 @@
...
+	if (line)
+		free(line);
+	return result;
+}
+
+static void sig_int(int signo)
+{
+	exiting = 1;
+}
+int main(int argc, char **argv)
```
anakryiko, line 128: empty line

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,232 @@
...
+		goto cleanup;
+	}
+
+	obj = bashreadline_bpf__open();
+	if (!obj) {
+		warn("failed to open BPF object\n");
+		goto cleanup;
+	}
+
+	err = bashreadline_bpf__load(obj);
```
anakryiko, line 167: you can combine these into single `bashreadline_bpf__open_and_load()`

libbpf-tools/Makefile
```diff
@@ -16,6 +16,7 @@ $(error Architecture $(ARCH) is not supported yet. Please open an issue)
 endif
 
 APPS = \
+	bashreadline \
```
chenhengqi, line 19: Please also add an entry to `.gitignore`.

libbpf-tools/bashreadline.bpf.c
chenhengqi, line 2: You should update this copyright statement.

chenhengqi: And use C style comment.

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,232 @@
...
+		goto cleanup;
+	}
+
+	func_off = get_elf_func_offset(readline_so_path, "readline");
+	if (func_off < 0) {
+		warn("cound not find readline in %s\n", readline_so_path);
+		goto cleanup;
+	}
+
+	link = bpf_program__attach_uprobe(obj->progs.printret, true, -1,
```
anakryiko, line 179: it's best to set it to `obj->links.printret = bpf_program__attach_uprobe(...)` so that skeleton's destroy() will automatically destroy the link as well. Simpler clean up that way.

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,232 @@
...
+					  readline_so_path, func_off);
```
anakryiko, line 180: don't know if that Github UI rendering it misaligned or not, but please make sure that argument on the new line is aligned with the first argument after (

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,232 @@
...
+	if (err) {
+		warn("failed to attach readline: %d\n", err);
+		goto cleanup;
+	}
+
+	fd = bpf_map__fd(obj->maps.events);
+	if (fd < 0) {
+		warn("failed to open perf fd: %d\n", fd);
+		goto cleanup;
+	}
```
anakryiko, line 191: this can't happen if skeleton was loaded successfully, don't bother checking

libbpf-tools/bashreadline.h
```diff
@@ -0,0 +1,13 @@
...
+#define MAX_LINE_SIZE 80
+
+struct str_t {
+	__u64 pid;
```
chenhengqi, line 9: __u32 ?

libbpf-tools/bashreadline.bpf.c
```diff
@@ -0,0 +1,35 @@
...
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, sizeof(__u32));
+} events SEC(".maps");
+
+SEC("uretprobe/readline")
+int printret(struct pt_regs *ctx) {
+	struct str_t data  = {};
+	char comm[TASK_COMM_LEN] = {};
+	u32 pid;
```
chenhengqi, line 19: You can assign the pid field directly. Please add an empty line below.

libbpf-tools/bashreadline.bpf.c
```diff
@@ -0,0 +1,35 @@
...
+	if (!PT_REGS_RC(ctx))
+		return 0;
+	pid = bpf_get_current_pid_tgid() >> 32;
+	data.pid = pid;
+	bpf_probe_read_user(&data.str, sizeof(data.str) - 1, (void *)PT_REGS_RC(ctx));
+	data.str[sizeof(data.str) - 1] = 0;
+
+	bpf_get_current_comm(&comm, sizeof(comm));
+	if (comm[0] == 'b' && comm[1] == 'a' && comm[2] == 's' && comm[3] == 'h' && comm[4] == 0 ) {
```
chenhengqi, line 28: You can do this filtering before `bpf_probe_read_user`.

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,232 @@
...
+	struct str_t *e = data;
+	struct tm *tm;
+	char ts[16];
+	time_t t;
+
+	time(&t);
+	tm = localtime(&t);
+	strftime(ts, sizeof(ts), "%H:%m:%S", tm);
+
+	printf("%-9s %-6lld %s\n", ts, e->pid, e->str);
```
chenhengqi, line 70: `%7d` for pid. Ubuntu has large pid which mess up the output.

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,232 @@
...
+cleanup:
+	if (line)
+		free(line);
+	return result;
+}
+
+static void sig_int(int signo)
+{
+	exiting = 1;
+}
```
chenhengqi, line 127: Add an empty line below.

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,232 @@
...
+	static const struct argp argp = {
+		.options = opts,
+		.parser = parse_arg,
+		.doc = argp_program_doc,
+	};
+	struct bashreadline_bpf *obj = NULL;
+	struct perf_buffer_opts pb_opts;
+	struct perf_buffer *pb = NULL;
+	struct bpf_link *link = NULL;
+	char *readline_so_path = NULL;
```
chenhengqi, line 139: No need to NULL-initialized these variables.

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,232 @@
...
+	err = bump_memlock_rlimit();
+	if (err) {
+		warn("failed to increase rlimit: %d\n", err);
+		goto cleanup;
+	}
+
+	obj = bashreadline_bpf__open();
+	if (!obj) {
+		warn("failed to open BPF object\n");
+		goto cleanup;
```
chenhengqi, line 164: Nothing to cleanup here, just return.

libbpf-tools/bashreadline.c
chenhengqi, line 179: Assign the return value to `obj->links.printret` and let the skeleton do cleanup for you.

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	fd = bpf_map__fd(obj->maps.events);
+	if (fd < 0) {
+		warn("failed to open perf fd: %d\n", fd);
+		goto cleanup;
+	}
+
+	err = bashreadline_bpf__attach(obj);
```
chenhengqi, line 193: Is this necessary ?

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,232 @@
...
+
+	link = bpf_program__attach_uprobe(obj->progs.printret, true, -1,
+					  readline_so_path, func_off);
+	err = libbpf_get_error(link);
+	if (err) {
+		warn("failed to attach readline: %d\n", err);
+		goto cleanup;
+	}
+
+	fd = bpf_map__fd(obj->maps.events);
```
chenhengqi, line 187: You don't need this fd, right ?

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,232 @@
...
+		if (sp == NULL)
+			goto cleanup;
+		path_start[sp - path_start] = 0;
+		if (strstr(path_start, "/libreadline.so") == NULL)
+			goto cleanup;
+		result = strdup(path_start);
+		break;
+	}
+
+cleanup:
```
chenhengqi, line 118: pclose(fp) ?

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,232 @@
...
+	 * Try to find libreadline.so if the symbol of readline is not
+	 * defined in bash
+	 */
+	fp = popen("ldd /bin/bash", "r");
+	if (fp == NULL)
+		goto cleanup;
+	while (getline(&line, &line_sz, fp) >= 0) {
+		if (strstr(line, "libreadline.so") == NULL)
+			continue;
+		sep = strstr(line, " => ");
```
chenhengqi, line 102: Use `sscanf` ?

chenhengqi (changes requested):

libbpf-tools/bashreadline.c
ThinkerYzu1: fd will be used to create perf_buffer later.

libbpf-tools/bashreadline.c
ThinkerYzu1: Although this is not a perfect parser, sscanf() is not adaptive for changes in the output format.  For example, the length of leading spaces may change.  Does that make sense?

libbpf-tools/bashreadline.c
ThinkerYzu1: Good catch!

libbpf-tools/bashreadline.c
ThinkerYzu1: You are right! This is not necessary.  I didn't know it until I tested it after you told me.

libbpf-tools/bashreadline.bpf.c
ThinkerYzu1: Both lines in C style?

libbpf-tools/bashreadline.c
ThinkerYzu1: It should be Github doing wrong.

libbpf-tools/bashreadline.bpf.c
ThinkerYzu1: So, even it is an uprobe?

libbpf-tools/bashreadline.bpf.c
ThinkerYzu1: What if we want to read a block of binary?  Use another API?

libbpf-tools/bashreadline.h
ThinkerYzu1: It should work.

libbpf-tools/bashreadline.c
ThinkerYzu1: readline_so_path should be released.

libbpf-tools/bashreadline.c
chenhengqi: It's OK. I don't have strong opinion on this.

chenhengqi (approved):
LGTM

libbpf-tools/bashreadline.bpf.c
anakryiko: yeah, uprobe and kprobe are no different in terms of input context, so BPF_KPROBE/BPF_KRETPROBE can be used for uprobe/uretprobe

libbpf-tools/bashreadline.bpf.c
anakryiko: oh, wait, I missed that this is bpf_probe_read, not bpf_probe_read_str(). Shouldn't bpf_probe_read_user_str() be used here?

libbpf-tools/bashreadline.bpf.c
```diff
@@ -0,0 +1,38 @@
...
+#define TASK_COMM_LEN 16
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, sizeof(__u32));
+} events SEC(".maps");
+
+SEC("uretprobe/readline")
+int BPF_KRETPROBE(printret) {
```
anakryiko, line 17: you can do

```
int BPF_KRETPROBE(printret, const void *ret) {
  /* use ret instead of PT_REGS_RC() */
}
```

libbpf-tools/bashreadline.bpf.c
```diff
@@ -0,0 +1,38 @@
...
+	if (!PT_REGS_RC(ctx))
+		return 0;
+
+	bpf_get_current_comm(&comm, sizeof(comm));
+	if (comm[0] != 'b' || comm[1] != 'a' || comm[2] != 's' || comm[3] != 'h' || comm[4] != 0 )
+		return 0;
+
+	pid = bpf_get_current_pid_tgid() >> 32;
+	data.pid = pid;
+	bpf_probe_read_user(&data.str, sizeof(data.str), (void *)PT_REGS_RC(ctx));
```
anakryiko, line 31: see my other comment, if it has to be `bpf_probe_read_user()`, then zero termination was correct, sorry about that. But I think it should be `bpf_probe_read_user_str()`. Can you please check if the string is a proper C string, or it's actually a collection of zero-separated strings?

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,218 @@
...
+		if (sep == NULL)
+			goto cleanup;
+		path_start = sep + 4;
+		while (isspace(*path_start))
+			path_start++;
+		sp = strstr(path_start, " ");
+		if (sp == NULL)
+			goto cleanup;
```
anakryiko, line 111: tbh, I'm a bit lost of what kind of string we are parsing here, can you please add a comment with an example? It would make it easier to follow and maybe someone can suggest a more succinct way to go about that

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,218 @@
...
+	if (err) {
+		warn("failed to attach readline: %d\n", err);
+		goto cleanup;
+	}
+
+	fd = bpf_map__fd(obj->maps.events);
+
+	pb_opts.sample_cb = handle_event;
+	pb_opts.lost_cb = handle_lost_events;
+	pb = perf_buffer__new(fd, PERF_BUFFER_PAGES, &pb_opts);
```
anakryiko, line 188: nit: you could have just inlined bpf_map__fd(obj->maps.events) here, it would be fine

libbpf-tools/bashreadline.bpf.c
ThinkerYzu1: bpf_probe_read_user_str() works well.  When a user types 80 or more chars, it cuts the string to 79 bytes and terminates it with a null char.   With bpf_probe_read_user() will return a string of 80 bytes without a null char.
It just happens to work while the memory is clean.

And, yes, it is a proper C string.

libbpf-tools/bashreadline.c
anakryiko: ok, so now that I see what we are parsing, I think sscanf actually makes total sense. You were worried about sscanf() not handling leading spaces. That's not true. If you specify format string as "%s => %s" sscanf() will skip any number of whitespaces (tabs and spaces, maybe some other whitespaces) between the first string, "=>", and second string. So I think sscanf("") makes sense here. It will make code much shorter and more understandable.

ThinkerYzu1:
I found with sscanf() it will be even more complicated.  sscanf() doesn't
know the size of buffers, it fills as many characters as possible.
To avoid buffer overflows, two buffers are allocated as big as 'line'
itself, and more checks are added to make sure everything works as expected.

chenhengqi:
FYI

```c
#include <stdio.h>
#include <stdbool.h>

int main()
{
	char line[256];
	FILE *fp;
	int ret;

	fp = popen("ldd /bin/bash", "r");
	if (!fp)
		return -1;

	while (true) {
		ret = fscanf(fp, "%*s => %s %*s\n", line);
		if (ret == EOF && feof(fp))
			break;
		if (ret != 1)
			continue;
		printf("%s\n", line);
	}

	return 0;
}

```

ThinkerYzu1:
Input can be very long. This can cause an overflow.
To do it correctly, a buffer big enough should be allocated.

But, after looking over the man page again, I found a feature I didn't know
before.
I can use

    char *line;
    fscanf(fp, "%*s => %ms", &line);

It will allocate a buffer for us.

ThinkerYzu1:
Just check this feature, '%ms", it seems to be a new feature introduced by
POSIX.1-2017 released by 2018.
There are always new things to learn.

anakryiko:
> Just check this feature, '%ms", it seems to be a new feature introduced by POSIX.1-2017 released by 2018. There are always new things to learn.

Yeah, but this will allocate memory for each line, which is suboptimal if there are many lines. But note that sscanf supports limiting the length of a string to read:

sscanf(str, "%3s", buf);

This will require buf of at least (3 + 1) characters. So I think in this case the simplest solution is to allocate local buffer of, say, 128 bytes, and read %127s string. Simple, no allocations, no freeing.

libbpf-tools/bashreadline.c
```diff
@@ -0,0 +1,212 @@
...
+	 *
+	 *      libreadline.so.8 => /usr/lib/libreadline.so.8 (0x00007b....)
+	 *
+	 * Here, it finds a line with libreadline.so and extracts the
+	 * path after the arrow, '=>', symbol.
+	 */
+	fp = popen("ldd /bin/bash", "r");
+	if (fp == NULL)
+		goto cleanup;
+	while ((input_n = fscanf(fp, "%*s => %127s", line)) != EOF) {
```
anakryiko, line 109: even if this works, it will work by accident. You need to read a full line first, and then do sscanf(). I thought we already discussed this? See the following:

```
$ git diff
diff --git a/libbpf-tools/bashreadline.c b/libbpf-tools/bashreadline.c
index efe41bb5603c..d32a7180e227 100644
--- a/libbpf-tools/bashreadline.c
+++ b/libbpf-tools/bashreadline.c
@@ -87,6 +87,7 @@ static char *find_readline_so()
        int input_n;

        func_off = get_elf_func_offset(bash_path, "readline");
+       func_off = -1;
        if (func_off >= 0)
                return strdup(bash_path);

@@ -107,11 +108,12 @@ static char *find_readline_so()
        if (fp == NULL)
                goto cleanup;
        while ((input_n = fscanf(fp, "%*s => %127s", line)) != EOF) {
+               printf("line is '%s'\n", line);
                if (input_n <= 0)
                        continue;
                if (strstr(line, "/libreadline.so")) {
                        result = strdup(line);
-                       break;
+                       //break;
                }
        }


$ ldd /bin/bash
        linux-vdso.so.1 (0x00007ffe019d4000)
        libtinfo.so.6 => /lib64/libtinfo.so.6 (0x00007fc8cf7eb000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007fc8cf5e7000)
        libc.so.6 => /lib64/libc.so.6 (0x00007fc8cf222000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fc8cfd42000)

$ sudo ./bashreadline
line is ''
line is ''
line is '/lib64/libtinfo.so.6'
line is '/lib64/libtinfo.so.6'
line is '/lib64/libdl.so.2'
line is '/lib64/libdl.so.2'
line is '/lib64/libc.so.6'
line is '/lib64/libc.so.6'
line is '/lib64/libc.so.6'
failed to find readline
```

Note the output of line. You should be getting only one instance of each library each time for this to work reliably, but clearly you are not. You fscanf() is just skipping at least one string from the input, but that's working by accident. You actually want to skip unrecognized line on each iteration, not parts of that line.

So please reimplement and validate it's working as expected.

anakryiko (approved):
looks correct now, thanks!

Issue #3682, opened 2021-10-29
[Proposal] Remove B frontend and P4 language support


davemarchevsky:
## What

The B language is a custom tracing language supported by BCC ([example program](https://github.com/iovisor/bcc/blob/master/tests/python/test_trace1.b), [frontend directory w/ lexer, parser, loader](https://github.com/iovisor/bcc/tree/master/src/cc/frontends/b)). [P4](https://en.wikipedia.org/wiki/P4_(programming_language)) is a packet processing DSL which bcc can translate into bcc-style BPF C programs ([example program](https://github.com/iovisor/bcc/blob/master/src/cc/frontends/p4/test/testprograms/basic_routing.p4), [frontend directory](https://github.com/iovisor/bcc/tree/master/src/cc/frontends/p4)). Both allow users to write high-level BPF programs.

The B support is implemented as a proper LLVM frontend which has some interaction with BCC's custom FrontendAction and loader functionality, while the P4 support is in the form of a transpiler that produces bcc-style C programs.

## Why

It's unclear if anyone is using these two languages to write BPF programs today or actively maintaining BPF programs written in these languages. Neither language's frontends have been touched in many years (active development seems to have been 2015-2017 for both) and none of the folks actively working on BCC tools and core functionality are familiar with them. The community overwhelmingly uses bcc- and libbpf-style C for tools.

I'm currently refactoring and simplifying BCC's internals to better match libbpf loader conventions. The internals have been stable for the past few years, so these large changes are likely to break B language support in subtle ways. This is a reasonable time to announce that B is no longer supported and rip out the integration. The alternative is continuing to make small fixes to ensure code compiles and tests pass, but not having confidence that edge cases are covered and things actually work. 

Although P4's transpilation is unlikely to break as a result of these changes, the other logic for removal still holds.

## When

This issue will remain open for at least a month before any action is taken

## Alternatives

Folks from the community volunteering to maintain the B frontend and P4 transpiler would be a viable alternative. Learning about current use cases for these languages might convince folks who actively maintain bcc to pay some attention to B and P4.

davemarchevsky:
#3759 did this

PR #3759, opened 2021-12-17
Remove support for B and P4 languages

davemarchevsky:
See #3682 for more detail.

These commits remove all of the P4 language code and vast majority of the B language code. There may be some vestigial B stuff in files like `bpf_module.cc`, but they can be removed in followups. There are likely some refactoring opportunities as well due to this, they aren't explored here so as to keep the PR as simple as possible.


4ast (approved):
Long overdue.

PR #3762, opened 2021-12-18
libbpf-tools: Add tcpstates

chenhengqi:
Add new libbpf tool tcpstates.

### DIFF
--- a/libbpf-tools/tcpstates.c
+++ b/libbpf-tools/tcpstates.c
@@ -1,0 +1,297 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+
+/*
+ * tcpstates    Trace TCP session state changes with durations.
+ * Copyright (c) 2021 Hengqi Chen
+ *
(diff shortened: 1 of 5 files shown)


libbpf-tools/tcpstates.c
```diff
@@ -0,0 +1,274 @@
...
+	struct tcpstates_bpf *obj;
+	int err, port_map_fd;
+	short port_num;
+	char *port;
+
+	err = argp_parse(&argp, argc, argv, 0, NULL, NULL);
+	if (err)
+		return err;
+
+	libbpf_set_strict_mode(LIBBPF_STRICT_ALL);
```
anakryiko, line 191: when I was updating all the tools with `libbpf_set_strict_mode()` I've noticed that a bunch of tools don't support verbose mode and don't set custom libbpf print callback. Please do that so that it's possible to see what libbpf is doing and/or debug more issues without having to patch and recompile tools (ideally we would add -v support in every tool that we missed this in). Thank you!

chenhengqi: Sure, will do.

PR #3769, opened 2021-12-20
docker + tests: Run tests using python3, refactor Dockerfile

davemarchevsky:
In #3707, I added ubuntu-20.04 to the list of OS's bcc-test uses to run
tests. I expected that this would result in the tests running on both
18.04 and 20.04. Unfortunately this change was effectively a no-op as
the tests are run in a Docker container and the os field in bcc-test.yml
dictates the type/version of the _host_ OS, not the container's OS.

So it's not necessary to run on both, just run on 20.04. It will be
necessary to modify Dockerfile.test to use both 18.04 and 20.04. To
prepare for this, move all python tests to use python3 interpreter, as
20.04 doesn't have python2 pip readily available.

Also, refactor Dockerfile.tests a bit so that it's possible to provide
ubuntu version and shortname as build input.

This commit does not result in the docker test container working/running
both 18.04 and 20.04, rather lays groundwork for future commits to do
so.

### DIFF
--- a/.github/workflows/bcc-test.yml
+++ b/.github/workflows/bcc-test.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-18.04, ubuntu-20.04] # 18.04.3 release has 5.0.0 kernel                                        
+        os: [ubuntu-20.04]
         env:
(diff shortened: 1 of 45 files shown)


davemarchevsky:
The two `18.04` will never run as this PR removes them . `20.04` tests give same signal.

PR #3770, opened 2021-12-20
docker: Bump default LLVM used for tests to 11

davemarchevsky:
This is another prep commit for running docker tests on both 18.04 and
20.04 versions of ubuntu. 20.04 doesn't have some LLVM 8 libs that the
docker build expects, so bump to 11.

Bumping to 11 causes test_disassembler to fail because it expects a map
w/ a certain format with FD 3, and the LLVM bump causes the map to be
FD 4. The purpose of the test is to ensure that the format of the
diassembled map is correct, not that it has a specific FD, so make some
small changes to the test so that it doesn't care what the FD number is.


davemarchevsky:
I believe "Publish to Dockerhub" is failing due to an active issue with docker registry: https://status.docker.com/pages/533c6539221ae15e3f000031

Will give it an hour or two and re-run

PR #3771, opened 2021-12-21
docker: Run tests on ubuntu-20.04 as well

davemarchevsky:
Move bcc-test github action's `matrix.os` to indicate the OS to run in
the container, not the test runner itself (just pin the latter to
ubuntu-20.04).

Also, fixup some tests that were failing when trying to run on 20.04
manually.

### DIFF
--- a/.github/workflows/bcc-test.yml
+++ b/.github/workflows/bcc-test.yml
@@ -8,10 +8,10 @@ on:
 
 jobs:
   test_bcc:
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
-        os: [ubuntu-20.04]
+        os: [{version: "18.04", nick: bionic}, {version: "20.04", nick: focal}]
         env:
         - TYPE: Debug
           PYTHON_TEST_LOGFILE: critical.log
@@ -25,7 +25,10 @@ jobs:
         ip addr
     - name: Build docker container with all deps
       run: |
-        docker build -t bcc-docker -f Dockerfile.tests .
+        docker build \
+        --build-arg UBUNTU_VERSION=${{ matrix.os.version }} \
+        --build-arg UBUNTU_SHORTNAME=${{ matrix.os.nick }} \
+        -t bcc-docker -f Dockerfile.tests .
     - name: Run bcc build
       env: ${{ matrix.env }}
       run: |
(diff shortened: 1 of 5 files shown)


tools/memleak.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
```
yonghong-song, line 1: Any particular reason we changed memleak.py to python3?

davemarchevsky: I made this change before adding the 

```
RUN if [ ! -f "/usr/bin/python" ]; then ln -s /bin/python3 /usr/bin/python; fi
```

to `Dockerfile.tests` in this PR. The `ubuntu-20.04` docker container doesn't have a `/usr/bin/python` by default, so `memleak` was failing to run at all.

I can probably get rid of the explicit python3 shebang here and elsewhere, but there maybe is a case for keeping it explicit:
* `ubuntu-20.04` container doesn't really support python2 to the extent we'd need to run tests with it. There doesn't seem to be any `pip` in official repo for py2 on that version. When I was experimenting with test updates I tried manually downloading-and-building pip for py2, but this felt unnecessarily complex relative to just using py3.
* If we want to continue to run tests on `ubuntu-18.04` container for foreseeable future, it would be nice to ensure that they're running on python3 so as not to be testing on different python versions without making that obvious to the user. So worth staying consistent with `20.04` for now, and potentially adding `18.04+python2` vs `18.04+python3` test runs later.

I don't feel strongly about this though. Main thing I was optimizing for here was using same dockerfile for both ubuntu versions with minimal changes

yonghong-song: cc @brendangregg I agree python3 is the trend, but from some posts in the mailing list, it looks like python2 is still used, so I think we should continue to support python2.

Agree we should try to test both in 18.04.

If we change to python3 explicitly for memleak.py, but not others, it will look strange and that is my question above.
Maybe we can drop this particular for now? Any test will fail?

davemarchevsky: i will give it a try. 

FWIW there was a mass conversion of python tests to explicit `python3` in #3769. I can try to change that back as well, but guessing it's not as big of a deal as they're all tests, not tools.

yonghong-song: I see tests are using python3. That is not critical as you said. But for tools, we have to be careful.

PR #3773, opened 2021-12-25
libbpf-tools: Support verbose mode and set custom libbpf print callback

chenhengqi:
Add verbose option and set custom libbpf print callback in all tools.
This would allow user to print libbpf debug log and aid the debugging
process.

### DIFF
--- a/libbpf-tools/ksnoop.c
+++ b/libbpf-tools/ksnoop.c
@@ -33,6 +33,7 @@ enum log_level {
 };
 
 static enum log_level log_level = WARN;
+static bool verbose = false;
 
 static __u32 filter_pid;
 static bool stack_mode;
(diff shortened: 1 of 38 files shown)


chenhengqi:
@davemarchevsky Please take a look.  Thanks.

libbpf-tools/ksnoop.c
```diff
@@ -251,7 +252,7 @@ static int get_func_btf(struct btf *btf, struct func *func)
 	return 0;
 }
 
-int predicate_to_value(char *predicate, struct value *val)
```
davemarchevsky, line 254: I'm fine with this change, but can you confirm that it's independent of the others in this file? If it's not, then I must be confused

davemarchevsky (approved):
LGTM aside from one question

libbpf-tools/ksnoop.c
chenhengqi: Yes, it's independent.

PR #3774, opened 2021-12-27
docs: Fixup cpudist pid example in man page

MdSalih:
-P prints all by process id
-p prints specific process id

### DIFF
--- a/man/man8/cpudist.8
+++ b/man/man8/cpudist.8
@@ -68,9 +68,9 @@ Print 1 second summaries, using milliseconds as units for the histogram, and inc
 #
 .B cpudist \-mT 1
 .TP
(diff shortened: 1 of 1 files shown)


man/man8/cpudist.8
```diff
@@ -70,7 +70,7 @@ Print 1 second summaries, using milliseconds as units for the histogram, and inc
 .TP
 Trace PID 186 only, 1 second summaries:
 #
-.B cpudist -P 185 1
+.B cpudist -p 185 1
```
chenhengqi, line 73: `185` here does not match the above description, please fix it.

MdSalih: fixed

Issue #2662, opened 2019-12-22
funccount and trace can't get address of struct function of go package


tty5:
/usr/share/bcc/tools/funccount './a.out:github.com/sirupsen/logrus.(*Logger).SetOutput'

No functions matched by pattern ^github.com/sirupsen/logrus.(.*Logger).SetOutput$

is there any solution?

tty5:
root cause is Go language will use paran in function name for struct's function.
Like github.com/sirupsen/logrus.(*Logger).SetOutput, but paran is
use to get sig part in function _parse_probe

i has submit PR https://github.com/iovisor/bcc/pull/2663 to add go function support

PR #3776, opened 2021-12-28
bcc/tools: Allow trace.py to trace golang functions

chenhengqi:
Golang function symbols can contain `(`, which is excluded
by the original implementation. Thus oolang function symbol like:

    example.com/some-project/foo.(*Bar).Buz

is not allowed. Let's remove this constraint. Now we can do:

    trace 'p:/path/to/binary:example.com/some-project/foo.(*Bar).Buz "%llx", arg1'

### DIFF
--- a/tools/trace.py
+++ b/tools/trace.py
@@ -110,7 +110,7 @@ class Probe(object):
                 #                                          opt. signature
                 #                               probespec       |      rest
                 #                               ---------  ----------   --
-                (spec, sig, rest) = re.match(r'([^ \t\(]+)(\([^\(]*\))?(.*)',
+                (spec, sig, rest) = re.match(r'([^ \t]+)(\([^\(]*\))?(.*)',
                                              text).groups()
 
                 self._parse_spec(spec)


yonghong-song:
@shadyabhi could you check whether the fixed trace.py by @chenhengqi works for you or not?

davemarchevsky:
Looks like #2662 is noting the same issue and #2663 is a similar attempt to fix.

I think this may break existing usecases, like [these examples](https://github.com/iovisor/bcc/blob/master/tools/trace_example.txt#L218-L234) from `trace_example.txt`:

```
trace 'p:c:open(char *filename) "opening %s", filename'
trace 'p::SyS_nanosleep(struct timespec *ts) "sleep for %lld ns", ts->tv_nsec'
```

Using the first command as an example, looks like before this PR the `spec` is `p:c:open`, while after this PR it's `p:c:open(char`.

Instead of requiring folks to add a space before the optional function signature, maybe we could:
* Have a special flag for golang that enables this logic
* Take advantage of the `.(*` pattern that these golang function symbols seem to have:

```
github.com/sirupsen/logrus.(*Logger).SetOutput
example.com/some-project/foo.(*Bar).Buz
```

to disambiguate, since the dot isn't a valid char for c symbols.

IMO the latter is a better path forward if it doesn't subtly break something I'm not aware of. #2663 looks to be doing something similar to this:

```
if there are more letter after ), it consider be a go function.
```

chenhengqi:
It's hard to handle Go's function signatures and parameters. I think it deserve a separate tool or just use bpftrace.

For go's gc toolchain, it use a different calling convention. To print something useful from the stack, we need type information, which make it hard to implement and use.

PR #3778, opened 2021-12-29
add fedora docker tests + move dockerfiles to a new dir

davemarchevsky:
Dockerfiles are cluttering up the main repo dir, so move them to a
newly-created 'docker' dir.

Add a fedora dockerfile and use it in bcc-test workflow.


davemarchevsky:
I'm going to merge this, but for now will not add the fedora tests to the list of "Required" tests that block merge. There are some issues that should be addressed before the fedora tests can be considered at parity with the Ubuntu tests:

* The "Run bcc build" step links against LLVM as a shared library (`-DENABLE_LLVM_SHARED=ON`), which Ubuntu tests don't do
* Two lua tests kept failing (`test_lua_uprobes`, `test_lua_standalone`) for reasons I couldn't track down. The equivalent python `uprobes` test passes on Fedora, so it seems to be a Fedora-specific thing. Added a `RUN_LUA_TESTS` CMake option (and set `-DRUN_LUA_TESTS=OFF` in "Run bcc build" step) for now. Better to have the rest of the tests pass and run on future PRs. Will create an issue to fix this.

/cc @yonghong-song

davemarchevsky:
I think the bullet point about failing lua tests was incorrect. On Ubuntu, the bcc build step says:

```
Could NOT find LuaJIT (missing: LUAJIT_LIBRARIES LUAJIT_INCLUDE_DIR) 
```

While on Fedora:

```
-- Found LuaJIT: /usr/lib64/libluajit-5.1.so;/usr/lib64/libdl.so;/usr/lib64/libm.so  
```

If LuaJIT is not found, Lua tests will not be run (see `tests/lua/CMakeLists.txt`). So it's possible that the tests don't work on Ubuntu either. Will investigate.

.github/workflows/bcc-test.yml
```diff
@@ -94,6 +94,86 @@ jobs:
         name: critical-tests-${{ matrix.env['TYPE'] }}-${{ matrix.os.version }}
         path: tests/python/critical.log
 
+  test_bcc_fedora:
+    runs-on: ubuntu-20.04
```
yonghong-song, line 98: test_bcc_fedora and runs-on ubuntu-20.04? I do see we are using `Dockerfile.fedora` below. So `ubuntu-20.04` is a copy-paste error?

davemarchevsky: No, this is intentional. `ubuntu-20.04` here describes the OS of the "GitHub actions runner" VM. List of supported runner OS's [here](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) doesn't include fedora.

The docker container is Fedora, though. So aside from the kernel still being `ubuntu-20.04`'s, tests are running using userspace and toolchains of Fedora.

yonghong-song: for ubuntu run:
>  Could NOT find LuaJIT (missing: LUAJIT_LIBRARIES LUAJIT_INCLUDE_DIR) 
luajit package may be missing in the system. Will need to figure how to get the package properly.

yonghong-song: > No, this is intentional. `ubuntu-20.04` here describes the OS of the "GitHub actions runner" VM. List of supported runner OS's [here](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) doesn't include fedora.
> 
> The docker container is Fedora, though. So aside from the kernel still being `ubuntu-20.04`'s, tests are running using userspace and toolchains of Fedora.

Okay, utunbu vm and fedora docker container. Make sense. thanks for explanation.

PR #3779, opened 2021-12-30
Build libbpf tools as part of fedora test run

davemarchevsky:
This makes no attempt to CMake-ify `libbpf-tools/Makefile`, so the build happens as a separate step

There's currently nothing using the generated executables to test, but ostensibly the `tests/python/test_tools_smoke.py` could be extended to do so. May be worth merging independently as there are lots of outstanding libbpf-tools PRs and quick feedback on whether build succeeds would be useful

Why fedora only? Installing `linux-tools-generic` package on Ubuntu, which contains `bpftool` causes problems since Ubuntu's `/usr/sbin/bpftool` is a script wrapper which tries to ensure that the installed bpftool (located in `/usr/lib/linux-tools/...`) kernel version matches the running kernel. Normally a useful thing to check. Also bpftools on 18.04 and 20.04 are too old to gen skel. All can be worked around, but needs some thought.


davemarchevsky:
I expect "Dummy commit" Fedora builds to fail, then builds off a revert commit to succeed

PR #3784, opened 2022-01-03
bcc: Fix array type handling due to llvm changes

chenhengqi:
The llvm commit aee49255074f ([0]) changes array type
from `int [4]` to `int[4]` (with space removed), which
breaks the assumption in BCC. This commit fixes this
issue and adds a comment to the related code. While at
it, also remove execution permission of file `table.py`.

  [0]: https://github.com/llvm/llvm-project/commit/aee49255074fd4ef38d97e6e70cbfbf2f9fd0fa7

### DIFF
--- a/src/python/bcc/table.py
+++ b/src/python/bcc/table.py
@@ -227,8 +227,8 @@ def _get_event_class(event_map):
                    'unsigned __int128' : (ct.c_ulonglong * 2),
                    'void *'            : ct.c_void_p }
 
-    # handle array types e.g. "int [16] foo"
-    array_type = re.compile(r"(.+) \[([0-9]+)\]$")
+    # handle array types e.g. "int [16]" or "char[16]"
+    array_type = re.compile(r"([^ ]+) ?\[([0-9]+)\]$")
 
     fields = []
     num_fields = lib.bpf_perf_event_fields(event_map.bpf.module, event_map._name)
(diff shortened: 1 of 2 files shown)


chenhengqi:
cc @yonghong-song

src/python/bcc/table.py
```diff
@@ -227,8 +227,8 @@ def _get_event_class(event_map):
                    'unsigned __int128' : (ct.c_ulonglong * 2),
                    'void *'            : ct.c_void_p }
 
-    # handle array types e.g. "int [16] foo"
-    array_type = re.compile(r"(.+) \[([0-9]+)\]$")
+    # handle array types e.g. "int [16]" or "char[16]"
+    array_type = re.compile(r"([^ ]+) ?\[([0-9]+)\]$")
```
yonghong-song, line 231: Does this match type like `struct foo[16]`?

chenhengqi: No, neither the original implementation.

yonghong-song: > No, neither the original implementation.

Okay, I see. I though `(.+) \[([0-9]+)\]$` may cover `struct foo [16]` with hoping `(.+) to cover `struct foo` but apparently this is not the case.

This functionality is used for python and we don't really have complicated python tools and that is probably reason we didn't hit `struct foo[16]` issue. I will merge this patch as it will fix a regression due to latest llvm. If we ever hit `struct foo[16]` issue later, we can address it then.

PR #3786, opened 2022-01-04
bcc: support BPF_MAP_TYPE_{INODE, TASK}_STORAGE maps

chenhengqi:
Add support for BPF_MAP_TYPE_{INODE, TASK}_STORAGE in BCC.
Like sk local storage, this commit allows creating inode/task
local storage using BPF_{INODE, TASK}_STORAGE macros, and
manipulating maps using map.{inode, task}_storage_get() and
map.{inode, task}_storage_delete() helpers.

Also add examples to demostrate BPF_{INODE, TASK}_STORAGE usage.


yonghong-song (commented):
Thanks for the implementation. This is really useful, esp. task local storage. Currently quiet some bcc tools use a hash map with tid as the key to preserve data between different bpf programs. We could use task local storage here.

So for example task_storate.py, I suggest to make it a little bit more omplex. For example, we can have two programs, one KFUNC and another KRETFUNC, we can use task local storage to pass some data (e.g., timestamp in order to calculate latency) from KFUNC program to KRETFUNC program. This example can serve as a template for bcc tools use case with task local storage. WDYT?

chenhengqi:
OK, will do.

chenhengqi:
Updated.

yonghong-song:
LGTM. Thanks!

PR #3787, opened 2022-01-04
Vendor deprecated btf_ext APIs and structs

davemarchevsky:
Some btf_ext-related APIs in libbpf are being deprecated because they
make incorrect assumptions. They're being used only by bcc currently, so
vendor them before they get deleted.

After / as part of #3660, may need to revisit the incorrect assumptions
being made here.

The functions and structs were ripped directly from libbpf with minimal
changes:
  * Change void* arithmetic to uint8_t
  * __u32 -> uint32_t and similar
  * Add a wrapping namespace
  * `rec_size` functions were not needed - just grab the rec_size
directly since type is no longer opaque to bcc

### DIFF
--- a/src/cc/bcc_btf.cc
+++ b/src/cc/bcc_btf.cc
@@ -21,10 +21,330 @@
 #include "libbpf.h"
 #include "bcc_libbpf_inc.h"
 #include <vector>
+#include <byteswap.h>
 
 #define BCC_MAX_ERRNO       4095
 #define BCC_IS_ERR_VALUE(x) ((x) >= (unsigned long)-BCC_MAX_ERRNO)
 #define BCC_IS_ERR(ptr) BCC_IS_ERR_VALUE((unsigned long)ptr)
+#ifndef offsetofend
+# define offsetofend(TYPE, FIELD) \
+	        (offsetof(TYPE, FIELD) + sizeof(((TYPE *)0)->FIELD))
+#endif
+
+namespace btf_ext_vendored {
+
+/* The minimum bpf_func_info checked by the loader */
+struct bpf_func_info_min {
+        uint32_t   insn_off;
+        uint32_t   type_id;
+};
+
+/* The minimum bpf_line_info checked by the loader */
+struct bpf_line_info_min {
+        uint32_t   insn_off;
+        uint32_t   file_name_off;
+        uint32_t   line_off;
+        uint32_t   line_col;
+};
+
+struct btf_ext_sec_setup_param {
+        uint32_t off;
+        uint32_t len;
+        uint32_t min_rec_size;
+        struct btf_ext_info *ext_info;
+        const char *desc;
+};
+
+static int btf_ext_setup_info(struct btf_ext *btf_ext,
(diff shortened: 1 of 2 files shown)


davemarchevsky:
need to remove __bcc prefix from things

src/cc/bcc_btf.cc
```diff
@@ -21,10 +21,330 @@
...
+        }
+
+        err = btf_ext_setup_core_relos(btf_ext);
+        if (err)
+                goto done;
+
+done:
+        if (err) {
+                btf_ext__free(btf_ext);
+                return (struct btf_ext*)err;
```
yonghong-song, line 276: err is 'int' type, converting to 'struct btf_ext *' will trigger the warning:
```
/home/yhs/work/bcc/src/cc/bcc_btf.cc: In function ‘btf_ext_vendored::btf_ext* btf_ext_vendored::btf_ext__new(const uint8_t*, uint32_t)’:                                            
/home/yhs/work/bcc/src/cc/bcc_btf.cc:276:41: warning: cast to pointer from integer of different size [-Wint-to-pointer-cast]                                                        
                 return (struct btf_ext*)err;                                                                                                                                       
                                         ^~~ 
```

You can do `(struct btf_ext *)(long)err`.

yonghong-song (commented):
btf_load() is also deprecated.
```
/home/yhs/work/bcc/src/cc/bcc_btf.cc:600:20: warning: ‘int btf__load(btf*)’ is deprecated: libbpf v0.6+: use btf__load_into_kernel instead [-Wdeprecated-declarations]
```

Maybe to check to use btf__load_into_kernel()?

If we want to accommodate the external libbpf build, you can copy btf__load() function as well. Hopefully, eventually all these (bpf_ext* and btf__load) are gone and handled by libbpf library directly.

I did a simple test with profile.py with this patch and dump program xlated code. It appears the patch is working and I can see proper source annotation in xlated code.

chenhengqi:
Three libbpf APIs used by BCC are deprecated.

* bpf_create_map_xattr -> bpf_map_create
* bpf_load_program_xattr -> bpf_prog_load
* btf__load -> btf__load_into_kernel

@davemarchevsky  Are you going to address these in this PR ?

Or I may pick this up.

davemarchevsky:
> Maybe to check to use btf__load_into_kernel()?

> @davemarchevsky Are you going to address these in this PR ?

I think those migrations away from other deprecated stuff should be addressed in a separate PR. @chenhengqi if you'd like to do it plz feel free, otherwise I will pick up late this week / early next.

yonghong-song:
> I think those migrations away from other deprecated stuff should be addressed in a separate PR. @chenhengqi if you'd like to do it plz feel free, otherwise I will pick up late this week / early next.

okay. let us address btf__load API in the separate PR. Please fix the warning and I think we will be good to go.

src/cc/bcc_btf.cc
davemarchevsky: fixed, but with a `uintptr_t` cast instead of `long`

Issue #3793, opened 2022-01-06
[Feature request] Bulk polling API for perf ring buffer


yzhao1012:
Perf ring buffer (or perf buffer) created with BPF_PERF_OUTPUT(<name>). BPF runtime can submit events to perf buffer via perf_submit() API. Userspace call PerfBuffer::poll() to invoke callbacks assigned during the initialization of the perf buffer.

IIUC, the perf buffer polling will repetitively invoke the callback on each and every event in the perf buffer. That can be wasteful when there is a lot of events flowing through the perf buffer.

Instead, I am wondering if a bulk polling API would be more efficient.

chenhengqi:
Maybe you are looking for [BPF ring buffer](https://nakryiko.com/posts/bpf-ringbuf/#bpf-ringbuf-data-notification-control).

yzhao1012:
Thanks for the pointer, @chenhengqi !

I read the article, it offers a lot of new information for me. We'll explore ringbuf.
But for now, our existing product need to support 4.14 kernel which has no ringbuf.
So we have to look for alternatives in perfbuf (or perf ring buffer as I mentioned above, I believe they are referring to the same thing).

I noticed that in the referred article ([BPF ringbuf: data notification control](https://nakryiko.com/posts/bpf-ringbuf/#bpf-ringbuf-data-notification-control)), perfbuf also has a sampled notification option, as explained below:

_Perfbuf handles that with the ability to set up sampled notification, in which case only every Nth sample will send a notification. You can do that when creating a BPF perfbuf map from the user-space._

I tried to look for any relevant APIs in https://github.com/iovisor/bcc/blob/155d8ab0f3b8ea0e542570dcd98842daadaf22da/src/cc/libbpf.h
https://github.com/iovisor/bcc/blob/155d8ab0f3b8ea0e542570dcd98842daadaf22da/src/cc/api/BPFTable.h
But couldn't find obvious matches.

Are you aware of how to configure sampled notification for perfbuf? Any pointers to the kernel API, and/or BCC APIs?

chenhengqi:
@yzhao1012  I just opened #3801 , can you give it a try ?

yzhao1012:
Thanks Hengqi! I saw that. I'll try it ASAP.

PR #3801, opened 2022-01-08
bcc: Allow specify wakeup_events for perf buffer

chenhengqi:
This commit adds a new option `wakeup_events` to the open_perf_buffer API.
This provides an alternative to solve #3793.

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -1408,8 +1408,22 @@ int bpf_attach_lsm(int prog_fd)
 
 void * bpf_open_perf_buffer(perf_reader_raw_cb raw_cb,
                             perf_reader_lost_cb lost_cb, void *cb_cookie,
-                            int pid, int cpu, int page_cnt) {
-  int pfd;
+                            int pid, int cpu, int page_cnt)
+{
+  struct bcc_perf_buffer_opts opts = {
+    .pid = pid,
+    .cpu = cpu,
+    .wakeup_events = 1,
+  };
+
+  return bpf_open_perf_buffer_opts(raw_cb, lost_cb, cb_cookie, page_cnt, &opts);
+}
+
+void * bpf_open_perf_buffer_opts(perf_reader_raw_cb raw_cb,
+                            perf_reader_lost_cb lost_cb, void *cb_cookie,
+                            int page_cnt, struct bcc_perf_buffer_opts *opts)
+{
+  int pfd, pid = opts->pid, cpu = opts->cpu;
   struct perf_event_attr attr = {};
   struct perf_reader *reader = NULL;
 
@@ -1421,7 +1435,7 @@ void * bpf_open_perf_buffer(perf_reader_raw_cb raw_cb,
   attr.type = PERF_TYPE_SOFTWARE;
   attr.sample_type = PERF_SAMPLE_RAW;
   attr.sample_period = 1;
-  attr.wakeup_events = 1;
+  attr.wakeup_events = opts->wakeup_events;
   pfd = syscall(__NR_perf_event_open, &attr, pid, cpu, -1, PERF_FLAG_FD_CLOEXEC);
   if (pfd < 0) {
     fprintf(stderr, "perf_event_open: %s\n", strerror(errno));
--- a/src/cc/libbpf.h
+++ b/src/cc/libbpf.h
@@ -35,6 +35,12 @@ enum bpf_probe_attach_type {
 	BPF_PROBE_RETURN
 };
 
+struct bcc_perf_buffer_opts {
+  int pid;
+  int cpu;
+  int wakeup_events;
+};
+
 int bcc_create_map(enum bpf_map_type map_type, const char *name,
                    int key_size, int value_size, int max_entries,
                    int map_flags);
@@ -106,6 +112,10 @@ int kernel_struct_has_field(const char *struct_name, const char *field_name);
 void * bpf_open_perf_buffer(perf_reader_raw_cb raw_cb,
                             perf_reader_lost_cb lost_cb, void *cb_cookie,
                             int pid, int cpu, int page_cnt);
+
+void * bpf_open_perf_buffer_opts(perf_reader_raw_cb raw_cb,
+                            perf_reader_lost_cb lost_cb, void *cb_cookie,
+                            int page_cnt, struct bcc_perf_buffer_opts *opts);
 
 /* attached a prog expressed by progfd to the device specified in dev_name */
 int bpf_attach_xdp(const char *dev_name, int progfd, uint32_t flags);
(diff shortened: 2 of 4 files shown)


davemarchevsky:
Is my understanding correct here?

If `wakeup_events` is set to something like `10`,  epoll won't trigger on the fd until 10 events have been collected (ignoring `epoll_wait`'s `timeout` for now). When 10 events have been collected, [`perf_buffer__poll`](https://github.com/libbpf/libbpf/blob/master/src/libbpf.c#L11158), will end up calling the handler callback once for each sample. So the handler cb doesn't need to be updated to handle multiple events if `wakeup_events > 1`.

Do we need to do anything special to ensure the 'remainder' events are handled? e.g. `wakeup_events = 10`, 9 events are collected but 10th never happens.

yonghong-song:
> Do we need to do anything special to ensure the 'remainder' events are handled? e.g. wakeup_events = 10, 9 events are collected but 10th never happens.

does epoll timeout will serve this purpose? If the number 10 event never comes up, timeout eventually happens and the user space should process it, right?

chenhengqi:
The callchain is:
```c
bpf_perf_event_output
    __bpf_perf_event_output
        perf_event_output
            __perf_event_output
                perf_output_sample
                perf_output_end
                    perf_output_put_handle
```

In `perf_output_sample`, `rb->wakeup` is updated only when `events >= wakeup_events`:
```c
	if (!event->attr.watermark) {
		int wakeup_events = event->attr.wakeup_events;


		if (wakeup_events) {
			struct perf_buffer *rb = handle->rb;
			int events = local_inc_return(&rb->events);


			if (events >= wakeup_events) {
				local_sub(wakeup_events, &rb->events);
				local_inc(&rb->wakeup);
			}
		}
	}
```

In `perf_output_put_handle`, `perf_output_wakeup` is called after `rb->wakeup` changed:
```c
	if (handle->wakeup != local_read(&rb->wakeup))
		perf_output_wakeup(handle);
```

And here is `perf_output_wakeup`:
```c
static void perf_output_wakeup(struct perf_output_handle *handle)
{
	atomic_set(&handle->rb->poll, EPOLLIN);


	handle->event->pending_wakeup = 1;
	irq_work_queue(&handle->event->pending);
}
```

IIRC, the timeout won't take effect. And the users have to handle the `remainder` events themselves.

According to my test, each sample triggers the callback handler once, so we don't have to change the handler.

chenhengqi:
The callchain is:
```c
bpf_perf_event_output
    __bpf_perf_event_output
        perf_event_output
            __perf_event_output
                perf_output_sample
                perf_output_end
                    perf_output_put_handle
```

In `perf_output_sample`, `rb->wakeup` is updated only when `events >= wakeup_events`:
```c
	if (!event->attr.watermark) {
		int wakeup_events = event->attr.wakeup_events;


		if (wakeup_events) {
			struct perf_buffer *rb = handle->rb;
			int events = local_inc_return(&rb->events);


			if (events >= wakeup_events) {
				local_sub(wakeup_events, &rb->events);
				local_inc(&rb->wakeup);
			}
		}
	}
```

In `perf_output_put_handle`, `perf_output_wakeup` is called after `rb->wakeup` changed:
```c
	if (handle->wakeup != local_read(&rb->wakeup))
		perf_output_wakeup(handle);
```

And here is `perf_output_wakeup`:
```c
static void perf_output_wakeup(struct perf_output_handle *handle)
{
	atomic_set(&handle->rb->poll, EPOLLIN);


	handle->event->pending_wakeup = 1;
	irq_work_queue(&handle->event->pending);
}
```

IIRC, the timeout won't take effect. And the users have to handle the `remainder` events themselves.

According to my test, each sample triggers the callback handler once, so we don't have to change the handler.

davemarchevsky:
@yonghong-song 

> If the number 10 event never comes up, timeout eventually happens and the user space should process it, right?

I think we're on the same page. I didn't have a very clear idea of the scenarios I was worried about. After some more thought I think there's nothing to be worried about. For posterity, though:

* Programs that use perf_events and run for a specified amount of time, like `examples/PyPerf`, already do an explicit [poll of the perf buffer](https://github.com/iovisor/bcc/blob/master/examples/cpp/pyperf/PyPerfUtil.cc#L266-L281) after detaching the event to catch 'remainder' events. So `wakeup_events = 10` scenario is handled.
* Tools that collect data until killed by terminal input, like `libbpf-tools/execsnoop` don't do an explicit extra poll, but are also not attempting to collect a precise time period's worth of events, so it's not a problem if `remainder` events aren't handled. Ostensibly if someone is setting `wakeup_events > 1` in their tool they're tracing an event that occurs often. 

So no problem from my end.

@chenhengqi 

> IIRC, the timeout won't take effect. And the users have to handle the remainder events themselves.

When you say "timeout won't take effect", are you referring to the case where `>= wakeup_events` events are collected, or more generally? Just want to confirm that it's the former, because some programs (like `PyPerf` in first bullet above) do a poll w/ `0` timeout to grab remainder events.

chenhengqi:
@davemarchevsky  

Yes, you are right.

By `timeout won't take effect`, I mean that timeout does not affect the behavior of PerfBuffer itself.

davemarchevsky:
LGTM

yzhao1012:
Context: I filed #3793 

IIUC, the updated API needs to be exposed to `ebpf::BPFPerfBuffer` as well, for users of that API to be able to change the `wakeup_events`.

But one thing I think my understanding does not match the reality:
* It seems there is no change to the callback function signature?
* I assumed that the callback needs to be changed to handle multiple events. Today the callback looks like: `typedef void (*perf_reader_raw_cb)(void *cb_cookie, void *raw, int raw_size);` I'd imagine it turns to something like `typedef void (*perf_reader_raw_cb)(void *cb_cookie, void **raw, int *raw_sizes, int len);` which gives the number of events, and the size of each individual event.

So here I assumed that, by waking up every (for example) 10 events, the 10 events are handled by 1 invocation of the callback on all events in the perf ring buffer (perfbuffer, not the new bpf ringbuf).

But it seems that this actually only invoke the callback on the 10th event? If that's the case, then we cannot use it, because we do need to handle all events, missing any of them would not be allowed.

Instead, we only want to minimize the overhead of repetitively invoking callback on events. Thus #3793

davemarchevsky:
> But it seems that this actually only invoke the callback on the 10th event? 

In the `wakeup_events=10` case, the callback will be invoked on every event, but the perf buffer's fd will not indicate to `epoll` that it has data ready to read until it contains 10 events. When it does indicate that it's ready to be read, all of the data in the buffer will be read and handled.

I see what you mean with the "handle many events per callback instead of 1" point, but I don't think this improve performance significantly relative to exposing `wakeup_events`. Bumping `wakeup_events` from `1` to `10` would reduce context switches due to poll 10x which would dwarf the perf win from reducing callback function call overhead. 

Also, note that the samples in the perf buffer are not of fixed size and may loop around from the 'end' to 'beginning' of the ring buffer, so when pulling many samples out it's necessary to poke at each one to find size and possibly do some `memcpy`ing to deal with looped sample (see [`perf_event_read_simple` in libbpf](https://github.com/libbpf/libbpf/blob/master/src/libbpf.c#L10683)). You can probably create a modified `perf_event_read_simple` that creates a batch of events for a callback to deal with but I'm not sure if it would be generally useful.

> IIUC, the updated API needs to be exposed to ebpf::BPFPerfBuffer as well, for users of that API to be able to change the wakeup_events.

Yep. This should be added, as well as a test validating this `wakeup_events` stuff.

While messing around with this PR's changes today I noticed that we're missing a `perf_buffer__consume` equivalent, which is necessary to handle 'remainder' events I mentioned in previous comments, as well as some other opportunities to simplify, so thanks for bumping this!

PR #3805, opened 2022-01-11
Add perf_reader_consume

davemarchevsky:
In order to read 'remainder' events from perf buffers w/ 'wakeup_events 
\> 1', an API to force reading from a program's perf buffers regardless
of whether their fds are in "ready to read" state is necessary. This
commit introduces such an API, perf_reader_consume, modeled after
libbpf's perf_buffer__consume.

Future work will refactor bcc to use libbpf's perf buffer
support as much as possible instead of duplicating functionality. For
now, since #3801 was commited let's add this piece of missing
functionality.

---

Explanation: Some of my understanding of 'remainder' event handling from #3801 was incorrect:

> Programs that use perf_events and run for a specified amount of time, like examples/PyPerf, already do an explicit poll of the perf buffer after detaching the event to catch 'remainder' samples. So wakeup_events = 10 scenario is handled.

My belief that adding a `poll()` w/ `timeout=0` would result in remainder events being handled regardless of `wakeup_events` setting was incorrect. Such a poll will only guarantee handling of remainder samples if `wakeup_events = 1`, because the buffer's fd will be marked readable (`POLLIN`/`EPOLLIN`) anytime a sample is written to the buffer. To properly handle remainder events for all cases, it's necessary to add an API that attempts to read regardless of what the FD says.

Proof:

Consider the following patch:
```python
+++ b/tests/python/test_usdt.py
@@ -202,18 +202,19 @@ int do_trace5(struct pt_regs *ctx) {
 
         def print_event5(cpu, data, size):
             event = ct.cast(data, ct.POINTER(Data5)).contents
-            print("%s" % event.v1)
+            print("testtesttest%s" % event.v1)
```

This will result in test_usdt output like:
```
31: Ran 1 test in 4.067s                                                                                                                                                                                               31:                                                                                                                                                                                                                    31: OK                                                                                                                                                                                                                 31: b'str4'                                                                                                                                                                                                            
31: testtesttestb'str5'                                                                                                                                                                                                
31: b'str6'                                                                                                                                                                                                            
31: testtesttestb'str7'                                                                                                                                                                                                
31: b'str8'                                                                                                                                                                                                            
31: testtesttestb'str9'                                                                                                                                                                                                
31: Running from kernel directory at: /lib/modules/5.11.0-44-generic/build                                                                                                                                             
31/41 Test #31: py_test_usdt .....................   Passed    4.19 sec                                                                                                                                                
```

If the perf buffer that's using `print_event5` as a sample handler cb is modified like so:
```python
+++ b/tests/python/test_usdt.py
@@ -202,18 +202,19 @@ int do_trace5(struct pt_regs *ctx) {
 
         def print_event5(cpu, data, size):
             event = ct.cast(data, ct.POINTER(Data5)).contents
-            print("%s" % event.v1)
+            print("testtesttest%s" % event.v1)
 
         # loop with callback to print_event
         b["event1"].open_perf_buffer(print_event1)
         b["event2"].open_perf_buffer(print_event2)
         b["event3"].open_perf_buffer(print_event3)
         b["event4"].open_perf_buffer(print_event4)
-        b["event5"].open_perf_buffer(print_event5)
+        b["event5"].open_perf_buffer(print_event5, wakeup_events=2)
```

Output will look like:

```
31: Ran 1 test in 4.084s                                                                                                                                                                                               31:                                                                                                                                                                                                                    31: OK                                                                                                                                                                                                                 
31: b'str4'                                                                                                                                                                                                            
31: b'str6'                                                                                                
31: testtesttestb'str5'                                                                                                                                                                                                
31: testtesttestb'str7'                                                                                                                                                                                                
31: b'str8'                                                                                                                                                                                                            
31: Running from kernel directory at: /lib/modules/5.11.0-44-generic/build                                                                                                                                             
31/41 Test #31: py_test_usdt .....................   Passed    4.28 sec                  
```

Note that `print_event5` is now called twice in succession after the `event5` perf buffer has 2 samples. But `str9` is now never printed because the buffer never again contains 2 samples.

Doing something like:

```python
+++ b/tests/python/test_usdt.py
@@ -202,18 +202,19 @@ int do_trace5(struct pt_regs *ctx) {
 
         def print_event5(cpu, data, size):
             event = ct.cast(data, ct.POINTER(Data5)).contents
-            print("%s" % event.v1)
+            print("testtesttest%s" % event.v1)
 
         # loop with callback to print_event
         b["event1"].open_perf_buffer(print_event1)
         b["event2"].open_perf_buffer(print_event2)
         b["event3"].open_perf_buffer(print_event3)
         b["event4"].open_perf_buffer(print_event4)
-        b["event5"].open_perf_buffer(print_event5)
+        b["event5"].open_perf_buffer(print_event5, wakeup_events=2)
 
         # three iterations to make sure we get some probes and have time to process them
         for i in range(3):
             b.perf_buffer_poll()
+        b.perf_buffer_poll(timeout=0)
         self.assertTrue(self.evt_st_1 == 1 and self.evt_st_2 == 1 and self.evt_st_3 == 1)
```

Does not fix the problem because the fd is not marked readable, so timing out early doesn't help. Replacing the new `perf_buffer_poll` line w/ a`perf_buffer_consume()` call fixes the issue, resulting in output like:
```
31: Ran 1 test in 4.069s
31: 
31: OK
31: b'str4'
31: b'str6'
31: testtesttestb'str5'
31: testtesttestb'str7'
31: b'str8'
31: testtesttestb'str9'
31: Running from kernel directory at: /lib/modules/5.11.0-44-generic/build
31/41 Test #31: py_test_usdt .....................   Passed    4.27 sec
```

I will add a separate test instead of modifying `test_usdt` because it's testing entirely different things and doesn't care about ordering. 

/cc @chenhengqi


yonghong-song:
LGTM. Thanks @davemarchevsky for very detailed explanation.

PR #3806, opened 2022-01-11
Add rwengine=OFF test run, disable tests we expect to fail during it

davemarchevsky:
I noticed this the other day. Just pushing this PR up to get signal from GH actions. On my local box, running the tests with this commit results in these failures:

```
The following tests FAILED:                                                                                                                                                                  
          3 - test_libbcc (Failed)                                                            
         16 - py_test_clang (Failed)                                                          
```

### DIFF
--- a/.github/workflows/bcc-test.yml
+++ b/.github/workflows/bcc-test.yml
@@ -15,8 +15,13 @@ jobs:
         env:
         - TYPE: Debug
           PYTHON_TEST_LOGFILE: critical.log
+          RW_ENGINE_ENABLED: ON
+        - TYPE: Debug
+          PYTHON_TEST_LOGFILE: critical.log
+          RW_ENGINE_ENABLED: OFF
         - TYPE: Release
           PYTHON_TEST_LOGFILE: critical.log
+          RW_ENGINE_ENABLED: ON
     steps:
     - uses: actions/checkout@v2
     - name: System info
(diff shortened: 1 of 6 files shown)


davemarchevsky:
This PR now contains two commits:

```
Disable tests that will fail when rwengine=off
When ENABLE_LLVM_NATIVECODEGEN is OFF, the bpf module rw_engine code
will not be used. This is the code that generates custom sscanf and
printf functions for BPF map defs. These generated functions are used
to serialize arguments to update_value (and similar calls which take
string as input) to the correct key / value types.

It's still possible to use update_value and its ilk without rw_engine
enabled, just need to use the variants that take raw bytes and do
serialization manually. In the future we can add some simple
sscanf/printfs for common types that can be used without LLVM
generation.

For now, when rw_engine is disabled, it's fine to skip tests that
require these autogenerated functions.
```

```
Add RW_ENGINE_ENABLED=OFF test run for ubuntu Debug build
```

@yonghong-song I'm going to merge this and update the "required statuses", as this changes the name of some tests. I'm getting ready to push up a rebased version of #3660 and these failing tests were confusing me as I thought they were related to that PR's changes.

davemarchevsky:
tests failing after this was merged, but i don't think it's related:

```
Download action repository 'actions/checkout@v2' (SHA:ec3a7ce113134d7a93b817d10a8272cb61118579)
Warning: Failed to download action 'https://api.github.com/repos/actions/checkout/tarball/ec3a7ce113134d7a93b817d10a8272cb61118579'. Error: Response status code does not indicate success: 503 (Service Unavailable).
Warning: Back off 11.036 seconds before retry.
Warning: Failed to download action 'https://api.github.com/repos/actions/checkout/tarball/ec3a7ce113134d7a93b817d10a8272cb61118579'. Error: Response status code does not indicate success: 503 (Service Unavailable).
Warning: Back off 11.264 seconds before retry.
Error: Response status code does not indicate success: 503 (Service Unavailable).
```

Probably a transient issue on GH's end.

PR #3817, opened 2022-01-16
bcc: Update perf buffer C++ API

chenhengqi:
Like in #3801 and #3805, this patch adds a consume API
which allows user to force reading from perf buffer and
adds wake_events parameter on open_all_cpu API.

### DIFF
--- a/src/cc/api/BPFTable.cc
+++ b/src/cc/api/BPFTable.cc
@@ -397,21 +397,21 @@ BPFPerfBuffer::BPFPerfBuffer(const TableDesc& desc)
                                 "' is not a perf buffer");
 }
 
-StatusTuple BPFPerfBuffer::open_on_cpu(perf_reader_raw_cb cb,
-                                       perf_reader_lost_cb lost_cb, int cpu,
-                                       void* cb_cookie, int page_cnt) {
-  if (cpu_readers_.find(cpu) != cpu_readers_.end())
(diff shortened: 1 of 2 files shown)


src/cc/api/BPFTable.cc
```diff
@@ -397,21 +397,21 @@ BPFPerfBuffer::BPFPerfBuffer(const TableDesc& desc)
...
 }
 
-StatusTuple BPFPerfBuffer::open_on_cpu(perf_reader_raw_cb cb,
-                                       perf_reader_lost_cb lost_cb, int cpu,
-                                       void* cb_cookie, int page_cnt) {
-  if (cpu_readers_.find(cpu) != cpu_readers_.end())
-    return StatusTuple(-1, "Perf buffer already open on CPU %d", cpu);
+StatusTuple BPFPerfBuffer::open_on_cpu(perf_reader_raw_cb cb, perf_reader_lost_cb lost_cb,
+                                       void* cb_cookie, int page_cnt,
+                                       struct bcc_perf_buffer_opts *opts) {
```
davemarchevsky, line 402: can you add a null pointer check for `opts` here, or make it a reference?

davemarchevsky (changes requested):

src/cc/api/BPFTable.cc
chenhengqi: OK, make it a reference.

davemarchevsky (approved):

PR #3818, opened 2022-01-17
Walk user/kernel stack in trace_entry rather than trace_return

xingfeng2510:
Walk user/kernel stack in trace_entry rather than trace_return, show the specified function on top of a stack.

### before (kretprobe_trampoline on top of stack)
```
# ./funcslower.py -K vfs_write
...
    b'kretprobe_trampoline'
    b'__x64_sys_write'
    b'do_syscall_64'
    b'entry_SYSCALL_64_after_hwframe'
```

### after (vfs_write on top of stack)
```
# ./funcslower.py -K vfs_write
...
    b'vfs_write'
    b'__x64_sys_write'
    b'do_syscall_64'
    b'entry_SYSCALL_64_after_hwframe'
```

### DIFF
--- a/tools/funcslower.py
+++ b/tools/funcslower.py
@@ -85,6 +85,13 @@ struct entry_t {
 #ifdef GRAB_ARGS
     u64 args[6];
 #endif
+#ifdef USER_STACKS
+    int user_stack_id;
+#endif
+#ifdef KERNEL_STACKS
+    int kernel_stack_id;
+    u64 kernel_ip;
+#endif
 };
 
 struct data_t {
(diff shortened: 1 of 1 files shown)


yonghong-song:
LGTM. The kretprobe stack issue is fixed by the patch set:
   https://lore.kernel.org/all/163163030719.489837.2236069935502195491.stgit@devnote2/
which is available at 5.15.

Luckily for funcslower.py, we have both kprobe and kretprobe, so it is okay to collect stack trace with
kprobe to workaround the bug.

PR #3820, opened 2022-01-19
bcc/tools: fix -u <int> flag for execsnoop. UID of 0 always fails.

sanderfoobar:
https://github.com/iovisor/bcc/blob/b1d87968aceed0a905f1ce32ac8c8d010a07d5e8/tools/execsnoop.py#L216-L220

This clause fails when `args.uid` is `0`. [parse_uid()](https://github.com/iovisor/bcc/blob/b1d87968aceed0a905f1ce32ac8c8d010a07d5e8/tools/execsnoop.py#L33) returns int.

### DIFF
--- a/tools/execsnoop.py
+++ b/tools/execsnoop.py
@@ -213,7 +213,7 @@ int do_ret_sys_execve(struct pt_regs *ctx)
 
 bpf_text = bpf_text.replace("MAXARG", args.max_args)
 
-if args.uid:
+if isinstance(args.uid, int):
     bpf_text = bpf_text.replace('UID_FILTER',
         'if (uid != %s) { return 0; }' % args.uid)
 else:


tools/execsnoop.py
```diff
@@ -213,7 +213,7 @@ def parse_uid(user):
 
 bpf_text = bpf_text.replace("MAXARG", args.max_args)
 
-if args.uid:
+if isinstance(args.uid, int):
```
chenhengqi, line 216: You can use `args.uid is not None`.

sanderfoobar: `isinstance()` is needed, anything other than `int` will result in failure to compile the BPF module. 

https://github.com/iovisor/bcc/blob/b1d87968aceed0a905f1ce32ac8c8d010a07d5e8/tools/execsnoop.py#L216-L218

chenhengqi: I don't think so.

https://github.com/iovisor/bcc/blob/b1d87968aceed0a905f1ce32ac8c8d010a07d5e8/tools/execsnoop.py#L33-L46

sanderfoobar: Yes, I noticed `parse_uid` may return `[str, int]`, which is wrong. Noe that the BPF filter only accepts an int; `if (uid != %s)` hence the `int` enforcement.

chenhengqi: How do you repro this error ?

sanderfoobar: 1. the if branch only supports numbers
2. thus we enforce numbers

Any other implementation would allow BPF module to fail. Your suggestion `args.uid is not None` implies "this if branch can work with anything other than `None`. This is unrelated to what `parse_uid()` is doing. Python is a dynamic language, another programmer in a few weeks can mistakenly modify `args.uid` such that it is not an `int` in the impression that the BPF would still work because of how the if clause is constructed.

Anyway, who cares. Merge or PR something different.

sanderfoobar: yes it also wrong, since we care about the type, like I tried to explain.

I guess `execsnoop` will be broken for `-u 0`. Too bad.

chenhengqi: I don't see the point here. 

`args.uid is not None` should work, no ?

PR #3822, opened 2022-01-24
perf_reader: perf_reader_event_read will consume more than a one ring buffer worth of data

JamesMBartlett:
I noticed that in a single call to `perf_reader_event_read`, the perf_reader can end up reading more than a full ring buffer's worth of data. As long as the consumer side (perf_reader_raw_cb) is slower than the producer side (ebpf events from the kernel), then the ring buffer will fill up and the ring buffer's head position will always be behind its tail. Since the consumer is slower than the producer, as soon as the perf_reader updates the ring buffer tail position, the kernel will quickly write a new event and increment head. Because the perf_reader_event_read loop, reloads the ring buffer's head position on each iteration, its possible for this function to loop forever. This change just makes it so that the head position is read once at the start of the loop, and then the loop reads all the data up until this initial head position. The downside of this is that if data is written by the kernel during the `perf_reader_event_read` call it won't be read until the next call to `perf_reader_event_read`, but this seems acceptable to me compared to unbounded runtime of `perf_reader_event_read`.

I also added a test that demonstrates the behaviour above. The test has a thread that constantly writes to `/dev/null`, and then on the main thread it deploys a ebpf program that pushes 32kB of data to the ring buffer on every single write call. It also has a sleep in the callback function for the probe, that forces the consumer to be much slower than the producer. Then in a separate thread it calls `perf_buffer->poll(0)` and records how much data is received by the callback. Without the changes in this diff, this test would show that more than a full ring buffer's worth of data (times the number of perf_readers' the poll call touched) was received by the callback. I wasn't sure what kernel version perf buffer's are supported from, but based on `test_perf_event.cc` I assumed that 4.13 should be good. (Happy to remove the test from this diff if its not something you want, but thought it was useful to demonstrate the behaviour I'm trying to fix)

Looked for a style guide, but couldn't find one, so let me know if there are style issues.

cc. @oazizi000


yonghong-song:
@JamesMBartlett I think we probably cannot make this change, bailing out once it wrapped to the 'head', as the default. This may cause losing more events and possible performance issue for the application. cc @davemarchevsky

JamesMBartlett:
Hi @yonghong-song, thanks for looking at my diff.

My understanding of the situation is there are two cases:

In one case the producer is faster than the consumer. In this case, no matter what you do you are going to be losing events. With my change you will lose events and return control flow to the caller, but without my changes you will also lose events (because head will catch up to tail and the kernel won't push more events into the buffer) but `poll()` could also potentially loop forever (depending on how much slower the consumer is) and never return control flow to the calling application.

In the other case, what you would expect in "normal" operation, the consumer is faster than the producer. In this case, you would expect not to lose any events. This case should not be affected by my changes. Since each call to `poll()` returns an arbitrary amount of data from the calling applications perspective, all calling applications have to have logic to continually re-call `poll()`. So it shouldn't matter whether you get the event in the first, or second, etc call to `poll()` since the consumer is faster than the producer, you won't lose any of the events. 

I could be missing something here, so let me know if so, but based on this understanding I don't see where my change would cause more lost events than existing code.

To ground this in our use case, I created this change because I noticed in our application that a single call to `poll()` was leading to an infinite loop.

yonghong-song:
The current implementation might be good for the 'burst' case (producer sometimes faster than consumer, but not last long). Not sure how perf handles this. cc @liu-song-6 do you know?

liu-song-6:
IIUC, perf-top will repeat reading until the ring buffer is empty. I guess it is similar to current implementation in bcc. 
This was a bigger issue for perf-record. `perf record --aio` mode was introduced to consume data from ring buffer faster.

yzhao1012:
> IIUC, perf-top will repeat reading until the ring buffer is empty. I guess it is similar to current implementation in bcc.
IIUC, this is to say that perf-record's own polling works just like bcc. I think that means perf-top is not affected by this PR, right?

> This was a bigger issue for perf-record. `perf record --aio` mode was introduced to consume data from ring buffer faster.
How exactly `perf record --aio` works?

liu-song-6:
> > IIUC, perf-top will repeat reading until the ring buffer is empty. I guess it is similar to current implementation in bcc.
> 
> IIUC, this is to say that perf-record's own polling works just like bcc. I think that means perf-top is not affected by this PR, right?

I don't think perf-top is affected by this PR, as perf doesn't use bcc. 

> 
> > This was a bigger issue for perf-record. `perf record --aio` mode was introduced to consume data from ring buffer faster.
> 
> How exactly `perf record --aio` works?
It uses aio_write from aio.h.

yonghong-song:
I checked perf builtin-trace.c and builtin-top.c, they both tried to read until no data is available. In builtin-trace.c perf read, it handles interrupt (^C) to stop the reading and I think this is consistent with 'perf record' is doing.

I think we probably should preserve the existing default behavior. Typically if producer is faster, people will adjust sampling rate to make it slow not to overrun the consumer. So we really have typical case where producer is not faster than consumer. The default behavior here handles such a case well and can handle occasional burst.

To handle producer faster than consumer case for possible infinite loop case, I think we can introduce a flag for this. Any comments?

PR #3744, opened 2021-12-10
libbpf-tools: Fix bio tools with 5.16 kernel

chenhengqi:
The tools biosnoop and biostacks are broken due to kernel change ([0]).
blk_account_io_{start, done} were renamed to __blk_account_io_{start, done},
and the symbols gone from vmlinux BTF. Fix them by checking symbol existence.

  [0]: https://github.com/torvalds/linux/commit/be6bfe36db1795babe9d92178a47b2e02193cb0f

Signed-off-by: Hengqi Chen <chenhengqi@outlook.com>

PR #3827, opened 2022-01-27
biolatpcts: Build fixes on recent kernels

htejun:
* `struct request` definition recently moved from blkdev.h to blk-mq.h
  breaking both tools/biolatpcts and examples/tracing/biolatpcts. Fix them
  by also including blk-mq.h.

* blk_account_io_done() got split into two parts - inline condition checks
  and the actual accounting with the latter now done in
  __blk_account_io_done(). The kprobe attachment needs to be conditionalized
  to work across the change. tools/biolatpcts was already updated but
  examples/tracing/biolatpcts wasn't. Fix it.

Signed-off-by: Tejun Heo <tj@kernel.org>

### DIFF
--- a/examples/tracing/biolatpcts.py
+++ b/examples/tracing/biolatpcts.py
@@ -11,6 +11,7 @@ from time import sleep
 
 bpf_source = """
 #include <linux/blk_types.h>
+#include <linux/blk-mq.h>
 #include <linux/blkdev.h>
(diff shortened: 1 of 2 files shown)


davemarchevsky (approved):

chenhengqi:
@davemarchevsky  Please also take a look at #3744 .

PR #3837, opened 2022-02-08
extern BPF_TABLE_PINNED to include an optional flags argument

yonghong-song:
BPF_TABLE_PINNED has a feature to create a map and pin it if
the pinned map doesn't exist. Some maps, e.g., bpf_sk_storage map,
requires a non-zero flag for map creation. This patch extended
the BPF_TABLE_PINNED map to include an optional flags argument
for any pinned map which my require a non-zero flags.

### DIFF
--- a/src/cc/export/helpers.h
+++ b/src/cc/export/helpers.h
@@ -170,8 +170,17 @@ struct _name##_table_t __##_name
 #define BPF_TABLE(_table_type, _key_type, _leaf_type, _name, _max_entries) \
 BPF_F_TABLE(_table_type, _key_type, _leaf_type, _name, _max_entries, 0)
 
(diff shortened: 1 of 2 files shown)


yonghong-song:
Failure stack:
  File "/bcc/tests/python/test_tools_smoke.py", line 400, in test_uthreads
    self.run_with_int("lib/uthreads.py %d" % os.getpid())
  File "/bcc/tests/python/test_tools_smoke.py", line 59, in run_with_int
    self.assertTrue((rc == 0 and allow_early) or rc == 124
AssertionError: False is not true : rc was 1
Command was expected to do one of:
	Be killed by SIGINT

The failure is on fedora only and ubuntu works fine. It seems being related to usdt and will need further debug.

PR #3838, opened 2022-02-08
Use page offsets to distinguish between locations

iii-i:
When attaching to a probe that is inlined multiple times without
specifying a pid (which is not always convenient, e.g. when a target
process is not running yet), the following exception occurs:

    Exception: can't generate USDT probe arguments; possible cause is missing pid when a probe in a shared object has multiple locations

This is because eBPF programs distinguish between different locations
by doing a switch on an instruction pointer, and in order to get full
probe addresses in presence of ASLR, one has to choose a specific
process and inspect its /proc/{pid}/maps.

However, since shared objects are page-aligned, when all of the probe
locations have different page offsets, these offsets uniquely identify
locations. Adjust Probe::usdt_getarg() to generate a switch on page
offset if this is the case.

If some locations have the same page offset, an instruction pointer
switch is used, so this patch does not fully lift the restriction.
However, it helps greatly with one-off investigations.

Signed-off-by: Ilya Leoshkevich <iii@linux.ibm.com>

### DIFF
--- a/src/cc/usdt/usdt.cc
+++ b/src/cc/usdt/usdt.cc
@@ -175,6 +175,11 @@ bool Probe::usdt_getarg(std::ostream &stream, const std::string& probe_func) {
   if (arg_count == 0)
     return true;
 
+  uint64_t page_size = sysconf(_SC_PAGESIZE);
+  std::unordered_set<int> page_offsets;
+  for (Location &location : locations_)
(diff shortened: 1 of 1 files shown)


yonghong-song:
LGTM. I think this should cover majority of cases where the same probe is in multiple locations without pid.

PR #3840, opened 2022-02-09
tools/sslsniff: trace extra libraries

bacher09:
This adds `--extra-lib` flag, which might be quite useful if an application is using custom TLS libraries that is shipped with application or in case of containerized applications. Also want to mention that bcc can determine path to the library [automatically by inspecting](https://github.com/iovisor/bcc/blob/e83019bdf6c400b589e69c7d18092e38088f89a8/src/cc/bcc_proc.c#L419) `/proc/$pid/maps` (if pid has been specified), but this not always works. For example, if application hasn't been started yet (and you need to trace launching process), if you want to trace multiple processes at the same time, or if application hasn't loaded that library at that time.

This flag can specified multiple times (which allows to provide list of libraries), for example:

```
./sslsniff --extra-lib openssl:/path/to/libssl.so --extra-lib nss:/path/to/libnspr4.so --extra-lib gnutils:/path/to/libgnutls.so
```
Also tracing of system wide libraries could be disabled (using flags that already implemented in `./sslsniff`), so it would be possible to just trace calls for specified libraries.

### DIFF
--- a/tools/sslsniff_example.txt
+++ b/tools/sslsniff_example.txt
@@ -160,12 +160,18 @@ READ/RECV    0.000049464        nginx            7081    1      0.004
 
 ----- END DATA -----
 
+Tracing few extra libraries (useful for docker containers and other isolated
+apps)
+
+# ./sslsniff.py --extra-lib openssl:/var/lib/docker/overlay2/l/S4EMHE/lib/libssl.so.1.1
+
+
 
 USAGE message:
 
 usage: sslsniff.py [-h] [-p PID] [-u UID] [-x] [-c COMM] [-o] [-g] [-n] [-d]
                    [--hexdump] [--max-buffer-size MAX_BUFFER_SIZE] [-l]
-                   [--handshake]
+                   [--handshake] [--extra-lib EXTRA_LIB]
 
 Sniff SSL data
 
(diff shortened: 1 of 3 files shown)


bacher09:
Also wanted to say regarding test [that fails](https://github.com/iovisor/bcc/blob/master/tests/python/test_tools_smoke.py#L313) on `sslsniff.py`, it fails because github actions environment have no `nss` library, running it with `--no-nss` should fix the issue (although it would skip testing some functionality, which might be bad).

tools/sslsniff_example.txt
```diff
@@ -160,12 +160,18 @@ READ/RECV    0.000049464        nginx            7081    1      0.004
 
 ----- END DATA -----
 
+Tracing few extra libraries (useful for docker containers and other isolated
+apps)
+
+# ./sslsniff.py --extra-lib /var/lib/docker/overlay2/l/S4EMHE/lib/libssl.so.1.1
```
yonghong-song, line 166: openssl:/var/lib/docker/overlay2/l/S4EMHE/lib/libssl.so.1.1?

yonghong-song (commented):
LGTM. Only one minor comment.

tools/sslsniff_example.txt
bacher09: Yeah you right, don't even know how I missed this :shrug: .

PR #3844, opened 2022-02-11
exitsnoop --max-args MAX_ARGS

nenadnoveljic:
Added the option to show program arguments if the option --max-args is specified.

### DIFF
--- a/tools/exitsnoop.py
+++ b/tools/exitsnoop.py
@@ -1,4 +1,4 @@
-#/usr/bin/python
+#!/usr/bin/python
 # @lint-avoid-python-3-compatibility-imports
 from __future__ import print_function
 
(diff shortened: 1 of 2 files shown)


chenhengqi:
Why this change ? Why not use execsnoop to do that ?

nenadnoveljic:
You're right - you can correlate the execsnoop and exitsnoop outputs over PID. However, I've had a case in which was  more convenient to get the arguments in the exitsnoop output. I needed to capture the parameters of a failed execution which happens very rarely - once every few days - and the number of good executions was very high. So instead of running execsnoop over days and taking care of the log rotation and retention, I was running exitsnoop with -x (only failed executions) and captured the arguments.

yonghong-song:
@brendangregg Any comments for this patch? Looks like this patch solves a real production problem. But I feel in exitsnoop tracing exec syscall seems weird. But on the other hand, we still trace 'exit' here. The extra 'exec' information is just used to provide more information when certain 'exit' event happens.

nenadnoveljic:
exec is traced only if the new parameter --mag-args is specified. Default is 0.

PR #3858, opened 2022-02-18
libbpf-tools: Use __u16 for bindsnoop protocol.

eiffel-fl:
Hi.

I was playing with CO-RE version of `bindsnoop` and I figured out the protocol is stored in an `__u8`.
Nothing really serious but sadly this can not suffice to store some protocol (like `MPTCP`).
So, I preferred to open a PR suggesting this change or at least make people aware of it.

Best regards.

### DIFF
--- a/libbpf-tools/bindsnoop.h
+++ b/libbpf-tools/bindsnoop.h
@@ -11,8 +11,8 @@ struct bind_event {
 	__u32 bound_dev_if;
 	int ret;
 	__u16 port;
+	__u16 proto;
 	__u8 opts;
-	__u8 proto;
 	__u8 ver;
 	char task[TASK_COMM_LEN];
 };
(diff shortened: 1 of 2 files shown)


libbpf-tools/bindsnoop.h
```diff
@@ -12,7 +12,7 @@ struct bind_event {
 	int ret;
 	__u16 port;
 	__u8 opts;
-	__u8 proto;
+	__u16 proto;
```
chenhengqi, line 15: Please reorder this field to avoid padding.

eiffel-fl: Normally it is done, but I got 64 bytes for this struct with and without reordering of this field.
I also removed the uneeded include of `maps.bpf.h`.

chenhengqi (approved):
LGTM. Thanks.

davemarchevsky (approved):

eiffel-fl:
Thank you for the merge!

PR #3864, opened 2022-02-19
Decode symname for exception message when symbol cannot be found

viniciusd:
symname is kept as a byte-sring. As such, when the exception message is created, it generates something like:
```
could not determine address of symbol b'function__entry'
```
By decoding the symname variable, the user will get the following message instead:
```
could not determine address of symbol function__entry
```

I took it one step further and added the module to the error message. An example of the new message:
```
could not determine address of symbol function__entry in /usr/bin/python
```

This will make it more clear for users as well as hide the implementation detail of having byte-strings, which shouldn't be exposed to users who simply rely on the tool instead of coding it themselves.

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -957,7 +957,8 @@ class BPF(object):
             ct.cast(None, ct.POINTER(bcc_symbol_option)),
             ct.byref(sym),
         ) < 0:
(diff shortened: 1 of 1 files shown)


davemarchevsky:
LGTM. Can you squash the commits down to one?

viniciusd:
> LGTM. Can you squash the commits down to one?

Squashed 👍

PR #3868, opened 2022-02-21
libbpf-tools: remove redundant slash

wangyuntao:
Since the dir function returns a path which ends with a trailing slash,
there is no need to add it again.

Signed-off-by: Yuntao Wang <ytcoode@gmail.com>


davemarchevsky (approved):
test failure for `test_bcc_fedora` looks unrelated. Coincidentally `test_bcc_fedora`'s `Run libbpf-tools build` step adequately tests this change

LGTM

Issue #3869, opened 2022-02-21
bio tools fail to compile, incomplete definition of type 'struct request'


cmurf:
Stack track from `biosnoop` `biolatency` and `biotop` are the same on both Fedora 35 and 36 (pre-release), with bcc-tools-0.21.0-1.fc35.x86_64 and bcc-tools-0.22.0-3.fc36.x86_64. Nothing found in dmesg.

```
$ sudo /usr/share/bcc/tools/biosnoop
/virtual/main.c:54:24: error: incomplete definition of type 'struct request'
        .data_len = req->__data_len
                    ~~~^
include/linux/blkdev.h:24:8: note: forward declaration of 'struct request'
struct request;
       ^
/virtual/main.c:89:26: error: incomplete definition of type 'struct request'
        data.sector = req->__sector;
                      ~~~^
include/linux/blkdev.h:24:8: note: forward declaration of 'struct request'
struct request;
       ^
/virtual/main.c:91:38: error: incomplete definition of type 'struct request'
        struct gendisk *rq_disk = req->rq_disk;
                                  ~~~^
include/linux/blkdev.h:24:8: note: forward declaration of 'struct request'
struct request;
       ^
/virtual/main.c:108:26: error: incomplete definition of type 'struct request'
    data.rwflag = !!((req->cmd_flags & REQ_OP_MASK) == REQ_OP_WRITE);
                      ~~~^
include/linux/blkdev.h:24:8: note: forward declaration of 'struct request'
struct request;
       ^
4 errors generated.
Traceback (most recent call last):
  File "/usr/share/bcc/tools/biosnoop", line 165, in <module>
    b = BPF(text=bpf_text)
  File "/usr/lib/python3.10/site-packages/bcc/__init__.py", line 452, in __init__
    raise Exception("Failed to compile BPF module %s" % (src_file or "<text>"))
Exception: Failed to compile BPF module <text>
$ sudo /usr/share/bcc/tools/biolatency
cannot attach kprobe, probe entry may not exist
Traceback (most recent call last):
  File "/usr/share/bcc/tools/biolatency", line 176, in <module>
    b.attach_kprobe(event="blk_account_io_done",
  File "/usr/lib/python3.10/site-packages/bcc/__init__.py", line 803, in attach_kprobe
    raise Exception("Failed to attach BPF program %s to kprobe %s" %
Exception: Failed to attach BPF program b'trace_req_done' to kprobe b'blk_account_io_done'
[...]

Issue #3825, opened 2022-01-27
tutorial bcc python: `disksnoop.py` not working


irevoire:
Hello,

I'm following the [tutorial bcc python](https://github.com/iovisor/bcc/blob/master/docs/tutorial_bcc_python_developer.md#lesson-6-disksnooppy) and when I try to execute the disksnooppy example I get this error:
```
/virtual/main.c:20:37: error: incomplete definition of type 'struct request'
                bpf_trace_printk("%d %x %d\n", req->__data_len,
                                               ~~~^
include/linux/blkdev.h:24:8: note: forward declaration of 'struct request'
struct request;
       ^
/virtual/main.c:21:10: error: incomplete definition of type 'struct request'
                    req->cmd_flags, delta / 1000);
                    ~~~^
include/linux/blkdev.h:24:8: note: forward declaration of 'struct request'
struct request;
       ^
2 errors generated.
Traceback (most recent call last):
  File "/root/bcc/./examples/tracing/disksnoop.py", line 20, in <module>
    b = BPF(text="""
  File "/usr/lib/python3.10/site-packages/bcc/__init__.py", line 479, in __init__
    raise Exception("Failed to compile BPF module %s" % (src_file or "<text>"))
Exception: Failed to compile BPF module <text>
```

And the thing is, the only occurrence of `blkdev.h` on my computer is found in `/tmp/kheaders-5.16.1-arch1-1/include/linux/blkdev.h`.
And it effectively starts with a sequence of empty structs:
```C
# head -n 35 /tmp/kheaders-5.16.1-arch1-1/include/linux/blkdev.h
/* SPDX-License-Identifier: GPL-2.0 */
#ifndef _LINUX_BLKDEV_H
#define _LINUX_BLKDEV_H

#include <linux/sched.h>
#include <linux/genhd.h>
#include <linux/list.h>
#include <linux/llist.h>
#include <linux/minmax.h>
#include <linux/timer.h>
#include <linux/workqueue.h>
#include <linux/wait.h>
#include <linux/bio.h>
#include <linux/gfp.h>
#include <linux/rcupdate.h>
#include <linux/percpu-refcount.h>
#include <linux/blkzoned.h>
#include <linux/sbitmap.h>

struct module;
struct request_queue;
struct elevator_queue;
struct blk_trace;
struct request;
struct sg_io_hdr;
struct blkcg_gq;
struct blk_flush_queue;
struct kiocb;
struct pr_ops;
struct rq_qos;
[...]

PR #3877, opened 2022-02-23
tools: include blk-mq.h in bio tools

jeromemarchand:
Kernel commit 24b83deb29b ("block: move struct request to blk-mq.h")
has moved struct request  from blkdev.h to blk-mq.h. It results in
several bio tools to fail with errors of the following type:

error: incomplete definition of type 'struct request'

Since blk-mq.h had always included blkdev.h. it is safe to simply
replace the inclusion of blkdev.h by blk-mq-h. It works on both older
and newer kernel.

Fixes: #3869

Signed-off-by: Jerome Marchand <jmarchan@redhat.com>

### DIFF
--- a/tools/biolatency.py
+++ b/tools/biolatency.py
@@ -64,7 +64,7 @@ if args.flags and args.disks:
 # define BPF program
 bpf_text = """
 #include <uapi/linux/ptrace.h>
-#include <linux/blkdev.h>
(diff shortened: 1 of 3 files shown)


chenhengqi (approved):
LGTM. This should also close #3825.

yonghong-song:
LGTM. Thanks for simpler solution!

PR #3880, opened 2022-02-24
usdt: support xmm registers as args for x64

davemarchevsky:
Support for using xmm registers for USDT args was added to systemtap in
early 2021 (commit 04c99d0d0267f574fa60044c96933b0dd3846aa1 added xmm0-7
and eaa15b047688175a94e3ae796529785a3a0af208 added xmm8-15). As a result
these registers are showing up in probe descriptors on Fedora.

pthread_start probe description on Ubuntu 20.04:
```
  stapsdt              0x00000052       NT_STAPSDT (SystemTap probe descriptors)
    Provider: libpthread
    Name: pthread_start
    Location: 0x00000000000095e9, Base: 0x000000000001922c, Semaphore: 0x0000000000000000
    Arguments: 8@%rax 8@1600(%rax) 8@1608(%rax)
```

And in the Fedora test container:

```
  stapsdt              0x00000053       NT_STAPSDT (SystemTap probe descriptors)
    Provider: libpthread
    Name: pthread_start
    Location: 0x0000000000009280, Base: 0x0000000000016cbc, Semaphore: 0x0000000000000000
    Arguments: 8@%xmm0 8@1600(%rax) 8@1608(%rax)
```

bcc doesn't recognize these registers so it's unable to generate code to fetch arguments which use these registers for storage. Add support for XMM0-15.

This should fix `lib/uthreads.py` test, which uses `pthread_start` and therefore fails to generate program on Fedora.

### DIFF
--- a/src/cc/usdt.h
+++ b/src/cc/usdt.h
@@ -166,6 +166,22 @@ private:
     X64_REG_14,
     X64_REG_15,
     X64_REG_RIP,
+    X64_REG_XMM0,
+    X64_REG_XMM1,
+    X64_REG_XMM2,
+    X64_REG_XMM3,
+    X64_REG_XMM4,
+    X64_REG_XMM5,
+    X64_REG_XMM6,
+    X64_REG_XMM7,
+    X64_REG_XMM8,
+    X64_REG_XMM9,
+    X64_REG_XMM10,
+    X64_REG_XMM11,
+    X64_REG_XMM12,
+    X64_REG_XMM13,
+    X64_REG_XMM14,
+    X64_REG_XMM15,
   };
 
   struct RegInfo {
--- a/src/cc/usdt/usdt_args.cc
+++ b/src/cc/usdt/usdt_args.cc
@@ -69,7 +69,13 @@ bool Argument::assign_to_local(std::ostream &stream,
   }
 
   if (!deref_offset_) {
-    tfm::format(stream, "%s = ctx->%s;", local_name, *base_register_name_);
+    if(base_register_name_->substr(0,3) == "xmm") {
+      // TODO: When we can read xmm registers from BPF, update this to read
+      // the actual value
+      tfm::format(stream, "%s = 0;", local_name);
+    } else {
+      tfm::format(stream, "%s = ctx->%s;", local_name, *base_register_name_);
+    }
     // Put a compiler barrier to prevent optimization
     // like llvm SimplifyCFG SinkThenElseCodeToEnd
     // Volatile marking is not sufficient to prevent such optimization.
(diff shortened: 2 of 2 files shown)


davemarchevsky:
This is currently just a preliminary fix and isn't in a mergeable state. TODO:

- [ ] Add test which actually uses the xmm arg to confirm correct value is fetched
- [ ] Confirm that pointer usecase works w/ xmm regs
- [ ] Consider a different approach for `base_register_name_->substr(0,3) == "xmm"` check: `Argument` class is currently platform-agnostic, so maybe this platform-specific custom logic can be added in a more obvious/elegant way

yonghong-song:
@davemarchevsky Thanks for the fix. The CI is green now. Will wait for the next revision after you addressed some of your above listed items.

yonghong-song:
xmm registers typically used to pass floating point values, it probably can carry integer value too. Do you know in this case xmm carries a floating or int value?

davemarchevsky:
> xmm registers typically used to pass floating point values, it probably can carry integer value too. Do you know in this case xmm carries a floating or int value?

I think for USDT the register is considered to hold arbitrary bytes. In this specific case I think it's holding a pointer.

But, more importantly, this PR won't work as intended. It's adding a x86-specific `__asm__` statement to the "get argument" helpers, which are compiled to BPF bytecode and executed there. Not sure how this passes verifier or what it would do if called, but definitely wouldn't grab the `xmm` value in current state.

To fix this properly we'd need to add a bpf helper which reads this register, or more likely arbitrary register. Or we could read and add more regs to the program `ctx` ahead of time, but that seems like overkill. 

A third option would be to only support USDTs which use regs in `pt_regs` to pass arguments, but given that the change to use `xmm` is in a lib that most users have no control over, and users might want to attach to USDTs in other libs which they have no control over, that feels like a bad solution that'd break user experience.

davemarchevsky:
Separately from that larger problem, I can modify this diff to just set value of USDT arg to 0 if the probe wants to pass value through `xmm`. Then tests will be unbroken.

yonghong-song:
> But, more importantly, this PR won't work as intended. It's adding a x86-specific __asm__ statement to the "get argument" helpers, which are compiled to BPF bytecode and executed there. Not sure how this passes verifier or what it would do if called, but definitely wouldn't grab the xmm value in current state.

I tried a simple example like below:
```
$ cat t.c
int test() {
  __asm__ __volatile__("movq xmm1, xmm2" ::);
  return 0;
}
$ clang -target bpf -O2 -g -c t.c
t.c:2:24: error: invalid register/token name
  __asm__ __volatile__("movq xmm1, xmm2" ::);
                       ^
<inline asm>:1:2: note: instantiated into assembly here
        movq xmm1, xmm2
        ^
1 error generated.
$
```

As expected compilation failed. Maybe the added 'movq' code is a dead code? You can use debug=4 to show the source. I agree with you for the approach below:
 > Separately from that larger problem, I can modify this diff to just set value of USDT arg to 0 if the probe wants to pass 
 > value through xmm. Then tests will be unbroken.
Let us first work around the issue and we can deal with xmm registers properly (maybe with a bpf helper) later.

PR #3885, opened 2022-02-25
Add doc on probe name defined by TRACEPOINT_PROBE

yzhao1012:
Tested seem to work.

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -220,9 +220,10 @@ TRACEPOINT_PROBE(random, urandom_read) {
 ```
 
 This instruments the tracepoint `random:urandom_read tracepoint`, and prints the tracepoint argument ```got_bits```.
-The name of the probe function is `tracepoint__random__urandom_read`.
-This tracepoint probe can be attached with:
+When using Python API, this probe is automatically attached to the right tracepoint target.
+For C++, this tracepoint probe can be attached by specifying the tracepoint target and function name explicitly:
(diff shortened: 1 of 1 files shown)


justicezyx:
> attach_tracepoint

IIUC, what you stated only applies to the Python API. We tried the BCC::BPF API, attach_tracepoint() is necessary.

Did I miss something?
If not, then the question becomes what's the expected approach to use tracepoint in BCC's BPF C++ APIs. If not attaching through this hidden behavior (which I agree), what's the other way of attaching tracepoint?

Sorry for late reply.

chenhengqi:
OK, I see. The Python and C++ APIs don't have the same implementation.

docs/reference_guide.md
```diff
@@ -216,7 +219,10 @@ TRACEPOINT_PROBE(random, urandom_read) {
 }
 ```
 
-This instruments the random:urandom_read tracepoint, and prints the tracepoint argument ```got_bits```.
+This instruments the tracepoint `random:urandom_read tracepoint`, and prints the tracepoint argument ```got_bits```.
+The name of the probe function is `tracepoint__random__urandom_read`.
+This tracepoint probe can be attached with:
```
yonghong-song, line 224: You can make it explicit that explicit tracepoint attachment is only needed for C++? For exmaple, like below
```
For C++, this tracepoint probe can be attached with:
...
```

yzhao1012: Done, in addition to mentioning C++, I added a statement on the expectation when using the Python API.

PR #3886, opened 2022-02-25
tools: fix klockstat when DEBUG_LOCK_ALLOC is set

### LINKED ISSUE #3882: klockstat Uses Incorrect kprobe When CONFIG_DEBUG_LOCK_ALLOC is Set
When the kernel config `DEBUG_LOCK_ALLOC` is set, the kprobe for `mutex_lock` appears to change to `mutex_lock_nested` which means that `klockstat` will not be able to set the correct kernel probe (here: https://github.com/iovisor/bcc/blob/master/tools/klockstat.py#L432-L433) and will experience the following error:

```
# klockstat -i 5 -n 10
cannot attach kprobe, probe entry may not exist
  File "/usr/lib/python-exec/python3.10/klockstat", line 432, in <module>
    b.attach_kretprobe(event="mutex_lock", fn_name="mutex_lock_return")
  File "/usr/lib/python3.10/site-packages/bcc/__init__.py", line 872, in attach_kretprobe
    raise Exception("Failed to attach BPF program %s to kretprobe %s" %
Exception: Failed to attach BPF program b'mutex_lock_return' to kretprobe b'mutex_lock'
```

Looking at `/sys/kernel/debug/tracing/error_log` shows:
```
[86294.825300] trace_kprobe: error: Invalid probed address or symbol
  Command: r:kprobes/r_mutex_lock_bcc_62628 mutex_lock
```

However, patching `klockstat` to instead attach to the `mutex_lock_nested` kprobe allows `klockstat` to work again.

Patch below:
```
$ cat klockstat.patch
--- old/klockstat       2022-02-24 20:50:19.055305306 +0000
+++ new/klockstat       2022-02-24 20:50:59.636482177 +0000
@@ -429,8 +429,8 @@

 if not is_support_kfunc:
     b.attach_kprobe(event="mutex_unlock",  fn_name="mutex_unlock_enter")
-    b.attach_kretprobe(event="mutex_lock", fn_name="mutex_lock_return")
-    b.attach_kprobe(event="mutex_lock",    fn_name="mutex_lock_enter")
+    b.attach_kretprobe(event="mutex_lock_nested", fn_name="mutex_lock_return")
+    b.attach_kprobe(event="mutex_lock_nested",    fn_name="mutex_lock_enter")

 enabled = b.get_table("enabled");
```

Not sure exactly the best way to check for this that is within the convention of this repo or I would submit the PR myself as this seems pretty straight forward.

aauren:
When `DEBUG_LOCK_ALLOC` is set, mutex_lock is changed to mutex_lock_nested, which means that attaching the kprobe statically to mutex_lock won't work for all kernel configurations.

This change detects the available kprobe functions exposed by the kernel and ensures that the kprobe is attached correctly.

I tested this on a machines with `DEBUG_LOCK_ALLOC` set and without `DEBUG_LOCK_ALLOC` set and this change works correctly on both.

Fixes #3882

### DIFF
--- a/tools/klockstat.py
+++ b/tools/klockstat.py
@@ -367,9 +367,30 @@ KFUNC_PROBE(mutex_lock, void *lock)
 
 """
 
+program_kfunc_nested = """
+KFUNC_PROBE(mutex_unlock, void *lock)
+{
+    return do_mutex_unlock_enter();
+}
+
+KRETFUNC_PROBE(mutex_lock_nested, void *lock, int ret)
+{
+    return do_mutex_lock_return();
+}
+
+KFUNC_PROBE(mutex_lock_nested, void *lock)
+{
+    return do_mutex_lock_enter(ctx, 3);
+}
+
+"""
+
 is_support_kfunc = BPF.support_kfunc()
 if is_support_kfunc:
-    program += program_kfunc
+    if BPF.get_kprobe_functions(b"mutex_lock_nested"):
+        program += program_kfunc_nested
+    else:
+        program += program_kfunc
 else:
     program += program_kprobe
 
@@ -428,9 +449,14 @@ program = program.replace('STACK_STORAGE_SIZE', str(args.stack_storage_size))
 b = BPF(text=program)
 
 if not is_support_kfunc:
-    b.attach_kprobe(event="mutex_unlock",  fn_name="mutex_unlock_enter")
-    b.attach_kretprobe(event="mutex_lock", fn_name="mutex_lock_return")
-    b.attach_kprobe(event="mutex_lock",    fn_name="mutex_lock_enter")
+    b.attach_kprobe(event="mutex_unlock", fn_name="mutex_unlock_enter")
+    # Depending on whether DEBUG_LOCK_ALLOC is set, the proper kprobe may be either mutex_lock or mutex_lock_nested
+    if BPF.get_kprobe_functions(b"mutex_lock_nested"):
+        b.attach_kretprobe(event="mutex_lock_nested", fn_name="mutex_lock_return")
+        b.attach_kprobe(event="mutex_lock_nested", fn_name="mutex_lock_enter")
+    else:
+        b.attach_kretprobe(event="mutex_lock", fn_name="mutex_lock_return")
+        b.attach_kprobe(event="mutex_lock", fn_name="mutex_lock_enter")
 
 enabled = b.get_table("enabled");
 


tools/klockstat.py
```diff
@@ -427,10 +435,10 @@ def display(sort, maxs, totals, counts):
 
 b = BPF(text=program)
 
-if not is_support_kfunc:
-    b.attach_kprobe(event="mutex_unlock",  fn_name="mutex_unlock_enter")
-    b.attach_kretprobe(event="mutex_lock", fn_name="mutex_lock_return")
-    b.attach_kprobe(event="mutex_lock",    fn_name="mutex_lock_enter")
+if not is_support_kfunc and kprobe_mutex_lock_func:
```
aauren, line 438: I tried to look around at surrounding usages of `get_kprobe_functions` to take my clue about error reporting. From what I could tell most of them just attach to whatever `get_kprobe_functions` finds and if it doesn't find anything, they just silently don't attach.

So that's what I implemented here. But I'm definitely willing to add error reporting if that would be better received.

tools/klockstat.py
```diff
@@ -373,6 +373,14 @@ def stack_id_err(stack_id):
 else:
     program += program_kprobe
 
+# Depending on whether DEBUG_LOCK_ALLOC is set, the proper kprobe may be either mutex_lock or mutex_lock_nested
+global kprobe_mutex_lock_func
+if not is_support_kfunc:
+    kprobe_mutex_lock_funcs = BPF.get_kprobe_functions(b'^mutex_lock(_nested)?$')
```
chenhengqi, line 379: Why regular expression here ?

Please refer to https://github.com/iovisor/bcc/commit/11614bcacdecd4d1f7015bb0f0311bb709207991

For kfunc, you should do the same check.

aauren: Updated to follow the same convention as the linked commit.

chenhengqi: Please do the same for kfunc.

aauren: Sorry, that I misunderstood you about the kfunc change. I didn't realize above that you were commenting on two separate changes that needed to be made.

So, I've done an updated version that I _think_ addresses the kfunc issue that you mentioned. However, because of the way the machines that I have access to define `bpf_trampoline_link_prog` none of my systems actually use the kfuncs, so I don't have a way to test it. Additionally, I'm relatively new to BPF stuff.

So the patch that I've updated on this MR is really just a guess, sorry if I get it wrong or took it the wrong way.

PR #3889, opened 2022-03-02
libbpf-tools: Add experimental BTFGen integration

mauriciovasquezbernal:
This PR enables an experimental integration with [BTFGen](https://lore.kernel.org/bpf/20220215225856.671072-1-mauricio@kinvolk.io/). The goal of this is to extend the range of systems where these tools can be run, in this case to allow running some of the tools in systems that don't expose BTF information. 

### How to use?

Snip from README.md:

```bash
# be sure that new bpftool submodule is updated
$ git submodule update --init --recursive

# compile bcc tools enabling this support
$  ENABLE_MIN_CORE_BTFS=1 make -j$(nproc)

# execute the tools in a system without CONFIG_DEBUG_INFO_BTF
$ cat /boot/config-$(uname -r) | grep CONFIG_DEBUG_INFO_BTF | wc -l
0

$ sudo ./execsnoop 
[...]

$ sudo ./opensnoop 
[...]
```

### FAQ:

All the following numbers were gathered on a AMD Ryzen 7 3700X, 32GB RAM, SN750 NMVe system. 

#### Why disabled by default?

Cloning btfhub-archive and generating the reduced BTF files requires some extra time. Since not all users are interested in this feature let's keep it disabled by default. 

#### What's the compilation time overhead? 

The overhead is about 30 seconds when compiling with `-j16`:

```
$ make clean
  CLEAN    
$ time make -j16 > /dev/null

real	0m4,546s
user	0m30,908s
sys	0m3,138s

$ make clean
  CLEAN    
$ time BTF_HUB_ARCHIVE=... ENABLE_MIN_CORE_BTFS=1 make -j16 > /dev/null

real	0m27,962s
user	3m50,755s
sys	0m18,659s
```

#### What's the size overhead of this? 

It's about 100KB per binary:

```
# without this 
$ ls -lh execsnoop opensnoop
-rwxr-xr-x 1 mvb mvb 1,7M Feb 28 11:00 execsnoop
-rwxr-xr-x 1 mvb mvb 1,6M Feb 28 11:00 opensnoop

# with this 
$ ls -lh execsnoop opensnoop
-rwxr-xr-x 1 mvb mvb 1,8M Feb 28 11:00 execsnoop
-rwxr-xr-x 1 mvb mvb 1,7M Feb 28 11:00 opensnoop
```

#### Why to compress the files?

The size of the uncompressed BTF files is about 3.3MB, compressed they are only ~100KB. 

```
$ find .output/min_core_btfs/ -name "*.btf" | xargs du -ch | tail -n 1
3,3M	total
$ ls -lh .output/min_core_btfs.tar.gz 
-rw-r--r-- 1 mvb mvb 87K Feb 28 11:00 .output/min_core_btfs.tar.gz
```

Also libz is already a dependency of these tools, so we can use it a runtime without introducing a new one. 

The compression ratio is so high because many of the generated BTF files are the same. Even if we can try to fix it directly in bpftool, it's easier in the time being to use a compression process. 

#### What's the runtime overhead?

Decompressing and saving the file takes around ~10ms. On systems that expose BTF the overhead should be minimal as there's some logic to return early avoiding decompressing and so on.

#### Why do you say only **some** tools?

Tools that use fentry/fexit hooks can't work with this because they require the kernel to expose BTF information.

#### Why only execsnoop and opensnoop are updated? 

I'll update other tools once this approach makes sense for the maintainers.

### TODO

- [X] Use bpftool as submodule
- [X] Update makefile to clone BTFHub repo
- [X] Implement support in other tools

/cc @anakryiko


anakryiko:
Hi! Thanks for working on this!

I didn't manage to get it working locally for some reason (I think I followed your instructions but I got no BTFs). Makefile was also failing until I did `mkdir -p .output/min_core_btfs`, so you might want to fix that.

But here are some overall thoughts:
1. For the btfhub-archive. I see that it contains BTFs for x86-64 and arm64 intermixed. It's quite unfortunate that I need to download both architectures even though I will use just one architecture. Do you think it make sense to have mulitple repos instead, one for each supported architecture? This way makefile can be taught to download only architecture you need.
2. Ideally, if minimized BTFs are requested, it would be great for Makefile to download it, if necessary. I think it should be pretty easy to support efficiently by git cloning into some local .btfs directory, and then doing git pull (to fetch any updates). It would be especially great if we structure repos in a way that allows downloading only single architecture.
3. There are some improvements that can be done in a way that we embed that tar file into application. I'd use objcopy to create ELF object file and link it into final application. Then we can use __weak extern symbol to detect if application was built with embedded BTFs or not. In the application itself we can trim it down to a very simple:
```
   if (!ensure_core_btf(&opts)) {
       fprintf(stderr, "Couldn't fetch necessary BTF for CO-RE!\n");
       return 1;
   }
```
where opts are bpf_object_open_opts.
4. We might want to also add a way to specify custom BTF not by file path but directly as array of bytes (e.g., .custom_btf_data and .custom_btf_sz as an alternative to .custom_btf_path). Mostly to avoid dealing with temporary files. But this is minor, we can figure this out later.

Now, to the question about bpftool. Given bpftool is now mirrored into Github repo, I think we should add bpftool as submodule and build it from sources. This bypasses any need to pre-checking it, any architecture questions. We'll just build it on demand. WDYT?

mauriciovasquezbernal:
> I didn't manage to get it working locally for some reason (I think I followed your instructions but I got no BTFs). 

Perhaps you're running a system with an older glibc and bpftool can't be run there? Or you forgot to uncompress the BTF files? 

> Makefile was also failing until I did `mkdir -p .output/min_core_btfs`, so you might want to fix that.

That's weird, the Makefile should create it https://github.com/kinvolk/bcc/blob/mauricio/btfgen-integration/libbpf-tools/Makefile#L135

>1. For the btfhub-archive. I see that it contains BTFs for x86-64 and arm64 intermixed. It's quite unfortunate that I need to download both architectures even though I will use just one architecture. Do you think it make sense to have mulitple repos instead, one for each supported architecture? This way makefile can be taught to download only architecture you need.
> 

Even if I agree with this, I'm not in charge of BTFHub, let's ping @rafaeldtinoco on this one. 

>2. Ideally, if minimized BTFs are requested, it would be great for Makefile to download it, if necessary. I think it should be pretty easy to support efficiently by git cloning into some local .btfs directory, and then doing git pull (to fetch any updates). It would be especially great if we structure repos in a way that allows downloading only single architecture.
> 

I tried to do it but hit a problem. I'm using a variable 

```
SOURCE_BTF_FILES = $(shell find $(BTF_HUB_ARCHIVE)/ -iregex ".*$(subst x86,x86_64,$(ARCH)).*" -type f -name '*.btf')
```

to get the list of BTF files and create dependencies between them and the .bpf.o files. This variable is populated when the Makefile is read, so if I include a step to clone btfhub that variable will be empty. Right now I don't have a clear solution for this on mind, do you any idea? 

>     3. There are some improvements that can be done in a way that we embed that tar file into application. I'd use objcopy to create ELF object file and link it into final application. Then we can use __weak extern symbol to detect if application was built with embedded BTFs or not. In the application itself we can trim it down to a very simple:
> 
> 
> ```
>    if (!ensure_core_btf(&opts)) {
>        fprintf(stderr, "Couldn't fetch necessary BTF for CO-RE!\n");
>        return 1;
>    }
> ```
> 
> where opts are bpf_object_open_opts.

I used `ld` of `objcopy`, I think there's not a big difference in this case. Also, I didn't consider support for cross-compiling, anyway I think it's not supported now. 

> 4. We might want to also add a way to specify custom BTF not by file path but directly as array of bytes (e.g., .custom_btf_data and .custom_btf_sz as an alternative to .custom_btf_path). Mostly to avoid dealing with temporary files. But this is minor, we can figure this out later.
> 

I agree. Is it better to receive an array of bytes or would it be better to receive the `struct btf` and let parsing to the caller? 

> Now, to the question about bpftool. Given bpftool is now mirrored into Github repo, I think we should add bpftool as submodule and build it from sources. This bypasses any need to pre-checking it, any architecture questions. We'll just build it on demand. WDYT?

I agree, it's not so nice to have a binary checked on repo. I'll update it once btfgen makes it to the mirror.

anakryiko:
> > I didn't manage to get it working locally for some reason (I think I followed your instructions but I got no BTFs).
> 
> Perhaps you're running a system with an older glibc and bpftool can't be run there? Or you forgot to uncompress the BTF files?

No, it was my usual dev server, where everything BPF related works just fine, but I think I missed the uncompression step, which probably explains why I didn't get any BTFs packaged.

> 
> > Makefile was also failing until I did `mkdir -p .output/min_core_btfs`, so you might want to fix that.
> 
> That's weird, the Makefile should create it https://github.com/kinvolk/bcc/blob/mauricio/btfgen-integration/libbpf-tools/Makefile#L135
> 
> > 1. For the btfhub-archive. I see that it contains BTFs for x86-64 and arm64 intermixed. It's quite unfortunate that I need to download both architectures even though I will use just one architecture. Do you think it make sense to have mulitple repos instead, one for each supported architecture? This way makefile can be taught to download only architecture you need.
> 
> Even if I agree with this, I'm not in charge of BTFHub, let's ping @rafaeldtinoco on this one.

@rafaeldtinoco, any opinion? I'm trying to minimize the amount of data that needs to be downloaded during build if btfgen is enabled. Having per-architecture repository or one repository with per-architecture submodule (if that's more convenient) would make sure we don't download unnecessary gigabytes of data.

> 
> > 2. Ideally, if minimized BTFs are requested, it would be great for Makefile to download it, if necessary. I think it should be pretty easy to support efficiently by git cloning into some local .btfs directory, and then doing git pull (to fetch any updates). It would be especially great if we structure repos in a way that allows downloading only single architecture.
> 
> I tried to do it but hit a problem. I'm using a variable
> 
> ```
> SOURCE_BTF_FILES = $(shell find $(BTF_HUB_ARCHIVE)/ -iregex ".*$(subst x86,x86_64,$(ARCH)).*" -type f -name '*.btf')
> ```
> 
> to get the list of BTF files and create dependencies between them and the .bpf.o files. This variable is populated when the Makefile is read, so if I include a step to clone btfhub that variable will be empty. Right now I don't have a clear solution for this on mind, do you any idea?

You are trying to do it declaratively though Makefile rules. It would be easier and simpler to do it imperatively. When we need to rebuild tools, do an explicit step of checking out or git pull'ing latest BTFs, then find all relevant files, minimize BTFs, and then link them into final binaries.

> 
> > 3. There are some improvements that can be done in a way that we embed that tar file into application. I'd use objcopy to create ELF object file and link it into final application. Then we can use __weak extern symbol to detect if application was built with embedded BTFs or not. In the application itself we can trim it down to a very simple:
> > ```
> >    if (!ensure_core_btf(&opts)) {
> >        fprintf(stderr, "Couldn't fetch necessary BTF for CO-RE!\n");
> >        return 1;
> >    }
> > ```
> > where opts are bpf_object_open_opts.
> 
> I used `ld` of `objcopy`, I think there's not a big difference in this case. Also, I didn't consider support for cross-compiling, anyway I think it's not supported now.
> 

Yeah, I cross-compiling is always harder and should be solved later (if there is enough interest). 

> > 4. We might want to also add a way to specify custom BTF not by file path but directly as array of bytes (e.g., .custom_btf_data and .custom_btf_sz as an alternative to .custom_btf_path). Mostly to avoid dealing with temporary files. But this is minor, we can figure this out later.
> 
> I agree. Is it better to receive an array of bytes or would it be better to receive the `struct btf` and let parsing to the caller?

I think raw bytes are move convenient (especially for applications that will embed minimized BTFs as is, without compression). It's analogous to `bpf_object__open_file()` vs `bpf_object__open_mem()`. We have open_file variant with custom_btf_path, we can have open_mem variant as well.

> 
> > Now, to the question about bpftool. Given bpftool is now mirrored into Github repo, I think we should add bpftool as submodule and build it from sources. This bypasses any need to pre-checking it, any architecture questions. We'll just build it on demand. WDYT?
> 
> I agree, it's not so nice to have a binary checked on repo. I'll update it once btfgen makes it to the mirror.

@qmonnet can you please help with the sync of bpftool into Github to unblock this for btfgen stuff?

qmonnet:
> @qmonnet can you please help with the sync of bpftool into Github to unblock this for btfgen stuff?

Sure thing! I was waiting for your latest libbpf update, so I'd pick [the `reallocarray` fix](https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git/commit/?id=a19df7139440258e02126f1c795ba64932a8e949) automatically and avoid breaking the build when reusing libbpf's CHECKPOINT_COMMIT to sync up. I saw that you updated yesterday, and I just updated bpftool's mirror based on that, so we're all good.

libbpf-tools/min_core_btf_helpers.c
```diff
@@ -0,0 +1,197 @@
...
+			*dst = realloc(*dst, size);
+		}
+	} while (ret != Z_STREAM_END);
+
+	*dst_size = strm.total_out;
+
+out:
+	/* clean up and return */
+	inflateEnd(&strm);
+	return ret == Z_STREAM_END ? Z_OK : Z_DATA_ERROR;
```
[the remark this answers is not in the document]
mauriciovasquezbernal: It freed by `ensure_core_btf` after writing it to a file.

libbpf-tools/min_core_btf_helpers.c
```diff
@@ -0,0 +1,197 @@
...
+
+	strncpy(info.kernel_release, u.release, FIELD_LEN);
+	strncpy(info.arch, u.machine, FIELD_LEN);
+
+	f = fopen("/etc/os-release", "r");
+	if (!f)
+		return info;
+
+	while ((read = getline(&line, &len, f)) != -1) {
+		if (strncmp(id_prefix, line, strlen(id_prefix)) == 0) {
```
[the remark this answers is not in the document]
mauriciovasquezbernal: What do you mean exactly by using `fscanf()`?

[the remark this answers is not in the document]
chenhengqi: sscanf ?

https://github.com/iovisor/bcc/blob/03e66ed5790f2ec1ad01ab2b17f10a749b6dd019/libbpf-tools/bashreadline.c#L122-L129

mauriciovasquezbernal:
I handled some of the feedback and updated all the tools that are compatible with it. I'll try to handle missing comments shortly.

libbpf-tools/min_core_btf_helpers.c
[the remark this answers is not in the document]
anakryiko: e.g., (from selftests's trace_helpers):
```
while (fscanf(f, "%llx %c %499s%*[^\n]\n", &value, &type, name) > 0) {
    ...
}
```

You'll need a different format string, of course.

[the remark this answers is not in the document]
mauriciovasquezbernal: I don't think that `fscanf()` can be used in the `while` loop. I need to read the whole line and then check if this starts with the prefix we're interested in. I used `sscanf()` as suggested by @chenhengqi and I think the result is much better than before. WDYT?

mauriciovasquezbernal:
> You are trying to do it declaratively though Makefile rules. It would be easier and simpler to do it imperatively. When we need to rebuild tools, do an explicit step of checking out or git pull'ing latest BTFs, then find all relevant files, minimize BTFs, and then link them into final binaries.

I ended up splitting the makefile in two, moving the btfgen specific logic to a new one. The first makefile clones and updates the btfhub repository (if needed) and btfgen.mk creates all "reduced" BTF files using declarative rules.

libbpf-tools/Makefile
```diff
@@ -1,21 +1,23 @@
...
+export BPFTOOL ?= $(OUTPUT)/bpftool/bpftool
 LIBBPF_SRC := $(abspath ../src/cc/libbpf/src)
 LIBBPF_OBJ := $(abspath $(OUTPUT)/libbpf.a)
 INCLUDES := -I$(OUTPUT) -I../src/cc/libbpf/include/uapi
 CFLAGS := -g -O2 -Wall
 INSTALL ?= install
 prefix ?= /usr/local
-ARCH := $(shell uname -m | sed 's/x86_64/x86/' | sed 's/aarch64/arm64/' | sed 's/ppc64le/powerpc/' | sed 's/mips.*/mips/')
+export ARCH := $(shell uname -m | sed 's/x86_64/x86/' | sed 's/aarch64/arm64/' | sed 's/ppc64le/powerpc/' | sed 's/mips.*/mips/')
+export BTF_HUB_ARCHIVE ?= $(shell echo $$HOME/.local/share/btfhub-archive)
```
anakryiko, line 14: nit: can we have exports on a separate line? exported intermixed with non-exported variables are hard to see, but on the other hand break the reading flow a bit

libbpf-tools/Makefile
```diff
@@ -91,6 +98,11 @@ $(OUTPUT) $(OUTPUT)/libbpf:
 	$(call msg,MKDIR,$@)
 	$(Q)mkdir -p $@
 
+.PHONY: BPFTOOL
+BPFTOOL:
```
anakryiko, line 102: why all caps?

libbpf-tools/Makefile
```diff
@@ -91,6 +98,11 @@ $(OUTPUT) $(OUTPUT)/libbpf:
...
+	$(Q)mkdir -p $(OUTPUT)/bpftool
+	$(Q)OUTPUT=$(OUTPUT)/bpftool/ $(MAKE) -C $(BPFTOOL_SRC)
```
anakryiko, line 104: make accepts such definitions directly, I think it's better to use this way:

```
$(Q)$(MAKE) OUTPUT=$(OUTPUT)/bpftool/ -C $(BPFTOOL_SRC)
```

libbpf-tools/Makefile
anakryiko, line 14: I also think that having repo-local cache for btfhub is better than using $HOME, wdyt?

libbpf-tools/Makefile
```diff
@@ -111,6 +123,22 @@ $(OUTPUT)/%.bpf.o: %.bpf.c $(LIBBPF_OBJ) $(wildcard %.h) $(ARCH)/vmlinux.h | $(O
...
+	$(Q)cd $(BTF_HUB_ARCHIVE) && git pull
+
+$(BTF_HUB_ARCHIVE):
+	$(call msg,GIT CLONE,$@)
+	$(Q)git clone https://github.com/aquasecurity/btfhub-archive/ $(BTF_HUB_ARCHIVE)
+
+BPF_O_FILES = $(patsubst %,$(OUTPUT)/%.bpf.o,$(APPS))
+
+$(OUTPUT)/min_core_btf_tar.o: $(BPF_O_FILES) update_btfhub
+	$(Q)$(MAKE) -f btfgen.mk
```
anakryiko, line 139: naming nit: btfgen.mk -> Makefile.btfgen? It's much easier to spot makefile that actually spells out "Makefile" in the name

libbpf-tools/Makefile
```diff
@@ -111,6 +123,22 @@ $(OUTPUT)/%.bpf.o: %.bpf.c $(LIBBPF_OBJ) $(wildcard %.h) $(ARCH)/vmlinux.h | $(O
 		     -I$(ARCH)/ $(INCLUDES) -c $(filter %.c,$^) -o $@ &&      \
 	$(LLVM_STRIP) -g $@
 
+ifdef ENABLE_MIN_CORE_BTFS
+
+.PHONY: update_btfhub
+update_btfhub: $(BTF_HUB_ARCHIVE)
```
anakryiko, line 129: might be forgetting some Make quirk, but why extra rule if you can just do that cd + git pull right there in min_core_btf_tar.o rule?

libbpf-tools/README.md
```diff
@@ -92,3 +92,30 @@ CONFIG_DEBUG_INFO=y
...
+are shipped with the tools in order to provide the needed information to
+perform the CO-RE relocations when loading the eBPF programs.
+
+If you haven't cloned the
+[btfhub-archive](https://github.com/aquasecurity/btfhub) repository, you
+can run make and it'll clone it for you into the `$HOME/.local/share`
+directory:
+
+```bash
+ENABLE_MIN_CORE_BTFS=1 make -j$(nproc)
```
anakryiko, line 113: similar here, it seems customary to do it like this:

```
make ENABLE_MIN_CORE_BTFS=1 -j$(nproc)
```

There are also subtle semantic differences, as this way ENABLE_MIN_CORE_BTFS will override any `ENABLE_MIN_CORE_BTFS ?= <my default value>` definitions, while the `ENABLE_MIN_CORE_BTFS=1 make` won't (at least that's why I remember from times when I had to deal with Makefiles a bit more).

libbpf-tools/README.md
```diff
@@ -92,3 +92,30 @@ CONFIG_DEBUG_INFO=y
...
+```
+
+If you have a local copy of such repository, you can pass it's location
+to avoid cloning it again:
+
+```bash
+BTF_HUB_ARCHIVE=<path_to_btfhub-archive> ENABLE_MIN_CORE_BTFS=1 make -j$(nproc)
```
anakryiko, line 120: same here, let's put make in front

libbpf-tools/Makefile
```diff
@@ -111,6 +123,22 @@ $(OUTPUT)/%.bpf.o: %.bpf.c $(LIBBPF_OBJ) $(wildcard %.h) $(ARCH)/vmlinux.h | $(O
...
+	$(Q)cd $(BTF_HUB_ARCHIVE) && git pull
+
+$(BTF_HUB_ARCHIVE):
+	$(call msg,GIT CLONE,$@)
+	$(Q)git clone https://github.com/aquasecurity/btfhub-archive/ $(BTF_HUB_ARCHIVE)
+
+BPF_O_FILES = $(patsubst %,$(OUTPUT)/%.bpf.o,$(APPS))
```
anakryiko, line 136: given this definition is used in one place in one rule, let's just inline it (at least for now)?

libbpf-tools/btfgen.mk
```diff
@@ -0,0 +1,40 @@
...
+endif
+
+$(BTF_HUB_ARCHIVE)/%.btf: $(BTF_HUB_ARCHIVE)/%.btf.tar.xz
+	$(call msg,UNTAR,$@)
+	$(Q)tar xvfJ $< -C "$(@D)" > /dev/null
+	$(Q)touch $@
+
+$(MIN_CORE_BTF_FILES): $(BPF_O_FILES)
+
+# Create reduced version of BTF files to be embeed within the tools executables
```
anakryiko, line 25: typo: embedded

libbpf-tools/fsslower.c
```diff
@@ -451,5 +459,10 @@ int main(int argc, char **argv)
 	perf_buffer__free(pb);
 	fsslower_bpf__destroy(skel);
 
+	if (open_opts.btf_custom_path) {
+		unlink(open_opts.btf_custom_path);
+		free((void *)open_opts.btf_custom_path);
+	}
+
```
anakryiko, line 466: give this is repeated everywhere, let's have a counter-part to ensure_core_btf() and move this into "cleanup_core_btf()" helper?

libbpf-tools/main.c
```diff
@@ -0,0 +1,32 @@
+#include <stdio.h>
+
+extern char _binary_min_core_btfs_tar_gz_start[] __attribute__((weak));
```
anakryiko, line 3: is this some debug leftover file?

libbpf-tools/min_core_btf_helpers.c
```diff
@@ -0,0 +1,231 @@
...
+	char version[2];
+	char uname[32];
+	char gname[32];
+	char devmajor[8];
+	char devminor[8];
+	char prefix[155];
+	char padding[12];
+};
+
+char *tar_file_start(struct tar_header *tar, const char *name, int *length)
```
anakryiko, line 153: please mark all the functions static in this file

libbpf-tools/oomkill.c
```diff
@@ -16,6 +16,7 @@
 #include <bpf/libbpf.h>
 #include "oomkill.skel.h"
 #include "oomkill.h"
+#include "min_core_btf_helpers.h"
```
anakryiko, line 19: sorry, another naming nit: I'd generalize this header to be just "btf_helpers.h" so that we can use it for any other BTF-related helpers in the future (e.g., if we need some BTF-based feature probing helpers or something like that). "min_core_btf_helpers.h" is a bit too specific.

anakryiko (changes requested):
Sorry, this dropped off my radar for a while, sorry for later review.

I did the first pass, I intend to play a bit more with it.

One more thing besides my comments, I didn't notice any .gitignore changes, we'll need to make sure all the temporary stuff is gitignored properly.

anakryiko (changes requested):
So few more things after I played with this locally.

- Ignore what I said about .gitignore, you've put everything under .output, so all good.

- Some dependencies are off in Makefile, I get a lot of errors similar to below:
```
/bin/sh: /data/users/andriin/bcc/libbpf-tools/.output/bpftool/bpftool: No such file or directory
make[1]: *** [btfgen.mk:29: /data/users/andriin/bcc/libbpf-tools/.output/min_core_btfs/amzn/2/x86_64/4.14.101-91.76.amzn2.x86_64.btf] Error 127
```
Please check that and fix.

But even when it does succeed building, I don't see `_binary_min_core_btfs_tar_gz_start` symbol and running under GDB I get -EOPNOTSUPP from ensure_core_btf(). How did you test this? Am  I missing some extra steps. I just did this:

```
$ make ENABLE_MIN_CORE_BTFS=1 -j90
```

anakryiko:
>     1. For the btfhub-archive. I see that it contains BTFs for x86-64 and arm64 intermixed. It's quite unfortunate that I need to download both architectures even though I will use just one architecture. Do you think it make sense to have mulitple repos instead, one for each supported architecture? This way makefile can be taught to download only architecture you need.
> 

@rafaeldtinoco, ping? Any thoughts?

rafaeldtinoco:
> >     1. For the btfhub-archive. I see that it contains BTFs for x86-64 and arm64 intermixed. It's quite unfortunate that I need to download both architectures even though I will use just one architecture. Do you think it make sense to have mulitple repos instead, one for each supported architecture? This way makefile can be taught to download only architecture you need.
> > 
> 
> @rafaeldtinoco, ping? Any thoughts?

Sorry, was recovering from surgery, I'll visit this tomorrow and provide some feedback.

libbpf-tools/Makefile
mauriciovasquezbernal: I'm not sure what option is better. I'll move it to the repo.

mauriciovasquezbernal (commented):
> Some dependencies are off in Makefile, I get a lot of errors similar to below:

There was a missing dependency on bpftool. 

> How did you test this?

I tested on a clean VM (Ubuntu 20.04.4) and the compilation worked fine by just using `make ENABLE_MIN_CORE_BTFS=1 -j$(nproc)`. 

Are the BTF files being generated? 
```
$ find .output/min_core_btfs/ -name "*.btf" | xargs du -ch | tail -n 1
3.3M	total
```

Are symbols present on ` .output/min_core_btf_tar.o`?  

```
$ objdump --syms .output/min_core_btf_tar.o

.output/min_core_btf_tar.o:     file format elf64-x86-64

SYMBOL TABLE:
0000000000000000 l    d  .data	0000000000000000 .data
0000000000015f39 g       .data	0000000000000000 _binary_min_core_btfs_tar_gz_end
0000000000015f39 g       *ABS*	0000000000000000 _binary_min_core_btfs_tar_gz_size
0000000000000000 g       .data	0000000000000000 _binary_min_core_btfs_tar_gz_start
```

> $ make ENABLE_MIN_CORE_BTFS=1 -j90

I'm only trying with 8 or 16 cores. Perhaps there was still some dependencies problems in the Makefile that don't have effect with a lower number of cores :thinking:. Also, have you tried to run `make clean` and compile again?

anakryiko:
there were BTFs downloaded, but there were no symbols. After pulling your latest changes I do see symbols now, so much have been some race condition in Makefile or something.

libbpf-tools/.gitignore
```diff
@@ -1,4 +1,5 @@
 /.output
+/btfhub-archive
```
anakryiko, line 2: nit: let's call it just "btfhub"? archive implies it's outdated and not maintained anymore. So for consistency with MAkefile, let's call it "btf-hub" here and $BTF_HUB in Makefile?

libbpf-tools/Makefile
```diff
@@ -1,21 +1,25 @@
 # SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
-OUTPUT := .output
+# export variables that are used in Makefile.btfgen as well.
+export OUTPUT := $(abspath .output)
+export BPFTOOL ?= $(OUTPUT)/bpftool/bpftool
+export ARCH := $(shell uname -m | sed 's/x86_64/x86/' | sed 's/aarch64/arm64/' | sed 's/ppc64le/powerpc/' | sed 's/mips.*/mips/')
+export BTF_HUB_ARCHIVE ?= $(abspath btfhub-archive)
```
anakryiko, line 6: I mean, it's fine, but I meant to split variable declaration and exporting:

```
OUTPUT = ...
BPFTOOL ?= ...
...

export OUTPUT BPFTOOL ARCH BTF_HUB_ARCHIVE
```

libbpf-tools/Makefile
```diff
@@ -1,21 +1,25 @@
...
 INSTALL ?= install
 prefix ?= /usr/local
-ARCH := $(shell uname -m | sed 's/x86_64/x86/' | sed 's/aarch64/arm64/' | sed 's/ppc64le/powerpc/' | sed 's/mips.*/mips/')
 
 ifeq ($(wildcard $(ARCH)/),)
 $(error Architecture $(ARCH) is not supported yet. Please open an issue)
 endif
 
-APPS = \
+export APPS = \
```
anakryiko, line 22: same, please don't export variable as you are defining it. Just do all the exports after variables are defined

libbpf-tools/Makefile
```diff
@@ -68,8 +72,13 @@ COMMON_OBJ = \
 	$(OUTPUT)/errno_helpers.o \
 	$(OUTPUT)/map_helpers.o \
 	$(OUTPUT)/uprobe_helpers.o \
+	$(OUTPUT)/btf_helpers.o \
 	#
 
+ifdef ENABLE_MIN_CORE_BTFS
+COMMON_OBJ += $(OUTPUT)/min_core_btf_tar.o
+endif
```
anakryiko, line 80: you can do it more succinctly:

```
$(OUTPUT)/btf_helpers.o \
$(if $(ENABLE_MIN_CORE_BTFS),$(OUTPUT)/min_core_btf_tar.o) \
#
```

libbpf-tools/Makefile
```diff
@@ -111,6 +125,16 @@ $(OUTPUT)/%.bpf.o: %.bpf.c $(LIBBPF_OBJ) $(wildcard %.h) $(ARCH)/vmlinux.h | $(O
...
+ifdef ENABLE_MIN_CORE_BTFS
+$(BTF_HUB_ARCHIVE):
+	$(call msg,GIT CLONE,$@)
+	$(Q)git clone -q https://github.com/aquasecurity/btfhub-archive/ $(BTF_HUB_ARCHIVE)
+
+$(OUTPUT)/min_core_btf_tar.o: $(patsubst %,$(OUTPUT)/%.bpf.o,$(APPS)) $(BTF_HUB_ARCHIVE) | bpftool
+	$(Q)cd $(BTF_HUB_ARCHIVE) && git pull
+	$(Q)$(MAKE) -f Makefile.btfgen
+endif
+
```
anakryiko, line 137: this has a downside of not checking if anything changed about btf-hub. I did this locally and it seems to work well plus it defines "btf-hub" command to fetch it, even if we don't specify ENABLE_MIN_CORE_BTFS.

```diff
diff --git a/libbpf-tools/Makefile b/libbpf-tools/Makefile
index caf39b6409c1..6426262ea395 100644
--- a/libbpf-tools/Makefile
+++ b/libbpf-tools/Makefile
@@ -125,13 +125,13 @@ $(OUTPUT)/%.bpf.o: %.bpf.c $(LIBBPF_OBJ) $(wildcard %.h) $(ARCH)/vmlinux.h | $(O
                     -I$(ARCH)/ $(INCLUDES) -c $(filter %.c,$^) -o $@ &&      \
        $(LLVM_STRIP) -g $@

-ifdef ENABLE_MIN_CORE_BTFS
-$(BTF_HUB_ARCHIVE):
-       $(call msg,GIT CLONE,$@)
-       $(Q)git clone -q https://github.com/aquasecurity/btfhub-archive/ $(BTF_HUB_ARCHIVE)
-
-$(OUTPUT)/min_core_btf_tar.o: $(patsubst %,$(OUTPUT)/%.bpf.o,$(APPS)) $(BTF_HUB_ARCHIVE) | bpftool
+btf-hub: force
+       $(call msg,GIT,$@)
+       $(Q)[ -d "$(BTF_HUB_ARCHIVE)" ] || git clone -q https://github.com/aquasecurity/btfhub-archive/ $(BTF_HUB_ARCHIVE)
        $(Q)cd $(BTF_HUB_ARCHIVE) && git pull
+
+ifdef ENABLE_MIN_CORE_BTFS
+$(OUTPUT)/min_core_btf_tar.o: $(patsubst %,$(OUTPUT)/%.bpf.o,$(APPS)) btf-hub | bpftool
        $(Q)$(MAKE) -f Makefile.btfgen
 endif

@@ -157,6 +157,9 @@ install: $(APPS) $(APP_ALIASES)
        $(Q)$(INSTALL) $(APPS) $(DESTDIR)$(prefix)/bin
        $(Q)cp -a $(APP_ALIASES) $(DESTDIR)$(prefix)/bin

+.PHONY: force
+force:
+
```

This will always try to fetch latest btfhub version before building.

libbpf-tools/btf_helpers.c
```diff
@@ -0,0 +1,242 @@
...
+	if (!info)
+		return -errno;
+
+	ret = mkstemp(btf_path);
+	if (ret < 0)
+		return -errno;
+
+	dst = fdopen(ret, "wb");
+	if (!dst)
+		return -errno;
```
anakryiko, line 194: you are leaking os_info here

anakryiko (changes requested):
I think it's almost ready, just few remaining small issues.

libbpf-tools/btf_helpers.c
```diff
@@ -0,0 +1,246 @@
...
+	if (uname(&u) == -1)
+		return NULL;
+
+	f = fopen("/etc/os-release", "r");
+	if (!f)
+		return NULL;
+
+	info = calloc(1, sizeof(*info));
+	if (!info)
+		return NULL;
```
anakryiko, line 45: hm.. another leak, f is not closed, missed last time

libbpf-tools/.gitignore
rafaeldtinoco: Yep, we've split the scripts/tooling/docs from the actual archive. There were major discussion where to keep the BTF files back then, and, to be honest, github was what made it to have a higher reach (among projects)... Unfortunately we don't keep different arches (maybe they could be kept as diff branches). I'm just afraid of breaking other projects (as we currently have 4 or 5 projects relying in BTFHUB now).

anakryiko (approved):
There is one more tiny leak of resources, please fix. But otherwise looks and seem to work great (I don't have system without BTF to check that part, though).

@davemarchevsky, @yonghong-song with that minor fix it should be ready to go in

libbpf-tools/.gitignore
anakryiko: oh, separate branches is an interesting idea! You can still master branch that merge all the arch-specific branches each time you get an update. So no one should break, but anyone who wants only architecture-specific BTFs can do that without paying for all the other architectures.

rickysarraf:
Thank you @mika for pointing this PR.

In general, project wide, do we have some table about the compatibility with different versions of external dependencies ? If I remember correct, @copyninja was already having some integration issues when building recent versions of `bcc + libbpf` for Debian.

davemarchevsky:
LGTM

@mauriciovasquezbernal can you rebase? Will press button after

mauriciovasquezbernal:
@davemarchevsky I rebased and pushed.

PR #3893, opened 2022-03-03
libbpf-tools: Add biotop tool.

eiffel-fl:
Hi.

I converted `biotop` from BCC to CO-RE.
I tried to stick with the BCC behavior.

I just have two questions:
1. I only probed on some functions, is this OK? Because I think we cannot try to probe like it it is done here:
https://github.com/iovisor/bcc/blob/ca8240205e175e94cef3c70c3b5c5accbe51512f/tools/biotop.py#L178
2. I am not really satisfied with the way I handle disk name. Sadly it is a bit complicated to use a hash map in C. So if someone sees a way to improve this, feel free to share :D.

Best regards.

### DIFF
--- a/libbpf-tools/biotop.bpf.c
+++ b/libbpf-tools/biotop.bpf.c
@@ -1,0 +1,106 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2022 Francis Laniel <flaniel@linux.microsoft.com>
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_tracing.h>
+
+#include "biotop.h"
+#include "maps.bpf.h"
+#include "core_fixes.bpf.h"
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 10240);
+	__type(key, struct request *);
+	__type(value, struct start_req_t);
+} start SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 10240);
+	__type(key, struct request *);
+	__type(value, struct who_t);
+} whobyreq SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 10240);
+	__type(key, struct info_t);
+	__type(value, struct val_t);
+} counts SEC(".maps");
+
+SEC("kprobe")
+int BPF_KPROBE(blk_account_io_start, struct request *req)
+{
+	struct who_t who = {};
+
+	/* cache PID and comm by-req */
+	bpf_get_current_comm(&who.name, sizeof(who.name));
+	who.pid = bpf_get_current_pid_tgid() >> 32;
+	bpf_map_update_elem(&whobyreq, &req, &who, 0);
+
+	return 0;
+}
+
+SEC("kprobe/blk_mq_start_request")
+int BPF_KPROBE(blk_mq_start_request, struct request *req)
+{
+	/* time block I/O */
+	struct start_req_t start_req;
+
+	start_req.ts = bpf_ktime_get_ns();
+	start_req.data_len = BPF_CORE_READ(req, __data_len);
+
+	bpf_map_update_elem(&start, &req, &start_req, 0);
+	return 0;
+}
+
+SEC("kprobe")
+int BPF_KPROBE(blk_account_io_done, struct request *req, u64 now)
+{
+	struct val_t *valp, zero = {};
+	struct info_t info = {};
+	struct start_req_t *startp;
+	unsigned int cmd_flags;
+	struct gendisk *disk;
+	struct who_t *whop;
+	u64 delta_us;
+
+	/* fetch timestamp and calculate delta */
+	startp = bpf_map_lookup_elem(&start, &req);
+	if (!startp)
+		return 0;    /* missed tracing issue */
+
+	delta_us = (bpf_ktime_get_ns() - startp->ts) / 1000;
+
+	/* setup info_t key */
+	cmd_flags = BPF_CORE_READ(req, cmd_flags);
+
+	disk = get_disk(req);
+	info.major = BPF_CORE_READ(disk, major);
+	info.minor = BPF_CORE_READ(disk, first_minor);
+	info.rwflag = !!((cmd_flags & REQ_OP_MASK) == REQ_OP_WRITE);
+
+	whop = bpf_map_lookup_elem(&whobyreq, &req);
+	if (whop) {
+		info.pid = whop->pid;
+		__builtin_memcpy(&info.name, whop->name, sizeof(info.name));
+	}
+
+	valp = bpf_map_lookup_or_try_init(&counts, &info, &zero);
+
+	if (valp) {
+		/* save stats */
+		valp->us += delta_us;
+		valp->bytes += startp->data_len;
+		valp->io++;
+	}
+
+	bpf_map_delete_elem(&start, &req);
+	bpf_map_delete_elem(&whobyreq, &req);
+
+	return 0;
+}
+
+char LICENSE[] SEC("license") = "GPL";
--- a/libbpf-tools/biotop.c
+++ b/libbpf-tools/biotop.c
@@ -1,0 +1,451 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+
+/*
+ * biotop Trace block I/O by process.
+ * Copyright (c) 2022 Francis Laniel <flaniel@linux.microsoft.com>
+ *
+ * Based on biotop(8) from BCC by Brendan Gregg.
+ * 03-Mar-2022   Francis Laniel   Created this.
+ */
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+#include <argp.h>
+#include <errno.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+#include <unistd.h>
+
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+#include "biotop.h"
+#include "biotop.skel.h"
+#include "trace_helpers.h"
+
+#define warn(...) fprintf(stderr, __VA_ARGS__)
+#define OUTPUT_ROWS_LIMIT 10240
+
+enum SORT {
+	ALL,
+	IO,
+	BYTES,
+	TIME,
+};
+
+struct disk {
+	int major;
+	int minor;
+	char name[256];
+};
+
+struct vector {
+	size_t nr;
+	size_t capacity;
+	void **elems;
+};
+
+int grow_vector(struct vector *vector) {
+	if (vector->nr >= vector->capacity) {
+		void **reallocated;
+
+		if (!vector->capacity)
+			vector->capacity = 1;
+		else
+			vector->capacity *= 2;
+
+		reallocated = reallocarray(vector->elems, vector->capacity, sizeof(*vector->elems));
+		if (!reallocated)
+			return -1;
+
+		vector->elems = reallocated;
+	}
+
+	return 0;
+}
+
+void free_vector(struct vector vector) {
+	for (size_t i = 0; i < vector.nr; i++)
+		if (vector.elems[i] != NULL)
+			free(vector.elems[i]);
+	free(vector.elems);
+}
+
+struct vector disks = {};
+
+static volatile sig_atomic_t exiting = 0;
+
+static bool clear_screen = true;
+static int output_rows = 20;
+static int sort_by = ALL;
+static int interval = 1;
+static int count = 99999999;
+static bool verbose = false;
+
+const char *argp_program_version = "biotop 0.1";
+const char *argp_program_bug_address =
+	"https://github.com/iovisor/bcc/tree/master/libbpf-tools";
+const char argp_program_doc[] =
+"Trace file reads/writes by process.\n"
+"\n"
+"USAGE: biotop [-h] [interval] [count]\n"
+"\n"
+"EXAMPLES:\n"
+"    biotop            # file I/O top, refresh every 1s\n"
+"    biotop 5 10       # 5s summaries, 10 times\n";
+
+static const struct argp_option opts[] = {
+	{ "noclear", 'C', NULL, 0, "Don't clear the screen" },
+	{ "sort", 's', "SORT", 0, "Sort columns, default all [all, io, bytes, time]" },
+	{ "rows", 'r', "ROWS", 0, "Maximum rows to print, default 20" },
+	{ "verbose", 'v', NULL, 0, "Verbose debug output" },
+	{ NULL, 'h', NULL, OPTION_HIDDEN, "Show the full help" },
+	{},
+};
+
+static error_t parse_arg(int key, char *arg, struct argp_state *state)
+{
+	long rows;
+	static int pos_args;
+
+	switch (key) {
+	case 'C':
+		clear_screen = false;
+		break;
+	case 's':
+		if (!strcmp(arg, "all")) {
+			sort_by = ALL;
+		} else if (!strcmp(arg, "io")) {
+			sort_by = IO;
+		} else if (!strcmp(arg, "bytes")) {
+			sort_by = BYTES;
+		} else if (!strcmp(arg, "time")) {
+			sort_by = TIME;
+		} else {
+			warn("invalid sort method: %s\n", arg);
+			argp_usage(state);
+		}
+		break;
+	case 'r':
+		errno = 0;
+		rows = strtol(arg, NULL, 10);
+		if (errno || rows <= 0) {
+			warn("invalid rows: %s\n", arg);
+			argp_usage(state);
+		}
+		output_rows = rows;
+		if (output_rows > OUTPUT_ROWS_LIMIT)
+			output_rows = OUTPUT_ROWS_LIMIT;
+		break;
+	case 'v':
+		verbose = true;
+		break;
+	case 'h':
+		argp_state_help(state, stderr, ARGP_HELP_STD_HELP);
+		break;
+	case ARGP_KEY_ARG:
+		errno = 0;
+		if (pos_args == 0) {
+			interval = strtol(arg, NULL, 10);
+			if (errno || interval <= 0) {
+				warn("invalid interval\n");
+				argp_usage(state);
+			}
+		} else if (pos_args == 1) {
+			count = strtol(arg, NULL, 10);
+			if (errno || count <= 0) {
+				warn("invalid count\n");
+				argp_usage(state);
+			}
+		} else {
+			warn("unrecognized positional argument: %s\n", arg);
+			argp_usage(state);
+		}
+		pos_args++;
+		break;
+	default:
+		return ARGP_ERR_UNKNOWN;
+	}
+	return 0;
+}
+
+static int libbpf_print_fn(enum libbpf_print_level level, const char *format, va_list args)
+{
+	if (level == LIBBPF_DEBUG && !verbose)
+		return 0;
+	return vfprintf(stderr, format, args);
+}
+
+static void sig_int(int signo)
+{
+	exiting = 1;
+}
+
+struct data_t {
+	struct info_t key;
+	struct val_t value;
+};
+
+static int sort_column(const void *obj1, const void *obj2)
+{
+	struct data_t *d1 = (struct data_t *) obj1;
+	struct data_t *d2 = (struct data_t *) obj2;
+
+	struct val_t *s1 = &d1->value;
+	struct val_t *s2 = &d2->value;
+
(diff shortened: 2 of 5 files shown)


libbpf-tools/biotop.c
```diff
@@ -0,0 +1,375 @@
...
+{
+	exiting = 1;
+}
+
+static int sort_column(const void *obj1, const void *obj2)
+{
+	struct val_t *s1 = (struct val_t *) obj1;
+	struct val_t *s2 = (struct val_t *) obj2;
+
+	if (sort_by == IO) {
```
chenhengqi, line 156: Drop the surrounding `{}`.

chenhengqi:
> Hi.
> 
> I converted `biotop` from BCC to CO-RE. I tried to stick with the BCC behavior.
> 
> I just have two questions:
> 
> 1. I only probed on some functions, is this OK? Because I think we cannot try to probe like it it is done here:
>    https://github.com/iovisor/bcc/blob/ca8240205e175e94cef3c70c3b5c5accbe51512f/tools/biotop.py#L178

Why not ? You just need to check which symbol exists and do bpf_program__set_attach_target.

> 2. I am not really satisfied with the way I handle disk name. Sadly it is a bit complicated to use a hash map in C. So if someone sees a way to improve this, feel free to share :D.
> 

This reminds me some discussions about introduce Rust-based libbpf-tools to BCC. @yonghong-song  WDYT ?

> Best regards.

libbpf-tools/biotop.bpf.c
```diff
@@ -0,0 +1,106 @@
...
+} whobyreq SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 10240);
+	__type(key, struct info_t);
+	__type(value, struct val_t);
+} counts SEC(".maps");
+
+SEC("kprobe/blk_account_io_start")
```
mfrw, line 31: In the current kernel , the function(s) are inline [0] and instead `__blk_account_io_start` seems to be a good choice here. Do we still want to add this instead of `__blk_account_io_start` ?

[0]: [inline hot paths of blk_account_io_*()](https://lore.kernel.org/all/b0662a636bd4cc7b4f84c9d0a41efa46a688ef13.1633781740.git.asml.silence@gmail.com/)

`be6bfe36db17 block: inline hot paths of blk_account_io_*()`

libbpf-tools/biotop.bpf.c
```diff
@@ -0,0 +1,106 @@
...
+	struct start_req_t start_req;
+
+	start_req.ts = bpf_ktime_get_ns();
+	start_req.data_len = BPF_CORE_READ(req, __data_len);
+
+	bpf_map_update_elem(&start, &req, &start_req, 0);
+	return 0;
+}
+
+SEC("kprobe/blk_account_io_done")
```
mfrw, line 58: Same as above :)

mfrw (commented):
A few clarification requests :)

eiffel-fl:
Hi.

Thank you for your reviews:

>Why not ? You just need to check which symbol exists and do bpf_program__set_attach_target.

I was not aware of this function, I will then add then add all the functions and probe some of them only if symbol exists.

> This reminds me some discussions about introduce Rust-based libbpf-tools to BCC. @yonghong-song WDYT ?

I am not familiar to Rust, thus converting the C code to Rust would take me a bit of time.
But I would be happy to learn it for this contribution.

Best regards.

libbpf-tools/biotop.bpf.c
eiffel-fl: Hi.

I saw some of these functions were inlined into the kernel but I was not aware of how to probe only some functions (see my comment below).
I will add them and try to probe them only if their symbols exists (to filter out inlined functions).


Best regards.

chenhengqi: You can use:

https://github.com/iovisor/bcc/blob/8f40d6f57a8d94e7aee74ce358572d34d31b4ed4/libbpf-tools/trace_helpers.h#L20

libbpf-tools/biotop.c
```diff
@@ -0,0 +1,392 @@
...
+		err = bpf_map_lookup_elem(fd, &keys[rows], &values[rows]);
+		if (err) {
+			warn("bpf_map_lookup_elem failed: %s\n", strerror(errno));
+			return err;
+		}
+		prev_key = &keys[rows];
+		rows++;
+	}
+
+	qsort(values, rows, sizeof(struct val_t), sort_column);
```
eiffel-fl, line 265: Fix this.

libbpf-tools/biotop.c
```diff
@@ -0,0 +1,392 @@
...
+	disk_number = 0;
+	zero = 0;
+	while (getline(&line, &zero, fp) != -1) {
+		struct disk disk;
+
+		if (sscanf(line, "%d %d %s", &disk.major, &disk.minor,
+			disk.name) != 3)
+			continue;
+
+		disks = realloc(disks, (disk_number + 1) * sizeof *disks);
```
mauriciovasquezbernal, line 187: nit: you can double the size (instead of increasing by 1) avoid doing a realloc() for each iteration of the loop.

mauriciovasquezbernal: Missing check that disk is not NULL after realloc(). (nit: you'll need to use a temp variable to avoid leaking disks is the reallocation fails).

libbpf-tools/biotop.c
```diff
@@ -0,0 +1,392 @@
...
+
+static void parse_disk_stat(void)
+{
+	FILE *fp;
+	char *line;
+	size_t zero;
+	size_t disk_number;
+
+	fp = fopen("/proc/diskstats", "r");
+	if (fp == NULL)
```
mauriciovasquezbernal, line 175: nit: `if (!fp)`

libbpf-tools/biotop.c
```diff
@@ -0,0 +1,392 @@
...
+	fclose(fp);
+
+	disk_number++;
+	disks = realloc(disks, disk_number * sizeof *disks);
+	disks[disk_number] = NULL;
+}
+
+static char *search_disk_name(int major, int minor)
+{
+	for(size_t i = 0; disks[i] != NULL; i++)
```
mauriciovasquezbernal, line 205: nit: add space after `for` -> `for (size...`

libbpf-tools/biotop.c
```diff
@@ -0,0 +1,392 @@
...
+	while (getline(&line, &zero, fp) != -1) {
+		struct disk disk;
+
+		if (sscanf(line, "%d %d %s", &disk.major, &disk.minor,
+			disk.name) != 3)
+			continue;
+
+		disks = realloc(disks, (disk_number + 1) * sizeof *disks);
+
+		disks[disk_number] = malloc(sizeof disk);
```
mauriciovasquezbernal, line 189: nit: missing check that malloc() doesnt' return NULL. Same in other places.

libbpf-tools/biotop.c
```diff
@@ -0,0 +1,392 @@
...
+		disk_number++;
+	}
+
+	free(line);
+	fclose(fp);
+
+	disk_number++;
+	disks = realloc(disks, disk_number * sizeof *disks);
+	disks[disk_number] = NULL;
+}
```
mauriciovasquezbernal, line 201: I think it'll be simpler if you use a linked list for `disks`, then you won't have to realloc() the array containing pointers and so on.

eiffel-fl:
This comments are targeting coding style.
Is there a rule regarding coding style for `*.c` files (I know `*.bpf.c` files should follow that of the kernel)?

libbpf-tools/biotop.c
eiffel-fl: Using a linked list would increase the time of looping through because I will not get access from the cache.
Anyway, I stated [I am not satisfied with this solution](https://github.com/iovisor/bcc/pull/3893#issuecomment-1059718669), so I would like to wait if we decide to switch from C to Rust for this control plane.

eiffel-fl:
I am still not happy with the proposed solution but I addressed the above comments.
I think, if bcc maintainers are OK with this, we can merge this version with a C control plane and I will come back later (in another PR) with a Rust one.

libbpf-tools/biotop.bpf.c
```diff
@@ -0,0 +1,108 @@
...
+
+	/* setup info_t key */
+	cmd_flags = BPF_CORE_READ(req, cmd_flags);
+
+	info.major = BPF_CORE_READ(req, rq_disk, major);
+	info.minor = BPF_CORE_READ(req, rq_disk, first_minor);
+	info.rwflag = !!((cmd_flags & REQ_OP_MASK) == REQ_OP_WRITE);
+
+	whop = bpf_map_lookup_elem(&whobyreq, &req);
+	if (!whop) {
```
mauriciovasquezbernal, line 83: When does this case happen? Should this program continue the processing even if we don't have the request on `whobyreq`?

libbpf-tools/biotop.bpf.c
```diff
@@ -0,0 +1,108 @@
...
+		valp = bpf_map_lookup_elem(&counts, &info);
+		if (!valp)
+			bpf_map_update_elem(&counts, &info, &zero, BPF_NOEXIST);
```
mauriciovasquezbernal, line 86: I think there is a missing lookup, otherwise `valp` will be NULL at L95 and the counters for this iteration won't be added. 

I see there is a `bpf_map_lookup_or_try_init()` helper defined in `maps.bpf.h`, it'll simplify this code.

mauriciovasquezbernal (commented):
Two comments regarding ebpf code.

libbpf-tools/biotop.bpf.c
eiffel-fl: Thank you for it, the code is clearer now!

libbpf-tools/biotop.bpf.c
eiffel-fl: I am not really sure if this can happen.
Nonetheless, this check was already present in [`biotop.py`](https://github.com/iovisor/bcc/blob/15ed997b82fa452c0ddd243fb26dff4e6e9c222f/tools/biotop.py#L171) and I tried to stick with the original code.

libbpf-tools/biotop.bpf.c
```diff
@@ -0,0 +1,104 @@
...
+	__type(key, struct info_t);
+	__type(value, struct val_t);
+} counts SEC(".maps");
+
+SEC("kprobe/blk_account_io_start")
+int BPF_KPROBE(blk_account_io_start, struct request *req)
+{
+	struct who_t who = {};
+
+	// cache PID and comm by-req
```
chenhengqi, line 37: Again, no C++ style comments.

libbpf-tools/biotop.bpf.c
```diff
@@ -0,0 +1,104 @@
...
+	if (!startp)
+		return 0;    /* missed tracing issue */
+
+	delta_us = (bpf_ktime_get_ns() - startp->ts) / 1000;
+
+	/* setup info_t key */
+	cmd_flags = BPF_CORE_READ(req, cmd_flags);
+
+	info.major = BPF_CORE_READ(req, rq_disk, major);
+	info.minor = BPF_CORE_READ(req, rq_disk, first_minor);
```
chenhengqi, line 80: I opened a PR to handle this `request->rq_disk` properly.

https://github.com/iovisor/bcc/pull/4088/commits/0651edec65fec303438af99de96388a3e911f3ec

You can reuse it after it landed.

libbpf-tools/biotop.bpf.c
```diff
@@ -0,0 +1,108 @@
...
+	bpf_map_update_elem(&start, &req, &start_req, 0);
+	return 0;
+}
+
+SEC("kprobe/blk_account_io_done")
+int BPF_KPROBE(blk_account_io_done, struct request *req, u64 now)
+{
+	struct val_t *valp, zero = {};
+	struct info_t info = {};
+
```
chenhengqi, line 65: Remove this empty line.

libbpf-tools/biotop.c
```diff
@@ -0,0 +1,419 @@
...
+	TIME,
+};
+
+struct disk {
+	int major;
+	int minor;
+	char name[256];
+};
+
+struct disk **disks = NULL;
```
chenhengqi, line 41: Could this be statically allocated ? Like:
```
struct disk disks[1024];
```

So that we can avoid realloc and free below.

libbpf-tools/biotop.c
```diff
@@ -0,0 +1,419 @@
...
+		err = print_stat(obj);
+		if (err)
+			goto cleanup;
+
+		count--;
+		if (exiting || !count)
+			goto cleanup;
+	}
+
+cleanup:
```
chenhengqi, line 414: `free(ksyms)` ?

libbpf-tools/biotop.c
```diff
@@ -0,0 +1,428 @@
...
+
+	ksyms = ksyms__load();
+	if (!ksyms) {
+		err = -ENOMEM;
+		warn("failed to load kallsyms\n");
+		goto cleanup;
+	}
+
+	if (ksyms__get_symbol(ksyms, "__blk_account_io_start"))
+		bpf_program__set_attach_target(obj->progs.blk_account_io_start, 0, "kprobe/__blk_account_io_start");
```
chenhengqi, line 377: IIRC, bpf_program__set_attach_target does NOT apply to kprobe.

libbpf-tools/biotop.c
```diff
@@ -0,0 +1,438 @@
...
+		bpf_program__attach_kprobe(obj->progs.blk_account_io_start, false, "__blk_account_io_start");
+	if (ksyms__get_symbol(ksyms, "blk_account_io_start"))
+		bpf_program__attach_kprobe(obj->progs.blk_account_io_start, false, "blk_account_io_start");
+
+	if (ksyms__get_symbol(ksyms, "__blk_account_io_done"))
+		bpf_program__attach_kprobe(obj->progs.blk_account_io_done, false, "__blk_account_io_done");
+	if (ksyms__get_symbol(ksyms, "blk_account_io_done"))
+		bpf_program__attach_kprobe(obj->progs.blk_account_io_done, false, "blk_account_io_done");
+
+	err = biotop_bpf__attach(obj);
```
chenhengqi, line 402: With `bpf_program__attach_kprobe`, you don't need `biotop_bpf__attach` any more.

You should check result of `bpf_program__attach_kprobe` and free the resources on exit.

For simplicity, you can have four dummy BPF programs, so the generated skeleton will have four bpf_link s, assign the attach results to these links and call `biotop_bpf__destroy` on exit, done.

chenhengqi: @eiffel-fl  This looks incorrect to me. Please fix.

chenhengqi:
```diff
diff --git a/libbpf-tools/biotop.bpf.c b/libbpf-tools/biotop.bpf.c
index 0d418cd5..226e32d3 100644
--- a/libbpf-tools/biotop.bpf.c
+++ b/libbpf-tools/biotop.bpf.c
@@ -9,7 +9,6 @@
 #include "maps.bpf.h"
 #include "core_fixes.bpf.h"
 
-
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
 	__uint(max_entries, 10240);
@@ -31,7 +30,7 @@ struct {
 	__type(value, struct val_t);
 } counts SEC(".maps");
 
-SEC("kprobe/blk_account_io_start")
+SEC("kprobe")
 int BPF_KPROBE(blk_account_io_start, struct request *req)
 {
 	struct who_t who = {};
@@ -57,7 +56,7 @@ int BPF_KPROBE(blk_mq_start_request, struct request *req)
 	return 0;
 }
 
-SEC("kprobe/blk_account_io_done")
+SEC("kprobe")
 int BPF_KPROBE(blk_account_io_done, struct request *req, u64 now)
 {
 	struct val_t *valp, zero = {};
diff --git a/libbpf-tools/biotop.c b/libbpf-tools/biotop.c
index 98f2fffd..2dc64658 100644
--- a/libbpf-tools/biotop.c
+++ b/libbpf-tools/biotop.c
@@ -7,6 +7,9 @@
  * Based on biotop(8) from BCC by Brendan Gregg.
  * 03-Mar-2022   Francis Laniel   Created this.
  */
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
 #include <argp.h>
 #include <errno.h>
 #include <signal.h>
@@ -70,7 +73,7 @@ void free_vector(struct vector vector) {
 	free(vector.elems);
 }
 
-struct vector disks = { 0 };
+struct vector disks = {};
 
 static volatile sig_atomic_t exiting = 0;
 
@@ -281,7 +284,7 @@ static int print_stat(struct biotop_bpf *obj)
 			printf("%8s loadavg: %s\n", ts, buf);
 		fclose(f);
 	}
-	printf("%-6s %-16s %1s %-3s %-3s %-8s %5s %7s %6s\n",
+	printf("%-7s %-16s %1s %-3s %-3s %-8s %5s %7s %6s\n",
 	       "PID", "COMM", "D", "MAJ", "MIN", "DISK", "I/O", "Kbytes", "AVGms");
 
 	while (1) {
@@ -319,7 +322,7 @@ static int print_stat(struct biotop_bpf *obj)
 		major = key->major;
 		minor = key->minor;
 
-		printf("%-6d %-16s %1s %-3d %-3d %-8s %5d %7lld %6.2f\n",
+		printf("%-7d %-16s %1s %-3d %-3d %-8s %5d %7lld %6.2f\n",
 		       key->pid, key->name, key->rwflag ? "W": "R",
 		       major, minor, search_disk_name(major, minor),
 		       value->io, value->bytes / 1024, avg_ms);
@@ -357,7 +360,6 @@ int main(int argc, char **argv)
 		.parser = parse_arg,
 		.doc = argp_program_doc,
 	};
-	struct bpf_link *probes[4] = { NULL };
 	struct biotop_bpf *obj;
 	struct ksyms *ksyms;
 	int err;
@@ -390,34 +392,22 @@ int main(int argc, char **argv)
 		goto cleanup;
 	}
 
-	if (ksyms__get_symbol(ksyms, "__blk_account_io_start")) {
-		probes[0] = bpf_program__attach_kprobe(obj->progs.blk_account_io_start, false, "__blk_account_io_start");
-		if (!probes[0]) {
-			warn("failed to load kprobe __blk_account_io_start\n");
-			goto cleanup;
-		}
-	}
-	if (ksyms__get_symbol(ksyms, "blk_account_io_start")) {
-		probes[1] = bpf_program__attach_kprobe(obj->progs.blk_account_io_start, false, "blk_account_io_start");
-		if (!probes[1]) {
-			warn("failed to load kprobe blk_account_io_start\n");
-			goto cleanup;
-		}
+	if (ksyms__get_symbol(ksyms, "__blk_account_io_start"))
+		obj->links.blk_account_io_start = bpf_program__attach_kprobe(obj->progs.blk_account_io_start, false, "__blk_account_io_start");
+	else
+		obj->links.blk_account_io_start = bpf_program__attach_kprobe(obj->progs.blk_account_io_start, false, "blk_account_io_start");
+	if (!obj->links.blk_account_io_start) {
+		warn("failed to attach blk_account_io_start\n");
+		goto cleanup;
 	}
 
-	if (ksyms__get_symbol(ksyms, "__blk_account_io_done")) {
-		probes[2] = bpf_program__attach_kprobe(obj->progs.blk_account_io_done, false, "__blk_account_io_done");
-		if (!probes[2]) {
-			warn("failed to load kprobe __blk_account_io_done\n");
-			goto cleanup;
-		}
-	}
-	if (ksyms__get_symbol(ksyms, "blk_account_io_done")) {
-		probes[3] = bpf_program__attach_kprobe(obj->progs.blk_account_io_done, false, "blk_account_io_done");
-		if (!probes[3]) {
-			warn("failed to load kprobe blk_account_io_done\n");
-			goto cleanup;
-		}
+	if (ksyms__get_symbol(ksyms, "__blk_account_io_done"))
+		obj->links.blk_account_io_done = bpf_program__attach_kprobe(obj->progs.blk_account_io_done, false, "__blk_account_io_done");
+	else
+		obj->links.blk_account_io_done = bpf_program__attach_kprobe(obj->progs.blk_account_io_done, false, "blk_account_io_done");
+	if (!obj->links.blk_account_io_done) {
+		warn("failed to attach blk_account_io_done\n");
+		goto cleanup;
 	}
 
 	err = biotop_bpf__attach(obj);
@@ -453,9 +443,6 @@ int main(int argc, char **argv)
 cleanup:
 	ksyms__free(ksyms);
 	free_vector(disks);
-	for (int i = 0; i < sizeof probes / sizeof probes[0]; i++)
-		if (probes[i])
-			bpf_link__destroy(probes[i]);
 	biotop_bpf__destroy(obj);
 
 	return err != 0;

```

chenhengqi: Try this small diff.

chenhengqi: Please paste the error here. By making `SEC("kprobe")`, we tell libbpf not to perform auto-attach, it should work with the latest libbpf submodule. Otherwise, calling `biotop_bpf__attach` will make them attach twice.

chenhengqi (approved):
LGTM.

PR #3894, opened 2022-03-03
libbpf-tool: Add tcptop

eiffel-fl:
Hi.

In this PR, I converted `tcptop` from BCC to CO-RE.

Best regards.

### DIFF
--- a/libbpf-tools/tcptop.bpf.c
+++ b/libbpf-tools/tcptop.bpf.c
@@ -1,0 +1,127 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2022 Francis Laniel <flaniel@linux.microsoft.com>
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_tracing.h>
+#include <bpf/bpf_endian.h>
+
+#include "tcptop.h"
+
+/* Taken from kernel include/linux/socket.h. */
+#define AF_INET		2	/* Internet IP Protocol 	*/
+#define AF_INET6	10	/* IP version 6			*/
+
+const volatile bool filter_cg = false;
+const volatile pid_t target_pid = -1;
+const volatile int target_family = -1;
+
+struct {
+	__uint(type, BPF_MAP_TYPE_CGROUP_ARRAY);
+	__type(key, u32);
+	__type(value, u32);
+	__uint(max_entries, 1);
+} cgroup_map SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 10240);
+	__type(key, struct ip_key_t);
+	__type(value, struct traffic_t);
+} ip_map SEC(".maps");
+
+static int probe_ip(bool receiving, struct sock *sk, size_t size)
+{
+	struct ip_key_t ip_key = {};
+	struct traffic_t *trafficp;
+	u16 family;
+	u32 pid;
+
+	if (filter_cg && !bpf_current_task_under_cgroup(&cgroup_map, 0))
+		return 0;
+
+	pid = bpf_get_current_pid_tgid() >> 32;
+	if (target_pid != -1 && target_pid != pid)
+		return 0;
+
+	family = BPF_CORE_READ(sk, __sk_common.skc_family);
+	if (target_family != -1 && target_family != family)
+		return 0;
+
+	/* drop */
+	if (family != AF_INET && family != AF_INET6)
+		return 0;
+
+	ip_key.pid = pid;
+	bpf_get_current_comm(&ip_key.name, sizeof(ip_key.name));
+	ip_key.lport = BPF_CORE_READ(sk, __sk_common.skc_num);
+	ip_key.dport = bpf_ntohs(BPF_CORE_READ(sk, __sk_common.skc_dport));
+	ip_key.family = family;
+
+	if (family == AF_INET) {
+		bpf_probe_read_kernel(&ip_key.saddr,
+				      sizeof(sk->__sk_common.skc_rcv_saddr),
+				      &sk->__sk_common.skc_rcv_saddr);
+		bpf_probe_read_kernel(&ip_key.daddr,
+				      sizeof(sk->__sk_common.skc_daddr),
+				      &sk->__sk_common.skc_daddr);
+	} else {
+		/*
+		 * family == AF_INET6,
+		 * we already checked above family is correct.
+		 */
+		bpf_probe_read_kernel(&ip_key.saddr,
+				      sizeof(sk->__sk_common.skc_v6_rcv_saddr.in6_u.u6_addr32),
+				      &sk->__sk_common.skc_v6_rcv_saddr.in6_u.u6_addr32);
+		bpf_probe_read_kernel(&ip_key.daddr,
+				      sizeof(sk->__sk_common.skc_v6_daddr.in6_u.u6_addr32),
+				      &sk->__sk_common.skc_v6_daddr.in6_u.u6_addr32);
+	}
+
+	trafficp = bpf_map_lookup_elem(&ip_map, &ip_key);
+	if (!trafficp) {
+		struct traffic_t zero;
+
+		if (receiving) {
+			zero.sent = 0;
+			zero.received = size;
+		} else {
+			zero.sent = size;
+			zero.received = 0;
+		}
+
+		bpf_map_update_elem(&ip_map, &ip_key, &zero, BPF_NOEXIST);
+	} else {
+		if (receiving)
+			trafficp->received += size;
+		else
+			trafficp->sent += size;
+
+		bpf_map_update_elem(&ip_map, &ip_key, trafficp, BPF_EXIST);
+	}
+
+	return 0;
+}
+
+SEC("kprobe/tcp_sendmsg")
+int BPF_KPROBE(tcp_sendmsg, struct sock *sk, struct msghdr *msg, size_t size)
+{
+	return probe_ip(false, sk, size);
+}
+
+/*
+ * tcp_recvmsg() would be obvious to trace, but is less suitable because:
+ * - we'd need to trace both entry and return, to have both sock and size
+ * - misses tcp_read_sock() traffic
+ * we'd much prefer tracepoints once they are available.
+ */
+SEC("kprobe/tcp_cleanup_rbuf")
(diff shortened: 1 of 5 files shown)


libbpf-tools/tcptop.h
```diff
@@ -0,0 +1,30 @@
...
+#ifndef __TCPTOP_H
+#define __TCPTOP_H
+
+#ifndef AF_INET
+#define AF_INET		2	/* Internet IP Protocol 	*/
+#endif /* AF_INET */
+
+#ifndef AF_INET6
+#define AF_INET6	10	/* IP version 6			*/
+#endif /* AF_INET6 */
```
chenhengqi, line 11: Move these to bpf.c ?

libbpf-tools/tcptop.h
```diff
@@ -0,0 +1,30 @@
...
+#define TASK_COMM_LEN 16
+
+struct ip_key_t {
+	unsigned __int128 saddr;
+	unsigned __int128 daddr;
+	__u32 pid;
+	char name[TASK_COMM_LEN];
+	__u16 lport;
+	__u16 dport;
+	__u32 __pad__;
```
chenhengqi, line 22: Why this padding here ?

libbpf-tools/tcptop.h
eiffel-fl: Done, thank you for spotting it.

libbpf-tools/tcptop.bpf.c
```diff
@@ -0,0 +1,139 @@
...
+// Copyright (c) 2022 Francis Laniel <flaniel@linux.microsoft.com>
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_tracing.h>
+#include <bpf/bpf_endian.h>
+
+#include "tcptop.h"
+
+#ifndef AF_INET
```
chenhengqi, line 11: No need these`#ifndef`.

libbpf-tools/tcptop.bpf.c
```diff
@@ -0,0 +1,139 @@
...
+	__type(key, struct ip_key_t);
+	__type(value, struct traffic_t);
+} ipv4_bytes SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 10240);
+	__type(key, struct ip_key_t);
+	__type(value, struct traffic_t);
+} ipv6_bytes SEC(".maps");
```
chenhengqi, line 42: Why do you need two separate maps for the same purpose ? I think just one is fine.

libbpf-tools/tcptop.bpf.c
```diff
@@ -0,0 +1,139 @@
...
+
+static int probe_ip(bool receiving, struct sock *sk, size_t size)
+{
+	struct bpf_map *ip_map;
+	struct ip_key_t ip_key;
+
+	if (filter_cg && !bpf_current_task_under_cgroup(&cgroup_map, 0))
+		return 0;
+
+	u32 pid = bpf_get_current_pid_tgid() >> 32;
```
chenhengqi, line 52: Let's stick to kernel code style, declare variables first. :)

libbpf-tools/tcptop.bpf.c
```diff
@@ -0,0 +1,139 @@
...
+		return 0;
+
+	u16 family = BPF_CORE_READ(sk, __sk_common.skc_family);
+	if (target_family != -1 && target_family != family)
+		return 0;
+
+	if (family != AF_INET && family != AF_INET6)
+		return 0;
+
+	__builtin_memset(&ip_key, 0, sizeof(ip_key));
```
chenhengqi, line 63: You just need `struct ip_key_t ip_key = {};`.

libbpf-tools/tcptop.c
```diff
@@ -0,0 +1,420 @@
...
+{
+	exiting = 1;
+}
+
+static int sort_column(const void *obj1, const void *obj2)
+{
+	struct traffic_t *s1 = (struct traffic_t *)obj1;
+	struct traffic_t *s2 = (struct traffic_t *)obj2;
+
+	if (sort_by == SENT) {
```
chenhengqi, line 187: Remove the surrounding `{}`.

libbpf-tools/tcptop.c
```diff
@@ -0,0 +1,420 @@
...
+	int err;
+
+	err = argp_parse(&argp, argc, argv, 0, NULL, NULL);
+	if (err)
+		return err;
+
+	libbpf_set_strict_mode(LIBBPF_STRICT_ALL);
+	libbpf_set_print(libbpf_print_fn);
+
+	if (ipv4_only && ipv6_only) {
```
chenhengqi, line 339: Check this on `argp_parse` ?

libbpf-tools/tcptop.bpf.c
eiffel-fl: This was one my thinking during the writing of this CO-RE version.
I have several possibilities:
1. If I do not use two separates maps, then I will need to add a version field to differentiate.
And, in userspace, when I will loop the map, I will either need to loop it twices (one for IPv4 and for IPv6) or keep an array of pointer for IPv6 (I will need to realloc this array each time I discover a new IPv4).
This is due to the fact where IPv4 and IPv6 are printed separately (and IPv4 are printed first).
2. With two separate maps, I can first loop the one IPv4 then that of IPv6.

The second way is simpler for userspace code but I understand it will take more resources.
Should I go with the first way?

chenhengqi: I don't have strong opinion about this.

But why the order matters ?

eiffel-fl: > But why the order matters ?

I tried to keep the same output as the python version of this tool which first prints IPv4.

chenhengqi: I think you can do a sort and then print in userspace.

eiffel-fl: Oh right! I did not think about the `sort`!
I will test and come back to you!

eiffel-fl: Thinking to it, I will sadly need to do two sorts:
1. One for IP family.
2. Another to print the top value first.

Nonetheless, your comment made me realizing this code is wrong:
https://github.com/iovisor/bcc/pull/3894/commits/762af004a4c3a7e3aa1a5e4d236af1d4753db3ae#diff-f3977b11bd3764a632191fb362db857b4523541c43d4294091a78fa86346ead7R262
I will tackle this with a new structure.

chenhengqi: I think you can sort by IP family and values in one sort, no ?

Like:
```c
obj1.af == obj2.af ? obj1.val < obj2.val : obj1.af < obj2.af;
```

eiffel-fl: Good idea!
Sorry, I was thinking to do two calls to `qsort()` while I can do like your proposal.
I updated the code to add this!

libbpf-tools/tcptop.bpf.c
```diff
@@ -0,0 +1,123 @@
...
+	bpf_get_current_comm(&ip_key.name, sizeof(ip_key.name));
+	ip_key.lport = BPF_CORE_READ(sk, __sk_common.skc_num);
+	ip_key.dport = bpf_ntohs(BPF_CORE_READ(sk, __sk_common.skc_dport));
+	ip_key.family = family;
+
+	if (family == AF_INET) {
+		bpf_probe_read_kernel(&ip_key.saddr, sizeof(ip_key.saddr),
+				      &sk->__sk_common.skc_rcv_saddr);
+		bpf_probe_read_kernel(&ip_key.daddr, sizeof(ip_key.daddr),
+				      &sk->__sk_common.skc_daddr);
```
mauriciovasquezbernal, line 65: Why is it reading `sizeof(ip_key.daddr)` if this is IPv4 and hence should only read 4 bytes? @chenhengqi you did the same for bindnsoop, is it correct?

libbpf-tools/tcptop.c
```diff
@@ -0,0 +1,420 @@
...
+const char *argp_program_version = "tcptop 0.1";
+const char *argp_program_bug_address =
+	"https://github.com/iovisor/bcc/tree/master/libbpf-tools";
+const char argp_program_doc[] =
+"Trace sending and received operation over IP.\n"
+"\n"
+"USAGE: tcptop [-h] [-p PID] [interval] [count]\n"
+"\n"
+"EXAMPLES:\n"
+"    tcptop            # file I/O top, refresh every 1s\n"
```
mauriciovasquezbernal, line 65: file I/O top -> tcp top?

libbpf-tools/tcptop.bpf.c
chenhengqi: Yep, reading 4 bytes is 4 better.

eiffel-fl: Good catch, bad copy/pasting.

libbpf-tools/tcptop.c
```diff
@@ -0,0 +1,420 @@
...
+		char saddr[INET6_ADDRSTRLEN];
+		char daddr[INET6_ADDRSTRLEN];
+		char saddr_port[size];
+		char daddr_port[size];
+
+		inet_ntop(key->family, &key->saddr, saddr, INET6_ADDRSTRLEN);
+		inet_ntop(key->family, &key->daddr, daddr, INET6_ADDRSTRLEN);
+
+		snprintf(saddr_port, size, "%s:%d", saddr, key->lport);
+		snprintf(daddr_port, size, "%s:%d", daddr, key->dport);
```
chenhengqi, line 292: This is not necessary.

libbpf-tools/tcptop.c
```diff
@@ -0,0 +1,420 @@
...
+
+		printf(key->family == AF_INET
+					? "%-6d %-12.12s %-21s %-21s %6ld %6ld\n"
+					: "%-6d %-12.12s %-51s %-51s %6ld %6ld\n",
+					key->pid, key->name, saddr_port, daddr_port,
+					value->received / 1024, value->sent / 1024);
```
chenhengqi, line 298: FYI, we have a elegant way to handle this:

https://github.com/iovisor/bcc/blob/c54336ef2b26c7a244604efb8741a5c5b3f0b968/libbpf-tools/tcplife.c#L116-L136

chenhengqi:
@eiffel-fl Please rebase and push again. Thanks.

PR #3899, opened 2022-03-07
add tcp congestion status duration statistic tool

jackygam2001:
add tcp congestion contol status duration statistic tool,  and it can be used to evalute the networking and congestion algorithm performace.


iamkafai:
The use case makes sense to me.  The implementation is not straightforward though because of the kernel missing a tracepoint in tcp_set_ca_state(), so it has to resort to kprobe and kretprobe on all functions calling tcp_set_ca_state(). 

".set_state" of "struct tcp_congestion_ops" is implemented for some popular cc like cubic, bbr, and dctcp.  That will be a more stable kprobe points.

If you are interested, a tracepoint in tcp_set_ca_state() may be a good add to the kernel considering it is not the hot path.   This future kernel work does not need to derail the current bcc pull request.

jackygam2001:
> The use case makes sense to me. The implementation is not straightforward though because of the kernel missing a tracepoint in tcp_set_ca_state(), so it has to resort to kprobe and kretprobe on all functions calling tcp_set_ca_state().
> 
> ".set_state" of "struct tcp_congestion_ops" is implemented for some popular cc like cubic, bbr, and dctcp. That will be a more stable kprobe points.
> 
> If you are interested, a tracepoint in tcp_set_ca_state() may be a good add to the kernel considering it is not the hot path. This future kernel work does not need to derail the current bcc pull request.

yes, currently the upstream kernel only has inline function tcp_set_ca_state(),  which is optimized by compiler when building and can not be probed.  So in my implementation i have to kprobe and kretprobe on all functions calling it.

brendangregg:
Thanks; Several comments:

- Name is too long; should be tcpcong or something.

- It has 20 k{ret}probes plus a lot of struct digging, which would make it by far the most brittle tool in all of bcc. Most kprobe tools use between 1 and 6 kprobes. The prior biggest is zfsdist/zfsslower, which has 16, but only attaches to 6 at a time depending on the software version, and it also does no struct digging. (While BTF can help solve struct members moving, it can't solve logic changes.) Also, looking at these tcp functions:

```
grep attach_k.*probe tcpcongestdura.py 
b.attach_kprobe(event="tcp_try_keep_open",
b.attach_kretprobe(event="tcp_try_keep_open",
b.attach_kprobe(event="tcp_enter_cwr", fn_name="trace_entry_tcp_enter_cwr")
b.attach_kretprobe(event="tcp_enter_cwr",
b.attach_kprobe(event="tcp_process_tlp_ack",
b.attach_kretprobe(event="tcp_process_tlp_ack",
b.attach_kprobe(event="tcp_enter_recovery",
b.attach_kretprobe(event="tcp_enter_recovery",
b.attach_kprobe(event="tcp_enter_loss", fn_name="trace_entry_tcp_enter_loss")
b.attach_kretprobe(event="tcp_enter_loss",
b.attach_kprobe(event="tcp_simple_retransmit",
b.attach_kretprobe(event="tcp_simple_retransmit",
b.attach_kprobe(event="tcp_try_undo_recovery",
b.attach_kretprobe(event="tcp_try_undo_recovery",
b.attach_kprobe(event="tcp_try_undo_loss",
b.attach_kretprobe(event="tcp_try_undo_loss",
b.attach_kprobe(event="tcp_fastretrans_alert",
b.attach_kretprobe(event="tcp_fastretrans_alert",
b.attach_kprobe(event="tcp_disconnect", fn_name="trace_entry_tcp_enter_open")
b.attach_kretprobe(event="tcp_disconnect",
```

I guess this is only really tracing 10 functions as it's doing both kprobe and kretprobe.

Things like tcp_disconnect() I'm not too worried about as it's in tcp_prot, and isn't likely going to change, and tcp_enter_loss() is in tcp.h. But others like tcp_try_undo_loss() feel like gritty internals. I'm ok with one or two such functions in a tool, but if it ends up being several it's a pain to maintain.

So my two questions are: How many of these are unstable internal functions, and does it need to trace them all or can you pop up a stack frame and trace some more-stable parent function to do the same job? I'm not saying this is a blocker, I just want to be thoughtful about the future maintenance.

- How does it compare to tcp_probe and other TCP congestion analysis tools? (Other people will no doubt ask.)

- Output columns look misaligned by one.

- Default output is > 80 chars, but fortunately only by 5. This does make it hard (and sometimes impossible) to include the output in things like published books and printed articles, but I don't see an easy fix.

- Columns are in "us", but the numbers are so big it's confusing. Would it make sense to make them "ms" by default? Have a switch to output "us".

- I think the opening example in the examples file is missing a sentence or more to explain why the output matters. Which of these columns is evidence of a problem, such that clients are blocked waiting? Help a newcomer interpret what they are seeing.

- These numbers are great, but how would I double check that they are right? Have you already got some fields in nstat/netstat or something to compare them to, at least as a sanity check? (They won't have latency totals, but they may have counts.) Just looking for a way to sanity check the output.

jackygam2001:
@brendangregg ,
thanks for your comments!   
For comment 2  your are right.  I have checked kernel code, most of cc status change functions are called by 5 functions(tcp_fastretrans_alert, tcp_enter_cwr, tcp_enter_loss, tcp_enter_recovery, tcp_process_tlp_ack), so i will update the code in next pull request.    I think @iamkafai 's advice is very good, i will try to add tracepoint for tcp_set_ca_state to upstream kernel.
For comment 3  i think the tcp_probe is for trace the connection's congestion window update, but this tool is for trace the congestion control status change and duration.  you know the cwnd's update may certainly not cause status's change, so  they are two different tools.
For comment 8 i don't use the nstat/netstat to validate the tool,  as those tools has only summary for all tcp sockets.  Actually i use packetdrill script which was provided by google on github to check the output of my tool.
For other comments, i will update the code in next pull request.    thanks very much!

tools/tcpcong.py
```diff
@@ -0,0 +1,484 @@
...
+        FILTER_LPORT
+        FILTER_DPORT
+        ipv6_val.ipv6_key.dport = ntohs(dport);
+        ipv6_val.cong_state = cong_status.cong_stat + 1;
+        start_ipv6.update(&key, &ipv6_val);
+    }
+    return 0;
+}
+
+int ret_state_update_func(struct pt_regs *ctx, struct sock *sk)
```
yonghong-song, line 197: This won't work, for return probe, we don't have 'struct sock *sk' available. To get 'struct sock *sk', you need to save it in the hash table with key tid or task ptr, and retrieve it in the kretprobe.

Do we have potential for nested bpf program execution here, one is in process context and another in softirq context? In such nested case, the bpf program in softirq context will not get executed due to kernel internal checking, so we could miss some events.

Alternatively, you can kfunc/kretfunc which does give you function parameters in kretfunc. It is only available recent kernels though.

tools/tcpcong.py
```diff
@@ -0,0 +1,484 @@
...
+                STORE
+            }
+        }
+        start_ipv6.delete(&key);
+    }
+    return 0;
+}
+
+
+"""
```
yonghong-song, line 283: the above two empty lines can be removed.

yonghong-song (commented):
A lot of inlining COULD happen for the probe functions used in this tool. So I strongly suggest you to implement the tracepoint as Martin suggested. The tool can still proceed. But with tracepoint in the kernel, we can have an alternative better implementation for future kernels.

jackygam2001:
@yonghong-song ,
thanks for your comments,  yes, i will prepare the tracepoint for tcp_set_ca_state patch to upstream kernel.  And  will update my code for other comments in next commit.  thanks very much!

tools/tcpcong.py
```diff
@@ -0,0 +1,484 @@
...
+# @lint-avoid-python-3-compatibility-imports
+#
+# tcpcong  Measure tcp congestion control status duration.
+#           For Linux, uses BCC, eBPF.
+#
+# USAGE: tcpcong [-h] [-T] [-L] [-R] [-m] [-d] [interval] [outputs]
+#
+#
+#
+# Copyright (c) ping gan.
```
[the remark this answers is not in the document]
jackygam2001: yes

tools/tcpcong.py
jackygam2001: okay, should store the paremeter 'struct sock *sk' in a hash table.
yes, your worry is reasonable, if there is  nested execution between softirq context  and process context ,  we should need another  context  type(softirq, process) to record , then can cover this case.  does bcc have a function to get the context type? if so, please tell me and i will add code for this case.  thanks very much!

tools/tcpcong.py
```diff
@@ -0,0 +1,484 @@
...
+        bpf_probe_read_kernel(&ipv6_val.ipv6_key.daddr,
+            sizeof(ipv6_val.ipv6_key.daddr),
+            &sk->__sk_common.skc_v6_daddr.in6_u.u6_addr32);
+        ipv6_val.ipv6_key.lport = sk->__sk_common.skc_num;
+        dport = sk->__sk_common.skc_dport;
+        dport = ntohs(dport);
+        lport = ipv6_val.ipv6_key.lport;
+        FILTER_LPORT
+        FILTER_DPORT
+        ipv6_val.ipv6_key.dport = ntohs(dport);
```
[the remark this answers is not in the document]
jackygam2001: for ipv6, ntohs has been done before fileter, so the assignment statement need not do again, will update in next commit.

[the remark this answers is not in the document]
jackygam2001: for ipv6, ntohs has been done before fileter, so the assignment statement need not do again, will update in next commit.

tests/python/test_tools_smoke.py
```diff
@@ -358,6 +358,10 @@ def test_tcpdrop(self):
     def test_tcptop(self):
         self.run_with_duration("tcptop.py 1 1")
 
+    @skipUnless(kernel_version_ge(4,4), "requires kernel >= 4.4")
```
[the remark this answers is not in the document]
jackygam2001: will remove this skip

tools/tcpcong.py
yonghong-song: `context type(softirq, process) ` won't work since this is kprobe and in the kernel (kernel/trace/bpf_trace.c) we have
```
        if (unlikely(__this_cpu_inc_return(bpf_prog_active) != 1)) {
                /*
                 * since some bpf program is already running on this cpu,
                 * don't call into another bpf program (same or different)
                 * and don't send kprobe event into ring-buffer,
                 * so return zero here
                 */
                ret = 0;
                goto out;
        }
```
Basically nested kprobe programs not allowed.

kfunc approach is better and it allows nested bpf programs as long as they are different programs. But kfunc will not work for earlier kernels.

Maybe putting a warning/caveat in the beginning of tools is okay.

jackygam2001: @yonghong-song , thanks for your comments. okay, i will add the kfunc/kretfunc for the tools when it's available.

tools/tcpcong_example.txt
```diff
@@ -0,0 +1,463 @@
...
+  -L LOCALPORT, --localport LOCALPORT
+                        trace local ports only
+  -R REMOTEPORT, --remoteport REMOTEPORT
+                        trace the dest ports only
+  -T, --timestamp       include timestamp on output
+  -d, --dist            show distributions as histograms
+  -u, --microseconds    output in microseconds
+
+examples:
+    ./tcpcong                 #show tcp congestion status duration
```
yonghong-song, line 458: Let us put a space between '\#' and 'show'. The same for other examples below.

tools/tcpcong_example.txt
```diff
@@ -0,0 +1,463 @@
...
+192.168.219.3/39112   192.168.219.1/19225   500792  846    9297    489264  0      1850
+192.168.219.3/39120   192.168.219.1/19225   518700  94     749     480171  0      1967
+
+07:39:46
+LAddrPort            RAddrPort             Open_us Dod_us Rcov_us Cwr_us  Los_us Chgs 
+192.168.219.3/39070   192.168.219.1/19225   587340  5324   37035   370066  0      1602
+192.168.219.3/39098   192.168.219.1/19225   532986  5630   31624   345336  0      1319
+192.168.219.3/39112   192.168.219.1/19225   481936  1129   6244    510235  0      1909
+192.168.219.3/39120   192.168.219.1/19225   507196  316    6200    485737  0      1957
+
```
yonghong-song, line 235: Could you have at least one ipv6 examples so we could see what it looks like?

yonghong-song (commented):
Most of okay. A few minor comments.

man/man8/tcpcong.8
```diff
@@ -0,0 +1,135 @@
+.TH tcpcong 8  "2022-01-27" "USER COMMANDS"
+.SH NAME
+tcpcong \- Measure tcp congestion state duration. Uses Linux eBPF/bcc.
+.SH SYNOPSIS
+.B tcpcong [\-h] [\-T] [\-L] [\-R] [\-u] [\-d] [interval] [outputs]
+.SH DESCRIPTION
+this tool measures tcp sockets congestion control status duration, and
+prints a summary one state by one along with total state change number.
```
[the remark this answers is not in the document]
jackygam2001: okay, will update it in next commit

tools/tcpcong.py
```diff
@@ -0,0 +1,564 @@
...
+    if len(rports) == 2:
+        start_rport = int(rports[0])
+        end_rport = int(rports[1])
+    else:
+        start_rport = int(rports[0])
+        end_rport = int(rports[0])
+if start_rport > end_rport:
+    tmp = start_rport
+    start_rport = end_rport
+    end_rport = tmp
```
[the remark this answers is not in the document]
jackygam2001: yes, it works well, when using '30000-40000' or '40000-30000', both parameters are okay.

tools/tcpcong.py
```diff
@@ -0,0 +1,564 @@
...
+#include <bcc/proto.h>
+#include <net/tcp.h>
+#include <net/inet_connection_sock.h>
+
+typedef struct ipv4_flow_key {
+    u32 saddr;
+    u32 daddr;
+    u16 lport;
+    u16 dport;
+}ipv4_flow_key_t;
```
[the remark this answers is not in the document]
jackygam2001: okay, will update

tools/tcpcong.py
```diff
@@ -0,0 +1,564 @@
...
+BPF_HASH(ipv4_stat, ipv4_flow_key_t, data_val_t);
+BPF_HASH(ipv6_stat, ipv6_flow_key_t, data_val_t);
+
+HIST_TABLE
+
+static int entry_state_update_func(struct sock *sk)
+{
+    u16 dport = 0, lport = 0;
+    u32 tid = bpf_get_current_pid_tgid();
+    u32 pid = (bpf_get_current_pid_tgid() >> 32);
```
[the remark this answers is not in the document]
jackygam2001: will remove it

tools/tcpcong_example.txt
jackygam2001: okay, i will add it in example file.

tools/tcpcong_example.txt
jackygam2001: okay, will update it

tools/tcpcong_example.txt
```diff
@@ -0,0 +1,463 @@
+Demonstrations of tcpcongestdura, the Linux eBPF/bcc version.
```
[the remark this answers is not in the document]
jackygam2001: sorry, miss one, i will update it.

tools/tcpcong.py
```diff
@@ -0,0 +1,564 @@
...
+    unsigned __int128 daddr;
+    u16 lport;
+    u16 dport;
+}ipv6_flow_key_t;
+
+typedef struct process_key {
+    char comm[TASK_COMM_LEN];
+    u32  pid;
+    u32  tid;
+}process_key_t;
```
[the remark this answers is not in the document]
jackygam2001: hi @yonghong-song ,
i have tried in SMP platform, when there are  multiple processes whose tid is 0 and name are such that swapper/0, swapper/1, swapper/2...., then only the tid can not work well. so i add the comm in the key struct, then it work well.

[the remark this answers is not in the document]
yonghong-song: > hi @yonghong-song , i have tried in SMP platform, when there are multiple processes whose tid is 0 and name are such that swapper/0, swapper/1, swapper/2...., then only the tid can not work well. so i add the comm in the key struct, then it work well.

okay, tid + comm should be fine. The tcpconnect is in process context so tid is never 0, hence doesn't have issue. Alternatively, using task pointer (bpf_get_current_task()) should be also okay.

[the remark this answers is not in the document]
jackygam2001: removed pid in process_key struct.

tools/tcpcong.py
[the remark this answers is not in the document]
jackygam2001: done in commit [b59d237](https://github.com/iovisor/bcc/pull/3899/commits/b59d237f062a72aa5b978e287f9ff404ff5ca5d3)

tools/tcpcong.py
```diff
@@ -0,0 +1,559 @@
...
+    return 0;
+}
+
+static int ret_state_update_func(struct sock *sk)
+{
+    u64 ts, ts1;
+    u16 family, last_cong_state;
+    u16 dport = 0, lport = 0;
+    u32 tid = bpf_get_current_pid_tgid();
+    process_key_t key = {0};
```
chenhengqi, line 201: If the kernel function is really triggered under process context, you can use tid as key, no need to obtain task comm here.

But I guess the function will be triggered under softirq context, I think you can simply use the struct sock itself as key.

chenhengqi: I played with this tool locally, all tid are zero :)

jackygam2001: hi @chenhengqi ,
Do you mean all the state update functions are executed in softirq context?  for your suggestion if  kfunc/kretfunc are not available,  we should use kprobe/kretprobe, right? but when using kprobe/kretprobe we should also use process_key to get the parameter struct sock in kretprobe function. so why should we add extra key type(struct sock) for the hash tables?

chenhengqi: > I think you can simply use the struct sock itself as key.

Sorry for the noise.

chenhengqi: Some (if not all) of these functions are run under softirq context, you will get tid=0 and comm=swapper.
Use these values as key is not reliable. You may consider using processor id as key.

jackygam2001: hi ,
yes, some functions are executed in softirq context.  but if the tid is 0, the pid is also 0.  i think tid + comm is enough as the comm is different for these swapper process.

yonghong-song: Right the swapper comm typically is swapper/<cpu_id>, so we should be fine here.

jackygam2001: hi @yonghong-song ,
yes,  thanks for your comment!

PR #3903, opened 2022-03-10
biolatency, biolatpcts, biosnoop, biotop: Build fix for v5.17+

htejun:
During 5.17 dev cycle, the kernel dropped request->rq_disk. It can now be
accessed through request->q->disk. Fix the python ones in tools/. There are
more usages in other places which need to be fixed too.

Signed-off-by: Tejun Heo <tj@kernel.org>


yonghong-song:
There are a couple of test failures on ubuntu 18.04 and fedora. fedora failure is known and ubuntu 18.04 failure is related to mdflush.py and sslsniff.py, which needs further check. This patch looks good to me.

PR #3915, opened 2022-03-21
tools: Unify PID column width (at most 7 chars)

xingfeng2510:
The "kernel.pid_max" sysctl is now bumped to 4194304 by default, so 6 chars not enough for PID colume. Some tools use %6d, some use %7d. This patch try to unify PID column width by %7d (at most 7 chars).

```
# sysctl -a | grep pid_max
kernel.pid_max = 4194304
```

### DIFF
--- a/tools/bashreadline.py
+++ b/tools/bashreadline.py
@@ -68,11 +68,11 @@ b = BPF(text=bpf_text)
 b.attach_uretprobe(name=name, sym="readline", fn_name="printret")
 
 # header
-print("%-9s %-6s %s" % ("TIME", "PID", "COMMAND"))
+print("%-9s %-7s %s" % ("TIME", "PID", "COMMAND"))
 
 def print_event(cpu, data, size):
     event = b["events"].event(data)
-    print("%-9s %-6d %s" % (strftime("%H:%M:%S"), event.pid,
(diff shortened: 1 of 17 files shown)


chenhengqi (commented):
LGTM

yonghong-song:
The patch looks good  to me. But test_bcc_fedora test failed with something like below,
```
29: timeout: failed to run command ‘/bcc/tools/argdist.py’: No such file or directory
29: .timeout: failed to run command ‘/bcc/tools/bashreadline.py’: No such file or directory
29: Ftimeout: failed to run command ‘/bcc/tools/bindsnoop.py’: No such file or directory
29: Ftimeout: failed to run command ‘/bcc/tools/biolatency.py’: No such file or directory
29: Ftimeout: failed to run command ‘/bcc/tools/biosnoop.py’: No such file or directory
29: Ftimeout: failed to run command ‘/bcc/tools/biotop.py’: No such file or directory
29: Ftimeout: failed to run command ‘/bcc/tools/biotop.py’: No such file or directory
29: Ftimeout: failed to run command ‘/bcc/tools/bpflist.py’: No such file or directory
...
```
The failure is not related to this patch. @davemarchevsky could you take a look at the failure before we merge this patch?

davemarchevsky:
Test failure is definitely unrelated, am investigating. PR LGTM, don't think it needs to be held up on account of the test.

Issue #3781, opened 2022-01-01
Broken libbpf tools on arm64


chenhengqi:
Open this issue to keep track of broken libbpf tools on arm64.

* [ ] biosnoop
* [ ] biostacks
* [ ] cachestat
* [ ] execsnoop
* [ ] ext4slower/ext4dist
* [ ] numamove
* [ ] readahead
* [ ] solisten
* [ ] statsnoop
* [ ] tcpconnlat
* [ ] vfsstat

ChrisMcMStone:
Hi @chenhengqi, could this be related to my issue here: https://github.com/iovisor/bcc/issues/3783?

chenhengqi:
I don't think so, will need some time to investigate.

PR #3917, opened 2022-03-22
libbpf-tools: improve fentry_exists check

cobrien7:
On architectures that lack support for fentry programs, tools should
fall back to using kprobes even if the kernel version is new enough to
include BPF_TRACE_FENTRY, but fentry_exists() cannot currently detect
this case.

Add an additional check verifying that attaching a BPF_TRACE_FENTRY
program can actually succeed.

Signed-off-by: Connor O'Brien <connoro@google.com>

### DIFF
--- a/libbpf-tools/trace_helpers.c
+++ b/libbpf-tools/trace_helpers.c
@@ -15,6 +15,7 @@
 #include <fcntl.h>
 #include <sys/resource.h>
 #include <time.h>
+#include <bpf/bpf.h>
 #include <bpf/btf.h>
 #include <bpf/libbpf.h>
 #include <limits.h>
@@ -990,14 +991,33 @@ bool is_kernel_module(const char *name)
 	return found;
 }
 
-bool fentry_exists(const char *name, const char *mod)
+static bool fentry_try_attach(int id)
+{
+	struct bpf_insn insns[] = { { .code = BPF_JMP | BPF_EXIT } };
+	LIBBPF_OPTS(bpf_prog_load_opts, opts);
+	int prog_fd, attach_fd;
+
+	opts.expected_attach_type = BPF_TRACE_FENTRY;
+	opts.attach_btf_id = id,
+
+	prog_fd = bpf_prog_load(BPF_PROG_TYPE_TRACING, "test", NULL, insns, 1, &opts);
+	if (prog_fd < 0)
+		return false;
+
+	attach_fd = bpf_raw_tracepoint_open(NULL, prog_fd);
+	if (attach_fd >= 0)
+		close(attach_fd);
+
+	close(prog_fd);
+	return attach_fd >= 0;
+}
+
+bool fentry_can_attach(const char *name, const char *mod)
 {
 	const char sysfs_vmlinux[] = "/sys/kernel/btf/vmlinux";
 	struct btf *base, *btf = NULL;
-	const struct btf_type *type;
-	const struct btf_enum *e;
 	char sysfs_mod[80];
-	int id = -1, i, err;
+	int id = -1, err;
 
 	base = btf__parse(sysfs_vmlinux, NULL);
 	if (!base) {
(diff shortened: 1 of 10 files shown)


cobrien7:
Are these failed checks a regression caused by this change? It looks like the failing tests involve the python tools, so I'm not sure I see how this would affect them.

libbpf-tools/trace_helpers.c
```diff
@@ -990,6 +991,28 @@ bool is_kernel_module(const char *name)
 	return found;
 }
 
+static bool fentry_can_attach(int id)
+{
+	struct bpf_insn insns[] = { { .code = BPF_JMP | BPF_EXIT } };
+	struct bpf_prog_load_opts opts;
```
chenhengqi, line 997: Declare `opts` using `LIBBPF_OPTS`, it takes care of assigning the sz field and ensure all fields are zero-initialized.

libbpf-tools/trace_helpers.c
```diff
@@ -1042,7 +1065,7 @@ bool fentry_exists(const char *name, const char *mod)
 err_out:
 	btf__free(btf);
 	btf__free(base);
-	return id > 0;
+	return id > 0 ? fentry_can_attach(id) : false;
```
chenhengqi, line 1068: I think this commit will also close #3781 :)

davemarchevsky:
> Are these failed checks a regression caused by this change? It looks like the failing tests involve the python tools, so I'm not sure I see how this would affect them.

Agreed, test failures look unrelated. Will look into them separately.

davemarchevsky:
Looks like these are the tools which are affected by this change:

```
cachestat.c
fsdist.c
fsslower.c
klockstat.c
solisten.c
tcprtt.c
tcpsynbl.c
vfsstat.c
```

Just listing for posterity

libbpf-tools/trace_helpers.c
```diff
@@ -990,6 +991,28 @@ bool is_kernel_module(const char *name)
...
+
+	attach_fd = bpf_raw_tracepoint_open(NULL, prog_fd);
+	if (attach_fd >= 0)
+		close(attach_fd);
+
+	close(prog_fd);
+	return attach_fd >= 0;
+}
+
 bool fentry_exists(const char *name, const char *mod)
```
davemarchevsky, line 1016: Since this change tries to load and attach a program with `BPF_TRACE_FENTRY` attach type, can we remove the searching done elsewhere in this function for `bpf_attach_type` enum and its `BPF_TRACE_FENTRY` member?

libbpf-tools/trace_helpers.c
davemarchevsky, line 1068: nit: This can be `id > 0 && fentry_can_attach(id)`

libbpf-tools/trace_helpers.c
davemarchevsky: Also, since the helper function you're adding is doing more significant / invasive checking, perhaps renaming `fentry_exists` to `fentry_can_attach` would make what's being done more obvious to callers, especially if the "test for existence of enum member" logic can be removed.

Could then call the function you're adding something like `fentry_try_attach` or merge the two functions - I don't feel strongly about this though.

davemarchevsky (changes requested):

libbpf-tools/trace_helpers.c
cobrien7: It should help, but I don't think all of the tools listed there use fentry_exists to fall back to kprobes

libbpf-tools/trace_helpers.c
cobrien7: Done

libbpf-tools/trace_helpers.c
cobrien7: Done - removed the search and renamed to `fentry_can_attach` and `fentry_try_attach`.

PR #3919, opened 2022-03-22
bcc: add method to close file descriptors

terceiro:
I have a long running process that attaches BPF programs to containers,
and I want those programs to stay attached after the BPF object goes out
of scope. After some time, the open file descriptors pile up and my
process starts failing since it cannot open new files.

This close() method allows me to create a BPF object, attach my
function, then close the associated file descriptors without detaching
the program.

Signed-off-by: Antonio Terceiro <antonio.terceiro@linaro.org>

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -1736,6 +1736,20 @@ class BPF(object):
     def donothing(self):
         """the do nothing exit handler"""
 
+
+    def close(self):
+        """close(self)
+
+        Closes all associated files descriptors. Attached BPF programs are not
+        detached.
+        """
+        for name, fn in list(self.funcs.items()):
+            os.close(fn.fd)
+            del self.funcs[name]
+        if self.module:
+            lib.bpf_module_destroy(self.module)
+            self.module = None
+
     def cleanup(self):
         # Clean up opened probes
         for k, v in list(self.kprobe_fds.items()):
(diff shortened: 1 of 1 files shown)


chenhengqi:
> I want those programs to stay attached after the BPF object goes out
of scope.

IIUC, `BPF::__exit__` is called when BPF object goes out of scope. It will trigger a cleanup call and finally detach functions.

> After some time, the open file descriptors pile up

Is this related to BPF object itself ?

davemarchevsky:
@chenhengqi 

> IIUC, BPF::\_\_exit\_\_ is called when BPF object goes out of scope. It will trigger a cleanup call and finally detach functions.

I think this is only the case when the Python API's `BPF` object is used as a context manager, not a more general 'goes out of scope' or 'is GC'd'.

@terceiro To confirm my understanding, does the below sound correct to you?

* You have a program using the Python API's `BPF` object, the program runs ~forever and loads/attaches some BPF progs to many containerized processes which are shorter-lived than the daemon
* You want the BPF progs to remain attached until the containerized process exits, at which point the cgroup/socket/whatever resource will be cleaned up. If there are no other open references to program(s) attached to these resources, [they'll be cleaned up](https://facebookmicrosites.github.io/bpf/blog/2018/08/31/object-lifetime.html) at this time as well.
* You don't anticipate needing to interact with these progs from your long-running program after attaching, want to avoid going over open fd limit, and want to rely on the kernel's refcounting to clean up the progs you're attaching

davemarchevsky (approved):

PR #3922, opened 2022-03-24
tools: Flush stdout explicitly in event loop

xingfeng2510:
When redirect output of hardirqs/softirqs to a file, line buffering (_IOLBF) switched to full buffering (_IOFBF), it may take dozens of seconds to see last output. hardirqs/softirqs write relatively less chars than other tools (such as biosnoop/opensnoop) each second, buffer filling causes poor real-time. For example:

```
/usr/share/bcc/tools/hardirqs -CT 1 > hardirqs.out
```
```
tail -f hardirqs.out

First Output
<after 30s>
Second Output
<after 34s>
Third Output
...
```

As ttysnoop/memleak/biolatpcts, this patch use stdout.flush() to flush stdout explicitly in event loop.

xxxsnoop tools (compactsnoop/drsnoop/mountsnoop/syncsnoop) with low frequency usually, also flush them.

### DIFF
--- a/tools/compactsnoop.py
+++ b/tools/compactsnoop.py
@@ -18,6 +18,7 @@ from bcc import BPF
 import argparse
 import platform
 from datetime import datetime, timedelta
+import sys
 
 # arguments
 examples = """examples:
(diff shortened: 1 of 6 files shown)


chenhengqi:
The simplest solution to this problem would be using `python -u`.

yonghong-song:
> The simplest solution to this problem would be using `python -u`.

I think this patch is okay. Most people probably won't use `python -u` as I didn't realize and use it as well.

reinierpost:
I had the same issue with `tcplife` (#4329).

The issue is not that the tool buffers its output, but that I didn't expect it to do so in the first place.

PR #3923, opened 2022-03-24
tools/softirqs: Add event counting support

xingfeng2510:
Sometimes, I just want counts of soft irq events instead of timing. hardirqs has already added event counting support in #1460. This patch try to follow it.

```
# ./softirqs.py -C
Tracing soft irq events... Hit Ctrl-C to end.
^C
SOFTIRQ          TOTAL_count
block                      5
tasklet                    6
net_rx                   402
sched                   5251
rcu                     5748
timer                   9530
```

### DIFF
--- a/man/man8/softirqs.8
+++ b/man/man8/softirqs.8
@@ -2,7 +2,7 @@
 .SH NAME
 softirqs \- Measure soft IRQ (soft interrupt) event time. Uses Linux eBPF/bcc.
 .SH SYNOPSIS
-.B softirqs [\-h] [\-T] [\-N] [\-d] [interval] [count]
+.B softirqs [\-h] [\-T] [\-N] [\-C] [\-d] [\-c CPU] [interval] [count]
 .SH DESCRIPTION
 This summarizes the time spent servicing soft IRQs (soft interrupts), and can
 show this time as either totals or histogram distributions. A system-wide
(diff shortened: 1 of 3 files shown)


man/man8/softirqs.8
```diff
@@ -2,7 +2,7 @@
 .SH NAME
 softirqs \- Measure soft IRQ (soft interrupt) event time. Uses Linux eBPF/bcc.
 .SH SYNOPSIS
-.B softirqs [\-h] [\-T] [\-N] [\-d] [interval] [count]
+.B softirqs [\-h] [\-T] [\-N] [\-C] [\-d] [interval] [outputs]
```
yonghong-song, line 5: I would like we keep 'count' here so it is consistent with other tools.

tools/softirqs.py
```diff
@@ -22,6 +23,7 @@
 # arguments
 examples = """examples:
     ./softirqs            # sum soft irq event time
+    ./softirqs -C         # show soft irq event counts
```
yonghong-song, line 26: to avoid referring to count, just say 'show the number of soft irq events'. The '-C' option is okay to me.

man/man8/softirqs.8
xingfeng2510: done

tools/softirqs.py
xingfeng2510: Done. btw, args.count is conflicted with positional argument 'count', use args.events instead.

xingfeng2510:
@yonghong-song Thanks for your suggestions. I've fixed, please take a look again.

PR #3691, opened 2021-11-08
libbpf-tools: Add mdflush

chenhengqi:
Add libbpf-tool mdflush, this is a replacement for BCC mdflush.
This is part of the solution to #3650.

PR #3926, opened 2022-03-30
libbpf-tools: Fix dropped request->rq_disk for kernel 5.17+

xingfeng2510:
Since 5.17, The Linux kernel has dropped request->rq_disk, use request->q->disk instead. tools/bioxxx had been fixed not long ago, this patch try to fix libbpf-tools/bioxxx.

### DIFF
--- a/libbpf-tools/biolatency.bpf.c
+++ b/libbpf-tools/biolatency.bpf.c
@@ -19,6 +19,10 @@ const volatile bool targ_ms = false;
 const volatile bool filter_dev = false;
 const volatile __u32 targ_dev = 0;
 
+struct request_queue___x {
+	struct gendisk *disk;
+} __attribute__((preserve_access_index));
+
 struct {
 	__uint(type, BPF_MAP_TYPE_CGROUP_ARRAY);
 	__type(key, u32);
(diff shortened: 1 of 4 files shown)


libbpf-tools/biolatency.bpf.c
```diff
@@ -53,9 +57,15 @@ int trace_rq_start(struct request *rq, int issue)
 	u64 ts = bpf_ktime_get_ns();
 
 	if (filter_dev) {
-		struct gendisk *disk = BPF_CORE_READ(rq, rq_disk);
+		struct request_queue *queue = BPF_CORE_READ(rq, q);
+		struct gendisk *disk;
 		u32 dev;
 
+		if (LINUX_KERNEL_VERSION >= KERNEL_VERSION(5, 17, 0))
```
chenhengqi, line 64: We can fix this without kconfig version check. Please refer to #3691 .

xingfeng2510: done, thanks.

xingfeng2510:
sorry, I should add __attribute__((preserve_access_index)) later.

xingfeng2510:

libbpf-tools/biolatency.bpf.c
```diff
@@ -53,9 +57,15 @@ int trace_rq_start(struct request *rq, int issue)
 	u64 ts = bpf_ktime_get_ns();
 
 	if (filter_dev) {
-		struct gendisk *disk = BPF_CORE_READ(rq, rq_disk);
+		struct request_queue___x *q = (void *)BPF_CORE_READ(rq, q);
```
yonghong-song, line 60: I don't think this works for old kernel. See #3691 for an example.

yonghong-song, line 60: I don't think this works for old kernel. See #3691 for an example.

xingfeng2510: I referred to #3691 before creating this pull request. req->q first appears since kernel 3.0, and remains unchanged till now. So BPF_CORE_READ(rq, q) will always return a non-NULL pointer.

```
struct request {
    struct request_queue *q;
    ...
};
```

There is another way, define request__x, but it's not simple as current way: 

```
struct request___x {
	struct gendisk *disk;
} __attribute__((preserve_access_index));

```

yonghong-song: Okay, thanks for explanation. looks like req->q is always valid for supported kernels so we are fine here.

PR #3936, opened 2022-04-07
tools/mdflush: Fix CI build failure

xingfeng2510:
CI failed due to the following error (in mdflush.py, bio->bi_disk not found). I guess CI runtime env (such as kernel version) had changed in recent days.

```
error: no member named 'bi_disk' in 'struct bio'
     struct gendisk *bi_disk = bio->bi_disk;
                              ~~~  ^
1 error generated.
   File "/bcc/tools/mdflush.py", line 54, in <module>
```

Before kernel 4.14, stay the same, just use bio->bi_bdev->bd_disk.
Since kernel 4.14, macro bio_dev begin to appear, which refers to bio->bi_disk. 
Since kernel 5.12, macro bio_dev has changed to bio->bi_bdev->bd_disk (bio->bi_bdev changed back). So, just use '#ifdef bio_dev' could not make the right choice under higher kernel version.

This patch try to fix this error. Please take a look. @davemarchevsky @yonghong-song

### DIFF
--- a/tools/mdflush.py
+++ b/tools/mdflush.py
@@ -15,8 +15,8 @@ from __future__ import print_function
 from bcc import BPF
 from time import strftime
 
-# load BPF program
-b = BPF(text="""
+# define BPF program
+bpf_text="""
 #include <uapi/linux/ptrace.h>
 #include <linux/sched.h>
 #include <linux/genhd.h>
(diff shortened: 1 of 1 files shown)


xingfeng2510:
@davemarchevsky My pleasure, It's very kind of you.

PR #3938, opened 2022-04-09
tools/tcpconnect: Support IPv6 DNS

### LINKED ISSUE #3933: tcpconnect DNS correlation doesn't work with IPv6 DNS
`tcpconnect -d` reports `No DNS Query` when DNS queries are sent over IPv6 instead of IPv4.

Can be reproduced by setting `2606:4700:4700::64` in resolv.conf.

It is able to correlate AAAA queries.

chenhengqi:
Add IPv6 DNS support for tcpconnect.
Fixes #3933.

### DIFF
--- a/tools/tcpconnect.py
+++ b/tools/tcpconnect.py
@@ -322,6 +322,31 @@ delete_and_return:
     return 0;
 }
 
+#include <uapi/linux/udp.h>
+
+int trace_udpv6_recvmsg(struct pt_regs *ctx)
+{
+    struct sk_buff *skb = (struct sk_buff *)PT_REGS_PARM2(ctx);
+    struct udphdr *hdr = (void*)skb->head + skb->transport_header;
+    struct dns_data_t *event;
+    int zero = 0;
+    void *data;
+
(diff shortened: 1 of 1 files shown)


tools/tcpconnect.py
```diff
@@ -322,6 +322,31 @@
...
+        return 0;
+
+    /* skip UDP header */
+    data = skb->data + 8;
+
+    event = dns_data.lookup(&zero);
+    if (!event)
+        return 0;
+
+    bpf_probe_read(event->pkt, sizeof(event->pkt), data);
```
yonghong-song, line 346: Could we have some garbage here? Is it possible we check the udp packet size and only copy min(udp_size, sizeof(event->pkt))?

chenhengqi: Do you mean something like this ?

1st record(8 bytes): XXXX XXXX
2nd record(6 bytes): YYYY YYXX    <- the last two bytes is from last record

It could be.

But DNS resource record format has a field specifying the record length and dnslib parses this correctly (unless the record is truncated).

yonghong-song: Okay. that is fine then since the data is parsed correctly in user space library.

PR #3939, opened 2022-04-12
Add rwsem support to klockstat

namhyung:
Hello,

This PR adds rwsem support to libbpf-tools/klockstat.  For kernel lock debugging, I found it useful to track rwsem as well as mutex.  This doesn't distinguish read and write to rwsem in the logic but users can see it in the caller info.

Example:
```
$ sudo klockstat -n 3 -s 3
Tracing mutex/rwsem lock events...  Hit Ctrl-C to end
^C

                       mutex_lock+0x5      1422      791      15628      1125030
                  do_epoll_wait+0x1ce
            do_epoll_pwait.part.0+0xc
                              Max PID 2441, COMM murdockd

                       mutex_lock+0x5      1699      465      14471       790264
                  do_epoll_wait+0x1ce
            __x64_sys_epoll_wait+0x60
                              Max PID 1864873, COMM svw_NetThreadEv

                       mutex_lock+0x5      1559      144       2106       224575
                      pipe_write+0x47
                 new_sync_write+0x199
                              Max PID 2441, COMM murdockd

                down_read_trylock+0x5   1888876        5    9109738      9444384
                   trylock_super+0x16
           __writeback_inodes_wb+0x8d
                              Max PID 3744442, COMM kworker/u145:6

                       mutex_lock+0x5   2350695       15    2593128     35260434
        bpf_tracing_prog_attach+0x34b
        bpf_raw_tracepoint_open+0x1c4
                              Max PID 3748530, COMM klockstat

                       mutex_lock+0x5   2510333        8    2590270     20082666
       bpf_trampoline_link_prog+0x18b
        bpf_tracing_prog_attach+0x1e8
                              Max PID 3748530, COMM klockstat
Exiting trace of mutex/rwsem locks
```

### DIFF
--- a/libbpf-tools/klockstat.c
+++ b/libbpf-tools/klockstat.c
@@ -64,12 +64,12 @@ static struct prog_env {
 	.iterations = 99999999,
 };
 
-const char *argp_program_version = "klockstat 0.1";
+const char *argp_program_version = "klockstat 0.2";
 const char *argp_program_bug_address =
 	"https://github.com/iovisor/bcc/tree/master/libbpf-tools";
 static const char args_doc[] = "FUNCTION";
 static const char program_doc[] =
-"Trace mutex lock acquisition and hold times, in nsec\n"
+"Trace mutex/sem lock acquisition and hold times, in nsec\n"
 "\n"
 "Usage: klockstat [-hRTv] [-p PID] [-t TID] [-c FUNC] [-L LOCK] [-n NR_LOCKS]\n"
 "                 [-s NR_STACKS] [-S SORT] [-d DURATION] [-i INTERVAL]\n"
@@ -549,6 +549,26 @@ int main(int argc, char **argv)
 					       "mutex_lock_killable_nested");
 	}
 
+	if (fentry_can_attach("down_read_nested", NULL)) {
+		bpf_program__set_attach_target(obj->progs.down_read, 0,
+					       "down_read_nested");
+		bpf_program__set_attach_target(obj->progs.down_read_exit, 0,
+					       "down_read_nested");
+		bpf_program__set_attach_target(obj->progs.down_read_killable, 0,
+					       "down_read_killable_nested");
+		bpf_program__set_attach_target(obj->progs.down_read_killable_exit, 0,
+					       "down_read_killable_nested");
+
+		bpf_program__set_attach_target(obj->progs.down_write, 0,
+					       "down_write_nested");
+		bpf_program__set_attach_target(obj->progs.down_write_exit, 0,
+					       "down_write_nested");
+		bpf_program__set_attach_target(obj->progs.down_write_killable, 0,
+					       "down_write_killable_nested");
+		bpf_program__set_attach_target(obj->progs.down_write_killable_exit, 0,
+					       "down_write_killable_nested");
+	}
+
 	err = klockstat_bpf__load(obj);
 	if (err) {
 		warn("failed to load BPF object\n");
@@ -560,7 +580,7 @@ int main(int argc, char **argv)
 		goto cleanup;
 	}
 
-	printf("Tracing mutex lock events...  Hit Ctrl-C to end\n");
+	printf("Tracing mutex/sem lock events...  Hit Ctrl-C to end\n");
 
 	for (i = 0; i < env.iterations && !exiting; i++) {
 		sleep(env.interval);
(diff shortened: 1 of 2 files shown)


namhyung:
The check failure on Fedora was not related to this change.

```
Step 5/7 : RUN ln -s $(readlink /usr/bin/python3) /usr/bin/python
 ---> Running in 039641fabbda
ln: failed to create symbolic link '/usr/bin/python': File exists
The command '/bin/sh -c ln -s $(readlink /usr/bin/python3) /usr/bin/python' returned a non-zero code: 1
```

chenhengqi (commented):
LGTM

libbpf-tools/klockstat.c
```diff
@@ -64,12 +64,12 @@ static struct prog_env {
...
 };
 
-const char *argp_program_version = "klockstat 0.1";
+const char *argp_program_version = "klockstat 0.2";
 const char *argp_program_bug_address =
 	"https://github.com/iovisor/bcc/tree/master/libbpf-tools";
 static const char args_doc[] = "FUNCTION";
 static const char program_doc[] =
-"Trace mutex lock acquisition and hold times, in nsec\n"
+"Trace mutex/rwsem lock acquisition and hold times, in nsec\n"
```
brho, line 72: optional: i imagine the next thing someone will add is generic semaphore support, so maybe rename "rwsem" with "sem'?  otherwise we're hoping whoever adds full semaphores will remember to change this text.

libbpf-tools/klockstat.c
```diff
@@ -549,6 +549,26 @@ int main(int argc, char **argv)
 					       "mutex_lock_killable_nested");
 	}
 
+	if (fentry_can_attach("down_read_nested", NULL)) {
```
brho, line 552: this nested stuff is a little worrisome.  i see you're doing the same thing as the code did for mutex_lock, but i'm not familiar with that change either.  are we tracing actual nested locks, or is that some kprobe-naming issue?

klockstat.bpf has an assumption that a given task won't try to grab the same lock more than once at a time.  if it does, you'll get corrupted results.

that's not a reason to block the rwsem change, but it is a concern with any use of nested locks (mutex or otherwise).

brho (commented):
Can you add a comment around L152 about how with semaphores, there are multiple threads "holding the lock" at the same time?

When you originally told me about adding sem support, I was a little worried about assumptions of having a single lockholder.  I guess that was from older versions of the code in my head, since the existing code can handle multiple lock holders (or what *appeared* to be multiple lockholders, based on the comment in account()).

libbpf-tools/klockstat.c
namhyung: Sure, I'll add the comment and rename to "sem".

libbpf-tools/klockstat.c
namhyung: I believe even for the nested lock paths, they will see different lock instances at each invocation.  It's about the lock classes (in lockdep) rather than instances.

And yeah, it's more about the kprobe-naming issue.  When CONFIG_DEBUG_LOCK_ALLOC is defined, the regular `mutex_lock` will be renamed to `mutex_lock_nested`.  On the other hand, if the config option is not selected, `mutex_lock_nested` will be renamed to `mutex_lock`.

namhyung:
This time the failure is like below:

```
In file included from /bcc/src/cc/libbpf.c:57:
/bcc/src/cc/bcc_libbpf_inc.h:8:11: fatal error: libbpf/src/bpf.h: No such file or directory
    8 | # include "libbpf/src/bpf.h"
      |           ^~~~~~~~~~~~~~~~~~
compilation terminated.
```

brho (approved):
LGTM, thanks.

not sure about the build errors - maybe some thing with libbpf got updated?

one gotcha is that src/cc/libbpf is a git submodule, so you periodically need to update it.  try

```
git submodule update --init --recursive
```

yonghong-song:
I will merge this one.  Could you also port the change to bcc/tools/klockstat.py?

yonghong-song:
I will merge this one.  Could you also port the change to bcc/tools/klockstat.py?

namhyung:
Sure, I'll take a look.

brho:
On 2022-04-25 at 11:31 Namhyung Kim ***@***.***> wrote:
> I believe even for the nested lock paths, they will see different lock instances at each invocation.  It's about the lock classes (in lockdep) rather than instances.

great, thanks.  the name "nested" threw me off for a second, but it's
all __mutex_lock under the hood.

PR #3929, opened 2022-04-01
Support additional ino field seen on openSUSE 15.4 (to parse the prog_tag)

chwillso:
Without this fix (or one like it): one ends up getting this warning when any bpf program loads
```
WARNING: cannot get prog tag, ignore saving source with program tag 
prog tag mismatch 71d4822dd06c5b8e 1 
```
thus one can't get the source of the bpf programs output to /var/tmp/bcc

Seen on openSUSE 15.4 for bpf programs.

Other OSes cat /proc/self/fdinfo/*
```
pos:    0
flags:  02000002
mnt_id: 15
prog_type:      2
```
On openSUSE SP4 cat /proc/self/fdinfo/*
```
pos:    0
flags:  02000002
mnt_id: 15
ino:       11319
prog_type:      2
```
The additional field causes the prog_tag not to be retrieved properly.

PR #3943, opened 2022-04-13
adapted `bpf_prog_get_tag` to new kernel version

wineway:
refer to [patch](https://lore.kernel.org/linux-mm/20210701015449.VAmBDscYI%25akpm@linux-foundation.org/), `ino` was added to fdinfo, which make `bpf_prog_get_tag` get `prog_jited` and returned an incorrect value as prog tag


chenhengqi:
Some one already fix this in #3929 .

wineway:
> Someone already fix this in #3929 .

Sorry, I didn't notice it 😂

PR #3944, opened 2022-04-14
libbpf-tools/klockstat: Remove compilation warnings and consequent se…

ekyooo:
…gfault

There are compilation warnings and consequent segfault.
Relocate the definition of _GNU_SOURCE to the top of the header to fix it.

klockstat.c:404:12: warning: implicit declaration of function ‘reallocarray’; did you mean ‘realloc’? [-Wimplicit-function-declaration]
		stats = reallocarray(stats, stats_sz, sizeof(void *));
						^~~~~~~~~~~~
						realloc
klockstat.c:404:10: warning: assignment makes pointer from integer without a cast [-Wint-conversion]
		stats = reallocarray(stats, stats_sz, sizeof(void *));

Signed-off-by: Eunseon Lee <es.lee@lge.com>

### DIFF
--- a/libbpf-tools/klockstat.c
+++ b/libbpf-tools/klockstat.c
@@ -10,13 +10,13 @@
  *     when '-s' > 1 (otherwise it's cluttered).
  * - does not reset stats each interval by default. Can request with -R.
  */
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
 #include <argp.h>
 #include <errno.h>
 #include <signal.h>
 #include <stdio.h>
-#ifndef _GNU_SOURCE
-#define _GNU_SOURCE
-#endif
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>


chenhengqi:
Why does the order matter ? :)

ekyooo:
> Why does the order matter ? :)

reallocarray is included in stdlib.h if __USE_GNU is defined.
```
552 #ifdef __USE_GNU
558 extern void *reallocarray (void *__ptr, size_t __nmemb, size_t __size)
559      __THROW __attribute_warn_unused_result__;
560 #endif
```
__USE_GNU is defined in features.h if _GNU_SOURCE is defined.
```
375 #ifdef  _GNU_SOURCE
376 # define __USE_GNU      1
377 #endif
```
features.h is included when including <stdio.h> without _GNU_SOURCE before defining _GNU_SOURCE in klockstat.
And when include stdlib.h after defining _GNU_SOURCE then features.h will not be included again by header guard.
As a result, reallocarray is not provided in stdlib.h because __USE_GNU is not defined.
I'll add it to the commit message if you like.

(tested gcc version: 
gcc version 7.5.0 (Ubuntu/Linaro 7.5.0-3ubuntu1\~18.04)
gcc version 7.5.0 (Ubuntu 7.5.0-3ubuntu1~18.04))

chenhengqi:
Ah, that makes sense. Thanks for the detailed explanation.

chenhengqi (approved):
LGTM.

PR #3956, opened 2022-04-24
Fix segfault and warnings for LLVM 15

chenhengqi:
When BCC is compiled with LLVM 15, running BCC tools will cause segfault.
The core dump reveals that the segfault is triggered by the calls to
`getPointerElementType()`. This is because LLVM is moving towards
`Opaque Pointer` ([0]). This commit follows the migration guide to fix
the issue.

  [0]: https://llvm.org/docs/OpaquePointers.html

### DIFF
--- a/src/cc/bpf_module_rw_engine.cc
+++ b/src/cc/bpf_module_rw_engine.cc
@@ -49,7 +49,12 @@ void BPFModule::cleanup_rw_engine() {
 
 static LoadInst *createLoad(IRBuilder<> &B, Value *addr, bool isVolatile = false)
 {
-#if LLVM_MAJOR_VERSION >= 14
+#if LLVM_MAJOR_VERSION >= 15
+  if (isa<AllocaInst>(addr))
+    return B.CreateLoad(dyn_cast<AllocaInst>(addr)->getAllocatedType(), addr, isVolatile);
+  else
+    return B.CreateLoad(addr->getType(), addr, isVolatile);
+#elif LLVM_MAJOR_VERSION >= 14
   return B.CreateLoad(addr->getType()->getPointerElementType(), addr, isVolatile);
 #else
   return B.CreateLoad(addr, isVolatile);
@@ -58,7 +63,12 @@ static LoadInst *createLoad(IRBuilder<> &B, Value *addr, bool isVolatile = false
 
 static Value *createInBoundsGEP(IRBuilder<> &B, Value *ptr, ArrayRef<Value *>idxlist)
 {
-#if LLVM_MAJOR_VERSION >= 14
+#if LLVM_MAJOR_VERSION >= 15
+  if (isa<GlobalValue>(ptr))
+    return B.CreateInBoundsGEP(dyn_cast<GlobalValue>(ptr)->getValueType(), ptr, idxlist);
+  else
+    return B.CreateInBoundsGEP(ptr->getType(), ptr, idxlist);
+#elif LLVM_MAJOR_VERSION >= 14
   return B.CreateInBoundsGEP(ptr->getType()->getScalarType()->getPointerElementType(),
                              ptr, idxlist);
 #else
@@ -116,7 +126,7 @@ static void finish_sscanf(IRBuilder<> &B, vector<Value *> *args, string *fmt,
   B.SetInsertPoint(label_false);
   // s = &s[nread];
   B.CreateStore(
-      createInBoundsGEP(B, createLoad(B, sptr), createLoad(B, nread, true)), sptr);
+      createInBoundsGEP(B, createLoad(B, sptr), {createLoad(B, nread, true)}), sptr);
 
   args->resize(2);
   fmt->clear();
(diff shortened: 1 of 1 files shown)


yonghong-song:
@chenhengqi thanks for investigation. I tried the patch, looks it doesn't solve the issue,
```
$ sudo python3.6 ./profile.py
Sampling at 49 Hertz of all threads by user + kernel stack... Hit Ctrl-C to end.
python3.6: ../include/llvm/IR/Instructions.h:922: llvm::Type* llvm::checkGEPType(llvm::Type*): Assertion `Ty && "Invalid GetElementPtrInst indices for type!"' failed.
Aborted
```
I haven't looked at the issue yet. If you got time, could you take a look?

chenhengqi:
Will do.

yonghong-song:
I still have the same issues. The following is my llvm build command line,
```
    -DLLVM_ENABLE_PROJECTS="clang;lld" \
    -DLLVM_ENABLE_ASSERTIONS=ON \
    -DLLVM_ENABLE_ZLIB=OFF \

ninja && ninja install
```
I guess I enabled `-DLLVM_ENABLE_ASSERTIONS=ON` which caused the assertion.

chenhengqi:
Still work in progress.

Here is the doc I found related to these issues.
    https://llvm.org/docs/OpaquePointers.html

chenhengqi:
@yonghong-song Please take a look again.

src/cc/bpf_module_rw_engine.cc
```diff
@@ -50,7 +50,10 @@ void BPFModule::cleanup_rw_engine() {
 static LoadInst *createLoad(IRBuilder<> &B, Value *addr, bool isVolatile = false)
 {
 #if LLVM_MAJOR_VERSION >= 14
-  return B.CreateLoad(addr->getType()->getPointerElementType(), addr, isVolatile);
+  if (isa<AllocaInst>(addr))
+    return B.CreateLoad(dyn_cast<AllocaInst>(addr)->getAllocatedType(), addr, isVolatile);
+  else
+    return B.CreateLoad(addr->getType(), addr, isVolatile);
```
yonghong-song, line 56: I thought this patch is to fix llvm15 issue. Is llvm14 change needed as well?

chenhengqi: https://github.com/llvm/llvm-project/blob/llvmorg-14.0.0/llvm/include/llvm/IR/Type.h#L369-L373

`getPointerElementType` is deprecated since llvm14. And this change also applies to llvm14.

src/cc/bpf_module_rw_engine.cc
```diff
@@ -400,10 +405,12 @@ int BPFModule::annotate() {
     table_names_[table.name] = id++;
     GlobalValue *gvar = mod_->getNamedValue(table.name);
     if (!gvar) continue;
-    if (PointerType *pt = dyn_cast<PointerType>(gvar->getType())) {
-#if LLVM_MAJOR_VERSION >= 15
-      StructType *st = dyn_cast<StructType>(pt->getPointerElementType());
+#if LLVM_MAJOR_VERSION >= 14
```
chenhengqi, line 408: For consistency, change this to `>= 14`, see https://github.com/llvm/llvm-project/blob/llvmorg-14.0.0/llvm/lib/Transforms/Instrumentation/DataFlowSanitizer.cpp#L249-L258

yonghong-song: I checked the llvm14 release branch, the below is git blame result:
```
6ff18b08e6bf8 (George Balatsouras      2021-03-02 14:56:16 -0800  249) static StringRef getGlobalTypeString(const GlobalValue &G) {
b7dd329f2f3ff (Alexey Samsonov         2014-07-09 19:40:08 +0000  250)   // Types of GlobalVariables are always pointer types.
5f6eaac6115a4 (Manuel Jacob            2016-01-16 20:30:46 +0000  251)   Type *GType = G.getValueType();
858d38557857b (Eric Christopher        2020-06-20 00:24:57 -0700  252)   // For now we support excluding struct types only.
b7dd329f2f3ff (Alexey Samsonov         2014-07-09 19:40:08 +0000  253)   if (StructType *SGType = dyn_cast<StructType>(GType)) {
b7dd329f2f3ff (Alexey Samsonov         2014-07-09 19:40:08 +0000  254)     if (!SGType->isLiteral())
b7dd329f2f3ff (Alexey Samsonov         2014-07-09 19:40:08 +0000  255)       return SGType->getName();
b7dd329f2f3ff (Alexey Samsonov         2014-07-09 19:40:08 +0000  256)   }
b7dd329f2f3ff (Alexey Samsonov         2014-07-09 19:40:08 +0000  257)   return "<unknown type>";
b7dd329f2f3ff (Alexey Samsonov         2014-07-09 19:40:08 +0000  258) }

```

The changes are all pretty old.

I also checked the patch with a build of llvm14 release branch release/14.x. This patch actually caused the assertion error. With this patch llvm14 is okay.

So I suspect we should guard all changes in this patch with llvm15.

chenhengqi: OK, just make the changes specific to llvm15.

yonghong-song:
LGTM. Thanks for working on this!

PR #3958, opened 2022-04-25
tools: Fix misuse of BPF.kernel_struct_has_field in bio* tools

xingfeng2510:
BPF.kernel_struct_has_field possible return values: -1 (not support BTF), 1 (has a specified field), 0 (has no specified field).
This patch try to fix current flaws.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,25 @@ enable_testing()
 
 # populate submodules (libbpf)
 if(NOT CMAKE_USE_LIBBPF_PACKAGE)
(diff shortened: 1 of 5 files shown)


xingfeng2510:
btw, CI failed due to git additional safe.directory check option. 

```
fatal: unsafe repository ('/bcc' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory /bcc
CMake Warning at CMakeLists.txt:27 (message):
  Failed to update submodule libbpf
```

`git config --global --add safe.directory /bcc` or `git config --global --add safe.directory '*'` not working, this patch try to add root/libbpf/bpftool source directories into safe.directory.

PR #3964, opened 2022-04-26
bitesize: Fix the spelling mistake of historgram to histogram

nagarevaibhav:


nagarevaibhav:
Previous output:
~~~
DESCRIPTION
       Show I/O distribution for requested block sizes, by process name.

       This works by tracing block:block_rq_issue and prints a historgram of I/O size.

       Since this uses BPF, only the root user can use this tool.
~~~

Now:
~~~
DESCRIPTION
       Show I/O distribution for requested block sizes, by process name.

       This works by tracing block:block_rq_issue and prints a histogram of I/O size.

       Since this uses BPF, only the root user can use this tool.
~~~

nagarevaibhav:
Hello,

I hope you're safe and in great health

Thank you for merging!

PR #3968, opened 2022-04-28
tools: Use u32 as pid field type

xingfeng2510:
Some tools still use u64 as pid field type, this patch try to unify them.

### DIFF
--- a/tools/btrfsslower.py
+++ b/tools/btrfsslower.py
@@ -87,10 +87,10 @@ struct data_t {
     // XXX: switch some to u32's when supported
     u64 ts_us;
     u64 type;
-    u64 size;
+    u32 size;
     u64 offset;
     u64 delta_us;
-    u64 pid;
+    u32 pid;
     char task[TASK_COMM_LEN];
     char file[DNAME_INLINE_LEN];
 };
@@ -219,9 +219,11 @@ static int trace_return(struct pt_regs *ctx, int type)
         return 0;
 
     // populate output struct
-    u32 size = PT_REGS_RC(ctx);
-    struct data_t data = {.type = type, .size = size, .delta_us = delta_us,
-        .pid = pid};
+    struct data_t data = {};
+    data.type = type;
+    data.size = PT_REGS_RC(ctx);
+    data.delta_us = delta_us;
+    data.pid = pid;
     data.ts_us = ts / 1000;
     data.offset = valp->offset;
     bpf_get_current_comm(&data.task, sizeof(data.task));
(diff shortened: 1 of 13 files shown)


tools/btrfsslower.py
```diff
@@ -220,8 +220,8 @@
 
     // populate output struct
     u32 size = PT_REGS_RC(ctx);
-    struct data_t data = {.type = type, .size = size, .delta_us = delta_us,
-        .pid = pid};
+    struct data_t data = { .type = type, .size = size, .delta_us = delta_us };
+    data.pid = pid;
```
yonghong-song, line 224: any particular reason why original .pid = pid in the struct initialization doesn't work? The same for a few instances below.

xingfeng2510: when use u32 as pid field type, original .pid = pid in the struct initialization causes an error:
`invalid indirect read from stack off -104+92 size 96`
sizeof(data_t) equals 96 bytes rather than 92 (due to structure member alignment).

The BPF verifier in the kernel checks the stack boundary that a BPF program does not access outside of boundary or uninitialized stack area. Using struct with the padding as a map value, will cause invalid indirect read from stack failure on bpf_prog_load().

As [struct data_t exitsnoop](https://github.com/iovisor/bcc/blob/master/tools/exitsnoop.py#L113), remove struct padding with aligning members by using __attribute__((packed)) may be a better way.

yonghong-song: Using packed attribute is not a good idea. Previous aligned load/store (load 4 bytes) may become four load-one-byte instructions. This may cause verification failure.

The reason for failure is due to uninitialized padding. Let us change these cases like below
```
   struct t a;
   __builtin_memset(&a, 0, sizeof(struct t));
  a.f1 = ...;
  a.f2 = ...;
```
This should resolve uninitialized padding issue and also preserve aligned load/stores.

yonghong-song: The semantics in the above two difference approaches are actually different. As you mention, the `= {}` approach will do zero initialization for all members. But it is possible that implicit padding bytes will be random numbers and not set with zero. The `__builtin_memset` approach explicitly sets all fields *and* padding to zero.

But based on my past experience, clang should be able to use `__builtin_memset` for `= {}` case, and gcc may not. So I will merge your code as it is. If in the future, clang does change its behavior, we can convert *all* `= {}` approach to `__builtin_memset`.

xingfeng2510: Thank you for your detailed and clear explanation.

Issue #3702, opened 2021-11-15
CGROUP_SOCK_ADDR support?


guedou:
First of all, thanks a lot for `bcc`. It its an awesome tool to easily experiment with `eBPF`!

I recently needed to load an `CGROUP_SOCK_ADDR` program from Python. It turns out that it not currently possible as the `expected_attach_type` member of the `bpf_load_program_attr` structure must be filled before calling `bcc_prog_load_xattr()`. I successfully modified `BPFModule::bcc_func_load()` and I am no able to load `CGROUP_SOCK_ADDR` programs.

I am willing to provide a patch, but I need advice to do it. Shall I modify `BPFModule::bcc_func_load()` to accept an ``expected_attach_type`  argument (i.e. that will break the current API), or shall I add `BPFModule::bcc_func_load_xattr()` that will get a pointer to a `bpf_load_program_attr` structure  (i.e. that will be more generic and future proof)?

yonghong-song:
At this point, let us add an extra optional parameter to bcc_func_load() (after `flag`).
```
  int bcc_func_load(int prog_type, const char *name,
                    const struct bpf_insn *insns, int prog_len,
                    const char *license, unsigned kern_version,
                    int log_level, char *log_buf, unsigned log_buf_size,
                    const char *dev_name = nullptr,
                    unsigned flags = 0);
```

We are in the process to trying to use more libbpf repo API functions, so let us avoid creating new API functions at this point.

guedou:
Thanks a lot for you quick answer. Let me know if you want me to test the new bcc_func_load() function.

yonghong-song:
@guedou  Since this is for your use case, maybe you can try to implement the functionality to ensure it works for you?

PR #3969, opened 2022-04-28
Support the expected_attach_type attribute

guedou:
Here is a patch that addresses https://github.com/iovisor/bcc/issues/3702

It works fine for my use case and I would like to  gather feedback.

Thanks

### DIFF
--- a/src/cc/api/BPF.cc
+++ b/src/cc/api/BPF.cc
@@ -671,7 +671,7 @@ int BPF::poll_perf_buffer(const std::string& name, int timeout_ms) {
 }
 
 StatusTuple BPF::load_func(const std::string& func_name, bpf_prog_type type,
-                           int& fd, unsigned flags) {
+                           int& fd, unsigned flags, bpf_attach_type attach_type) {
   if (funcs_.find(func_name) != funcs_.end()) {
     fd = funcs_[func_name];
     return StatusTuple::OK();
@@ -692,7 +692,7 @@ StatusTuple BPF::load_func(const std::string& func_name, bpf_prog_type type,
   fd = bpf_module_->bcc_func_load(type, func_name.c_str(),
                      reinterpret_cast<struct bpf_insn*>(func_start), func_size,
                      bpf_module_->license(), bpf_module_->kern_version(),
-                     log_level, nullptr, 0, nullptr, flags);
+                     log_level, nullptr, 0, nullptr, flags, attach_type);
 
   if (fd < 0)
     return StatusTuple(-1, "Failed to load %s: %d", func_name.c_str(), fd);
--- a/src/cc/api/BPF.cc
+++ b/src/cc/api/BPF.cc
@@ -671,7 +671,7 @@ int BPF::poll_perf_buffer(const std::string& name, int timeout_ms) {
 }
 
 StatusTuple BPF::load_func(const std::string& func_name, bpf_prog_type type,
-                           int& fd, unsigned flags, bpf_attach_type attach_type) {
+                           int& fd, unsigned flags, bpf_attach_type expected_attach_type) {
   if (funcs_.find(func_name) != funcs_.end()) {
     fd = funcs_[func_name];
     return StatusTuple::OK();
@@ -692,7 +692,7 @@ StatusTuple BPF::load_func(const std::string& func_name, bpf_prog_type type,
   fd = bpf_module_->bcc_func_load(type, func_name.c_str(),
                      reinterpret_cast<struct bpf_insn*>(func_start), func_size,
                      bpf_module_->license(), bpf_module_->kern_version(),
-                     log_level, nullptr, 0, nullptr, flags, attach_type);
+                     log_level, nullptr, 0, nullptr, flags, expected_attach_type);
 
   if (fd < 0)
     return StatusTuple(-1, "Failed to load %s: %d", func_name.c_str(), fd);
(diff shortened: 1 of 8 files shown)


src/cc/api/BPF.h
```diff
@@ -263,7 +263,7 @@ class BPF {
   int poll_perf_buffer(const std::string& name, int timeout_ms = -1);
 
   StatusTuple load_func(const std::string& func_name, enum bpf_prog_type type,
-                        int& fd, unsigned flags = 0);
+                        int& fd, unsigned flags = 0, enum bpf_attach_type = (bpf_attach_type) __MAX_BPF_ATTACH_TYPE);
```
yonghong-song, line 266: can we use macro MAX_BPF_ATTACH_TYPE instead of __MAX_BPF_ATTACH_TYPE?

src/cc/bpf_module.cc
```diff
@@ -929,13 +929,16 @@ int BPFModule::bcc_func_load(int prog_type, const char *name,
...
                 int log_level, char *log_buf, unsigned log_buf_size,
-                const char *dev_name, unsigned flags) {
+                const char *dev_name, unsigned flags, int attach_type) {
   struct bpf_load_program_attr attr = {};
   unsigned func_info_cnt, line_info_cnt, finfo_rec_size, linfo_rec_size;
   void *func_info = NULL, *line_info = NULL;
   int ret;
 
   attr.prog_type = (enum bpf_prog_type)prog_type;
+  if (attach_type != __MAX_BPF_ATTACH_TYPE) {
```
yonghong-song, line 939: Again, MAX_BPF_ATTACH_TYPE?

src/cc/api/BPF.cc
```diff
@@ -671,7 +671,7 @@ int BPF::poll_perf_buffer(const std::string& name, int timeout_ms) {
 }
 
 StatusTuple BPF::load_func(const std::string& func_name, bpf_prog_type type,
-                           int& fd, unsigned flags) {
+                           int& fd, unsigned flags, bpf_attach_type attach_type) {
```
yonghong-song, line 674: It would be good to change the argument name to expected_attach_type. This is to load_func and not really do any attachment. The expected_attach_type is also consistent with the field name in the attribute.

src/python/bcc/__init__.py
```diff
@@ -495,7 +495,7 @@ def load_funcs(self, prog_type=KPROBE):
 
         return fns
 
-    def load_func(self, func_name, prog_type, device = None):
+    def load_func(self, func_name, prog_type, device = None, attach_type = 38):
```
yonghong-song, line 498: Can we define a call BPFAttachType to define some <name, value> pairs? Using 38 is kind of cryptic.

yonghong-song (commented):
The concept looks good. Just added some comments.

src/python/bcc/__init__.py
guedou: Thanks a lot for your feedback.

Could you explain what changes you have in mind? I did not understand what you are suggesting.

BTW, 38 is `MAX_BPF_ATTACH_TYPE`.

src/cc/api/BPF.h
```diff
@@ -263,7 +263,7 @@ class BPF {
   int poll_perf_buffer(const std::string& name, int timeout_ms = -1);
 
   StatusTuple load_func(const std::string& func_name, enum bpf_prog_type type,
-                        int& fd, unsigned flags = 0);
+                        int& fd, unsigned flags = 0, enum bpf_attach_type = (bpf_attach_type) MAX_BPF_ATTACH_TYPE);
```
yonghong-song, line 266: I think that using MAX_BPF_ATTACH_TYPE here might not the best option. Note that MAX_BPF_ATTACH_TYPE equal to the last ATTACH_TYPE + 1. Say it is 38. If in the future, a new attach type is added, the MAX_BPF_ATTACH_TYPE becomes 39.

So if an old bcc package may not work properly with newer kernel. So I think the default value as -1 probably a better choice.

src/cc/bpf_module.cc
```diff
@@ -929,13 +929,16 @@ int BPFModule::bcc_func_load(int prog_type, const char *name,
...
                 int log_level, char *log_buf, unsigned log_buf_size,
-                const char *dev_name, unsigned flags) {
+                const char *dev_name, unsigned flags, int expected_attach_type) {
   struct bpf_load_program_attr attr = {};
   unsigned func_info_cnt, line_info_cnt, finfo_rec_size, linfo_rec_size;
   void *func_info = NULL, *line_info = NULL;
   int ret;
 
   attr.prog_type = (enum bpf_prog_type)prog_type;
+  if (expected_attach_type != MAX_BPF_ATTACH_TYPE) {
```
yonghong-song, line 939: again, replace MAX_BPF_ATTACH_TYPE with -1 might be a better choice.

src/cc/bpf_module.h
```diff
@@ -144,7 +145,7 @@ class BPFModule {
                     const char *license, unsigned kern_version,
                     int log_level, char *log_buf, unsigned log_buf_size,
                     const char *dev_name = nullptr,
-                    unsigned flags = 0);
+                    unsigned flags = 0, int attach_type = MAX_BPF_ATTACH_TYPE);
```
yonghong-song, line 148: MAX_BPF_ATTACH_TYPE => -1?

src/python/bcc/__init__.py
yonghong-song: MAX_BPF_ATTACH_TYPE => -1?

guedou:
Thanks for the comments. I pushed a commit that uses `-1` as the default value.

Issue #3808, opened 2022-01-13
[Proposal] Bump minimum LLVM version to 11


davemarchevsky:
Currently there is no explicitly-stated minimum version of LLVM that bcc supports linking against. Based on `INSTALL.md` and @yonghong-song's [mailing list response](https://lists.iovisor.org/g/iovisor-dev/message/1930) from late 2020, which states

> bcc is supposed to work for all major llvm releases since llvm 3.7.

3.7 seems to be the implicit minimum version. [3.7 is from mid-2015](https://github.com/llvm/llvm-project/tree/release/3.7.x).

__I propose that we bump the minimum supported version of LLVM to 11 (released late 2020).__

__Why?__ 

`bcc` is one of the more widely-deployed projects in the BPF ecosystem, the tools specifically are many folks' introduction to BPF. In order to keep `bcc` working in as many environments as possible with minimum fuss we've historically tried to support a variety of older LLVMs. There are a few reasons why I think this is no longer the best path forward:

* __Working around LLVM API changes across so many versions is tedious at best and at worst actively prevents progress__. Sprinkling some [preprocessor conditionals](https://github.com/iovisor/bcc/search?q=LLVM_MAJOR_VERSION) is not the end of the world, generally LLVM APIs and functionality we rely on don't change too much across versions. This isn't the case currently, though: my ["Keep track of BPF progs using function instead of section"](https://github.com/iovisor/bcc/pull/3660) refactor won't work on older LLVMs due to an implementation difference. Landing that refactor is a necessary part of the "bcc -> libbpf" loader conversion project, which will result in an easier-to-maintain `bcc` with access to more cutting-edge libbpf features (e.g. global variables).
* __Old LLVMs can produce inefficient or broken BPF bytecode__. Although a C compiler from 2015 is not usually considered 'old', there's been significant work on the BPF backend since then, including many logic fixes to ensure that the BPF verifier will accept the generated bytecode. BPF maintainers have mentioned hearing from confused users whose BPF programs - generated by `bcc` - are rejected by the verifier, with upgrading to a more recent LLVM fixing the issue. This results in frustration and wasted time on the users' part ("I'm just trying to get this simple thing to work") and maintainers ("Didn't we fix this bad bytecode generation two years ago?"). 

__Why Now?__

[...]

PR #3977, opened 2022-05-04
Fix some redundant spaces in the INSTALL.md

JadeFlute0127:
Fix some redundant spaces in the INSTALL.md
some command line like : "curl  -LO  http://releases.llvm.org/7.0.1/llvm-7.0.1.src.tar.xz" should be "curl -LO http://releases.llvm.org/7.0.1/llvm-7.0.1.src.tar.xz".

And other:
When "make install bcc" in CentOS(7.X) system, we need the versions of llvm and clang to be greater than or equal to 10.0.0. Otherwise, an error will be reported as follows:
```
/tmp/bcc/src/cc/bpf_module.cc: In member function ‘virtual void ebpf::MyMemoryManager::notifyObjectLoaded(llvm::ExecutionEngine*, const llvm::object::ObjectFile&)’:
/tmp/bcc/src/cc/bpf_module.cc:108:46: error: no matching function for call to ‘llvm::object::SectionRef::getName() const’
       auto sec_name = section.get()->getName();
```

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -511,7 +511,7 @@ mkdir llvm-build
 
 cd llvm-build
 cmake3 -G "Unix Makefiles" -DLLVM_TARGETS_TO_BUILD="BPF;X86" \
-  -DCMAKE_BUILD_TYPE=Release ../llvm-10.0.10.src
+  -DCMAKE_BUILD_TYPE=Release ../llvm-10.0.0.src
 make
 sudo make install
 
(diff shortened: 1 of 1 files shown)


chenhengqi:
> And other: When "make install bcc" in CentOS system, we need the versions of llvm and clang to be greater than or equal to 10.0.0. Otherwise, an error will be reported as follows:
> 
> ```
> /tmp/bcc/src/cc/bpf_module.cc: In member function ‘virtual void ebpf::MyMemoryManager::notifyObjectLoaded(llvm::ExecutionEngine*, const llvm::object::ObjectFile&)’:
> /tmp/bcc/src/cc/bpf_module.cc:108:46: error: no matching function for call to ‘llvm::object::SectionRef::getName() const’
>        auto sec_name = section.get()->getName();
> ```

Before #3808  is landed, I think this compilation error should be addressed.

yonghong-song:
Since we intend to use more recent llvm compilers, changing to llvm10 from llvm7 looks good to me.

yonghong-song:
Since we intend to use more recent llvm compilers, changing to llvm10 from llvm7 looks good to me.

yonghong-song:
Could you squash 4 commits into one with proper commit message?

yonghong-song:
Could you squash 4 commits into one with proper commit message?

JadeFlute0127:
OK, I will close this PR and resubmit a new proper one

PR #3982, opened 2022-05-10
libbpf-tools: Add initial support to run execsnoop on arm64.

eiffel-fl:
Hi.

In this PR, I changed CO-RE `execsnoop` to use `K(RET)PROBE` on arm64 instead of tracepoint.
Indeed, it seems there is a [bug](https://marc.info/?l=linux-kernel&m=165210957108106&w=1) in kernel for arm64 where tracepoint `syscalls:sys_exit_execve` will never be triggered due to syscall number being forgotten in [`start_thread_common()`](https://elixir.bootlin.com/linux/v5.18-rc6/source/arch/arm64/include/asm/processor.h#L254).

Sadly, `execsnoop` on arm64 will not display the `exec` args as I think I need to deal with a loop without finite number of iterations:

```bash
$ uname -m
aarch64
$ sudo ./execsnoop 
runc             3017180 74876    0 
exe              3017187 3017180   0 
5                3017187 3017180   0 
gadgettracerman  3017189 3017180   0
```

### DIFF
--- a/libbpf-tools/execsnoop.bpf.c
+++ b/libbpf-tools/execsnoop.bpf.c
@@ -2,6 +2,7 @@
 #include <vmlinux.h>
 #include <bpf/bpf_helpers.h>
 #include <bpf/bpf_core_read.h>
+#include <bpf/bpf_tracing.h>
 #include "execsnoop.h"
 
 const volatile bool filter_cg = false;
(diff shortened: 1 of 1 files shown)


libbpf-tools/execsnoop.bpf.c
```diff
@@ -35,22 +36,33 @@ static __always_inline bool valid_uid(uid_t uid) {
 	return uid != INVALID_UID;
 }
 
+#ifdef __TARGET_ARCH_arm64
+SEC("kprobe/do_execveat_common.isra.0")
```
chenhengqi, line 40: This is fragile. Let's handle this difference in userspace.

You can write two BPF programs and enable them when available.

chenhengqi: You can always use kprobe for aarch64.

chenhengqi: > Sadly, with kprobe I think we will not be able to get command line arguments.

Why ?

chenhengqi: I see, it's a struct.

PR #3984, opened 2022-05-11
libbpf-tools/klockstat update

namhyung:
A couple of updates of klockstat in libbpf-tools.  The most important change is the per-thread mode (`-P` option) to track lock wait/hold stats for each task.  Also changed time format to have units.

```
$ sudo klockstat -n 5 -d 3 -P
Tracing mutex/sem lock events...  Hit Ctrl-C to end


                Tid              Comm  Avg Wait    Count   Max Wait   Total Wait
               8428   IPDL Background    4.8 us      213   379.0 us       1.0 ms
               8369    IPC I/O Parent    2.0 us      889   194.5 us       1.8 ms
             139151     IPC I/O Child    2.0 us      929   148.0 us       1.8 ms
               8400    GLXVsyncThread    3.1 us      182    25.9 us     559.9 us
             139148       Web Content    1.5 us     1013    24.0 us       1.5 ms

                Tid              Comm  Avg Hold    Count   Max Hold   Total Hold
             366280    kworker/u16:32   21.5 us     1760    31.1 ms      37.9 ms
               1056         in:imklog    4.0 ms      760    17.9 ms       3.0 s 
             377600         klockstat    1.0 ms      180     2.9 ms     188.6 ms
               5436            chrome   50.8 us       13   490.8 us     660.1 us
               2916       gnome-shell    4.4 us     1182   451.8 us       5.2 ms
Exiting trace of mutex/sem locks
```

### DIFF
--- a/libbpf-tools/klockstat.bpf.c
+++ b/libbpf-tools/klockstat.bpf.c
@@ -14,6 +14,7 @@
 const volatile pid_t targ_tgid = 0;
 const volatile pid_t targ_pid = 0;
 void *const volatile targ_lock = NULL;
+const volatile int per_thread = 0;
 
 struct {
 	__uint(type, BPF_MAP_TYPE_STACK_TRACE);
@@ -147,6 +148,10 @@ static void account(struct lockholder_info *li)
 {
 	struct lock_stat *ls;
 	u64 delta;
+	u32 key = li->stack_id;
+
+	if (per_thread)
+		key = li->task_id;
 
 	/*
 	 * Multiple threads may have the same stack_id.  Even though we are
@@ -155,15 +160,19 @@ static void account(struct lockholder_info *li)
 	 * by multiple readers at the same time.  They will be accounted as
 	 * the same lock, which is what we want, but we need to use atomics to
 	 * avoid corruption, especially for the total_time variables.
+	 * But it should be ok for per-thread since it's not racy anymore.
 	 */
-	ls = bpf_map_lookup_elem(&stat_map, &li->stack_id);
+	ls = bpf_map_lookup_elem(&stat_map, &key);
 	if (!ls) {
 		struct lock_stat fresh = {0};
 
-		bpf_map_update_elem(&stat_map, &li->stack_id, &fresh, BPF_ANY);
-		ls = bpf_map_lookup_elem(&stat_map, &li->stack_id);
+		bpf_map_update_elem(&stat_map, &key, &fresh, BPF_ANY);
+		ls = bpf_map_lookup_elem(&stat_map, &key);
 		if (!ls)
 			return;
+
+		if (per_thread)
+			bpf_get_current_comm(ls->acq_max_comm, TASK_COMM_LEN);
 	}
 
 	delta = li->acq_at - li->try_at;
@@ -176,7 +185,8 @@ static void account(struct lockholder_info *li)
 		 * Potentially racy, if multiple threads think they are the max,
 		 * so you may get a clobbered write.
 		 */
-		bpf_get_current_comm(ls->acq_max_comm, TASK_COMM_LEN);
+		if (!per_thread)
+			bpf_get_current_comm(ls->acq_max_comm, TASK_COMM_LEN);
 	}
 
 	delta = li->rel_at - li->acq_at;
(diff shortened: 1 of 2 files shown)


libbpf-tools/klockstat.c
```diff
@@ -368,25 +390,30 @@ static void print_acq_stat(struct ksyms *ksyms, struct stack_stat *ss,
...
 	printf("%37s %9s %8llu %10s %12s\n",
-	       symname(ksyms, ss->bt[0], buf, sizeof(buf)),
+	       print_caller(buf, sizeof(buf), ksyms, ss),
 	       print_time(avg, sizeof(avg), ss->ls.acq_total_time / ss->ls.acq_count),
 	       ss->ls.acq_count,
 	       print_time(max, sizeof(max), ss->ls.acq_max_time),
 	       print_time(tot, sizeof(tot), ss->ls.acq_total_time));
 	for (i = 1; i < nr_stack_entries; i++) {
-		if (!ss->bt[i])
+		if (!ss->bt[i] || env.per_thread)
```
brho, line 399: if you enforced that nr_stack_entries == 1 in arg parsing, you could skip these special cases for env.per_thread.

if you want to keep the env.per_thread check, i'd be tempted to just return from print_{acq,hld}_stat() before the stack entry parsing.

(here and below).

libbpf-tools/klockstat.bpf.c
```diff
@@ -147,6 +148,10 @@ static void account(struct lockholder_info *li)
 {
 	struct lock_stat *ls;
 	u64 delta;
+	u32 key = li->stack_id;
+
+	if (per_thread)
+		key = li->task_id;
 
 	/*
```
brho, line 156: maybe move this comment to where you assign key = li->stack_id?  since we aren't racy when dealing with task_id.

libbpf-tools/klockstat.bpf.c
```diff
@@ -156,15 +161,18 @@ static void account(struct lockholder_info *li)
...
-		bpf_map_update_elem(&stat_map, &li->stack_id, &fresh, BPF_ANY);
-		ls = bpf_map_lookup_elem(&stat_map, &li->stack_id);
+		bpf_map_update_elem(&stat_map, &key, &fresh, BPF_ANY);
+		ls = bpf_map_lookup_elem(&stat_map, &key);
 		if (!ls)
 			return;
-	}
+
+		if (per_thread)
+			bpf_get_current_comm(ls->acq_max_comm, TASK_COMM_LEN);
```
brho, line 174: nice.  much easier to tuck it away in the lock_stat.  we already have the space available, and its probably not worth making a per-thread pid-to-comm map.

libbpf-tools/klockstat.c
brho: or maybe have completely separate print_* functions for per-thread tracking vs per-lock (stackframe) tracking.

libbpf-tools/klockstat.bpf.c
```diff
@@ -147,6 +148,10 @@ static void account(struct lockholder_info *li)
 {
 	struct lock_stat *ls;
 	u64 delta;
+	u32 key = li->stack_id;
+
+	if (per_thread)
+		key = li->task_id;
```
brho, line 154: there's a slight risk of running out of space in the stat_map if you run klockstat too long, since there's no reaping of old tasks.  (and you don't want to, since you want the info from whoever was running but exited).  but not worth worrying about.

also, if the user is running with -R (reset after printing), it *will* reap.  so this would only be an issue if there were max_entries processes per output interval.  (which can happen).

namhyung: Right, maybe it's worth add a counter how many it missed due to the lack of storage.

libbpf-tools/klockstat.bpf.c
namhyung: Probably.. but I prefer not as it mentions the stack-id in the beginning. Or I can append something like this?

"But it should be ok for per-thread mode since it's not racy anymore."

libbpf-tools/klockstat.bpf.c
namhyung: Right.

libbpf-tools/klockstat.c
```diff
@@ -229,6 +232,9 @@ static error_t parse_arg(int key, char *arg, struct argp_state *state)
 	case 'T':
 		env->timestamp = true;
 		break;
+	case 'P':
+		env->per_thread = true;
+		break;
```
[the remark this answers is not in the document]
namhyung: Good idea.  The nr_locks still works with tasks to limit how many it prints.

libbpf-tools/klockstat.c
namhyung: Yeah, it'd be better to split.

namhyung (commented):
Thanks for the review!

libbpf-tools/klockstat.bpf.c
brho: SGTM

brho (approved):

brho:
LGTM

namhyung:
Ping!  Can anybody take a look and possibly merge this?

PR #3987, opened 2022-05-12
make BPFHashTable::get_table_offline check key is found

wineway:
refer to [man](https://man7.org/linux/man-pages/man2/bpf.2.html)
> If key is found, the operation returns zero and sets the next_key pointer to the key of the next element.  If key is not found, the operation returns zero and sets the next_key pointer to the key of the first element. 

when `KeyType` of BPFTable mismatches the actual key type, `bpf_get_next_key` maybe not find the key, and trap in an infinity loop

we should detect this situation and throw an error to user.


yonghong-song:
@wineway Do you have an actual case for this infinite loop? Is it possible to have an infinite loop theoretically yet? But I would argument it is unlikely since this needs map update/delete rate pretty high. Also this patch won't solve the problem. For example, what if the first element is deleted and when get_next_key got an illegal key and go to the original second element.

wineway:
> @wineway Do you have an actual case for this infinite loop? Is it possible to have an infinite loop theoretically yet? 

@yonghong-song  yes, I found this case when key type of bpf_map in bpf code is u64 and I set it to u32  in my cpp code by mistake, then the program trap in an infinite loop

>  what if the first element is deleted and when get_next_key got an illegal key and go to the original second element.

I think this is another problem, if this case happened, the `BPFTable::get_table_offline` result is incorrect, and shouldn't return directly, maybe we can solve this case by retry?

yonghong-song:
This patch added batch interface in bcc.
```
commit 748d6b51b1cb3090fb83980f5ba018ece48f656c
Author: Emilien Gobillot <emilien.gobillot@gmail.com>
Date:   Sun Apr 25 19:18:50 2021 +0200

    add bpf_map_lookup_batch and bpf_map_delete_batch in bcc (#3363)
    
     . add bpf_map_lookup_batch and bpf_map_delete_batch in bcc
     . add test_map_batch_ops.py to test batch lookup and delete on map
     . add items_lookup_batch() and items_delete_batch() in the reference guide
     . add keys as an optional argument to items_delete_batch
```

It might be a good idea to implement get_table_offline() with trying to use batch interface if it is available in the kernel. If not, let us maintain the current approach.
Going forward, as people use new kernels, the current approach (which may cause infinite loop) will not be used any more.

wineway:
> It might be a good idea to implement get_table_offline() with trying to use batch interface if it is available in the kernel. If not, let us maintain the current approach. Going forward, as people use new kernels, the current approach (which may cause infinite loop) will not be used any more.

@yonghong-song I found that we can check `sizeof(KeyType) == desc.key_size` to avoid this memory out of bound, and I reimplement `get_table_offline` by using `bpf_map_lookup_batch`.

I am still trying to find out how to run tests in my dev env :(
but I think I can push first and ask some feedback :)

PR #3990, opened 2022-05-12
cachetop: Add missing entry of CMD in the FIELDS section

nagarevaibhav:
Man page before:
~~~
FIELDS
       PID    Process ID of the process causing the cache activity.

       UID    User ID of the process causing the cache activity.

       HITS   Number of page cache hits.
~~~

Man page now:
~~~
FIELDS
       PID    Process ID of the process causing the cache activity.

       UID    User ID of the process causing the cache activity.

       CMD    Name of the process.
~~~


davemarchevsky:
Can you remove the unrelated commits from this PR?>

PR #3991, opened 2022-05-13
bcc/usdt: fix parse double-reg-indirect addressing mode in arguments …

maqiao-mq:
…on AArch64

For bpftrace testcase usdt_args.c, it will generates usdt arguments
as follow:
> stapsdt              0x00000036       NT_STAPSDT (SystemTap probe descriptors)
>    Provider: usdt_args
>    Name: index_8
>    Location: 0x0000000000400724, Base: 0x0000000000400810, Semaphore: 0x0000000000000000
>    Arguments: -1@[x0, x1]

Such double-reg-indirect addressing mode is not supported yet, so
bpftrace will report such error:
> Parse error:
>     -1@[x0, x1]
> -----------^
> ERROR: couldn't get argument 0 for ./usdt_args:usdt_args:index_8

This patch fixed it.

Signed-off-by: Qiao Ma <mqaio@linux.alibaba.com>

### DIFF
--- a/src/cc/usdt/usdt_args.cc
+++ b/src/cc/usdt/usdt_args.cc
@@ -190,16 +190,27 @@ bool ArgumentParser_aarch64::parse_size(ssize_t pos, ssize_t &new_pos,
 }
 
 bool ArgumentParser_aarch64::parse_mem(ssize_t pos, ssize_t &new_pos,
-                                       std::string &reg_name,
-                                       optional<int> *offset) {
-  if (parse_register(pos, new_pos, reg_name) == false)
+                                       Argument *dest) {
+  std::string base_reg_name, index_reg_name;
+
+  if (parse_register(pos, new_pos, base_reg_name) == false)
     return false;
+  dest->base_register_name_ = base_reg_name;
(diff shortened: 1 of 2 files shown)


yonghong-song:

src/cc/usdt/usdt_args.cc
```diff
@@ -190,16 +190,25 @@ bool ArgumentParser_aarch64::parse_size(ssize_t pos, ssize_t &new_pos,
...
     return false;
+  dest->base_register_name_ = base_reg_name;
 
   if (arg_[new_pos] == ',') {
     pos = new_pos + 1;
-    new_pos = parse_number(pos, offset);
-    if (new_pos == pos)
-      return error_return(pos, pos);
+    new_pos = parse_number(pos, &dest->deref_offset_);
+    if (new_pos == pos) {
```
davemarchevsky, line 203: Can you add a comment here like `// offset isn't a number, so must be a reg`?

davemarchevsky (approved):
Aside from comment nit, LGTM

PR #3997, opened 2022-05-17
add actual send msg size for tcptop

jackygam2001:
add actual send msg size for tcptop

### DIFF
--- a/tools/tcptop.py
+++ b/tools/tcptop.py
@@ -110,8 +110,9 @@ struct ipv6_key_t {
 };
 BPF_HASH(ipv6_send_bytes, struct ipv6_key_t);
 BPF_HASH(ipv6_recv_bytes, struct ipv6_key_t);
+BPF_HASH(sock_store, u32, struct sock *);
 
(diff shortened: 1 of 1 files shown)


tools/tcptop.py
```diff
@@ -141,26 +154,56 @@ def range_check(string):
...
     }
+    sock_store.delete(&tid);
     // else drop
 
     return 0;
 }
 
+int kretprobe__tcp_sendmsg(struct pt_regs *ctx)
+{
+    int size = PT_REGS_RC(ctx);
```
yonghong-song, line 173: Should we check `size <= 0` here? If that is the case, we just return 0? Similar to the original implementation, it is not needed to clear the sock_store map since next time when tid is reused, it will be updated first.

jackygam2001: > Should we check `size <= 0` here? If that is the case, we just return 0? Similar to the original implementation, it is not needed to clear the sock_store map since next time when tid is reused, it will be updated first.

thanks for your good suggestion. yes if the size <=0,  it need not to call the tcp_sendstat function. will update it   in next commit.

PR #4011, opened 2022-05-24
tools/offwaketime: Fix concurrency issue caused by idle threads

xingfeng2510:
Fix concurrency issue caused by idle threads.

There is an example output: target process swapper/82 0 blocked time is 18446744073743394 us.

```
waker:            0
    --               --
    b'finish_task_switch'
    b'schedule_idle'
    b'do_idle'
    b'cpu_startup_entry'
    b'start_secondary'
    b'secondary_startup_64'
    target:          swapper/82 0
        18446744073743394
```

Also, I think idle threads should be excluded by this tool, which may pollution output.

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -1005,7 +1005,7 @@ class BPF(object):
                 if os.path.isdir(evt_dir):
                     tp = ("%s:%s" % (category, event))
                     if re.match(tp_re.decode(), tp):
-                        results.append(tp)
+                        results.append(tp.encode())
         return results
 
     @staticmethod


yonghong-song:
In `Fix concurrency issue caused by idle threads`, could you explain what *concurrency* issue you are referring to?
I think in general, we should display the transition caused by idle thread as well. User can do filtering based on tgid/pid if they need. Also in general, you won't see lots of 'idle' thread transition. If the system is really idle, then we shouldn't probably care too much of offwaketime. How much you see 'idle' thread transition in your system?

yonghong-song:
In `Fix concurrency issue caused by idle threads`, could you explain what *concurrency* issue you are referring to?
I think in general, we should display the transition caused by idle thread as well. User can do filtering based on tgid/pid if they need. Also in general, you won't see lots of 'idle' thread transition. If the system is really idle, then we shouldn't probably care too much of offwaketime. How much you see 'idle' thread transition in your system?

PR #4015, opened 2022-05-26
libbpf-tools/bitesize: inline trace_rq_issue

cobrien7:
When running with 5.10 kernel, bitesize fails verification as a result
of passing a modified ctx pointer into trace_rq_issue. Forcing the
compiler to inline trace_rq_issue fixes the problem.

Signed-off-by: Connor O'Brien <connoro@google.com>

### DIFF
--- a/libbpf-tools/bitesize.bpf.c
+++ b/libbpf-tools/bitesize.bpf.c
@@ -38,7 +38,7 @@ static __always_inline bool comm_allowed(const char *comm)
 	return true;
 }
 
-static int trace_rq_issue(struct request *rq)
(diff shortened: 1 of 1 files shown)


yonghong-song:
Your change looks okay. But could you have a little more details on the commit message, esp. w.r.t. `passing a modified ctx pointer into trace_rq_issue`. Could you show the assembly code and verification failure log so people can see what is the exact problem?

yonghong-song:
Your change looks okay. But could you have a little more details on the commit message, esp. w.r.t. `passing a modified ctx pointer into trace_rq_issue`. Could you show the assembly code and verification failure log so people can see what is the exact problem?

PR #4026, opened 2022-05-31
tools/biolatency: Add disk filter support

xingfeng2510:
Sometimes, I just want to focus on a specified disk rather than all disks or per-disk. Refer to libbpf-tools/biolatency, this patch try to add disk filter support.


yonghong-song:
LGTM. Thanks!

PR #4046, opened 2022-06-10
Update the tracefs path.

tiann:
The tracefs is mounted at `/sys/kernel/tracing` after 4.1 kernel. For backward compatibility, when mounting the debugfs file system, the tracefs file system will be automatically mounted at `/sys/kernel/debug/tracing`; But when CONFIG_TRACEFS_DISABLE_AUTOMOUNT is enabled, the automount wouldn't happen and the `/sys/kernel/debug/tracing` is not exist.

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -205,7 +205,7 @@ This is a macro that instruments the tracepoint defined by *category*:*event*.
 The tracepoint name is `<category>:<event>`.
 The probe function name is `tracepoint__<category>__<event>`.
 
-Arguments are available in an ```args``` struct, which are the tracepoint arguments. One way to list these is to cat the relevant format file under /sys/kernel/debug/tracing/events/*category*/*event*/format.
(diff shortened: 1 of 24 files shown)


chenhengqi:
Anyway, you can mount tracefs to `/sys/kernel/debug/tracing` manually without this change, no ?

Or parse `/proc/mounts` ?

tiann:
> Anyway, you can mount tracefs to `/sys/kernel/debug/tracing` manually without this change, no ?

According to this change: https://osdn.net/projects/android-x86/scm/git/kernel/commits/412bd71850dbd2d78ee3bbd1e43e3ebe3c9bd819, there isn't any mount point in `/sys/kernel/debug/tracing` when the automount is disabled, we can't mount it manually and we can't create a node in sysfs even we are root.

> Or parse `/proc/mounts` ?

This is a better way but it needs lots of change

evilpan:
By changing **tracefs** location from debugfs to `/sys/kernel/tracing` would break some older Linux distributions. Perhaps parsing `/proc/mounts` is the only option to refactor the code.

PR #4047, opened 2022-06-11
make deadlock can run everywhere

mackong:


yonghong-song:
Could you have a good commit message to explain the problem you are facing and how this patch addressed your issue?

4383:
I confirm that I face the same problem on my side.
I patched my BCC setup with the proposed patch and it helped me to get rid of this error.

The commit message seems explicit to me, but maybe some words could be added to highlight the problem like suggested by @yonghong-song. Can't hurt.

Thanks @mackong for proposing this fix. Hopefully it will be merged.

PR #4053, opened 2022-06-15
libbpf-tools: fix tcpconnect build errors

aspsk:
With the latest clang and bpftool the build of tcpconnect fails as
follows (output patched a bit for readability):

    bpftool gen skeleton tcpconnect.bpf.o > tcpconnect.skel.h
    Error: Something is wrong for .rodata's variable #1: need offset 0, already at 4.

This happens because the filter_ports variable is declared using a ".rodata hack":

    SEC(".rodata") int filter_ports[MAX_PORTS]

This breaks with the recent clang, as the filter_ports variable is placed into
the '.rodata,aw' section. Older clang would put it into '.rodata,a' section
where all 'const volatile' variables are placed. The result is that the
filter_ports variable has a wrong offset of 0 in BTF_KIND_DATASEC.

To hack the hack we can declare the variable as

    SEC(".rodata") const int filter_ports[MAX_PORTS]

but, instead, we now can just declare it as

    const volatile int filter_ports[MAX_PORTS]

In fact, this was already done in a02663be ("libbpf-tools: update bpftool and
fix .rodata hack"), but a later commit f8ac3c6c ("libbpf-tools: fix tcpconnect
compile errors") reverted the change without any comments.

### DIFF
--- a/libbpf-tools/tcpconnect.bpf.c
+++ b/libbpf-tools/tcpconnect.bpf.c
@@ -11,7 +11,7 @@
 #include "maps.bpf.h"
 #include "tcpconnect.h"
 
-SEC(".rodata") int filter_ports[MAX_PORTS];
+const volatile int filter_ports[MAX_PORTS];
 const volatile int filter_ports_len = 0;
 const volatile uid_t filter_uid = -1;
 const volatile pid_t filter_pid = 0;


aspsk:
cc: @anakryiko @yonghong-song This looks like a bug in BTF generation: variables `x` and `y` defined as `SEC(".rodata") const int x` and `SEC(".rodata") int y` would eventually get into the same `DATASEC .rodata`, but from the ELF perspective they live in different sections:
```
$ llvm-readelf -S .output/example.bpf.o
...
[ 3] .rodata           PROGBITS        0000000000000000 000040 000004 00  WA  0   0  4
[ 4] .rodata           PROGBITS        0000000000000000 000044 000004 00   A  0   0  4
```

anakryiko:
> cc: @anakryiko @yonghong-song This looks like a bug in BTF generation: variables `x` and `y` defined as `SEC(".rodata") const int x` and `SEC(".rodata") int y` would eventually get into the same `DATASEC .rodata`, but from the ELF perspective they live in different sections:
> 
> ```
> $ llvm-readelf -S .output/example.bpf.o
> ...
> [ 3] .rodata           PROGBITS        0000000000000000 000040 000004 00  WA  0   0  4
> [ 4] .rodata           PROGBITS        0000000000000000 000044 000004 00   A  0   0  4
> ```

isn't this some sort of ELF violation to have two different sections with the same name? I think artificial SEC(".rodata") for non-const variable is a pretty ugly hack, so `const volatile int filter_ports[..];` makes more sense.

But I'd imagine that compiler would complain about this "duplicate" .rodata. @yonghong-song  can you please check when you get a chance?

anakryiko (approved):

yonghong-song:
For
```
SEC(".rodata") int filter_ports[MAX_PORTS];
const volatile int filter_ports_len = 0;
const volatile uid_t filter_uid = -1;
const volatile pid_t filter_pid = 0;
const volatile bool do_count = 0;
```
The latest clang generates two sections:
```
  [ 3] .rodata           PROGBITS        0000000000000000 000040 00000c 00   A  0   0  4
  [ 4] .rodata           PROGBITS        0000000000000000 00004c 000100 00  WA  0   0  4
```
The first ([3]) is readonly for three `const volatile` variables. The second ([4]) permits W (write).
Unfortunately, because they have the same section name, in BTF generation. All these 5 variables
are put in the *same* datasec. But both filter_ports and filter_ports_len have the same section
offset 0. This will cause the problem during skeleton generation since the same offset for two
different variables are not allowed. The fix to change SEC usage to const volatile is the correct fix.

PR #4060, opened 2022-06-17
tools/llcstat: Add TID info support

yezhem:
The tools/llcstat only collect info pre-process, sometimes, I just want to focus on per-thread rather than per-process.
This patch tries to add tid info support.

### DIFF
--- a/libbpf-tools/llcstat.bpf.c
+++ b/libbpf-tools/llcstat.bpf.c
@@ -3,36 +3,43 @@
 #include <vmlinux.h>
 #include <bpf/bpf_helpers.h>
 #include <bpf/bpf_tracing.h>
+#include "maps.bpf.h"
 #include "llcstat.h"
 
 #define MAX_ENTRIES	10240
 
+const volatile bool targ_per_thread = false;
+
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
 	__uint(max_entries, MAX_ENTRIES);
-	__type(key, u64);
-	__type(value, struct info);
+	__type(key, struct key_info);
+	__type(value, struct value_info);
 } infos SEC(".maps");
 
 static __always_inline
 int trace_event(__u64 sample_period, bool miss)
 {
-	u64 pid = bpf_get_current_pid_tgid();
-	u32 cpu = bpf_get_smp_processor_id();
-	struct info *infop, info = {};
-	u64 key = pid << 32 | cpu;
+	struct key_info key = {};
+	struct value_info *infop, zero = {};
 
-	infop = bpf_map_lookup_elem(&infos, &key);
-	if (!infop) {
-		bpf_get_current_comm(info.comm, sizeof(info.comm));
-		infop = &info;
-	}
+	u64 pid_tgid = bpf_get_current_pid_tgid();
+	key.cpu = bpf_get_smp_processor_id();
+	key.pid = pid_tgid >> 32;
+	if (targ_per_thread)
+		key.tid = (u32)pid_tgid;
+	else
+		key.tid = key.pid;
+
+	infop = bpf_map_lookup_or_try_init(&infos, &key, &zero);
+	if (!infop)
+		return 0;
 	if (miss)
 		infop->miss += sample_period;
 	else
 		infop->ref += sample_period;
-	if (infop == &info)
-		bpf_map_update_elem(&infos, &key, infop, 0);
+	bpf_get_current_comm(infop->comm, sizeof(infop->comm));
+
 	return 0;
 }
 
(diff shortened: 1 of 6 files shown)


davemarchevsky (changes requested):
This seems reasonable, but I'd like to see:
* This functionality gated behind a flag. We want to preserve current "aggregate by pid only" behavior for folks by default
* Similar functionality added to `libbpf-tools`'s `llcstat`

libbpf-tools/llcstat.bpf.c
```diff
@@ -3,36 +3,42 @@
...
-	}
+	struct key_info key = {};
+	struct value_info *infop, zero = {};
+
+	key.cpu = bpf_get_smp_processor_id();
+	key.pid = bpf_get_current_pid_tgid() >> 32;
+	if (targ_per_thread)
+		key.tid = (u32)bpf_get_current_pid_tgid();
+	else
+		key.tid = key.pid;
```
chenhengqi, line 31: `key.tid = 0;` ?

yezhem: if not set targ_per_thread, do not use key.tid filed, default key.tid = key.pid

yezhem:
> This seems reasonable, but I'd like to see:
> 
> * This functionality gated behind a flag. We want to preserve current "aggregate by pid only" behavior for folks by default
> * Similar functionality added to `libbpf-tools`'s `llcstat`

pls review

yezhem:
> This seems reasonable, but I'd like to see:
> 
> * This functionality gated behind a flag. We want to preserve current "aggregate by pid only" behavior for folks by default
> * Similar functionality added to `libbpf-tools`'s `llcstat`

@davemarchevsky pls review

tools/llcstat.py
```diff
@@ -98,22 +117,39 @@
...
-
-print('PID      NAME             CPU     REFERENCE         MISS    HIT%')
+    if args.tid:
+        miss_count[(k.pid, k.tid, k.cpu, k.name)] = v.value
+    else:
+        miss_count[(k.pid, k.cpu, k.name)] = v.value
+
+header_text = 'PID      NAME             CPU     REFERENCE         MISS    HIT%'
+format_text = '{:<8d} {:<16s} {:<4d} {:>12d} {:>12d} {:>6.2f}%'
+if args.tid:
```
davemarchevsky, line 127: Can all this printing code be refactored to avoid so many `if args.tid` everywhere?

Something like 
```python
header_text = 'PID   '
if args.tid:
  header_text += 'TID   '
header_text += 'NAME       CPU ....'
```
and similar gradual buildup of the formatted output?

This way when further optional args are added which change the output format it'll be straightforward to incorporate them

libbpf-tools/llcstat.c
```diff
@@ -143,11 +150,20 @@ static void print_map(struct bpf_map *map)
...
-		pid = next_key >> 32;
-		cpu = next_key;
-		printf("%-8u %-16s %-4u %12llu %12llu %6.2f%%\n", pid, info.comm,
-			cpu, info.ref, info.miss, info.ref > 0 ?
-			hit * 1.0 / info.ref * 100 : 0);
+		cpu = next_key.cpu;
+		pid = next_key.pid;
+		tid = next_key.tid;
+		if (env.per_thread) {
+			printf("%-8u %-8u %-16s %-4u %12llu %12llu %6.2f%%\n",
```
davemarchevsky, line 157: similar to the other comment in this review pass, it should be possible to build up this formatted output more gradually such that the if statement only does tid-specific work

tools/llcstat.py
```diff
@@ -41,16 +46,19 @@
 
 struct key_t {
     int cpu;
-    int pid;
+    u32 pid;
+    TIDINFO
```
davemarchevsky, line 50: Can we just keep `u32 tid` here for both use-tid and dont-use-tid cases? I'd like to limit use of `bpf_text.replace` as it makes the program harder to understand

tools/llcstat.py
```diff
@@ -41,16 +46,19 @@
...
 
 BPF_HASH(ref_count, struct key_t);
 BPF_HASH(miss_count, struct key_t);
 
 static inline __attribute__((always_inline)) void get_key(struct key_t* key) {
+    u64 pid_tgid = bpf_get_current_pid_tgid();
     key->cpu = bpf_get_smp_processor_id();
-    key->pid = bpf_get_current_pid_tgid() >> 32;
+    key->pid = pid_tgid >> 32;
+    GET_TID
```
davemarchevsky, line 61: similarly to the other `bpf_text.replace`-related comment, making the replacement as simple as possible will make this easier to understand

```python
if (GET_TID)
  key->tid = (u32)pid_tgid;
```

have the replacement just be `1` or `0`
```

davemarchevsky (changes requested):
aside from minor nits, looking pretty good

tools/llcstat.py
yezhem: DONE

davemarchevsky (approved):

PR #4062, opened 2022-06-21
sync with libbpf + backport `struct bpf_create_map_attr`

chantra:
Since https://github.com/libbpf/libbpf/commit/7e8d4234ac0b24492dd473a2ba80204e175b87df, this structure is gone from libbpf.

BCC uses it as a structure to pass around `bcc_create_map_xattr` and `libbpf_bpf_map_create`.
The alternative would be to modify both libbpf_bpf_map_create and bcc_create_map_xattr to take each arguments, which I am not sure it would be any better.
 Renamed the struct from `bpf_create_map_xattr` to `bcc_create_map_xattr` to better reflect this is a bcc-provided struct, not bpf anymore.

### DIFF
--- a/src/cc/bpf_module.cc
+++ b/src/cc/bpf_module.cc
@@ -404,7 +404,7 @@ int BPFModule::create_maps(std::map<std::string, std::pair<int, int>> &map_tids,
     }
 
     if (pinned_id <= 0) {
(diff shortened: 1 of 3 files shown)


chantra:
5/11 passing checks, this seems to align with other PRs (such as #4055), so I suppose it is fine.

PR #4065, opened 2022-06-21
tools/biosnoop: Add disk filter support

xingfeng2510:
It may be useful to focus on a specified disk rather than all disks. Refer to libbpf-tools/biosnoop, this patch try to add disk filter support.

@yonghong-song Please have a look, thanks.


yonghong-song:
This will make biosnoop.py compatible with libbpf-tools version. LGTM. thanks!

PR #4066, opened 2022-06-21
libbpf-tools: Various fixes for old kernels

chenhengqi:
Running some libbpf tools on old kernels (e.g. kernel v5.4) results in error.
This is mainly because the BPF verifiers shipped with old kernels does NOT
smart enough to do dead code elimination and range check, etc. Or missing some
features.

Fixing them by making the BPF verifier happy and do feature detection.

### DIFF
--- a/libbpf-tools/tcpconnlat.bpf.c
+++ b/libbpf-tools/tcpconnlat.bpf.c
@@ -31,7 +31,7 @@ struct {
 	__uint(value_size, sizeof(u32));
 } events SEC(".maps");
 
-static __always_inline int trace_connect(struct sock *sk)
+static int trace_connect(struct sock *sk)
 {
 	u32 tgid = bpf_get_current_pid_tgid() >> 32;
 	struct piddata piddata = {};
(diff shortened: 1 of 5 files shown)


libbpf-tools/tcpconnlat.bpf.c
```diff
@@ -85,12 +72,12 @@ int BPF_PROG(tcp_rcv_state_process, struct sock *sk)
 			sizeof(event.comm));
 	event.ts_us = ts / 1000;
 	event.tgid = piddatap->tgid;
-	event.lport = sk->__sk_common.skc_num;
-	event.dport = sk->__sk_common.skc_dport;
-	event.af = sk->__sk_common.skc_family;
+	event.lport = BPF_CORE_READ(sk, __sk_common.skc_num);
```
zgsbughammer, line 75: I wonder any specific reason for choosing BPF_CORE_READ?  Cuz i think  BTF-enabled prog should be given top prio in terms of performance.  does kernel 5.4 not support btf relocation for this or sth else?

PR #4070, opened 2022-06-23
libbpf-tools: fix fentry_try_attach bug

zgsbughammer:
the bpf verifier would complain on program exit without initing R0 but the buggy implementation makes up one BPF_EXIT instruction only, which would be rejected by the verifier, and what's worse is that it does not log any err about the try-to-load failure, which makes all tools use this api would silently fall back to kprobe unexpectedly.

this patch fixs it and also prints verbose message about the result of the try.

for instance, if the [init R0 instruction](https://github.com/iovisor/bcc/pull/4070/files#diff-52b43275661eff7b28907ff06ec5011abea3309ea24f5b54fe1cbc079ee0c23aR999) were not added, once we call the fixed version of **fentry_try_attach** , you would get the following hints:
```sh 
failed to try attaching to fentry: 0: (95) exit
R0 !read_ok
processed 1 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
```
Signed-off-by: Ze Gao <ze_gao@outlook.com>

### DIFF
--- a/libbpf-tools/trace_helpers.c
+++ b/libbpf-tools/trace_helpers.c
@@ -993,16 +993,25 @@ bool is_kernel_module(const char *name)
 
 static bool fentry_try_attach(int id)
 {
-	struct bpf_insn insns[] = { { .code = BPF_JMP | BPF_EXIT } };
-	LIBBPF_OPTS(bpf_prog_load_opts, opts);
 	int prog_fd, attach_fd;
+	char error[4096];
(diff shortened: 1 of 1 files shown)


yonghong-song:
The patch looks good to me. Yes, some earlier version of fentry bpf program return value is not enforced in verifier and later it is enforced with 0.

PR #4076, opened 2022-06-27
libbpf-tools: Fix biopattern with recent tracepoint change

chenhengqi:
After kernel commit d5869fdc189f ("block: introduce block_rq_error tracepoint"),
tracepoint block_rq_complete now shares the same argument struct as
`struct trace_event_raw_block_rq_completion` with tracepoint block_rq_error.
Because of that, now biopattern is broken because `struct trace_event_raw_block_rq_complete`
is disappeared from kernel BTF. Fix it by checking type existence.

### DIFF
--- a/libbpf-tools/biopattern.bpf.c
+++ b/libbpf-tools/biopattern.bpf.c
@@ -5,6 +5,7 @@
 #include <bpf/bpf_tracing.h>
 #include "biopattern.h"
 #include "maps.bpf.h"
+#include "core_fixes.bpf.h"
 
 const volatile bool filter_dev = false;
 const volatile __u32 targ_dev = 0;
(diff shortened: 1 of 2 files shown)


libbpf-tools/biopattern.bpf.c
```diff
@@ -17,12 +18,24 @@ struct {
...
+	u32 nr_sector;
+	u32 dev;
+
+	if (has_block_rq_completion()) {
+		struct trace_event_raw_block_rq_completion___x *ctx = args;
+		sector = BPF_CORE_READ(ctx, sector);
+		nr_sector = BPF_CORE_READ(ctx, nr_sector);
+		dev = BPF_CORE_READ(ctx, dev);
+	} else {
+		struct trace_event_raw_block_rq_complete *ctx = args;
```
yonghong-song, line 34: What if the vmlinux.h is updated with newer kernel and trace_event_raw_block_rq_complete type is not available?
Should we define this type in the header file as well?

chenhengqi: Yes, that will greatly improve the portability.

But I would like to defer that to another PR, and include other type definitions (for struct task_struct etc).

yonghong-song: Okay, sounds good. We can have follow-up patch once vmlinux.h is updated in the future.

PR #4078, opened 2022-06-29
tools/biotop: Fix output format specifier

xingfeng2510:
biotop output column 'Kbytes' looks not good. One example is as follows (first tow suffixed with .0, last one is 0.0078125):
```
PID     COMM             D MAJ MIN DISK       I/O  Kbytes  AVGms
941343  mysqld           W 253 0   vda       1262 56476.0   2.00
447     jbd2/vda1-8      W 253 0   vda        547  6712.0   0.89
0                        R 0   0   ?            1 0.0078125   0.34
```

This patch try to fix it by modifying output format specifier. Please have a look @davemarchevsky, thanks.

### DIFF
--- a/tools/biotop.py
+++ b/tools/biotop.py
@@ -259,7 +259,7 @@ while 1:
 
         # print line
         avg_ms = (float(v.us) / 1000) / v.io
-        print("%-7d %-16s %1s %-3d %-3d %-8s %5s %7s %6.2f" % (k.pid,
+        print("%-7d %-16s %1s %-3d %-3d %-8s %5s %7d %6.2f" % (k.pid,
             k.name.decode('utf-8', 'replace'), "W" if k.rwflag else "R",
             k.major, k.minor, diskname, v.io, v.bytes / 1024, avg_ms))
 


yonghong-song:
What is the output with this patch? I actually got a failure,
```
  File "biotop.py", line 245, in <module>
    "D", "MAJ", "MIN", "DISK", "I/O", "Kbytes", "AVGms"))
TypeError: %d format: a number is required, not str
```

xingfeng2510:
An example of output:

```
15:29:32 loadavg: 0.00 0.00 0.00 1/479 812152

PID     COMM             D MAJ MIN DISK       I/O  Kbytes  AVGms
811530  kworker/u32:2    W 253 0   vda          2       8   0.73
807227  kworker/u32:0    W 253 0   vda          1       4   0.83
```

type(v.bytes) is `<class 'int'>`.  What's your python3-bcc version？

yonghong-song:
Okay. I tried again. My python version is 3.6 and looks like it works now. Not sure why it doesn't work previously.

Think about the issue again. I think current implementation looks okay. For example,
```
PID     COMM             D MAJ MIN DISK       I/O  Kbytes  AVGms
941343  mysqld           W 253 0   vda       1262 56476.0   2.00
447     jbd2/vda1-8      W 253 0   vda        547  6712.0   0.89
0                        R 0   0   ?            1 0.0078125   0.34
```
If we use '%d' instead of '%s', the third raw will have 'Kbytes' (0) but 'AVGms' (0.34). That doesn't really make sense.

xingfeng2510:
Okay, thanks. I closed this patch.

PR #4088, opened 2022-07-04
Add RISC-V support for libbpf tools

chenhengqi:


dlan17 (commented):
sounds there are three independent commits combined in this single PR, I'm not sure if this is acceptable by bcc project, but I'd rather see them as separated

1) can you document the full command how to generate vmlinux.h? so people will do same update next time
2) can we also follow the convention as other Architecture, to name the file as vmlinux_${version}.h and have a symbol link to vmlinux.h?

thanks, I really appreciate!

chenhengqi:
> sounds there are three independent commits combined in this single PR, I'm not sure if this is acceptable by bcc project, but I'd rather see them as separated
> 

Hi, the first patch is a preparation for the RISC-V support. 
The second patch is can be dropped, it is used to exercise the first patch.

> 1. can you document the full command how to generate vmlinux.h? so people will do same update next time

```shell
sudo bpftool btf dump file /sys/kernel/btf/vmlinux format c > vmlinux.h
```

Not sure if it's worth documenting.

> 2. can we also follow the convention as other Architecture, to name the file as vmlinux_${version}.h and have a symbol link to vmlinux.h?
> 
> thanks, I really appreciate!

Updated.

davemarchevsky:
LGTM, but I have no access to a riscv system or VM, so can't validate the final commit.

Regardless, will merge and we can fix later if anything isn't right for riscv

davemarchevsky:
(FWIW multiple commits in a single PR is totally fine as long as they can be applied independently and each has a good summary)

PR #4094, opened 2022-07-06
sync libbpf

chantra:
sync with libbpf as of 2022-07-05


yonghong-song:
bpftool sync caused libbpf-tools compilation error. Let us have two patches: one for libbpf sync and the other for bpftool sync.

chantra:
> bpftool sync caused libbpf-tools compilation error. Let us have two patches: one for libbpf sync and the other for bpftool sync.

ok, I will limit that to libbpf, which is what I am the most interested about ATM.

PR #4101, opened 2022-07-11
[libbpf-tools] core_fixes: Add 'old' triple-under structs

davemarchevsky:
Currently, `core_fixes.bpf.h` helper fns which deal with structs whose
fields changed after a specific commit contain a triple-underscore
struct definition which contains the new field, e.g.

```
struct task_struct___x {
        unsigned int __state;
} __attribute__((preserve_access_index));
```

This allows the `bpf_core_field_exists` and `BPF_CORE_READ` calls
dealing with the new field to compile even if the `vmlinux.h` the user
is compiling against contains the 'old' struct definition (`state`
instead of `__state` in the above example).

If the `vmlinux.h` the user is compiling against has the 'new' struct
def, currently `core_fixes.bpf.h` compilation will fail because the old
field definition (`state`) does not exist. Let's fix this by also having
a `___old` struct def, and returning NULL or 0 if neither old or new
fields exist.

### DIFF
--- a/libbpf-tools/core_fixes.bpf.h
+++ b/libbpf-tools/core_fixes.bpf.h
@@ -13,17 +13,24 @@
  * see:
  *     https://github.com/torvalds/linux/commit/2f064a59a1
  */
-struct task_struct___x {
+struct task_struct___new {
 	unsigned int __state;
+} __attribute__((preserve_access_index));
+
+struct task_struct___old {
+	volatile long state;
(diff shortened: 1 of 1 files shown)


davemarchevsky:
@chenhengqi what do you think of this change?

chenhengqi:
I have changes similar to this in #4088 .

davemarchevsky:
Ah, excellent, I will close this in favor of that then

PR #4102, opened 2022-07-12
tools/profile: add multi process/thread support

yezhem:
Just like the perf tool, support multi process or thread may be useful.

### DIFF
--- a/tools/profile.py
+++ b/tools/profile.py
@@ -51,6 +51,13 @@ def positive_int(val):
         raise argparse.ArgumentTypeError("must be positive")
     return ival
 
+def positive_int_list(val):
+    vlist = val.split(",")
+    if len(vlist) <= 0:
+        raise argparse.ArgumentTypeError("must be an integer list")
+
+    return [positive_int(v) for v in vlist]
+
 def positive_nonzero_int(val):
     ival = positive_int(val)
     if ival == 0:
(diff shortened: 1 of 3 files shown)


yonghong-song:
Do you actually have a use case for this?

yezhem:
> 

Yes, but in a productive environment, can`t paste the use case example.

yezhem:
> Do you actually have a use case for this?

@yonghong-song Yes, but in a productive environment, can`t paste the use case example.

yonghong-song:
> > Do you actually have a use case for this?
> 
> @yonghong-song Yes, but in a productive environment, can`t paste the use case example.

okay. thanks for clarification.

tools/profile_example.txt
```diff
@@ -161,6 +161,35 @@ but we're lacking the symbol translation. This is a common for all profilers
 on Linux, and is usually fixable. See the DEBUGGING section of the profile(8)
 man page.
 
+You can also profile different process:
+# ./profile -p 2040 1316151
```
yonghong-song, line 165: should the option look like `-p 2040 -p 1316151`? What is the syntax `perf` allows multiple pid/tid's?

yezhem: perf just like: -p 2040,1316151 
should change to it?

yonghong-song: Yes, it would be good to following perf conventions.

yonghong-song: Please also mention perf convention in your commit message.

yezhem: resolved, pls review it

tools/profile.py
```diff
@@ -81,9 +88,9 @@ def stack_id_err(stack_id):
     formatter_class=argparse.RawDescriptionHelpFormatter,
     epilog=examples)
 thread_group = parser.add_mutually_exclusive_group()
-thread_group.add_argument("-p", "--pid", type=positive_int,
+thread_group.add_argument("-p", "--pid", type=positive_int_list,
     help="profile process with this PID only")
```
yonghong-song, line 92: could you change this help to 'profile process with one or more comma separated PIDs only'

tools/profile.py
```diff
@@ -81,9 +88,9 @@ def stack_id_err(stack_id):
...
-thread_group.add_argument("-L", "--tid", type=positive_int,
+thread_group.add_argument("-L", "--tid", type=positive_int_list,
     help="profile thread with this TID only")
```
yonghong-song, line 94: change this to 'prrofile thread with one or more comman separated TIDs only'.

yonghong-song (commented):
LGTM. Thanks. A couple of minor comments. Could you also update man page for '-p' and '-L' option?

tools/profile.py
yezhem: @yonghong-song Thanks for your review, pls check the modification.

PR #4108, opened 2022-07-15
docs:add install guide for wsl users

aseaday:


INSTALL.md
```diff
@@ -270,6 +271,33 @@ sudo docker run --rm -it --privileged \
   alpine:3.12
 ```
 
+## WSL - Binary
```
davemarchevsky, line 274: Can you put `(Windows Subsystem for Linux)` here as well? Will make this easier to search for

davemarchevsky (changes requested):
I don't have access to a WSL environment to validate, but happy to merge this in anyways if it worked for you. Just one nit that I think should be addressed

INSTALL.md
aseaday: thx! PTAL.

PR #4116, opened 2022-07-18
riscv:fix user_regs_struct macro issue

### LINKED ISSUE #4110: RISC-V: cachestat tool fails to run: "no member named 'pc' in 'struct pt_regs'"
```
root@riscv64-test:/# ./cachestat
/virtual/main.c:13:14: error: no member named 'pc' in 'struct pt_regs'
    key.ip = PT_REGS_IP(ctx);
             ^~~~~~~~~~~~~~~
/virtual/include/bcc/helpers.h:1376:29: note: expanded from macro 'PT_REGS_IP'
#define PT_REGS_IP(x) ((x)->pc)
                       ~~~  ^
1 error generated.
  File "/usr/share/bcc/tools/./cachestat", line 96, in <module>
    b = BPF(text=bpf_text)
  File "/usr/lib/python3/dist-packages/bcc/__init__.py", line 476, in __init__
```

System: QEMU VM with riscv64 CPU, OS: Ubuntu 22.04 with kernel 5.15.0-1014-generic.

If I understand it correctly, 'epc' should be used instead of 'pc', same as in arch/riscv/include/asm/ptrace.h.

```
diff --git a/src/cc/export/helpers.h b/src/cc/export/helpers.h
index 71ed8c8a..0f6e59ef 100644
--- a/src/cc/export/helpers.h
+++ b/src/cc/export/helpers.h
@@ -1373,7 +1373,7 @@ int bpf_usdt_readarg_p(int argc, struct pt_regs *ctx, void *buf, u64 len) asm("l
 #define PT_REGS_FP(x) ((x)->s0) /* Works only with CONFIG_FRAME_POINTER */
 #define PT_REGS_RC(x) ((x)->a0)
 #define PT_REGS_SP(x) ((x)->sp)
-#define PT_REGS_IP(x) ((x)->pc)
+#define PT_REGS_IP(x) ((x)->epc)
 #else
 #error "bcc does not support this platform yet"
 #endif
```

Right?
@xmzzz, what do you think?

xmzzz:
Fix #4110 . With this patch, running tools/cachestat on riscv64 can get the expected result.

```
$ sudo ./cachestat 
    HITS   MISSES  DIRTIES HITRATIO   BUFFERS_MB  CACHED_MB
       0        0        0    0.00%           52       5869
       0        0        0    0.00%           52       5869
       0        0        0    0.00%           52       5869
       0        0        0    0.00%           52       5869
       0        0        0    0.00%           52       5869
     618        0        0  100.00%           52       5869
       0        0        0    0.00%           52       5869
       0        0        0    0.00%           52       5869
       0        0        0    0.00%           52       5869
       0        0        0    0.00%           52       5869
       0        0        0    0.00%           52       5869
       0        0        0    0.00%           52       5869
       0        0        0    0.00%           52       5869
     607        0        0  100.00%           52       5869
       0        0        0    0.00%           52       5869
       0        0        0    0.00%           52       5869
       0        0        0    0.00%           52       5869
       0        0        0    0.00%           52       5869
       0        0        0    0.00%           52       5869
       0        0        0    0.00%           52       5869
^C       0        0        0    0.00%           52       5869
Detaching...
```

Signed-off-by: Mingzheng Xing <xingmingzheng@iscas.ac.cn>

### DIFF
--- a/src/cc/export/helpers.h
+++ b/src/cc/export/helpers.h
@@ -1363,17 +1363,19 @@ int bpf_usdt_readarg_p(int argc, struct pt_regs *ctx, void *buf, u64 len) asm("l
 #define PT_REGS_SP(x) ((x)->regs[29])
 #define PT_REGS_IP(x) ((x)->cp0_epc)
 #elif defined(bpf_target_riscv64)
-#define PT_REGS_PARM1(x) ((x)->a0)
-#define PT_REGS_PARM2(x) ((x)->a1)
-#define PT_REGS_PARM3(x) ((x)->a2)
-#define PT_REGS_PARM4(x) ((x)->a3)
-#define PT_REGS_PARM5(x) ((x)->a4)
-#define PT_REGS_PARM6(x) ((x)->a5)
-#define PT_REGS_RET(x) ((x)->ra)
-#define PT_REGS_FP(x) ((x)->s0) /* Works only with CONFIG_FRAME_POINTER */
-#define PT_REGS_RC(x) ((x)->a0)
-#define PT_REGS_SP(x) ((x)->sp)
-#define PT_REGS_IP(x) ((x)->pc)
+/* riscv64 provides struct user_pt_regs instead of struct pt_regs to userspace */
+#define __PT_REGS_CAST(x) ((const struct user_regs_struct *)(x))
+#define PT_REGS_PARM1(x) (__PT_REGS_CAST(x)->a0)
+#define PT_REGS_PARM2(x) (__PT_REGS_CAST(x)->a1)
+#define PT_REGS_PARM3(x) (__PT_REGS_CAST(x)->a2)
+#define PT_REGS_PARM4(x) (__PT_REGS_CAST(x)->a3)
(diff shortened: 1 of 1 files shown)


chenhengqi:
Looks like arm64 also need to be fixed.

euspectre:
@xmzzz I tried your patch in my VM - yes, `cachestat` runs now and seems to work OK.

RISC-V, Ubuntu 22.04, kernel 5.15.0-1014-generic.

Thank you for a quick fix!

xmzzz:
> Looks like arm64 also need to be fixed.

Yes, but I don't have an arm64 environment here, so I can't test and verify it. So wait for someone in need to fix it.
thanks. ^0^

xmzzz:
> @xmzzz I tried your patch in my VM - yes, `cachestat` runs now and seems to work OK.
> 
> RISC-V, Ubuntu 22.04, kernel 5.15.0-1014-generic.
> 
> Thank you for a quick fix!

Thanks for your feedback. ^0^

PR #4118, opened 2022-07-19
riscv: using bpf_probe_read_{kernel|user} on riscv64

xmzzz:
According to the bcc Reference Guide [[0]](https://github.com/iovisor/bcc/blob/master/docs/reference_guide.md#rewriter), for non-s390, the rewriter will use bpf_probe_read() for implicit memory access. This is because some existing users may have implicit memory accesses to access user memory, so using bpf_probe_read_kernel() will cause their applications to fail.

In riscv64, we should use the new bpf_probe_read_{kernel|user} function instead of the old bpf_probe_read(). riscv64 does not have overlapping address spaces, but does not have ARCH_HAS_NON_OVERLAPPING_ADDRESS_SPACE set like x86. And bpf_probe_read() is not available in riscv64, so it will cause an error.

This patch works fine to correctly use bpf_probe_read_{kernel|user}() on riscv64 until issues that may break existing user programs are resolved.

Now, running tools/cachestat on riscv64 can get the expected result.
```
$ sudo ./execsnoop
ls               2148    943       0 /usr/bin/ls --color=auto
ls               2151    943       0 /usr/bin/ls --color=auto
ls               2152    943       0 /usr/bin/ls --color=auto -a
ifconfig         2153    943       0 /usr/sbin/ifconfig
^C
```

[0] https://github.com/iovisor/bcc/blob/master/docs/reference_guide.md#rewriter

Signed-off-by: Mingzheng Xing <xingmingzheng@iscas.ac.cn>

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -336,7 +336,7 @@ ProbeVisitor::ProbeVisitor(ASTContext &C, Rewriter &rewriter,
   C(C), rewriter_(rewriter), m_(m), ctx_(nullptr), track_helpers_(track_helpers),
   addrof_stmt_(nullptr), is_addrof_(false) {
   const char **calling_conv_regs = get_call_conv();
-  has_overlap_kuaddr_ = calling_conv_regs == calling_conv_regs_s390x;
+  cannot_fall_back_safely = (calling_conv_regs == calling_conv_regs_s390x || calling_conv_regs == calling_conv_regs_riscv64);
 }
 
 bool ProbeVisitor::assignsExtPtr(Expr *E, int *nbDerefs) {
@@ -509,7 +509,7 @@ bool ProbeVisitor::VisitUnaryOperator(UnaryOperator *E) {
   memb_visited_.insert(E);
   string pre, post;
   pre = "({ typeof(" + E->getType().getAsString() + ") _val; __builtin_memset(&_val, 0, sizeof(_val));";
-  if (has_overlap_kuaddr_)
+  if (cannot_fall_back_safely)
     pre += " bpf_probe_read_kernel(&_val, sizeof(_val), (u64)";
   else
     pre += " bpf_probe_read(&_val, sizeof(_val), (u64)";
@@ -573,7 +573,7 @@ bool ProbeVisitor::VisitMemberExpr(MemberExpr *E) {
   string base_type = base->getType()->getPointeeType().getAsString();
   string pre, post;
   pre = "({ typeof(" + E->getType().getAsString() + ") _val; __builtin_memset(&_val, 0, sizeof(_val));";
-  if (has_overlap_kuaddr_)
+  if (cannot_fall_back_safely)
     pre += " bpf_probe_read_kernel(&_val, sizeof(_val), (u64)&";
   else
     pre += " bpf_probe_read(&_val, sizeof(_val), (u64)&";
@@ -627,7 +627,7 @@ bool ProbeVisitor::VisitArraySubscriptExpr(ArraySubscriptExpr *E) {
     return true;
 
   pre = "({ typeof(" + E->getType().getAsString() + ") _val; __builtin_memset(&_val, 0, sizeof(_val));";
-  if (has_overlap_kuaddr_)
+  if (cannot_fall_back_safely)
     pre += " bpf_probe_read_kernel(&_val, sizeof(_val), (u64)((";
   else
     pre += " bpf_probe_read(&_val, sizeof(_val), (u64)((";
@@ -724,7 +724,7 @@ DiagnosticBuilder ProbeVisitor::error(SourceLocation loc, const char (&fmt)[N]) 
 BTypeVisitor::BTypeVisitor(ASTContext &C, BFrontendAction &fe)
     : C(C), diag_(C.getDiagnostics()), fe_(fe), rewriter_(fe.rewriter()), out_(llvm::errs()) {
   const char **calling_conv_regs = get_call_conv();
-  has_overlap_kuaddr_ = calling_conv_regs == calling_conv_regs_s390x;
+  cannot_fall_back_safely = (calling_conv_regs == calling_conv_regs_s390x || calling_conv_regs == calling_conv_regs_riscv64);
 }
 
 void BTypeVisitor::genParamDirectAssign(FunctionDecl *D, string& preamble,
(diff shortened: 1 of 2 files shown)


euspectre:
I tested a similar approach (see https://github.com/iovisor/bcc/pull/4085#issuecomment-1179190706 and follow-ups) and it should work.

However, the way `bpf_probe_read()` functions are used in bcc became not that clear over time. I suggested a more invasive solution that could make the code easier to understand: https://github.com/iovisor/bcc/pull/4109.

There are some concerns based on the issue from 2020, however:
https://github.com/iovisor/bcc/pull/4109#issuecomment-1186953758
https://github.com/iovisor/bcc/pull/4109#issuecomment-1186964849

If these concerns are still valid, the solution by @xmzzz  seems preferable, of course.
If no tool should use implicit `bpf_probe_read*` to access user data via pointers from the kernel structures, I'd suggest to straighten up the usage of `bpf_probe_read*` in bcc. The code would be easier to maintain and understand that way.

xmzzz:
> I tested a similar approach (see [#4085 (comment)](https://github.com/iovisor/bcc/pull/4085#issuecomment-1179190706) and follow-ups) and it should work.

Wow, that's true. Only the variable names are different.

I am still new to bcc and I'm having some confusion in understanding the issues related to bpf_probe_read() that I overlooked it was from your suggestion, sorry about that.

> However, the way `bpf_probe_read()` functions are used in bcc became not that clear over time. I suggested a more invasive solution that could make the code easier to understand: #4109.
> 
> There are some concerns based on the issue from 2020, however: [#4109 (comment)](https://github.com/iovisor/bcc/pull/4109#issuecomment-1186953758) [#4109 (comment)](https://github.com/iovisor/bcc/pull/4109#issuecomment-1186964849)
> 
> If these concerns are still valid, the solution by @xmzzz seems preferable, of course. On the other hand, if no tool should use implicit `bpf_probe_read*` calls to access user data via pointers from the kernel structures, I'd suggest to straighten up the usage of `bpf_probe_read*` in bcc. The code would be easier to maintain and understand that way.


Yes, I agree. If these concerns have been invalidated, maybe it's time to straighten out the use of bpf_probe_read* in bcc.

I have learnt so much, thanks to you @euspectre .

yonghong-song:
The patch looks good to me. 
I do agree that we should improve rewriter to detect to use bpf_probe_read_kernel() or to use bpf_probe_read_user(). clang15 and kernel >= 5.18 started to support btf_type_tags.
   https://lore.kernel.org/r/20220127154600.652613-1-yhs@fb.com

From bcc perspective, the bpf program could annotate necessary memory access with __user attribute so rewriter can detect such attribute in btf and then use bpf_probe_read_user() properly. This will requires clang >= 15. Also, This should be enabled only if the kernel support bpf_probe_read_kernel() and bpf_probe_read_user().

yonghong-song:
The following is an example about using btf_type_tag
```
[yhs@devbig309.ftw3 ~/tmp1] cat t.c                                                                  
#define __user __attribute__((btf_type_tag("user")))                                                 
int foo(int __user *p) {                                                                             
        return *p;
}
[yhs@devbig309.ftw3 ~/tmp1] clang -c -O2 -g -Xclang -ast-dump t.c
TranslationUnitDecl 0x8451648 <<invalid sloc>> <invalid sloc>
|-TypedefDecl 0x8451e70 <<invalid sloc>> <invalid sloc> implicit __int128_t '__int128'
| `-BuiltinType 0x8451c10 '__int128'
|-TypedefDecl 0x8451ee0 <<invalid sloc>> <invalid sloc> implicit __uint128_t 'unsigned __int128'
| `-BuiltinType 0x8451c30 'unsigned __int128'
|-TypedefDecl 0x84521e8 <<invalid sloc>> <invalid sloc> implicit __NSConstantString 'struct __NSConstantString_tag'
| `-RecordType 0x8451fc0 'struct __NSConstantString_tag'
|   `-Record 0x8451f38 '__NSConstantString_tag'
|-TypedefDecl 0x8452280 <<invalid sloc>> <invalid sloc> implicit __builtin_ms_va_list 'char *'
| `-PointerType 0x8452240 'char *'
|   `-BuiltinType 0x84516f0 'char'
|-TypedefDecl 0x8452578 <<invalid sloc>> <invalid sloc> implicit __builtin_va_list 'struct __va_list_tag[1]'
| `-ConstantArrayType 0x8452520 'struct __va_list_tag[1]' 1 
|   `-RecordType 0x8452360 'struct __va_list_tag'
|     `-Record 0x84522d8 '__va_list_tag'
`-FunctionDecl 0x84b02a0 <t.c:2:1, line:4:1> line:2:5 foo 'int (int  btf_type_tag(user)*)'
  |-ParmVarDecl 0x84b01a8 <col:9, col:21> col:21 used p 'int  btf_type_tag(user)*'
  `-CompoundStmt 0x84b0408 <col:24, line:4:1>
    `-ReturnStmt 0x84b03f8 <line:3:2, col:10>
      `-ImplicitCastExpr 0x84b03e0 <col:9, col:10> 'int btf_type_tag(user)':'int' <LValueToRValue>
        `-UnaryOperator 0x84b03c8 <col:9, col:10> 'int btf_type_tag(user)':'int' lvalue prefix '*' cannot overflow
          `-ImplicitCastExpr 0x84b03b0 <col:10> 'int  btf_type_tag(user)*' <LValueToRValue>
            `-DeclRefExpr 0x84b0390 <col:10> 'int  btf_type_tag(user)*' lvalue ParmVar 0x84b01a8 'p'  'int  btf_type_tag(user)*'
```
You can see:
```
  |-ParmVarDecl 0x84b01a8 <col:9, col:21> col:21 used p 'int  btf_type_tag(user)*'
```
The parameber 'p' type is 'int  btf_type_tag(user)*'. The rewriter could parse the type info to get whether the pointer pointing to user memory or not. Of course, we need to actually implement in rewriter.

Also, explicitly annotate user memory access with __user is a good thing for bpf program.

xmzzz:
> The patch looks good to me. I do agree that we should improve rewriter to detect to use bpf_probe_read_kernel() or to use bpf_probe_read_user(). clang15 and kernel >= 5.18 started to support btf_type_tags. https://lore.kernel.org/r/20220127154600.652613-1-yhs@fb.com
> 
> From bcc perspective, the bpf program could annotate necessary memory access with __user attribute so rewriter can detect such attribute in AST and then use bpf_probe_read_user() properly. This will requires clang >= 15. Also, This should be enabled only if the kernel support bpf_probe_read_kernel() and bpf_probe_read_user().

Thank you for the information and an example, I will learn about it.  ^0^/

dlan17:
@yonghong-song

 hey, with this patch merged, the BCC tool is kind of usable for RISC-V 64 platform. do you have any estimated date that a new version release will come out? from downstream OS distribution's perspective, we're unsure to backport patches or wait for new release..

PR #4119, opened 2022-07-19
tools/runqslower.py: remove unused global variable rq.

linuszeng:
Global variable rq is unused, remove it.

### DIFF
--- a/tools/runqlat.py
+++ b/tools/runqlat.py
@@ -86,8 +86,6 @@ typedef struct pidns_key {
 BPF_HASH(start, u32);
 STORAGE
 
-struct rq;
-
 // record enqueue timestamp
 static int trace_enqueue(u32 tgid, u32 pid)
 {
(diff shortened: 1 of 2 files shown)


davemarchevsky (approved):
LGTM

Looks like `runqlat.py` also has a `struct rq;` fwd declaration while including `sched.h`. Can you remove it there too?

Also, curious: did you use some automated tool to notice this, or just eyeballs?

linuszeng:
> LGTM
> 
> Looks like `runqlat.py` also has a `struct rq;` fwd declaration while including `sched.h`. Can you remove it there too?
> 
> Also, curious: did you use some automated tool to notice this, or just eyeballs?

Fix the problem in 6703af11c7ecd51cb41002c62b0aff27a952186c. In the follow-up work, I will use static program analysis tools to scan for some defects on bcc. Thank you for your suggestions.

davemarchevsky (approved):

PR #4122, opened 2022-07-20
Move BPF UAPI import into `bcc_bpf_uapi_inc.h`

chantra:
and conditionally include the headers either from linux-headers or from libbpf source if
`HAVE_EXTERNAL_LIBBPF` is defined.

### DIFF
--- a/src/cc/bcc_bpf_uapi_inc.h
+++ b/src/cc/bcc_bpf_uapi_inc.h
@@ -1,0 +1,11 @@
+#pragma once
+
+#ifdef HAVE_EXTERNAL_LIBBPF
+# include <bpf/uapi/linux/bpf.h>
+# include <bpf/uapi/linux/bpf_common.h>
+# include <bpf/uapi/linux/btf.h>
+#else
+# include <linux/bpf.h>
+# include <linux/bpf_common.h>
+# include <linux/btf.h>
+#endif
--- a/src/cc/api/BPF.cc
+++ b/src/cc/api/BPF.cc
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 
-#include <linux/bpf.h>
 #include <linux/perf_event.h>
 #include <unistd.h>
 #include <cstdio>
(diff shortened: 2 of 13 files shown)


chenhengqi:
What's the purpose of this change ?

libbpf UAPI headers are not installed by default.

src/cc/bcc_bpf_uapi_inc.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#ifdef HAVE_EXTERNAL_LIBBPF
+# include <bpf/uapi/linux/bpf.h>
+# include <bpf/uapi/linux/bpf_common.h>
+# include <bpf/uapi/linux/btf.h>
```
yonghong-song, line 6: Is it possible that we have a cmake option like BCC_EXT_LIBBPF_INC_PREFIX for example
   BCC_EXT_LIBBPF_INC_PREFIX = "bpf/uapi"
so we can replace the above hardcoded prefix 'bpf/uapi' with the one passed from cmake system?

chantra:
> What's the purpose of this change ?

Fair point, I did a poor job at explaining the background here.

In our use-case, we get bcc without submodules and compile against libbpf's copy of the uapi headers given that the one from the platform could be very outdated.
The intent of this diff is to provide a way to compile bcc against those headers. Currently, we are rewriting the path to those headers from the sources. The goal was to make it transparent by mean of a define.

> 
> libbpf UAPI headers are not installed by default.

Ha, I had missed that. So `HAVE_EXTERNAL_LIBBPF` is prob not enough in itself and should probably be more something along the line:
```
if defined(HAVE_EXTERNAL_LIBBPF) && defined(USE_LIBBPF_UAPI)
...
...
```


FWIW, we have a way to patch our local cmake file (which is already heavily patched) to make this transparent, so I am happy to drop this if not deemed useful.

yonghong-song:
> FWIW, we have a way to patch our local cmake file (which is already heavily patched)
> to make this transparent, so I am happy to drop this if not deemed useful.

let us drop this patch then.

chantra:
dropping

PR #4124, opened 2022-07-22
tools/biosnoop: add a forgotten “else”

WuXianChangKuang:
“blk_start_request” and “blk_mq_start_request” should be chosen between the two.


chenhengqi:
I am not familiar with the block subsystem, but I don't see the two code paths overlap, they both have `trace_block_rq_issue`.

chenhengqi:
After reading the related kernel [patches](https://patchwork.kernel.org/project/linux-scsi/cover/20181031175922.8849-1-axboe@kernel.dk/), I think this change is incorrect.

Prior to kernel 5.0, there are two kinds of block drivers (single queue and multiple queue), so the original implementation is to handle these two scenarios.

The `if` here is to check that whether single queue related symbols exist.

Maybe we should use the block tracepoints here.

cc @yonghong-song

PR #4135, opened 2022-07-28
libbpf-tools: tcpconnect: take source port into consideration

aspsk:
Originally, the tcpconnect utility didn't display the source port when tracing
events and ignored it (intentionally) when counting new connections. Add a new
option -s (or --source-port) to display the source port when tracing, or to use
the source port as part of the key when counting connections. This option is
unset by default to provide the original output.


davemarchevsky:
LGTM. Do you think you may want to filter by source port at some point, similarly to dest port filtering this tool already has?

aspsk:
> Do you think you may want to filter by source port at some point, similarly to dest port filtering this tool already has?

I am not sure, as this tool is about looking at outgoing connections, so the source port is, typically, random. Will add, for sure, if there will be a use case

PR #4142, opened 2022-08-01
tools/mdflush: include blkdev.h instead of genhd.h

jeromemarchand:
In recent kernels, i.e. since commit 322cbb50de71 ("block: remove
genhd.h"), genhd.h header has been removed and its content moved to
blkdev.h. Since genhd.h has been included in blkdev.h since forever,
including blkdev instead of genhd in the mdflush tool works for both
older and newer kernel.

### DIFF
--- a/tools/mdflush.py
+++ b/tools/mdflush.py
@@ -19,7 +19,7 @@ from time import strftime
 bpf_text="""
 #include <uapi/linux/ptrace.h>
 #include <linux/sched.h>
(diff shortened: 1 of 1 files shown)


davemarchevsky:
You weren't kidding - `blkdev.h` has included `genhd.h` since the earliest git commit.

Thanks for including commit which removed it to make this easy to verify. LGTM.

PR #4144, opened 2022-08-02
tools/memleak: Fix concurrency issue caused by duplicate addresses in multiple processess

xingfeng2510:
An address may be used simultaneously by multiple processes, so itself may not be unique key for map 'allocs'. This patch try to fix it by combining pid with address as a unique key, please take a look, thanks. @yonghong-song

### DIFF
--- a/tools/memleak.py
+++ b/tools/memleak.py
@@ -131,6 +131,11 @@ if command is not None:
 bpf_source = """
 #include <uapi/linux/ptrace.h>
 
+struct alloc_key_t {
+    u64 pid;
+    u64 address;
+};
+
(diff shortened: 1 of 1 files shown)


yonghong-song:
Which command line you are using to trigger this issue (the same address used by multiple processes)? If I understand correctly, for memleak, without `-p <pid>`, kernel functions will be traced. With `-p <pid>` a single process will be traced. Did I miss anything?

xingfeng2510:
Sorry, I got it wrong. Tracing a single user process seems reasonable, there is no concurrency issue in such a situation. I closed this PR, thanks.

PR #4147, opened 2022-08-04
oomkill: Remove trailing newline from output

chenhengqi:

### DIFF
--- a/libbpf-tools/oomkill.c
+++ b/libbpf-tools/oomkill.c
@@ -78,11 +78,11 @@ static void handle_event(void *ctx, int cpu, void *data, __u32 data_sz)
 	strftime(ts, sizeof(ts), "%H:%M:%S", tm);
 
 	if (n)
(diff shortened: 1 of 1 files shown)


libbpf-tools/oomkill.c
```diff
@@ -78,11 +78,11 @@ static void handle_event(void *ctx, int cpu, void *data, __u32 data_sz)
 	strftime(ts, sizeof(ts), "%H:%M:%S", tm);
 
 	if (n)
-		printf("%s Triggered by PID %d (\"%s\"), OOM kill of PID %d (\"%s\"), %lld pages, loadavg: %s\n",
+		printf("%s Triggered by PID %d (\"%s\"), OOM kill of PID %d (\"%s\"), %lld pages, loadavg: %s",
```
yonghong-song, line 81: why removing '\n' here? The else branch still has '\n'. it seems '\n' makes sense since different oom's will be in different lines?

chenhengqi: The if branch print `/proc/loadavg` which has a newline already.

PR #4151, opened 2022-08-08
Add systemtap header and ruby to fedora docker

davemarchevsky:
The fedora USDT tests are failing because 1) ruby isn't installed on fedora by default, and 2) fedora's ruby doesn't (didn't) have USDTs when it was installed by default. So let's manually build it with USDTs.


davemarchevsky:
Goal of this PR was to get ubuntu and fedora tests failing for same reason, instead of fedora failing for USDT as well, so LGRM

PR #4152, opened 2022-08-09
tools/syscount: Use lock_xadd to guarantee atomicity of addition operations

xingfeng2510:
As libbpf-tools/syscount, use lock_xadd to guarantee atomicity of addition operations (without -P or -p, we can not guarantee atomicity).


yonghong-song:
LGTM

PR #4159, opened 2022-08-11
tools/execsnoop: Add PPID filter support

xingfeng2510:
- I have a agent which may invoke various external tools. Sometimes, I just want to trace children processes of this agent process.
- Only trace new processes of a bash process may be useful (keep an eye on command aliais).

```
# ./execsnoop.py -P $(pgrep -nx bash)
ls               884360  874065    0 /usr/bin/ls --color=auto
ip               884395  874065    0 /usr/sbin/ip a
ls               884427  874065    0 /usr/bin/ls --color=auto -l --color=auto
```

This patch try to add an option -P to do this, please take a look.


yonghong-song:
LGTM. Agree that this might be useful to trace children processes of a particular parent process.

PR #4163, opened 2022-08-11
[test][usdt_probes][bionic] kill child manually

chantra:
bionic has util-linux 2.31, `unshare` added support for `--kill-child` in 2.32.

This change remove the use of the argument `--kill-child` and instead kills the
child process manually before returning. While this is not going to take care
of all corner cases, it is probably more than enough for when the test is succeeding.


davemarchevsky:
Can you rebase this on latest master? Let's see what test state looks like

chantra:
did so, the unshare issue is not happening anymore, but there is still 1 test which is failing the overall run. this is unrelated to this change though.

PR #4164, opened 2022-08-11
[test][sockhash/map] Fix the test to work with kernels >= 5.15

chantra:
Those tests have started to fail since kernel 5.15.
The restriction was lifted in https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git/commit/?id=0c48eefae712c2fd91480346a07a1a9cd0f9470b

This diff makes the expected returned value to the call to `update_value` conditional
on the kernel version.

Tested on 5.15 (using a Ubuntu 22.04 host), which is representative of the kernel running in GH CI.
Also tested on Ubuntu 20.04 stock kernel:

```
$ docker run -ti \
                    --privileged \
                    --network=host \
                    --pid=host \
                    -v $(pwd):/bcc \
                    -v /sys/kernel/debug:/sys/kernel/debug:rw \
                    -v /lib/modules:/lib/modules:ro \
                    -v /usr/src:/usr/src:ro \
                    -e CTEST_OUTPUT_ON_FAILURE=1 \
                    bcc-docker-focal \
                    /bin/bash -c \
                    '/bcc/build/tests/wrapper.sh \
                        c_test_all sudo /bcc/build/tests/cc/test_libbcc "test sock*"'
===============================================================================
All tests passed (8 assertions in 2 test cases)

[22:40:55] chantra@focal:bcc git:(fix_sock_map_tests*) $ uname -a
Linux focal 5.4.0-122-generic #138-Ubuntu SMP Wed Jun 22 15:00:31 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,6 +97,10 @@ CMAKE_DEPENDENT_OPTION(ENABLE_CPP_API "Enable C++ API" ON "ENABLE_USDT" OFF)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 
+if(ENABLE_TESTS)
+  find_package(KernelHeaders)
+endif()
+
 if (CMAKE_USE_LIBBPF_PACKAGE)
   find_package(LibBpf)
 endif()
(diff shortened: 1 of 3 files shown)


chantra:
Hum... it seems that the version number is not properly detected when compiling. I am going to mock around with the test for a bit.

chantra:
This is good for review again.

davemarchevsky:
nice!

PR #4167, opened 2022-08-12
test_bpf_table: Re-add deletion of 'bpf' module

davemarchevsky:
This was removed in 6616092255f80fc0918e19e898f0abfe2e3c2840, presumably
because sanitizer complained.


chantra:
My mistake. I should have read the comment above the `delete bpf` to catch that the test was purposedly destroying the object and testing access to the map after that.

PR #4168, opened 2022-08-12
Revert "[test] Fix Wrapper script to properly handle arguments"

davemarchevsky:
This reverts commit 770a590e879dcc831bb68c4cc2b3fc1c1ee81674.

Now that "cc tests" are passing, the python section of "all tests" is
failing. Those tests run using wrapper.sh, so let's try reverting some
recent changes to see if that helps.


davemarchevsky:
@chantra the second revert results in all the py tests succeeding except `py_test_tools_smoke`, which is a better state than before

chantra:
I have a fix forward  with explanations in #4169.
From looking at the failing test (in bionic) I have a gut feeling that it is some python version difference.

PR #4173, opened 2022-08-13
update swapin.py

egobillot:
update swapin.py
 - measure time spent during the interval
 - add the swapout metrics

### DIFF
--- a/tools/swapin.py
+++ b/tools/swapin.py
@@ -4,8 +4,6 @@
 # swapin        Count swapins by process.
 #               For Linux, uses BCC, eBPF. Embedded C.
 #
-# TODO: add -s for total swapin time column (sum)
-#
 # Copyright (c) 2019 Brendan Gregg.
 # Licensed under the Apache License, Version 2.0 (the "License").
 # This was originally created for the BPF Performance Tools book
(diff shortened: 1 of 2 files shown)


yonghong-song:
The code change mostly fine. But since swapout is not implemented in the original code, I guess swapout in general is not that important. You should in the commit message to describe why you need swapout information.

Also, the tools name is called `swapin.py`. It would be odd to add swapout support in swapin.py. Maybe we should rename it as swap.py? If swapout is less useful (means useful in occasional case), an option is probably still necessary to control swapout.

If we indeed go with swap.py, if it is necessary to keep swapin.py, we can have swapin.py to
be alias to swap.py with necessary argument including disabling swapout.

cc @brendangregg

egobillot:
I have added the -o flag to enable or not the swap out metrics. 
I agree the tool is called swapin.py and if we keep the swap out  metrics it needs to be renamed. I agree with you proposition (to keep swapin.py and to add swap.py).

Yes the main information is the swap **in** to identify which process is slow down disk accesses. My point of view about the swap out information is that it will bring the complete view about the swap activities. 
I will let you decide what to do. No problem to remove the swap out code if you think those metrics are not really important.

PR #4176, opened 2022-08-14
tools/trace: Decode bytes to str and remove explicit ctype cast

xingfeng2510:
- remove b'' by decoding all field values whose format specifier are 's'.
- remove explicit ctype cast.

```
# ./trace.py 'c:open "%s, %d", arg1, arg2'

318915  318919  grpc_global_tim open             b'/etc/resolv.conf', 0
318915  318919  grpc_global_tim open             b'/etc/nsswitch.conf', 0
318915  318919  grpc_global_tim open             b'/etc/hosts', 0
```

### DIFF
--- a/tools/trace.py
+++ b/tools/trace.py
@@ -321,44 +321,6 @@ class Probe(object):
                     expr = expr.replace(alias, replacement)
                 return expr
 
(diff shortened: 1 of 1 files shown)


yonghong-song:
LGTM. Thanks for the cleanup.

PR #4177, opened 2022-08-14
GH Actions: Stop publishing to quay.io

davemarchevsky:
This hasn't succeeded in quite some time. Let's turn it off for now.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -7,66 +7,6 @@ on:
   pull_request:
 
 jobs:
-  publish_images:
-    # Optionally publish container images, guarded by the GitHub secret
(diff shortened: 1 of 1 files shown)


chantra:
This would be interesting to get working though. It would seriously speed up the CI by essentially skipping the build container step, which last I checked was about 6 min.

davemarchevsky:
@chantra there's a "Publish to Dockerhub" step which is likely doing the same thing and has working creds. Neither `quay.io` nor `dockerhub` publish steps have been truly maintained for a while and consistently make `master` commit signal red.

I'm totally on board for building container off master when new commit is added, then using that for PR tests. But until that's working I'd rather have master commit signal be useful.

PR #4179, opened 2022-08-15
Build CI containers

chantra:
Currently, it takes between 7 and 8 minutes to go through the step "Build docker container with all deps", which is time spent in setting up the CI essentially.
This step is quite long because on top of pulling packages, we also build ruby.

We can greatly reduce this time by using pre-made container images that our CI job will just pull with all deps installed.

The result can be seen in https://github.com/chantra/bcc/tree/faster_ci .
Building and publishing the containers workflow run : https://github.com/chantra/bcc/actions/runs/2862888407
and then a subsequent CI run pulling those containers straight from the registry: https://github.com/chantra/bcc/runs/7844899211?check_suite_focus=true

End result, the step to get our CI containers is in the 30 or 40 seconds. And there is also ways to make this even smaller by making the containers smaller.

This is the first patches in a series of 2 to achieve this.

With those patches I:
- move the docker files used for building CI containers to their own subdir (docker/build) and rename the file `Dockerfile.test` to `Dockerfile.ubuntu`, which will make the workflow logic easier.
- Create a new workflow that will build new containers images:
  - daily to ensure that we have an up to date container
  - on pushes to `master` to only create new images that align with master.
  
  
  This PR will be followed by another PR which will start using those images as demonstrated in https://github.com/chantra/bcc/runs/7844899211?check_suite_focus=true

PR #4180, opened 2022-08-15
[ci] make build CI use docker images from this repo registry

chantra:
Pull CI containers from GHCR and use them during build.

Example workflow run: https://github.com/chantra/bcc/runs/7844899499?check_suite_focus=true
Logs of the pull step: https://gist.github.com/chantra/b6a13223d530696bb00d40ce58c45fdb#file-gistfile1-txt-L540

depends on #4179

### DIFF
--- a/.github/workflows/bcc-test.yml
+++ b/.github/workflows/bcc-test.yml
@@ -6,12 +6,18 @@ on:
       - master
   pull_request:
 
+env:
+  # Use docker.io for Docker Hub if empty
+  REGISTRY: ghcr.io
+  # github.repository as <account>/<repo>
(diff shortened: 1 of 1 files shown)


chantra:
Currently contains all commits including the ones from #4179

Only the latest commit is needed.

chantra:
TWith #4191 fixed, I think this is good to go. The outcome from this patch is that we will reduce the CI full runs by 10 mines ( ~27 min down to ~18 min). Build/bcc tests feedback will be within 5 minutes instead of 15 minutes.

yonghong-song:
cc @davemarchevsky

PR #4186, opened 2022-08-17
Show the stacktrace of the userspace in opensnoop.

ThinkerYzu1:
With new option '-c', opensnoop will show the names and line numbers
of the two most inner callers in the userspace if there are.  It
uses BlazeSym to do symbolization.

Signed-off-by: Kui-Feng Lee <kuifeng@fb.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,14 +23,32 @@ endif()
 
 enable_testing()
 
+execute_process(COMMAND git config --global --add safe.directory ${CMAKE_CURRENT_SOURCE_DIR}
+                WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+                RESULT_VARIABLE CONFIG_RESULT)
+if(CONFIG_RESULT AND NOT CONFIG_RESULT EQUAL 0)
+  message(WARNING "Failed to add root source directory to safe.directory")
+endif()
+
+# populate submodule blazesym
+if(NOT NO_BLAZESYM)
+    execute_process(COMMAND git config --global --add safe.directory ${CMAKE_CURRENT_SOURCE_DIR}/libbpf-tools/blazesym
+                    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+                    RESULT_VARIABLE CONFIG_RESULT)
+    if(CONFIG_RESULT AND NOT CONFIG_RESULT EQUAL 0)
+      message(WARNING "Failed to add blazesym source directory to safe.directory")
+    endif()
+
+    execute_process(COMMAND git submodule update --init --recursive -- libbpf-tools/blazesym
+                    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+                    RESULT_VARIABLE UPDATE_RESULT)
+    if(UPDATE_RESULT AND NOT UPDATE_RESULT EQUAL 0)
+      message(WARNING "Failed to update submodule blazesym")
+    endif()
+endif()
+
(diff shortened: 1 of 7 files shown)


ThinkerYzu1:
The output of `opensnoop -c`  looks like
```
2669466 python3.8         -1   2 /etc/userdb
                                 files_add
                                 conf_files_list_strv_internal
2669466 python3.8         -1   2 /run/userdb
                                 files_add
                                 conf_files_list_strv_internal
2669466 python3.8         -1   2 /run/host/userdb
                                 files_add
                                 conf_files_list_strv_internal
```

It shows the two most inner callers of the syscall if there are.

davemarchevsky:
Probably the best way to solve the "dubious ownership" issue will be to modify `CMakeLists.txt` similar to what `libbpf` and `bpftool` do here: https://github.com/iovisor/bcc/blob/master/CMakeLists.txt#L34 . A little confusing perhaps since `libbpf-tools` dir is not using the `cmake` build system like rest of `bcc` repo.

Speaking of makefiles, it would be good to gate the `blazesym` build behind some flag and have that propagated to the `opensnoop.c` userspace runner, so that `-c` flag and added functionality aren't available if `blazesym` isn't built. Totally fine to have the `USE_BLAZESYM` or whatever on by default, but want to give folks an out if they don't want to have `rustc`/`cargo` dependency.  In long run if we migrate to blazesym by default we can just remove it.

ThinkerYzu1:
The error now is caused by the absence of cargo.  How can I make cargo/rustc available?

CMakeLists.txt
```diff
@@ -23,14 +23,32 @@ endif()
 
 enable_testing()
 
+# populate submodule blazesym
+if(NOT NO_BLAZESYM)
+    execute_process(COMMAND git config --global --add safe.directory ${CMAKE_CURRENT_SOURCE_DIR}
```
davemarchevsky, line 28: This `execute_process` statement and 

> Failed to add root source directory to safe.directory"

warning check should happen for both blazesym and libbpf submodule usage. I think it's fine for it to live outside of both `if(NOT NO_BLAZESYM)` and `if(NOT CMAKE_USE_LIBBPF_PACKAGE)` checks.

CMakeLists.txt
```diff
@@ -23,14 +23,32 @@ endif()
...
+                    RESULT_VARIABLE CONFIG_RESULT)
+    if(CONFIG_RESULT AND NOT CONFIG_RESULT EQUAL 0)
+      message(WARNING "Failed to add blazesym source directory to safe.directory")
+    endif()
+
+    execute_process(COMMAND git submodule update --init --recursive -- libbpf-tools/blazesym
+                    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+                    RESULT_VARIABLE UPDATE_RESULT)
+    if(UPDATE_RESULT AND NOT UPDATE_RESULT EQUAL 0)
+      message(WARNING "Failed to update submodule libbpf")
```
davemarchevsky, line 46: can you update the warning text here?

davemarchevsky (changes requested):

davemarchevsky:
re: rust and cargo, can you add this change to `docker/build/Dockerfile.fedora`:

```
+++ b/docker/build/Dockerfile.fedora
@@ -41,6 +41,10 @@ RUN dnf -y install \
        python3 \
        python3-pip

+RUN dnf -y install \
+       rust \
+       cargo
+
 RUN if [[ ! -e /usr/bin/python && -e /usr/bin/python3 ]]; then \
```

PR #4187, opened 2022-08-17
libbpf-tools/klockstat: do not account locks with missing try_at

bobrik:
This change prevents confusion from reporting unrealistic values:

```
              down_write_killable+0x5    5.4 m      5421   491.2 h      491.2 h
                do_mprotect_pkey+0xb2
              __x64_sys_mprotect+0x1b
                   do_syscall_64+0x3b
  entry_SYSCALL_64_after_hwframe+0x44
                              Max PID 1983644, COMM edgeworker/sbox
```

### DIFF
--- a/libbpf-tools/klockstat.bpf.c
+++ b/libbpf-tools/klockstat.bpf.c
@@ -175,6 +175,10 @@ static void account(struct lockholder_info *li)
 			bpf_get_current_comm(ls->acq_max_comm, TASK_COMM_LEN);
 	}
 
+	if (!li->try_at) {
+		return;
+	}
+
 	delta = li->acq_at - li->try_at;
 	__sync_fetch_and_add(&ls->acq_count, 1);
 	__sync_fetch_and_add(&ls->acq_total_time, delta);


bobrik:
cc @brho, @namhyung

libbpf-tools/klockstat.bpf.c
```diff
@@ -175,6 +175,10 @@ static void account(struct lockholder_info *li)
 			bpf_get_current_comm(ls->acq_max_comm, TASK_COMM_LEN);
 	}
 
+	if (!li->try_at) {
+		return;
+	}
```
namhyung, line 180: I'm curious how it cannot have try_at.  The `li` is only added in `lock_contended()` which set the try_at before adding it.

bobrik: I'm not sure, but the output from the description isn't even the most wild. Here's another:

```
$ sudo /tmp/klockstat -d 10 -c down_ -n 3 -s 10 -S hld_total -L 0xffff90296c51f768
Tracing mutex/sem lock events...  Hit Ctrl-C to end
 
 
              down_write_killable+0x5  945.1 h      5421 5123604.5 h  5123604.5 h
                do_mprotect_pkey+0xb2
              __x64_sys_mprotect+0x1b
                   do_syscall_64+0x3b
  entry_SYSCALL_64_after_hwframe+0x44
                              Max PID 1983644, COMM edgeworker/sbox
```

There's no way average wait was 945.1h when the measurement only went on for 10s.

namhyung: Could you please add a comment at least that we don't know how it can happen but it caused incorrect results?

brho: definitely weird, and i'm worried there's some other bug at play here.  

in both examples, the Max Wait == Total Wait.  i guess that'd due to rounding: the total is probably just a little more than the max, and the gigantic max dominates.

also, the Count is 5421, which seems odd that you had the same count twice.

also, even if we say "try_at" is wrong, the acq_at is garbage too: 5 million hours is > 500 years.

unlikely, but maybe bpf_ktime_get_ns() returned 0?  can you try and catch that with a bpf_printk or something?

they are both down_write_killable too - perhaps there's something wrong with that in the lock_aborted case?  i can imagine there being something that could corrupt li->acq_at (if there's a weird scenario where the same task mucks with the same lock?), but li->try_at should be set before it's ever in the map and never modified afterwards.

another option: since you can recreate this, can you add a little debugging to lockholder_info?  essentially it's a state machine, where we go from 
    lock_contended .-> lock_acquired -> lock_released -> deleted
                              |
                              \_->lock_aborted -> deleted

and maybe we can catch an unexpected path. e.g. we call lock_released when the tracking code thought it hadn't been acquired yet.  not that that would trigger this bug, but it'd be a clue as to what is wrong here.  we could also detect if a lock is contended while it is acquired (which would violate the uniqueness of lockholder_info).

i'm always slightly worried that we have more than one call path using a single lockholder_info, though you'd need the same pid and lock struct for that.  i thought that couldn't happen, but you never know.  though even if that happened, i still don't see why try_at == 0.

if that's a huge pain, let me know.  i might be able to put together the debug infrastructure, and then you can try it out.  alternatively, if you can give me a way to recreate this easily, i can try it on my end.

bobrik: I won't have time to look into this for a couple of weeks, but I can definitely try to catch this when I get back. We altered the code that was causing too much contention on `mmap_lock`, so this might not be so easily reproducible anymore, but there's no way to tell without trying.

There's no rush not merge this in the meantime.

bobrik: We rearchitected our problematic `mmap_sem` usage leading to excessive contention and the issue is no longer easily reproducible.

Let's close this for now.

PR #4192, opened 2022-08-22
tools/opensnoop: Allow to show full path for an open file with relative path

xingfeng2510:
List some practical scenarios from my daily work:
- When a process open an file with relative path, I also need to execute **lsof** or **ls -l /proc/pid/cwd** to view current directory (for example: locate MySQL or PostgreSQL config files), so it may be convenient to show full path for an open file with relative path.
- A process open an non-existent config file with relative path, then exits directly. Leave me no chance to view current directory of it.
- A process may change its current directory, and open files with relative path. Only show relative path could not identify full path.

Try to add option -F to solve this, please take a look, thanks. @yonghong-song 

For example:

shell 1:
```
./opensnoop.py -n cat -F
```

shell 2:
```
cd /root/bpf
cat example.c
```

shell 1 shows a full path: /root/bpf/example.c:
```
373114 cat                 3   0 /root/bpf/example.c
```

without -F, just show relative path: example.c

### DIFF
--- a/tools/opensnoop.py
+++ b/tools/opensnoop.py
@@ -4,7 +4,9 @@
 # opensnoop Trace open() syscalls.
 #           For Linux, uses BCC, eBPF. Embedded C.
 #
-# USAGE: opensnoop [-h] [-T] [-x] [-p PID] [-d DURATION] [-t TID] [-n NAME]
+# USAGE: opensnoop [-h] [-T] [-U] [-x] [-p PID] [-t TID]
+#                  [--cgroupmap CGROUPMAP] [--mntnsmap MNTNSMAP] [-u UID]
+#                  [-d DURATION] [-n NAME] [-F] [-e] [-f FLAG_FILTER]
 #
 # Copyright (c) 2015 Brendan Gregg.
 # Licensed under the Apache License, Version 2.0 (the "License")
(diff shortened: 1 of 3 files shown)


yonghong-song:
could you rebase on top of master and resubmit? There is a conflict with master branch.

tools/opensnoop.py
```diff
@@ -312,6 +338,41 @@
...
+    bpf_text = bpf_text.replace('SUBMIT_DATA', """
+    data.type = EVENT_ENTRY;
+    events.perf_submit(ctx, &data, sizeof(data));
+
+    if (data.name[0] != '/') { // relative path
+        struct task_struct *task;
+        struct dentry *dentry;
+        int i;
+
+        task = (struct task_struct *)bpf_get_current_task();
```
yonghong-song, line 352: I tried on my local environment and hit the following verifier failure.
```
78: (15) if r1 == 0x2f goto pc+28
 R0_w=inv(id=0) R1_w=inv(id=0,umax_value=255,var_off=(0x0; 0xff)) R6=ctx(id=0,off=0,imm=0) R7_w=ctx(id=0,off=0,imm=0) R8=inv(id=0) R9=invP0 R10=fp0 fp-8=mmmmmmmm fp-16=mmmmmmmm fp-24=mmmmmmmm fp-32=mmmmmmmm fp-40=mmmmmmmm fp-48=mmmmmmmm fp-56=mmmmmmmm fp-64=mmmmmmmm fp-72=mmmmmmmm fp-80=mmmmmmmm fp-88=mmmmmmmm fp-96=mmmmmmmm fp-104=mmmmmmmm fp-112=mmmmmmmm fp-120=mmmmmmmm fp-128=mmmmmmmm fp-136=mmmmmmmm fp-144=mmmmmmmm fp-152=mmmmmmmm fp-160=mmmmmmmm fp-168=mmmmmmmm fp-176=mmmmmmmm fp-184=mmmmmmmm fp-192=mmmmmmmm fp-200=mmmmmmmm fp-208=mmmmmmmm fp-216=mmmmmmmm fp-224=mmmmmmmm fp-232=mmmmmmmm fp-240=mmmmmmmm fp-248=mmmmmmmm fp-256=mmmmmmmm fp-264=mmmmmmmm fp-272=mmmmmmmm fp-280=mmmmmmmm fp-288=mmmmmmmm fp-296=mmmmmmmm fp-304=mmmmmmmm fp-312=fp fp-320=mmmmmmmm fp-328=mmmmmmmm
; task = (struct task_struct *)bpf_get_current_task();
79: (85) call bpf_get_current_task#35
80: (b7) r8 = 30
; dentry = task->fs->pwd.dentry;
81: (79) r1 = *(u64 *)(r0 +1880)
R0 invalid mem access 'inv'
```

Let us use `bpf_get_current_task_btf()` so it will work with recent kernels.

xingfeng2510: done

xingfeng2510:
Modify as suggested. Please take a look again, thanks.

Issue #3834, opened 2022-02-02
memleak: could not determine address of symbol malloc


hidva:
```
/usr/bin/python ../tools/memleak.py  -c "/root/hidva.com/rustplay2/target/debug/rustplay2" -O /lib64/libc-2.33.so
Executing '/root/project/rustplay2/target/debug/rustplay2' and tracing the resulting process.
Attaching to pid 964501, Ctrl+C to quit.
Traceback (most recent call last):
  File "/root/project/bcc/build/../tools/memleak.py", line 449, in <module>
    attach_probes("malloc")
  File "/root/project/bcc/build/../tools/memleak.py", line 437, in attach_probes
    bpf.attach_uprobe(name=obj, sym=sym,
  File "/root/project/bcc/build/src/python/bcc-python3/bcc/__init__.py", line 1361, in attach_uprobe
    (path, addr) = BPF._check_path_symbol(name, sym, addr, pid, sym_off)
  File "/root/project/bcc/build/src/python/bcc-python3/bcc/__init__.py", line 960, in _check_path_symbol
    raise Exception("could not determine address of symbol %s" % symname)
Exception: could not determine address of symbol b'malloc'
```

# Root Cause

rustplay2 only runs for 50ms, When `openelf(path="/proc/964798/root/lib64/libc-2.33.so")` executes, the process 964798 is terminated, so `/proc/964798/root` is cleared.

```
$ ls /proc/964798/root
ls: cannot read symbolic link '/proc/964798/root': No such file or directory
/proc/964798/root
$ ps -o pid,cmd 964798
    PID CMD
 964798 [rustplay2] <defunct>
```

hidva:
Should we suspend the process until all initialization is completed?

bacher09:
It's a behaviour you would get with any short running process, it's probably more reliable to attach uprobe before launching an application, but there is an issue with this approach, since application might load library from another path (that's why BCC is checking `/proc/$pid/maps`).

PR #4195, opened 2022-08-26
tools/memleak: support user-space tracing without a PID

ghedo:
Sometimes it is useful to attach user probes before starting a process
(e.g. see https://github.com/iovisor/bcc/issues/3834), however memleak
currently requires a PID in order to enable user-space probing.

This change adds a new option `--kernel-trace` to explicitly enable
kernel tracing, and allow user-space tracing without providing a PID.

In order to do this, a stack's PID need to be stored in the BPF map, in
order to be able to later resolve symbols correctly.

### DIFF
--- a/tools/memleak.py
+++ b/tools/memleak.py
@@ -5,7 +5,7 @@
 #
 # USAGE: memleak [-h] [-p PID] [-t] [-a] [-o OLDER] [-c COMMAND]
 #                [--combined-only] [--wa-missing-free] [-s SAMPLE_RATE]
-#                [-T TOP] [-z MIN_SIZE] [-Z MAX_SIZE] [-O OBJ]
+#                [-T TOP] [-z MIN_SIZE] [-Z MAX_SIZE] [-O OBJ] [--percpu] [-K]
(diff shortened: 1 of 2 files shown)


yonghong-song:
> This change adds a new option --kernel-trace to explicitly enable
kernel tracing, and allow user-space tracing without providing a PID.

This is a behavior change as kernel tracing now needs explicit option.
We cannot do this.

What is your expected command line for user space tracing?

ghedo:
The way I've been running this is something like `memcheck --obj /path/to/binary` and once the probes are attached I actually run `/path/to/binary`.

In order to avoid breaking previous use cases I guess I could add a `--user-trace` option rather than `--kernel-trace`, and then have user-space tracing enabled when either a PID is available, or `--user-trace` is passed. This way existing use cases should still work. Thoughts?

terryryan1431 (changes requested):
People are scandless

yonghong-song:
I think adding a `--user-trace` option is better. This way, we can maintain the backward compatibility.

Issue #4182, opened 2022-08-16
compactsnoop fails to compile on Fedora36/LLVM14


chantra:
When trying to run the compactsnoop test on Fedora, it fails with:
```
$ docker run -ti \                     
                    --privileged \
                    --network=host \
                    --pid=host \
                    -v $(pwd):/bcc \
                    -v /sys/kernel/debug:/sys/kernel/debug:rw \
                    -v /lib/modules:/lib/modules:ro \
                    -v /usr/src:/usr/src:ro \
                    -e CTEST_OUTPUT_ON_FAILURE=1 \
                    bcc-docker-f36 \
                    /bin/bash -c \
                    '/bcc/build/tests/wrapper.sh \
                        "py_test_map_in_map" "sudo"  "/bcc/tools/compactsnoop.py"'
Error at line 82: Unsupport signed division for DAG: 0x55c28143eb00: i64 = sdiv exact 0x55c28143ea30, Constant:i64<1472>, /virtual/main.c:82:17 @[ /virtual/main.c:159:17 @[ /virtual/main.c:192:5 ] ]Please convert to unsigned div/mod.
LLVM ERROR: Cannot select: 0x55c28143eb00: i64 = sdiv exact 0x55c28143ea30, Constant:i64<1472>, /virtual/main.c:82:17 @[ /virtual/main.c:159:17 @[ /virtual/main.c:192:5 ] ]
  0x55c28143ea30: i64 = sub 0x55c281fb0290, 0x55c28143ec38, /virtual/main.c:82:17 @[ /virtual/main.c:159:17 @[ /virtual/main.c:192:5 ] ]
    0x55c281fb0290: i64,ch = CopyFromReg 0x55c281c492f8, Register:i64 %0, /virtual/main.c:168:25
      0x55c281fb0a48: i64 = Register %0
    0x55c28143ec38: i64,ch = load<(dereferenceable load (s64) from %ir.2, !tbaa !2015)> 0x55c28143e890, FrameIndex:i64<0>, undef:i64, /virtual/main.c:82:19 @[ /virtual/main.c:159:17 @[ /virtual/main.c:192:5 ] ]
      0x55c28143e3b0: i64 = FrameIndex<0>
      0x55c281fb0568: i64 = undef
  0x55c28143ea98: i64 = Constant<1472>
In function: raw_tracepoint__mm_compaction_suitable
[...]

PR #4196, opened 2022-08-26
Fix a llvm signed division error for compactsnoop tool

yonghong-song:
Fix issue #4182.
llvm doesn't support signed division and an assertion error
will happen when the IR contains sdiv. The reason is
due to code
  zone - zone_pgdat->node_zones
where zone and zone_pgdat->node_zones are pointers to
struct zone (with size 1664). So effectively the above
is equivalent to
  ((void *)zone - (void *)zone_pgdat->node_zones)/sizeof(struct zone)
which is converted to sdiv insn.
llvm11 seems okay and I didn't investigate why. But llvm14 and
latest llvm16 failed with compiler fatal error.

To fix the issue let us replace the above '(void *)' as
'(u64)' and then the udiv will be used for the division.


chantra:
Thanks @yonghong-song

PR #4202, opened 2022-08-29
libbpf-tools: Update bpftool submodule

chenhengqi:
Closes #4200.

Signed-off-by: Hengqi Chen <chenhengqi@outlook.com>

PR #4204, opened 2022-08-29
[python][test] change traced syscall from clone to execve

### LINKED ISSUE #4203: test_percpu failing on Fedora 36
As part of updating what version of Fedora BCC can be tested against, it appears that `test_percpu.py` is failing on Fedora 36:
https://gist.github.com/chantra/d37f3daa969bdbc07862dc11f8384a38

The underlying reason is that the test expects calls to `clone`, but on Fedora 36, vfork is called instead:

```
[root@940cbfdf11e3 /]# strace -f python3 -c 'import os; os.popen("hostname");' 2>&1 | egrep '(clone|fork)'
vfork(strace: Process 46 attached
[pid    45] <... vfork resumed>)        = 46
```

Python version is:
```
[root@940cbfdf11e3 /]# python3 -V
Python 3.10.6
```

[`os.popen` uses `subprocess.Popen` under the hood](https://docs.python.org/3/library/os.html#os.popen)

This is due to a change in 3.10: https://docs.python.org/3/whatsnew/changelog.html#id150

> [bpo-35823](https://bugs.python.org/issue?@action=redirect&bpo=35823): Use vfork() instead of fork() for [subprocess.Popen()](https://docs.python.org/3/library/subprocess.html#subprocess.Popen) on Linux to improve performance in cases where it is deemed safe.

Per

> [gh-91401](https://github.com/python/cpython/issues/91401): Provide a fail-safe way to disable [subprocess](https://docs.python.org/3/library/subprocess.html#module-subprocess) use of vfork() via a private subprocess._USE_VFORK attribute. While there is currently no known need for this, if you find a need please only set it to False. File a CPython issue as to why you needed it and link to that from a comment in your code. This attribute is documented as a footnote in 3.11.

it can be disabled:

```
[root@940cbfdf11e3 /]# strace -f python3 -c 'import os; import subprocess; subprocess._USE_VFORK = False; os.popen("hostname");' 2>&1 | egrep '(clone|fork)'
clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLDstrace: Process 67 attached
[root@940cbfdf11e3 /]# strace -f python3 -c 'import os; import subprocess; os.popen("hostname");' 2>&1 | egrep '(clone|fork)'
vfork(strace: Process 73 attached
[pid    72] <... vfork resumed>)        = 73
```

regardless, this either need to be gated based on python version, which is sub-optimal and add more conditionals to the code... or use another syscall.

I am suggesting tracing `execve` instead.

chantra:
Fixes #4203

The issue goes into more details as to why this change is needed. The TL;DR is
that the implementation of `os.popen` change somewhere between python 3.9 and
3.10 and depending on which version of python is used, either `fork/clone` is
called, or `vfork`.

In both cases, we are going to end up calling `execve`. This change switches
the syscall traced from `clone` to `execve`.

At the same time, it is now a constant that can be used across the tests.

Test plan:
Before https://gist.github.com/chantra/d37f3daa969bdbc07862dc11f8384a38
After https://gist.github.com/chantra/2860e9812f00eaa0758ccdb5d3192689

### DIFF
--- a/tests/python/test_percpu.py
+++ b/tests/python/test_percpu.py
@@ -7,6 +7,8 @@ import unittest
 from bcc import BPF
 import multiprocessing
 
+MONITORED_SYSCALL="execve"
+
 class TestPercpu(unittest.TestCase):
 
     def setUp(self):
@@ -38,7 +40,7 @@ class TestPercpu(unittest.TestCase):
         """
         bpf_code = BPF(text=test_prog1)
         stats_map = bpf_code.get_table("stats")
-        event_name = bpf_code.get_syscall_fnname("clone")
+        event_name = bpf_code.get_syscall_fnname(MONITORED_SYSCALL)
         bpf_code.attach_kprobe(event=event_name, fn_name="hello_world")
         ini = stats_map.Leaf()
         for i in range(0, multiprocessing.cpu_count()):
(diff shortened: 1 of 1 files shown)


chantra:
The Fedora 34 failure would be fixed once #4202 is landed.

PR #4207, opened 2022-08-29
[test][python] temptatively attach to blk_start_request probe.

chantra:
`blk_start_request` is gone since kernel 5. This patch is in the same vein as https://github.com/iovisor/bcc/pull/4124
except that we only conditionally attach to `blk_start_request` but attach also attach to
`blk_mq_start_request`.

### DIFF
--- a/tests/python/test_trace3.py
+++ b/tests/python/test_trace3.py
@@ -16,11 +16,13 @@ if len(sys.argv) > 1:
 
 
 class TestBlkRequest(TestCase):
-    @mayFail("This fails on github actions environment, and needs to be fixed")
     def setUp(self):
         b = BPF(arg1, arg2, debug=0)
         self.latency = b.get_table("latency", c_uint, c_ulong)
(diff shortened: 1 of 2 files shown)


chantra:
Mind that this is currently based off #4204 .

@chenhengqi , @yonghong-song I could fix #4124 at the same time (e.g remove the conditional).

chenhengqi:
As I commented in #4124, I think we should trace both `blk_mq_start_request` and `blk_start_request`.

chantra:
> As I commented in #4124, I think we should trace both `blk_mq_start_request` and `blk_start_request`.

I am happy to fix this as part of this diff. @yonghong-song WDYT?

chantra:
@chenhengqi , I checked with Jens offline and he confirmed what you said. e.g for 4.x, a driver would either be using blk-mq or the legacy stack.

I have updated this PR with a revert of #4124.

PR #4211, opened 2022-09-01
[ci] Install netperf and iperf

chantra:
Some tests need those binaries to run. Currently those tests are allowed to
fail and we are possibly losing signal.
Once those packages are in the container image, we will be able to enforce
running those test and having them succeed.


chantra:
Added a commit on top to build (but not push) containers when a file in `docker/build` is modified. This will insure that changes to docker files are not breaking.

PR #4216, opened 2022-09-04
Use /sys/kernel/tracing when debugfs is not available

michalgr:
Tracefs might be available as /sys/kernel/tracing and not /sys/kernel/debug/tracing. This commit makes bcc check whether /sys/kernel/debug/tracing exists and use /sys/kernel/tracing if not.

### DIFF
--- a/src/cc/common.cc
+++ b/src/cc/common.cc
@@ -211,6 +211,19 @@ static inline field_kind_t _get_field_kind(std::string const& line,
   return field_kind_t::regular;
 }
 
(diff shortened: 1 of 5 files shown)


SeeFlowerX:
should we change here too ?
- https://github.com/iovisor/bcc/blob/f90126bb3770ea1bdd915ff3b47e451c6dde5c40/libbpf-tools/trace_helpers.c#L1065

chenhengqi:
@SeeFlowerX  Could you help with a new PR ?

PR #4219, opened 2022-09-06
tools/opensnoop: Allow to set size of the perf ring buffer

xingfeng2510:
open() syscalls often invoke at very high frequency in a production system. The fixed size (64 pages) of the perf ring buffer is not enough, which causes the output mixed with some warns, "Possibly lost xxx samples". Add option -b to allow to set size of the perf ring buffer. So we can use large buffer size to avoid event lost.

```
Possibly lost 2 samples
Possibly lost 1 samples
Possibly lost 3 samples
...
Possibly lost 1 samples
Possibly lost 1 samples
```

### DIFF
--- a/tools/opensnoop.py
+++ b/tools/opensnoop.py
@@ -7,6 +7,7 @@
 # USAGE: opensnoop [-h] [-T] [-U] [-x] [-p PID] [-t TID]
 #                  [--cgroupmap CGROUPMAP] [--mntnsmap MNTNSMAP] [-u UID]
 #                  [-d DURATION] [-n NAME] [-F] [-e] [-f FLAG_FILTER]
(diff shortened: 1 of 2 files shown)


brendangregg:
Thanks, I'm leaving a comment here for future perf ring buffer issues.

Ideally we do not add "-b"/"--buffer-pages" to individual tools but instead improve the buffering in the bcc library. See #1033.

xingfeng2510:
> Thanks, I'm leaving a comment here for future perf ring buffer issues.
> 
> Ideally we do not add "-b"/"--buffer-pages" to individual tools but instead improve the buffering in the bcc library. See #1033.

Thanks for your advice, I'll keep up with it.

PR #4221, opened 2022-09-06
Conditionally rebuild container

chantra:
When any Dockerfile is modified, we should rebuild the container before running the CI.
Failing to do so, we would need to:
1) land the docker file change in order to incorporate the change
2) land the charge that uses this new dockerfile change

This is cumbersome and it is hard to make sure that changes to the dockerfile are indeed correct.

This change conditionally pull the pre-built image if no dockerfile change happened. Otherwise
it reuses the "build_container" action to build the image.

A first commit will confirm that the old behaviour is unchanged. A subsequent commit will confirm that when a Dockerfile is modified, we rebuild the container.

### DIFF
--- a/.github/actions/build-container/action.yml
+++ b/.github/actions/build-container/action.yml
@@ -1,0 +1,47 @@
+name: "Build/Push container"
+description: "Build a BCC CI container and push it when not a pull-request."
+
+inputs:
+  os_distro:
+    description: "OS Disctribution. Ex: ubuntu"
+    required: true
+  os_version:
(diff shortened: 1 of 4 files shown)


chantra:
I believe test are currently broken due to  https://github.com/iovisor/bcc/commit/5acd0b5c29272ba6fc321dcb16292ee0ea4d4549#diff-f10061225cb5f9e51f4c93214b7cd3e62fea9697c3fa0d7ab7edf254e4ebadf2

A PR on my branch works just fine: https://github.com/chantra/bcc/pull/1

PR #4238, opened 2022-09-16
fix undefined reference reallocarray

maliubiao:
symbol reallocarray does not exists in ubuntu 16.04,   unable to make the project.

### DIFF
--- a/libbpf-tools/klockstat.c
+++ b/libbpf-tools/klockstat.c
@@ -496,6 +496,16 @@ static void print_hld_task(struct stack_stat *ss)
 	       print_time(tot, sizeof(tot), ss->ls.hld_total_time));
 }
 
+static inline void *reallocarray_handwrite(void *ptr, size_t nmemb, size_t size)
+{
+        size_t total;
+        if (size == 0 || nmemb > ULONG_MAX / size)
+                return NULL;
+        total = nmemb * size;
+        return realloc(ptr, total);
+}
+
+
 static int print_stats(struct ksyms *ksyms, int stack_map, int stat_map)
 {
 	struct stack_stat **stats, *ss;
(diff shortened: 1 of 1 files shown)


chenhengqi:
Don't think `klockstat` can run on Ubuntu 16.04, what's your kernel version ?

maliubiao:
> Don't think `klockstat` can run on Ubuntu 16.04, what's your kernel version ?

I'm trying to install a new kernel for ubuntu 16.04 to use the latest features of ebpf, but I found a compilation problem, if the new kernel can be installed successfully, this patch is useful, at least there are no compilation errors

chenhengqi:
So, instead of making this change, please try upgrade your kernel first.

maliubiao:
This is not a kernel issue, the "reallocarray" function did not exist in glibc before 2017, please refer to this link, https://github.com/bminor/glibc/commit/ 2e0bbbfbf95fc9e22692e93658a6fbdd2d4554da

yonghong-song:
actually libbpf has the same issue where reallocarray is not available in certain cases. So libbpf has the following implementation
```
/*
 * Re-implement glibc's reallocarray() for libbpf internal-only use.
 * reallocarray(), unfortunately, is not available in all versions of glibc,
 * so requires extra feature detection and using reallocarray() stub from
 * <tools/libc_compat.h> and COMPAT_NEED_REALLOCARRAY. All this complicates
 * build of libbpf unnecessarily and is just a maintenance burden. Instead,
 * it's trivial to implement libbpf-specific internal version and use it
 * throughout libbpf.           
 */
static inline void *libbpf_reallocarray(void *ptr, size_t nmemb, size_t size)
{
        size_t total;

#if __has_builtin(__builtin_mul_overflow)
        if (unlikely(__builtin_mul_overflow(nmemb, size, &total)))
                return NULL;
#else
        if (size == 0 || nmemb > ULONG_MAX / size)
                return NULL;
        total = nmemb * size;
#endif
        return realloc(ptr, total);
}
```

Unfortunately, the function is in `libbpf_internal.h` so we cannot use other this won't work if link with external libbpf libraries.

So I think this patch is okay.

PR #4242, opened 2022-09-19
Fix the uid not defined issue

wenlxie:


chenhengqi (approved):
LGTM, thanks for the fix.

PR #4250, opened 2022-09-26
Mini debug info

michalgr:
Add support for mini debug info.

This pull request adds support for mini debug info:
https://sourceware.org/gdb/onlinedocs/gdb/MiniDebugInfo.html

Some libraries contain a .gnu_debugdata section that stores lzma
compressed so file with additional symbol info. This pull request
integrates liblzma and teaches bcc how to use it to extract that
information whenever debug file is looked up.


yonghong-song:

src/cc/bcc_elf.c
```diff
@@ -646,6 +682,64 @@ static int find_debug_via_symfs(Elf *e, const char *path,
...
+
+  stream.next_in = gnu_debugdata;
+  stream.avail_in = gnu_debugdata_size;
+  stream.avail_out = 0;
+
+  while (ret == LZMA_OK && stream.avail_in > 0) {
+    if (stream.avail_out < LZMA_MIN_BUFFER_SIZE) {
+      decompressed_buffer_size += LZMA_MIN_BUFFER_SIZE;
+      decompressed = realloc(decompressed, decompressed_buffer_size);
+      if (decompressed == NULL) {
```
davemarchevsky, line 712: if `realloc` fails it will return `NULL` but won't free the buffer passed in in 1st arg, so you'll need to do some pointer juggling to free the original buffer here

src/cc/bcc_elf.c
```diff
@@ -47,19 +47,44 @@ static int openelf_fd(int fd, Elf **elf_out) {
...
 
-static int openelf(const char *path, Elf **elf_out, int *fd_out) {
-  *fd_out = open(path, O_RDONLY);
-  if (*fd_out < 0)
+struct elf_file {
+  Elf *elf;
+  int fd;
+};
+
+#define ELF_FILE_INIT \
```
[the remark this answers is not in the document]
michalgr: I made it a proper function.

src/cc/bcc_elf.c
```diff
@@ -55,21 +55,30 @@ struct elf_file {
 #define ELF_FILE_INIT \
   { .elf = NULL, .fd = -1 }
 
+static int elf_file_adapt_fd(int fd, struct elf_file *out) {
```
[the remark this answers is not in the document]
michalgr: Renamed. There is one `openelf_fd` usage (`bcc_elf_foreach_vdso_sym`) that does not care about closing the file descriptor, I keep `openelf_fd` call there. However, it appears that there is a `Elf` leak there which I am fixing in https://github.com/iovisor/bcc/pull/4250/commits/c5e8fb9a3fa1e6791c5ab503ef6c86df4937e421

cmake/FindLibLzma.cmake
```diff
@@ -0,0 +1,45 @@
+# - Try to find liblzma
+# Once done this will define
+#
+#  LIBLZMA_FOUND - system has liblzma
+#  LIBLZMA_INCLUDE_DIRS - the liblzma include directory
+#  LIBLZMA_LIBRARIES - Link these to use liblzma
+
+if (LIBLZMA_LIBRARIES AND LIBLZMA_INCLUDE_DIRS)
+    set (LibLzma_FIND_QUIETLY TRUE)
```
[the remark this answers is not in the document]
michalgr: I am a noob too :) Does anything check whether user provided `LIBLZMA_{LIBRARIES,INCLUDE_DIRS}` work ? I think when those variables are set by a user then `find_path` and `find_library` don't do anything and the purpose of setting quiet is to suppress messages like [this one](https://github.com/Kitware/CMake/blob/343685869716b2ac31876d3e4173ece46ea49efc/Modules/FindPackageHandleStandardArgs.cmake#L581).

I am not going to lie, changes in this file (including this if) are heavily inspired by other cmake files in this (and bpftrace) repos, like [FindLibElf.cmake](https://github.com/iovisor/bcc/blob/master/cmake/FindLibElf.cmake).

src/cc/bcc_elf.c
```diff
@@ -629,102 +649,107 @@ static char *find_debug_via_symfs(Elf *e, const char* path) {
...
-        find_debug_file(elf_file.elf, path, option->check_debug_file_crc);
-    if (debug_file) {
-      foreach_sym_core(debug_file, callback, callback_lazy, option, payload, 1);
-      free(debug_file);
+  if (option->use_debug_file) {
+    if (find_debug_file(elf_file.elf, path, option->check_debug_file_crc,
+                        &debug_elf_file) == 0) {
+      listsymbols(debug_elf_file.elf, callback, callback_lazy, payload, option,
+                  1);
+      bcc_elf_file_close(&debug_elf_file);
```
chenhengqi, line 861: Return ?

chenhengqi, line 861: Return ?

michalgr: That's not how it used to work. I am worried that some symbols might be defined only in the non-debug file.

chenhengqi: OK, I misread this.

davemarchevsky (approved):

PR #4252, opened 2022-09-27
tools: killsnoop: support target PID filter

pizhenwei:
Support '-t'/'--tpid' to filter the target PID to avoid message flood, ignore other processes we don't care.

### DIFF
--- a/man/man8/killsnoop.8
+++ b/man/man8/killsnoop.8
@@ -2,7 +2,7 @@
 .SH NAME
 killsnoop \- Trace signals issued by the kill() syscall. Uses Linux eBPF/bcc.
 .SH SYNOPSIS
-.B killsnoop [\-h] [\-x] [-p PID]
+.B killsnoop [\-h] [\-x] [-p PID] [-T PID]
 .SH DESCRIPTION
 killsnoop traces the kill() syscall, to show signals sent via this method. This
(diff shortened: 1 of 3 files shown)


pizhenwei:
Hi @devidasjadhav 
Thanks for your suggestion, force pushed a new versoin.

man/man8/killsnoop.8
```diff
@@ -27,7 +27,10 @@ Print usage message.
 Only print failed kill() syscalls.
 .TP
 \-p PID
-Trace this process ID only (filtered in-kernel).
+Trace this process ID only which is the sender of signal (filtered in-kernel).
+.TP
+\-t PID
+Trace this target process ID only which is the receiver of signal (filtered in-kernel).
```
yonghong-song, line 33: -t is commonly used for tid. So I would like this option to be -r/--rpid to imply it is for signal receiver.

pizhenwei: what about '-T/-tpid'?

yonghong-song: '-T/-tpid' is okay to me.

pizhenwei:
@yonghong-song 
Use '-T' instead of '-t', also append another change: auto detect the length of PID/TPID field.
I noticed that the master branch also hits auto test failed...

PR #4259, opened 2022-09-30
softirqs: Uniform capitalization of strings

Rtoax:
Before:

```bash
$ sudo ./softirqs.py  1 2 -C
Tracing soft irq events... Hit Ctrl-C to end.

SOFTIRQ          TOTAL_count
net_rx                     1
hi                        74
timer                    124
rcu                      178
sched                    194
```
After:
```bash
[rongtao@RT-NUC tools]$ sudo ./softirqs.py  1 2 -C
Tracing soft irq events... Hit Ctrl-C to end.

SOFTIRQ          TOTAL_COUNT
net_rx                     2
hi                        77
timer                    119
rcu                      122
sched                    188
```


chenhengqi:
It seems that these labels exists since the first day of these tools, I'd suggest just keep it as it is.

Rtoax:
Based on compatibility, you're right, but it still looks weird

PR #4262, opened 2022-10-02
libbpf-tools: Add support for bpf_ringbuf

chenhengqi:
This patch introduces bpf_ringbuf to the existing libbpf-tools. Currently, most tools use PERF_EVENT_ARRAY for event-based tracing. For newer kernels, we could use bpf_ringbuf instead, for better performance. In this commit, we introduce two new header files: compat.bpf.h and compat.h, for BPF programs and userspace program respectively, providing uniform APIs for perfbuf/ringbuf handling in a CO-RE way.

### DIFF
--- a/libbpf-tools/compat.bpf.h
+++ b/libbpf-tools/compat.bpf.h
@@ -1,0 +1,45 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+/* Copyright (c) 2022 Hengqi Chen */
+
+#ifndef __COMPAT_BPF_H
+#define __COMPAT_BPF_H
+
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+
+#define MAX_EVENT_SIZE		10240
+#define RINGBUF_SIZE		(1024 * 256)
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__uint(max_entries, 1);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, MAX_EVENT_SIZE);
+} heap SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_RINGBUF);
+	__uint(max_entries, RINGBUF_SIZE);
+} events SEC(".maps");
+
+static __always_inline void *reserve_buf(__u64 size)
+{
+	static const int zero = 0;
+
+	if (bpf_core_type_exists(struct bpf_ringbuf))
+		return bpf_ringbuf_reserve(&events, size, 0);
+
+	return bpf_map_lookup_elem(&heap, &zero);
+}
+
+static __always_inline long submit_buf(void *ctx, void *buf, __u64 size)
+{
+	if (bpf_core_type_exists(struct bpf_ringbuf)) {
+		bpf_ringbuf_submit(buf, 0);
+		return 0;
+	}
+
+	return bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, buf, size);
+}
+
+#endif /* __COMPAT_BPF_H */
--- a/libbpf-tools/compat.c
+++ b/libbpf-tools/compat.c
@@ -1,0 +1,115 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+/* Copyright (c) 2022 Hengqi Chen */
+
+#include "compat.h"
+#include "trace_helpers.h"
+#include <stdlib.h>
+#include <errno.h>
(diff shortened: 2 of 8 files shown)


chenhengqi:
@anakryiko Could you please help code reviewing this PR, thanks.

libbpf-tools/mountsnoop.bpf.c
```diff
@@ -57,16 +46,17 @@ static int probe_exit(void *ctx, int ret)
...
-	int zero = 0;
+	struct event *eventp;
+	struct arg *argp;
 
 	argp = bpf_map_lookup_elem(&args, &tid);
 	if (!argp)
 		return 0;
 
-	eventp = bpf_map_lookup_elem(&heap, &zero);
+	bpf_map_delete_elem(&args, &tid);
```
anakryiko, line 57: you shouldn't do this, you are deleting element that you are going to use afterwards. while nothing will crash, the same memory can be reused for another element, do a proper goto cleanup pattern

libbpf-tools/mountsnoop.c
```diff
@@ -277,6 +288,14 @@ int main(int argc, char **argv)
...
 	obj->rodata->target_pid = target_pid;
 
+	use_ringbuf = probe_ringbuf();
+	if (use_ringbuf) {
+		bpf_map__set_autocreate(obj->maps.events_perfbuf, false);
+		bpf_map__set_autocreate(obj->maps.heap, false);
+	} else {
+		bpf_map__set_autocreate(obj->maps.events_ringbuf, false);
+	}
+
```
anakryiko, line 298: I think this should be hidden inside bpf_buffer__new(), why each app should do this over and over again?

libbpf-tools/mountsnoop.c
```diff
@@ -289,11 +308,15 @@ int main(int argc, char **argv)
...
 
-	pb = perf_buffer__new(bpf_map__fd(obj->maps.events), PERF_BUFFER_PAGES,
-			      handle_event, handle_lost_events, NULL, NULL);
-	if (!pb) {
+	if (use_ringbuf)
+		buf = bpf_buffer__new(obj->maps.events_ringbuf,
+				      handle_ringbuf_event, NULL, NULL, NULL);
+	else
+		buf = bpf_buffer__new(obj->maps.events_perfbuf,
+				      handle_perfbuf_event, handle_lost_events, NULL, NULL);
```
anakryiko, line 316: similarly, why these differences, just make bpf_buffer__new() accept unified callback (no cpu arg, decide on int/void return result) and optional lost_events (which won't be ever called for ringbuf, unless you add atomic counters in compat.bpf.c, which might be not too bad of an idea). Handle such differences inside the compat code, basically

libbpf-tools/compat.bpf.h
```diff
@@ -0,0 +1,51 @@
...
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, sizeof(__u32));
+} events_perfbuf SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_RINGBUF);
+	__uint(max_entries, RINGBUF_SIZE);
+} events_ringbuf SEC(".maps");
```
anakryiko, line 29: note that you can use just one BPF map, you can default to RINGBUF or PERF_EVENT_ARRAY, doesn't matter. And then in your compat layer you can actually change the type of the map with bpf_map__set_type(), depending on what kernel supports. Then you'll just have to disable auto-create for heap map.

anakryiko (changes requested):
left few thoughts. It's good direction, but can be cleaned up a bit more and not require apps to do feature detection at all.

libbpf-tools/mountsnoop.c
chenhengqi: This should be done after skeleton load, so I added a new API for it.

libbpf-tools/compat.bpf.h
chenhengqi: Nice feature, got it.

libbpf-tools/compat.c
```diff
@@ -0,0 +1,115 @@
...
+static void perfbuf_sample_fn(void *ctx, int cpu, void *data, __u32 size)
+{
+	struct bpf_buffer *buffer = ctx;
+	bpf_buffer_sample_fn fn;
+
+	fn = buffer->fn;
+	if (!fn)
+		return;
+
+	fn(buffer->ctx, data, size);
```
anakryiko, line 29: add (void) to make it explicit that we ignore fn's  result

anakryiko (approved):
LGTM!

PR #4265, opened 2022-10-05
libbpf-tools: Fix klockstat to finish with -R

namhyung:
On extremely heavy lock contention, klockstat can be stuck in the loop
in print_stats() since it keeps adding new entries to the stat_map.

Instead of deleting the elements in the loop, process them all and then
delete them at the end.

### DIFF
--- a/libbpf-tools/klockstat.c
+++ b/libbpf-tools/klockstat.c
@@ -526,16 +526,9 @@ static int print_stats(struct ksyms *ksyms, int stack_map, int stat_map)
 			warn("Out of memory\n");
 			return -1;
 		}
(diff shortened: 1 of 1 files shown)


yonghong-song:
Indeed, the current implementation has an issue for
```
                        ret = bpf_map_lookup_and_delete_elem(stat_map,
                                                             &stack_id,
                                                             &ss->ls);
                        lookup_key = 0;
```

With `lookup_key = 0`, bpf_map_get_next_key will start from the beginning again and this may cause problems if there are intense new map elements.

Alternative is to use map_lookup_and_delete API, but it requires newer kernel.

PR #4270, opened 2022-10-11
Revert "Enable user symbol resolution on exited processes"

chenhengqi:
This partially reverts iovisor/bcc#4264.


chenhengqi:
cc @yonghong-song @lenticularis39 for awareness.

PR #4273, opened 2022-10-12
tools/vfsstat: Add PID filter support

xingfeng2510:
Currently, this tool traces system-wide VFS calls, which may be useful to focus on a process with lots of file I/O (MySQL、ClickHouse, and so on). Add option -p to filter specified PID.

### DIFF
--- a/tools/vfsstat.py
+++ b/tools/vfsstat.py
@@ -2,39 +2,47 @@
 # @lint-avoid-python-3-compatibility-imports
 #
 # vfsstat   Count some VFS calls.
-#           For Linux, uses BCC, eBPF. See .c file.
+#           For Linux, uses BCC, eBPF. Embedded C.
 #
 # Written as a basic example of counting multiple events as a stat tool.
 #
-# USAGE: vfsstat [interval [count]]
+# USAGE: vfsstat [-h] [-p PID] [interval] [count]
 #
 # Copyright (c) 2015 Brendan Gregg.
 # Licensed under the Apache License, Version 2.0 (the "License")
 #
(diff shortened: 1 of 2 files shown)


tools/vfsstat.py
```diff
@@ -51,34 +59,48 @@ def usage():
 
 BPF_ARRAY(stats, u64, S_MAXSTAT);
 
-static void stats_increment(int key) {
+static void stats_try_increment(int key) {
+    u32 pid = bpf_get_current_pid_tgid() >> 32;
```
chenhengqi, line 63: Move this line into `PID_FILTER`, so that we save a helper call if no PID filter applied.

xingfeng2510: done，thanks.

tools/vfsstat.py
```diff
@@ -2,39 +2,47 @@
...
+    ./vfsstat -p 185      # trace PID 185 only
+    ./vfsstat 1 10        # print 1 second summaries, 10 times
+"""
+parser = argparse.ArgumentParser(
+    description="Count some VFS calls.",
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    epilog=examples)
+parser.add_argument("-p", "--pid",
+    help="trace this PID only")
+parser.add_argument("interval", nargs="?", default=99999999,
```
chenhengqi, line 36: Let's keep the default interval to 1, so that it won't change the behavior of the tool. Other than that, looks good to me.

xingfeng2510: ok, it's my fault.

PR #4275, opened 2022-10-13
libbpf-tools: Enable biolatency on kernels without BPF trampoline

chenhengqi:
On kernels without BPF trampoline, let's switch to raw tracepoint instead.

cc @eiffel-fl


chenhengqi:
Thanks.

PR #4281, opened 2022-10-14
bcc: support building with external libbpf package and older uapi lin…

alan-maguire:
…ux/bpf.h

When building bcc with a relatively new packaged libbpf (0.8.1) and -DCMAKE_USE_LIBBPF_PACKAGE:BOOL=TRUE, multiple compilation failures are encountered due the fact the system uapi header in /usr/include/linux/bpf.h is not very recent (this is often the case for distros, which sync it via a kernel headers package quite conservatively due to use by glibc).

With libbpf built via git submodule, the uapi header included in the libbpf package is used, so here a similar approach is proposed for the external package build.  Instead of having to sync another file the already present compat/linux/virtual_bpf.h is used; we copy it to compat/linux/bpf.h (eliminating the string prefix/suffix on first/last lines).

From there, we ensure that places that assume the presence of the libbpf git submodule point at compat/ as a location to find the uapi header.

Signed-off-by: Alan Maguire <alan.maguire@oracle.com>

### DIFF
--- a/src/cc/CMakeLists.txt
+++ b/src/cc/CMakeLists.txt
@@ -15,6 +15,12 @@ endif (LIBDEBUGINFOD_FOUND)
 # todo: if check for kernel version
 if (CMAKE_USE_LIBBPF_PACKAGE AND LIBBPF_FOUND)
   include_directories(${LIBBPF_INCLUDE_DIRS})
+  # create up-to-date linux/bpf.h from virtual_bpf.h (remove string wrapper);
+  # when libbpf is built as a submodule we use its version of linux/bpf.h
+  # so this does similar for the libbpf package, removing reliance on the
+  # system uapi header which can be out of date.
+  execute_process(COMMAND sh -c "cd ${CMAKE_CURRENT_SOURCE_DIR}/compat/linux && grep -ve '\\*\\*\\*\\*' virtual_bpf.h > bpf.h")
+  include_directories(${CMAKE_CURRENT_SOURCE_DIR}/compat)
 else()
   include_directories(${CMAKE_CURRENT_SOURCE_DIR}/libbpf/include)
   include_directories(${CMAKE_CURRENT_SOURCE_DIR}/libbpf/include/uapi)
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ examples/cgroupid/cgroupid
 # Output from docker builds
 scripts/docker/output/
 /output/
+
+# UAPI header generated for libbpf package-based builds
+src/cc/compat/linux/bpf.h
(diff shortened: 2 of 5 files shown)


chenhengqi:
If users are building BCC with external libbpf package, I guess they are expected to install UAPI headers shipped with libbpf.

```shell
$ make install_uapi_headers
```

src/cc/CMakeLists.txt
```diff
@@ -15,6 +15,12 @@ endif (LIBDEBUGINFOD_FOUND)
 # todo: if check for kernel version
 if (CMAKE_USE_LIBBPF_PACKAGE AND LIBBPF_FOUND)
   include_directories(${LIBBPF_INCLUDE_DIRS})
+  # create up-to-date linux/bpf.h from virtual_bpf.h (remove string wrapper);
+  # when libbpf is built as a submodule we use its version of linux/bpf.h
+  # so this does similar for the libbpf package, removing reliance on the
+  # system uapi header which can be out of date.
+  execute_process(COMMAND sh -c "cd ${CMAKE_CURRENT_SOURCE_DIR}/compat/linux && grep -ve '\\*\\*\\*\\*' virtual_bpf.h > bpf.h")
```
chenhengqi, line 22: This generated `bpf.h` should be added to `.gitignore` ?

Another thought: Maybe we can allow users the specify the path to libbpf UAPI headers, something like:
```shell
$ cmake -DLIBBPF_UAPI_HEADERS_PATH=/path/to/headers ..
```

alan-maguire:
> If users are building BCC with external libbpf package, I guess they are expected to install UAPI headers shipped with libbpf.
> 
> ```shell
> $ make install_uapi_headers
> ```

The problem is that in most libbpf packages I've seen, they don't ship the uapi header file built with "make install_uapi_headers", because in many cases it will cause conflicts with a kernel-headers package which also delivers /usr/include/linux/bpf.h. 

My concern is that it is an extra step that the uses of the libbpf git submodule don't have to take; they get to use the uapi header bundled with libbpf when building. So I'm hoping we can figure out a way to handle the packaged libbpf similarly. Thanks!

src/cc/CMakeLists.txt
alan-maguire: good catch with the .gitignore, thanks! it would be nice to have a configurable path to uapi, but given that we have to keep compat/linux/virtual_bpf.h in sync already, generating a uapi header from it is a useful fallback mechanism for providing a uapi header internally for the bcc build I think. as mentioned below, it will often be the case that there is no up-to-date uapi header available, so I'm hoping we can cater to that case too.

chenhengqi:
Ah, I see. I am OK with this.

PR #4284, opened 2022-10-18
runqslower: create sufficiently large hash table to avoid missing tasks

ismhong:
We have encountered an issue while running runqslower for long run test, as you can check the log below. The workload of target platform is running auto test script and launching lots of applications periodically on Android TV. There are several abnormal records of run queue latency, ex. 1311763801.

We have investgated this issue and found the root cuase of this issue is the hash map to store en-queue start time of each task overflowed, and it seems cpudist also encountered such issue before and fix by following pull request.

https://github.com/iovisor/bcc/pull/2568

Apply the similar fix to runqslower, get the max_pid from /proc/sys/kernel/max_pid and pass to BPF C in kernel to create enough entries of hash map to store en-queue start time of each task.

```
(bcc)root@localhost:/# runqslower 100000 -P
Tracing run queue latency higher than 100000 us
TIME     COMM             TID           LAT(us) PREV COMM        PREV TID
23:42:45 glide-source-th  28986          161605 kworker/u9:3     12260
23:42:50 HTTPREQUEST_MAN  26249          107187 OMX_VideoInput   7152
23:42:50 migration/2      24             146799 OmxSideband      7147
23:42:50 rtk_post_worker  128            174145 migration/2      24
23:42:50 Binder:233_4     633            170457 runqslower       7086
23:42:58 GIBBON_SURFACE_  26233          103377 composer@2.4-se  457
23:43:14 Binder:581_1B    2202           113153 Binder:581_7     833
23:43:24 migration/0      15             107716 writer           7277
....
05:08:14 dumpLogs.sh      22242      1311763801 NBP_POOL-[1]     24233
05:08:14 dumpLogs.sh      22243      1311752373 appsrc7:src      24281
05:08:14 dumpLogs.sh      22248      1311760825 dumpLogs.sh      22193
```

Signed-off-by: Ism Hong <ism.hong@realtek.com>

### DIFF
--- a/tools/runqslower.py
+++ b/tools/runqslower.py
@@ -74,7 +74,7 @@ bpf_text = """
 #include <linux/nsproxy.h>
 #include <linux/pid_namespace.h>
 
-BPF_HASH(start, u32);
+BPF_ARRAY(start, u64, MAX_PID);
 
 struct data_t {
     u32 pid;
@@ -132,7 +132,7 @@ int trace_run(struct pt_regs *ctx, struct task_struct *prev)
 
     // fetch timestamp and calculate delta
     tsp = start.lookup(&pid);
-    if (tsp == 0) {
+    if ((tsp == 0) || (*tsp == 0)) {
         return 0;   // missed enqueue
     }
     delta_us = (bpf_ktime_get_ns() - *tsp) / 1000;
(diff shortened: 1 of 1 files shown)


tools/runqslower.py
```diff
@@ -132,7 +132,7 @@
 
     // fetch timestamp and calculate delta
     tsp = start.lookup(&pid);
-    if (tsp == 0) {
+    if ((tsp == 0) || (*tsp == 0)) {
```
chenhengqi, line 135: Why `*tsp == 0` check here and below ?

ismhong: Hi @chenhengqi 

Since this patch creates bpf map with pid_max entries from runtime information, I replace the original BPF_HASH with BPF_ARRAY for preventing hash collision and less array map lookup overhead as above.

All array elements are pre-allocated and zero initialized at init time, so we not only need to check `tsp == 0`, but also need to check it contains a valid timestamp. Does it make sense? Thank you.

chenhengqi: Sounds reasonable. Thanks.

PR #4285, opened 2022-10-19
tools/opensnoop: Bugfix for a full path contains multiple mountpoints

xingfeng2510:
For a pathname contains multiple mountpoints, -F option only dump till first mount point, not full path.

For example, an application try to open a.txt (located in /home/app/), while '/home' is a mount point, opensnoop will dump '... /app/a.txt' rather than '/home/app/a.txt'. 

This patch try to fix it.

### DIFF
--- a/tools/opensnoop.py
+++ b/tools/opensnoop.py
@@ -103,6 +103,7 @@ bpf_text = """
 #ifdef FULLPATH
 #include <linux/fs_struct.h>
 #include <linux/dcache.h>
+#include <linux/mount.h>
 
 #define MAX_ENTRIES 32
 
@@ -110,6 +111,20 @@ enum event_type {
     EVENT_ENTRY,
     EVENT_END,
 };
+
+struct mount {
+    struct hlist_node mnt_hash;
+    struct mount *mnt_parent;
+    struct dentry *mnt_mountpoint;
+    struct vfsmount mnt;
+    // ...
+};
+
+static inline struct mount *real_mount(struct vfsmount *mnt)
+{
+    void *__mptr = (void *)(mnt);
+    return (struct mount *)(__mptr - ((unsigned long)&((struct mount *)0)->mnt));
+}
 #endif
 
 struct val_t {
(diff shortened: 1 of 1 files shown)


tools/opensnoop.py
```diff
@@ -352,20 +367,37 @@
     if (data.name[0] != '/') { // relative path
         struct task_struct *task;
         struct dentry *dentry;
+        struct vfsmount *vfsmnt;
+        struct mount *mnt, *mnt_parent;
         int i;
 
-        task = (struct task_struct *)bpf_get_current_task_btf();
+        task = (struct task_struct *)bpf_get_current_task();
```
yonghong-song, line 374: We may have some verification with newer kernel. Which kernel did you test?

xingfeng2510: I tested on kernel 5.4/5.14/5.15. When I switched to bpf_get_current_task_btf(), it failed with such error messages:

```
84: (85) call bpf_get_current_task_btf#158
85: (b7) r6 = 31
; dentry = task->fs->pwd.dentry;
86: (79) r9 = *(u64 *)(r0 +2768)
; vfsmnt = task->fs->pwd.mnt;
87: (79) r8 = *(u64 *)(r9 +40)
; dentry = task->fs->pwd.dentry;
88: (07) r9 += 48
; mnt_parent = mnt->mnt_parent;
89: (79) r7 = *(u64 *)(r8 -16)
R8 is ptr_vfsmount invalid negative access: off=-16
processed 87 insns (limit 1000000) max_states_per_insn 0 total_states 4 peak_states 4 mark_read 3
```

xingfeng2510: btw, other tools still use bpf_get_current_task(), and bpf_get_current_task_btf() looks not stable, I didn't analyze it yet.

xingfeng2510:
any suggestions or feedback ?

dennisorlando:
should be addressed by #5153

xingfeng2510:
> should be addressed by #5153

#5153 unwind dentry by reading `/proc/self/mountinfo` frequently is inefficient. This pull request referred to kernel function `d_path`, which returns the path of a dentry and unwinds in kernel.

for example, when reading such a link file `/proc/793073/fd/4`, it returns full path, calls do_proc_readlink -> d_path
[d_path](https://elixir.bootlin.com/linux/v6.6.31/source/fs/proc/base.c#L1773)

```
# ls -l /proc/793073/fd/4
lrwx------ 1 root root 64 Nov 17 22:41 /proc/793073/fd/4 -> /usr/local/agent/log/agent.log
```

PR #4303, opened 2022-10-30
libbpf-tools/tcplife: fix filter by ports

mozillazg:
ebpf code depends on these variables when filtered by ports:

https://github.com/iovisor/bcc/blob/dcb34dc51a9336ff88c5a57305bd04758a25808c/libbpf-tools/tcplife.bpf.c#L60-L88


chenhengqi (approved):
Thanks for the fix.

PR #4313, opened 2022-11-03
don't quit due to ensure_core_btf failure

yuzhichang:
Sometimes `ensure_core_btf()` failure is not fatal.
I upgraded CentOS 7 kernel to `5.4.221-1.el7.elrepo.x86_64`, extracted `vmlinux.h` via `bpftool` and built libbpf-tools.
Several tools fail to start:
```
$ sudo ./execsnoop
failed to fetch necessary BTF for CO-RE: Operation not supported
```
With this patch, these tool work as expected.

PR #4315, opened 2022-11-04
libbpf-tools: Detect kernel/modules BTF using libbpf APIs

chenhengqi:
In #4313, there are situations that ensure_core_btf is failed to detect kernel BTF existence. Let's switch to libbpf APIs to perform a best effort searching.

### DIFF
--- a/libbpf-tools/compat.h
+++ b/libbpf-tools/compat.h
@@ -4,6 +4,8 @@
 #ifndef __COMPAT_H
 #define __COMPAT_H
 
+#include <stdlib.h>
+#include <limits.h>
 #include <sys/types.h>
 #include <linux/bpf.h>
 
@@ -20,5 +22,25 @@ int bpf_buffer__open(struct bpf_buffer *buffer, bpf_buffer_sample_fn sample_cb,
 		     bpf_buffer_lost_fn lost_cb, void *ctx);
 int bpf_buffer__poll(struct bpf_buffer *, int timeout_ms);
 void bpf_buffer__free(struct bpf_buffer *);
+
+/* taken from libbpf */
+#ifndef __has_builtin
+#define __has_builtin(x) 0
+#endif
+
+static inline void *libbpf_reallocarray(void *ptr, size_t nmemb, size_t size)
+{
+	size_t total;
+
+#if __has_builtin(__builtin_mul_overflow)
+	if (__builtin_mul_overflow(nmemb, size, &total))
+		return NULL;
+#else
+	if (size == 0 || nmemb > ULONG_MAX / size)
+		return NULL;
+	total = nmemb * size;
+#endif
+	return realloc(ptr, total);
+}
 
 #endif /* __COMPAT_H */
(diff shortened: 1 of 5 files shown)


chenhengqi:
@yuzhichang  Could you please check if this PR fix your problem ?

yuzhichang:
@chenhengqi By the way, the following patch is need:

```
eoi@ck16:/data01/zhichyu/bcc/libbpf-tools$ git diff
diff --git a/libbpf-tools/biotop.c b/libbpf-tools/biotop.c
index 1f61739..9815034 100644
--- a/libbpf-tools/biotop.c
+++ b/libbpf-tools/biotop.c
@@ -47,6 +47,15 @@ struct vector {
        void **elems;
 };
 
+static inline void *reallocarray_handwrite(void *ptr, size_t nmemb, size_t size)
+{
+        size_t total;
+        if (size == 0 || nmemb > ULONG_MAX / size)
+                return NULL;
+        total = nmemb * size;
+        return realloc(ptr, total);
+}
+
 int grow_vector(struct vector *vector) {
        if (vector->nr >= vector->capacity) {
                void **reallocated;
@@ -56,7 +65,7 @@ int grow_vector(struct vector *vector) {
                else
                        vector->capacity *= 2;
 
-               reallocated = reallocarray(vector->elems, vector->capacity, sizeof(*vector->elems));
+               reallocated = reallocarray_handwrite(vector->elems, vector->capacity, sizeof(*vector->elems));
                if (!reallocated)
                        return -1;

```

chenhengqi:
> 

Just added another commit to address this.

libbpf-tools/compat.h
```diff
@@ -21,4 +23,24 @@ int bpf_buffer__open(struct bpf_buffer *buffer, bpf_buffer_sample_fn sample_cb,
...
+#ifndef __has_builtin
+#define __has_builtin(x) 0
+#endif
+
+static inline void *libbpf_reallocarray(void *ptr, size_t nmemb, size_t size)
+{
+	size_t total;
+
+#if __has_builtin(__builtin_mul_overflow)
+	if (unlikely(__builtin_mul_overflow(nmemb, size, &total)))
```
yonghong-song, line 36: could you remove 'unlikely' here? The 'unlikely' is defined in libbpf_internal.h and not available to users using libbpf standard uapi functions.

This can be easily reproduced with 'make CC=clang' where the clang points to latest clang which does have __builtin_mul_overflow builtin.

```
In file included from compat.c:4:
./compat.h:36:6: warning: call to undeclared function 'unlikely'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
        if (unlikely(__builtin_mul_overflow(nmemb, size, &total)))
            ^
1 warning generated.
...
/usr/bin/ld: Dwarf Error: Invalid or unhandled FORM value: 0x25.
/home/yhs/work/bcc/libbpf-tools/.output/biotop.o: In function `grow_vector':
biotop.c:(.text+0x41): undefined reference to `unlikely'
/home/yhs/work/bcc/libbpf-tools/.output/biotop.o: In function `main':
```

chenhengqi: Done.

PR #4318, opened 2022-11-06
wakeuptime: remove unused global var `exiting`

woodpenker:
It seems that the global signal var `exiting` is of no use here, this PR is going to remove it.


chenhengqi:
Looks like `offcputime` also need this change, would you mind adding another commit for it ?

woodpenker:
> Looks like `offcputime` also need this change, would you mind adding another commit for it ?

Sure & updated.

PR #4327, opened 2022-11-11
Fix libbpf-tools/cpufreq

bjzhjing:
While trying the cpufreq tool under libbpf-tools, the improper initialization of data structure makes the result confusing, correct it, with an additional polish change for data display.

### DIFF
--- a/libbpf-tools/cpufreq.c
+++ b/libbpf-tools/cpufreq.c
@@ -132,7 +132,7 @@ static void sig_handler(int sig)
 {
 }
 
-static int init_freqs_hmz(__u32 *freqs_mhz, int nr_cpus)
+static int init_freqs_mhz(__u32 *freqs_mhz, int nr_cpus)
 {
 	char path[64];
 	FILE *f;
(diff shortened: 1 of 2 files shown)


libbpf-tools/cpufreq.c
```diff
@@ -155,6 +155,7 @@ static int init_freqs_hmz(__u32 *freqs_mhz, int nr_cpus)
 			fclose(f);
 			return -1;
 		}
+		freqs_mhz[i] /= 1000;
```
chenhengqi, line 158: Since you are here, please rename the function to `init_freqs_mhz` and add a comment to scaling_cur_freq saying that it is in kHZ.

libbpf-tools/trace_helpers.c
```diff
@@ -973,9 +973,11 @@ void print_linear_hist(unsigned int *vals, int vals_size, unsigned int base,
 	printf("     %-13s : count     distribution\n", val_type);
 	for (i = idx_min; i <= idx_max; i++) {
 		val = vals[i];
-		printf("        %-10d : %-8d |", base + i * step, val);
-		print_stars(val, val_max, stars_max);
-		printf("|\n");
+		if (val > 0) {
```
chenhengqi, line 976:
```c
if (!val)
    continue;
```
to avoid nestedness.

libbpf-tools/cpufreq.c
bjzhjing: Thanks for helping review. The change is done.

libbpf-tools/trace_helpers.c
bjzhjing: Good advice, thanks!

PR #4333, opened 2022-11-12
tools/filegone: add filegone to trace why file gone

curu:
file gone/vanished either been deleted or renamed

### DIFF
--- a/tools/filevanish.py
+++ b/tools/filevanish.py
@@ -1,0 +1,152 @@
+#!/usr/bin/python
+# @lint-avoid-python-3-compatibility-imports
+#
+# filevanish    Trace why file vanished
+#             For Linux, uses BCC, eBPF. Embedded C.
+#
+# USAGE: filevanish [-h] [-p PID]
+#
+# Licensed under the Apache License, Version 2.0 (the "License")
+#
+# 08-Nov-2022 Curu. modified from filelife
+
+from __future__ import print_function
+from bcc import BPF
+import argparse
+from time import strftime
+
+# arguments
+examples = """examples:
+    ./filevanish           # trace all stat() syscalls
+    ./filevanish -p 181    # only trace PID 181
+"""
+parser = argparse.ArgumentParser(
+    description="Trace stat() syscalls",
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    epilog=examples)
+parser.add_argument("-p", "--pid",
+    help="trace this PID only")
+parser.add_argument("--ebpf", action="store_true",
+    help=argparse.SUPPRESS)
+args = parser.parse_args()
+debug = 0
+
+# define BPF program
+bpf_text = """
+#include <uapi/linux/ptrace.h>
+#include <linux/fs.h>
+#include <linux/sched.h>
+
+struct data_t {
+    u32 pid;
+    u8 action;
+    char comm[TASK_COMM_LEN];
+    char fname[DNAME_INLINE_LEN];
+    char fname2[DNAME_INLINE_LEN];
+};
+
+BPF_HASH(birth, struct dentry *);
+BPF_PERF_OUTPUT(events);
+
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5, 12, 0)
+// trace file deletion and output details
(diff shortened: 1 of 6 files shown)


tools/filevanish.py
```diff
@@ -0,0 +1,152 @@
...
+
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5, 12, 0)
+// trace file deletion and output details
+int trace_unlink(struct pt_regs *ctx, struct inode *dir, struct dentry *dentry)
+#else
+int trace_unlink(struct pt_regs *ctx, struct user_namespace *ns, struct inode *dir, struct dentry *dentry)
+#endif
+{
+    struct data_t data = {};
+    u32 pid = bpf_get_current_pid_tgid();
```
chenhengqi, line 59: `pid` should be `bpf_get_current_pid_tgid() >> 32`.

tools/filevanish.py
```diff
@@ -0,0 +1,152 @@
...
+
+    FILTER
+
+    struct qstr d_name = dentry->d_name;
+    if (d_name.len == 0)
+        return 0;
+
+    if (bpf_get_current_comm(&data.comm, sizeof(data.comm)) == 0) {
```
chenhengqi, line 67: No need to check the return value of `bpf_get_current_comm()`, even it fails, we should capture this event.

tools/filevanish.py
```diff
@@ -0,0 +1,152 @@
...
+struct inode *new_dir, struct dentry *new_dentry)
+{
+#else
+int trace_rename(struct pt_regs *ctx, struct renamedata *rd)
+{
+    struct dentry *old_dentry = rd->old_dentry;
+    struct dentry *new_dentry = rd->new_dentry;
+#endif
+    struct data_t data = {};
+    u32 pid = bpf_get_current_pid_tgid();
```
chenhengqi, line 89: ditto

tools/filevanish.py
```diff
@@ -0,0 +1,152 @@
...
+        exit()
+
+# initialize BPF
+b = BPF(text=bpf_text)
+b.attach_kprobe(event="vfs_unlink", fn_name="trace_unlink")
+b.attach_kprobe(event="vfs_rmdir", fn_name="trace_unlink")
+b.attach_kprobe(event="vfs_rename", fn_name="trace_rename")
+
+# header
+print("%-8s %-7s %-16s %7s %s" % ("TIME", "PID", "COMM", "ACTION", "FILE"))
```
chenhengqi, line 135: `%6s` for `ACTION`.

tools/filevanish.py
```diff
@@ -0,0 +1,152 @@
...
+    u8 action;
+    char comm[TASK_COMM_LEN];
+    char fname[DNAME_INLINE_LEN];
+    char fname2[DNAME_INLINE_LEN];
+};
+
+BPF_HASH(birth, struct dentry *);
+BPF_PERF_OUTPUT(events);
+
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5, 12, 0)
```
chenhengqi, line 51: We could use BTF info to avoid this version check, see #4339.

brendangregg:
Thanks, some comments:

- I like it. Seems like a useful debugging tool, and a companion to filelife.
- Maybe filegone is a better name? Short is good.
- Needs a man page, and a link in README.md.
- I've come across a temporary file mechanism in libc/syscalls before, I forget the name, but seemed to be a special way to create temporary file objects that didn't go via the typical paths. Doesn't have to go in this version, but someday it would be good to trace it and add it to filelife and filegone.

tools/filevanish.py
curu: maybe change it later.
currently we need to keep compatibility with bcc versions below v0.23 which has no 'kernel_struct_has_field'

Rtoax: Hi, can we do this https://github.com/iovisor/bcc/pull/4346 ?

curu: > Hi, can we do this #4346 ?
RHEL/CentOS 7 comes with bcc version 0.10, which has no **kernel_struct_has_field**,   so the same script woud not run without modification.

so maybe we should drop support for that old version and use a legacy branch to do that?

Issue #4231, opened 2022-09-09
Allow libbpf-tools to run on old kernels


chenhengqi:
We have a bunch of libbpf tools require features like `fentry` / `tp_btf` which only available on kernel v5.5+.
Running these tools on old kernels result in EINVAL. Instead of error out, we can fallback to kprobe.
Open this ticket to keep track of this process:

* [ ] biolatency
* [ ] biosnoop
* [ ] biostacks
* [ ] bitesize
* [ ] cachestat
* [ ] cpudist
* [ ] cpufreq
* [ ] drsnoop
* [x] hardirqs
* [ ] mdflush
* [ ] numamove
* [ ] offcputime
* [ ] readahead
* [x] softirqs

If you are interested, please feel free to pick it up.

chenhengqi:
Using `raw_tp` is enough, this will make the tools running on older LTS kernels (like v4.19/v5.4).

woodpenker:
Do we consider providing a cmd flag such as `-btf` to let user provide their BTF file for the kernel, so that we can use tools even on 4.x kernels?

chenhengqi:
> Do we consider providing a cmd flag such as `-btf` to let user provide their BTF file for the kernel, so that we can use tools even on 4.x kernels?

We have such support at build time. See https://github.com/iovisor/bcc/pull/3889 .

woodpenker:
> > Do we consider providing a cmd flag such as `-btf` to let user provide their BTF file for the kernel, so that we can use tools even on 4.x kernels?
> 
> We have such support at build time. See #3889 .

It seems to rely on the BTF hub and won't support self-built kernels or some old kernels which can't be found in the  hub, eg:  `Ubuntu 4.18.0-22`. Since `LIBBPF_OPTS` is supported, how about letting the user make a choice?
example:
```c
if (env.btf_path) {
  LIBBPF_OPTS(bpf_object_open_opts, bpf_opts, .btf_custom_path = env.btf_path);
  skel = xxx_bpf__open_opts(&bpf_opts);
} else {
  skel = xxx_bpf__open();
}
```

chenhengqi:
Instead of touching every tool, I'd suggest providing the BTF path through environment variable and teach ensure_core_btf() to recognize it.

woodpenker:
> Instead of touching every tool, I'd suggest providing the BTF path through environment variable and teach ensure_core_btf() to recognize it.

[...]

PR #4334, opened 2022-11-12
offcputime: Allow offcputime to run on old kernels

woodpenker:
Allow offcputime to run on old kernels according to https://github.com/iovisor/bcc/issues/4231.
Test OK on Ubuntu with 5.4.0 kernel.

### DIFF
--- a/libbpf-tools/offcputime.bpf.c
+++ b/libbpf-tools/offcputime.bpf.c
@@ -44,21 +44,20 @@ struct {
 
 static bool allow_record(struct task_struct *t)
 {
-	if (targ_tgid != -1 && targ_tgid != t->tgid)
+	if (targ_tgid != -1 && targ_tgid != BPF_CORE_READ(t, tgid))
 		return false;
-	if (targ_pid != -1 && targ_pid != t->pid)
+	if (targ_pid != -1 && targ_pid != BPF_CORE_READ(t, pid))
 		return false;
-	if (user_threads_only && t->flags & PF_KTHREAD)
+	if (user_threads_only && BPF_CORE_READ(t, flags) & PF_KTHREAD)
 		return false;
-	else if (kernel_threads_only && !(t->flags & PF_KTHREAD))
+	else if (kernel_threads_only && !(BPF_CORE_READ(t, flags) & PF_KTHREAD))
 		return false;
 	if (state != -1 && get_task_state(t) != state)
 		return false;
 	return true;
 }
 
-SEC("tp_btf/sched_switch")
-int BPF_PROG(sched_switch, bool preempt, struct task_struct *prev, struct task_struct *next)
+static int handle_switch(u64 *ctx, struct task_struct *prev)
 {
 	struct internal_key *i_keyp, i_key;
 	struct val_t *valp, val;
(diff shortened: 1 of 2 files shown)


libbpf-tools/offcputime.bpf.c
```diff
@@ -107,4 +106,15 @@ int BPF_PROG(sched_switch, bool preempt, struct task_struct *prev, struct task_s
 	return 0;
 }
 
+SEC("tp_btf/sched_switch")
+int BPF_PROG(sched_switch_btf, bool preempt, struct task_struct *prev, struct task_struct *next) {
+	return handle_switch(ctx, prev);
+}
+
+SEC("kprobe/finish_task_switch")
+int BPF_PROG(sched_switch_tp) {
```
chenhengqi, line 115: Let's fallback to raw tracepoint instead. See https://github.com/iovisor/bcc/pull/4217.

libbpf-tools/offcputime.c
```diff
@@ -314,6 +314,12 @@ int main(int argc, char **argv)
 		return 1;
 	}
 
+	if (probe_tp_btf("sched_switch")) {
+		bpf_program__set_autoload(obj->progs.sched_switch_tp, false);
+	} else {
+		bpf_program__set_autoload(obj->progs.sched_switch_btf, false);
+	}
```
chenhengqi, line 321: We are following kernel code style guide, so drop the surrounding `{}`.

libbpf-tools/offcputime.bpf.c
woodpenker: Thanks for replay, but on my test env, the `"raw_tp/sched_switch"` won't work:
```
libbpf: sched_switch is not found in vmlinux BTF                                                                                                               
    entries_lock                                                                                                                                               
failed to get syms                                                                                                                                             
    -                 (0)                                                                                                                                      
        51                                                                                                                                                     
                                                                                                                                                               
    entries_lock                                                                                                                                               
failed to get syms
    -                 (0)
        4811

    entries_lock        
```
The comm is empty and PID is always 0, I think the `prev` is not correctly taken. Do you have any suggest?

libbpf-tools/offcputime.c
woodpenker: > We are following kernel code style guide, so drop the surrounding `{}`.

Do you mean to write it like this? 
```
	bpf_program__set_autoload(obj->progs.sched_switch_btf, false);
	if (probe_tp_btf("sched_switch")) {
		bpf_program__set_autoload(obj->progs.sched_switch_tp, false);
		bpf_program__set_autoload(obj->progs.sched_switch_btf, true);
	}
```

libbpf-tools/offcputime.bpf.c
woodpenker: > Let's fallback to raw tracepoint instead. See #4217.

I tried to make it work and succeed, but the call stacks still seem always belong to the tracepoint `raw_tp/sched_switch`.

Changed code:
```c
SEC("raw_tp/sched_switch")
int BPF_PROG(sched_switch_tp) {
	struct task_struct *prev = (struct task_struct *)((struct bpf_raw_tracepoint_args *)ctx)->args[1];
	return handle_switch(ctx, prev);
}
```
Output example:
```
    entries_lock                                                                                                                                    [250/15106]
    -                kworker/2:2 (2704182)                                                                                                                     
        9                                                                                                                                                      
                                                                                                                                                                       

    entries_lock                       
    wait4                              
    -                bash (2708107)
        13                             

    entries_lock                       
    -                ksoftirqd/0 (10)
        1                              
```

chenhengqi: Does #4217 work for you ?

woodpenker: > Does #4217 work for you ?

Everything works fine except the callstacks, so close this pr until I find the reason and solutions。

PR #4338, opened 2022-11-13
libbpf-tools/cpudist: Allow cpudist to run on old kernels

woodpenker:
According to https://github.com/iovisor/bcc/issues/4231

Test OK on Ubuntu 20.04 with 5.4.0-97-generic kernel like this:

```
$ sudo ./cpudist  
libbpf: sched_switch is not found in vmlinux BTF
^C
         0 -> 1          : 5        |**                                      |
         2 -> 3          : 6        |**                                      |
         4 -> 7          : 14       |*****                                   |
         8 -> 15         : 22       |*********                               |
        16 -> 31         : 29       |************                            |
        32 -> 63         : 28       |***********                             |
        64 -> 127        : 51       |*********************                   |
       128 -> 255        : 15       |******                                  |
       256 -> 511        : 17       |*******                                 |
       512 -> 1023       : 29       |************                            |
      1024 -> 2047       : 67       |****************************            |
      2048 -> 4095       : 89       |*************************************   |
      4096 -> 8191       : 90       |*************************************   |
      8192 -> 16383      : 95       |****************************************|

```

### DIFF
--- a/libbpf-tools/cpudist.bpf.c
+++ b/libbpf-tools/cpudist.bpf.c
@@ -71,8 +71,7 @@ static __always_inline void update_hist(struct task_struct *task,
 		histp = bpf_map_lookup_elem(&hists, &id);
 		if (!histp)
 			return;
-		bpf_probe_read_kernel_str(&histp->comm, sizeof(task->comm),
-					  task->comm);
+		BPF_CORE_READ_STR_INTO(&histp->comm, task, comm);
 	}
 	delta = ts - *tsp;
 	if (targ_ms)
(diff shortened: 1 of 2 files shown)


libbpf-tools/cpudist.bpf.c
```diff
@@ -71,8 +71,7 @@ static __always_inline void update_hist(struct task_struct *task,
 		histp = bpf_map_lookup_elem(&hists, &id);
 		if (!histp)
 			return;
-		bpf_probe_read_kernel_str(&histp->comm, sizeof(task->comm),
-					  task->comm);
+		BPF_CORE_READ_STR_INTO(&histp->comm, task, comm);
```
yonghong-song, line 74: this and below BPF_CORE_READ will actually call kernel helpers instead of direct memory access. But I understand this is necessary for raw_tp. I think it is okay for portability perspective. We have quite a few tools still using kprobe which uses BPF_CORE_READ as well.

woodpenker: Thanks for the review, do you mean this will have a latency cost to use BPF_CORE_READ instead of direct mem access? Maybe another way is to separate each other with independent functions containing the same logic?

PR #4339, opened 2022-11-13
tools: filelife: Check btf struct field for CO-RE and add vfs_open() 

Rtoax:
Since kernel commit 6521f8917082("namei: prepare for idmapped mounts"), the vfs_unlink() function add argument 'struct user_namespace'. And add vfs_open() probe if 'f_mode = FMODE_CREATED'.

### DIFF
--- a/tools/filelife.py
+++ b/tools/filelife.py
@@ -16,6 +16,7 @@
 #
 # 08-Feb-2015   Brendan Gregg   Created this.
 # 17-Feb-2016   Allan McAleavy updated for BPF_PERF_OUTPUT
+# 13-Nov-2022   Rong Tao        Check btf struct field for CO-RE and add vfs_open()
 
 from __future__ import print_function
 from bcc import BPF
@@ -24,11 +25,11 @@ from time import strftime
 
 # arguments
 examples = """examples:
-    ./filelife           # trace all stat() syscalls
+    ./filelife           # trace lifecycle of file(create->remove)
     ./filelife -p 181    # only trace PID 181
 """
 parser = argparse.ArgumentParser(
-    description="Trace stat() syscalls",
+    description="Trace lifecycle of file",
     formatter_class=argparse.RawDescriptionHelpFormatter,
     epilog=examples)
 parser.add_argument("-p", "--pid",
@@ -54,8 +55,7 @@ struct data_t {
 BPF_HASH(birth, struct dentry *);
 BPF_PERF_OUTPUT(events);
 
-// trace file creation time
-int trace_create(struct pt_regs *ctx, struct inode *dir, struct dentry *dentry)
+static int probe_dentry(struct pt_regs *ctx, struct dentry *dentry)
 {
     u32 pid = bpf_get_current_pid_tgid() >> 32;
     FILTER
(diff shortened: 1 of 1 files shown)


tools/filelife.py
```diff
@@ -109,12 +151,21 @@
...
+    bpf_text = bpf_text.replace('TRACE_UNLINK_FUNC', trace_unlink_text_new)
+else:
+    bpf_text = bpf_text.replace('TRACE_CREATE_FUNC', trace_create_text_old)
+    bpf_text = bpf_text.replace('TRACE_UNLINK_FUNC', trace_unlink_text_old)
+
 # initialize BPF
 b = BPF(text=bpf_text)
 b.attach_kprobe(event="vfs_create", fn_name="trace_create")
+# newer kernels may don't fire vfs_create, call vfs_open instead:
+b.attach_kprobe(event="vfs_open", fn_name="trace_open")
```
chenhengqi, line 165: Maybe we should also trace `vfs_tmpfile` ?

Rtoax: When I wrote this code, I also thought about vfs_tmpfile 🤝, thanks

Rtoax: I check `$ man 2 open`

```
O_TMPFILE (since Linux 3.11) Create an unnamed temporary regular file. Anything written to the resulting file will be lost when the last file descriptor is closed, unless the file is given a name.
```
vfs_tmpfile DON'T have dentry struct, thus, i think we don't need it.

Rtoax:
@chenhengqi Hi, Hengqi, Is my modification correct and can we continue?

tools/filelife.py
```diff
@@ -16,6 +16,8 @@
 #
 # 08-Feb-2015   Brendan Gregg   Created this.
 # 17-Feb-2016   Allan McAleavy updated for BPF_PERF_OUTPUT
+# 13-Nov-2022   Rong Tao        Update vfs_create/vfs_unlink argument in CO-RE
+#                               way, and add vfs_open for f_mode=FMODE_CREATED
```
yonghong-song, line 20: This patch is for bcc tools which do not need CO-RE, so please remove reference to CO-RE in the above.
Could you explain a little bit more in the commit message for how f_mode=FMODE_CREATE related to vfs_open?

Please also change subject line to remove CO-RE reference.

tools/filelife.py
```diff
@@ -109,12 +151,21 @@
     if args.ebpf:
         exit()
 
+if BPF.kernel_struct_has_field(b'renamedata', b'old_mnt_userns') == 1:
```
yonghong-song, line 154: I checked the original patch (https://lore.kernel.org/all/20210121131959.646623-15-christian.brauner@ubuntu.com/) this looks good to me.

tools/filelife.py
```diff
@@ -109,12 +151,21 @@
...
+    bpf_text = bpf_text.replace('TRACE_CREATE_FUNC', trace_create_text_new)
+    bpf_text = bpf_text.replace('TRACE_UNLINK_FUNC', trace_unlink_text_new)
+else:
+    bpf_text = bpf_text.replace('TRACE_CREATE_FUNC', trace_create_text_old)
+    bpf_text = bpf_text.replace('TRACE_UNLINK_FUNC', trace_unlink_text_old)
+
 # initialize BPF
 b = BPF(text=bpf_text)
 b.attach_kprobe(event="vfs_create", fn_name="trace_create")
+# newer kernels may don't fire vfs_create, call vfs_open instead:
```
yonghong-song, line 164: change to 'may not fire'. Any upstream commit can be referenced here?

PR #4346, opened 2022-11-19
tools/filegone: Check btf struct field instead of KERNEL_VERSION macro

Rtoax:
kernel commit 9fe61450972d("namei: introduce struct renamedata") introduce 'struct renamedata', at the same time, the function declarations changed.

Changes:
1. kernel_struct_has_field() function return true if struct exist and field set to and empty string('');
2. use kernel_struct_has_field() to determine if structure 'renamedata' exists, thus solving the vfs_rename using CO-RE way instead of using the kernel version number.

### DIFF
--- a/tools/old/filegone.py
+++ b/tools/old/filegone.py
@@ -9,7 +9,6 @@
 # Licensed under the Apache License, Version 2.0 (the "License")
 #
 # 08-Nov-2022 Curu. modified from filelife
-# 19-Nov-2022 Rong Tao Check btf struct field instead of KERNEL_VERSION macro.
 
 from __future__ import print_function
 from bcc import BPF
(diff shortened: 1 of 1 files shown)


Rtoax:
`trace_unlink()` CO-RE can fixed with this PR's way too, but, i think it's not belongs to kernel commit 9fe61450972d("namei: introduce struct renamedata"), so, make it another PR?

chenhengqi:
The two commits belongs to the same patch set, there is no need to check them separately.

I am uncomfortable about the change to `kernel_struct_has_field()`, just do `kernel_struct_has_field("renamedata", "old_mnt_userns")`.

Rtoax:
Thanks, Hengqi, i changed the code.

Rtoax:
Thanks, Hengqi, i changed the code.

tools/filegone.py
```diff
@@ -9,6 +9,8 @@
 # Licensed under the Apache License, Version 2.0 (the "License")
 #
 # 08-Nov-2022 Curu. modified from filelife
+# 19-Nov-2022 Rong Tao vfs_rename/unlink/rmdir: Use CO-RE way, instead of
+#                      KERNEL_VERSION macro
```
yonghong-song, line 13: Just say 'Check btf struct field instead of KERNEL_VERSION macro'.
The subject can be 'Check btf struct field field of KERNEL_VERSION macro'.

The btf struct field checking applies to all three functions, not just vfs_rename().

Rtoax: I changed the comment and subject, thanks Song.

PR #4356, opened 2022-11-22
libbpf-tools: update ppc vmlinux.h with bpf_perf_link, BPF_LINK_TYPE_PERF_EVENT

natoscott:
A BCC-based project I'm working on builds fine on x86_64 but fails on ppc. The cause is the slightly outdated vmlinux header for the latter platform, which is missing the bpf_perf_link structure and related macros.

I've regenerated a header using this following command (kernel 5.14.0) and updated the symlink:
bpftool btf dump file /sys/kernel/btf/vmlinux format c > vmlinux_5140.h

Signed-off-by: Nathan Scott <nathans@redhat.com>


libbpf-tools/powerpc/vmlinux.h
```diff
@@ -1 +1 @@
-vmlinux_510.h
\ No newline at end of file
+vmlinux_5140.h
```
chenhengqi, line 1: Should be `vmlinux_514.h`.

natoscott: Kernel version used was 5.14.0 though, not 5.1.4 ... so I think its correct as-is, no?

natoscott: Perhaps a better scheme would be something like 'vmlinux_5_14_0.h'?

chenhengqi: See the README file. 505 for 5.5, let's follow this convention.

natoscott: Aha - thanks for the pointer - will update.

PR #4359, opened 2022-11-23
Memory sanitizer workarounds

captain5050:


libbpf-tools/klockstat.c
```diff
@@ -179,6 +179,7 @@ static bool parse_sorts(struct prog_env *env, char *arg)
 	return parse_one_sort(env, arg);
 }
 
+ __attribute__((no_sanitize("memory")))
```
chenhengqi, line 182: Why we need this change ? Please squash your commits into one.

captain5050: The change is necessary to avoid klockstat failing with -fsanitize=memory (tested on clang):
```
==9588==WARNING: MemorySanitizer: use-of-uninitialized-value
     #0 0x55c0ee8b35b1 in parse_arg src/klockstat.c:220:25
     #1 0x7f7299390d7c in argp_parse (.../lib64/libc.so.6+0x144d7c)
     #2 0x55c0ee8ae9b1 in main src/klockstat.c:717:8
SUMMARY: MemorySanitizer: use-of-uninitialized-value src/klockstat.c:220:25 in parse_arg
```

The issue being that memory sanitizer doesn't intercept argp_parse and so fails to see the memory as initialized. Similarly with the switch from malloc to calloc.

The previous commits were generated by hitting sync on my tree in github. Not clear on how to avoid them showing up in the PR but a squash sounds wrong.

chenhengqi: Rebase, squash and force push should work.

chenhengqi: What's your clang version ?

PR #4363, opened 2022-11-26
Add initial guide and scripts for building bcc from specific commit

ArthurChiao:
1. Currently only covers Ubuntu 20.04
2. Also added several minor fixups/workarounds missing from the upstream

Signed-off-by: ArthurChiao <arthurchiao@hotmail.com>


ArthurChiao:
Sorry for disruption, please ignore this.

PR #4366, opened 2022-11-28
tools/ttysnoop: Fix wrong KERNEL_VERSION

michael-chuh:
Kernel commits 9bb48c82aced (:tty: implement write_iter") is merged from v5.10.11.

Signed-off-by: mickey_zhu <mickey_zhu@realsil.com.cn>

### DIFF
--- a/tools/ttysnoop.py
+++ b/tools/ttysnoop.py
@@ -120,7 +120,7 @@ static int do_tty_write(void *ctx, const char __user *buf, size_t count)
  * commit 9bb48c82aced (v5.11-rc4) tty: implement write_iter
  * changed arguments of tty_write function
  */
-#if LINUX_VERSION_CODE < KERNEL_VERSION(5, 11, 0)
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5, 10, 11)
 int kprobe__tty_write(struct pt_regs *ctx, struct file *file,
     const char __user *buf, size_t count)
 {


chenhengqi:
Commit 9bb48c82aced exists since v5.10-rc5, why v5.10.11 ?

michael-chuh:
Linux tags show it is added from v5.10-rc5, but version number is actually from v5.10.11.
Simply check the source code:
https://elixir.bootlin.com/linux/v5.10.11/source/drivers/tty/tty_io.c#L146
https://elixir.bootlin.com/linux/v5.10.10/source/drivers/tty/tty_io.c#L146

chenhengqi:
I see, the commit was merged in v5.11-rc5, and backported to 5.10.

chenhengqi:
https://cdn.kernel.org/pub/linux/kernel/v5.x/ChangeLog-5.10.11

tools/ttysnoop.py
```diff
@@ -162,6 +162,22 @@ def usage():
 #endif
 """
 
+kernel_release = open('/proc/sys/kernel/osrelease').read().split('.')
+if int(kernel_release[0]) >= 5 and int(kernel_release[1]) >= 8:
+    PERF_MODE = "USE_BPF_RING_BUF"
+else:
+    PERF_MODE = "USE_BPF_PERF_BUF"
```
chenhengqi, line 169: This is unreliable due to backport, currently, you can use `kernel_struct_has_field()` API.

Or maybe introduce a new API `kernel_has_struct()`.

michael-chuh: Replace with "BPF.kernel_struct_has_field(b'bpf_ringbuf', b'waitq') == 1" works.
One more question, it's my first time to pull request, there are already 2 commits, it will be 3 after I upload a refine patch. I was going to upload one patch for version fix, and another for ringbuf, how could I do? Will it be ok if I just upload later refine patch, or need I revert the ringbuf patch and create a new branch to upload?

yonghong-song: Let us do two separate commits: one for version fix and the other is ringbuf fix.

michael-chuh: Hi，ringbuf patch is reverted

PR #4372, opened 2022-12-01
tools/runqslower: replace invalid array map delete op by zero update

curu:
The BPF_ARRAY map type has no delete operation. which may cause long runq delay result because old timestamp can not be deleted

### DIFF
--- a/tools/runqslower.py
+++ b/tools/runqslower.py
@@ -150,7 +150,8 @@ int trace_run(struct pt_regs *ctx, struct task_struct *prev)
     // output
     events.perf_submit(ctx, &data, sizeof(data));
 
-    start.delete(&pid);
+    //array map has no delete method, set ts to 0 instead
+    *tsp = 0;
     return 0;
 }
 """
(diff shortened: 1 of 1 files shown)


chenhengqi:
cc @ismhong

tools/runqslower.py
```diff
@@ -150,7 +150,9 @@
     // output
     events.perf_submit(ctx, &data, sizeof(data));
 
-    start.delete(&pid);
+    //array map has no delete method, set ts to 0 instead
+    *tsp = 0;
+    start.update(&pid, tsp);
```
yonghong-song, line 155: Indeed, the array deleting elem operation will return -EINVAL.
In kernel/bpf/arraymap.c,
```
/* Called from syscall or from eBPF program */
static int array_map_delete_elem(struct bpf_map *map, void *key)
{       
        return -EINVAL;
}  
```

The above `start.update(...)` is not needed, `*tsp = 0` should already set ts to 0 in the array map.
The same for another case below.

ismhong:
Apologize for my carelessness.
I didn't realize there is no delete method for BPF_ARRAY when I created the PR, and thanks for @yonghong-song  to clarify it.

Hi @curu,
Could you please help to update the PR? Thank you for spotting this issue.
Or I can help to create a PR for this to fix my own mistake. :)

tools/runqslower.py
curu: thanks! removed  the update.

PR #4380, opened 2022-12-05
Add madvsnoop

irenge:
Hi I am learning bcc tool development.
I would like to request for your comments on this tool. 
I just borrowed some code and change
Thanks


yonghong-song (commented):
I think this tool is not necessary. `trace.py` should be able to capture all arguments for a particular tracepoint.

PR #4382, opened 2022-12-06
libbpf-tools/klockstat: Memory sanitizer workarounds

captain5050:
Disable memory sanitizer for the parse_arg libc callback. Ensure stack_stat appears initialized with calloc.

Replaces badly sync-ed PR #4359.

### DIFF
--- a/tools/ppchcalls.py
+++ b/tools/ppchcalls.py
@@ -94,7 +94,7 @@ hcall_table = {
     420: 'H_CREATE_RPT',
     424: 'H_REMOVE_RPT',
     428: 'H_REGISTER_RPAGES',
-    432: 'H_DISABLE_AND_GETC',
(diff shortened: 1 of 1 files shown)


yonghong-song:
@captain5050 could you explain the issue again in the commit message? If malloc is changed to calloc, does asan issue fixed or not? We should try to fix the asan-triggered issue, rather than adding sanitization attribute.

captain5050:
BPF system calls are not understood by memory sanitizer, as such memory passed to them is considered uninitialized even though we know the system call should initialize it. Using calloc rather than malloc is a workaround for this and ideally wouldn't be necessary. These changes are sufficient to make klockstat run cleanly with memory sanitizer but the issues raised by memory sanitizer were more issues with it than klockstat.

PR #4396, opened 2022-12-15
tools: add numasched: trace task switch between NUMA

Rtoax:
Tasks frequently switch between NUMA, often resulting in poor performance.

programs running on the same NUMA will have better performance (usually at the memory level). If processes are migrating between different CPUs and these CPUs belong to different NUMAs, this tool provides an effective means of tracing to optimize the program or configuration.

# For example:

## Terminal 1:

```
  $ numactl --hardware
  available: 4 nodes (0-3)
  node 0 cpus: 0 - 23
  node 1 cpus: 24 - 47
  node 2 cpus: 48 - 71
  node 3 cpus: 72 - 95
```

## Terminal 2:

```
  $ sudo ./numasched.py
  Tracing task NUMA switch...
  TIME     PID      SRC_NID     DST_NID  COMM
```

## Terminal 3:

```
  $ taskset -c 1 yes >/dev/null &
  $ taskset -p 0x1000000 $(pidof yes)
  $ taskset -p 0x1000000000000 $(pidof yes)
```

## Then, Terminal 2:

```
  $ sudo ./numasched.py
  Tracing task NUMA switch...
  TIME     PID      SRC_NID     DST_NID  COMM
  20:55:35 355842   0        -> 1        b'yes'
  20:55:50 355842   1        -> 2        b'yes'
```

### DIFF
--- a/tools/numasched.py
+++ b/tools/numasched.py
@@ -1,0 +1,146 @@
+#!/usr/bin/env python
+# @lint-avoid-python-3-compatibility-imports
+#
+# numasched  Trace task NUMA switch
+#            For Linux, uses BCC, eBPF.
+#
+# USAGE: numasched [-p PID] [-t TID] [-c COMM]
+#
+# This script tracks NUMA migrations of tasks, and in general, frequent
+# NUMA migrations can cause poor performance.
+#
+# Copyright 2022 CESTC, Co.
+# Licensed under the Apache License, Version 2.0 (the "License")
+#
+# 14-Dec-2022   Rong Tao    Created this.
+
+from __future__ import print_function
+from bcc import BPF
+import argparse
+from time import strftime
+from socket import inet_ntop, AF_INET, AF_INET6
+from struct import pack
+from time import sleep
+
+
+# arguments
+examples = """examples:
+    ./numasched             # trace all processes
+    ./numasched -p 185      # trace PID 185 only
+"""
+parser = argparse.ArgumentParser(
+    description="Trace task NUMA switch",
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    epilog=examples)
+parser.add_argument("-p", "--pid",
+    help="trace this PID only")
+parser.add_argument("-t", "--tid",
+    help="trace this TID only")
+parser.add_argument("-c", "--comm",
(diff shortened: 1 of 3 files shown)


tools/numasched.py
```diff
@@ -0,0 +1,141 @@
...
+    bpf_get_current_comm(&val.comm, sizeof(val.comm));
+    val.nid = new_nid;
+
+    valp = numaid_info.lookup(&pid);
+    if (!valp)
+        goto update;
+
+    old_nid = valp->nid;
+
+    if (old_nid != new_nid && !comm_strcmp(val.comm, valp->comm)) {
```
yonghong-song, line 95: any particular reason you compare based comm instead of tid ((u32)bpf_get_current_pid_tgid())?
do we need to handle cases like (one process has two threads, one thread moved to a different numa node while the other stays in the same numa node)?

tools/numasched.py
```diff
@@ -0,0 +1,141 @@
...
+else:
+    # always skip PID=0
+    bpf_text = bpf_text.replace('FILTER_PID', 'pid == 0')
+
+# process event
+def print_event(cpu, data, size):
+    event = b["events"].event(data)
+    print("%-8s %-8d %-8d -> %-8d %-8s" %
+        (strftime("%H:%M:%S"),
+        event.pid,
```
yonghong-song, line 123: I believe we need both pid and tid to capture the whole picture.

tools/numasched.py
Rtoax: Thanks, Song. When I wrote this tool, I didn't take the multithreaded process model into account, thank you for the review. I just submitted code that uses tid as a judgment condition and added --tid and --comm command line arguments. I have also revised the following modification suggestions, thank you.

tools/numasched.py
Rtoax: I added tid, thank you.

Rtoax:
Rebase to master!

Rtoax:
Rebase to master.

Rtoax:
Rebase to master.

yonghong-song:
LGTM. As a followup, could you add a link for the new tool to the README.md?

PR #4401, opened 2022-12-20
libbpf-tools: Add schedtimes tool.

curu:
schedtimes is a tool to show process run/sleep/block/queue time from scheduler view. 
based on schedtimes.stp from systemtap
see https://sourceware.org/systemtap/examples/process/schedtimes.stp.

sample output:

23:35:35
COMM             PID     RUN(us)    SLEEP(us)  BLOCK(us)  QUEUE(us)  TOTAL(us)
kworker/1:1      563024  7          1061942    0          4          1061954
Monitor Deflati  323936  8          1000103    0          7          1000119
ksmtuned         668135  959996     0          0          6          960002
kworker/7:4      562372  6          959991     0          4          960002


yonghong-song:
Could you clarify how this tool might help you to solve a production issue?

libbpf-tools/schedtimes.c
```diff
@@ -0,0 +1,358 @@
...
+
+    time(&t);
+    tm = localtime(&t);
+    strftime(ts, sizeof(ts), "%H:%M:%S", tm);
+    printf("%s\n", ts);
+    printf("%-16s %-7s %-10s %-10s %-10s %-10s %s\n",
+            "COMM", "PID", "RUN(us)", "SLEEP(us)", "BLOCK(us)", "QUEUE(us)", "TOTAL(us)");
+
+
+    while (!bpf_map_get_next_key(schedtimes_fd, &lookup_key, &next_key)){
```
yonghong-song, line 241: We should avoid using `bpf_map_get_next_key` which might have some issues if map delete happens in between. Please check batch_map_ops/dump_hash in map_helpers.c.

libbpf-tools/schedtimes.bpf.c
```diff
@@ -0,0 +1,200 @@
...
+        delta = state_ts.ts - state_tsp->ts;
+        if(state_tsp->state == STATE_SLEEPING)
+            update_times(p, delta, STATE_SLEEPING);
+        if(state_tsp->state == STATE_BLOCKED)
+            update_times(p, delta, STATE_BLOCKED);
+        bpf_map_delete_elem(&start, &pid);
+    }
+    //update queue time
+    state_ts.state = STATE_QUEUED;
+    bpf_map_update_elem(&start, &pid, &state_ts, BPF_ANY);
```
yonghong-song, line 160: We have bpf_map_delete_elem above followed by bpf_map_update_elem. There is no need for this. You can directly modify state_tsp contents if not null.

libbpf-tools/schedtimes.c
curu: changed,  the code is shortter, with overhead of extra keys  memory cost.
btw: there's  no other delete  in schedtimes

and I think maybe we should add another clear_hash helper ?

curu:
> Could you clarify how this tool might help you to solve a production issue?

we use this tool to find out cpu time distribution of target process,  for example,  if a request  costs 200ms on one server and 150s  on the other,  with schedtimes, we can tell if it's due to runq, io or mem block, or just sleeptime (maybe wait for backend rsp)  differ.

PR #4409, opened 2022-12-30
libbpf-tools/runqlen: Use rq instead of cfs_rq when getting nr_running

suiya001:
The rq->nr_running can more accurately reflect the queuing congestion of tasks on the current cpu, while cfs_rq->nr_running can only reflect a part of the group in the container environment.

Signed-off-by: Hailong Liu <liuhailong@linux.alibaba.com>

### DIFF
--- a/libbpf-tools/runqlen.bpf.c
+++ b/libbpf-tools/runqlen.bpf.c
@@ -18,7 +18,7 @@ int do_sample(struct bpf_perf_event_data *ctx)
 	u64 slot, cpu = 0;
 
 	task = (void*)bpf_get_current_task();
-	slot = BPF_CORE_READ(task, se.cfs_rq, nr_running);
+	slot = BPF_CORE_READ(task, se.cfs_rq, rq, nr_running);
 	/*
 	 * Calculate run queue length by subtracting the currently running task,
 	 * if present. len 0 == idle, len 1 == one running task.


yonghong-song:
You are right. Currently, we use `cfs_rq->nr_running` to account the queue length for each inidividual cfs_rq, while `cfs_rq->rq->nr_running` is the the queue length for that particular cpu. I think we should not change the existing behavior as it might still be useful for some people to identify the potential issue for a particular cfs_rq at e.g., cgroup level. I think you can add a new option `--percpu` for per-cpu queue length accounting.

suiya001:
> You are right. Currently, we use `cfs_rq->nr_running` to account the queue length for each inidividual cfs_rq, while `cfs_rq->rq->nr_running` is the the queue length for that particular cpu. I think we should not change the existing behavior as it might still be useful for some people to identify the potential issue for a particular cfs_rq at e.g., cgroup level. I think you can add a new option `--percpu` for per-cpu queue length accounting.

Ok, thanks for your suggestion.

suiya001:
@yonghong-song ,
Hi yonghong, I'm not sure that shoud I add a new option --percpu for per-cpu runqueue length accounting  or modify the existing options ”C“ which is used to print output the cfs_rq->nr_running of the specified cpu ?

PR #4424, opened 2023-01-07
tools/ttysnoop: Add ITER_UBUF support

Rtoax:
Since kernel commit fcb14cb1bdac("new iov_iter flavour - ITER_UBUF"), tty_write() will use ITER_UBUF. If do not add support for ITER_UBUF, ttysnoop will not work

### DIFF
--- a/tools/ttysnoop.py
+++ b/tools/ttysnoop.py
@@ -14,6 +14,7 @@
 #
 # 15-Oct-2016   Brendan Gregg   Created this.
 # 13-Dec-2022   Rong Tao        Detect whether kfunc is supported.
+# 07-Jan-2023   Rong Tao        Support ITER_UBUF(CO-RE way)
 
 from __future__ import print_function
 from bcc import BPF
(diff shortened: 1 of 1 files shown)


tools/ttysnoop.py
```diff
@@ -147,18 +148,27 @@ def usage():
...
-    buf   = kvec->iov_base;
-    count = kvec->iov_len;
-
-    return do_tty_write(ctx, kvec->iov_base, kvec->iov_len);
+    switch (from->iter_type) {
+    case ITER_IOVEC:
+        kvec  = from->kvec;
+        buf   = kvec->iov_base;
+        count = kvec->iov_len;
+        break;
```
yonghong-song, line 162: So the address space is in kernel space for ITER_IOVEC, and it is in user space for ITER_UBUF. If this is the case, I guess, we should have add a parameter to do_tty_write to differentiate these two cases since the corresponding bpf_probe_read_* functions will be different.

Rtoax: Hi, Song. I'm curious about kernel/user space case. Since always, `ttysnoop.py`'s probe of `buf` has `__user` macro. Does it's prove that we should use `bpf_probe_read_user()` in both case. By the way, i tried `bpf_probe_read_kernel()` for `ITER_IOVEC`, it's cause `ttysnoop.py` run failed.

AND, In the new commit just now, I optimized the program for CO-RE.

Thanks a lot.

PR #4426, opened 2023-01-09
Fix get_kprobe_functions

jeromemarchand:
get_kprobe_functions will not only return a function that matches the regular expression, but also any function that starts with a substrings that matches it. This is obviously not the intended behavior.
The issue is easily fixed by replacing re.match by re.fullmatch.


chenhengqi (approved):
LGTM.

Issue #4417, opened 2023-01-04
libbpf-tools:some demo tools excute failed!such as biotop tcpstaus filetop....


LinuxVWQH:
Hi，when i made and installed these tools in my aarch-64 OS,I got these problem:
_**BIOTOP**_
![image](https://user-images.githubusercontent.com/74600097/210477035-79cdc74e-64c6-420e-8be8-884ed3f98a90.png)
![image](https://user-images.githubusercontent.com/74600097/210477064-2c09e7f4-1a4b-4f96-bdae-5c2cb646942b.png)
![image](https://user-images.githubusercontent.com/74600097/210477080-ce07a1e4-cf0a-4881-9c3d-b9148c298f4e.png)
**_FILETOP_**
![image](https://user-images.githubusercontent.com/74600097/210477319-fb5b4430-c1e2-4f0f-9e5b-fb2204a26406.png)
![image](https://user-images.githubusercontent.com/74600097/210477340-d1a833dc-4715-4901-bcc5-1b09e0a21b88.png)
_**TCPSTATUS**_
![image](https://user-images.githubusercontent.com/74600097/210477445-1f5b5c6b-ba27-4104-afb1-f6b7e5d6a098.png)
![image](https://user-images.githubusercontent.com/74600097/210477473-fd77af85-823d-496b-8892-fe45271ca737.png)
![image](https://user-images.githubusercontent.com/74600097/210477496-fc56375e-3564-41ec-be5f-746f7c1e8883.png)
...
I try to take some measures but not useful,please help me,thanks!
PS1:
**my OS info:**
Linux e04-g4 5.10.110-yocto-standard #1 SMP PREEMPT Tue Jan 3 09:00:28 UTC 2023 aarch64 GNU/Linux
PS2:
**my kernel config:**
`CONFIG_TASKS_TRACE_RCU=y
CONFIG_BPF_SYSCALL=y
CONFIG_TRACEPOINTS=y
CONFIG_UPROBES=y
CONFIG_BINARY_PRINTF=y

CONFIG_DEBUG_INFO_BTF=y
CONFIG_STACKTRACE=y
CONFIG_NOP_TRACER=y
CONFIG_RING_BUFFER=y
CONFIG_EVENT_TRACING=y
CONFIG_CONTEXT_SWITCH_TRACER=y
CONFIG_TRACING=y
CONFIG_FTRACE=y
CONFIG_FTRACE_SYSCALLS=y
[...]

PR #4429, opened 2023-01-10
fix the Makefile of libbpf-tools,make it support cross compile

LinuxVWQH:
Hi,bro!
When i buited these tools in libbpf-tools with yocto 4.x,I got these errors:https://github.com/iovisor/bcc/issues/4417#issue-1518224690.    Finally,I found a bug of libbpf-toos Makefile,the Makefile used 'uname -m' to confirm the ARCH,I think this way is reasonable.such as i buit them with yocto on X86-ubuntu system,the process is a classic cross compile,the ARCH's  value should take the target system  architecture.So i commite my bugfix, i hope you can pass it:)

### DIFF
--- a/libbpf-tools/Makefile
+++ b/libbpf-tools/Makefile
@@ -90,7 +90,7 @@ APPS = \
 	#
 
 # export variables that are used in Makefile.btfgen as well.
-export OUTPUT BPFTOOL BTFHUB_ARCHIVE APPS
+export OUTPUT BPFTOOL ARCH BTFHUB_ARCHIVE APPS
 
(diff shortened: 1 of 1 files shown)


libbpf-tools/Makefile
```diff
@@ -90,7 +90,7 @@ APPS = \
 	#
 
 # export variables that are used in Makefile.btfgen as well.
-export OUTPUT BPFTOOL ARCH BTFHUB_ARCHIVE APPS
+export OUTPUT BPFTOOL BTFHUB_ARCHIVE APPS
```
chenhengqi, line 93: Why do you remove `ARCH` here ?

PR #4437, opened 2023-01-16
Add Linux distro-independent code, remove bash executable path hardcoding

expoli:
The original code hard-coded the path of the bash executable file, lacking platform independence, causing the program to exit due to path errors.

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -698,6 +698,15 @@ class BPF(object):
                 raise e
             blacklist = set([])
 
+        avail_filter_file = "%s/tracing/available_filter_functions" % DEBUGFS
+        try:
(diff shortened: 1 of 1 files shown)


tools/bashreadline.py
```diff
@@ -15,8 +15,10 @@
 #
 # 28-Jan-2016    Brendan Gregg   Created this.
 # 12-Feb-2016    Allan McAleavy migrated to BPF_PERF_OUTPUT
+# 16-Jan-2023    Expoli added distribution independent code
 
 from __future__ import print_function
+from distutils import spawn
```
chenhengqi, line 21: This introduces a deprecated dependency. Perhaps we can add a new option to allow user specify path of bash.

PR #4440, opened 2023-01-16
Implement symbol lookup in zipped libraries.

michalgr:
Implement symbol lookup in zipped libraries.

On Android dynamic linker is capable of loading libraries directly from zip archives, provided that they are uncompressed and 4kb aligned. This pull request implements ability to set probes and lookup symbols in such libraries. In particular:
- minimal ZIP parser is added (`bcc_zip.c`)
- `bcc_elf_file` is extended to represent zipped libraries as well
- zip and apk files are checked for libs when parsing process' maps
- `bpf_attach_uprobe` adjusts file offsets and paths accordingly when probing libs inside a zip file
- libraries inside zip files can be referred to using the same notation as implemented in simpleperf: "/path/to/a/zip!/path/within/a/zip". For example: `trace '/data/app/com.example/base.apk!/lib/arm64-v8a/libtest.so:some_function'` traces `some_function` defined in `lib/arm64-v8a/libtest.so` embedded in `/data/app/com.example/base.apk` archive.

Test plan:
- use trace to trace and symbolicate functions defined in a zipped library
- use bpftrace to trace and symbolicate functions defined in a zipped library

### DIFF
--- a/src/cc/bcc_proc.c
+++ b/src/cc/bcc_proc.c
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+#include "bcc_proc.h"
+
 #include <ctype.h>
 #include <dirent.h>
 #include <fcntl.h>
@@ -29,9 +31,9 @@
 #include <sys/types.h>
 #include <unistd.h>
 
-#include "bcc_perf_map.h"
-#include "bcc_proc.h"
 #include "bcc_elf.h"
+#include "bcc_perf_map.h"
+#include "bcc_zip.h"
 
 #ifdef __x86_64__
 // https://www.kernel.org/doc/Documentation/x86/x86_64/mm.txt
@@ -121,12 +123,49 @@ static char *_procutils_memfd_path(const int pid, const uint64_t inum) {
   return path;
 }
 
+static int _procfs_might_be_zip_path(const char *path) {
+  return strstr(path, ".zip") || strstr(path, ".apk");
+}
+
+static char *_procfs_find_zip_entry(const char *path, int pid,
+                                    uint32_t *offset) {
+  char ns_relative_path[PATH_MAX];
+  int rc = snprintf(ns_relative_path, sizeof(ns_relative_path),
+                    "/proc/%d/root%s", pid, path);
+  if (rc < 0 || rc >= sizeof(ns_relative_path)) {
+    return NULL;
+  }
+
+  struct bcc_zip_archive *archive = bcc_zip_archive_open(ns_relative_path);
+  if (archive == NULL) {
+    return NULL;
+  }
+
+  struct bcc_zip_entry entry;
+  if (bcc_zip_archive_find_entry_at_offset(archive, *offset, &entry) ||
+      entry.compression) {
+    bcc_zip_archive_close(archive);
+    return NULL;
+  }
+
+  char *result = malloc(strlen(path) + entry.name_length + 3);
+  if (result == NULL) {
+    bcc_zip_archive_close(archive);
+    return NULL;
+  }
+
+  sprintf(result, "%s!/%.*s", path, entry.name_length, entry.name);
+  *offset -= entry.data_offset;
+  bcc_zip_archive_close(archive);
+  return result;
+}
+
 // return: 0 -> callback returned < 0, stopped iterating
 //        -1 -> callback never indicated to stop
 int _procfs_maps_each_module(FILE *procmap, int pid,
                              bcc_procutils_modulecb callback, void *payload) {
   char buf[PATH_MAX + 1], perm[5];
-  char *name;
+  char *name, *resolved_name;
   mod_info mod;
   uint8_t enter_ns;
   while (true) {
@@ -148,14 +187,25 @@ int _procfs_maps_each_module(FILE *procmap, int pid,
     if (!bcc_mapping_is_file_backed(name))
       continue;
 
+    resolved_name = NULL;
     if (strstr(name, "/memfd:")) {
-      char *memfd_name = _procutils_memfd_path(pid, mod.inode);
-      if (memfd_name != NULL) {
-        strcpy(buf, memfd_name);
-        free(memfd_name);
-        mod.name = buf;
+      resolved_name = _procutils_memfd_path(pid, mod.inode);
+      if (resolved_name != NULL) {
         enter_ns = 0;
       }
+    } else if (_procfs_might_be_zip_path(mod.name)) {
+      uint32_t zip_entry_offset = mod.file_offset;
+      resolved_name = _procfs_find_zip_entry(mod.name, pid, &zip_entry_offset);
+      if (resolved_name != NULL) {
+        mod.file_offset = zip_entry_offset;
+      }
+    }
+
+    if (resolved_name != NULL) {
+      strncpy(buf, resolved_name, PATH_MAX);
+      buf[PATH_MAX] = 0;
+      free(resolved_name);
+      mod.name = buf;
     }
 
     if (callback(&mod, enter_ns, payload) < 0)
--- a/src/cc/bcc_zip.c
+++ b/src/cc/bcc_zip.c
@@ -1,0 +1,388 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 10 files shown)


src/cc/bcc_zip.c
```diff
@@ -0,0 +1,323 @@
...
+
+  size_t size = lseek(fd, 0, SEEK_END);
+  void* data = mmap(NULL, size, PROT_READ, MAP_PRIVATE, fd, 0);
+  close(fd);
+
+  if (data == MAP_FAILED) {
+    return NULL;
+  }
+
+  struct bcc_zip_archive* archive = malloc(sizeof(struct bcc_zip_archive));
```
chenhengqi, line 219: Check the return value of malloc ?

src/cc/bcc_zip.c
```diff
@@ -0,0 +1,357 @@
...
+
+static int find_central_directory(struct bcc_zip_archive* archive) {
+  if (archive->size <= END_OF_CD_RECORD_SIZE) {
+    return -1;
+  }
+
+  int rc = -1;
+  uint32_t offset = archive->size - END_OF_CD_RECORD_SIZE;
+  int64_t limit = (int64_t)offset - (1 << 16);
+  for (; offset >= 0 && offset > limit && rc == -1; offset--) {
```
davemarchevsky, line 198: To confirm my understanding: this loop is necessary because the variable-length "comment" field at end of "end of central directory record" makes it impossible to find the beginning of the record w/o searching?

If so, can you add a comment explaining that `limit = ... (1 << 16)` above is the earliest place "end of central directory record"  can be?

Also, assuming my earlier "check the magic number before memcpying" comment on `parse_end_of_central_directory` makes sense, it should be possible to avoid checking every byte offset since `END_OF_CD_RECORD_MAGIC` is 4 bytes

davemarchevsky (changes requested):

src/cc/bcc_proc.c
```diff
@@ -121,12 +123,44 @@ static char *_procutils_memfd_path(const int pid, const uint64_t inum) {
...
+  }
+
+  struct bcc_zip_entry entry;
+  if (bcc_zip_archive_find_entry_at_offset(archive, *offset, &entry) ||
+      entry.compression) {
+    bcc_zip_archive_close(archive);
+    return NULL;
+  }
+
+  char *result = malloc(strlen(path) + entry.name_length + 3);
```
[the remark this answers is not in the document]
michalgr: There are quite a few unchecked `malloc`s in this file, should I just add a `malloc_or_die` helper and use it instead? IMO when malloc starts failing we cannot guarantee the quality of delivered results and killing the process would be the way to recover.

src/cc/bcc_zip.c
```diff
@@ -0,0 +1,357 @@
...
+
+static int read_cd_file_header(struct bcc_zip_archive* archive,
+                               uint32_t* offset,
+                               struct central_directory_file_header* header) {
+  if (check_access(archive, *offset, CD_FILE_HEADER_SIZE)) {
+    return -1;
+  };
+
+  memcpy(header, archive->data + *offset, CD_FILE_HEADER_SIZE - 8);
+  // skip padding
```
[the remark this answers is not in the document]
michalgr: I can try that. I avoided packed structs in the original version of this pull request because `packed` increases complexity by an order of magnitude and might be an invitation for surprises now and in the future if this code evolves. In particular:
- I am not aware of a comprehensive list of guarantees and restrictions that reasonable compilers provide/impose on packed structs across all of the architectures we need to support. How do I ensure that misaligned accesses introduced by packed structs do not generate `SIGBUS` on arm processors ?
- it might be not enough to make all 3 structs packed. They contain 2 and 4 byte fields, but they can appear in memory in location that is 1 byte aligned. Is compiler allowed to assume that 2-byte field accesses are always aligned ? We might need to force even lower alignment.
- One of the pitfalls to avoid is capturing pointers to misaligned fields. This will be a potential danger for pull requests modifying this file in the future.
- comprehensive testing becomes much harder, we now need to test on multiple platforms.

`memcpy` is an alternative that makes the concerns above disappear. It does copy too much data though.

There is a hybrid approach. I can turn all the fields into arrays of 2 or 4 bytes and add helpers to memcpy those arrays to uint16 and uint32. As a result there should be no alignment requirements and we will not memcpy more than necessary.

Please tell me which way to go.

src/cc/bcc_zip.c
michalgr: This is correct.

Can you elaborate on not checking every byte offset ? Do you want me to check the magic number and only if it matches call parse_end_of_central_directory ?

src/cc/bcc_proc.c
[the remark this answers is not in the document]
davemarchevsky: I saw that you added `malloc_or_die` in 2nd pass. I'm not a fan of `abort` in library code. Could you please get rid of `malloc_or_die` and instead return whatever error-like thing makes sense in the function `malloc` is called in? (`NULL`, `-1`, etc.)

src/cc/bcc_zip.c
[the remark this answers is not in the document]
davemarchevsky: > I am not aware of a comprehensive list of guarantees and restrictions that reasonable compilers provide/impose on packed structs across all of the architectures we need to support. How do I ensure that misaligned accesses introduced by packed structs do not generate SIGBUS on arm processors ?

Based on my googling, recent ARM standards (starting with [`v6`](https://patchwork.kernel.org/project/linux-arm-kernel/patch/1352205711-15787-1-git-send-email-dave.martin@linaro.org/) apparently, although I'm by no means knowledgeable wrt. ARM) won't throw on unaligned access. Am I mistaken here?

Separately, according to @yonghong-song, modern compilers should be smart enough to turn unaligned loads into single-byte loads on platforms where the former would cause an exception. 

> it might be not enough to make all 3 structs packed. They contain 2 and 4 byte fields, but they can appear in memory in location that is 1 byte aligned. Is compiler allowed to assume that 2-byte field accesses are always aligned ? We might need to force even lower alignment.

I think `__attribute__((packed, aligned(1)))` would satisfy "can appear in memory in location that is 1 byte aligned" and "no empty space between struct fields". Regardless, I think the `unaligned_uint{16,32}_t` approach you took in 2nd pass is equivalent and readable, so no need to do anything further here.

> One of the pitfalls to avoid is capturing pointers to misaligned fields. This will be a potential danger for pull requests modifying this file in the future.

I'm not too worried about this as I don't think code outside of `bcc_zip.c` will be directly interacting with these structs.

> comprehensive testing becomes much harder, we now need to test on multiple platforms.

True, but we need to test `bcc` on multiple platforms already

src/cc/bcc_zip.c
davemarchevsky: >   Can you elaborate on not checking every byte offset ? Do you want me to check the magic number and only if it matches call parse_end_of_central_directory ?

By this I meant that you could use something like [Boyer-Moore algorithm's "bad character" rule](https://en.wikipedia.org/wiki/Boyer%E2%80%93Moore_string-search_algorithm) to skip over potential starting offsets.  TL;DR of the heuristic:

Assume we're looking for string `1234` in a string of digits (unknowns represented by `X`). First we'll compare to the end of the haystack string:

```
XXXXX1234234
        1234
```

`4234` != `1234`, but instead of comparing against `3423` next, since we know that the string we're searching for ends in `4`, and there was a `4` in wrong place in our initial comparison, we can jump to comparing against whatever index results in that `4` being end of the string. 

But I wasn't expecting you to write any of that yourself, but maybe consider using something like `memmem` helper. Regardless, I don't feel strongly about doing any change here until since there's no data showing that this is slow. Just looping back.

davemarchevsky (changes requested):
Sorry for the egregious review time. I think that this is good-to-go aside from the `malloc_or_die` comment.

tests/cc/test_zip.cc
```diff
@@ -0,0 +1,80 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <cstring>
+
+#include "bcc_zip.h"
+#include "catch.hpp"
+
+#define ENTRY_NAME "libdebuginfo_test_lib.so"
```
danielocfb, line 22: Would it make sense to have a test for a hierarchical name as well, such as `/lib/arm64-v8a/libtest.so` that you mentioned in the pull request?

michalgr: Good idea. I added tests with a file in a subdirectory.

src/cc/bcc_proc.c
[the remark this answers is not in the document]
michalgr: I removed `malloc_or_die` and now this function just returns `NULL` on malloc failures.

michalgr:
Changes since last iteration:
- eliminate `malloc_or_die`
- add more tests
- rebase

PR #4442, opened 2023-01-18
Truncate long uprobe names and append its hash for differentiation

chengruizhe:
Sometime uprobe names can be quite long, e.g. when it contains the path to a binary in a docker container. This pull request will truncate a long uprobe name to 208 characters and append a 16 character hash of the name to the end in order to differentiate between similar names.

### DIFF
--- a/src/cc/api/BPF.cc
+++ b/src/cc/api/BPF.cc
@@ -858,17 +858,8 @@ namespace {
 constexpr size_t kEventNameSizeLimit = 224;
 
 std::string shorten_event_name(const std::string& name) {
-  constexpr size_t kHashSuffixLen = 16;
-  std::string res;
-  res.reserve(kEventNameSizeLimit);
-  res.assign(name);
-  res.resize(kEventNameSizeLimit - kHashSuffixLen);
-
-  std::stringstream stream;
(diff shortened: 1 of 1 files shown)


src/cc/api/BPF.cc
```diff
@@ -853,6 +853,26 @@ bool BPF::add_module(std::string module)
...
+namespace {
+
+constexpr size_t kEventNameSizeLimit = 224;
+
+std::string shorten_event_name(const std::string& name) {
+  constexpr size_t kHashSuffixLen = 16;
+  std::string res;
+  res.reserve(kEventNameSizeLimit);
+  res.assign(name);
+  res.resize(kEventNameSizeLimit - kHashSuffixLen);
```
chenhengqi, line 865: This can be replaced by a simple `std::string::substr()`.

src/cc/api/BPF.cc
```diff
@@ -853,6 +853,26 @@ bool BPF::add_module(std::string module)
...
+
+  std::stringstream stream;
+  size_t hash = std::hash<std::string>{}(name);
+  stream << std::hex << hash;
```
chenhengqi, line 869: Use `uint_to_hex()` instead.

Issue #4438, opened 2023-01-16
tools/slabratetop: call to undeclared function 'slab_address'


Rtoax:
When my system uses SLUB And the kernel version >=5.16, there are the following problems:

```bash
$ sudo ./slabratetop.py
In file included from /virtual/main.c:13:
include/linux/slub_def.h:162:26: warning: call to undeclared function 'slab_address'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
        void *object = x - (x - slab_address(slab)) % cache->size;
                                ^
include/linux/slub_def.h:162:46: error: invalid operands to binary expression ('void *' and 'unsigned int')
        void *object = x - (x - slab_address(slab)) % cache->size;
                           ~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~
include/linux/slub_def.h:164:8: error: incomplete definition of type 'struct slab'
                (slab->objects - 1) * cache->size;
                 ~~~~^
include/linux/kasan.h:13:8: note: forward declaration of 'struct slab'
struct slab;
       ^
In file included from /virtual/main.c:13:
include/linux/slub_def.h:184:31: warning: call to undeclared function 'slab_address'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
        return __obj_to_index(cache, slab_address(slab), obj);
                                     ^
include/linux/slub_def.h:184:31: error: incompatible integer to pointer conversion passing 'int' to parameter of type 'void *' [-Wint-conversion]
        return __obj_to_index(cache, slab_address(slab), obj);
                                     ^~~~~~~~~~~~~~~~~~
include/linux/slub_def.h:173:14: note: passing argument to parameter 'addr' here
                                          void *addr, void *obj)
                                                ^
include/linux/slub_def.h:190:13: error: incomplete definition of type 'struct slab'
        return slab->objects;
               ~~~~^
include/linux/kasan.h:13:8: note: forward declaration of 'struct slab'
struct slab;
       ^
2 warnings and 4 errors generated.
Traceback (most recent call last):
  File "/home/rongtao/Git/bcc/tools/./slabratetop.py", line 111, in <module>
    b = BPF(text=bpf_text)
        ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/bcc/__init__.py", line 475, in __init__
    raise Exception("Failed to compile BPF module %s" % (src_file or "<text>"))
[...]

PR #4447, opened 2023-01-23
tools/slabratetop: Fix error: incomplete definition of type 'struct s…

Rtoax:
Try fix https://github.com/iovisor/bcc/issues/4438 with @yonghong-song 's solution.

### DIFF
--- a/tools/slabratetop.py
+++ b/tools/slabratetop.py
@@ -14,6 +14,9 @@
 # Licensed under the Apache License, Version 2.0 (the "License")
 #
 # 15-Oct-2016   Brendan Gregg   Created this.
+# 23-Jan-2023   Rong Tao        Introduce kernel internal data structure and
+#                               functions to temporarily solve problem for
+#                               >=5.16(TODO: fix this workaround)
 
 from __future__ import print_function
 from bcc import BPF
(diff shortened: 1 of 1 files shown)


tools/slabratetop.py
```diff
@@ -65,6 +68,81 @@
...
+#endif
+};
+
+#define slab_folio(s)           (_Generic((s),                          \
+        const struct slab *:    (const struct folio *)s,                \
+        struct slab *:          (struct folio *)s))
+
+static inline void *slab_address(const struct slab *slab)
+{
+        return folio_address(slab_folio(slab));
```
chenhengqi, line 143: Please fix the following warnings for kernel < 5.16.
```
/virtual/main.c:82:16: warning: implicit declaration of function 'folio_address' is invalid in C99 [-Wimplicit-function-declaration]
        return folio_address(slab_folio(slab));
               ^
/virtual/main.c:82:16: warning: incompatible integer to pointer conversion returning 'int' from a function with result type 'void *' [-Wint-conversion]
        return folio_address(slab_folio(slab));
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2 warnings generated.
```

Rtoax: Hi, Hengqi, Could we just make slab_address() return NULL? because it's never used in slabratetop.py, we don't have to solve folio_address() for different kernel version, do you agree with me?

Rtoax:
Hi, @yonghong-song @chenhengqi , I found the patch below to fix the problem, just like `struct memcg_cache_params {};` in the script. Please review the patches below to see if they are available

```diff
diff --git a/tools/slabratetop.py b/tools/slabratetop.py
index ac44b2bd..0427bf19 100755
--- a/tools/slabratetop.py
+++ b/tools/slabratetop.py
@@ -65,6 +65,13 @@ bpf_text = """
 // 5.9, but it does not hurt to have it here for versions 5.4 to 5.8.
 struct memcg_cache_params {};

+struct slab {};
+
+static inline void *slab_address(const struct slab *slab)
+{
+    return NULL;
+}
+
 #ifdef CONFIG_SLUB
 #include <linux/slub_def.h>
 #else
```

tools/slabratetop.py
Rtoax: This does solve the problem:

```diff
+static inline void *slab_address(const struct slab *slab)
+{
+    return NULL;
+}
```

yonghong-song:
Does this change work for pre-5.16 kernel? E.g., 5.15 kernel is not old enough so I would like the change can work for both pre-5.16 and later kernels. I am okay with the kernel version check since the structure definition does not change in the affecting kernel commit.

Rtoax:
> Does this change work for pre-5.16 kernel? E.g., 5.15 kernel is not old enough so I would like the change can work for both pre-5.16 and later kernels. I am okay with the kernel version check since the structure definition does not change in the affecting kernel commit.

Hi, Song, sorry for the late reply, I tried this modification and it works fine in the 5.14.0 kernel. Do I need to test on more kernel versions?

yonghong-song:
Indeed. I tried 5.12 and 5.19. Both works fine.

PR #4459, opened 2023-02-01
 examples/networking: add tcp_mon_block

agentzex:
tcp_mon_block uses netlink TC, kernel tracepoints and kprobes to monitor outgoing connections from given PIDs (usually HTTP web servers) and block connections to all addresses initiated from them (acting like an in-process firewall) , unless they are listed in allow_list. This way the process can mitigate connections to unexpected malicious addresses, like to a C2 server, happening after a successful exploitation of the process
Signed-off-by: Dor A <cybernpy@gmail.com >


agentzex:
Hello  :)
Are you still accepting PRs for new examples ?

PR #4464, opened 2023-02-06
tools/scsilatency: add scsilatency tool

weibang6liu:
We used BCC to monitor I/O stack performance on mobile(AArch64) platform, and found it's very efficient and use easily. Meanwhile, we found lack of some tools in multiple layers，then we devloped a tools for SCSI layer.
The mobile storage were UFS mostly, and UFS device follows SCSI protocol. This tool can be used for monitor device performance in SCSI layer.
We hope this tools can help more people.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,7 @@ pair of .c and .py files, and some are directories of files.
 - tools/[runqlat](tools/runqlat.py): Run queue (scheduler) latency as a histogram. [Examples](tools/runqlat_example.txt).
 - tools/[runqlen](tools/runqlen.py): Run queue length as a histogram. [Examples](tools/runqlen_example.txt).
 - tools/[runqslower](tools/runqslower.py): Trace long process scheduling delays. [Examples](tools/runqslower_example.txt).
+- tools/[scsilatency](tools/scsilatency.py): Summarize SCSI layer I/O latency as a histogram . [Examples](tools/scsilatency_example.txt).
 - tools/[shmsnoop](tools/shmsnoop.py): Trace System V shared memory syscalls. [Examples](tools/shmsnoop_example.txt).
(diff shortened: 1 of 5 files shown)


brendangregg:
I published scsilatency in BPF Performance Tools (2019) and online here https://github.com/brendangregg/bpf-perf-tools-book/blob/master/originals/Ch09_Disks/scsilatency.bt

It is indeed useful, but I did not put it in bcc as I didn't want to clutter up bcc with too many tools and create a maintenance burden. You've already had to deal with this, as you've encountered the Linux 4.7 change. :-)

Note that I broke down flags into more useful opcodes than just read or write:

```
        @opcode[0x00] = "TEST_UNIT_READY";
        @opcode[0x03] = "REQUEST_SENSE";
        @opcode[0x08] = "READ_6";
        @opcode[0x0a] = "WRITE_6";
        @opcode[0x0b] = "SEEK_6";
        @opcode[0x12] = "INQUIRY";
        @opcode[0x18] = "ERASE";
        @opcode[0x28] = "READ_10";
        @opcode[0x2a] = "WRITE_10";
        @opcode[0x2b] = "SEEK_10";
        @opcode[0x35] = "SYNCHRONIZE_CACHE";
```

If we include your version in bcc, I think it should also break down the flags fully with the -F option like the above, instead of just Read and Write.

Also please note from my tool:

```
 * Copyright (c) 2019 Brendan Gregg.
 * Licensed under the Apache License, Version 2.0 (the "License").
 * This was originally created for the BPF Performance Tools book
 * published by Addison Wesley. ISBN-13: 9780136554820
 * When copying or porting, include this comment.
```

If I were porting a tool, I'd include that comment; if I were borrowing or inspired by another tool, I'd have something shorter just to say thanks, either in the tool or in the man page.

weibang6liu:
> 

thank you for your reply and accept your opinion sincerely. I learned from your scsilatency.bt and removed other flags operations because of monitoring Read/Write only, I will add other flags and copyright note in later commit. 
thank you  again for your bpftrace guidance.

PR #4466, opened 2023-02-08
tools/scsisnoop: add scsisnoop tool

weibang6liu:
We used BCC to monitor I/O stack performance on mobile(AArch64) platform, and found it's very efficient and use easily. Meanwhile, we found lack of some tools in multiple layers，then we devloped a tools for SCSI layer.
The mobile storage were UFS mostly, and UFS device follows SCSI protocol. This tool can be used for monitor device performance in SCSI layer.
We hope this tool can help more people.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,7 @@ pair of .c and .py files, and some are directories of files.
 - tools/[runqlat](tools/runqlat.py): Run queue (scheduler) latency as a histogram. [Examples](tools/runqlat_example.txt).
 - tools/[runqlen](tools/runqlen.py): Run queue length as a histogram. [Examples](tools/runqlen_example.txt).
 - tools/[runqslower](tools/runqslower.py): Trace long process scheduling delays. [Examples](tools/runqslower_example.txt).
+- tools/[scsisnoop](tools/scsisnoop.py): Trace SCSI layer device I/O with PID and latency. [Examples](tools/scsisnoop_example.txt).
 - tools/[shmsnoop](tools/shmsnoop.py): Trace System V shared memory syscalls. [Examples](tools/shmsnoop_example.txt).
 - tools/[sofdsnoop](tools/sofdsnoop.py): Trace FDs passed through unix sockets. [Examples](tools/sofdsnoop_example.txt).
 - tools/[slabratetop](tools/slabratetop.py): Kernel SLAB/SLUB memory cache allocation rate top. [Examples](tools/slabratetop_example.txt).
--- a/man/man8/scsisnoop.8
+++ b/man/man8/scsisnoop.8
@@ -1,0 +1,80 @@
+.TH scsisnoop 8  "2022-10-10" "USER COMMANDS"
+.SH NAME
+scsisnoop \- Trace SCSI layer I/O and print details incl. issuing PID.
+.SH SYNOPSIS
+.B scsisnoop [\-h]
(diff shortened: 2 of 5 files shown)


brendangregg:
scsisnoop it rings a bell but I can't put my finger on the previous version (I might be thinking of scsi.d, written by other people for an older tracer around 2009, but I can't even find it online anymore). I like how this version fits within 80 chars. Please consider adding columns for host and status when using a -r (result) option (based on what I did in [scsiresult](https://github.com/brendangregg/bpf-perf-tools-book/blob/master/originals/Ch09_Disks/scsiresult.bt)), e.g.:

```
# ./scsisnoop -r
TIME(s)     COMM           PID    DISK    T SECTOR     BYTES  LAT(ms) HOST STAT
0.000000    kworker/u16:2  29466  sda     R 1867600    4096      0.21    0    0
0.000330    kworker/3:1H   193    sda     R 1881480    4096      0.19    0    0
0.001666    InputReader    1228   sda     R 1847944    155648    0.84    3    8
0.003512    InputReader    1228   sda     R 1848256    204800    0.96    0    0
```

Now, it isn't clear what host and stat mean at this point, but it fits within 80 chars which is immensely useful for various reasons. An extended version, say, -R, can do the translations and not worry about the 80 char limit:

```
# ./scsisnoop -R
TIME(s)     COMM           PID    DISK    T SECTOR     BYTES  LAT(ms)  HOST STATUS
0.000000    kworker/u16:2  29466  sda     R 1867600    4096      0.21  OK   GOOD
0.000330    kworker/3:1H   193    sda     R 1881480    4096      0.19  OK   GOOD
0.001666    InputReader    1228   sda     R 1847944    155648    0.84  TRANSPORT_DISRUPTED RESERVATION_CONFLICT
0.003512    InputReader    1228   sda     R 1848256    204800    0.96  OK   GOOD
```

Normally I'd put those as neat fixed-width columns that are right-justified (as it's the result; I left-justify input), but the content is so variable in this case (from "OK" to "INTERMEDIATE_CONDITION_MET") that I'd consider giving up and let them overflow, especially as they are the final columns on the right and won't perturb the columns on the left. So those columns in my examples are really "%-4s %s". I also do this for tools that include filesystem path names, which can be anything from a few charters to over a hundred: I put that as the final column as a %s. It's ok that this is a bit messy as most of the time I'd be using the -r and seeing a lot of "0    0"s.

weibang6liu:
> 

First of all, thanks brendan for support and appreciation for this tool.
You memtioned it's not clear for adding the Host and Status columns, maybe we can patch later when clear requirements. 
In my opinion, the current version is efficient and sufficient for observing SCSI IO status and latency.

Could you continue to review, and look forward to merge into project. 
Thank you.

weibang6liu:
hi @yonghong-song 
    For this commit,  almost request checks failed or cancelled, I have no idea. 
    I check a fail log, found the error was "AssertionError: False is not true : rc was 126 ", " timeout: failed to run command '/bcc/tools/scsisnoop.py': Permission denied" ,  I checked scsisnoop.py permission is 'a+x', I thought is enough for execution. 
  
could you tell me what to do for this? thank you.

brendangregg:
> maybe we can patch later when clear requirements.

I mocked up output for you, referenced example code, and explained column justification -- that's as clear as it gets.

PR #4469, opened 2023-02-08
Json format output for tools

tariromukute:
Hi,

We used bcc tools to do performance analysis of different virtualised applications, trying to understand how they use the underlying resource. We had to provide graphs for the publication. We found it easier to graph the results when the output is in json format (especially for distribution graphs). Since we had to do this for a number of tools and applications adding a json format output option for each of the tools made it easier. In the case that someone finds this option useful this PR:

1. Adds json format output option to the tools (`-j` or `--json`)
2. Adds test cases to validate the returned json output `tests/python/test_tools_json_output.py`

In our case we used plotly to create graphs, see the sample below.

hardirqs_count

### DIFF
--- a/tests/python/test_tools_json_output.py
+++ b/tests/python/test_tools_json_output.py
@@ -48,23 +48,20 @@ class SmokeTests(TestCase):
     # Use this for commands that don't have a built-in timeout, so we have
     # to Ctrl-C out of them by sending SIGINT. If that still doesn't stop
     # them, send a kill signal 5 seconds later.
-    def run_with_int(self, command, timeout=5, kill_timeout=5,
+    def run_with_int(self, command, output={}, timeout=5, kill_timeout=5,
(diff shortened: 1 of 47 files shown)


chenhengqi:
cc @brendangregg

yonghong-song:
I see this 'We had to provide graphs for the publication'. We probably cannot just add jason support if this intends to be used for research only. cc @brendangregg

tariromukute:
Hi @yonghong-song, thank you for checking this out. Yeah our use case was for research, however I assume this can be useful outside research e.g., streaming the logs in production to Promotheus and Grafana (although we would need to implement an exporter first). I would presume it will be easier to ETL the output of these tools when they are in json format for production visualisation of their results (especially for distribution graphs). 

I also noted that there was already a function `print_json_hist` which was not being used by the tools. I also presume there was an intention to have this function called from within the tools.

PR #4470, opened 2023-02-08
tools/f2fsslower: add f2fsslower for F2FS

weibang6liu:
Add f2fsslower tool, which can be used for tracing mobile platform IO operations in F2FS layer.
F2FS (Flash Friendly File System) is an open source flash file system 
designed specifically for NAND-based storage devices, which currently mainly used on mobile platforms.

### DIFF
--- a/tools/f2fsslower.py
+++ b/tools/f2fsslower.py
@@ -86,7 +86,7 @@ struct data_t {
     char file[DNAME_INLINE_LEN];
 };
 BPF_HASH(entryinfo, u64, struct val_t);
-BPF_PERF_OUTPUT(events);
+PERF_TABLE
 //
 // Store timestamp and size on entry
 //
@@ -202,7 +202,7 @@ static int trace_return(struct pt_regs *ctx, int type)
         return 0;
     bpf_probe_read_kernel(&data.file, sizeof(data.file), (void *)qs.name);
     // output
-    events.perf_submit(ctx, &data, sizeof(data));
+    PERF_OUTPUT_CTX
     return 0;
 }
 int trace_read_return(struct pt_regs *ctx)
@@ -246,6 +246,19 @@ if args.pid:
     bpf_text = bpf_text.replace('FILTER_PID', 'pid != %s' % pid)
 else:
     bpf_text = bpf_text.replace('FILTER_PID', '0')
+
+if BPF.kernel_struct_has_field(b'bpf_ringbuf', b'waitq') == 1:
+    PERF_MODE = "USE_BPF_RING_BUF"
+    bpf_text = bpf_text.replace('PERF_TABLE',
+                                'BPF_RINGBUF_OUTPUT(events, 64);')
+    bpf_text = bpf_text.replace('PERF_OUTPUT_CTX',
+                                'events.ringbuf_output(&data,sizeof(data),0);')
+else:
+    PERF_MODE = "USE_BPF_PERF_BUF"
+    bpf_text = bpf_text.replace('PERF_TABLE', 'BPF_PERF_OUTPUT(events);')
+    bpf_text = bpf_text.replace('PERF_OUTPUT_CTX',
+                                'events.perf_submit(ctx,&data,sizeof(data));')
+
 if debug or args.ebpf:
     print(bpf_text)
     if args.ebpf:
@@ -308,9 +321,18 @@ else:
           "BYTES", "OFF_KB", "LAT(ms)", "FILENAME"))
 
 # read events
-b["events"].open_perf_buffer(print_event, page_cnt=64)
+
+# loop with callback to print_event
+if PERF_MODE == "USE_BPF_RING_BUF":
+    b["events"].open_ring_buffer(print_event)
+else:
+    b["events"].open_perf_buffer(print_event, page_cnt=64)
+
 while 1:
     try:
-        b.perf_buffer_poll()
+        if PERF_MODE == "USE_BPF_RING_BUF":
+            b.ring_buffer_poll()
+        else:
+            b.perf_buffer_poll()
     except KeyboardInterrupt:
         exit()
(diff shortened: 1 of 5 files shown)


brendangregg:
Would f2fs have enough widespread use to justify inclusion in bcc? Where is it mostly adopted, is it Pixel phones?

I've published over 500 useful tracing tools (for various different tracers) but I've only included around 50 of them in bcc, because we'd like it to have the "best of" collection that is well maintained and working. I would guess there are over one hundred file systems for Linux, and over a dozen in mainline, so I'd really want to think about whether this is a commonly used file system or one of the niche ones.

If we determine this shouldn't be in bcc at the moment, I'd recommend creating a Samsung github repo called "f2fs eBPF tools" and put it there. It can always be published somewhere else first and then moved to bcc if there is demand.

netedwardwu:
```
Would f2fs have enough widespread use to justify inclusion in bcc? Where is it mostly adopted, is it Pixel phones?
```
Just sharing my experience...

As I know, f2fs is a popular file system in Android platform nowaday, since it becomes more mature.
Not only Android phones but also some Android TV devices have adopted it.
When comparing f2fs with ext4, the score from some benchmark tools(ex: Antutu) that f2fs is better than ext4.
Everybody likes to compare different products with benchmark score before buying some.

BTW, running f2fsslower on my Android platform, and it seems to be working fine.

Thanks

weibang6liu:
The F2FS system is currently widely used in Android mobile platforms, including mobile phones, TVs, and vehicle systems. 
 It is friendly to flash memory and makes it perform better. Compared with ext4, it has better random read and write performance.
 At present, the data partition of most mobile phone systems adopts F2FS to ensure a better experience.
 Google's Android official guidance, when prompting users to update the system, also recommends choosing F2FS, 
 see the link(**Partition requirements** part):https://source.android.com/docs/core/ota/dynamic-system-updates ,
 Starting from linux 3.8, F2FS has been officially added to the linux kernel.
 At the same time, there are currently many papers discussing the performance optimization of F2FS. 
 In general, F2FS is a popular file system on Android platform. 
thanks @netedwardwu for sharing.

tools/f2fsslower.py
```diff
@@ -0,0 +1,318 @@
...
+#define TRACE_WRITE        1
+#define TRACE_OPEN        2
+#define TRACE_FSYNC        3
+struct val_t {
+    u64 ts;
+    u64 offset;
+    struct file *fp;
+};
+struct data_t {
+    // XXX: switch some to u32's when supported
```
chenhengqi, line 80: ?

tools/f2fsslower.py
```diff
@@ -0,0 +1,318 @@
...
+
+# header
+if (csv):
+    print("ENDTIME_us,TASK,PID,TYPE,BYTES,OFFSET_b,LATENCY_us,FILE")
+else:
+    if min_ms == 0:
+        print("Tracing f2fs operations")
+    else:
+        print("Tracing f2fs operations slower than %d ms" % min_ms)
+    print("%-8s %-14s %-6s %1s %-7s %-8s %7s %s" % ("TIME", "COMM", "PID", "T",
```
chenhengqi, line 309: `%7s` for pid.

tools/f2fsslower.py
```diff
@@ -0,0 +1,318 @@
...
+if BPF.get_kprobe_functions(b'f2fs_file_read_iter'):
+    b.attach_kprobe(event="f2fs_file_read_iter", fn_name="trace_read_entry")
+else:
+    b.attach_kprobe(event="generic_file_read_iter", fn_name="trace_read_entry")
+b.attach_kprobe(event="f2fs_file_write_iter", fn_name="trace_write_entry")
+b.attach_kprobe(event="f2fs_file_open", fn_name="trace_open_entry")
+b.attach_kprobe(event="f2fs_sync_file", fn_name="trace_fsync_entry")
+if BPF.get_kprobe_functions(b'f2fs_file_read_iter'):
+    b.attach_kretprobe(event="f2fs_file_read_iter",
+                       fn_name="trace_read_return")
```
chenhengqi, line 293: Keep this in single line.

tools/f2fsslower.py
weibang6liu: ? 
 I thought "%-6s" is enough for pid.

tools/f2fsslower.py
weibang6liu: I want to keep this in one line, but if do that, it would't meet pep8("E501 line too long (80 > 79 characters)") rule.

tools/f2fsslower.py
```diff
@@ -0,0 +1,318 @@
...
+    ./f2fsslower 1           # trace operations slower than 1 ms
+    ./f2fsslower -j 1        # ... 1 ms, parsable output (csv)
+    ./f2fsslower 0           # trace all operations (warning: verbose)
+    ./f2fsslower -p 185      # trace PID 185 only
+"""
+parser = argparse.ArgumentParser(
+    description="Trace common f2fs file operations slower than a threshold",
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    epilog=examples)
+parser.add_argument("-j", "--csv", action="store_true",
```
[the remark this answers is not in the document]
weibang6liu: This suggestion has been adopted，see ‘6c3e9f5’ commit.

tools/f2fsslower.py
```diff
@@ -0,0 +1,318 @@
...
+else:
+    if min_ms == 0:
+        print("Tracing f2fs operations")
+    else:
+        print("Tracing f2fs operations slower than %d ms" % min_ms)
+    print("%-8s %-14s %-6s %1s %-7s %-8s %7s %s" % ("TIME", "COMM", "PID", "T",
+          "BYTES", "OFF_KB", "LAT(ms)", "FILENAME"))
+
+# read events
+b["events"].open_perf_buffer(print_event, page_cnt=64)
```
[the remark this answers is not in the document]
weibang6liu: Thank you for your sugguestion, I add ringbuf process on this commit [fcb1601](https://github.com/iovisor/bcc/pull/4470/commits/fcb1601afc5747bd3e34b6c6107b8f534851122e)

chenhengqi:
Close in favor of #4782 .

PR #4471, opened 2023-02-09
libbpf-tools/offcputime: Use CO-RE operation to read kernel data

ekyooo:
Use CO-RE API to read the value of a field from task_struct.

Signed-off-by: Eunseon Lee <es.lee@lge.com>

### DIFF
--- a/libbpf-tools/offcputime.bpf.c
+++ b/libbpf-tools/offcputime.bpf.c
@@ -66,15 +66,15 @@ int BPF_PROG(sched_switch, bool preempt, struct task_struct *prev, struct task_s
 	u32 pid;
 
 	if (allow_record(prev)) {
-		pid = prev->pid;
+		pid = BPF_CORE_READ(prev, pid);
 		/* To distinguish idle threads of different cores */
 		if (!pid)
 			pid = bpf_get_smp_processor_id();
 		i_key.key.pid = pid;
-		i_key.key.tgid = prev->tgid;
+		i_key.key.tgid = BPF_CORE_READ(prev, tgid);
 		i_key.start_ts = bpf_ktime_get_ns();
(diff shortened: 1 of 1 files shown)


libbpf-tools/offcputime.bpf.c
```diff
@@ -66,15 +66,15 @@ int BPF_PROG(sched_switch, bool preempt, struct task_struct *prev, struct task_s
 	u32 pid;
 
 	if (allow_record(prev)) {
-		pid = prev->pid;
+		pid = BPF_CORE_READ(prev, pid);
```
chenhengqi, line 69: `tp_btf` is BTF-awared, pointer chasing should work, right ?

ekyooo: Oh... I didn't know that. Thank you for the information. Does `tp_btf` support depend on kernel version?
A segfault occurred when running the default version of offfcputime on a Linux v5.4 kernel environment. So I switched tp_btf to tracepoint as below. So this patch was needed.
```
-SEC("tp_btf/sched_switch")
+SEC("tracepoint/sched/sched_switch")
```
Could you please tell me if `tp_btf` is supported in Linux v5.4 as well?
Thank you in advance.

ekyooo: Ok. I got information from another engineer that tp_btf is not supported in Linux v5.4. I abandon this patch. Thank you.

chenhengqi: See https://github.com/iovisor/bcc/issues/4231.

ekyooo: Oh..!! Thank you. I'll read it carefully.

PR #4484, opened 2023-02-15
libbpf-tools/cachestat: Introduce tracepoint writeback_dirty_{page,folio}

Rtoax:
Add tracepoint:writeback:writeback_dirty_{page,folio} support.

kernel commit 9fb0a7da0c52("writeback: add more tracepoints") add writeback_dirty_page tracepoint, kernel commit b9b0ff61eef5("mm/writeback: Convert tracing writeback_page_template to folios") replace writeback_dirty_page to writeback_dirty_folio.

Both folio_account_dirtied() and account_page_dirtied() are static functions and they may be gone during compilation.

### DIFF
--- a/libbpf-tools/cachestat.c
+++ b/libbpf-tools/cachestat.c
@@ -5,6 +5,7 @@
 //  8-Mar-2021   Wenbo Zhang   Created this.
 // 30-Jan-2023   Rong Tao      Add kprobe and use fentry_can_attach() decide
 //                             use fentry/kprobe
+// 15-Feb-2023   Rong Tao      Add tracepoint writeback_dirty_{page,folio}
 #include <argp.h>
 #include <signal.h>
 #include <stdio.h>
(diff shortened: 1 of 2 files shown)


libbpf-tools/cachestat.c
```diff
@@ -164,8 +165,20 @@ int main(int argc, char **argv)
 	}
 	if (kprobe_exists("folio_account_dirtied")) {
 		bpf_program__set_autoload(obj->progs.kprobe_account_page_dirtied, false);
-	} else {
+		bpf_program__set_autoload(obj->progs.tracepoint__writeback_dirty_folio, false);
```
chenhengqi, line 168: Please separate kprobe and tracepoint checks.

Rtoax: Hi, I'm not separate kprobe/tracepoint because those are same thing in diff kernel version.

* folio_account_dirtied:  v5.15
* account_page_dirtied: v2.6.29 - v5.15
* tp:writeback:writeback_dirty_page: v3.8 - v5.15
* tp:writeback:writeback_dirty_folio:  v5.15 - nowest

These observation points actually mean the same thing, so I put them together. Follow the if-else statement to choose one of them. Do you think it also makes some sense?

yonghong-song:
LGTM. Thanks!

PR #4486, opened 2023-02-17
bcc/python: Add ring_buffer_epoll_fd() API to get fd of ring buffer.

chaudron:
Add an API for Python programs to get the file descriptor of the ring buffer. This API is named ring_buffer_epoll_fd() and can be used to add async event handling to python programs.

For example the common loop in the examples:

  try:
      while 1:
          b.ring_buffer_poll()
  except KeyboardInterrupt:
      sys.exit()

Can be re-written as:

  ring_fd = b.ring_buffer_epoll_fd()
  epoll = select.epoll()
  epoll.register(ring_fd, select.EPOLLIN)

  try:
      while 1:
          events = epoll.poll()
          for fd, event in events:
              if fd == ring_fd:
                  b.ring_buffer_consume()
  except KeyboardInterrupt:
      sys.exit()

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -1696,6 +1696,12 @@ int bpf_consume_ringbuf(struct ring_buffer *rb) {
     return ring_buffer__consume(rb);
 }
 
+/* Get an fd that can be used to sleep until data is available in the
+ * ring(s). */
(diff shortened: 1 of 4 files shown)


chaudron:
Found a problem with this approach, will close the MR.

PR #4488, opened 2023-02-17
libbpf-tools: match update return type in bpf_map_try_or_init()

inwardvessel:
I'm finding an issue at times where the comparison: `err != -EEXIST` does not work as intended due to the type used to store `err`. There have been cases where `bpf_map_update_elem()` returns `-EEXIST` (-17), but the conditional is still entered.

Adding a log message inside the conditional like this:
```bpf_printk("update fail err %ld %d\n", err, err);```

can result in:
```... bpf_trace_printk: update fail err 4294967279, -17```

... which might show why using a `long` results in something other than the actual returned error value.

The type returned by `bpf_map_update_elem()` is `int`, so the change below keeps things consistent and avoids the unintended behavior above.

### DIFF
--- a/libbpf-tools/maps.bpf.h
+++ b/libbpf-tools/maps.bpf.h
@@ -10,7 +10,7 @@ static __always_inline void *
 bpf_map_lookup_or_try_init(void *map, const void *key, const void *init)
 {
 	void *val;
-	long err;
+	int err;
 
 	val = bpf_map_lookup_elem(map, key);
 	if (val)

### CODE REVIEW

**Comment by inwardvessel** [comment]
I'll close this, since @anakryiko pointed out to me that the return type of the underlying bpf syscall is `long`. seems I was looking at:
`int (*map_update_elem)(struct bpf_map *map, void *key, void *value, u64 flags);`
 in the bpf_map_ops struct  and assuming `int` was consistent all the way up.

I'm going to look into the generated bpf assembly some more to get a better understanding of the observed behavior.

**Comment by yonghong-song** [comment]
@inwardvessel right, the return type for bpf_map_update_elem() is long. In the above example, `bpf_printk` is a macro defined in bpf_helpers.h. In this case, it eventually resolved to `bpf_trace_printk` and eventually kernel function `bstr_printf` is called. In your above example,
```
 bpf_printk("update fail err %ld %d\n", err, err);
```
If bpf code is correctly generated which I think it is based on a simple example. The issue might be inside the kernel, particularly in `bstr_printf` or the interaction between `bpf_trace_printk` and `bstr_printf`. Please take a look. Thanks!

PR #4492, opened 2023-02-22
bcc: fix indentation error

liyueyi:
Fix indentation error in get_kprobe_functions(), it can lead to function name mismatch.

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -714,7 +714,7 @@ class BPF(object):
                 elif in_init_section == 1:
                     if fn == b'__init_end':
                         in_init_section = 2
-                    continue
+                        continue
                 # Skip all functions defined between __irqentry_text_start and
                 # __irqentry_text_end
                 if in_irq_section == 0:
@@ -730,7 +730,7 @@ class BPF(object):
                 elif in_irq_section == 1:
                     if fn == b'__irqentry_text_end':
                         in_irq_section = 2
-                    continue
+                        continue
                 # All functions defined as NOKPROBE_SYMBOL() start with the
                 # prefix _kbl_addr_*, blacklisting them by looking at the name
                 # allows to catch also those symbols that are defined in kernel


chenhengqi:
I think the original implementation is correct. What kind of mismatch have you met ?

liyueyi:
> 
I have tested 2 functions, here is result:
```sh
root@localhost:/# cat /proc/kallsyms |grep tcp_drop
ffffffe07cc2fd30 t tcp_drop
root@localhost:/# tcpdrop
ERROR: tcp_drop() kernel function not found or traceable. Older kernel versions not supported.
function tcp_drop() missed.
```
try another function 'ext4slower' :
add debug log:
```py
# Common file functions. See earlier comment about generic_file_read_iter().
if BPF.get_kprobe_functions(b'ext4_file_read_iter'):
    b.attach_kprobe(event="ext4_file_read_iter", fn_name="trace_read_entry")
else:
    print("get ext4_file_read_iter fail")
    b.attach_kprobe(event="generic_file_read_iter", fn_name="trace_read_entry")
```
test:
```sh
root@localhost:/# ext4slower
get ext4_file_read_iter fail
Tracing ext4 operations slower than 10 ms
TIME     COMM           PID    T BYTES   OFF_KB   LAT(ms) FILENAME
```

Perhaps I was too hasty  in this commit,  maybe it just not work in my enviroment.

chenhengqi:
Does your patch fix the `tcpdrop` problem ?

netedwardwu:
By your patch, symbol parsing will bypass below two conditions
```
                # Skip all functions defined between __init_begin and
                # __init_end
```
```
                # Skip all functions defined between __irqentry_text_start and
                # __irqentry_text_end
```
So please don't use this patch...did you check your /proc/kallsyms for detail?

For example:
https://github.com/iovisor/bcc/commit/a090b462533fdb7d873f2d63afca2bdabd04f0a5
ARM64 doesn't use the blacklist between  __irqentry_text_start and __irqentry_text_end which x86 did.

So what is your condition in /proc/kallsyms?

PR #4494, opened 2023-02-23
Bahamasbahamas patch 1

bahamasbahamas:
display the average time and count as in the funclatency tool below the histogram output


Bojun-Seo:
Could you add a bit more explanation about this PR?
- What was the original output/result before this change?
- After applying this PR, how does the output change?

Also, it would be great if you could include a clear description of each patch in the commit messages.
I think it’s better to have one commit = one logical change. Right now some commits seem to mix unrelated changes.
Lastly, we don’t usually include the author’s ID in the PR title, so it would be better to rephrase the title as well.

Thanks!

PR #4502, opened 2023-03-01
Memleak libbpf port

inwardvessel:
This is a conversion of the existing memleak.py tool.

The BPF program is structured similarly to the original. One addition is the explicit uprobe definitions. Another notable change is within the `update_statistics_{add,del}()` functions. There seemed to be a need to synchronize the incrementing (or decrementing) of two integer members on the same struct. To avoid having the userspace side potentially seeing inconsistent values, the two members in question were combined within a single u64 bitfield. The `combined_alloc_info` struct (see memleak.h) was changed to a `union` so that access to individual fields (via anonymous struct) and and the combined bits could be performed. It allowed for the atomic incrementing/decrementing of the two original members in question. Open to feedback on this.

The userspace program was set up to allow all the same command line options as the original. Some notable items are:
- using blazesym for address to symbol resolution
- using "libc.so.6" as the default object in which to attach uprobes, can specify otherwise on CLI
- makes use of global variables in the bpf program for config/settings
- disables/enables probes based on requested settings
- synchronizes the call to `exec()` until after bpf program has been attached to try so probes can't be missed
- added some macros to do code generation of setting up uprobes

### DIFF
--- a/libbpf-tools/memleak.bpf.c
+++ b/libbpf-tools/memleak.bpf.c
@@ -1,0 +1,388 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (c) 2023 Meta Platforms, Inc. and affiliates.
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+
+#include "maps.bpf.h"
+#include "memleak.h"
+
+const volatile size_t min_size = 0;
+const volatile size_t max_size = -1;
+const volatile size_t page_size = 4096;
+const volatile __u64 sample_rate = 1;
+const volatile bool trace_all = false;
+const volatile __u64 stack_flags = 0;
+const volatile bool wa_missing_free = false;
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__type(key, pid_t);
+	__type(value, u64);
+	__uint(max_entries, 10240);
+} sizes SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__type(key, u64); /* address */
+	__type(value, struct alloc_info);
+	__uint(max_entries, ALLOCS_MAX_ENTRIES);
+} allocs SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__type(key, u64); /* stack id */
+	__type(value, union combined_alloc_info);
+	__uint(max_entries, COMBINED_ALLOCS_MAX_ENTRIES);
+} combined_allocs SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__type(key, u64);
+	__type(value, u64);
+	__uint(max_entries, 10240);
+} memptrs SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_STACK_TRACE);
+	__type(key, u32);
+} stack_traces SEC(".maps");
+
+static union combined_alloc_info initial_cinfo;
+
+static void update_statistics_add(u64 stack_id, u64 sz)
+{
+	union combined_alloc_info *existing_cinfo;
+
+	existing_cinfo = bpf_map_lookup_or_try_init(&combined_allocs, &stack_id, &initial_cinfo);
+	if (!existing_cinfo)
+		return;
+
+	const union combined_alloc_info incremental_cinfo = {
+		.total_size = sz,
+		.number_of_allocs = 1
+	};
+
+	__sync_fetch_and_add(&existing_cinfo->bits, incremental_cinfo.bits);
+}
+
+static void update_statistics_del(u64 stack_id, u64 sz)
+{
+	union combined_alloc_info *existing_cinfo;
+
+	existing_cinfo = bpf_map_lookup_elem(&combined_allocs, &stack_id);
+	if (!existing_cinfo) {
+		bpf_printk("failed to lookup combined allocs\n");
+
+		return;
+	}
+
+	const union combined_alloc_info decremental_cinfo = {
+		.total_size = sz,
+		.number_of_allocs = 1
+	};
+
+	__sync_fetch_and_sub(&existing_cinfo->bits, decremental_cinfo.bits);
+}
+
+static int gen_alloc_enter(size_t size)
+{
+	if (size < min_size || size > max_size)
+		return 0;
+
+	if (sample_rate > 1) {
+		if (bpf_ktime_get_ns() % sample_rate != 0)
+			return 0;
+	}
+
+	const pid_t pid = bpf_get_current_pid_tgid() >> 32;
+	bpf_map_update_elem(&sizes, &pid, &size, BPF_ANY);
+
+	if (trace_all)
+		bpf_printk("alloc entered, size = %lu\n", size);
+
+	return 0;
+}
+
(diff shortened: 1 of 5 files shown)


inwardvessel:
saw a build failure on a stray blazesym symbol outside of a `USE_BLAZESYM` pre-processor block. pushed a fix for this. now when blazesym not in use, stack traces will only contain addresses.

anakryiko:
please squash and properly format all those 43 commits, unlikely that you meant to land "wip", "cleanup/debug" commits as is, right?

libbpf-tools/memleak.bpf.c
```diff
@@ -0,0 +1,392 @@
...
+	};
+
+	__sync_fetch_and_sub(&existing_cinfo->bits, decremental_cinfo.bits);
+}
+
+static __always_inline int gen_alloc_enter(size_t size)
+{
+	if (size < min_size || size > max_size) {
+		return 0;
+	}
```
anakryiko, line 94: nit: no need for `{}` for single line if

libbpf-tools/memleak.bpf.c
```diff
@@ -0,0 +1,392 @@
...
+
+	if (sample_rate > 1) {
+		if (bpf_ktime_get_ns() % sample_rate != 0) {
+			return 0;
```
anakryiko, line 98: ditto about {}

libbpf-tools/memleak.bpf.c
```diff
@@ -0,0 +1,392 @@
...
+		}
+	}
+
+	const pid_t pid = bpf_get_current_pid_tgid() >> 32;
+	const u64 size64 = size;
```
anakryiko, line 103: size (size_t) is u64 on BPF side, always, so no need for this

libbpf-tools/memleak.bpf.c
```diff
@@ -0,0 +1,392 @@
...
+	bpf_map_update_elem(&sizes, &pid, &size64, BPF_ANY);
+
+	if (trace_all)
+		bpf_printk("alloc entered, size = %lu\\n", size64);
```
anakryiko, line 107: `\\n` -> `\n` ?

libbpf-tools/memleak.bpf.c
```diff
@@ -0,0 +1,392 @@
...
+SEC("uretprobe")
+int BPF_KRETPROBE(realloc_exit)
+{
+	return gen_alloc_exit(ctx);
+}
+
+SEC("uprobe")
+int BPF_KPROBE(mmap_enter)
+{
+	size_t size = (size_t)PT_REGS_PARM2(ctx);
```
anakryiko, line 212: why not rely on BPF_KPROBE macro to fetch second arg?

libbpf-tools/memleak.bpf.c
```diff
@@ -0,0 +1,392 @@
...
+}
+
+SEC("uretprobe")
+int BPF_KRETPROBE(mmap_exit)
+{
+	return gen_alloc_exit(ctx);
+}
+
+SEC("uprobe")
+int BPF_KPROBE(munmap_enter, void *address) {
```
anakryiko, line 224: { should be on separate line, please check other functions to see if they need to be fixed as well

libbpf-tools/memleak.bpf.c
```diff
@@ -0,0 +1,392 @@
...
+	return gen_free_enter(address);
+}
+
+SEC("uprobe")
+int BPF_KPROBE(posix_memalign_enter, void **memptr, size_t alignment, size_t size)
+{
+	const u64 memptr64 = (u64)(size_t)memptr;
```
anakryiko, line 231: this probably won't work on 32-bit architectures properly. size_t is BPF architecture size, so always 64-bit. To support 32-bit arches properly we might need to use a parameter passed through global var that would specify pointer size

libbpf-tools/memleak.bpf.c
```diff
@@ -0,0 +1,392 @@
...
+
+SEC("uretprobe")
+int BPF_KRETPROBE(posix_memalign_exit)
+{
+	const u64 pid = bpf_get_current_pid_tgid() >> 32;
+	u64 *memptr64;
+	void *addr;
+
+	memptr64 = bpf_map_lookup_elem(&memptrs, &pid);
+
```
anakryiko, line 246: nit: unnecessary empty line?

libbpf-tools/memleak.bpf.c
```diff
@@ -0,0 +1,392 @@
...
+}
+
+SEC("uretprobe")
+int BPF_KRETPROBE(pvalloc_exit)
+{
+	return gen_alloc_exit(ctx);
+}
+
+SEC("tracepoint/kmem/kmalloc")
+int tracepoint__kmalloc(struct trace_event_raw_kmem_alloc *ctx)
```
anakryiko, line 309: this `tracepoint__` prefix is BCC necessity, but for libbpf is just a waste, let's drop this prefix altogether. It would be better to have memleak-specific short prefix to distinguish it from other programs and maps in the system. E.g. "mlk_" or even "memleak_" would be helpful

libbpf-tools/memleak.c
```diff
@@ -0,0 +1,972 @@
...
+		allocs = calloc(ALLOCS_MAX_ENTRIES, sizeof(*allocs));
+
+	if (!allocs) {
+		fprintf(stderr, "failed to allocate array\n");
+		ret = -ENOMEM;
+
+		goto cleanup;
+	}
+
+	libbpf_set_strict_mode(LIBBPF_STRICT_ALL);
```
anakryiko, line 317: no need, libbpf 1.0 is always in "strict mode"

libbpf-tools/memleak.c
```diff
@@ -0,0 +1,972 @@
...
+
+		break;
+	default:
+		return ARGP_ERR_UNKNOWN;
+	}
+
+	return 0;
+}
+
+int libbpf_print_fn(enum libbpf_print_level level, const char *format, va_list args)
```
anakryiko, line 537: static, same for a bunch of other funcs

libbpf-tools/memleak.c
```diff
@@ -0,0 +1,972 @@
...
+
+bool has_kernel_node_tracepoints()
+{
+	return tracepoint_exists("kmem", "kmalloc_node") &&
+		tracepoint_exists("kmem", "kmem_cache_alloc_node");
+}
+
+int disable_kernel_node_tracepoints(struct memleak_bpf *skel)
+{
+	if (bpf_program__set_autoload(skel->progs.tracepoint__kmalloc_node, false) ||
```
anakryiko, line 891: set_autoload can't fail if called before skeleton load, so unnecessary error checking here and in other places

anakryiko (commented):
@inwardvessel I left mostly some styling nits, and I only briefly looked through user-space parts. Overall looks great, good job converting a pretty nontrivial tool!

@chenhengqi it would be great if you could do a more thorough review, thank you!

libbpf-tools/memleak.c
inwardvessel: earlier in the file, the prototypes are static

anakryiko (approved):
it overall looks good to me. One more styling nit: please don't use C++-style comments, and instead utilize `/* */` C comments.

@yonghong-song, @chenhengqi, @davemarchevsky when you guys get a chance, I'd appreciate another set of eyes to point out any other issues. Thanks!

yonghong-song:
LGTM. Thanks!

PR #4520, opened 2023-03-16
Porting from distutils to setuptools

jeromemarchand:
Distutils has been deprecated in python 3.10 and will be removed from python 3.12. Distutils itself has for a long time recommended to use setuptools instead.

This patch migrate from distutils to setuptools using the migration advice from PEP 632:
https://peps.python.org/pep-0632/


jeromemarchand:
Ok. Looks like packaging.version is not a dropping replacement for distutils.version. Need to use packaging.version.Version instead

jeromemarchand:
And python3-packaging added to snapcraft.yml.

PR #4532, opened 2023-03-25
tools/cpudist: Bugfix incorrect cpu key when pid not equal to 0

xingfeng2510:
When pid != 0, current cpu should not be used as a part of key, it may run on different CPUs, which causes start.lookup(key) returns NULL. Just pid can be unique key, here we use an invalid value (0xFFFFFFFF) to fill cpu field of struct entry_key_t.

When pid == 0, still use current cpu as key to distinguish different swappers.

### DIFF
--- a/tools/cpudist.py
+++ b/tools/cpudist.py
@@ -96,7 +96,7 @@ static inline void store_start(u32 tgid, u32 pid, u32 cpu, u64 ts)
     if (IDLE_FILTER)
         return;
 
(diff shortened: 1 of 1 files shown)


chenhengqi (commented):
Good catch.

```
cpudist   Summarize on- and off-CPU time per task as a histogram.
```
Since the tool is captured the stats on a per-task basis, I think it's reasonable to make this change.

yonghong-song:
LGTM. Thanks!

PR #4536, opened 2023-03-27
libbpf-tools/memleak: don't kill process if attached

EspenG:
We should only kill traced process if it is a child process.


yonghong-song:
@inwardvessel could you help take a look at this patch?

inwardvessel (approved):
good one. thanks

PR #4543, opened 2023-03-31
tools/wakeuptime: Reverse the order of stack output

xingfeng2510:
libbpf-tools/wakeuptime based on bcc version (tools/wakeuptime), but it's stack output in reversed order. This patch tries to keep consistent with bcc version.

Also see: https://www.brendangregg.com/FlameGraphs/offcpuflamegraphs.html#Wakeup

Please take a look, thanks. @yonghong-song @chenhengqi

### DIFF
--- a/tools/wakeuptime.py
+++ b/tools/wakeuptime.py
@@ -9,6 +9,7 @@
 # Licensed under the Apache License, Version 2.0 (the "License")
 #
 # 14-Jan-2016	Brendan Gregg	Created this.
+# 03-Apr-2023	Rocky Xing   	Modified the order of stack output.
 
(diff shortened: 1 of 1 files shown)


yonghong-song:
@xingfeng2510 I think we should change bcc tools wakeuptime.py ordering. Typically for kernel stack, we have syscall entry at the bottom of the stack. The wakeuptime_example.txt seems also having this order. Could you help with this?

xingfeng2510:
@yonghong-song Got it, modified by your advice, thanks.

chenhengqi (approved):
LGTM.

Issue #455, opened 2016-03-29
[RFC] Alternate binding implementations


vmg:
From the BCC readme:

> In the future, more bindings besides python will likely be supported. Feel free to add support for the language of your choice and send a pull request!

This is very cool. Let's say that hypothetically GitHub (my employer) had a mostly feature complete `libbcc` toolkit written in LuaJIT, with very interesting performance characteristics because it's able to JIT-compile the C types for the FFI interface to the probes. We may also (hypothetically) have ported many of the existing examples and tools to the new toolkit as examples.

Would you folks be interested on keeping this toolkit "in-tree" (i.e. just like the Python wrapper is kept in this repository), or shall we rather open-source it eventually on its own repository?

drzaeus77:
It could make sense to have in-tree, but really it's about whichever makes the code look cleaner. For instance, I've been working on a Go-based tool that uses bcc (iovisor/iomodules), and that links against the C api that libbcc exposes, but obviously isn't in this same repo. The decision to split that out came not because of a desire to split the repos, but because the use case is different...it's not just language bindings but a completely orthogonal tool, which will have a life of its own.

So, whichever you think makes sense: language bindings in this repo or a separate tool, perhaps in a new repo under iovisor/. Or maybe a bit of both?

vmg:
Thanks for the context!

> The decision to split that out came not because of a desire to split the repos, but because the use case is different...it's not just language bindings but a completely orthogonal tool, which will have a life of its own.

In our case the tool is so far (faster) bindings to libbcc, so I'm going to open a PR and we can see how it looks. I agree obviously agree that complex tools using those bindings should be on a separate repository. :)

4ast:
we had an agreement to stabilize c++/c api of libbcc.so eventually, but today we're far from it. So at the moment any python/go/lua bindings on top of libbcc should really be part of bcc repo, otherwise they will keep breaking. If lua-based tools are going to be similar to python based then makes sense for them to live in tools directory as well, so both can be improved/bugfixed together. Even more complex stuff if it's building on top of libbcc.so can incubate here too. Just to be exposed to more eyes of the community. Small self-managed repos typically has less oversight, less codereviews, less critical feedback.

vmg:
This shipped and is being worked on. Closing this RFC!

PR #4550, opened 2023-04-04
libbpf-tools: Allow biostacks to run on old kernels

michael-chuh:
Fallback to use kprobes if kernel lacks support for fentry/fexit.


libbpf-tools/biostacks.bpf.c
```diff
@@ -110,4 +97,47 @@ int BPF_PROG(blk_account_io_done, struct request *rq)
 	return 0;
 }
 
+
+SEC("fentry/blk_account_io_start")
+int BPF_PROG(blk_account_io_start, struct request *rq)
+{
+	return trace_start(ctx, rq, false);
```
chenhengqi, line 104: Missing the `true` case. `fentry` progs should have as many as the `kprobe` ones.

libbpf-tools/biostacks.c
```diff
@@ -172,52 +172,56 @@ int main(int argc, char **argv)
 
 	obj->rodata->targ_ms = env.milliseconds;
 
-	if (fentry_can_attach("blk_account_io_start", NULL)) {
-		bpf_program__set_attach_target(obj->progs.blk_account_io_start, 0,
-					       "blk_account_io_start");
-		bpf_program__set_attach_target(obj->progs.blk_account_io_done, 0,
-					       "blk_account_io_done");
+	if (kprobe_exists("blk_account_io_start")) {
```
chenhengqi, line 175: The userspace part looks way to complicated, could you please simplify that ?

Let's first check if fentry is supported:
* if yes, disable all kprobe progs, and do auto-attach
* if no, disable all fentry progs, and do auto-attach
You may need to disable some progs if symbols not exist

chenhengqi: Same for #4549 and #4551.

libbpf-tools/biostacks.bpf.c
michael-chuh: blk_account_io_{start, done} were renamed to __blk_account_io_{start, done} from v5.16, see commit a184f095ecb4(libbpf-tools: Fix bio tools).
bpf_program__set_attach_target() can be used for that,  `fentry` progs are not always same as `kprobe`

chenhengqi: I mean `blk_account_io_merge_bio` is missing.

libbpf-tools/biostacks.c
michael-chuh: donw

michael-chuh: done

michael-chuh: Hi @chenhengqi
I did some simplification follow you suggestion, do you have some more advice?
And there is a failure in https://github.com/iovisor/bcc/pull/455, but I cann't see the cause of that request?

PR #4555, opened 2023-04-07
libbpf-tools/memleak: remove unused arg "PERCPU"

michael-chuh:
'memleak -P' is used to trace percpu allocations, no argument is needed


inwardvessel (changes requested):
nice. thank you. can you also update the usage text in `argp_args_doc`?

michael-chuh:
> nice. thank you. can you also update the usage text in `argp_args_doc`?

done

PR #4556, opened 2023-04-07
tools/funcslower: fix printing of folded stacks

jeromemarchand:
When trying to print folded stack, funcslower tries to join bytes to a string. Let's perform that operation with bytes only, and decode before printing.
Also, decode symbols name before printing for the default stack format, to avoid unsightly b'xxx' output.

It fixes the following error:

Exception ignored on calling ctypes callback function: <function PerfEventArray._open_perf_buffer.<locals>.raw_cb_ at 0x7f200541e5e0>
  File "/usr/lib/python3.9/site-packages/bcc/table.py", line 982, in raw_cb_
    callback(cpu, data, size)
  File "/usr/share/bcc/tools/funcslower", line 340, in print_event
    print_stack(event)
  File "/usr/share/bcc/tools/funcslower", line 324, in print_stack
    print("%s %d" % (";".join(line), 1))

### DIFF
--- a/tools/funcslower.py
+++ b/tools/funcslower.py
@@ -317,17 +317,17 @@ def print_stack(event):
         # print folded stack output
         user_stack = list(user_stack)
         kernel_stack = list(kernel_stack)
-        line = [event.comm.decode('utf-8', 'replace')] + \
+        line = [event.comm] + \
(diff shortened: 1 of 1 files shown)


chenhengqi (approved):
LGTM, thanks for the fix.

PR #4559, opened 2023-04-10
tools/profile: Allow to increment hash storage size to aviod missing keys

xingfeng2510:
`BPF_HASH(counts, struct key_t);` uses default hash storage size 10240, usually not enough to cover heavy workloads in a production environment (many cores and many running threads) for 5 seconds only, hash table becomes full and misses some keys. 

This patch try to improve the situation:
1. Add an warn if hash table is full and remind users to increase hash storage size
2. Add option `--hash-storage-size` to increase hash storage size
3. Use default value 40960 instread of 10240 (10240 often not enough, 40960 covers more cases) 

@yonghong-song @chenhengqi Please take a look, thanks.


yonghong-song:
LGTM. Thanks!

Issue #3897, opened 2022-03-04
Can't build Ubuntu 20.04 bcc docker container


dmacnet:
Building the bcc docker container fails for Ubuntu 20.04 LTS using the bcc repo HEAD of master or the v0.24.0 tagged commit. v0.23.0 and earlier work, so this error was introduced after that release.

```docker build -f docker/Dockerfile.ubuntu -t bcc:latest --build-arg OS_TAG=20.04 .```

It fails with a llvm error. `debian/control` forces it to use llvm 9, even though Ubuntu 20.04 LTS has llvm 10 in the apt repo, which does build bcc. The error trace ends like this:

```/tmp/debuild.BIfCG8/bcc/src/cc/bpf_module.cc: In member function 'virtual void ebpf::MyMemoryManager::notifyObjectLoaded(llvm::ExecutionEngine*, const llvm::object::ObjectFile&)':
/tmp/debuild.BIfCG8/bcc/src/cc/bpf_module.cc:108:46: error: no matching function for call to 'llvm::object::SectionRef::getName() const'
  108 |       auto sec_name = section.get()->getName();
      |                                              ^
In file included from /usr/lib/llvm-9/include/llvm/DebugInfo/DIContext.h:18,
                 from /usr/lib/llvm-9/include/llvm/ExecutionEngine/RuntimeDyld.h:18,
                 from /usr/lib/llvm-9/include/llvm/ExecutionEngine/RTDyldMemoryManager.h:18,
                 from /usr/lib/llvm-9/include/llvm/ExecutionEngine/SectionMemoryManager.h:19,
                 from /tmp/debuild.BIfCG8/bcc/src/cc/bpf_module.cc:22:
/usr/lib/llvm-9/include/llvm/Object/ObjectFile.h:437:24: note: candidate: 'std::error_code llvm::object::SectionRef::getName(llvm::StringRef&) const'
  437 | inline std::error_code SectionRef::getName(StringRef &Result) const {
      |                        ^~~~~~~~~~
/usr/lib/llvm-9/include/llvm/Object/ObjectFile.h:437:24: note:   candidate expects 1 argument, 0 provided
/tmp/debuild.BIfCG8/bcc/src/cc/bpf_module.cc: In member function 'int ebpf::BPFModule::finalize()':
[...]

PR #4562, opened 2023-04-11
Fix build error when using docker.

haosdent:
Refer the changes in https://github.com/iovisor/bcc/issues/3897 , also include the missing packages in Dockerfile.


yonghong-song:

PR #4570, opened 2023-04-14
kvmexitaarch64: introduce a tool to show kvm exit reason and count, d…

LangeTang:
…uration of kvm exiting

In the performance optimization of virtualization, the frequency of the KVM EXIT event often needs to be tracked. In addition, the time interval from the KVM EXIT event to the guest mode(kvm entry) is more important.These can help us find the bottleneck of performance. So KVMexitaarch64 is here.

The tool uses a BPF_HASH:to save the timestamp of KVM EXIT and ENTRY, Reason, the accumulated Exit duration, maximum value, minimum value, thread TID and names.


chenhengqi:
This should be integrated into the existing kvmexit tool.

PR #4571, opened 2023-04-14
workaround for evaluating errors from bpf helpers

inwardvessel:
We found out why in some cases the conditional branch was being entered:
```
if (err && err != -EEXIST)
	return 0;
```
...while the error was known to have been `-EEXIST`. 

If this criteria is met:
- the architecture is 64-bit
- the helper function return type is long
- the helper function returns the value of a call to a bpf_map_ops func
- the bpf_map_ops function return type is int
- the compiler inlines the helper function
- the compiler does not sign extend the result of the bpf_map_ops func

In the case of `-EEXIST` (-17), the resulting comparison will be done:
`cmp -17, 4294967279`
... since the sign bit was stuck at the 32nd bit of a 64-bit register. as a result the conditional branch will be entered.

we patched this upstream in bpf-next so that affected bpf_map_ops funcs will return `long`, but as a workaround for the meantime changing the storage type for the error is good enough to have the comparison above work as intended.

### DIFF
--- a/libbpf-tools/maps.bpf.h
+++ b/libbpf-tools/maps.bpf.h
@@ -10,7 +10,23 @@ static __always_inline void *
 bpf_map_lookup_or_try_init(void *map, const void *key, const void *init)
 {
 	void *val;
-	long err;
+	/* bpf helper functions like bpf_map_update_elem() below normally return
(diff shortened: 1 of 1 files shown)


yonghong-song:
LGTM. Thanks for the fix!

PR #4573, opened 2023-04-17
Add option to count tcp accept number by PID with default 5s interval…

pingzhaozz:
Add option to count tcp accept number by PID with default 5s interval or a specified interval.

For example:
$ ./tcpaccept -c -i 2
Tracing accept count... Hit Ctrl-C to end
PID     COMM         Counts
1331419 nginx         26
1331411 nginx         24
1331401 nginx         24
1331477 nginx         22
1331516 nginx         16
1331478 nginx         12
1331415 nginx         8
1331427 nginx         4
1331481 nginx         2

### DIFF
--- a/tools/tcpaccept.py
+++ b/tools/tcpaccept.py
@@ -22,7 +22,7 @@ from socket import inet_ntop, AF_INET, AF_INET6
 from struct import pack
 import argparse
 from bcc.utils import printb
-from time import strftime
+from time import sleep, strftime
 
 # arguments
 examples = """examples:
@@ -34,6 +34,7 @@ examples = """examples:
     ./tcpaccept --mntnsmap mappath   # only trace mount namespaces in the map
     ./tcpaccept -4        # trace IPv4 family
     ./tcpaccept -6        # trace IPv6 family
+    ./tcpaccept -c -i 2   # count tcp accept number by PID every 2s interval
 """
 parser = argparse.ArgumentParser(
     description="Trace TCP accepts",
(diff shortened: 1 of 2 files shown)


pingzhaozz:
Hi @chenhengqi @davemarchevsky 
Could you help review and provide comments about this change? Thanks a lot.

yonghong-song:
@pingzhaozz tcpaccept.py mostly helps when it printed the details of the sock acceptance. Could you explain how the new flag you added will help production debug?

pingzhaozz:
> @pingzhaozz tcpaccept.py mostly helps when it printed the details of the sock acceptance. Could you explain how the new flag you added will help production debug?

Hi Yonghong, the new flag can count the tcp accept number with PID by which could show the load balance of the requests distribution, especially when cpu pinning used. Also it can help find the un-balance case and help further analysis to improve the system capability. For Nginx as an example, it can show the connection number of each work process since the accept work process will also handle the following requests of this connection.

yonghong-song (commented):
could you also update the man page?

yonghong-song:
Also could you add a test to test_tools_smoke.py, there already exists one for tcp_accept.py, you can add another one with your newly added options.

pingzhaozz:
Hi Yonghong, Thanks for your comments that's valuable! I've updated the code according to your comments. Pls help review the latest commit.

tools/tcpaccept.py
```diff
@@ -258,24 +295,50 @@ def print_ipv6_event(cpu, data, size):
...
 
 # header
-if args.time:
-    print("%-9s" % ("TIME"), end="")
-if args.timestamp:
-    print("%-9s" % ("TIME(s)"), end="")
-print("%-7s %-12s %-2s %-16s %-5s %-16s %-5s" % ("PID", "COMM", "IP", "RADDR",
-    "RPORT", "LADDR", "LPORT"))
+if args.count:
+    print("Tracing accept count... Hit Ctrl-C to end")
```
chenhengqi, line 311: This printing is conditional, but on BPF side, the data is captured unconditionally.
That is, we pay extra overhead if we don't specify `-c`.

pingzhaozz: It's a good point. This may introduce some overhead in eBPF code. Since the overhead is only at tcp accept with eBFP monitor, I think it should be acceptable?

PR #4576, opened 2023-04-18
kvmexit.py: add duration of KVM EXIT state for outputing

LangeTang:
In performance analysis, it is important to track the duration of threads in the EXIT state, its important. so, add new item @EXIT_TIME_AVG for Outputing, its mean the average of all interval from kvm_exit to the next kvm_entry. unit:ns

### DIFF
--- a/tools/kvmexit.py
+++ b/tools/kvmexit.py
@@ -5,10 +5,11 @@
 # Display the exit_reason and its statistics of each vm exit
 # for all vcpus of all virtual machines. For example:
 # $./kvmexit.py
-#  PID      TID      KVM_EXIT_REASON                     COUNT
-#  1273551  1273568  EXIT_REASON_MSR_WRITE               6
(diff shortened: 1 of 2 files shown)


LangeTang:
Hi @chenhengqi

Could you help to review it ? Thanks

yonghong-song:
@ShirleyFei based on your experience, do you think tracking the duration of threads in EXIT state is something useful to support with kvmexit tool?

LangeTang:
virtual machine vcpu is not run when in KVM EXIT state. It is necessary to know its duration in performance analysis.This is my opinion.

LangeTang:
virtual machine vcpu is not run when in KVM EXIT state. It is necessary to know its duration in performance analysis.This is my opinion. @yonghong-song @ShirleyFei

LangeTang:
@yonghong-song Will this requests must be reviewed by the original author?

PR #4578, opened 2023-04-20
libbpf-tools/memleak: support ksyms and syms_cache

ShenChen1:
This commit allows the use of 'ksyms' for kernel symbols and 'syms_cache' for userspace application symbols when 'blazesym' is not supported.

### DIFF
--- a/libbpf-tools/memleak.c
+++ b/libbpf-tools/memleak.c
@@ -133,9 +133,12 @@ static int event_notify(int fd, uint64_t event);
 static pid_t fork_sync_exec(const char *command, int fd);
 
 #ifdef USE_BLAZESYM
-static void print_stack_frame(size_t frame, uint64_t addr, const blazesym_csym *sym);
+static void print_stack_frame_by_blazesym(size_t frame, uint64_t addr, const blazesym_csym *sym);
+static void print_stack_frames_by_blazesym();
+#else
+static void print_stack_frames_by_ksyms();
+static void print_stack_frames_by_syms_cache();
 #endif
-
 static int print_stack_frames(struct allocation *allocs, size_t nr_allocs, int stack_traces_fd);
 
 static int alloc_size_compare(const void *a, const void *b);
@@ -210,7 +213,11 @@ static int child_exec_event_fd = -1;
 #ifdef USE_BLAZESYM
 static blazesym *symbolizer;
 static sym_src_cfg src_cfg;
+#else
+struct syms_cache *syms_cache;
+struct ksyms *ksyms;
 #endif
+static void (*print_stack_frames_func)();
 
 static uint64_t *stack;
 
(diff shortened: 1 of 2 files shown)


ShenChen1:
Hi @inwardvessel @chenhengqi 

Could you help to review it ?
Thanks in advance :)

libbpf-tools/memleak.c
```diff
@@ -133,9 +133,11 @@ static int event_notify(int fd, uint64_t event);
 static pid_t fork_sync_exec(const char *command, int fd);
 
 #ifdef USE_BLAZESYM
-static void print_stack_frame(size_t frame, uint64_t addr, const blazesym_csym *sym);
```
inwardvessel, line 136: can this remain as statically declared?

libbpf-tools/memleak.c
```diff
@@ -395,6 +401,30 @@ int main(int argc, char *argv[])
...
+
+			goto cleanup;
+		}
+		print_stack_frames_func = print_stack_frames_by_ksyms;
+	} else {
+		syms_cache = syms_cache__new(0);
+		if (!syms_cache) {
+			fprintf(stderr, "Failed to create syms_cache\n");
+
+			goto cleanup;
```
inwardvessel, line 424: can you set some non-zero error value before jumping to cleanup (here and above)?

libbpf-tools/memleak.c
```diff
@@ -668,53 +787,7 @@ int print_stack_frames(struct allocation *allocs, size_t nr_allocs, int stack_tr
...
-		for (size_t j = 0; j < env.perf_max_stack_depth; ++j) {
-			const uint64_t addr = stack[j];
-
-			if (addr == 0)
-				break;
-
-			printf("\t%zu [<%016lx>]\n", j, addr);
-		}
-#endif
+		(*print_stack_frames_func)();
```
inwardvessel, line 790: I wonder if we can validate the existence of this function somewhere early on. there's a comment "post-processing and validation of env settings" where checks are done on env values

inwardvessel (approved):
appreciate this addition. left some minor comments.

libbpf-tools/memleak.c
ShenChen1: Done

libbpf-tools/memleak.c
ShenChen1: Done

libbpf-tools/memleak.c
ShenChen1: Hi @inwardvessel 

Thanks for the comments.

I assume that the place you mentioned is too early ? `print_stack_frames_func` is determined by whether the initialization of various `symbolizers` is successful at runtime.
And based on the current implementation, `print_stack_frames_func` should always be valid, otherwise, the program would have jumped to `cleanup` already.

libbpf-tools/memleak.c
inwardvessel: hmm, not a big deal but I wouldn't have expected a permission error. is the implication that privileges are needed for these initializations?

libbpf-tools/memleak.c
inwardvessel: fair enough

libbpf-tools/memleak.c
ShenChen1: Nice view, my first thought was to use a generic errno since there could be various kind of error from low-level parts.

But now I assume `ENOMEM` would be the better one, because `blazesym_new`/`ksyms__load`/`syms_cache__new` could be treated as a kind of memory allocation failure.

what do you think about `ENOMEM` ? Or do you have another proposal ?

inwardvessel: I think that will be ok for now since it's somewhat related. let's get this feature merged in and we can change the error later on if we think of something better in the future.

inwardvessel (approved):
@yonghong-song this is good to merge

PR #4589, opened 2023-04-24
tcprtt: Consolidate kprobe and fentry implementations

mauriciovasquezbernal:
Both implementations are identical, then they can be consolidated into a single function that is reused by both programs.

### DIFF
--- a/libbpf-tools/tcprtt.bpf.c
+++ b/libbpf-tools/tcprtt.bpf.c
@@ -29,8 +29,7 @@ struct {
 
 static struct hist zero;
 
-SEC("fentry/tcp_rcv_established")
-int BPF_PROG(tcp_rcv, struct sock *sk)
+static int handle_tcp_rcv_established(struct sock *sk)
 {
 	const struct inet_sock *inet = (struct inet_sock *)(sk);
 	struct tcp_sock *ts;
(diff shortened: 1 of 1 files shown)


libbpf-tools/tcprtt.bpf.c
```diff
@@ -29,8 +29,7 @@ struct {
 
 static struct hist zero;
 
-SEC("fentry/tcp_rcv_established")
-int BPF_PROG(tcp_rcv, struct sock *sk)
+static __always_inline int handle_tcp_rcv_established(struct sock *sk)
```
chenhengqi, line 32: `__always_inline` is not necessary.

chenhengqi (changes requested):
Please use BPF_CORE_READ for all `inet->xxx` and `sk->xxx`. They cause failure on old kernels.

libbpf-tools/tcprtt.bpf.c
mauriciovasquezbernal: We hit some issues with this in Inspektor Gadget: https://github.com/inspektor-gadget/inspektor-gadget/pull/1463. I can remove that here if you think it's not an issue for this case.

mauriciovasquezbernal:
> Please use BPF_CORE_READ for all `inet->xxx` and `sk->xxx`. They cause failure on old kernels.

Oh right. Updated!

libbpf-tools/tcprtt.bpf.c
chenhengqi: According to this comment, we should not hit this issue in tracing progs.

https://github.com/torvalds/linux/blob/219d54332a09e8d8741c1e1982f5eae56099de85/kernel/bpf/verifier.c#L6004-L6007

mauriciovasquezbernal: Thanks for the link. I removed it.

chenhengqi (approved):
LGTM.

PR #4608, opened 2023-05-10
libbpf-tools: use fentry in funclatency

brho:
For tracing kernel functions, funclatency will attempt to use fentry and fallback to kprobes if that fails.  I left the option to force the use of kprobes if you'd like.

Under load, fentry is more efficient:

With kprobes:

cpu:/tmp/b# ./funclatency -u -k bpf_spin_lock
Tracing bpf_spin_lock.  Hit Ctrl-C to exit
^C
     usec                : count    distribution
         0 -> 1          : 565159   |****************************************|
         2 -> 3          : 216204   |***************                         |
         4 -> 7          : 306316   |*********************                   |
         8 -> 15         : 416990   |*****************************           |
        16 -> 31         : 415110   |*****************************           |
        32 -> 63         : 170298   |************                            |
        64 -> 127        : 29726    |**                                      |
       128 -> 255        : 427962   |******************************          |
       256 -> 511        : 233207   |****************                        |
       512 -> 1023       : 453      |                                        |
Exiting trace of bpf_spin_lock

With fentry:
cpu:/tmp/b# ./funclatency -u bpf_spin_lock
Tracing bpf_spin_lock.  Hit Ctrl-C to exit
^C
     usec                : count    distribution
         0 -> 1          : 6371351  |****************************************+|
         2 -> 3          : 1100093  |******                                  |
         4 -> 7          : 518197   |***                                     |
         8 -> 15         : 174881   |*                                       |
        16 -> 31         : 25022    |                                        |
        32 -> 63         : 1553     |                                        |
        64 -> 127        : 1312     |                                        |
       128 -> 255        : 2376     |                                        |
Exiting trace of bpf_spin_lock

### DIFF
--- a/libbpf-tools/funclatency.bpf.c
+++ b/libbpf-tools/funclatency.bpf.c
@@ -28,8 +28,7 @@ struct {
 
 __u32 hist[MAX_SLOTS] = {};
 
-SEC("kprobe/dummy_kprobe")
-int BPF_KPROBE(dummy_kprobe)
+static void entry(void)
 {
 	u64 id = bpf_get_current_pid_tgid();
 	u32 tgid = id >> 32;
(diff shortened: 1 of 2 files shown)


namhyung:
I wonder what makes the difference.  Isn't kprobe optimized well?  Is it specific to the BPF spin lock implementation?

brho:
> I wonder what makes the difference. Isn't kprobe optimized well? Is it specific to the BPF spin lock implementation?

it's not specific to bpf_spin_lock.  i tried a few normal functions too.  

my guess is something related to cache lines bouncing around.  the trace from my commit message came from a 64 core / 128 cpu AMD socket, and a lot of the cpus were hitting that function.  if i run it on something less heavily contended, i don't notice the difference between kprobe and fentry.

glancing through the kprobes code, i see stuff like "optimize kprobes", which on x86 replaces the int3 trap with a jump.  not sure if that was happening either.  and maybe the scalability problem was related to cache misses via the trap?

might also have been the kretprobe and not the kprobe.  not sure on the exact mechanism for the retprobe.

brho:
i hacked up a version where i used kprobes + fexit, and another with fentry + kretprobe.  

the performance / scalability issue is with kretprobes.  kprobe + fexit was fine.

i glanced at kretprobes, and they modify the stack to jump to a trampoline.  not sure about FEXIT.  from poking around, my guess is the fentry/fexit run the func completetly off an intermediate function/trampoline synthesized/jitted, so that all the exit paths are handled without needing to muck with the stack.

yonghong-song:
fentry/fexit is more modern than kprobe/kretprobe. fexit has three advantages over kretprobe: (1) kretprobe will incur two interrupts while fexit incurs none, (2) kretprobe might miss events if too many kretprobe's. fexit does not, and (3), kretprobe only have 'ret' available to bpf program while fexit has all input arguments plus 'ret' value.

So using fentry/fexit is better than kprobe/kretprobe.

This patch looks good to me. Thanks!

PR #4615, opened 2023-05-24
libbpf-tools/opensnoop: Display mode.

eiffel-fl:
Hi.

In this contribution, I added the displyaing of file permission mode to `opensnoop` when extended information are requested:

```
$ sudo ./opensnoop -e
PID    COMM              FD ERR FLAGS    MODE PATH
125035 touch              3   0 02000000 000 /etc/ld.so.cache
125035 touch              3   0 02000000 000 /lib/x86_64-linux-gnu/libc.so.6
125035 touch              3   0 02000000 000 /usr/lib/locale/locale-archive
125035 touch              3   0 00004501 666 /tmp/quux
4274   zsh                3   0 00001501 666 /tmp/corge
125072 zsh                3   0 00001501 666 /dev/null
125106 cat                3   0 02000000 000 /etc/ld.so.cache
125106 cat                3   0 02000000 000 /lib/x86_64-linux-gnu/libc.so.6
125106 cat                3   0 02000000 000 /usr/lib/locale/locale-archive
125106 cat                3   0 00000000 000 /tmp/corge
```

Note that, this information is meaningful only when `open*()` is used to create a file, otherwise its value is 0.

### DIFF
--- a/libbpf-tools/opensnoop.bpf.c
+++ b/libbpf-tools/opensnoop.bpf.c
@@ -47,7 +47,7 @@ bool trace_allowed(u32 tgid, u32 pid)
 }
 
 static __always_inline
-int trace_enter(const char *filename, int flags)
+int trace_enter(const char *filename, int flags, __u16 mode)
 {
 	u64 id = bpf_get_current_pid_tgid();
 	/* use kernel terminology here for tgid/pid: */
@@ -59,6 +59,7 @@ int trace_enter(const char *filename, int flags)
 		struct args_t args = {};
 		args.fname = filename;
 		args.flags = flags;
+		args.mode = mode;
 		bpf_map_update_elem(&start, &pid, &args, 0);
 	}
 	return 0;
@@ -67,13 +68,13 @@ int trace_enter(const char *filename, int flags)
 SEC("tracepoint/syscalls/sys_enter_open")
 int tracepoint__syscalls__sys_enter_open(struct trace_event_raw_sys_enter* ctx)
 {
-	return trace_enter((const char *)ctx->args[0], (int)ctx->args[1]);
+	return trace_enter((const char *)ctx->args[0], (int)ctx->args[1], (__u16)ctx->args[2]);
 }
 
 SEC("tracepoint/syscalls/sys_enter_openat")
 int tracepoint__syscalls__sys_enter_openat(struct trace_event_raw_sys_enter* ctx)
 {
-	return trace_enter((const char *)ctx->args[1], (int)ctx->args[2]);
+	return trace_enter((const char *)ctx->args[1], (int)ctx->args[2], (__u16)ctx->args[3]);
 }
 
 static __always_inline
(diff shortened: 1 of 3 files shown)


libbpf-tools/opensnoop.h
```diff
@@ -18,6 +19,7 @@ struct event {
 	uid_t uid;
 	int ret;
 	int flags;
+	__u16 mode;
```
chenhengqi, line 22: This should be added to the end of the struct. Otherwise it introduces some hole inside the struct.

chenhengqi: I think `rwx` would be better.

libbpf-tools/opensnoop.bpf.c
```diff
@@ -46,8 +46,8 @@ bool trace_allowed(u32 tgid, u32 pid)
 	return true;
 }
 
-SEC("tracepoint/syscalls/sys_enter_open")
-int tracepoint__syscalls__sys_enter_open(struct trace_event_raw_sys_enter* ctx)
+static __always_inline
+int trace_enter(const char *filename, int flags, __u16 mode)
```
chenhengqi, line 50: We can drop this `__always_inline` and keep these in single line.

libbpf-tools/opensnoop.c
```diff
@@ -250,9 +288,20 @@ void handle_event(void *ctx, int cpu, void *data, __u32 data_sz)
 	printf("%-6d %-16s %3d %3d ", e->pid, e->comm, fd, err);
 	sps_cnt += 7 + 17 + 4 + 4;
 	if (env.extended) {
-		printf("%08o ", e->flags);
-		sps_cnt += 9;
+		// 11 due to 3 times "rwx", one for UID/GID or sticky bit and one for '\0'.
+		char *ret = malloc(11 * sizeof(char));
```
chenhengqi, line 292: `malloc` is unnecessary ? Just use stack variable.

libbpf-tools/opensnoop.c
```diff
@@ -199,6 +200,43 @@ static void sig_int(int signo)
...
+		{ S_IXUSR, 'x'},
+		{ S_IRGRP, 'r'},
+		{ S_IWGRP, 'w'},
+		{ S_IXGRP, 'x'},
+		{ S_IROTH, 'r'},
+		{ S_IWOTH, 'w'},
+		{ S_IXOTH, 'x'},
+	};
+
+	// Deal with rwx for user, group and others.
```
chenhengqi, line 227: No C++ style comments, please.

chenhengqi (commented):
I've tested this tool, but some output look strange:
```
2863755 cat                3   0 02000000 ---------- /etc/ld.so.cache
2863755 cat                3   0 02000000 ---------- /lib64/libc.so.6
2863755 cat                3   0 02000000 ---------- /lib64/libdl.so.2
```

Haven't dug into that.

PR #4622, opened 2023-06-01
libbpf-tools: Add support for execsnoop on arm64.

eiffel-fl:
Hi.

On old `arm64` kernels, there is a bug where [syscall number is forgotten](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=de6921856f99) when a new thread is created.
So, this is not possible to use `sys_exit_execve` on these kernels:

```bash
$ git checkout de6921856f99~1
# Just jump before the fix and compile the kernel.
root@vm-arm64:~# lscpu | head -1
Architecture:                    aarch64
root@vm-arm64:~# uname -a
Linux vm-arm64 5.19.0-rc3-00001-g9f6a503d5238 #3 SMP PREEMPT Thu Jun 1 17:26:04 WEST 2023 aarch64 GNU/Linux
root@vm-arm64:~# ./share/linux/linux/tools/perf/perf record -ag -e 'syscalls:sys_exit_execve' -e 'syscalls:sys_enter_execve' &
[1] 381
root@vm-arm64:~# ls > /dev/null
root@vm-arm64:~# fg
./share/linux/linux/tools/perf/perf record -ag -e 'syscalls:sys_exit_execve' -e 'syscalls:sys_enter_execve'
^C[ perf record: Woken up 1 times to write data ]
[ perf record: Captured and wrote 0.060 MB perf.data (1 samples) ]

root@vm-arm64:~# ./share/linux/linux/tools/perf/perf script
bash   382 [003]  1064.274111: syscalls:sys_enter_execve: filename: 0xaaaaec27b>
                   a8188 [unknown] (/usr/lib/aarch64-linux-gnu/libc-2.31.so)
                   4b060 [unknown] (/usr/bin/bash)
                   4e0ec [unknown] (/usr/bin/bash)
                   4e850 [unknown] (/usr/bin/bash)
                   359e4 [unknown] (/usr/bin/bash)
                   32f90 [unknown] (/usr/bin/bash)
                   24218 [unknown] (/usr/lib/aarch64-linux-gnu/libc-2.31.so)
                   34494 [unknown] (/usr/bin/bash)
```

As you can see, the above `perf script` does not contain `sys_exit_execve` due to the bug describe above.

So, the solution is to use a `kretprobe` as exit point:

```bash
root@vm-arm64:~# lscpu | head -1
Architecture:                    aarch64
root@vm-arm64:~# uname -a
Linux vm-arm64 5.19.0-rc3-00001-g9f6a503d5238 #3 SMP PREEMPT Thu Jun 1 17:26:04 WEST 2023 aarch64 GNU/Linux
root@vm-arm64:~# ./share/kinvolk/bcc/libbpf-tools/execsnoop &
[1] 289
root@vm-arm64:~# PCOMM            PID    PPID   RET ARGS
ls -al > /dev/null
ls               290    276      0 /usr/bin/ls -al 
```

### DIFF
--- a/libbpf-tools/execsnoop.bpf.c
+++ b/libbpf-tools/execsnoop.bpf.c
@@ -2,6 +2,9 @@
 #include <vmlinux.h>
 #include <bpf/bpf_helpers.h>
 #include <bpf/bpf_core_read.h>
+#ifdef __TARGET_ARCH_arm64
+#include <bpf/bpf_tracing.h>
+#endif /* __TARGET_ARCH_arm64 */
 #include "execsnoop.h"
 
 const volatile bool filter_cg = false;
@@ -108,8 +111,14 @@ int tracepoint__syscalls__sys_enter_execve(struct trace_event_raw_sys_enter* ctx
 	return 0;
 }
 
+
+#ifdef __TARGET_ARCH_arm64
+SEC("kretprobe/do_execveat_common.isra.0")
+int BPF_KRETPROBE(execveat_exit)
+#else /* !__TARGET_ARCH_arm64 */
 SEC("tracepoint/syscalls/sys_exit_execve")
 int tracepoint__syscalls__sys_exit_execve(struct trace_event_raw_sys_exit* ctx)
+#endif /* !__TARGET_ARCH_arm64 */
 {
 	u64 id;
 	pid_t pid;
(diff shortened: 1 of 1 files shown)


libbpf-tools/execsnoop.bpf.c
```diff
@@ -108,8 +111,14 @@ int tracepoint__syscalls__sys_enter_execve(struct trace_event_raw_sys_enter* ctx
 	return 0;
 }
 
+
+#ifdef __TARGET_ARCH_arm64
+SEC("kretprobe/do_execveat_common.isra.0")
```
chenhengqi, line 116: Can we handle this difference from user space ? Symbols with `isra` suffix is quite unreliable.

yonghong-song: Two things.
First, we should not put 'do_execveat_common.isra.0' into the section names. The function is not reliable. It depends on compiler optimization. It may be gone in the next kernel release. User space needs to handle this
to ensure this function exists before attaching to it.
Second, this is a hack for buggy kernel and these buggy LTS kernels have been fixed. So there is really no need to put such a workaround to the tool. We don't want to support any intermediate revisions of the stable kernel.

PR #4626, opened 2023-06-07
Fix TypeError: a bytes-like object is required, not 'str'

sancppp:


sancppp:
I found some small errors while studying the tutorial_bcc_python_developer.md document. I am a github newbie, so please point out if this PR is not standardized enough. Thanks!

PR #4633, opened 2023-06-12
argdist: print process name COMM

tariromukute:
Hi,

Thank you for this project.

I would want to use `argdist` but instead of getting the PID I need the process name. The reason being I have short lived processes, and I want to inspect the sockets created by the processes if they are non-blocking. To make it easier to identify the process I want the tool to print the process name, supporting something like `argdist -c -C 't:syscalls:sys_enter_accept4():int,char*,int:args->fd,$COMM,args->flags&00004000'`. I managed to make some code changes in this PR but I am facing an issue with the output, see below

```
t:syscalls:sys_enter_accept4():int,char*,int:args->fd,$COMM,args->flags&00004000
	COUNT      EVENT
	2          args->fd = 9, $COMM = b'nrf', args->flags&00004000 = 2048
	4          args->fd = 9, $COMM = b'nrf', args->flags&00004000 = 2048
	4          args->fd = 9, $COMM = b'nrf', args->flags&00004000 = 2048
	4          args->fd = 9, $COMM = b'nrf', args->flags&00004000 = 2048
```

Although the COMM is successfully extracted, it results in what should be the same count being separate. I am assuming this is due to the pointer to the COMM being saved in the BPF_HASH map. I would appreciate pointers to resolving this.

Thanks,

### DIFF
--- a/tools/argdist.py
+++ b/tools/argdist.py
@@ -128,12 +128,9 @@ u64 __time = bpf_ktime_get_ns();
         
         def _generate_comm_prefix(self):
                 text = """
-enum {
-        TASK_COMM_LEN = 16,
-};
 struct val_t {
         u32 pid;
-        char name[TASK_COMM_LEN];
+        char name[sizeof(struct __string_t)];
 };
 struct val_t val = {.pid = (bpf_get_current_pid_tgid() >> 32) };
 bpf_get_current_comm(&val.name, sizeof(val.name));
--- a/tools/argdist.py
+++ b/tools/argdist.py
@@ -583,6 +583,9 @@ argdist -p 1005 -H 'r:c:read()'
 argdist -C 'r::__vfs_read():u32:$PID:$latency > 100000'
         Print frequency of reads by process where the latency was >0.1ms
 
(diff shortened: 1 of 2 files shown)


yonghong-song:
The tool itself is argdist (argument distribution) and 'comm' is not part of the argument, so we probably should not add 'comm' in it. You can have a customized private version for that purpose.
You can use option '--ebpf' to dump the generated code to debug your issue.

tariromukute:
Thanks @yonghong-song for the pointer. I found what was causing the misbehavior leveraging the --ebpf flag. 

When the key was being created `bpf_probe_read_kernel(&__key.v0.s, sizeof(__key.v0.s), (void *)&val.name);` the size of the `v0` variable is equivalent to `sizeof(struct __string_t)`. I had set the size of the process name (`val.name`) to `TASK_COMM_LEN = 16`. This mismatch was causing the difference in the key even when the process name was the same. Changing the size of `val.name` to `sizeof(struct __string_t)` resolved the issue.

I am using this customised version for my work, however I updated the tool's documentation for the PR to merged. I think other users might find the flexibility useful.

Thanks.

PR #4644, opened 2023-06-20
libbpf-tools: Avoid virtual memory area [uprobes] warning

michael-chuh:
Refer to kernel commit 704bde3cc26a ("uprobes: Use vm_special_mapping to name the XOL vma"), processes setup the [uprobes] special virtual memory area with uprobe events to excute original instruction out of line.

open_elf() would throw a warning of 'Could not open [uprobes]' when attaching uprobes, so remove '[uprobes]' from get_elf_type() to avoid it.

This warning can be reproduced with `memleak -c <command>` or `memleak -p <pid>`, for example
```
~ # memleak -p 57
using default object: libc.so.6
using page size: 4096
tracing kernel: false
Tracing outstanding memory allocs...  Hit Ctrl-C to end
[7:21:3] Top 0 stacks with outstanding allocations:
[7:21:8] Top 3 stacks with outstanding allocations:
4096 bytes in 1 allocations from stack
Could not open [uprobes]
        0 [<00ffffff980d6904>] _IO_file_doallocate+0x5a [/lib/libc.so.6]
        1 [<00fffffffffff000>] <null sym>
```

### DIFF
--- a/libbpf-tools/trace_helpers.c
+++ b/libbpf-tools/trace_helpers.c
@@ -243,6 +243,7 @@ static bool is_file_backed(const char *mapname)
 		STARTS_WITH(mapname, "[stack") ||
 		STARTS_WITH(mapname, "/SYSV") ||
 		STARTS_WITH(mapname, "[heap]") ||
+		STARTS_WITH(mapname, "[uprobes]") ||
 		STARTS_WITH(mapname, "[vsyscall]"));
 }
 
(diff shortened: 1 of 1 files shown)


yonghong-song:
Your code is probably correct. But it would be great if you can add a simple reproducer in the commit message and github summary. It serves two purpose. First, bcc users will know when the issue might happen and why. Second, reviewer can run the reproducer to verify the behavior before and after this patch. Thanks!

libbpf-tools/trace_helpers.c
```diff
@@ -256,6 +256,11 @@ static bool is_vdso(const char *path)
 	return !strcmp(path, "[vdso]");
 }
 
+static bool is_uprobes(const char *path)
+{
+	return !strcmp(path, "[uprobes]");
+}
+
```
JackieLiu1, line 263: is_file_backed also need this.

michael-chuh: Thanks for reviewing, is_file_backed is patched.
A simple reproducer is also added to commit message.

PR #4653, opened 2023-06-26
cpufreq: read initial freqs only of valid links

carofe82:
It results in a targeted freqs initialization, and prevents errors such as missing cpu freq sysfs file of not present, or offline cpus.

### DIFF
--- a/libbpf-tools/cpufreq.c
+++ b/libbpf-tools/cpufreq.c
@@ -132,13 +132,16 @@ static void sig_handler(int sig)
 {
 }
 
-static int init_freqs_mhz(__u32 *freqs_mhz, int nr_cpus)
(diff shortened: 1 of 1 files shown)


carofe82:
I'm having a hard time figuring out the relationship between those test failures and my PR. It looks like they were  interrupted, probably after timing out.

yonghong-song:
> I'm having a hard time figuring out the relationship between those test failures and my PR. It looks like they were interrupted, probably after timing out.

The test failure is indeed irrelevant to this patch.

carofe82:
> > I'm having a hard time figuring out the relationship between those test failures and my PR. It looks like they were interrupted, probably after timing out.
> 
> The test failure is indeed irrelevant to this patch.

@yonghong-song , if there is anything I can do to help troubleshoot let me know.

PR #4659, opened 2023-06-29
libbpf-tools: add block_io_{start,done} tracepoints support to bio tools

michael-chuh:
Some bio tools fail to kprobe blk_account_io_{start,done} after v5.17, because they become inlined, see [0]. To fix this issue, tracepoints blick_io_{start,done} are introcuded in kernel, see[1].

Update related bio tools to support new tracepoints, and also simplify attach.

[0] Kernel commit 450b7879e345 (block: move blk_account_io_{start,done}
    to blk-mq.c)
[1] Kernel commit 5a80bd075f3b (block: introduce
    block_io_start/block_io_done tracepoints)

PR #4662, opened 2023-06-29
libbpf-tools: fallback to tracefs mount point if debugfs is not mounted

michael-chuh:
When testing https://github.com/iovisor/bcc/pull/4659, tracepoint_exists() always return false.
My kernel config shows `CONFIG_DEBUG_FS is not set`, so `/sys/kernel/debug/tracing` is not exist.
Follow libbpf to add `/sys/kernel/tracing` support.


JackieLiu1 (commented):
most of them copy from libbpf.c, LGTM

PR #4664, opened 2023-07-01
libbpf-tools: Allow tcppktlat to run on old kernels

chenhengqi:
Currently, tcppktlat can only run on kernel v5.12+ because of the usages of bpf_get_socket_cookie().
This commit gets the tool run on old kernels in a CO-RE way. While at it, also adds tcp_destroy_sock tracepoint. This is necessary since there are cases where the socket does not get the chance to reach the tcp_rcv_space_adjust tracepoint and entries in start map never get removed.

cc @ethercflow

### DIFF
--- a/libbpf-tools/core_fixes.bpf.h
+++ b/libbpf-tools/core_fixes.bpf.h
@@ -215,4 +215,26 @@ static __always_inline bool has_kmem_alloc(void)
 	return false;
 }
 
+/**
+ * The bpf_get_socket_cookie helper is landed since kernel v4.12，
+ * but only available for tracing programs since kernel v5.12
+ * via commit c5dbb89fc2ac("bpf: Expose bpf_get_socket_cookie to tracing programs").
+ * Since the helper is used to provide a unique socket identifier,
+ * we could use the sock itself as the identifier if the helper is not available.
+ * Here, we use BPF_FUNC_check_mtu to check the availability of the helper
+ * since they are both introduced in v5.12.
+ *
+ * see:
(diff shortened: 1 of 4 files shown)


libbpf-tools/core_fixes.bpf.h
```diff
@@ -215,4 +215,26 @@ static __always_inline bool has_kmem_alloc(void)
...
+ * since they are both introduced in v5.12.
+ *
+ * see:
+ *    https://github.com/torvalds/linux/commit/91b8270f2a4d
+ *    https://github.com/torvalds/linux/commit/c5dbb89fc2ac
+ *    https://github.com/torvalds/linux/commit/34b2021cc616
+ */
+static __always_inline __u64 get_sock_ident(struct sock *sk)
+{
+	if (bpf_core_enum_value_exists(enum bpf_func_id, BPF_FUNC_check_mtu)) {
```
ethercflow, line 234: From https://github.com/torvalds/linux/commit/34b2021cc616, I saw 
> bpf: Add BPF-helper for MTU checking
This BPF-helper bpf_check_mtu() works for both XDP and TC-BPF programs.

So it can be used for tracing BPF prog?

ethercflow (approved):
Rest LGTM

libbpf-tools/core_fixes.bpf.h
chenhengqi: Probably not, but we don't use it, just check its existence.

ethercflow: Oh I see what you mean :)

PR #4367, opened 2022-11-29
tools/statsnoop: Add some stat() variants

xingfeng2510:
Add some stat() variants (statx/newlstat).

PR #4668, opened 2023-07-10
tools/statsnoop: bring back the syscall__ prefix

wenjianhn:
fn_name of a syscall should start with "syscall__".

See
https://github.com/iovisor/bcc/blob/v0.28.0/src/cc/frontends/clang/b_frontend_action.cc#L818

Without the patch, file names will not be printed.

Fixes: c743fcb17b46 ("tools/statsnoop: Add some stat() variants (#4367)")


wenjianhn:
@xingfeng2510 please take a review.

PR #4669, opened 2023-07-10
libbpf-tools: add futexctn

ethercflow:
Although systemtap already has a futexes.stp tool, it doesn't feel very easy to use. For example, there are only maximum, minimum and average values, but no delay distribution information. In addition, although the mutex address is behind, there is no call stack information, which is inconvenient to check when you cannot use gdb to attach the process.

This tool identifies lock contention information by filtering futex `FUTEX_WAIT` operations and retrieves the call stack information when entering lock contention. The lock contention counter is partitioned based on the call stack, enabling users to locate all path information involved in this lock contention using the printed lock address, which serves as the basis for optimization.

```
➜ ✗ sudo ./futexctn -p 381676 1 1 -mT

15:07:13


fmu[381694] lock 0x55bb40a44060 contended 61 times, 16 avg msecs [max: 21 msecs, min 16 msecs]
    -
    __lll_lock_wait_private
    bar
    foo
    thread_func
    pthread_condattr_setpshared
    -
     msecs               : count    distribution
        16 -> 31         : 61       |****************************************|


fmu[381689] lock 0x55bb40a44060 contended 61 times, 16 avg msecs [max: 21 msecs, min 16 msecs]
    -
    __lll_lock_wait_private
    bar
    foo
    thread_func
    pthread_condattr_setpshared
    -
     msecs               : count    distribution
        16 -> 31         : 61       |****************************************|
```

### DIFF
--- a/libbpf-tools/futexctn.bpf.c
+++ b/libbpf-tools/futexctn.bpf.c
@@ -1,0 +1,131 @@
+// SPDX-License-Identifier: GPL-2.0
+/* Copyright (c) 2023 Wenbo Zhang */
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_tracing.h>
+
+#include "futexctn.h"
+#include "bits.bpf.h"
+#include "maps.bpf.h"
+
+#define MAX_ENTRIES	10240
+
+#define FUTEX_WAIT		0
+#define FUTEX_PRIVATE_FLAG	128
+#define FUTEX_CLOCK_REALTIME	256
+#define FUTEX_CMD_MASK		~(FUTEX_PRIVATE_FLAG | FUTEX_CLOCK_REALTIME)
+
+const volatile bool targ_summary = false;
+const volatile bool targ_ms = false;
+const volatile __u64 targ_lock = 0;
+const volatile pid_t targ_pid = 0;
+const volatile pid_t targ_tid = 0;
+
+struct val_t {
+	u64 ts;
+	u64 uaddr;
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_ENTRIES);
+	__type(key, u64);
+	__type(value, struct val_t);
+} start SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_STACK_TRACE);
+	__uint(key_size, sizeof(u32));
+} stackmap SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_ENTRIES);
+	__type(key, struct hist_key);
+	__type(value, struct hist);
+} hists SEC(".maps");
+
+static struct hist initial_hist = {};
+
+SEC("tracepoint/syscalls/sys_enter_futex")
+int futex_enter(struct trace_event_raw_sys_enter *ctx)
+{
+	struct val_t v = {};
+	u64 pid_tgid;
+	u32 tid;
+
+	if (((int)ctx->args[1] & FUTEX_CMD_MASK) != FUTEX_WAIT)
+		return 0;
+
+	pid_tgid = bpf_get_current_pid_tgid();
+	tid = (__u32)pid_tgid;
+	if (targ_pid && targ_pid != pid_tgid >> 32)
+		return 0;
+	if (targ_tid && targ_tid != tid)
+		return 0;
+	v.uaddr = ctx->args[0];
+	if (targ_lock && targ_lock != v.uaddr)
+		return 0;
+	v.ts = bpf_ktime_get_ns();
+	bpf_map_update_elem(&start, &pid_tgid, &v, BPF_ANY);
+	return 0;
+}
+
+SEC("tracepoint/syscalls/sys_exit_futex")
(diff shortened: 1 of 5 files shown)


chenhengqi:
Could you explain how does this tool work, especially the `contended` counter ?

libbpf-tools/futexctn.bpf.c
```diff
@@ -0,0 +1,120 @@
...
+	if (((int)ctx->args[1] & FUTEX_CMD_MASK) != FUTEX_WAIT)
+		return 0;
+
+	pid_tgid = bpf_get_current_pid_tgid();
+	tid = (__u32)pid_tgid;
+	if (targ_pid && targ_pid != pid_tgid >> 32)
+		return 0;
+	if (targ_tid && targ_tid != tid)
+		return 0;
+	v.uaddr = ctx->args[0];
```
chenhengqi, line 65: We can't store a pointer in a 32-bit value.

ethercflow: The `mutex` is defined as below (in [glibc 2.35](https://github.com/bminor/glibc/blob/4290aed05135ae4c0272006442d147f2155e70d7/sysdeps/x86/nptl/bits/struct_mutex.h#L22)):

```c
struct __pthread_mutex_s
{
  int __lock;
  unsigned int __count;
  int __owner;
#ifdef __x86_64__
  unsigned int __nusers;
...
}
```
So we can see the value of `&__lock` is also the address of `pthread_mutex_t *mutex`. The glibc will pass it to the syscall `futex`: 
```c
futex_wait ((unsigned int *) &mutex->__data.__lock, oldval, ...
```

So It's used as ((unsigned long)(REC->uaddr)).

ethercflow:
> Could you explain how does this tool work, especially the `contended` counter ?

Done

chenhengqi:
* Since this tool is working on a single process basis, `FUTEX_WAIT_PRIVATE` should be used instead of `FUTEX_WAIT`.
> When executing a futex operation that requests to block a thread,
       the kernel will block only if the futex word has the value that
       the calling thread supplied (as one of the arguments of the
       futex() call) as the expected value of the futex word.
* According to the manpage, I guess more work should be done to check whether there is a contention or not.

libbpf-tools/futexctn.bpf.c
```diff
@@ -0,0 +1,126 @@
...
+	v.ts = bpf_ktime_get_ns();
+	bpf_map_update_elem(&start, &pid_tgid, &v, 0);
+	return 0;
+}
+
+SEC("tracepoint/syscalls/sys_exit_futex")
+int futex_exit(struct trace_event_raw_sys_enter *ctx)
```
yonghong-song, line 72: The argument type does not sounds right. Should it be `trace_event_raw_sys_exit`?

```
struct trace_event_raw_sys_exit {
        struct trace_entry ent;
        long id;
        long ret;       
        char __data[0]; 
};
```

Also, since we have 'ret' value here, should we check whether the syscall is successfully or not, for either WAIT or WAKE? We should only care successful case, right?

ethercflow: > The argument type does not sounds right. Should it be trace_event_raw_sys_exit?

Yes, it should be `trace_event_raw_sys_exit`.

> Also, since we have 'ret' value here, should we check whether the syscall is successfully or not, for either WAIT or WAKE? We should only care successful case, right?

Agree, successful case is more make sence.

@yonghong-song Both are fixed, thanks!

ethercflow:
> Since this tool is working on a single process basis, FUTEX_WAIT_PRIVATE should be used instead of FUTEX_WAIT.

@chenhengqi  Glibc (pthread) only use `FUTEX_WAIT`, so we need to add comments to notice this?

ethercflow:
Hi @chenhengqi @yonghong-song, do you have any other suggestions for modification?

yonghong-song:
Similar to other tools, please guard your usage of blazesym with macro USE_BLAZESYM as it may not be available in all systems.

ethercflow:
> Similar to other tools, please guard your usage of blazesym with macro USE_BLAZESYM as it may not be available in all systems.

@yonghong-song Fixed, thanks!

ethercflow:
Hi @yonghong-song @chenhengqi , sorry to bother you. May I ask if this PR be merged? Is there anything else I need to do?

libbpf-tools/futexctn.bpf.c
```diff
@@ -0,0 +1,131 @@
...
+		slot = MAX_SLOTS - 1;
+	__sync_fetch_and_add(&histp->slots[slot], 1);
+	__sync_fetch_and_add(&histp->contended, 1);
+	__sync_fetch_and_add(&histp->total_elapsed, delta);
+	min = __sync_fetch_and_add(&histp->min, 0);
+	if (!min || min > delta)
+		__sync_val_compare_and_swap(&histp->min, min, delta);
+	max = __sync_fetch_and_add(&histp->max, 0);
+	if (max < delta)
+		__sync_val_compare_and_swap(&histp->max, max, delta);
```
chenhengqi, line 123: This only available on kernel 5.12+, maybe we can use `bpf_spin_lock` instead.

chenhengqi: `bpf_spin_lock` is not available to tracing progs, how can we get this work on old kernels ? @yonghong-song

ethercflow: @yonghong-song Get it, thanks!

PR #4674, opened 2023-07-18
libbpf-tools/execsnoop: fix ret type of bpf_probe_read_user*.

opsnull:
The [bpf_probe_read_user_str* manpage](https://man7.org/linux/man-pages/man7/bpf-helpers.7.html) says: 
> On success, the strictly positive length of the output string, including the trailing NUL character. On error, a negative value.

### DIFF
--- a/libbpf-tools/execsnoop.bpf.c
+++ b/libbpf-tools/execsnoop.bpf.c
@@ -40,7 +40,7 @@ int tracepoint__syscalls__sys_enter_execve(struct trace_event_raw_sys_enter* ctx
 {
 	u64 id;
 	pid_t pid, tgid;
-	unsigned int ret;
+	int ret;
 	struct event *event;
 	struct task_struct *task;
 	const char **args = (const char **)(ctx->args[1]);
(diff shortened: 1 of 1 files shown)


yonghong-song:
The current code seems okay. The code churn with the same functionality seems not necessary.
The following is an explanation.

For helper like `bpf_probe_read_user*`, return a negative value means failure. But typical all negative error
code is with range of [-4096, -1], so a 'unsigned int' means the error code is a pretty big unsigned int number,
much larger than typical argsize, so comparison like 'unsigned_ret > argsize' means a failure.

Also, if `bpf_probe_read_user*` failed, the dest buffer will be zero out, it can be tested as well in indicate that
it might be a true end, or it might be an error case. In any case, the arg processing can be stopped.

opsnull:
Thanks for the explanation! Although the result is correct, the code is not straightforward to understand. Moreover, these two cases do not seem to conform to general programming practices, such as comparing signed and unsigned types at the same time, not directly checking the return value of the function but relying on internal implementation details.

libbpf-tools/execsnoop.bpf.c
```diff
@@ -73,6 +73,9 @@ int tracepoint__syscalls__sys_enter_execve(struct trace_event_raw_sys_enter* ctx
 	event->args_size = 0;
 
 	ret = bpf_probe_read_user_str(event->args, ARGSIZE, (const char*)ctx->args[0]);
+	if (ret < 0) {
+		return 0
```
yonghong-song, line 77: Okay, I changed my mind. Let us try to code based on uapi specification.
Have you compiled the code before submit the patch? There is a missing ';' in the above return statement.

opsnull: Aha, sloppy, fixed, compile and self-test pass.

PR #4682, opened 2023-07-25
cc/python-api: Add PID parameter to `perf_event_open`

0x161e-swei:
Related to the discussion here: https://github.com/iovisor/bcc/discussions/4673#discussion-5405903

The first commit in this PR adds the `pid` parameter to the `perf_event_open` in both cc and python API.   
The second commit in this PR edits the existing cc and python tests to pass in the PID of the test process as an argument to `perf_event_open`.


yonghong-song:
Thanks. LGTM.

PR #4691, opened 2023-08-01
tools: Add support for the new block_io_* tracepoints

### LINKED ISSUE #4261: `biotop` and `biosnoop` do not work under 5.19 kernel due to missing `blk_account_io_start` kprobe
On the latest Arch Linux installation, and Fedora 36 with 5.19 kernel, the kprobes `blk_account_io_start` as well as `__blk_account_io_start` are missing, as seen from the `/proc/kallsyms` file. As a result, both `biotop` and `biosnoop` fail to start with the following error messages:
`Exception: Failed to attach BPF program b'trace_pid_start' to kprobe b'blk_account_io_start', it's not traceable (either non-existing, inlined, or marked as "notrace")
`

jeromemarchand:
The bio tools currently depends on blk_account_io_done/start functions that can be inlined. To fix that, a couple of tracepoints have been added upstream (block:block_io_start/done). This patch add the support for those tracepoints when they are available.

Unfortunately, the bio tools relies on data that is not available to the tracepoints (mostly the struct request). So the tracepoints can't be used as drop in replacement for blk_account_io_*. Main difference, is that we can't use the struct request as the hash key anymore, so it now uses the couple (dev_t, sector) for that purpose.

For the biolatency tool, the -F option is disabled when only the tracepoints are available because the flags are not all accessible from the tracepoints. Otherwise, all features of the tools should remain.

Closes #4261

### DIFF
--- a/tools/biolatency.py
+++ b/tools/biolatency.py
@@ -11,6 +11,7 @@
 #
 # 20-Sep-2015   Brendan Gregg   Created this.
 # 31-Mar-2022   Rocky Xing      Added disk filter support.
+# 01-Aug-2023   Jerome Marchand Added support for block tracepoints
 
 from __future__ import print_function
 from bcc import BPF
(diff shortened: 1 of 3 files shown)


yonghong-song:
LGTM. Thanks!

PR #4696, opened 2023-08-06
fix. backport compilation with llvm9 and earlier

pagict:
llvm::SectionRef changed the `getName` interface since LLVM10, in particular, in commit llvm/llvm-project@a11d302fa0.

In llvm-9 and early version, `getName` required a StringRef reference as a parameter, and returned a std::error_code.

### DIFF
--- a/src/cc/bpf_module.cc
+++ b/src/cc/bpf_module.cc
@@ -122,10 +122,17 @@ class MyMemoryManager : public SectionMemoryManager {
       if (!section)
         continue;
 
+#if LLVM_MAJOR_VERSION >= 10
       auto sec_name = section.get()->getName();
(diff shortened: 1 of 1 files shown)


yonghong-song:
llvm9 is pretty old. Any particular reason why you are still using it? For bcc, we really recommend to use llvm >= 11.

pagict:
The Ubuntu 18.04 uses llvm-6, with kernel 5.4.0, by default, which means bcc would fail compiling.

Yes, ubuntu 18.04 has been released for 5 years, but I believe there are still a lot of production machine run this version.

yonghong-song:
Okay, I guess the change does not hurt either. So I will merge.

PR #4699, opened 2023-08-09
tools/statsnoop: Add more stat() variansts

wenjianhn:
The underlying system call employed by the Golang Stat() wrapper function is newfstatat().
See https://cs.opensource.google/go/go/+/refs/heads/master:src/syscall/zsyscall_linux_amd64.go;l=1453;drc=94e40d5603797fe4d9302421682bb5b997d9c568

### DIFF
--- a/tools/statsnoop.py
+++ b/tools/statsnoop.py
@@ -128,7 +128,7 @@ b = BPF(text=bpf_text)
 def try_attach_syscall_probes(syscall):
     syscall_fnname = b.get_syscall_fnname(syscall)
     if BPF.ksymname(syscall_fnname) != -1:
-        if syscall == "statx":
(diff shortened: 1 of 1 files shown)


tools/statsnoop.py
```diff
@@ -139,6 +139,8 @@ def try_attach_syscall_probes(syscall):
 try_attach_syscall_probes("statfs")
 try_attach_syscall_probes("newstat")
 try_attach_syscall_probes("newlstat")
+try_attach_syscall_probes("fstatat64")
```
yonghong-song, line 142: How about this one? should it belong to `yscall__statx_entry`?

wenjianhn: Yes. It should reuse the `statx_entry`.

PR #4704, opened 2023-08-11
Misc Ipv6 fixes

jeromemarchand:
Quite a few networking tools have trouble with IPv6. Either it's buggy and or there is no IPv6 support it at all.

Here a re a few fixes. Some more should come.


yonghong-song:
Thanks for working on this. I checked the affected tools and they are all working fine.

yonghong-song:
I won't consider 26->39 change as bug. '26' still works. just if the width is more than '26', the format will be a little bit mis-aligned with other entries.

PR #4707, opened 2023-08-16
examples/perf/ipc: Use generic hardware events

sandip4n:
The use of hardcoded raw events breaks compatibility as these are tied to specific processors. On other processors, it could also be the case that the raw event codes represent something completely different. Hence, use the generic events for cycles and instructions as these are available on most platforms with a core PMU.


chenhengqi (approved):
LGTM.

Refs:
* https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/perf_event.h
* https://man7.org/linux/man-pages/man2/perf_event_open.2.html
* https://perfmon-events.intel.com/

PR #4709, opened 2023-08-16
tools/profile: add support for PID-namespacing

lucab:
This adds translation logic to perform PID remapping across PID-namespaces.
It is now possible to profile a target process from within a nested PID-namespace (e.g. in a container).

### DIFF
--- a/tools/profile.py
+++ b/tools/profile.py
@@ -231,7 +231,6 @@ try:
     bpf_text = bpf_text.replace('USE_PIDNS', "1")
     bpf_text = bpf_text.replace('PIDNS_DEV', str(devinfo.st_dev))
     bpf_text = bpf_text.replace('PIDNS_INO', str(devinfo.st_ino))
-    pidns_translation = "1"
 except:
     bpf_text = bpf_text.replace('USE_PIDNS', "0")
     bpf_text = bpf_text.replace('PIDNS_DEV', "0")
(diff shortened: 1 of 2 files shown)


tools/profile.py
```diff
@@ -216,6 +225,18 @@ def stack_id_err(stack_id):
 }
 """
 
+# pid-namespace translation
+try:
+    devinfo = os.stat("/proc/self/ns/pid")
+    bpf_text = bpf_text.replace('USE_PIDNS', "1")
+    bpf_text = bpf_text.replace('PIDNS_DEV', str(devinfo.st_dev))
+    bpf_text = bpf_text.replace('PIDNS_INO', str(devinfo.st_ino))
+    pidns_translation = "1"
```
chenhengqi, line 234: `pidns_translation` is unused.

lucab: Ah right sorry, it's leftover from a previous iteration of the patch, I dropped it now.

yonghong-song:
Can we add an option for this new functionality, something like '--with-pid-namespace'? Note that with '--with-pid-namespace', the command line arguments like pid/tid should be the one in the pid namespace to make the tool work.
With introducing new option, please add a few examples in the example file and update the man page as well.

yonghong-song:
Okay, @lucab Your argument above indeed makes sense. Without this symbolization will be incorrect in child pid namespace.

lucab:
@yonghong-song good to hear, and thanks for taking the time to review this!

PR #4713, opened 2023-08-22
Add BCC script for Calculating Scheduler latency and Decision latency

vineethribm:
This patch adds the bcc script for calculating the Scheduler latency and Scheduler decision latency.

scheduler latency is the amount of time a task waits on a cpu before it actually gets a chance to run.

scheduler decision latency is the time taken by the cpu in deciding the processor on which the task needs to be wokenup.

Usage: scheduler_decision_latency.py [-h] [-c COMM]

comm if given, the script collects stats only for that comm, if not given then collects stats across the entire system

Sample output:

Press ctrl-c to end
^CThe minimum and maximum scheduler latency is 0 and 49 microseconds
The avg scheduler latency is 3.475 microseconds
The 50 th %ile scheduler latency is 2 microseconds
The 90 th %ile scheduler latency is 7 microseconds
The 95 th %ile scheduler latency is 12 microseconds
The 99 th %ile scheduler latency is 20 microseconds
The 99.99 th %ile scheduler latency is 49 microseconds

The minimum and maximum decision latency is 0 and 14 microseconds
The avg decision latency is 2.213 microseconds
The 50 th %ile decision latency is 1 microseconds
The 90 th %ile decision latency is 7 microseconds
The 95 th %ile decision latency is 9 microseconds
The 99 th %ile decision latency is 9 microseconds
The 99.99 th %ile decision latency is 14 microseconds

### DIFF
--- a/tools/scheduler_decision_latency.py
+++ b/tools/scheduler_decision_latency.py
@@ -1,0 +1,136 @@
+#!/usr/bin/python
+# scheduler_decision_latency    collect stats for scheduler and decision latencies
+#                                across the entire system or for a specific comm
+#
+# scheduler latency is the amount of time a task waits on a cpu before it actually
+# gets a chance to run.
+#
+# decision latency is the time taken by the cpu to decide on which cpu the task
+# needs to be woken up.
+#
+# USAGE: scheduler_decision_latency.py [-h] [-c COMM]
+#
+# 23-Aug-2023 Madadi Vineeth Reddy created this.
+
+import argparse
+from bcc import BPF
+import math
+import signal
+
+print("Press ctrl-c to end")
+
+parser = argparse.ArgumentParser()
+parser.add_argument('-c','--comm',help="Collects stats only for the comm specified",default="")
+args = parser.parse_args()
+
+b = BPF(text="""
+#include <linux/sched.h>
+BPF_HASH(sl_ptime);
+BPF_HASH(dl_ptime);
+struct sl_data {
+    u64 sched_latency;
+    char comm[TASK_COMM_LEN];
+};
+struct dl_data {
+    u64 dec_latency;
+    char comm[TASK_COMM_LEN];
+};
+
+BPF_PERF_OUTPUT(sl_events);
+BPF_PERF_OUTPUT(dl_events);
+TRACEPOINT_PROBE(sched,sched_waking)
+{
+    u64 time = bpf_ktime_get_ns();
+    time = time / 1000;
+    u64 proc_id = args->pid;
+    dl_ptime.update(&proc_id,&time);
+    return 0; 
+}
+
+TRACEPOINT_PROBE(sched,sched_wakeup)
+{
+    u64 time = bpf_ktime_get_ns();
+    time = time / 1000;
+    u64 proc_id = args->pid;
+    sl_ptime.update(&proc_id,&time);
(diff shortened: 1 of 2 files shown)


yonghong-song:
Could you explain if you have collected the data described in your above summary what you will do next with these data?

vineethribm:
> Could you explain if you have collected the data described in your above summary what you will do next with these data?

Hi,
Yes, I have collected the data that I described in the above summary.

I had previously used this data to test the[ EEVDF ](https://lwn.net/Articles/925371/) patchset (on x86 and power machines) that brings new scheduling strategy as it mainly aims at reducing the latency values.

This script will be helpful to test any scheduler related patches before and after the patch to gain insights into 
* The time taken for a cpu to place a task on the runqueue (decision latency)
* The time spent by a task waiting on runqueue of a cpu before it actually got chance to run (scheduler latency)

Thanks!

brendangregg:
Didn't read CONTRIBUTING_SCRIPTS.md.

You can't write accurate tracing tools without a lot of attention to detail, as it's so easy to mess up and have a tool that appears to work but gets something wrong under closer scrutiny, or under different workloads. These tools also run as root in production environments, another reason to be careful. So when I see that the contributing documentation hasn't even been followed, I get pretty nervous to begin with.

If there is a strong reason to have percentiles, then I'd add a switch to runqlat instead of inventing a new tool. But given runqlat and runqslower, I don't really see a strong reason to even add this -- how did this solve the patch evaluation in a way that the existing tools didn't?

vineethribm:
Sorry, I was not aware of  CONTRIBUTING_SCRIPTS.md. I had gone through it now and will follow them for further contributions.

And also gone through the runqlat tool. My script contains two parts, scheduler latency (already implemented in runqlat) and scheduler decision latency.

### Regarding 'scheduler decision latency'
* It is the time taken for the processor to decide on which cpu runqueue the task needs to be wokenup .
* In terms of tracepoints, it is the time between sched_waking and sched_wakeup
* This metric will be useful in case of patches affecting `try_to_wake_up` code path (in measuring time taken to do so, so that it can be further optimized). As `try_to_wake_up` code path aims at choosing the cpu on which the task needs to be wokenup.

So, is it a good idea to add this option of measuring `scheduler decision latency` as part of `runqlat` tool itself or as an example script in /example? Would love to hear your thoughts on this @brendangregg 

Thanks!

brendangregg:
Can you provide a production use case of scheduler decision latency being a problem?

If it was frequently a problem, it would should up in a CPU flame graph (sampled stack traces). This would have to be for latency outliers (e.g., 99.99% percentile) caused by scheduler decision latency. Is there an example?

vineethribm:
Hi @brendangregg 
Regarding a production use case, recently our team had a issue reported about high latency in client-server model and as part of analysing, scheduler and decision latency scripts were used to analyze the high tail latency among other things.

Thanks!

brendangregg:
I still don't think this has been answered:

> how did this solve the patch evaluation in a way that the existing tools didn't?

By existing tools I don't just mean bcc tools, I mean all scheduling tools. Do you know what they all are?

vineethribm:
Hi @brendangregg , sorry for the late reply. 
Do you mean tools like mpstat, top, sar or benchmarks like schbench that give the wakeup latency?

Thanks!

PR #4715, opened 2023-08-23
libbpf-tools/filelife: Check btf struct field for CO-RE

Rtoax:
commit abf08576afe3("fs: port vfs_*() helpers to struct mnt_idmap") [0] use `struct mnt_idmap *new_mnt_idmap` instead of `struct user_namespace * old_mnt_userns`.

[0] https://github.com/torvalds/linux/commit/abf08576afe3


JackieLiu1 (commented):
Looks good to me

PR #4720, opened 2023-08-28
cc: remove a spurious check for UID 0

lucab:
This drops an hardcoded check for the root user-ID before accessing `/proc/kallsyms`, fixing a bug which is preventing proper kernel symbols resolution by non-root users.
The `/proc/kallsyms` file is generally available to non-privileged users:
```
> runuser -u nobody -- wc -l /proc/kallsyms

217514 /proc/kallsyms
```

Its content is further controlled by the `CAP_SYSLOG` capability
through `/proc/sys/kernel/kptr_restrict`:
https://man7.org/linux/man-pages/man7/capabilities.7.html

As such, it is better to directly open it and rely on `fopen()`
for privileges checking.

### DIFF
--- a/src/cc/bcc_proc.c
+++ b/src/cc/bcc_proc.c
@@ -262,10 +262,6 @@ int bcc_procutils_each_ksym(bcc_procutils_ksymcb callback, void *payload) {
   FILE *kallsyms;
   unsigned long long addr;
 
-  /* root is needed to list ksym addresses */
-  if (geteuid() != 0)
-    return -1;
-
   kallsyms = fopen("/proc/kallsyms", "r");
   if (!kallsyms)
     return -1;


yonghong-song:
Make sense. In many systems, 'root' privilege is not needed to read `/proc/kallsyms`.

chenhengqi:
> fixing a bug which is preventing proper kernel symbols resolution by non-root users.

@lucab  What kind of bug here ?

chenhengqi:
What's your use case ? 

IIRC, a non-root user won't pass the following check.

https://github.com/iovisor/bcc/blob/ec49363e2e9daec026ee6cae4c5fc316f8fab0ff/src/cc/bcc_proc.c#L271

chenhengqi:
Fine. Instead of removing the check, I guess we could replace it with a capabilities check ?

lucab:
@chenhengqi that's not really needed, the current logic already degrades gracefully in all cases:
 * profiler running with `CAP_SYSLOG` can open `/proc/kallsyms` and read all symbols plus addresses. It will properly capture a profile and symbolize kernel frames.
 * profiler running without `CAP_SYSLOG` can open `/proc/kallsyms` and read all symbols, but all the addresses are zeroed. These all-zeroes addresses are filtered out by the code you linked above. The profiler will properly capture a profile, but all kernel frame will be marked as `unknown`.

PR #4721, opened 2023-08-29
libbpf-tools: Modify error message on uprobe_helpers

Bojun-Seo:
function `get_pid_lib_path` on file `libbpf-tools/uprobe_helpers.c` prints error message like followings if it failed to open `proc_pid_maps`
"No such pid 1234"
But fopen can be failed even though pid `1234` is alive. So change error message to
"Failed to open /proc/1234/maps"

### DIFF
--- a/libbpf-tools/uprobe_helpers.c
+++ b/libbpf-tools/uprobe_helpers.c
@@ -65,7 +65,7 @@ int get_pid_lib_path(pid_t pid, const char *lib, char *path, size_t path_sz)
 	}
 	maps = fopen(proc_pid_maps, "r");
 	if (!maps) {
-		warn("No such pid %d\n", pid);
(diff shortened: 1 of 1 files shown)


yonghong-song:
Could you give a detailed situation when /proc/<pid>/maps open is failing with a valid pid?

Bojun-Seo:
In case if the process doesn't have permission but tries to open /proc/pid/maps

yonghong-song:
> In case if the process doesn't have permission but tries to open /proc/pid/maps
I understand this. I actually want to know your case when /proc/pid/maps cannot be accessed? What kind of changes you made to the /proc/pid/maps or system in general. By default, all user pid maps can be accessed by bpf program since typically bpf tool is running with root privilege.

PR #4736, opened 2023-09-12
Core swapin

matthew-olson-intel:


yonghong-song:
This tool tried to count the number of swap_in's. But actually, such information can already be got by taskstats.
```
/usr/include/linux/taskstats.h:
        /* Delay waiting for page fault I/O (swap in only) */
        __u64   swapin_count;
        __u64   swapin_delay_total;
```
and it is much cheaper compared to bpf approach and swapin_delay_total can give actual 'time cost'
for all these swapin_count which is important to determine the real cost.

PR #4737, opened 2023-09-13
Use LLVM_VERSION_MAJOR from llvm-config.h instead of inferred LLVM_MAJOR_VERSION

chantra:
Instead of inferring this from within the CMakeList.txt file, we can pull it directly from llvm-config.h.


chantra:
hum... I am not sure those errors are related.  Forcing an update to re-trigger build.

chantra:
ok, tests that previously failed, now succeed. Tests that previously succeeded, now fail.

src/cc/bcc_debug.cc
```diff
@@ -19,7 +19,7 @@
 #include <tuple>
 #include <vector>
 
-#if LLVM_MAJOR_VERSION >= 15
```
chenhengqi, line 22: Too much code churn, could we avoid this change here ?

chantra: @chenhengqi what do you mean by too much code churn?

This is essentially a `sed s/LLVM_MAJOR_VERSION/LLVM_VERSION_MAJOR/` on the files, e.g from internally defined to the one provided by llvm itself.

The main reason for doing this is that when using alternate build systems such as buck, bazel... the LLVM version get automatically discovered at build time without needing some special handling by the build system.
Given that this is readily available, we can use it directly instead of re-inventing the wheel.

chenhengqi: > without needing some special handling by the build system

But we still rely on cmake to find the LLVM libraries, no ?

chantra: In a build agnostic system, the build system will find the llvm libraries (cmake, make, bazel, buck....), yes.
LLVM library itself (via llvm-config.h https://github.com/llvm-mirror/llvm/blob/2c4ca6832fa6b306ee6a7010bfb80a3f2596f824/include/llvm/Config/llvm-config.h.cmake#L69 ) exposes the major version (and more).

Currently, this is inferred from https://github.com/iovisor/bcc/blob/ec49363e2e9daec026ee6cae4c5fc316f8fab0ff/src/cc/CMakeLists.txt#L47 

I am replacing this by llvm native defines.

davemarchevsky (approved):

PR #4743, opened 2023-09-20
Adding memory zones for Power server

dubeyabhishek:
config PPC_BOOK3S_64 skips setting ZONE_DMA for
server processors. NORMAL and MOVABLE zones are
available on Power.


dubeyabhishek:
@chenhengqi can we close this PR also?

dubeyabhishek:
@yonghong-song This PR lists the correct memory zones for PPC64 processors. Can we close this PR?

PR #4745, opened 2023-09-20
Test multiple LLVM versions

chantra:
Currently, only llvm-11 get tested on Ubuntu, on Fedora, it is whatever is the default llvm package of that Fedora release.

Given the checks on LLVM versions in the code base, it would be useful to test different LLVM versions too.

This change allow installing multiple versions when building the Ubuntu docker image, and when building bcc, LLVM_ROOT is used to force building against a specific llvm version.

Ubuntu bionic had failure installing llvm 15, so I ma skipping it for that version of ubuntu.
On the other hand, I am not sure how much value testing on Bionic has. Given the tests are running in docker, the kernel bcc is tested against are the same so.
Also Bionic reached end of standard support https://ubuntu.com/18-04 .
Removing this Ubuntu flavor would make sense in my opinion. Happy to do this on top of this stack.

### DIFF
--- a/.github/workflows/bcc-test.yml
+++ b/.github/workflows/bcc-test.yml
@@ -132,7 +132,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        os: [{distro: "fedora", version: "34", nick: "f34"}, {distro: "fedora", version: "36", nick: "f36"}]
+        os: [{distro: "fedora", version: "38", nick: "f38"}]
         env:
         - TYPE: Debug
           PYTHON_TEST_LOGFILE: critical.log
--- a/.github/workflows/bcc-test.yml
+++ b/.github/workflows/bcc-test.yml
@@ -21,7 +21,8 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        os: [{distro: "ubuntu", version: "18.04", nick: bionic}, {distro: "ubuntu", version: "20.04", nick: focal}]
+        os: [{distro: "ubuntu", version: "18.04", nick: bionic, installed_llvm_versions: "11 12"}, {distro: "ubuntu", version: "20.04", nick: focal, installed_llvm_versions: "11 12 15"}]
+        llvm_version: [11, 12, 15]
         env:
         - TYPE: Debug
           PYTHON_TEST_LOGFILE: critical.log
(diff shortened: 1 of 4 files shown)


.github/workflows/bcc-test.yml
```diff
@@ -21,7 +21,8 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        os: [{distro: "ubuntu", version: "18.04", nick: bionic}, {distro: "ubuntu", version: "20.04", nick: focal}]
+        os: [{distro: "ubuntu", version: "18.04", nick: bionic, installed_llvm_versions: "11 12"}, {distro: "ubuntu", version: "20.04", nick: focal, installed_llvm_versions: "11 12 15"}]
```
davemarchevsky, line 24: IIUC, `installed_llvm_versions` / `llvm_versions` is used to populate almost everything, while `llvm_version` here is just needed for a few things:
  * job name below (`name: Run bcc build - llvm-${{ matrix.llvm_version }}`)
  * `LLVM_ROOT` below

Do these need to be kept in sync? e.g. here you have `os:` line w/ `installed_llvm_versions: "11 12"` and `installed_llvm_versions: "11 12 15"` - does the `llvm_version:` line added one line below need to contain all the `installed_llvm_versions` ?

davemarchevsky: Aside from this comment, everything in this PR makes sense to me, I think it's good to go

chantra: so... this is kind of messy, but I did not find a better way to fix this. Well, a simple way would be to retire Ubuntu 18.04.

```
{distro: "ubuntu", version: "20.04", nick: focal, installed_llvm_versions: "11 12 15"}
```
Essentially drives the container creation/usage, this `installed_llvm_versions` indeed specify that we want to install LLVM versions 11, 12, and 15 in the container.

The `llvm_version` below drives the matrix.
The problem is that GH will generate a combination of all os/llvm_version/env tuples. There is no way to associate a separate `llvm_version` for ubuntu bionic, or infer the llvm_versions from the installed_llvm_versions.
Which explains there is the `exclude` specifically for ubuntu bionic and llvm 15.

> llvm_version: line added one line below need to contain all the installed_llvm_versions

it is rather the opposite.... `installed_llvm_versions` needs to be a superset of `llvm_version`. The former drives what is installed during docker build in the container, the latter, which of those specific versions we are going to use.

In the PR I commented:
> Ubuntu bionic had failure installing llvm 15, so I ma skipping it for that version of ubuntu.

If we just retire bionic, I think it would simplify this workflow.

Separately, I think the container logic could be aggregated in a separate action that can be re-used eliminating code duplication. I will give this a try.

davemarchevsky: Thanks for the explanation!

> If we just retire bionic, I think it would simplify this workflow.

No objection from me, but let's not hold up this PR for it

chantra: let me follow up with the retirement from that build. It does not make much sense to keep it when the distro itself is EOL.

PR #4747, opened 2023-09-21
memleak: use tid instead of tgid as map key

### LINKED ISSUE #4741: memleak has concurrency problem?
``` c
static int gen_alloc_exit2(void *ctx, u64 address)
{
	const pid_t pid = bpf_get_current_pid_tgid() >> 32;
	struct alloc_info info;

	const u64* size = bpf_map_lookup_elem(&sizes, &pid);
	if (!size)
		return 0; // missed alloc entry
```

the pid is tgid, so if a process has multi thread alloc memory in the same time, this pid key in sizes map will conflict?

inwardvessel:
some of the maps were using tgid as key but should really be using tid so that multithreaded programs can record consistent allocation data.

fixes #4741

### DIFF
--- a/libbpf-tools/memleak.bpf.c
+++ b/libbpf-tools/memleak.bpf.c
@@ -18,7 +18,7 @@ const volatile bool wa_missing_free = false;
 
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
-	__type(key, pid_t);
+	__type(key, u32);
 	__type(value, u64);
 	__uint(max_entries, 10240);
 } sizes SEC(".maps");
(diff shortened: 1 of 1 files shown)


yangguosdxl (changes requested):
I think you need use the hole pid composed by tgid and  tid, this is the only id for all process's thread.otherwise, different process thread may conflict the 32 bits thread id?

chenhengqi:
> I think you need use the hole pid composed by tgid and tid, this is the only id for all process's thread.otherwise, different process thread may conflict the 32 bits thread id?

No, tid is unique.

PR #4752, opened 2023-10-01
tools/gotlskeylog: trace TLS secrets in golang app

bacher09:
This adds a new tool: `gotlskeylog` that can be used to intercept TLS secrets in a [SSLKEYLOGFILE](https://datatracker.ietf.org/doc/html/draft-thomson-tls-keylogfile-01) format that can be used in a Wireshark to decrypt TLS traffic. The main use case is sniffing Golang applications.

Currently, this supports only **amd64** architecture and applications that compiled with a relatively recent enough compiler, since golang ABI is different on i386 (arguments on a stack) and I haven't tested it on arm. Older versions of golang compiler used stack on amd64 as well.

Typical use is starting `tcpdump` or a `wireshark`, while intercepting TLS secrets and using these secrets to decrypt TLS sessions of a tracing application.


bacher09:
I've made another [version](https://github.com/bacher09/bcc/blob/gotlskeylog_x86/tools/gotlskeylog.py) that supports both 32bit binaries (in a 64 bit OS) and 64 bit ones. Although, I'm uncertain if it's a good idea to add it here, since most of `uprobe` based tools here's aren't supporting 32 bit binaries.

chenhengqi:
Have you tried [eCapture](https://github.com/gojue/ecapture) ?

bacher09:
> Have you tried [eCapture](https://github.com/gojue/ecapture) ?

Thanks, I just tried it, and it seems better almost in all ways, so I'm going to close this **PR**.

There are only 2 differences, it supports tracing only a single binary, so if you want to trace multiple ones, you need to launch multiple instances, while my version **can trace** multiple golang binaries at the same time using a same eBPF program. It also aren't supporting 32 bit binaries on x86_64, but I did this mostly for fun, so I don't really need this.

PR #4760, opened 2023-10-12
libbpf-tools: Fix misaligned pointer accesses in exitsnoop

captain5050:
The perf buffer in exit snoop doesn't maintain 8 byte alignment for start_time and exit_time in struct event. When building with "-fsanitize=alignment -fsanitize-undefined-trap-on-error" failures happen in handle_event. Fix these by copying the event from the perf buffer before accessing.

### DIFF
--- a/libbpf-tools/exitsnoop.c
+++ b/libbpf-tools/exitsnoop.c
@@ -120,12 +120,19 @@ static void sig_int(int signo)
 
 static void handle_event(void *ctx, int cpu, void *data, __u32 data_sz)
 {
-	struct event *e = data;
+	struct event e;
 	time_t t;
 	struct tm *tm;
 	char ts[32];
 	double age;
 	int sig, coredump;
+
+	if (data_sz < sizeof(e)) {
+		printf("Error: packet too small\n");
+		return;
+	}
+	/* Copy data as alignment in the perf buffer isn't guaranteed. */
+	memcpy(&e, data, sizeof(e));
 
 	if (emit_timestamp) {
 		time(&t);
@@ -134,18 +141,18 @@ static void handle_event(void *ctx, int cpu, void *data, __u32 data_sz)
 		printf("%8s ", ts);
 	}
 
-	age = (e->exit_time - e->start_time) / 1e9;
+	age = (e.exit_time - e.start_time) / 1e9;
 	printf("%-16s %-7d %-7d %-7d %-7.2f ",
-	       e->comm, e->pid, e->ppid, e->tid, age);
+	       e.comm, e.pid, e.ppid, e.tid, age);
 
-	if (!e->sig) {
-		if (!e->exit_code)
+	if (!e.sig) {
+		if (!e.exit_code)
 			printf("0\n");
 		else
-			printf("code %d\n", e->exit_code);
+			printf("code %d\n", e.exit_code);
 	} else {
-		sig = e->sig & 0x7f;
-		coredump = e->sig & 0x80;
+		sig = e.sig & 0x7f;
+		coredump = e.sig & 0x80;
 		if (sig)
 			printf("signal %d (%s)", sig, strsignal(sig));
 		if (coredump)


MaskRay:
Thanks for the patch.

> -fsanitize=alignment -fsanitize-undefined-trap-on-error

Consider changing the deprecated `-fsanitize-undefined-trap-on-error` to `-fsanitize-trap=undefined` in the description to discourage new uses of `-fsanitize-undefined-trap-on-error`:) (`-fsanitize-trap=undefined` is in Clang for a long time but relatively new in GCC.)

chenhengqi:
> The perf buffer in exit snoop doesn't maintain 8 byte alignment for start_time and exit_time in struct event.

Please include the error message here.

IIRC, raw perf events are 64-bit aligned, though it could contain trailing garbage.

captain5050:
> > The perf buffer in exit snoop doesn't maintain 8 byte alignment for start_time and exit_time in struct event.
> 
> Please include the error message here.
> 
> IIRC, raw perf events are 64-bit aligned, though it could contain trailing garbage.

There's no particular error message except SIGILL:

```
$ make EXTRA_CFLAGS="-fsanitize=alignment -fsanitize-trap=undefined -O0 -g" -j16 exitsnoop
  CC       exitsnoop.o
  BINARY   exitsnoop
$ sudo ./exitsnoop PCOMM            PID     PPID    TID     AGE(s)  EXIT_CODE
Illegal instruction
$ sudo gdb --args ./exitsnoop 
GNU gdb (Debian 13.2-1) 13.2
Copyright (C) 2023 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./exitsnoop...
(gdb) r
Starting program: /home/irogers/bcc/libbpf-tools/exitsnoop 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
PCOMM            PID     PPID    TID     AGE(s)  EXIT_CODE

Program received signal SIGILL, Illegal instruction.
0x000055555555d14e in handle_event (ctx=0x0, cpu=3, data=0x7ffff7d5700c, data_sz=60)
    at exitsnoop.c:137
137             age = (e->exit_time - e->start_time) / 1e9;
(gdb) bt
#0  0x000055555555d14e in handle_event (ctx=0x0, cpu=3, data=0x7ffff7d5700c, data_sz=60)
    at exitsnoop.c:137
#1  0x000055555559cee5 in perf_buffer__process_record (e=0x7ffff7d57000, ctx=0x5555555d5200)
    at libbpf.c:11913
#2  0x000055555559ba73 in perf_event_read_simple (mmap_mem=0x7ffff7d56000, mmap_size=65536, 
    page_size=4096, copy_mem=0x5555555d5210, copy_size=0x5555555d5218, 
    fn=0x55555559cd7d <perf_buffer__process_record>, private_data=0x5555555d5200) at libbpf.c:11541
#3  0x000055555559d04a in perf_buffer__process_records (pb=0x5555555d4f80, cpu_buf=0x5555555d5200)
    at libbpf.c:11935
#4  0x000055555559d14f in perf_buffer__poll (pb=0x5555555d4f80, timeout_ms=100) at libbpf.c:11960
#5  0x000055555555d88f in main (argc=1, argv=0x7fffffffe408) at exitsnoop.c:246
(gdb) p/x e
$1 = 0x7ffff7d5700c
(gdb) p/x (long)e & ~7
$2 = 0x7ffff7d57008
```

I see the BPF ring buffer 8 byte aligning:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/kernel/bpf/ringbuf.c#n414

I see perf events manually aligning themselves to 8 bytes:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/kernel/events/core.c#n8525

I don't see alignment in the perf code:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/kernel/events/ring_buffer.c#n149
except for bits of adhoc alignment:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/kernel/events/core.c#n7773

MaskRay:
FWIW [C11 6.3.2.3p7](https://port70.net/~nsz/c/c11/n1570.html#6.3.2.3p7) says

> A pointer to an object type may be converted to a pointer to a different object type. If the resulting pointer is not correctly aligned for the referenced type, the behavior is undefined.

yonghong-song:
I can reproduce the issue with gcc or clang. After some investigation, I guess the reason is below. 
Now that the perf record we are collecting here are RAW perf record. Look at /usr/include/linux/perf_event.h, we have
```
         *      # The RAW record below is opaque data wrt the ABI
         *      #
         *      # That is, the ABI doesn't make any promises wrt to
         *      # the stability of its content, it may vary depending
         *      # on event, hardware, kernel version and phase of
         *      # the moon.
         *      #
         *      # In other words, PERF_SAMPLE_RAW contents are not an ABI.
         *      #
         *
         *      { u32                   size;   
         *        char                  data[size];}&& PERF_SAMPLE_RAW
```

For each RAW record, first is a perf_event_header
```
struct perf_event_header {
        __u32   type;
        __u16   misc;
        __u16   size;
};
```
and then followed by perf record type specific fields.

So for sampling type, we have total 12 bytes (8 for 'perf_event_header' and 4 byte for above 'size') before the real data. This is aligning with above core stack as well (esp. perf_buffer__process_record() and handle_event()).

So it looks like we not only have this issue for exitsnoop, but for other tools as well (e.g., biosnoop, drsnoop, etc.).
Could you check other tools as well and fix them similarly? If 'struct event' alignment is 4, we should be okay. Otherwise, as you patch implemented, copy is necessary.

chenhengqi:
I found that the simplest way to fix the issue is:
```diff
diff --git a/libbpf-tools/exitsnoop.h b/libbpf-tools/exitsnoop.h
index e42d156f..08b5f595 100644
--- a/libbpf-tools/exitsnoop.h
+++ b/libbpf-tools/exitsnoop.h
@@ -4,7 +4,7 @@
 
 #define TASK_COMM_LEN  16
 
-struct event {
+struct __attribute__((__packed__)) event {
        __u64 start_time;
        __u64 exit_time;
        __u32 pid;
```

yonghong-song:
Thanks @chenhengqi for the suggestion. Indeed, adding '__attribute__((packed))' will prevent sanitizer to check alignment so we won't have issue any more. arm64 and x86 should handle misaligned load/store efficiently (almost the same as aligned). Some architecture does not support efficient misalign access (e.g., MIPS, SPARC?) but there should be a hardware interrupt to do the work automatically. @captain5050 could you change the code based on @chenhengqi 's suggestion?

captain5050:
Thanks for the feedback and particularly the perf sample details!

Using attribute packed is less than ideal because:

1. the variables in the struct have no padding between them and so they are naturally "packed". We're using packed for a side-effect which should at least be commented.
2. packed isn't sticky, so taking the address of a struct variable loses the "packed" notion and the compiler will assume again that it is aligned or else undefined behavior (UB).

For these reasons it is better to memcpy, the compiler able to reason the src of the memcpy is unaligned but afterward things will be aligned without a risk of introducing UB.

> Could you check other tools as well and fix them similarly?

The change is simple, I'm happy to help. I haven't observed other failures so I may have a challenge to reproduce and check a fix. Given that I'd need to learn each tool to produce a fix, would it be possible for the maintainer of the tool to fix it? Fixing other tools doesn't need to gate this patch. Thanks again!

chenhengqi:
`memcpy` is suboptimal for efficiency.

yonghong-song:
I did some further checking. Looks bcc does memcpy as well.
   https://github.com/iovisor/bcc/blob/master/src/cc/perf_reader.c#L187-L191
bpftrace uses bcc codes for perf handling so bpftrace also uses memcpy.

The below is libbpf sample_fn callback function:
```
typedef void (*perf_buffer_sample_fn)(void *ctx, int cpu,
                                      void *data, __u32 size);
```
The `data` is `void *` type, which actually means its alignment is unknown (although in practice it is 4-byte aligned).
Also libbpf  tools might be used in architectures other than x64, so maybe doing memcpy is the right thing if the 'data' handling requires 8-byte alignment. Let me check with a few other kernel folks (esp. what is the concern during the design in libbpf).

yonghong-song:
Sync'ed with @anakryiko Since the data pointer (as below) has type 'void *' which implies alignment of 1 although advanced users might be dig out that actual alignment is 4, application doing memcpy() sounds a good idea to
make the code portable. Typically the data size should not be that big so the overhead should be very small.

Will merge this patch and other similar cases (biosnoop, drsnoop, etc.) can be handled later.

anakryiko:
Oh, I'd handle it in a code a bit more efficiently and nicely. Basically just add this on top and don't change anything else:

```
struct event e_copy;
struct *e;

if (data % alignof(*e)) {
    memcpy(&e_copy, data, sizeof(*e));
    e = &e_copy;
} else {
    e = data;
}
```

yonghong-song:
@captain5050 could you have a followup with the above code suggested by @anakryiko? This code pattern, which has minimum diff's can be used for other tools in the future. Thanks!

captain5050:
So if I try this in godbolt I find the code is worse with both gcc and clang. When there is an unconditional memcpy both gcc and clang can remove it, whereas the conditional one leads to larger code: https://godbolt.org/z/1sj4o85aa


```
int foo2(void *buf)
{
    struct abc abc;
    memcpy(&abc, buf, sizeof(abc));
    if (abc.a > abc.b)
        return 1;
    return -1;
}
```

```
foo2(void*):                              # @foo2(void*)
        movq    (%rdi), %rax
        xorl    %ecx, %ecx
        cmpq    8(%rdi), %rax
        setg    %cl
        leal    -1(,%rcx,2), %eax
        retq
```

```
int foo4(void *buf)
{
    struct abc abc_copy;
    struct abc *abc;
    if ((unsigned long)buf % alignof(abc)) {
       memcpy(&abc_copy, buf, sizeof(abc_copy));
       abc = &abc_copy;
    } else {
        abc = reinterpret_cast<struct abc*>(buf);
    }
    if (abc->a > abc->b)
        return 1;
    return -1;
}
```

```
foo4(void*):                              # @foo4(void*)
        testb   $7, %dil
        je      .LBB3_2
        movq    16(%rdi), %rax
        movq    %rax, -8(%rsp)
        movups  (%rdi), %xmm0
        movaps  %xmm0, -24(%rsp)
        leaq    -24(%rsp), %rdi
.LBB3_2:
        movq    (%rdi), %rax
        xorl    %ecx, %ecx
        cmpq    8(%rdi), %rax
        setg    %cl
        leal    -1(,%rcx,2), %eax
        retq
```

I think I favor the memcpy and changing '->' to '.' , I find the code easier to read with less cognitive load, the compiler does a better job with it. Unfortunately the '->' to '.' makes the change a little large and more invasive.

anakryiko:
How can compiler remove memcpy() in foo2() if there is no guarantee that pointer is aligned? It just opts to do unaligned 8-byte reads from memory?

I mean, I don't particularly care, I was just trying to help minimize changes.

anakryiko:
Interesting! Thanks, @yonghong-song!

PR #4762, opened 2023-10-16
python: Fix device or resource busy for `trace_pipe`

ShawnZhong:
Currently, the following code throws an error `Device or resource busy: '/sys/kernel/debug/tracing/trace_pipe'`

```py
from bcc import BPF

BPF(text="int kprobe__sys_execve(void *ctx) { bpf_trace_printk(\"execve\\n\"); return 0; }").trace_readline()
BPF(text="int kprobe__sys_clone(void *ctx) { bpf_trace_printk(\"clone\\n\"); return 0; }").trace_readline()
```

```
  File "/usr/lib/python3/dist-packages/bcc/__init__.py", line 1546, in trace_readline
    trace = self.trace_open(nonblocking)
  File "/usr/lib/python3/dist-packages/bcc/__init__.py", line 1497, in trace_open
    _tracefile = open("%s/trace_pipe" % TRACEFS, "rb")
OSError: [Errno 16] Device or resource busy: '/sys/kernel/debug/tracing/trace_pipe'
```

The same behavior will also happen in an IPython/Jupyter Notebook environment where multiple BPF objects are created during one execution. 

This is because each BPF object opens its own `/sys/kernel/debug/tracing/trace_pipe` file, preventing the other BPF object from opening it. 

This PR makes the `tracefile` variable as a module variable to avoid multiple handles to the same file.


yonghong-song:
Theoretically it is possible. But I cannot reproduce the issue with your example.
I tried with
```
$ cat hello_world.py 
#!/usr/bin/python
from bcc import BPF

BPF(text='int kprobe____x64_sys_clone(void *ctx) { bpf_trace_printk("Hello, World1!\\n"); return 0; }').trace_print()
BPF(text='int kprobe____x64_sys_nanosleep(void *ctx) { bpf_trace_printk("Hello, World2!\\n"); return 0; }').trace_print()
```
and it only dumps for clone syscall.
Change the code to be
```
#!/usr/bin/python
from bcc import BPF

BPF(text='int kprobe____x64_sys_clone(void *ctx) { bpf_trace_printk("Hello, World1!\\n"); return 0; }').trace_readline()
BPF(text='int kprobe____x64_sys_nanosleep(void *ctx) { bpf_trace_printk("Hello, World2!\\n"); return 0; }').trace_readline()
```
cannot reproduce the issue either.

I guess to trigger this, you need two *live* bpf programs in the same python runtime context and both programs are trying to do bpf_trace_printk. But please do provide a reproducible test case.

4ast (changes requested):
131 commits?! Something wrong with this PR.

PR #4763, opened 2023-10-16
tools/trace: don't raise an exception in a ctype callback

### LINKED ISSUE #3049: /usr/share/bcc/tools/trace cannot exit from loops
On python 3.8.5 and kernel 5.8
root@intel-x86-64:~# /usr/share/bcc/tools/trace do_sys_openat2 -M 1
179     179     systemd-journal do_sys_openat2   
Exception ignored on calling ctypes callback function: <function PerfEventArray._open_perf_buffer.<locals>.raw_cb_ at 0x7f2a75f060d0>
  File "/usr/lib64/python3.8/site-packages/bcc/table.py", line 685, in raw_cb_
    callback(cpu, data, size)
  File "/usr/share/bcc/tools/trace", line 630, in print_event
    exit()
  File "/usr/lib64/python3.8/_sitebuiltins.py", line 26, in __call__
    raise SystemExit(code)
SystemExit: None
The above repeated endlessly

The following change could make it work, but not sure if it's the best way to do.

diff --git a/tools/trace.py b/tools/trace.py
index 7a61594f..d09dd2a8 100755
--- a/tools/trace.py
+++ b/tools/trace.py
@@ -627,7 +627,7 @@ BPF_PERF_OUTPUT(%s);
                 Probe.event_count += 1
                 if Probe.max_events is not None and \
                    Probe.event_count >= Probe.max_events:
"-                        exit()"
"+                        os._exit(0)"
                 sys.stdout.flush()
 
"         def attach(self, bpf, verbose):"

jeromemarchand:
To exit the tool when the maximal number of event is reached (-M option), the tool currently call exit(), which raise a SystemExit exception. The handling of exception from ctype callback doesn't seem straightforward and dependent on python version.

This patch avoid the issue altogether by using a global variable instead.

Closes #3049

### DIFF
--- a/tools/trace.py
+++ b/tools/trace.py
@@ -43,6 +43,7 @@ class Probe(object):
         build_id_enabled = False
         aggregate = False
         symcount = {}
+        done = False
 
         @classmethod
         def configure(cls, args):
@@ -635,7 +636,7 @@ BPF_PERF_OUTPUT(%s);
                     if self.aggregate:
                         self.print_aggregate_events()
                     sys.stdout.flush()
-                    exit()
+                    Probe.done = True;
 
         def attach(self, bpf, verbose):
                 if len(self.library) == 0:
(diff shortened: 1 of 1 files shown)


yonghong-song:
I still cannot reproduce the issue in my local environment. But looks some people do hit this issue and the fix looks good to me. So I will go ahead to merge it. Thanks!

PR #4766, opened 2023-10-17
memleak print a stackcollapsed report for generate flamegraph

pagict:
By reporting combined allocations at exit, with stack collapsed format, we can later easily generate a flame graph of potential memory leaks.
With this flame graph, leaks are more friendly to inspect at userland's call stacks.


pagict:
also fixed issue #4767

yonghong-song:
@pagict there is conflict. Could you rebase on top of master?
The patch intends to dump collapsed stack in order to generate flame graph. It would be good you can
give a detailed example, e.g., with a specific collapsed stack, with which tool, you can generate what
flame graph. and then you can demonstrate that flame graph is more useful to spot memleak issues.

PR #4770, opened 2023-10-18
Fix compilation issues of bcc on Alibaba Cloud Linux 3

suxb201:
When trying to compile bcc on Alibaba Cloud Linux 3, I found several issues that I've fixed in this PR:

- **LLVM10 Compilation**: Alibaba Cloud Linux 3 is compatible with CentOS 8 and does not support epel-release and centos-release-scl. This means we need to manually compile llvm10.
- **Clang Search Directory**: There was a problem with the search directory for clang. I've added `/usr/local/lib/` to fix this.
- **Python Environment with Conda**: I use conda to manage my Python environment. I found that bcc would install to `/usr/lib/python3.9/site-packages/` by default, not `/root/miniconda3/lib/python3.9/site-packages/`. By removing the `--prefix` argument, bcc now installs correctly in different environments. Without this change, bcc would end up in the wrong Python directory and using `python setup.py install --prefix " "`.

I appreciate any review and feedback.

### DIFF
--- a/src/python/CMakeLists.txt
+++ b/src/python/CMakeLists.txt
@@ -37,14 +37,10 @@ foreach(PY_CMD ${PYTHON_CMD})
   )
   add_custom_target(bcc_py_${PY_CMD_ESCAPED} ALL DEPENDS ${PIP_INSTALLABLE})
 
-  if(NOT PYTHON_PREFIX)
-     set(PYTHON_PREFIX, ${CMAKE_INSTALL_PREFIX} )
-  endif()
-
   install(
     CODE "
       execute_process(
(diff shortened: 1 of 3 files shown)


src/python/CMakeLists.txt
```diff
@@ -37,14 +37,10 @@ foreach(PY_CMD ${PYTHON_CMD})
   )
   add_custom_target(bcc_py_${PY_CMD_ESCAPED} ALL DEPENDS ${PIP_INSTALLABLE})
 
-  if(NOT PYTHON_PREFIX)
-     set(PYTHON_PREFIX, ${CMAKE_INSTALL_PREFIX} )
-  endif()
```
yonghong-song, line 42: Can you specify the PYTHON_PREFIX explicitly in order to install at your desired place?
The following patch
  https://github.com/iovisor/bcc/commit/6813fbce8718bede5075272cdc3ab9314eb91b5c
introduced the feature.

PR #4772, opened 2023-10-20
biolatency: Fix kernel version check.

ntnx-aleksa:
The tracepoint definition change in v5.11-rc1 mentioned in the biolatency program seems to have been [backported into v5.10.137](https://elixir.bootlin.com/linux/v5.10.137/source/include/trace/events/block.h#L206).
This small fix allows the program to load and work on kernels between v5.10.137 and v5.11.


yonghong-song:
We have a better fix already merged. The pull request link is https://github.com/iovisor/bcc/pull/4748. Let us close this pull request then.

PR #4773, opened 2023-10-20
Minor fixes for vfscount and xfsslower

gautammenghani:
This PR fixes 2 minor issues:
1. vfsscount_example : A typo in example usage of help
2. xfsslower : Print utf-8 decoded data instead of bytes

### DIFF
--- a/tools/vfscount_example.txt
+++ b/tools/vfscount_example.txt
@@ -51,5 +51,5 @@ You can edit the script to customize what kernel functions are matched.
 
 Full usage:
 
-# ./vfsstat -h
-USAGE: ./vfsstat [time]
+# ./vfscount.py -h
+USAGE: ./vfscount.py [time]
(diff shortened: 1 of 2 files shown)


yonghong-song:
Could you add more details in commit message, esp. for 'xfsslower' change? What command line you are using? what kind of output you observed? This will allow people to reproduce the issue and judge whether the fix is good or not.

gautammenghani:
I've already mentioned the commands used and output observed - https://github.com/iovisor/bcc/pull/4773/commits/244a35e62a22f8cbbce6df3e9b7004e012c18644

yonghong-song:
> I've already mentioned the commands used and output observed - [244a35e](https://github.com/iovisor/bcc/commit/244a35e62a22f8cbbce6df3e9b7004e012c18644)

Thanks. It will be good for you to add similar messages in github pull request description. Typically people will first look at pull request description to understand what is the problem and how the solution resolves the problem etc.

PR #4776, opened 2023-10-23
libbpf-tools: Fixes for 32-bit platforms

jeez:
Fix build errors when compiling for 32-bit platforms by making the code more portable.

### DIFF
--- a/libbpf-tools/bindsnoop.c
+++ b/libbpf-tools/bindsnoop.c
@@ -168,9 +168,9 @@ static void handle_event(void *ctx, int cpu, void *data, __u32 data_sz)
 		i++;
 	}
 	if (e->ver == 4) {
-		inet_ntop(AF_INET, &e->addr, addr, sizeof(addr));
+		inet_ntop(AF_INET, e->addr, addr, sizeof(addr));
 	} else {
-		inet_ntop(AF_INET6, &e->addr, addr, sizeof(addr));
+		inet_ntop(AF_INET6, e->addr, addr, sizeof(addr));
 	}
 	printf("%-7d %-16s %-3d %-5s %-5s %-4d %-5d %-48s\n",
 	       e->pid, e->task, e->ret, proto, opts, e->bound_dev_if, e->port, addr);
--- a/libbpf-tools/trace_helpers.c
+++ b/libbpf-tools/trace_helpers.c
@@ -7,6 +7,7 @@
 #define _GNU_SOURCE
 #endif
 #include <ctype.h>
+#include <inttypes.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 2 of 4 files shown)


jeez:
cc @anakryiko too since you've reviewed similar fixes in the past.

libbpf-tools/trace_helpers.c
```diff
@@ -552,7 +553,8 @@ static int create_tmp_vdso_image(struct dso *dso)
 		return -1;
 
 	while (true) {
-		ret = fscanf(f, "%lx-%lx %*s %*x %*x:%*x %*u%[^\n]",
+		ret = fscanf(f,
+			     "%" PRIx64 "-%" PRIx64 "%*s %*x %*x:%*x %*u%[^\n]",
```
anakryiko, line 557: I really dislike those PRIxxx macros, in libbpf we just explicitly cast to (long long) and use %lld, can you please do the same here?

libbpf-tools/trace_helpers.c
```diff
@@ -669,10 +671,11 @@ struct syms *syms__load_file(const char *fname)
 		goto err_out;
 
 	while (true) {
-		ret = fscanf(f, "%lx-%lx %4s %lx %lx:%lx %lu%[^\n]",
-			     &map.start_addr, &map.end_addr, perm,
-			     &map.file_off, &map.dev_major,
-			     &map.dev_minor, &map.inode, buf);
+		ret = fscanf(f,
+			     "%" PRIx64 "-%" PRIx64 "%4s %" PRIx64 "%" PRIx64
+			     ":%" PRIx64 "%" PRIu64 "%[^\n]",
```
anakryiko, line 676: ditto

anakryiko (changes requested):

libbpf-tools/trace_helpers.c
jeez: I don't love them either but these are used in several other source files in this code base. Shouldn't we keep things consistent here?

anakryiko: where do you see it used in libbpf-tools tools?

libbpf-tools/bindsnoop.c
```diff
@@ -168,10 +168,10 @@ static void handle_event(void *ctx, int cpu, void *data, __u32 data_sz)
 		i++;
 	}
 	if (e->ver == 4) {
-		inet_ntop(AF_INET, &e->addr, addr, sizeof(addr));
-	} else {
-		inet_ntop(AF_INET6, &e->addr, addr, sizeof(addr));
-	}
+		inet_ntop(AF_INET, e->addr, addr, sizeof(addr));
+        } else {
```
chenhengqi, line 172: Please use tab for indention.

libbpf-tools/trace_helpers.c
jeez: I meant in the root project, bcc.

libbpf-tools/bindsnoop.c
jeez: will fix in v2, thanks

libbpf-tools/trace_helpers.c
jeez: fixed in v2

anakryiko (approved):
LGTM, thanks.

PR #4782, opened 2023-10-26
f2fsslower:This tool attaches to the entry and return points of speci…

zhangting9756:
add new tool tracing IO operation slower than a given threshold for f2fs file system in Andriod and Linux

### DIFF
--- a/tools/f2fsslower.py
+++ b/tools/f2fsslower.py
@@ -1,0 +1,338 @@
+#!/usr/bin/python
+# SPDX-License-Identifier: <SPDX License Expression>
+# @lint-avoid-python-3-compatibility-imports
+#
+# f2fsslower  Trace slow f2fs operations.
+#              For Linux, uses BCC, eBPF.
+#
+# USAGE: f2fsslower [-h] [-s] [-p PID] [min_ms]
+#
+# This script traces common f2fs file operations: reads, writes, opens, and
+# syncs. It measures the time spent in these operations, and prints details
+# for each that exceeded a threshold.
+#
+# WARNING: This adds low-overhead instrumentation to these f2fs operations,
+# including reads and writes from the file system cache. Such reads and writes
+# can be very frequent (depending on the workload; eg, 1M/sec), at which
+# point the overhead of this tool (even if it prints no "slower" events) can
+# begin to become significant.
+#
+# By default, a minimum millisecond threshold of 10 is used.
+#
+# Copyright (c) 2022, Samsung Electronics.  All rights reserved.
+# Licensed under the Apache License, Version 2.0 (the "License")
+# thanks for Brendan Gregg's ext4slower
+# (https://github.com/iovisor/bcc/blob/master/tools/ext4slower.py) reference.
+#
+# 15-Aug-2022   Ting Zhang   Created this.
+
+from __future__ import print_function
+from bcc import BPF
+import argparse
+from time import strftime
+
+# symbols
+kallsyms = "/proc/kallsyms"
+
+# arguments
+examples = """examples:
+    ./f2fsslower             # trace operations slower than 10 ms (default)
+    ./f2fsslower 1           # trace operations slower than 1 ms
+    ./f2fsslower -s 1        # ... 1 ms, parsable output (csv)
+    ./f2fsslower 0           # trace all operations (warning: verbose)
+    ./f2fsslower -p 185      # trace PID 185 only
+"""
+parser = argparse.ArgumentParser(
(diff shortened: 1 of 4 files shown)


chenhengqi:
cc @brendangregg  @netedwardwu @weibang6liu

tools/f2fsslower.py
```diff
@@ -0,0 +1,338 @@
...
+#
+# 15-Aug-2022   Ting Zhang   Created this.
+
+from __future__ import print_function
+from bcc import BPF
+import argparse
+from time import strftime
+
+# symbols
+kallsyms = "/proc/kallsyms"
```
chenhengqi, line 35: Move close to where we use it.

tools/f2fsslower.py
```diff
@@ -0,0 +1,338 @@
...
+    u32 pid = id >> 32; // PID is higher part
+    valp = entryinfo.lookup(&id);
+    if (valp == 0) {
+        // missed tracing issue or filtered
+        return 0;
+    }
+    // calculate delta
+    u64 ts = bpf_ktime_get_ns();
+    u64 delta_us = (ts - valp->ts) / 1000;
+    entryinfo.delete(&id);
```
chenhengqi, line 184: Do NOT delete the element prematurely. This should be done after perf submit calls.

tools/f2fsslower.py
```diff
@@ -0,0 +1,338 @@
...
+    return trace_return(ctx, TRACE_WRITE);
+}
+int trace_open_return(struct pt_regs *ctx)
+{
+    return trace_return(ctx, TRACE_OPEN);
+}
+int trace_fsync_return(struct pt_regs *ctx)
+{
+    return trace_return(ctx, TRACE_FSYNC);
+}
```
chenhengqi, line 223: Please add an empty line between functions.

tools/f2fsslower.py
```diff
@@ -0,0 +1,338 @@
...
+          event.task.decode('utf-8', 'replace'), event.pid, type, event.size,
+          event.offset / 1024, float(event.delta_us) / 1000,
+          event.file.decode('utf-8', 'replace')))
+
+# initialize BPF
+b = BPF(text=bpf_text)
+
+# Common file functions. See earlier comment about generic_file_read_iter().
+if BPF.get_kprobe_functions(b'f2fs_file_read_iter'):
+    b.attach_kprobe(event="f2fs_file_read_iter", fn_name="trace_read_entry")
```
chenhengqi, line 296: `attach_kretprobe` here we can save an `if` block below.

tools/f2fsslower.py
```diff
@@ -0,0 +1,338 @@
...
+    b["events"].open_perf_buffer(print_event, page_cnt=64)
+
+while 1:
+    try:
+        if PERF_MODE == "USE_BPF_RING_BUF":
+            b.ring_buffer_poll()
+        else:
+            b.perf_buffer_poll()
+    except KeyboardInterrupt:
+        exit()
```
chenhengqi, line 338: No newline below ?

chenhengqi (changes requested):
Please reword the subject line and commit message properly.

tools/f2fsslower.py
zhangting9756: I've modified the code based on review advice and invite you review again,thank you

tools/f2fsslower.py
```diff
@@ -0,0 +1,340 @@
...
+    data.ts_us = ts / 1000;
+    data.offset = valp->offset;
+    bpf_get_current_comm(&data.task, sizeof(data.task));
+    // workaround (rewriter should handle file to d_name in one step):
+    struct dentry *de = NULL;
+    struct qstr qs = {};
+    de = valp->fp->f_path.dentry;
+    qs = de->d_name;
+    if (qs.len == 0)
+        return 0;
```
chenhengqi, line 201: `entryinfo.delete(&id);` before returning.

zhangting9756: Thank you for reviewing again，code had modifed based on comments

chenhengqi: This is not what I asked for. 🤦 

`entryinfo.delete(&id);` can be done only if `valp` is not referenced any more.

zhangting9756: Thanks  a lot, code had modified for resource clean-up before returning

zhangting9756:
@chenhengqi @yonghong-song @weibang6liu will you review this bcc tool and commit？

PR #4784, opened 2023-10-27
Added support for f2fs to fsdist and fsslower tools.

pchengcui:
F2FS (FlashFriendly File System) is an open source flash file system，designed specifically for NAND-based storage devices and currently used primarily on mobile platforms. 

With the idea of CO-RE (Compile Once, Run Everywhere) being proposed and implemented, the portability of BPF has been improved, and the application of BPF on mobile platforms has become more and more widespread. 

Therefore, I have added support for F2FS to fsdist and fsslower. 

Thank you for your review.


chenhengqi (approved):
LGTM. Please also add an install target in Makefile.

PR #4785, opened 2023-10-27
libbpf-tools/gethostlatency: Resolve the path of libc for different namespaces

KentaTada:
gethostlatency.c cannot handle the path of libc inside other mnt namespaces.
If PID is specified, this commit resolves the path of libc from /proc/[PID]/root.


KentaTada:
@chenhengqi 
I want to use your tool for applications inside a container.
Could you review this change if you have some time?

PR #4789, opened 2023-10-31
tools/*.py: fix invalid escape sequences in regexps

hhoffstaette:
With python >=3.8 many tools emit warnings on startup:
```
/usr/sbin/dcstat:91: SyntaxWarning: invalid escape sequence '\d'
  b.attach_kprobe(event_re="^lookup_fast$|^lookup_fast.constprop.*.\d$", fn_name="count_fast")
```
Fix this by ~properly escaping backslashes~ using raw strings for regular expressions.

### DIFF
--- a/tools/cpudist.py
+++ b/tools/cpudist.py
@@ -208,7 +208,7 @@ if debug or args.ebpf:
 max_pid = int(open("/proc/sys/kernel/pid_max").read())
 
 b = BPF(text=bpf_text, cflags=["-DMAX_PID=%d" % max_pid])
-b.attach_kprobe(event_re="^finish_task_switch$|^finish_task_switch\.isra\.\d$",
+b.attach_kprobe(event_re=r'^finish_task_switch$|^finish_task_switch\.isra\.\d$',
                 fn_name="sched_switch")
 
 print("Tracing %s-CPU time... Hit Ctrl-C to end." %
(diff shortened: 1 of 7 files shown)


tools/cpudist.py
```diff
@@ -208,7 +208,7 @@
 max_pid = int(open("/proc/sys/kernel/pid_max").read())
 
 b = BPF(text=bpf_text, cflags=["-DMAX_PID=%d" % max_pid])
-b.attach_kprobe(event_re="^finish_task_switch$|^finish_task_switch\.isra\.\d$",
+b.attach_kprobe(event_re="^finish_task_switch$|^finish_task_switch\\.isra\\.\\d$",
```
chenhengqi, line 211: Looks correct according to https://github.com/python/cpython/pull/99011, though I don't get this warning with Python 3.10 on Ubuntu.

Could we use raw string here, for both Python 2 and Python 3 ?  `\\` looks ugly.

hhoffstaette: I actually considered raw strings first but figured proper escapes were more idiomatic.. 😅
But ok - I will convert the regexps to raw strings if you prefer. Thanks for taking a look!

hhoffstaette: Update pushed.

hhoffstaette:
> I somehow cannot reproduce the warning with my local env (python 3.9). But I verified that the change works for 3.6 and 3.9 in my local env. Also the changed format is what python3 recommended https://docs.python.org/3/library/re.html.

Hm.. I'm on 3.12 already and when I searched for the warning, various results indicated that this happens since 3.8. Might only be since 3.10, but whatever - as you said it's the right thing to do anyway. Thanks for merging. :)

PR #4791, opened 2023-11-05
tools/filelife: Fix error tracking caused by failed unlink

Rtoax:
When we try to delete a file with a normal user after creating a file as root (sudo), the deletion of the file fails, but the kprobe:vfs_unlink is triggered, which leads to incorrect file lifecycle tracking.

Reproducing the issue:

```
  Terminal 1:

    $ sudo ./filelife.py

  Terminal 2:

    $ sudo touch /etc/a.txt
    $ rm /etc/a.txt
    rm: remove write-protected regular empty file '/etc/a.txt'? y
    rm: cannot remove '/etc/a.txt': Permission denied

  Then, terminal 1:

    $ sudo ./filelife.py
    TIME     PID     COMM             AGE(s)  FILE
    12:24:11 4988    rm               7.13    a.txt   <<< wrong statistic
```

However, file 'a.txt' is not removed. This commit solves this problem.

### DIFF
--- a/tools/filelife.py
+++ b/tools/filelife.py
@@ -17,6 +17,7 @@
 # 08-Feb-2015   Brendan Gregg   Created this.
 # 17-Feb-2016   Allan McAleavy updated for BPF_PERF_OUTPUT
 # 13-Nov-2022   Rong Tao        Check btf struct field for CO-RE and add vfs_open()
+# 05-Nov-2023   Rong Tao        Support unlink failed
 
 from __future__ import print_function
 from bcc import BPF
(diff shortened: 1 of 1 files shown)


tools/filelife.py
```diff
@@ -95,7 +99,8 @@
 TRACE_UNLINK_FUNC
 {
     struct data_t data = {};
-    u32 pid = bpf_get_current_pid_tgid() >> 32;
+    u64 pid_tgid = bpf_get_current_pid_tgid();
```
yonghong-song, line 102: Let us 'u32 tid = bpf_get_current_pid_tgid()' instead. Just search 'tid' in tools directory and you can find in most (if not all) cases, 'tid' is used as the key in the hash table between func entry and func return. Using 'tid' can be consistent with other tools. Otherwise the patch looks good. Indeed, we should check vfs_unlink return value to report unlink unless unlinking is really happening.

Rtoax: Fix it , thanks, Song! and fix the following warning by ` birth.delete((struct dentry **)&data->dentry);`:

```
$ sudo ./filelife.py 
/virtual/main.c:113:18: warning: incompatible pointer types passing 'void **' to parameter of type 'struct dentry **' [-Wincompatible-pointer-types]
  113 |     birth.delete(&data->dentry);
      |                  ^~~~~~~~~~~~~
1 warning generated.
```

PR #4793, opened 2023-11-05
tools: Add dirwatch

Rtoax:
We often need to monitor which processes in a directory have created and modified files, especially in the /etc directory. For example, sometimes the /etc/fstab is deleted by mistake, which causes the system to fail to reboot.

The dirwatch tool can monitor the creation and deletion of all files in a single directory, and can even capture the parent process when the verbose (-V) message is displayed.

Terminal 1:

```
    $ sudo ./dirwatch.py -D /etc/
```

Terminal 2:

```
    $ sudo mkdir -p /etc/a/b/c/d/e/
    $ sudo touch /etc/a/b/c/d/e/readme
    $ sudo rm -rf /etc/a
```

Then, Terminal 1 shows:

```
    $ sudo ./dirwatch.py -D /etc/
    Tracing file remove ... Hit Ctrl-C to end
    TIME     PID      COMM         OPERATE  INODE        FILEPATH
    15:40:35 28094    mkdir        MKDIR    2015         /etc//a
    15:40:35 28094    mkdir        MKDIR    67426298     /etc//a/b
    15:40:35 28094    mkdir        MKDIR    134492307    /etc//a/b/c
    15:40:35 28094    mkdir        MKDIR    201858033    /etc//a/b/c/d
    15:40:35 28094    mkdir        MKDIR    2058         /etc//a/b/c/d/e
    15:40:46 28100    touch        CREATE   2059         /etc//a/b/c/d/e/readme
    15:40:57 28107    rm           UNLINK   2059         /etc//a/b/c/d/e/readme
    15:40:57 28107    rm           RMDIR    2058         /etc//a/b/c/d/e
    15:40:57 28107    rm           RMDIR    201858033    /etc//a/b/c/d
    15:40:57 28107    rm           RMDIR    134492307    /etc//a/b/c
    15:40:57 28107    rm           RMDIR    67426298     /etc//a/b
    15:40:57 28107    rm           RMDIR    2015         /etc//a
```

TODO: Not support symbol link yet.


yonghong-song:
could you check whether inotify() syscall can do similar thing or not?

Rtoax:
> could you check whether inotify() syscall can do similar thing or not?

Yes, `inotify` also can do the same thing.

brendangregg:
> We often need to monitor which processes in a directory have created and modified files, especially in the /etc directory

This sounds like outdated Unix security practices from the 1980s. I'd also want to see how it compares to other file notification frameworks, in detail.

This tool passes all VFS events to user space (including two strings, comm and pcomm) and does filtering there. At least it filters uint64 inodes, but this whole approach is inefficient.

The reason I was using BPF, and not straight-up perf to do these tools, was efficiency. A tool like this can just be done in perf. It's a tool using best practices from yesteryear. Thanks but I'd reject this.

PR #4795, opened 2023-11-06
avxcputime: Summarize AVX-512 cputime per task as a histogram

yezhiyong30:
The AVX-512 instruction set in x86 can accelerate FPU execution but also causes core turbo frequency drop that impacts sibling CPUs. This can affect performance of other workloads running on the sibling CPU. Detecting AVX-512 usage is thus critical in cloud environments for placement.

This tool summarizes AVX-512 cputime as a histogram, showing the amount of CPU time consumed by AVX-512 per-task. This provides valuable insights - it can identify processes and CPUs executing AVX-512 to avoid scheduling sensitive jobs together. The cputime distribution can also help debug AVX-512 performance issues.

This program is based on this Linux kernel patch that tracks per-task AVX-512 usage:

https://github.com/torvalds/linux/commit/2f7726f955572e587d5f50fbe9b2deed5334bd90

It uses BPF to measure the time between a task's AVX-512 timestamps to calculate the CPU time in AVX-512 mode. The aggregated cputime distribution is printed for analysis.


yezhiyong30:
Can you help me review it? I will improve it with any comments.
@brendangregg @goldshtn @devidasjadhav @drzaeus77 @yonghong-song @4ast

brendangregg (commented):
I'm not familiar with tracing x86_fpu_regs_deactivated, so I'll assume you've dug into that and understand how it all works.

Note that there are AVX-512 PMCs but I don't think they are easily findable. https://github.com/intel/PerfSpect recently added support for AVX-512 metrics, although I don't think it can emit a per-PID breakdown or the context-switch event runtimes. (It's also unlikly to work in most cloud environments due to a lack of the PMU.)

I think this tool should be a good starting point. Future enhancements could add more detail about how the system is affected (power, clockspeed, CPI) during AVX-512 runs, likely requiring PMCs.

yezhiyong30:
Hi @brendangregg ,

I've updated the PR. Changes are:

- Rename `avxcputime` to `avxtime`.
-  Added explanation of AVX-512 time measurement method.
> Within the specified time period, it tracks the time spent from being scheduled
on the CPU until the FPU registers are deactivated for each task that uses AVX-512.

Welcome to review it again. Any comments can help me improve this patch.

PR #4797, opened 2023-11-07
tools/memleak: Add --sort option to report ordered in size or count

Bojun-Seo:
Add `--sort` option to report ordered in size or count. Available sort key list: size, count
`size` is a default value.

Usage:
./memleak.py --sort count
./memleak.py --sort size

### DIFF
--- a/tools/memleak.py
+++ b/tools/memleak.py
@@ -39,6 +39,12 @@ def run_command_get_pid(command):
         p = subprocess.Popen(command.split())
         return p.pid
 
+sort_keys = ["size", "count"]
+alloc_sort_map = {sort_keys[0]: lambda a: a.size,
+                  sort_keys[1]: lambda a: a.count};
+combined_sort_map = {sort_keys[0]: lambda a: -a[1].total_size,
+                     sort_keys[1]: lambda a: -a[1].number_of_allocs};
+
 examples = """
 EXAMPLES:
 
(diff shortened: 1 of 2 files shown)


tools/memleak.py
```diff
@@ -119,6 +121,12 @@ def run_command_get_pid(command):
 min_size = args.min_size
 max_size = args.max_size
 obj = args.obj
+sort_key = args.sort
+
```
ekyooo, line 125: How about defining and using this dictionary object?
sort_keys = {"size": lambda a: a.size,
 "count: lambda a: a.count}

tools/memleak.py
```diff
@@ -119,6 +121,12 @@ def run_command_get_pid(command):
...
+if sort_key != "size" and sort_key != "count":
```
ekyooo, line 126: if sort_key not in sort_keys:

tools/memleak.py
```diff
@@ -522,17 +530,33 @@ def print_outstanding():
...
+        to_show = None
+        if sort_key == "size":
+                to_show = sorted(alloc_info.values(),
+                                 key=lambda a: a.size)[-top_stacks:]
+        elif sort_key == "count":
+                to_show = sorted(alloc_info.values(),
+                                 key=lambda a: a.count)[-top_stacks:]
+        else:
+                print("Fatal: reached unreachable code");
+                exit()
```
ekyooo, line 542: to_show = sorted(alloc_info.values(), key = sort_keys[sort_key]) [-top_stacks:]

tools/memleak.py
Bojun-Seo: I changed the code as you suggested. Thanks!

tools/memleak.py
Bojun-Seo: Looks much better, I changed my code.

tools/memleak.py
Bojun-Seo: Awesome! The code becomes one line. Thanks!
I applied your suggestion.

PR #4803, opened 2023-11-15
test: Preserving relative path for tools directory

dubeyabhishek:
The absolute path set for TOOLS_DIR may not always be defined for manual runs, adding back the relative path removed in commit 9a45c3c.


yonghong-song:
Agree. This makes local testing easier.

PR #4810, opened 2023-11-24
libbpf-tools/biotop：Add PID filter support

zhangting9756:
-p option for tracking a specific pid in tools/biotop which filters target PID to avoid message flooding.
So, adding -p option to filter pid is necessary.

### DIFF
--- a/libbpf-tools/biotop.bpf.c
+++ b/libbpf-tools/biotop.bpf.c
@@ -9,6 +9,8 @@
 #include "maps.bpf.h"
 #include "core_fixes.bpf.h"
 
+const volatile pid_t target_pid = 0;
+
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
 	__uint(max_entries, 10240);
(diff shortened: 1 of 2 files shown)


zhangting9756:
@pchengcui

pchengcui:
@chenhengqi @yonghong-song can you review this PR ? thanks.

yonghong-song:
I think this pid filtering thing may not work in such cases, e.g., blk_account_io_start() is often not called in sycall process context. When an IO operation is initialized by user space, the kernel in general will do blk_queue_bio() to queue the IO and returns to the user space. Later blk_account_io_start() will be done by per-cpu IO thread (maybe something else). So I am not sure this pid based filtering works or not for blk_account_io_start(). Similarly, the same case for other functions.

pchengcui:
> I think this pid filtering thing may not work in such cases, e.g., blk_account_io_start() is often not called in sycall process context. When an IO operation is initialized by user space, the kernel in general will do blk_queue_bio() to queue the IO and returns to the user space. Later blk_account_io_start() will be done by per-cpu IO thread (maybe something else). So I am not sure this pid based filtering works or not for blk_account_io_start(). Similarly, the same case for other functions.

Thanks for the reply, PID may be lost in this case. However, this situation is in the minority, and I think it is acceptable. This situation may also exist in the corresponding /tools/biotop.py.

yonghong-song:
Okay, will merge the patch since we have biotop precedence.

PR #4817, opened 2023-12-01
examples/tracing: lbr.py trace cond branches executed by syscalls

lorenzocarrozzo:
This script uses the Last Branch Record (LBR) buffer to trace conditional branches executed by syscalls. 
The script takes a snapshot of the LBR and displays the to/ from address to the user.
This can be useful if you want are trying to trace the execution path of a syscall which is returning an error.
If this script is not too niche I would be happy to submit it as a tool as well.

### DIFF
--- a/examples/tracing/lbr.py
+++ b/examples/tracing/lbr.py
@@ -1,0 +1,281 @@
+#!/usr/bin/python
+#
+# lbr.py
+#
+# Trace conditional branches executed by syscalls using the Last Branch Record
+# Buffer (LBR)
+#
+# REQUIRES:
+#   Linux 5.16+ (bpf_get_branch_snapshot support)
+#
+# Copyright (c) 2023 Bytedance Inc.
+#
+# Author(s):
+#   Lorenzo Carrozzo <lorenzocarrozzo@bytedance.com>
+
+from __future__ import absolute_import, print_function, unicode_literals
+from bcc import BPF
+from bcc import PerfType, PerfHWConfig, PerfEventSampleFormat, Perf
+import argparse
+from sys import exit
+from pathlib import Path
+from subprocess import Popen, PIPE
+
+# Number of LBR entries and output tags
+lbr_cnt = 32
+num_entries_tag = 'lbr_total_entries:'
+entry_tag = 'lbr_entry:'
+
+# BPF program text
+bpf_text = """
+#include <uapi/linux/perf_event.h>
+#include <uapi/linux/ptrace.h>
+
+// Define arguments passed to tracepoint
+struct params {
+        short common_type;
+        unsigned char common_flags;
+        unsigned char common_preempt_count;
+        int common_pid;
+        int __syscall_nr;
+        long ret;
+};
+
+struct perf_branch_entry_buf {
+    struct perf_branch_entry entries[ENTRY_CNT];
+};
+
+BPF_PERCPU_ARRAY(branch_entry, struct perf_branch_entry_buf, 1);
+
+// Function to use with tracepoint
+int disp_snapshot_tp(struct params *p) {
+    unsigned buf_size = sizeof(struct perf_branch_entry_buf), idx = 0;
+    struct perf_branch_entry_buf *buf;
+
+    buf = branch_entry.lookup(&idx);
+    if (!buf)
+        return 0;
+
+    int total_entries = bpf_get_branch_snapshot(buf, buf_size, 0);
+    total_entries /= sizeof(struct perf_branch_entry);
+
+    if (true T_R_COND P_COND) {
+        bpf_trace_printk("NUM_ENTRIES%d", total_entries);
+
+        for (int i = 0; i < ENTRY_CNT; i++) {
+            bpf_trace_printk("ENTRY%pS --> %pS", buf->entries[i].from,
+                             buf->entries[i].to);
+        }
+    }
+
+    return 0;
+}
+
+// Function to use with kretprobe
+int disp_snapshot_krp(struct pt_regs *p) {
+    unsigned buf_size = sizeof(struct perf_branch_entry_buf), idx = 0;
+    struct perf_branch_entry_buf *buf;
+
+    buf = branch_entry.lookup(&idx);
+    if (!buf)
+        return 0;
+
+    int total_entries = bpf_get_branch_snapshot(buf, buf_size, 0);
+    total_entries /= sizeof(struct perf_branch_entry);
+
+    if (true K_R_COND P_COND) {
+        bpf_trace_printk("NUM_ENTRIES%d", total_entries);
+
+        for (int i = 0; i < ENTRY_CNT; i++) {
+            bpf_trace_printk("ENTRY%pS --> %pS", buf->entries[i].from,
+                             buf->entries[i].to);
+        }
+    }
+
+    return 0;
+}
+"""
+
+# Parse arguments
+examples = """
+examples:
+    ./lbr -t <syscall>       # the syscall to attach the exit tracepoint to
+    ./lbr -k <syscall>       # the syscall to attach a kretprobe to
+    ./lbr -r <value>         # filter by syscall's return value
+    ./lbr -p <pid>           # filter by program pid
+    ./lbr -b <path/to/bin>   # kernel to search addresses in using addr2line
+    ./lbr -d                 # show debug strings
+"""
+parser = argparse.ArgumentParser(
+    description="Trace conditional branches using LBR.",
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    epilog=examples)
+parser.add_argument("-t", "--tracepoint", type=str, metavar="SYSCALL",
+                    help="the syscall to attach the exit tracepoint to")
+parser.add_argument("-k", "--kretprobe", type=str, metavar="SYSCALL",
+                    help="the syscall to attach a kretprobe to", )
+parser.add_argument("-r", "--ret_value", type=int, metavar="VALUE",
+                    help="filter by syscall's return value")
+parser.add_argument("-p", "--pid", type=int,
+                    help="filter by pid")
+parser.add_argument("-e", "--extend", action="store_true",
+                    help="extend output width so entry addresses are on on \
+                    one line")
+parser.add_argument("-b", "--bin", type=str,
+                    help="the binary to search address in")
+parser.add_argument("-d", "--debug", action="store_true",
+                    help="print out bpf program text")
+args = parser.parse_args()
+
+# Check that tracepoint or kretprobe is given
+if args.tracepoint is None and args.kretprobe is None:
+    print('Error tracepoint or kretprobe is required')
+    parser.print_help()
+    exit(1)
+elif args.tracepoint == args.kretprobe:
+    print('Warning it is not recommend to attach to a syscall`s tracepoint \n \
+          and kretprobe at the same time!!!')
+
+# Check binary is valid if provided
+if args.bin is not None and not Path(args.bin).is_file():
+    print('Error binary path is invalid')
+    parser.print_help()
+    exit(1)
+
+# Replace conditions based on arguments
+if args.ret_value is not None:
+    bpf_text = bpf_text.replace('T_R_COND', f'&& p->ret == {args.ret_value}')
+    bpf_text = bpf_text.replace('K_R_COND',
+                                f'&& PT_REGS_RC(p) == {args.ret_value}')
+
+if args.pid is not None:
+    bpf_text = bpf_text.replace('P_COND',
+                                f'&& (bpf_get_current_pid_tgid() >> 32) == \
+                                {args.pid}')
+
+# Remove any unused tags not used
+bpf_text = bpf_text.replace('T_R_COND', '')
+bpf_text = bpf_text.replace('K_R_COND', '')
+bpf_text = bpf_text.replace('P_COND', '')
+
+# Replace other globals
+bpf_text = bpf_text.replace('ENTRY_CNT', str(lbr_cnt))
+bpf_text = bpf_text.replace('NUM_ENTRIES', num_entries_tag)
+bpf_text = bpf_text.replace('ENTRY', entry_tag)
+
+# Print out completed bpf program text
+if args.debug:
+    print(bpf_text)
+
+# Load bpf program
+bpf_prog = BPF(text=bpf_text)
+
+# Open perf event
+# Filters are defined in perf_event.h
+# PERF_SAMPLE_BRANCH_KERNEL | PERF_SAMPLE_BRANCH_USER | PERF_SAMPLE_BRANCH_COND
+attr = Perf.perf_event_attr()
+attr.config = PerfHWConfig.CPU_CYCLES
+attr.type = PerfType.HARDWARE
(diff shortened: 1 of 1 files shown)


yonghong-song:
It seems a useful example. But I tried and already found a few errors:
```
[~/tmp]$ sudo python3.6 lbr.py -t clone                                                                                                                
/virtual/main.c:21:31: warning: only %d %u %x %ld %lu %lx %lld %llu %llx %p %s conversion specifiers allowed                                                              
   21 |             bpf_trace_printk("%pS --> %pS", entries[i].from, entries[i].to);                                                                                      
      |                               ^                                                                                                                                   
/virtual/main.c:34:31: warning: only %d %u %x %ld %lu %lx %lld %llu %llx %p %s conversion specifiers allowed                                                              
   34 |             bpf_trace_printk("%pS --> %pS", entries[i].from, entries[i].to);                                                                                      
      |                               ^                                                                                                                                   
2 warnings generated.                                                                                                                                                     
Warning: Setting field pid on perf_event_attr that isn't part of the ctype - pid won't make it to perf_event_open                                                         
                                                                                                                                                                          
                                                                                                                                                                          
Tracing logical branches... Hit Ctrl-C to end.                                                                                                                            
                                                                                                                                                                          
i  | T/F  | Addresses                            |                                                                                                                        
-------------------------------------------------|                                                                                                                        
0  | From | trace_call_bpf+0x76/0x240            |                                                                                                                        
   |---------------------------------------------|                                                                                                                        
   | To   | trace_call_bpf+0xb6/0x240            |                                                                                                                        
-------------------------------------------------| 
...
  File "lbr.py", line 238, in <module>
    print_snapshot()
  File "lbr.py", line 194, in print_snapshot
    to_addrs.append(addrs[1])
IndexError: list index out of range
[~/tmp]$ sudo python3.6 lbr.py -k clone                            
/virtual/main.c:21:31: warning: only %d %u %x %ld %lu %lx %lld %llu %llx %p %s conversion specifiers allowed
   21 |             bpf_trace_printk("%pS --> %pS", entries[i].from, entries[i].to);
      |                               ^
/virtual/main.c:34:31: warning: only %d %u %x %ld %lu %lx %lld %llu %llx %p %s conversion specifiers allowed
   34 |             bpf_trace_printk("%pS --> %pS", entries[i].from, entries[i].to);
      |                               ^
2 warnings generated.
Warning: Setting field pid on perf_event_attr that isn't part of the ctype - pid won't make it to perf_event_open


Tracing logical branches... Hit Ctrl-C to end.

  File "lbr.py", line 238, in <module>
    print_snapshot()
  File "lbr.py", line 186, in print_snapshot
    total_entries = int(msg)
ValueError: invalid literal for int() with base 10: b'enqueue_task_fair+0x109a/0x1670 --> enqueue_task_fair+0x4ac/0x1670'
[~/tmp]$
```

I didn't try other commands. Could you take a look?

lorenzocarrozzo:
> It seems a useful example. But I tried and already found a few errors:
> 
> ```
> [~/tmp]$ sudo python3.6 lbr.py -t clone                                                                                                                
> /virtual/main.c:21:31: warning: only %d %u %x %ld %lu %lx %lld %llu %llx %p %s conversion specifiers allowed                                                              
>    21 |             bpf_trace_printk("%pS --> %pS", entries[i].from, entries[i].to);                                                                                      
>       |                               ^                                                                                                                                   
> /virtual/main.c:34:31: warning: only %d %u %x %ld %lu %lx %lld %llu %llx %p %s conversion specifiers allowed                                                              
>    34 |             bpf_trace_printk("%pS --> %pS", entries[i].from, entries[i].to);                                                                                      
>       |                               ^                                                                                                                                   
> 2 warnings generated.                                                                                                                                                     
> Warning: Setting field pid on perf_event_attr that isn't part of the ctype - pid won't make it to perf_event_open                                                         
>                                                                                                                                                                           
>                                                                                                                                                                           
> Tracing logical branches... Hit Ctrl-C to end.                                                                                                                            
>                                                                                                                                                                           
> i  | T/F  | Addresses                            |                                                                                                                        
> -------------------------------------------------|                                                                                                                        
> 0  | From | trace_call_bpf+0x76/0x240            |                                                                                                                        
>    |---------------------------------------------|                                                                                                                        
>    | To   | trace_call_bpf+0xb6/0x240            |                                                                                                                        
> -------------------------------------------------| 
> ...
> Traceback (most recent call last):
>   File "lbr.py", line 238, in <module>
>     print_snapshot()
>   File "lbr.py", line 194, in print_snapshot
>     to_addrs.append(addrs[1])
> IndexError: list index out of range
> [~/tmp]$ sudo python3.6 lbr.py -k clone                            
> /virtual/main.c:21:31: warning: only %d %u %x %ld %lu %lx %lld %llu %llx %p %s conversion specifiers allowed
>    21 |             bpf_trace_printk("%pS --> %pS", entries[i].from, entries[i].to);
>       |                               ^
> /virtual/main.c:34:31: warning: only %d %u %x %ld %lu %lx %lld %llu %llx %p %s conversion specifiers allowed
>    34 |             bpf_trace_printk("%pS --> %pS", entries[i].from, entries[i].to);
>       |                               ^
> 2 warnings generated.
> Warning: Setting field pid on perf_event_attr that isn't part of the ctype - pid won't make it to perf_event_open
> 
> 
> Tracing logical branches... Hit Ctrl-C to end.
> 
> Traceback (most recent call last):
>   File "lbr.py", line 238, in <module>
>     print_snapshot()
>   File "lbr.py", line 186, in print_snapshot
>     total_entries = int(msg)
> ValueError: invalid literal for int() with base 10: b'enqueue_task_fair+0x109a/0x1670 --> enqueue_task_fair+0x4ac/0x1670'
> [~/tmp]$
> ```
> 
> I didn't try other commands. Could you take a look?

Apologies these crashes were to do with me failing to take into account the concurrency of the bpf trace_pipe.
I’ve added some tags to the output of the lbr script to allow it to distinguish between the origin of the messages it reads from the bpf trace_pipe.

lorenzocarrozzo:
As for the warning about using `%pS` I was wondering if you know if there is a specific reason as to why it is not included in BCC’s allowed conversion specifiers?
I can change it to %p to get rid of the warning but this ultimately makes the tools output less useful.

lorenzocarrozzo:
Also the warning about the pid field is due to me being unable to use Perf.perf_event_open() to setup my perf event as its arguments are too strict.
Specifically I need access to the attr.sample_type and attr.branch_sample_type.
Would it be best if I suggest adding these options in another patch?

yonghong-song:
> As for the warning about using `%pS` I was wondering if you know if there is a specific reason as to why it is not included in BCC’s allowed conversion specifiers? I can change it to %p to get rid of the warning but this ultimately makes the tools output less useful.

The supported kernel bpf_trace_printk format can be found in the following kernel file
   https://github.com/torvalds/linux/blob/master/kernel/bpf/helpers.c#L815
The function is bpf_bprintf_prepare().

Could you help improve bcc to cover more supported strings? The related file is b_frontend_action.cc, function checkFormatSpecifiers.

Currently, warning does not impact functionality, but we still want to warn issues for incorrect format other users won't have idea why a particular bpf_trace_printk won't print anything.

yonghong-song:
> Also the warning about the pid field is due to me being unable to use Perf.perf_event_open() to setup my perf event as its arguments are too strict. Specifically I need access to the attr.sample_type and attr.branch_sample_type. Would it be best if I suggest adding these options in another patch?

Yes, please. A separate patch to add sample_type/branch_sample_type in perf_event_open() API looks okay to me.

lorenzocarrozzo:
I have submitted two new patches which should fix the warnings.
When they are merged I will update the script to use the new API.

lorenzocarrozzo:
I have updated the script to use Perf.perf_custom_event_open()

examples/tracing/lbr.py
```diff
@@ -0,0 +1,267 @@
...
+from __future__ import absolute_import, print_function, unicode_literals
+from bcc import BPF, utils
+from bcc import PerfType, PerfHWConfig, PerfEventSampleFormat, Perf
+import argparse
+from sys import exit
+from pathlib import Path
+from subprocess import Popen, PIPE
+
+# Number of LBR entries and output tags
+lbr_cnt = 20  # Should be 32 but limited by bpf stack size
```
yonghong-song, line 25: I think we can create a percpu map with one element and the element size can be big enough to support 32 entries. This will make example more useful to people wanting to get lbr record in their bpf programs.

lorenzocarrozzo: Ive tried doing something similar in the past like so.
`struct lbr_entries {
    struct perf_branch_entry entries[ENTRY_CNT];
};

BPF_PERCPU_ARRAY(pc_arr, struct lbr_entries, 1);

// Function to use with tracepoint
int disp_snapshot_tp(struct params *p) {
    int zero = 0;
    if (pc_arr.lookup(&zero) == NULL) {
        return 1;
    }
    struct lbr_entries *l_entries = pc_arr.lookup(&zero);
    struct perf_branch_entry *entries = l_entries->entries;

    int total_entries = bpf_get_branch_snapshot(entries, sizeof(*entries), 0);
    total_entries /= sizeof(struct perf_branch_entry);

    if (true T_R_COND P_COND) {

        bpf_trace_printk("NUM_ENTRIES%d", total_entries);

        for (int i = 0; i < ENTRY_CNT; i++) {
            bpf_trace_printk("ENTRY%pS --> %pS", entries[i].from,
                             entries[i].to);
        }
    }

    return 0;
}`

But I get the following error which Im not sure how to interpret.
Screenshot 2024-01-05 at 10 44 42

lorenzocarrozzo: In the kernel self test gets around the limited stack size issue by initialising the array of entries as a global variable.
https://github.com/torvalds/linux/blob/master/tools/testing/selftests/bpf/progs/get_branch_snapshot.c

Is there a way of replicating this method in BCC as the per cpu array does not seem to work?

yonghong-song: The following code works for me:

```
diff --git a/examples/tracing/lbr.py b/examples/tracing/lbr.py
index 3d77051b..633d3bef 100644
--- a/examples/tracing/lbr.py
+++ b/examples/tracing/lbr.py
@@ -22,7 +22,7 @@ from pathlib import Path
 from subprocess import Popen, PIPE
 
 # Number of LBR entries and output tags
-lbr_cnt = 20  # Should be 32 but limited by bpf stack size
+lbr_cnt = 32
 num_entries_tag = 'lbr_total_entries:'
 entry_tag = 'lbr_entry:'
 
@@ -41,11 +41,22 @@ struct params {
         long ret;
 };
 
+struct perf_branch_entry_buf {
+    struct perf_branch_entry entries[ENTRY_CNT];
+};
+
+BPF_PERCPU_ARRAY(branch_entry, struct perf_branch_entry_buf, 1);
+
 // Function to use with tracepoint
 int disp_snapshot_tp(struct params *p) {
-    struct perf_branch_entry entries[ENTRY_CNT] = {};
+    unsigned buf_size = sizeof(struct perf_branch_entry_buf), idx = 0;
+    struct perf_branch_entry_buf *buf;
 
-    int total_entries = bpf_get_branch_snapshot(entries, sizeof(entries), 0);
+    buf = branch_entry.lookup(&idx);
+    if (!buf)
+        return 0;
+
+    int total_entries = bpf_get_branch_snapshot(buf, buf_size, 0);
     total_entries /= sizeof(struct perf_branch_entry);
 
     if (true T_R_COND P_COND) {
@@ -53,8 +64,8 @@ int disp_snapshot_tp(struct params *p) {
         bpf_trace_printk("NUM_ENTRIES%d", total_entries);
 
         for (int i = 0; i < ENTRY_CNT; i++) {
-            bpf_trace_printk("ENTRY%pS --> %pS", entries[i].from,
-                             entries[i].to);
+            bpf_trace_printk("ENTRY%pS --> %pS", buf->entries[i].from,
+                             buf->entries[i].to);
         }
     }
 
@@ -63,9 +74,14 @@ int disp_snapshot_tp(struct params *p) {
 
 // Function to use with kretprobe
 int disp_snapshot_krp(struct pt_regs *p) {
-    struct perf_branch_entry entries[ENTRY_CNT] = {};
+    unsigned buf_size = sizeof(struct perf_branch_entry_buf), idx = 0;
+    struct perf_branch_entry_buf *buf;
+
+    buf = branch_entry.lookup(&idx);
+    if (!buf)
+        return 0;
 
-    int total_entries = bpf_get_branch_snapshot(entries, sizeof(entries), 0);
+    int total_entries = bpf_get_branch_snapshot(buf, buf_size, 0);
     total_entries /= sizeof(struct perf_branch_entry);
 
     if (true K_R_COND P_COND) {
@@ -73,8 +89,8 @@ int disp_snapshot_krp(struct pt_regs *p) {
         bpf_trace_printk("NUM_ENTRIES%d", total_entries);
 
         for (int i = 0; i < ENTRY_CNT; i++) {
-            bpf_trace_printk("ENTRY%pS --> %pS", entries[i].from,
-                             entries[i].to);
+            bpf_trace_printk("ENTRY%pS --> %pS", buf->entries[i].from,
+                             buf->entries[i].to);
         }
     }
```

To trace nanosleep syscall,
```
...
i  | T/F  | Addresses                            |                                                                                                            
-------------------------------------------------|                                                                                                            
0  | From | trace_call_bpf+0x76/0x240            |                                                                                                            
   |---------------------------------------------|                                                                                                            
   | To   | trace_call_bpf+0xb6/0x240            |                                                                                                            
-------------------------------------------------|                                                                                                            
1  | From | perf_trace_buf_alloc+0x38/0xb0       |                                                                                                            
   |---------------------------------------------|                                                                                                            
   | To   | perf_trace_buf_alloc+0x53/0xb0       |                                                                                                            
-------------------------------------------------|                                                                                                            
2  | From | perf_syscall_exit+0x3b/0x160         |                                                                                                            
   |---------------------------------------------|                                                                                                            
   | To   | perf_syscall_exit+0x64/0x160         | 
...
-------------------------------------------------|                             
13 | From | finish_task_switch+0x1d1/0x2d0       |                             
   |---------------------------------------------|                             
   | To   | finish_task_switch+0x7d/0x2d0        |                             
-------------------------------------------------|                             
14 | From | perf_pmu_sched_task+0xfd/0x130       |                             
   |---------------------------------------------|                             
   | To   | perf_pmu_sched_task+0x102/0x130      |                             
-------------------------------------------------|                             
15 | From | 0x0                                  |                             
   |---------------------------------------------|                             
   | To   | 0x0                                  |                             
-------------------------------------------------|                             
16 | From | 0x0                                  |                             
   |---------------------------------------------|                             
   | To   | 0x0                                  |                             
-------------------------------------------------|     
...
-------------------------------------------------|                             
30 | From | 0x0                                  |                             
   |---------------------------------------------|                             
   | To   | 0x0                                  |                             
-------------------------------------------------|                             
31 | From | 0x0                                  |                             
   |---------------------------------------------|                             
   | To   | 0x0                                  |                             
-------------------------------------------------|  
```

If the above does not work, then your kernel probably too old.

lorenzocarrozzo: While the changes I had tried weren't fully correct. It seems like the error was generated as I was checking that the that the first item in the percpu array was not null rather than checking that the pointer was not null.

I have added the correct changes now to the patch.

PR #4819, opened 2023-12-05
Fix unaligned memory accesses within the perf ring buffer

captain5050:
As with exitsnoop fixed in:
https://github.com/iovisor/bcc/pull/4760/

Fix other instances of unaligned memory accesses by memcpy-ing the event structs.


yonghong-song:
LGTM. Thanks!

PR #4836, opened 2023-12-11
libbpf-tools: Fix llx printf modifiers in futexctn

captain5050:
Fix llx printf modifiers by using inttypes.h's PRIx64.


yonghong-song:
In libbpf-tools/javagc.c, we already have
```
javagc.c:       while (fscanf(f, "%zx-%zx %s %zx %*s %*d%[^\n]\n",
```
So '%zx' should work in your case. To preserve the code consistency, could you use '%zx' style instead?

yonghong-song:
BTW, this patch has conflict with latest futexctn.c. You need to rebase your patch on the top of latest master branch.

Issue #4842, opened 2023-12-14
Debian build broken because of changelog divergence with last tag


ader1990:
Hello,

Because of commit https://github.com/iovisor/bcc/commit/7cdf90e9e040a1eda9ae24405f4b2dfb389d0ac1 which adds a 
changelog entry for version 0.29.0 and the existence of tag 0.29.1 in the repository, at this moment, the Debian .deb build fails.

Build error log:

```
#10 28.62 + tar -A -f bcc_0.29.1.orig.tar bcc_libbpf_0.29.1.orig.tar
#10 28.63 + gzip bcc_0.29.1.orig.tar
#10 29.52 + tar xf bcc_0.29.1.orig.tar.gz
#10 29.71 + cd bcc
#10 29.71 + debuild=debuild
#10 29.71 + [[ release = \t\e\s\t ]]
#10 29.71 + [[ release = \n\i\g\h\t\l\y ]]
#10 29.71 + DEB_BUILD_OPTIONS='nocheck parallel=1'
#10 29.71 + debuild -us -uc
#10 29.80 This package has a Debian revision number but there does not seem to be
#10 29.80 an appropriate original tar file or .orig directory in the parent directory;
#10 29.80 (expected one of bcc_0.29.0.orig.tar.gz, bcc_0.29.0.orig.tar.bz2,

```

One option is to add a changelog entry in the debian/changelog file for the 0.29.1 version.

Thank you.

PR #4845, opened 2023-12-15
Fix debian changelog and Python3 library path

ader1990:
When the tag 0.29.1 was added to the repo, the debian changelog file needed one entry as well, otherwise the .deb build fails.
Also, in github actions environment, the Python3 library path is usr/lib/python3*.

Fixes: https://github.com/iovisor/bcc/issues/4842

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,10 @@
+bcc (0.29.1-1) unstable; urgency=low
+
+  * Fix Ubuntu 22.04 deb build
+  * Fix Ubuntu 22.04 Docker image build
+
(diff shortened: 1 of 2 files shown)


debian/changelog
```diff
@@ -1,3 +1,10 @@
+bcc (0.29.1-1) unstable; urgency=low
+
+  * Fix Ubuntu 22.04 deb build
+  * Fix Ubuntu 220.04 Docker image build
```
chenhengqi, line 4: 220.04 => 22.04 ?

PR #3263, opened 2021-02-05
Add install target to libbpf-tools

olsajiri:
We plan to put those tools in separate rpm, so we need a way to install them.

Adding install target with standard DESTDIR and prefix make variables.

PR #4848, opened 2023-12-17
libbpf-tools: Allow to install apps with prefixed names

jirutka:
Fedora adds prefix bpf- to the tool names, mainly to distinguish them from the old tools written in Python. Alpine Linux is going to do the same, but for a slightly different reason - there are dozens of these tools, so it's a good idea to add them a common prefix, to make it easier to recognize them among all other commands.

See-Also: https://github.com/iovisor/bcc/pull/3263

See-Also: https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/35688

### DIFF
--- a/libbpf-tools/Makefile
+++ b/libbpf-tools/Makefile
@@ -13,6 +13,7 @@ CFLAGS := -g -O2 -Wall
 BPFCFLAGS := -g -O2 -Wall
 INSTALL ?= install
 prefix ?= /usr/local
+bindir := $(prefix)/bin
 ARCH ?= $(shell uname -m | sed 's/x86_64/x86/' | sed 's/aarch64/arm64/' \
 			 | sed 's/ppc64le/powerpc/' | sed 's/mips.*/mips/' \
 			 | sed 's/riscv64/riscv/' | sed 's/loongarch.*/loongarch/' \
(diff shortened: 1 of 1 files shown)


libbpf-tools/Makefile
```diff
@@ -222,21 +223,21 @@ $(LIBBPF_OBJ): $(wildcard $(LIBBPF_SRC)/*.[ch]) | $(OUTPUT)/libbpf
 
 $(FSSLOWER_ALIASES): fsslower
 	$(call msg,SYMLINK,$@)
-	$(Q)ln -f -s $^ $@
+	$(Q)ln -f -s $(APP_PREFIX)$^ $@
```
chenhengqi, line 226: What is `APP_PREFIX` ?

jirutka: This is a variable that can be defined to prepend a prefix to the app binary names (e.g. `APP_PREFIX=bpf-`).

chenhengqi: I mean, should we add something like:
```makefile
APP_PREFIX ?= bpf-
```

jirutka: If you wanna make it the default then yes.

chenhengqi:
cc @anakryiko

libbpf-tools/Makefile
anakryiko: I'd probably leave it as empty by default, for cases when people build locally from source

anakryiko (approved):
LGTM

PR #4850, opened 2023-12-18
fix c_ulong type handle and decimal ip address bugs

singchia:
- fix c_ulong type handle logic at line 112
- fix ip decimal to human bug at line 101

which should be handled in python way

From: 
```python
def decimal_to_human(input_value):
    input_value = int(input_value)
    hex_value = hex(input_value)[2:]
    pt3 = literal_eval((str('0x'+str(hex_value[-2:]))))
    pt2 = literal_eval((str('0x'+str(hex_value[-4:-2]))))
    pt1 = literal_eval((str('0x'+str(hex_value[-6:-4]))))
    pt0 = literal_eval((str('0x'+str(hex_value[-8:-6])))) <- BUG here
    result = str(pt0)+'.'+str(pt1)+'.'+str(pt2)+'.'+str(pt3)
    return result

try:
    while True :
        time.sleep(OUTPUT_INTERVAL)
        packet_cnt_output = packet_cnt.items()
        output_len = len(packet_cnt_output)
        print('\n')
        for i in range(0,output_len):
            if (len(str(packet_cnt_output[i][0]))) != 30: <- BUG here
                continue
            temp = int(str(packet_cnt_output[i][0])[8:-2]) # initial output omitted from the kernel space program
            temp = int(str(bin(temp))[2:]) # raw file
            src = int(str(temp)[:32],2) # part1 
            dst = int(str(temp)[32:],2)
            pkt_num = str(packet_cnt_output[i][1])[7:-1]
```

To:
```python
def decimal_to_human(input_value):
    try:
        decimal_ip = int(input_value)
        ip_string = str(ipaddress.IPv4Address(decimal_ip))
        return ip_string
    except ValueError:
        return "Invalid input"

try:
    while True :
        time.sleep(OUTPUT_INTERVAL)
        packet_cnt_output = packet_cnt.items()
        output_len = len(packet_cnt_output)
        current_time = datetime.now()
        formatted_time = current_time.strftime("%Y-%m-%d %H:%M:%S")
        if output_len != 0:
            print('\ncurrent packet nums:')
        
        for i in range(0,output_len):
            srcdst = packet_cnt_output[i][0].value
            src = (srcdst >> 32) & 0xFFFFFFFF
            dst = srcdst & 0xFFFFFFFF
            pkt_num = packet_cnt_output[i][1].value
```

### DIFF
--- a/examples/networking/net_monitor.py
+++ b/examples/networking/net_monitor.py
@@ -44,7 +44,6 @@ int packet_monitor(struct __sk_buff *skb) {
     u64 pass_value = 0;
     
     struct ethernet_t *ethernet = cursor_advance(cursor, sizeof(*ethernet));
-
     struct ip_t *ip = cursor_advance(cursor, sizeof(*ip));
     if (ip->nextp != IP_TCP) 
     {
@@ -80,6 +79,9 @@ import sys
 import socket
 import os
 import struct
+import ipaddress
+import ctypes
+from datetime import datetime
 
 OUTPUT_INTERVAL = 1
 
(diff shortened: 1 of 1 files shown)


yonghong-song:
Since you are touching this file, could you add the following check in the bpf program?

```
     struct ethernet_t *ethernet = cursor_advance(cursor, sizeof(*ethernet));
+    if (ip->ver != 4)
+        return 0;
+
```
The example assumes ipv4 so let us enforce it in the bpf program.

singchia:
> Since you are touching this file, could you add the following check in the bpf program?
> 
> ```
>      struct ethernet_t *ethernet = cursor_advance(cursor, sizeof(*ethernet));
>      struct ip_t *ip = cursor_advance(cursor, sizeof(*ip));
> +    if (ip->ver != 4)
> +        return 0;
> +
> ```
> 
> The example assumes ipv4 so let us enforce it in the bpf program.

sure

PR #4867, opened 2024-01-09
tools/memleak: Use compact size of key for sizes and memptrs maps

Bojun-Seo:
As `sizes` and `memptrs` map are to store the temporal information of each thread, use only `tid` portion of the return value of `bpf_get_current_pid_tgid` as a key of `sizes` and `memptrs` map. This change could reduce the key size of each map from 64 bits to 32 bits. The lower 32 bits portion of the return value of `bpf_get_current_pid_tgid` is kernel's view of the `pid`, which in user space is usually presented as the `tid`.

Please refer this document for further information:
https://github.com/iovisor/bcc/blob/master/docs/reference_guide.md#4-bpf_get_current_pid_tgid

### DIFF
--- a/tools/memleak.py
+++ b/tools/memleak.py
@@ -161,9 +161,9 @@ struct combined_alloc_info_t {
         u64 number_of_allocs;
 };
 
-BPF_HASH(sizes, u64);
+BPF_HASH(sizes, u32, u64);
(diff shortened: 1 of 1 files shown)


yonghong-song:
Right, tid should work.

PR #4892, opened 2024-01-30
Set minimum supported llvm version to 12

yonghong-song:
Currently, memleak.py uses some atomic operations in order to have precise counting, e.g. __sync_fetch_and_sub(), etc. and these atomic operations are only supported with llvm >= 12. Previously, we set llvm minimum support version is 11. Let us boost it to 12 so more useful llvm features can be used by bpf programs.


chantra:
LGTM

PR #4897, opened 2024-02-02
add workqueue latency observation tool

jackygam2001:
add tool to observe work's waiting latency on kernel's workqueue


tests/python/test_tools_smoke.py
```diff
@@ -424,6 +424,10 @@ def test_vfsstat(self):
     def test_wakeuptime(self):
         self.run_with_duration("wakeuptime.py 1")
 
+    @skipUnless(kernel_version_ge(4,7), "requires kernel >= 4.7")
+    def test_wqlat(self):
+        self.run_with_int("wqlat.py 1 1")
```
yonghong-song, line 429: Please add 'allow_early=True' to run_with_int since the command line may finish earlier.

tools/wqlat.py
```diff
@@ -0,0 +1,181 @@
...
+args = parser.parse_args()
+countdown = int(args.count)
+if args.nanoseconds:
+    factor = 1
+    label = "nsecs"
+else:
+    factor = 1000
+    label = "usecs"
+debug = 0
+if args.wqname and len(args.wqname) >= 24:
```
yonghong-song, line 56: In kernel, one of workqueue is "events_freezable_power_efficient". Could you increase the length limit from 24 to say 40? Hopefully verification still okay. If not, we might be able to compare e.g., first batch of 8 bytes, after that doing byte-vs-byte comparison.

tools/wqlat.py
```diff
@@ -0,0 +1,181 @@
...
+"""
+
+# code substitutions
+bpf_text = bpf_text.replace('FACTOR', 'delta /= %d;' % factor)
+if args.workqueues:
+    bpf_key_text = """
+    typedef struct wq_key {
+        char wq_name[WQ_NAME_LEN];
+        u64 slot;
+    }wq_key_t;
```
yonghong-song, line 123: Please put a space between "}" and "wq_key_t".

tools/wqlat_example.txt
```diff
@@ -0,0 +1,418 @@
...
+         0 -> 1          : 0        |                                        |
+         2 -> 3          : 0        |                                        |
+         4 -> 7          : 4        |****************************************|
+         8 -> 15         : 1        |**********                              |
+
+wqname = xfs-cil/dm-4
+     usecs               : count     distribution
+         0 -> 1          : 0        |                                        |
+         2 -> 3          : 0        |                                        |
+         4 -> 7          : 1        |****************************************|
```
yonghong-song, line 367: The above dump with 'wqname' is too long, you can cut short with some '......' in between.

yonghong-song (commented):
I think this is a useful tool. Sometimes during production we do wonder whether workqueue latency is too long which amplifies the chance for race condition, etc.

Two things:
  First, could you add the tool with brief description into README.md (CPU and Scheduler Tools)
  Second, sometimes we want to identify which original task is enqueued which caused long latency. So it might be useful to add another distribution factor (tid (together with its comm)).
  
 What do you think?

tools/wqlat_example.txt
jackygam2001: okay, will do in next commit

tools/wqlat.py
jackygam2001: will do in next commit

tests/python/test_tools_smoke.py
jackygam2001: will do that.

tools/wqlat.py
jackygam2001: I'm not sure whether length 40 is okay since kernel's workqueue name length is 24, anyway I will try; but if that doesn't work, can I limit the wqname length to 16? I think length 16 can match most cases.

yonghong-song:
> 
> thanks for your suggestions, and I don't fully understand your second suggestion, do you mean adding a '-T TID' option to filter the workqueue's source thread ID for this tool?

Currently we have '-W' option to print histogram based on different workqueues. So we have
```
   workqueue1:
      <histogram>
   workqueue2:
     <histogram>
   ...
```

What I means is to add '-P' option to print histogram based on PIDs. (I think PID (process id) granularity is good enough).
```
   pid1:
      <histogram>
   pid2:
      <histogram>
```

If both -W and -P are specified, we can have
```
   workqueue1:
      pid1:
         <histogram>
      pid2:
         <histogram>
   workqueue2:
      pid3:
          <histogram>
      pid4:
          <histogram>
   ...
```

Similar to '-W', '-p <pid>' is also supported, so we only care a particular pid. So we have support of the following combinations:
```
   . -W && -P
   . -W && -p <pid>
   . -w <workqueue> && -P
   . -w <workqueue> && -p <pid>
```
What do you think?

tools/wqlat.py
yonghong-song: > I'm not sure whether length 40 is okay since kernel's workqueue name length is 24, anyway I will try; but if that doesn't work, can I limit the wqname length to 16? I think length 16 can match most cases.

Sorry, I double checked the kernel code and indeed the workqueue name length is 24.
```
  /* in kernel/workqueue.c */
  WQ_NAME_LEN             = 24
```
So your current change looks good to me.

jackygam2001:
> > thanks for your suggestions, and I don't fully understand your second suggestion, do you mean adding a '-T TID' option to filter the workqueue's source thread ID for this tool?
> 
> Currently we have '-W' option to print histogram based on different workqueues. So we have
> 
> ```
>    workqueue1:
>       <histogram>
>    workqueue2:
>      <histogram>
>    ...
> ```
> 
> What I means is to add '-P' option to print histogram based on PIDs. (I think PID (process id) granularity is good enough).
> 
> ```
>    pid1:
>       <histogram>
>    pid2:
>       <histogram>
> ```
> 
> If both -W and -P are specified, we can have
> 
> ```
>    workqueue1:
>       pid1:
>          <histogram>
>       pid2:
>          <histogram>
>    workqueue2:
>       pid3:
>           <histogram>
>       pid4:
>           <histogram>
>    ...
> ```
> 
> Similar to '-W', '-p ' is also supported, so we only care a particular pid. So we have support of the following combinations:
> 
> ```
>    . -W && -P
>    . -W && -p <pid>
>    . -w <workqueue> && -P
>    . -w <workqueue> && -p <pid>
> ```
> 
> What do you think?

-P option you mentioned in some cases does not make senses since kernel may commit the work to workqueue in interrupt context; and in the case  histogram base on PID is not the real process which commit the work, right?

yonghong-song:
> -P option you mentioned in some cases does not make senses since kernel may commit the work to workqueue in interrupt context; and in the case histogram base on PID is not the real process which commit the work, right?

Good point. I guess let us not do this PID thing now. I will do some investigation about different percentages from process context, softirq context, or others. I agree if majority is not from process context, then PID histogram probably not useful. Even if quite some workqueue works from process context, we might still want to filter out those not from process context.

libbpf-tools/Makefile
```diff
@@ -70,6 +70,7 @@ APPS = \
 	numamove \
 	offcputime \
 	oomkill \
+	profile \
```
yonghong-song, line 73: This and others mean your working directory is not clean. Could you rebase on top of master branch?

jackygam2001: have done, thanks for your reminding!

PR #4903, opened 2024-02-06
tools/bashreadline: Fix bashreadline

sancppp:
This Pull Request fixes the issue where the bashreadline tool would lose the first bash command upon startup.

At runtime, bash blocks at the readline function in bash::readline.c. Upon receiving the Enter key signal, it calls the readline_internal_teardown function to retrieve the user's input. Therefore, changing the hook function from readline to readline_internal_teardown resolves the issue of losing the first bash command after startup of bashreadline.

The specific approach is to open the symbol table of bash and check if there is a symbol for readline_internal_teardown function. If not, continue to use the readline function as the hook position.

### DIFF
--- a/docker/Dockerfile.ubuntu
+++ b/docker/Dockerfile.ubuntu
@@ -24,5 +24,5 @@ COPY --from=builder /root/bcc/*.deb /root/bcc/
 RUN \
   apt-get update -y && \
   DEBIAN_FRONTEND=noninteractive apt-get install -y python3 python3-pip python-is-python3 binutils libelf1 kmod  && \
-  pip3 install dnslib cachetools  && \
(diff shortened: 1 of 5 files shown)


yonghong-song:
Could you give more details about how to reproduce `lose the first bash command upon startup`? That is, give a sequence of commands where the result will be different with vs. without your patch.

sancppp:
> Could you give more details about how to reproduce `lose the first bash command upon startup`? That is, give a sequence of commands where the result will be different with vs. without your patch.


https://github.com/iovisor/bcc/assets/67474179/5330ffd7-d873-4918-997f-388541452871


https://github.com/iovisor/bcc/assets/67474179/3b675585-43d9-4980-8525-21d69a8c3dc8

PR #4904, opened 2024-02-06
tools/execsnoop: display which cpu the processes running on

markyangcc:
In SMP systems, knowing which specific CPU a process is running on can be useful for troubleshooting.

Test it,
```
[root@anylinux ~]# taskset -c 0 sleep 100 &
[1] 2199347
[root@anylinux ~]# taskset -c 1 sleep 100 &
[2] 2199348
```

Now, execsnoop with '-C' option displays that one sleep process is running on CPU 0, the other is running on CPU 1, as expected.

```
[root@anylinux ~]# /usr/share/bcc/tools/execsnoop -C
PCOMM            PID     PPID    CPU  RET ARGS
taskset          2199347 2199222 0      0 /usr/bin/taskset -c 0 sleep 100
sleep            2199347 2199222 0      0 /usr/bin/sleep 100
taskset          2199348 2199222 4      0 /usr/bin/taskset -c 1 sleep 100
sleep            2199348 2199222 1      0 /usr/bin/sleep 100
```

### DIFF
--- a/tools/execsnoop.py
+++ b/tools/execsnoop.py
@@ -54,6 +54,7 @@ examples = """examples:
     ./execsnoop -T                   # include time (HH:MM:SS)
     ./execsnoop -P 181               # only trace new processes whose parent PID is 181
     ./execsnoop -U                   # include UID
+    ./execsnoop -C                   # include CPU
     ./execsnoop -u 1000              # only trace UID 1000
     ./execsnoop -u user              # get user UID and trace only them
     ./execsnoop -t                   # include timestamps
@@ -90,6 +91,8 @@ parser.add_argument("-l", "--line",
     help="only print commands where arg contains this line (regex)")
 parser.add_argument("-U", "--print-uid", action="store_true",
     help="print UID column")
+parser.add_argument("-C", "--print-cpu", action="store_true",
+    help="print CPU column")
 parser.add_argument("--max-args", default="20",
     help="maximum number of arguments parsed and displayed, defaults to 20")
 parser.add_argument("-P", "--ppid",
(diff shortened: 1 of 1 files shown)


tools/execsnoop.py
```diff
@@ -207,6 +208,7 @@ def parse_uid(user):
     // as the real_parent->tgid.
     // We use the get_ppid function as a fallback in those cases. (#1883)
     data.ppid = task->real_parent->tgid;
+    data.cpu = task->cpu;
```
yonghong-song, line 211: Which kernel version you are testing? I checked latest upstream and 5.19, the field name for 'task' is 'on_cpu' and not 'cpu'. Please ensure your change works for all supported kernels (from e.g. 4.11 to recent kernels).

markyangcc: Yes, 'cpu' filed was changed on recent kernels. It was move back to 'thread_info' since linux 5.16
Link: https://github.com/torvalds/linux/commit/bcf9033e5449
Now, BTF filed check or buildtime filed check is added to handle this change.

yonghong-song:
can we add this 'cpu' display under an option (e.g., '-c')? The reason is in most cases, the value will be 0 and we should not display them. People can use -c if they have pinned-to-cpu task and they want to monitor them.

tools/execsnoop.py
```diff
@@ -272,7 +275,9 @@ def check_cpu_filed():
     print("%-8s" % ("TIME(s)"), end="")
 if args.print_uid:
     print("%-6s" % ("UID"), end="")
-print("%-16s %-7s %-7s %-4s %3s %s" % ("PCOMM", "PID", "PPID", "CPU", "RET", "ARGS"))
+if args.print_cpu:
+    print("%-6s" % ("CPU"), end="")
+print("%-16s %-7s %-7s %3s %s" % ("PCOMM", "PID", "PPID", "RET", "ARGS"))
```
yonghong-song, line 280: I would like CPU column after 'PPID' and before 'RET' and this confirms to most other bcc tools where PCOMM/PID/PPID is displayed earlier in columns.

markyangcc: Thanks reviewing, I rearranged the CPU column display position. Now it's after 'PPID' and before 'RET'.

PR #4912, opened 2024-02-16
tcpdrop: add support for filtering by netns

tych0:
We are concerned about drops from a specific container, so add support for supplying an explicit netns to scope to (or as a convenience, a pid).


tools/tcpdrop.py
```diff
@@ -193,6 +200,16 @@
 else:
     bpf_text = bpf_text.replace('FILTER_FAMILY', '')
 
+if args.pid_netns != 0:
+    args.netns_id = os.stat('/proc/{}/ns/net'.format(args.pid_netns)).st_ino
```
yonghong-song, line 204: Let us add a check here. If both pid_netns and netns_id are not zero, return a failure.

PR #4920, opened 2024-02-27
libbpf-tools: fix syscall tracepoints

jeromemarchand:
In BPF programs, syscall tracepoints returns a local syscall_tp_t structure that's meant to mimic syscall_trace_enter/exit structures. It just happen to have the same offsets for nr and args fields as struct trace_event_raw_sys_enter/exit. However, a change in the trace_entry structure in the linux-rt broke that assumption (ea622076b76f "sched: Add support for lazy preemption").

Program calling syscall tracepoint should have used the right structure in the first place. This patch replace the trace_event_raw_sys_* by their syscall_trace_* counterpart when calling a syscall tracepoint.

### DIFF
--- a/libbpf-tools/execsnoop.bpf.c
+++ b/libbpf-tools/execsnoop.bpf.c
@@ -36,7 +36,7 @@ static __always_inline bool valid_uid(uid_t uid) {
 }
 
 SEC("tracepoint/syscalls/sys_enter_execve")
-int tracepoint__syscalls__sys_enter_execve(struct trace_event_raw_sys_enter* ctx)
+int tracepoint__syscalls__sys_enter_execve(struct syscall_trace_enter* ctx)
 {
 	u64 id;
 	pid_t pid, tgid;
@@ -112,7 +112,7 @@ int tracepoint__syscalls__sys_enter_execve(struct trace_event_raw_sys_enter* ctx
 }
 
 SEC("tracepoint/syscalls/sys_exit_execve")
-int tracepoint__syscalls__sys_exit_execve(struct trace_event_raw_sys_exit* ctx)
+int tracepoint__syscalls__sys_exit_execve(struct syscall_trace_exit* ctx)
 {
 	u64 id;
 	pid_t pid;
--- a/libbpf-tools/futexctn.bpf.c
+++ b/libbpf-tools/futexctn.bpf.c
@@ -49,7 +49,7 @@ struct {
 static struct hist initial_hist = {};
 
 SEC("tracepoint/syscalls/sys_enter_futex")
-int futex_enter(struct trace_event_raw_sys_enter *ctx)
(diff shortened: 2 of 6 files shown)


jeromemarchand:
@sm00th Please have a look.

sm00th:
Looks good, thank you.

yonghong-song:
https://lore.kernel.org/r/20230801075222.7717-1-ykaliuta@redhat.com (commit d3c4db86c711 "tracing: bpf: use struct trace_entry in struct syscall_tp_t") should fix this issue but it only available in recent kernels.

Regarding trace_event_raw_sys_* structs, we have
```
struct trace_event_raw_sys_enter {
        struct trace_entry ent;
        long int id;
        long unsigned int args[6];
        char __data[0];
}; 
```
So from input parameter perspective, it looks good to me. Or I miss some critical things?

jeromemarchand:
Commit d3c4db86c711 fixed the issue for bpf programs using trace_event_raw_sys_*, but other issues remained.
Since then, commit ba8ea72388a1 ("bpf: Change syscall_nr type to int in struct syscall_tp_t") changes syscall_tp_t again:
```
diff --git a/kernel/trace/trace_syscalls.c b/kernel/trace/trace_syscalls.c
index de753403cdaf..9c581d6da843 100644
--- a/kernel/trace/trace_syscalls.c
+++ b/kernel/trace/trace_syscalls.c
@@ -556,7 +556,7 @@ static int perf_call_bpf_enter(struct trace_event_call *call, struct pt_regs *re
 {
        struct syscall_tp_t {
                struct trace_entry ent;
-               unsigned long syscall_nr;
+               int syscall_nr;
                unsigned long args[SYSCALL_DEFINE_MAXARGS];
        } __aligned(8) param;
        int i;
@@ -661,7 +661,7 @@ static int perf_call_bpf_exit(struct trace_event_call *call, struct pt_regs *reg
 {
        struct syscall_tp_t {
                struct trace_entry ent;
-               unsigned long syscall_nr;
+               int syscall_nr;
                unsigned long ret;
        } __aligned(8) param;
 
```

sm00th:
More background here: https://lore.kernel.org/lkml/20231005123413.GA488417@alecto.usersys.redhat.com/t/#m5d5dcd2e4dee03c76506158163ba8dcac40901c3

jeromemarchand:
One more thing, the format of bpf tracepoints structure should be taken from `/sys/kernel/tracing/events/*/*/format` and the formats of raw_syscalls and syscall are different.  One has a `long id` and the other a `int syscall_nr`. The args field used to be at the same offset, but that changed in RT kernels with the addition of common_preempt_lazy_count.

yonghong-song:
@jeromemarchand @sm00th Thanks for explanation! I missed that syscall_nr is changed from long to int and the potential alignment change here.

PR #4928, opened 2024-03-06
clang: check header ownership

brendangregg:
Example testing with a brendan-owned /tmp/kheaders file (note the "ERROR:" message):

```
~/bcc/build$ sudo /usr/share/bcc/tools/biosnoop
ERROR: header file ownership unexpected: /tmp/kheaders-5.15.47-internal
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
  File "/usr/share/bcc/tools/biosnoop", line 335, in <module>
    b = BPF(text=bpf_text)
  File "/usr/lib/python3/dist-packages/bcc-0.1.5+6cd27218-py3.10.egg/bcc/__init__.py", line 479, in __init__
~/bcc/build$ ls -lhd /tmp/kheaders-5.15.47-internal
drwxrwxr-x 2 brendan dev 4.0K Mar  6 02:50 /tmp/kheaders-5.15.47-internal
```

No error when chown'd back to root.

### DIFF
--- a/src/cc/frontends/clang/kbuild_helper.cc
+++ b/src/cc/frontends/clang/kbuild_helper.cc
@@ -140,15 +140,22 @@ int KBuildHelper::get_flags(const char *uname_machine, vector<string> *cflags) {
   return 0;
 }
 
-static inline int file_exists(const char *f)
+static inline int file_exists_and_ownedby(const char *f, uid_t uid)
 {
   struct stat buffer;
-  return (stat(f, &buffer) == 0);
+  int ret;
+  if ((ret = stat(f, &buffer)) == 0) {
+    if (buffer.st_uid != uid) {
+      std::cout << "ERROR: header file ownership unexpected: " << std::string(f) << "\n";
+      return -1;
+    }
+  }
+  return ret;
 }
 
 static inline int proc_kheaders_exists(void)
 {
-  return file_exists(PROC_KHEADERS_PATH);
+  return file_exists_and_ownedby(PROC_KHEADERS_PATH, 0);
 }
 
 static inline const char *get_tmp_dir() {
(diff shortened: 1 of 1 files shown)


src/cc/frontends/clang/kbuild_helper.cc
```diff
@@ -140,15 +140,22 @@ int KBuildHelper::get_flags(const char *uname_machine, vector<string> *cflags) {
...
 
-static inline int file_exists(const char *f)
+static inline int file_exists_and_ownedby(const char *f, uid_t uid)
 {
   struct stat buffer;
-  return (stat(f, &buffer) == 0);
+  int ret;
+  if ((ret = stat(f, &buffer)) == 0) {
+    if (buffer.st_uid != uid) {
+      std::cout << "ERROR: header file ownership unexpected: " << std::string(f) << "\n";
```
jordalgo, line 149: Nit: is it useful to also print out the expected `uid` and the one in the file: `st_uid`?

jordalgo:
Do we need to check the ld cache ownership too? https://github.com/iovisor/bcc/blob/master/src/cc/bcc_proc.c#L400

src/cc/frontends/clang/kbuild_helper.cc
brendangregg: I think this is close enough: says this filename has this type of error. End-user should likely see this and run `ls -l`. There's related codepaths that can error with no such hint, which is pretty hard to debug, which is why I put a big clue in. :-)

brendangregg:
Thanks, although the ld cache is only in /etc, right? /etc should already be locked down, so we wouldn't need to check. If unprivileged users can create files in /etc, I imagine there's a ton of other issues.

jordalgo (approved):

jeromemarchand:
I'm afraid this is not fixed yet. The example above fails only because it does not contain the include/linux/kconfig.h file.
If a user builds a valid kheaders tree, with possibly a few chosen alteration, we're going to show an error mesage, but we're still going to use the compromised headers.
```
$ sudo /usr/share/bcc/tools/execsnoop 
ERROR: header file ownership unexpected: /tmp/kheaders-6.8.0-63.fc41.x86_64
ls               1690    1426      0 /usr/bin/ls --color=auto
```

PR #4940, opened 2024-03-18
Fix slabratetop.py for 6.8 kernel

yonghong-song:
slabratetop.py failed with 6.8 kernel with
```    
      /virtual/main.c:100:10: fatal error: 'linux/slub_def.h' file not found
        100 | #include <linux/slub_def.h>
            |          ^~~~~~~~~~~~~~~~~~
      1 error generated.
```
    
Actually 'kmem_cache' is also not available to bcc any more.
We already copied 'slab' data structure to bcc, copying yet
another one (probably with others) will make things too
complex and error prone.
    
Let us use newly introduced kernel_struct_field_size_offset() API.
We could get some data structure (field, size) information
from kernel BTF. This enables us to work around header file
not available issue.

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -1540,6 +1540,34 @@ static int find_member_by_name(struct btf *btf, const struct btf_type *btf_type,
   return 0;
 }
 
+static unsigned long long find_member_size_offset(struct btf *btf, const struct btf_type *btf_type,
+						  const char *field_name) {
+  const struct btf_member *btf_member = btf_members(btf_type);
+  unsigned long long ret;
+  int i;
+
+  for (i = 0; i < btf_vlen(btf_type); i++, btf_member++) {
+    const char *name = btf__name_by_offset(btf, btf_member->name_off);
+    if (!strcmp(name, field_name)) {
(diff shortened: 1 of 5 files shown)


yonghong-song:
cc @jeromemarchand

4ast (commented):
this is fragile. Why do we need this .py based tool when we already have one based on vmlinux.h, libbpf, and CORE. Doing mini-CORE like logic specific for this tool will be challenging to maintain. I'd just delete this slabratetop.py

yonghong-song:
> this is fragile. Why do we need this .py based tool when we already have one based on vmlinux.h, libbpf, and CORE. Doing mini-CORE like logic specific for this tool will be challenging to maintain. I'd just delete this slabratetop.py

Okay, I will abandon this patch. But since this patch is to address an issue with 6.8 kernel which is just release a week ago and has not made into the main distribution yet. I think we do not need to delete it now. Instead of this patch, I will have another patch to warn user that >= 6.8 release won't work and suggest user to use libbpf tool version slabratetop. The tool itself will be moved to 'old' directory at some point in the future when most users won't be able to use it any more.

PR #4943, opened 2024-03-20
fix deb build/docker ubuntu

stonith:
- Fixes typo in `debian/control` which was causing deb builds to fail
- add llvm-12-dev package to ubuntu dockerfile


yonghong-song:
LGTM. Thanks for the fix. At some point, we should upgrade the compiler from llvm12 to say llvm17.

PR #4944, opened 2024-03-21
libbpf-tools/gethostlatency: correct section name

ShawnZhong:
`handle_entry` and `handle_return` are attached to uprobes `getaddrinfo`, `gethostbyname`, and `gethostbyname2`, so it doesn't make sense to have the section name be `kprobe`

### DIFF
--- a/libbpf-tools/gethostlatency.bpf.c
+++ b/libbpf-tools/gethostlatency.bpf.c
@@ -60,13 +60,13 @@ static int probe_return(struct pt_regs *ctx)
 	return 0;
 }
 
-SEC("kprobe/handle_entry")
+SEC("uprobe")
 int BPF_KPROBE(handle_entry)
 {
 	return probe_entry(ctx);
 }
 
-SEC("kretprobe/handle_return")
+SEC("uretprobe")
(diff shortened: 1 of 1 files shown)


libbpf-tools/gethostlatency.bpf.c
```diff
@@ -60,13 +60,13 @@ static int probe_return(struct pt_regs *ctx)
 	return 0;
 }
 
-SEC("kprobe/handle_entry")
+SEC("uprobe")
 int BPF_KPROBE(handle_entry)
```
chenhengqi, line 64: We can also replace `BPF_KPROBE` with `BPF_UPROBE`.

ShawnZhong: I thought about this as well. I didn't make the change because I don't see any other usage of `BPF_UPROBE` in this codebase. 

FYI, the macros `BPF_UPROBE` and `BPF_URETPROBE` are added in [libbpf v1.2.0](https://github.com/libbpf/libbpf/releases/tag/v1.2.0) released on May 1, 2023. That's not an issue if people build libbpf-tools using the libbpf submodule under `src/cc/libbpf`.

PR #4948, opened 2024-03-26
remove llvm-12-dev from debian package dependency

stonith:
llvm-12-dev shouldn't be required for the deb packages and saves ~200+MB in the resulting container


yonghong-song:
Do you have any reference/pointer to show debian package build does not need llvm?

PR #4964, opened 2024-04-15
libbpf-tools: add string helpers, filter ksyms by regex

myhro:
The goal is offer the basis of what is required, in terms of regex handling, to implement the C equivalent of what `b.attach_kprobe(event_re=...)` does in Python. The first client that will be making use of it is a CO-RE port of `vfscount`:

- https://github.com/iovisor/bcc/pull/4965

This could also be useful to, for instance, add regex support to the `libbpf-tools/funclatency` tool which is currently lacking this feature when compared to the Python version:

https://github.com/iovisor/bcc/blob/c0e9b562675bcd18eff9443af4e0d2cc55e38738/libbpf-tools/funclatency.c#L7-L9


myhro:
Closing this PR as there wasn't activity in it for ~2 years.

PR #4979, opened 2024-04-30
tools/old/profile.py: Backport the page_offset from tools/profile.py

bernhardkaindl:
Fix `tools/old/profile.py` for older kernels like 4.19 (tested):

This is done using a plain an unmodified backport the of the preprocessor conditions for getting the `page_offset` for different kernel versions and architectures from tools/profile.py.

It should fix getting the page_offset for the full range of versions and architectures supported by the conditions from `tools/profile.py`, but I can't test all of them.

However, as `tools/profile.py` is the default profile tool packaged by distros like Fedora, the method used in `tools/profile.py` should be a lot better than the very limited old method of getting the page_offset.

### DIFF
--- a/tools/old/profile.py
+++ b/tools/old/profile.py
@@ -162,13 +162,26 @@ PERF_TRACE_EVENT {
         struct pt_regs regs = {};
         bpf_probe_read(&regs, sizeof(regs), (void *)REGS_LOCATION);
         u64 ip = PT_REGS_IP(&regs);
+        u64 page_offset;
 
         // if ip isn't sane, leave key ips as zero for later checking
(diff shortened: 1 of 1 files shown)


bernhardkaindl (commented):
With #4974 merged, I can use `tools/profile.py` instead of `tools/old/profile.py`.

@yonghong-song:
Because of this, I'm open to closing it if you'd rather not like to change `tools/old/profile.py`

yonghong-song:
>  Because of this, I'm open to closing it if you'd rather not like to change tools/old/profile.py
Yes, it would be great if you can just use tools/profile.py instead of tools/old/profile.py.

PR #4987, opened 2024-05-06
libbpf tcptop: Fix "unrecognized bpf_ld_imm64 insn" on 4.19

bernhardkaindl:
Fix the BPF load error `unrecognized bpf_ld_imm64 insn` in `tcptop` on 4.19 by converting `.rodata` config variables to a map.

`tcptop` can use CO-RE including on 4.19 when a detached `vmlinux` BTF file is provided.
It can be generated from a DWARF `vmlinux` using `pahole`:
```sh
pahole -J;objcopy --only-section=.BTF vmlinux /boot/vmlinux-$(uname -r)
```

### DIFF
--- a/libbpf-tools/tcptop.bpf.c
+++ b/libbpf-tools/tcptop.bpf.c
@@ -12,9 +12,12 @@
 #define AF_INET		2	/* Internet IP Protocol 	*/
 #define AF_INET6	10	/* IP version 6			*/
 
-const volatile bool filter_cg = false;
-const volatile pid_t target_pid = -1;
-const volatile int target_family = -1;
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY);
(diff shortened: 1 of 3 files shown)


libbpf-tools/tcptop.bpf.c
```diff
@@ -12,9 +12,12 @@
 #define AF_INET		2	/* Internet IP Protocol 	*/
 #define AF_INET6	10	/* IP version 6			*/
 
-const volatile bool filter_cg = false;
-const volatile pid_t target_pid = -1;
-const volatile int target_family = -1;
```
yonghong-song, line 17: The above coding pattern has been used in most of libbpf-tools bpf programs. I would rather to keep this. Changing to use map does not sound good compared to more elegant way in the above. 4.19 is quite old. Maybe you could folk the repo and make change in your private repo?

PR #4990, opened 2024-05-08
libbpf-tools/profile: Batch lookup support for hash map

ekyooo:
Use bpf_map_lookup_batch() instead of bpf_map_get_next_key() to read the hash map, if available.


libbpf-tools/profile.c
```diff
@@ -268,13 +270,61 @@ static int cmp_counts(const void *a, const void *b)
...
 
+	if (batch_map_ops) {
+		err = read_batch_counts_map(fd, items, count);
+		if (err < 0 && err == -EINVAL) {
+			/* fall back to a racy variant */
+			batch_map_ops = false;
+		} else {
+			return err;
+		}
+	}
```
yonghong-song, line 326: In map_helpers.c, we have dump_hash(). Could you try to use that function?

ekyooo: Yes, there is duplicate code in this patch.
read_counts_map() in profile.c returns data as an array of type "struct key_ext_t", which is a combination of the keys and values ​​from the hash map. This is the difference from the design of dump_hash() in map_helpers.c.

In the current patch, the combination is only needed for lookup_batch. However, using map_helpers' dump_hash() requires combination (data copying) for both the batch and iteration cases. Perhaps that's why syscount.c doesn't use map_helpers either. Could you tell me what you think about it?

There is also a way to utilize only dump_hash_batch() of map_helpers.c after removing static.
Thank you.

yonghong-song: I think dump_hash() should work. It will cover both batch and non-batch case. I suggest to replace non-batch case with dump_hash(). Could you give a try?

ekyooo: Of course, I'll update it.

libbpf-tools/profile.c
```diff
@@ -270,26 +271,29 @@ static int cmp_counts(const void *a, const void *b)
...
-
-		if (items[i].v == 0)
-			continue;
+	err = dump_hash(fd, keys, key_size, vals, value_size, &n, &empty);
+	if (err)
+		return err;
 
-		lookup_key = &items[i].k;
-		i++;
+	for (i = 0; i < n; i++) {
```
ekyooo, line 287: In previous patches, maps were read directly into the "items" variable, so this copy would not work unless it was in a batch. Now we use a little more memory and CPU. However, this change is good for eliminating duplicate code.

PR #4997, opened 2024-05-09
Narrow libbcc dependency from llvm-dev to just libllvm

bobrik:
It works with list libllvm, so no need to pull in the whole thing, which is 600MiB worth of packages.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -20,7 +20,7 @@ Package: libbcc
 Architecture: any
 Provides: libbpfcc, libbpfcc-dev
 Conflicts: libbpfcc, libbpfcc-dev
-Depends: libc6, libstdc++6, libelf1, llvm-12-dev
+Depends: libc6, libstdc++6, libelf1, libllvm12
 # add 'libdebuginfod1' to Depends if built with libdebuginfod support
 Description: Shared Library for BPF Compiler Collection (BCC)
  Shared Library for BPF Compiler Collection to control BPF programs


bobrik:
@shunghsiyu, what exactly broke for you prior to #4921?

shunghsiyu:
> It works with list libllvm, so no need to pull in the whole thing, which is 600MiB worth of packages.
> @shunghsiyu, what exactly broke for you prior to #4921?

The problem for me is that using bcc header files (packaged as `bcc-devel` on RPM systems) requires `llvm-devel` to be on installed as well since 6f11bf7e2806658c4bd69415b921cedf85d9ebfe, otherwise compilation will fail.

debian/control does not have a separate package to host the bcc header files so I ended up placing it along with the main bcc package, which is admittedly not ideal.

shunghsiyu:
From a packaging point of view I'd much prefer we revert 6f11bf7e2806658c4bd69415b921cedf85d9ebfe to keep the dependency at a minimum.

That said I only know two users of `bcc-devel`: one is [Sysinternals/ProcMon-for-Linux](https://github.com/Sysinternals/ProcMon-for-Linux) mentioned above, and the other is [bpftrace/bpftrace](https://github.com/bpftrace/bpftrace). bpftrace already pulls `llvm-devel` anyway, while ProcMon-for-Linux doesn't get packaged everywhere; so this is perhaps a bit of bikeshedding.

@chantra what do you think?

chantra:
On Sun, May 19, 2024 at 11:02 PM Shung-Hsi Yu ***@***.***>
wrote:

> From a packaging point of view I'd much prefer we revert 6f11bf7
> <https://github.com/iovisor/bcc/commit/6f11bf7e2806658c4bd69415b921cedf85d9ebfe>
> to keep the dependency at a minimum.
>

For packaging, this should only be a build dependency right:
https://www.debian.org/doc/debian-policy/ch-relationships.html ?
The package itself has a pretty minimal footprint (26kb on Ubuntu per
https://packages.ubuntu.com/noble/llvm-dev). Probably this pulls more deps,
but only on the build machine.

Quickly checking the issue, I think moving llvm-dev to a “Build-Depends”
would solve the problem for machines where bcc gets installed.

Would that be a good enough middle ground?

That said I only know two users of bcc-devel: one is

chantra:
If bcc-devel requires llvm-devel, I suppose the problem is because we pull llvm-config.h in bcc_module.h which is public? Moving llvm-config.h to private only headers should do the job.

Can you list the public headers and we can work on moving this include out of the way.

shunghsiyu:
> If bcc-devel requires llvm-devel, I suppose the problem is because we pull llvm-config.h in bcc_module.h which is public? Moving llvm-config.h to private only headers should do the job.

I think that will work

> Can you list the public headers and we can work on moving this include out of the way.

The public headers I founds are:
- BPF.h
- BPFTable.h
- bcc_common.h
- bcc_elf.h
- bcc_exception.h
- bcc_proc.h
- bcc_syms.h
- bcc_usdt.h
- bcc_version.h
- bpf_module.h
- file_desc.h
- libbpf.h
- perf_reader.h
- table_desc.h
- table_storage.h

So bpf_module.h should be the one that need modification.

chantra:
#5018 should fix this issue. @shunghsiyu could you confirm the package change are fine? cc @yonghong-song

PR #5003, opened 2024-05-15
Add filters to filetop

Hannibal404:
Add filter to reverse the output order of filetop, and filters to trace either read or write operations at a time

### DIFF
--- a/tools/filetop.py
+++ b/tools/filetop.py
@@ -4,7 +4,8 @@
 # filetop  file reads and writes by process.
 #          For Linux, uses BCC, eBPF.
 #
-# USAGE: filetop.py [-h] [-C] [-r MAXROWS] [interval] [count]
+# USAGE: filetop.py [-h] [-a] [-C] [-r MAXROWS] [-p PID] [--read-only]
+#                   [--write-only] [interval] [count]
 #
 # This uses in-kernel eBPF maps to store per process summaries for efficiency.
 #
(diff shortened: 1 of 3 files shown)


yonghong-song:
Could you explain why you add these filters and why they are useful? The existing output should already cover what you try to implement.

yonghong-song:
Also, for not-single-character option, please use '--option' format.

Hannibal404:
> Could you explain why you add these filters and why they are useful? The existing output should already cover what you try to implement.

The filters implemented allow for finer searching for any files that we need to look for. In case we know the file we are looking for has very few operations, sorting the the reversed order for the same would help. Similarly tracking only read/write operations would help reduce the clutter when a large number of files is there


> Also, for not-single-character option, please use '--option' format.

Updated

tools/filetop.py
```diff
@@ -164,8 +174,13 @@
 
 # initialize BPF
 b = BPF(text=bpf_text)
-b.attach_kprobe(event="vfs_read", fn_name="trace_read_entry")
-b.attach_kprobe(event="vfs_write", fn_name="trace_write_entry")
+if not args.writeonly:
```
yonghong-song, line 177: write_only?

tools/filetop.py
```diff
@@ -164,8 +174,13 @@
...
+    b.attach_kprobe(event="vfs_read", fn_name="trace_read_entry")
+if not args.readonly:
```
yonghong-song, line 179: read_only?

tools/filetop.py
```diff
@@ -164,8 +174,13 @@
...
+    b.attach_kprobe(event="vfs_write", fn_name="trace_write_entry")
+if args.readonly and args.writeonly:
```
yonghong-song, line 181: read_only and write_only?

tools/filetop.py
```diff
@@ -4,7 +4,8 @@
 # filetop  file reads and writes by process.
 #          For Linux, uses BCC, eBPF.
 #
-# USAGE: filetop.py [-h] [-C] [-r MAXROWS] [interval] [count]
+# USAGE: filetop.py [-h] [-a] [-C] [-r MAXROWS] [-p PID] [--reverse]
+#                   [--read-only] [--write-only] [interval] [count]
```
yonghong-song, line 8: I can see write-only and read-only are useful since sorting can be done based on either read or write but not both.
But I don't like --reverse. Currently, the sorting is in descending order which is the same for other tools e.g.  biotop.py.

tools/filetop.py
Hannibal404: I kept the flags this way to handle the cases when neither --read-only/--write-only are passed, or when both are passed. If you prefer I can make it throw an error when both are passed instead

yonghong-song: Let us do without --read-only/--write-only, both will handle. If both --read-only/--write-only, let us throw an error.

Hannibal404: Updated

tools/filetop.py
Hannibal404: Removed --reverse

Issue #5013, opened 2024-05-23
`bcc_procutils_which_so(libname, pid)` finds library unused by the process


gukoff:
## Problem

When debugging a specific process, I might need to set the probes in the shared library that this process is using.

The function `bcc_procutils_which_so(libname, pid)` seemingly helps with this. But in reality, it also looks for the library in the ldconfig cache.

It potentially returns the libraries unrelated to the given process even if I pass a `pid` to this function, which breaks my expectations as a library user. Example: https://github.com/sumerc/gilstats.py/pull/1

## Potential solutions

1. **Backwards-compatible**: export a more specialized function `bcc_procutils_which_so_in_process(libname, pid)`.
2. **Backwards-incompatible**: force `bcc_procutils_which_so(..., pid)` to search only the libraries of the given process when `pid != 0`.

Also, here's an implementation of the backwards-compatible solution: https://github.com/iovisor/bcc/pull/5014

PR #5014, opened 2024-05-23
Extend `bcc_proc` API. Allow to limit search to specific pid.

gukoff:
Fixes https://github.com/iovisor/bcc/issues/5013

- Extend `bcc_proc` API. Allow to limit search to specific pid.
- Also extend the Python binding with the same goal.
- The API changes are backwards-compatible.
- Also added a couple of boundary checks for `memcpy`

### DIFF
--- a/src/python/bcc/libbcc.py
+++ b/src/python/bcc/libbcc.py
@@ -233,8 +233,6 @@ class bcc_symbol_option(ct.Structure):
 
 lib.bcc_procutils_which_so_in_process.restype = ct.POINTER(ct.c_char)
 lib.bcc_procutils_which_so_in_process.argtypes = [ct.c_char_p, ct.c_int]
-lib.bcc_procutils_which_so_in_ldconfig_cache.restype = ct.POINTER(ct.c_char)
-lib.bcc_procutils_which_so_in_ldconfig_cache.argtypes = [ct.c_char_p]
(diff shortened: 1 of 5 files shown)


gukoff:
The failed test seems unrelated to this change; please retry?

src/python/bcc/libbcc.py
```diff
@@ -231,6 +231,10 @@ class bcc_symbol_option(ct.Structure):
             ('use_symbol_type', ct.c_uint),
         ]
 
+lib.bcc_procutils_which_so_in_process.restype = ct.POINTER(ct.c_char)
+lib.bcc_procutils_which_so_in_process.argtypes = [ct.c_char_p, ct.c_int]
+lib.bcc_procutils_which_so_in_ldconfig_cache.restype = ct.POINTER(ct.c_char)
+lib.bcc_procutils_which_so_in_ldconfig_cache.argtypes = [ct.c_char_p]
```
yonghong-song, line 237: Are the above lib.bcc_procutils_which_so_in_ldconfig_cache necessary or not? find_library(libname, 0) should cover this.

yonghong-song (commented):
Agree that we do not want to break backward compatibility. So add a new parameter to find_library looks good to me.

src/python/bcc/libbcc.py
gukoff: Fair enough! No need to expose this function.

PR #5035, opened 2024-06-19
Add the line "setlinebuf(stdout);" to avoid data output is not timely

zhangzihengya:
It is recommended to add "setlinebuf(stdout);" to all tools.In order to avoid data output is not timely.If I can, I'll add this line of code to the other tools.


yonghong-song:
Do you have a specific use case for this patch?

zhangzihengya:
> Do you have a specific use case for this patch?

At present, there is no concrete case, that is, in programming, I feel that adding this line of code is more friendly and real-time for the output, and has no other impact on other aspects of the code. I wonder if you are willing to join this line of code?

PR #5041, opened 2024-06-25
Updating Debian source installation build tool versions

VishnuVelayuthan:
Was having a problem using clang version 16 while building bcc from source in a fresh debian bookworm container. Found the solution to be downgrading my clang version to 13. Updated the Install.README to reflect those changes to the installation steps.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -345,11 +345,12 @@ deb-src http://deb.debian.org/debian sid main contrib non-free
 apt-get update
 # According to https://packages.debian.org/source/sid/bpfcc,
 # BCC build dependencies:
-sudo apt-get install arping bison clang-format cmake dh-python \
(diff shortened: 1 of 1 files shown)


yonghong-song:
Maybe cmake/clang_libs.cmake needs some tweak to make clang16 work. Unfortunately, I am not able to reproduce the issue in my environment with clang cmake
```
    -DLLVM_ENABLE_PROJECTS="clang;llvm;compiler-rt;lld" \
    -DLLVM_ENABLE_ASSERTIONS=ON \
    -DLLVM_ENABLE_ZLIB=ON \
```
I think  your clang16 build definitely has a different cmake command line than mine.

c2main:
After trying your solution, I just `sudo apt install libpolly-16-dev` instead, and it works now.

PR #5049, opened 2024-07-02
tcpconnect: increase comm field size

benjaminp:
The max size of comm is 16 bytes.


libbpf-tools/tcpconnect.c
```diff
@@ -296,7 +296,7 @@ static void print_events_header()
 		printf("%-9s", "TIME(s)");
 	if (env.print_uid)
 		printf("%-6s", "UID");
-	printf("%-6s %-12s %-2s %-16s %-16s",
+	printf("%-6s %-16s %-2s %-16s %-16s",
```
yonghong-song, line 299: The format in handle_event() should also be updated.

PR #5089, opened 2024-08-24
bcc: Use `-Werror=undef`

### LINKED ISSUE #5019: Use `-Werror=undef` when compiling bcc to catch issues when a macro is not defined.
A macro which is not defined will default to 0, this can make it challenging to debug issues as the compiler will happily go ahead and treat an undefined macro to 0.

Here is an example case where it caught an issue: https://github.com/bpftrace/bpftrace/pull/2950/commits/a0eefb6491431d90457c38107a7724f4303e8312

ShawnZhong:
FIx #5019


Bojun-Seo:
How about adding some explanation about this commit in the commit message?
You can write the whole reason in it, or you can just leave a link like this PR description.
(e.g. `Fix https://github.com/iovisor/bcc/issues/5019`)
I prefer the former by the way.

src/cc/usdt/usdt.cc
```diff
@@ -36,15 +36,15 @@ Location::Location(uint64_t addr, const std::string &bin_path, const char *arg_f
     : address_(addr),
       bin_path_(bin_path) {
 
-#ifdef __aarch64__
+#if defined(__aarch64__)
```
Bojun-Seo, line 39: Is it necessary to change `#ifdef A` into `#if defined(A)` ?

yonghong-song: Agree.  ifdef -> defined(...) change seems unnecessary.

PR #5106, opened 2024-09-14
Adding ability to truncate output

ShawnTuatara:
If there are too many tcp connections then the top doesn't work well as it scrolls the shell window. This `-o` option lets you limit the output to however many lines you want.

### DIFF
--- a/tools/tcptop.py
+++ b/tools/tcptop.py
@@ -61,6 +61,8 @@ parser.add_argument("-S", "--nosummary", action="store_true",
     help="skip system summary line")
 parser.add_argument("-p", "--pid",
     help="trace this PID only")
+parser.add_argument("-o", "--outputlines", default=-1, type=range_check,
+    help="Number of lines to output per interval")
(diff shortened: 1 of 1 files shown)


Bojun-Seo:
Can you explain the differences between `-o` option and position argument `count` option?

And it will be helpful if you add some examples about
(1) What you cannot do without the `-o` option
(2) What you can do with the `-o` option

ShawnTuatara:
Count is how many times to loop through and output all the connections. The output lines reduces how many lines to show for each loop. When the output lines isn't specified then if there is more connections than screen / terminal lines the top of the list, and header lines, scrolls off the screen. In most cases users only want to see the top connections along with the header information.

Bojun-Seo:
Thanks for the explanation.

There is a `-T, --top` option in `syscount`, `argdist`, `memleak` and `ppchcalls` tools which is similar feature in this patch.
So I suggest to change the option name from `-o` to `-T` as other tool does.

yonghong-song:
Agree. option -T is more consistent with other tools.

PR #5118, opened 2024-10-07
fix(disksnoop): register to blk_mq_complete_request

saiintbrisson:
As mentioned in #4691, `disksnoop`, like `biosnoop`, rely on functions that can be inlined, like `blk_account_io_done`. This change adds a new function `blk_mq_complete_request` as a fallback in case that the others were inlined. I haven't changed this to support tracepoint because the tutorial guide would have to be considerably rearranged to make it follow a logical order, as the first mention of tracepoint is on lesson 12.

### DIFF
--- a/docs/tutorial_bcc_python_developer.md
+++ b/docs/tutorial_bcc_python_developer.md
@@ -224,22 +224,29 @@ void trace_completion(struct pt_regs *ctx, struct request *req) {
 	}
 }
 """)
-if BPF.get_kprobe_functions(b'blk_start_request'):
-        b.attach_kprobe(event="blk_start_request", fn_name="trace_start")
 b.attach_kprobe(event="blk_mq_start_request", fn_name="trace_start")
+if BPF.get_kprobe_functions(b'blk_start_request'):
+	b.attach_kprobe(event="blk_start_request", fn_name="trace_start")
+
 if BPF.get_kprobe_functions(b'__blk_account_io_done'):
     b.attach_kprobe(event="__blk_account_io_done", fn_name="trace_completion")
-else:
+elif BPF.get_kprobe_functions(b'blk_account_io_done'):
     b.attach_kprobe(event="blk_account_io_done", fn_name="trace_completion")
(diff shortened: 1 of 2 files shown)


docs/tutorial_bcc_python_developer.md
```diff
@@ -224,22 +224,29 @@ void trace_completion(struct pt_regs *ctx, struct request *req) {
 	}
 }
 """)
-if BPF.get_kprobe_functions(b'blk_start_request'):
-        b.attach_kprobe(event="blk_start_request", fn_name="trace_start")
 b.attach_kprobe(event="blk_mq_start_request", fn_name="trace_start")
+if BPF.get_kprobe_functions(b'blk_start_request'):
+	b.attach_kprobe(event="blk_start_request", fn_name="trace_start")
```
yonghong-song, line 229: Why the above change?

docs/tutorial_bcc_python_developer.md
```diff
@@ -224,22 +224,29 @@ void trace_completion(struct pt_regs *ctx, struct request *req) {
...
+
 if BPF.get_kprobe_functions(b'__blk_account_io_done'):
     b.attach_kprobe(event="__blk_account_io_done", fn_name="trace_completion")
-else:
+elif BPF.get_kprobe_functions(b'blk_account_io_done'):
     b.attach_kprobe(event="blk_account_io_done", fn_name="trace_completion")
+elif BPF.get_kprobe_functions(b'blk_mq_complete_request'):
+    b.attach_kprobe(event="blk_mq_complete_request", fn_name="trace_completion")
```
yonghong-song, line 236: Let us just use tracepoint block:block_rq_complete here. It is more stable.

yonghong-song:
for
> I haven't changed this to support tracepoint because the tutorial guide would have to be considerably rearranged to make it follow a logical order, as the first mention of tracepoint is on lesson 12

Let us just use tracepoint. It should be fine. It would be good that example can represent reasonable recent kernel internals.

PR #5119, opened 2024-10-08
libbpf-tools/profile: Remove unnecessary parameter

Bojun-Seo:
Parameter `freq` of `open_and_attach_perf_event` is not in use.  So remove it.


ekyooo (commented):
LGTM. Thank you.

PR #5120, opened 2024-10-08
libbpf-tools/profile: Check if `nr_cpus` is zero

Bojun-Seo:


libbpf-tools/profile.c
```diff
@@ -584,7 +584,7 @@ int main(int argc, char **argv)
 	libbpf_set_print(libbpf_print_fn);
 
 	nr_cpus = libbpf_num_possible_cpus();
-	if (nr_cpus < 0) {
+	if (nr_cpus <= 0) {
 		printf("failed to get # of possible cpus: '%s'!\n",
 		       strerror(-nr_cpus));
```
ekyooo, line 589: I agree with you that a result of 0 from libbpf_num_possible_cpus can be treated as a failure case. 

However, on line 589, strerror(0) will print 'Success'. Please check this. Thank you.

Bojun-Seo: I agree with you. So I changed the code. I hope you to review it again.
Thanks for the review.

ekyooo: How about using just one if statement for the failure case?

Bojun-Seo: I changed the code to use one if statement for the failure case.

libbpf-tools/profile.c
```diff
@@ -584,9 +584,9 @@ int main(int argc, char **argv)
 	libbpf_set_print(libbpf_print_fn);
 
 	nr_cpus = libbpf_num_possible_cpus();
-	if (nr_cpus < 0) {
+	if (nr_cpus <= 0) {
```
yonghong-song, line 587: The current code is correct. libbpf_num_possible_cpus() never returns 0. 
In libbpf, we have
```
int parse_cpu_mask_file(const char *fcpu, bool **mask, int *mask_sz)
{
        int fd, err = 0, len;
        char buf[128];

        fd = open(fcpu, O_RDONLY | O_CLOEXEC);
        if (fd < 0) {
                err = -errno;
                pr_warn("Failed to open cpu mask file %s: %d\n", fcpu, err);
                return err;
        }
        len = read(fd, buf, sizeof(buf));
        close(fd);      
        if (len <= 0) { 
                err = len ? -errno : -EINVAL;
                pr_warn("Failed to read cpu mask from %s: %d\n", fcpu, err);
                return err;
        }
        if (len >= sizeof(buf)) {
                pr_warn("CPU mask is too big in file %s\n", fcpu);
                return -E2BIG;
        }
        buf[len] = '\0';
                
        return parse_cpu_mask_str(buf, mask, mask_sz);
}
```

if there is no cpu, negative value (err number) will be return.

Bojun-Seo: I think adding this kind of error handling code can make the program more robust and less dependent on other modules. But as you said, since `libbpf_num_possible_cpus` never returns zero, I'll close this PR.

PR #5122, opened 2024-10-13
tools/tcpcong: Reduce code redundancy

kknjh:
1: Ports that do not meet the conditions in function "entry_state_update_func" will not be stored in the "start_ipv4" or "start_ipv6", and the "ret_state_update_func" function will not obtain values, so there is no need to perform port filtering again.

2: Sock_store can be directly deleted in function "ret_func".

### DIFF
--- a/tools/tcpcong.py
+++ b/tools/tcpcong.py
@@ -199,7 +199,6 @@ static int ret_state_update_func(struct sock *sk)
 {
     u64 ts, ts1;
     u16 family, last_cong_state;
-    u16 dport = 0, lport = 0;
     u32 tid = bpf_get_current_pid_tgid();
     process_key_t key = {0};
     bpf_get_current_comm(&key.comm, sizeof(key.comm));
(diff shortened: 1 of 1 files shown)


tools/tcpcong.py
```diff
@@ -274,7 +265,7 @@
         }
         start_ipv6.delete(&key);
     }
-    SOCK_STORE_DEL
```
yonghong-song, line 277: How about the following prog?
```
KRETFUNC_PROBE(tcp_fastretrans_alert, struct sock *sk)
{
    return ret_state_update_func(sk);
}
```

kknjh: if is_support_kfunc:
        bpf_text += kfunc_program
        bpf_text = bpf_text.replace('SOCK_STORE_DEF', '')
        bpf_text = bpf_text.replace('SOCK_STORE_ADD', '')
        bpf_text = bpf_text.replace('SOCK_STORE_DEL', '')
    else:
        bpf_text += kprobe_program
        bpf_text = bpf_text.replace('SOCK_STORE_DEF',
                       'BPF_HASH(sock_store, process_key_t, struct sock *);')
        bpf_text = bpf_text.replace('SOCK_STORE_ADD',
                       'sock_store.update(&key, &sk);')
        bpf_text = bpf_text.replace('SOCK_STORE_DEL',
                       'sock_store.delete(&key);')

It does not use SOCK_STORE, so it does not affect

yonghong-song: Okay, I see. Indeed this is the case.

PR #5134, opened 2024-11-07
tools/threadsnoop: Add -h option

Bojun-Seo:
Import argparser and add basic implementation for this. Now user can use `-h` option for threadsnoop tool.

### DIFF
--- a/tools/threadsnoop.py
+++ b/tools/threadsnoop.py
@@ -14,6 +14,20 @@
 
 from __future__ import print_function
 from bcc import BPF
+import argparse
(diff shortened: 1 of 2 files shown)


ekyooo (commented):
How about including the tool name in the commit title like the other commits?
"tools/threadsnoop: Add -h..." 

Could you also check this in your other commits?

Bojun-Seo:
@ekyooo 
Sounds reasonable. I'll change the description and commit message as you suggested. Thanks.

PR #5136, opened 2024-11-07
tools/softirqs: Print results in descending order

Bojun-Seo:
Users are usually interested in events that happen frequently and consume more time. So, print the results in descending order.


yonghong-song:
Make sense. Thanks!

PR #5153, opened 2024-11-22
Add fs mount point identification in 'tools/opensnoop.py'

dennisorlando:
Hi :)
Not sure if this addresses existing issues, a quick look didn't give me any result..
Anyways: there are a ton of filesystem roots in the linux filetree which mess up the output of opensnoop.py.
E.g. when working with files in /tmp, let's say `/tmp/averynicefile`, only `/averynicefile` gets displayed since that file is in the "root" `/` of a tmpfs file system.
This simple commit reconstructs the mountpoint information by passing the device identifier (dev_t) field of the BPF's PWD directory to the userspace python program, which retrieves the mountpoint information by grepping /proc/self/mountinfo

### DIFF
--- a/tools/opensnoop.py
+++ b/tools/opensnoop.py
@@ -103,12 +103,15 @@ bpf_text = """
 #ifdef FULLPATH
 #include <linux/fs_struct.h>
 #include <linux/dcache.h>
+#include <linux/fs.h>
+#include <linux/mount.h>
 
 #define MAX_ENTRIES 32
 
 enum event_type {
     EVENT_ENTRY,
     EVENT_END,
+    EVENT_MOUNTPOINT,
 };
 #endif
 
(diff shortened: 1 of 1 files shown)


dennisorlando:
Update: I just stumbled upon #4285 but that's literally 2 years old... + it's outdated: `struct vfsmount` doesn't appear to include `mnt_parent` anymore

xingfeng2510:
> Update: I just stumbled upon #4285 but that's literally 2 years old... + it's outdated: `struct vfsmount` doesn't appear to include `mnt_parent` anymore

Not outdate actually, `struct vfsmount` never includes `mnt_parent`, the fixup still works now, you should try it.

btw, unwind dentry by reading `/proc/self/mountinfo` frequently is inefficient,  #4285 referred to kernel implementation, do dentry unwinding in kernel.

Issue #5091, opened 2024-08-25
Build failure with LTO: examples/cpp/pyperf/PyPerfUtil.cc:29:20: error: ‘PYPERF_BPF_PROGRAM’ violates the C++ One Definition Rule [-Werror=odr]


hhoffstaette:
Gentoo found a problem when building with LTO, see [here for the bug](https://bugs.gentoo.org/938491).

I have reproduced this with 0.31.0 and despite the fact that PyPerf is "only" an example, I am reporting it since it often indicates header confusion/mixups.
```
[156/156] : && /usr/bin/x86_64-pc-linux-gnu-g++ -pipe -march=native -O2 -flto -fuse-linker-plugin -Wall  -fPIC -Wl,-O1,--as-needed,-z,now,-z,pack-relative-relocs -flto=auto -fuse-linker-plugin    -rdynamic examples/cpp/pyperf/CMakeFiles/PyPerf.dir/PyPerf.cc.o examples/cpp/pyperf/CMakeFiles/PyPerf.dir/PyPerfUtil.cc.o examples/cpp/pyperf/CMakeFiles/PyPerf.dir/PyPerfBPFProgram.cc.o examples/cpp/pyperf/CMakeFiles/PyPerf.dir/PyPerfLoggingHelper.cc.o examples/cpp/pyperf/CMakeFiles/PyPerf.dir/PyPerfDefaultPrinter.cc.o examples/cpp/pyperf/CMakeFiles/PyPerf.dir/Py36Offsets.cc.o -o examples/cpp/pyperf/PyPerf  -Wl,-rpath,/tmp/portage/dev-util/bcc-0.31.0/work/bcc-0.31.0_build/src/cc:/usr/lib/llvm/18/lib64  src/cc/libbcc.a  src/cc/libbcc.so.0.31.0  src/cc/libbcc-loader-static.a  -lelf  -lz  src/cc/frontends/clang/libclang_frontend.a  -Wl,--whole-archive  /usr/lib/llvm/18/lib64/libclang-cpp.so  /usr/lib/llvm/18/lib64/libLLVM.so.18.1  -Wl,--no-whole-archive  -lelf  -llzma  -lbpf  src/cc/usdt/libusdt-static.a  src/cc/api/libapi-static.a && :
[...]

PR #5160, opened 2024-11-26
examples/cpp/pyperf: Make external symbols match their definitions

hhoffstaette:
This is not just good hygiene but also silences valid complaints about mismatched definitions when building with LTO.

This is one part of fixing #5091


hhoffstaette:
This test failure has nothing to do with the proposed change..

chenhengqi (approved):

PR #5171, opened 2024-12-15
complete dump Cpu profiling

Jason-Ltiger:


yonghong-song:
Please give detailed explanation in 'Description' on why changes. This is important for reviewers to judge whether the change makes sense or not.

yonghong-song:
Please use English for comments.

PR #5180, opened 2025-01-01
tools/compactsnoop: Add aarch64 support

kknjh:


tools/compactsnoop.py
```diff
@@ -308,6 +309,14 @@ def zone_idx_to_str(idx):
                     {
                         0: "ZONE_NORMAL",
                         1: "ZONE_MOVABLE"
+                    },
+            'aarch64':
```
ekyooo, line 313: It appears that the notice '# NOTICE: consider only x86_64 servers' in line 299 is no longer valid.

yonghong-song: Thanks @ekyooo. the patch looks good. But let us remove the outdated line 299 as in the above comment.

kknjh: Already deleted

ekyooo:
LGTM, thanks

PR #5181, opened 2025-01-02
profile (libbpf): tool enhacements

glima:
This enhances the profile tool with two main tracks:

- offline symbol resolution capabilities and
- PSI threshould trigger mode, for always-running profiling capabilities

### DIFF
--- a/libbpf-tools/profile.c
+++ b/libbpf-tools/profile.c
@@ -66,7 +66,7 @@ struct fmt_t stacktrace_formats[] = {
 	{ true, ";", "", "-" }		/* folded */
 };
 
-#define pr_format(str, fmt)		printf("%s%s%s", fmt->prefix, str, fmt->suffix)
+#define pr_format(fd, str, fmt)		dprintf(fd, "%s%s%s", fmt->prefix, str, fmt->suffix)
 
 #define DSO_TIMER_INTERVAL_MS 100
 
@@ -93,6 +93,7 @@ int poll_fds[POLL_FD_COUNT];
 static struct env {
 	pid_t pids[MAX_PID_NR];
 	pid_t tids[MAX_TID_NR];
+	char *output_path;
 	bool user_stacks_only;
 	bool kernel_stacks_only;
 	bool addrs_only;
@@ -131,6 +132,7 @@ const char argp_program_doc[] =
 "    profile 5           # profile at 49 Hertz for 5 seconds only\n"
 "    profile -f          # output in folded format for flame graphs\n"
 "    profile -P          # run in an infinite loop, triggered by PSI watermarks\n"
+"    profile -o /tmp/f   # save output to that file (not stdout)\n"
 "    profile -p 185      # only profile process with PID 185\n"
 "    profile -L 185      # only profile thread with TID 185\n"
 "    profile -U          # only show user space stacks (no kernel)\n"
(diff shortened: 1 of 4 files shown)


libbpf-tools/profile.c
```diff
@@ -106,6 +107,7 @@ const char argp_program_doc[] =
 "    profile -p 185      # only profile process with PID 185\n"
 "    profile -L 185      # only profile thread with TID 185\n"
 "    profile -U          # only show user space stacks (no kernel)\n"
+"    profile -A          # only output addresses\n"
```
ekyooo, line 139: Can this not be achieved using the BPF_F_USER_BUILD_ID flag of the bpf_get_stackid helper function?

glima: OK, finally time for this. Not sure I follow, though, I'm afraid. bpf_get_stackid will only honor the following flags, it seems:
- BPF_F_SKIP_FIELD_MASK
- BPF_F_USER_STACK
- BPF_F_FAST_STACK_CMP
- BPF_F_REUSE_STACKID.

What did I miss? Are you talking about employing a different set of bpf helpers at the BPF program level, should that mode be in place, that would get to the same effect? But the parent profile.c wants to walk symbol resolution, through syms__map_addr(), regardless, doesn't it?

ekyooo: Oh, it's not bpf_get_stackid but bpf_get_stack. Sorry for the confusion. Please refer to:
https://youtu.be/20SO5thkvhI?list=PLbzoR-pLrL6oj1rVTXLnV7cOuetvjKn9q&t=145

glima: I see what you mean. Yeah, even for my uses, I intend to downstream-fork the symbol resolution phase to adapt to company-only flows of symbol resolution.

BUILD ID could indeed be one of the things serialized there. However, can that be an *addition* to this? This mode is already useful for some as-is, right? Not everybody will have BUILD-ID annotaded binaries for their systems, to begin with.

I will revisit and think about BUILD-ID addition when I refine it internally (making something more generic to add here). Is that sound? Or do you want me to focus on something I missed? 

Thanks.

glima: @ekyooo happy to accomodate, though, if you work with me deeper in your idea :) If my take is beneficial, gentle ping.

ekyooo: If there is no BUILD-ID, you can perform offline symbol resolving using the module name and module base offset that are outputted by the -v option.

However, since this information can vary with each build version, I think that using the build-id method could be more practical and has lower maintenance costs.

This is my opinion as a contributor, not a reviewer.

glima: I'm not sure where you are going here now, TBH. -v will only give you extra path and an offset

e.g.     0x00007fae6af9068b (/usr/lib64/firefox/libxul.so+0x479068b)

But the full context of start, end, file_offset etc is needed to actualy find the symbols (see my print_dso_info(), which serializes that). I just replicated the same c logic in python, for the offline resolver.

Build ID matching or not, more context than -v mode is needed.

glima: OK, not sure I'll have time to agree/fix on the suggestions here. Any interest in the rest of the patch (everything but the -A mode)?

PR #5189, opened 2025-01-16
tools/cpuunclaimed: Remove unused python methods

kiraskyler:
python methods `getpid, system` from os  module unused at tools/cpuunclaimed


ekyooo (commented):
LGTM for the change. 
However, a better title would be: 
'tools/cpuunclaimed: Remove unused python method imports'

thanks

kiraskyler:
> LGTM for the change. However, a better title would be: 'tools/cpuunclaimed: Remove unused python method imports'
> 
> thanks

Thank you for your correction. I have revised it according to your recommendation

PR #5193, opened 2025-01-17
syms: Initialize ModulePath::fd_ to invalid FD

danobi:
~ModulePath() does `if (fd_ > 0) close fd_;`. But ModulePath constructor has early return codepath for `!enter_ns` which does not initialize `fd_`.

So `fd_` can potentially have junk value. That junk value could be any FD downstream. bcc cannot be closing random FDs.

See https://dxuuu.xyz/flaky.html for a writeup for the implications.

### DIFF
--- a/src/cc/syms.h
+++ b/src/cc/syms.h
@@ -130,7 +130,7 @@ class ProcSyms : SymbolCache {
     // process by storing a file descriptor created from openat(2) if possible
     // if openat fails, falls back to process-dependent path with /proc/.../root
    private:
-    int fd_;
+    int fd_ = -1;
     std::string proc_root_path_;
     std::string path_;
 


danobi:
Here is a (basically) unconditional `enter_ns=false` call: https://github.com/iovisor/bcc/blob/fea5b15d745a5029055771dc370a7329e43627e0/src/cc/bcc_proc.c#L246-L255

4ast:
Nice catch and writeup.
cc @lenticularis39

danobi:
Can we get a patch release with this? This bug is really scary to have running around in the wild.

cc @yonghong-song

yonghong-song:
> Can we get a patch release with this? This bug is really scary to have running around in the wild.
> 
> cc @yonghong-song

Sure. Will do.

yonghong-song:
@danobi  See the new release tag v0.33.0 has been created.

danobi:
Thank you!

lenticularis39:
Nice catch, I must have missed that one during the multiple iterations of the implementation of `ModulePath`.

PR #5196, opened 2025-01-19
tools/opensnoop: Display mode for -e, --extended_fields

Rtoax:
When a program creates a file with mode=0000, it cannot even access the file
    itself. It would be helpful if we could track the mode value. so we can know
    who did it.
    
    Example:
    
        open("a.txt", O_WRONLY | O_EXCL | O_CREAT, 0000);
    
    Then:
    
        $ ls -l a.txt
        ----------. 1 rongtao rongtao 0 Jan 24 09:07 a.txt
        $ cat a.txt
        cat: a.txt: Permission denied
    
        $ sudo ./opensnoop.py -e
        PID    COMM               FD ERR FLAGS    MODE PATH
        673067 open                3   0 00000301 0000 a.txt
                                                  ^^^^

### DIFF
--- a/tools/opensnoop.py
+++ b/tools/opensnoop.py
@@ -99,6 +99,7 @@ for flag in args.flag_filter or []:
 bpf_text = """
 #include <uapi/linux/ptrace.h>
 #include <uapi/linux/limits.h>
+#include <linux/fcntl.h>
 #include <linux/sched.h>
 #ifdef FULLPATH
 #include <linux/fs_struct.h>
@@ -117,6 +118,7 @@ struct val_t {
     char comm[TASK_COMM_LEN];
     const char *fname;
     int flags; // EXTENDED_STRUCT_MEMBER
+    u32 mode; // EXTENDED_STRUCT_MEMBER
 };
 
 struct data_t {
@@ -130,6 +132,7 @@ struct data_t {
 #endif
     char name[NAME_MAX];
     int flags; // EXTENDED_STRUCT_MEMBER
+    u32 mode; // EXTENDED_STRUCT_MEMBER
 };
 
 BPF_PERF_OUTPUT(events);
@@ -158,6 +161,7 @@ int trace_return(struct pt_regs *ctx)
     data.ts = tsp / 1000;
     data.uid = bpf_get_current_uid_gid();
     data.flags = valp->flags; // EXTENDED_STRUCT_MEMBER
+    data.mode = valp->mode; // EXTENDED_STRUCT_MEMBER
     data.ret = PT_REGS_RC(ctx);
 
     SUBMIT_DATA
@@ -169,12 +173,15 @@ int trace_return(struct pt_regs *ctx)
 """
 
 bpf_text_kprobe_header_open = """
-int syscall__trace_entry_open(struct pt_regs *ctx, const char __user *filename, int flags)
+int syscall__trace_entry_open(struct pt_regs *ctx, const char __user *filename,
+                              int flags, u32 mode)
 {
 """
 
 bpf_text_kprobe_header_openat = """
-int syscall__trace_entry_openat(struct pt_regs *ctx, int dfd, const char __user *filename, int flags)
+int syscall__trace_entry_openat(struct pt_regs *ctx, int dfd,
+                                const char __user *filename, int flags,
+                                u32 mode)
 {
 """
 
@@ -183,6 +190,10 @@ bpf_text_kprobe_header_openat2 = """
 int syscall__trace_entry_openat2(struct pt_regs *ctx, int dfd, const char __user *filename, struct open_how *how)
 {
     int flags = how->flags;
+    u32 mode = 0;
+
+    if (flags & (O_CREAT | O_TMPFILE))
+        mode = how->mode;
 """
 
 bpf_text_kprobe_body = """
@@ -204,6 +215,7 @@ bpf_text_kprobe_body = """
         val.id = id;
         val.fname = filename;
         val.flags = flags; // EXTENDED_STRUCT_MEMBER
+        val.mode = mode; // EXTENDED_STRUCT_MEMBER
         infotmp.update(&id, &val);
     }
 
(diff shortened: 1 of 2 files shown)


ekyooo:
Could you include in the commit message when it is useful to output MODE information, which can be verified outside of the tool, within the tool itself?

Rtoax:
> Could you include in the commit message when it is useful to output MODE information, which can be verified outside of the tool, within the tool itself?

Thanks, I just add the extra information.

Rtoax:
Rebase to master

tools/opensnoop.py
```diff
@@ -217,8 +226,18 @@
...
     int flags = PT_REGS_PARM2(regs);
+    unsigned short mode = 0;
+
+    /**
+     * open(2): The mode argument must be supplied if O_CREAT or O_TMPFILE is
+     * specified in flags; if it is not supplied, some arbitrary bytes from
+     * the stack will be applied as the file mode.
+     */
+    if (flags & (O_CREAT | O_TMPFILE))
+        mode = PT_REGS_PARM3(regs);
```
Rtoax, line 237: I just fix an `random mode value` error, see open(2).

```
bcc/tools$ sudo ./opensnoop.py -e
PID    COMM               FD ERR FLAGS    MODE PATH
698289 pgrep               5   0 00000000 160043 /proc/uptime
698289 pgrep               5   0 00000000 160043 /proc/1329/stat
698289 pgrep               5   0 00000000 160043 /proc/1329/status
698289 pgrep               5   0 00000000 160043 /proc/1329/cmdline
698289 pgrep               5   0 00000000 160043 /proc/1329/cgroup
698289 pgrep              -1   2 00000000 160043 /proc/1329/ctty
635232 Memory       243   0 02000000 160225 /proc/698079/status
635232 Memory       243   0 02000000 160225 /proc/698079/stat
635232 Memory       243   0 02000000 160225 /proc/698079/status
```

tools/opensnoop.py
```diff
@@ -169,12 +173,15 @@
 """
 
 bpf_text_kprobe_header_open = """
-int syscall__trace_entry_open(struct pt_regs *ctx, const char __user *filename, int flags)
+int syscall__trace_entry_open(struct pt_regs *ctx, const char __user *filename,
+                              int flags, unsigned short mode)
```
yonghong-song, line 177: The mode for 'open' should be an int type, right?
See https://man7.org/linux/man-pages/man3/mode_t.3type.html

tools/opensnoop.py
```diff
@@ -245,13 +269,16 @@
...
+    unsigned short mode;
 
     bpf_probe_read_user(&how, sizeof(struct open_how), (struct open_how*)PT_REGS_PARM3(regs));
     flags = how.flags;
+    mode = how.mode;
 #else
 KRETFUNC_PROBE(FNNAME, int dfd, const char __user *filename, struct open_how __user *how, int ret)
 {
     int flags = how->flags;
+    unsigned short mode = how->mode;
```
yonghong-song, line 281: The flag and mode types in struct open_how:
```
struct open_how {
        __u64 flags;
        __u64 mode;
        __u64 resolve;
};
```
Also, should we check O_CREAT | O_TMPFILE flags before saving how->mode?
The same for other open_how usages.

yonghong-song (commented):
Although for struct open_how, it uses __u64 for flags and mode, but 
in reality, mode value should be still within 32bit range.

So I suggest to change 'mode' type to __u32 (unsigned int) instead of 'unsigned short'.

tools/opensnoop_example.txt
```diff
@@ -156,30 +156,31 @@ to the '-n' option.
...
-28512  sshd                3   0 02100000 /lib/x86_64-linux-gnu/libaudit.so.1
-28512  sshd                3   0 02100000 /lib/x86_64-linux-gnu/libpam.so.0
-28512  sshd                3   0 02100000 /lib/x86_64-linux-gnu/libselinux.so.1
-28512  sshd                3   0 02100000 /lib/x86_64-linux-gnu/libsystemd.so.0
-28512  sshd                3   0 02100000 /usr/lib/x86_64-linux-gnu/libcrypto.so.1.0.2
-28512  sshd                3   0 02100000 /lib/x86_64-linux-gnu/libutil.so.1
+# ./opensnoop.py -e
+PID    COMM               FD ERR FLAGS    MODE PATH
+312769 cgroupify           5   0 02204000 0000 .
+312769 cgroupify           5   0 00000000 0000 312814/cgroup.procs
```
ekyooo, line 162: The value 0000 for MODE has different meanings depending on the flag value. 
In the case where a file is created, 0000 has a valid meaning, but when it is not a file creation case, 0000 represents the default value of the mode variable.
How about using a distinct expression (such as a blank space or "n/a") instead of 0000 for clarity and user convenience?

Rtoax: Thanks, you're right, just fix it like:

```diff
+# ./opensnoop.py -e
+PID    COMM               FD ERR FLAGS    MODE PATH
+12458  open                3   0 02000000 n/a  /etc/ld.so.cache
+12458  open                3   0 02000000 n/a  /lib/x86_64-linux-gnu/libc.so.6
+12458  open                3   0 00000301 0664 tmp.txt
```

tools/opensnoop_example.txt
```diff
@@ -156,30 +156,27 @@ to the '-n' option.
...
-PID    COMM               FD ERR FLAGS    PATH
-28084  clear_console       3   0 00100002 /dev/tty
-28084  clear_console      -1  13 00100002 /dev/tty0
-28084  clear_console      -1  13 00100001 /dev/tty0
-28084  clear_console      -1  13 00100002 /dev/console
-28084  clear_console      -1  13 00100001 /dev/console
-28051  sshd                8   0 02100002 /var/run/utmp
-28051  sshd                7   0 00100001 /var/log/wtmp
+# ./opensnoop.py -e -f O_WRONLY -f O_RDWR
+PID    COMM               FD ERR FLAGS    MODE PATH
```
ekyooo, line 176: duplicated on line 177.

tools/opensnoop.py
```diff
@@ -453,7 +491,12 @@ def print_event(cpu, data, size):
                     event.comm, fd_s, err), nl="")
 
             if args.extended_fields:
-                printb(b"%08o " % event.flags, nl="")
+                # If neither O_CREAT nor O_TMPFILE is specified in flags, then
+                # mode is ignored, see open(2).
+                if event.mode == 0:
```
ekyooo, line 496: > but when it is not a file creation case, 0000 represents the default value of the mode variable.

This is the only case where I suggested using a distinct expression.
If you handle all the zero values this way, won't it fail to meet your requirements below?

> When a program creates a file with mode=0000, it cannot even access the file
itself. It would be helpful if we could track the mode value. so we can know
who did it.

Rtoax: Thanks, please review!

Rtoax:
Rebase to master  https://github.com/iovisor/bcc/commit/0dc42822d88e977d5fa58ad3f259b626e610b5c7

PR #5203, opened 2025-01-30
libbpf-tools/klockstat: Search for correct stack offset when printing

tohojo:
The klockstat BPF code contains a hard-coded offset (of 4) for the
captured stack entries, which is supposed to get rid of the topmost
entries that are just locking functions. However, when running the
klockstat tool on RHEL, this offset is not correct, leading to output
like the following:

                       mutex_lock+0x5    1.1 us        3     1.8 us       3.2 us
                       mutex_lock+0x5    825 ns        5     1.6 us       4.1 us
                       mutex_lock+0x5    1.1 us        4     1.5 us       4.3 us

                       mutex_lock+0x5   18.0 ms        3    23.0 ms      54.0 ms
                       mutex_lock+0x5    1.7 ms       15     2.2 ms      25.1 ms
                       mutex_lock+0x5    1.7 ms       15     2.2 ms      25.1 ms

To fix this, add logic in the userspace component of the utility to
walk the stack frames, skipping any entries that correspond to any of
the kprobe hooks that the tool attaches to. This fixes the output
on RHEL, without affecting other systems where the offset works
correctly.

Signed-off-by: Toke Høiland-Jørgensen <toke@toke.dk>

### DIFF
--- a/libbpf-tools/klockstat.c
+++ b/libbpf-tools/klockstat.c
@@ -31,6 +31,7 @@
 #include "trace_helpers.h"
 
 #define warn(...) fprintf(stderr, __VA_ARGS__)
+#define ARRAY_SIZE(x) (sizeof(x) / sizeof(*(x)))
 
 enum {
 	SORT_ACQ_MAX,
@@ -92,6 +93,29 @@ static const char program_doc[] =
 "  klockstat -s 6                # display 6 stack entries per lock\n"
 "  klockstat -P                  # print stats per thread\n"
 ;
+
+static const char *lock_ksym_names[] = {
+	"mutex_lock",
+	"mutex_lock_nested",
+	"mutex_lock_interruptible",
+	"mutex_lock_interruptible_nested",
+	"mutex_lock_killable",
+	"mutex_lock_killable_nested",
+	"mutex_trylock",
+	"down_read",
+	"down_read_nested",
+	"down_read_interruptible",
+	"down_read_killable",
+	"down_read_killable_nested",
+	"down_read_trylock",
+	"down_write",
+	"down_write_nested",
+	"down_write_killable",
+	"down_write_killable_nested",
+	"down_write_trylock",
+};
+
+static unsigned long lock_ksym_addr[ARRAY_SIZE(lock_ksym_names)];
 
 static const struct argp_option opts[] = {
 	{ "pid", 'p', "PID", 0, "Filter by process ID", 0 },
@@ -359,6 +383,38 @@ static char *symname(struct ksyms *ksyms, uint64_t pc, char *buf, size_t n)
 	return buf;
 }
 
+static int resolve_lock_ksyms(struct ksyms *ksyms)
+{
+	const struct ksym *ksym;
+	int i, j = 0;
+
+	for (i = 0; i < ARRAY_SIZE(lock_ksym_names); i++) {
+		ksym = ksyms__get_symbol(ksyms, lock_ksym_names[i]);
+		if (!ksym)
+			continue;
+
+		lock_ksym_addr[j++] = ksym->addr;
+	}
+
+	return !j;
+}
+
+static int find_stack_offset(struct ksyms *ksyms, struct stack_stat *ss)
+{
+	const struct ksym *ksym;
+        int i, j;
+
+	for (i = 0; i < PERF_MAX_STACK_DEPTH && ss->bt[i]; i++) {
+		ksym = ksyms__map_addr(ksyms, ss->bt[i]);
+
+		for (j = 0; j < ARRAY_SIZE(lock_ksym_addr) && lock_ksym_addr[j]; j++)
+			if (ksym->addr == lock_ksym_addr[j])
+				return i + 1;
+	}
+
+	return 0;
+}
+
 static char *print_caller(char *buf, int size, struct stack_stat *ss)
 {
 	snprintf(buf, size, "%u  %16s", ss->stack_id, ss->ls.acq_max_comm);
@@ -404,6 +460,7 @@ static void print_acq_header(void)
 static void print_acq_stat(struct ksyms *ksyms, struct stack_stat *ss,
 			   int nr_stack_entries)
 {
+	int offset = find_stack_offset(ksyms, ss);
 	char buf[40];
 	char avg[40];
 	char max[40];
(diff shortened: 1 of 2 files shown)


libbpf-tools/klockstat.c
```diff
@@ -93,6 +94,21 @@ static const char program_doc[] =
...
+	"mutex_lock_killable",
+	"mutex_trylock",
+	"down_read",
+	"down_read_interruptible",
+	"down_read_killable",
+	"down_read_trylock",
+	"down_write",
+	"down_write_killable",
+	"down_write_trylock",
+	"isig",
```
yonghong-song, line 109: Why adding 'isig' here?

libbpf-tools/klockstat.c
```diff
@@ -359,6 +375,26 @@ static char *symname(struct ksyms *ksyms, uint64_t pc, char *buf, size_t n)
...
+
+	for (i = 0; i < PERF_MAX_STACK_DEPTH; i++) {
+		if (!ss->bt[i])
+			break;
+
+		ksym  = ksyms__map_addr(ksyms, ss->bt[i]);
+
+		for (j = 0; j < ARRAY_SIZE(internal_ksym_names); j++)
+			if (!strcmp(ksym->name, internal_ksym_names[j]))
+				ret = i + 1;
```
yonghong-song, line 392: Should we break out if 'ret = i + 1' is not triggered?

libbpf-tools/klockstat.c
tohojo: Hmm, okay, so those 'isig' names were looking weird in the output when I was playing around with the filtering. So I assumed it was an artifact from the probing. But looking at the full traces:
```

                       mutex_lock+0x5    823 ns        1     823 ns       823 ns
                            isig+0xa5
       n_tty_receive_signal_char+0x16
      n_tty_receive_buf_standard+0xd1
                  __receive_buf+0x1fb
        n_tty_receive_buf_common+0x8f
    tty_port_default_receive_buf+0x3d
                  flush_to_ldisc+0x85
               process_one_work+0x1e5
                   worker_thread+0x50
                              Max PID 63507, COMM kworker/u97:1, Lock no-ksym (0xffffb3a36643f2a0)
```

...it seems 'isig' is actually a function name in the tty code; so I'll get rid of that :)

libbpf-tools/klockstat.c
tohojo: Hmm, no, I don't think so; the loop is trying to find the top-most instance of one of the names in the list, so to do that we'll keep matching while walking the whole stack. I.e., I'm not assuming that it's definitely an off-by-one thing here...

yonghong-song: @tohojo Could you show an stack example before and after the change for the above case? It will give a visible example so we can decide what is the proper way for the stack trace.

tohojo: Sorry for the delay. While investigating this further I discovered that this only affects older versions of RHEL (9.4 in this case), which outputs a kinda odd-looking upper stack trace, that also happens to be one entry longer before getting to the `mutex_lock` entry.

I'm including three different stack traces below (from `klockstat -L ftrace_lock`), obtained by modifying the klockstat in-kernel program to change the captured stack offset to 0 (and disable the userspace filtering in this PR).

In all cases, the change in this PR correctly trims the stack trace (even when it's not done in the kernel) to start just above mutex_lock. If we break out of the loop as you suggest, the filter only works if the mutex_lock entry is the topmost one. So I think it makes the most sense to keep looping; WDYT?

For RHEL-9.4:
```
         __init_scratch_end+0xe3361f5    387 ns        5     629 ns       1.9 us
          bpf_get_stackid_raw_tp+0x73
         __init_scratch_end+0xe3361f5
    virtio_balloon_driver_init+0x2042
                       mutex_lock+0x5
__modify_ftrace_direct_multi.part.0+0x1
      modify_ftrace_direct_multi+0x8c
          bpf_trampoline_update+0x183
      __bpf_trampoline_link_prog+0xcc
        bpf_trampoline_link_prog+0x27
        bpf_tracing_prog_attach+0x312
                    link_create+0x196
                      __sys_bpf+0x5fe
                   __x64_sys_bpf+0x1a
                   do_syscall_64+0x59
  entry_SYSCALL_64_after_hwframe+0x77
                              Max PID 23590, COMM klockstat, Lock ftrace_lock (0xffffffffb084fd00)
```

For RHEL-9.6:
```
         __init_scratch_end+0xc4cc92e    258 ns        5     278 ns       1.3 us
         __init_scratch_end+0xc4cc92e
         __init_scratch_end+0xc4ce752
                       mutex_lock+0x5
__modify_ftrace_direct_multi.part.0+0x1
      modify_ftrace_direct_multi+0x8c
          bpf_trampoline_update+0x192
      __bpf_trampoline_link_prog+0xcc
        bpf_trampoline_link_prog+0x27
        bpf_tracing_prog_attach+0x312
                    link_create+0x19f
                      __sys_bpf+0x8d2
                   __x64_sys_bpf+0x1a
                   do_syscall_64+0x5f
  entry_SYSCALL_64_after_hwframe+0x78
                              Max PID 23109, COMM klockstat, Lock ftrace_lock (0xffffffffb2658cc0)
```

Upstream kernel 6.13:
```
bpf_prog_6deef7357e7b4530_sd_fw_ingress    667 ns        5     796 ns       3.3 us
bpf_prog_6deef7357e7b4530_sd_fw_ingress
bpf_prog_6deef7357e7b4530_sd_fw_ingress
                       mutex_lock+0x9
   __modify_ftrace_direct.part.0+0x18
            modify_ftrace_direct+0xa9
          bpf_trampoline_update+0x266
      __bpf_trampoline_link_prog+0xdd
        bpf_trampoline_link_prog+0x32
        bpf_tracing_prog_attach+0x329
                     __sys_bpf+0x190f
                   __x64_sys_bpf+0x21
                   do_syscall_64+0x82
  entry_SYSCALL_64_after_hwframe+0x76
                              Max PID 427582, COMM klockstat, Lock ftrace_lock (0xffffffff96f43480)
```

libbpf-tools/klockstat.c
```diff
@@ -93,6 +94,20 @@ static const char program_doc[] =
 "  klockstat -P                  # print stats per thread\n"
 ;
 
+static const char *internal_ksym_names[] = {
```
ekyooo, line 97: Comparing the address of each symbol appears to be more cost-effective than comparing the symbol strings.

ekyooo (commented):
FYI, it seems that other methods are also being attempted to resolve the same issue. I'm not sure if it's in progress, but you might want to check it out:
https://www.youtube.com/watch?v=P3KgDhnYohY&list=PLTfHzETcek4S1XwF8KRXGDpr93j5lNnyc&t=526s

libbpf-tools/klockstat.c
```diff
@@ -359,6 +374,26 @@ static char *symname(struct ksyms *ksyms, uint64_t pc, char *buf, size_t n)
 	return buf;
 }
 
+static int find_stack_start(struct ksyms *ksyms, struct stack_stat *ss)
+{
+	const struct ksym *ksym;
+	int ret = 0;
+        int i, j;
+
+	for (i = 0; i < PERF_MAX_STACK_DEPTH; i++) {
```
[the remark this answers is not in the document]
tohojo: Hmm, I don't think running through a dozen or so entries on each print is a big deal? But we could certainly get rid of the filtering on the kernel side to cut down on this. @yonghong-song WDYT?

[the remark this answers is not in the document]
ekyooo: Does 'dozen on each print' take into account the maximum number of entries in the map (or env.nr_locks) and the maximum depth?

tohojo:
Alright, switched to pre-resolving the lock names and doing a symbol address comparison in the loop, and got rid of the offset from the kernel side. Also fixed `-c` - PTAL :)

PR #5216, opened 2025-02-19
libbpf-tools: signoop: Support real-time signals

Rtoax:
For example:

    $ sudo ./sigsnoop --name
    TIME     PID     COMM             SIG          TPID    RESULT
    10:09:34 431623  SIGRTMAX         SIGRTMIN+15  431623  0
    10:09:36 0       swapper/6        SIGRTMIN     2610    0
    10:13:46 0       swapper/11       SIGRTMIN     2610    0
    10:13:47 432111  SIGRTMAX         SIGRTMAX-10  432111  0

The test program:

```c
#include <stdio.h>
#include <signal.h>
#include <unistd.h>
#include <errno.h>

#define LIBC_SIGRTMIN		(__libc_current_sigrtmin())
#define LIBC_SIGRTMAX		(__libc_current_sigrtmax())

void handler(int sig, siginfo_t *si, void *context)
{
	printf("Received signal %d with value %d\n", sig, si->si_value.sival_int);
}

int main(int argc, char *argv[])
{
	struct sigaction sa;
	sigset_t mask;
	int rtoff = 20;

	printf("SIGRTMIN = %d, LIBC_SIGRTMIN = %d\n", SIGRTMIN, LIBC_SIGRTMIN);
	printf("SIGRTMAX = %d, LIBC_SIGRTMAX = %d\n", SIGRTMAX, LIBC_SIGRTMAX);

	/* Set up signal handler */
	sa.sa_flags = SA_SIGINFO;
	sa.sa_sigaction = handler;
	sigemptyset(&sa.sa_mask);
	sigaction(SIGRTMIN + rtoff, &sa, NULL);

	/* Block the signal temporarily */
	sigemptyset(&mask);
	sigaddset(&mask, SIGRTMIN + rtoff);
	sigprocmask(SIG_SETMASK, &mask, NULL);

	/* Send a real-time signal with additional data */
	union sigval sv;
	sv.sival_int = 520;
	sigqueue(getpid(), SIGRTMIN + rtoff, sv);

	/* Unblock the signal */
	sigprocmask(SIG_UNBLOCK, &mask, NULL);

	/* Wait for signal to be handled */
	printf("Press ctrl-C to end.\n");
	pause();

	return 0;
}
```

### DIFF
--- a/libbpf-tools/sigsnoop.c
+++ b/libbpf-tools/sigsnoop.c
@@ -63,6 +63,39 @@ static const char *sig_name[] = {
 	[29] = "SIGIO",
 	[30] = "SIGPWR",
 	[31] = "SIGSYS",
+	[32] = "SIGNAL-32", /* SIGRTMIN in kernel */
+	[33] = "SIGNAL-33",
+	[34] = "SIGNAL-34",
+	[35] = "SIGNAL-35",
+	[36] = "SIGNAL-36",
+	[37] = "SIGNAL-37",
+	[38] = "SIGNAL-38",
+	[39] = "SIGNAL-39",
+	[40] = "SIGNAL-40",
+	[41] = "SIGNAL-41",
+	[42] = "SIGNAL-42",
+	[43] = "SIGNAL-43",
+	[44] = "SIGNAL-44",
+	[45] = "SIGNAL-45",
+	[46] = "SIGNAL-46",
+	[47] = "SIGNAL-47",
+	[48] = "SIGNAL-48",
+	[49] = "SIGNAL-49",
+	[50] = "SIGNAL-50",
+	[51] = "SIGNAL-51",
+	[52] = "SIGNAL-52",
+	[53] = "SIGNAL-53",
+	[54] = "SIGNAL-54",
+	[55] = "SIGNAL-55",
+	[56] = "SIGNAL-56",
+	[57] = "SIGNAL-57",
+	[58] = "SIGNAL-58",
+	[59] = "SIGNAL-59",
+	[60] = "SIGNAL-60",
+	[61] = "SIGNAL-61",
+	[62] = "SIGNAL-62",
+	[63] = "SIGNAL-63",
+	[64] = "SIGNAL-64", /* SIGRTMAX */
 };
 
 const char *argp_program_version = "sigsnoop 0.1";
@@ -172,10 +205,10 @@ static void handle_event(void *ctx, int cpu, void *data, __u32 data_sz)
 	tm = localtime(&t);
 	strftime(ts, sizeof(ts), "%H:%M:%S", tm);
 	if (signal_name && e->sig < ARRAY_SIZE(sig_name))
-		printf("%-8s %-7d %-16s %-9s %-7d %-6d\n",
+		printf("%-8s %-7d %-16s %-12s %-7d %-6d\n",
 		       ts, e->pid, e->comm, sig_name[e->sig], e->tpid, e->ret);
 	else
-		printf("%-8s %-7d %-16s %-9d %-7d %-6d\n",
+		printf("%-8s %-7d %-16s %-12d %-7d %-6d\n",
 		       ts, e->pid, e->comm, e->sig, e->tpid, e->ret);
 }
 
@@ -248,7 +281,7 @@ int main(int argc, char **argv)
 		goto cleanup;
 	}
 
-	printf("%-8s %-7s %-16s %-9s %-7s %-6s\n",
+	printf("%-8s %-7s %-16s %-12s %-7s %-6s\n",
 	       "TIME", "PID", "COMM", "SIG", "TPID", "RESULT");
 
 	while (!exiting) {


libbpf-tools/sigsnoop.c
```diff
@@ -63,6 +63,39 @@ static const char *sig_name[] = {
 	[29] = "SIGIO",
 	[30] = "SIGPWR",
 	[31] = "SIGSYS",
+	[32] = "N/A",
+	[33] = "N/A",
+	[34] = "SIGRTMIN",
```
chenhengqi, line 68: The man page says that `SIGRTMIN` is not constant.
Maybe:
```c
[SIGRTMIN] = "SIGRTMIN"
[SIGRTMIN+1] = "SIGRTMIN+1"
```

Rtoax: Thanks, @chenhengqi , SIGRT{MIN,MAX} is not constant in glibc header, thus, I get these two constant values ​​in Makefile, add two new macro definitions, and replace them in sigsnoop.c. please review, thanks again.

libbpf-tools/sigsnoop.c
```diff
@@ -63,6 +63,54 @@ static const char *sig_name[] = {
 	[29] = "SIGIO",
 	[30] = "SIGPWR",
 	[31] = "SIGSYS",
+	/**
+	 * manual signal(7)
+	 * The Linux kernel supports a range of 33 different real-time signals,
+	 * numbered 32 to 64. However, the glibc POSIX threads implementation
+	 * internally uses two or three real-time signals, and adjusts the
+	 * value of SIGRTMIN suitably (to 34 or 35).
```
yonghong-song, line 71: I pulled in latest gilbc and find the SIGRTMIN is 34.
```
sysdeps/unix/sysv/linux/internal-signals.h:#define RESERVED_SIGRT  2
```

Could you provide a reference where SIGRTMIN == 35?

libbpf-tools/sigsnoop.c
```diff
@@ -63,6 +63,54 @@ static const char *sig_name[] = {
...
+	 */
+#if CONST_SIGRTMIN == 35
+	[32] = "SIGRTMIN-3",
+	[33] = "SIGRTMIN-2",
+	[34] = "SIGRTMIN-1",
+else
```
yonghong-song, line 77: Let us `#elif CONST_SIGRTMIN == 34` here and later add `#else` branch like
```
error "Unsupported CONST_SIGRTMIN value"
```
to trigger a compilation error is CONST_SIGRTMIN is not 34/35.

libbpf-tools/sigsnoop.c
Rtoax: Thanks, in signal(7) manual

   Real-time signals
       Starting  with  Linux 2.2, Linux supports real-time signals as originally defined
       in the POSIX.1b real-time extensions (and now  included  in  POSIX.1-2001).   The
       range of supported real-time signals is defined by the macros SIGRTMIN and SIGRT‐
       MAX.   POSIX.1-2001  requires  that an implementation support at least _POSIX_RT‐
       SIG_MAX (8) real-time signals.

       The Linux kernel supports a range of 33 different real-time signals, numbered  32
       to  64.  However, the glibc POSIX threads implementation internally uses two (for
       NPTL) or three (for LinuxThreads) real-time signals (see  pthreads(7)),  and  ad‐
       justs  the value of `SIGRTMIN suitably (to 34 or 35)`.  Because the range of avail‐
       able real-time signals varies according to  the  glibc  threading  implementation
       (and  this  variation can occur at run time according to the available kernel and
       glibc), and indeed the range of real-time signals  varies  across  UNIX  systems,
       programs  should  never  refer to real-time signals using hard-coded numbers, but
       instead should always refer to real-time signals using the  notation  SIGRTMIN+n,
       and include suitable (run-time) checks that SIGRTMIN+n does not exceed SIGRTMAX.

libbpf-tools/sigsnoop.c
Rtoax: Modified in https://github.com/iovisor/bcc/compare/5d2579656c281d6718ae00abce1facca6749ce6b..7c66533ced0291d97a578889fbe6ed4126d554d5, thanks.

libbpf-tools/sigsnoop.c
Rtoax: Found another POSIX libc https://git.musl-libc.org/cgit/musl/tree/src/signal/sigrtmin.c

libbpf-tools/Makefile
```diff
@@ -158,6 +158,18 @@ ifeq ($(USE_BLAZESYM),1)
 LDFLAGS += $(OUTPUT)/libblazesym.a -lrt -lpthread -ldl
 endif
 
+# In sigsnoop, SIGRTMAX and SIGRTMIN are used as array indexes. However,
+# SIGRTMAX and SIGRTMIN are not constants in glibc, so we add constants.
```
Rtoax, line 162: seems glibc is not totally obey the POSIX rule

image

libbpf-tools/sigsnoop.c
```diff
@@ -63,6 +63,56 @@ static const char *sig_name[] = {
...
+	[CONST_SIGRTMIN+2] = "SIGRTMIN+2",
+	[CONST_SIGRTMIN+3] = "SIGRTMIN+3",
+	[CONST_SIGRTMIN+4] = "SIGRTMIN+4",
+	[CONST_SIGRTMIN+5] = "SIGRTMIN+5",
+	[CONST_SIGRTMIN+6] = "SIGRTMIN+6",
+	[CONST_SIGRTMIN+7] = "SIGRTMIN+7",
+	[CONST_SIGRTMIN+8] = "SIGRTMIN+8",
+	[CONST_SIGRTMIN+9] = "SIGRTMIN+9",
+	[CONST_SIGRTMIN+10] = "SIGRTMIN+10",
+	[CONST_SIGRTMIN+11] = "SIGRTMIN+11",
```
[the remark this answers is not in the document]
Rtoax: The suggestion here is different from your suggestion above to "initialize the array with constants". Which method is better? thanks.

libbpf-tools/sigsnoop.c
```diff
@@ -63,6 +63,56 @@ static const char *sig_name[] = {
...
+	 * value of SIGRTMIN suitably (to 34 or 35).
+	 */
+#if CONST_SIGRTMIN == 35
+	[32] = "SIGRTMIN-3",
+	[33] = "SIGRTMIN-2",
+	[34] = "SIGRTMIN-1",
+#elif CONST_SIGRTMIN == 34
+	[32] = "SIGRTMIN-2",
+	[33] = "SIGRTMIN-1",
+#else
```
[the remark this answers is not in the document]
Rtoax: Still check `34` and `35` here, or add `-Wno-initializer-overrides` in Makefile?

libbpf-tools/Makefile
```diff
@@ -158,6 +158,18 @@ ifeq ($(USE_BLAZESYM),1)
...
+SIGRT_PROBE := '\#include <stdio.h>\n'
+SIGRT_PROBE += '\#include <signal.h>\n'
+SIGRT_PROBE += 'int main(void) {\n'
+SIGRT_PROBE += '	printf("-DCONST_SIGRTMIN=%d -DCONST_SIGRTMAX=%d\\n", SIGRTMIN, SIGRTMAX);\n'
+SIGRT_PROBE += '	return 0;\n'
+SIGRT_PROBE += '}'
+
+CFLAGS += $(shell printf '%b\n' $(SIGRT_PROBE) | \
+	$(CC) -xc -Wall - -o sig.out 2>&1 && ./sig.out && rm -f sig.out)
+
```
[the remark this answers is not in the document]
chenhengqi: You can use `SIGRTMIN` directly, just don't assume it is 34 or 35.

libbpf-tools/sigsnoop.c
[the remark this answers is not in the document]
chenhengqi: I mean something like:
```c
for (int i = SIGRTMIN; i <= SIGRTMAX; i++)
		sig_name[SIGMIN] = sprintf("SINRTMIN+%d", i - SIGRTMIN);
```

libbpf-tools/Makefile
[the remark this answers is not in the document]
Rtoax: SIGRTMIN is not compile-time const

```
  CC       sigsnoop.o
In file included from sigsnoop.c:12:
sigsnoop.c:83:10: error: nonconstant array index in initializer
   83 |         [SIGRTMIN] = "SIGRTMIN",
      |          ^~~~~~~~
sigsnoop.c:83:10: note: (near initialization for ‘sig_name’)
```

libbpf-tools/sigsnoop.c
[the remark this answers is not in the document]
Rtoax: Good idea.  Or should we use the kernel value directly without considering the occupancy of the RT signal by the user program?

```c
/**
 * - linux kernel SIGRTMIN is 32.
 * - POSIX SIGRTMIN is 34 or 35, see signal(7)
 *   - glibc is 32: sysdeps/unix/sysv/linux/internal-signals.h:#define RESERVED_SIGRT  2
 *   - musl-libc is 35: https://git.musl-libc.org/cgit/musl/tree/src/signal/sigrtmin.c
 */
#define LIBC_SIGRTMIN		(__libc_current_sigrtmin())
#define LIBC_SIGRTMAX		(__libc_current_sigrtmax())

/* glibc: sysdeps/unix/sysv/linux/internal-signals.h */
#if defined(__GLIBC__)
#define SIGCANCEL	__SIGRTMIN
#define SIGTIMER	SIGCANCEL
#define SIGSETXID	33
#endif

/* musl-libc: src/internal/pthread_impl.h */
#if defined(__MUSL__)
#define SIGTIMER 32
#define SIGCANCEL 33
#define SIGSYNCCALL 34
#endif
```

libbpf-tools/Makefile
Rtoax: @chenhengqi Hi, hengqi, SIGRTMIN is not compile-time const.

libbpf-tools/sigsnoop.c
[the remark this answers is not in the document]
yonghong-song: From signal 32 to 63, let us use the kernel value directly. Basically, signal 32 to 63 have names like "SIGNAL-32", ... "SIGNAL-63".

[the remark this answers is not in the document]
Rtoax: Thanks @chenhengqi @yonghong-song , sorry for the late reply, please review.

PR #5224, opened 2025-03-04
updated functionality and modified bashreadline script/txt

skipmcgee:
I have worked through the contributing scripts file and I think this pr meets all desired criteria. I tknow this script hasn't been contributed to in a long time, but I think these mods would help a user of this tool.

This pr adds the following functionality to the bashreadline script and includes appropriate updates in the example.txt:
1) Date updated from time to datetime
2) Adds UID of command which seems essential to script relevance
3) Adds json output for additional tool compatibility
4) Adds logging to a file output as an argument

### DIFF
--- a/tools/bashreadline.py
+++ b/tools/bashreadline.py
@@ -1,9 +1,9 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 #
-# bashreadline  Print entered bash commands from all running shells.
+# bashreadline Print entered bash commands from all running shells.
 #               For Linux, uses BCC, eBPF. Embedded C.
 #
-# USAGE: bashreadline [-s SHARED]
+# USAGE: bashreadline [-h] [-s [SHARED]] [-j] [-o [OUTPUT]]
 # This works by tracing the readline() function using a uretprobe (uprobes).
 # When you failed to run the script directly with error:
 # `Exception: could not determine address of symbol b'readline'`,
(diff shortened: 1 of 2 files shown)


skipmcgee:
@yonghong-song i looked through these tests but I don’t see much related to my PR - is there any action you want from me?

yonghong-song:
Tests are not related to your patch. I yet got time to investigate test failure. But will do it soon.

yonghong-song:
You didn't really provide strong use case for new features. For example, for UID, I don't know how it will useful for people just want to trace bashreadline thing. They can search pid in 'ps' to find all additional information is needed and I suspect the current output should be sufficient in most cases. There is no need to json output as well as the output from the script is likely the one user will check.

Again, please provide strong reason to add new options.

skipmcgee:
Thanks, that is a great point. Let me provide my intentions and use-cases. These are really just a couple of simple items that I think any user would prefer to have available as they make using the script more beneficial / easier.
1) Time changed to date+time: this helps a script user better correlate the events displayed by the script - for example if you output to a file it helps a user correlate the output to the appropriate day and time.
2) Added UID:
  a) this proves that the script does in fact accomplish its stated goal to monitor the commands input by all bash shells
  b) enables a user of the script to easily track those commands to multiple or individual users (without an extra call like ps)
3) Added JSON output:
  a) this enables a user to display the output more nicely via a utility like jq
  b) this enables the script output to be more easily consumed by other tools
4) file output: this is nice to have if you want to collect data from this script for a period of time. A user could always redirect the output, but a built-in file output option makes the script more user-friendly

skipmcgee:
@yonghong-song just wanted to follow up here if you have time to review my comments?

skipmcgee:
following up here @yonghong-song

PR #5226, opened 2025-03-08
libbpf-tools/statsnoop: Display syscall name

Rtoax:
Display the system call name is more helpful for analysis, like:

    $ sudo ./statsnoop
    PID     COMM                 RET  ERR  SYSCALL    PATH
    83459   code                 -1   2    statx      /home/rongtao/.config/Code/User/tasks.json
    2807    terminator           0    0    newfstatat /usr/share/icons
    2807    terminator           -1   2    newfstatat /home/rongtao/.local/share/flatpak/exports/share/pixmaps
    2807    terminator           -1   2    newfstatat /usr/local/share/pixmaps
    2807    terminator           0    0    newfstatat /usr/share/pixmaps
    83445   code                 -1   2    statx      /home/sdb/Git/5.10.13/.vscode/c_cpp_properties.json
    2525    pool                 0    0    newfstatat /etc/krb5.conf
    88285   ls                   0    0    statfs     /sys/fs/selinux
    88285   ls                   0    0    statx      listdir.c

### DIFF
--- a/libbpf-tools/statsnoop.c
+++ b/libbpf-tools/statsnoop.c
@@ -25,6 +25,7 @@ static volatile sig_atomic_t exiting = 0;
 static pid_t target_pid = 0;
 static bool trace_failed_only = false;
 static bool emit_timestamp = false;
+static bool emit_sysname = false;
 static bool verbose = false;
 
 const char *argp_program_version = "statsnoop 0.1";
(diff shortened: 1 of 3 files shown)


libbpf-tools/statsnoop.c
```diff
@@ -121,7 +130,8 @@ static void handle_event(void *ctx, int cpu, void *data, __u32 data_sz)
 		ts = (double)(e.ts_ns - start_timestamp) / 1000000000;
 		printf("%-14.9f ", ts);
 	}
-	printf("%-7d %-20s %-4d %-4d %-s\n", e.pid, e.comm, fd, err, e.pathname);
+	printf("%-7d %-20s %-4d %-4d %-10s %-s\n", e.pid, e.comm, fd, err,
```
chenhengqi, line 133: Should we add a new option for syscall name so that the default output is the same as the Python tool ?

Rtoax: Thanks, You're right, just add -s argument.

PR #5250, opened 2025-03-23
tcpdrop: fix unexpected exit if tcp_drop() missing

ioworker0:
This commit fixes an issue where tcpdrop would unexpectedly exit when the
tcp_drop() function is not found, even if the 'skb:kfree_skb' tracepoint
is traceable. Now, tcpdrop will continue running with a warning message
in that case.

### DIFF
--- a/tools/tcpdrop.py
+++ b/tools/tcpdrop.py
@@ -238,22 +238,23 @@ def print_ipv6_event(cpu, data, size):
         print("\t%s" % sym)
     print("")
 
+kfree_skb_traceable = False
+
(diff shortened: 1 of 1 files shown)


ioworker0:
Hi there,

After commit b93c0a9, tcpdrop is broken on Rocky Linux v9.4 (kernel: 5.14.0-503.22.1.el9.x86_64) ;(

To: @chenhengqi 
Cc: @seizethedave

seizethedave:
Ah, it looks like I misinterpreted the situation when that warning was logged.

PR #5252, opened 2025-03-23
tcpdrop: add support for dumping TCP drop reasons

ioworker0:
This commit adds support for dumping TCP drop reasons. It maps these reasons
to human-readable strings, making it easier to understand why TCP packets
are dropped. In cases where the drop reason is not supported, 'NOT_SPECIFIED'
will be displayed.

### DIFF
--- a/tools/tcpdrop.py
+++ b/tools/tcpdrop.py
@@ -17,6 +17,7 @@
 #
 # 30-May-2018   Brendan Gregg   Created this.
 # 15-Jun-2022   Rong Tao        Add tracepoint:skb:kfree_skb
+# 23-Mar-2025   Lance Yang      Dump drop reason
 
 from __future__ import print_function
 from bcc import BPF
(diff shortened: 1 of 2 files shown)


tools/tcpdrop.py
```diff
@@ -213,55 +219,179 @@
 else:
     bpf_text = bpf_text.replace('FILTER_NETNS', '')
 
+# the reasons of skb drop
+drop_reasons = {
+    0: "SKB_NOT_DROPPED_YET",
+    1: "SKB_CONSUMED",
+    2: "SKB_DROP_REASON_NOT_SPECIFIED",
```
chenhengqi, line 226: May we can omit the `SKB_DROP_REASIN_` prefix.

ioworker0: Good catch! 

I was thinking the same, so let's remove the prefix from that ;)

ioworker0: Done.

ioworker0:
Gently ping @chenhengqi

chenhengqi:
Just merged, thanks.

PR #5257, opened 2025-03-27
libbpf-tools/readahead: Fix low page usage count since v5.15

B106Roger:
Since Linux 5.15, `mark_page_accessed` is no longer called in `filemap_read`, which causes increased counts of unused page cache.
(For more detail, please refers to `mark_page_accessed` in Linux Patch1 and `filemap_read` in Linux Patch2.)

This patch attempts to attach `mark_accessed` to `folio_mark_accessed` if it is available; otherwise, it falls back to using `mark_page_accessed`.

Linux Patch1: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=76580b65 
Linux Patch2: https://github.com/torvalds/linux/commit/6020c204be997e3f5129839ff9c801800fb4336e#diff-6f046052d9af59ceb36616faffd8b1920c28195619803cfdb25df7271200ba5e

### DIFF
--- a/libbpf-tools/readahead.bpf.c
+++ b/libbpf-tools/readahead.bpf.c
@@ -80,8 +80,8 @@ int BPF_PROG(do_page_cache_ra_ret)
 	return 0;
 }
 
-SEC("fentry/mark_page_accessed")
-int BPF_PROG(mark_page_accessed, struct page *page)
+static __always_inline
+int mark_accessed(struct page *page)
 {
 	u64 *tsp, slot, ts = bpf_ktime_get_ns();
 	s64 delta;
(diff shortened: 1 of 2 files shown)


libbpf-tools/readahead.bpf.c
```diff
@@ -104,4 +104,16 @@ int BPF_PROG(mark_page_accessed, struct page *page)
 	return 0;
 }
 
+SEC("fentry/folio_mark_accessed")
+int BPF_PROG(folio_mark_accessed, struct folio *folio)
+{
+	mark_accessed(&folio->page);
```
yonghong-song, line 110: 'return mark_accessed(&folio->page)'? Similar issue below.

B106Roger: Thanks for pointing that out!
Please take a look and let me know if there's anything else I should address. Thanks again!

PR #5259, opened 2025-03-27
offcputime.py and profile.py can now be stopped by SIGUSR1 signal

jirka-h:
Hello,

I was missing an option to stop `offcputime.py ` and `profile.py` when the benchmark stops. 

I have added an option to send the SIGUSR1 signal to stop these applications. Here is the pseudocode showing a usage from Bash script:

```
# Start the profile tool in the background
/usr/share/bcc/tools/profile.py -adf > profile.folded &
PROFILE_PID=$!

# Run the benchmark
./run_benchmark   # Replace with the actual benchmark command

# Stop profile with SIGUSR1. It will shut down correctly, producing a full output. 
kill -USR1 $PROFILE_PID

# Wait for profile.py to finish
wait $PROFILE_PID
```

Please consider adding this feature. 

Thanks a lot
Jirka

### DIFF
--- a/tools/offcputime.py
+++ b/tools/offcputime.py
@@ -14,6 +14,7 @@
 # 13-Jan-2016	Brendan Gregg	Created this.
 # 27-Mar-2023	Rocky Xing      Added option to show symbol offsets.
 # 04-Apr-2023   Rocky Xing      Updated default stack storage size.
+# 27-Mar-2025   Jirka Hladky    Gracefully stops upon receiving the SIGUSR1 signal, making it suitable for automated workflows.
 
 from __future__ import print_function
 from bcc import BPF
@@ -301,6 +302,17 @@ def print_warn_event(cpu, data, size):
           % (event.pid, event.tgid, event.t_start, event.t_end),
           file=stderr)
 
+# Custom exception to break the sleep
+class SleepInterruptException(Exception):
+    pass
+
+# Signal handler to raise the exception when SIGUSR1 is received
+def signal_handler(signum, frame):
+    raise SleepInterruptException
+
+# Trap SIGUSR1
+signal.signal(signal.SIGUSR1, signal_handler)
+
 b["warn_events"].open_perf_buffer(print_warn_event)
 try:
     duration_ms = duration * 1000
@@ -310,7 +322,7 @@ try:
         if elapsed_ms >= duration_ms:
             break
         b.perf_buffer_poll(timeout=duration_ms - elapsed_ms)
-except KeyboardInterrupt:
+except (KeyboardInterrupt, SleepInterruptException):
     # as cleanup can take many seconds, trap Ctrl-C:
     signal.signal(signal.SIGINT, signal_ignore)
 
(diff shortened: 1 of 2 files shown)


yonghong-song:
I don't know what is the purpose for your benchmarking, maybe testing bpf overhead? For offcputime.py and profile.py, ctrl-C is needed to end the tool (after printing out the necessary information). You can just filter out these two tools when you are doing your benchmarking. I don't want to add these SIGUSR1 things for benchmarking purpose.

jirka-h:
Hi, these changes are not to benchmark the BPF overhead. The purpose is to **automate** the collection of call traces in the folded format for Flamegraphs generation. 

With `perf record`, I collect the call traces like this:
```
perd record --call-graph fp -o ${out_name}.perf -- ./test_binary
perf script -i "${out_name}.perf" | "${SCRIPT_DIR}/FlameGraph/stackcollapse-perf.pl" > "${out_name}.perf-folded"
"${SCRIPT_DIR}/FlameGraph/flamegraph.pl" --minwidth 1 "${out_name}.perf-folded" > "${out_name}.svg"
```
The point here is that I collect the traces just for the `test_binary`. Sometimes, I need to use `-a` to collect the traces from the whole system. The key idea is that it's fully automated, and you collect the traces only while `test_binary` is running.

The issue with `perf-record` is its significant overhead, which leads to the collection of vast amounts of data that must later be filtered. To address this, Brendan Gregg developed these scripts. However, the current limitation is that there is no way to use these tools from a fully automated script. You either need to specify the duration **in advance** or run it in the foreground, stopping the collection with Ctrl-C. Both approaches are impractical. 

SIGUSR1 is a simple method to stop the collection of call traces as soon as the test program ends. The perf-record command above can be rewritten as follows:

```
/usr/share/bcc/tools/profile.py -adf > profile.folded &
PROFILE_PID=$!

# Run the test
./test_binary   # Collect call traces for your binary

# Stop profile with SIGUSR1. This is to ensure the profiling will stop instantly after the test_binary finishes.
kill -USR1 $PROFILE_PID

# Wait for profile to finish
wait $PROFILE_PID

#Create Flamegraphs
"${SCRIPT_DIR}/FlameGraph/flamegraph.pl" --colors=java --minwidth 1 "profile.folded" > "${out_name}.svg"
```

Perhaps this approach is too cumbersome, and we should implement a method similar to `perf-record`, where you are allowed to specify the program to be launched and profiled and run it like this?

```
/usr/share/bcc/tools/profile.py -adf -o profile.folded -- ./test_binary
```

Please share your thoughts. I believe a method similar to `perf-record` would be the best solution. If you agree, I can prepare the PR. Nonetheless, I still think it would be valuable to have a way to stop `profile.py` when running in the background and output the data collected so far. Perhaps instead of using SIGUSR1, we should switch to the more standard signal SIGINT? At the moment, SIGINT stops profile.py instantly, and no output is generated. 

Thank you!
Jirka

jirka-h:
Hello,

any feedback on my last comment/proposal? 

Thanks
Jirka

seekstar:
Same issue here. It seems that offcputime doesn't respond to SIGINT if it is run with sudo in a script.

Update: It seems that SIGINT is handled specially in `sh`. I rewrote my previous `sh` script with python, and then SIGINT can gracefully interrupt offcputime. I think it would be better if offcputime can be gracefully interrupted with SIGTERM, so that we don't have to rely on python.

PR #5263, opened 2025-03-31
libbpf-tools: syscall_helpers.c: support loongarch64 and add more generic syscall

Rtoax:
Display syscall name on loongarch64, instead of fallback to '[unknown: %u]'.


chenhengqi (commented):
LGTM. Thanks.

PR #5264, opened 2025-04-01
libbpf-tools: syscall_helpers.c: Fix incorrect syscall name

Rtoax:
In [1] commit dc24158af21e ("libbpf-tools: syscount: Add syscall lookup table for arm64 and riscv"), an incorrect system call name was introduced, probably because __NR_syscalls was used directly to obtain the name "syscalls", while in fact it was a count.

Link: https://github.com/iovisor/bcc/commit/dc24158af21e [1]


chenhengqi:
cc @ismhong

chenhengqi:
cc @ismhong

ismhong:
LGTM.

Sorry for my carelessness, it seems I miss this item while generating table by script. Please check kernel source include/uapi/asm-generic/unistd.h [1]. Thanks @Rtoax to fix it.

[1]: https://github.com/torvalds/linux/blob/38fec10eb60d687e30c8c6b5420d86e8149f7557/include/uapi/asm-generic/unistd.h#L791

PR #5269, opened 2025-04-07
uprobe: Fix failed to attach BPF to uprobe error

kknjh:
./ugc <pid>
Sometimes errors may occur:
Traceback (most recent call last):
  File "/usr/share/bcc/tools/lib/./ugc", line 216, in <module>
    bpf = BPF(text=program, usdt_contexts=[usdt])
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/bpfcc/__init__.py", line 512, in __init__
    usdt_context.attach_uprobes(self, attach_usdt_ignore_pid)
  File "/usr/lib/python3.11/site-packages/bpfcc/usdt.py", line 209, in attach_uprobes
    bpf.attach_uprobe(name=binpath, fn_name=fn_name,
  File "/usr/lib/python3.11/site-packages/bpfcc/__init__.py", line 1417, in attach_uprobe
    raise Exception("Failed to attach BPF to uprobe")
Exception: Failed to attach BPF to uprobe

The error location is at IS_LIGHNED(ref_ctr_offset, sizeof(short)) of function __uprobe_register. This is because the input of the bpf_attach_uprobe function in libbpf.c has increased, but lib.bpf_attach_uprobe has not been modified,
resulting in ref_ctr_offset being a random number and potentially causing an error.

Fixes: 50f2009c7521 ("uprobe: Add ref_cnt_offset arg to bpf_attach_uprobe")

PR #5271, opened 2025-04-08
libbpf: Fix event name too long error

kknjh:
When the event name is too long, such as p:uprobes/p__proc_124566_root_usr_lib64_libruby_so_3_2_2_0xda3ac_124566_bcc_131967, an error will occur when executing the write function of function create_probe_event.

The kernel error path is
probes_write
  trace_parse_run_command
    create_or_delete_trace_uprobe
      trace_uprobe_create
        trace_probe_create
          __trace_uprobe_create
            traceprobe_parse_event_name
		else if (len >= MAX_EVENT_NAME_LEN)
Requires less than 64 bytes.

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -1365,11 +1365,11 @@ class BPF(object):
 
     def _get_uprobe_evname(self, prefix, path, addr, pid):
         if pid == -1:
-            return b"%s_%s_0x%x" % (prefix, self._probe_repl.sub(b"_", path), addr)
+            return b"%s_%s_0x%x" % (prefix, self._probe_repl.sub(b"_", os.path.basename(path)), addr)
         else:
             # if pid is valid, put pid in the name, so different pid
             # can have different event names
-            return b"%s_%s_0x%x_%d" % (prefix, self._probe_repl.sub(b"_", path), addr, pid)
+            return b"%s_%s_0x%x_%d" % (prefix, self._probe_repl.sub(b"_", os.path.basename(path)), addr, pid)
 
     def attach_uprobe(self, name=b"", sym=b"", sym_re=b"", addr=None,
             fn_name=b"", pid=-1, sym_off=0):


src/python/bcc/__init__.py
```diff
@@ -1365,11 +1365,11 @@ def sym_cb(sym_name, addr):
 
     def _get_uprobe_evname(self, prefix, path, addr, pid):
         if pid == -1:
-            return b"%s_%s_0x%x" % (prefix, self._probe_repl.sub(b"_", path), addr)
+            return b"%s_%s_0x%x" % (prefix, self._probe_repl.sub(b"_", os.path.basename(path)), addr)
```
chenhengqi, line 1368: What's your kernel version ?

BTW, the current libbpf implementation has the same issue.

kknjh: 6.6 Kernel and arm64 architecture.
```static int bpf_attach_probe(int progfd, enum bpf_probe_attach_type attach_type,
                            const char *ev_name, const char *config1, const char* event_type,
                            uint64_t offset, pid_t pid, int maxactive,
                            uint32_t ref_ctr_offset)
{
  int kfd, pfd = -1;
  char buf[PATH_MAX], fname[256], kprobe_events[PATH_MAX];
  bool is_kprobe = strncmp("kprobe", event_type, 6) == 0;

  if (maxactive <= 0)
    // Try create the [k,u]probe Perf Event with perf_event_open API.
    pfd = bpf_try_perf_event_open_with_probe(config1, offset, pid, event_type,
                                             attach_type != BPF_PROBE_ENTRY,
                                             ref_ctr_offset);

  // If failed, most likely Kernel doesn't support the perf_kprobe PMU
  // (e12f03d "perf/core: Implement the 'perf_kprobe' PMU") yet.
  // Try create the event using tracefs.
  if (pfd < 0) {
    if (create_probe_event(buf, ev_name, attach_type, config1, offset,
                           event_type, pid, maxactive) < 0)
      goto error;
      ......
      ```
The call to bpf_try_perf_event_open_with_probe should have succeeded, but it failed due to the issue with ref_ctr_offset(https://github.com/iovisor/bcc/pull/5269). However, the subsequent create_probe_event also failed, so this issue was discovered

> BTW, the current libbpf implementation has the same issue.

Thank you for your reminder. I have also submitted a patch to libbpf

yonghong-song:
Thanks. The change looks good to me.

PR #5274, opened 2025-04-09
Python-BCC: support fmod_ret attaching method, and add a useful tool for mptcp

Dwyane-Yan:
Multipath TCP (MPTCP) serves as an enhancement to the conventional TCP
protocol, enabling a single transport-layer connection to leverage
multiple network interfaces. This capability makes MPTCP advantageous
for applications requiring bandwidth consolidation, seamless failover
mechanisms, and more robust connectivity solutions.

Linux kernel starts to support MPTCP since v5.6, and it provides a
fmod_ret interface 'update_socket_protocol' to force applications using
MPTCP instead of TCP without modifyiing its code.

So these patches provide a tool named 'mptcpify' which can achieve this.
Using python-BCC is a more easy way for future development, so it is so
important to support 'fmod_ret' in python-BCC.

The first patch is suggested by Yonghong:
‘’‘
On Sun, 2024-08-25 at 21:05 -0700, Yonghong Song wrote:
...
\> Gang Yan, could you explore to add fmod_ret support in bcc? It should
\> be similar to kfunc/kretfunc support. I am happy to review your patches.
’‘’
when I try to make a commit to kernel. (Link attached below)
https://patchwork.kernel.org/project/netdevbpf/patch/tencent_1E619C9E44C8C4B2B713A0D6DD45B92BF70A@qq.com/

@matttbe from the MPTCP kernel team had a look at the new tool.

### DIFF
--- a/src/cc/libbpf.c
+++ b/src/cc/libbpf.c
@@ -1540,6 +1540,78 @@ static int find_member_by_name(struct btf *btf, const struct btf_type *btf_type,
   return 0;
 }
 
+static int find_val_in_enum(struct btf *btf, const struct btf_type *btf_type, const char * val_name)
+{
+  const struct btf_enum *enums = btf_enum(btf_type);
+  uint16_t vlen = BTF_INFO_VLEN(btf_type->info);
+  int found = 0;
+
+  for (uint16_t i = 0; i < vlen; i++) {
+    const struct btf_enum *e = &enums[i];
+    const char *member_name = btf__name_by_offset(btf, e->name_off);
+
+    if (member_name && strcmp(member_name, val_name) == 0) {
+      found = 1;
+      break;
+    }
+  }
+  return found;
+}
+
+static int find_enum_by_name(struct btf *btf, const char *enum_name, const char *val_name)
+{
+  const struct btf_type *btf_type;
+  int ret, btf_id;
+
+  btf_id = btf__find_by_name_kind(btf, enum_name, BTF_KIND_ENUM);
+  if (btf_id < 0) {
+    ret = -1;
+  } else {
+    btf_type = btf__type_by_id(btf, btf_id);
+    ret = find_val_in_enum(btf, btf_type, val_name);
+  }
+  return ret;
+}
+
+static int find_enum_by_anony(struct btf *btf, const char *val_name)
+{
+  __u32 nr_types;
+  int ret = 0;
+
+  nr_types = btf__type_cnt(btf);
+
+  for (int i = 1; i < nr_types; i++) {
+    const struct btf_type *btf_type = btf__type_by_id(btf, i);
+
+    if (!btf_type || btf_kind(btf_type) != BTF_KIND_ENUM || btf_type->name_off != 0)
+      continue;
+
+    ret = find_val_in_enum(btf, btf_type, val_name);
+    if (ret)
+      break;
+  }
+  return ret;
+}
+
+int kernel_enum_has_val(const char *enum_name, const char *val_name)
+{
+  struct btf *btf;
+  int ret;
+
+  btf = btf__load_vmlinux_btf();
+  ret = libbpf_get_error(btf);
+  if (ret)
+    return -1;
+
+  if (!enum_name || strlen(enum_name) == 0)
+    ret = find_enum_by_anony(btf, val_name);
+  else
+    ret = find_enum_by_name(btf, enum_name, val_name);
+
+  btf__free(btf);
+  return ret;
+}
+
 int kernel_struct_has_field(const char *struct_name, const char *field_name)
 {
   const struct btf_type *btf_type;
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -461,6 +461,7 @@ class BPF(object):
         self.raw_tracepoint_fds = {}
         self.kfunc_entry_fds = {}
         self.kfunc_exit_fds = {}
+        self.fmod_ret_fds = {}
         self.lsm_fds = {}
         self.perf_buffers = {}
         self.open_perf_events = {}
@@ -1157,6 +1158,12 @@ class BPF(object):
             return True
         return False
 
+    @staticmethod
+    def support_fmod_ret():
+        # Checking the existence of enum BPF_MODIFY_RETURN as the
+        # condition of support_fmod_ret.
+        return BPF.kernel_enum_has_val("bpf_attach_type", "BPF_MODIFY_RETURN") == 1
+
     def detach_kfunc(self, fn_name=b""):
         fn_name = _assert_is_bytes(fn_name)
         fn_name = BPF.add_prefix(b"kfunc__", fn_name)
@@ -1165,6 +1172,15 @@ class BPF(object):
             raise Exception("Kernel entry func %s is not attached" % fn_name)
         os.close(self.kfunc_entry_fds[fn_name])
         del self.kfunc_entry_fds[fn_name]
+
+    def detach_fmod_ret(self, fn_name=b""):
+        fn_name = _assert_is_bytes(fn_name)
+        fn_name = BPF.add_prefix(b"kmod_ret__", fn_name)
+
+        if fn_name not in self.fmod_ret_fds:
+            raise Exception("Fmod_ret func %s is not attached" % fn_name)
+        os.close(self.fmod_ret_fds[fn_name])
+        del self.fmod_ret_fds[fn_name]
 
     def detach_kretfunc(self, fn_name=b""):
         fn_name = _assert_is_bytes(fn_name)
@@ -1187,6 +1203,22 @@ class BPF(object):
         if fd < 0:
             raise Exception("Failed to attach BPF to entry kernel func")
         self.kfunc_entry_fds[fn_name] = fd
+        return self
+
+    def attach_fmod_ret(self, fn_name=b""):
+        fn_name = _assert_is_bytes(fn_name)
+        fn_name = BPF.add_prefix(b"kmod_ret__", fn_name)
+
+        if fn_name in self.fmod_ret_fds:
+            raise Exception("Fmod_ret func %s has been attached" % fn_name)
+
+        fn = self.load_func(fn_name, BPF.TRACING)
+        fd = lib.bpf_attach_kfunc(fn.fd)
+
+        if fd < 0:
+            raise Exception("Failed to attach BPF to fmod_ret kernel func")
+        self.fmod_ret_fds[fn_name] = fd
+
         return self
 
     def attach_kretfunc(self, fn_name=b""):
@@ -1830,6 +1862,8 @@ class BPF(object):
             self.detach_kretfunc(k)
         for k, v in list(self.lsm_fds.items()):
             self.detach_lsm(k)
+        for k, v in list(self.fmod_ret_fds.items()):
+            self.detach_fmod_ret(k)
 
         # Clean up opened perf ring buffer and perf events
         table_keys = list(self.tables.keys())
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -1252,6 +1252,12 @@ class BPF(object):
         field_name = _assert_is_bytes(field_name)
         return lib.kernel_struct_has_field(struct_name, field_name)
 
+    @staticmethod
+    def kernel_enum_has_val(enum_name, value_name):
+        enum_name = _assert_is_bytes(enum_name)
+        value_name = _assert_is_bytes(value_name)
+        return lib.kernel_enum_has_val(enum_name, value_name)
+
     def detach_tracepoint(self, tp=b""):
         """detach_tracepoint(tp="")
 
--- a/tools/mptcpify.py
+++ b/tools/mptcpify.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python
+#
+# mptcpify Make the applications to use MPTCP.
+#           For Linux, uses BCC, eBPF. Embedded C.
+#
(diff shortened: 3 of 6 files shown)


src/python/bcc/__init__.py
```diff
@@ -1157,6 +1158,12 @@ def support_lsm():
             return True
         return False
 
+    @staticmethod
+    def support_fmod_ret():
+        # It is not clear what can be used to check if 'fmod_ret'
+        # is supported by the kernel. Assuming checking kfunc is enough
```
chenhengqi, line 1164: This could be done by checking the existence of enum `BPF_MODIFY_RETURN`.
See `kernel_struct_has_field()` for reference.

Dwyane-Yan: Thanks a lot, that sounds a great idea, I'll try it later

tools/mptcpify.py
chenhengqi, line 1: I am not familiar with MPTCP. How to test this tool ?

matttbe: In theory, something like this:

```shell
$ curl https://check.mptcp.dev
You are not using MPTCP.
$ sudo python3 mptcpify.py -t curl &
$ curl https://check.mptcp.dev
You are using MPTCP.
```

The last command should be equivalent to `curl --mptcp https://check.mptcp.dev`. If even this last command doesn't work, that could be because MPTCP is somehow explicitly blocked in your network.

Dwyane-Yan: Here is another method.

A script named 'iperf.sh' to verify.
[iperf.txt](https://github.com/user-attachments/files/19682062/iperf.txt)
download first, and rename it to 'iperf.sh'

not using MPTCP:
> $ sudo ./iperf.sh

[ ID] Interval           Transfer     Bitrate
[  5]   0.00-1.00   sec  11.4 MBytes  95.3 Mbits/sec

using MPTCP
> $ sudo python3 mptcpify.py -t iperf3
> $ sudo ./iperf.sh

[ ID] Interval           Transfer     Bitrate
[  5]   0.00-1.00   sec  87.0 MBytes   729 Mbits/sec

You can see an obvious enhancement in column 'Bitrate'

matttbe: You can also list MPTCP sockets using `ss -Ma`, checking MIB counters with `nstat`, or packet traces. For more details about that: https://www.mptcp.dev/debugging.html

tools/mptcpify.txt
```diff
@@ -0,0 +1,21 @@
+Demonstrations of mptcpify, the Linux eBPF/bcc version.
```
yonghong-song, line 1: Please change mptcpify.txt to mptcpify_example.txt.

Please also add mptcpify_example.txt to README.md file at the root directory.

tools/mptcpify.txt
```diff
@@ -0,0 +1,21 @@
...
+USAGE message:
+
+usage: sudo python ./mptcpify.py [-h] -t TARGETS
+
+mptcpify try to force applications to use MPTCP instead of TCP
+
+options:
+  -h, --help            show this help message and exit
+  -t TARGETS, --targets TARGETS
+                        use ',' for multi targets, eg: 'iperf3,rsync'
```
yonghong-song, line 21: Just the above usage description is not enough. In the above, possible targets like iperf3/rsync, etc.

Please give detailed examples how it could be applied. In the above, you already have some examples like
```
    $ sudo ./iperf.sh
[ ID] Interval Transfer Bitrate
[ 5] 0.00-1.00 sec 11.4 MBytes 95.3 Mbits/sec

using MPTCP
    $ sudo python3 mptcpify.py -t iperf3
    $ sudo ./iperf.sh
[ ID] Interval Transfer Bitrate
[ 5] 0.00-1.00 sec 87.0 MBytes 729 Mbits/sec
```

Please have some examples like the above so people can easily have all the command lines to reproduce.

tools/mptcpify.py
yonghong-song: This is also a good example to be listed in mptcpify_example.txt file.

src/python/bcc/__init__.py
Dwyane-Yan: Sorry for the late reply.

'kernel_struct_has_field' seems not support type 'enum'. So I add a new interface named 'kernel_enum_has_val' in the first commit. Can you take a look?  

Thanks!

tools/mptcpify.txt
Dwyane-Yan: Hi, Yonghong:

So nice to receieve your suggestions.

I've made some modification in the new patch.  Please take a look.

By the way, I met a problem when compiling source code:

cmake ..
> CMake Warning at cmake/version.cmake:31 (message):
> Could not extract major/minor/patch from revision 128-NOTFOUND+00062fe4
> Call Stack (most recent call first):
> CMakeLists.txt:95 (include)

make -j:
> /usr/lib/python3/dist-packages/setuptools/dist.py:509: SetuptoolsDeprecationWarning: Invalid version: '128-NOTFOUND+00062fe4'.

Do you know why is '128-NOTFOUND+00062fe4'? How can I fix this problem.

Further more, I noticed the CI-test reports compiling error in fedora, but the error message (LUA BCC) seems not related to this PR.

src/cc/libbpf.c
```diff
@@ -1540,6 +1540,42 @@ static int find_member_by_name(struct btf *btf, const struct btf_type *btf_type,
...
+  const struct btf_type *btf_type;
+  struct btf *btf;
+  int ret, btf_id;
+
+  btf = btf__load_vmlinux_btf();
+  ret = libbpf_get_error(btf);
+  if (ret)
+    return -1;
+
+  btf_id = btf__find_by_name_kind(btf, enum_name, BTF_KIND_ENUM);
```
chenhengqi, line 1554: I think this does not handle unnamed enum correctly.

src/python/bcc/__init__.py
```diff
@@ -1157,6 +1158,12 @@ def support_lsm():
             return True
         return False
 
+    @staticmethod
+    def support_fmod_ret():
```
chenhengqi, line 1162: This method is not called anywhere.

src/python/bcc/__init__.py
```diff
@@ -1252,6 +1284,12 @@ def kernel_struct_has_field(struct_name, field_name):
         field_name = _assert_is_bytes(field_name)
         return lib.kernel_struct_has_field(struct_name, field_name)
 
+    @staticmethod
+    def kernel_enum_has_val(enum_name, value_name):
+        enum_name = _assert_is_bytes(enum_name)
+        value_name = _assert_is_bytes(value_name)
+        return lib.kernel_enum_has_val(enum_name, value_name)
```
chenhengqi, line 1291: This should be in the first commit instead. Type annotations for `lib.kernel_enum_has_val` is missing in libbcc.py.

tools/mptcpify.py
```diff
@@ -0,0 +1,92 @@
...
+    args_list = [t.strip() for t in args_str.targets.split(',')]
+else:
+    mode = 0
+
+if (not BPF.support_fmod_ret()):
+    print("Your kernel version is too old,"
+          " fmod_ret method is only support kernel v5.7 and later.")
+    exit()
+
+TASK_COMM_LEN = 18
```
chenhengqi, line 40: https://elixir.bootlin.com/linux/v6.13.7/source/include/linux/sched.h#L306

src/python/bcc/__init__.py
Dwyane-Yan: In fact, it is called in mptcpify.py:

if (not BPF.support_fmod_ret()):
    print("Your kernel version is too old,"
          " fmod_ret method is only support kernel v5.7 and later.")
    exit() 

before using fmod_ret, it should check ther kernel whether support 'fmod_ret'

src/python/bcc/__init__.py
Dwyane-Yan: Thanks for that suggestions. I'll fix in the next  series of patches.

tools/mptcpify.py
Dwyane-Yan: Thank you for your reminder. We will fix this issue later.

src/cc/libbpf.c
Dwyane-Yan: I tried an unnamed enum named 'YGYG', and the function can be returned in:
> if (btf_id < 0) {
>      ret = -1;
>      goto cleanup;
> }

Do you still have any other thoughts about this?

src/python/bcc/__init__.py
```diff
@@ -1252,6 +1284,12 @@ def kernel_struct_has_field(struct_name, field_name):
         field_name = _assert_is_bytes(field_name)
         return lib.kernel_struct_has_field(struct_name, field_name)
 
+     @staticmethod
+     def kernel_enum_has_val(enum_name, value_name):
+         enum_name = _assert_is_bytes(enum_name)
+         value_name = _assert_is_bytes(value_name)
+         return lib.kernel_enum_has_val(enum_name, value_name)
```
matttbe, line 1291: @Dwyane-Yan it looks like you changed the indentation by accident by adding an extra whitespace in the last version.

Dwyane-Yan: Hi Matt,

Thanks a lot, have fixed in the new commit!

src/cc/libbpf.c
chenhengqi: I mean something like:
```
enum {
	TCP_ESTABLISHED = 1,
	TCP_SYN_SENT,
	TCP_SYN_RECV,
	TCP_FIN_WAIT1,
	TCP_FIN_WAIT2,
	TCP_TIME_WAIT,
	TCP_CLOSE,
	TCP_CLOSE_WAIT,
	TCP_LAST_ACK,
	TCP_LISTEN,
	TCP_CLOSING,	/* Now a valid state */
	TCP_NEW_SYN_RECV,
	TCP_BOUND_INACTIVE, /* Pseudo-state for inet_diag */

	TCP_MAX_STATES	/* Leave at the end! */
};
```

How do you get `TCP_SYN_SENT` ?

Dwyane-Yan: Thanks, I got you.

After trying, I think the function below can work well in this situation. By using 'name_off == 0' to identifying the anonymous enum type.

```
static int find_enum_by_anony(struct btf *btf, const char *val_name)
{
  int ret = 0;
  __u32 nr_types;

  nr_types = btf__type_cnt(btf);

  for (int i = 1; i < nr_types; i++) {
    const struct btf_type *btf_type = btf__type_by_id(btf, i);

    if (!btf_type || btf_kind(btf_type) != BTF_KIND_ENUM)
      continue;

    if (btf_type->name_off != 0)
            continue;

    const struct btf_enum *enums = btf_enum(btf_type);
    uint16_t vlen = BTF_INFO_VLEN(btf_type->info);
    for (uint16_t i = 0; i < vlen; i++) {
      const struct btf_enum *e = &enums[i];
      const char *member_name = btf__name_by_offset(btf, e->name_off);

      if (member_name && strcmp(member_name, val_name) == 0) {
        ret = 1;
        break;
      }
    }
  }
  return ret;
}
```

But could you give a suggestion about which solution is more acceptable in BCC ? By adding a new function to split this two situation (anoymous?) or using the same interface, just searching all the 'enum type' in BTF to find the field, the difference attached below:
Solution1:
```
int kernel_enum_has_val(const char *enum_name, const char *val_name)
  ...
  if (enum_name)
    ret = find_enum_by_name(btf, enum_name, val_name);
  else
    ret = find_enum_by_anony(btf, val_name);
  ...
```
Solution2:
```
int kernel_enum_has_val(const char *val_name)
  ...
  search all the 'enum type' in BTF, to find 'val_name'
  ...
```

chenhengqi: I prefer option 1. With enum_name, it will be a little faster.

Dwyane-Yan:
Hi @chenhengqi and @yonghong-song,

It was truly delightful to receive your valuable suggestions regarding this pull request. I'm pleased to inform you that all the issues have now been addressed. Would you mind taking another look at it when you have a moment?

Thank you so much！

src/cc/libbpf.c
```diff
@@ -1540,6 +1540,77 @@ static int find_member_by_name(struct btf *btf, const struct btf_type *btf_type,
...
+  int ret;
+
+  btf = btf__load_vmlinux_btf();
+  ret = libbpf_get_error(btf);
+  if (ret)
+    return -1;
+  if (enum_name)
+    ret = find_enum_by_name(btf, enum_name, val_name);
+  else
+    ret = find_enum_by_anony(btf, val_name);
```
chenhengqi, line 1608: Looking at the Python API, I think this should be:
```c
  if (!enum_name || !enum_name[0])
    ret = find_enum_by_anony(btf, val_name);
  else
    ret = find_enum_by_name(btf, enum_name, val_name);
```

src/python/bcc/__init__.py
```diff
@@ -1252,6 +1252,12 @@ def kernel_struct_has_field(struct_name, field_name):
         field_name = _assert_is_bytes(field_name)
         return lib.kernel_struct_has_field(struct_name, field_name)
 
+    @staticmethod
+    def kernel_enum_has_val(enum_name, value_name):
+        enum_name = _assert_is_bytes(enum_name)
```
chenhengqi, line 1289: `enum_name` could be `None`, right ?

Dwyane-Yan:
Hi, @chenhengqi , 
Many tests has been skipped in CI, should it be triggered again?
Thanks!

chenhengqi:
Just merged. The CI need to be fixed. Will investigate.

Issue #5282, opened 2025-04-15
The issue of the PID consistently being -1 in the parameters of the `bcc_resolve_symname` method


guolifu:
1. When using the bpf.attach_uprobe method in the BCC API, I encounter an issue where bcc fails to function as intended when the first parameter is c.

`ERROR : attach error: Unable to attach uprobe for binary c symbol malloc addr 0 offset 0 using alloc_enter`

I found that the issue arose due to an unexpected value in the binary_path parameter of bpf_attach_uprobe.

2. After bypassing this issue, the specified PID process events still cannot be collected as expected

```
int res = bcc_resolve_symname(binary_path.c_str(), symbol.c_str(),
                                symbol_addr, -1, nullptr, &output);
```

PR #5283, opened 2025-04-15
Improved the implementation of bpf.attach_uprobe

guolifu:
Improved the implementation of bpf.attach_uprobe
﻿
Related issue: [#2147](https://github.com/iovisor/bcc/issues/5282)
﻿﻿
What this PR does:
﻿
This PR resolved the issue where bcc failed to operate as anticipated when utilizing the uprobe method via the c-API.
﻿
Why we need it:
﻿
This issue was blocked when I used BCC through C-API, so I suggest fixing it.

 This might not be the best way to resolve the issue.😉
﻿
﻿

### DIFF
--- a/src/cc/api/BPF.cc
+++ b/src/cc/api/BPF.cc
@@ -270,7 +270,7 @@ StatusTuple BPF::attach_uprobe(const std::string& binary_path,
 
   std::string module;
   uint64_t offset;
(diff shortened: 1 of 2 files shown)


yonghong-song:
Please provide a concrete reproducer to demonstrate what is the issue, and how your patch can resolve it. Thanks!

PR #5287, opened 2025-04-16
clang: Fix pointer dereference on big-endian machines

sumanthkorikkar:
When pointer is dereferenced, it ends up with bpf_probe_read_kernel(). However, the data type of source and destination might differ in size for genParamIndirectAssign() and this can cause endianness issue.

Fix the pointer dereference by adjusting offsets of source and destination, when there is a size disparity. Also, as a small optimization, perform memset() only on the required destination part, considering the endianness.

* This is specifically needed for the genParamIndirectAssign(), where size disparity could happen. With the patch, the code generated is shown below:

__attribute__((section(".bpf.fn.syscall__kill")))
int syscall__kill(struct pt_regs *ctx)
{
 struct pt_regs * __ctx = (void *)ctx->gprs[2];
 int tpid; BCC_PROBE_READ(&tpid, &__ctx->orig_gpr2, bpf_probe_read_kernel);
 int sig; BCC_PROBE_READ(&sig, &__ctx->gprs[3], bpf_probe_read_kernel);
...

Here ctx->gprs[3] is 64 bit and sig is 32 bit and there is a size disparity.

* However, in other usecases like VisitMemberExpr() etc, The code generated is as below:
if (flags & O_CREAT || (flags & O_TMPFILE) == O_TMPFILE) mode = ({
    typeof(__u64) _val; __builtin_memset(&_val, 0, sizeof(_val));
    bpf_probe_read_kernel(&_val, sizeof(_val), (void *)&how->mode);
_val; });

The code uses typeof(__u64) to create a temporary variable (_val) that matches the size of the data being read. This prevents mismatch between the source and destination sizes and hence changes are not necessary in these usecases.

The fix is based on linux commit:
2e2c6d3fb383 ("selftests/bpf: Fix test_core_reloc_mods on big-endian machines")

### DIFF
--- a/src/cc/frontends/clang/b_frontend_action.cc
+++ b/src/cc/frontends/clang/b_frontend_action.cc
@@ -363,6 +363,7 @@ ProbeVisitor::ProbeVisitor(ASTContext &C, Rewriter &rewriter,
   addrof_stmt_(nullptr), is_addrof_(false) {
   const char **calling_conv_regs = get_call_conv();
   cannot_fall_back_safely = (calling_conv_regs == calling_conv_regs_s390x || calling_conv_regs == calling_conv_regs_riscv64);
+  probe_read_func = cannot_fall_back_safely ? "bpf_probe_read_kernel" : "bpf_probe_read";
 }
 
 bool ProbeVisitor::assignsExtPtr(Expr *E, int *nbDerefs) {
@@ -776,6 +777,7 @@ BTypeVisitor::BTypeVisitor(ASTContext &C, BFrontendAction &fe)
     : C(C), diag_(C.getDiagnostics()), fe_(fe), rewriter_(fe.rewriter()), out_(llvm::errs()) {
   const char **calling_conv_regs = get_call_conv();
   cannot_fall_back_safely = (calling_conv_regs == calling_conv_regs_s390x || calling_conv_regs == calling_conv_regs_riscv64);
+  probe_read_func = cannot_fall_back_safely ? "bpf_probe_read_kernel" : "bpf_probe_read";
 }
 
 void BTypeVisitor::genParamDirectAssign(FunctionDecl *D, string& preamble,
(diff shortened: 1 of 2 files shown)


src/cc/frontends/clang/b_frontend_action.cc
```diff
@@ -1826,6 +1824,20 @@ void BFrontendAction::DoMiscWorkAround() {
   else {
     probefunc = "";
   }
+  probefunc += "#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__\n"
+    "#define BCC_PROBE_READ(dst, src, probe_read_func) ({ \\\n"
+    "  int __sz = sizeof(*(dst)) < sizeof(*(src)) ? sizeof(*(dst)) : sizeof(*(src)); \\\n"
+    "  __builtin_memset((char *)(dst) + __sz, 0, sizeof(*(dst)) - __sz); \\\n"
+    "  probe_read_func(dst, sizeof(*(dst)), src); \\\n"
```
yonghong-song, line 1831: Are lines 1828 and 1829 necessary?

Let us say *dst size is 4 and *src size is 8, __sz = 4, so effectively, the code is `probe_read_func(...)` as in the above.

Let us say *dst size is 8 and *src size is 4, __sz = 4. So we will have __builtin_memset((char *)dst + 4, 0, 4) which will set to 0 for last 4 bytes of *dst. But we will have probe_read_func(dst, 8, src) which will overwrite the last 4 bytes of *dst.

So I think we should just have probe_read_func(dst, sizeof(*(dst)), src).

src/cc/frontends/clang/b_frontend_action.cc
```diff
@@ -1826,6 +1824,20 @@ void BFrontendAction::DoMiscWorkAround() {
...
+    "})\n"
+    "#else\n"
+    "#define BCC_PROBE_READ(dst, src, probe_read_func) ({ \\\n"
+    "  int __sz = sizeof(*(dst)) < sizeof(*(src)) ? sizeof(*(dst)) : sizeof(*(src)); \\\n"
+    "  __builtin_memset((char *)(dst), 0, sizeof(*(dst)) - __sz); \\\n"
+    "  probe_read_func((char *)(dst) + sizeof(*(dst)) - __sz, __sz, \\\n"
+    "  (const char *)(src) + sizeof(*(src)) - __sz); \\\n"
```
yonghong-song, line 1838: I checked. This part is good for big endian.

src/cc/frontends/clang/b_frontend_action.cc
sumanthkorikkar: agree. Will do so.

sumanthkorikkar:
Added v2

PR #5310, opened 2025-05-18
Fix the issue where xfsslower cannot track write operations on files opened with the O_SYNC flag

KylinsecWuGeng:
This patch fixes an issue in xfsslower where write operation traces are overwritten when files are opened with O_SYNC flag. Currently, when a file is opened with O_SYNC, xfsslower fails to track write operations and only shows sync operations.
For example, when testing XFS I/O using this fio command:
[xxx@yyy ~]# fio --name=dsync-write --group_reporting --ioengine=psync --filename=./test --direct=1 --sync=1 --rw=write --bs=32k --size=4G --iodepth=1 --thread=1 --numjobs=1
And tracing with:
[xxx@yyy ~]# /usr/share/bcc/tools/xfsslower 0
The output only shows fsync operations without any write traces:
14:57:16 fio 189921 S 0 0 0.00 test
14:57:16 fio 189921 S 0 0 0.00 test
14:57:16 fio 189921 S 0 0 0.04 test
The root cause is that for O_SYNC files, xfs_file_write_iter calls xfs_file_fsync internally. This causes both trace_rw_entry and trace_fsync_entry to execute sequentially in the same process context.Since xfsslower uses BPF_HASH(entryinfo, u64, struct val_t) where entryinfo only uses pid and tid as keys, the entryinfo entry created by trace_rw_entry gets overwritten by trace_fsync_entry, making it impossible to track xfs_file_write_iter latency.
The solution introduces a new key structure:
struct key_t {
u64 ptid; // pid and tid
u32 type; // operation type
};
All kprobe trace functions in xfsslower have been updated to work with the new key structure. After this change, xfsslower correctly tracks both xfs_file_write_iter and xfs_file_fsync latencies:
15:22:13 fio 190028 S 0 0 0.00 test
15:22:13 fio 190028 W 32768 2293504 0.05 test
15:22:13 fio 190028 S 0 0 0.00 test
15:22:13 fio 190028 W 32768 2293536 0.05 test
See the code changes in this commit for implementation details.

### DIFF
--- a/tools/xfsslower.py
+++ b/tools/xfsslower.py
@@ -69,7 +69,7 @@ bpf_text = """
 #define TRACE_FSYNC     3
 
 struct key_t {
-    u64 ptid; //pid and tid
+    u64 id;
     u32 type;
 };
 
@@ -108,7 +108,7 @@ static int trace_rw_entry(struct pt_regs *ctx, struct kiocb *iocb, int type)
         return 0;
 
     struct key_t key = {};
-    key.ptid = id;
+    key.id = id;
     key.type = type;
 
     // store filep and timestamp by id
@@ -143,7 +143,7 @@ int trace_open_entry(struct pt_regs *ctx, struct inode *inode,
         return 0;
     
     struct key_t key = {};
-    key.ptid = id;
+    key.id = id;
     key.type = TRACE_OPEN;
 
     // store filep and timestamp by id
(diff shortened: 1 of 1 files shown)


tools/xfsslower.py
```diff
@@ -68,6 +68,11 @@
 #define TRACE_OPEN      2
 #define TRACE_FSYNC     3
 
+struct key_t {
+    u64 ptid; //pid and tid
```
yonghong-song, line 72: Let us use 'id' instead of 'ptid'. We already have something like 'u64 id = bpf_get_current_pid_tgid();' in the code.

KylinsecWuGeng: OK, I've updated the PR per your suggestions

PR #5313, opened 2025-05-19
libbpf-tools/memleak: Extract print_headers

ekyooo:
Extract the print_headers logic into a separate function to improve code maintainability and extensibility.

### DIFF
--- a/libbpf-tools/memleak.c
+++ b/libbpf-tools/memleak.c
@@ -251,6 +251,16 @@ static struct allocation *allocs;
 
 static const char default_object[] = "libc.so.6";
 
+static void print_headers()
+{
+	if (env.kernel_trace)
+		printf("Attaching to kernel allocators");
+	else
+		printf("Attaching to pid %d", env.pid);
+
+	printf(", Ctrl+C to quit.\n");
+}
+
 int main(int argc, char *argv[])
 {
 	int ret = 0;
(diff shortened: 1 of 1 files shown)


libbpf-tools/memleak.c
```diff
@@ -283,14 +293,10 @@ int main(int argc, char *argv[])
 		return 1;
 	}
 
-	if (!strlen(env.object)) {
-		printf("using default object: %s\n", default_object);
```
ekyooo, line 287: This information is only valid when the -p option is used and should be modified to be printed only in that case if needed. 
As its necessity is unclear and to align with memleak Python's header format, it's removed.
Please provide feedback if necessary.

libbpf-tools/memleak.c
```diff
@@ -283,14 +293,10 @@ int main(int argc, char *argv[])
...
+	if (!strlen(env.object))
 		strncpy(env.object, default_object, sizeof(env.object) - 1);
-	}
 
 	env.page_size = sysconf(_SC_PAGE_SIZE);
-	printf("using page size: %ld\n", env.page_size);
```
ekyooo, line 292: This information is only valid when the -p option is not used and should be modified to be printed only in that case if needed.
As its necessity is unclear and to align with memleak Python's header format, it's removed.
Please provide feedback if necessary.

yonghong-song:
The patch LGTM. I agree that the messages you deleted are not essential and should not appear in the output.

PR #5314, opened 2025-05-21
libbpf-tools/syscall_helers: Modify syscall_name to return error code

ekyooo:
Allow callers to handle errors via return value instead of parsing the buffer.


yonghong-song:
syscall_name() is used in syscount.c, we should make change there with the new signature.

ekyooo:
> syscall_name() is used in syscount.c, we should make change there with the new signature.

You're right, syscall_name() is used in syscount.c. However, since syscount.c does not use the return value of syscall_name, this change does not affect the behavior of syscount.c.

Ideally, I think it would be a slightly better design if syscall_name() does not store `[unknown: xx]` in buf upon failure, but instead only returns the failure result, leaving the handling of the failure (e.g., printing `[unknown: xx]`) to the caller.
If you want, I will modify it this way and apply it to syscount as well.

Thank you.

yonghong-song:
> Ideally, I think it would be a slightly better design if syscall_name() does not store `[unknown: xx]` in buf upon failure, but instead only returns the failure result, leaving the handling of the failure (e.g., printing `[unknown: xx]`) to the caller. If you want, I will modify it this way and apply it to syscount as well.

@ekyooo Yes, I agree with you. I also prefer that syscall_name() will not store '[unknown:xx]' in case of failure. Rather, it is the tool itself to decide what to print.

ekyooo:
Changed as agreed.

Thank you.

PR #5323, opened 2025-05-30
libbpf-tools: opensnoop: add full-path argument -F

Rtoax:
Like commit [1], support display fullpath, for example:

    $ cd /var/tmp/
    $ touch a.out

    $ sudo ./opensnoop -F
    106779 touch              3   0 /var/tmp/a.out
                                    ^^^^^^^^^^^^^^

Link: https://github.com/iovisor/bcc/commit/c110a4dd0c8f

### DIFF
--- a/libbpf-tools/opensnoop.bpf.c
+++ b/libbpf-tools/opensnoop.bpf.c
@@ -2,7 +2,9 @@
 // Copyright (c) 2019 Facebook
 // Copyright (c) 2020 Netflix
 #include <vmlinux.h>
+#include <bpf/bpf_core_read.h>
 #include <bpf/bpf_helpers.h>
+#include "compat.bpf.h"
 #include "opensnoop.h"
 
 #ifndef O_CREAT
@@ -16,6 +18,7 @@ const volatile pid_t targ_pid = 0;
 const volatile pid_t targ_tgid = 0;
 const volatile uid_t targ_uid = 0;
 const volatile bool targ_failed = false;
+const volatile bool full_path = false;
 
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
@@ -23,12 +26,6 @@ struct {
 	__type(key, u32);
 	__type(value, struct args_t);
 } start SEC(".maps");
-
-struct {
-	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
-	__uint(key_size, sizeof(u32));
-	__uint(value_size, sizeof(u32));
-} events SEC(".maps");
 
 static __always_inline bool valid_uid(uid_t uid) {
 	return uid != INVALID_UID;
@@ -116,7 +113,7 @@ int tracepoint__syscalls__sys_enter_openat2(struct syscall_trace_enter* ctx)
 static __always_inline
 int trace_exit(struct syscall_trace_exit* ctx)
 {
-	struct event event = {};
+	struct event *eventp;
 	struct args_t *ap;
 	uintptr_t stack[3];
 	int ret;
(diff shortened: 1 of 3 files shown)


libbpf-tools/opensnoop.bpf.c
```diff
@@ -149,7 +150,34 @@ int trace_exit(struct syscall_trace_exit* ctx)
...
+						(void *)dentry->d_name.name);
+			event.type = EVENT_ENTRY;
+			bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU,
+						&event, sizeof(event));
+
+			if (dentry == dentry->d_parent)
+				break;
+
+			dentry = dentry->d_parent;
+		}
```
yonghong-song, line 187: Each loop iteration will send out an event, which is too much and unnecessary. Please take a look at kernel implementation with bpf_probe_read_kernel_str which is the right fit.

https://github.com/torvalds/linux/blob/master/tools/testing/selftests/bpf/progs/profiler.inc.h#L204-L229

Rtoax: Thanks a lot, i'll try it. :)

Rtoax: Maybe opensnoop.py could use bpf_probe_read_kernel_str too, right? https://github.com/iovisor/bcc/pull/5324

yonghong-song: I didn't check the code yet, but yes bpf_probe_read_kernel_str should be better.

Rtoax: @yonghong-song Just pushed, please review, thanks

libbpf-tools/opensnoop.c
```diff
@@ -268,7 +267,13 @@ void handle_event(void *ctx, int cpu, void *data, __u32 data_sz)
 			printf("%08o %04o ", e.flags, e.mode);
 		sps_cnt += 9;
 	}
-	printf("%s\n", e.fname);
+	if (env.full_path) {
+		for (int depth = e.path_depth; depth >= 0; depth--)
+			printf("%s%s", "/\0" + (e.path_depth == 0),
```
yonghong-song: Could you add some comments what the above mean? It is not obvious just from the code.

libbpf-tools/opensnoop.bpf.c
```diff
@@ -129,29 +126,64 @@ int trace_exit(struct syscall_trace_exit* ctx)
...
+		void *payload = eventp->fname;
+
+
+		task = (struct task_struct *)bpf_get_current_task_btf();
+		dentry = BPF_CORE_READ(task, fs, pwd.dentry);
+
+		for (depth = 1, payload += NAME_MAX; depth < MAX_PATH_DEPTH; depth++) {
+			filepart_length =
+				bpf_probe_read_kernel_str(payload, NAME_MAX,
+						BPF_CORE_READ(dentry, d_name.name));
```
yonghong-song, line 169: We should check the return value of filepart_length. It is possible that bpf_probe_read_kernel_str may return a negative value in which case we should not explore any parent path.

libbpf-tools/opensnoop.c
```diff
@@ -268,7 +267,13 @@ void handle_event(void *ctx, int cpu, void *data, __u32 data_sz)
...
+					(char *)&e.fname[NAME_MAX * depth]);
```
yonghong-song: For the above loop, it is possible that we may not reach the root during path traversal. For example, MAX_PATH_DEPTH has been reached or bpf_probe_read_kernel_str() failed, in such cases, if it is not the leaf path, we should not add '/'. For example, for a correct path `/usr/lib/locale/en_US.UTF-8/LC_MESSAGES`, if the bpf only able to get `lib/locale/en_US.UTF-8/LC_MESSAGES`, we should not have proceeding `/`.

yonghong-song (commented):
Overall, look good. There are some minor comments.

libbpf-tools/opensnoop.c
Rtoax: Fixed, thanks.

libbpf-tools/opensnoop.bpf.c
Rtoax: Fixed, thanks.

PR #5324, opened 2025-05-30
tools/opensnoop: support mount parent full-path

Rtoax:
Like d_path()/prepend_path() [1] did, get mount parent full-path.

For example, /dev/sda mount on /home/sda:

    $ lsblk
    NAME        MAJ:MIN RM   SIZE RO TYPE MOUNTPOINTS
    sda           8:0    0   1.8T  0 disk /home/sda

and open file under /home/sda/:

    $ cd /home/sda/git-repos/tst-linux
    $ touch a.out

Old tracing, only get the mount root:

    $ sudo ./opensnoop.py -F
    114293 touch               3   0 /git-repos/test-linux/a.out

This patch get real full path:

    $ sudo ./opensnoop.py -F
    114110 touch               3   0 /home/sda/git-repos/test-linux/a.out
                                     ^^^^^^^^^

Link: https://github.com/torvalds/linux/blob/master/fs/d_path.c [1]

### DIFF
--- a/tools/opensnoop.py
+++ b/tools/opensnoop.py
@@ -103,6 +103,8 @@ bpf_text = """
 #include <linux/sched.h>
 #ifdef FULLPATH
 #include <linux/fs_struct.h>
+#include <linux/fs.h>
+#include <linux/mount.h>
 #include <linux/dcache.h>
 
 #define MAX_ENTRIES 32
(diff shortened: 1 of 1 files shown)


Rtoax:
Perhaps we need to first address the issue of multiple processing of events caused by the full-path patch, which was introduced in c110a4dd0c8f8e15e3107f3a0807683a81657cbf.  Each loop iteration will send out an event, which is too much and unnecessary.  cc @yonghong-song

yonghong-song:
Sounds good. Let us resolve https://github.com/iovisor/bcc/pull/5323 first and then deal with this patch.

Rtoax:
> Perhaps we need to first address the issue of multiple processing of events caused by the full-path patch, which was introduced in [c110a4d](https://github.com/iovisor/bcc/commit/c110a4dd0c8f8e15e3107f3a0807683a81657cbf). Each loop iteration will send out an event, which is too much and unnecessary. cc @yonghong-song

Close this PR and fix this first.

Issue #4843, opened 2023-12-14
Detaching uprobes with invalid `/proc/fs` binary path after process termination


benkilimnik:
Say we have a uprobe attached to a binary accessed via `/proc/<other-program-pid>/root/normal/path/to/binary`. When the process terminates, the pid gets cleaned up so that the `/proc/fs` filepath is no longer valid. To clean up the dangling FD, we need to call `detach_uprobe`, but this fails because [`check_binary_symbol`](https://github.com/iovisor/bcc/blob/8bc151f94e10be5e922f2972ceb1e319a1ec9d38/src/cc/api/BPF.cc#L516C7-L516C7) can't read the file.

Is there a way to make the detach API work with proc/fs based file paths? Do we need a mount-namespace aware file path resolver or is there an easier way?

benkilimnik:
Another way of phrasing this question: how do we detach uprobes for short-lived processes/containers?

etep:
Could a BCC maintainer please comment on this?

It would be neat to have BCC track the open probes so that the binary can be removed from the file system and the probe can still be detached. For example, BCC could return a probe descriptor which internally could map to the open FDs. When the user calls detach probe, internally BCC uses the probe descriptor to find the aforesaid FDs and then closes them.

yonghong-song:
For uprobe, current bcc has a mapping 'ev_name -> fd'. When detaching happens, bcc will follow 'ev_name -> fd' mapping and find the 'fd' and do detaching. I need to double check whether 'fd' (created at 'attach'/perf_event_open point) is still valid or not for short-lived process. Do you have an example to show the issue?

etep:
We will craft an example. Thank you!

aayush-ap:
any update on this issue ?

PR #5325, opened 2025-06-03
Added functionality to detach all uprobes for a binary

aayush-ap:
Added a new method in BCC

`StatusTuple BPF::detach_all_uprobes_for_binary(const std::string& binary_path)`

This method detaches all uprobe probes that are associated with the given binary path. It works by matching the sanitized binary path against the event names of all currently attached uprobes, and detaches any that match. Importantly, this function does not check if the binary file still exists on disk; it simply removes all matching uprobes from the internal tracking and from the kernel.

This is useful for cleaning up probes that may otherwise remain attached if the binary is deleted or moved, or if you want to forcibly remove all uprobes for a specific binary regardless of its current presence on the filesystem.

#### Automatic Cleanup When a Binary is Deleted:
If a monitored binary is removed from the filesystem, any uprobes attached to it will remain active in the kernel unless explicitly detached. This function allows you to clean up all such probes, preventing resource (fd) leaks and potential errors.
####  Bulk Detachment:
When updating or replacing a binary, you may want to remove all associated uprobes before re-attaching new ones. This function provides a simple way to do so without needing to track individual probe details.

### Example:
Suppose you are tracing a user-space binary /proc/<program-pid>/exe and that binary is deleted still you can call:

`bpf.detach_all_uprobes_for_binary("/proc/<program-pid>/exe");`

This will remove all uprobes associated with stoped program-pid , even if the file no longer exists, ensuring no stale probes are left behind.

With this functionality https://github.com/iovisor/bcc/issues/4843 will fixed

### DIFF
--- a/src/cc/api/BPF.cc
+++ b/src/cc/api/BPF.cc
@@ -528,6 +528,35 @@ StatusTuple BPF::detach_uprobe(const std::string& binary_path,
   return StatusTuple::OK();
 }
 
+// Detach all uprobes associated with the given binary path.
+// This function cleans up all matching uprobes without checking if the binary file exists.
+// It simply matches the sanitized binary path in the event names and detaches them.
+
+StatusTuple BPF::detach_all_uprobes_for_binary(const std::string& binary_path) {
+  bool has_error = false;
+  std::string error_msg;
+  std::vector<std::string> to_detach;
+
+  // Sanitize the binary path as used in event names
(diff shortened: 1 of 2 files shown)


aayush-ap:
@yonghong-song @drzaeus77 @4ast @davemarchevsky 
can anyone of you please review this PR.

yonghong-song:
> @yonghong-song @drzaeus77 @4ast @davemarchevsky can anyone of you please review this PR.

LGTM. Thanks!

PR #5328, opened 2025-06-10
libbpf-tools: runqslower: add -c option to trace specified comm

chudihuang:


libbpf-tools/runqslower.bpf.c
```diff
@@ -26,8 +27,19 @@ struct {
 	__uint(value_size, sizeof(u32));
 } events SEC(".maps");
 
+static __always_inline bool comm_allowed(const char *comm)
+{
+        int i;
+
+        for (i = 0; i < TASK_COMM_LEN && targ_comm[i] != '\0'; i++) {
```
ekyooo: How about using bpf_strncmp provided by the BPF helper?

libbpf-tools/runqslower.bpf.c
```diff
@@ -37,6 +49,8 @@ static int trace_enqueue(u32 tgid, u32 pid)
 		return 0;
 	if (targ_pid && targ_pid != pid)
 		return 0;
+	  if (!comm_allowed(comm))
```
ekyooo: Please check the indentation.

libbpf-tools/runqslower.c
```diff
@@ -64,6 +68,11 @@ static error_t parse_arg(int key, char *arg, struct argp_state *state)
 	case 'v':
 		env.verbose = true;
 		break;
+	case 'c':
+                env.comm = arg;
+                len = strlen(arg) + 1;
+                env.comm_len = len > TASK_COMM_LEN ? TASK_COMM_LEN : len;
```
ekyooo: I think we need to inform the tool's users about the COMM length limit. If only part of the user's input argument is used, that information should also be conveyed so that users can predict the tool's behavior."

libbpf-tools/runqslower.bpf.c
```diff
@@ -99,13 +116,18 @@ int BPF_PROG(sched_switch, bool preempt, struct task_struct *prev, struct task_s
 SEC("raw_tp/sched_wakeup")
 int BPF_PROG(handle_sched_wakeup, struct task_struct *p)
 {
-	return trace_enqueue(BPF_CORE_READ(p, tgid), BPF_CORE_READ(p, pid));
+	char comm[TASK_COMM_LEN] = {0};
```
ekyooo: A blank line after the declaration section, please.

libbpf-tools/runqslower.bpf.c
```diff
@@ -99,13 +116,18 @@ int BPF_PROG(sched_switch, bool preempt, struct task_struct *prev, struct task_s
...
+	BPF_CORE_READ_STR_INTO(&comm, p, comm);
+	return trace_enqueue(BPF_CORE_READ(p, tgid), BPF_CORE_READ(p, pid), comm);
 }
 
 SEC("raw_tp/sched_wakeup_new")
 int BPF_PROG(handle_sched_wakeup_new, struct task_struct *p)
 {
-	return trace_enqueue(BPF_CORE_READ(p, tgid), BPF_CORE_READ(p, pid));
+
+        char comm[TASK_COMM_LEN] = {0};
```
ekyooo: A blank line after the declaration section, please.

libbpf-tools/runqslower.c
```diff
@@ -53,7 +57,7 @@ static const struct argp_option opts[] = {
 
 static error_t parse_arg(int key, char *arg, struct argp_state *state)
 {
-	static int pos_args;
+	static int pos_args,len;
```
ekyooo: Rather than grouping it with unrelated variables, it would be better to declare it on a separate line.

libbpf-tools/runqslower.c
ekyooo: Personally, I think I'd use the MIN macro, but I'll leave it to your discretion.

libbpf-tools/runqslower.c
```diff
@@ -64,6 +68,11 @@ static error_t parse_arg(int key, char *arg, struct argp_state *state)
 	case 'v':
 		env.verbose = true;
 		break;
+	case 'c':
+                env.comm = arg;
```
ekyooo: Would you like to check if it is safer to use strdup for string-type arguments?

libbpf-tools/runqslower.c
```diff
@@ -20,6 +20,8 @@ static volatile sig_atomic_t exiting = 0;
 struct env {
 	pid_t pid;
 	pid_t tid;
+	int comm_len;
```
ekyooo: How about using TASK_COMM_LEN instead of this variable?

libbpf-tools/runqslower.bpf.c
chudihuang: The comm_allowed function is designed to match processes whose command name prefix matches the string specified by the -c option. However, it appears that bpf_strncmp may not be well-suited for this particular prefix-matching scenario.

chudihuang:
@ekyooo
I have addressed the review comments and fixed the code. Please re-review:
- Fixed variable declaration/indentation issues
- Added length limit warnings for comm parameters

libbpf-tools/runqslower.c
```diff
@@ -33,18 +36,20 @@ const char *argp_program_bug_address =
...
 "    runqslower 1000    # trace latency higher than 1000 us\n"
 "    runqslower -p 123  # trace pid 123\n"
 "    runqslower -t 123  # trace tid 123 (use for threads only)\n"
+"    runqslower -c bash # trace bash only (max comm name length is 15 chars)\n"
 "    runqslower -P      # also show previous task name and TID\n";
 
 static const struct argp_option opts[] = {
 	{ "pid", 'p', "PID", 0, "Process PID to trace", 0 },
 	{ "tid", 't', "TID", 0, "Thread TID to trace", 0 },
+	{ "comm",  'c', "COMM",  0, "Trace this comm only", 0 },
```
ekyooo: If this patch is intended to "filter by process name prefix," it is not consistent with the option description.

libbpf-tools/runqslower.bpf.c
```diff
@@ -26,8 +27,19 @@ struct {
...
+	for (i = 0; i < TASK_COMM_LEN && targ_comm[i] != '\0'; i++) {
+		if (comm[i] != targ_comm[i])
+			return false;
+	}
+	return true;
+}
+
 /* record enqueue timestamp */
-static int trace_enqueue(u32 tgid, u32 pid)
+static int trace_enqueue(u32 tgid, u32 pid,const char *comm)
```
ekyooo: Please always match the coding style of the base code.

libbpf-tools/runqslower.c
```diff
@@ -20,6 +21,8 @@ static volatile sig_atomic_t exiting = 0;
 struct env {
 	pid_t pid;
 	pid_t tid;
+	int task_comm_len;
```
ekyooo: If it’s needed, standardizing names such as comm/comm_len and task_comm/task_comm_len seems to improve readability.

ekyooo (commented):
There were also comments regarding the title.

libbpf-tools/runqslower.bpf.c
ekyooo: Even in that case, I found that using bpf_strncmp works as shown below.
Sharing for reference.

```
+const volatile __u32 targ_comm_len = 0;
 const volatile __u64 min_us = 0;
 const volatile pid_t targ_pid = 0;
 const volatile pid_t targ_tgid = 0;
@@ -31,11 +32,17 @@ static __always_inline bool comm_allowed(const char *comm)
 {
        int i;

+       if (targ_comm_len)
+               return 0 == bpf_strncmp(comm, targ_comm_len, targ_comm);
+       else
+               return true;
```

chudihuang: Hello, while `bpf_strncmp` offers a more modern implementation approach, I'd like to propose maintaining the current manual comparison method to preserve backward compatibility. This consideration stems from the fact that the `bpf_strncmp` helper was introduced in kernel 5.17+ (commit: 8efd0d9c316). if using bpf_strncmp meaning adoption would limit the tool's compatibility with earlier kernel versions.

chudihuang:
@ekyooo 
I have addressed the review comments and fixed the code. Please re-review:
1. Update option description.
2. Added length limit warnings for comm parameters
3. Fix coding style

ekyooo:
Sorry, could you please double-check before requesting a review?

chudihuang:
@ekyooo
    Apologies for the oversight. I've thoroughly re-examined the changes and confirmed their correctness. Should any issues remain, please don't hesitate to point them out - I'll address them immediately.

ekyooo:
@chudihuang 
Two of my review comments did not receive your feedback. Could you provide feedback on them? They might affect the current code, so it would be good to proceed with the review after receiving your feedback.

libbpf-tools/runqslower.bpf.c
```diff
@@ -48,10 +62,13 @@ static int handle_switch(void *ctx, struct task_struct *prev, struct task_struct
...
 	u32 pid;
+	char comm[TASK_COMM_LEN] = {0};
 
 	/* ivcsw: treat like an enqueue event and store timestamp */
-	if (get_task_state(prev) == TASK_RUNNING)
-		trace_enqueue(BPF_CORE_READ(prev, tgid), BPF_CORE_READ(prev, pid));
+	if (get_task_state(prev) == TASK_RUNNING) {
+		BPF_CORE_READ_STR_INTO(&comm, prev, comm);
+		trace_enqueue(BPF_CORE_READ(prev, tgid), BPF_CORE_READ(prev, pid), comm);
+	}
```
ekyooo: Remove tailing whitespace. Set your editor to auto-trim whitespace, please.

libbpf-tools/runqslower.c
```diff
@@ -169,6 +182,12 @@ int main(int argc, char **argv)
 	}
 
 	/* initialize global data (filtering options) */
+	if (env.comm) {
+		size_t copy_len = strnlen(env.comm, TASK_COMM_LEN - 1);
+		memcpy((char*)obj->rodata->targ_comm, env.comm, copy_len);
+		obj->rodata->targ_comm[copy_len] = '\0';
```
ekyooo: How about changing the three lines to a single line like this?
```
snprintf(obj->rodata->targ_comm, TASK_COMM_LEN, "%s", env.comm);
```

libbpf-tools/runqslower.c
```diff
@@ -9,6 +9,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
+#include <sys/param.h>
```
ekyooo: not needed.

libbpf-tools/runqslower.bpf.c
ekyooo: The code you referenced from bitesize was created in 2020. In my opinion, it would be better to write the code based on the latest version or, if backward compatibility is needed, to branch according to the target version. However, I won't insist on it.

chudihuang:
@ekyooo 
Thank you for the detailed code review. I have addressed the review comments and fixed the code. Please re-review:
1. Remove tailing whitespace. 
2. Consolidated three lines into a single statement as recommended.
3. Removed redundant #include <sys/param.h> declaration.

ekyooo:
@chudihuang 
There’s still some code that doesn’t follow the base coding style. How about finding and fixing it yourself to get the hang of it for future patches?

chudihuang:
@ekyooo 
Sorry for the formatting issues in the previous version. I've fixed two more style issues:

1. Replaced spaces with tabs for consistent indentation:
 ```diff
        if (!comm_allowed(comm))
-                return 0;
+               return 0;
 ```
2. ​​Added missing spaces​​ after that ',' :
 ```diff
- fprintf(stderr,"Warning: ...", env.comm,TASK_COMM_LEN-1, ...);
+ fprintf(stderr, "Warning: ...", env.comm, TASK_COMM_LEN - 1, ...);
 ```
Also fixed the line wrapping for better readability. Changes have been pushed. Please re-review!

libbpf-tools/runqslower.c
```diff
@@ -169,6 +180,10 @@ int main(int argc, char **argv)
 	}
 
 	/* initialize global data (filtering options) */
+	if (env.comm)
+		snprintf(obj->rodata->targ_comm, TASK_COMM_LEN, "%s", env.comm);
+
+
```
ekyooo: Remove unnecessary blank lines.

libbpf-tools/runqslower.c
```diff
@@ -64,6 +68,13 @@ static error_t parse_arg(int key, char *arg, struct argp_state *state)
 	case 'v':
 		env.verbose = true;
 		break;
+	case 'c':
+		env.comm = strdup(arg);
+		len = strlen(arg);
+		if (len >= TASK_COMM_LEN)
```
ekyooo: If the len variable is only used once, is it really necessary to declare it? I'm curious about your thoughts.

ekyooo (commented):
Reviewers (not me) have to review a lot of patches. I think doing your best to reduce the reviewers' burden before requesting a review will help you get your review done.

chudihuang:
Sorry for the formatting issues. Perhaps we should consider implementing a tool like the Linux kernel's checkpatch.pl to automate code style validation?

libbpf-tools/runqslower.bpf.c
```diff
@@ -48,10 +62,13 @@ static int handle_switch(void *ctx, struct task_struct *prev, struct task_struct
 	struct event event = {};
 	u64 *tsp, delta_us;
 	u32 pid;
+	char comm[TASK_COMM_LEN] = {0};
```
chenhengqi: We prefer `char comm[TASK_COMM_LEN] = {};`.

libbpf-tools/runqslower.bpf.c
```diff
@@ -99,13 +116,19 @@ int BPF_PROG(sched_switch, bool preempt, struct task_struct *prev, struct task_s
 SEC("raw_tp/sched_wakeup")
 int BPF_PROG(handle_sched_wakeup, struct task_struct *p)
 {
-	return trace_enqueue(BPF_CORE_READ(p, tgid), BPF_CORE_READ(p, pid));
+	char comm[TASK_COMM_LEN] = {0};
+
+	BPF_CORE_READ_STR_INTO(&comm, p, comm);
```
chenhengqi: We have to do string copy even if `targ_comm` is not set.
Consider adding a feature flag, if not set, pass `NULL` instead.

chudihuang:
@chenhengqi 
     I've implemented your suggestions in the latest update.

chenhengqi (approved):
LGTM overall. But let's wait for input from @ekyooo .

ekyooo:
@chudihuang 
LGTM too, thanks for the effort.

> automate code style validation?

I think it's a good suggestion. If you're interested, it would be great if you could contribute.
Thank you.

@chenhengqi 
Thank you. :smile:

PR #5333, opened 2025-06-13
libbpf-tools/opensnoop: support mount full-path

Rtoax:
Like d_path()/prepend_path() [1] did, get mount parent full-path. Although full-path support is added in [2], when there is a mount point in the path, [2] does not display the true full-path, for example:

/dev/sda mount on /home/sda/:

    $ mount | grep sda
    /dev/sda on /home/sda type xfs ...

    $ cd /home/sda/rongtao/
    $ touch a.out

    $ sudo ./opensnoop -F | grep touch -w
    216942 touch              3   0 /rongtao/a.out
                                    ^^^^^^^^^^^^^^

This patch will handle the mount points, tracking the real full-path:

    $ sudo ./opensnoop -F
    216568 touch              3   0 /home/sda/rongtao/a.out
                                    ^^^^^^^^^

[1] https://github.com/torvalds/linux/blob/master/fs/d_path.c
[2] https://github.com/iovisor/bcc/commit/b1c5548d4763


libbpf-tools/opensnoop.bpf.c
```diff
@@ -173,13 +178,34 @@ int trace_exit(struct syscall_trace_exit* ctx)
...
 			if (filepart_length > NAME_MAX)
 				break;
 
+			parent_dentry = BPF_CORE_READ(dentry, d_parent);
+
+			if (dentry == parent_dentry || dentry == mnt_root) {
+				struct mount *mnt_parent;
+				mnt_parent = BPF_CORE_READ(mnt, mnt_parent);
+
+				if (dentry != mnt_root)
```
yonghong-song: This probably not needed.

libbpf-tools/opensnoop.bpf.c
```diff
@@ -173,13 +178,34 @@ int trace_exit(struct syscall_trace_exit* ctx)
...
+					break;
+				if (mnt != mnt_parent) {
+					dentry = BPF_CORE_READ(mnt, mnt_mountpoint);
```
yonghong-song, line 191: IIUC, correctly, dentry here is a 'hidden' dentry to facilitate path traversal. So after this, we should do `dentry = BPF_CORE_READ(dentry, d_parent);`?

Please look at this upstream patch https://lore.kernel.org/bpf/20250611220220.3681382-2-song@kernel.org/ to double check.

libbpf-tools/opensnoop.bpf.c
Rtoax: Yep, fixed,thanks.

libbpf-tools/opensnoop.bpf.c
Rtoax: I tried, but tracking the wrong path

```diff
diff --git a/libbpf-tools/opensnoop.bpf.c b/libbpf-tools/opensnoop.bpf.c
index 64d931a8..08510e42 100644
--- a/libbpf-tools/opensnoop.bpf.c
+++ b/libbpf-tools/opensnoop.bpf.c
@@ -189,6 +189,7 @@ int trace_exit(struct syscall_trace_exit* ctx)
 
                                if (mnt != mnt_parent) {
                                        dentry = BPF_CORE_READ(mnt, mnt_mountpoint);
+                                       dentry = BPF_CORE_READ(dentry, d_parent);
 
                                        mnt = mnt_parent;
                                        vfsmnt = &mnt->mnt;
```

For example:

```
    $ mount | grep sda
    /dev/sda on /home/sda type xfs ...

    /home/sda/rongtao/a.out
          ^^^
          [1]

    /* get 'rongtao' mount point 'sda' */
[1] dentry = BPF_CORE_READ(mnt, mnt_mountpoint);
[2] dentry = BPF_CORE_READ(dentry, d_parent);
```

so, i think there is no need [2] after [1], right?

yonghong-song: I tried a nested mount example. You are right, the [2] is not needed.

PR #5334, opened 2025-06-13
tools/opensnoop: one event is enough for full-path

Rtoax:
Each loop iteration of dentry will send out an event [1], which is too much and unnecessary.

1. Use ringbuf instead of perf buffer. Because the event transmitted at one time is large and the stack space is insufficient, the reservation function of ringbuf is used. And, ringbuf is wildly supported.
2. Each full-path transmits only one event, eliminating unnecessary events. It's good for further developement.

[1] https://github.com/iovisor/bcc/commit/c110a4dd0c8f

Cc: Yonghong Song <yonghong.song@linux.dev>

### DIFF
--- a/tools/opensnoop.py
+++ b/tools/opensnoop.py
@@ -104,14 +104,10 @@ bpf_text = """
 #ifdef FULLPATH
 #include <linux/fs_struct.h>
 #include <linux/dcache.h>
-
-#define MAX_ENTRIES 32
-
-enum event_type {
-    EVENT_ENTRY,
-    EVENT_END,
-};
 #endif
+
+#define NAME_MAX 255
+#define MAX_ENTRIES 32
 
 struct val_t {
     u64 id;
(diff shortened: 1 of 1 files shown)


Rtoax:
@yonghong-song Try use a single event to handle full-path, instead of each loop iteration send out an event.  (relate closed-not-merged PR https://github.com/iovisor/bcc/pull/5324), please review, thanks.

yonghong-song:
Generally looks good. But I see the following output
```
250682 FuncSched         399   0 //sys/fs/cgroup/system.slice/memory.high
250682 FuncSched         399   0 //sys/fs/cgroup/system.slice/memory.max
250682 FuncSched         399   0 //sys/fs/cgroup/system.slice/scribed.service/cgroup.controllers
...
640792 tmux: server       32   0 //proc/3403369/cmdline
1831796 load-monitor      130   0 //proc/stat
321262 FuncSched        1296   0 //sys/fs/cgroup/system.slice/cpu.pressure
```
The double prefix `//' is not right.

Rtoax:
> Generally looks good. But I see the following output
> 
> ```
> 250682 FuncSched         399   0 //sys/fs/cgroup/system.slice/memory.high
> 250682 FuncSched         399   0 //sys/fs/cgroup/system.slice/memory.max
> 250682 FuncSched         399   0 //sys/fs/cgroup/system.slice/scribed.service/cgroup.controllers
> ...
> 640792 tmux: server       32   0 //proc/3403369/cmdline
> 1831796 load-monitor      130   0 //proc/stat
> 321262 FuncSched        1296   0 //sys/fs/cgroup/system.slice/cpu.pressure
> ```
> 
> The double prefix `//' is not right.

Fixed. it's because if `touch /home/xxxx`, an extra '/' will added. Thanks :)

PR #5342, opened 2025-06-23
tools/runqlat: Dynamically size pid/pidns histogram map

ismhong:
When using `runqlat` with the `--pids` or `--pidnss` flags, the BPF histogram for tracking per-process or per-namespace latency is created with a fixed default size of 64 entries.

On systems with a large number of processes or active namespaces, this small, fixed size can be easily exhausted. This leads to silent data loss for any PIDs or PID namespaces beyond the 64th entry, making the tool's output incomplete and misleading.

To fix this, read the maximum possible PID and PID Namespace values from `/proc/sys/kernel/pid_max` and `/proc/sys/user/max_user_namespaces` respectively. Pass these values to the BPF program to dynamically size the histogram map, ensuring it is large enough to avoid dropping data.

This approach is similar to the one implemented for cpudist as discussed in PR #2568.

Link: https://github.com/iovisor/bcc/pull/2568


tools/runqlat.py
```diff
@@ -289,8 +289,11 @@
     if args.ebpf:
         exit()
 
+max_pid = int(open("/proc/sys/kernel/pid_max").read())
+max_pidns = int(open("/proc/sys/user/max_user_namespaces").read())
```
yonghong-song: Should the above be /proc/sys/user/max_pid_namespaces? Did I miss anything?

ismhong: Hi @yonghong-song 
Thanks for review. You are right, sorry for my carelessness.
https://docs.kernel.org/admin-guide/sysctl/user.html
I have fixed it, please help to review it, thanks.

PR #5344, opened 2025-06-30
libbpf-tools: trace_helpers: add str_loadavg()

Rtoax:
There are many tools read /proc/loadavg, this commit add a comment function str_loadavg().


libbpf-tools/trace_helpers.c
```diff
@@ -1311,3 +1311,30 @@ int str_to_long(const char *src, void *dest)
...
+	struct tm *tm;
+	char ts[16], avg[64];
+	FILE *f = fopen("/proc/loadavg", "r");
+	if (!f) {
+		perror("fopen(/proc/loadavg)");
+		return NULL;
+	}
+
+	int n = fread(avg, 1, sizeof(avg), f);
+	if (n) {
```
yonghong-song: I would like str_loadavg() to return a bool var to indicate whether putting loadavg in buf is successful or not. This will make code easier to understand and less error prone.

For example, if the above 'if (n)' is false, the return value should be false, then there is no need to print buf at all. For the current code, since buf is not 'memset' yet, garbage could be printed out.

Rtoax: Fixed, thanks :)

libbpf-tools/trace_helpers.c
```diff
@@ -1311,3 +1311,30 @@ int str_to_long(const char *src, void *dest)
 
 	return errno;
 }
+
+bool str_loadavg(char *buf, size_t buf_len, bool ts_pfx)
```
ekyooo: Please add invalid argument check for buf.
The common API has diverse callers, so defensive coding seems safer.

ekyooo: The error causes for returning 'false' seem varied. It would be helpful if the called function could return the cause so the calling function can handle logic based on it.

ekyooo: The name ts_pfx isn't very intuitive. Renaming the variable or adding comments to make it self-explanatory without reading the function body would be helpful.

libbpf-tools/trace_helpers.c
```diff
@@ -1311,3 +1311,30 @@ int str_to_long(const char *src, void *dest)
...
+{
+	time_t t;
+	struct tm *tm;
+	char ts[16], avg[64];
+	FILE *f = fopen("/proc/loadavg", "r");
```
ekyooo: a blank line please

libbpf-tools/biotop.c
```diff
@@ -301,24 +301,12 @@ static int read_stat(struct biotop_bpf *obj, struct data_t *datas, __u32 *count)
...
-		strftime(ts, sizeof(ts), "%H:%M:%S", tm);
-		memset(buf, 0, sizeof(buf));
-		n = fread(buf, 1, sizeof(buf), f);
-		if (n)
-			printf("%8s loadavg: %s\n", ts, buf);
-		fclose(f);
-	}
+	int i, err = 0, rows = OUTPUT_ROWS_LIMIT;
+
+	printf("%s\n", str_loadavg(buf, sizeof(buf), true) ? buf : "N/A");
```
ekyooo: Intentional change to the tool's output format?

libbpf-tools/trace_helpers.c
```diff
@@ -1311,3 +1311,30 @@ int str_to_long(const char *src, void *dest)
...
+	if (!f) {
+		perror("fopen(/proc/loadavg)");
```
ekyooo: Common APIs (helper functions) typically signal error conditions through return values rather than output.

libbpf-tools/trace_helpers.c
```diff
@@ -1311,3 +1311,30 @@ int str_to_long(const char *src, void *dest)
...
+		return false;
+	}
+
+	int n = fread(avg, 1, sizeof(avg), f);
+	if (n) {
```
ekyooo: Handle it as a failure value and return an appropriate error code?

libbpf-tools/trace_helpers.c
```diff
@@ -1311,3 +1311,30 @@ int str_to_long(const char *src, void *dest)
...
+		memset(buf, 0, buf_len);
+		if (ts_pfx) {
+			time(&t);
+			tm = localtime(&t);
+			strftime(ts, sizeof(ts), "%H:%M:%S", tm);
+			snprintf(buf, buf_len, "%8s loadavg: %s", ts, avg);
```
ekyooo: Could we handle cases where the caller provides a buffer of insufficient size?

Rtoax:
@ekyooo Thanks for all the advise, and also thanks @yonghong-song , please review.

libbpf-tools/filetop.c
```diff
@@ -182,26 +182,13 @@ static int sort_column(const void *obj1, const void *obj2)
...
-		time(&t);
-		tm = localtime(&t);
-		strftime(ts, sizeof(ts), "%H:%M:%S", tm);
-		memset(buf, 0, sizeof(buf));
-		n = fread(buf, 1, sizeof(buf), f);
-		if (n)
-			printf("%8s loadavg: %s\n", ts, buf);
-		fclose(f);
-	}
+	printf("%s\n", str_loadavg(buf, sizeof(buf), true) ? "N/A" : buf);
```
ekyooo: Intentional change to the tool's output format?
Same question for other tool changes.
If intentional, how about including it in the commit message?

libbpf-tools/trace_helpers.c
```diff
@@ -1311,3 +1311,47 @@ int str_to_long(const char *src, void *dest)
...
+	char ts[16], avg[64];
+	FILE *f;
+
+	if (!buf || buf_len == 0)
+		return -EINVAL;
+
+	f = fopen("/proc/loadavg", "r");
+	if (!f) {
+		return -errno;
+	}
```
ekyooo: Parentheses are no longer needed.

libbpf-tools/trace_helpers.c
```diff
@@ -1311,3 +1311,47 @@ int str_to_long(const char *src, void *dest)
 
 	return errno;
 }
+
+/**
```
ekyooo: The function's description comments are in both .h and .c files, but it would be better to consolidate them into one place.

libbpf-tools/trace_helpers.c
```diff
@@ -1311,3 +1311,47 @@ int str_to_long(const char *src, void *dest)
...
+	if (!n) {
+		err = -errno;
+		goto cleanup;
+	}
+
+	memset(buf, 0, buf_len);
+	if (ts_pfx) {
+		time(&t);
+		tm = localtime(&t);
+		strftime(ts, sizeof(ts), "%H:%M:%S", tm);
```
ekyooo, line 1345: The timestamp-related code is used in several tools, not just when reading /proc/loadavg.
Instead of including it within this function, how about separating it into a utility function and letting the caller invoke it as needed?

ekyooo: The loadavg and timestamp are orthogonal pieces of information, so it makes sense to keep their handling separate. 
The timestamp-related code is used in several tools, not just when reading /proc/loadavg. 
Instead of including it within this function, how about separating it into a utility function and letting the caller invoke it as needed? This would also improve maintainability.

libbpf-tools/filetop.c
Rtoax: Fixed, thanks.

Rtoax:
@ekyooo add `str_timestamp()` and apply to all tools, please review.

Rtoax:
rebase to master.

libbpf-tools/trace_helpers.c
```diff
@@ -1311,3 +1311,50 @@ int str_to_long(const char *src, void *dest)
...
+		return -errno;
+
+	n = fread(avg, 1, sizeof(avg), f);
+	if (!n) {
+		err = -errno;
+		goto cleanup;
+	}
+
+	memset(buf, 0, buf_len);
+	n = snprintf(buf, buf_len, "loadavg: %s", avg);
```
ekyooo, line 1334: avg may not be null-terminated after fread, which can cause undefined behavior with %s in snprintf?

libbpf-tools/trace_helpers.c
```diff
@@ -1311,3 +1311,50 @@ int str_to_long(const char *src, void *dest)
...
+	if (!f)
+		return -errno;
+
+	n = fread(avg, 1, sizeof(avg), f);
+	if (!n) {
+		err = -errno;
+		goto cleanup;
+	}
+
+	memset(buf, 0, buf_len);
```
ekyooo: memset is unnecessary, as snprintf null-terminates the buffer.
Please refer to the snprintf man page.

libbpf-tools/trace_helpers.c
```diff
@@ -1311,3 +1311,50 @@ int str_to_long(const char *src, void *dest)
...
+	n = snprintf(buf, buf_len, "loadavg: %s", avg);
+
+	if (n >= buf_len)
+		err = -ERANGE;
+	else
+		buf[n] = '\0';
```
ekyooo: not needed according to the snprintf man.

libbpf-tools/bashreadline.c
```diff
@@ -74,13 +74,9 @@ static int libbpf_print_fn(enum libbpf_print_level level, const char *format, va
...
 {
 	struct str_t *e = data;
-	struct tm *tm;
 	char ts[16];
-	time_t t;
 
-	time(&t);
-	tm = localtime(&t);
-	strftime(ts, sizeof(ts), "%H:%m:%S", tm);
+	str_timestamp("%H:%m:%S", ts, sizeof(ts));
```
chenhengqi: `%m` ?

chenhengqi (commented):
Nice cleanup. Thanks.

libbpf-tools/trace_helpers.c
Rtoax: Fixed, thanks.

libbpf-tools/trace_helpers.c
Rtoax: Fixed, thanks.

libbpf-tools/trace_helpers.c
Rtoax: Fixed, thanks.

libbpf-tools/bashreadline.c
Rtoax: Sorry about this, fixed, please review, thanks.

libbpf-tools/trace_helpers.c
```diff
@@ -1311,3 +1311,47 @@ int str_to_long(const char *src, void *dest)
 
 	return errno;
 }
+
+int str_loadavg(char *buf, size_t buf_len)
+{
+	int n, err = 0;
+	char avg[64] = {0};
```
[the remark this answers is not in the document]
ekyooo: Sorry, I'm confused and sleepy. Will remove the comment again.

libbpf-tools/trace_helpers.c
```diff
@@ -1311,3 +1311,47 @@ int str_to_long(const char *src, void *dest)
...
+{
+	time_t t;
+	struct tm *tm;
+
+	if (!format || !buf || buf_len == 0)
+		return -EINVAL;
+
+	time(&t);
+	tm = localtime(&t);
+	if (tm == NULL)
```
ekyooo: Trivial, but `!tm` aligns with the format of line 1349, making it look nicer.

libbpf-tools/biotop.c
```diff
@@ -301,24 +301,15 @@ static int read_stat(struct biotop_bpf *obj, struct data_t *datas, __u32 *count)
...
-		if (n)
-			printf("%8s loadavg: %s\n", ts, buf);
-		fclose(f);
-	}
+	int i, err = 0, rows = OUTPUT_ROWS_LIMIT;
+
+	err = str_loadavg(loadavg, sizeof(loadavg));
+	err += str_timestamp("%H:%M:%S", ts, sizeof(ts)) <= 0;
+	if (!err)
+		printf("%s %s\n", ts, loadavg);
```
ekyooo: Was the change to the tool's output format intentional?
original `%8s` looks generally suitable for ts as it aligns columns.

It seems there have been a few updates to the tool's output code without notification, which don't align with the commit descriptions, causing confusion.
If intentional, could you share the reasoning in the commit message or elsewhere? If not, it might help to be more cautious in the future.
This feedback also applies to similar code in this PR.

libbpf-tools/trace_helpers.c
```diff
@@ -1311,3 +1311,47 @@ int str_to_long(const char *src, void *dest)
...
+	}
+
+	n = snprintf(buf, buf_len, "loadavg: %s", avg);
+
+	if (n >= buf_len)
+		err = -ERANGE;
+
+cleanup:
+	fclose(f);
+	return err;
```
ekyooo: How about having str_loadavg return the string length? It would let the caller handle the result more accurately and maintain consistency with str_timestamp, I think.

libbpf-tools/trace_helpers.c
Rtoax: Fixed , thanks.

libbpf-tools/biotop.c
Rtoax: Yep, sorry, the bashreadline timestamp use wrong %m (month), it's should be %M, add information to commit lines. thank you, and thanks to @chenhengqi .

libbpf-tools/trace_helpers.c
Rtoax: OK, Done.

Rtoax:
Rebase to master.

Rtoax:
Rebase to master.

libbpf-tools/slabratetop.c
```diff
@@ -169,26 +169,16 @@ static int sort_column(const void *obj1, const void *obj2)
...
-		strftime(ts, sizeof(ts), "%H:%M:%S", tm);
-		memset(buf, 0 , sizeof(buf));
-		n = fread(buf, 1, sizeof(buf), f);
-		if (n)
-			printf("%8s loadavg: %s\n", ts, buf);
-		fclose(f);
-	}
+	err = str_loadavg(loadavg, sizeof(loadavg));
+	err += str_timestamp("%H:%M:%S", ts, sizeof(ts)) <= 0;
+	if (!err)
```
ekyooo, line 180: The str_loadavg API return value has changed, requiring updates to the caller-side code.
It is currently not working.
Please check other tools as well.

Rtoax: Sorry about that, fixed.

libbpf-tools/trace_helpers.c
```diff
@@ -1311,3 +1311,47 @@ int str_to_long(const char *src, void *dest)
...
+	FILE *f;
+
+	if (!buf || buf_len == 0)
+		return -EINVAL;
+
+	f = fopen("/proc/loadavg", "r");
+	if (!f)
+		return -errno;
+
+	n = fread(avg, 1, sizeof(avg), f);
```
ekyooo: Reading sizeof(buf) bytes with fread may overwrite the null terminator (\0), potentially causing unsafe behavior.
How about using `n = fread(avg, 1, sizeof(avg) - 1, f);` instead?

libbpf-tools/slabratetop.c
```diff
@@ -169,26 +169,16 @@ static int sort_column(const void *obj1, const void *obj2)
...
-		tm = localtime(&t);
-		strftime(ts, sizeof(ts), "%H:%M:%S", tm);
-		memset(buf, 0 , sizeof(buf));
-		n = fread(buf, 1, sizeof(buf), f);
-		if (n)
-			printf("%8s loadavg: %s\n", ts, buf);
-		fclose(f);
-	}
+	err = str_loadavg(loadavg, sizeof(loadavg)) <= 0;
+	err = err ?: str_timestamp("%H:%M:%S", ts, sizeof(ts)) <= 0;
```
ekyooo: (<=) seems less readable to me, but it could be a matter of personal taste. 

You may consider simplifying this logic for readability, especially since the err variable is not used afterward. For example:
```
if (str_loadavg(buf, sizeof(buf)) > 0 && str_timestamp("%H:%M:%S", ts, sizeof(ts)) > 0)
    printf("%8s %s\n", ts, buf);
```

Alternatively, you might prefer using parentheses for clarity:
```
err = err ?: (str_timestamp("%H:%M:%S", ts, sizeof(ts)) <= 0);
```
Both styles are just suggestions and not mandatory. Please feel free to choose whichever approach you find most suitable.
There is no significant difference from the original code, so this is just a minor stylistic suggestion.

libbpf-tools/trace_helpers.c
Rtoax: Indeed, thanks.

libbpf-tools/slabratetop.c
Rtoax: Thanks, i choice the second one, it's more readable, thanks!

Rtoax:
> From my viewpoint, we’re nearly done.
> I appreciate your patience with my iterative requests.
> Thanks!

Totally OK, more patience, higher quality. thanks.

ekyooo:
LGTM. Thanks for nice cleanup.
I hope this work can be helpful for your subsequent common API tasks.
Thank you again.

ekyooo (approved):

PR #5345, opened 2025-06-30
tools: introduce path helpers and filelife,filegone support file path

Rtoax:
Do the same thing https://github.com/iovisor/bcc/pull/5340 to tools.

### add 

- tools/path_helpers.bpf.c: add bpf helpers to get full-path from dentry;
- tools/path_helpers.py: to parse full-path in userspace with python;
- tools/full_path.h: define C full-path struct/macro;

### filelife,filegone

- use ring-buffer instead of perf-buffer;
- support full-path with -F argument;

### DIFF
--- a/tools/filegone.py
+++ b/tools/filegone.py
@@ -11,6 +11,7 @@
 # 08-Nov-2022 Curu. modified from filelife
 # 19-Nov-2022 Rong Tao Check btf struct field instead of KERNEL_VERSION macro.
 # 05-Nov-2023 Rong Tao Support rename/unlink failed situation.
+# 01-Jul-2025 Rong Tao Support file path
 
 from __future__ import print_function
 from bcc import BPF
@@ -20,6 +21,7 @@ from time import strftime
 # arguments
 examples = """examples:
     ./filegone           # trace all file gone events
+    ./filegone -P        # show file path
     ./filegone -p 181    # only trace PID 181
 """
 parser = argparse.ArgumentParser(
@@ -28,6 +30,8 @@ parser = argparse.ArgumentParser(
     epilog=examples)
 parser.add_argument("-p", "--pid",
     help="trace this PID only")
+parser.add_argument("-P", "--path", action="store_true",
+    help="show file path")
 parser.add_argument("--ebpf", action="store_true",
     help=argparse.SUPPRESS)
 args = parser.parse_args()
@@ -37,14 +41,25 @@ debug = 0
 bpf_text = """
 #include <uapi/linux/ptrace.h>
 #include <linux/fs.h>
+#include <linux/fs_struct.h>
 #include <linux/sched.h>
+#ifdef FULLPATH
+INCLUDE_FULL_PATH_H
+INCLUDE_PATH_HELPERS_BPF_H
+#endif
 
 struct data_t {
     u32 pid;
     u8 action;
     char comm[TASK_COMM_LEN];
-    char fname[DNAME_INLINE_LEN];
-    char fname2[DNAME_INLINE_LEN];
+    u32 path_depth, path_depth2;
+#ifdef FULLPATH
+    FULL_PATH_FIELD(fname);
+    FULL_PATH_FIELD(fname2);
+#else
+    char fname[NAME_MAX];
+    char fname2[NAME_MAX];
+#endif
 };
 
 struct entry_t {
@@ -52,6 +67,7 @@ struct entry_t {
     u8 action;
     struct {
         char name[DNAME_INLINE_LEN];
+        struct dentry *dentry;
     } old, new;
 };
 
@@ -79,6 +95,7 @@ TRACE_VFS_UNLINK_FUNC
 
     entry.pid = pid;
     entry.action = 'D';
+    entry.old.dentry = dentry;
     get_dentry_name((char **)&entry.old.name, dentry);
 
     currentry.update(&tid, &entry);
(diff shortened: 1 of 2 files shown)


Rtoax:
Rebase to master.

Rtoax:
Rebase to master.

ekyooo (commented):
I have a few questions about the overall structure first.

1. Is the use of bpf_map_ringbuf mandatory for path_helper?

2. If not mandatory, can we split the application commits for each tool (e.g., filegone, filelife) into two separate changes?
(1) One commit for replacing the perf buffer map with the ring buffer map
(2) Another commit for actually applying path_helper
I believe separating the structural (infrastructure) change (ring buffer migration) from the functional change (full-path helper adoption) would make the history easier to manage and help with future maintenance. Also, having a dedicated commit for path_helper integration will serve as a clear reference for other developers who may want to apply it to additional tools later.

3. Would it be possible to separate the application of path_helper to opensnoop into its own commit? The first commit is quite large, so splitting out the path_helper integration—similar to what was done for the other tools—would make the code easier to review and understand.

4. Is there a specific reason for using -F (full-path) instead of -P/--path? Personally, when I see the -F option, it’s not immediately clear that it refers to “path output,” and the meaning of “F” isn’t very intuitive. I’m curious about the reasoning behind this choice.

Thank you.

Rtoax:
@ekyooo Thanks a lot, i'll try to split the commits, please give me some time.

Rtoax:
@ekyooo  There are 6 commits ( minimize code modifications per commit )

 1. Introduce path_helpers (without modifying opensnoop.py) 
 2. Modify opensnoop.py 
 3. filelife perf-buffer replaced with ring-buffer (required, ringbuffer_reserve required) 
 4. filelife implementation file path (-P) 
 5. Replace perf-buffer of fileone with ring-buffer (required, ringbuffer_reserve required) 
 6. filebone implementation file path (-P) 

 Please review the code again, thank you

Rtoax:
rebase to master

ekyooo:
>Is the use of bpf_map_ringbuf mandatory for path_helper?

Could you answer my question with clear reasoning?

To clarify the buffer type in https://github.com/iovisor/bcc/pull/5340:

1. You used compat.h in filelife, which provides compatibility between perf buffer and ring buffer.
2. You stated in the commit message that you replaced perf buffer with ring buffer.
3. What is true?

The use of the ring buffer is a modern change, but it excludes users of kernel versions below 5.8.
Should we reduce the tool's overall usability for an optional feature?
Before deciding, I need to verify the justification for this change.

Rtoax:
> > Is the use of bpf_map_ringbuf mandatory for path_helper?
> 
> Could you answer my question with clear reasoning?
> 
> To clarify the buffer type in #5340:
> 
>     1. You used compat.h in filelife, which provides compatibility between perf buffer and ring buffer.
> 
>     2. You stated in the commit message that you replaced perf buffer with ring buffer.
> 
>     3. What is true?

Thanks for your reply, the https://github.com/iovisor/bcc/pull/5340 is for `libbpf-tools/`, this PR for `tools/`, those are different. This PR only replace perfbuffer with ring buffer.

> 
> 
> The use of the ring buffer is a modern change, but it excludes users of kernel versions below 5.8. Should we reduce the tool's overall usability for an optional feature? Before deciding, I need to verify the justification for this change.

Indeed, thanks, you're right, maybe we should introduce compat.h like method for `tools/` ?

ekyooo:
I have a question: was perf_buffer also verified to work correctly in PR #5340?

If so, I think it would be beneficial to apply compat.h to the Python tools as well. Even without using path_helper, this change would likely benefit multiple tools. Thanks!

Rtoax:
Thank you for your reply, I will confirm this ;)

Rtoax:
> I have a question: was perf_buffer also verified to work correctly in PR #5340?

Sorry about the late reply, i tried this change, it's not works.

```diff
diff --git a/libbpf-tools/compat.c b/libbpf-tools/compat.c
index 7d932ef30591..31065e178984 100644
--- a/libbpf-tools/compat.c
+++ b/libbpf-tools/compat.c
@@ -35,7 +35,7 @@ struct bpf_buffer *bpf_buffer__new(struct bpf_map *events, struct bpf_map *heap)
        bool use_ringbuf;
        int type;
 
-       use_ringbuf = probe_ringbuf();
+       use_ringbuf = 0;
        if (use_ringbuf) {
                bpf_map__set_autocreate(heap, false);
                type = BPF_MAP_TYPE_RINGBUF;
```

```
$ sudo ./filelife 
libbpf: map 'events': failed to create: -EINVAL
libbpf: failed to load BPF skeleton 'filelife_bpf': -EINVAL
failed to load BPF object: -22
```

ekyooo:
> > I have a question: was perf_buffer also verified to work correctly in PR #5340?
> 
> Sorry about the late reply, i tried this change, it's not works.

You need to disable the ringbuffer not only for `.c` files but also for `.bpf.c` files if you want to force the use of the perf buffer. For example:

```c
--- a/libbpf-tools/compat.bpf.h
+++ b/libbpf-tools/compat.bpf.h
@@ -26,19 +26,21 @@ static __always_inline void *reserve_buf(__u64 size)
 {
        static const int zero = 0;
 
+#if 0
        if (bpf_core_type_exists(struct bpf_ringbuf))
                return bpf_ringbuf_reserve(&events, size, 0);
-
+#endif
        return bpf_map_lookup_elem(&heap, &zero);
 }
 
 static __always_inline long submit_buf(void *ctx, void *buf, __u64 size)
 {
+#if 0
        if (bpf_core_type_exists(struct bpf_ringbuf)) {
                bpf_ringbuf_submit(buf, 0);
                return 0;
        }
-
+#endif
        return bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, buf, size);
 }
```

By forcing the use of the perf buffer with the above approach, `opensnoop` and `oomkill` work correctly, but `filelife` does not. Could you investigate the cause?

Rtoax:
> By forcing the use of the perf buffer with the above approach, `opensnoop` and `oomkill` work correctly, but `filelife` does not. Could you investigate the cause?

Sure, please see this PR https://github.com/iovisor/bcc/pull/5432 , please review.

PR #5373, opened 2025-07-17
tools/biolatency: Restore tracepoint-first probe logic

ismhong:
On Linux kernels v5.16 and newer, kprobe attachment for biolatency can fail. The `blk_account_io_done` function became an inline function in v5.16 and was later removed in v6.4. This causes biolatency to fail with an "attach to kprobe failed" error on recent kernels like v6.6.

Commit 947bd74 ("tools/{biolatency,biosnoop,biotop}: Use tracepoint first") originally established the intended behavior to prefer tracepoints.  However, commit 3630581
("tools/{biolatency,biosnoop,biotop}: use TRACEPOINT_PROBE() for tracepoints") inadvertently changed the probe attachment logic, causing it to prioritize kprobes over tracepoints. When the necessary kprobes are not found, the tool fails instead of falling back to available tracepoints.

This patch fixes the issue by restoring the original logic, ensuring that the tool checks for available tracepoints before attempting to attach kprobes. This resolves the failure on modern kernels and aligns with the tool's intended design.

Fixes: 3630581 ("tools/{biolatency,biosnoop,biotop}: use TRACEPOINT_PROBE() for tracepoints")

Change-Id: I7ed3e61d629c033e0f87dc736d18949c81c25d58

### DIFF
--- a/tools/biolatency.py
+++ b/tools/biolatency.py
@@ -267,28 +267,28 @@ kprobe_done = None
 tp_done = None
 
 if args.queued:
-    if BPF.get_kprobe_functions(b'__blk_account_io_start'):
-        kprobe_start.add("__blk_account_io_start")
-    elif BPF.get_kprobe_functions(b'blk_account_io_start'):
-        kprobe_start.add("blk_account_io_start")
-    elif BPF.tracepoint_exists("block", "block_io_start"):
+    if BPF.tracepoint_exists("block", "block_io_start"):
         tp_start = "block_io_start"
     elif BPF.tracepoint_exists("block", "block_bio_queue"):
         tp_start = "block_bio_queue"
+    elif BPF.get_kprobe_functions(b'__blk_account_io_start'):
+        kprobe_start.add("__blk_account_io_start")
+    elif BPF.get_kprobe_functions(b'blk_account_io_start'):
+        kprobe_start.add("blk_account_io_start")
 
 else:
     if BPF.get_kprobe_functions(b'blk_start_request'):
         kprobe_start.add("blk_start_request")
     kprobe_start.add("blk_mq_start_request")
 
-if BPF.get_kprobe_functions(b'__blk_account_io_done'):
-    kprobe_done = "__blk_account_io_done"
-elif BPF.get_kprobe_functions(b'blk_account_io_done'):
-    kprobe_done = "blk_account_io_done"
-elif BPF.tracepoint_exists("block", "block_io_done"):
+if BPF.tracepoint_exists("block", "block_io_done"):
     tp_done = "block_io_done"
 elif BPF.tracepoint_exists("block", "block_rq_complete"):
     tp_done = "block_rq_complete"
+elif BPF.get_kprobe_functions(b'__blk_account_io_done'):
+    kprobe_done = "__blk_account_io_done"
+elif BPF.get_kprobe_functions(b'blk_account_io_done'):
+    kprobe_done = "blk_account_io_done"
 
 if args.flags and (tp_start or tp_done):
             # Some flags are accessible in the rwbs field (RAHEAD, SYNC and META)


ismhong:
Hi @jeromemarchand 

Could you please help to review this PR? Thanks.

jeromemarchand:
If `blk_account_io_done` is inlined, then the tool would use the tracepoint instead unless BPF.get_kprobe_functions() is broken somehow. On my machine with a 6.12 kernel and the `blk_account_io_done` kprobe unavailable, the tools safely default to tracepoint in that case. In what exact configuration do you see the tool fails?
The kprobes offers more information than the tracepoints and because of that, using the tracepoint prevent the use of the -F option.

ismhong:
Hi @jeromemarchand 

I'm encountering the issue on my Yocto 6.6 arm64 platform. The function `blk_account_io_done` is inlined, yet it still appears in `/proc/kallsyms`.
The implementation of `BPF.get_kprobe_functions()` checks for the function's existence in `/proc/kallsyms`. In my case, this check returns True for `blk_account_io_done`, causing it to proceed with a kprobe attachment. However, this attachment fails because the function is not actually probeable, as confirmed by its absence from `/sys/kernel/tracing/available_filter_functions`.
Please see the system details and failure log below.
```
(bcc)root@bleedingedge-rtd1619b-mini:/# cat /proc/kallsyms | grep account_io_done
ffffffc0804b6480 t blk_account_io_done

(bcc)root@bleedingedge-rtd1619b-mini:/# cat /sys/kernel/tracing/available_filter_functions | grep account_io_done

(bcc)root@bleedingedge-rtd1619b-mini:/# uname -a
Linux bleedingedge-rtd1619b-mini 6.6.54-yocto-standard-00234-ga4ee8082cbf9 #1 SMP PREEMPT Tue Oct  8 03:03:40 UTC 2024 aarch64 Linux

(bcc)root@bleedingedge-rtd1619b-mini:/# runqlat
libbpf: failed to find valid kernel BTF
/virtual/main.c:92:15: error: no member named 'state' in 'struct task_struct'; did you mean 'stats'?
    if (prev->state == TASK_RUNNING) {
              ^~~~~
              stats
include/linux/sched.h:561:34: note: 'stats' declared here
        struct sched_statistics         stats;
                                        ^
/virtual/main.c:92:21: error: invalid operands to binary expression ('struct sched_statistics' and 'int')
    if (prev->state == TASK_RUNNING) {
        ~~~~~~~~~~~ ^  ~~~~~~~~~~~~
2 errors generated.
  File "/usr/share/bcc/tools//runqlat", line 296, in <module>
    "-DMAX_PIDNS=%d" % max_pidns])
  File "/usr/lib/python2.7/site-packages/bcc/__init__.py", line 476, in __init__
```

jeromemarchand:
Interesting. If the function is not in /sys/kernel/tracing/available_filter_functions, it should be filtered out by `BPF.get_kprobe_functions()`, at least since commit e0291c8d ("Filter with available_filter_functions in BPF.get_kprobe_functions()").

jeromemarchand:
I'm gonna try to reproduce the issue on a arm64 machine.

jeromemarchand:
That was inconclusive. On my system the function was only partially inlined and the function doesn't appear in kallsyms but a subfunction does:
``` 
# grep blk_account_io_done /proc/kallsyms 
ffffa5d7c43a55f0 t blk_account_io_done.part.0
```

But looking at the current code of `get_kprobe_functions()` it seems it should work in your case too. Do you by any chance use recent bcc-tools with an older bcc?

Does `funccount -r 'blk.*'` or `funccount -r 'vfs.*'` works for you?

ismhong:
Hi @jeromemarchand 
Thanks to help digging it, and you should be right as the current BCC version I use is v0.25 (host pre-packed for Embedded Linux :D). As you mentioned it should be filter out by checking available_filter_functions. I will try to upgrade BCC runtime first, and update status later. Thank you.

ismhong:
Hi @jeromemarchand 
After cherry-pick the commit https://github.com/iovisor/bcc/commit/e0291c8d3a16e18ba163fd55e44d0860ef1223a1 ("Filter with available_filter_functions in BPF.get_kprobe_functions()"), the `BPF.get_kprobe_functions()` won't report `blk_account_io_done` exist and fix the error I encountered.

I will close this PR, thanks for help again.

Issue #5352, opened 2025-07-04
Incorrect calculation of recv throughput in tcptop.py


sc07kvm:
tcp_cleanup_rbuf is called several times in a loop inside tcp_recvmsg. At the same time, it is passed the accumulated parameter "copied". Thus, the same data is taken into account several times.

```c
int tcp_recvmsg(...) {
     ...
     int copied = 0;
     ...
     do {
         ...
         tcp_cleanup_rbuf(sk, copied);
         ...
         copied += used;
         ...
     } while (len > 0);
     
     tcp_cleanup_rbuf(sk, copied);
     return copied;
     ...
}
```

chenhengqi:
Could you elaborate ? I don't see tcpdrop do anything to do with tcp_cleanup_rbuf.

sc07kvm:
Of course, on newer kernels(>=5.11) this happens via the tcp_recvmsg_locked function, which is inlined:
```c
static int tcp_recvmsg_locked() {
    ...
    do {
        ...
        tcp_cleanup_rbuf(...);
        ...
    } while (...);
     
    tcp_cleanup_rbuf(...);
    ...
}

int tcp_recvmsg(...) {
    ...
    ret = tcp_recvmsg_locked(...);
    ...
}
```

If you look at older versions, you can see the call directly in tcp_recvmsg - https://github.com/torvalds/linux/blob/2c85ebc57b3e1817b6ce1a6b703928e113a90442/net/ipv4/tcp.c#L2159

chenhengqi:
How does this affect the behavior of tcpdrop.py ?

sc07kvm:
tcpdrop.py hooks **tcp_cleanup_rbuf** to count the amount of data received based on the **"copied"** argument:
```c
int kprobe__tcp_cleanup_rbuf(..., int copied) {
    ...
    ipv4_recv_bytes.increment(ipv4_key, copied);
    ....
}
```

For example, if **tcp_cleanup_rbuf** function is called twice within a single **tcp_recvmsg** call with **"copied"** argumen of 10 and 20, the **ipv4_recv_bytes** counter will end up being incremented by 30 instead of 20.:
```
--> tcp_recvmsg()
    --> tcp_cleanup_rbuf(..., copied = 10)
    --> tcp_cleanup_rbuf(..., copied = 20)
```

chenhengqi:
OK, could you help with a PR ?

sc07kvm:
I have no idea how to fix this at all

[...]

PR #5375, opened 2025-07-20
tools/tcptop: rewrite using fentry/fexit for accurate recv statistics

Sutter099:
This PR improves the accuracy of TCP receive statistics in `tcptop`.

Previously, `tcptop` used a kprobe on `tcp_cleanup_rbuf` to track receive traffic. However, this function may be called multiple times in a single `tcp_recvmsg()` call, with an accumulating `copied` parameter. This can lead to overcounting.

This issue was reported by @sc07kvm:
https://github.com/iovisor/bcc/issues/5352

To resolve this, I followed the suggestion from @yonghong-song:

    That is really unfortunate. But since it is not accurate any way, so I think we could move current tcptop.py to old directory and add fentry/fexit support in tcptop.py and add comments in the beginning of the tool to mention it only works for kernel >= 5.5.

As a result:
- The original kprobe-based `tcptop.py` has been moved to `tools/old/`
- A new version based on `fentry/fexit` hooks is introduced
- Send/receive tracking is unified under `tcp_stat()` with shared logic
- Receive stats now hook into `tcp_recvmsg` with proper entry/exit coordination

### DIFF
--- a/tools/tcptop.py
+++ b/tools/tcptop.py
@@ -166,16 +166,15 @@ static int tcp_sendstat(int size)
     return 0;
 }
 
-int tcp_send_ret(struct pt_regs *ctx)
+KRETFUNC_PROBE(tcp_sendmsg, struct sock *sk, struct msghdr *msg, size_t size, int ret)
 {
-    int size = PT_REGS_RC(ctx);
-    if (size > 0)
-        return tcp_sendstat(size);
+    if (ret > 0)
+        return tcp_sendstat(ret);
     else
         return 0;
 }
 
-int tcp_send_entry(struct pt_regs *ctx, struct sock *sk)
+KFUNC_PROBE(tcp_sendmsg, struct sock *sk, struct msghdr *msg, size_t size)
 {
     if (container_should_be_filtered()) {
         return 0;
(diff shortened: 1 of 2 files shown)


yonghong-song:
Thanks! I checked the goal and also run the new tcptop.py and LGTM.

tools/tcptop.py
```diff
@@ -195,47 +209,25 @@ def range_check(string):
  * - misses tcp_read_sock() traffic
  * we'd much prefer tracepoints once they are available.
  */
-int kprobe__tcp_cleanup_rbuf(struct pt_regs *ctx, struct sock *sk, int copied)
+KRETFUNC_PROBE(tcp_recvmsg, struct sock *sk, struct msghdr *msg, size_t len, int flags, int *addr_len, int ret)
```
sc07kvm, line 212: Will only work with 5.19. Since the function signature has changed: https://github.com/torvalds/linux/commit/ec095263a965720e1ca39db1d9c5cd47846c789b

Sutter099: Indeed, only works for linux over v5.19. Change the comment from v5.5 to v5.19?

sc07kvm: or support both options =)

PR #5380, opened 2025-07-29
python: tcp.py: add state2str() and apply to tools

Rtoax:
There are two tools/{tcpretrans.py,tcpstates.py} display the TCP state, and those code already exist in bcc/tcp.py, thus, this patch remove the duplicate code and add bcc.tcp.state2str() function.


src/python/bcc/tcp.py
```diff
@@ -56,3 +56,9 @@ def flags2str(flags):
     if flags & TCPHDR_CWR:
         arr.append("CWR")
     return "|".join(arr)
+
+def state2str(state):
+    if state in tcpstate:
+        return tcpstate[state]
+    else:
+        return str(state)
```
ekyooo: It can be made more concise as follows:
```
def state2str(state):
    return _tcpstate.get(state, str(state))
```

tools/tcpretrans.py
```diff
@@ -362,9 +347,9 @@ def print_ipv4_event(cpu, data, size):
         "%s:%s" % (inet_ntop(AF_INET, pack('I', event.daddr)), event.dport)),
         end='')
     if args.sequence:
-        print(" %-12s %s" % (tcpstate[event.state], event.seq))
+        print(" %-12s %s" % (tcp.tcpstate[event.state], event.seq))
```
ekyooo: How about using the state2str function you created?

src/python/bcc/tcp.py
```diff
@@ -56,3 +56,9 @@ def flags2str(flags):
     if flags & TCPHDR_CWR:
         arr.append("CWR")
     return "|".join(arr)
+
+def state2str(state):
+    if state in tcpstate:
+        return tcpstate[state]
```
ekyooo: How about renaming tcpstate to _tcpstate on line 16 to conventionally indicate that it is for module-internal use?
This way, we can guide users to access the state only through the state2str method instead of direct variable access, which helps future maintenance easier if the internal implementation changes. 

```
_tcpstate = {
    1: "ESTABLISHED",
    2: "SYN_SENT",
    3: "SYN_RECV",
    4: "FIN_WAIT1",
    5: "FIN_WAIT2",
    6: "TIME_WAIT",
    7: "CLOSE",
    8: "CLOSE_WAIT",
    9: "LAST_ACK",
    10: "LISTEN",
    11: "CLOSING",
    12: "NEW_SYN_RECV",
}
```

FYI, tcpdrop.py also has code that directly accesses tcp.tcpstate.

src/python/bcc/tcp.py
Rtoax: Fixed, thanks.

src/python/bcc/tcp.py
Rtoax: Fixed, thanks.

PR #5381, opened 2025-07-30
tools/kvmexit: implement AMD support and parallel post processing

mtp401:
This PR implements a few enhancements for the kvmexit tool.

First, it adds support for AMD's virtualization extensions (SVM) so the tool now works for both AMD and Intel based VMs.

Second, it parallelizes post processing of the VM exit counts in order to reduce the amount of time needed for this step on high core count machines. I included this data in the commit message for the relevant commit but on a 640 core machine this feature reduces the time taken to post process the results from ~37 minutes to ~5 minutes.

Finally, it adds a new `--debug` argument and disables BPF compilation warnings if this argument is not specified. I have been running this tool within a few different environments and having a variable set of warnings emitted in those different contexts makes automated data collection challenging. Disabling BPF compilations warnings makes this more straightforward and reduces visual noise when running this tool manually.

### DIFF
--- a/tools/kvmexit.py
+++ b/tools/kvmexit.py
@@ -26,6 +26,7 @@
 # 31-Aug-2021  Fei Li <lifei.shirley@bytedance.com>  Initial implementation.
 # 28-Jul-2025  Matt Pelland <mpelland@akamai.com>    Implement support for AMD.
 # 28-Jul-2025  Matt Pelland <mpelland@akamai.com>    Parallelize postprocessing.
+# 28-Jul-2025  Matt Pelland <mpelland@akamai.com>    Silence compiler warnings.
 
 from __future__ import print_function
 from time import sleep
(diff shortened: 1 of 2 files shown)


tools/kvmexit.py
```diff
@@ -163,97 +162,193 @@ def valid_args_list(args):
...
-    "N/A",
-    "N/A",
-    "N/A",
-    "N/A",
-    "N/A",
-    "BUS_LOCK",
-    "NOTIFY",
-)
+# Defines Intel (VMX) VM exit reason codes. Keep this in sync with
+# tools/arch/x86/include/uapi/asm/vmx.h.
```
chenhengqi: `tools` ?

tools/kvmexit.py
```diff
@@ -163,97 +162,193 @@ def valid_args_list(args):
...
 try:
-    # Currently, only adapte on intel architecture
-    cmd = "cat /proc/cpuinfo | grep vendor_id | head -n 1"
-    arch_info = subprocess.check_output(cmd, shell=True).strip()
-    if b"Intel" in arch_info:
-        pass
-    else:
-        raise Exception("Currently we only support Intel architecture, please do expansion if needs more.")
+    cmd = "grep vendor_id /proc/cpuinfo | head -n 1"
+    cpu_vendor_raw = subprocess.check_output(cmd, shell=True).decode()
```
chenhengqi: How about just read and process the `/proc/cpuinfo` file without a shell ?

tools/kvmexit.py
mtp401: Good catch, fixed in the latest iteration.

tools/kvmexit.py
mtp401: I switched to a Python native implementation in the latest version.

PR #5383, opened 2025-08-04
Removing trailing white spaces and fixing typo

Bojun-Seo:
Trailing whitespaces are removed from all files, empty lines and image files are preserved. This patch is created using the following shell commands:
```
find . -type f -not -path '*/\.*' -exec sed -i '/^$/!s/[ \t]*$//' {} \;
```

fixing typo
Before: its possible to ~
After: it's possible to ~

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -2616,7 +2616,7 @@ cannot call GPL only function from proprietary program
 
 eBPF program compilation needs kernel sources or kernel headers with headers
 compiled. In case your kernel sources are at a non-standard location where BCC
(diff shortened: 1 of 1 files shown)


chenhengqi:
I personally don't like this kind of whitespace changes(for non first-time contributor). Please invest your efforts on something more meaningful.

yonghong-song:
Okay. I will merge this since it does makes format (esp. at 'git' level) better across the repo. Hopefully in the future, we do not have such format issues.

Agree with @chenhengqi that actually fixing the issue and had new features will be much better.

PR #5391, opened 2025-08-18
memleak:support the malloc and free uprobe of third-party allocation libraries

xpbob:
We found that third-party allocation libraries have custom probe analysis allocation instead of the default malloc, such as jemalloc 5.3.0. Inline malloc fastpath into operator new（https://github.com/jemalloc/jemalloc/commit/edbfe6912c1b7e8b561dfee1b058425de6c06285#diff-e84aa11907127083071fc3046183dd381618df08a67a609128f5d750874ab251）. The observed uprobe has changed. Add a way to customize the uprobe entry to support third-party allocation libraries

### DIFF
--- a/tools/memleak.py
+++ b/tools/memleak.py
@@ -117,6 +117,10 @@ parser.add_argument("--sort", type=str, default="size",
         help="report sorted in given key; available key list: size, count")
 parser.add_argument("--symbols-prefix", type=str,
         help="memory allocator symbols prefix")
+parser.add_argument("--costom-malloc-uprobe", type=str,
+        help="costom memory allocator uprobe")
+parser.add_argument("--costom-free-uprobe", type=str,
+        help="costom memory free uprobe")
 
 args = parser.parse_args()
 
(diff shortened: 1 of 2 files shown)


xpbob:
The failed task was not caused by the current modification

ekyooo (commented):
The -S option in libbpf-tools/memleak.c seems to provide the same functionality. 
The option differs slightly from this patch. 
How about using the same option in both the Python and libbpf-tools versions of memleak? Would there be any issues with this approach?

tools/memleak.py
```diff
@@ -117,6 +117,10 @@ def run_command_get_pid(command):
         help="report sorted in given key; available key list: size, count")
 parser.add_argument("--symbols-prefix", type=str,
         help="memory allocator symbols prefix")
+parser.add_argument("--costom-malloc-uprobe", type=str,
```
Bojun-Seo, line 120: `costom`? please check typo.

xpbob:
thanks for review
Let's turn off pr first. There are better solutions

PR #5395, opened 2025-08-29
clang: Handle empty kheaders directory gracefully, fix race condition

### LINKED ISSUE #4213: Inconsistent `/tmp/kheaders` directory makes bcc fail
I just spent way too long debugging why `execsnoop` failed with this error:

```
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
  File "/nix/store/7dgdpsd8lrclqzl4s5vmlm8hwm0y91d2-bcc-0.24.0/share/bcc/tools/.execsnoop-wrapped", line 228, in <module>
    b = BPF(text=bpf_text)
  File "/nix/store/7dgdpsd8lrclqzl4s5vmlm8hwm0y91d2-bcc-0.24.0/lib/python3.10/site-packages/bcc/__init__.py", line 475, in __init__
```

The reason was that I had a directory `/tmp/kheaders-5.18.15` containing nothing but an empty `include/uapi` directory. I have no idea how this happened. `rm -r /tmp/kheaders-5.18.15` fixed it.

A few ideas on what needs to be done:

- make sure the kheaders directory is always in a consistent state when the program exits
- improve the existence check, à la https://github.com/iovisor/bcc/pull/3588

I'm on Linux 5.18.15, NixOS unstable, BCC 0.24.0.

Lekensteyn:
Do not assume an existing `/tmp/kheaders-$(uname -r)` directory to contain headers. By default, systemd removes files older than 10 days via tmpfiles config. If a header file is missing, proceed by removing the old directory before attempting a new extraction.

This also fixes a vulnerability where bcc could be convinced to use a kheaders directory of an unprivileged user:

 1. Attacker creates a kheaders dir symlink to a non-existing file. This bypasses file_exists and non-root user checks in get_proc_kheaders.
 2. extract_kheaders proceeds to extract a headers tarball to a temporary directory. Within this race window, an attacker can remove the symlink, and replace it by a directory under its control.
 3. extract_kheaders returns success even though the directory is under control of an attacker:

        // fails because the attacker created the directory.
        ret = rename(dirpath_tmp, dirpath.c_str());
        if (ret)
          // OOPS: return code is overwritten when tmp dir is removed
          ret = system(("rm -rf " + std::string(dirpath_tmp)).c_str());

Fix by ignoring temporary directory removal failures, and by checking whether a parallel process created a legitimate privileged directory with the expected header files.

All privileged directory checks do not follow symlinks. Remove the owner check for PROC_KHEADERS_PATH as we assume /sys to be trusted.

Fixes #4213

Fixes: 008ea09e8911 ("clang: check header ownership (#4928)")

### DIFF
--- a/src/cc/frontends/clang/kbuild_helper.cc
+++ b/src/cc/frontends/clang/kbuild_helper.cc
@@ -153,22 +153,31 @@ static inline bool file_exists(const char *f)
   return (stat(f, &buffer) == 0);
 }
 
-static inline bool file_exists_and_ownedby(const char *f, uid_t uid)
+static inline int check_privileged_dir(const char *f)
 {
-  struct stat buffer;
-  int ret = stat(f, &buffer) == 0;
-  if (ret) {
-    if (buffer.st_uid != uid) {
-      std::cout << "ERROR: header file ownership unexpected: " << std::string(f) << "\n";
-      return false;
-    }
+  struct stat buf;
+  if (lstat(f, &buf) != 0)
+    return -errno;
+
+  if (!S_ISDIR(buf.st_mode)) {
+    std::cout << "ERROR: not a directory: " << std::string(f) << "\n";
+    return -ENOTDIR;
   }
-  return ret;
+  if (buf.st_uid != 0) {
+    std::cout << "ERROR: not owned by root: " << std::string(f) << "\n";
+    return -EACCES;
+  }
+  return 0;
 }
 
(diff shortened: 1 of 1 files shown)


Ecordonnier:
@Lekensteyn do you need help in getting this merged? I'm guessing the failing CI needs to be fixed before the PR can be reviewed?

PR #5399, opened 2025-09-17
Fix exit codes in BCC python tools for reliable error handling

HoyeonRhee:
Some BCC tools currently exit with status code 0 even when encountering
fatal errors. This makes it impossible to distinguish between success and
failure, which makes smoke tests (and packaging checks) unreliable.

Examples are:
- No kprobe or tracepoint event attached
- Missing required symbols in /proc/kallsyms
- Invalid or conflicting command-line options

This PR updates the BCC python tools to consistently return exit(1) in
these cases. This ensures errors are clearly reported and allows the
tools to be used more effectively in automated workflows such as smoke
testing.

### DIFF
--- a/tools/biolatency.py
+++ b/tools/biolatency.py
@@ -65,7 +65,7 @@ debug = 0
 
 if args.flags and args.disks:
     print("ERROR: can only use -D or -F. Exiting.")
-    exit()
+    exit(1)
 
 # define BPF program
 bpf_text = """
--- a/tools/biolatency.py
+++ b/tools/biolatency.py
@@ -294,7 +294,7 @@ if args.flags and (tp_start or tp_done):
             # Some flags are accessible in the rwbs field (RAHEAD, SYNC and META)
             # but other aren't. Disable the -F option for tracepoint for now.
             print("ERROR: blk_account_io_start/done probes not available. Can't use -F.")
-            exit()
+            exit(1)
 
 if tp_start:
     bpf_text += tp_start_text.replace("START_TP", tp_start)
--- a/tools/biosnoop.py
+++ b/tools/biosnoop.py
@@ -344,7 +344,7 @@ if not tp_start:
         b.attach_kprobe(event="blk_account_io_start", fn_name="trace_pid_start")
     else:
         print("ERROR: No found any block io start probe/tp.")
-        exit()
+        exit(1)
 
 if BPF.get_kprobe_functions(b'blk_start_request'):
     b.attach_kprobe(event="blk_start_request", fn_name="trace_req_start")
@@ -357,7 +357,7 @@ if not tp_done:
         b.attach_kprobe(event="blk_account_io_done", fn_name="trace_req_completion")
     else:
         print("ERROR: No found any block io done probe/tp.")
-        exit()
+        exit(1)
 
 
 # header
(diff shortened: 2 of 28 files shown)


tools/biolatency.py
```diff
@@ -65,7 +65,7 @@
 
 if args.flags and args.disks:
     print("ERROR: can only use -D or -F. Exiting.")
-    exit()
+    exit(1)
```
ekyooo, line 68: How about differentiating between usage errors (e.g., invalid option) and environmental failures (e.g., /proc/kallsyms missing) instead of collapsing everything into exit(1)?

I think this preserves backward compatibility, avoids magic numbers, and keeps expansion centralized in tools/exitcodes.py.

Example implementation: tools/exitcodes.py
```
"""
Basic exit status codes for BCC command-line tools.

0: Success
1: General runtime / environment / internal error
2: Command-line / usage error (argument/option parsing error, argparse SystemExit, etc.)

If future expansion (finer-grained exit codes) is needed, add new constants here.
This module is not for handling internal exceptions inside the bcc.* library;
it is intended to standardize process exit policy for the standalone tools.
"""

class ExitCodes:
    SUCCESS = 0
    GENERAL_ERROR = 1
    USAGE_ERROR = 2
```
 
For future needs, we can expand error handling step by step as new requirements emerge as the following example:
```
     # Optional refined codes (subset of GENERAL_ERROR)
    MISSING_RESOURCE = 3
    PERMISSION_DENIED = 4
    UNSUPPORTED_KERNEL_FEATURE = 5
    VERIFIER_REJECTED = 6
```

What do you think?

HoyeonRhee: While consistency is a reasonable goal, adding a centralized `exitcodes.py` for the Python tools is unlikely to help in practice.

Current BCC Python layer does not expose structured error types. (Aside from `USDTException`, everything is a generic `Exception`.) As a result, tools cannot reliably map conditions like `UNSUPPORTED_KERNEL_FEATURE` or `VERIFIER_REJECTED` to distinct exit codes. 

In practice, each tool fails for different reasons, so `exitcodes.py` would need to be updated whenever a new tool-specific case appears, creating ongoing maintenance burden without clear benefit.

The intended use here is packaging and smoke tests, where the useful distinction is simply success versus failure. Users typically read error messages, not numeric codes. Ensuring errors return a nonzero exit code already solves the practical problem. 

Finer-grained codes do not add clear value unless the library first exposes well-defined exception types that tools can depend on.

ekyooo: This patch fixes the immediate smoke test issue and is acceptable as-is.  
My comments focus on preparing for future extensibility.

1. Constants (0/1 or 0/1/2):
Replacing magic numbers with constants improves readability, simplifies refactoring, and enables easier future extension with minimal maintenance overhead.

2. Further Refinement (Future Consideration):
I agree with your point that the current BCC Python layer does not expose structured error types, which limits the immediate utility of more detailed exit codes.
While current smoke tests may not require anything beyond 0/1, in the long term, leveraging more granular exit codes could strengthen test coverage and improve maintainability.
For example, distinguishing between unsupported kernel features and environment-related issues could reduce triage time and streamline CI results.
I believe that new exit codes should only be added when specific test cases or recurring failure patterns justify them. If such needs arise, they could also provide a strong case for improving the Python library’s structured exception handling.
Defining exitcode.py in its current form could serve as a practical starting point for iterating on this approach. It allows us to explore and validate the potential benefits of more detailed exit codes in a controlled manner.

Given the current scope (smoke test fix), I’m fine accepting this patch as it stands.

PR #5400, opened 2025-09-19
tools/filetop: Add --recursive for directory filter

wswsmao:
The commit https://github.com/iovisor/bcc/commit/137bd5fbc6f7c779ef81f8627fe69552ce4c297b introduced a directory filter. However, files under subdirectories were not being traced. 
This PR adds a --recursive option which, while remaining backward-compatible, enables recursive filtering so that files in subdirectories are also matched.

### DIFF
--- a/tools/filetop.py
+++ b/tools/filetop.py
@@ -24,14 +24,15 @@ from subprocess import call
 
 # arguments
 examples = """examples:
-    ./filetop                 # file I/O top, 1 second refresh
-    ./filetop -C              # don't clear the screen
-    ./filetop -p 181          # PID 181 only
-    ./filetop -d /home/user   # trace files in /home/user directory only
-    ./filetop 5               # 5 second summaries
-    ./filetop 5 10            # 5 second summaries, 10 times only
-    ./filetop 5 --read-only   # 5 second summaries, only read operations traced
-    ./filetop 5 --write-only  # 5 second summaries, only write operations traced
+    ./filetop                    # file I/O top, 1 second refresh
+    ./filetop -C                 # don't clear the screen
+    ./filetop -p 181             # PID 181 only
+    ./filetop -d /home/user      # trace files in /home/user directory only
+    ./filetop -d /home/user -R   # trace files in /home/user and subdirectories
+    ./filetop 5                  # 5 second summaries
+    ./filetop 5 10               # 5 second summaries, 10 times only
+    ./filetop 5 --read-only      # 5 second summaries, only read operations traced
+    ./filetop 5 --write-only     # 5 second summaries, only write operations traced
 """
 parser = argparse.ArgumentParser(
     description="File reads and writes by process",
@@ -60,6 +61,8 @@ parser.add_argument("--ebpf", action="store_true",
     help=argparse.SUPPRESS)
 parser.add_argument("-d", "--directory", type=str,
     help="trace this directory only")
+parser.add_argument("-R", "--recursive", action="store_true",
+    help="when used with -d, also trace files in subdirectories")
 
 args = parser.parse_args()
 interval = int(args.interval)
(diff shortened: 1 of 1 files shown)


tools/filetop.py
```diff
@@ -180,12 +183,25 @@
 if args.directory:
     try:
         directory_inode = os.lstat(args.directory)[stat.ST_INO]
-        print(f'Tracing directory: {args.directory} (Inode: {directory_inode})')
-        bpf_text = bpf_text.replace('DIRECTORY_FILTER',  'file->f_path.dentry->d_parent->d_inode->i_ino != %d' % directory_inode)
+        if args.recursive:
+            print(f'Tracing directory recursively: {args.directory} (Inode: {directory_inode})')
+            directory_filter = ("({ struct dentry *pde = de; int found = 0; "
+                               "for (int i = 0; i < 50; i++) { "
```
ekyooo: 1. Please replace the magic number 50 with a macro or constant for clarity and maintainability.

2. Is there a specific reason why 50 is chosen as the max depth? Could you also include your thoughts on performance, BPF instruction count, and verifier safety?

tools/filetop.py
```diff
@@ -180,12 +183,25 @@
 if args.directory:
     try:
         directory_inode = os.lstat(args.directory)[stat.ST_INO]
-        print(f'Tracing directory: {args.directory} (Inode: {directory_inode})')
-        bpf_text = bpf_text.replace('DIRECTORY_FILTER',  'file->f_path.dentry->d_parent->d_inode->i_ino != %d' % directory_inode)
+        if args.recursive:
+            print(f'Tracing directory recursively: {args.directory} (Inode: {directory_inode})')
+            directory_filter = ("({ struct dentry *pde = de; int found = 0; "
```
ekyooo: The directory_filter code for the recursive case is hard to read. 
Could you please refactor it for better readability and maintainability?

wswsmao: done

tools/filetop.py
wswsmao: 1、done
2、I switched the bound to 32 and made it a macro FILETOP_MAX_DENTRY_DEPTH. 32 aligns with existing path-walking limits in this repo (e.g., libbpf-tools uses MAX_PATH_DEPTH=32) and is well above typical directory depths we see in practice.

tools/filetop.py
```diff
@@ -80,6 +83,28 @@
 #undef DNAME_INLINE_LEN
 #define DNAME_INLINE_LEN __BCC_DNAME_INLINE_LEN
 
+// Limit dentry parent walk depth to satisfy eBPF verifier loop
+#ifndef FILETOP_MAX_DENTRY_DEPTH
+#define FILETOP_MAX_DENTRY_DEPTH 32
```
ekyooo: While FILETOP_MAX_DENTRY_DEPTH is below 32 in most cases, there is still a chance that it could exceed this limit in certain scenarios. If that happens, the tool's behavior might deviate from the user's expectations. Would it be worth notifying users about this limitation to avoid potential confusion?

wswsmao: done, add max depth log in `--help` and `Tracing directory recursively` log

tools/filetop.py
```diff
@@ -80,6 +86,28 @@
...
+// Limit dentry parent walk depth to satisfy eBPF verifier loop
+#ifndef MAX_DENTRY_DEPTH
+#define MAX_DENTRY_DEPTH MAX_DENTRY_DEPTH_VALUE
+#endif
+
+static __always_inline int not_under_inode(struct dentry *de, unsigned long target_ino)
+{
+    struct dentry *pde = de;
+    int found = 0;
+    #pragma unroll
```
ekyooo, line 98: Please add a blank line after the variable declaration section.

tools/filetop.py
```diff
@@ -180,12 +211,24 @@
...
+            print(f'Tracing directory: {args.directory} (Inode: {directory_inode})')
+            directory_filter = "de->d_parent->d_inode->i_ino != %d" % directory_inode
+
+        bpf_text = bpf_text.replace('DIRECTORY_FILTER', directory_filter)
     except (FileNotFoundError, PermissionError) as e:
         print(f'Error accessing directory {args.directory}: {e}')
         exit(1)
 else:
+    if args.recursive:
+        print("Error: --recursive can only be used with -d/--directory option")
```
ekyooo, line 230: How about `-R/--recursive`?

tools/filetop.py
```diff
@@ -60,6 +64,8 @@
     help=argparse.SUPPRESS)
 parser.add_argument("-d", "--directory", type=str,
     help="trace this directory only")
+parser.add_argument("-R", "--recursive", action="store_true",
+    help=f"when used with -d, also trace files in subdirectories (max depth {MAX_DENTRY_DEPTH_VAL})")
```
ekyooo, line 68: It would be better to make it clearer that this is only available when using -d. The current wording makes it seem like there could be cases where it's possible to use without -d. 
For example, "This feature is only supported when used with the -d option, tracing files inside subdirectories."

tools/filetop.py
ekyooo, line 98: Is `#pragma unroll` (and `__always_inline`) intended?
please refer to:
https://lwn.net/Articles/1017116/

PR #5401, opened 2025-09-28
Avoid using deprecated string triple TargetRegistry overloads

boomanaiden154:
These functions calls are deprecated and will most likely be removed in LLVM version 23.

### DIFF
--- a/src/cc/bcc_debug.cc
+++ b/src/cc/bcc_debug.cc
@@ -126,21 +126,34 @@ void SourceDebugger::dump() {
     return;
   }
 
+#if LLVM_VERSION_MAJOR >= 22
+  std::unique_ptr<MCRegisterInfo> MRI(T->createMCRegInfo(TheTriple));
+#else
   std::unique_ptr<MCRegisterInfo> MRI(T->createMCRegInfo(TripleStr));
+#endif
   if (!MRI) {
     errs() << "Debug Error: cannot get register info\n";
     return;
   }
 
   MCTargetOptions MCOptions;
+#if LLVM_VERSION_MAJOR >= 22
+  std::unique_ptr<MCAsmInfo> MAI(T->createMCAsmInfo(*MRI, TheTriple, MCOptions));
+#else
   std::unique_ptr<MCAsmInfo> MAI(T->createMCAsmInfo(*MRI, TripleStr, MCOptions));
+#endif
   if (!MAI) {
     errs() << "Debug Error: cannot get assembly info\n";
     return;
   }
 
+#if LLVM_VERSION_MAJOR >+ 22
+  std::unique_ptr<MCSubtargetInfo> STI(
+      T->createMCSubtargetInfo(TheTriple, "", ""));
+#else
   std::unique_ptr<MCSubtargetInfo> STI(
       T->createMCSubtargetInfo(TripleStr, "", ""));
+#endif
   MCObjectFileInfo MOFI;
 #if LLVM_VERSION_MAJOR >= 13
   MCContext Ctx(TheTriple, MAI.get(), MRI.get(), STI.get(), nullptr);


src/cc/bcc_debug.cc
```diff
@@ -126,21 +126,34 @@ void SourceDebugger::dump() {
...
+  std::unique_ptr<MCAsmInfo> MAI(T->createMCAsmInfo(*MRI, TheTriple, MCOptions));
+#else
   std::unique_ptr<MCAsmInfo> MAI(T->createMCAsmInfo(*MRI, TripleStr, MCOptions));
+#endif
   if (!MAI) {
     errs() << "Debug Error: cannot get assembly info\n";
     return;
   }
 
+#if LLVM_VERSION_MAJOR >+ 22
```
ekyooo: `>=` ?

src/cc/bcc_debug.cc
```diff
@@ -126,21 +126,34 @@ void SourceDebugger::dump() {
...
   std::unique_ptr<MCRegisterInfo> MRI(T->createMCRegInfo(TripleStr));
+#endif
   if (!MRI) {
     errs() << "Debug Error: cannot get register info\n";
     return;
   }
 
   MCTargetOptions MCOptions;
+#if LLVM_VERSION_MAJOR >= 22
+  std::unique_ptr<MCAsmInfo> MAI(T->createMCAsmInfo(*MRI, TheTriple, MCOptions));
```
ekyooo: Do you happen to know which LLVM version first introduced support for the Triple-based overloads?

I'm asking because if our minimum supported LLVM version already includes them, we could simplify the code by removing the conditional.

Thank you.

src/cc/bcc_debug.cc
boomanaiden154: Fixed.

src/cc/bcc_debug.cc
boomanaiden154: The triple overloads for these functions was only introduced in LLVM 22 which hasn't had a release cut yet.

ekyooo: Thank you for the information.
I’ve checked the details in LLVM's "commit 67823469bd1c5f5416c4b009a895dcaffa971b69."

1. Could this approach also be applied to lookupTarget? It seems this function also has a version that accepts a Triple type.
2. What do you think about defining a macro like the example below to minimize conditional branching?
```
#if LLVM_VERSION_MAJOR >= 22
#define TRIPLE_ARG TheTriple
#else
#define TRIPLE_ARG TripleStr
#endif
```

boomanaiden154: > Could this approach also be applied to lookupTarget? It seems this function also has a version that accepts a Triple type.

It could, but that one is not yet deprecated and there needs to be some cleanup done upstream first before that one is ready. I'll post another PR when I get to that.

> What do you think about defining a macro like the example below to minimize conditional branching?

I'm hesitant to do it with a macro given it doesn't respect scope. We're passing around local variables, so have something scoped keeps things correct. I don't think there are enough instances here to justify a macro/substitution.

boomanaiden154: Also threw `lookupTarget` in here too since we're deprecating that one when LLVM 22 branches as well.

ekyooo: > Also threw lookupTarget in here too since we're deprecating that one when LLVM 22 branches as well.

Good, thank you.


> I'm hesitant to do it with a macro given it doesn't respect scope. We're passing around local variables, so have something scoped keeps things correct. I don't think there are enough instances here to justify a macro/substitution.

My purpose is to reduce duplicated code and make code maintenance easier, so any appropriate method that satisfies that purpose would be fine.

What do you think about the following approach?

```
#if LLVM_VERSION_MAJOR >= 22
  const auto &TripleArg = TheTriple;     // llvm::Triple
#else
  const auto &TripleArg = TripleStr;     // std::string
#endif
```

boomanaiden154: Updated.

ekyooo:
LGTM. Thank you.

PR #5402, opened 2025-10-06
libbpf-tools/syncsnoop: Add support for sync_file_range2

jeromemarchand:
Some architectures like ppc64 don't use the sync_file_range syscall, but sync_file_range2 instead and that makes syncsnoop fails on those arches. Add support for it.

Fixes the following error:
libbpf: failed to determine tracepoint 'syscalls/sys_enter_sync_file_range' perf event ID: No such file or directory libbpf: prog 'tracepoint__syscalls__sys_enter_sync_file_range': failed to create tracepoint 'syscalls/sys_enter_sync_file_range' perf event: No such file or directory libbpf: prog 'tracepoint__syscalls__sys_enter_sync_file_range': failed to auto-attach: -2 failed to attach BPF object

### DIFF
--- a/libbpf-tools/syncsnoop.bpf.c
+++ b/libbpf-tools/syncsnoop.bpf.c
@@ -52,6 +52,18 @@ void tracepoint__syscalls__sys_enter_sync_file_range(struct trace_event_raw_sys_
 	__syscall(ctx, SYS_SYNC_FILE_RANGE);
 }
 
+SEC("tracepoint/syscalls/sys_enter_sync_file_range2")
+void tracepoint__syscalls__sys_enter_sync_file_range2(struct trace_event_raw_sys_enter *ctx)
+{
+	__syscall(ctx, SYS_SYNC_FILE_RANGE2);
+}
+
+SEC("tracepoint/syscalls/sys_enter_arm_sync_file_range")
+void tracepoint__syscalls__sys_enter_arm_sync_file_range(struct trace_event_raw_sys_enter *ctx)
+{
+	__syscall(ctx, SYS_ARM_SYNC_FILE_RANGE);
+}
+
 SEC("tracepoint/syscalls/sys_enter_syncfs")
 void tracepoint__syscalls__sys_enter_syncfs(struct trace_event_raw_sys_enter *ctx)
 {
--- a/libbpf-tools/syncsnoop.c
+++ b/libbpf-tools/syncsnoop.c
@@ -10,6 +10,7 @@
 #include <bpf/libbpf.h>
 #include "syncsnoop.h"
 #include "syncsnoop.skel.h"
+#include "trace_helpers.h"
 
 #define PERF_BUFFER_PAGES       16
 #define PERF_POLL_TIMEOUT_MS    100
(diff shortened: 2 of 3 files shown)


libbpf-tools/syncsnoop.c
```diff
@@ -101,9 +102,21 @@ int main(int argc, char **argv)
 
 	libbpf_set_print(libbpf_print_fn);
 
-	obj = syncsnoop_bpf__open_and_load();
-	if (!obj) {
-		fprintf(stderr, "failed to open and load BPF object\n");
+        obj = syncsnoop_bpf__open();
```
ekyooo: Please check the indentation and blank line overall.

libbpf-tools/syncsnoop.c
```diff
@@ -101,9 +102,21 @@ int main(int argc, char **argv)
...
+        if (!obj) {
+		fprintf(stderr, "failed to open BPF object\n");
+		return 1;
+	}
+
+        if (!tracepoint_exists("syscalls", "sys_enter__sync_file_range"))
```
ekyooo: Please check the number of underscores.

libbpf-tools/syncsnoop.c
```diff
@@ -101,9 +102,21 @@ int main(int argc, char **argv)
...
+		bpf_program__set_autoload(obj->progs.tracepoint__syscalls__sys_enter_sync_file_range, false);
+        if (!tracepoint_exists("syscalls", "sys_enter__sync_file_range2"))
```
ekyooo: What about using else here?

libbpf-tools/syncsnoop.c
jeromemarchand: Hmm. there is something wrong with my editor setting ATM. Will fix.

libbpf-tools/syncsnoop.c
jeromemarchand: Oops! Will fix.

libbpf-tools/syncsnoop.c
jeromemarchand: This might not work on arm system with an older kernel where this syscall was called arm_sync_file_range().
On the other end, not failing here, means the tools will silently skip file range synchronization events.
Best solution would probably to add that third syscall here too.

libbpf-tools/syncsnoop.bpf.c
```diff
@@ -52,6 +52,18 @@ void tracepoint__syscalls__sys_enter_sync_file_range(struct trace_event_raw_sys_
...
+SEC("tracepoint/syscalls/sys_enter_sync_file_range2")
+void tracepoint__syscalls__sys_enter_sync_file_range2(struct trace_event_raw_sys_enter *ctx)
+{
+	__syscall(ctx, SYS_SYNC_FILE_RANGE2);
+}
+
+SEC("tracepoint/syscalls/sys_enter_arm_sync_file_range")
+void tracepoint__syscalls__sys_enter_arm_sync_file_range(struct trace_event_raw_sys_enter *ctx)
+{
+	__syscall(ctx, SYS_SYNC_FILE_RANGE);
```
ekyooo: SYS_ARM_SYNC_FILE_RANGE?

libbpf-tools/syncsnoop.c
```diff
@@ -101,9 +102,22 @@ int main(int argc, char **argv)
...
 
-	obj = syncsnoop_bpf__open_and_load();
+	obj = syncsnoop_bpf__open();
 	if (!obj) {
-		fprintf(stderr, "failed to open and load BPF object\n");
+		fprintf(stderr, "failed to open BPF object\n");
+		return 1;
+	}
+
+	if (!tracepoint_exists("syscalls", "sys_enter_sync_file_range"))
```
ekyooo, line 111: How about failing as the tool previously did when none of the tracepoints are found?

libbpf-tools/syncsnoop.bpf.c
jeromemarchand: Indeed.

libbpf-tools/syncsnoop.c
jeromemarchand: I assume you mean if none of the *sync_file_range* tracepoints are found. It still should fail if any other tracepoints are missing.

PR #5403, opened 2025-10-07
tools: Don't use the old bpf_probe_read() helper.

jeromemarchand:
The old bpf_probe_read() helper doesn't work on s390 arch. Use bpf_probe_read_user/kernel() instead.

### DIFF
--- a/tools/tcpconnect.py
+++ b/tools/tcpconnect.py
@@ -314,7 +314,7 @@ int trace_udp_ret_recvmsg(struct pt_regs *ctx)
         return 0;
 
     void *iovbase = msghdr->msg_iter.IOV_FIELD->iov_base;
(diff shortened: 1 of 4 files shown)


tools/tcpconnect.py
```diff
@@ -314,7 +314,7 @@
         return 0;
 
     void *iovbase = msghdr->msg_iter.IOV_FIELD->iov_base;
-    bpf_probe_read(data->pkt, buflen, iovbase);
+    bpf_probe_read_kernel(data->pkt, buflen, iovbase);
```
ekyooo: `bpf_probe_read_user` ?

jeromemarchand: Ah indeed, the the iov_base field of strutc iovec has a __user qualifier. Will fix.

Issue #5316, opened 2025-05-26
the bitfield calculation of sk_protocol is wrong due to which tools like 'tcpaccept' are not working


suresh2514:
Testing 'tcpaccept' on a 6.12 kernel, I see its not printing any 'accept' calls.
Debugging further,  we can see the `protocol` bitfield calculation is wrong and
thus it always returns at:

145     if (protocol != IPPROTO_TCP)
146         return 0;

May be the offset calculation needs a review ?.  What I see is:

From: ./tools/tcpaccept.py
+++
125     int gso_max_segs_offset = offsetof(struct sock, sk_gso_max_segs);
126     int sk_lingertime_offset = offsetof(struct sock, sk_lingertime);
127 
128     if (sk_lingertime_offset - gso_max_segs_offset == 4)
129         // 4.10+ with little endian
130 #if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
131         protocol = *(u8 *)((u64)&newsk->sk_gso_max_segs - 3);
132     else
133         // pre-4.10 with little endian
134         protocol = *(u8 *)((u64)&newsk->sk_wmem_queued - 3);
135 #elif __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
136         // 4.10+ with big endian
137         protocol = *(u8 *)((u64)&newsk->sk_gso_max_segs - 1);
138     else
139         // pre-4.10 with big endian
140         protocol = *(u8 *)((u64)&newsk->sk_wmem_queued - 1);
141 #else
142 # error "Fix your compiler's __BYTE_ORDER__?!"
143 #endif
144 
145     if (protocol != IPPROTO_TCP)
146         return 0;
+++

In my 6.12  kernel, the offset is large value:

+++
crash> struct -o sock |grep -e sk_gso_max_segs -e sk_lingertime
  [0x22a] u16 sk_gso_max_segs;
  [0x240] unsigned long sk_lingertime;
+++

If I directly use sk_protocol bitfield, then the script works....

PR #5405, opened 2025-10-07
tools/tcpaccept: Fix tcpaccept on recent kernel

jeromemarchand:
The tcpaccept use the relative offset of gso_max_segs and sk_lingertime_offset to check whether sk_protocol is its own field of part of a bitfield and find it's location. This is not very robust.

Use BPF.kernel_struct_has_field() to find out whether it's its own field and revert to the old workaround if it's part of a bitfield.

Closes: #5316

### DIFF
--- a/tools/tcpaccept.py
+++ b/tools/tcpaccept.py
@@ -117,42 +117,8 @@ int kretprobe__inet_csk_accept(struct pt_regs *ctx)
 
     // check this is TCP
     u16 protocol = 0;
-    // workaround for reading the sk_protocol bitfield:
 
-    // Following comments add by Joe Yin:
-    // Unfortunately,it can not work since Linux 4.10,
-    // because the sk_wmem_queued is not following the bitfield of sk_protocol.
-    // And the following member is sk_gso_max_segs.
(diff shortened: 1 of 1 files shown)


tools/tcpaccept.py
```diff
@@ -216,6 +222,11 @@
     bpf_text = bpf_text.replace('##FILTER_FAMILY##',
         'if (family != AF_INET6) { return 0; }')
 
+if BPF.kernel_struct_has_field("sock", "sk_protocol") == 1:
```
ekyooo, line 230: Is the existing offset == 2 fallback still necessary for cases where the sk_protocol field exists, but the absence of BTF information causes the kernel_struct_has_field() check to fail?

jeromemarchand: Good point. I didn't think of that.

chenhengqi:
Thanks for the fix.

But I think at some point we may retire all python-based tools.
The libbpf-tools are production ready and can be shipped without extra dependencies.
These kind of issues raised by kernel breaking changes can also be fixed by CO-RE easily.

PR #5408, opened 2025-10-12
Fix available_filter_functions path: replace DEBUGFS with TRACEFS

spevnev:


spievniev:
Hello @ekyooo,
Some BCC tools still don't work on newer kernels (where tracing has been moved from debug fs) because one tracing fs access hasn't been updated.
Could you review and merge it so that the tools work again?

PR #5411, opened 2025-10-21
Fix `available_filter_functions` file path?

iago-lito:
I had to tweak this today to get `offcputime` working.. does it make sense?

### DIFF
--- a/src/python/bcc/__init__.py
+++ b/src/python/bcc/__init__.py
@@ -726,7 +726,7 @@ class BPF(object):
                 raise e
             blacklist = set([])
 
-        avail_filter_file = "%s/tracing/available_filter_functions" % DEBUGFS
+        avail_filter_file = "%s/available_filter_functions" % TRACEFS
         try:
             with open(avail_filter_file, "rb") as avail_filter_f:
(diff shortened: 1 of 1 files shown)


iago-lito:
Possibly related: https://github.com/bpftrace/bpftrace/issues/1338

src/python/bcc/__init__.py
```diff
@@ -726,7 +726,7 @@ def get_kprobe_functions(event_re):
                 raise e
             blacklist = set([])
 
-        avail_filter_file = "%s/tracing/available_filter_functions" % DEBUGFS
+        avail_filter_file = "%s/available_filter_functions" % TRACEFS
```
chenhengqi, line 729: Use `os.stat()` to check the existence of tracefs/debugfs instead.

iago-lito: A test with `os.path.exists` is already done on line 50 up there. I figured that test covered the check logic and that this line (729) was a wrong application of it?
In any case, I for sure don't understand the thing at play here. Where is `available_filter_functions` supposed to live? Does it depend on the distro? What are `tracefs/debugfs` and how are they related?

ekyooo:
This has just been merged via #5408.
Thank you.

Issue #5412, opened 2025-10-23
Wrong shared library resolved when multiple SONAMEs exist (add support for versioned SONAMEs in uprobe)


HoyeonRhee:
## Wrong shared library resolved when multiple SONAMEs exist

BCC attach_uprobe resolve shared libraries by name (e.g. `ssl`), and this might resolve to the wrong library if multiple versions of the same SONAME exist in `/etc/ld.so.cache`.
This happens because BCC’s resolver always picks the first match in the cache, which can be a completely different library version or even a different implementation (e.g., `LibreSSL` instead of `OpenSSL`).

This is a general issue affecting any library with multiple SONAME entries (e.g., `libssl`, `libcrypto`, etc.).

### Reproducible Command

```bash
$ sudo docker run -dit --privileged --name tw opensuse/tumbleweed bash 
$ sudo docker exec tw zypper install -y --no-recommends openssl libssl59 bcc-tools kernel-default-devel 
$ sudo docker exec -d tw bash -c ' 
  while true; do
    printf "GET / HTTP/1.1\r\nHost: google.com\r\nConnection: close\r\n\r\n" \
    | openssl s_client -connect google.com:443 -servername google.com -quiet > /dev/null 2>&1
    sleep 1
  done'
$ sudo docker exec -it tw bash
# for i in $(rpm -qa | grep lib | grep ssl); do echo $i ; rpm -ql $i | grep '/libssl'; done
libopenssl3-3.5.3-1.1.x86_64
/usr/lib64/libssl.so.3
/usr/lib64/libssl.so.3.5.3
libopenssl-3-fips-provider-3.5.3-1.1.x86_64
libssl59-4.1.0-1.1.x86_64
/usr/lib64/libssl.so.59
/usr/lib64/libssl.so.59.0.1
# LD_DEBUG=libs openssl version
     573:     find library=libssl.so.3 [0]; searching
     573:      search cache=/etc/ld.so.cache
     573:       trying file=/lib64/libssl.so.3
     573:     calling init: /lib64/libssl.so.3

# ldconfig -p | egrep 'cache|ssl\.'
171 libs found in cache '/etc/ld.so.cache'
        libssl.so.59 (libc6,x86-64) => /lib64/libssl.so.59
        libssl.so.3 (libc6,x86-64) => /lib64/libssl.so.3
# /usr/share/bcc/tools/sslsniff -g -n  
FUNC         TIME(s)            COMM             PID     LEN 
^C
[...]

PR #5413, opened 2025-10-23
cc: support versioned SONAME in shared library resolution

HoyeonRhee:
bcc_procutils_which_so() and related helpers previously handled only short
names like "ssl", constructing "lib%s.so" and picking the first match from
/etc/ld.so.cache, which could select the wrong library (e.g. libssl.so.59
instead of libssl.so.3).

This patch adds support for versioned SONAMEs containing .so.x (e.g.
ssl.so.3.5.3, crypto.so.3), treating them as exact matches rather than
"lib%s.so". It ensures correct resolution on systems with multiple SONAME
variants (e.g. OpenSSL vs LibreSSL) and maintains compatibility with 
library short-name lookups

Downstream tools like bpftrace also benefit from this change, allowing
versioned SONAMEs in attachpoints such as uprobe:libssl.so.3:SSL_write,
ensuring probes attach to the correct library when several versions coexist.

Closes: #5412 
Related: bpftrace/bpftrace#4760

### DIFF
--- a/tests/cc/test_c_api.cc
+++ b/tests/cc/test_c_api.cc
@@ -70,6 +70,30 @@ TEST_CASE("shared object resolution using loaded libraries with the generalized 
   free(libelf);
 }
 
+TEST_CASE("versioned shared object resolution with the generalized function", "[c_api]") {
+  char *libc = bcc_procutils_which_so("c.so.6", 0);
+  REQUIRE(libc);
+  REQUIRE(libc[0] == '/');
+  REQUIRE(string(libc).find("libc.so.6") != string::npos);
+  free(libc);
+}
+
+TEST_CASE("versioned shared object resolution using loaded libraries", "[c_api]") {
+  char *libc = bcc_procutils_which_so_in_process("c.so.6", getpid());
+  REQUIRE(libc);
+  REQUIRE(libc[0] == '/');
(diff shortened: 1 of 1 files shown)


docs/reference_guide.md
```diff
@@ -1723,7 +1723,7 @@ By default `pid` is set to -1, indicating the `uprobe` will be attached to all p
 For libraries, the uprobe will attach to the version of the library used by the process if `pid` was given.
 For how `pid` is used, see examples in [funcinterval](https://github.com/iovisor/bcc/blob/78423e1667db202012bbb032c567589175a2796c/tools/funcinterval.py#L155-L156).
 
-Libraries can be given in the name argument without the lib prefix, or with the full path (/usr/lib/...). Binaries can be given only with the full path (/bin/sh).
+Libraries can be given in the name argument without the lib prefix, optionally with a versioned SONAME (c.so.6), or with the full path (/usr/lib/...). Binaries can be given only with the full path (/bin/sh).
```
jordalgo: nit: 
```suggestion
Libraries can be given in the name argument without the lib prefix, with a versioned SONAME (c.so.6), or with the full path (/usr/lib/...). Binaries can be given only with the full path (/bin/sh).
```

HoyeonRhee: Your suggestion sounds more natural!
I'll update right away! Thank you for the review!

PR #5416, opened 2025-10-28
Delete the element from the `starts` map after it has been used.

mogoreanu:
I think it has been an omission, as opposed to an optimization, in the original change that introduced this. The python version of the `funclatency` has the delete after the lookup.

Before the change:
./funclatency nvme_setup_cmd
Tracing nvme_setup_cmd.  Hit Ctrl-C to exit
^C
     nsec                : count    distribution
       512 -> 1023       : 243      |****************************************|
      1024 -> 2047       : 43       |*******                                 |
      2048 -> 4095       : 59       |*********                               |
      4096 -> 8191       : 26       |****                                    |

After the change:
./funclatency nvme_setup_cmd
Tracing nvme_setup_cmd.  Hit Ctrl-C to exit
^C
     nsec                : count    distribution
       256 -> 511        : 4        |                                        |
       512 -> 1023       : 624      |****************************************|
      1024 -> 2047       : 39       |**                                      |
      2048 -> 4095       : 98       |******                                  |
      4096 -> 8191       : 71       |****                                    |
      8192 -> 16383      : 1        |                                        |

### DIFF
--- a/libbpf-tools/funclatency.bpf.c
+++ b/libbpf-tools/funclatency.bpf.c
@@ -88,6 +88,8 @@ static void exit(void)
 	if (slot >= MAX_SLOTS)
 		slot = MAX_SLOTS - 1;
 	__sync_fetch_and_add(&hist[slot], 1);
+
+	bpf_map_delete_elem(&starts, &pid);
 }
 
 SEC("fexit/dummy_fexit")


libbpf-tools/funclatency.bpf.c
```diff
@@ -72,6 +72,7 @@ static void exit(void)
 	start = bpf_map_lookup_elem(&starts, &pid);
 	if (!start)
 		return;
+	bpf_map_delete_elem(&starts, &pid);
```
chenhengqi: This cleanup should be done at the end of this function.

mogoreanu: Done? PTAL

mogoreanu: Please.

chenhengqi: Please squash the three commits into one. Other than that, looks good to me.

mogoreanu: @chenhengqi Please take another look, took me awhile to squash. Please bear with me.

ekyooo:
LGTM.
The patch correctly prevents memory leaks.

Note: If the traced function is recursive, it looks like race conditions could occur (same TID overwrites `starts` during nested calls), and the delete operation wouldn't be guaranteed in such cases. Within current architecture constraints, this is acceptable. 
For recursive function support, per-CPU maps or stack-based tracking could be considered as future improvements.

Thank you.

PR #5414, opened 2025-10-24
Handle shallow clone build failure with Github Workflow (ensure PEP 440)

HoyeonRhee:
bcc's GitHub workflow has been failing for a while when building from
a shallow clone (git clone --depth 1). In this case, git_describe()
from CMake returns an invalid version string (128-NOTFOUND-deadbeef).

Older setuptools versions used in Fedora 38 tolerated this format, but
newer setuptools in Ubuntu 24.04 enforces PEP 440 compliance and treats
it as an error, breaking the CI build.

This updates cmake/version.cmake to detect this case and fall back to a
compliant `0.0.0+<hash>` version when tags are missing. The commit hash is
kept for traceability, and behavior with full clones remains unchanged.

```bash
/usr/lib/python3/dist-packages/setuptools/command/egg_info.py:131: SetuptoolsDeprecationWarning: Invalid version: '128-NOTFOUND+bb1a55f3'.
!!

        ********************************************************************************
        Version '128-NOTFOUND+bb1a55f3' is not valid according to PEP 440.

        Please make sure to specify a valid version for your package.
        Also note that future releases of setuptools may halt the build process
        if an invalid version is given.

        This deprecation is overdue, please update your project and remove deprecated
        calls to avoid build errors in the future.

        See https://peps.python.org/pep-0440/ for details.
        ********************************************************************************

!!
  return _normalization.best_effort_version(tagged)
Traceback (most recent call last):
  File "/bcc/build/src/python/bcc-python3/setup.py", line 11, in <module>
    setup(name='bcc',
  File "/usr/lib/python3/dist-packages/setuptools/__init__.py", line 107, in setup
    return distutils.core.setup(**attrs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/setuptools/_distutils/core.py", line 185, in setup
    return run_commands(dist)
           ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/setuptools/_distutils/core.py", line 201, in run_commands
    dist.run_commands()
  File "/usr/lib/python3/dist-packages/setuptools/_distutils/dist.py", line 969, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python3/dist-packages/setuptools/dist.py", line 1233, in run_command
    super().run_command(command)
  File "/usr/lib/python3/dist-packages/setuptools/_distutils/dist.py", line 988, in run_command
    cmd_obj.run()
[...]

PR #5428, opened 2025-11-25
Fix BCC CI: shallow-clone versioning, clang/tools_smoke test regressions

HoyeonRhee:
bcc CI has been red from shallow-clone builds emitting PEP 440–invalid
"128-NOTFOUND-…" versions, and from py_test_clang/py_test_tools_smoke
breakage on newer kernels. This series supersedes the prior PR #5414.

By first commit, cmake/version.cmake now falls back to a compliant 
'0.0.0+\<hash\>' when tags are missing, satisfying newer python setuptools
in shallow-clone workflows.

The clang tests are fixed to use blk-mq (rq->q->disk in place of rq_disk)
and the percpu-counter layout for mm->rss_stat[MM_ANONPAGES].count,
and kmalloc kprobe are replaced with kmalloc_noprof for recent kernels.
[1] [2] [3]

Additionally, smoke tests now only runs btrfs/f2fs tools when those 
modules are loaded, and softirqslower.py is fixed with missing exec bit to
avoid error. 

Together these changes clear the py_test_clang and py_test_tools_smoke
failures and restore CI.

1: commit torvalds/linux@f3fa33acca9f ("block: remove the ->rq_disk field in struct request")
2: commit torvalds/linux@f1a7941243c1 ("mm: convert mm's rss stats into percpu_counter") 
3: commit torvalds/linux@7bd230a26648 ("mm/slab: enable slab allocation tagging for kmalloc and friends")


cmake/version.cmake
```diff
@@ -5,14 +5,19 @@ if(NOT REVISION)
...
   git_describe(GIT_TAG_LAST "--abbrev=0" "--tags")
-  git_get_exact_tag(GIT_TAG_EXACT)
-  string(SUBSTRING "${GIT_TAG_LAST}+${GIT_SHA1_SHORT}" 1 -1 REVISION)
-  if(GIT_TAG_EXACT)
-    string(SUBSTRING "${GIT_TAG_EXACT}" 1 -1 REVISION)
-    message(STATUS "Currently on Git tag ${GIT_TAG_EXACT}")
-  else ()
-    message(STATUS "Latest recognized Git tag is ${GIT_TAG_LAST}")
-    set(GIT_TAG_EXACT "")
+  if(GIT_TAG_LAST MATCHES "-128-NOTFOUND")
```
ekyooo: Sorry for the late response.

On second thought, while the fallback approach is effective for CI tests, I think release builds should treat missing git tags as an error to ensure proper version control.

I tried to find a way to clone only commits since the last git tag (using git shallow-exclude, shallow-since, etc.) so we could handle both CI tests and release builds with the same code, but couldn't find a clean solution. Distinguishing between these two cases might be a reasonable approach.

HoyeonRhee: Thank you for looking into this.

Every CMake build without an explicit `-DCMAKE_BUILD_TYPE=` is treated as a Release build:
https://github.com/iovisor/bcc/blob/380ee018bbd04e0300c811bebd88c84da59f715f/CMakeLists.txt#L7-L9

Regarding the suggestion to enforce git tags for Release builds, this would cause tarball builds to fail unless `REVISION` is provided (see `cmake/version.cmake#L3`) or `-DCMAKE_BUILD_TYPE=Debug` is specified. 

Distributions also build from tarballs [1] [2], but they inject `REVISION` in their spec files.

Regular users, however, do not have git metadata in a tarball, so a plain cmake will always fail unless they manually set `REVISION` or use a `Debug` build.

1: https://src.fedoraproject.org/rpms/bcc/blob/rawhide/f/bcc.spec#_130
2: https://build.opensuse.org/projects/Base:System/packages/bcc/files/bcc.spec

HoyeonRhee:
Updated `GIT_TAG_LAST` tag match with  `-NOTFOUND` placeholder to cover other edge cases defined in
[iovisor/bcc/blob/master/cmake/GitRevisionDescription.cmake](https://github.com/iovisor/bcc/blob/HEAD/cmake/GetGitRevisionDescription.cmake).

tests/python/test_clang.py
```diff
@@ -1310,8 +1310,8 @@ def test_jump_table(self):
     if (!rq->start_time_ns)
       return 0;
 
-    if (!rq->rq_disk || rq->rq_disk->major != 5 ||
-        rq->rq_disk->first_minor != 6)
+    struct gendisk *disk = rq->q->disk;
```
ekyooo: Is a NULL check needed for rq->q?

ekyooo: Please follow declaration-first style.

tests/python/test_clang.py
```diff
@@ -1298,7 +1298,7 @@ def test_packed_structure(self):
     def test_jump_table(self):
         text = b"""
 #include <linux/blk_types.h>
-#include <linux/blkdev.h>
+#include <linux/blk-mq.h>
```
ekyooo, line 1307: Does this patch require updating the @skipUnless kernel version requirement at line 1297?

tests/python/test_clang.py
```diff
@@ -1249,7 +1249,7 @@ def test_probe_read_array_accesses8(self):
         text = b"""
 #include <linux/mm_types.h>
 int test(struct pt_regs *ctx, struct mm_struct *mm) {
-    return mm->rss_stat.count[MM_ANONPAGES].counter;
+    return mm->rss_stat[MM_ANONPAGES].count;
```
ekyooo, line 1258: Does the rss_stat test need @skipUnless?

tests/python/test_clang.py
```diff
@@ -423,7 +423,7 @@ def test_probe_simple_assign(self):
 #include <linux/gfp.h>
 struct leaf { size_t size; };
 BPF_HASH(simple_map, u32, struct leaf);
-int kprobe____kmalloc(struct pt_regs *ctx, size_t size) {
+int kprobe____kmalloc_noprof(struct pt_regs *ctx, size_t size) {
```
ekyooo, line 430: Does the kmalloc_noprof test need @skipUnless as well?

cmake/version.cmake
```diff
@@ -5,14 +5,19 @@ if(NOT REVISION)
...
-  string(SUBSTRING "${GIT_TAG_LAST}+${GIT_SHA1_SHORT}" 1 -1 REVISION)
-  if(GIT_TAG_EXACT)
-    string(SUBSTRING "${GIT_TAG_EXACT}" 1 -1 REVISION)
-    message(STATUS "Currently on Git tag ${GIT_TAG_EXACT}")
-  else ()
-    message(STATUS "Latest recognized Git tag is ${GIT_TAG_LAST}")
-    set(GIT_TAG_EXACT "")
+  if(GIT_TAG_LAST MATCHES "-NOTFOUND")
+    set(REVISION "0.0.0+${GIT_SHA1_SHORT}")
+    message(STATUS "Using PEP 440 compliant version for shallow clone")
```
ekyooo: How about making the message more informative? e.g.,
`message(STATUS "No valid tag reachable for shallow clone, using fallback 0.0.0+${GIT_SHA1_SHORT}")`

cmake/version.cmake
ekyooo: Thank you for the information about tarball cases. Agreed.

ekyooo:
> Updated `GIT_TAG_LAST` tag match with `-NOTFOUND` placeholder to cover other edge cases defined in [iovisor/bcc/blob/master/cmake/GitRevisionDescription.cmake](https://github.com/iovisor/bcc/blob/HEAD/cmake/GetGitRevisionDescription.cmake).

It's good.

cmake/version.cmake
HoyeonRhee: Good point. Since a tarball isn’t a shallow clone, I’ll adjust it to:

"No valid Git tag found, using fallback 0.0.0+${GIT_SHA1_SHORT}"

Thanks!

tests/python/test_clang.py
HoyeonRhee: This code is used as a standalone test, so it isn’t attached to any events.
In normal blk-mq paths `blk_mq_start_request`, rq->q should always be initialized, so a NULL here would indicate a kernel issue.

Still, if you think adding a NULL check improves clarity, I’m happy to include it.

HoyeonRhee: Got it. But to follow the `rq->q` NULL check, I’ll drop the separate declaration.

tests/python/test_clang.py
HoyeonRhee: Got it. I’ll update the @skipUnless accordingly.

tests/python/test_clang.py
HoyeonRhee: I’ll adjust as suggested.

tests/python/test_clang.py
HoyeonRhee: Got it. I’ll update the @skipUnless accordingly.

tests/python/test_clang.py
ekyooo: > In normal blk-mq paths blk_mq_start_request, rq->q should always be initialized, so a NULL here would indicate a kernel issue.

Thanks for the clarification. Then, I agree the NULL check is unnecessary.

tests/python/test_tools_smoke.py
```diff
@@ -181,7 +187,10 @@ def test_ext4slower(self):
 
     @skipUnless(kernel_version_ge(4,4), "requires kernel >= 4.4")
     def test_f2fsslower(self):
-        self.run_with_int("f2fsslower.py", allow_early=True)
+        if (self.kmod_loaded("f2fs")):
```
ekyooo, line 190: The current `pass` statement might mislead users into thinking the f2fsslower test executed successfully when it's actually skipped. 

Could you use `@skipUnless` or `self.skipTest()` to explicitly mark it as skipped in the test results?  For example: 
`skipped 'f2fs module not loaded'`

This would make the test status clearer.

HoyeonRhee: Thanks for the review!
Will apply it.

tests/python/test_clang.py
ekyooo: Could you please check this comment?

ekyooo:
>>> In normal blk-mq paths blk_mq_start_request, rq->q should always be initialized, so a NULL here would indicate a kernel issue.

>> Thanks for the clarification. Then, I agree the NULL check is unnecessary.


> Could you please check this comment?

Other than the above comment, LGTM. 
Thank you for your contribution to improving the tests and for your patience. 
If you'd prefer, I can make the changes mentioned in that comment myself, as I realize it may be a bit tedious. Just let me know!

tests/python/test_clang.py
HoyeonRhee: Sorry for the late reply. Changes reverted as requested.

HoyeonRhee:
Sorry for the late reply. Changes reverted as requested.

PR #5442, opened 2026-01-08
libbpf-tools/ksnoop: Fix memory leaks

kknjh:
cmd_info and cmd_trace return directly when the parse_traces function returns a negative value, but parse_traces may also fail at the parse_trace step. Therefore, we add the release logic here to address this scenario.

### DIFF
--- a/libbpf-tools/ksnoop.c
+++ b/libbpf-tools/ksnoop.c
@@ -639,7 +639,7 @@ static int parse_traces(int argc, char **argv, struct trace *traces)
 
 	for (i = 0; i < argc; i++) {
 		if (parse_trace(argv[i], &(traces[i])))
-			return -EINVAL;
+			return -i - 1;
 
 		if (!stack_mode || i == 0)
 			continue;
@@ -668,6 +668,19 @@ static int alloc_traces(int argc, struct trace **traces)
 	return 0;
 }
 
+static void free_traces(struct trace *traces, __u8 num)
+{
+	__u8 i;
+
+	for (i = 0; i < num; ++i) {
+		btf_dump__free(traces[i].dump);
+		if (traces[i].btf != vmlinux_btf)
+			btf__free(traces[i].btf);
+	}
+	btf__free(vmlinux_btf);
+	free(traces);
+}
+
 static int cmd_info(int argc, char **argv)
 {
 	struct trace *traces = NULL;
@@ -681,8 +694,8 @@ static int cmd_info(int argc, char **argv)
 
 	nr_traces = parse_traces(argc, argv, traces);
 	if (nr_traces < 0) {
-		free(traces);
-		return nr_traces;
+		free_traces(traces, -nr_traces);
+		return -EINVAL;
 	}
 
 	for (i = 0; i < nr_traces; i++) {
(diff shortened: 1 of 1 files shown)


Bojun-Seo:
Great catch!

Applying this patch as-is might be a good approach, but the issue seems to stem from how the code is structured, making such problems more likely to occur. Instead of directly applying this patch, what if we restructure the function's role? That way, the overall structure would be better, and it would be easier to maintain.

I suggest modifying the `parse_traces` function so that it only retains its core functionality of parsing traces, while the memory allocation for this function is handled by the user—specifically, by the `cmd_info` and `cmd_trace` functions. By doing this, the functions responsible for allocating memory would also handle its deallocation, reducing the chance of developer errors and making maintenance easier.

What do you think?

kknjh:
> Great catch!
> 
> Applying this patch as-is might be a good approach, but the issue seems to stem from how the code is structured, making such problems more likely to occur. Instead of directly applying this patch, what if we restructure the function's role? That way, the overall structure would be better, and it would be easier to maintain.
> 
> I suggest modifying the `parse_traces` function so that it only retains its core functionality of parsing traces, while the memory allocation for this function is handled by the user—specifically, by the `cmd_info` and `cmd_trace` functions. By doing this, the functions responsible for allocating memory would also handle its deallocation, reducing the chance of developer errors and making maintenance easier.
> 
> What do you think?

Modified, thank you.

Bojun-Seo:
LGTM!

libbpf-tools/ksnoop.c
```diff
@@ -649,28 +664,26 @@ static int parse_traces(int argc, char **argv, struct trace **traces)
...
 
-	nr_traces = parse_traces(argc, argv, &traces);
-	if (nr_traces < 0)
+	err = alloc_traces(argc, &traces);
+	if (err < 0)
+		return err;
+
+	nr_traces = parse_traces(argc, argv, traces);
+	if (nr_traces < 0) {
+		free(traces);
```
ekyooo: Dynamically allocated fields within each trace structure are not being properly cleaned up?
In parse_trace() at line 557:
`trace->dump = btf_dump__new(trace->btf, trace_printf, NULL, NULL);`

Bojun-Seo: Another great catch!
While it is still a memory leak like the original one, I think it would be better to handle this fix in a separate patch or PR.
Since it is happening at a separate location in the code.

libbpf-tools/ksnoop.c
```diff
@@ -668,6 +668,13 @@ static int alloc_traces(int argc, struct trace **traces)
 	return 0;
 }
 
+static void free_traces(struct trace *traces)
+{
+	btf_dump__free(traces->dump);
+	btf__free(traces->btf);
```
Bojun-Seo: Great catch! This is another good one.

The `traces->btf` is allocated via `btf__load_vmlinux_btf()`, which returns a `struct btf *` that the caller is responsible for freeing with `btf__free()`.

Reference
https://docs.ebpf.io/ebpf-library/libbpf/userspace/btf__load_vmlinux_btf/

Bojun-Seo (commented):
Finally, this patch series fixes three memory leaks and improves the overall resource management structure of `ksnoop`.

PR #5443, opened 2026-01-08
libbpf-tools/offcputime: Fix the unit conversion error

kknjh:
The units of min_block_time and max_block_time are microseconds, yet they are directly passed to min_block_ns and max_block_ns (which expect nanoseconds).

### DIFF
--- a/libbpf-tools/offcputime.c
+++ b/libbpf-tools/offcputime.c
@@ -363,8 +363,8 @@ int main(int argc, char **argv)
 	obj->rodata->user_threads_only = env.user_threads_only;
 	obj->rodata->kernel_threads_only = env.kernel_threads_only;
 	obj->rodata->state = env.state;
-	obj->rodata->min_block_ns = env.min_block_time;
-	obj->rodata->max_block_ns = env.max_block_time;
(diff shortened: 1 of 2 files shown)


libbpf-tools/offcputime.c
```diff
@@ -363,8 +363,8 @@ int main(int argc, char **argv)
 	obj->rodata->user_threads_only = env.user_threads_only;
 	obj->rodata->kernel_threads_only = env.kernel_threads_only;
 	obj->rodata->state = env.state;
-	obj->rodata->min_block_ns = env.min_block_time;
-	obj->rodata->max_block_ns = env.max_block_time;
+	obj->rodata->min_block_ns = env.min_block_time * 1000;
+	obj->rodata->max_block_ns = env.max_block_time * 1000;
```
ekyooo: It has a potential overflow risk when multiplying by 1000.

Is nanosecond precision really necessary here? How about using microseconds directly in the BPF code as well? This would eliminate the overflow risk and avoid unnecessary unit conversion.

Please refer to offcputime.py for reference.

kknjh: Modified, thank you.

PR #5463, opened 2026-02-09
libbpf-tools: add EXTRA_BPFCFLAGS build flag

aaronjheng:


Bojun-Seo:
Please add the purpose and the use cases for this PR on the commit message.

aaronjheng:
I was trying to package libbpf-tools in nixpkgs. The include path should be set for linux headers. As a matter of fact, I ultimately achieved it by modifying `C_INCLUDE_PATH`, so the current PR is no longer necessary.

Thanks for your response. Closing.

PR #5491, opened 2026-03-08
folly/tracing: Add ARM and AArch64 support to static tracepoints

### LINKED ISSUE #5354: USDT Probe Test Failures on ARM64 Architecture
Hi all,

While running BCC test cases on an ARM64 Yocto build, I encountered several failures related to USDT probes with the following error message:
"Unable to find USDT libbcc_test:sample_lib_probe_1 from binary PID 600 for probe on_event."

The failures seem to be due to the sample probes not being generated on ARM64. The tests that failed include:

cc test finding a probe in our own process

cc test find a probe in our own binary with C++ API

cc test find probes in our own binary with C++ API

cc test find a probe in our process with C++ API

cc test find a probe in our process' shared libs with C++ API

cc test uprobe refcnt semaphore activation

cc test usdt partial init with fail init_usdt

python py_test_usdt

python py_test_usdt2

python py_test_usdt3

When I inspected the binaries with readelf, I found no NOP instructions, no probe points, and no .note.stapsdt section.

Looking at [this link](https://github.com/linux-usdt/libstapsdt/blob/main/src/sdtnote.c#L9), it appears that libstapsdt only supports x86 architectures. Could this be the reason why the tests are failing on ARM64?

hsadinen:
This resolves USDT probe test failures on ARM64 platforms.
Without these changes, the .note.stapsdt section containing probe
information was missing entirely on ARM64, causing test failures when
attempting to find and attach to USDT probes in the BCC test suite.

Fixes: #5354


ekyooo:
Could you add an appropriate prefix to the commit message? (ex. test/python: xxx)

hsadinen:
@ekyooo, I have updated commit message with appropriate prefix.

PR #5497, opened 2026-03-24
libbpf-tools: Always escape tab and newline chars in execsnoop

ReillyBrogan:
5340c21bc94c199cae55e891e6dbff972a6f0df6 made it so that the Python execsnoop always escaped newline characters which is required to avoid breaking tools that parse the execsnoop stdout when command args contain newlines. This applies the same change to the libbpf-tools version as well as additionally escaping tab characters.

Specifically I noticed this difference in behavior when porting system76-scheduler to libbpf execsnoop. It would probably be better to have a json output instead but this at least gets system76-scheduler working without additional changes.

### DIFF
--- a/libbpf-tools/execsnoop.c
+++ b/libbpf-tools/execsnoop.c
@@ -171,11 +171,15 @@ static void time_since_start()
 	printf("%-8.3f", time_diff);
 }
 
-static void inline quoted_symbol(char c) {
+static void inline escape_symbol(char c, bool quote) {
 	switch(c) {
 		case '"':
-			putchar('\\');
-			putchar('"');
+			if (quote) {
+				putchar('\\');
+				putchar('"');
+			} else {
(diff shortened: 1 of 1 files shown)


ekyooo:
- I confirmed that the AI review feedback is valid.
- \r can cause a similar issue to \n. Should it also be included in the escape targets?
- Escaping \t and \n was already applied in quote mode, but not in non-quote mode. As I understand it, this PR extends that behavior to non-quote mode and also includes some refactoring. Could you update the commit message to reflect that more accurately?

ekyooo (commented):
Do you still plan to address the review comments on this PR? If not, would you be okay with me taking over?

ReillyBrogan:
> Do you still plan to address the review comments on this PR? If not, would you be okay with me taking over?

Go for it!

PR #5508, opened 2026-05-05
libbpf-tools/mountsnoop: Fix a compiler error in probe_exit()

namhyung:
## Description

This PR is to fix the following build error
```
  libbpf-tools/mountsnoop.bpf.c:79:10: error: enumeration value 'OP_MIN' not handled in switch [-Werror,-Wswitch]
     79 |         switch (argp->op) {
        |                 ^~~~~~~~
  1 error generated.
```

## Why this approach

It'd be simpler to add `default` label rather than handle missing cases.

---

## Checklist

- [X] Commit prefix matches changed area (e.g., `tools/toolname:`, `libbpf-tools/toolname:`, `src/cc:`, `docs:`, `build:`, `tests/python:`)
- [X] Commit body explains **why** this change is needed

**For new tools only**
- [ ] README.md entry added

---

### DIFF
--- a/libbpf-tools/mountsnoop.bpf.c
+++ b/libbpf-tools/mountsnoop.bpf.c
@@ -130,6 +130,8 @@ static int probe_exit(void *ctx, int ret)
 					sizeof(eventp->move_mount.to_pathname),
 					argp->sys.move_mount.to_pathname);
 		break;
+	default:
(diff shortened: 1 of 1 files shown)


ekyooo (approved):
Thank you.

PR #5530, opened 2026-06-23
cc: Synchronize so flags for sw_64 architecture from glibc

zhangzikang01:
Synchronize the shared library flag bits for sw_64 from glibc to resolve errors in some bcc tools caused by their inability to recognize the libc library.

glibc reference path: sysdeps/generic/ldconfig.h

Signed-off-by: zhangzikang01  zhangzikang@kylinos.cn


ekyooo:
1. Please specify what in glibc you are referencing in sysdeps/generic/ldconfig.h, since it is not clear from the current description.
2. Please fix the actual commit message (not just PR description) to follow the required prefix and include the exact rationale.
3. Please rebase this branch onto latest master.

PR #5538, opened 2026-08-10
bcc_zip: Convert zip header fields from little endian to host order

sumanthkorikkar:
The following bpftrace zip test fails on s390:
bpftrace -e 'uprobe:"./testprogs/archive.zip!/true":main {} begin { exit() }'

This test sets a uprobe on the main function of the true program that is stored in the archive.

The zip file format specs mentions that all multibyte fields (uint16_t and uint32_t) are stored in little endian format. However, the original code in bcc_zip.c was reading these values directly without converting them to the host systems byte order [1] (see sections 4.3.3 and 4.4.1.1)

[1] https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT

Use le16toh() and le32toh() when reading unaligned zip header fields in bcc_zip.c to ensure correct parsing on big endian systems.

Reviewed-by: Ilya Leoshkevich <iii@linux.ibm.com>
Signed-off-by: Sumanth Korikkar <sumanthk@linux.ibm.com>

### DIFF
--- a/src/cc/bcc_zip.c
+++ b/src/cc/bcc_zip.c
@@ -16,6 +16,7 @@
 
 #include "bcc_zip.h"
 
+#include <endian.h>
 #include <fcntl.h>
 #include <limits.h>
 #include <stdint.h>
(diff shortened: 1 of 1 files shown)


ekyooo (approved):
Thank you.

PR #5546, opened 2026-08-21
tools/killsnoop: migrate to tracepoints

### LINKED ISSUE #3592: update killsnoop to use tracepoints

codebyangelo:
Fixes #3592

This commit updates the python killsnoop tool to use tracepoints for the kill, tkill, and tgkill syscalls. Previously, it used kprobes on sys_kill, which broke on Linux 5.11 due to kernel changes. This brings the python tool in parity with the libbpf-tools C version (sigsnoop), providing a more robust and stable tracing mechanism.

 ## Description
    Fixes #3592.

    This PR migrates the Python `tools/killsnoop.py` script from using `kprobes` to using
  `tracepoints`.

    Previously, the tool relied on kprobing `sys_kill`. Starting with Linux 5.11,
  architectural changes to syscall wrappers caused this kprobe approach to fail. This PR
  removes the fragile kprobe logic and replaces it with robust `TRACEPOINT_PROBE` macros for
  `syscalls:sys_enter_kill`, `tkill`, and `tgkill` (and their respective `sys_exit`
  counterparts).

## Why this approach
    Tracepoints provide a stable, kernel-maintained ABI for tracing syscalls that is immune
  to the underlying function renaming and `pt_regs` mapping changes that break kprobes.

    Additionally, the `libbpf-tools` C equivalent (`sigsnoop`) had already successfully
  migrated to this exact tracepoint approach, but the original Python tool was left behind.
  This PR brings the Python implementation back into parity and restores its functionality on
  all modern kernels.

---

## Checklist

**For new tools only**
- [ ] README.md entry added

---

### DIFF
--- a/tools/killsnoop.py
+++ b/tools/killsnoop.py
@@ -143,8 +143,6 @@ TRACEPOINT_PROBE(syscalls, sys_exit_tgkill)
 {
     return probe_exit(args, args->ret);
 }
-;
-
 
 """
 
--- a/tools/killsnoop.py
+++ b/tools/killsnoop.py
@@ -67,7 +67,8 @@ struct data_t {
 BPF_HASH(infotmp, u32, struct val_t);
 BPF_PERF_OUTPUT(events);
 
-int syscall__kill(struct pt_regs *ctx, int tpid, int sig)
+
+static int probe_entry(u32 tpid, int sig)
 {
     u64 pid_tgid = bpf_get_current_pid_tgid();
     u32 pid = pid_tgid >> 32;
@@ -85,9 +86,24 @@ int syscall__kill(struct pt_regs *ctx, int tpid, int sig)
     }
 
     return 0;
-};
+}
 
-int do_ret_sys_kill(struct pt_regs *ctx)
+TRACEPOINT_PROBE(syscalls, sys_enter_kill)
+{
+    return probe_entry(args->pid, args->sig);
+}
+
+TRACEPOINT_PROBE(syscalls, sys_enter_tkill)
+{
+    return probe_entry(args->pid, args->sig);
+}
+
+TRACEPOINT_PROBE(syscalls, sys_enter_tgkill)
+{
+    return probe_entry(args->pid, args->sig);
+}
+
+static int probe_exit(void *ctx, int ret)
 {
     struct data_t data = {};
     struct val_t *valp;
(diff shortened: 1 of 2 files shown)


tools/killsnoop.py
```diff
@@ -85,9 +86,24 @@
     }
 
     return 0;
-};
+}
+
+TRACEPOINT_PROBE(syscalls, sys_enter_kill)
```
ekyooo, line 91: TRACEPOINT_PROBE-based BPF programs require Linux 4.7 or later. Please update the test_killsnoop() kernel gate from 4.4 to 4.7, consistent with the existing tracepoint tests.

for example,
```
--- a/tests/python/test_tools_smoke.py
+++ b/tests/python/test_tools_smoke.py
@@
-    @skipUnless(kernel_version_ge(4,4), "requires kernel >= 4.4")
+    @skipUnless(kernel_version_ge(4,7), "requires kernel >= 4.7")
```

tools/killsnoop.py
```diff
@@ -104,14 +120,32 @@
...
+TRACEPOINT_PROBE(syscalls, sys_exit_tkill)
+{
+    return probe_exit(args, args->ret);
+}
+
+TRACEPOINT_PROBE(syscalls, sys_exit_tgkill)
+{
+    return probe_exit(args, args->ret);
+}
+;
```
ekyooo: Please remove the extra blank lines and semicolon.

ekyooo (commented):
Thank you.

tools/killsnoop.py
codebyangelo: Thank you for the guidance. Done.

ekyooo (approved):

PR #5553, opened 2026-09-07
src/python: fix RIP-relative USDT argument formatting

cuishuang:
## Description

`USDTProbeArgument._format()` compares the ctypes `c_char_p` `base_register_name` value with the Python string `"ip"`.

On Python 3, `c_char_p` values are returned as `bytes`, so the comparison is always false for RIP-relative USDT arguments. The RIP-relative formatting path therefore reports:

    unrecognized argument format

instead of formatting the argument as:

    *(&symbol + offset)

This change compares the value with `b"ip"` and adds a regression test covering the RIP-relative symbol format.

## Why this approach

The change keeps the existing ctypes byte representation and matches the encoding handling already used elsewhere in `usdt.py`. It is a minimal Python-only fix with no C++ API or ABI changes.

---

## Checklist

**For new tools only**
- [ ] README.md entry added

---


ekyooo:
Thanks

ekyooo:
The newly added test is failing in CI. Could you please investigate?

cuishuang:
> The newly added test is failing in CI. Could you please investigate?

@ekyooo  Modified. Please review it again.

PR #5555, opened 2026-09-10
src/cc: silence clang warning caused by typeof(x) expressions

hhoffstaette:
## Description
```
In file included from /virtual/main.c:4:
In file included from include/net/tcp.h:34:
In file included from include/net/xfrm.h:26:
include/net/ip6_fib.h:256:10: warning: default initialization of an object of type 'typeof (f6i->expires)' (aka 'const unsigned long')
      leaves the object uninitialized [-Wdefault-const-init-var-unsafe]
  256 |                 return time_after(jiffies, f6i->expires);
      |                        ^
include/linux/jiffies.h:75:3: note: expanded from macro 'time_after'
   75 |          typecheck(unsigned long, b) && \
      |          ^
include/linux/typecheck.h:8:12: note: expanded from macro 'typecheck'
    8 |         typeof(x) __dummy2; \
      |                   ^
1 warning generated.
```
Add -Wno-default-const-init-var-unsafe to silence this.

## Why this approach

We cannot fix kernel headers or the semantics of `typeof()` expressions.

---

## Checklist

**For new tools only**
- [ ] README.md entry added


hhoffstaette:
The Ubuntu test failure has nothing to do with the patch, it's a generic python explosion related to nss (if I read the error log correctly).

ekyooo:
Thanks

PR #5557, opened 2026-09-12
libbpf-tools/execsnoop: also trace execveat

### LINKED ISSUE #5556: libbpf-tools/execsnoop: processes started via execveat(2) are never reported
I noticed that `libbpf-tools/execsnoop` only hooks `sys_enter_execve` and `sys_exit_execve`. So if a process is started with `execveat(2)` (that's what `fexecve(3)` uses under the hood), it never shows up. No error, no warning, the line is just not there.

To reproduce: build execsnoop from master, run `sudo ./execsnoop`, and in another terminal run a tiny program that does `syscall(SYS_execveat, AT_FDCWD, "/bin/true", argv, envp, 0)`. You'll see the test program itself get exec'd, but not the exec it does through execveat.

This is what I got on Ubuntu 24.04, kernel `6.17.0-1022-azure`, x86_64 (same on aarch64):

```
true             3845   3825     0 /bin/true EXECVE_MARKER
execveat_test    3846   3825     0 /tmp/execveat_test
```

The second line is the test program starting normally. The `/bin/true EXECVEAT_MARKER` exec it then performs via execveat is missing.

I have this running in CI if you want to see the full thing, including the test program source: https://github.com/ojuschugh1/bcc/actions/runs/34631964095

I have a fix ready and will open a PR in a bit.

ojuschugh1:
## Description

Fixes #5556

execsnoop only hooks the execve tracepoints, so anything started with `execveat(2)` (which is what `fexecve(3)` uses) never shows up. No error, the line just isn't there.

The fix moves the enter and exit handler bodies into shared inline functions and adds programs for `sys_enter_execveat` and `sys_exit_execveat`. The only real difference on the enter side is where the arguments live: execveat has the filename in `args[1]` and argv in `args[2]`. Nothing changes in userspace since the skeleton attaches every program on its own.

I tested this on Ubuntu 24.04 (kernel `6.17.0-1022-azure`) on both x86_64 and aarch64, building execsnoop from master and then from this branch and running the same little test program against each. On master the execveat call is missed. With this patch it shows up, and normal execve events look the same as before:

```
true             3987   3963     0 /bin/true EXECVE_MARKER
execveat_test    3988   3963     0 /tmp/execveat_test
true             3988   3963     0 /bin/true EXECVEAT_MARKER
```

Full before/after run, including the test program source: https://github.com/ojuschugh1/bcc/actions/runs/34631964095

I also checked that the refactor didn't change the existing programs: `sys_exit_execve` compiles to byte-identical bytecode before and after, and `sys_enter_execve` has the same instruction count with only register allocation moving around.

One known gap: if execveat is called with `AT_EMPTY_PATH` and a file descriptor, the args column shows an empty pathname. PCOMM still tells you what ran. Turning the fd back into a path needs a fair bit more code, so I'd rather leave that for a separate change if people want it.

The Python `tools/execsnoop.py` has the same blind spot. Happy to do that one too if there's interest.

## Why this approach

Sharing the handler bodies keeps the two syscalls in step, so any future fix to argument parsing lands on both, and the diff stays small enough to check against the original.

## Checklist

- [x] Commit prefix matches changed area
- [x] Commit body explains why this change is needed


ekyooo (commented):
Could you add a commit message body similar to the PR description? It would help other developers understand the change when reading the Git history.

ojuschugh1:
Done, added the commit body. Thanks for taking a look.

ojuschugh1:
hey @ekyooo did you get chance to review this PR.

PR #5558, opened 2026-09-14
libbpf-tools/execsnoop: trace execveat() syscalls

tychoyang:
## Description

`execsnoop` currently only attaches to `sys_enter_execve`/`sys_exit_execve`, so any process
that launches a new program via `execveat(2)` instead of `execve(2)` is completely invisible
to it. `execveat()` is a separate syscall (own syscall number, own tracepoints) — it's not a
variant of `execve()` that gets caught by the same probes.

Real callers that only go through `execveat()`:
- glibc's `fexecve(fd, argv, envp)` (kernel >= 3.19) translates directly to
  `execveat(fd, "", argv, envp, AT_EMPTY_PATH)`.
- Container runtimes (runc/containerd) commonly exec via an already-opened file descriptor
  instead of a path, to avoid TOCTOU races between checking and executing a file.

This PR adds `sys_enter_execveat`/`sys_exit_execveat` tracepoint programs so these calls are
captured too, using the exact same event format, filters (`-u`/`-x`/`--max-args`/`-c`) and
output as the existing `execve` path.

I verified this by writing a small test program that:
- calls `fexecve()` on an fd opened from `/bin/true` (success path), and
- calls `execveat()` with an invalid fd (failure path, `EBADF`).

Before this patch, `execsnoop` produced **zero** output for either call. After this patch,
both are reported correctly (the success case shows `COMM` as `"3"`, since the kernel
synthesizes `/dev/fd/<fd>` as the filename when an empty path + `AT_EMPTY_PATH` is used, and
takes its basename for `comm`; the failure case shows `RET=-9` i.e. `-EBADF`).

## Why this approach

- **Mirror the existing `execve` handling instead of introducing shared helpers.** The new
  `tracepoint__syscalls__sys_enter_execveat` / `sys_exit_execveat` functions are structurally
  identical to their `execve` counterparts, reusing the same `execs`/`events`/`cgroup_map`
  maps, `empty_event`, `valid_uid()`, and all existing filtering knobs (`filter_cg`,
  `targ_uid`, `ignore_failed`, `max_args`). This keeps the diff additive-only and avoids
  touching any existing, already-reviewed code path.
- **Only the `ctx->args[]` offsets differ.** Per the tracepoint format
  (`sys_enter_execveat`: `fd, filename, argv, envp, flags`), `filename`/`argv` sit one slot
  further right than in `sys_enter_execve` (`filename, argv, envp`), so the offsets are shifted
  by 1 (`args[0]`→`args[1]` for filename read, `args[1]`→`args[2]` for argv). Everything else
  (argv capture loop, `LAST_ARG` bound check, `#pragma unroll`, comm/ppid/uid capture) is
  copied verbatim.
- **Considered but rejected: a kprobe/kretprobe on `do_execveat_common()`.** Function-level
  probes would require per-kernel-version signature handling and are far more fragile than the
  stable `syscalls:sys_{enter,exit}_execveat` tracepoint ABI, which hasn't changed since it was
  introduced in 3.19.
- **Considered but rejected: refactoring `execve`/`execveat` into one shared static function.**
  This would have touched the existing `sys_enter_execve`/`sys_exit_execve` bodies, which is
  unnecessary risk for a change whose sole goal is closing a coverage gap; keeping the two
  pairs independent makes the diff trivially reviewable (pure addition, no behavior change to
  existing code).

---

## Checklist

**For new tools only**
- [ ] README.md entry added

<sub>(Not a new tool — this PR only extends the existing `execsnoop` tool's coverage, so the "For new tools only" items don't apply. Existing man page / README / smoke test for `execsnoop` remain accurate as-is since the CLI, output format and options are unchanged.)</sub>


ekyooo:
This duplicates the preceding PR #5557. 
Thanks.

